PR #2, opened 2016-02-13
fix android build instructions (use gradle instead of ant)

Hinsbart:
Btw, it's totally awesome that we can easily contribute to the documentation now.
Good job :)


akien-mga:
Thanks! :) Using PRs instead of simply editing wiki pages online will likely have some overhead cost for newcomers, but for regular contributors it will indeed be much simpler IMO.

PR #5, opened 2016-02-17
DampedSpringJoint2D: explain stiffness and damping

JoshuaGrams:


akien-mga:
Thanks! The class reference should not be edited directly here though, [as mentioned in the README](https://github.com/godotengine/godot-docs#editing-existing-pages) (I'll see how to make it clearer). You need to add those changes directly to `doc/base/classes.xml` in the main Godot source repo: https://github.com/godotengine/godot/tree/master/doc/base

See http://docs.godotengine.org/en/latest/contributing/reference_filling_work.html

akien-mga:
Otherwise it's a great addition, looking forward to merging it in the main godot repo :)

JoshuaGrams:
Ah, bother. Sorry about that. If they're autogenerated, can you automatically add a big warning comment at the top of each file? That's the usual thing to do, and the only thing that I've ever found to work at all reliably...

akien-mga:
That's a good idea, I'll add this to the parser that generates the files, thanks :)

Issue #16, opened 2016-02-23
Out of date tutorials (primarily screenshots)


Fisto69:
when attempting to get into Godot via 

http://docs.godotengine.org/en/latest/tutorials/step_by_step/scenes_and_nodes.html  

i quickly noticed the screenshots are outdated, this causes a couple problems.  first being that if the screenshots are outdated, my confidence in the fact the text itself is up-to-date and will allow me to learn the software without a hitch is not very high.  

and two, arguably the most important for me is the fact that the lack of up-to-date screenshots can cause a problem for a good albiet small percentage of users, i myself, like others that exist use visual references to help learn.. and out-of-date screenshots will make that a bit more difficult. not impossible, but just a bit harder

StraToN:
I agree. I already noticed this in the proofreading available here : https://public.etherpad-mozilla.org/p/godot-docs-proofreading
... which is not finished yet, by the way.
But yes, definitely, we will have to pdate screenshots everywhere in the docs.

Fisto69:
oh sweet, any ETA on when the tutorials should be updated? id wager 2 weeks to maybe a month or so? 

this is assuming the tutorials/website are taking low priority over everything else with the recent big release.

akien-mga:
Check the link above what you just posted?

akien-mga:
Ah maybe you meant for the whole documentation. Can't give an ETA, better use the documentation as is, and see it improve little by little. It's a never-ending work :)

Fisto69:
ah, yes i meant as a whole. i was unsure if you had an actual team dedicated to keeping the tutorials and just all documentation up-to-date or not.

however, im going to close this since it this particular question has been answered and solved.

much appreciated fella's.

akien-mga:
Well let's keep it open as a reminder, the way you formulated the title does make it more general, so let's try to fix all screenshots :)

Fisto69:
oh okay, that is a good point.  my bad.

  fix all the things! ;P.

akien-mga:
I just merged the changes for the Scenes and nodes tutorial, the website should be updated in 5 min or so.

Fisto69:
oh sweet, thats great news.  thanks a bunch.

NathanLovato:
We've got to redo a lot of stuff for Godot 3. Closing the issue for now as it's old and idle - if you have specific screenshots that need update, feel free to open a new issue and list them

PR #17, opened 2016-02-23
Update images in tutorial "Scenes and nodes" to 2.0

flurick:
Solves #16


StraToN:
Thanks a lot !
@akien-mga Should we remove old screenshots?

akien-mga:
Thanks a lot, those screenshots are great.

Could you merge all commits into one to clean up the git history a bit?

To do so, you can run in your master branch:
```git rebase -i HEAD~7```

You should get an editor with a list of the last 7 commits (yours).
Replace "pick" by "fixup" for the last 6 commits, and save, that should merge them all into the first commit.

Then to be sure check the ``git log``, and if it looks good (one commit), you can force push your changes to the remote:
```git push -f```

Alternative you could also ```git rm``` the old images as @StraToN suggested (we don't support Godot < 1.0 anymore, so no need to keep those around). But I can also do it after merging this PR if need be :)

akien-mga:
If the above seems like black magic to you, I can also do it myself, I know git can be daunting ;)

akien-mga:
I've merged in your changes manually in this single commit: 7bb4816
(+ I removed the old screenshots)

PR #19, opened 2016-02-24
Better image for playscene button

techtonik:
See https://github.com/godotengine/godot/issues/3822


StraToN:
Thank you !
Is the new image used somewhere already?

techtonik:
@StraToN the old image which is renamed to topnav is unused. The new image is used here http://docs.godotengine.org/en/latest/tutorials/step_by_step/scenes_and_nodes.html#editor

PR #25, opened 2016-02-27
Some fixes to "Scenes and Nodes"

her001:
Adjust the text describing location of scene tree and inspector to match new defaults, spelling and grammar fixes.


her001:
Should we completely avoid contractions? I think so, but I only corrected this with the fewest number of changes.

Perhaps it could be reworded from "let's change it to" to something like "we will change it to."

PR #27, opened 2016-02-27
Update scripting_continued.rst

IllusionCatalyst:
Corrected some typos


akien-mga:
Thanks! Note that as far as the commit log is concerned, GitHub's default suggestion for "Update $filename" is not really pertinent, better to set a meaningful title yourself such "Fix some typos in scripting_continued tutorial".

PR #32, opened 2016-02-28
Add links and other fixes to  "Simple 2D Game"

her001:


her001:
Okay, I'll add those in...

PR #37, opened 2016-02-29
Fix using tiles tutorial 

genete:


akien-mga:
Thanks! Please use more descriptive commit titles next time, GitHub's editor's suggestion is meaningless :) Basically what you used as commit description would have been fine directly as commit title.

PR #38, opened 2016-02-29
Update using_tilemaps.rst

genete:
Make 'Introduction' a chapter of 'Using tilemaps' and not a main documentation entry


akien-mga:
Nice catch.

PR #42, opened 2016-03-01
Update documentation for class_meshinstace.rst

MinnowMike:
MeshInstance contains the set(String property, var value) property for setting morph shapes from GDScript, but it isn't documented. I stumbled over a year old reference to it in the doc's of the old site. I've added reference to this function in the member function list and the member function description of this document.


akien-mga:
MeshInstance likely inherit the ``set`` method from the [Object class](http://docs.godotengine.org/en/latest/classes/class_object.html#class-object), like all Nodes in Godot.

Furthermore, the documentation in classes/ should not be updated here, but directly in the source XML file in Godot's source repo, as mentioned:
- in the README: https://github.com/godotengine/godot-docs#editing-existing-pages
- and in the header of every file: https://raw.githubusercontent.com/MinnowMike/godot-docs/548ad9bf6b1a68a608c4109c325ba5b22a230469/classes/class_meshinstance.rst

PR #48, opened 2016-03-02
Description of limit functions

MeiHouwang:


akien-mga:
Thanks, but the documentation in classes/ should not be updated here, but directly in the source XML file in Godot's source repo, as mentioned:
- in the README: https://github.com/godotengine/godot-docs#editing-existing-pages
- and in the header of every file, e.g.: https://raw.githubusercontent.com/godotengine/godot-docs/master/classes/class_camera2d.rst

PR #60, opened 2016-03-04
Update requirements to include JDK

vsrinu26f:


akien-mga:
Does it work with OpenJDK 8? I wouldn't want to specify the old JDK 6 explicitly if newer versions work too, users tend to take it to the letter and try to install JDK 6 even if it's not available anymore on their recent distro.

vsrinu26f:
Yes it is working with OpenJDK 8

akien-mga:
Well then it would be better to say "OpenJDK 6 or later" :)

akien-mga:
I've cherry-picked your commits, squashed them and improve them in c6b2e38.

PR #65, opened 2016-03-06
A place for general performance or optimization tips

vsrinu26f:


akien-mga:
Please discuss such things in an issue first, adding table of contents does not really bring much :)
We need to discuss what should be the overall structure *if* we want to add such content in the docs.

PR #86, opened 2016-03-12
Fix typo and extra word in "Custom GUI controls"

her001:


StraToN:
I'd have kept the "left" word you removed as it relies to the mouse button. One can not click a button with another mouse button than left button, AFAIK.

her001:
@StraToN Control._input_event() can take any event. The default controls might only take left clicks, but this is for custom controls, which should be able to take anything as long as the mouse is over it. I could be wrong...

PR #95, opened 2016-03-17
I removed "porn" from High Dynamic Range tutorial

unfa:
I guess it's really not needed there and  just might be offending for some people.


akien-mga:
It doesn't add much indeed, and we should stay SFW, there are kids using Godot ;)

PR #98, opened 2016-03-17
Fix formatting

BlaXpirit:
These things don't render as expected

### DIFF
--- a/reference/gdscript.rst
+++ b/reference/gdscript.rst
@@ -493,7 +493,7 @@ in the class' node parent, accessible with ``get_parent``:
     # two parents up (etc)
     get_parent().get_parent().function(args)
 
-To call a function in a *base class* (i.e. one ``extend``ed in your current class), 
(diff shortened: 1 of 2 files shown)


reference/gdscript.rst
```diff
@@ -493,7 +493,7 @@ in the class' node parent, accessible with ``get_parent``:
     # two parents up (etc)
     get_parent().get_parent().function(args)
 
-To call a function in a *base class* (i.e. one ``extend``ed in your current class), 
+To call a function in a *base class* (i.e. one ``extend``-ed in your current class),
```
akien-mga: I'm not sure it will work, I fear it works only with spaces on both sides :/ But we'll see.

BlaXpirit: I checked that this fixes it at least in GitHub.

akien-mga: Ah, good to know :)

PR #103, opened 2016-03-21
Remove of duplicated entry

jazn:
Remove of duplicated entry for 
elif (what==NOTIFICATION_THEME_CHANGED):


akien-mga:
Thanks. Love your avatar :p

PR #115, opened 2016-03-29
Fix abbreviation in "SSL certificates"

her001:


akien-mga:
I'd remove the comma too while at it :P
Not important though hehe.

PR #117, opened 2016-03-30
Remove unneeded indent in "HTTP client class"

her001:


tutorials/http_client_class.rst
```diff
@@ -29,7 +29,7 @@ It will connect and fetch a website.
 
 
         while( http.get_status()==HTTPClient.STATUS_CONNECTING or http.get_status()==HTTPClient.STATUS_RESOLVING):
-                    # Wait until resolved and connected
+        # Wait until resolved and connected
```
akien-mga: You removed too many spaces ;) It's a while block, so it should be:
```
while condition:
    # comment
    do_stuff()
    # other comment
```

akien-mga: Though actually, now that I read the comment, I think it should go before the while block as it explains the condition, wdyt?

PR #124, opened 2016-04-17
FreeBSD specific build instructions

erikarn:
hi!

I just built the latest godot master on -HEAD (rather than installing from freebsd-ports, which is godot 1.1) and I thought it would be good to explicitly state the build instructions for FreeBSD.

This adds exactly that!

### DIFF
--- a/reference/_compiling.rst
+++ b/reference/_compiling.rst
@@ -8,6 +8,7 @@ Compiling
    introduction_to_the_buildsystem
    compiling_for_windows
    compiling_for_linux
+   compiling_for_freebsd
    compiling_for_osx
    compiling_for_android
    compiling_for_ios
--- a/reference/compiling_for_freebsd.rst
+++ b/reference/compiling_for_freebsd.rst
@@ -1,0 +1,98 @@
+.. _doc_compiling_for_freebsd:
+
+Compiling for FreeBSD
+=====================
+
+.. highlight:: shell
+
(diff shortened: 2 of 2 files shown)


StraToN:
After reflection with @akien-mga we agreed that this PR should find its way into the currently named "Compiling for linux" page. Please read below before your hair raises on your skin :)

HOWEVER......
While we all agree to say that *BSD != GNU/Linux, we generally tend to use (in the source code actually) the appellation "X11" for all UNIX-based systems (GNU/Linux, *BSD, etc). 
The documentation currently doesn't reflect this. This is bad, and we need to change its name.

So, we are going to change the current name of this page from ```compiling_for_linux.rst``` to ```compiling_for_x11.rst``` and change its name accordingly. Thus, we will not need the creation of a new page.

I invite you to cherry-pick your PR and insert it in the existing page as soon as it is renamed. Thanks a lot !

erikarn:
np! thanks!


-a

akien-mga:
Could we merge the FreeBSD-specific bits in https://github.com/godotengine/godot-docs/blob/master/reference/compiling_for_x11.rst ? (And maybe add OpenBSD instructions too if they differ, or note them as FreeBSD/OpenBSD if the instructions are the same)

StraToN:
Yeah, let's do it. This PR has been here for too long, let's include it.

PR #131, opened 2016-04-24
Grammar and link fixes for "InputEvent" and "Simple 2D Game"

jdknox:
``simple_2d_game.rst`` had a broken link, so I updated it along with ``inputevent.rst``.

I also updated the sample code to be more uniform with respect to spaces around operators, and modified the variables/class names within the text to display as inline code (e.g., ``pad_size`` in ``simple_2d_game.rst``).  Are these good practices?

One additional question: is it easier to merge lots of file changes at the same time, or is limiting the pull request to a few files easier?

### DIFF
--- a/tutorials/step_by_step/simple_2d_game.rst
+++ b/tutorials/step_by_step/simple_2d_game.rst
@@ -8,7 +8,7 @@ Pong
 
 In this simple tutorial, a basic game of Pong will be created. There are
 plenty of more complex examples in the demos included with the engine,
-but this should get introduced to basic functionality for 2D Games.
+but this should get one introduced to basic functionality for 2D Games.
 
 Assets
 ~~~~~~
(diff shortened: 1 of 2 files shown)


tutorials/step_by_step/simple_2d_game.rst
```diff
@@ -55,9 +55,10 @@ is defined, then input methods that trigger it are added.
 Open the project properties dialog again, but this time move to the
 "Input Map" tab.
 
+.. role:: python(code)
```
akien-mga: The changes look good, but why define a role here? Note that all code snippets are defined as python code by default unless specified otherwise.

jdknox: Oops, sorry about that.  I was testing different roles, but then realized the default one was better.  But I forgot to remove that line before committing.

PR #150, opened 2016-05-20
Updated images in Pong tutorial

ArcFutahito:
Fixed outdated images in Godot Pong tutorial


akien-mga:
Thanks! In the end I redid them myself, because if people get confused about a missing checkbox, I can't imagine how they would feel with a screenshot from 2.1-dev without buttons in the scene panel :p
See 66d2836fc9a482521a41eb9bd53bf5bc818fe10c

PR #152, opened 2016-05-24
Revamp Windows compilation tutorial

Griefchief:
The old tutorial was incomplete and out of date.
Adds a lot of new info on how Visual Studio works now.

### DIFF
--- a/reference/compiling_for_windows.rst
+++ b/reference/compiling_for_windows.rst
@@ -10,9 +10,10 @@ Requirements
 
 For compiling under Windows, the following is required:
 
--  `Visual C++ <http://www.microsoft.com/visualstudio>`__, Visual C++
-   Express compiler or Visual Studio Community (recommended) at least
-   the 2013 version (12.0) up to 2015 (14.0). **Make sure you get a
+-  Visual C++, `Visual
+   Studio Community <https://www.visualstudio.com/en-us/products/visual-studio-community-vs.aspx>`__
+   (recommended), at least the 2013 version (12.0) up to 2015 (14.0).
+   **If you're using Express, make sure you get/have a
    version that can compile for C++, Desktop**.
 -  `Python 2.7+ <https://www.python.org/downloads/>`__ (3.0 is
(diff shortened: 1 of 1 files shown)


vnen:
Thanks, I've never got the time to work on this and you did a better job than I would.

I just don't really agree with this: 

> However Visual Studio 2015 and some others seems to not install any of
 +the above[...]

VS2015 install those prompts just fine, even the community edition. IME is VS2013 that gives more trouble in this matter. You might want to say: "Visual Studio sometimes do not install any or some of the above[...]"

Griefchief:
@vnen My version of VS2015 CE installed only the dev console... VS2013 for seme reason seemed to have installed all prompts except the x86 Native...

That said, you are correct the sentence was a bit to strong, so i've ammended the commit to be a bit more neutral (first and last paragraph):

```
However Visual Studio sometimes seems to not install some of the above
shortcuts, except the Developer Console at these locations that are
automatically searched by the start menu search option:


   Win 7:
   C:\ProgramData\Microsoft\Windows\Start Menu\Programs\Visual Studio 2015\Visual Studio Tools
   C:\ProgramData\Microsoft\Windows\Start Menu\Programs\Visual Studio 2013\Visual Studio Tools
   
If you found the Developer Console, it will do for now to create a 32
bit version of Godot, but if you want the 64 bit version, you might need
to setup the prompts manually for easy access.

If you don't see some of the shortcuts , "How the prompts actually work"
section bellow will explain how to setup these prompts if you need them.
```

Thank you for reviewing the tut :)

Griefchief:
By the way, how should corrections like this be handled in a PR? I amended the commit so that there is only one commit in the PR, but is this the general way of doing it for PR-s, unless there is a specific need for 2+ commits?

Or do you squash commits yourselves when you pull into the repo (if you want history of the changes)?

akien-mga:
@Griefchief I like it best with only one commit when there is no specific need for more indeed. For the docs I'm not too regarding, but on the main source repo I'm more demanding ;) So now it's perfect with your self squashed commit.

PR #155, opened 2016-05-26
add how to configure Xcode as your IDE

IndicaInkwell:


StraToN:
Why does the PR contain a link to imgur album? The images should be part of the PR, so we don't rely on external image sharing service.

IndicaInkwell:
I'll remove that and add the images to the repo. Also, reformat the text. Looks like rst doesn't have indented lists?

akien-mga:
> Also, reformat the text. Looks like rst doesn't have indented lists?

It does have it, but some empty newlines are needed for a proper separation: http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#bullet-lists (like for any indented block in rst as I understand it).

IndicaInkwell:
Ah! I wish I had read that. Ended up formatting it like the QtCreator entry, which I think looks better, so it worked out.

PR #163, opened 2016-05-31
Update vector_math.rst

MerKki:
There was an error in the "Perpendicular Vectors" example.

The example for rotating right would have rotated left and the example for rotating left would have rotated right.
Additionally the variable name for rotating left was also v_right.
I also changed the x and y values to fit the image in the example so there is an direct correlation.

### DIFF
--- a/tutorials/vector_math.rst
+++ b/tutorials/vector_math.rst
@@ -165,11 +165,13 @@ Example:
 
 ::
 
-    var v = Vector2(0,1)
+    var v = Vector2(2,4)
+    
     # rotate right (clockwise)
-    var v_right = Vector2(-v.y, v.x)
-    # rotate left (counter-clockwise)
     var v_right = Vector2(v.y, -v.x)
+    
+    # rotate left (counter-clockwise)
+    var v_left = Vector2(-v.y, v.x)
 
 This is a handy trick that is often of use. It is impossible to do with
 3D vectors, because there are an infinite amount of perpendicular


bojidar-bg:
To be fair, the example is correct, but the image has the wrong coordinate system. Y should go down, and X should go right.

akien-mga:
We should decide whether this tutorial is meant to teach generic vector maths with a typical cartesian system with (X+) to the right and (Y+) to the top, or if we should stick to the convention of Godot/the gaming industry with downward Y+. If we use the latter, we should fix the image (and maybe other ones in this tutorial, I haven't checked).

StraToN:
OpenGL uses an origin point located at bottom-left corner, but Godot reverts this by using top-left corner. So I'd personally agree to fix images so they correspond to Godot representation system.

OvermindDL1:
@akien-mga: In all my OGL work I've never seen any engine use Y+ as down, where is this from and what sense does it make?

StraToN:
@OvermindDL1 I think the reason is that GUI development sometimes uses such representation. Since Godot uses both, it is possible that this choice was made.
However, it is not possible to change this now, as this would break all game produced with Godot since this representation was first used.

ObaniGemini:
I think this PR should be closed. Another would be opened then to fix images. Am I wrong ?

StraToN:
I agree. Lets's wait for the new renderer way to do in order to make deep changes. Until that, we should simply propose new illustrations.

Issue #174, opened 2016-06-11
Update dependency list for X11


vnen:
Since godotengine/godot#4905 there's a new dependency for `xrandr`, so it should be added the documentation. It's called `libxrandr-dev` in Ubuntu and `libxrandr` in Arch, but I don't know about the other distros.

vnen:
Ping @Hinsbart :wink:

akien-mga:
`"pkconfig(xrandr)"` should work on any pkgconfig-aware distro (Mageia for sure, Fedora probably).

akien-mga:
Fixed by #175.

PR #175, opened 2016-06-12
Add xrandr to dependency list for x11 compilation

Hinsbart:
TODO: list package name for Mageia

### DIFF
--- a/reference/compiling_for_x11.rst
+++ b/reference/compiling_for_x11.rst
@@ -15,7 +15,7 @@ required:
 -  Python 2.7+ (3.0 is untested as of now)
 -  SCons build system
 -  pkg-config (used to detect the dependencies below)
--  X11, Xcursor and Xinerama development libraries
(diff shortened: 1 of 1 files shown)


vnen:
Almost closes #174 😄.

I don't know Mageia but for the logic it might be `pkgconfig(xrandr)`.

Hinsbart:
Might be, but I'm not sure either :D

neikeq:
It's `libxrandr-dev` on Ubuntu, is it not?

vnen:
@neikeq yes, it is.

Hinsbart:
Right

akien-mga:
Yeah `pkgconfig(xrandr)` would work for Mageia (like on any proper RPM-based distro: https://www.rpmfind.net/linux/rpm2html/search.php?query=pkgconfig%28xrandr%29)

akien-mga:
Merged and added for Mageia too: 282b1b7

PR #177, opened 2016-06-15
Updating invalid call

zackurben:
Updating `get_viewport_rect()` to `get_viewport().get_rect().size`, to fix example errors.

### DIFF
--- a/tutorials/engine/mouse_and_input_coordinates.rst
+++ b/tutorials/engine/mouse_and_input_coordinates.rst
@@ -37,7 +37,7 @@ for example:
 
        # Print the size of the viewport
 
-       print("Viewport Resolution is: ",get_viewport_rect().size)
+       print("Viewport Resolution is: ", get_viewport().get_rect().size)
 
     func _ready():
         set_process_input(true)


akien-mga:
`get_viewport_rect()` should work fine if you're extending `CanvasItem`: https://github.com/godotengine/godot/blob/4866713bc3ab58cd22ef5090a75b8a20e3b35e7d/scene/2d/canvas_item.cpp#L1097

But if `get_viewport().get_rect()` works for all nodes in the tree, it might be better yes.

akien-mga:
Actually the two calls are not equivalent: `get_viewport_rect()` calls `get_viewport().get_visible_rect()`, and the definition of the visible_rect is not the same as the plain rect: https://github.com/godotengine/godot/blob/b75fb6674b844d177d6d07526dd4d05de74294f5/scene/main/viewport.cpp#L728-L752

zackurben:
@akien-mga Ah I see, I was not extending `CanvasItem` so I couldn't use `get_viewport_rect()`. Based on the runtime errors, I wrongfully assumed it was an old example :)

Thanks for the clarification.

PR #188, opened 2016-06-27
Unity to godot page

### LINKED ISSUE #166: Guide for switching from Unity to Godot
Since there is a huge inflow of new/potential users right now, it would be nice to have a guide for transitioning them from Unity's way of thinking to the Godot way of thinking.

Adapted from IRC (meeting on -devel, 2-nd of June):
**[reduz]** as you and many probably noticed, the change of license in Unity generated a lot of people angry and an unexpected large amount of new Godot users and discussions of Godot as an alternative in many places
**[reduz]** I´ve seen two things used as argument from people that would like to switch, one is of course support for C#, which I guess we´ll have to eventually implement. The second is documentation or tutorials on how to move workflow from Unity to Godot. Some sort of ¨map¨ of how something you do in Unity is done in Godot
**[reduz]** do you think we could assemble some sort of mini team to work on this kind of documentation? I know Unreal has a large article about this for Unity users, but anything that could ease their transition, specially for advanced users, while at the same time being able to show how we do many of the same things in Godot better would be fantastic

StraToN:
Please review and/or discuss missing parts. :)
Resolves #166


reference/unity_to_godot.rst
```diff
@@ -0,0 +1,177 @@
...
+
+As described before, Godot scene system allows splitting scenes in smaller scenes. Since each scene and subscene is actually one scene file in the project, we recommend organizing your project a bit differently. This wiki provides a page for this: `Project Organization <engine/project_organization.html>`_.
+
+
+Where are my prefabs?
+---------------------
+
+The concept of prefabs as provided by Unity is a 'template' element of the scene. It is reusable, and each instance of the prefab that exists in the scene has an existence of its own, but all of them have the same properties as defined by the prefab.
+
+Godot does not provide prefabs as such, but this functionality is here again filled thanks to its scene system: as we saw the scene system is organized as a tree. Godot allows you to save a subtree of a scene as its own scene, thus saved in its own file. This new scene can then be instanced any times you want. Any change you make to this new, separate scene will be applied to the instance. However, any change you make to the instance will not have any impact on the 'template' scene.
```
djrm: any times you want -> as many times as you want
to the instance -> to its instances

reference/unity_to_godot.rst
```diff
@@ -0,0 +1,177 @@
...
+| OS (editor)       | Windows, OSX, Linux (unofficial)                                                  | Windows, X11 (Linux, *BSD), Haiku, OSX                                                                         |
++-------------------+-----------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------+
+| OS (export)       | | Desktop: Windows, Linux/SteamOS, OSX                                            | | Desktop: Windows, X11, OSX                                                                                   |
+|                   | | Mobile: Android, iOS, Windows Phone, Tizen,                                     | | Mobile: Android, iOS, Blackberry (deprecated)                                                                |
+|                   | | Web: WebGL                                                                      | | Web: HTML5 (via emscripten, broken)                                                                          |
```
reduz: Both Unity and Godot export to WebGL

reference/unity_to_godot.rst
```diff
@@ -0,0 +1,177 @@
...
+
+This guide provides an overview of Godot Engine from the viewpoint of a Unity user, and aims to help you migrate your existing Unity experience into the world of Godot.
+
+Differences
+-----------
+
++-------------------+-----------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------+
+|                   | Unity                                                                             | Godot                                                                                                          |
++===================+===================================================================================+================================================================================================================+
+| License           | proprietary                                                                       | MIT License                                                                                                    |
```
reduz: I wold explain license extents a bit more, like:

Proprietary, closed, free license with revenue caps and usage restrictions
vs
MIT License, free and fully open souce without any restriction

reference/unity_to_godot.rst
```diff
@@ -0,0 +1,177 @@
...
++-------------------+-----------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------+
+| OS (editor)       | Windows, OSX, Linux (unofficial)                                                  | Windows, X11 (Linux, *BSD), Haiku, OSX                                                                         |
```
reduz: unofficial and unsupported

reference/unity_to_godot.rst
```diff
@@ -0,0 +1,177 @@
...
+|                   | | Consoles: PS4, PS Vita, XBox One, XBox 360, WiiU, 3DS                           |                                                                                                                |
+|                   | | VR: Occulus Rift, SteamVR, Google Cardboard, Playstation VR, Gear VR, HoloLens  |                                                                                                                |
+|                   | | TV: AndroidTV, Samsung SMARTTV, tvOS                                            |                                                                                                                |
++-------------------+-----------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------+
+| Scene system      | Entity-Component System (ECS) (GameObject > Component).                           | Scene tree and nodes, allowing scenes to be nested and/or inherit other scenes                                 |
```
reduz: I would add that Unity has prefabs too

reference/unity_to_godot.rst
```diff
@@ -0,0 +1,177 @@
...
++-------------------+-----------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------+
+| Third-party tools | Visual Studio or SharpEditor                                                      | | Android SDK for Android export                                                                               |
+|                   |                                                                                   | | External editors are possible                                                                                |
++-------------------+-----------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------+
+| Killer features   | | Huge community                                                                  | | Live-editing                                                                                                 |
```
reduz: Not sure if I would add those as killer features, as people will think Unity has this too.
I would write something like.

Scene System, Animation Pipeline, Easy to write Shaders, Integrate debugging that also runs on device.

reference/unity_to_godot.rst
```diff
@@ -0,0 +1,177 @@
...
+
+Design
+^^^^^^
+
+As you may know already, Unity supports 2 scripting languages for its API: C# and Javascript (called UnityScript). Both languages can be used in the same project (but not in the same file, of course). Choosing one instead of the other is a matter of personal taste, as performances seem not to be affected that much by the use of Javascript as long as the project remains small. C# benefits from its integration with Visual Studio and other specific features, such as static typing.
+
+Godot provides its own scripting language: GDScript. This language borrows its syntax to Python, but is not related to it. If you wonder about why GDScript and not Lua, C# or any other, please read `GDScript <reference/gdscript>`_ and `FAQ <reference/faq>`_ pages. GDScript is strongly attached to Godot API, but it is really easy to learn: between 1 evening for an experimented programmer and 1 week for a complete beginner.
+
+Unity allows you to attach as many scripts as you want to a GameObject. Each script adds a behaviour to the GameObject: for example, you can attach a script to a  so that the player can control it, and another that controls its specific game logic.
```
djrm: you can attach a script to a ???  so that the player can control

reference/unity_to_godot.rst
```diff
@@ -0,0 +1,177 @@
...
+
+As you can see, there are some similarities between both editors: first of course, the scene which allows you to display and organize your game elements on screen. Second, the inspector which allows you to edit the parameters of the selected object. Also, the Hierarchy panel in Unity finds its correspondance with Godot's Scene tree panel, although we will see in the next section the differences between how Unity and Godot manage scenes and scene elements.
+
+Unity and Godot both directly use the folder architecture you define for your project. There is no real good solution, and every folder architecture will work for both engines. You can thus continue with your current way to organize your project. However, we have some advices to share a bit further, in order to match your project architecture and Godot scene system.
+
+To finish, the toolbar at the top of the screen is also similar between Unity and Godot. However, Godot provides some new buttons aside the classic "Play", "Pause" and "Stop" buttons. These buttons are introduced in the `Toolbar <introduction/editor#toolbar>`_ section as they a bit out of the scope of this page.
+
```
reduz: I would use a simpler and more to the point language, and trying to remark what Godot advantages are, like:

While both may seem similar, there are many differences below the surface. Both let you organize the project using the filesystem, but Godot approach is simpler, with a single configuration file, minimalist text format, and no metadata. All this contributes to Godot being much friendlier to VCS systems such as Git, Subversion or Mercurial.

The scene panel is similar to the hierarchy but, as each node has a specific function, the approach used by Godot is more visually descriptive. In other words, It's easier to guess how a scene works at a glance.

The inspector in Godot is more minimalist and designed to only show properties. Thanks to this, objects can export a much larger amount of useful parameters to the user, without having to hide functionality in language APIs. As a plus, Godot allows animating any of those properties visually, so changing colors, textures, enumerations or even links to resources in real-time is possible without involving code.

Finally, the toolbar at the top of the screen is similar in the sense that it allows controlling the project playback, but projects in Godot run in a separate window, as they don't execute inside the editor (but the tree and objects can still be explored in the debugger window). This approach has the disadvantage that the running game can't be explored from different ways (though this may be supported in the future, and displaying collision gizmos in the running game is already possible), but in exchange it has several advantages:

- Running the project and closing it is very fast (Unity has to save, run the project, close the project and then reload the previous state).
- Live editing is a lot more useful, because changes done to the editor take effect immediately in the game, and are not lost (nor have to be synced) when the game is closed. This allows fantastic workflows, like creating levels while you play them. You can even program while the game is running and scripts are reloaded automatically.
-The editor is more stable, because the game runs in a separate process

Finally, the top toolbar includes a menu for remote debugging. These options make it very easy to deploy to a device (connected phone, tablet or browser via HTML5), and debug / live edit the running game after it was exported.

reference/unity_to_godot.rst
```diff
@@ -0,0 +1,177 @@
...
+The scene system
+----------------
+
+This is the most important difference between Unity and Godot, and actually the favourite feature of most Godot users.
+
+Unity's scene system generally consist in embedding all the required assets in a scene, and link them together by setting components and scripts to them.
```
reduz: I would try to remove extra words such as "generally" as much as possible, and text short and concise

reference/unity_to_godot.rst
```diff
@@ -0,0 +1,177 @@
...
+To finish, the toolbar at the top of the screen is also similar between Unity and Godot. However, Godot provides some new buttons aside the classic "Play", "Pause" and "Stop" buttons. These buttons are introduced in the `Toolbar <introduction/editor#toolbar>`_ section as they a bit out of the scope of this page.
+
+The scene system
+----------------
+
+This is the most important difference between Unity and Godot, and actually the favourite feature of most Godot users.
+
+Unity's scene system generally consist in embedding all the required assets in a scene, and link them together by setting components and scripts to them. 
+
+Godot's scene system is different: it actually consists in a tree made of nodes. Each node serves a purpose: Sprite, Mesh, Light... Basically, this is similar to Unity scene system. However, each node can have multiple children, which make each a subscene of the main scene. This means you can compose a whole scene with different scenes, stored in different files.
```
reduz: I would simplifly it like:

A scene in Godot is simple, it consists of a tree made of nodes. A node could be seen as a GameObject with one component and one behavior and, as an extra limitation, scenes can only have one root node.

While this restriction can be seen as arbitrary at first, it makes the whole system a lot more flexible and powerful for the following reasons:

- One function per node makes scenes more descriptive at a glance. It's very easy in Godot for someone who didn't create a scene to understand how it works by just taking a quick look at it. 
- It allows for a divide and conquer approach to create game content like no other tool. Scenes can be edited separately and then instanced and nested at any level. 
- Even inheriting a scene is possible. Make a base enemy scene with common functions (targeting, explosion, hitpoint display, etc), and inherit it to create every kind of enemy. Import entire scenes from your favorite 3D DCC, then inherit them to do local modifications while preserving the original for re-import.
-Teamwork is easier than with any other game engine, as it's very easy to divide a project in scenes. Artist, programmer, game designer and level designer will work most of the time in their own scenes without stepping on each other's toes.

StraToN:
Whew, thanks @reduz and @djrm for these remarks. 
To both of you, I suggest to directly commit on the branch when you see some mistakes. Regarding your more general remarks (about the style and also points to insist), we can keep it here ;)

I'm on it. Thanks again!

reference/unity_to_godot.rst
```diff
@@ -0,0 +1,187 @@
...
+|                   | | Consoles: PS4, PS Vita, XBox One, XBox 360, WiiU, 3DS                           |                                                                                                                |
+|                   | | VR: Occulus Rift, SteamVR, Google Cardboard, Playstation VR, Gear VR, HoloLens  |                                                                                                                |
+|                   | | TV: AndroidTV, Samsung SMARTTV, tvOS                                            |                                                                                                                |
++-------------------+-----------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------+
+| Scene system      | | Entity-Component System (ECS) (GameObject > Component).                         | Scene tree and nodes, allowing scenes to be nested and/or inherit other scenes                                 |
```
OvermindDL1: Just as a note, Unity is not ECS in any way shape or form, it is entirely Component/Scene based.  ECS is where you have data strongly separated from code, most generally in a Dataflow pattern (since ECS is a superset of Dataflow after all).  Unity has absolutely no Dataflow style in it whatsoever (which is one of its major speed hits).  Unity's design is entirely ComponentScene.

StraToN:
ping.

PR #196, opened 2016-07-01
Info on how to compile ios fat binary 32bit + 64bit

kubecz3k:


kubecz3k:
@akien-mga I'm not sure if this is default filename, took this from the build script you send me when I encouraged this problem for the first time...  it wont hurt to manually indicate binary name in this case I guess (user will be able to see similarities to 64 build)

kubecz3k:
Especially since lipo is using this file, so the user looking at lipo line will know better whats happening

akien-mga:
Yeah it's the default filename. Fact is, I'm not sure the argument makes sense at all, I don't think our SConscript has support for specifying a specific binary name this way, so it's likely ignored.

kubecz3k:
Oh I see.. so those names are made by default depending on architecture and release target?

kubecz3k:
Interestingly when I put different name than those, scons is telling 'dont know how to Make File target 'different file name path'

akien-mga:
Yes those names are composed by default by the SConscript, typically `godot.$platform.$target.$bits[.$extension]`

PR #197, opened 2016-07-01
Added missing Rect2 built in type

djrm:


bojidar-bg:
Looks good, except you probably have to put a comma after it, so it is like it was before :smile:

djrm:
@bojidar-bg it doesnt matter, the previous comma (a typo by the way) is irrelevant in python.

PR #201, opened 2016-07-05
Update String.capitalize() documentation

pkowal1982:
Original doc was describing to_upper()


akien-mga:
Thanks, but as mentioned in the README.md and the header of all files in `classes/`, the class reference should not be updated here as it's generated automatically, but it should be updated directly in https://github.com/godotengine/godot/blob/master/doc/base/classes.xml

Sorry about the inconvenience. I won't merge this PR here as it will be overwritten next time the class reference is sync with the source code, but it would be great if you could add those enhancements to the XML file linked above.

PR #207, opened 2016-07-22
Fix typo in ScrollContainer documentation

29jm:


29jm:
Now that I think about it, shouldn't the "Edit on Github" button in the docs lead to doc/base/classes.xml instead of this file? Or maybe changes made here are applied to classes.xml too?

akien-mga:
Yeah the "Edit on GitHub" button needs to be fix for the class reference, I have just no clue how to do this :)

PR #212, opened 2016-07-26
New logo for the docs

djrm:


reduz:
oooh...

bojidar-bg:
Much better :smile:

akien-mga:
@djrm How about using a small version of the Godot icon with those glasses for the "Documentation" icon in the script editor? :)

djrm:
That would be nice, will do it later

PR #214, opened 2016-07-29
Update - Compiling for Android

KelinciFX:
- Updated to 2.1 development
- Easier to understand (I think I'm a little bit biased)

:-)

### DIFF
--- a/reference/compiling_for_android.rst
+++ b/reference/compiling_for_android.rst
@@ -16,21 +16,29 @@ Also, you still need to do all the steps mentioned in the
 :ref:`doc_exporting_for_android` tutorial before attempting your custom
 export template.
 
-Requirements
-------------
+A. Requirements
+---------------
 
 For compiling under Windows, Linux or OSX, the following is required:
 
--  Python 2.7+ (3.0 is untested as of now).
--  SCons build system.
+-  Python 2.7+ [3.0 is untested as of now]
+-  pywin32 [Optional, Windows only. Applied when -j(number of cores) specified]
+-  SCons build system
 -  Android SDK version 19 [Note: Please install all Tools and Extras of sdk manager]
 -  Android build tools version 19.1
--  Android NDK
--  Gradle
--  OpenJDK 6 or later (or Oracle JDK 6 or later)
+-  Android NDK r10e or later
+-  Gradle [will be downloaded and installed when compiling for android if it doesn't exist]
+-  JDK 6+ (Oracle's releases) or OpenJDK 6+
 
-Setting up SCons
-----------------
+B. Setting up Build System
+--------------------------------------------------------
+
+1) Scons : please, refer to their respective docs
+
+2) Gradle : will be automatically setup upon packing APK if 
+we do not have gradle installed previously (may require working internet connection)
+
+3) Android :
 
 Set the environment variable ANDROID_HOME to point to the Android
 SDK.
@@ -38,17 +46,20 @@ SDK.
 Set the environment variable ANDROID_NDK_ROOT to point to the
 Android NDK.
 
-To set those environment variables on Windows, press Windows+R, type
-"control system", then click on **Advanced system settings** in the left
-pane, then click on **Environment variables** on the window that
-appears.
+- on Windows :
+press Windows+R, type "control system", then click on **Advanced system settings**
+in the left pane, then click on **Environment variables** on the window that appears.
 
-To set those environment variables on Linux, use
+- on Unix (eg. Linux and Mac OS):
+Open up Terminal and then execute commands below :
 ``export ANDROID_HOME=/path/to/android-sdk`` and
-``export ANDROID_NDK_ROOT=/path/to/android-ndk``.
+``export ANDROID_NDK_ROOT=/path/to/android-ndk``
 
-Compiling
----------
+C. Compiling
+------------
+
+C1. Compiling the Engine
+------------------------
 
 Go to the root dir of the engine source code and type:
 
@@ -56,53 +67,30 @@ Go to the root dir of the engine source code and type:
 
     C:\godot> scons platform=android
 
-This should result in a regular .so in ``\bin`` folder as if it was
+This should result in a regular .so file (built for ARM platform) in ``\bin`` folder, as if it was
 compiled with flags: ``tools=no target=debug``. The resulting file will
-be huge because it will contain all debug symbols, so for next builds,
-using ``target=release_debug`` or ``target=release`` is recommended.
-
-Copy the .so to the ``libs/armeabi`` Android folder (or symlink if you are
-in Linux or OSX). Note: Git does not support empty directories so you
-will have to create it if it does not exist:
+be huge because it will contain all debug symbols.
 
 ::
 
-    C:\godot> mkdir platform/android/java/libs
-    C:\godot> mkdir platform/android/java/libs/armeabi
-
-Then copy:
-
-::
-
-    C:\godot> copy bin/libgodot.android.<version>.so platform/android/java/libs/armeabi/libgodot_android.so
-
-Or alternatively, if you are under a Unix system you can symlink:
-
-::
-
-    user@host:~/godot$ ln -s bin/libgodot.android.<version>.so platform/android/java/libs/armeabi/libgodot_android.so
-
-Remember that only *one* of libgodot_android.so must exist for each
-platform, for each build type (release, debug, etc), it must be
-replaced.
-
-**Note**: The file inside ``libs/armeabi`` must be renamed to
-**"libgodot_android.so"**, or else unsatisfied link error will happen
-at runtime.
+    C:\godot> bin\libgodot.android.debug.armv7.neon.so
 
 If you also want to include support for x86 Android, add the following
-compile flag: ``android_arch=x86``, then copy/symlink the resulting binary to
-the ``x86`` folder:
+compile flag: ``android_arch=x86``.
 
 ::
 
-    C:\godot> copy bin/libgodot.android.<version>.x86.so platform/android/java/libs/x86/libgodot_android.so
+    C:\godot> scons platform=android android_arch=x86
+
+::
+
+    C:\godot> bin\libgodot.android.debug.x86.neon.so
 
 This will create a fat binary that works in both platforms, but will add
-about 6 megabytes to the APK.
+about more than 6 megabytes to the APK.
 
-Toolchain
----------
+Note on Toolchain
+-----------------
 
 We usually try to keep the Godot Android build code up to date, but
 Google changes their toolchain versions very often, so if compilation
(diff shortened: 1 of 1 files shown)


reference/compiling_for_android.rst
```diff
@@ -215,14 +198,14 @@ failure should be presented there.
...
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 If the application runs but exits immediately, there might be one of the
 following reasons:
 
--  libgodot_android.so is not in ``libs/armeabi``
+-  libgodot_android.so is not in ``lib/armeabi-v7a``
 -  Device does not support armv7 (try compiling yourself for armv6)
--  Device is Intel, and apk is compiled for ARM.
+-  Device is ARM, and apk is compiled for intel.
```
akien-mga: That's indeed biased. Most devices are ARM, and the default build options are for ARM, so in the troubleshooting section, one should mention Intel as it was.

reference/compiling_for_android.rst
```diff
@@ -215,14 +198,14 @@ failure should be presented there.
...
 
-Application exits immediately
+2.) Application exits immediately
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 If the application runs but exits immediately, there might be one of the
 following reasons:
 
--  libgodot_android.so is not in ``libs/armeabi``
+-  libgodot_android.so is not in ``lib/armeabi-v7a``
```
akien-mga: It's `libs/`..

reference/compiling_for_android.rst
```diff
@@ -16,93 +16,81 @@ Also, you still need to do all the steps mentioned in the
 :ref:`doc_exporting_for_android` tutorial before attempting your custom
 export template.
 
-Requirements
-------------
+A. Requirements
```
akien-mga: I'm not for adding those, the table of contents does this job already. And it's inconsistent with the rest of the documentation, and impossible to maintain.

reference/compiling_for_android.rst
```diff
@@ -16,93 +16,81 @@ Also, you still need to do all the steps mentioned in the
...
+-  pywin32 [Optional, Windows only. Applied when -j(number of cores) specified]
+-  SCons build system
 -  Android SDK version 19 [Note: Please install all Tools and Extras of sdk manager]
 -  Android build tools version 19.1
--  Android NDK
--  Gradle
--  OpenJDK 6 or later (or Oracle JDK 6 or later)
+-  Android NDK r10e or later
+-  Gradle [will be downloaded and installed when compiling for android if it doesn't exist]
+-  JDK 6+ (Oracle's releases) or OpenJDK 6+
```
akien-mga: Why? I think it better to put forward the open source implementation. This change is gratuitous.

reference/compiling_for_android.rst
```diff
@@ -16,93 +16,81 @@ Also, you still need to do all the steps mentioned in the
...
+
+B. Setting up Build System
+--------------------------------------------------------
+
+1) Scons : please, refer to their respective docs
```
akien-mga: Why?

reference/compiling_for_android.rst
```diff
@@ -16,93 +16,81 @@ Also, you still need to do all the steps mentioned in the
...
-This should result in a regular .so in ``\bin`` folder as if it was
+This should result in a regular .so file (built for ARM platform) in ``\bin`` folder, as if it was
 compiled with flags: ``tools=no target=debug``. The resulting file will
-be huge because it will contain all debug symbols, so for next builds,
-using ``target=release_debug`` or ``target=release`` is recommended.
-
-Copy the .so to the ``libs/armeabi`` Android folder (or symlink if you are
-in Linux or OSX). Note: Git does not support empty directories so you
-will have to create it if it does not exist:
+be huge because it will contain all debug symbols.
```
akien-mga: So no information anymore about the `target` parameter? The previous way to present things was bad (why make users build a binary during 15 minutes to finally tell them that they should use other parameter?), but removing the last part makes it worse :)

akien-mga:
The updates to the state as of 2.1 are good, but I'm not convinced by the "Easier to understand" parts.

reference/compiling_for_android.rst
KelinciFX: Sorry, late to reply (been having a lunch time)

It's not about which one dominate the market. Intel devices (which of course powered with x86 cpu) CAN run apk which compiled for ARM devices. On opposite side, ARM devices CAN'T run apk which compiled for Intel devices.

So, Device is Intel, and apk is compiled for ARM == Not a trouble, as Intel devices could be able to run Android application which build for ARM arch.

:-)

reference/compiling_for_android.rst
KelinciFX: Not correct. Mine is the correct one.

Should I fire up a screenshot?

:-)

akien-mga: Then it's a bug in the Windows buildsystem. For the official binaries built on OSX: https://github.com/GodotBuilder/godot-builds/blob/master/.travis.yml#L161

akien-mga: And it's also `libs/{release,debug}/armeabi-v7a` on Unix.

reference/compiling_for_android.rst
KelinciFX: It's up to you and the guidelines.

From my experience in guide writing, my style usually easier to understand. It's ok to not believe it. Just something like Code Folding, easier to grasp the whole concept/idea of a script.

But, again... If you can't accept it, I will follow previous style / yours.

:-)

bojidar-bg: It is acceptable only if you do the same style on all other pages, and then ensure that all future PRs and changes don't break the ordering, and finally tell users how to fix merge conflicts that would stem from it. :wink:

akien-mga: Well it kind of defeats the purpose of using Sphinx if we can't rely on it to generate the table of contents, but have to do it ourselves by appending "I. A. 1. a. α." to all headings based on their level. If we do want such section identifiers (I'm not sure of the gain for online docs with mostly short pages), I'm pretty sure there's a sphinx configuration option to get them generated automatically. Doing them manually is no-go.

reference/compiling_for_android.rst
KelinciFX: We are all know that Godot Engine is Open Source. Fortunately, I can confirm of being an Open Source fans. Dangerously, Godot is not just for Open Source people, instead for both Open Source and/or Propietary people, coming from GameMaker Studio, Unity, GameSalad, Construct, and so on.

OpenJDK 6+ or JDK 6+ == JDK 6+ or OpenJDK 6+

I'm sorry, but I believe it has nothing to do with which one should be implemented first (A first, B later), as it's not something like scons first and then gradle. It's about choice, which saying we can use A or B (B or C).

Now, why I re-arrange the position from OpenJDK or JDK to JDK or OpenJDK? It's simply because : 

- I guess JDK (mostly windows, mac users?) way more popular than OpenJDK. So, readers (especially new comers) could understand it quickly.
- Android SDK Docs does not even mention about OpenJDK (while now we are discussing about polishing android related docs, right?) The one mentioned by Google team for Windows, Mac, and Linux users is Oracle Java Development Kit (JDK) only, and none for OpenJDK (yeah, I can agree we should boom Google offices, for not "loving" OpenSource community.) 😄 

Again... It's all up to your style and understanding, because you're the boss after my prestigious men, Juan Linietsky and Ariel Manzur.

:-)

reference/compiling_for_android.rst
KelinciFX: Why not asking more elaborate?

Scons? the docs does mention it in the title, but the description is mostly about Android SDK and NDK. So, I managed it better by doing so (not?) :-)

But, I think I should write "refer to other docs" and linking it to other docs, instead of just saying "refer to their respective docs"

NDK r10e or later? Afaik, NDK older than r10e doesn't have arm-linux-androideabi-4.9 (while detect.py is now at 4.9), instead they have arm-linux-androideabi-4.8

:-)

reference/compiling_for_android.rst
KelinciFX: The idea is we should distinguish between Compiling the Engine for Android and Compiling Android Export Template. Despite they are mostly similar, it has a difference which could potentially lead a new comer being confused.

- target parameter : I think I should make it clearer
- Copy the .so bla bla bla ---> not exist : I believe it's not used anymore as randomshaper PR make it automagically happen.
- removing the last part makes it worse : As usual.. It's all up to the boss...

:-)

reference/compiling_for_android.rst
akien-mga: Well my point is that referring users to the SCons documentation does not make sense here. The SCons documentation is for people who need to develop a buildsystem using SCons. I think you're right to rename the section from `Setting up SCons` to `Setting up the buildsystem`, but then there's no need to mention SCons at all. What the docs should mention is what the users need to do.

> NDK r10e or later? Afaik, NDK older than r10e doesn't have arm-linux-androideabi-4.9 (while detect.py is now at 4.9), instead they have arm-linux-androideabi-4.8

No problem with that :)

reference/compiling_for_android.rst
akien-mga: Well as of now the engine (`tools=yes`) cannot be built for Android, only the export templates (`tools=no`).

Removing the information about copying the .so is good of course, but what I complain about is removing ", so for next builds, using ``target=release_debug`` or ``target=release`` is recommended." and keeping only "The resulting file will be huge because it will contain all debug symbols."

As it stands now, we just tell users:
- build with default parameters
- yeah it's use because it contains debug info, sorry about that! :)

So we should either tell them:

Version 1 (what was before)

- build with default parameters
- yeah it's use because it contains debug info, sorry about that! :)
- now use these custom parameters to actually build what you want (release and release_debug templates)

Version 2 (what would be better)

- build with those custom parameters to get the release and debug templates

reference/compiling_for_android.rst
KelinciFX: There is no bug in Windows buildsystem. Unix is the same as Windows one.
Both Windows and Unix has libs/{release,debug}/armeabi-v7a

The phrase below :
If the application runs but exits immediately, there might be one of the following reasons:
libgodot_android.so is not in ``lib/armeabi-v7a``

Doesn't refer to Gradle based Godot Android project, instead it refers to Exported Apk(s), in this case android_debug.apk and android_release.apk

Gradle based Godot Android project VS Exported Apks
libs/{release,debug}/armeabi-v7a  -- VS --  lib/armeabi-v7a

:-)

reference/compiling_for_android.rst
KelinciFX: I don't know which one of above is better.

If I was Juan, Ariel, or Other Docs Squad, I will just leaving Compiling Android Export Template only, and deleting Compiling the Engine for Android. The advantages are:
- Less confused
- It's straight to usability and what really matter (Android Export Template). Compiling with target=debug has little to no use.

:-)

reference/compiling_for_android.rst
akien-mga: Oh, my bad, thanks for the clarification.

reference/compiling_for_android.rst
akien-mga: Well yes, the current docs are bogus. Having a "Compiling" section and then "Compiling the export templates" makes no sense, as they both refer to the same thing (it's only possible to compile export templates). Not sure what/when it was made this way.

KelinciFX:
Revised by #216 . It's now safe to close this PR.

:-)

PR #217, opened 2016-08-03
Update class_raycast.rst

Marqin:
clarification that set_cast_to needs LOCAL point and get_collision_point gives GLOBAL point

### DIFF
--- a/classes/class_raycast.rst
+++ b/classes/class_raycast.rst
@@ -83,6 +83,8 @@ Member Function Description
 
 - :ref:`Vector3<class_vector3>`  **get_collision_point**  **(** **)** const
 
+Returns collision point. This point is in **global** coordinate system.
+
 .. _class_RayCast_is_colliding:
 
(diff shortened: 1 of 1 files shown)


akien-mga:
Could you make this PR on doc/base/classes.xml of the main repo, which is used to generate this documentation? See https://github.com/godotengine/godot-docs#editing-existing-pages

Marqin:
Oh. I pressed "edit this" on documentation page for that class, and this is what I got.

akien-mga:
Yeah I still haven't checked how to remove this link. It's good to have on the tutorials, but it should be removed on the whole classes/ folder. I've put a comment about this in each class .rst file + in the README, but you're not the first one to make the mistake, this link is really bad.

Marqin:
Ok. I will do this the proper way.

akien-mga:
I might have found how to remove the link from all pages. Until we do a proper template to filter out the classes/ folder, it might do the trick.

PR #221, opened 2016-08-10
Fixes for "Scripting" page under "Step-by-step" tutorial

landon912:
I updated the page to correctly reflect the current state of the editor interface. This change involved updating all the images, and rewording small portions of the tutorial to reflect the recent changes in workflow.


akien-mga:
Awesome work, thanks! My only concern is that the `img/signals.png` image has now a huge height, and it might look weird in the docs. But this can be improved later on (maybe by cutting it and showing only the two relevant areas side by side).

landon912:
Thanks for the quick reply. Here is the updated image in a new pull request. [(#222)](https://github.com/godotengine/godot-docs/pull/222)

PR #222, opened 2016-08-10
Small Image Fix for "img/signals.png"

landon912:
Made "img/signals.png" much smaller in vertical height and corrected which node was selected(From "Panel" to "Button")


akien-mga:
Awesome! :)

PR #223, opened 2016-08-10
document adding custom CXXFLAGS to Custom Modules

### LINKED ISSUE #137: Document how to add compile options only for the module
Need documentation to explain how to add compile options for the custom module only so they don't affect the whole engine. As discussed in godotengine/godot#4490.

Marqin:
fix #137

### DIFF
--- a/reference/custom_modules_in_c++.rst
+++ b/reference/custom_modules_in_c++.rst
@@ -158,6 +158,19 @@ this module:
 
     env.add_source_files(env.modules_sources,"*.cpp") # just add all cpp files to the build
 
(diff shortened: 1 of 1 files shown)


Marqin:
if you try doing it as documented earlier, in `config.py` it would break whole Godot build

vnen:
I guess this solves #137.

Marqin:
Still, many flags will break build because `.h` files included from Godot has many bugs, so flags like `-Werror` will break 😢 

In that case it's good to put some of module's mechanics in separate compilation unit (eg. compiled from hand via make) and link it via
`module_env.Library(source = 'something.o')`

PR #226, opened 2016-08-18
Update making_plugins.rst

tuga3d:
In the custom_node example the filenames are wrong (they are relative to the custom dock example).


bojidar-bg:
Looks good by me, @akien-mga

PR #238, opened 2016-09-03
Added mesa-libGLU-devel to Fedora oneliner

NiclasEriksen:
Build failed with this message:
`In file included from ./drivers/gles2/rasterizer_gles2.h:51:0,
                 from platform/x11/os_x11.cpp:30:
drivers/gl_context/GL/glew.h:1202:24: fatal error: GL/glu.h: No such file or directory
 #    include <GL/glu.h>`

mesa-libGLU-devel was missing.

### DIFF
--- a/reference/compiling_for_x11.rst
+++ b/reference/compiling_for_x11.rst
@@ -30,7 +30,8 @@ Distro-specific oneliners
 | **Fedora**    | ::                                                                                                         |
 |               |                                                                                                            |
 |               |     sudo dnf install scons pkgconfig libX11-devel libXcursor-devel libXrandr-devel libXinerama-devel \     |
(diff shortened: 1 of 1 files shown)


akien-mga:
What version are you building? AFAIR the dependency on GLU was unnecessary and removed for 2.1.

NiclasEriksen:
v2.2.alpha on linux x64.

vnen:
GLEW include the glu.h header. I believe there's a preprocessor definition to avoid it but Godot doesn't define it.

akien-mga:
Then we should probably define it, as we apparently don't use any of its features, so we don't like against it: 587826f

vnen:
I just noticed that the other oneliners actually have GLU.

PR #242, opened 2016-09-13
mention UWP limitations on Xbox One

Marqin:
taken from https://msdn.microsoft.com/en-us/windows/uwp/xbox-apps/system-resource-allocation


akien-mga:
I wonder if this wouldn't be better suited in the `Exporting for UWP` page? https://github.com/godotengine/godot-docs/blob/master/tutorials/asset_pipeline/exporting_for_uwp.rst

Marqin:
ok

PR #243, opened 2016-09-13
mention UWP limitations on Xbox One

Marqin:
taken from https://msdn.microsoft.com/en-us/windows/uwp/xbox-apps/system-resource-allocation


Marqin:
Also I'm not sure if 

>Apps that exceed these memory requirements will encounter memory allocation failures. For more information about monitoring memory use, see the MemoryManager class reference.

Is implemented properly. Maybe mention that?

Marqin:
@vnen probably knows

PR #248, opened 2016-09-22
Explain enums and mention ternary-if in GDScript reference

bojidar-bg:
Comment if you feel like I'm missing something (or if the rst is wrong...)

### DIFF
--- a/reference/gdscript.rst
+++ b/reference/gdscript.rst
@@ -75,6 +75,11 @@ here's a simple example of how GDScript looks.
     const answer = 42
     const thename = "Charly"
 
+    # enums (Godot 2.2+)
+
+    enum {UNIT_NEUTRAL, UNIT_ENEMY, UNIT_ALLY}
+    enum Named {THING_1, THING_2, ANOTHER_THING = -1}
+
     # built-in vector types
 
     var v2 = Vector2(1, 2)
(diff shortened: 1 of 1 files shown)


reference/gdscript.rst
```diff
@@ -516,6 +525,34 @@ expressions and must be assigned on initialization.
...
     const f = sin(20)  # sin() can be used in constant expressions
     const g = x + 20  # invalid; this is not a constant expression!
+    
+Enums
+^^^^^
+
+*Note, only available in Godot 2.2 or higher.*
+
+Enums are basically a shorthand for constants, and are pretty useful if you
+want to assing consecuitive integers to some constant.
```
akien-mga: *assign, *consecutive

bojidar-bg: ouch...

bojidar-bg: Fixed!

reference/gdscript.rst
```diff
@@ -516,6 +525,34 @@ expressions and must be assigned on initialization.
...
+    const TILE_FLOOR = 1
+    const TILE_SPIKE = 2
+    const TILE_TELEPORT = 3
+
+    enum State {STATE_IDLE, STATE_JUMP = 5, STATE_SHOOT}
+    # Is the same as:
+    const STATE_IDLE = 0
+    const STATE_JUMP = 5
+    const STATE_SHOOT = 6
+    const State = {STATE_IDLE = 0, STATE_JUMP = 5, STATE_SHOOT = 6}
```
Sslaxx: This is a little confusing , I think. I assume the `const State` line is not meant to be bunched up with the `const STATE_` statements, so perhaps you're missing a `# Is also the same as:` type line?

vnen: I think it is like this. You can use either `STATE_IDLE` or `State.STATE_IDLE`. Though this syntax wouldn't really work in GDScript, I believe it should be: 

```GDScript
const State = {"STATE_IDLE" = 0, "STATE_JUMP" = 5, "STATE_SHOOT" = 6}
```

bojidar-bg: @vnen My syntax would work in GDScript... yours won't, except if you replace the `=` signs with `:` :D

vnen: Oh, I stand corrected then. This quirk I didn't know.

akien-mga: Yeah, GDScript supports both Python-style (`{"key": value}`) and Lua-style (`{key = value}`) dictionary declarations: http://docs.godotengine.org/en/latest/reference/gdscript.html#dictionary

PR #262, opened 2016-10-06
Replaced two Kinematic Character documentation images

Bonfi96:
Any comment or suggestion about the images? Suggest any edit you think should be done!


akien-mga:
It looks great to me @Bonfi96.

PR #265, opened 2016-10-08
Updated an image of the raycast page

Bonfi96:
Should I upload the svg original files somewhere?


akien-mga:
Good idea, let's make an `img/src` folder for that.

PR #266, opened 2016-10-08
Replaced images of the Vector Math page and added sources

Bonfi96:


Bonfi96:
If there's anything else that needs a fix feel free to report it

akien-mga:
It's perfect, thanks. Nice upgrade from the MS Paint artwork :D

PR #272, opened 2016-10-17
Added example showing target=release_debug

raould:
see https://github.com/godotengine/godot/issues/2095


reference/introduction_to_the_buildsystem.rst
```diff
@@ -86,6 +86,20 @@ Example for using 4 cores:
 
     user@host:~/godot$ scons platform=x11 -j 4
 
+The default build is not heavily optimized, to allow
```
akien-mga: Actually not optimized at all, it's a debug build :)

Could you make a new section for that? You put it at the right place, but it shouldn't be under "Platform selection", best make a new "Optimized build" section.

raould:
ok i will update as soon as i can :-)

akien-mga:
It seems you rebased but did not do the changes mentioned?

akien-mga:
> It seems you rebased but did not do the changes mentioned?

Ping.

raould:
working on it now...

PR #273, opened 2016-10-21
Added example of export (PackedScene)

Tybobobo:
Added export (PackedScene), which can be used to select a scene file from your files.


eska014:
Exporting resources is already covered, so please add it in there, otherwise it looks like this is a special, separate export method

```
# Export can also take a resource type to use as a hint

export(Texture) var character_face
```

PR #275, opened 2016-10-22
Warning+fix about compiling in Ubuntu >= 16.10

hetdegon:
Ubuntu 16.10 ships with GCC 6.2.0 as default, which currently causes issues. Added a warning and instructions to install and force usage of GCC 5.4.X at the moment of writing.

### DIFF
--- a/reference/compiling_for_x11.rst
+++ b/reference/compiling_for_x11.rst
@@ -79,6 +79,19 @@ If you wish to compile using Clang rather than GCC, use this command:
 ::
 
     user@host:~/godot$ scons platform=x11 use_llvm=yes
+    
+If you are using Ubuntu >=16.10, it includes GCC 6.2 as default,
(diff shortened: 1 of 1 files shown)


akien-mga:
That's a good idea, though I'm not fond of always writing Ubuntu-centric documentation, there are tons of other worthy distros in the wild. Could you instead mention that GCC 6 is not supported for now, and that people should try to install GCC 5 or clang and specify the CXX compiler accordingly? Then you can give the one-liner for Ubuntu users if really needed, but the paragraph should not start as "If you're using Ubuntu X" - many distros use GCC 6 out of the box now.

hetdegon:
There's a small disclaimer above, but it's pretty easy to miss. Maybe I should put it in some glaring warning box or something?

akien-mga:
Superseded by e4660c9.

PR #290, opened 2016-11-01
Update compiling_for_osx.rst

songproducer:
Updating compile instructions for MacOS

### DIFF
--- a/reference/compiling_for_osx.rst
+++ b/reference/compiling_for_osx.rst
@@ -22,7 +22,7 @@ Start a terminal, go to the root dir of the engine source code and type:
 
 ::
 
-    user@host:~/godot$ scons platform=osx
+    user@host:~/godot$ scons platform=osx bits=64 target=release_debug
 
 If all goes well, the resulting binary executable will be placed in the
 "bin" subdirectory. This executable file contains the whole engine and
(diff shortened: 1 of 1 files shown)


reference/compiling_for_osx.rst
```diff
@@ -22,19 +22,19 @@ Start a terminal, go to the root dir of the engine source code and type:
...
 manager.
 
 To create an .app like in the official builds, you need to use the template
-located in ``tools/Godot.app``. Typically:
+located in ``tools/dist/osx_tools.app/Contents/MacOS/Godot.app``. Typically:
 
 ::
 
-    user@host:~/godot$ cp -r tools/Godot.app .
+    user@host:~/godot$ cp -r tools/dist/osx_tools.app/Contents/MacOS/Godot.app .
```
akien-mga: Here it should be `cp -r tools/dist/osx_tools.app Godot.app`

reference/compiling_for_osx.rst
```diff
@@ -22,19 +22,19 @@ Start a terminal, go to the root dir of the engine source code and type:
 
 ::
 
-    user@host:~/godot$ scons platform=osx
+    user@host:~/godot$ scons platform=osx bits=64 target=release_debug
```
akien-mga: I'd prefer we don't specify these arguments, people can also compile with `bits=32` or `bits=fat` (the latter is used for official binaries), and `target=debug` if they want. Instead, a small comment could be added before `user@host:~/godot$ cp bin/godot.osx.opt.tools.64 Godot.app/Contents/MacOS/Godot` to say that the binary name may different depending on build options used with scons.

akien-mga:
Superseded by #297.

PR #291, opened 2016-11-02
Rename WinRT platform to UWP

vnen:
Update the docs with the change introduced in godotengine/godot#7019


StraToN:
Seems good for me, although I wonder if users who are not used to this name yet will feel lost?

PR #299, opened 2016-11-09
Short tutorial on boot splash

ZX-WT:
This is a common question, so I decided to add this to the doc. 

I have trouble in syncing my fork with this master branch, just add this short tutorial on boot splash.


akien-mga:
Something went wrong in your upload, you seem to have changed stuff in the gridmap tutorial. Also, the screenshots of the options you used are a bit too low-res, it's hard to read what the options are. It would be better if you downsize your Godot instance before opening the options to get a smaller window, and grab only the relevant section of the screen in native resolution.

Could you try to rebase/redo the PR to fix this? Also, I would advise against using GitHub's drag and drop upload for that, it's IMO a very bad decision from GitHub to try to make version control like Dropbox. Learning the basics of git is not easy, but it's definitely worth it in the long run.

ZX-WT:
I am the one who started the gridmap tutorial, maybe someone updated it and I don't why I can't merge the latest doc to my fork... Then, I know I cannot rely on github and go for git, I manage to fix the issue by accidentally replaced the gridmap tutorial. I knew something went wrong but don't know how to fix it
sigh, I will start fresh this time and solely rely on git to avoid the problem, fixing those pics are easy, just wait

ZX-WT:
You can close this, I created a new fork instead and created new pull request, I know this is not a good way to fix things, I will learn to use git in better ways, it is so tempting to use the drag and drop for lazy users like me, github is so evil in providing me this option ;p I will ignore this function from now on and rely on git instead.

PR #305, opened 2016-11-20
Fixed some parts where code was rendered as normal text

popenke:
Fixed some code parts that where rendered as normal text. There is also one variable - 'direction' - that is not anywhere in code examples and I don't know how fix it.


popenke:
I can delete this branch right now? Sorry, it's my first contribution.

vnen:
@popenke yes, you can delete the branch

PR #307, opened 2016-11-20
fix-7160-broken-code-blocks

avril-gh:
some lines of code was broken in half by CRLF which then was detected as `end-of-code-block`.
Rest of code which was after mentioned CRLF was displayed as normal text outside of its code block.


akien-mga:
Hopefully it was not CRLF but LF, we're using Unix line endings :)

PR #308, opened 2016-11-20
Document how to build custom C++ module as shared library

touilleMan:
Following https://github.com/godotengine/godot/issues/3936#issuecomment-260715374

There is still something I not sure about:

```
    # Finally notify the main env it has our shared lirary as a new dependency.
    # To do so, SCons wants the name of the lib with it custom suffixes
    # (e.g. ".x11.tools.64") but without the final ".so".
    # We pass this along with the directory of our library to the main env.
    shared_lib_shim = shared_lib[0].name.rsplit('.', 1)[0]
    env.Append(LIBS=[shared_lib_shim])
    env.Append(LIBPATH=['#bin'])
```

During the linking phase of godot, we have to pass a `-lmy_module.so` attribute and a `-L/path/to/my_module`. If we provide the path withing the `-l` attribute (what scons does when passing `env.Append(LIBS=[shared_lib])`), the binary fails to load the .so.
However to do so I have to rely on ` shared_lib[0].name.rsplit('.', 1)[0]` which look like a hack to me. Is there any better way to do ?

### DIFF
--- a/reference/custom_modules_in_c++.rst
+++ b/reference/custom_modules_in_c++.rst
@@ -220,6 +220,106 @@ now do:
 
 And the output will be ``60``.
 
+Improving the build system for development
+------------------------------------------
+
+So far we defined a clean and simple SCsub that allows us to add the sources
+of our new module as part of the Godot binary.
+
+This static approach is fine when we want to build a release version of our
+game given we want all the modules in a single binary.
+
+However the trade-of is every single change means a full recompilation of the
+game. Even if SCons is able to detect and recompile only the file that have
+changed, finding such files and eventually linking the final binary is a
+long and costly part.
+
+The solution to avoid such a cost is to build our own module as a shared
+library that will be dynamically loaded when starting our game's binary.
+
+.. code:: python
+
+    # SCsub
+    Import('env')
+
+    sources = [
+        "register_types.cpp",
+        "sumator.cpp"
+    ]
+
+    # First, create a custom env for the shared library.
+    module_env = env.Clone()
+    module_env.Append(CXXFLAGS='-fPIC')  # Needed to compile shared library
+    # We don't want godot's dependencies to be injected into our shared library.
+    module_env['LIBS'] = []
+
+    # Now define the shared library. Note that by default it would be built
+    # into the module's folder, however it's better to output it into `bin`
+    # next to the godot binary.
+    shared_lib = module_env.SharedLibrary(target='#bin/sumator', source=sources)
+
+    # Finally notify the main env it has our shared lirary as a new dependency.
+    # To do so, SCons wants the name of the lib with it custom suffixes
+    # (e.g. ".x11.tools.64") but without the final ".so".
+    # We pass this along with the directory of our library to the main env.
+    shared_lib_shim = shared_lib[0].name.rsplit('.', 1)[0]
+    env.Append(LIBS=[shared_lib_shim])
+    env.Append(LIBPATH=['#bin'])
+
+Once compiled, we should end up with a ``bin`` directory containing both the
+``godot*`` binary and our ``libsumator*.so``. However given the .so is not in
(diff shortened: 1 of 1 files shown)


ScotFlux:
Pretty raaaddd. You've just got Godot 🏆 and 🎖 as well, my man... Congratulations!

bojidar-bg:
@ScotFlux beware, your comments are quickly getting into my GH autoignore list :wink:

ScotFlux:
@bojidar-bg DO IT AS YOU LIKE, MAN!!! I DON'T CARE!!! I DON'T KNOW WHAT'S YOUR PROBLEM, YOU ALWAYS ACT AS AN ENEMY TO ME. SO, JUST DO IT!!!

@reduz I suggest you to change Godot Engine to Bojidar Engine, so I can safely stop my self in spreading / promoting / suggesting Godot Engine to game developers out there. I believe I never being a rude guy, instead I always want to make laugh, joke, and smile, so people will feel that Godot Engine development is enjoyable. I want to see happy and healthy environment here.

bojidar-bg:
Well, I'm also making jokes, why don't you laugh instead of being angry :disappointed:. I was just trying to say that you posted nearly the same comment on 3-4 PRs, nothing more...

touilleMan:
@bojidar-bg, @ScotFlux since there is still people here, maybe you can help 😄 

I've noticed a dangerous behavior

this works:
```
 user@host:~/godot$ scons sumator_shared=yes bin/sumator.x11.tools.64.so
```

this seems to work, but in fact doesn't do anything unless the .so has been deleted
```
 user@host:~/godot$ scons sumator_shared=yes /home/user/godot/bin/sumator.x11.tools.64.so
```

The reason is `bin/sumator.x11.tools.64.so` is the name of the target of the shared library as specified in the scons script (the  `target='#bin/sumatort' stuff` so scons can map the dependencies.
On the other hand, if we only give a regular path scons just make sure the file is present and if so doesn't check for it dependancies.

You can add `--tree=all` option and run the two scons commands to see clearly the trouble.

I think this could lead to serious trouble when people mix with this and have the feeling the build system is somehow broken (the .so is sometime not rebuilding properly). So we should find a way to fix this or at least put a big warning in this tutorial...

bojidar-bg:
Well, maybe you should just document that one cannot use absolute file paths with this setup?

arcanis:
By any chance, has someone been able to follow this guide on Windows? I get a lot of undefined references on the godot symbols, and that's it 😢

touilleMan:
@arcanis this thing is pretty outdated, the new fancy way to do this is to use GDnative (so you no longer have to compile your module within godot)

akien-mga:
See also https://github.com/godotengine/godot/issues/11766 which mentions linking issues. Should we just revert the additions from this PR, since it's no longer a recommended method for 3.0 (while the rest of the doc for static modules is still valid)?

touilleMan:
@akien-mga the only usecase I see for this now GDNative is available is as a recipe for Godot core developers when working on big module with frequent recompilations (in order to avoid the loooong linking process). Like maybe the mono module.

arcanis:
Fwiw my experimentation made me to believe that this was not a viable solution on Windows (DLL need the compiled base app to reference symbols from the main app, and the main app needs the compiled DLL to reference symbols from it, so the circular dependency makes things pretty hairy. There's apparently a way to somehow generate a manifest file to fix this, but I haven't been able to find documentation yet).

However, what I did was to move my native module into its own static library. The HUGE compilation time I had (more than 15 minutes for each compilation) was immediately reduced to less than a minute. Funnily, the reason for this is not the linking, but rather the `ar q` command: for some reason it takes an enormous amount of time to execute (the linking is comparatively pretty fast). Moving my files into their own library avoided the archive to have to be rebuilt for each compilation. So if you want to tackle the compilation time, my advice would be to change the build system to compile each native module separately, so that any of them can be recompiled independently.

PR #315, opened 2016-12-16
Fixed some commands for scons

josempans:
Changed platform to iphone but with x86 arch for simulator
Added arch parameter to 32 and 64 targets to get the correct files.

### DIFF
--- a/reference/compiling_for_ios.rst
+++ b/reference/compiling_for_ios.rst
@@ -19,13 +19,13 @@ Open a Terminal, go to the root dir of the engine source code and type:
 
 ::
 
-    $ scons p=iphone bin/godot.iphone.debug
+    $ scons p=iphone target=debug
 
 for a debug build, or:
 
 ::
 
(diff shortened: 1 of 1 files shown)


josempans:
Corrections to the whole compiling document, not only the universal runtime issue. There is no need for include the output name.

reference/compiling_for_ios.rst
```diff
@@ -44,9 +44,9 @@ It can be done in three steps, first compile 32 bit version, then compile 64 bit
 
 ::
 
-    $ scons p=iphone tools=no bits=32 target=release 
-    $ scons p=iphone tools=no bits=64 target=release
-    $ lipo -create bin/godot.iphone.opt.32 arm64 bin/godot.iphone.opt.64 -output bin/godot.iphone.opt.universal
+    $ scons p=iphone tools=no bits=32 target=release arch=arm
```
akien-mga: The `bits` argument should also be removed.

josempans: good, I didn't tried without it

PR #319, opened 2016-12-31
Fixed all instances of "bellow" to "below"

jacobspeicher:
I don't know why it's displaying as 435 additions and 6 deletions, I only deleted 6 of the letter L.


akien-mga:
That's because you copy pasted the whole file in itself again, compare the original: https://github.com/godotengine/godot-docs/blob/master/reference/compiling_for_windows.rst with yours: https://raw.githubusercontent.com/godotengine/godot-docs/master/reference/compiling_for_windows.rst (see how "Compiling for Windows" shows up twice).

PR #324, opened 2017-01-08
Fix code snippet in class_editorsettings

touilleMan:


akien-mga:
See the header of the file:
```
```
:)

The corresponding code in classes.xml probably lacks the `[codeblock]` tags.

touilleMan:
Thanks for pointing me the right way :smile: 
https://github.com/godotengine/godot/pull/7468

akien-mga:
Thanks :)

PR #335, opened 2017-01-30
grammar fix

cbscribe:
it's -> its


cbscribe:
Found more, so see my other pull request.

PR #336, opened 2017-01-30
Cbscribe grammar patches1

cbscribe:
Noticed a bunch of incorrect uses of "it's" in the docs when I was reading them, so I wrote a script to find them (hopefully) all.

### DIFF
--- a/classes/class_@gdscript.rst
+++ b/classes/class_@gdscript.rst
@@ -157,7 +157,7 @@ Member Functions
 Numeric Constants
 -----------------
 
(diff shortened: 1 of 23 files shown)


akien-mga:
Note that the changes made to `class_*.rst` files will be overwritten if they're not fixed in https://github.com/godotengine/godot/blob/master/doc/base/classes.xml, as they are generated from that file.

cbscribe:
Thanks - I'll issue a pull request for those too.

PR #337, opened 2017-01-30
Add function remove_line in RichTextLabel

Dumbeldor:
Add function : https://github.com/godotengine/godot/pull/7653


bojidar-bg:
Quoting the edited file:
> .. Generated automatically by doc/tools/makerst.py in Godot's source tree.
> .. DO NOT EDIT THIS FILE, but the doc/base/classes.xml source instead.

Quoting @akien-mga:
> Note that the changes made to class_*.rst files will be overwritten if they're not fixed in https://github.com/godotengine/godot/blob/master/doc/base/classes.xml, as they are generated from that file.

I guess that can be closed.

PR #358, opened 2017-03-20
Improve physics doc

ombre5733:
- Fixed some typos.
- Added some additional text.
- Reordered sections: StaticBody2D, KinematicBody2D and RigidBody2D belong together as well as CollisionShape2D and CollisionPolygon2D.

### DIFF
--- a/tutorials/2d/physics_introduction.rst
+++ b/tutorials/2d/physics_introduction.rst
@@ -60,9 +60,10 @@ motion casting, etc.
 Transforming shapes
 ~~~~~~~~~~~~~~~~~~~
 
-As seen before in the collide functions, 2D shapes in godot can be
+As seen before in the collide functions, 2D shapes in Godot can be
 transformed by using a regular :ref:`Matrix32 <class_Matrix32>`
-transform, meaning they can check collision while scaled, moved and
+transform, meaning the functions can check for collisions while the 
+shapes are scaled, moved and
 rotated. The only limitation to this is that shapes with curved sections
 (such as circle and capsule) can only be scaled uniformly. This means
 that circle or capsule shapes scaled in the form of an ellipse **will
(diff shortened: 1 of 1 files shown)


akien-mga:
Same here, you should remove commits 94c5166 and 4dd735d from this branch using `git rebase -i HEAD~3`.

ombre5733:
Done.

akien-mga:
The reordering is a good idea, but I'd go a step further, as it's still somewhat unclear. Adding another level of heading would likely make things clearer:
```
Collision objects
-----------------

// intro, present the typical outline for collision objects with:
- Any CollisionObject2D
  * CollisionShape2D or CollisionPolygon2D

Collision shapes
~~~~~~~~~~~~~

CollisionShape2D
^^^^^^^^^^^^^^

CollisionPolygon2D
^^^^^^^^^^^^^^

Triggers
^^^^^^^^

CollisionObject2D
~~~~~~~~~~~~

StaticBody2D
^^^^^^^^^^^^^^

KinematicBody2D
^^^^^^^^^^^^^^^

RigidBody2D
^^^^^^^^^^^^^
```

Though in a second step it might be worth moving all this + the collision reporting section to its own tutorial page, to make that "introduction" clearer.

ombre5733:
@akien-mga Ok, this change will need some more time. 

When we are at it, I would like to clarify more things in the documentation:

- Why is there `StaticBody2D` and `KinematicBody2D`, when the same functionality can be achieved with the `Mode` flag of `KinematicBody2D`? Reading just the docs, having just a `KinematicBody2D` with appropriate mode would be sufficient.
- Why is there no `CharacterBody2D` but a flag for it in `KinematicBody2D`? The table near the end shows a `CharacterBody` but Godot has none, so this needs to be fixed as well.
- Does it make sense that `Trigger` is a flag of shapes rather than the `CollisionObject2D`? I always look at the object first to set this flag only to remember later on that I have to change the shape/polygon.

akien-mga:
Let's merge this one as is and maybe move the discussion to a separate issue, @StraToN is working on a refactoring of the way the docs are organized, so files will be moved around a bit, which would introduce conflicts in pending PRs.

Regarding your questions, I will answer soon-ish :)

PR #359, opened 2017-03-23
Custom C++ Module example snippets

ssj71:
just some clarifications and examples that would have been helpful to me as I'm trying to figure this out.


akien-mga:
Merged manually and improved as 447b439.

PR #361, opened 2017-03-28
Remove misspelled duplicate text ('conves shapes')

fleskesvor:


akien-mga:
Merged manually as 4f9d2ba as we moved the file around so the patch wouldn't apply anymore.

PR #364, opened 2017-04-04
Fixed typo

jascase901:


akien-mga:
Merged as 3d01287 together with #365, as it was a bit overkill to get 4 commits (including two merge commits) for a 2-char change :D

PR #365, opened 2017-04-04
fixed typo

jascase901:


jascase901:
On the contrary, if you shape is small  ->  On the contrary, if your shape is small

akien-mga:
Merged as 3d01287 together with #364, as it was a bit overkill to get 4 commits (including two merge commits) for a 2-char change :D

PR #371, opened 2017-04-15
Update compiling_for_osx.rst

Melix19:
Update "Compiling for osx" to use the new reordered folders.

### DIFF
--- a/development/compiling/compiling_for_osx.rst
+++ b/development/compiling/compiling_for_osx.rst
@@ -35,9 +35,9 @@ supports both 32-bit and 64-bit architectures):
 
 ::
 
-    user@host:~/godot$ cp -r tools/dist/osx_tools.app ./Godot.app
+    user@host:~/godot$ cp -r misc/dist/osx_tools.app ./Godot.app
     user@host:~/godot$ mkdir -p Godot.app/Contents/MacOS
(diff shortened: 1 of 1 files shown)


development/compiling/compiling_for_osx.rst
```diff
@@ -35,9 +35,9 @@ supports both 32-bit and 64-bit architectures):
 
 ::
 
-    user@host:~/godot$ cp -r tools/dist/osx_tools.app ./Godot.app
+    user@host:~/godot$ cp -r misc/dist/osx_tools.app ./Godot.app
     user@host:~/godot$ mkdir -p Godot.app/Contents/MacOS
-    user@host:~/godot$ cp bin/godot.osx.opt.tools.fat Godot.app/Contents/MacOS/Godot
+    user@host:~/godot$ cp bin/godot.osx.tools.fat Godot.app/Contents/MacOS/Godot
```
akien-mga: The `opt.` part should be kept, usually if you're making an `.app` it's better to use the optimized binary. What's missing though is a mention in the previous paragraph that scons should be run with `target=release_debug`.

akien-mga:
Superseded by 2d7b6ac. (But thanks for the patch!)

PR #379, opened 2017-05-06
Change minor wording

tlahn:


hubbyist:
I think it should be "complete control **_over_** every single piece of it".

tlahn:
Sounds reasonable. I'll drop this and recreate another change. See #380 Thanks

PR #385, opened 2017-05-17
Add instructions for getting the source code to compile

Bugsbane:
Add how to get the actual source code needed for compiling so people can follow along in a more step by step fashion.

### DIFF
--- a/development/compiling/compiling_for_x11.rst
+++ b/development/compiling/compiling_for_x11.rst
@@ -72,6 +72,18 @@ Distro-specific oneliners
 |               |         libxrandr-dev                                                                                      |
 +---------------+------------------------------------------------------------------------------------------------------------+
 
+Get the Source Code
+-------------------
+
+To get the latest (unstable) current source code, in a terminal enter:
+
+::
+
+    git clone https://github.com/godotengine/godot.git
+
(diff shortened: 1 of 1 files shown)


akien-mga:
I don't think it is the right place for that. Fetching the source code is relevant for every single platform, not just for x11, so it should go on http://docs.godotengine.org/en/stable/development/compiling/introduction_to_the_buildsystem.html or maybe a new page.

Bugsbane:
Ok, but can we at least add a link to the page that describes how to get the source code? I would suggest adding this link to the "how to compile" pages of each platform.

As someone new to Godot, I looked up "how to compile godot" and Google lead me to this page, which then told me to start compiling without any hints as to where I could get the source from. Instructions on how to compile obviously aren't much use without source code.

akien-mga:
Yeah, it should be on a common page with a link to it in all platform-specific pages.

We can use the nice `seealso` markup for that:
```
.. seealso:

    To compile the engine, you will first need to `download the source code <doc_download_source>`_.
```

JuanNeedsToApologize:
Definitely will need to learn how to re-compile if you want to change the executable icon!

mhilbrunner:
@Bugsbane Want to add a comon page as described by akien above? You can use `git commit --amend` or close this PR and add a new one.

PR #396, opened 2017-06-21
Specify location of "Install Export Templates"

demipixel:
Just got a little lost looking through the top left menus in the editor while reading the docs.


vnen:
This does not apply to `master` though.

akien-mga:
Merged manually in the 2.1 branch as 544c06b.

PR #401, opened 2017-07-04
Fixed wrong method name

AndreaCatania:


akien-mga:
Though that's correct, there are still many references to the old "type" API instead of the new "class" one in the file. It would be nice to fix them all.

akien-mga:
Superseded by 5aceaeb.

PR #406, opened 2017-07-07
fix broken link to project organization

clofresh:


akien-mga:
To avoid such issues the link should have used a reference to `_doc_project_organization` in the first place, which is kept by Sphinx even if the file is moved around.

Could you change the link from <code>\`Project Organization <../workflow/project_setup/project_organization.html>\`_</code> to <code>:ref:\`doc_project_organization\`</code> instead?

(cc @StraToN who used the wrong relative link in the first place ;))

clofresh:
Ah good to know about `:ref:`. Updated the PR

PR #407, opened 2017-07-08
Changed Description

henriiquecampos:
Changed the description to better fit the node behaviors, since it can add margins to every direction


akien-mga:
See a few lines above:

> .. Generated automatically by doc/tools/makerst.py in Godot's source tree.

:)

PR #412, opened 2017-07-16
Update some pages for Godot 3

### LINKED ISSUE #408: Inconsistency in documentation regarding dictionaries
In the article http://docs.godotengine.org/en/stable/learning/scripting/gdscript/gdscript_advanced.html#doc-gdscript-more-efficiently
there is a sample source, where it is stated that

`# same example, lua-style support`
`...`
`d.mother = "rebecca" # this doesn't work (use syntax below to add a key:value pair)`
`d["mother"] = "rebecca" # this works`

but in the article http://docs.godotengine.org/en/stable/learning/scripting/gdscript/gdscript_basics.html#doc-gdscript
is another sample source, which reads exactly opposite:

`var d = {} # create an empty Dictionary`
`d.Waiting = 14 # add String "Waiting" as a key and assign the value 14 to it`
`...`

So, is it allowed to create new keys in dictionaries using the following syntax:

`dictionary.new_key = "new value"`?

I'm new to GScript, so I cannot tell which one is valid or invalid neither I will take responsibility for submitting changes on github (possible these were for different versions of engine/language?)...

Noshyaar:
Fixes #408

### DIFF
--- a/about/introduction.rst
+++ b/about/introduction.rst
@@ -54,7 +54,7 @@ open source `Sphinx <http://www.sphinx-doc.org>`_ and `ReadTheDocs
 
 .. note:: You can contribute to Godot's documentation by opening issue tickets
           or sending patches via pull requests on its GitHub
-          `source repository <http://github.com/godotengine/godot-docs>`_.
+          `source repository <https://github.com/godotengine/godot-docs>`_.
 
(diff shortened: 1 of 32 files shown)


akien-mga:
Awesome work, thanks!

PR #419, opened 2017-07-25
Separate `extends` keyword from `is` keyword.

### LINKED ISSUE #403: GDScript `extends` was replaced by `is` for type checking
The [GDSctipt reference](http://docs.godotengine.org/en/latest/learning/scripting/gdscript/gdscript_basics.html) wasn't updated to reflect the addition of the `is` operator to check the instance type.

StraToN:
Fixes #403

### DIFF
--- a/classes/class_editorplugin.rst
+++ b/classes/class_editorplugin.rst
@@ -162,7 +162,7 @@ Add a custom type, which will appear in the list of nodes or resources. An icon 
 
 When given node or resource is selected, the base type will be instanced (ie, "Spatial", "Control", "Resource"), then the script will be loaded and set to this object.
 
-You can use the :ref:`EditorPlugin.handles<class_EditorPlugin_handles>` to check if your custom object is being edited by checking the script or using 'extends' keyword.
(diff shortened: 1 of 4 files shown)


classes/class_editorplugin.rst
```diff
@@ -162,7 +162,7 @@ Add a custom type, which will appear in the list of nodes or resources. An icon
 
 When given node or resource is selected, the base type will be instanced (ie, "Spatial", "Control", "Resource"), then the script will be loaded and set to this object.
 
-You can use the :ref:`EditorPlugin.handles<class_EditorPlugin_handles>` to check if your custom object is being edited by checking the script or using 'extends' keyword.
```
akien-mga: You're co-maintainer of this repo and don't read its README? ;) https://github.com/godotengine/godot-docs/blob/master/README.md#editing-existing-pages

And file header:
>  .. Generated automatically by doc/tools/makerst.py in Godot's source tree.

It doesn't hurt anyway, just think about making the change in classes.xml or it will be overridden next time we sync.

StraToN: Haha! Crap, I just followed my list of grep results. :D

PR #432, opened 2017-08-01
update script

liungkejin:
script api updated, _gui_input() has been dropped


vnen:
It's actually the opposite: `_input_event()` got renamed to `_gui_input()`. Those docs are about the Godot 3.0 version.

liungkejin:
Okay....

PR #445, opened 2017-08-24
Fix matrix multiplication

Meseira:
- Replace two erroneous addition symbols by multiplication ones in "Matrix multiplication"
- Replace erroneous addition symbols by multiplication ones in "Matrix tips" and remove the useless pair of parentheses. Indeed, matrix multiplication satisfies the associative property.

### DIFF
--- a/learning/features/math/matrices_and_transforms.rst
+++ b/learning/features/math/matrices_and_transforms.rst
@@ -373,7 +373,7 @@ However, this is not the same:
     # yields a different results
     pos = (transform1 * transform2).xform(pos)
 
-Because in matrix math, A + B is not the same as B + A.
+Because in matrix math, A * B is not the same as B * A.
 
 Multiplication by inverse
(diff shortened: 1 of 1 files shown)


learning/features/math/matrices_and_transforms.rst
```diff
@@ -373,7 +373,7 @@ However, this is not the same:
     # yields a different results
     pos = (transform1 * transform2).xform(pos)
 
-Because in matrix math, A + B is not the same as B + A.
+Because in matrix math, A * B is not the same as B * A.
```
akien-mga: I wonder how such bogus doc could have been written in the first place :P

It's not the first time we have such fixes to those math docs, I think more than a human error, it might actually have been a fallback of converting from GitHub wiki's Markdown to OpenProject wiki's Textile and finally to Sphinx's reST. :)

PR #446, opened 2017-08-26
Fix Viewport.size type is Vector2 not Rect2

letheed:


Noshyaar:
> Editing existing pages
> 
> To edit an existing page, just locate its .rst source file and open it in your favourite text editor. You can then commit the changes, push them to your fork and make a pull request. **Note that the pages in classes/should not be edited here, they are automatically generated from Godot's XML class reference.**

vnen:
Also, in this case the type is generated from the bindings, so that need to be corrected in code if it's wrong.

akien-mga:
Fixed by https://github.com/godotengine/godot/pull/10680.

PR #453, opened 2017-09-01
Added tabs for multiple programming languages

djrm:
allows this
screenshot from 2017-08-31 23-57-27


akien-mga:
Awesome! For the reference, that's using this extension: https://github.com/djungelorm/sphinx-tabs

akien-mga:
BTW, it looks like we might be able to rely on pip to install the extension, and thus not need to bundle it ourselves: http://docs.readthedocs.io/en/latest/faq.html?highlight=requirements.txt#my-project-isn-t-building-with-autodoc
But for that time being, what works works :)

reduz:
This is great, did you check that we can use an image for visual script?

akien-mga:
> This is great, did you check that we can use an image for visual script?

Yes, check the example here (scroll down): http://docs.godotengine.org/en/latest/learning/step_by_step/scripting.html#handling-a-signal

neikeq:
Would be nice if we could have something like this for the class reference as well. Not idea what would be the best way to identify `[code]` blocks for translating them though.

akien-mga:
We can easily parse the XML and output the correct reST markup to use that extension, but indeed we need to decide on how we want to implement code examples in the classref. The current `[codeblock]` is a hack and not very scalable.

djrm:
@akien-mga i could not make it wotk with pip, maybe because im declaring ./extensions as extensions folder, in order to have the gdscript parser.

hpvb:
I think you need [CDATA] blocks if you want to allow the forbidden characters in XML at all though...

PR #455, opened 2017-09-06
Step-by-Step Scenes and Nodes 3.0 fixes

cvvs:
Several fixes found while following the godot step-by-step tutorial:
* Settings is access from Project
* Main Scene is under Run

Do I need to indicate the space in "Main Scene" with an underscore? Should I keep the lower case?

### DIFF
--- a/learning/step_by_step/scenes_and_nodes.rst
+++ b/learning/step_by_step/scenes_and_nodes.rst
@@ -180,10 +180,10 @@ you can change in this file to alter how a project executes. To simplify this
 process, Godot provides a project settings dialog, which acts as a sort of
 frontend to editing a project.godot file.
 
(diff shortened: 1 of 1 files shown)


akien-mga:
`application/run/main_scene` is actually the name of the property, and how it's referenced in your project.godot:
```
[application]

run/main_scene="res://Control.tscn"
```
So I think it makes sense to keep it like this. Note that unlike markdown, \` doesn't delimit inline code but links in reST. For inline code, you need to double it: \`\`some code\`\`.

cvvs:
Yes, but the instructions are for changing this from the GUI, not editing the .godot file.

NathanLovato:
Checked, and yes the tut only covers the UI. Merging. Note this page will likely be replaced in the +/- near future, so no need to further edit it.

PR #469, opened 2017-09-14
use new get_node ($) in intro

toger5:


about/introduction.rst
```diff
@@ -6,7 +6,7 @@ Introduction
 ::
 
     func _ready():
-        get_node("Label").set_text("Hello world!")
+        $Label.set_text("Hello world!")
```
vnen: I think it would be better to also use members, so `$Label.text = "Hello World!"`.

toger5: lol obviously...

toger5:
updated...

cbscribe:
I like it!

PR #477, opened 2017-09-15
Fixed broken README link to XML class references

rminderhoud:


NathanLovato:
Thanks for the fix!
@Noshyaar in this case, do we merge directly or is there something else to do to keep the git history clean?

NathanLovato:
I guess it's okay in the docs repo?

Noshyaar:
@NathanLovato 
- Check if it's rebased.
- Check if the commit log is meaningful enough.
- Inform the author if the email doesn't match their Github Account. (rarely happens)

In this case, I think it was okay to merge.
(I don't have merge permission in this repo btw.)

PR #479, opened 2017-09-16
GDScript style guide

cbscribe:
Goal: provide a centralized location for "good" GDScript coding practices.

Should be used for all code samples in documentation (mostly is already, but we should have a place to point new contributors).

### DIFF
--- a/learning/scripting/gdscript/gdscript_styleguide.rst
+++ b/learning/scripting/gdscript/gdscript_styleguide.rst
@@ -1,0 +1,180 @@
+.. _doc_gdscript_styleguide:
+
+GDScript Style Guide
+====================
+
+Description
+-----------
+
+This styleguide lists conventions to write elegant GDScript. The goal is
+to encourage writing clean, readable code and promote consistency across
+projects, discussions, and tutorials. Hopefully, this will also
+encourage development of auto-formatting tools.
+
+Since GDScript is close to Python, this guide is inspired by Python's
+`PEP 8 <https://www.python.org/dev/peps/pep-0008/>`__ programming
+styleguide.
+
+.. note:: Godot's built-in script editor uses a lot of these conventions
+          by default. Let it help you.
+
+Code Structure
+--------------
+
+Indentation
+~~~~~~~~~~~
+
+Indent type: Tabs *(editor default)*
+
+Indent size: 4 *(editor default)*
+
+Each indent level should be one greater than the block containing it.
+
+**Good**
+
+::
+
+    for i in range(10):
+        print("hello")
+
+**Bad**
+
+::
+    
+    for i in range(10):
+      print("hello")
+      
+    for i in range(10):
+            print("hello")
+
+Use 2 indent levels to distinguish continuation lines from
+regular code blocks.
+
+**Good**
+
+::
+
+    effect.interpolate_property(sprite, 'transform/scale',
+                sprite.get_scale(), Vector2(2.0, 2.0), 0.3,
+                Tween.TRANS_QUAD, Tween.EASE_OUT)
+
+**Bad**
+
+::
+
+    effect.interpolate_property(sprite, 'transform/scale',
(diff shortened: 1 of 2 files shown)


learning/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -0,0 +1,187 @@
...
+This styleguide lists conventions to write elegant GDScript. The goal is
+to encourage writing clean, readable code and promote consistency across
+projects, discussions, and tutorials. Hopefully, this will also
+encourage development of auto-formatting tools.
+
+Since GDScript is close to Python, this guide is inspired by Python's
+`PEP 8 <https://www.python.org/dev/peps/pep-0008/>`__ programming
+styleguide.
+
+*Note:* Godot's built-in script editor uses a lot of these conventions
```
NathanLovato: Use the docs' note block
`.. note::`
If you keep it with multiple lines you'll have to create a block with 4 spaces as indent
```
.. note::

    Godot's built-in script editor uses a lot of these conventions
    by default. Let it help you.
```

NathanLovato: Aside from that I think it's good

cbscribe: Good one - I did that in my tutorial, but hadn't gone back and added them to this one

learning/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -0,0 +1,187 @@
...
+**Bad**
+
+::
+    for i in range(10):
+      print("hello")
+      
+    for i in range(10):
+            print("hello")
+
+Use at least 2 indent levels to distinguish continuation lines from
```
akien-mga: I would remove "at least".

learning/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -0,0 +1,187 @@
...
+        Tween.TRANS_QUAD, Tween.EASE_OUT)
+
+Blank lines
+~~~~~~~~~~~
+
+Surround functions and class definitions by a blank line.
+
+Use one blank line inside functions to separate logical sections.
+
+One Statement per line
```
akien-mga: Inconsistent capitalization ;) ("Statement" and "line")

learning/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -0,0 +1,187 @@
...
+
+**Never!**
+
+::
+
+    x        = 100
+    y        = 100
+    velocity = 500
+
+**Exceptions**
```
akien-mga: Not sure about that one. "When using multiple operators in an expression" is not clear. Does that mean that we should write `c = a * a` but `c = a*a + b*b`? That's not consistent.

It should be either space around all operators, or around all operators but `*` and `/`. It's up for discussion, but I guess the former is easier to explain and enforce (and still looks fine).

I tended to recommend the latter (no space around `*`), but consider something like: `position += delta*direction*speed`. That's not readable. `position += delta * direction * speed` is much better.

cbscribe: Yeah, this example came from PEP8. I get the intent - the thinking was probably something around order of operations. I can see how `c = a * a + b * b` requires more thought than `c = a*a + b*b`, but it's probably too fuzzy a distinction to put in a style guide. I'll just take it out.

learning/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -0,0 +1,179 @@
...
+**Good**
+
+::
+
+    for i in range(10):
+        print("hello")
+
+**Bad**
+
+::
```
pwab: Missing linebreak so the code block doesn't render

cbscribe: Fixed.

PR #480, opened 2017-09-16
Added templates subfolder instructions for Godot >= 3

kjav:
This commit adds a note for users of Godot 3.0 and above, in which it is necessary to copy the templates into a versioned subfolder of the global templates directory.

It's pretty likely that those compiling their own templates will also be building newer versions of Godot, such as 3.0-alpha, and so I think it is important to mention this in the documentation.


development/compiling/compiling_for_android.rst
```diff
@@ -157,6 +157,13 @@ The templates folder can be located in:
 -  Linux: ``/home/[username]/.godot/templates``
 -  Mac OSX: ``/users/[username]/.godot/templates``
 
+Note: If using Godot version 3 or above, the templates must be put into the
```
Noshyaar: Please use note template. `.. note::`

kjav: Done :)

kjav:
I've changed the notes to use the "note" template.

akien-mga:
More than a note for 3.0+, you should edit the references to the templates folder. This documentation you're editing is the documentation for 3.0 and later already, so you can remove the reference to the previous system and only document how it works from now on.

kjav:
I've updated the docs to only instruct people how to install templates in V3.0 and above. My .vimrc also removed some trailing whitespace which is a plus!

mhilbrunner:
@kjav Thanks for your first contribution. Your changes have conflicts with the current master. Would you be willing to solve the conflicts and squash your commits into one? If you need assistance, feel free to ask in IRC or Discord.

Otherwise, I may try to take some time to cherry-pick your changes and get them into the repo.

mhilbrunner:
Thanks for updating it and thank you for your contribution!

PR #484, opened 2017-09-20
Update exporting_for_uwp.rst

samvila:
if 'publisherName' doesn't use the full sintax: "CN=publisherName" it will give an error. In order to avoid this error this is the right command:

MakeCert /n "CN=publisherName" /r /h 0 /eku "1.3.6.1.5.5.7.3.3,1.3.6.1.4.1.311.10.3.13" /e expirationDate /sv MyKey.pvk MyKey.cer

### DIFF
--- a/learning/workflow/export/exporting_for_uwp.rst
+++ b/learning/workflow/export/exporting_for_uwp.rst
@@ -35,7 +35,7 @@ You can get more detailed instructions from `Microsoft documentation
 
 First, run ``MakeCert`` to create a private key::
 
-    MakeCert /n publisherName /r /h 0 /eku "1.3.6.1.5.5.7.3.3,1.3.6.1.4.1.311.10.3.13" /e expirationDate /sv MyKey.pvk MyKey.cer
+    MakeCert /n "CN=publisherName" /r /h 0 /eku "1.3.6.1.5.5.7.3.3,1.3.6.1.4.1.311.10.3.13" /e expirationDate /sv MyKey.pvk MyKey.cer
 
 Where ``publisherName`` matches the Publisher Name of your package and
 ``expirationDate`` is in the ``mm/dd/yyyy`` format.


vnen:
IMO it's best to explain what `publisherName` is, since it's stated right after that it needs to match the "Publisher" attribute on the package manifest (maybe that part is not clear). Also, you can have other parameters on the Publisher string, like organization name and address. This section is mostly based on the [Microsoft official documentation](https://msdn.microsoft.com/en-us/library/windows/desktop/jj835832(v=vs.85).aspx).

For instance, check the expected format in the Publisher attribute in the "Remarks" section of [the Package Identity documentation](https://docs.microsoft.com/en-us/uwp/schemas/appxpackage/appxmanifestschema/element-identity#remarks). There it shows the right format, it can have more stuff than just the `CN`.

samvila:
Hi @vnen , I'm afraid I have to disagree with your statement. running the line like this is going to throw an error no matter what:

MakeCert /n publisherName /r /h 0 /eku "1.3.6.1.5.5.7.3.3,1.3.6.1.4.1.311.10.3.13" /e expirationDate /sv MyKey.pvk MyKey.cer

If we want to avoid frustration on the users we should provide an example that works perfectly at first without having to do further research. We are not Microsoft and that means we can do our own documentation in a much simplified way and something that just works. Just by adding the correct command doesn't make things worst in my opinion. You can add any other additional information if you want but Godot docs already explain about matching the publisher name, the problem is that the example command is throwing an error which will cause frustration to new users. Why not making things easier if we can?

vnen:
Both `publisherName` and `expirationDate` are variables in this command. So yeah, if you are just copying and pasting, it'll give errors unless you properly replace the variables. My point really is that `CN=publisherName` is not quite a proper solution to this issue. I have two proposals:

1. Make clear that `publisherName` and `expirationDate` are variables, and explain well their formats. Maybe using `$publisherName` instead.
2. Use a complete example instead, and warn the user to change the sample values. For instance: `MakeCert /n "CN=Godot Engine" /r /h 0 /eku "1.3.6.1.5.5.7.3.3,1.3.6.1.4.1.311.10.3.13" /e 
"12/31/2017" /sv MyKey.pvk MyKey.cer`.

In any case, I think we need to add a paragraph before the MakeCert command explaining how to make the Publisher Name.

samvila:
I think the second solution would work best, it's far more clear and it will avoid any human error factors, that said I will keep the explanation about customizing the publisher name and all the rest of the related information.

Keplar1995:
Voting to close this.. not a big issue whatsoever. Downvoted too.

samvila:
I really think not changing this is a mistake, all users trying this method
for the first time will have to face this issue as well as a big amount of
frustration, also they will need to spend time looking for a solution or
having to do some additional research. I always thought that one of the
targets on Godot is to lower the level of frustration as lowest as
possible. For these reasons please consider make the life of Godot users a
little bit easier.

Thanks

akien-mga:
@samvila Nevermind that comment, that was a spam comment unrelated to this issue.

mhilbrunner:
@samvila Thanks for your first contribution!

Would you be willing to add an explanation for the variables and build an example like vnen suggested above? Please use "ExampleGame" or "YourGame" or something along those lines instead of "Godot", though. You can update your PR via `commit --amend` or simply create a new PR and close this one.

If you aren't, I'll try to take some time the next days to adapt your changes and add some of vnen's suggestions. Thanks!

mhilbrunner:
Looked into this to see how we could improve this. However, any improvements would essentially just duplicate the Microsoft documentation.

After further review, the page currently already points to https://msdn.microsoft.com/en-us/library/windows/desktop/jj835832(v=vs.85).aspx, where this is explained in far more detail, as is stated in the instructions.

For this reasons, I'll leave this as is. But thanks for bringing this issue to our attention :)

PR #489, opened 2017-09-24
Minor language fixes to Spatial Material documentation

eluadyl:
Made minor changes to spatial_material.rst to fix some errors in language. Not sure what 'microsurfacetting' is so left it as is. Please review.

### DIFF
--- a/learning/features/3d/spatial_material.rst
+++ b/learning/features/3d/spatial_material.rst
@@ -6,13 +6,13 @@ Spatial Material
 Introduction
 ------------
 
-For Godot 3, instead of following the trend and focusing in shader graphs,
-we put most of the work offering a default material that covers far
-most use cases. This replaces the old "FixedMaterial" in Godot 2.x
+For Godot 3, instead of following the trend and focusing on shader graphs,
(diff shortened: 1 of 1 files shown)


cbscribe:
Since the only google hit for "microsurfacetting" is that doc, I suspect it's not a word! :)
I suspect they meant "microsurfacing".

eluadyl:
That's what I suspect too but maybe they are coining a new term ;).

cbscribe:
<hint>
You know, the class ref for SpatialMaterial is very bare:
http://docs.godotengine.org/en/latest/classes/class_spatialmaterial.html
</hint>

eluadyl:
Microsufracing (previously 'microsufacetting') is used in the doc to describe a physical property of real world objects, the very small surface features which affect how light reacts with the object's surface. It's not a reference to a function in the SpatialMaterial class. I don't understand how the <hint>...</hint> is supposed to help me.

NathanLovato:
Microsurface is a cg term, but I'd definitely recommend roughness, which is a lot more descriptive and popular. In the meantime, thanks for the fixes! merging now

PR #497, opened 2017-09-30
Renamed fixed_process to physics_process

AndreaCatania:
Related to this godotengine/godot#11659

### DIFF
--- a/learning/step_by_step/scripting_continued.rst
+++ b/learning/step_by_step/scripting_continued.rst
@@ -11,7 +11,7 @@ functions, so there is no need to check for writing code that runs all
 the time. Additionally, a lot can be done with animation players.
 
 However, it is still a very common case to have a script process on every
-frame. There are two types of processing: idle processing and fixed
+frame. There are two types of processing: idle processing and physics
 processing.
 
 Idle processing is activated automatically when the method :ref:`Node._process() <class_Node__process>`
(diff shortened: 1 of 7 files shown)


AndreaCatania:
cc @NathanLovato

learning/step_by_step/scripting_continued.rst
```diff
@@ -29,9 +29,13 @@ frames per second (FPS) of the application:
 The delta parameter describes the time elapsed (in seconds, as
 floating point) since the previous call to "_process()".
 
-Fixed processing is similar, but only needed for synchronization with
-the physics engine. It is called a fixed amount of times per second
-(defined in the Project Settings, but 60 by default).
+Physics processing is similar, but should be used for all the processes that
```
NathanLovato: Reworked, closer to the [writing guidelines](docs.godotengine.org/en/latest/community/contributing/docs_writing_guidelines.html)
`Physics processing is similar, but it should be used for all the processes that must happen before each physics step. For example, to move a character.
It is always runs before a physics step and it is called at fixed time intervals, 60 times per second by default. Change the value in the Project Settings.`

AndreaCatania:
@NathanLovato I did it

NathanLovato:
A few typos remaining, but let's merge it - the article needs editing anyway

NathanLovato:
Thanks for your work!

PR #498, opened 2017-09-30
Fixes to "Your First Game" tutorial.

grking:
A couple of minor corrections and adds note to enable the "Playing" property so mob animations play.


cbscribe:
Thanks for the fixes!

PR #500, opened 2017-10-01
Fixed script error on Line 51

rollergo11:
Going through the step-by-step and found that L51:
    text = (str(accum) # text is a built-in label property
should be 
    text = (str(accum)) # text is a built-in label property.
Missing a parentheses.

### DIFF
--- a/learning/step_by_step/scripting_continued.rst
+++ b/learning/step_by_step/scripting_continued.rst
@@ -48,7 +48,7 @@ with the following script:
 
     func _process(delta):
         accum += delta
-        text = (str(accum) # text is a built-in label property
+        text = (str(accum)) # text is a built-in label property
 
 Which will show a counter increasing each frame.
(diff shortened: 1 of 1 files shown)


learning/step_by_step/scripting_continued.rst
```diff
@@ -48,7 +48,7 @@ with the following script:
 
     func _process(delta):
         accum += delta
-        text = (str(accum) # text is a built-in label property
+        text = (str(accum)) # text is a built-in label property
```
cbscribe: That's too many parentheses. It would look better as 

    text = str(accum)

rollergo11:
Yeah that would be better. Thanks.

NathanLovato:
Hey we can't merge it as is. Can you squash the 2 commits together? It's a bit overkill to remove a single paren.

PR #503, opened 2017-10-02
Updated image of the Viewports docs + source svg

Bonfi96:


toger5:
For the history:
Maybe you could rename the commit message to be more meaningful, of course you uploaded a file, that is correct 😝 but the message should be phrased so that you know what was changed by reading it.

`png for Viewport docs (stretch)`
`svg source for Viewport docs (stretch)`

actually your pr name is really good ;)

also it could be one commit I think?

if you want to make it one commit with a new message:
`git reset HEAD~2` (resets the commits but keeps the changes. the changes are unstaged)
`git commit -a -m "new commit msg"`
`git push -f origin master` (pushes and overrides the current branch. without -f it would give you an error that the remote does not accept the changes since they override something. because the two 
`Add files via upload` commits are missing since they got reseted)

Just another hint. I would recommend using a different branch than master for PR's
so that you can work on multiple things at the same time. It also has the advantage that master (locally) is never ahead of whatever is happening on the github remote repository. So you can always easily pull on master and there would be merge conflicts. than you can rebase all your pr branches onto you local master...

Maybe I did the git stuff too in detail. Don't get offended by it. Even if you knew most of it I decided that it does not hurt to make it more detailed. I always was confused when needing to change git stuff...

Bonfi96:
I guess I need to read a git guide or book 😞 That's the kind of thing that happens when you think you got how git works but you really didn't. Your comment was a gradient from "Yeah, that makes sense..." to "I have no idea what he's talking about".
I forked the repo on github and uploaded the files from the web interface, one by one in different folders, so that resulted in two commits. I'll try to merge the commits, hope I don't blow up something in the meantime. Thanks a lot for your time!

toger5:
@Bonfi96 I'm super sorry that it was my comment was confusing...
why don't you jump on irc and I can explain it in more detail/ guide you trough it ;)

Bonfi96:
Your comment is perfectly fine! I just lack the knowledge to fully understand it. It's going well, if I'm having issues I'll jump on irc.
Thanks again!

toger5:
> I guess I need to read a git guide or book 😞

basically working on godot is much better than a git guide ;) big projects will make you pretty good with it since they actually rely on it. other than a one man project where you commit from time to time just to have the option to roll back ;)

NathanLovato:
You may use `git commit --amend` to modify your commit message. It's the easiest solution. You'll be dropped in Vim, so if you have any trouble drop by on IRC or something

Bonfi96:
@NathanLovato it should be fine now, I merged the two commits and edited the message with the help of @toger5

NathanLovato:
Great job, thanks for your work!

akien-mga:
Note that we have http://docs.godotengine.org/en/latest/community/contributing/pr_workflow.html as a guide to Git (and in particular the workflow we're using). It's not perfect but it should give a good introduction :)

Bonfi96:
Thanks, I already found it the day after, better later than never : ( Though it was nice to learn having a "real repo" scenario :D

PR #504, opened 2017-10-03
Reduced img folder size by 60% with pngquant

NathanLovato:
Following on a discussion with @akien-mga, here's the img files batch-optimized. From now on, we should advise contributors to use at least png-out or some lossy png optim library, so we don't have to overwrite them later.

- Saved 70-80% on all pngs
- Reduced robi .gif title screen animation by >90%
- In total, reduced the img size folder by >60%


toger5:
It leads to a bunch of artefacts. here are screenshots of some of them:

screen shot 2017-10-03 at 14 37 27
screen shot 2017-10-03 at 14 38 00
screen shot 2017-10-03 at 14 38 37

Are we reaching storage limits for hosting?
Or is it just for a little bit faster image loading for the ppl reading the docs?

NathanLovato:
The compression can be a bit heavy at times. I don't want to manually compress all 530 pics 1 by 1 though - the batch compression job alone took a good 4 hours already.
Contributors should really optimize their pictures for the web. A single 200kb pic alone is 6 seconds of extra loading time on a 3g connection - when you have a page with 10+ true color pngs, it can take a while to load and eat up your mobile data.

toger5:
I'm not sure how many ppl will use the docs on mobile data?
and if so there is also the option to download.
I'm all for optimising, but I think even more important is the quality of the content.

NathanLovato:
Mh, I went through the files quickly, and it looks like a bug - artifacts may show up on 1/100 pics, but the compression should barely be noticeable. And the majority's messed up. Gonna redo it. Thanks for the heads up.
Even on public wifi or when many use the same router (at school, in a shared house...), it makes a big difference.

toger5:
@NathanLovato yea you are right reading the docs on abus ride might actually be quite common practise...

NathanLovato:
Fixed! I went through the pics and now they're correct. Found out the issue: ImageOptimizer runs several libs one after the other to get the lowest size possible. If you set it to lossy, it runs 5-6 passes of lossy compression, leading to these artifacts.

I found out about the image size because at the school I was teaching last week + the accommodation I stayed at the internet was slow - down to a crawl at times in the apartment ;)

Calinou:
This is a really good writeup on image compression, put together by Google: https://images.guide/

TL;DR:

- Use [guetzli](https://github.com/google/guetzli) to optimize JPEG images. It is *very* slow (usually several minutes per image, if not more), but provides excellent results. This makes it suited for static content.
- Use [pngquant](https://pngquant.org/) to optimize PNGs while preserving their quality. Alternatively, use [oxipng](https://github.com/shssoichiro/oxipng), I've had good results with that one (it's multi-threaded, so it is quite fast).
- Use [giflossy](https://github.com/pornel/giflossy) to optimize GIFs. However, usage of videos (MP4 + WebM) should be investigated, as videos are much more efficient than animated GIFs.

PR #507, opened 2017-10-07
Fix missing examples and remove broken highlighting in docs guidelines.

Nallebeorn:
Also use [code][/code] for variables etc. in examples, and list the "8 banned words" in the summary.

### DIFF
--- a/community/contributing/docs_writing_guidelines.rst
+++ b/community/contributing/docs_writing_guidelines.rst
@@ -15,7 +15,7 @@ In summary, always try to:
 2. Use precise action verbs
 3. Avoid verbs that end in -ing
 4. Remove unnecessary adverbs and adjectives.
-5. Ban these 8 words
+5. Ban these 8 words: obvious, simple, basic, easy, actual, just, clear, and however
 6. Use explicit references
 7. Use 's to show possession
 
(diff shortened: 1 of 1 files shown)


NathanLovato:
Thanks! Need someone who has more experience with rst to review - it's weird the highlight doesn't work.

Nallebeorn:
That would be a good idea. The problem, as I understand it, was that the highlight directive is meant to be it's own command (which affects all following code examples in the file), but it was apparently used as a parameter for the code snippets. The result being that the code snippets ended up not being code snippets at all.

I'm not sure why md highlighting was used in the first place, though. I mean, the examples aren't markdown at all.

NathanLovato:
> I'm not sure why md highlighting was used in the first place

To get highlights for \*\*ClassNames\*\*. I don't know RST too well, that's it. And the initial PR passed reviews without too much trouble.

NathanLovato:
Okay let's merge this, seems good to me.

akien-mga:
The `.. highlight::` directive is not to start a code block (it's not like <code>```bash</code> for example), it defines the highlight style of all code blocks below it until the next `.. highlight::` directive, if any.

The default highlight for all the docs is gdscript, so you only need to specify something different when it's not gdscript code.

PR #520, opened 2017-10-15
Update simple_2d_game.rst

owengalenjones:
Without setting the Stretch Mode the game window was resulting in (1280, 800) on my MBP not sure if this is hardware specific or not.  This results in the game being pretty broken and I'm sure would be confusing to new users.

gd

You can see the window extends for many pixels after the divider Sprite ends.

Through Googling I was able to find the Stretch Mode setting and found that it _mostly_ fixed the issue, there is still about 100px height that is "outside" the game window where the ball will continue  before bouncing.

Not sure if this is the correct fix, just what worked for me.

### DIFF
--- a/learning/step_by_step/simple_2d_game.rst
+++ b/learning/step_by_step/simple_2d_game.rst
@@ -25,7 +25,8 @@ Scene setup
 To pay homage to the olden times, the game will be in 640x400 pixels
 resolution. This can be configured in the Project Settings (see
 :ref:`doc_scenes_and_nodes-configuring_the_project`) under the Scene/Project
-settings menu. The default background color should be set to black:
+settings menu. Set Stretch Mode in Display to 2d.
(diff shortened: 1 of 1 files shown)


NathanLovato:
Hey thanks for the fix, but this page has phased out of the docs. It's been replaced with @cbscribe's [Your first game](http://docs.godotengine.org/en/latest/learning/step_by_step/your_first_game.html) tutorial which is a lot larger. Closing for now.

pwab:
Shouldn't this fix be considered for [2.1 branch](https://github.com/godotengine/godot-docs/tree/2.1)? Or will the _Your first game_ tutorial be backported someday?

I just think that as long as the 2.1.x docs are online they should be maintained.

NathanLovato:
Sure, but then please open a PR for the 2.1 branch - right now it's on master.

PR #521, opened 2017-10-15
Update two outdated statements in the modules and plugins docs

NullConstant:
This specifically fixes two things:

- Removes the notice in `development/plugins/making_plugins.rst` that the tutorial in question only applies to 2.1 (which has since been released, but the notice was never removed).

- Updates a paragraph in `development/cpp/custom_modules_in_cpp.rst` mentioning that only 2 modules exist by default - way more exist nowadays than that, so it felt appropriate to reflect that.

(Not sure if this should be split/squashed...)


NathanLovato:
Hey there! Thanks for the contribution.

PR #534, opened 2017-10-16
Tip about using constants/enums outside the class

zinkkrysty:
As a beginner, I was confused about the fact that constants and enums cannot be accessed directly on the class, and you need to access them through an instance. I added an example to docs. Feel free to suggest another place to put this, however I think it is quite important for beginners to build the right habit for this.

Since this is my first contribution 🤞 , let me know if I'm doing something wrong.

### DIFF
--- a/learning/scripting/gdscript/gdscript_basics.rst
+++ b/learning/scripting/gdscript/gdscript_basics.rst
@@ -548,6 +548,8 @@ constant dictionary of that name.
     const STATE_JUMP = 5
     const STATE_SHOOT = 6
     const State = {STATE_IDLE = 0, STATE_JUMP = 5, STATE_SHOOT = 6}
+    
+**Note**: Since GDScript does not have class variables, to access constans or enums (defined for example with ``enum State {STATE_IDLE, STATE_JUMP = 5, STATE_SHOOT}``) from another class/script you need to access them through the instance: e.g. ``MyScript.instance().State.STATE_SHOOT``
 
 
 Functions


akien-mga:
Adding a mention to this detail in the docs is fine, but this sentence is a bit misleading and encourages a bad practice.

It's misleading because GDScript does have class variables (whatever that means, but classes do have member variables), and constants from another class (subclass in the same script, or instanced class) can be accessed, e.g. https://github.com/KOBUGE-Games/project-kumquat/blob/master/levels/level.gd#L61

The difference is that in the case you are trying to document, you are trying to access data from `MyScript` - you don't define it so it's a bad example for beginners, but as an experienced user I can imagine that it's defined as `const MyScript = preload("res://myscript.gd")`. What you need to describe is therefore not how to access members from a class, but that the result of that preload is not a class but a preloaded Resource, and that this resource needs to be instanced to be used (including accessing its packed data).

There's some documentation about this here: http://docs.godotengine.org/en/stable/learning/scripting/gdscript/gdscript_basics.html?highlight=preload
So that's likely where a note about accessing static members from the resource should be added.

Then, encouraging people to instance resources just to access a constant is a bad practice. Most of the time you will be accessing an already instanced node which uses this script (e.g. `get_node("Player").State.STATE_SHOOT`).

akien-mga:
But thanks for your contribution, you're raising an important point about something that needs to be documented :) I'll have a look asap to see how to best document it, as it's not a trivial topic.

zinkkrysty:
Yes yes you are totally right! 

And I was using the wrong term: by class variables (habit from Ruby) I meant static members.

So I agree, that what I was looking for, is how to reference static members from the loaded resource.

I have indeed more to learn about this. And as I am learning, it would be really great to have a "best practices" manual with tons of examples to get familiarised with this. Thanks a lot for the explanation and the resources, it's a great help!

akien-mga:
> And I was using the wrong term: by class variables (habit from Ruby) I meant static members.

Note that I'm not sure myself about what the exact term is, I learned most of what I know of programming by doing, so I'm a bit light on the theory :P

PR #536, opened 2017-10-16
Write Code a life bar step-by-step tutorial

NathanLovato:
Contains the tutorial + the demo
Not proofed by a reviewer yet, but if it's okay for you, can we merge so that people can directly edit the file to fix typos?

PR #545, opened 2017-10-19
pressed is replace by is_pressed()

dfleury2:
and may be "ev" can be "event" too, to match the default param name of _gui_input
tapped = !tapped  is more interactive too :)


NathanLovato:
Hey there! The GUI tutorial is out of date and the new one, that completely replaces it, is just waiting for a merge: https://github.com/godotengine/godot-docs/pull/536
Thanks for the help but I won't merge now considering this page is disappearing very soon

PR #552, opened 2017-10-22
Slight doc tweak to Curve2D#interpolate.

nanodeath:
t=1.0 isn't the last vertex -- it's the idx+1 vertex, which I'm calling the second vertex. "next vertex" might also work, but I thought this would be more consistent with "first vertex".

### DIFF
--- a/classes/class_curve2d.rst
+++ b/classes/class_curve2d.rst
@@ -124,7 +124,7 @@ Returns the position of the vertex "idx". If the index is out of bounds, the fun
 
 - :ref:`Vector2<class_vector2>`  **interpolate**  **(** :ref:`int<class_int>` idx, :ref:`float<class_float>` t  **)** const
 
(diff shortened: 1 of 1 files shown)


cbscribe:
The class reference must be edited in the main branch. The .rst files here are automatically generated.

For instructions on how to contribute to the class reference, see:
- http://docs.godotengine.org/en/latest/community/contributing/updating_the_class_reference.html
- http://docs.godotengine.org/en/latest/community/contributing/docs_writing_guidelines.html

nanodeath:
Why even have an "Edit on GitHub" link if the resulting pull request is going to be automatically closed?

cbscribe:
It wasn't automatically closed. I closed it and told you why: these are auto-generated files and your changes would be erased the next time the docs are rebuilt. I'm sure that "Edit on Github" link exists on all Read the Docs pages, that doesn't mean it's always the best way to make changes.

It's only the class reference that this applies to. If you read the contribution guide, you'll see the class reference lives at https://github.com/godotengine/godot/tree/master/doc/classes and is autogenerated XML so that it can be displayed in the web manual as well as the in-editor help UI.

If you're interested in helping out the docs team, we hang out in the Godot Discord's #documentation channel, where you can ask questions if you need any pointers.

nanodeath:
Alright...thanks for the clarification.

PR #553, opened 2017-10-23
Document member and signal link BBCode tags

sheepandshepherd:
Currently the `[method methodname]` tag is documented, but `member` and `signal` can be linked the same way with their own tags, which aren't mentioned in the BBCode-style tag table.

Also replace the method tag example, because it was using a method that has been renamed (`Node2D.set_pos`).

### DIFF
--- a/community/contributing/updating_the_class_reference.rst
+++ b/community/contributing/updating_the_class_reference.rst
@@ -228,9 +228,17 @@ Godot's class reference supports BBcode-like tags. They add nice formatting to t
 +===========================+================================+===================================+============================================+
 | [Class]                   | Link a class                   | Move the [Sprite].                | Move the :ref:`class_sprite`.              |
 +---------------------------+--------------------------------+-----------------------------------+--------------------------------------------+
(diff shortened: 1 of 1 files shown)


community/contributing/updating_the_class_reference.rst
```diff
@@ -228,9 +228,17 @@ Godot's class reference supports BBcode-like tags. They add nice formatting to t
 +===========================+================================+===================================+============================================+
 | [Class]                   | Link a class                   | Move the [Sprite].                | Move the :ref:`class_sprite`.              |
 +---------------------------+--------------------------------+-----------------------------------+--------------------------------------------+
-| [method methodname]       | Link to a method in this class | See [method set_pos].             | See :ref:`set_pos <class_node2d_set_pos>`. |
+| [method methodname]       | Link to a method in this class | Call [method hide].               | See :ref:`hide <class_spatial_hide>`.      |
```
cbscribe: Good job replacing "set_pos"!

Issue #526, opened 2017-10-15
Overview of the Animation Tools


NathanLovato:
Show what's possible with the anim tools. It's not only for visual anim: used to control any property (callbacks to the set() method)
- the basics (keys, AnimPlayer and path to properties, etc.)
- animating colliders
- animating "hidden" properties (inside resources, e.g. shaders)
- function calls
- the api's most useful functions

dploeger:
I'll cover the basics about the AnimationPlayer UI.

NathanLovato:
Edited your comment with the parts you'll write

dploeger:
The draft will live here: https://github.com/dploeger/godot-docs/blob/master/526/features-animation-introduction-2d.md

dploeger:
Reviewed the draft and updated some things based on the doc writer guidelines. Could you please review it now?

NathanLovato:
Reviewing now

dploeger:
Updated the documentation based on your suggestions. Please review again.

NathanLovato:
It's a lot better already! Still some long bullet lists, but as it is it's good to add to the official docs imo - already a lot better than previously available content. Thank you very much for your hard work!

PR #556, opened 2017-10-25
Added introduction to 2d animation #526

dploeger:


dploeger:
@NathanLovato Can't you just [squash them in the merge](https://github.com/blog/2141-squash-your-commits)?

learning/features/animation/introduction_2d.rst
```diff
@@ -328,9 +328,7 @@ Godot's animation engine doesn't stop here. If you're already
 comfortable with Godot's scripting language GDScript and API you know
 that each node type is a class and has a bunch of callable functions.
 
-For example, the
-```SamplePlayer2D`` <http://docs.godotengine.org/en/stable/classes/class_sampleplayer2d.html>`__
-node type has a function to play a sample.
+For example, the ``SamplePlayer2D`` node type has a function to play a sample.
```
akien-mga: It was good to make it a reference to the page, but it should not be hardcoded. Here's how to do it:
```
For example, the :ref:`class_SamplePlayer2D` node type has...
```
This will automatically be converted to a relative link to `/classes/class_sampleplayer2d.html` (without the domain and the language and branch namespaces, which we don't want to hardcode).

akien-mga: The `class_SamplePlayer2D` comes from the identifier at the top of the corresponding page: https://raw.githubusercontent.com/godotengine/godot-docs/2.1/classes/class_sampleplayer2d.rst (without the leading underscore).

akien-mga:
Note that this page covers Godot 2.1, while we want documentation for 3.0 in priority. For example `SamplePlayer2D` doesn't exist in 3.0.

It's still worth merging and then cherry-picking for the 2.1 branch, but then it should be reworked for 3.0 in the master branch.

learning/features/animation/introduction_2d.rst
akien-mga: > <pre>For example, the :ref:`class_SamplePlayer2D` node type has...</pre>

BTW, in the <code>:ref:`some_reference`</code>, the link's caption will be taken from the title of the page. If you need to specify a custom caption (not needed here), you can use <code>:ref:`caption <some_reference>`</code>.

dploeger:
Ah, right. I'll fix that.

dploeger:
I'll update the SamplePlayer2D-example in a second pull request then.

dploeger:
This should work. There was also a reference to AnimationPlayer at the top, which I reworked into a ref-link.

dploeger:
Hmm. Let me close this pull request and make a new one for the 2.1 branch.

PR #557, opened 2017-10-26
Fixed comment about looping over Dictionaries

Petelavi:


akien-mga:
No, that comment is correct.

akien-mga:
And for future contributions: if you delete your repo and GitHub account right after making a PR, it's less likely to be merged as you don't appear as a trustworthy contributor (for such a patch it doesn't matter much, but it's a general rule).

PR #559, opened 2017-10-26
Introduction to 2d animation added. #526

dploeger:


dploeger:
@akien-mga @NathanLovato here's the 2.1-one.

PR #561, opened 2017-10-28
Proofreading: Change most usages of "specially" to "especially"

cooperra:
I know this is a nitpick, but when I was reading the documentation, many usages of "specially" popped out to me as odd. In my opinion, they'd sound better as "especially". Maybe there are others that feel the same way?

### DIFF
--- a/classes/class_animationplayer.rst
+++ b/classes/class_animationplayer.rst
@@ -138,7 +138,7 @@ Member Variables
 Numeric Constants
 -----------------
 
-- **ANIMATION_PROCESS_PHYSICS** = **0** --- Process animation during the physics process. This is specially useful when animating physics bodies.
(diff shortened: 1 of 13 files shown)


classes/class_animationplayer.rst
```diff
@@ -138,7 +138,7 @@ Member Variables
 Numeric Constants
 -----------------
 
-- **ANIMATION_PROCESS_PHYSICS** = **0** --- Process animation during the physics process. This is specially useful when animating physics bodies.
+- **ANIMATION_PROCESS_PHYSICS** = **0** --- Process animation during the physics process. This is especially useful when animating physics bodies.
```
akien-mga: This one needs to modified in the source XML: https://github.com/godotengine/godot/blob/master/doc/classes/AnimationPlayer.xml#L384
We can still merge it, but it will be reverted next time we sync the docs if not fixed upstream.

cooperra: Upstream change merged! :smile:

godotengine/godot#12547

PR #565, opened 2017-10-29
Fixed a typo in line 44

Athomield:
The line was     
platform/x11/  // make that platform/osx/ is you're using OS X
 and changed to     
platform/x11/  // make that platform/osx/   if   you're using OS X


akien-mga:
Note that you don't have to explain the changes in the commit log when they're obvious ;)

Issue #566, opened 2017-11-01
instancing.zip outdated


mattsp1290:
Going through the Step By Step learning guide, I noticed instancing.zip is outdated. It can be found at http://docs.godotengine.org/en/latest/learning/step_by_step/instancing.html

I was able to work around this by exporting to 3.0 from the latest 2.X build. Once in 3.0 I needed to add textures to the sprite nodes in both scenes.

PR #567, opened 2017-11-01
Update instancing.zip to 3.0

mattsp1290:
This is to resolve https://github.com/godotengine/godot-docs/issues/566


NathanLovato:
Tested in a recent build, works like a charm. Thanks!

Issue #184, opened 2016-06-25
Old patch for typos to review


akien-mga:
@brettle had made a patch against the GH wiki (in Markdown) but it was sadly never merged.

Now the docs are in reST so the patch obviously can't be applied, and many of those fixes have probably already been done, but it would be great if someone could check them and see if some are still relevant:
https://gist.github.com/brettle/10007085

(raw patch: https://gist.githubusercontent.com/brettle/10007085/raw/6a4f14d57d34af504a2e0bc4e037a91974c03000/godot.wiki-readability.patch)

(Moved from https://github.com/godotengine/godot/issues/256)

akien-mga:
I went through the first part of that patch up to line 400 (out of 1400), some changes were already applied while others were still relevant, so I tried to apply them in aa520a7.

It's tedious work though, so I'll stop here and let potential contributors finish the job if they're interested :) (if not we'll just close that issue, its improvements will end up in the main tree little by little eventually).

pwab:
I started working on this and filtered the gist with a regex (in case this is of interest: `/(\W[-+].*([A-Za-z])\w+.*)/g`) to just show the lines with + and - so I can search through the docs line by line.
(*This lead to a point where I had to drop some of the code improvements because my regex skills aren't sufficient. But I think code comments/improvements is a broader issue.*)

Pasted the filtered diff into a gist:  https://gist.github.com/pwab/c46b15b4a126731a6530191c39239f7d
(You can see, that there are only about 200+ changes to check.)

Some of the changes were already merged by other commits so the necessary amount of work isn't that high. I will work on this subject later on if I find the time.

pwab:
Well, I just did it and found something useful while jumping through the whole documentation :smile: - but most of the stuff was already changed with other merges.

The suggestions are mainly targeted at the current stable branch, because the step-by-step guide of the latest branch is being worked on. At the end I picked some general suggestions into an extra category.

---

## [A] Suggested changes 

### 1 [step_by_step/scenes_and_nodes](http://docs.godotengine.org/en/stable/learning/step_by_step/scenes_and_nodes.html)

#### 1.1 Has a name...

```patch
-*  Has a name.
-*  Has editable properties.
-*  Can receive a callback to process every frame.
-*  Can be extended (to have more functions).
-*  Can be added other nodes as children.

+*  It has a name.
+*  It has editable properties.
+*  It can receive a callback to process every frame.
+*  It can be extended (to have more functions).
[...]

PR #568, opened 2017-11-01
Fixed several typos

jaggygames:
As suggested in #184


akien-mga:
Thanks. Luckily we can squash commits while merging... :D

PR #569, opened 2017-11-01
Update kinematic_character_2d.rst

Ryozuki:
Latest godot doesn't need set_physics_process(true), it is enabled by default if it finds the callback.

Also the method move has been renamed to move_and_slide or move_and_collide which makes half the tutorial useless.

### DIFF
--- a/learning/features/physics/kinematic_character_2d.rst
+++ b/learning/features/physics/kinematic_character_2d.rst
@@ -44,7 +44,7 @@ Basically, the oldschool way of handling collisions (which is not
 necessarily simpler under the hood, but well hidden and presented as a
 nice and simple API).
 
-Fixed process
+Physics process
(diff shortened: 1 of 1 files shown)


NathanLovato:
@cbscribe is working on a replacement tutorial that will probably go a lot more in depth. This one is likely going to the trash bin by the 3.0 release.

cbscribe:
I wasn't actually redoing this one, but rather the "Physics introduction" doc in the same section. Agreed this one needs a lot of updating as well, and there could be a lot of overlap between the two.

I'm about 90% done, but I've had to re-do some examples a couple of times because of alpha changes. I was waiting for alpha2 to revise once again, so I'll work on that in the next few days.

NathanLovato:
Okay then let's merge for now. Thanks!

PR #572, opened 2017-11-03
Vector intro revisions

cbscribe:
I've rewritten the vector math intro, with the 

- Clean up the grammar
- Remove the "jokes" and non-relevant asides
- Add additional examples with code snippets
- Replaced images (the old ones used right-handed coordinates)
- Move advanced topics (SAT, etc.) to a separate "Advanced Vector Math" document. They don't belong in an introductory text.

Still TODO: Rewrite the advanced & matrices docs.


NathanLovato:
After a quick review, looks good so far. Do you want to do another PR for the advanced part? This way I can merge faster, and we can have other people proof this first part if need be.

cbscribe:
Yes, that's the idea. This PR just moves the advanced content to a separate file.

PR #578, opened 2017-11-06
Fix typos and incorrect or incomplete instructions

mattsp1290:


learning/step_by_step/your_first_game.rst
```diff
@@ -101,10 +101,10 @@ Frames" side of the panel:
 .. image:: img/spriteframes_panel2.png
 
 
-Finally, add a :ref:`CollisionShape2D <class_CollisionShape2D>` as a child 
-of the ``Player``. This will determine the player's "hitbox", or the 
-bounds of its collision area.  For this character, a ``CapsuleShape2D`` 
-gives the best fit, so next to "Shape" in the Inspector, click 
+Finally, add a :ref:`CollisionShape2D <class_CollisionShape2D>` as a child
+of the ``Player``. This will determine the player's "hitbox", or the
+bounds of its collision area. Rename the node to ``Collision``. For this character, a ``CapsuleShape2D``
```
cbscribe: Renaming the collision shape node is not really necessary, since we're not accessing it directly.

cbscribe: This needs to be fixed by changing the screenshot (with the node renamed) to match the text. I will upload a new image. Please remove this change and I can merge the rest of this PR.

mattsp1290: @cbscribe I have made this change. One line 351 of your first game, was that the correct syntax for referencing menus? On 442 of the same file, was that the correct name for that menu? I did not see any hover text for it, but that seemed contextually accurate.

cbscribe: The "Snapping Options" button does have hover text, but it's new since this tutorial was written, so yes your change is now correct.

PR #579, opened 2017-11-07
Update for @gdscript.rst - added wrapi, wrapf functions

Chaosus:


akien-mga:
No need to make PRs for that, the .rst is auto generated and I update it all every now and then.

Chaosus:
Ah, sorry I remember this

akien-mga:
I'm about to push a sync of all classes, so the wrapi/wrapf docs should be up in a minute. Thanks for the poke :)

PR #586, opened 2017-11-13
Fix Step By Step Design UI instructions

mattsp1290:


NathanLovato:
Thanks a lot! Haven't had time to do it since they replaced the anchor menu with Layout.

Issue #447, opened 2017-08-27
Locales page is inconsistent with the engine.


ghost:
(Master branch)

Compare:
https://github.com/godotengine/godot/blob/master/core/translation.cpp
http://docs.godotengine.org/en/latest/learning/features/misc/locales.html

pwab:
Fixed by #590

PR #590, opened 2017-11-17
Updated Locales doc with missing language codes.

Hylek:
Updated the Locale section of documentation with the missing locale codes found in the translation.cpp file of the Godot engine, responding to issue #447.


JuanNeedsToApologize:
Not an issue or high priority at all. Juan told me on reddit the focus should be on 3d only.

akien-mga:
Oops sorry for the huge delay. It looks great :)

PR #595, opened 2017-11-21
Edited vector_math.rst (added clarifications)

amirea:
commented some code and clarifying a few things

### DIFF
--- a/learning/features/math/vector_math.rst
+++ b/learning/features/math/vector_math.rst
@@ -65,7 +65,7 @@ corner of the screen, so to place a 2D node named ``Node2D`` 400 pixels to the r
 
 ::
 
-    $Node2D.position = Vector2(400, 300)  # "$" fetches the node by name from the tree
+    $Node2D.position = Vector2(400, 300)
 
 Godot supports both :ref:`Vector2 <class_Vector2>` and
 :ref:`Vector3 <class_Vector3>` for 2D and 3D usage respectively. The same
@@ -123,7 +123,7 @@ Let's look at two common uses for vector addition and subtraction.
 
 - Movement
 
-A vector can represent **any** quantity with a magnitude and direction. Typical examples are: position, direction, velocity, acceleration, force etc. In
+A vector can represent **any** quantity with a magnitude and direction. Typical examples are: position, velocity, acceleration and force. In
 this image, the spaceship at step 1 has a position vector of ``(1,3)`` and
 a velocity vector of ``(2,1)``. The velocity vector represents how far the
 ship moves each step. We can find the position for step 2 by adding
@@ -198,7 +198,7 @@ to handle this. Here is a GDScript example of the diagram above using a
 
 ::
 
-    var collision = move_and_collide(velocity * delta)  # object "collision" retains information about the collision
+    var collision = move_and_collide(velocity * delta)  # object "collision" contains information about the collision
     if collision:
         var reflect = collision.remainder.bounce(collision.normal)
         velocity = velocity.bounce(collision.normal)
@@ -243,7 +243,7 @@ Facing
 
 We can use this fact to detect whether an object is facing toward another
 object. In the diagram below, the player ``P`` is trying to avoid the
-zombies ``A`` and ``B``. Assuming zombies have a field of view of **180°**, can they see the player?
+zombies ``A`` and ``B``. Assuming a zombie's field of view is **180°**, can they see the player?
 
 .. image:: img/vector_facing2.png
 
--- a/learning/features/math/vector_math.rst
+++ b/learning/features/math/vector_math.rst
@@ -123,7 +123,7 @@ Let's look at two common uses for vector addition and subtraction.
 
 - Movement
 
-A vector can represent **any** quantity with a magnitude and direction. Typical examples are: position, velocity, acceleration and force. In
+A vector can represent **any** quantity with a magnitude and direction. Typical examples are: position, velocity, acceleration, and force. In
 this image, the spaceship at step 1 has a position vector of ``(1,3)`` and
 a velocity vector of ``(2,1)``. The velocity vector represents how far the
 ship moves each step. We can find the position for step 2 by adding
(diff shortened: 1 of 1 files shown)


learning/features/math/vector_math.rst
```diff
@@ -123,7 +123,7 @@ Let's look at two common uses for vector addition and subtraction.
 
 - Movement
 
-A vector can represent **any** quantity with a magnitude and direction. In
+A vector can represent **any** quantity with a magnitude and direction. Typical examples are: position, direction, velocity, acceleration, force etc. In
```
her001: I agree that adding examples here is helpful, but I think that the wording can be improved. Having direction listed as an example when we just defined a vector as "magnitude and direction" may be confusing. In addition, I think that the "etc." at the end is not necessary, as this has already been established to be a limited list of examples.

I would probably write it like so: "Typical examples are position, velocity, acceleration, and force."

her001: > ...I wanted to suggest some kind of a comprehensive use case...

I don't agree that the list should try to be comprehensive in any way, it just needs to give an idea of what they are and how they are useful. What you have does that pretty well in my opinion, it just needs to be cleaned up a bit.

> Would it help if we say "Typical use cases..."?

I would say that changing the beginning to "Typical use cases include..." can work too. 

> Or maybe move the list in another part of the document?

I think the placement in the doc is appropriate as it is.

learning/features/math/vector_math.rst
```diff
@@ -198,7 +198,7 @@ to handle this. Here is a GDScript example of the diagram above using a
 
 ::
 
-    var collision = move_and_collide(velocity * delta)
+    var collision = move_and_collide(velocity * delta)  # object "collision" retains information about the collision
```
cbscribe: I would say "contains" instead of "retains"

learning/features/math/vector_math.rst
```diff
@@ -243,7 +243,7 @@ Facing
 
 We can use this fact to detect whether an object is facing toward another
 object. In the diagram below, the player ``P`` is trying to avoid the
-zombies ``A`` and ``B``. Can the zombies see the player?
+zombies ``A`` and ``B``. Assuming zombies have a field of view of **180°**, can they see the player?
```
cbscribe: "have a field of view of" sounds awkward. I would instead use "Assuming a zombie's field of view is **180°**, can it see the player?"

learning/features/math/vector_math.rst
cbscribe: Agreed, there are too many use cases to be comprehensive. Remember the audience here. This doc typically isn't going to be read by users as their entry point to Godot. They're going to wind up here when they hit their first Vector2 and don't know what that is.

learning/features/math/vector_math.rst
```diff
@@ -60,12 +60,12 @@ Vector Operations
...
 refer to a vector, but for convenience programmers typically use the
 coordinate notation. For example, in Godot the origin is the top-left
-corner of the screen, so to place a 2D node 400 pixels to the right and
+corner of the screen, so to place a 2D node named ``Node2D`` 400 pixels to the right and
 300 pixels down, use the following code:
 
 ::
 
-    $Node2D.position = Vector2(400, 300)
+    $Node2D.position = Vector2(400, 300)  # "$" fetches the node by name from the tree
```
cbscribe: I don't think this is especially helpful, especially after your addition above regarding the node name. If you feel like you need to explain "$" because you're thinking of an absolute beginner, then you can't use "tree" without explaining it either.

amirea: I searched for the symbol `$` in the documentation within Godot and couldn't find it. I actually found it on the internet in an answer. That's why I thought I'd mention what it does.

As far as I'm concerned, I already know about it so I'm OK with what you suggest.

amirea:
OK guys, whenever you're ready :)

learning/features/math/vector_math.rst
cbscribe: The GDScript intro doc still has to be updated. Now that we're in beta, it will be a good time because there aren't likely to be any more changes to basic GDScript.

amirea:
I've also tried my best with https://github.com/godotengine/godot-docs/blob/master/learning/features/math/matrices_and_transforms.rst but I don't even understand more than half of it despite my matrices knowledge.
I don't know where I can raise this issue. I'm positive that if *I* don't understand it, beginners definitely won't.

cbscribe:
@amirea you should make the suggested changes and amend your commit. That way the commit log will still credit you as the author.

And yes, the matrix doc is in need of a rewrite as well. FYI, https://github.com/godotengine/godot-docs/issues is the ideal place to make suggestions and point out things like that. Before you post there, make sure you look through the existing issues to see if you're duplicating one that's already there.

learning/features/math/vector_math.rst
amirea: Can I use "Assuming a zombie's field of view is 180°, can they see the player?"

cbscribe: I'm not sure what the preferred pronoun is for a zombie, so I guess so. :)

amirea: there's an image with 2 zombies, out of which only one has the player in its FOV.

cbscribe: Yes, but the question isn't applying to the group, it's applying to each individual zombie. How about "Assuming each zombie's..." ?

amirea:
Done correcting. I've re-commited on my master. I'm not sure if this alone notifies you guys or I have to let you know in here.

learning/features/math/vector_math.rst
```diff
@@ -123,7 +123,7 @@ Let's look at two common uses for vector addition and subtraction.
 
 - Movement
 
-A vector can represent **any** quantity with a magnitude and direction. Typical examples are: position, direction, velocity, acceleration, force etc. In
+A vector can represent **any** quantity with a magnitude and direction. Typical examples are: position, velocity, acceleration and force. In
```
cbscribe: Missing an Oxford comma here. Our doc standards are to use it: "acceleration, and force."

Issue #531, opened 2017-10-15
Revise Scripting Introduction


cbscribe:
- Update text and descriptions
- Use tabs to show code examples in GDScript, VS, C#

mhilbrunner:
Seems this was done :)

PR #597, opened 2017-11-22
Add C# tabs to the first scripting tutorial page.

PJB3005:
Here's a screenshot (slightly out of date):

By the way, the lack of VS tabs isn't an issue. The tab system seems to handle it well.

### DIFF
--- a/learning/step_by_step/scripting.rst
+++ b/learning/step_by_step/scripting.rst
@@ -262,7 +262,9 @@ The final script should look basically like this:
 
     using Godot;
 
-    public class SayHello : Panel
+    // IMPORTANT: the name of the class MUST match the filename exactly.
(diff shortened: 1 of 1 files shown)


reduz:
this is great

cbscribe:
Agreed, this is great! However, the GDScript snippets are no longer good examples for idiomatic GDScript in 3.0. For example, we should be preferring:

    $Label.text = "HELLO!"

over

    get_node("Label").text="HELLO!"

I'm working on a rewrite of this doc to update to 3.0, but I think we can go ahead and merge your changes, and I'll update from there.

cbscribe:
I'm starting on the rewrite of this doc, so I'm going to merge this and use it as the basis for the new version. See #531 for status/comments

PR #608, opened 2017-11-25
Replace "guards" with "enemies" for consistency

davidpol:
The Groups section shows how to create an "enemies" group but later on uses "guards" as the group's name.


JuanNeedsToApologize:
Not a high priority at all. The focus should be mainly on 3d. Monster AI is not a high priority, voting to close this.

MoraleMuffin5:
Damn, $3,000/m is still not enough to merge a PR. Some really good leadership in Godot

PR #618, opened 2017-11-30
Clean up 93d3009

capnm:
References do not work in the code blocks,
a simple paragraph is much easier to read anyway.


capnm:
ping @cbscribe

cbscribe:
Good catch - thanks!

PR #900, opened 2017-12-13
Asset workflow/Import process: Mention MD5 checksum for automatic rei…

pschnorbus:
…orts are performed only if the MD5 checksum has changed.

This avoids confusion when someone tries to trigger a reimport by touching the resource file or exports a resource file with unchanged contents to the project.


NathanLovato:
Thanks!

Issue #902, opened 2017-12-14
Singletons (AutoLoad) example project outdated


johannesgunnar:
Over at [Singletons (Autoload) for Godot 3](http://docs.godotengine.org/en/latest/learning/step_by_step/singletons_autoload.html?highlight=singletons%20autoload) the [example download](http://docs.godotengine.org/en/latest/_downloads/autoload.zip) does not follow the example code and instructions in the documentation. The example project is still Godot 2 based, while the documentation has been updated for Godot 3.

Problems

- project example uses .scn files while the example code on the page talks about .tscn.
- project example does not have project.godot file / project example has not been converted to Godot 3

See the linked [QA thread](https://godotengine.org/qa/21209/singletons-autoload-documentation-incorrect-or-outdated?show=21209#q21209) for more information.

mhilbrunner:
Fixed by Supersed by https://github.com/godotengine/godot-docs/pull/1481

PR #911, opened 2017-12-19
Incorrect Code Sample Scene Extension

Ceryni:
The scene extension of the [example scenes](http://docs.godotengine.org/en/latest/_downloads/autoload.zip) were created using '.scn' instead of '.tscn'. The sample code incorrectly referred to the '.tscn' extension resulting in a crash.


Ceryni:
Apologies, I just looked around and found Issue #902. I will close this PR in favor of a quick redo of the whole thing in Godot 3.0 Beta1.

PR #922, opened 2017-12-26
Update ui_main_menu.rst

Mikerhinos:
Updated link to point to the resolution managing / stretching page, was pointing nowhere.

### DIFF
--- a/learning/step_by_step/ui_main_menu.rst
+++ b/learning/step_by_step/ui_main_menu.rst
@@ -294,5 +294,4 @@ You can change the scale mode in the project settings: click the Project menu ->
 Settings. In the window's left column, look for the Display category.
 Click on the Window sub-category. On the right side of the window,
 you'll find a Stretch section. The three settings, Mode, Aspect, and
-Shrink, control the screen size. For more information, see `Project
-Settings - Multiple resolutions <http://docs.godotengine.org/en/latest/learning/features/viewports/multiple_resolutions.html>`__.
+Shrink, control the screen size. For more information, see `:ref:'Project Settings<doc_multiple_resolutions>'`__.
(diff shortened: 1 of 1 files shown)


skyace65:
When linking to other pages in the docs you shouldn't be using a direct link, doing this can lead to problems in the  future. For example, when 3.0 becomes stable and work begins on 3.1 documentation, that page link will send you to the 3.1 page even if you were on 3.0. The proper syntax to use is a ref, here's an example from the Unity to Godot page

:ref:`this page <doc_scripting_continued>`

In this code, clicking the words "this page" sends you to the scripting continued page of the documentation. the code that goes inside the <> can be found at the top of the page you're trying to link to, in the raw file. On the doc scripting page, at the very top of the file is this

.. _doc_scripting_continued:

just ignore the dots and the first underscore, and you get what goes inside the <>. So in your code you want to go to the multiple resolutions page, at the top of that file is 

.. _doc_multiple_resolutions:

so your code should look like this

:ref:'Project Settings<doc_multiple_resolutions>'

skyace65:
When linking to other pages in the docs you shouldn't be using a direct link, doing this can lead to problems in the  future. For example, when 3.0 becomes stable and work begins on 3.1 documentation, that page link will send you to the 3.1 page even if you were on 3.0. The proper syntax to use is a ref, here's an example from the Unity to Godot page

`:ref:'this page <doc_scripting_continued>' `

Use a backtick instead of single quotes. I'm using single quotes in here because backtick is how github reads code examples

In this code, clicking the words "this page" sends you to the scripting continued page of the documentation. the code that goes inside the <> can be found at the top of the page you're trying to link to, in the raw file. On the doc scripting page, at the very top of the file is this

`.. _doc_scripting_continued:`

just ignore the dots and the first underscore, and you get what goes inside the <>. So in your code you want to go to the multiple resolutions page, at the top of that file is 

`.. _doc_multiple_resolutions:`

so your code should look like this

`:ref:'Project Settings<doc_multiple_resolutions>'`

Mikerhinos:
Got it thanks for the information, I'm still in the 90's :nerd_face:

NathanLovato:
Thanks!

PR #928, opened 2017-12-28
Fixes a bug in the example code in object type. Ref<MyReference> myre…

### LINKED ISSUE #506: Ref<MyReference> example does not compile
http://docs.godotengine.org/en/latest/development/cpp/object_class.html#id2

The error below is basically

Ref<MyReference> myref = memnew(MyReference);
```


modules/summator/summator.cpp:44:17: error: use of undeclared identifier 'call_func'
    this->_cb = call_func(args, 1, err);
                ^
modules/summator/summator.cpp:62:15: error: use of overloaded operator '=' is ambiguous (with operand types 'Ref<FuncRef>' and 'FuncRef *')
    this->_cb = memnew(FuncRef);
    ~~~~~~~~~ ^ ~~~~~~~~~~~~~~~
core/reference.h:144:7: note: candidate function
        void operator=(const Ref &p_from) {
             ^
core/reference.h:177:7: note: candidate function
        void operator=(const Variant &p_variant) {
```

I believe the proper way to use reference is
```
Ref<MyReference> myref = Ref<MyReference>(memnew(MyReference));
```

hungrymonkey:
…f = Ref<MyReference>(memnew(MyReference));

The statement is a compile error which emits  use of overloaded operator '=' is ambiguous (with operand types 'Ref<FuncRef>' and 'FuncRef *')

This commit uses neikeq suggestion of Ref<MyReference> myref(memnew(MyReference));

Fixes #506

### DIFF
--- a/development/cpp/object_class.rst
+++ b/development/cpp/object_class.rst
@@ -242,7 +242,7 @@ Declaring them must be done using Ref<> template. For example:
         GDCLASS(MyReference, Reference);
     };
 
-    Ref<MyReference> myref = memnew(MyReference);
+    Ref<MyReference> myref(memnew(MyReference));
 
 ``myref`` is reference counted. It will be freed when no more Ref<>
 templates point to it.


hungrymonkey:
Should I also change the stable version too?

Issue #923, opened 2017-12-27
Why isn't `COLOR` both in and out in canvas_item fragment shaders?


Zylann:
Godot 3.0

I noticed the doc says `COLOR` is an `in` parameter of fragment shaders. However the only way I could output pixel color in such a shader is to **assign** a value to `COLOR`.
`COLOR` should be marked as `in and out`, or a separate variable has to be created.

There is also no example of a `canvas_item` fragment shader (only `spatial`), so you can check this in the latest 2D shaders demo projects.

mhilbrunner:
Fixed in master, COLOR is markes as **out** for CanvasItem fragment shaders, and marked as **in** for Spatial fragment shaders, where ALBEDO is the color output (marked as **out**).

PR #930, opened 2017-12-29
Fix shading language variable

Peter-Lavigne:
After some testing, it looks like the COLOR variable in canvas_item fragment shaders is write-only, not read only. I'm able to set the COLOR variable using a vec4, and if I sample from the texture, can modify the color of a pixel. Running the code 'COLOR = COLOR;', along with any other modification of COLOR, makes the image entirely white.

Is there any way I could verify that the docs for shader variables are correct? I'd like to learn the shading language but have had a tough time writing basic shaders so far.


mhilbrunner:
Relevant: #923

Have to look into this some more

Issue #63, opened 2016-03-05
Rename "GDScript more efficiently"


Griatch:
The tutorial [GDScript more efficiently](https://github.com/godotengine/godot-docs/blob/master/reference/gdscript_more_efficiently.rst) strikes me as misleadingly named. The title (and introduction) suggests that it's a guide to optimizing GDScript code (which would be interesting and give more insight into the language's underpinnings). But apart from some examples of GDScript usage that seems to have little to do with efficient use, there is no real discussion of performance from the perspective of a GDScript user. Rather, this tutorial is squarely aimed at exemplifying GDScript code to developers coming from statically typed languages like C or Java. 

So I suggest renaming this tutorial to something more informative like "_GDScript compared to static languages_".

akien-mga:
I wouldn't make it a "compared to static languages" page. From a quick glance, it does give example of how similar stuff is done in static languages, but the purpose is more to show how to use dynamic languages in the most efficient way (not as performance-wise, but for code quality).

I'm not against changing the name though if we come up with something that describes it better.

Griatch:
@akien-mga 

It's certainly dominated by a side-by-side comparison between static languages and GDScript, but there are also a few mention of specific dynamic features like duck typing. So how about 
"*An introduction to dynamic languages*" ?

noidexe:
What about "Writing idiomatic GDScript"?

NathanLovato:
Fixed!

PR #949, opened 2018-01-05
Change ambiguous section titles

### LINKED ISSUE #933: The "Developing in C++" section name is ambiguous/confusing
This came up on the Godot Discord server the other day.

obraz

Basically, it's not immediately clear that the section refers to extending the engine itself (with C++), as opposed to using C++ for scripting (which is a separate matter), so I think it'd be appropriate to alter the name to more accurately represent what the section is meant to cover.

NullConstant:
Changes "Developing in C++" to "Engine development", and "GDScript more efficiently" to "GDScript: An introduction to dynamic languages".

Admittedly, the latter is a bit wordy (and thus open to suggestions), but it was similarly deemed to be a bit inaccurate/confusing (see #63), but the issue was kind of left in the dust.

Closes #933, #63.

### DIFF
--- a/development/cpp/index.rst
+++ b/development/cpp/index.rst
@@ -1,4 +1,4 @@
-Engine development 
+Engine development
 ==================
 
 .. toctree::
--- a/development/cpp/index.rst
+++ b/development/cpp/index.rst
@@ -1,5 +1,5 @@
-Developing in C++
-=================
+Engine development 
+==================
 
 .. toctree::
    :maxdepth: 1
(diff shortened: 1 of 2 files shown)


development/cpp/index.rst
```diff
@@ -1,5 +1,5 @@
-Developing in C++
-=================
+Engine development
```
akien-mga: There's a trailing space here (not sure that it matters much, but RST is very touchy about formatting).

NathanLovato:
Thanks!

PR #950, opened 2018-01-05
Updated parts of the documentation to use gender neutral pronouns.

FelixMaxwell:
Also corrected a few typos.


akien-mga:
Thanks, looks good.

PR #955, opened 2018-01-05
Update viewports pages images

skyace65:
Updates 3 images in the two viewports pages. Also updated the docs to say where specifically the resolutions settings are in the project settings.


NathanLovato:
Thanks :)

PR #958, opened 2018-01-06
Update documentation for Godot 2

emilhem:
See issue #408 for the issue fixed with this commit.

### DIFF
--- a/learning/scripting/gdscript/gdscript_advanced.rst
+++ b/learning/scripting/gdscript/gdscript_advanced.rst
@@ -310,9 +310,9 @@ states and quick structs:
 
     # indexing
 
-    d.mother = "rebecca" # this doesn't work (use syntax below to add a key:value pair)
+    d.mother = "rebecca" # this works
     d["mother"] = "rebecca" # this works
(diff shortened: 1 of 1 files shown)


learning/scripting/gdscript/gdscript_advanced.rst
```diff
@@ -310,9 +310,9 @@ states and quick structs:
 
     # indexing
 
-    d.mother = "rebecca" # this doesn't work (use syntax below to add a key:value pair)
+    d.mother = "rebecca" # this works
     d["mother"] = "rebecca" # this works
-    d.name = "caroline" # if key exists, assignment does work, this is why it's like a quick struct.
+    d.name = "caroline" # this works, this is why it's like a quick struct.
```
akien-mga: Why keep two examples doing the same thing? The way it was written was like "haha, this doesn't work! but this does!", but if everything works, just show the two ways to index and that's it :)

emilhem: Well, I think that it is beneficial to show that it's the same thing when changing an index as when adding one.

Issue #395, opened 2017-06-18
Add an EditorSettings reference


vnen:
The Project Settings is quite extensive. Having a reference page explaining briefly what each options does would be quite helpful. The same applies to the Editor Settings.

*Bugsquad edit: ProjectSettings now has a reference, but it still has to be added for EditorSettings.*

NathanLovato:
Absolutely. At the same time, they might benefit from some reviewed grouping/reorganization first. In any case, a quick rundown of the most important settings would be helpful, to get started.

akien-mga:
I agree with Nathan, we need to review the list and do necessary renamings/groupings before the 3.0 release (as it will break the support of previous settings).

pwab:
I listed the entries of these two settings dialogs in two seperate gists for myself. Maybe it is helpful here too:
Project Settings: https://gist.github.com/pwab/6ff702daefbc3aab20d673d823e1778d
Editor Settings: https://gist.github.com/pwab/08fae71b19a6118c68eddcaed4dd038f

Some points came to my mind, while doing so:
- The Project Settings uses snake_case (with underscores) and the Editor Settings are written in CamelCase (with whitespaces). This should be unified.
- The only settings-section which uses categories is the Blackberry-section in the Editor Settings. This should be utilised more often. (for example in the Text Editor-section)
- The Settings-Dialogs aren't real windows so it is not possible to resize them, switch them in taskbar/with Alt+Tab or move them to a second monitor. It would be quite comfortable while changing settings, because changes are instantly applied (i.e. Editor colors). I'm sure there must be an issue for that, but I couldn't find one - sorry if I missed it.
- An explanation of each setting entry could be shown at the bottom of the settings dialog (like the description in the new node dialog)

NathanLovato:
@pwab Do you still have the files? The gists aren't available anymore. Also, did you write some script to get the list from the source or what-not? That'd be useful because they're bound to evolve over time. I'm sure we can find contributors interested in working on this reference.

pwab:
Yeah @NathanLovato I still have the files.  I tried to rework the lists into a rst-table structure and started updating the items. But it wasn't ready yet (I'm not sure if I ever touched the v3-project-settings again) so be aware that the content of these files isn't up to date.

You can find the gists here (uploaded them again):
- [godot_2_editor_settings](https://gist.github.com/pwab/f6522fe7c8ba37f4dc0c42cc71ca0ec8)
[...]

PR #965, opened 2018-01-09
added godot3 editor settings table

ZX-WT:
Table for the editor settings, updated from the file in #395. I am surprised that this one has no problem to preview it on github :). Next thing is to do is fill in the blanks.


mhilbrunner:
I'm not sure adding version information at the beginning is a good idea. Looks outdated quickly even if nothing changed, and has to be kept up to date while the information is there on GitHub anyway. IMO, this is what a VCS is for.

Opinions? Thanks for taking the time by the way!

ZX-WT:
I agree with you, I just keep it because it was in the original file, :P,  I will go and delete it. It would be great if other people can fill in the table as well, there are quite a lot of info to input. :)

pwab:
@mhilbrunner 
> I'm not sure adding version information at the beginning is a good idea. Looks outdated quickly even if nothing changed, and has to be kept up to date while the information is there on GitHub anyway. IMO, this is what a VCS is for.

As I was the original author of the file I can tell you why it was there in the first place:
I created a gist for these files (editor and project settings) and wanted to know when I touched the file last time. At that time I didn't know that gists could be updated  😄 

So you are totally right - this should not be included.

mhilbrunner:
Closing this for now, as it was discussed and the only sane way to do this is by automatically parsing the data and adding descriptions to avoid having to keep the list up to date in multiple places; this would be too much of a burden for contributors. Thank you for taking the time and getting the ball rolling on this :)

We will probably use this as inspiration for the script. See discussion here: https://github.com/godotengine/godot-docs/issues/395.

PR #969, opened 2018-01-10
Update Instancing doc and images

cbscribe:
Rewrite of learning/step_by_step/instancing, including:

- Updated images (was using Godot 0.99!)
- Updated project zip
- Text and language cleanup

TODO: I didn't change the scene diagrams at the top, but I think that can wait for the [Scenes and nodes](http://docs.godotengine.org/en/latest/learning/step_by_step/scenes_and_nodes.html)
rewrite so that they'll match. If no one else does that rewrite before I get a chance, I'll change them at the same time.


learning/step_by_step/instancing.rst
```diff
@@ -3,109 +3,113 @@
...
-they can have only one single node as the tree root.
+In :ref:`doc_scenes_and_nodes` you learned that a scene is a collection of
+nodes organized in a tree structure, with a single node as the tree root.
 
 .. image:: img/tree.png
 
-Recall that a scene can be created and saved to disk. You can create and save
-as many scenes as you desire.
+You can create as many scenes as you like and save them to disk. Scenes 
+saved in this manner are called "Packed Scenes" and have a ``.tscn`` filename
```
willnationsdev: Maybe also mention the binary `.scn` filetype here, in parentheses or something?

cbscribe: Maybe, although it may be a bit out of scope for an introductory doc. It might be beter to link to the "Resources" doc which has details on PackedScene and saving/loading from disk.

mhilbrunner:
Nice! Found no issues. :)

PR #973, opened 2018-01-12
[DOCS] Remove semi-obsolete slide code

speakk:
Remove semi-obsolete custom move and slide code, and replace it with the modern in-built move_and_slide function.

### DIFF
--- a/learning/features/physics/kinematic_character_2d.rst
+++ b/learning/features/physics/kinematic_character_2d.rst
@@ -142,7 +142,6 @@ This adds simple walking support by pressing left and right:
 
     const GRAVITY = 200.0
     const WALK_SPEED = 200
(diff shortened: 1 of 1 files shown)


mhilbrunner:
Thanks for pointing this out, looks fine, although maybe we should add a short note explaining:
>const NORMAL_UP = Vector2(0, -1)
What do you think?

akien-mga:
> const NORMAL_UP = Vector2(0, -1)

I wouldn't introduce a constant for that. Just use it directly, but add a comment before `move_and_slide` explaining what the second argument means.

speakk:
Okay, I added a comment instead of a CONST, how about now? Thanks for the suggestions.

mhilbrunner:
Thanks for your contribution!

PR #983, opened 2018-01-17
added extension link for vscode

SShehriyar:
Added a link to the Visual Studio Marketplace Extension for Godot development.

### DIFF
--- a/development/cpp/configuring_an_ide.rst
+++ b/development/cpp/configuring_an_ide.rst
@@ -191,7 +191,13 @@ Test it:
 - It should break at the point!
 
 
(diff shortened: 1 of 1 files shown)


akien-mga:
While I agree that it would be nice to have this documented somewhere, this is not the right place. This page is about configuring an IDE for engine development (in C++), while this plugin is to use VS Code for scripting.

pwab:
I am not sure if this is something related to the docs. How about adding it to [awesome-godot](https://github.com/Calinou/awesome-godot)?

SShehriyar:
Alright I'll add it to another place. Thank you :)

PR #996, opened 2018-01-20
Updating incorrect cross product information.

Omicron666:
Updating incorrect cross product information.
TODO: update inverted picture (as illustrating following code).


cbscribe:
Thanks! Do you want to add the image or should I merge now?

PR #1004, opened 2018-01-21
Make it easier to build pages for Windows users

Xrayez:
Added a batch script to simplify the process of building html pages on
Windows platform. The script was auto-generated by Sphinx.

Allows typing `make html` directly from the command line just like in Linux.

~~Created .gitattributes for the *.bat files to be commited with CRLF in~~
~~order to prevent possible incorrect interpretation.~~

### DIFF
--- a/make.bat
+++ b/make.bat
@@ -1,0 +1,36 @@
+@ECHO OFF
+
+pushd %~dp0
+
+REM Command file for Sphinx documentation on Windows
+
+if "%SPHINXBUILD%" == "" (
+	set SPHINXBUILD=sphinx-build
+)
+set SOURCEDIR=.
+set BUILDDIR=_build
+set SPHINXPROJ=Test
+
+if "%1" == "" goto help
+
+%SPHINXBUILD% >NUL 2>NUL
+if errorlevel 9009 (
+	echo.
+	echo.The 'sphinx-build' command was not found. Make sure you have Sphinx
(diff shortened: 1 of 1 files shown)


akien-mga:
> Created .gitattributes for the *.bat files to be commited with CRLF in
order to prevent possible incorrect interpretation.

Is this really needed? AFAIK Git on Windows is configured by default to autoconvert to CRLF, and I'd prefer not to enforce CRLF files on non-Windows (since Linux and macOS Git are not configured to modify the EOL by default AFAIR).

Xrayez:
> Git on Windows is configured by default to autoconvert to CRLF

What about those who configured Git on Windows to autoconvert to LF by default instead?

Either way, I checked out the `make.bat` file with LF and it doesn't seem to break anything compared to CRLF, not to mention that there might be additional issues on Linux and macOS Git that I wasn't aware of. Should I remove `.gitattributes` then?

akien-mga:
> Either way, I checked out the make.bat file with LF and it doesn't seem to break anything compared to CRLF, not to mention that there might be additional issues on Linux and macOS Git that I wasn't aware of. Should I remove .gitattributes then?

If it still works fine as LF on Windows for those who considered their Git this way, I'd prefer we remove the .gitattributes yeah. No exceptions unless really needed.

Xrayez:
Yep, reasonable enough, updated.

PR #1007, opened 2018-01-22
Update 'Mesh generation with heightmap and shaders' tutorial to 3.0

antoniotorresm:
- Updated tutorial text to make sense with Godot 3.0.
- Replaced old screenshots with 3.0 ones.
- Deleted screenshots that are now unnecessary.
- Updated the shaders code.


adsmalley:
New to Godot and am going through the "Vertex displacement with shaders" tutorial mentioned above, but think I am misunderstanding something in it.   Seems to work fine until I get to the "Noise" section.  then when I cut and paste the sections in, I'm not seeing what is shown (as there are no screen shots for each addition within the Noise section, but at the end I am seeing what appears to be just a flat plane with no variation), but at the end I am seeing a flat plane instead of anything that appears to have variations in it.  

Being new to Godot and this sort of an engine, I'm assuming that I'm doing something wrong.  Do the Noise functions ("float hash(vec2 p)..." and "float noise(vec2 x)..." replace the "void vertex()..." function previously mentioned, or are they nested within or are they pasted onto the end of the "void vertex()..." function?  

Or is it because I am using a much "larger" area (instead of using "Size (2,2)" I used "Size (10,10) and instead of using "Subdivide Width/Depth" of 32 I used 64), so the variations are too small to see?

Just want to better understand what I am doing wrong relative to what the tutorial is trying to show.  Thanks in advance for any guidance that can be provided.

Issue #1011, opened 2018-01-23
Make doc version and link to other doc versions more prominent


mhilbrunner:
The box for switching to other doc versions in the lower left menu is maybe a little to subtle?

See discussion on https://github.com/godotengine/godot/issues/15791

Not sure how much we can influence that or if we can do something about this.

mhilbrunner:
This is old, and we got the large infobox at the top added somewhere along the way :)

PR #1015, opened 2018-01-24
Improve docs structure

NathanLovato:
Change Step by Step in Getting Started
Remove Engine Features category, move into a new Tutorials top-level section

Built and it works locally. I reordered the categories a bit: moved math and more specific things like 'platform-specific' or 'misc' at the bottom of the list, and moved essential themes like 'audio', 'input' up the list. It now starts with 2d and 3d.
There's no change in identifiers so it should be easy to redirect the pages


NathanLovato:
Now it looks like 
image

akien-mga:
Looks fine, but I wouldn't put Math so far down. It *is* a critical topic that users need to get familiar with sooner than later, so I'd put it between Networking and Platform-specific.

NathanLovato:
Moved it back up. If it's okay I'll fix conflicts and we can proceed with the merge and redirect

akien-mga:
It's fine by me. CC @cbscribe @mhilbrunner @StraToN for review.

cbscribe:
* Love "Getting Started"
* In "Your First Game" I refer back to to "Scripting" since it was earlier in the "Step-by-step" progression. Moving "Scripting" out of "Step-by-step" will mean some subtle wording changes. We'll need to keep an eye out for similar situations in other docs.

reduz:
looks fine, but you will have to change all links in class xml
documentation that points to these files

StraToN:
I'm fine with the changes. Green light :)

akien-mga:
@NathanLovato I'll let you update redirects as needed in the RTD backend.

> looks fine, but you will have to change all links in class xml documentation that points to these files

Updated with https://github.com/godotengine/godot/commit/f2e3825616ce67078159adc3d591e4c6e368983a but it would be good for someone to double-check them.

I've just created (again) the `3.0` branch, so we should soon have http://docs.godotengine.org/en/3.0/ online. And as previously, RTD will play clever and redirect to it for the `stable` namespace... but I guess it's almost true now - we just need to do something for #1011 to prevent confusion.

NathanLovato:
Doing the redirect now

PR #1030, opened 2018-01-26
Nintendo Switch, PlayStation 4 and Xbox One Support

eska014:
It's just a syntax fix

### DIFF
--- a/tutorials/platform/consoles.rst
+++ b/tutorials/platform/consoles.rst
@@ -27,6 +27,6 @@ your games to consoles.
 
 Following is the list of providers:
 
-* .. _a Lone Wolf Technology: http://www.lonewolftechnology.com/ offers
+* `Lone Wolf Technology <http://www.lonewolftechnology.com/>`_ offers
   Switch, PS4 and XBox One porting and publishing of Godot games.
 


mhilbrunner:
Was moved to Tutorials->Platform-specific.

reduz:
I can't see it, also it's not a tuttorial IMO. I don't think anyone will go look for it there.
I think it's ok where it was, as we don't have an exporting section. 

I also prefer it's a visible, as a lot of users are interested about this and we don't offer this information. 
Many users have expressed interest in using Godot, but they don't use it because they believe exporting to consoles is not possible, even if it's not the immediate of their priorities.

akien-mga:
We do have an exporting section, and it should go there: http://docs.godotengine.org/en/stable/learning/workflow/export/index.html

akien-mga:
That was the *stable* link, here's *latest*: http://docs.godotengine.org/en/latest/getting_started/workflow/export/index.html

reduz:
@akien-mga Ah, totally forgot about it. I would do away with Project Workflow, then move
* Project Setup
* Asset Workflow

To getting started

and

* Exporting

to tutorials.

Issue #967, opened 2018-01-10
Write guidelines for screenshots


Calinou:
The current editor and project (2D/3D) screenshots tend to be inconsistent, so we should write some guidelines for screenshots that are intended to be featured in the documentation.

Some suggestions:
- Always use the default editor theme and font settings.
- Window borders should not be visible unless mandatory, e.g. for a window management tutorial.
- For highlighting regions of the editor (such as buttons or a panel), use a 2 pixel-thick bright red (`#ff4444`) outline without rounded or beveled corners. For large regions where an outline would be hard to see, use a semi-transparent bright red rectangle with 40% opacity.
- *For 3D screenshots:* Use 4× MSAA.
- *For 3D screenshots:* Enable anisotropic filtering on the project's textures and set the anisotropic filter quality to 16× in the Project Settings.
- Take all screenshots in hiDPI?
  - This will make them larger and will require handling to scale them down (it should be fairly easy, it's just a CSS change) but will look much nicer on hiDPI displays such as Retina.
  - This would also make contributing harder for people with low-resolution displays (< 1920×1080), as they wouldn't be able to fit everything on screen at 2× scaling to take screenshots.

Feel free to discuss below. :smiley:

pwab:
I totally agree with that. Just two comments:

> For highlighting regions of the editor (such as buttons or a panel), use a 2 pixel-thick bright red (#ff4444) outline without rounded or beveled corners. For large regions where an outline would be hard to see, use a semi-transparent bright red rectangle with 40% opacity.

- What about multiple regions? Especially when it comes to "_1. click here and 2. click there_". (Or something extreme like [this](http://docs.godotengine.org/en/latest/_images/godot-gui-overlay.png) found [here](http://docs.godotengine.org/en/latest/learning/editor/unity_to_godot.html))
- How to handle text on images?

> This would also make contributing harder for people with low-resolution displays 

- A simple workaround would be to ask for help while submitting a PR. One can simply take a screenshot as a placeholder and ask other people with larger displays to redo it.

pwab:
And I also want to add something regarding [a11y](https://en.wikipedia.org/wiki/Computer_accessibility):
If we choose to stick to this combination (red regions on default editor theme) people with _Protanopia_ (specific kind of [color blindness](https://en.wikipedia.org/wiki/Color_blindness)) could have a hard time looking at the pictures.

**Example:**
[...]

PR #1031, opened 2018-01-27
Add Image Contribution Guidelines

skyace65:
Adds image contribution guidelines to the doc based on what was discussed in issue #967. A color to use for highlighting a region of the editor is not mentioned since we never agreed on one. Closes issue #967


NathanLovato:
Thanks much! If it's okay for you I'll merge now, and we can discuss and settle down on one or several suggested colors in a new issue?

PR #1036, opened 2018-01-27
Fixed typo in ui game user interface tutorial

arawkins:
I was working through the UI tutorial and I think I found a typo. I found the Full Rect option in the Layout menu, not the Anchor menu.


NathanLovato:
Thanks, it's part of the things that changed during Godot 3 development after the tut landed in the docs

PR #1049, opened 2018-01-29
added pulseaudio-devel to solus as a necessary package

### LINKED ISSUE #905: Add pulseaudio-devel package to SolusOs dependecies on docs
Hello, after new Godot build in Solus Os, Godot was complaining about sound driver (drivers not found fall back to dummy driver) but pulseaudio was installed by default and working correctly with other software. 

SOLUTION: install pulseaudio-devel (or pulseaudio-devel32 for 32bit godot) package and rebuild Godot.

Please add _pulseaudio-devel_ package to SolusOs dependecies on docs website.

hungrymonkey:
Closes #905

### DIFF
--- a/development/compiling/compiling_for_x11.rst
+++ b/development/compiling/compiling_for_x11.rst
@@ -71,7 +71,7 @@ Distro-specific oneliners
 | **Solus**     | ::                                                                                                         |
 |               |                                                                                                            |
 |               |     sudo eopkg install -c system.devel scons libxcursor-devel libxinerama-devel libxi-devel \              |
-|               |         libxrandr-devel mesalib-devel libglu alsa-lib pulseaudio freetype2-devel                           |
+|               |         libxrandr-devel mesalib-devel libglu alsa-lib pulseaudio freetype2-devel pulseaudio-devel          |
 +---------------+------------------------------------------------------------------------------------------------------------+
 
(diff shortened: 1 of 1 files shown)


hungrymonkey:
I want to say that I do not have a solus machine. I am basing my fixes on the suggestion in the issue.

hungrymonkey:
Although I have written the commit for pulseaudio package, I am not sure about unhappy alsa only users. I am wonder if I should ever made the change at all. Please consider before merging this commit. This engine supports both Alsa and pulseaudio

StraToN:
IIRC, ALSA users need `alsa-lib` and don't need `pulseaudio-devel` while PulseAudio users need both.
Wouldn't it be better then to inform PulseAudio users more clearly that they need `pulseaudio-devel`.

I am not so sure about the idea of "forcing" users to install a dependency they don't really require.

hungrymonkey:
@StraToN that is why i mention it in the commit message. Alsa users will no doubt yell. I am just writing a pull request to fix a bug. 

I am not sure how I should add if/else dependencies in godot docs.

akien-mga:
That's fine, we also tell to install pulseaudio in all other oneliners. Users who know they don't want pulseaudio are clever enough to adapt the oneliner.

PR #1050, opened 2018-01-29
Various small fixes for the docs

YeldhamDev:
- Fixed several title underline sizes.
- Fixed LOTS of undefined labels.
- Some miscellaneous fixes here and there (character escaping, etc).

There where some problems that I could not fix however:

- In the **Design a title screen** tutorial, there's no link for the download of the final menu, I suppose that this was made by @NathanLovato, correct?
- Undefined labels at **DCC-created models** and **Generated geometry** inside **Introduction to 3D** weren't fixed, because this sections are in need of some rewriting, which is beyond the scope of this pull.
- Undefined label in the **InputEvent** tutorial, as I'm not sure to where it should be redirecting.
- Undefined labels in the **Matrices and Transforms** tutorial, as Matrix3 is supposed to be Basis, but there is already a Basis section for something different in the page.
- Removal of a reference to an **Exporting images** guide that doesn't exist as far as I'm aware of.

### DIFF
--- a/about/faq.rst
+++ b/about/faq.rst
@@ -37,13 +37,13 @@ Which platforms are supported by Godot?
 
 * Windows
 * Mac OS X
-* X11 (Linux, *BSD)
+* X11 (Linux, \*BSD)
 
 **For exporting your games:**
 
 * Windows (and UWP)
 * Mac OS X
-* X11 (Linux, *BSD)
+* X11 (Linux, \*BSD)
 * Android
 * iOS
 * Web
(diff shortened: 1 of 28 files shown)


about/faq.rst
```diff
@@ -37,20 +37,20 @@ Which platforms are supported by Godot?
 
 * Windows
 * Mac OS X
-* X11 (Linux, *BSD)
+* X11 (GNU/Linux, \*BSD)
```
akien-mga: I don't want to start a political debate, but we're using "Linux" to refer to the operating system everywhere, I don't think we should change now. We can have an issue to discuss it if you want, but that's out of the scope of "small fixes", as we'd have to enforce a *GNU plus Linux* policy everywhere to be consistent.

YeldhamDev: Well, the Asset Library guide (http://docs.godotengine.org/en/latest/tutorials/assetlib/what_is_assetlib.html) is written as GNU/Linux, so I tried my luck.

But I understand, changing it.

NullConstant: Hm, that's on me then since I wrote that, unaware of a docs-wide "Linux over GNU/Linux" policy (even though I use Linux over GNU/Linux myself). Might as well change that there's while you're at it.
@YeldhamDev

YeldhamDev:
@akien-mga Removed GNUs.

Forgive me Stallman.

PR #1063, opened 2018-01-31
Spelling correction

jcs224:


vnen:
Is there a guideline specifying whether to use American or British spelling?

jcs224:
Ah, I didn't realize (or should I say realise?) that. American here 😛 just looked it up. I know most of the differences (-our vs -or, -ise vs -ize), and I leave those alone, but I didn't know about this one. British English is more pervasive, and probably what most non-Americans learn. I'll be more careful next time.

In this specific case, the British version uses licence is a noun, but license as a verb. Americans use license for both noun and verb. Also, GitHub tries to correct my spelling if I try any British spelling.

This is actually an important question, especially for American contributors.

jcs224:
I just took a look at @reduz's last blog post and it is definitely using American spelling. Maybe he has some guidance on it?

vnen:
@reduz uses American spelling but @akien-mga prefers British. I'd say to go with American because 1) it's already used more often in the docs and 2) the API itself is in American spelling (there's a `Color` type not `Colour`).

Maybe this discussion can be moved to its own issue.

jcs224:
Got it, thanks @vnen!

jcs224:
Created issue #1066

PR #1071, opened 2018-02-01
Remove Bullet Physics from module example

skyace65:
Removes Bullet from the module example of external libraries since it was integrated in 3.0.


mhilbrunner:
Sorry, was faster this time :) Already done

PR #1086, opened 2018-02-03
Fix typo (`players` to `layers`)

dominicfreeston:
I'm not super familiar with Godot yet, but I'm guessing this should refer to `animation layers` not `animation players`. Feel free to reject otherwise :)


NathanLovato:
No, there's no animation layers. It's animation players indeed.

akien-mga:
> Maybe it should be AnimationPlayers instead for clarity?

Plurals of Node names are always a bit tricky. It could be:
<pre>
``AnimationPlayer``s
</pre>
if that works, otherwise `the <a href="link to the docs page>AnimationPlayer</a> node`.

PR #1095, opened 2018-02-04
Large copyedit of "getting_started/step_by_step/scripting.html"

gemkibeju:
I've edited [getting_started/step_by_step/scripting.html](https://godot.readthedocs.io/en/3.0/getting_started/step_by_step/scripting.html) for grammar, sentence structure, and typography. I've aimed to increase clarity and make the tutorial easier to understand.

I found this tutorial hard to follow because it assumes you know where in the Panel script to write all the functions, as well as failing to elaborate what the default script template does. I solved this issue by telling users to delete the template code under `extends panel`, so they have a blank slate to work with.

In general I've only ever worked in Python, so GDScript is foreign to me. Fortunately this means I'm not biased with the curse of knowledge, so I can look at every tutorial with a fresh pair of eyes and improve them from the perspective of a beginner.

Also, first pull request in a while.

### DIFF
--- a/getting_started/step_by_step/scripting.rst
+++ b/getting_started/step_by_step/scripting.rst
@@ -29,7 +29,7 @@ to its high integration with Godot:
 * Loads and compiles blazingly fast.
 * The editor integration is a pleasure to work with, with code completion for nodes, signals, and many other items pertaining to the scene being edited.
 * Has vector types built-in (such as Vectors, transforms, etc.), making it efficient for heavy use of linear algebra.
-* Supports multiple threads as efficiently as statically typed languages - one of the limitations that made us avoid VMs such as Lua, Squirrel, etc).
+* Supports multiple threads as efficiently as statically typed languages - one of the limitations that made us avoid VMs such as Lua, Squirrel, etc.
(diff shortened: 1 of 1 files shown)


cbscribe:
Thanks for doing this, but I think this doc is more in need of a rewrite rather than a simple grammar clean-up. It's a bit schizophrenic: it states that GDScript and VisualScript are the "main" languages, but the copy is very GDScript focused, and the examples show GDScript and C# (no VisualScript), with no mention of how to attach a C# or VS script instead of GDScript.

I'm cool with the changes - grammar improvements are always good, but a full rethink of the doc should be discussed. It's something I've wanted to do for a while but haven't had the time, I'm afraid.

mhilbrunner:
Thanks for doing this, looks like a vast improvement. :)

I agree with both of you that there needs a way gentler introduction to GDScript (and the other scripting options). If there is no issue for this yet, please open one so we don't forget this @cbscribe / @gemkibeju?

Going to merge this for now, even if eventually big parts are going to be replaced, as the fixes are better than the current state, especially as this is an important page and probably one of the first that gets looked at by new users.

cbscribe:
See #531

PR #1108, opened 2018-02-06
Fixed wrong userdata path for unix-like systems in "Data Paths"

YeldhamDev:
This is the current directory for it in GNU/Linux distros at least. I'm just supposing that it's the same for macOS.


akien-mga:
It's not the same in macOS, no. I'll fix that tonight as it needs more details for all platforms.

PR #1114, opened 2018-02-07
Fixed build command for shared library summator

Simonah88:
I can't seem to get the example to work as is on Ubunto 16.04. With these changes it works perfectly as a shared library but without it I get complaints about x11. Then if platform=x11 is added I get:

[  0%] scons: *** Do not know how to make File target `bin/summator.x11.tools.64.so' (/home/simon/projects/godot/bin/summator.x11.tools.64.so).  Stop.

Very unfamiliar with c++/scons/godot, not sure if docs should be changed or if it is a bug. ( I am running off of godot master ) 

Thanks,

Simon

### DIFF
--- a/development/cpp/custom_modules_in_cpp.rst
+++ b/development/cpp/custom_modules_in_cpp.rst
@@ -338,7 +338,7 @@ shared module as target in the scons command:
 
 ::
 
-    user@host:~/godot$ scons summator_shared=yes bin/summator.x11.tools.64.so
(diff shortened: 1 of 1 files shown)


NathanLovato:
Need someone else to check this. @vnen maybe you'll know how it works?

vnen:
I don't really know how this works (tried to make it work on Windows but failed). But this change makes sense since `platform` is a required argument and AFAIK SCons accepts the target file as a argument.

PR #1118, opened 2018-02-08
Fix typos in Control the game's UI with code

daglilley:
Fixed a typo.

Also fixed instances of `tween` mistakenly written as `Tween` in code samples. The samples were also written correctly elsewhere on the page. But these miscapitalised instances result in invalid calls (`Nonexistant function`).


NathanLovato:
Thanks! Seemed I've missed this one after changing the source code.

Issue #1084, opened 2018-02-03
Discussing a "godot-docs" style guide and several instances of somewhat off-putting writing


ghost:
I've been learning Godot and going through the documentation tutorial under the "step by step" section. I do notice, though, that some of the grammar, placement of punctuation, and sentence structure feels somewhat "off" when read by a native English speaker such as myself, as if it was written by contributors with a near-native understanding of English but without the experience to detect every edge case that comes from writing at length.

Some examples from the [scripting section](https://godot.readthedocs.io/en/3.0/getting_started/step_by_step/scripting.html) would be:

> Actually, in GDScript, the file itself represents the class, so in this case, the class name field is not editable.

Most of the time you don't want that many commas in a single sentence, as every comma reduces your reading speed (or "flow") and putting too many of them in a row makes the sentence harder to understand due to the pauses that naturally occur while reading commas. I would remove the "Actually," as well as the comma after "case".

> Note: A common misunderstanding in this tutorial is how get_node(path) works.

This sentence is vague; it can mean that the tutorial itself misunderstands how "get_node(path)" works, or the reader misunderstands. I would rephrase it as "A common misunderstanding when reading this tutorial..."

> **GDNative / C++**
> Finally, this is one of our brightest additions for the 3.0 release. GDNative allows scripting in C++ without requiring to recompile (or even restart) Godot.

Saying that "this is one of our brightest additions" is a bit redundant when the header already shows what the addition is, being GDNative. The second sentence uses "requiring" when most writers would use "requiring you", "needing", or "having" (less formal than "needing"). I would also phrase the first sentence as "Finally, one of our brightest...". These suggestions sound more "natural" in English, but there's no scientific definition for what sounds natural, so you'll just have to trust me on that 🙂

I know it sounds petty to nitpick like this, but when you see a lot of small issues over the course of several pages worth of documents, it adds up to make a jarring experience for the reader. 

### The solution:

The solution to the above would usually involve making fixing up the page and pushing an edit to it, much like Wikipedia. Unfortunately the GitHub model of needing to approve every pull request leads to a bureaucracy where maintainers would argue over minor details with the pull requester and would either reject the request or have to wait significant amounts of time in order for the requester to edit their contribution. 

So instead of bickering over who has the best idea of how the documentation should be written, it would be nice to develop a style guide for contributors, so that we can all agree on the best way to write for the Godot documentation and reference the guide for all of our writing style disputes.

Such issues could include:

[...]

PR #1126, opened 2018-02-10
Updated grammar in instancing_continued.rst

natehitchcock:
There was no need for that hyphen

### DIFF
--- a/getting_started/step_by_step/instancing_continued.rst
+++ b/getting_started/step_by_step/instancing_continued.rst
@@ -44,7 +44,8 @@ A lot of time spent in programming games (or software in general) is on
 designing an architecture and fitting game components to that architecture.
 Designing based on scenes replaces that approach and makes development much
 faster and more straightforward, allowing you to concentrate on the game logic
-itself. Because most game components map directly to a scene, using a design-based on scene instantiation means little other architectural code is needed.
+itself. Because most game components map directly to a scene, using a design
+based on scene instantiation means little other architectural code is needed.
 
 Let's take a look at one more, somewhat more complex, example of an open-world
 type game with lots of assets and nested elements:


NathanLovato:
It's not a grammar error. You may also want to check the ongoing discussion on a docs styleguide: https://github.com/godotengine/godot-docs/issues/1084#issuecomment-362859836.

natehitchcock:
Sorry, are you saying that 'design-based' is a [phrasal-adjective](https://practicaltypography.com/hyphens-and-dashes.html) in this case? I would dispute that. In the phrase 'using a design-based on scene instantiation...', 'on' is not a noun, and can't be described with an adjective. Maybe the phrase was supposed to be 'using a design-based scene instantiation'? Or maybe we should just remove this whole fragment, making the final sentence 'Because most game components map directly to a scene, little other architectural code is needed.'

NathanLovato:
My bad, I read too quickly and saw it as a `design-based project`

NathanLovato:
Thanks for the heads up, merging now

NathanLovato:
Ah it's in the 3.0 branch, crap. Right now to fix the docs redirecting you to the wrong version it's forced to go to the 3.0 version, but all fixes should go to the master one, named `latest` on the docs: http://docs.godotengine.org/en/latest/

natehitchcock:
Ah, I was using the 'Edit on GitHub' link in the docs, I'll just make my changes directly from GitHub in the future. Sorry about that.

Additionally, maybe the 'Edit on GitHub' link could point to the master branch? There would be minor discrepancies, but the flow for correcting typos or misinformation would be smoother.

NathanLovato:
> I didn't even realize I was on v3.0.
Don't worry, I didn't either ^^

PR #1137, opened 2018-02-12
Create PR template

mhilbrunner:


akien-mga:
Doesn't seem to work as intended, most of the information is missing and the first two paragraphs are included in the PR description:
image

mhilbrunner:
akien: I changed it later in https://github.com/godotengine/godot-docs/commit/7fba674fcfd32152f2b24bedd79f31717047cc3d because it's annoying if it's too much text and also from testing... it sadly isn't **that** useful, as you pointed out: At that point it's mostly too late anyway. Really have to see if we can fix the "Edit this on GitHub" link somehow (https://github.com/godotengine/godot-docs/issues/1138)

PR #1143, opened 2018-02-13
Updating resource restriction docs for Xbox One

jbuschdev:
Microsoft recently lifted the restrictions of UWP games on Xbox One if submitted through the "Xbox Live Creators Program" to allow nearly the full power of the Xbox One.


akien-mga:
Thanks!

PR #1147, opened 2018-02-15
Direct to InputEventJoypadButton for using buttons

sfotm:

### DIFF
--- a/classes/class_inputeventjoypadmotion.rst
+++ b/classes/class_inputeventjoypadmotion.rst
@@ -14,7 +14,7 @@ InputEventJoypadMotion
 Brief Description
 -----------------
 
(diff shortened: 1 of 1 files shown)


vnen:
Thanks for the contribution, but notice that 1) PRs should be made against the master branch (as stated in the template) and 2) This a generated file and should not be changed directly as stated in the file's header:

>   .. Generated automatically by doc/tools/makerst.py in Godot's source tree.
.. DO NOT EDIT THIS FILE, but the InputEventJoypadMotion.xml source instead.
.. The source is found in doc/classes or modules/\<name>/doc_classes.

This should be edited in the XML from Godot's main repository.

sfotm:
Alright, thanks for filling me in. I'll close this one and read a bit more thoroughly next time.

PR #1149, opened 2018-02-16
Corrected physics Introduction using move()

walomator:
In the [Physics Introduction](https://godot.readthedocs.io/en/stable/classes/class_viewport.html) page of the docs, it is incorrectly reported that KinematicBody2D uses move(), when in fact it uses move_and_slide() as of version 3.0. This does not apply to earlier versions of Godot.


NathanLovato:
Thanks!

NathanLovato:
Note you should work on http://docs.godotengine.org/en/latest/, not on the 3.0 branch. Gonna cherry pick the changes with git for you.

PR #1155, opened 2018-02-17
Assetlib fixes again

RiverMesa:


RiverMesa:
Gah, I screwed this one up, hold on.

PR #1162, opened 2018-02-19
Links for everything that needs to be installed

StJohn3D:
I really appreciate the android sdk link, and honestly had never heard of Open JDK or Oracle JDK before today. So a quick google search brought me to the link I've put in this proposed change. If it's correct, and the placement is good then please accept this PR.

### DIFF
--- a/getting_started/workflow/export/exporting_for_android.rst
+++ b/getting_started/workflow/export/exporting_for_android.rst
@@ -15,7 +15,7 @@ http://developer.android.com/sdk/index.html
 Install OpenJDK or Oracle JDK
 -----------------------------
 
-Download and install OpenJDK or Oracle JDK. Version 6 and 8 are known to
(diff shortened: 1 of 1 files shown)


akien-mga:
Would be better to put the link under `OpenJDK`, and add a similar link under `Oracle JDK`, to clear the confusion as you mentioned.

StJohn3D:
Do you have those links then? Like I said this is all new to me. I thought I found a link to a page with links to both the Open and the Oracle versions. From that page it sounds like you download either `Open` or `Oracle` depending on your OS. Is that right?

StJohn3D:
So for windows it looks like I have to go here. https://github.com/ojdkbuild/ojdkbuild
Because I'd like to stay as far away from Oracle and Java as I can.

mhilbrunner:
Added links in https://github.com/godotengine/godot-docs/commit/8d879a8a45c5c6c923ed064ebec953bdea8444a1. Thank you!

PR #1182, opened 2018-02-26
Optimize 3D FPS tutorial ZIP file sizes

mhilbrunner:
Both https://github.com/godotengine/godot-docs/blob/master/tutorials/3d/fps_tutorial/files/Godot_FPS_Finished.zip and https://github.com/godotengine/godot-docs/blob/master/tutorials/3d/fps_tutorial/files/Godot_FPS_Starter.zip were above 75mb, mostly because of a single texture with about 45mb and the .import directory being included. (Which of course also contained a copy, doubling the size)

The texture is used for a skybox in the demo. I changed the size down from 8k x 4k to 4k x 2k, and used a PNG compressor on it. This trades some quality for a lot of space and isn't much worse, IMO. Should be fine for a demo.

The zipped project is now 6mb, down from 75mb.

Things to consider:
- Maybe use Git LFS on this repo
- Maybe move the demo .zips to the demo repo
- We need a better process when reviewing PRs, this should have been caught by one of us, instead of bloating the Git history by about 140mb; I missed it too


cbscribe:
@TwistedTwigleg 

Also, we should make sure the `.import` folder is deleted from zipped projects, where .gitignore is not applied.

Calinou:
> Maybe use Git LFS on this repo

GitHub puts heavy restrictions on Git LFS usage in free repositories (1 GB of file size *and* 1 GB monthly bandwidth), so it's not really usable.

TwistedTwigleg:
Sorry guys! I didn’t even think of checking the zip file sizes. My apologizes, I’ll try to do better next time.

> ... instead of bloating bloating the Git history by about 140mb; ....

Can we remove the commits from the commit history to reduce the bloat? While likely not considered a good practice (any time we are editing the history we need to be careful) but that way we could then just make a new PR with the tutorial at a reduced size.

Also, should we just link to the FPS tutorial Github repository directly instead of attaching the zip files? We could link to a release with the relevant zip files, and that would both get the zip files out of the documentation repository, and would also make file size matter less (I’m not sure if there are limits on Github releases or not).

What do you guys think?

TwistedTwigleg:
@mhilbrunner Sorry for the late reply.
Tampering with the history is a risky buiseness since it is really easy for things to go wrong. I was only suggesting because it would remove the 8K image from the git history. I totally agree with letting the others decide :+1: 

Also, I had no idea it was an 8K image (Somehow I missed that part in the issue details) :scream: Huge apologizes for missing it. I'll fix it on the FPS repository tomorrow, so hopefully we will not have this issue again.

mhilbrunner:
No worries!

PR #1199, opened 2018-03-03
Setup docs for translation

### LINKED ISSUE #1057: Any update about translations?
Just read on the official documentation:

> For the moment, only English documentation is provided. We want to propose localized documentation in the future, but this is not a priority for now. 
(...)
When the English documentation is ready for translations, we will provide tools to simplify the work of translators by tracking changes to the English docs that they should translate on their end.

Any discussion going on regarding this? Even if it is low priority, localized documentation can make Godot user base grow a lot. (And is one of the few ways I can contribute, really want to help 😃)

Juankz:
This setup uses Spanish as an example which can be replicated with any other [sphinx supported language](http://www.sphinx-doc.org/en/stable/config.html#confval-language).

.po files were generated using `sphinx-intl` (http://www.sphinx-doc.org/en/stable/intl.html quick guide steps) and are located under the `locale/` directory.

New languages can be added (and updated) using `$ sphinx-intl update -p _build/gettext -l [lang]`.

Translated files: `locale/about/introduction.po`

May close #1057


akien-mga:
Nice work! Some early thoughts:

- `.mo` files should not be added to git, you should instead add `*.mo` to `.gitignore`. They're binary files meant to be generated while compiling the docs.

- Having all translations in the same repo as the English source may prove a burden on documentation contributors, if PRs are flooded with translation updates that most can't review. Moreover, [RTD specifies](https://docs.readthedocs.io/en/latest/localization.html#project-with-multiple-translations) that each locale must have its own RTD project, so if we keep them all in the same repo, it means that each commit (e.g. to Spanish translation) will trigger a rebuild of all locales.
That said, if we put the PO files on Weblate and only sync them manually all at once, once in a while (as I do for the engine translation), it might be ok.

Juankz:
Good. Lets see this points one by one:

- `.mo` files have been deleted and added to the `.gitignore` file.
- The idea is to use Weblate as is being used for the engine. For reference the [linux mint installation guide](http://linuxmint-installation-guide.readthedocs.io/en/latest/) has all its translations on the [same repo](https://github.com/linuxmint/doc-installation-guide). Although Godot documentation is far longer and complex. I can't decide whether or not would be better to split it on several repos.
- Class reference has been excluded. Now when getting new translatable messages one must use `make -e SPHINXOPTS="-t i18n" gettext` instead of simply `make gettext`

mhilbrunner:
@akien-mga Some thoughts:

- Translation files indeed need to live in their own repo. **godot-docs-i18n**? We can merge the config and setup changes for sphinx here, but the text files should definitely go into their own repo.

- We **really** need to discuss whether and how we want to *officially* support this. By doing so, and having it on **docs.godotengine.org**, the quality of the translations reflect on the entire engine.

Consider these scenarios:

- I'm a spanish Godot user. Cool, spanish docs.
However only a third of the text seems to be translated, rendering it a weird hodgepodge of english and spanish. Some spanish information seems wrong, outdated or even contradicts what some of the english text says (and the english text itself already is outdated in some places).

- A contributor sends a PR. We merge it. What do we do for the translations? Somebody needs to update them, keep them in sync. We'd need translation teams for each language we support, which IMO is hard to do with the size we're currently at.

To be clear, I think having translated docs would be nice and *could* be worth it, even considering this, but we need to discuss how the processes will work and how to minimize the potential downsides.

vnen:
If we're using gettext, then it should happen the same way as in the interface translation I think.

* When someone sends a PR to add/change text, Weblate will notify that there are new strings to translate.
* If the original source changes, the translations will be marked as `fuzzy`. In the editor we ignore those and show the original instead, but I'm not sure if we have that control with RTD/Sphinx. Also, I don't know if translation happens at a paragraph basis or what, that makes a difference when part of the text is not translated (but not much we can do about that either).
* Translation teams are responsible to organize themselves. They can use the available community channels to achieve that. Many have already a local Facebook group or something that they can use too. Weblate also has a place for comments and discussion about specific translations.

I think that the translations need to be official, at least in the hosting part. Otherwise it will happen over and over again that someone starts an third-party translation venture and abandon it halfway. But we can make clear that translations are maintained by the community and we can't assess their quality. I don't see many complaints about the editor interface translation (though that requires smaller effort).

akien-mga:
Sorry for the delay - I'll start playing with this branch locally and see how we would handle the POT generation and PO updates practically, as well as publishing the localised versions.

akien-mga:
Thanks for your work and the references, I've merged the config changes as cf22d19 (with some style fixes), and I'll leave out the .po files for now.

I'll have a look at setting up another repo for localized docs, so that we can sync PO files there as we want without messing with the documentation writing work here. Probably something similar to https://github.com/phpmyadmin/localized_docs, though I hope less convoluted.

PR #1205, opened 2018-03-04
stop StartTimer in game_over() 

KellyThomas:
Brings documentation into line with https://github.com/kidscancode/Godot3_dodge/pull/5


mhilbrunner:
cc @cbscribe

cbscribe:
This isn't necessary, as StartTimer should be set as OneShot in the Inspector.  See http://docs.godotengine.org/en/latest/getting_started/step_by_step/your_first_game.html#main-scene

PR #1207, opened 2018-03-05
small clarification

Nickerchen:
the original phrasing was hard to understand for a newbie and made some small problem


mhilbrunner:
Hi! As changes to the docs should go into the `master` branch instead of the `3.0` branch (which is kept in manually), I added your clarification to the `master` branch:
https://github.com/godotengine/godot-docs/commit/556040b7fe2f3b02d00e262b9dcd611e91db9884

Thanks for your contribution!

PR #1208, opened 2018-03-05
Added note on exporting from command line

sdfgeoff:
This has some duplicate information with:
http://docs.godotengine.org/en/3.0/getting_started/editor/command_line_tutorial.html#exporting

But also adds some new things. I also feel that the obvious place for this information is under "Export" rather than under "Command Line" as that is where a human would look for the information on ... exporting.

### DIFF
--- a/getting_started/workflow/export/exporting_projects.rst
+++ b/getting_started/workflow/export/exporting_projects.rst
@@ -132,3 +132,31 @@ creates the package. There are 3 different modes for exporting:
    *really* useful for games distributed on optical media.
 
 .. image:: img/expselected.png
+
+
+
+Export from Command Line
+------------------------
+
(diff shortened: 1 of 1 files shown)


getting_started/workflow/export/exporting_projects.rst
```diff
@@ -132,3 +132,31 @@ creates the package. There are 3 different modes for exporting:
    *really* useful for games distributed on optical media.
 
 .. image:: img/expselected.png
+
+
+
+Export from Command Line
+------------------------
+
+In production it is useful to automate builds, and godot supports this
```
mhilbrunner: Godot should be capitalized

sdfgeoff: Fixed

PR #1223, opened 2018-03-12
Added windows_icon tutorial

ZX-WT:
A simple tutorial about changing the icon for windows. It provides the necessary stuff, however, I would like to improve it a little more. Can anyone recommend a tool for exporting .ico file, better be FOSS. It seems gimp doesn't support it, is .ico some kind of format that is actually an existing format but dresses in different clothing?

Also, provide some requirement for the image like the max. dimensions etc if possible, I have no experience about ico . 

I haven't test it with wine myself but I suppose it works normally, correct me if you know about it.

I find out that you can use a png instead of ico but it only changes the icon in the title bar, no idea what's happening but interesting :)


vnen:
Gimp does support the `.ico` format. I was using it in my tests when I added the feature. I don't remember exactly how you should do it, but I think it needs one layer for each icon size (the `.ico` contains icons in many sizes inside the same file).

The icon in the title bar actually uses the icon defined in project settings, which is a PNG file.

getting_started/workflow/export/changing_applicationIcon_for_windows.rst
```diff
@@ -0,0 +1,52 @@
+.. _doc_changing_applicationIcon_for_windows:
+
+Changing applciation icon for windows
+==========================================
+
+By default, the exported game icon will be the godot icon, most likely you will like to change the icon for your game. There are two types of icon that can be changed, the file icon and the taskbar icon.
```
mhilbrunner: Capitalize **G**odot, two types of icon**s**

getting_started/workflow/export/changing_applicationIcon_for_windows.rst
```diff
@@ -0,0 +1,52 @@
...
+----------------------------------
+The file icon is the icon that you click on to start the game.
+
+.. image:: img/icon_file_icon.png
+
+Before selecting it in the export options, you will need to install an extra tool called rcedit.
+You can download it here:
+https://github.com/electron/rcedit/releases
+
+After downloading, you need to tell godot the path of rcedit on your computer. Go to Editor>Editor Settings>Export>Windows. Select the rcedit exectuable by clicking the folder icon in the rcedit entry.
```
mhilbrunner: Capitalize **G**odot

ZX-WT:
I overlooked the option in gimp, indeed, it can export ico and I find a online tutorial about it, just added a link to it. I will rebase it if everything is ok.

PR #1234, opened 2018-03-15
improve 3d workspace intro (added description about the [perspective] button)

ZX-WT:
I added a description about the [perspective] button as it is often neglect by users.


mhilbrunner:
Thanks. :)

PR #1237, opened 2018-03-15
Replaced using the select mode tool by changing the Size property

Rubonnek:
I'm not sure what the select mode tool meant, but as far as I can tell I was able to do the same thing by changing the Size property of the Rect.


mhilbrunner:
Thank you!

PR #1238, opened 2018-03-16
Docs error

tecbeast42:
Using Godot 3.0.2 the export switch will not recognize if you dont use `--` instead of `-`


mhilbrunner:
Thanks for spotting this and helping us fix this :) I manually made those small fixes in the `master` branch, as your PR points to the `3.0` branch. Thanks again for your contribution!

PR #1248, opened 2018-03-22
cleaner language on 'exporting for UWP' doc page

kn00tcn:
i first noticed a malformed sentence at the top, then decided to clean up some of the grammar scattered across the rest of the page, also turned the note paragraph at the end into a note box

### DIFF
--- a/getting_started/workflow/export/exporting_for_uwp.rst
+++ b/getting_started/workflow/export/exporting_for_uwp.rst
@@ -4,15 +4,15 @@ Exporting for Universal Windows Platform
 ========================================
 
 There's no extra requirement to export an ``.appx`` package that can be
-installed as a Windows App or submited to the Windows Store. Exporting
-packages also works from any platform, not only on Windows.
+installed as a Windows App or submited to the Windows Store. Exporting UWP
+packages also works from any platform, not only from Windows.
 
 However, if you want to install and run the app, you need to sign it with a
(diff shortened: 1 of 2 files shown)


kn00tcn:
i'm not familiar with windows dev, but i tried to find information on `signtool.exe` since it was suddenly mentioned without a source like the first two commands

it seemed like it's part of the SDK or VS, so i mentioned that

certutil seems to be part of regular windows installations, so it doesnt need a reference

mhilbrunner:
Thank you for cleaning this up :) Could you create the same PR against the `master` branch? We sync `3.0` from `master` manually.

kn00tcn:
do you mean create a duplicate pull request or can i just press the edit button & change to master? i dont know if the index.rst file appearing after i used the edit button is correct

mhilbrunner:
The issue with using the edit button is that you get 48df918 in your PR too. I can cherry-pick just your commit (766ce05) but you're probably not getting the proper credit for the commit and this PR then.

If you don't mind I'll do that, but otherwise it would probably be easiest to just create a duplicate PR against `master` and close this one. Your choice :)

kn00tcn:
ahh, i really did read that commented message when creating pull requests, why did i think i was on master back then...

i came from the 'edit' button directly on the doc page, which points to 3 of course, but this isnt specific for it

made it again https://github.com/godotengine/godot-docs/pull/1260

PR #1251, opened 2018-03-22
Fix dictionary declaration in 'Saving games' tutorial

antoniotorresm:
For some reason the dict was not declared correctly. Also changed `get_pos()` to `position` for to match 3.0 way of getting members.


mhilbrunner:
Thanks!

PR #1255, opened 2018-03-23
3.0 Docs Update

pixelprogrammer:
Updated the docs to fix one of the method names which causes an error.

In Godot 3.0 (not sure if in earlier versions) the function name is `_on_Button_pressed` not `_on_button_pressed`. The latter throws an error with method not found.

In the Docs it reads:

By default, the method name will contain the emitting node’s name (“Button” in this case), resulting in “_on_[EmitterNode]_[signal_name]”.

### DIFF
--- a/getting_started/step_by_step/scripting.rst
+++ b/getting_started/step_by_step/scripting.rst
@@ -259,7 +259,7 @@ Next, write a function which will be called when the button is pressed:
 .. tabs::
  .. code-tab:: gdscript GDScript
 
-    func _on_button_pressed():  
+    func _on_Button_pressed():  
         get_node("Label").text = "HELLO!"
 
  .. code-tab:: csharp
@@ -293,7 +293,7 @@ The final script should look like this:
 
     extends Panel
 
-    func _on_button_pressed():
+    func _on_Button_pressed():
         get_node("Label").text = "HELLO!"
 
     func _ready():
--- a/index.rst
+++ b/index.rst
@@ -3,12 +3,13 @@
    You can adapt this file completely to your liking, but it should at least
    contain the root `toctree` directive.
 
-Godot Docs – *master* branch
-============================
+Godot Docs – *3.0* branch
+=========================
 
-.. attention:: This is the documentation for the unstable (master) branch.
-               Looking for the documentation of the current **stable** branch?
-               `Have a look here <http://docs.godotengine.org/en/stable>`_.
+.. tip:: This is the documentation for the stable 3.0 branch.
+         Looking for the documentation of the current **development** branch?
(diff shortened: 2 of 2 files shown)


pwab:
**First of all:**
You should create your PR for master branch because changes are cherry-picked for the 3.0 branch.

**Regarding your proposed change:**
You are right that the function-name generated by Godot uses `B` instead of `b` because it just copies the Nodes name (which is "Button"). And when the tutorial tries to implement the same behaviour by code it uses `b` instead.

**But** I'm not sure why this triggers an error for you. Because if you read the next lines the binding is made in the correct way:

```gdscript
func _on_button_pressed():
    get_node("Label").text = "HELLO!"

func _ready():
        get_node("Button").connect("pressed", self, "_on_button_pressed")
```

pixelprogrammer:
@pwab 
Ok, I revisited the project and there were a few things that were going on which were causing me to think the issues were coming elsewhere. So this was more of an error on my part than the documentation.

The event was attached from the Node Signals section from the right panel. Also, in my script the event wasn't binding because I had this instead:

```
func _read():
    # This wasn't running because I am pretty sure the function is supposed to be _ready() not _read()
    get_node("Button").connect("pressed", self, "_on_button_pressed")
```

So the function wasn't binding despite originally being the _on_button_pressed() name originally. Must have deleted a letter without knowing it. So this is where my original issue was coming from. Sorry about that.

I can confirm that the original documented code version works but there was definitely some confusion with the generated function name and the customized one. Because as a reader I was following along and had that function generated for me which was working. But modifying the code so it was the same as the documentation at that point was where the error started happening for me.

I think the part after connecting the signal from the right panel in the documentation would be a good time to test it so far to see if it works, then disconnect the signal and then show how to do it manually. Or, you could connect another signal manually to show both ways to go about it. Anyways, thats another discussion and this PR wasn't helpful so I may now walk away in shame.

pwab:
@pixelprogrammer 

> Anyways, thats another discussion and this PR wasn't helpful so I will now walk away in shame.

Oh no you don't have to. This kind of feedback regarding little details is very much appreciated. I can clearly see that this point in the tutorial can lead to confusion if the reader is not following word by word. And to be honest: that might be the behavior of the average newbie. So we should look out for options to improve this because this is what makes great documentation.

> Or, you could connect another signal manually to show both ways to go about it.

This is a great idea. This could prevent this kind of mistake. Just add another button (which automatically gets another name like Button2) and then connect it manually. Would you mind creating a PR for this so we could discuss it further there?

mhilbrunner:
Feel free to open a PR for the suggestion as discussed above, if you want to. :) Closing this one for now.

PR #1256, opened 2018-03-23
Update importing_images.rst

hubslave:
Corrected a mistake in Compress Lossy (PNG) field, any compression that is changing the colors on an image is Lossy at best + it's bad to lie to developers.
DIFF MAP comparing original with PNG compressed image as proof: http://www71.zippyshare.com/v/MKHmzMkK/file.html

### DIFF
--- a/getting_started/workflow/assets/importing_images.rst
+++ b/getting_started/workflow/assets/importing_images.rst
@@ -45,7 +45,7 @@ In this table, each of the four options are described together with their
 advantages and disadvantages ( |good| = Best, |bad| =Worst ):
 
 +----------------+------------------------+---------------------------+-------------------------+------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


mhilbrunner:
Lossless mode should indeed be lossless. If it is not for you, this is probably a bug in Godot, not a documentation issue.

In this case:
- Please open an issue on the *godot* repository, not *godot-docs*
- Provide the exact steps you did, including the options, what you expected and what you got instead
- Submit the above image and any other data you think is useful (you can upload images directly on GitHub)

Closing this one.

hubslave:
If you consider this a bug a note should be placed in the documentation about it until it's fixed, also if this is not the correct place to post the PR then you should fix the link on top of http://docs.godotengine.org/en/3.0/getting_started/workflow/assets/importing_images.html to point to the correct one ...

PR #1261, opened 2018-03-26
Step by step fixes

josephskelly:
Hello.

This is a small edit to the step by step tutorial. I fixed an inaccurate image description in the introduction to the ui system page and updated the section of the main menu tutorial that creates the MarginContainer. The Full Rect Layout step should be before setting any margins as that step resets the margins to zero. Also, the right and bottom margin values should be negative.

Thanks. This is my first ever pull request, and I'm excited to help. Let me know if I made any mistakes.

PR #1262, opened 2018-03-27
Revert commit in Step by Step Fixes #1261

josephskelly:
I'm pretty embarassed, but I messed up a commit in a previous pull request (#1261). This PR just reverts the commit that contained the error.

There was some confusion on my part about which margins the tutorial was defining. Both margins were in the control section as described in the tutorial, but I had not been working with the custom constants as instructed. Sorry for the confusion. I will definitely try to be more thorough in the future.


mhilbrunner:
No worries, it happens; just be thorough :) Thanks for fixing it!

PR #1266, opened 2018-03-28
Fixes for builtin shader functions

Chaosus:
Various fixes for builtin shader functions..


bojidar-bg:
I would personally prefer them to be listed as "Inverse hyperbolic sine", or maybe "Area (inverse hyperbolic) sine".

Chaosus:
@bojidar-bg Ok, done

PR #1272, opened 2018-03-30
Add detail to doc for add_surface_from_arrays

garyo:
This clarified things for me, so maybe it's useful for others.


mhilbrunner:
Hey! Thanks for clarifying this more :) Unfortunately, the class reference `.rst` files are autogenerated from `.xml` files that live in the original Godot engine repository.

So instead of this PR, you'd need to open a PR there and edit https://github.com/godotengine/godot/blob/master/doc/classes/ArrayMesh.xml if it's not too much of a bother :)

garyo:
ok, I'll resubmit against the XML code (yuck! :-) )

PR #1274, opened 2018-03-31
Fix link

fahmyrm:
The link for "main menu tutorial" was pointing at the current page instead of the tutorial


mhilbrunner:
Thanks for pointing this out! Could you create a new PR with this fix against the `master` branch? PRs should go in master by default, 3.0 is synced manually by us. I know this can be confusing because the ReadTheDoc link points to the 3.0 branch on GitHub. Sorry for the hassle!

If it's too much of a ask, I can also manually do the small fix in the `master` branch myself, but that way out won't get the credit. :)

fahmyrm:
Done. Thanks for your help!

Issue #1247, opened 2018-03-21
Compiling: wrong templates dir (linux)


LinuxUserGD:
see https://github.com/godotengine/godot/issues/17661
According to volzhs, the templates directory should be ```/home/[username]/.local/share/godot/templates/[gd-version]/``` and not ```/home/[username]/.godot/templates```

Docs:
- https://godot.readthedocs.io/en/latest/development/compiling/compiling_for_x11.html#building-export-templates
- https://godot.readthedocs.io/en/latest/development/compiling/compiling_for_android.html#installing-the-templates

deleteaccount1234567890:
WHERE IS TEMPLATE - I have tried - no success :( Because It throws error like 
![screenshot from 2018-10-10 23-32-00](https://user-images.githubusercontent.com/3496136/46767263-bff64e00-cce4-11e8-8392-23eb5af7b400.png)

Why do you nerve me - I already updated from github since yesterday - And I want build single executable but I can't get templates - I already try step to step - I feel nerving if I have no templates. It is really crazy.

Please tell me correct because I want try Godot.

PR #1276, opened 2018-03-31
fix x11 export templates dir

LinuxUserGD:
#1247


mhilbrunner:
Thanks for the fix :)

deleteaccount1234567890:
Since April 2018? But now it happens :(
screenshot from 2018-10-10 23-32-00
How do I understand that?

PR #1277, opened 2018-03-31
Added rigid body doc

AndreaCatania:
I don't know if the syntax to show the new page on site is correctly so please check it before merge.
Also sorry for eventual english errors

### DIFF
--- a/tutorials/physics/rigid_body.rst
+++ b/tutorials/physics/rigid_body.rst
@@ -1,0 +1,52 @@
+.. _doc_rigid_body
+
+Rigid Body
+==========
+
+What is a rigid body?
+---------------------
+
+A rigid body is one that is directly controlled by the physics engine in order to simulate the behavior of physical objects.
+In order to define the shape of the body, it must have one or more :ref:Shape <class_Shape>objects assigned. Note that setting the position of these shapes will affect the body's center of mass.
+
+How to control rigid body
+-------------------------
+
+A rigid body's behavior can be altered by setting its properties such as friction, mass, bounce, etc. These properties can be set in the Inspector or via code. See :ref:RigidBody <class_RigidBody> for the full list of properties and their effects.
+
+There are several ways to control a rigid body's movement, depending on your desired application.
+
+If you only need to place a rigid body once, for example to set its initial location, you can use the methods provided by the :ref:Spatial <class_Spatial> node, such as ``set_global_transform()`` or ``look_at()``. However, these functions can not be called every frame or the physics engine will not be able to correctly simulate the body's state.
+As an example, consider a rigid body that you want to rotate so that it points towards another object. A common mistake when implementing this kind of behavior is to use ``look_at()`` every frame, which breaks the physics simulation. Below, we'll demonstrate how to implement this correctly.
+
+The fact that you can't use ``set_global_transform()`` or ``look_at()`` methods doesn't mean that you can't have full control of a rigid body. Instead, you can control it by using the ``_integrate_forces()`` callback. In this function, you can add *forces*, apply *impulses*, or set the *velocity* in order to achieve any movement you desire.
+
+Look at function
+----------------
+
+As described above, using the Spatial node's ``look_at()`` function can't be used each frame to follow a target.
+Here is a custom ``look_at()`` function that will work reliably with rigid bodies:
+
+.. tabs::
+ .. code-tab:: gdscript GDScript
+
+    extends RigidBody
+
+    func look_follow(state, current_transform, target_position):
+	
+	var up_dir = Vector3(0, 1, 0);
+	var cur_dir = current_transform.basis.xform(Vector3(0, 0, 1))
+	var target_dir = (target_position - current_transform.origin).normalized()
+	var rotation_angle = acos(cur_dir.x) - acos(target_dir.x)
+	
+	state.set_angular_velocity(up_dir * (rotation_angle / state.get_step()))
+
+    func _integrate_forces(state):
+	
+	var target_position = $"my_target_spatial_node".get_global_transform().origin
+	look_follow(state, get_global_transform(), target_position)
+
+
+This function uses the rigid body's ``set_angular_velocity()`` method to rotate the body. It first calculates the difference between the current and desired angle and then adds the velocity needed to rotate by that amount in one frame's time.
+
+.. note:: This script will not work with rigid bodies in *character mode* because in this mode the body's rotation is locked. In this case, you would have to rotate the attached mesh node instead using the standard Spatial methods.
--- a/tutorials/physics/index.rst
+++ b/tutorials/physics/index.rst
@@ -6,5 +6,6 @@ Physics
    :name: toc-learn-features-physics
 
    physics_introduction
+   rigid_body
    ray-casting
    kinematic_character_2d
--- a/tutorials/physics/physics_introduction.rst
+++ b/tutorials/physics/physics_introduction.rst
@@ -45,7 +45,7 @@ The other three bodies extend from :ref:`PhysicsBody2D <class_PhysicsBody2D>`:
 - :ref:`RigidBody2D <class_RigidBody2D>`
     This is the node that implements simulated 2D physics. You do not control a
     ``RigidBody2D`` directly, but instead you apply forces to it (gravity, impulses,
-    etc.) and the physics engine calculates the resulting movement.
+    etc.) and the physics engine calculates the resulting movement. :ref:`Read more about using rigid bodies. <_doc_rigid_body>`
 
 - :ref:`KinematicBody2D <class_KinematicBody2D>`
     A body that provides collision detection, but no physics. All movement and


cbscribe:
I think there's quite a bit of overlap here with http://docs.godotengine.org/en/3.0/tutorials/physics/physics_introduction.html
which has an overview of the three PhysicsBody2D nodes.

AndreaCatania:
yes but this overview is not sufficient

cbscribe:
Agreed. I just mean it might be useful to link them together - from introduction to "further details" and yours can serve as the in-depth article about the body.

AndreaCatania:
correctly, let me edit it

AndreaCatania:
Is this correct to set a link? :ref:`Read more about rigid body <_doc_rigid_body>`

cbscribe:
Looks good. I can do a review of the language later today.

AndreaCatania:
I did it, Thanks you

tutorials/physics/rigid_body.rst
```diff
@@ -0,0 +1,51 @@
+.. _doc_rigid_body
+
+Rigid Body
+=========
+
+What is a rigid body?
+--------------------
+
+Rigid body is a rigid (non deformable) physical object that is directly controlled by physics engine in order to simulate phisical behaviour of objects.
```
cbscribe: You shouldn't define a term with the same term. It's not a physical object, it's a *simulation* of one.

A rigid body is one that is directly controlled by the physics engine in order to simulate the behavior of physical objects.

tutorials/physics/rigid_body.rst
```diff
@@ -0,0 +1,51 @@
...
+It's necessary to add one or more shapes that define the final shape of rigid body, you can set different position of these primitive shapes and this affect the COM (center of mass) of body.
```
cbscribe: In order to define the shape of the body, it must have one or more :ref:`Shape <class_Shape>`objects assigned. Note that setting the position of these shapes will affect the body's center of mass.

tutorials/physics/rigid_body.rst
```diff
@@ -0,0 +1,51 @@
+.. _doc_rigid_body
+
+Rigid Body
+=========
```
cbscribe: To prevent build warnings, the heading delimiters must match the length of the section title.

tutorials/physics/rigid_body.rst
```diff
@@ -0,0 +1,51 @@
...
+
+What is a rigid body?
+--------------------
```
cbscribe: Add one more -

tutorials/physics/rigid_body.rst
```diff
@@ -0,0 +1,51 @@
...
+
+Rigid body is a rigid (non deformable) physical object that is directly controlled by physics engine in order to simulate phisical behaviour of objects.
+It's necessary to add one or more shapes that define the final shape of rigid body, you can set different position of these primitive shapes and this affect the COM (center of mass) of body.
+
+How to control rigid body
+--------------------------
```
cbscribe: Controlling rigid bodies
-----------------------

tutorials/physics/rigid_body.rst
```diff
@@ -0,0 +1,51 @@
...
+
+To alter the rigid body behaviour you can set some parameters like damping, friction, mass, bounce, gravity scale, etc..
```
cbscribe: A rigid body's behavior can be altered by setting its properties such as friction, mass, bounce, etc. These properties can be set in the Inspector or via code. See :ref:`RigidBody <class_RigidBody>` for the full list of properties and their effects.

tutorials/physics/rigid_body.rst
```diff
@@ -0,0 +1,51 @@
...
+
+While to control the movement of the rigid body, you have many ways and you have to choose the correct one depending of the case.
```
cbscribe: There are several ways to control a rigid body's movement, depending on your desired application.

tutorials/physics/rigid_body.rst
```diff
@@ -0,0 +1,51 @@
...
+
+When you need to move a rigid body one time you can use all functions defined by *spatial node*, for example ``set_global_transform`` or ``look_at``. These functions are perfectly fine to create a teleport but must not be called each frame to move the rigid body because in this case the physics engine will not be able to handle it correctly and the physics simulation will not be correct.
```
cbscribe: If you only need to place a rigid body once, for example to set its initial location, you can use the methods provided by the :ref:`Spatial <class_Spatial>` node, such as ``set_global_transform()`` or ``look_at()``. However, these functions can not be called every frame or the physics engine will not be able to correctly simulate the body's state.

tutorials/physics/rigid_body.rst
```diff
@@ -0,0 +1,51 @@
...
+For example one common mistake is to use ``look_at`` each frame, this break physics similation. Few line below I'll show how to implement ``look_at`` correctly.
```
cbscribe: As an example, consider a rigid body that you want to rotate so that it points towards another object. A common mistake when implementing this kind of behavior is to use ``look_at()`` every frame, which breaks the physics simulation. Below, we'll demonstrate how to implement this correctly.

tutorials/physics/rigid_body.rst
```diff
@@ -0,0 +1,51 @@
...
+
+The fact that you can't use ``set_global_transform`` or ``look_at`` functions doesn't mean that you can't have full control of rigid body, rather by adding forces, impulse, or by setting velocity, in the ``_integrate_forces(state)`` you will be able to achieve any movement you desire, even if it's not physically possible!
```
cbscribe: The fact that you can't use ``set_global_transform()`` or ``look_at()`` methods doesn't mean that you can't have full control of a rigid body. Instead, you can control it by using the ``_integrate_forces()`` callback. In this function, you can add forces, apply, impulses, or set the velocity in order to achieve any movement you desire.

tutorials/physics/rigid_body.rst
```diff
@@ -0,0 +1,51 @@
...
+
+Look at function
+----------------
+
+One of the most required feature is the look_at function, this function is defined by spatial node and as I said before it can't be used in each frame to follow a target with sight.
```
cbscribe: As described above, using the Spatial node's ``look_at()`` function can't be used each frame to follow a target.

tutorials/physics/rigid_body.rst
```diff
@@ -0,0 +1,51 @@
...
+Even if this function *seems to work* on the old physics engine use it is not correct and is not intended to work. So here I've prepared a custom ``look_at`` that will work with rigid body.
```
cbscribe: <*I don't think it's necessary here to talk about the "old" physics engine. For most readers, they won't know what that means*>

Here is a custom ``look_at()`` function that will work reliably with rigid bodies:

tutorials/physics/rigid_body.rst
```diff
@@ -0,0 +1,51 @@
...
+
+.. tabs::
+ .. code-tab:: gdscript GDScript
+
+    extends RigidBody
+
+    func look_follow(state, current_transform, target_location):
+	var up_dir = Vector3(0,1,0);
```
cbscribe: `var up_dir = Vector3(0, 1, 0)`

tutorials/physics/rigid_body.rst
```diff
@@ -0,0 +1,51 @@
...
+	var cur_dir = current_transform.basis.xform(Vector3(0, 0, 1))
+	var target_dir = (target_location-current_transform.origin).normalized()
```
cbscribe: `var target_dir = (target_location - current_transform.origin).normalized()`

tutorials/physics/rigid_body.rst
```diff
@@ -0,0 +1,51 @@
...
+	var rotation_angle = acos(cur_dir.x) - acos(target_dir.x)
+	
+	state.set_angular_velocity(up_dir * (rotation_angle / state.get_step()))
+
+    func _integrate_forces(state):
+	
+	var target = $"../Bullets5/MonitoringArea".get_global_transform().origin
```
cbscribe: This is confusing. What are the `Bullets5` and `MonitoringArea` nodes? I think we need a description before the code describing the node setup for this demonstration.

tutorials/physics/rigid_body.rst
```diff
@@ -0,0 +1,51 @@
...
+	look_follow(state, get_global_transform(), target)
+
+
+As you can see this function ``look_follow`` internally use the function ``set_angular_velocity`` to rotate the rigid body, it basically take the difference of current orientation and target position and add the velocity required to rotate the rigid body within 1 frame.
```
cbscribe: This function uses the rigid body's ``set_angular_velocity()``method to rotate the body. It first calculates the difference between the current and desired angle and then adds the velocity needed to rotate by that amount in one frame's time.

tutorials/physics/rigid_body.rst
```diff
@@ -0,0 +1,51 @@
...
+
+*IMPORTANT* This script doesn't work with rigid body in *character mode* because when in this mode the rigid body is locked and can't be rotated. In this case you have to rotate the mesh node that are under the rigid body and since the mesh is not a rigid body you can use all functions defined by spatial node to move it (like: look_at, set_global_transform, etc...)
```
cbscribe: <*You can use Sphinx tags for info blocks*>

.. note:: This script will not work with rigid bodies in *character mode* because in this mode the body's rotation is locked. In this case, you would have to rotate the attached mesh node instead using the standard Spatial methods.

AndreaCatania:
Thanks you, I did it

mhilbrunner:
Looks good. Thank you for taking the time to write some documentation on this :)

Issue #1281, opened 2018-04-01
Add C# to GUI skinning page


skyace65:
The next page I'm going to add C# code to is the GUI skinning page.

PR #1282, opened 2018-04-01
Add C# to GUI Skinning page

skyace65:
Adds C# code to GUI Skinning page with Code Tabs. Closes issue #1281

### DIFF
--- a/tutorials/gui/gui_skinning.rst
+++ b/tutorials/gui/gui_skinning.rst
@@ -67,13 +67,22 @@ Every option is associated to:
 
 An example usage:
 
-::
+.. tabs::
+ .. code-tab:: gdscript GDScript
 
     var t = Theme.new()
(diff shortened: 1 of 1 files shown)


tutorials/gui/gui_skinning.rst
```diff
@@ -67,14 +67,23 @@ Every option is associated to:
...
 
     var t = Theme.new()
     t.set_color("font_color", "Label", Color(1.0, 1.0, 1.0))
 
     var l = Label.new()
     l.set_theme(t)
 
+ .. code-tab:: csharp
+
+    Theme t = new Theme();
```
mysticfall: I'd suggest to use `var` like ```var t = new Theme();``` as it's a more common practice and it looks similar to the GDScript counterpart. The same goes for the following lines.

skyace65:
Should be fixed now

PR #1285, opened 2018-04-02
Fix link

fahmyrm:
The link for "main menu tutorial" was pointing at the current page instead of the tutorial


fahmyrm:
Oops, it seems I did this wrong. Sorry.

PR #1297, opened 2018-04-04
Added C# examples & info for the Your First Game example

exts:

### DIFF
--- a/getting_started/step_by_step/your_first_game.rst
+++ b/getting_started/step_by_step/your_first_game.rst
@@ -419,7 +419,7 @@ Add this code to the function:
         // for the sake of this example, but it's better to create a class var
         // then assign the variable inside _Ready()
         var collisionShape2D = GetNode("CollisionShape2D") as CollisionShape2D;
-        if(collisionShape2D != null)
+        if (collisionShape2D != null)
         {
             collisionShape2D.Disabled = true;
         }
@@ -449,7 +449,7 @@ the player when starting a new game.
         Show();
         
         var collisionShape2D = GetNode("CollisionShape2D") as CollisionShape2D;
-        if(collisionShape2D != null)
+        if (collisionShape2D != null)
         {
             collisionShape2D.Disabled = false;
         }
--- a/getting_started/step_by_step/your_first_game.rst
+++ b/getting_started/step_by_step/your_first_game.rst
@@ -254,7 +254,7 @@ or ``false`` if it isn't.
             velocity.y -= 1;
         }
     
-        var animatedSprite = GetNode("AnimatedSprite") as AnimatedSprite;
+        var animatedSprite = (AnimatedSprite) GetNode("AnimatedSprite");
         if (velocity.Length() > 0) {
             velocity = velocity.Normalized() * Speed;
             animatedSprite.Play();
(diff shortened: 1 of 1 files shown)


mhilbrunner:
cc @mysticfall

getting_started/step_by_step/your_first_game.rst
```diff
@@ -140,35 +143,59 @@ node, so we'll add a script. Click the ``Player`` node and click the
...
     extends Area2D
 
     export (int) var SPEED  # how fast the player will move (pixels/sec)
     var screensize  # size of the game window
+ 
+ .. code-tab:: csharp
+
+    public class Player : Area2D
+    {
+        [Export] public int SPEED = 0;
```
mysticfall: To follow the coding convention they should be `Speed` and `screenSize` respective, but it seems their GDScript counterparts were inconsistent in naming at the first place.

Maybe we should change GDScript part as well?

getting_started/step_by_step/your_first_game.rst
```diff
@@ -206,6 +234,36 @@ or ``false`` if it isn't.
         else:
             $AnimatedSprite.stop()
 
+ .. code-tab:: csharp
+
+    public override void _Process(float delta)
+    {
+        var velocity = new Vector2();
+        if(Input.IsActionPressed("ui_right")) {
```
mysticfall: It's trivial, but I guess we used a space between `if` and `(` in other places, so I suppose it'd be better to keep the conding style consistent across the documentation.

getting_started/step_by_step/your_first_game.rst
```diff
@@ -206,6 +234,36 @@ or ``false`` if it isn't.
...
+        
+        if(Input.IsActionPressed("ui_down")) {
+            velocity.y += 1;
+        }
+
+        if(Input.IsActionPressed("ui_up")) {
+            velocity.y -= 1;
+        }
+
+        var animated_sprite = GetNode("AnimatedSprite") as AnimatedSprite;
```
mysticfall: Maybe we should use PascalCase here? And maybe we could also consider changing the safe cast to the direct cast, since we depend on an assumption that it wouldn't be null and has the right type in this case.

getting_started/step_by_step/your_first_game.rst
```diff
@@ -267,16 +336,32 @@ Let's place this code at the end of our ``_process()`` function:
...
 
-::
+.. tabs::
+ .. code-tab:: gdscript GDScript
     
     hide()
 
+ .. code-tab:: csharp
+
+    this.Hide();
```
mysticfall: I'm not too sure about this, but my IDE complains if I use `this.` when it's not an extention method call.

getting_started/step_by_step/your_first_game.rst
```diff
@@ -311,13 +401,27 @@ settings - Godot will automatically create a function called
...
+ .. code-tab:: gdscript GDScript
 
     func _on_Player_body_entered( body ):
         hide() # Player disappears after being hit
         emit_signal("hit")
         $CollisionShape2D.disabled = true
 
+ .. code-tab:: csharp
+
+    public void _on_Player_body_entered(Godot.Object body)
```
mysticfall: I think we don't really need to follow the GDScript naming convention here, but maybe it could make sense in signal callbacks. I'd like to hear others might think about this issue.

mysticfall:
Most of my comments are about coding style or naming convention issues and I only added a comment at the first occurrence when there are many similar cases.

Thanks much for the contribution! :)

mysticfall:
It looks good to me, except for changing safe casts to direct ones. I heard that @exts is working on it now, so we better wait for him to update the PR before merging it.

exts:
Alright, should be good to go.

mysticfall:
LGTM. Thanks again for the contribution! :+1:

mhilbrunner:
Thanks for your contribution :)

KellyThomas:
@exts If you have your project files that match this write up available you should consider putting them in https://github.com/godotengine/godot-demo-projects

PR #1315, opened 2018-04-05
Update custom_mouse_cursor.rst

guilhermefelipecgs:
We don't have restrictions about file format.


guilhermefelipecgs:
I'll try remove the size restriction

PR #1323, opened 2018-04-06
Update ui_introduction_to_the_ui_system.rst

GetLocalPlayer:
There's no TextureFrame class in Godot 3. I think it should be TextureRect.


mhilbrunner:
Thanks for your contribution!

PR #1325, opened 2018-04-07
FPS Tutorial Part 2 Revisions

drambique:

### DIFF
--- a/tutorials/3d/fps_tutorial/part_two.rst
+++ b/tutorials/3d/fps_tutorial/part_two.rst
@@ -22,7 +22,7 @@ Making a system to handle animations
 ------------------------------------
 
 First we need a way to handle changing animations. Open up ``Player.tscn`` and select the :ref:`AnimationPlayer <class_AnimationPlayer>`
-Node (``Player``->``Rotation_helper``->``Model``->``AnimationPlayer``).
+Node (``Player`` -> ``Rotation_helper`` -> ``Model`` -> ``AnimationPlayer``).
 
 Create a new script called ``AnimationPlayer_Manager.gd`` and attach that to the :ref:`AnimationPlayer <class_AnimationPlayer>`.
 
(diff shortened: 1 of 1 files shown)


drambique:
In comparison to the first part of the tutorial, this part is very hard to follow.

In the section about plotting points on the animation timeline to mark when the gun fires, in the project, the points already exists with the correct names.

There is no control mapping for switching weapons in this section so the weapons cannot be tested when this part of the tutorial is complete.

Copying the new physics process code to the project defines how the run key should work twice because it is already defined in the first part of the project under a different function.

In addition, I could not get the weapons to work after doing the tutorial.

cbscribe:
> I'm thinking of also changing all of the code to use $ instead of get_node, and enforcing a consistent naming style for variables and nodes, so everything is nice and consistent. I'm not sure if that will make things easier, but at least it should be a little less jarring.

I would advocate for this. Consistent code style in all the demos is important. We have a GDScript style guide and we've been slowly fixing all the old code snippets in the docs to conform. A lot of the demo repo has been cleaned up as well.

mhilbrunner:
Thanks for the review TwistedTwigleg, merging. And I agree with cbscribe above, consistent style is important :)

drambique:
> What parts are hard to follow? Any suggestions on how to make them easier?

The files provided don't match the tutorial, for example, the animation points already existing when the tutorial clearly wants the reader to make them themselves which leads to confusion.

It would also be a big help if there was a way to preserve tabs in the documentation to make it easier to copy and paste large sections of code to the editor.

TwistedTwigleg:
Thanks drambique! 

> The files provided don't match the tutorial, for example, the animation points already existing when the tutorial clearly wants the reader to make them themselves which leads to confusion.

I'll do my best to update the files provided to better follow the tutorial. I'm thinking about using the code from part 4 in parts 1 through 3 (and then removing part 4) so the code is both easier to understand and easier to maintain. While I'm doing that I'll try to better keep the ZIP files inline with the tutorial

> It would also be a big help if there was a way to preserve tabs in the documentation to make it easier to copy and paste large sections of code to the editor.

Yeah, I agree. There is a way in the editor to convert spaces to tabs, but I only stumbled upon it recently. Maybe a section on converting from spaces to tabs (and back again) would be helpful?

drambique:
> Maybe a section on converting from spaces to tabs (and back again) would be helpful?
Yes, I personally would find this very useful.

PR #1334, opened 2018-04-08
rotations tutorial

tagcup:
Initial WIP doc on rotations. Once this becomes available on docs site, people will hopefully read it and give some feedback about the parts that need clarifications and I can update it.

I'll eventually add transforms, including scale, polar decompostion and translation. I wanted to get rotatation part out because that's the real nontrivial part


cbscribe:
I haven't read through the doc yet, but one issue:

I don't think changing from `imgmath` to `mathjax` is a good idea, at least not without extensive testing. `mathjax` renders the formulas on the browser side using Javascript, which may break some offline doc generators. When we added formula support earlier this year, the thought was that generating the formula images at build time was a safer path. Better to have the overhead on the server side than the client side, imho.

tagcup:
I already expressed my views about it [here](https://github.com/godotengine/godot/issues/17951).

Mathjax is well tested and extensively used on physics journal websites for years. It also doesn't produce hundreds of blurry images.

I prefer to use mathjax only.

tagcup:
Mathjax works online just as fine, it'd just need to be bundled.

cbscribe:
If it's not going to break anything, then I don't have any objections. I just raise the issue because I know there are users generating offline docs in pdf and other formats.

> I prefer to use mathjax only.

Agreed we should only use one or the other.

tagcup:
Ah, I see, offline in that sense would definitely be a problem. Is it not possible to use mathjax for html builds and imgmath for others?

tagcup:
I mean it seems it's just a matter of changing a single word in conf.py

tagcup:
Actually, hold on, Sphinx should use latex for pdf builds and shouldn't need javascript/mathjax at all. Let me check

tagcup:
Well, as it turns out, the latexpdf build is already broken by a myriad of other issues.
But I don't think math is one of them. This is my guess (I couldn't test it) but I think for latexpdf, sphinx just emits the math equations as-is, and those are already in latex. latexpdf shouldn't have anything to do with mathjax or imgmath.

tagcup:
I just checked and this is indeed explicitly mentioned in the [docs](http://www.sphinx-doc.org/en/master/ext/math.html#module-sphinx.ext.mathjax):

> This is the de-facto standard for plain-text math notation and has the added advantage that no further translation is necessary when building LaTeX output.

tagcup:
> We should certainly merge it at some point, but I wouldn't include it in the online manual as-is.

Very well. I wish somebody told about this before I started writing it. Anyway, I'm not going to waste anymore time on this then.

NathanLovato:
I just said **I** wouldn't have it **as-is**, not that it shouldn't be merged at all. That's quite the contrary:

> You can at least remove the majority of adverbs and split many sentences in two or three. This would greatly improve readability without changing any of the meaning. Just the style.

All tutorials go through reviews and some proofing. As a professional writer myself I've always had someone else proof my work, be it for blogs, magazines or here on Godot's docs.

NathanLovato:
Note I didn't use the request changes feature either so that another reviewer could merge it if they didn't agree with me.

But I mean what I said: it could use a proofing pass to make it accessible to more readers.

Reviewers are also always available to answer questions and give reviews as you write a page.

mhilbrunner:
@tagcup As suggested above, if you are willing to recreate the PR, we could merge it and could have a look over it later, without you needing to *waste any more time on it*.

tagcup:
You also just said it's too technical etc etc. If you go back and read your original post you can see how it can be parsed as this requires major revision *and* simplifications.

Anyway. If someone wants to merge this, be my guest. I'm fine otherwise too. But I'm not going to spend anymore time on it.

NathanLovato:
> You also just said it's too technical etc etc.

I suggested to change:

> Just the style

Allow me to rephrase: you seem to be very comfortable with math. The writing style is that of someone who probably studied and/or works with math. But the content sure is interesting!

By splitting long sentences, removing nesting within sentences (parentheses), and removing redundant words it will become easier to read for everyone. The goal isn't to spoil your work or to change what you're teaching there.

Anyway merging now.

tagcup:
I have no experience teaching anything below undergrad level, so it probably shows. It's also not polished as I mentioned elsewhere before submitting it. Hopefully others in the community will pick up the slack and improve it.

I can also add explanations/clarifications if/when we get some feedback

NathanLovato:
Yeah no worries it's all good. I opened an issue to see if someone's up for the task and also for future discussion. Tagging you there so people can ask you questions if need be

PR #1337, opened 2018-04-08
Fix wrong command in command line tutorial.

karl-run:
Banged my head against the wall for a while because of this.


mhilbrunner:
Oh, you're right. Thanks for fixing it!

akien-mga:
Nice PR number :)

PR #1351, opened 2018-04-10
Minor changes to rotations docs.

tagcup:
Also removed the redundant SLERP discussion (which contained an awful typo: it used the symbols we introduced to described the 3D vectors and angles to describe
the 4D ones).


tagcup:
I'm removing the geometric SLERP discussion because it's not well described, and needs a geometric description of quaternions as points on 4D sphere with pictures to be self-contained (such that we can place these two 4D vectors on great circle, and rotate one vector onto the other while the angle of one goes onto the other linearly; but those vectors and angles need to be clearly described with new symbols!). It is quite tangential to the algebraic approach of the entire text

NathanLovato:
Thanks!

tagcup:
@mhilbrunner This, as well as #1356 should be cherrypicked, it contains important corrections to what is merged to 3.0 branch.

mhilbrunner:
These should already have been in cherrypicked into 3.0 in https://github.com/godotengine/godot-docs/commit/9332e20d01d2f60ecadd9a1ab64feac101e5b7c4 and https://github.com/godotengine/godot-docs/commit/258371816c175bf82a77470ee1755b83cffc0d34

tagcup:
@mhilbrunner Oh, OK, that's interesting because the changes don't show up in http://docs.godotengine.org/en/stable/tutorials/math/rotations.html Is this a problem related to Read the Docs then?

mhilbrunner:
@vnen Ah thanks, that makes sense.

PR #1362, opened 2018-04-14
Add C# code to KinematicBody

skyace65:
Adds C# code to the KinematicBody page. This PR is WIP because we're not really sure what the C# equivalent to `preload` is and one code example doesn't work without that.

### DIFF
--- a/tutorials/physics/using_kinematic_body_2d.rst
+++ b/tutorials/physics/using_kinematic_body_2d.rst
@@ -116,7 +116,14 @@ the same collision response:
 
  .. code-tab:: csharp
 
-    Coming soon
+    // using MoveAndCollide
+    var collision = MoveAndCollide(velocity * delta);
+    if (collision != null)
+    {
+        velocity = velocity.Slide(collision.Normal);
+    }
+    // using MoveAndSlide
+    velocity = MoveAndSlide(velocity);
 
 Anything you do with ``move_and_slide()`` can also be done with ``move_and_collide()``,
 it just might take a little more code. However, as we'll see in the examples below,
@@ -170,7 +177,42 @@ Attach a script to the KinematicBody2D and add the following code:
 
  .. code-tab:: csharp
 
-    Coming soon
+    using Godot;
+    using System;
+
+    public class KBExample : KinematicBody2D
+    { 
+        public int Speed = 250;
+        private Vector2 _velocity = new Vector2();
+
+        public void getInput()
+        {
+            // Detect up/down/left/right keystate and only move when pressed
+            _velocity = new Vector2();
+            if (Input.IsActionPressed("ui_right"))
+            {
+                _velocity.x += 1;
+            }
+            if (Input.IsActionPressed("ui_left"))
+            {
+                _velocity.x -= 1;
+            }
+            if (Input.IsActionPressed("ui_down"))
+            {
+                _velocity.y += 1;
+            }
+            if (Input.IsActionPressed("ui_up"))
+            {
+                _velocity.y -= 1;
+            }
+        }
+
+        public override void _PhysicsProcess(float delta)
+        {
+            getInput();
+            MoveAndCollide(velocity * delta);
+        }
+    }
 
 
 Run this scene and you'll see that ``move_and_collide()`` works as expected, moving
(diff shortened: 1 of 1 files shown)


karroffel:
@skyace65 `preload()` is a thing in GDScript that's deeply integrated into the parser of the language, there's no way to do that in other languages not made especially for/in Godot.

So instead of loading the resource at GDScript parse time, in C# you should use `ResourceLoader.Load()` instead. (I think, it's the same with GDNative, I assume that that's the same for C#)

KellyThomas:
You should be able to use this to load the bullet scene:

    PackedScene bullet = (PackedScene) GD.Load("res://Bullet.tscn");

skyace65:
@mysticfall The PR is finished, if you could review this that would be great!

tutorials/physics/using_kinematic_body_2d.rst
```diff
@@ -170,7 +177,42 @@ Attach a script to the KinematicBody2D and add the following code:
...
  .. code-tab:: csharp
 
-    Coming soon
+    using Godot;
+    using System;
+
+    public class KBExample : KinematicBody2D
+    { 
+        int speed = 250;
+        Vector2 velocity = new Vector2();
```
mysticfall: Probably we should make `speed` a public property, and `velocity` a private field (with respective naming changes).

On a side note, I think declaring `velocity` as a member variable is sort of a bad practice. But I can see how it's duplicated from the GDScript counter part, so probably it needs to be dealt with a separate PR if we ever fix it.

tutorials/physics/using_kinematic_body_2d.rst
```diff
@@ -272,7 +360,39 @@ And the code for the Bullet:
...
+            {
+                velocity = velocity.Bounce(collsion.Normal);
+                if (collsion.Collider.HasMethod("Hit"))
+                {
+                    collsion.Collider.Hit();
+                }
+            }
+        }
+
+        public void onVisibilityNotifier2DScreenExited()
```
mysticfall: According to a discussion in #1365, it seems that we decided to follow PascalCase for signal callbacks. So, I believe we should treat it like any other methods regarding the naming convention, making it start with a upper case letter.

tutorials/physics/using_kinematic_body_2d.rst
```diff
@@ -337,7 +457,51 @@ Here's the code for the player body:
 
  .. code-tab:: csharp
 
-    Coming soon
+    using Godot;
+    using System;
+
+    public class KBExample : KinematicBody2D
+    { 
+        [Export] int runSpeed = 100;
```
mysticfall: I guess you missed `public` here :)

mysticfall:
@skyace65 Made a few comments regarding some inconsistencies I could find. Thanks for the heads up :)

skyace65:
@mysticfall Updated the PR, can you check that I made the changes properly?

tutorials/physics/using_kinematic_body_2d.rst
```diff
@@ -170,7 +177,42 @@ Attach a script to the KinematicBody2D and add the following code:
...
  .. code-tab:: csharp
 
-    Coming soon
+    using Godot;
+    using System;
+
+    public class KBExample : KinematicBody2D
+    { 
+        public int Speed = 250;
+        private Vector2 velocity = new Vector2();
```
mysticfall: Could you change it to `_velocity`? We recommend all private fields to be in camelCase with `_` prefix now.

tutorials/physics/using_kinematic_body_2d.rst
```diff
@@ -242,7 +284,53 @@ uses the mouse pointer. Here is the code for the Player, using ``move_and_slide(
 
  .. code-tab:: csharp
 
-    Coming soon
+    using Godot;
+    using System;
+
+    public class KBExample : KinematicBody2D
+    { 
+        PackedScene bullet = (PackedScene)GD.Load("res://Bullet.tscn");
```
mysticfall: It's better to make access modifiers explicit, and in this case I guess we better use `private PackedScene _bullet` instead.

And the same goes for `public int Speed`, and `private Vector2 _velocity`.

tutorials/physics/using_kinematic_body_2d.rst
```diff
@@ -272,7 +360,39 @@ And the code for the Bullet:
 
  .. code-tab:: csharp
 
-    Coming soon
+    using Godot;
+    using System;
+
+    public class Bullet : KinematicBody2D
+    { 
+        int speed = 750;
```
mysticfall: Same as above :)

tutorials/physics/using_kinematic_body_2d.rst
```diff
@@ -337,7 +457,51 @@ Here's the code for the player body:
 
  .. code-tab:: csharp
 
-    Coming soon
+    using Godot;
+    using System;
+
+    public class KBExample : KinematicBody2D
+    { 
+        [Export] public int runSpeed = 100;
```
mysticfall: public properties/fields need to be in PascalCase.

mysticfall:
I only commented on the first occurrence of a similar issue, but it seems it has caused some confusion. I updated my comments to mark all such occurrences now. Sorry!

skyace65:
Should be fixed now

mysticfall:
It looks good to me. Thanks much, as always :)

skyace65:
Thanks for your review :) @mhilbrunner feel free to merge whenever.

cbscribe:
Thanks!

PR #1377, opened 2018-04-19
fix transform_camera image not showing up

acfaruk:


acfaruk:
hmm.. it seems like the fix didn't work on the [hosted docs?](
http://docs.godotengine.org/en/latest/tutorials/3d/using_transforms.html#introducing-transforms)

Seems to work fine on the [GitHub Preview](https://github.com/godotengine/godot-docs/blob/master/tutorials/3d/using_transforms.rst#introducing-transforms) though... Locally it was working too.

mhilbrunner:
Works now, was probably not synced/updated yet or still cached. :)

PR #1381, opened 2018-04-20
fix typos in discussion of c sharp signals

KellyThomas:


paulloz:
👌 
Maybe we can take the opportunity of this PR to also fix the double `;` in *Using the AS operator*?

KellyThomas:
done

PR #1382, opened 2018-04-20
c sharp samples for autoloads

KellyThomas:

### DIFF
--- a/getting_started/step_by_step/singletons_autoload.rst
+++ b/getting_started/step_by_step/singletons_autoload.rst
@@ -46,16 +46,28 @@ are loaded.
 
 This means that any node can access a singleton named "playervariables" with:
 
-::
+.. tabs::
+ .. code-tab:: gdscript GDScript
 
    var player_vars = get_node("/root/playervariables")
    player_vars.health
 
+ .. code-tab:: csharp
+ 
+    var playerVariables = (PlayerVariables)GetNode("/root/PlayerVariables");
+    playerVariables.Health -= 10; // instance field
+
 Or even simpler using the name directly:
 
-::
+.. tabs::
+ .. code-tab:: gdscript GDScript
 
    playervariables.health
+
+ .. code-tab:: csharp
+
+    // Static members can be accessed by using the class name
+    PlayerVariables.Health -= 10;
 
 Custom scene switcher
 ---------------------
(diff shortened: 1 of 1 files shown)


getting_started/step_by_step/singletons_autoload.rst
```diff
@@ -151,6 +181,40 @@ current scene and replaces it with the requested one.
...
+        // This function will usually be called from a signal callback,
+        // or some other function from the running scene.
+        // Deleting the current scene at this point might be
+        // a bad idea, because it may be inside of a callback or function of it.
+        // The worst case will be a crash or unexpected behavior.
+
+        // The way around this is deferring the load to a later time, when
+        // it is ensured that no code from the current scene is running:
+
+        CallDeferred("DeferredGotoScene", path);
```
mysticfall: It's not mandatory, but probably we could consider using `nameof(DeferredGotoScene)` here, since it's a good practice which is still supported in C# 6.

getting_started/step_by_step/singletons_autoload.rst
```diff
@@ -114,10 +127,27 @@ loaded!
...
+    using Godot;
+    using System;
+
+    public class Global : Godot.Node
+    {
+        public Node CurrentScene { get; set; }
+
+        public override void _Ready()
+        {
+            var root = GetTree().GetRoot();
```
mysticfall: I think it's better to use `SceneTree root` since the type info is not immediately available from the right side of the assignment.

getting_started/step_by_step/singletons_autoload.rst
```diff
@@ -151,6 +181,40 @@ current scene and replaces it with the requested one.
...
+        CallDeferred("DeferredGotoScene", path);
+    }
+
+    public void DeferredGotoScene(string path)
+    {
+        // Immediately free the current scene, there is no risk here.
+        CurrentScene.Free();
+
+        // Load a new scene
+        var nextScene = (PackedScene) GD.Load(path);
```
mysticfall: It's just a tiny little nitpick, but it seems you didn't use a space between a cast and its associated identifier in your another PR, which conforms to the style of the engine source. So I guess it'd be better to keep it consistent here too.

KellyThomas:
Thanks @mysticfall, these items have been addressed.

mysticfall:
@KellyThomas Could you remove spaces after casts in L57, L243, and L263 also? Other than that, it looks good to me :)

KellyThomas:
Will do.

KellyThomas:
OK, those last three should be taken care of now.

mysticfall:
Looks good to me. Thanks much, as always! :)

Issue #1296, opened 2018-04-04
Create official Godot C# style guide


mhilbrunner:

paulloz:
This is some coding conventions from what I saw in the current documentation.

* Public fields, properties, methods should be named using `UpperCamelCase`
* Private fields, properties, methods and local variables should be named using `lowerCamelCase`
* Opening and closing brackets should be on their own lines
* Brackets can be spared if there's only one line inside the `if`, `while`, `for` etc.
* `this.` should be avoided if possible (I don't like this one but that's how everything is written right now)
* Local variables should be declared with `var` and not the actual type (same as above if you ask me)

KellyThomas:
At the moment the demo projects have signal handlers named `OnSignal()` and `OnClassSignal()`.  Once we have decided on a name scheme we should apply it consistently in this project and integrate it into the Connect Signal dialogue.

https://github.com/godotengine/godot-demo-projects/blob/master/mono/monkey_pong/Wall.cs#L6

    public void OnWallAreaEntered(Area2D area)

https://github.com/godotengine/godot-demo-projects/blob/master/mono/monkey_pong/Paddle.cs#L26

    public void OnAreaEntered(Area2D area)

neikeq:
We should this naming convention: https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members

> * Private fields, properties, methods and local variables should be named using `lowerCamelCase`

- Properties and methods should be named using `PascalCasel` regardless of their accessibility level.
- Locals should use `camelCasel`.
- Private and internal fields are not mentioned in the guidelines, so I would be fine with `camelCasel` or either `_camelCasel` or `_PascalCasel` (both with an underscore prefix). Use properties instead of public or protected fields.

> * Local variables should be declared with `var` and not the actual type (same as above if you ask me)

Only where it makes the code more readable. e.g.:

``` C#
// ACCEPTABLE?
var node = GetNode("..."); // Node node = GetNode("...");
// OK
var floor = new StaticBody();
var sprite = GetNode("...") as Sprite;
foreach (var child in children) { ... }
```

``` C#
// NOT GOOD
var count = 10; // int count = 10;
var result = task.Result; // Whatever result = task.Result;
```

mysticfall:
[...]

PR #1383, opened 2018-04-20
c sharp samples for custom GUI controls

KellyThomas:

### DIFF
--- a/tutorials/gui/custom_gui_controls.rst
+++ b/tutorials/gui/custom_gui_controls.rst
@@ -42,13 +42,28 @@ indicate that this is the currently focused control. To check for this
 status, the :ref:`Control.has_focus() <class_Control_has_focus>` method
 exists. Example
 
-::
+.. tabs::
+ .. code-tab:: gdscript GDScript
 
     func _draw():
         if has_focus():
              draw_selected()
         else:
              draw_normal()
+
+ .. code-tab:: csharp
+
+    public override void _Draw()
+    {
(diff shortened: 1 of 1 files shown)


tutorials/gui/custom_gui_controls.rst
```diff
@@ -100,14 +131,39 @@ This function is
...
+            {
+                GD.Print("Left mouse button was pressed!");
+            }
+        }
+    }
+
+    // or if using c# 7 or greater
+    public override void _GuiInput(InputEvent @event)
+    {
+        if (@event is InputEventMouseButton mbe && mbe.ButtonIndex == (int)ButtonList.Left && mbe.Pressed)
```
mysticfall: Is it alright to use C# 7 feature (pattern matching, in this case) in samples? I remember you recommended stay with C# 6 in engine sources (https://github.com/godotengine/godot-docs/issues/1296#issuecomment-379631679) but if it's ok with samples, I guess we can also follow the same pattern in L150.

KellyThomas: Here there are two implementations L145-155 are code that should be compatible with all c# versions.

Then L157-165 show how it could be approached if pattern matching is available.

If this is unclear I will be more explicit in the description .

mysticfall: No, it was my fault that I totally missed that comment. Sorry, shouldn't have done any code reading right after I wake up in the morning :blush:

PR #1394, opened 2018-04-23
#1296: Create official Godot C# style guide

mysticfall:
Initial draft of the official C# style guide (#1296).

### DIFF
--- a/getting_started/scripting/c_sharp/c_sharp_style_guide.rst
+++ b/getting_started/scripting/c_sharp/c_sharp_style_guide.rst
@@ -1,0 +1,305 @@
+.. _doc_c_sharp_styleguide:
+
+Style Guide
+===========
+
+Having well-defined and consistent coding conventions is important for every project, and Godot
+is no exception to this rule.
+
+This page contains a coding style guide which is followed by developers and contributors of Godot
+itself. As such, it is mainly intended for those who want to contribute to the project, but since
+the conventions and guidelines mentioned in this article are those most widely adopted by the users
+of the language, we encourage you to do the same, especially if you do not have such a guide yet.
+
+.. note:: This article is by no means an exhaustive guide on how to follow the standard coding
+        conventions or best practices. If you feel unsure of an aspect which is not covered here,
+        please refer to more comprehensive documentation, such as
+        `C# Coding Conventions <https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions>`_ or
+        `Framework Design Guidelines <https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/naming-guidelines>`_.
+
+Language Specification
+----------------------
+
+Currently, Godot uses C# version 6.0 in its engine and example source code. So, before we move to
+a newer version, care must be taken to avoid mixing language features only available in C# 7.0 or
+later, such as pattern matching or expression-bodied members inside get/set accessors.
+
+For detailed information of C# features in different versions, please see
+`What's New in C# <https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/>`_.
+
+Formatting Conventions
+----------------------
+
+* If you create a new file, make sure that it uses linefeed (*LF*) characters to break lines, not *CRLF* or *CR*.
+* Use UTF-8 encoding without a byte order mark (BOM <https://en.wikipedia.org/wiki/Byte_order_mark>).
+* Use 4 spaces instead of tabs for indentation (which is referred to as 'soft tabs').
+
+Line Breaks and Blank Lines
+---------------------------
+
+For a general indentation rule, follow `The 'Allman Style' <https://en.wikipedia.org/wiki/Indentation_style#Allman_style>`_
+which recommends placing the brace associated with a control statement on the next line, indented to
+the same level:
+
+.. code-block:: csharp
+
+    // Use this style:
+    if (x > 0)
+    {
+        DoSomething();
+    }
+
+    // NOT this:
+    if (x > 0) {
+        DoSomething();
+    }
+
+However, you may choose to omit line breaks inside brackets,
+
+* For simple property accessors.
+* For simple object, array, or collection initializers.
+* For abstract auto property, indexer, or event declarations.
+
+.. code-block:: csharp
+
+    // You may put the brackets in a single line in following cases:
+    public interface MyInterface
+    {
+        int MyProperty { get; set; }
+    }
+
+    public class MyClass : ParentClass
+    {
+        public int Value
+        {
+            get { return 0; }
+            set
+            {
+                ArrayValue = new [] {value};
+            }
+        }
+    }
+
+Insert a blank line,
+
+* After *using* statement list.
+* Between method, properties, and inner type declarations.
+
+Field and constant declarations can be grouped together according to relevance. In that case, consider
+inserting a blank line between the groups for easier reading.
+
+Avoid inserting a blank line,
+
+* After an opening bracket ('{').
+* Before a closing bracket ('}').
+* After a comment block, or a single line comment.
+* Adjacent to another blank line.
+
+.. code-block:: csharp
+
+    using System;
+    using Godot;
+                                                  // Blank line after using list.
+    public class MyClass
+    {                                             // No blank line after '{'.
+        public enum MyEnum
+        {
+            Value,
+            AnotherValue                          // No blank line before '}'.
+        }
+                                                  // Blank line around inner types.
+        public const int SomeConstant = 1;
+        public const int AnotherConstant = 2;
+
+        private Vector3 _x;
+        private Vector3 _y;                       // Related constants or fields can be
+                                                  // grouped together.
+        private float _width;
+        private float _height;
+
+        public int MyProperty { get; set; }
+                                                  // Blank line around properties.
+        public void MyMethod()
+        {
+            // Some comment.
+            AnotherMethod();                      // No blank line after a comment.
+        }
+                                                  // Blank line around methods.
+        public void AnotherMethod()
+        {
+        }
+    }
+
+Consider breaking a line when it's longer than 100 characters. And it's also a good practice to
+insert a line feed (LF) character at the end of a file because some utilities have trouble
+recognizing the last line without it (i.e. Linux's *cat* command).
+
+Using Spaces
+------------
+
+Insert a space,
+
+* Around a binary and tertiary operator.
+* Between an opening parenthesis and *if*, *for*, *foreach*, *catch*, *while*, *lock* or *using* keywords.
+* Before and within a single line accessor block.
+* Between accessors in a single line accessor block.
+* After a comma.
+* After a semi-colon in a *for* statement.
+* After a colon in a single line *case* statement.
+* Around a colon in a type declaration.
+* Around a lambda arrow.
+* After a single line comment symbol ('//'), and before it if used at the end of a line.
+
+Do not use a space,
+
+* After a type cast parentheses.
+* Within single line initializer braces.
+
+The following example shows a proper use of spaces, according to some of the the above mentioned conventions:
+
+.. code-block:: csharp
+
+    public class MyClass<A, B> : Parent<A, B>
+    {
+        public float MyProperty { get; set; }
+
+        public float AnotherProperty
+        {
+            get { return MyProperty; }
+        }
+
+        public void MyMethod()
+        {
+            int[] values = {1, 2, 3, 4}; // No space within initializer brackets.
+            int sum = 0;
+
+            // Single line comment.
+            for (int i = 0; i < values.Length; i++)
+            {
+                switch (i)
+                {
+                    case 3: return;
+                    default:
+                        sum += i > 2 ? 0 : 1;
+                        break;
+                }
+            }
+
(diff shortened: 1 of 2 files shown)


getting_started/scripting/c_sharp/c_sharp_style_guide.rst
```diff
@@ -0,0 +1,298 @@
+.. _doc_c_sharp:
```
mhilbrunner: better use _doc_c_sharp_styleguide or something similar

getting_started/scripting/c_sharp/c_sharp_style_guide.rst
```diff
@@ -0,0 +1,298 @@
...
+a newer version, care must be taken to avoid mixing language features only available in C# 7.0 or
+later, such as pattern matching or expression-bodied members inside get/set accessors.
+
+For detailed information of C# features in different versions, please see
+`What's New in C# <https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/>`_.
+
+Formatting Conventions
+----------------------
+
+* If you create a new file, make sure that it uses linefeed(*LF*) characters to break lines, not *CRLF* or *CR*.
```
mhilbrunner: missing space before the opening (

getting_started/scripting/c_sharp/c_sharp_style_guide.rst
```diff
@@ -0,0 +1,298 @@
...
+* Use UTF-8 encoding, if your code contains non-ASCII characters,
```
mhilbrunner: UTF-8 *without byte order mark (BOM)*, *always* (drop the non-ASCII characters condition, IMO)

getting_started/scripting/c_sharp/c_sharp_style_guide.rst
```diff
@@ -0,0 +1,298 @@
...
+            AnotherMethod();                      // No blank line after a comment.
+        }
+                                                  // Blank line around methods.
+        public void AnotherMethod()
+        {
+        }
+    }
+
+Consider breaking a line when it's longer than 120 characters. And it's also a good practice to
+insert a line feed(LF) character at the end of a file because some utilities have trouble
```
mhilbrunner: missing space before (

getting_started/scripting/c_sharp/c_sharp_style_guide.rst
```diff
@@ -0,0 +1,298 @@
...
+Rather than,
+
+.. code-block:: csharp
+
+    FindNode()?.Change(wpnDmg);
+
+Implicitly Typed Local Variables
+--------------------------------
+
+Consider using implicitly typing(*'var'*) for declaration of a local variable, but do so
```
mhilbrunner: missing space before (

getting_started/scripting/c_sharp/c_sharp_style_guide.rst
```diff
@@ -0,0 +1,298 @@
...
+        }
+    }
+
+Insert a blank line,
+
+* After *using* statement list.
+* Between method, properties, and inner type declarations.
+
+Field and constant declarations can be grouped together according to relevance. In that case, consider
+inserting a blank line between the groups for easier reading.
```
KellyThomas: Should we suggest a blank line between static and instance fields?

KellyThomas:
Should we be taking a stance on non-private fields?

neikeq:
@KellyThomas I commented on about this on #1296

> Use properties instead of public or protected instance fields.

getting_started/scripting/c_sharp/c_sharp_style_guide.rst
```diff
@@ -0,0 +1,298 @@
...
+            }
+
+            i += (int)MyProperty; // No space after a type cast.
+        }
+    }
+
+Naming Conventions
+------------------
+
+Use upper camel case (or *'Pascal case'*) for all namespaces, type names and member level identifiers
```
aaronfranke: It's best when the name itself is visually descriptive. I suggest that we write it as `PascalCase`

getting_started/scripting/c_sharp/c_sharp_style_guide.rst
```diff
@@ -0,0 +1,298 @@
...
+(i.e. methods, properties, constants, events) except for private fields:
```
aaronfranke: Comma before "except"

mysticfall:
Updated the PR with all the suggestions given above. Thanks :)

getting_started/scripting/c_sharp/c_sharp_style_guide.rst
```diff
@@ -0,0 +1,299 @@
...
+later, such as pattern matching or expression-bodied members inside get/set accessors.
+
+For detailed information of C# features in different versions, please see
+`What's New in C# <https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/>`_.
+
+Formatting Conventions
+----------------------
+
+* If you create a new file, make sure that it uses linefeed (*LF*) characters to break lines, not *CRLF* or *CR*.
+* Use UTF-8 encoding without a byte order mark (BOM <https://en.wikipedia.org/wiki/Byte_order_mark>),
```
aaronfranke: Change comma to a period.

getting_started/scripting/c_sharp/c_sharp_style_guide.rst
```diff
@@ -0,0 +1,299 @@
...
+
+    for (var i = 0; i < 10; i++)
+    {
+    }
+
+Other Considerations
+--------------------
+
+ * Use explicit access modifiers.
+ * Use properties instead of non-private fields.
```
aaronfranke: Why? And does this mean we should change the fields such as `real_t x` in `Vector2` to a property?

I don't really see any reason `public real_t x;` is bad. It should always be changeable from anywhere, right?

mysticfall: @aaronfranke I believe it's an old convention shared with Java devs, and the most important reason could be that if you expose something as a property, you can always change the underlying implementation (i.e. adding validation or returning a calculated value) later without affecting existing users.

In that particular case with `Vector2`, I agree that using public fields won't likely to cause any practical trouble.

But I don't see any reason not use properties in that case either, so I think it's worth to include it in the guide. If we don't want to change exising code in `Vector2` and other similar cases, maybe we can change the phrase to read it more like a recommendation rather than a hard and fast rule.

getting_started/scripting/c_sharp/c_sharp_style_guide.rst
mysticfall: Oops, I'll edit it in my next PR update.

getting_started/scripting/c_sharp/c_sharp_style_guide.rst
mysticfall: I just found a relevant comment in the C# guide from MS : 
> Generally, you should use fields only for variables that have private or protected accessibility. Data that your class exposes to client code should be provided through methods, properties and indexers. By using these constructs for indirect access to internal fields, you can guard against invalid input values.

* https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/fields

On a side note, it's interesting to see that it makes an exception for protected fields also. Personally, I've always assumed protected members to be a part of the API reference so I wonder if the practice differs among C# developers.

aaronfranke: Typically wouldn't you know in advance if the field/property would need the ability to have a differently coded implementation? IMO it'd be overcomplicated/possibly confusing to use properties everywhere, even when a field would do.

mysticfall: In general, yes. But probably we shouldn't shut out such possibilities of replacing such fields with some native methods, or adding extra codes like cache flushing or validation to them later. 

It also allows extracting an interface later, and can make the code more consistent when there are a few of such members that need to be restricted for update.

I don't have too strong an opinion about changing existing fields in such types like `Vector2`, but I think it's a good idea to include it among other recommendations.

aaronfranke: How about we make the text "Consider using ..."?

mysticfall: It sounds reasonable to me, especially if we are not going to change all such instances in existing code. I'd like to hear other people's thoughts before proceeding though.

mhilbrunner: "Prefer properties over public fields."?

KellyThomas: I think that there is value in having a consistent API.

Once we have finished discussion on this style guide I hope that it is applied consistently across the public API before 3.1.

With that in mind we will have to revisit the example `real_t x`, at least to change it to `real_t X`.  It doesn't seem like much more work to make it `real_t X { get; set; }` while we are at it.

The public API should be binding once final as we won't want to break people's projects.  This is a classic use case for properties over fields.

neikeq: No, use properties instead of public or protected instance fields. The reasons for using properties over fields are all over the internet, the main one being encapsulation.

Now, if you have a good reason or no other choice, go on. There are exceptions for every rule.

aaronfranke: Noteworthy: Never using public fields means that `camelCase` without underscores will only be used for local variables and method arguments.

mysticfall: So, should I change the relevant line ("Use properties instead of non-private fields.") to a more suggestive form, or I better leave it as it is?

I'll update the PR as soon as it becomes clear on this one.

mhilbrunner: @mysticfall I'd let neikeq make the call, so we can leave as is or change to the more explicit "Use properties instead of public or protected instance fields." instead. Shall I merge this, then? We can keep editing it forever anyway, but it's now been up for a week to review and all interested parties seem to have had a glance at it.

aaronfranke: I would absolutely be fine with merging as-is.

getting_started/scripting/c_sharp/c_sharp_style_guide.rst
```diff
@@ -0,0 +1,299 @@
...
+ * Use modifiers in this order: *'public/protected/private/internal virtual/override/abstract/new static readonly'*.
+ * Avoid using fully qualified names or *'this.'* prefix for members when it's not necessary.
+ * Remove unused *'using'* statements and unnecessary parentheses.
+ * Consider omitting default initial value for a type.
+ * Consider using null-conditional operators or type initializers to make the code more compact.
+ * Use safe cast when you expect a *'null'* value, and use direct cast otherwise.
```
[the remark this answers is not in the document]
mysticfall: Sorry, I must have been low on caffeine when I wrote that :sweat_smile: Updated with a proper instruction.

getting_started/scripting/c_sharp/c_sharp_style_guide.rst
```diff
@@ -0,0 +1,299 @@
...
+
+    var value = (int)speed;
+
+    // But not for these:
+
+    var value = GetValue();
+
+    var speed = 1;
+
+    var text = "Some value";
```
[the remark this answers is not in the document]
mysticfall: I agree. And as it seems that the only differences to the MS guide are variables inside a for-loop and numeric literals, I also included the former while retaining the latter with a bit of an explanation.

Please let me know if anyone want to change it further.

getting_started/scripting/c_sharp/c_sharp_style_guide.rst
```diff
@@ -0,0 +1,299 @@
...
+            // Some comment.
+            AnotherMethod();                      // No blank line after a comment.
+        }
+                                                  // Blank line around methods.
+        public void AnotherMethod()
+        {
+        }
+    }
+
+Consider breaking a line when it's longer than 120 characters. And it's also a good practice to
```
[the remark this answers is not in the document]
mysticfall: As it's only a generic suggestion, I just used the default value of the IDE I'm using. Now I changed it to 100, so people can choose either of them.

neikeq:
Thanks!

getting_started/scripting/c_sharp/c_sharp_style_guide.rst
```diff
@@ -0,0 +1,299 @@
...
+
+For detailed information of C# features in different versions, please see
+`What's New in C# <https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/>`_.
+
+Formatting Conventions
+----------------------
+
+* If you create a new file, make sure that it uses linefeed (*LF*) characters to break lines, not *CRLF* or *CR*.
+* Use UTF-8 encoding without a byte order mark (BOM <https://en.wikipedia.org/wiki/Byte_order_mark>),
+* Use 4-character indents with spaces instead of tabs (which is sometimes called as 'soft tabs').
```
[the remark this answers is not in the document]
Alumniminium: * Use tabs instead of 4-character indents with spaces (which is sometimes called correct).

PR #1396, opened 2018-04-24
FPS/Part1: Fix missing param in move_and_slide

Noshyaar:
This only affects master so it shouldn't be cherrypicked.


mhilbrunner:
cc @TwistedTwigleg

TwistedTwigleg:
Looks good!

All of the zip files going up from part 1 will also need updating, since the move_and_slide is present in all of them, but that can be a separate PR if needed.

mhilbrunner:
Will merge this then. Thanks!

PR #1399, opened 2018-04-24
Update class_tween.rst

rebertmatheus:


cbscribe:
Thanks for your contribution,  but class reference documents are built automatically from the XML files output by the engine, and must be edited in the main repo. For example, Tween is found here: https://github.com/godotengine/godot/blob/master/doc/classes/Tween.xml

Any changes made here will be overwritten.

See the following link for instructions on how to properly contribute to the API documentation:
http://docs.godotengine.org/en/latest/community/contributing/updating_the_class_reference.html

PR #1406, opened 2018-04-28
Fix extensions on singletons autoload code

Sliii:
http://docs.godotengine.org/en/3.0/getting_started/step_by_step/singletons_autoload.html had provided code referencing .tscn files, but it should be .scn to match the contents of autoload.zip provided on that page.

This is my first GitHub pull request and I have no idea what I'm doing.


cbscribe:
Instead of committing just your change to your branch, you pulled a bunch of changes from master into your branch and committed them too. See the list of commits above.

What you need to do is *only* commit the change to the one file you modified.

Please read this to understand the correct workflow for committing: http://docs.godotengine.org/en/latest/community/contributing/pr_workflow.html

That said, the whole singleton document also needs rewriting as it uses outdated screenshots, and using .scn isn't recommended.

Sliii:
Didn't see there was a page about pull requests, just clicked the "Edit Page" link at the top. Thank you for the information!

PR #1420, opened 2018-05-03
Update singletons_autoload.rst

ivancek:


ivancek:
Just a quick warning that the scene files provided in the .zip are .scn not .tscn so the scene load results in null

PR #1438, opened 2018-05-11
fixing Arc polygon function example for GDScript

mani144:
GDScript fixing an error in the example: Arc polygon function there is missing ")" on deg2rad.


akien-mga:
Thanks, looks good.

PR #1441, opened 2018-05-12
Few fixes in C# documentation

Chaosus:

### DIFF
--- a/getting_started/scripting/c_sharp/c_sharp_differences.rst
+++ b/getting_started/scripting/c_sharp/c_sharp_differences.rst
@@ -27,7 +27,7 @@ Math
 
 Math functions like ``abs``, ``acos``, ``asin``, ``atan`` and ``atan2`` are
 located under ``Mathf`` instead of in global scope.
-``PI`` is ``Mathf.PI``
+``PI`` is ``Mathf.Pi``
 
 Random
 ------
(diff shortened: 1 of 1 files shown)


getting_started/scripting/c_sharp/c_sharp_differences.rst
```diff
@@ -223,11 +223,10 @@ Other differences:
 GDScript          C#
 ================  ==================================================================
 Color8            Color.Color8
-is_inf            float.IsInfinity
-is_nan            float.IsNaN
```
neikeq: Why remove this?

getting_started/scripting/c_sharp/c_sharp_differences.rst
```diff
@@ -223,11 +223,10 @@ Other differences:
...
 dict2inst         ? TODO
 inst2dict         ? TODO
 load              GD.load which is the same as ResourceLoader.load
+yield             await ToSignal(object, signal)
```
neikeq: The left column should be "yield on signal", since yield can also be used in GDScript to simply suspend a function to manually resume it.
I would add some description to the right part instead of just a code example. At least, to make it clear that it's not exactly the same as yield on signal in the way they work internally.

getting_started/scripting/c_sharp/c_sharp_differences.rst
Chaosus: Mathf have them now

getting_started/scripting/c_sharp/c_sharp_differences.rst
Chaosus: What description ? Say - I will add

getting_started/scripting/c_sharp/c_sharp_differences.rst
neikeq: This shouldn't be in Mathf. What PR added it?

PR #1442, opened 2018-05-13
Add exporting tutorial for "Dodge the Creeps"

cbscribe:
Please comment and make suggestions.


mhilbrunner:
>However, it's not practical to ask your friends to download Godot

What?! If they don't already have Godot installed, they're probably not *real* friends!

This is a cool addition, and I didn't spot anything while reading it twice. So, merging! And thanks!

mhilbrunner:
Ah, damn, of course as soon as I pressed the button I realized: this should probably not be `_exporting:`, but something like `_doc_exporting_your_first_game`. Will fix :)

PR #1444, opened 2018-05-13
Add link for getting older versions of Mono

### LINKED ISSUE #1417: C# requirements need to be more explicit.
So i just spent half a day debugging an issue as to why the "is" examples in input weren't working.

Turns out, that it's easy to have it use msbuild.exe 14 instead of 15, and to fix it i tried to download the 2017 ms build tools, but that didn't work, until i found out only installing VS community would make it work. 

The documents only state:
`
To use C# in Godot you must have Mono installed (at least version 5.2), as well as MSBuild (at least version 15.0) which should come with the Mono installation.`

Which doesn't mention anything about the .net sdk, or VS community.  Oh right this is for windows as well. 

So i propose that we have links to VS community and a line about the build tools not working for this. 
Kelly thomas's comment here was the only one that helped me getting the correct version of Msbuild working.

https://github.com/godotengine/godot/issues/18152

skyace65:
Adds a link in the C# basics page to the Mono page on installing older versions of Mono on Linux. Closes #1417


mhilbrunner:
IMO, also add a link to https://download.mono-project.com/archive/ (download of older versions, mostly useful for Windows/OS X).
Also, Linux should be capitalized.

skyace65:
Added that link, and capitalized Linux.

skyace65:
Added that link, and capitalized Linux.

PR #1445, opened 2018-05-13
use `await ToSignal()` for game over timing

KellyThomas:
@exts I saw @Chaosus documenting this method in https://github.com/godotengine/godot-docs/pull/1441 and thought it should be used here.

### DIFF
--- a/getting_started/step_by_step/your_first_game.rst
+++ b/getting_started/step_by_step/your_first_game.rst
@@ -968,14 +968,12 @@ temporarily, such as "Get Ready". On the ``MessageTimer``, set the
 
     async public void ShowGameOver()
     {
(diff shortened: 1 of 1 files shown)


Chaosus:
Note you should also mark method `async` in order to correct work

cbscribe:
I don't have Mono set up atm. Can someone confirm this works as expected?

paulloz:
Works for me in 3.0.2

exts:
Works, maybe add a comment explaining what ToSignal does for the reader.

exts:
> However there is no equivalent explanation of yield for GDScript users, so I question if this is the right place.

I kind of disagree with that mentality personally, but the change works and it can be merged

PR #1458, opened 2018-05-20
added tip about updating source files in Qt Creator

Valentactive:
Added a tip about updating the source files (and folder) after pulling the latest commits from GitHub or a fork of Godot. Also added screenshots for the "**Importing the project**" section. 

Since I'm not a native English speaker feel free to correct me and/or suggest changes. This is my first contribution to the docs, I hope formatting is ok.

**Hint:** In the mean time I did another commit which added some images for the importing of the project and setting the build and run commands in QT Creator (I also changed one line to match the screenshot that was taken on Linux in this case.)

Oh and btw I fixed a typo it's called "Qt Creator". I make that mistake also all the time, but it's called everywhere "Qt Creator" in the IDE (About dialog, etc.).

I squashed those two commits into one, my changes only touch the "**Qt Creator**" section.


mhilbrunner:
Thank you!

PR #1469, opened 2018-05-25
Demonstrate coroutine's `completed` signal

Zirak:
The PR may be fudged up by forking from the 3.0 branch, if it's a problem and I'll create another one from master.


Zirak:
Holy cow, yes, this did not go as planned. Will fork from master. Sorry for the spam.

PR #1472, opened 2018-05-25
Update your_first_game.rst

jaynabonne:
Corrected HUD coordinates based on what I had to use when I tried the tutorial. Otherwise things were off screen, off center, etc in weird ways. Plus the "layout" fields seemed opposite for the Start button and the message text.


mhilbrunner:
cc @cbscribe

cbscribe:
FYI, changes should be made to `master` branch, where the anchors have been corrected. See http://docs.godotengine.org/en/latest/getting_started/step_by_step/your_first_game.html#scorelabel

PR #1486, opened 2018-05-31
Fixed missing backslash key in KeyList enum

Kloppie5:
The explanation for the backslash key didn't properly escape the backslash


StraToN:
Please read https://github.com/godotengine/godot-docs#editing-existing-pages
-> No editing on `classes/*` pages, they are automatically generated from Godot's XML class references.

akien-mga:
Actually this needs to be fixed in the script used to convert the XML to RST: https://github.com/godotengine/godot/blob/master/doc/tools/makerst.py

PR #1488, opened 2018-06-01
Tutorial for a very basic C++ GDNative plugin

BastiaanOlij:
This is a GDNative C++ tutorial based on nativescript 1.0 that people can use to develop GDNative plugins for Godot 3.0 (and should still work in 3.1).

Once nativescript 1.1 is official (release of 3.1) I'll do an updated version.


tutorials/plugins/gdnative/gdnative-cpp-example.rst
```diff
@@ -0,0 +1,337 @@
+.. _doc_gdnative_cpp_example:
+
+GDNative C++ example
+====================
+
+Introduction
+------------
+This tutorial builds on top of the information given in the :ref:`GDNative C example <doc_gdnative_c_example>` so I highly recommend you read that first.
+
+The C++ bindings for GDNative are build on top of the nativescript GDNative API and provide a nicer way to "extend" nodes in Godot using C++. Basically this is the equivilant to writing GDScript scripts but in C++.
```
mhilbrunner: equivalent

tutorials/plugins/gdnative/gdnative-cpp-example.rst
```diff
@@ -0,0 +1,337 @@
+.. _doc_gdnative_cpp_example:
+
+GDNative C++ example
+====================
+
+Introduction
+------------
+This tutorial builds on top of the information given in the :ref:`GDNative C example <doc_gdnative_c_example>` so I highly recommend you read that first.
```
mhilbrunner: I = we

tutorials/plugins/gdnative/gdnative-cpp-example.rst
```diff
@@ -0,0 +1,337 @@
...
+
+The C++ bindings for GDNative are build on top of the nativescript GDNative API and provide a nicer way to "extend" nodes in Godot using C++. Basically this is the equivilant to writing GDScript scripts but in C++.
+
+We'll be looking at nativescript 1.0 which is available in Godot 3.0. Godot 3.1 will see the introduction of nativescript 1.1 which has a number of improvements. We'll update this tutorial once that is officially released but the overal structure is much the same.
+
+You can download the full example we'll be creating in this tutorial here: https://github.com/BastiaanOlij/gdnative_cpp_example
```
mhilbrunner: Should this be moved to the official demos repo?

tutorials/plugins/gdnative/gdnative-cpp-example.rst
```diff
@@ -0,0 +1,337 @@
...
+
+Setting up your project
+-----------------------
+There are a few prerequisits you'll need:
```
mhilbrunner: prerequisites

Also, link to http://docs.godotengine.org/en/latest/development/compiling/index.html here?

tutorials/plugins/gdnative/gdnative-cpp-example.rst
```diff
@@ -0,0 +1,337 @@
...
+
+1) A Godot 3.x executable
+2) a C++ compiler
+3) scons as a build tool
+4) a copy of the godot_headers repository you can find here: https://github.com/GodotNativeTools/godot_headers
+5) a copy of the godot_cpp repository you can find here: https://github.com/GodotNativeTools/godot-cpp
+
+You can just download these repositories from github or let git do all the work for you. I've started to submodule them into my project folder. This is how I usually start my projects:
```
mhilbrunner: GitHub

tutorials/plugins/gdnative/gdnative-cpp-example.rst
```diff
@@ -0,0 +1,337 @@
...
+    git submodule init
+    git submodule update
+
+This will trigger downloading these two repositories into your project folder.
+
+**note** godot-cpp submodules godot_header but there is currently a permissions issue. Once that is resolved you can use that instead of having your own headers submodule. 
+
+Building the C++ bindings
+-------------------------
+Now that we've downloaded our prerequisits it is time to build our C++ bindings.
```
mhilbrunner: prerequisites

tutorials/plugins/gdnative/gdnative-cpp-example.rst
```diff
@@ -0,0 +1,337 @@
...
+
+Now the repository contains a copy of the meta data for the current Godot release (3.0.2 a the time of writing this) but if you need to build these bindings for a newer version of Godot you simply call the godot executable:
```
mhilbrunner: IMO remove version mention, will be outdated too fast

tutorials/plugins/gdnative/gdnative-cpp-example.rst
```diff
@@ -0,0 +1,337 @@
...
+
+.. code-block:: none
+
+    godot --gdnative-generate-json-api godot_api.json
+
+And place the resulting godot_api.json file in our godot-cpp folder.
+
+Now to generate and compile our bindings we do (do chose either windows, linux or osx):
```
mhilbrunner: Windows, Linux, OSX

tutorials/plugins/gdnative/gdnative-cpp-example.rst
```diff
@@ -0,0 +1,337 @@
...
+
+.. code-block:: none
+
+    cd godot-cpp
+    scons platform=windows/linux/osx headers=../godot_headers generate_bindings=yes
+    cd ..
+
+This step will take awhile but at the end of it, you should have static libraries you can compile into your project stored in godot-cpp/bin.
```
mhilbrunner: a while

tutorials/plugins/gdnative/gdnative-cpp-example.rst
```diff
@@ -0,0 +1,337 @@
...
+    _sections_unfolded = [ "Resource" ]
+
+This is a standard Godot resource and you could just create that directly inside of your scene but having this as a resource file makes life a lot easier in (re)using your native_script. 
+The important bits here are that we're pointing to our gdnlib file so Godot knows which dynamic library contains our native_script, and the ``class_name`` which identifies the natice_script in our plugin we want to use.
+
+Ok, we're all setup. Time to jump back into Godot. We load up the main scene we created way back in the beginning and now we add a Sprite to our scene:
+
+.. image:: img/gdnative_cpp_nodes.png
+
+We're going to assign our Godot logo to this sprite as our texture, turn of centered, and drag our ``gdexample.gdns`` file onto the script property of our sprite:
```
mhilbrunner: turn off

tutorials/plugins/gdnative/gdnative-cpp-example.rst
```diff
@@ -0,0 +1,337 @@
...
+
+.. image:: img/gdnative_cpp_sprite.png
+
+And we're ready to run our project:
+
+.. image:: img/gdnative_cpp_animated.gif
+
+Next steps
+----------
+Obviously the above is a very basic example just to get you setup but I hope it shows the basics. You can build upon this example to create full fledged scripts to control the nodes in Godot but using C++ as the language.
```
mhilbrunner: I = we

tutorials/plugins/gdnative/gdnative-cpp-example.rst
BastiaanOlij: The problem I had with gdnative-demos is that it no longer follows the structure I'm building, but it does make more sense to place it there.  It also makes it easier to add this to the asset library.

That said, @karroffel may have something to say about this too, I've often wondered if it doesn't make sense to have a GDNative Demos organisation on github (wouldn't it be nice if Github allowed you to group a set of repos) and have each demo be its own repo instead of having one repo with all the demos.

BastiaanOlij:
@mhilbrunner thanks man!

willnationsdev:
Wasn't there a recent change that allows a .tres to be used in place of the .gdnlib? I might mention that, and possibly show a picture of the alternative means of creating the resource / how it works. That way people can try both.

Also, I remember seeing that the basic function call was one that returns a Variant and let's you pass in a Variant (potentially an Array of parameters). When you register your own custom methods, how does one define multiple parameters? What do the register method calls look like? Do you need to specify parameter types or anything? What about signals, constants, or properties based on getters and setters? These are all things I would expect a docs page to cover for users.

mhilbrunner:
@BastiaanOlij Do you want to expand this wrt the stuff mentioned by @willnationsdev above or shall I merge now and you / others can do so in further PRs?

willnationsdev:
@mhilbrunner I'd probably just merge it so that others can go ahead and submit their own PRs.

BastiaanOlij:
@willnationsdev I don't know about using the .tres extension but that would be for gdns files, not gdnlib files I would assume. I don't think that one is very important.

As for parameters, what you state is mostly important for C based externals where you need to declare far more. With C++ that is indeed how things work internally but it's nicely taken care off in the bindings library. Signals, properties, etc. I would do follow up tutorials for. I really want the introduction to be just about the basics or you just end up overloading people with info. Each of those would be a chapter in their own right. "How to do signals in GDNative C++", "How to add a property in GDNative C++", etc.

@mhilbrunner I think merging it as is would be great

karroffel:
@willnationsdev @BastiaanOlij yes I made a commit that allows to have `GDNativeLibrary` resources as subresources and be able to save them as .tres. I don't know if that change is in 3.0.3 or will only be in 3.1.


Doesn't look like it has been cherry-picked #17965

karroffel:
Also @BastiaanOlij I think I want to fix the rights issue in the godot-cpp repository first before this gets merged, it will simplify the whole process *a lot*. So I would wait a little with the merge for that change

BastiaanOlij:
@karroffel tbh I actually liked it better when they were two separate submodules, nested submodules are icky :wink:

willnationsdev:
@BastiaanOlij eh? Thought your tutorial only showed a single-layer pair of sub-modules for godot-cpp and godot_headers. Is there another one somewhere?

BastiaanOlij:
@willnationsdev indeed it does, but only because Karroffel used the wrong url when including godot_headers inside of the godot-cpp project.

Once fixed up we won't need godot_headers as it will already be included in godot-cpp BUT you'll need to do a recursive submodule update

BastiaanOlij:
And for people who don't want to use git but just download the zip that can be a pain.

willnationsdev:
@BastiaanOlij I just noticed that you never actually detail the name of the final .cpp file with the `init/terminate` methods. I assume it's `init.cpp`, but it's never mentioned.

willnationsdev:
In your instructions, you tell users to name the demo folder "Demo", but the `final_lib_path` in the SConstruct file states that its initialized value (to which additional content is appended) is "demo/bin". This seems like it would lead to a bug.

I would also recommend making the "libgdexample" text an optional parameter that way users can simply specify it on the command line without having to edit the SConstruct file directly.

willnationsdev:
Perhaps modifying the `final_lib_path = "demo/bin"` line to be this instead?

    project_path = ARGUMENTS.get("project", ".")
    final_lib_path = ARGUMENTS.get("lib", project_path + "/bin/")

And then modifying the instructions for the command line to request that users specify `project=demo` or `project=Demo`, whichever.

willnationsdev:
Would also be a good idea to briefly go over all of the various command line options and what they are used for (in case someone doesn't have the know-how to parse the SConstruct file).

willnationsdev:
We should also endeavor to include the `vs_proj=yes` parameter again, which would mean needing to include an if statement that uses the scons Builder. I'm not sure exactly what all the differences are between the various parameters though / what we'd need to include.

    barsrcs = ['bar.cpp']
    barincs = ['bar.h']
    barlocalincs = ['StdAfx.h']
    barresources = ['bar.rc','resource.h']
    barmisc = ['bar_readme.txt']
    
    dll = env.SharedLibrary(target = 'bar.dll', source = barsrcs)
    
    env.MSVSProject(
        target = 'Bar' + env['MSVSPROJECTSUFFIX'],
        srcs = barsrcs,
        incs = barincs,
        localincs = barlocalincs,
        resources = barresources,
        misc = barmisc,
        buildtarget = dll,
        variant = 'Release')

I assume the SConstruct file itself could be in the `misc` group, but I don't know the difference between the includes and local includes, nor what any `resources` might be, if any.

So, maybe something like...

    if (ARGUMENTS.get("vs_proj", "no") == "yes":
        # do stuff to build project

What do you think?

BastiaanOlij:
@willnationsdev I'm not sure what you're doing but you're going down the wrong path.

Demo is the folder into which you place your godot project. In contrast to the old C example I've started doing this because else your plugins source code starts to get exported along with your project. 

You generally place the dlls into a subfolder, not in its root, and possibly into a subfolder in that folder per platform to organise things as the dynamic library often have the same name. 

Same with the bits, which are really important and should also be picked up when you build godot-cpp (are you sure you're not using an old copy of that repo?). You'll end up with a 32bit windows version, a 64bit windows version, a 64bit osx version, etc.etc.etc., they all need to have a unique name to link against. It used to always compile to the same name but that was changed weeks if not months ago.

As for the VCProject output, I don't use that so I don't know tbh

willnationsdev:
@BastiaanOlij I wasn't disparaging your use of having a project subfolder or building a library in another subfolder within there. Rather, I was thinking about how 1) the "Demo" and "demo/bin" references I mentioned use different capitalization, so users might have created a Demo folder per your instructions, but not have things build right (haven't tried it yet) because it tries to find "demo/bin" rather than "Demo/bin". 2) make it so that the project name and target library name are not hardcoded values, but rather additional parameters (figured "." would be the most suitable default value for a project path and then putting the library at bin subfolder within the project location would be appropriate. 3) looking at the parameters for the scons command, you can specify what name you would like to give to the built `cpp_bindings_library`, but if you directly type in `godot-cpp.x11.64` (like I expected to do), it results in the wrong file name since it appends the file extensions for the user. Usually when I'm writing out my own custom filename, I supply the file extension(s), so this tripped me up and I didn't even notice it was happening for a long time. This could be avoided by clearly warning users or by having logic that detects the number of dots in the final portion of the path and only appending the extensions if ones don't already exist. I'm not an expert on that stuff though, so if there are better ways of handling that, then by all means...

BastiaanOlij:
@willnationsdev ah! 

1. I totally missed the point of case, I work primairily on windows nowadays which hardly cares about case, so its a mistake I often make. Yes I should probably go through the tutorial and just make everything lower case.

2. maybe variables at the top of the scons project, not sure if they should be arguments. The scons files belongs to your project, I tend to write them so I can just run ```scons``` and any arguments are just when I need to do something different. Note that it scans for cpp files in the src folder, I don't like naming individual files if I can help it. So the only setting really is the final library name.

3. very different approach to me, the scons file for the godot-cpp makes up the name in the same way this scons file for the sample project does, so its a fire and forget thing. I don't want to have to think of a name, just standardise it and let it do its thing. I build godot-cpp with the same arguments as the sample project, and it works.

willnationsdev:
@BastiaanOlij 

1. Cool

2. I'd prefer it if they were arguments, that way people can define defaults by editing the file directly, but tutorials can just tell people what parameters to supply to make things work. It makes the process more accessible to folks imo. Supplying command line arguments is much faster and easier than having to launch VS Code or Atom or something to scan through a foreign file, locate the lines you need, and then edit and save the file according to the description you're given in the instructions, before FINALLY returning to the command line to execute the next instruction.

3. Fair enough. I just think the documentation could do with a clear description of the possible parameters, what they are for, and how they should be formatted, so that people (like me) don't get tripped up by misconstruing the purpose, misunderstanding the proper format, or not knowing the name / existence of a given parameter.

mhilbrunner:
Read it again, I'll merge it now, as it seems like a valuable addition already - feel free to submit further PRs to continue working on it :)

Thanks for spending time on the docs!

BastiaanOlij:
@willnationsdev yup, it's always hard to keep in mind how much knowledge someone has and especially when they are used to a completely different way of setting things up. In the end I feel there just needs to be a way so someone has a starting point, otherwise you just end up with a tutorial bloated with stuff.  I'm trying to keep things aligned with how Godot itself is setup so I can just direct someone to the existing documentation on how to compile Godot on the different platforms and what tools they need.  
This tutorial should be about how you write the actual plugin. Some people might not even want to use scons and have the skills to setup their own build files.

@mhilbrunner thanks, I'll probably submit some more PRs over time to tweak things, some of wills suggestions are definitely worth adding, just been a busy this week so agreed, better to make available to people what we already have and add stuff later on

PR #1493, opened 2018-06-03
Correct API mention is_networking_server -> is_network_server

gwilymk:
In the tutorial for the high level multiplayer, it mentions the API is_networking_server. The API is actually is_network_server. This PR corrects that mistake.


mhilbrunner:
Thank you!

Issue #1483, opened 2018-05-31
Global url list/file and tokenizing all urls.


romeojulietthotel:
While fixing some urls (404s, redirects and broken) I realize that it's a cumbersome task to keep them fresh.

A more human scalable way would be to maintain a master list of urls and then instead of inserting a url into each doc we insert a static token. Then via sphinx the token is translated to the url that matches the token. The token never changes and only the global list would need updating. We could make it so that the tokens are human readable and not abstracted too far away from the url.

The global list would contain the url and the token. This would mean just one file or urls would need to be edited and then the sphinx build would do the rest.

romeojulietthotel:
There was a built in extension in Sphinx for this already. See PR https://github.com/godotengine/godot-docs/pull/1494.

mhilbrunner:
Closing this; see discussion in https://github.com/godotengine/godot-docs/pull/1494.

PR #1494, opened 2018-06-03
This allows all urls to be kept in one file and should make things easier

romeojulietthotel:
It's a start to ... https://github.com/godotengine/godot-docs/issues/1483

I modified one _.rst_ to show how it looks but there are, of course, a lot more urls to convert.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -11,7 +11,14 @@ needs_sphinx = '1.3'
 
 # Sphinx extension module names and templates location
 sys.path.append(os.path.abspath('extensions'))
-extensions = ['gdscript', 'sphinx_tabs.tabs', 'sphinx.ext.mathjax']
+extensions = [
+    'sphinx.ext.extlinks',
+    'sphinx.ext.mathjax',
+
+    'gdscript',
+    'sphinx_tabs.tabs'
+]
+
 templates_path = ['_templates']
 
 # You can specify multiple suffix as a list of string: ['.rst', '.md']
@@ -109,3 +116,10 @@ gettext_compact = False
 # Exclude class reference when marked with tag i18n.
 if tags.has('i18n'):
     exclude_patterns = ['classes']
+
+# -- Extension interface -------------------------------------------------------
+
+extlinks = {
+            'issue': ('https://github.com/godotengine/godot/issues/%s','#'),
+            'godothome': ('https://godotengine.org''%s',None)
+}
--- a/development/compiling/introduction_to_the_buildsystem.rst
+++ b/development/compiling/introduction_to_the_buildsystem.rst
@@ -92,7 +92,7 @@ Example for using 4 cores:
 
     user@host:~/godot$ scons platform=x11 -j 4
 
-Note that there are currently `issues <https://github.com/godotengine/godot/issues/5182>`__ with parallel builds for at least some users, so if you are running into errors, try building without the ``-j`` parameter.
+Note that there are currently issues (:issue:`5182`) with parallel builds for at least some users, so if you are running into errors, try building without the ``-j`` parameter.
 
 Resulting binary
 ----------------
@@ -125,7 +125,7 @@ Just copy that binary to wherever you like, as it contains the
 project manager, editor and all means to execute the game. However, it
 lacks the data to export it to the different platforms. For that the
 export templates are needed (which can be either downloaded from
-`godotengine.org <https://godotengine.org/>`__, or you can build them yourself).
+:godothome:`/`, or you can build them yourself).
 
 Aside from that, there are a few standard options that can be set in all
 build targets, and which will be explained below.
@@ -199,9 +199,9 @@ Export templates
 ----------------
 
 Official export templates are downloaded from the Godot Engine site:
-`godotengine.org <https://godotengine.org/>`__. However, you might want
-to build them yourself (in case you want newer ones, you are using custom
-modules, or simply don't trust your own shadow).
+:godothome:`/`. However, you might want to build them yourself (in case you 
+want newer ones, you are using custom modules, or simply don't trust
+your own shadow).
 
 If you download the official export templates package and unzip it, you
 will notice that most are just optimized binaries or packages for each


romeojulietthotel:
> Not sure yet what is easier this way - is :godothome:`/` really better than `godotengine.org <https://godotengine.org/>`__?

It is better, the point of the change has to do with checking links and making that easier if they change, etc. You only need to edit one file to maintain fresh links. The recent commit I made updated links and it was not a trivial task.

> Also, how would this look for more complicated URLs? The longer the URL, and the less widely used it is inside the docs, the more cumbersome this approach seems to be.

Why? Please explain this, I don't understand.

> It also means editing the docs becomes harder, as you now need to edit multiple files to change a link (which is, for example, hard to do with the GitHub web UI).

Only if you're adding a new url. In that case you add it to one file and place the token in the other file.

> I'm undecided.

Time saving and simplicity is good.

mhilbrunner:
>Why? Please explain this, I don't understand.

Take for example this page: http://docs.godotengine.org/en/3.0/community/contributing/code_style_guidelines.html

There are a lot of URLs that are mostly only used on this page. We'd need to create something like the following:

- `:godot-clangformat` - https://github.com/godotengine/godot/blob/master/.clang-format
- `:llvm-clangformat` - http://clang.llvm.org/docs/ClangFormat.html
- `:llvm-download:` - http://llvm.org/releases/download.html
- `:llvm-repo:` - http://apt.llvm.org/
- `:qt-creator-clangformat` - http://doc.qt.io/qtcreator/creator-beautifier.html
- `:vscode-clangformat` - https://marketplace.visualstudio.com/items?itemName=xaver.clang-format
- `:vs-clangformat` - https://marketplace.visualstudio.com/items?itemName=LLVMExtensions.ClangFormat
- `:vim-clangformat:` - https://github.com/rhysd/vim-clang-format
- `:pep8:` - https://www.python.org/dev/peps/pep-0008/
- `:autopep8:` - https://pypi.python.org/pypi/autopep8

Not against it, but also not sure how much it improves the status quo. Editing two files instead of one probably isn't a big ask, but on the other hand we already get feedback that contributing to the docs is too cumbersome. For the main repo, this isn't as bad, as there the contributors are mostly experienced developers.

For the docs, people expect to be able to edit the content as easily as in a wiki. We already don't give them that, but rather demand they know:

- git / GitHub & PR workflow, including, in the worst case, things like `git commit --amend && git push -f`
- knowing how to differentiate branches and commit against master, not other branches
- RST
- ideally, having Python/Sphinx locally and building the page to see how their formatting looks for tables etc.
- CLI basics

We should be careful about not introducing any more magic.

Not saying this *would* make it worse. I'm just... not sure.

romeojulietthotel:
> There are a lot of URLs that are mostly only used on this page. We'd need to create something like the following:
> 
> :godot-clangformat - https://github.com/godotengine/godot/blob/master/.clang-format
> :llvm-clangformat - http://clang.llvm.org/docs/ClangFormat.html
> :llvm-download: - http://llvm.org/releases/download.html
> :llvm-repo: - http://apt.llvm.org/
> :qt-creator-clangformat - http://doc.qt.io/qtcreator/creator-beautifier.html
> :vscode-clangformat - https://marketplace.visualstudio.com/items?itemName=xaver.clang-format
> :vs-clangformat - https://marketplace.visualstudio.com/items?itemName=LLVMExtensions.ClangFormat
> :vim-clangformat: - https://github.com/rhysd/vim-clang-format
> :pep8: - https://www.python.org/dev/peps/pep-0008/
> :autopep8: - https://pypi.python.org/pypi/autopep8

###### (Ironically some of those are redirects or 404 IIRC....)
But one would only need to edit two files for new urls. Otherwise for an issue, for example, you just enter this :issue:`1234`

As for the workflow for contributing to docs being cumbersome that is a larger separate issue. 
When was the last time that the urls throughout the docs were refreshed? (before my recent commit)
That's the question to ask and this makes that easier and means there is a higher probability that urls stay fresh. What's the goal, ease of contribution or good docs? Good docs should come first.

romeojulietthotel:
> Not necessarily bad, not sure if it solves a problem either: a global search and replace also works in the rare case you need to replace URLs in multiple files.

How would you have done a search and replace here:

https://github.com/godotengine/godot-docs/commit/a6f126540334cc718f07a88c0dfd232ad8242d80

NathanLovato:
> How would you have done a search and replace here:

Okay allow me to rephrase: in the case you need to change the same URL in multiple places. E.g.: `https://github.com/godotengine/godot/issues/` or `https://godotengine.org`.

The reason I'm not sold out-right is most people who come to contribute won't check for guidelines or anything. I think neither mhil or I are are against the change or saying it's bad.

romeojulietthotel:
> If someone wants to take the time to maintain it, it can be good.

The PR is all about easing maintenance, just consider that recent commit of mine to fix urls.   I don't know how long the urls were that way but it was non-trivial to fix them all. But if they were managed this way, it would have been trivial.

This repo is not super busy so if new docs come in with urls the reviewer can point the contributor to the _conf.py_ examples and ask that _conf.py_ is updated too.

akien-mga:
I'm not sure about this, it's already difficult enough for documentation writers to use reST's weird inline links syntax properly, so if we need to use such tokens instead of register all links manually in `conf.py`, this could be quite bothersome.

I'd also prefer external links not to be in `conf.py`, which is the Sphinx configuration file. If we do go with this option, the links should be split in another python file.

IMO, if we have a tool to check and extract dead links, it's easy enough to do a global search and replace in all files. A sed script can do the trick.
I understand the aim, but we should make sure that the solution is not worse than the problem.

NathanLovato:
Ah if we could have a tool to check dead links, that'd be great indeed. The point @romeojulietthotel brought up is in his last PR, there were a few broken links that only appear once in the docs, so they're hard to track.

mhilbrunner:
Closing this as there does not seem to be a consensus on moving to this system. While it would make it easier to edit URLs in one central place, that only is useful for links that are used often - which then, ideally, shouldn't need to change often if they are used so frequently and are thus that important.

On the other hand, this would make the already complex system with Sphinx/RST files even more complicated to new contributors. We need to reduce friction for contributions as much as possible, even if that makes checking for dead links harder (or more time intensive).

Thanks for your contribution and for your time!

PR #1503, opened 2018-06-07
GDNative/C++: Parameterized scons

willnationsdev:
Okay, several changes here.

One is that I noticed the "demo/bin" path was being hardcoded in the scons file. This isn't really a good idea.

The other items of note are that any value which was previously hardcoded in the scons file has now been converted into a parameter of its own. This was particularly needed because the scons file was incompatible with the structure indicated by the tutorial. The tutorial assumed the demo project was in a subfolder (in "./demo"), but the sources were assumed to be in the local src folder ( "." and "./src" ) rather than in the demo project's respective folders ( "./demo" and "./demo/src" ).

Instead, you can now supply "project=demo" in the example usage and the `project_path` variable gets assigned. The `project_path` and its relative subfolder `src` are automatically included in the sources (so you can still keep the SConstruct file in the parent folder where `demo`, `godot-cpp`, and `godot_headers` are). I just tested this stuff on my windows machine and it all works fine.

I also enabled users to specify the name of the resulting dynamic library file (so you can say "name=libgdexample" to get "libgdexample.ext").

It will also automatically filter out the extensions from the cpp_bindings_library_path. That way, if the user accidentally says "godot-cpp.64.windows.lib" or something, it'll strip out all of the extensions and output the correct phrasing.

Please review this and let me know what you all think.

@BastiaanOlij @karroffel 

I would ideally like to build off of this SConstruct file to create a plugin for building C++ libraries using scons in Godot.

### DIFF
--- a/tutorials/plugins/gdnative/gdnative-cpp-example.rst
+++ b/tutorials/plugins/gdnative/gdnative-cpp-example.rst
@@ -202,15 +202,23 @@ This is the part I can't really make pretty. Just use the ``SConstruct`` file be
     import os, subprocess
     
     # Local dependency paths, adapt them to your setup
-    godot_headers_path = ARGUMENTS.get("headers", "godot_headers/")
-    cpp_bindings_path = ARGUMENTS.get("cpp_bindings_path", "godot-cpp/")
+    godot_headers_path = ARGUMENTS.get("headers", "godot_headers")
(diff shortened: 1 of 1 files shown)


willnationsdev:
I also plan to amend the tutorial so that it clarifies what all of the parameters are / what their format should be.

willnationsdev:
Closing this one since the meaningful changes are being diverted to other PRs.

PR #1504, opened 2018-06-08
Just fixed some small things

BastiaanOlij:
Fixed to small mistakes. Also renamed Demo to demo (so @willnationsdev we're doubling up on that one :wink: )

@karroffel I've removed the remark about godot_headers inside of godot-cpp. I'll rewrite that section when I have more time and can test through the steps.

### DIFF
--- a/tutorials/plugins/gdnative/gdnative-cpp-example.rst
+++ b/tutorials/plugins/gdnative/gdnative-cpp-example.rst
@@ -45,9 +45,7 @@ If you downloaded or cloned my example from the link specified in the introducti
     git submodule init
     git submodule update
 
-This will trigger downloading these two repositories into your project folder.
-
-**note** godot-cpp submodules godot_header but there is currently a permissions issue. Once that is resolved you can use that instead of having your own headers submodule. 
+This will trigger downloading these two repositories into your project folder. 
 
 Building the C++ bindings
 -------------------------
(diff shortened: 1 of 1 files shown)


willnationsdev:
Okay, this looks good to me. Part of the reason for my desire to make changes in the other PR stemmed from the fact that it seemed your instructions indicated that the `src` folder was a subfolder of the demo project folder, not the gdnative module folder.

    To start we want to create an empty Godot project...for our example we'll create a project in a folder called "demo" inside of our GDNative modules folder structure.
    
    *Inside our demo* we'll create a scene with a single Node of type Node called "Main" and we'll save this as main.tscn. We'll come back to that later.
    
    *We're also going to create a subfolder called ``src``* into which we'll place our source files.

Obviously, when I did this, the `SConstruct` file failed because it was including "."/"src" instead of "demo"/"demo/src".

This could be fixed by also making it more clear in the instructions that the `src` subfolder is in the gdnative module folder and not the project folder.

----

If these changes are made, then I actually don't see any reason to keep my changes from the other PR. Instead, I can create an alternate PR that simply introduces entirely new tutorial pages for making a more advanced scons file / how to operate a scons file / how to use advanced NativeScript C++ functionality (though that I might need help on, lol).

I've tested the process of building the necessary engine resources through the editor too, rather than making them directly as text files (.gdnlib and .gdns). Should we perhaps create "spoiler" sections that can be expanded into a more detailed overview of how to use those windows (the whole process of creating each file?). I feel that would make it more "complete" without resulting in the page blowing up with content.

willnationsdev:
Seems like spoiler sections aren't supported in .rst directly. You'd have to have some kind of javascript triggered by the presence of a class generated by the .rst in order to signify a "spoiler" or "see more" button of some kind. Anyway, way more complicated than what I was hoping for. Guess you could just make another tutorial page out of it and then link to that page as another option, to keep the size of this tutorial down.

willnationsdev:
Also, if you REALLY want the scons file to be dedicated to just this tutorial and reserve a more complicated scons file for a subsequent tutorial, then you might as well hardcode all of the relevant content aside from the os and just make assumptions about where all of the files are located. You can outline, very clearly to the reader, where each of their files should reside relative to one another at the end of the process.

PR #1505, opened 2018-06-08
Used hardcoded, downloadable scons for C++ tut

willnationsdev:
@BastiaanOlij This should remove the complexity of the build process from the user almost entirely. They never even have to look at any of the build script contents. Using hardcoded values only, with a clearer indication of the intended folder structure, it keeps the purpose of the tutorial entirely focused on getting the user through the fundamental process of downloading the bindings, building, creating the engine resources, and running a game with scripted gameplay.

We can then use subsequent tutorials for more complex behaviors.

### DIFF
--- a/tutorials/plugins/gdnative/files/cpp_example/SConstruct
+++ b/tutorials/plugins/gdnative/files/cpp_example/SConstruct
@@ -1,0 +1,47 @@
+#!python
+import os
+
+target = ARGUMENTS.get("target", "debug")
+platform = ARGUMENTS.get("platform", "windows")
+bits = ARGUMENTS.get("bits", 64)
+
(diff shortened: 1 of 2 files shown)


BastiaanOlij:
I like that idea, that indeed keeps the focus on actually writing the module and leaves the complexity of the SConstruct file away. 

Having a straight forward SConstruct file downloadable purely for this protect helps someone who's just following the tutorial to reproduce the end result.

We can always have a more complex SConstruct file as part of the full demo you can download or as I said before, have a separate tutorial purely focusing on making a more generic SConstruct setup for GDNative.

mhilbrunner:
Thanks :)

PR #1507, opened 2018-06-08
Fix typo error in matrices_and_transforms.rst

Chaosus:
There is no `basis_xform`, but you can access to it using `basis.xform`


Chaosus:
Oops.. looks like its existed property in Transform2D.

PR #1516, opened 2018-06-14
Add escn exporter to doc

sdfgeoff:
I think this contains all the information from the simedis wiki page.


sdfgeoff:
This was meant to be against a different fork. I don't know why it ended up here.

PR #1528, opened 2018-06-20
documentation on inheritance and constructors

razvanc-r:
Tried covering the possible scenarios with an example. It probably needs
some more refinement. Fixes godotengine/godot#19644

PS. my text editor auto-trims trailing white space but I don't think that's an issue :P

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_basics.rst
+++ b/getting_started/scripting/gdscript/gdscript_basics.rst
@@ -946,12 +946,53 @@ The class constructor, called on class instantiation, is named ``_init``.
 As mentioned earlier, the constructors of parent classes are called automatically when
 inheriting a class. So there is usually no need to call ``._init()`` explicitly.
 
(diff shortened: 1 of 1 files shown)


getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -946,13 +946,54 @@ The class constructor, called on class instantiation, is named ``_init``.
...
+        entity = e
+
+    func enter(m):
+        message = m
+
+
+    # Idle.gd (inheriting class)
+    extends "State.gd"
+
+    func _init(e=null, m).(e):
```
vnen: If you add a default value to an argument, the next arguments also need a default value.

razvanc-r: Indeed! Nice catch

mhilbrunner:
Thank you :)

PR #1531, opened 2018-06-20
Configuring an IDE: Add intro and details for scons path with Qt Creator

akien-mga:
Also moved Qt Creator up as it's the IDE used by most core developers.
Some users saw Kdevelop first and expected it to be the recommended one.

### DIFF
--- a/development/cpp/configuring_an_ide.rst
+++ b/development/cpp/configuring_an_ide.rst
@@ -6,11 +6,96 @@ Configuring an IDE
 We assume that you already `cloned <https://github.com/godotengine/godot>`_
 and :ref:`compiled <toc-devel-compiling>` Godot.
 
+You can easily develop Godot with any text editor and by invoking ``scons``
+on the command line, but if you want to work with an IDE (Integrated
+Development Environment), here are setup instructions for some popular ones:
+
+- :ref:`Qt Creator <doc_configuring_an_ide_qtcreator>` (all desktop platforms)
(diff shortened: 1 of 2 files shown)


development/cpp/configuring_an_ide.rst
```diff
@@ -140,6 +92,67 @@ clicking the "Apply Filter" button. A simple click on "Ok" and you're ready to c
 
 .. image:: img/qtcreator-edit-files-dialog.png
 
+Kdevelop
+--------
+
+It is a free, open source IDE for Linux, Solaris, FreeBSD, Mac OS X and other Unix flavors.
```
OvermindDL1: KDevelop also supports Windows, this should be updated:  https://www.kdevelop.org/download

akien-mga: Thanks, updated.

OvermindDL1:
As another aside, KDeveop uses CMake as it's base project format (and has for a long while now), isn't there any way of having scons produce a CMakeLists.txt file so the user doesn't have to go through that pain of setting up a custom build system in it, because that is quite a pain?

mhilbrunner:
@OvermindDL1 Couldn't find something like that, this is the only thing Google pointed me to: http://scons-dev.scons.narkive.com/fmxmIl1O/generate-cmakelists-with-scons

OvermindDL1:
> Couldn't find something like that

Hmm, how sad...  Eh, if only godot used cmake, especially as how ubiquitous cmake has become (even boost is switching to it, which I find odd considering how hard up they were on b2...).

PR #1537, opened 2018-06-22
Fixed incorrect documentation for MacOS compiling

clayheaton:
Line 40 contained a typo, referring to `godot.osx.opt.tools.64` instead of `godot.osx.tools.64`.


mhilbrunner:
Thank you!

PR #1539, opened 2018-06-24
Fix reference to Scene tab in the 'Design a title screen' page

monkeyhybrid:
On line 170 of the 'Design a title screen page', the use of Inspector is mistakenly used to refer to what should be the Scene tab. Additionally, the sentence is also missing a closing full stop.

> Then, double click on each of the nodes in the Inspector to rename them

should be

> Then, double click on each of the nodes in the Scene tab to rename them.


mhilbrunner:
Thank you!

Issue #1509, opened 2018-06-09
Update Mesh Generation Tutorial


clayjohn:
I have always been bothered by the Mesh Generation tutorial. It requires the users to use both Gimp and Blender, and then the Godot aspect of the tutorial is very short. 

I would like to rewrite the tutorial to be contained purely within Godot. Using Godot to make a subdivided quad and then using noise in the shader to offset the vertices. I think this way the tutorial could be way more Godot-centered and also be more useful for people learning shaders and stuff.

If anyone has any other ideas or reasons why the tutorial should stay the same please let me know! Otherwise im going to make a PR in the next couple days with a revised tutorial.

StraToN:
I wrote this tutorial back for Godot 2.1. At this time, Godot 3.0 was not there and direct geometry was kind of on the exit way.

So, in order to demonstrate a larger workflow involving external tools (GIMP and Blender, but I remember stating you can use your favourite softs) and importing into Godot. I agree using Blender for a simple quad is overkill though but at the time of writing I had no choice.

No problem if you wish to update this tutorial. I am not even certain it teaches the *right* way to do, but I believe it fulfills its primary objective: learn geometry shading basics in Godot. 

Your proposal (using noise in the shader to offset the vertices) is very interesting, but I don't think it is an update to this tutorial, but an extension :) .

Looking forward to read your PR!

clayjohn:
Thanks for replying! I hope I didn't sound too rude in my original post, looking back at it now I think I come off kinda rude and I didn't intend that. 

Your right, when you would've originally written it Godot didn't have the ability to make a subdivided quad in-engine. So now is the time to update that!

I think I will actually change the scope of the tutorial a bit. Instead of calling it "Mesh generation with heightmap and shaders" I am thinking of calling it "vertex displacement with shaders" because a) the tutorial isn't teaching mesh generation, and b) I plan on replacing the heightmap with noise that way the tutorial can focus more on writing shaders and less on importing and using textures.

clayjohn:
This can be closed now that #1540 is merged

PR #1540, opened 2018-06-25
Rewrote Mesh Generation Tutorial

clayjohn:
As per issue #1509 I have rewritten the tutorial "mesh generation with heightmap and shaders". I have taken out the parts that use GIMP and Blender and instead chosen to generate the noise in the shader instead of using a heightmap. Accordingly I renamed the tutorial 'vertex displacement with shaders'

The tutorial now teaches the user:

- How to create and subdivide a Plane Mesh
- How to create and assign a material to a Mesh
- How to write a Shader that displaces the vertices of a Mesh
- How to pass values (Uniforms) into a Shader to update the Mesh in realtime
- How to approximate normals from a height function
- How to use a light with a custom material

I can see it cannot be automatically merged, I have been working on the tutorial for a long time and so have not kept up with recent commits. I will update the branch with the new commits right away. 

Feedback and comments are welcome. Please let me know what you think!


mhilbrunner:
Thanks for such a large contribution :)

Issue #1538, opened 2018-06-23
explain shader *_PIXEL_SIZE better


razcore-art:
For example for `TEXTURE_PIXEL_SIZE` it says *Default 2D texture pixel size.* which I guess it's fine, but thought it means a `vec2` with the size of the texture in pixels say `vec2(64.0, 64.0)` for the default `icon.png` asset. It actually represents the pixel size in UV coordinates, for the `icon.png` image this means `vec2(0.015625, 0.015625)` or `1/64`, took me a while to figure this out... perhaps it's a better way to saying it.

clayjohn:
What about 'Size of individual pixels in texture'? Or maybe it would be better to do a more mathematical definition like 'The inverse of the Texture resolution'?

clayjohn:
@razcore-art its definitely accurate, but I think people who are new to graphics will have a tough time understanding what that means. What do you think about "normalized size of pixel in texture"?

razcore-art:
@DarkEnchanter1 I think that's even more technical and it isn't really correct, cause it's not UV of texel, it's more like "normalized texel size", but still I for one did even know texels existed until now.
@clayjohn I think your definition is very similar to mine, actually I don't know why I said "UV-normalized". Ya I'd go with your definition if there's no other objections.

mhilbrunner:
I would actually prefer the description to include something like your above example:
>for the Godot icon.png image (64x64) this means vec2(0.015625, 0.015625) or 1/64

clayjohn:
I agree that it is much more clear to give an example. My only concern is being able to fit all that in the space provided. My understanding is that you only have the size of that little box. 

One idea would be to explain that in a sentence below the table of inputs and outputs. I think many of the terms would benefit from having a short example at the bottom as well.

mhilbrunner:
Those table cells should expand fine with more content.

razcore-art:
cool, then I'd just add *Example: for an image of size 64x32 pixels, `TEXTURE_PIXEL_SIZE will be vec2(1/64, 1/32)*, I don't want it to depend on specific files cause files can change and no one is gonna keep track of this stuff :)

clayjohn:
Sounds great! Ill amend the PR.

clayjohn:
Looks like you can't expand the table. Im getting an error while building the docs.

PR #1546, opened 2018-06-28
Clarified texture_pixel_size description in shading language

clayjohn:
Just a small PR to address issue #1538. 

We settled on "Normalized size of single pixel in texture", if anyone thinks that there is a clearer explanation then lets discuss.


mhilbrunner:
Thanks! Left a comment in the linked issue with a suggestion

PR #1548, opened 2018-06-29
fixed confusion in your_first_game tutorial

mrwanashraf:
if you head just before ``func _on_MobTimer_timeout():`` section in the tutorial you will find out that there are no steps for connecting the `timeout` signal to the `MobTimer` node which makes the one who follows the tutorial becomes confused because then if i blindly follow the tutorial i would just write the code that is shown in ``func _on_MobTimer_timeout():`` and won't connect the signal as there was no step required for it in the tutorial.


cbscribe:
I think you've messed up your branch - you'll need to fix it so it only contains only your changes.

As for your description, the paragraph right before you're referring to says:

> Now connect the timeout() signal of each of the Timer nodes.

PR #1549, opened 2018-06-29
fixed confusion in your_first_game tutorial

mrwanashraf:
earlier today i created my first pull request ever in godot and in github in general and i unintentionally caused conflict in the branch, so in this PR I have resolved the conflict.

here is my previous [PR](https://github.com/godotengine/godot-docs/pull/1548) that caused the conflict.

quoting from the previous PR:

>if you head just before func _on_MobTimer_timeout(): section in the tutorial you will find out that there is no steps for connecting the timeout signal to the MobTimer node which makes the one who follows the tutorial becomes confused because then if i blindly follow the tutorial i would just write the code that is shown in func _on_MobTimer_timeout(): and won't connect the signal as there was no step required for it in the tutorial.

the commit is kinda messy due to the new lines apologies for that.

the original commit is from line 769 till line 772

### DIFF
--- a/getting_started/step_by_step/your_first_game.rst
+++ b/getting_started/step_by_step/your_first_game.rst
@@ -140,7 +140,7 @@ node, so we'll add a script. Click the ``Player`` node and click the
 In the script settings window, you can leave the default settings alone. Just
 click "Create":
 
-.. note:: If you're creating a C# script or other languages, select the 
+.. note:: If you're creating a C# script or other languages, select the
             language from the `language` drop down menu before hitting create.
 
 .. image:: img/attach_node_window.png
@@ -157,14 +157,14 @@ Start by declaring the member variables this object will need:
 
     export (int) var speed  # How fast the player will move (pixels/sec).
     var screensize  # Size of the game window.
- 
+
  .. code-tab:: csharp
 
     public class Player : Area2D
     {
-        [Export] 
+        [Export]
         public int Speed = 0; // How fast the player will move (pixels/sec).
-        
+
         private Vector2 _screenSize; // Size of the game window.
     }
 
@@ -179,7 +179,7 @@ the ``Player`` node and set the speed property to ``400``.
 
 .. image:: img/export_variable.png
 
-The ``_ready()`` function is called when a node enters the scene tree, 
+The ``_ready()`` function is called when a node enters the scene tree,
 which is a good time to find the size of the game window:
 
 .. tabs::
@@ -193,7 +193,7 @@ which is a good time to find the size of the game window:
     public override void _Ready()
     {
         _screenSize = GetViewport().GetSize();
-    }   
+    }
 
 Now we can use the ``_process()`` function to define what the player will do.
 ``_process()`` is called every frame, so we'll use it to update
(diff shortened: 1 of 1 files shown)


getting_started/step_by_step/your_first_game.rst
```diff
@@ -766,6 +766,11 @@ well as its position.
 Note that a new instance must be added to the scene using
 ``add_child()``.
 
+Now click on ``MobTimer`` in the scene window then head to inspector window,
```
cbscribe: This seems like a strange place for this. The previous paragraph mentions `_on_MobTimer_timeout()`, which is created when you connect the signal. I understand you want to be more explicit about connecting the signal, but if you're going to add this, it should be earlier. Probably around line 729 where the three connections are made, but only the other two are explicitly described.

mrwanashraf: @cbscribe  you have a point right there and i totally agree with you, but for me i think its more natural ( for me to understand and follow along with the tutorial without getting lost or confused ) to put it this way as after it there would be the step of actually writing the code for that function.

cbscribe: OK, but then to keep the tutorial's flow, you'd have to do more rewriting than just inserting this paragraph.

You also have to change the text above in the two places:  1) line 729, where you're already told to create the connection (after your change you're now being told twice, which is arguably more confusing), and 2) line 760, where the connected method is mentioned as already existing. It's not "natural" to discuss these things out of order. 

As it is written, you connect all three signals, then fill each of them out one at a time. The mob signal just has more description inserted because more is going on. Maybe I could see moving the description until after the code, but MobTimer is no different from StartTimer or ScoreTimer, which are filled out the same way.

mrwanashraf: i have made a new commit and it added a little hint to avoid the confusion on line 729 instead of inserting that paragraph on line 769, does it make it better now ?

getting_started/step_by_step/your_first_game.rst
```diff
@@ -709,24 +709,24 @@ function to set everything up for a new game:
...
         var startTimer = (Timer) GetNode("StartTimer");
         var startPosition = (Position2D) GetNode("StartPosition");
-        
+
         player.Start(startPosition.Position);
         startTimer.Start();
     }
 
-Now connect the ``timeout()`` signal of each of the Timer nodes.
+Now connect the ``timeout()`` signal of each of the Timer nodes (``StartTimer``, ``ScoreTimer`` & ``MobTimer``).
```
cbscribe: Please change the "&" to ", and" and I'll happily merge.

getting_started/step_by_step/your_first_game.rst
```diff
@@ -709,24 +709,24 @@ function to set everything up for a new game:
...
         var startTimer = (Timer) GetNode("StartTimer");
         var startPosition = (Position2D) GetNode("StartPosition");
-        
+
         player.Start(startPosition.Position);
         startTimer.Start();
     }
 
-Now connect the ``timeout()`` signal of each of the Timer nodes.
+Now connect the ``timeout()`` signal of each of the Timer nodes (``StartTimer``, ``ScoreTimer``,and ``MobTimer``).
```
cbscribe: You missed the space after the comma. Then can you please squash the commits so there's only one? It helps keep the commit log clean.

getting_started/step_by_step/your_first_game.rst
```diff
@@ -726,7 +726,7 @@ function to set everything up for a new game:
         startTimer.Start();
     }
 
-Now connect the ``timeout()`` signal of each of the Timer nodes (``StartTimer``, ``ScoreTimer`` & ``MobTimer``).
+Now connect the ``timeout()`` signal of each of the Timer nodes (``StartTimer``, ``ScoreTimer``,and ``MobTimer``).
```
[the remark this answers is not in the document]
mrwanashraf: @cbscribe  i tried squashing the commits and it really worked but then i found out it logged another commit so i thought that reverting would help removing the past commits and keeping the log clean but it didn't, anyways the last commit (fc666a8) is supposed to replace (060bf34), sorry about these mistakes I'm new to PR.

cbscribe:
Thanks for your help! Working with git will get easier with practice. Please continue to contribute.

mrwanashraf:
@cbscribe  thanks for your patience ^^, i will keep practicing and hopefully next PR will be less boring.

PR #1552, opened 2018-06-30
Update config example function signature

rminderhoud:
The previous example caused godot to display the following warning:

`Warning: module 'module' uses a deprecated 'can_build' signature in its config.py file, it should be 'can_build(env, platform)'.`


akien-mga:
Thanks, that's correct.

PR #1555, opened 2018-07-01
Spelling: Various rework

comradekingu:


comradekingu:
Most of these spotted by https://hosted.weblate.org/user/yurchor/

NVIDIA as per https://www.nvidia.com/nb-no/geforce/

development/file_formats/tscn.rst
```diff
@@ -372,7 +372,7 @@ An animation resource consists of tracks. Besides, it has 'length', 'loop' and
 
 1) :code:`length` and :code:`step` are both time in seconds
 
-Each track is described by a list of key, value pairt in the format :code:`tracks/Id/Attribute`,
+Each track is described by a list of key, value pair in the format :code:`tracks/Id/Attribute`,
```
comradekingu: Not 100% sure here

clayjohn:
Most of this is really great! I would like to point out that you should not be capitalizing 'e.g.' and 'i.e.' Unless they are at the beginning of a sentence, in which case you only capitalize the first letter. 

Another thing to note is that you should only capitalize after a colon when the following word must be capitalized (if its an acronym for example). Apparently [some style guides](https://www.grammarly.com/blog/capitalization-after-colons/) allow for words after a colon to be capitalized sometimes, but it seems to be the exception not the rule.

tutorials/viewports/viewports.rst
```diff
@@ -24,8 +24,8 @@ skybox.
 
 :ref:`Viewports <class_Viewport>` have a variety of use cases including: 
 
-- Rendering 3d objects within a 2d game
-- Rendering 2d elements in a 3d game
+- Rendering 3d objects within a 2D game
```
clayjohn: The first '3d' should also be capitalized. Same goes with the line below.

comradekingu:
@clayjohn I am sure i have seen "E.G" as valid alternative spelling somewhere. Just when I had made sense of truncation in English. Big versal following colon is however valid, and makes for fewer errors in translation.

clayjohn:
I have never seen 'e.g.' capitalized personally. I looked it up to make sure and every resource I found said to spell it lowercase. Maybe this is a language difference thing too?

As for the colon if it helps with translations then its a good idea. From what I could find the convention is mixed depending on where you are, so it shouldn't be an issue anyway. Its just a convention I am unfamiliar with.

community/contributing/pr_workflow.rst
```diff
@@ -54,11 +54,11 @@ The branches on the Git repository are organized as follows:
    occurs. As a development branch, it can be unstable
    and is not meant for use in production. This is where PRs should be done
    in priority.
--  The stable branches are named after their version, e.g. ``3.0`` and ``2.1``.
+-  The stable branches are named after their version, E.G. ``3.0`` and ``2.1``.
```
akien-mga: I've never seen "i.e." and "e.g." spelled with capital letters.

https://simple.wikipedia.org/wiki/Exempli_gratia
https://simple.wikipedia.org/wiki/List_of_Latin_phrases_(I)#id_est

community/contributing/pr_workflow.rst
```diff
@@ -122,10 +122,10 @@ You only need to do the above steps once, as long as you keep that local
 metadata is hidden in its ``.git`` subfolder).
 
 .. note:: *Branch it, pull it, code it, stage it, commit, push it, rebase
-          it... technologic.*
+          it… technologic.*
 
           This bad take on Daft Punk's *Technologic* shows the general
-          conception Git beginners have of its workflow: lots of strange
+          conception Git beginners have of its workflow: Lots of strange
```
akien-mga: As far as I know you don't start with a capital letter after ":" nor ";".

community/contributing/pr_workflow.rst
```diff
@@ -418,9 +418,9 @@ will raise an error:
     $ git push origin better-project-manager
     To https://github.com/akien-mga/godot
      ! [rejected]        better-project-manager -> better-project-manager (non-fast-forward)
-    error: failed to push some refs to 'https://akien-mga@github.com/akien-mga/godot'
+    error: Failed to push some refs to 'https://akien-mga@github.com/akien-mga/godot'
```
akien-mga: No, we're not changing the case in copy-pasted error messages from other programs...

development/file_formats/tscn.rst
```diff
@@ -63,7 +63,7 @@ so on. For example, a spatial node looks like:
 The Scene Tree
 --------------
 
-The scene tree is made up of ... nodes! The heading of each node consists of
+The scene tree is made up of … nodes! The heading of each node consists of
```
akien-mga: IMO this conversion should be handled by the theme itself, not in the raw text files. But it's up for discussion.

akien-mga:
Some of those changes were merged already with #1554 from @yurchor. I don't agree with all changes, so I guess I'll have a go at squashing and rebasing myself, reverting the unnecessary changes. Lots of good typo fixes though.

akien-mga:
Merged manually as eb6bdf40. Thanks for the contribution!

For future changes, please open a discussion prior to enforcing new policies such as "capital letter after colon", "three dots should be written as unicode character" and "e.g./i.e. should be capitalized".
On one hand such changes might not be consensual and thus need discussion to define our standard, and on the other hand they need to be applied consistently in the whole documentation, and not just in a handful of pages.

PR #1558, opened 2018-07-02
Removed duplicated function in button hooking GDScript example in

hemagx:
step_by_step/scripting.rst
The final example at the mentioned tutorial had the function _on_Button_pressed as _on_button_pressed in the GDScript
example, this commit shall remove the duplicated function _on_button_pressed
Fixes #1556

Signed-off-by: Ibrahim Zidan <brahem@aotsw.com>

PR #1572, opened 2018-07-06
Removed double function from GDScript example

bbcooper29:
_on_Button_pressed() function was doubled in final GDScript example.


vnen:
This is the same as #1558, which wasn't cherry-picked yet.

bbcooper29:
Sorry, didn't see.

PR #1575, opened 2018-07-07
Change comment for `s.queue_free()`

jonwolski:
In the previous section, we document `free` as being immediate, and here we contrast that with `queue_free`. However, each function has the same comment implying that both `free` and `queue_free` are immediate.

I believe the point of this documentation is to point out that `queue_free` is _not_ necessarily immediate and will wait until it is safe to free the node.

### DIFF
--- a/getting_started/step_by_step/scripting_continued.rst
+++ b/getting_started/step_by_step/scripting_continued.rst
@@ -327,13 +327,13 @@ This erases the node safely during idle.
  .. code-tab:: gdscript GDScript
 
     func _someaction():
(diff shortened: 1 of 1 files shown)


mhilbrunner:
Thanks for this, and congrats to your first docs contribution :)

YeldhamDev:
"becomse"?

jonwolski:
D'oh. @YeldhamDev, good catch. I'll issue an additional PR to clean up after myself. It won't be until later tonight (GMT-8)

PR #1582, opened 2018-07-10
Improved the grammar in some places.

NigelTicknor:
Reworded incorrect uses of "less" to make for easier reading.


mhilbrunner:
Thank you!

PR #1588, opened 2018-07-12
Synced FPS tutorial zip files with the change from PR #1580

### LINKED ISSUE #1563: Move_And_Slide no longer works for the FPS tutorial
Hey guys!

While I was investigating why the FPS project does not open in Godot 3.0.4 for [this issue](https://github.com/TwistedTwigleg/Godot_FPS_Tutorial/issues/11), when I realized the code for moving the player no longer works. Currently the code for `Move_And_Slide` is the following:

`vel = move_and_slide(vel,Vector3(0,1,0), true, 0.05, 4, deg2rad(MAX_SLOPE_ANGLE))`

The problem is the new additional argument (the `true` argument in the statement above) was moved to the end and made an optional argument, meaning we can use the same code we were using before:

`vel = move_and_slide(vel,Vector3(0,1,0), 0.05, 4, deg2rad(MAX_SLOPE_ANGLE))`

And then as a added bonus, the code will work with Godot 3.0 again!

I'll try to get around to fixing this, updating the zip files, and making a PR updating the GitHub repositories when I'm back form vacation (if no one else has got it by then). I just wanted to make this issue here so anyone who tries to go through the tutorial and stumbles upon this issue while I'm gone will know what's happening.

(Unrelated, but I figured out why you cannot open the project in Godot 3.0.4, and it totally makes sense! You cannot open projects opened in future versions of Godot. Sadly, this took me awhile to figure out 🤣 )

TwistedTwigleg:
Now the zip files should work with the latest daily Godot build. Fully closes #1563


mhilbrunner:
Awesome :)

PR #1591, opened 2018-07-13
FPS Tutorial Part05-06 general phrase corrections

Socterean:
Corrections highlights:
1.  Removed the round brackets from the code added to Player.gd when implementing grenades at line 351 and 353 so now the code follows the general coding standard used in the tutorial and removes any confusion on why they are there for the new users
2. Changed the order in if/else from Turret.gd script in fire_bullet() function because in the code explanation at first it is explained about using a RayCast to shoot at the player and only after that is explained using Bullets to shot at the player, so it would create confusion trying to read the explanation following the code as long as they where swapped
3. Other minor typo corrections and changes to improve the tutorial readability and to be more easily understood by users following this tutorial

NOTE: At line 488 when adding changes to the Player.gd script to implement "Grabbing and throwing objects" ability the "if ray_result:" expression is a bit misleading because no such coding convention was used prior to that and no extra explanation was given so we can understand what that line is actually doing. I personally didn't understood why the line is there and what does it do, and I am thinking that new users with lesser coding experience would be confused on what is actually happening and how to reuse and adapt the code for their own projects.

### DIFF
--- a/tutorials/3d/fps_tutorial/part_five.rst
+++ b/tutorials/3d/fps_tutorial/part_five.rst
@@ -323,7 +323,7 @@ Okay, now let's start making the grenades work with the player. Add the followin
 * ``sticky_grenade_scene``: The sticky grenade scene we worked on earlier.
 * ``GRENADE_THROW_FORCE``: The force at which the player will throw the grenades at.
          
-Most of these variables are similar to how we have out weapons set up.
+Most of these variables are similar to how we have our weapons set up.
 
 .. tip:: While it's possible to make a more modular grenade system, I found it was not worth the additional complexity for just two grenades.
          If you were going to make a more complex FPS with more grenades, you'd likely want to make a system for grenades similar to how we have the weapons set up.
@@ -348,9 +348,9 @@ Now we need to add some code in ``_process_input`` Add the following to ``_proce
             grenade_amounts[current_grenade] -= 1
             
             var grenade_clone
-            if (current_grenade == "Grenade"):
+            if current_grenade == "Grenade":
                 grenade_clone = grenade_scene.instance()
-            elif (current_grenade == "Sticky Grenade"):
+            elif current_grenade == "Sticky Grenade":
                 grenade_clone = sticky_grenade_scene.instance()
                 # Sticky grenades will stick to the player if we do not pass ourselves
                 grenade_clone.player_body = self
@@ -447,7 +447,7 @@ any additional grenades.
 
 ______
 
-Now you should be able to throw grenades now! Go give it a try!
+Now you should be able to throw grenades! Go give it a try!
 
 
 Adding the ability to grab and throw RigidBody nodes to the player
@@ -485,7 +485,7 @@ With that done, all we need to do is add some code to ``process_input``:
             var ray_to = ray_from + camera.project_ray_normal(center_position) * OBJECT_GRAB_RAY_DISTANCE
             
             var ray_result = state.intersect_ray(ray_from, ray_to, [self, $Rotation_Helper/Gun_Fire_Points/Knife_Point/Area])
-            if ray_result:
+            if ray_result != null:
                 if ray_result["collider"] is RigidBody:
                     grabbed_object = ray_result["collider"]
                     grabbed_object.mode = RigidBody.MODE_STATIC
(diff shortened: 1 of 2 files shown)


Socterean:
4. Also I have fixed the node at line 887 so now it is displayed properly in a Note box

TwistedTwigleg:
Thanks Socterean! The changes look good!

>NOTE: At line 488 when adding changes to the Player.gd script to implement "Grabbing and throwing objects" ability the "if ray_result:" expression is a bit misleading because no such coding convention was used prior to that and no extra explanation was given so we can understand what that line is actually doing.

I probably should have explained what that line is doing, as I agree it can be confusing if you have not gone through the [ray-casting section](http://docs.godotengine.org/en/latest/tutorials/physics/ray-casting.html) of the documentation.

As for what the code does, I will try to explain:

The code sends a `Raycast` from the center of the `Camera`, going `OBJECT_GRAB_DISTANCE` far in length. The `Raycast` has two exceptions, the player's `KinematicBody` and the knife's `Area` node, so they do not interfere with the results of the `Raycast`.

The results of the `Raycast` are stored in `ray_result`. This follows the coding convention in the ray-casting section of the documentation, though as I said earlier I should have explained this part in the tutorial.

If `state.interesct_ray` does not find anything, it returns `null`, so we can check if the `Raycast` found anything by simply calling `if ray_result` as it will return true if `ray_result` is not equal to `null`. From there we check to see if the collider the `Raycast` collided with is a `RigidBody`, and if it is we do additional processing so the player grabs it.

Hopefully that explains what is going on in that section of code. At some point I (or someone else) should probably add an explanation for that bit of code, as I can totally see why it would be confusing.

Thanks again for the changes!

tutorials/3d/fps_tutorial/part_five.rst
```diff
@@ -533,29 +533,29 @@ themselves or the knife's collision :ref:`Area <class_Area>`.
 Then we check to see if we got a result back from the ray. If we have, we then see if the collider the ray collided with is a :ref:`RigidBody <class_RigidBody>`.
 
 If the ray collided with a :ref:`RigidBody <class_RigidBody>`, we set ``grabbed_object`` to the collider the ray collided with. We then set the mode on
-the :ref:`RigidBody <class_RigidBody>` we collided with to ``MODE_STATIC`` so it does not move.
+the :ref:`RigidBody <class_RigidBody>` we collided with to ``MODE_STATIC`` so it does't move in our hands.
```
TwistedTwigleg: `does't` -> `doesn't`

Socterean: I will fix it as soon as I'm done with Part06 in the next hour or so, thanks for pointing it out to me.

Socterean:
About "if ray_result:", it would not work if written like this "if ray_result != null:" ? because I believe if it works it would lift some of the confusion even without major changes to the explanation because it is a bit self-explanatory.

TwistedTwigleg:
> About "if ray_result:", it would not work if written like "if ray_result != null:" ?

It should still work. It should probably be tested (just to make sure) but it should work since both are supposed to be checking the same thing, the first is just a shorter (and more confusing) way to achieve the same result.

If it works, I have no problem with the change! We'd need to update the ZIP flies for the tutorial, but that's not a big problem (and we can change the zips in a separate PR)

Thanks again for the corrections!

Socterean:
I have tested "if ray_result != null" and it seems to work as intended. I will make the change to be more clear and more easy to understand for beginners following this tutorial if that it's okay.

Socterean:
With this I believe I have completed this tutorial's general phrase corrections. And I believe now it should be error free. If some of my modifications are incorrect please feel free to point them out to me and I will modify them accordingly. And thank you for being this open with my endeavor to contributing to the Godot project. I hope that all this would benefit many users that are trying to learn how to use the engine and would help them get going in their own projects. And also thank you @TwistedTwigleg for making this great tutorial in the first place, it has helped me a lot in learning GDScript and Godot.

TwistedTwigleg:
> I have tested "if ray_result != null" and it seems to work as intended. I will make the change to be more clear and more easy to understand for beginners following this tutorial if that it's okay.

The change is cool with me.

> With this I believe I have completed this tutorial’s general phrase corrections (except for part01 which I plan to review when I have some more spare time).

Nice! Thanks so much for doing this @Socterean! I really appreciate the time and effort you have put into these pull requests!

cbscribe:
Is this ready to go? If so, please squash the commits - thanks!

Socterean:
Squashing done, now this is ready to go, thanks!

mhilbrunner:
Thanks for this :)

PR #1593, opened 2018-07-15
Clarify class constructor call order

wlycdgrfromflatiron:


cbscribe:
It looks like something went wrong with your commit history here. Please see http://docs.godotengine.org/en/latest/community/contributing/pr_workflow.html for the proper workflow.  I'll close this one and you can re-submit.

PR #1595, opened 2018-07-20
Fix a few typos in part 1 of the FPS tutorial.

rcorre:
For "Let's look is ``process_movement`` next", the following text is
actually referring to "process_input", and even says "and now back to
process_input" later in the same section.


mhilbrunner:
Thanks for this contribution :)

PR #1600, opened 2018-07-21
Updating deprecated pywin32 link

kavika13:
Used to be on Sourceforge, but if you check the link then it points you at their github releases page


akien-mga:
Thanks!

PR #1619, opened 2018-07-29
Added some info about the branches in the GDNative repositories

BastiaanOlij:
Added some explanation to both the C and C++ documentation for GDNative modules that the godot_headers and godot-cpp repository now have branches. The correct branch needs to be used that matches the version of Godot you're working against. GDNative modules build for older versions of Godot will keep working on newer versions but not the other way around.


BastiaanOlij:
Rebased my changes, @Calinou you might want to recheck this, I think I've managed to keep your improvements in tact but there are some changes that you might want to give some input and corrections on :)

PR #1631, opened 2018-07-31
Fixed script block

arrowgamer:
A script block was not displaying correctly due to incorrect indentation


cbscribe:
Good fix - thanks for your contribution!

PR #1633, opened 2018-07-31
Improve introduction of 2D transforms

FeralBytes:
Removed the odd wording intended to imply that the document is discussing very low level details.


akien-mga:
Thanks!

PR #1639, opened 2018-08-02
Add mac-specific and vscode-specific considerations

lawsonry:
Setting up Godot for development on a Mac was very frustrating for a few people and me. We spent way too long trying to figure out why it wasn't working, only to realize that the latest mono stable release is a gray button instead of the bright blue button on the Mono download page. Additionally trying to use `brew` or anything else to install Mono will not get you the version you need. 

I updated to docs to just point out the version discrepancies that I ran into inadvertently. This should help future Mac users not pull their hair out trying to get started with Godot and C#.


cbscribe:
Thanks!

PR #1645, opened 2018-08-03
Update GDscript operators.

### LINKED ISSUE #1524: Update GDScript operators table
-> http://docs.godotengine.org/en/latest/getting_started/scripting/gdscript/gdscript_basics.html#operators

Qwertystop on Discord:
>It's a list of operators with general meanings, but look at the row for + -. It just says addition and subtraction – I found by experimentation that + also concatenates arrays, which is useful because there's no method for it. Are there other places + is useful? Maybe, maybe not, the docs don't seem to say. Unary - (negation) is also undocumented, though most users will use it without thinking about it on numeric literals, and hopefully from there realize it works on numeric variables as well. Another example: get_node doesn't document the $ sigil; that's only in the tutorial.

Also, modulo is missing. Should be documented (and say that it is `int` only)

FeralBytes:
Fixes #1524

I would definitely like feedback on this one. I am not sure if I documented "$" correctly. I believe this shorthand only works for children nodes correct?
Of course please let me know if there are any other errors. Thank you.


cbscribe:
`$` works with any node path, not just child nodes - it's just a shorthand for `get_node()`. For example, you can do

    $"../Node/Node"

FeralBytes:
@cbscribe updated.

PR #1646, opened 2018-08-03
Improve writing style in "GDNative C++ example"

Calinou:
This improves the writing style in the **GDNative C++ example** page, making it easier to read. This also shortens the Git submodule initialization commands into a single one, `git submodule --init update`.


NathanLovato:
Thanks!

PR #1647, opened 2018-08-04
Add intro to shaders video series (with Bastiaan Olij)

NathanLovato:
We've made a 7 videos series with @BastiaanOlij to get people started with shaders. I asked Akien if he wanted it in the docs, and the lad said yes. So here it is! Added to the toc tree, I built the page to check that everything's fine.

----

By the way, building the docs in sphinx with the `-j` option doesn't work for me. I have a 2-core CPU, 4 threads, but regardless of the option sphinx will only build on 1 thread/1 core. Is that a limitation of html builds? I understand it should build the menus and all first to append them to every page, but using all the available cpu power wouldn't hurt...


cbscribe:
This is a great series. As it stands it's just a link to the videos, do you have any plans to expand the writeup?

Re: the -j option, I looked into this a while back. From what I recall it only applies to the write stage of the build, which is the least CPU-intensive. It happens after all of the extension processing, which is mostly singlethreaded.

BastiaanOlij:
@cbscribe seeing that building from clean source without -j8 takes about 30 minutes on my computer while it takes about 1 minute with this switch, I don't think that assertion is right :)

It does compile multiple source files in parallel and it does turn out to make a huge difference.

cbscribe:
Hmm, well I only get one thread and I've *never* seen it come close to one minute. As Nathan said, -j seems to have no effect for me. The googling I did suggested that single threading was the case, but there wasn't a lot of information to be had in a casual search.

Are you really getting one minute compiles after a `make clean`? That must be nice!

NathanLovato:
I thought I had pywin32 on the path but noticed compiling Godot from scratch was much slower than last time yesterday, and it only used a single thread, so that may be it!

To answer the text question, no, no plans for that. The videos and sources are there if anyone ever wants to do it. But otherwise it takes too much time, and I don't want to take on the maintenance

cbscribe:
OK, well let's put this in and if anyone wants to take on transcription, they can add it.

PR #1653, opened 2018-08-05
Add note about target=release_debug

FeralBytes:
To help out future users adding extensions like:
https://www.reddit.com/r/godot/comments/94qyc3/a_note_about_building_from_source_from_a/

### DIFF
--- a/development/compiling/compiling_for_windows.rst
+++ b/development/compiling/compiling_for_windows.rst
@@ -176,6 +176,11 @@ root directory of the engine source code and type:
 
     C:\godot> scons platform=windows
 
+.. note:: If you are compiling Godot for production use and not to hack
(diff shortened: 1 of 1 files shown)


neikeq:
I thinks this would be misleading. There is not much point in building templates with `release_debug`, it's meant for the editor. In fact, there is no option for `release_debug` templates in the editor. You want either `debug` or `release`.

neikeq:
The only reason to have `release_build` templates is if you want an optimized build but with better errors.

Calinou:
@neikeq I think the notice is about building the editor with `release_debug`, not the export templates, which is why I suggested moving it.

FeralBytes:
Good catch @neikeq and @Calinou, sorry, I should have quality checked my work.

PR #1659, opened 2018-08-05
Document more feature tags

Calinou:
This adds more tags to the Feature tags page, including the [recently-added `editor` and `standalone` feature tags](https://github.com/godotengine/godot/pull/20464). I also clarified what the `debug` feature tag exactly does — it's also effective when running a project from an editor build.


cbscribe:
Looks great!

PR #1672, opened 2018-08-07
Add clarifying paragraph on encryption page

vurpo:
I wouldn't want to mislead people into thinking that encrypting and decrypting a save file locally on the player's computer can prevent editing and sharing of save files. It can't. The only way to prevent unauthorized editing and sharing of save data is to keep the save data on a remote server, and then running the game logic on that server instead of locally on the player's computer.

### DIFF
--- a/tutorials/io/encrypting_save_games.rst
+++ b/tutorials/io/encrypting_save_games.rst
@@ -29,6 +29,13 @@ their own irrelevance would again take over their life.
 No, we definitely do not want this to happen, so let's see how to
 encrypt savegames and protect the world order.
 
+(NOTE: This method can not really prevent players from editing their savegames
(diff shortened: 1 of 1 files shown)


tutorials/io/encrypting_save_games.rst
```diff
@@ -29,6 +29,13 @@ their own irrelevance would again take over their life.
 No, we definitely do not want this to happen, so let's see how to
 encrypt savegames and protect the world order.
 
+(NOTE: This method can not really prevent players from editing their savegames
```
akien-mga: I'd move it at the end of the page to avoid spoiling the fun of the "Why?" section, and use the `.. note::` markup so that it appears as a note box.

mhilbrunner:
Agree with Akien about moving this to the end of the page.

PR #1674, opened 2018-08-09
Minor fixes to the FAQ page

YeldhamDev:

PR #1675, opened 2018-08-09
Fix List item formatting

lawsonry:
Fixed a list item format issue when RST is compiled by just removing it and turning it into a declarative statement.


lawsonry:
Just saw #1674, which fixes this issue and also addresses some other reST semantics. Closing this in favor of that one.

PR #1677, opened 2018-08-10
Remove some dead links

R3ality:
Happened upon some dead links while browsing the docs, opted to remove them due to ...
- was not able to locate the page/section of the docs where they should (used to?) have been referring to. It may have been something which has since been refactored/removed.
- and feel that the step-by-step section of the docs should ...
  - limit the amount of links to a minimum so to not confuse the reader (ie. to avoid questions such as "do I need to read that part before I continue with this part?", "is that part just optional extra reading if in case I want to get more deep knowledge?" or "if I don't read that part now, will it be linked again at a later point in the document?"
  - and not link at all in cases where the target is also part of the step-by-step guide (since the intent is to have the reader eventually reach that anyway, step by step. Or if the linked information is relevant to the page on hand, it should be included in this page or relocated to a page prior to this page)

### DIFF
--- a/getting_started/step_by_step/ui_introduction_to_the_ui_system.rst
+++ b/getting_started/step_by_step/ui_introduction_to_the_ui_system.rst
@@ -24,9 +24,6 @@ This guide will get you started with UI design. You will learn:
    containers
 -  The five most common containers
 
-To learn how to control the interface and connect it to other scripts,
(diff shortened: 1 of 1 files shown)


cbscribe:
These links seem to have been updated/removed in master, so I'm closing.

PR #1681, opened 2018-08-13
Improve the "Changing application icon for Windows" page

Calinou:
Advice specific to Windows XP was removed as Godot does not support Windows XP.


NathanLovato:
Thanks!

PR #1682, opened 2018-08-14
Added note about matrix builds from another matrix(in shaders)

Chaosus:
This is recently added feature for shaders.


mhilbrunner:
Yay, more shader docs!

PR #1686, opened 2018-08-17
Update min and target sdk numbers

FeralBytes:


cbscribe:
The template link is also broken, but I'll fix that after merge. Thanks.

Issue #1666, opened 2018-08-06
Add instancing by code link to "step by step instancing"


FeralBytes:
I plan to create  PR to add a link to [Scripting Continued: Instancing](http://docs.godotengine.org/en/latest/getting_started/step_by_step/scripting_continued.html#instancing-scenes) To make this option more obvious to those searching for instancing.

PR #1687, opened 2018-08-17
Add instancing scenes by code.

FeralBytes:
Does my ref work correctly, sorry still learning ReStructed text.


FeralBytes:
Closes #1666

PR #1690, opened 2018-08-19
Changed headers to headers_dir to match godot-cpp

cameron195:
The headers argument in godot-cpp's SConstruct file was renamed to headers_dir, but this change was not reflected by the tutorial.


cbscribe:
Thanks for your contribution!

PR #1694, opened 2018-08-20
Fix FRAGCOORD in shading_language.rst

Chaosus:
1. Place FRAGCOORD of Spatial Fragment on top of the list(for unification view with other lists)
2. Added missing FRAGCOORD for spatial light shader(recently added to master)
3. Fix vec2 -> vec4 FRAGCOORD in CanvasItem Light list


akien-mga:
Thanks!

PR #1695, opened 2018-08-21
Added example of raw "for" loop usage

Chaosus:


Calinou:
This could look confusing at the first glance. `range()` should probably be the recommended syntax since it's more readable and is much more commonly seen in other languages, such as Python.

PR #1701, opened 2018-08-24
Typo in "class_undoredo.rst"

char0xff:


vnen:
You shouldn't edit this file here since it's generated, as stated in the beginning of the file:

```
.. DO NOT EDIT THIS FILE, but the UndoRedo.xml source instead.
```

This should be edited int the XML doc in the main Godot repository. The same source is used for the class reference in-editor.

PR #1705, opened 2018-08-26
Remove text for add_icon_item that describes a required argument as optional

jshorty:
With 3.0.1 this argument is required, and calling `add_icon_item` with only a `texture` and `label` raises:
```
Invalid call to function 'add_icon_item' in base 'OptionButton': Expected 3 arguments.
```


cbscribe:
This is a good catch, but class docs are autogenerated, and must be edited in the main repo. See [Updating the Class Reference](http://docs.godotengine.org/en/3.0/community/contributing/updating_the_class_reference.html).

jshorty:
Gotcha— looks like it's getting resolved by actually making the argument optional:
https://github.com/godotengine/godot/issues/18686

PR #1706, opened 2018-08-26
Explain order in which nodes receive input events

romlok:
Some clarifying documentation for people who like me who misunderstood how input events propagate through the scene tree.

Ref: godotengine/godot#21042


mhilbrunner:
Thanks for this clarification! :)

PR #1717, opened 2018-08-31
Renamed simple.gdns occurrences to Simple.gdns

Turupawn:
The tutorial was using both uppercase and lowercase names interchangeably that led me to errors.
In this PR, I went for the uppercase name `Simple.gdns` to match the 4th gif of the tutorial.


vnen:
The consistence is very important indeed, since most OSes have case-sensitive filesystems. However I prefer using lowercase names for all files. I know that this would mean redoing the GIF, so I understand if it's easier to change only the text to match.

Turupawn:
@vnen Yeah I got the error because I'm using Linux.
And I agree, I also prefer lowercase. So I redid the gif and did #1718 where everything is lowercase.
Should we drop this PR and use the other one instead?

akien-mga:
Superseded by #1718.

PR #1718, opened 2018-09-01
Simple.gdns renamed to simple.gdns on GDNative C tutorial

Turupawn:
This is a follow up of #1717 but in this case I renamed lowercase on both the text and gif.


akien-mga:
Looks good, thanks!

PR #1721, opened 2018-09-02
Mention the 32x32 image requirement in the custom mouse cursor tutorial

voithos:
Fixes https://github.com/godotengine/godot/issues/17018


voithos:
Ah, thanks! Hadn't seen that.

PR #1724, opened 2018-09-02
rename var in setget section to avoid confusion

Razzeeyy:
changed my_var to myvar so that example actually compilable and doesn't cause a confusion to the shallow reader

Hello!
I was reading the docs and I found this part a bit confusing (especially if it would be a newcomer with less prior knowledge of the idiom), so I decided to try and fix that part so it's less confusing.
Thank you for such a great engine.

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_basics.rst
+++ b/getting_started/scripting/gdscript/gdscript_basics.rst
@@ -1228,13 +1228,13 @@ with the new value. Vice-versa, when ``variable`` is accessed, the *getter* func
 
 ::
 
-    var myvar setget my_var_set, my_var_get
+    var my_var setget my_var_set, my_var_get
 
     func my_var_set(new_value):
-        myvar = new_value
+        my_var = new_value
 
     func my_var_get():
-        return myvar # Getter must return a value.
+        return my_var # Getter must return a value.
 
 Either of the *setter* or *getter* functions can be omitted:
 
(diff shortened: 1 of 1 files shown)


getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1231,10 +1231,10 @@ with the new value. Vice-versa, when ``variable`` is accessed, the *getter* func
     var myvar setget my_var_set, my_var_get
 
     func my_var_set(new_value):
-        my_var = new_value
+        myvar = new_value
```
vnen: It seems to me it was supposed to be `my_var`. It's the previous declaration that is wrong. For instance, there are other references to `my_var` in the next code block.

neikeq: Agree, I would rename the declaration instead.

Razzeeyy: good catch! lemme fix that once more 😅

cbscribe:
Thanks! Just a note for future contributions - please squash your commits into one. See http://docs.godotengine.org/en/latest/community/contributing/pr_workflow.html if you're not sure how. I'll squash & commit this one.

akien-mga:
Note that this was merged in 3.0, not master. It needs to be cherry-picked for master.

Razzeeyy:
Thanks for squashing it for me.
Sorry I wasn't yet aware of that feature.
I'll try to make use of that feature in the future commits.

I'm glad there is always something new to learn from other people 😏

PR #1732, opened 2018-09-12
Update custom_godot_servers.rst

hungrymonkey:
"but" doesn't seem correct in this context. I wonder how I missed this word.
Before: It is possible to emit signals to GDScript but calling the GDScript dummy object.
After: It is possible to emit signals to GDScript by calling the GDScript dummy object.


hungrymonkey:
Thanks for thanking care of my mistakes.

PR #1737, opened 2018-09-15
Pong tutorial was modified for Godot 3

Ashafix:
As the title says the Pong Tutorial which was written for Godot 2 was modified to run with Godot 3.

Other changes:

- Some minor clean up in the code
- Add the complete code at the end
- Remove the background color section

It looks OK when I ran Sphinx on it.


cbscribe:
This tutorial was left with 2.1 because the general consensus among the docs team was that it is a not a good introduction to Godot. Unlike the "Dodge the Creeps" game, it doesn't demonstrate the node-based project structure very well, and instead does nearly everything in code. Nor does it demonstrate important concepts like instancing, signals, or UI.

If the project were to be reworked (for example to use Godot's collision system instead of position checks in GDScript) then perhaps it could be placed in the godot-demos collection. I don't, however, think it belongs in the "Getting Started" section.

Ashafix:
closing the PR per @cbscribe comments.

PR #1738, opened 2018-09-16
JDK 9 causes gradelw build to fail.

FeralBytes:
Fixing documentation.


FeralBytes:
@Calinou Added JDK 10.

PR #1746, opened 2018-09-18
Fix typo in FPS tutorial part 1

### LINKED ISSUE #1739: Minor Typo in FPS Tutorial Part 1
'In 3D'>'FPS tutorial'>'Part 1'>'Making the FPS movement logic', there is a typo in the (!)Note regarding the `MOUSE_SENSITIVITY` being purposely declared in capitals despite not being const.

> "You may have noticed that` MOUSE_SENSITIVITY` is written in all caps like the other constants, but **is** MOUSE_SENSITIVITY is not a constant."

Calinou:
This closes #1739.

Anyway, I think it should be a constant multiplied by a variable factor (which would be the custom sensitivity setting). Making a variable with an uppercased name not constant sounds like a bad practice to me.

### DIFF
--- a/tutorials/3d/fps_tutorial/part_one.rst
+++ b/tutorials/3d/fps_tutorial/part_one.rst
@@ -274,7 +274,7 @@ You can tweak many of these variables to get different results. For example, by 
 increasing ``JUMP_SPEED`` you can get a more 'floaty' feeling character.
 Feel free to experiment!
 
-.. note:: You may have noticed that ``MOUSE_SENSITIVITY`` is written in all caps like the other constants, but is ``MOUSE_SENSITIVITY`` is not a constant.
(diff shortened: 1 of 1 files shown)


TwistedTwigleg:
Thanks for the CC @mhilbrunner 

> @Calinou said:
> Anyway, I think it should be a constant multiplied by a variable factor (which would be the custom sensitivity setting). Naming a non-constant variable in all-uppercase sounds like bad practice to me.

I agree! Multiplying by a variable factor should work great and will solve the problem nicely.

The only reason it is a constant variable but all uppercase is because it originally was a constant variable, but in a later part of the FPS tutorial I realized I needed it to be a normal variable, and so changed it. When I did the rewrite I decided to just keep it a normal variable but still all uppercase, but in hindsight I should have named it constantly with the other variables.

Anyway, all this is to say I agree with changing it back into a constant and then using a variable factor for changing mouse sensitivity. Also, Thanks for fixing the typo!

PR #1752, opened 2018-09-20
Improve the "Internationalizing games" page

Calinou:
This also replaces screenshots with smaller (and more readable) ones.


tutorials/i18n/internationalizing_games.rst
```diff
@@ -6,58 +6,57 @@ Internationalizing games
...
 ------------
 
-Sería excelente que el mundo hablara solo un idioma. Unfortunately for
-us developers, that is not the case. While not generally a big
-requirement when developing indie or niche games, it is also common
-that games going into a more massive market require localization.
-
-Godot offers many tools to make this process more straightforward, so
-this tutorial is more like a collection of tips and tricks.
+*Sería excelente que el mundo hablara solo un idioma.* Unfortunately for
```
pwab: Would it be possible to change the first sentence into something like this:

---

_Sería excelente que el mundo hablara solo un idioma. (It would be excellent if the world spoke only one language)_

Unfortunately for us developers, that is not the case [...]

---

Because I had to translate that first sentence with a bit of effort to understand what that actually means.

mhilbrunner:
I like pwab's suggestion above, besides that, this looks good to go.

NathanLovato:
Thanks for your work! No update since 2 weeks so I'll merge now and change the suggestion from @pwab manually.

PR #1756, opened 2018-09-21
add info about restoring purchases on iOS

jandrewlong:


mhilbrunner:
Thanks for this addition!

Issue #1744, opened 2018-09-18
Compiling for Android Instructions are Wrong


FeralBytes:
I am opening this issue for myself to resolve.
This is how the directions should be simplified.
https://github.com/Kopfenheim/godot-gpgs
Maintainers please assign me for this issue.

Need to add Godotlib.call_defered() as an example and also need to explain that no more than 5 arguments can be sent at a time.

A PR has been opened to fix the documentation below.
From MySingleton.java.
```
import android.content.Intent;
import javax.microedition.khronos.opengles.GL10;

public int myFunction(String p_str) {
		// a function to bind
                //Added return statement that was missing.
		return 1;
	}

//SDK.initializeHere();

protected void onMainActivityResult(int requestCode, int resultCode, Intent data) {}
// For API 21 and above you will want the below if PR#22465 gets approved.
// https://github.com/godotengine/godot/pull/22465
protected void onMainRequestPermissionsResult(int requestCode, String[] permissions, int[] grantResults) {}

protected void onGLDrawFrame(GL10 gl) {}
protected void onGLSurfaceChanged(GL10 gl, int width, int height) {} // singletons will always miss first onGLSurfaceChanged call
```

PR #1767, opened 2018-09-28
Fix the Java Singleton example so it works

FeralBytes:
Reference Issue #1744
Also added the new onMainRequestPermissionsResult()


FeralBytes:
More Changes coming soon, please wait to review.

PR #1768, opened 2018-09-29
Add a small tutorial on using Resource scripts.

willnationsdev:
I can't vouch (at all) for the C# snippets in here, so someone more familiar with Godot's C# syntax will have to proofread those. I'm also not sure if my links are executed properly and whether I can safely nest code snippets within a note or warning (as I've done here).

This should finally help people get a better idea of how useful Resource scripts can be, how to make them, and how to use them.

### DIFF
--- a/getting_started/step_by_step/resources.rst
+++ b/getting_started/step_by_step/resources.rst
@@ -158,12 +158,260 @@ Freeing resources
 Resource extends from :ref:`Reference <class_Reference>`.
 As such, when a resource is no longer in use, it will automatically free
 itself. Since, in most cases, Resources are contained in Nodes, scripts
-or other resources, when a node is removed or freed, all the children
+or other resources, when a node is removed or freed, all the owned
 resources are freed too.
 
 Scripting
 ---------
 
-Like any object in Godot, not just nodes, resources can be scripted,
-too. However, there isn't generally much of an advantage, as resources
-are just data containers.
+Like any Object in Godot, users can also script Resources. As Resources,
+Resource scripts inherit the ability to freely translate between object
+properties and serialized text or binary data (*.tres, *.res). They also
+inherit the reference-counting memory management from the Reference type.
+
+This comes with many distinct advantages over alternative data
+structures such as JSON, CSV, or custom TXT files. Users can only import these 
+assets as a :ref:`Dictionary <class_Dictionary>` (JSON) or as a 
+:ref:`File <class_File>` to parse. What sets Resources apart is their 
+inheritance of :ref:`Object <class_Object>`, :ref:`Reference <class_Reference>`,
+and :ref:`Resource <class_Resource>` features:
+
+- They can define constants, so constants from other data fields or objects are not needed.
+
+- They can define methods, including setter/getter methods for properties. This allows for abstraction and encapsulation of the underlying data. If the Resource script's structure needs to change, the game using the Resource need not also change.
+
+- They can define signals, so Resources can trigger responses to changes in the data they manage.
+
+- They have defined properties, so users know 100% that their data will exist.
+
+- Resource auto-serialization and deserialization is a built-in Godot Engine feature. Users do not need to implement custom logic to import/export a resource file's data.
+
+- Resources can even serialize sub-Resources recursively, meaning users can design even more sophisticated data structures.
+
+- Users can save Resources as version-control-friendly text files (\*.tres). Upon exporting a game, Godot serializes resource files as binary files (\*.res) for increased speed and compression.
+
+- Godot Engine's Inspector renders and edits Resource files out-of-the-box. As such, users often do not need to implement custom logic to visualize or edit their data. To do so, double-click the resource file in the FileSystem dock or click the folder icon in the Inspector and open the file in the dialog.
+
+- They can extend **other** resource types besides just the base Resource.
+
+.. warning::
+    
+    Resources and Dictionaries are also different in that both are passed by reference,
+    but only Resources are reference-counted. This means that if a Dictionary is passed
+    between objects, and the first object is deleted, it will invalidate the second
+    object's variable. This is not so for Resources which won't free their memory until
+    *all* instances are gone.
+
+    .. tabs::
+      .. code-tab:: gdscript GDScript
+
+        extends Node
+
+        class MyObject:
+            extends Object
+            var dict = {}
+
+        func _ready():
+            var obj1 = MyObject.new()
+            var obj2 = MyObject.new()
+            obj1.dict.greeting = "hello"
+            obj2.dict = obj1.dict             # obj2.dict now references obj1's Dictionary
+            obj1.free()                       # obj1 is freed and the Dictionary too!
+            print(obj2.dict.greeting)         # Error! 'greeting' index accessed on null instance!
+
+            # to avoid this, we must manually duplicate the Dictionary
+            obj1 = MyObject.new()
+            obj1.dict.greeting = "hello"
+            obj2.dict = obj1.dict.duplicate() # now we are passing a copy, not a reference
+            obj1.free()                       # obj2's Dictionary still exists
+            print(obj2.dict.greeting)         # prints 'hello'
+
+Godot makes it easy to create custom Resources in the Inspector.
+
+1. Create a plain Resource object in the Inspector. This can even be a type that derives Resource, so long as your script is extending that type.
+2. Set the ``script`` property in the Inspector to be your script.
+
+The Inspector will now display your Resource script's custom properties. If one edits
+those values and saves the resource, the Inspector serializes the custom properties
+too! To save a resource from the Inspector, click the Inspector's tools menu (top right),
+and select "Save" or "Save As...".
+
+If the script's language supports `script classes <https://godot.readthedocs.io/en/latest/getting_started/step_by_step/scripting_continued.html#register-scripts-as-classes>`__, 
+then it streamlines the process. Defining a name for your script alone will add it to
+the Inspector's creation dialog. This will auto-add your script to the Resource
+object you create.
+
+Let's see some examples.
+
+.. tabs::
+  .. code-tab:: gdscript GDScript
+
+    # bot_stats.gd
+    extends Resource
+    export(int) var health
+    export(Resource) var sub_resource
+    export(Array, String) var strings
+
+    func _init(p_health = 0, p_sub_resource = null, p_strings = []):
+        health = p_health
+        sub_resource = p_sub_resource
+        strings = p_strings
+
+    # bot.gd
+    extends KinematicBody
+
+    export(Resource) var stats
+
+    func _ready():
+        # uses an implicit, duck-typed interface for any 'health'-compatible resources
+        if stats:
+            print(stats.health) # prints '10'
+  .. code-tab:: csharp
+
+        // BotStats.cs
+        using System;
+        using Godot;
+
+        namespace ExampleProject {
+            public class BotStats : Resource
+            {
+                [Export]
+                public int Health { get; set; }
+
+                [Export]
+                public Resource SubResource { get; set; }
+
+                [Export]
+                public String[] Strings { get; set; }
+
+                public BotStats(int health = 0, Resource subResource = null, String[] strings = null)
+                {
+                    Health = health;
+                    SubResource = subResource;
+                    Strings = strings ?? new String[0];
+                }
+            }
+        }
+
+        // Bot.cs
+        using System;
+        using Godot;
+
+        namespace ExampleProject {
+            public class Bot : KinematicBody
+            {
+                [Export]
+                public Resource Stats;
+
+                public override void _Ready()
+                {
+                    if (Stats != null && Stats is BotStats) {
+                        Godot.print((Stats as BotStats).Health);
+                    }
+                }
+            }
+        }
+
+.. note::
+
+    Resource scripts are similar to Unity's ScriptableObjects. The Inspector 
+    provides built-in support for custom resources. If desired though, users 
+    can even design their own Control-based tool scripts and combine them 
+    with an :ref:`EditorPlugin <class_EditorPlugin>` to create custom 
+    visualizations and editors for their data.
+
+    Unreal Engine 4's DataTables and CurveTables are also easy to recreate with
+    Resource scripts. DataTables are a String mapped to a custom struct, similar
+    to a Dictionary mapping a String to a secondary custom Resource script.
+
+    .. tabs::
+      .. code-tab:: gdscript GDScript
+
+        # bot_stats_table.gd
+        extends Resource
+
+        const BotStats = preload("bot_stats.gd") 
+
+        var data = {
+            "GodotBot": BotStats.new(10), # creates instance with 10 health
+            "DifferentBot": BotStats.new(20) # a different one with 20 health
+        }
+
+        func _init():
+            print(data)
+      .. code-tab:: csharp
+
+        using System;
+        using Godot;
+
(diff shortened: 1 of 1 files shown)


mhilbrunner:
Thanks for this! I'll proofread - FYI, building the docs yourself locally is a great way to verify how it looks in the output, and isn't too hard: Look at the [Building with Sphinx](https://github.com/godotengine/godot-docs#building-with-sphinx) section of the README in this repository.

willnationsdev:
@mhilbrunner I got it rendering locally (yay!), so I've fixed up a bunch of stuff with the syntax, however I can't seem to figure out how I'm supposed to make the "script classes" link work. It's supposed to link to the "Scripting Continued" page at the "Register scripts as classes" header, but all the variations I've tried (like :ref:\`script classes <scripting_continued#register-scripts-as-classes>\`) don't seem to work.

getting_started/step_by_step/resources.rst
```diff
@@ -158,12 +158,173 @@ Freeing resources
...
+
+- They can define signals, so Resources can trigger responses to changes in the data they manage.
+
+- They have defined properties, so users know 100% that their data will exist.
+
+- Resource auto-serialization and deserialization is a built-in Godot Engine feature. Users do not need to implement custom logic to import/export a resource file's data.
+
+- Resources can even serialize sub-Resources recursively, meaning users can design even more sophisticated data structures.
+
+- Users can save Resources as version-control-friendly text files (\*.tres). Upon exporting a game, Godot serializes resource files as binary files (\*.res) for increased compression.
```
LikeLakers2: Aren't `*.res` files also faster to load?

willnationsdev: Yes, because the same amount of data is packed into a smaller space, so it doesn't take as long to parse the data as input.

willnationsdev: I suppose I could do "for increased speed and compression" instead.

getting_started/step_by_step/resources.rst
```diff
@@ -158,12 +158,173 @@ Freeing resources
...
+    const BotStats = preload("res://bot_stats.gd")
+
+    func _init():
+        if not stats:
+            stats = BotStats.new(10)
+    
+    func _ready():
+        print(stats.health) # prints '10'
+  ..code-tab:: csharp
+        using Godot;
```
LikeLakers2: I think you forgot to add a C# version of `bot_stats.gd`.

willnationsdev: Oh yeah, guess I could add that. Like I said though, I doubt any of the C# code is really accurate since I haven't used Godot C# at all.

getting_started/step_by_step/resources.rst
```diff
@@ -158,12 +158,173 @@ Freeing resources
...
+    
+    CurveTables are the same thing except mapped to an Array of float values
+    or a :ref:`Curve <class_Curve>`/:ref:`Curve2D <class_Curve2D>` resource object.
+
+.. warning::
+
+    With GDScript, it only supports Resource classes at the top level. Defining
+    a Resource subclass, i.e. an inner class, within a GDScript file will have no 
+    effect. If one attempts to save an instance of a Resource subclass, none of 
+    the scripted properties will serialize to the saved file.
```
LikeLakers2: This sounds like a bug to me -- I've saved `ShaderMaterial`s with `Shader` resources inside them before. I don't see why this should be any different.

getting_started/step_by_step/resources.rst
```diff
@@ -158,12 +158,173 @@ Freeing resources
...
+            public override void _Ready()
+            {
+                Godot.print(stats.health);
+            }
+        }
+
+.. note::
+
+   Resource scripts are similar to Unity's ScriptableObjects. However, unlike in
+   Unity, there is no need to implement IMGUI logic to enable users to create
```
williamd1k0: In fact, there is no need to implement immediate gui or custom editor to use a "common" ScriptableObject.

The current workflow (I tested it in Unity 2018.3):

1. Create your `ScriptableObject` class:
```c#
public class CustomData : ScriptableObject {
    public string data;
}
```
2. Add it to `Assets->Create` menu:
```c#
// This is equivalent to `class_name` in Godot 3.1+
[CreateAssetMenu(menuName="MyCustomStuff/CustomData")]
public class CustomData : ScriptableObject {
    public string data;
}
```
3. Create a component that will use your `ScriptableObject`:
```c#
public class MyComponent : MonoBehaviour {
    public CustomData data;
}
```
4. Create your asset using the `Assets->Create` menu;
5. Drag and drop it to the `data` field in the inspector.

getting_started/step_by_step/resources.rst
willnationsdev: You can have sub resources within scripts, but if you try to define an entirely new type of Resource *as a subclass within another script*, then Godot isn't going to identify the path for the script on that resource as "path to the file that has the Resource subclass AND somehow mark that you need to assign the constant value of the subclass Script resource as the script value for the created \*.tres. You know, something like "res://file.gd#MyResource" or what-have-you.

willnationsdev: Like, if I make a Resource object and set it's script to be not another script file, but a Resource class that's a subclass of a script file (where the script file could derive ANYTHING), it won't start serializing those script properties properly or anything.

getting_started/step_by_step/resources.rst
willnationsdev: Oh, interesting. That's news to me. Great that they added that feature! Appreciate the heads-up. I'll go ahead and take out that tidbit then.

getting_started/step_by_step/resources.rst
LikeLakers2: Thing is, the way this issue is being described in this warning makes it come off as a bug. It sounds like something Godot *could* do, and that it probably should do, but for one reason or another just... doesn't do.

Maybe try to explain why this is the case, at least?

willnationsdev: Sure, sounds good.

getting_started/step_by_step/resources.rst
williamd1k0: Unity docs and changelogs are pretty confusing so I don't know in which version this started to work this way :sweat_smile:

willnationsdev:
@williamd1k0 @LikeLakers2 Updated the document to reflect all of the changes.

getting_started/step_by_step/resources.rst
LikeLakers2: Since you added a C# version, mind resolving this conversation? I can't seem to do so myself.

willnationsdev:
Okay, finally figured out how to make the script classes link work (just doing an absolute link via URL rather than relative one with :ref:).

getting_started/step_by_step/resources.rst
```diff
@@ -158,12 +158,201 @@ Freeing resources
...
 ---------
 
-Like any object in Godot, not just nodes, resources can be scripted,
-too. However, there isn't generally much of an advantage, as resources
-are just data containers.
+Like any Object in Godot, users can also script Resources. Resource
+scripts grant all the same benefits as the base Resource object. They
+inherit both reference-counting memory management and automatic serialization
+between binary or text data (\*.res, \*.tres) and in-memory object
+properties.
```
mhilbrunner: *Resource scripts grant all the same benefits as the base Resource object.
They inherit both reference-counting memory management and automatic serialization
between binary or text data (\*.res, \*.tres) and in-memory object properties.*

This is IMO very complex. We need to avoid overwhelming users with cognitive overload :)

How about:

*Resource scripts inherit all the same benefits as the base Resource object:
Automatic memory management using reference counting, automatic serialization
between binary or text data (\*.res, \*.tres) and in-memory object properties.*

getting_started/step_by_step/resources.rst
```diff
@@ -158,12 +158,201 @@ Freeing resources
...
+        using System;
+        using Godot;
+
+        public class BotStats : Resource
+        {
+            [Export]
+            public int health;
+
+            [Export]
+            public Resource subResource;
```
exts: http://docs.godotengine.org/en/latest/getting_started/scripting/c_sharp/c_sharp_style_guide.html#naming-conventions

public methods/properties need to be PascalCase. Update this for every property.

getting_started/step_by_step/resources.rst
```diff
@@ -158,12 +158,201 @@ Freeing resources
...
+                "GodotBot": BotStats.new(10), # creates instance with 10 health
+                "DifferentBot": BotStats.new(20) # a different one with 20 health
+            }
+            print(dict)
+      .. code-tab:: csharp
+        using Godot;
+
+        public class BotStatsTable : Resource
+        {
+            private Godot.Dictionary<String, BotStats> stats = new Godot.Dictionary<String, BotStats>();
```
exts: Use underscore prefix for private members http://docs.godotengine.org/en/latest/getting_started/scripting/c_sharp/c_sharp_style_guide.html#naming-conventions

getting_started/step_by_step/resources.rst
```diff
@@ -158,12 +158,201 @@ Freeing resources
...
+        }
+
+        // Bot.cs
+        using System;
+        using Godot;
+
+        public class Bot : KinematicBody
+        {
+            [Export]
+            public BotStats stats;
```
exts: change to PascalCase http://docs.godotengine.org/en/latest/getting_started/scripting/c_sharp/c_sharp_style_guide.html#naming-conventions

getting_started/step_by_step/resources.rst
```diff
@@ -158,12 +158,201 @@ Freeing resources
...
+        public class BotStats : Resource
+        {
+            [Export]
+            public int health;
+
+            [Export]
+            public Resource subResource;
+
+            [Export]
+            public String[] strArray;
```
exts: Change to PascalCase http://docs.godotengine.org/en/latest/getting_started/scripting/c_sharp/c_sharp_style_guide.html#naming-conventions

getting_started/step_by_step/resources.rst
```diff
@@ -158,12 +158,201 @@ Freeing resources
...
+        print(stats.health) # prints '10'
+  ..code-tab:: csharp
+        // BotStats.cs
+        using System;
+        using Godot;
+
+        public class BotStats : Resource
+        {
+            [Export]
+            public int health;
```
exts: Change to PascalCase http://docs.godotengine.org/en/latest/getting_started/scripting/c_sharp/c_sharp_style_guide.html#naming-conventions

willnationsdev:
@mhilbrunner @exts updated to reflect the syntax changes and the rewording of the introductory paragraph. I have also added an additional warning and code segment for GDScript (not even sure if a C# example would apply for the warning since it has to do with Dictionary vs. Resource memory management and I don't know if C# GC's Dictionaries). I also fixed a few rendering errors that were preventing the examples from showing up.

willnationsdev:
Oh, also, I realized that I've been mixing and matching my C# code examples' syntax for allocating objects, but I don't see anything in the docs to indicate what it looks like in the style guide or API differences. Which of these would be used?

    MyResource res = new MyResource(10);
    MyResource res = MyResource.new(10);

getting_started/step_by_step/resources.rst
```diff
@@ -158,12 +158,259 @@ Freeing resources
...
+
+        namespace ExampleProject {
+            public class Bot : KinematicBody
+            {
+                [Export]
+                public BotStats Stats;
+
+                public override void _Init()
+                {
+                    if (!Stats)
```
KellyThomas: To perform a null check we should use:

    if (stats == null)

getting_started/step_by_step/resources.rst
```diff
@@ -158,12 +158,259 @@ Freeing resources
...
+                    {
+                        Stats = (ResourceLoader::load("BotStats.cs") as BotStats).new(10);
```
KellyThomas: If we are using the constructor as suggested above we can use:

    stats = new BotStats(10);

getting_started/step_by_step/resources.rst
```diff
@@ -158,12 +158,259 @@ Freeing resources
...
+        using System;
+        using Godot;
+
+        namespace ExampleProject {
+            public class Bot : KinematicBody
+            {
+                [Export]
+                public BotStats Stats;
+
+                public override void _Init()
```
KellyThomas: While testing this code I found that `_Init()` isn't being called (https://github.com/godotengine/godot/issues/22633).

getting_started/step_by_step/resources.rst
```diff
@@ -158,12 +158,259 @@ Freeing resources
...
+
+    const BotStats = preload("res://bot_stats.gd")
+
+    func _init():
+        if not stats:
+            stats = BotStats.new(10)
+    
+    func _ready():
+        print(stats.health) # prints '10'
+  .. code-tab:: csharp
```
KellyThomas: Is this code-tab's code indented more than the others?

getting_started/step_by_step/resources.rst
```diff
@@ -158,12 +158,259 @@ Freeing resources
...
+
+        // Bot.cs
+        using System;
+        using Godot;
+
+        namespace ExampleProject {
+            public class Bot : KinematicBody
+            {
+                [Export]
+                public BotStats Stats;
```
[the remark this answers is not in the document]
willnationsdev: If there isn't a way to export the type (scripted types may not be supported this way?), perhaps something where we export `Resource`, but then do a runtime check that the type extends `BotStats`.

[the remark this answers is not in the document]
KellyThomas: BotStats is not displayed in the Create New Resource dialog either.
Its unclear to me how to create one using the editor GUI.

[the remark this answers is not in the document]
willnationsdev: Yeah, this seems like a deeper issue than I realized. https://github.com/godotengine/godot/issues/22641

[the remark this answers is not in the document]
willnationsdev: As for right now, you basically HAVE to export it as an engine type (Resource or some derivative that is still in the engine) and then check if the runtime instance is a script at runtime.

getting_started/step_by_step/resources.rst
willnationsdev: Should be fixed now.

willnationsdev:
Updated to reflect all changes and relevant Issues.

getting_started/step_by_step/resources.rst
```diff
@@ -158,12 +158,260 @@ Freeing resources
...
+            # this will NOT serialize the `value` property.
+            ResourceSaver.save("res://my_res.tres", my_res) 
+      .. code-tab:: csharp
+        using System;
+        using Godot;
+
+        public class MyNode : Node
+        {
+
+            public class MyResource : Resource {
```
KellyThomas: Its all looking good now, except this brace would be best on the next line.

getting_started/step_by_step/resources.rst
```diff
@@ -158,12 +158,260 @@ Freeing resources
...
+        namespace ExampleProject {
+            public class Bot : KinematicBody
+            {
+                [Export]
+                public Resource Stats;
+
+                public override void _Ready()
+                {
+                    if (Stats != null && Stats is BotStats) {
+                        Godot.print((Stats as BotStats).Health);
```
KellyThomas: Also this should be `GD.Print()`.

NathanLovato:
Thanks much for your work!

NathanLovato:
Sorry, just added the two small changes!

PR #1769, opened 2018-09-29
Fixed wrong result in class_float.rst

FlamyAT:
In the float-string parser documentation 1e-3 should return 0.001 and not -0.001.


LikeLakers2:
You should be editing https://github.com/godotengine/godot/blob/master/doc/classes/float.xml over on the Godot repo, not the float doc here.

FlamyAT:
Go it! I already did edit it, thanks for telling me

PR #1784, opened 2018-10-02
Fix mistakes in "Running code in the editor" page

YeldhamDev:
C'mon people! It doesn't take that much to type `make html` in a terminal to see if your page is OK!

Also, don't ask why the diff is treating it as a complete page revamp, beats me.


LikeLakers2:
> It doesn't take that much to type `make html` in a terminal to see if your page is OK!

Don't even need to do that, you could use an online rst editor like <http://rst.ninjs.org/>, lol

wingedadventurer:
My apologies for the non-loading gif. I did `make html` and it worked fine, but I've renamed the image afterwards and failed to properly rename the image path in the article.

@YeldhamDev Thanks for fixing it. 🙂

NathanLovato:
Thanks!

PR #1785, opened 2018-10-03
Fix GDScript#signals ref in Unity to Godot page

williamd1k0:
I was writing some things on the Unity to Godot page and accidentally clicked on this link that coincidentally was broken. :sweat_smile:

Also applicable to 3.0


NathanLovato:
Thanks!

PR #1793, opened 2018-10-04
Fix Scons not finding library using Linux

MrMinimal:
Because libraries built with Linux contain ".debug" in their filename
Scons would not find the library to compile the Cpp GDNative example ([source](https://old.reddit.com/r/godot/comments/9l5h7k/hitting_erros_while_doing_the_cpp_native_tutorial/)).

Fixes this error and updates the tutorial accordingly.
".debug" in the library name is omitted when compiling in release mode.

### DIFF
--- a/tutorials/plugins/gdnative/files/cpp_example/SConstruct
+++ b/tutorials/plugins/gdnative/files/cpp_example/SConstruct
@@ -7,6 +7,11 @@ bits = ARGUMENTS.get("bits", 64)
 
 final_lib_path = 'demo/bin/'
 
+# Linux has "debug" in library names, if they are compiled in debug mode
+# An example name for a debug library is: libgodot-cpp.linux.debug.64.a
+# In release mode, the library is named libgodot-cpp.linux.64.a
+linux_debug_suffix = ''
+
 # This makes sure to keep the session environment variables on windows, 
 # that way you can run scons in a vs 2017 prompt and it will find all the required tools
 env = Environment()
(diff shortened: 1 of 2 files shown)


mhilbrunner:
cc @willnationsdev

willnationsdev:
@MrMinimal 

1. Because "debug" is the default value of the "target" variable, I would remove the explicit parameter definition that you added to the README since the goal is to keep it as simple/streamlined as possible.
2. The changes you've suggested might(?) not be complete since, when I look at the [godot-cpp SConstruct file](https://github.com/GodotNativeTools/godot-cpp/blob/master/SConstruct#L121), it looks like all "target" values are now added to the resulting library name (not exclusive to "linux" or "debug" builds, but for all platforms and targets).

akien-mga:
It's not specific to Linux, any binary built by Godot's build system has `debug`, `opt` or `opt.debug` in its name depending on the `target` used.

akien-mga:
It's not specific to Linux, any binary built by Godot's build system has `debug`, `opt` or `opt.debug` in its name depending on the `target` used.

willnationsdev:
@akien-mga Based on the SConstruct file's contents, it looks like "debug" and "release" are the only used values so "opt" and "opt.debug" aren't' referenced. Not sure if that needs to be changed or not. Either way, the documentation README is still out of sync with the code, so people will be wondering why stuff isn't working.

@MrMinimal Feel free to tackle the rest of it if you feel up to it (it's your PR after all).

MrMinimal:
@willnationsdev I removed the `target` parameter from the command for consiceness. What other changes do I have to make in your opinion? Are "opt" and "opt.debug" of any relevance?
And do you mean by README the .rst files or am I still missing something?
Thanks for your feedback!

akien-mga:
As mentioned above, the changes made to the build script don't make sense, `debug` is not Linux-specific. So I don't see much point merging this as it would make things more confusing.

CC @karroffel

akien-mga:
Superseded by #1942.

PR #1796, opened 2018-10-05
Removed an unnecessary and controversial line

Warlaan:
I am not even going to start a discussion about GDscript vs. C#, but this line is unnecessary at this place and will only lead to false impressions. If you are in a company that intends to use C# you won't need this information. If you don't know about C# yet there's a good chance that you will be better off using GDscript.

### DIFF
--- a/getting_started/step_by_step/scripting.rst
+++ b/getting_started/step_by_step/scripting.rst
@@ -62,10 +62,6 @@ a generous donation from Microsoft.
 It has an excellent tradeoff between performance and ease of use,
 although one must be aware of its garbage collector.
 
-C# is usually the best choice for companies. The large amount of
-programmers familiar with it means less time can be spent learning
-Godot and more time can be spent programming with it.
-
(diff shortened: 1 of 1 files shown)


LikeLakers2:
It's not exactly controversial, since I imagine many professional developers going into game development will know C# already, and thus will probably prefer to use it compared to learning GDScript. But yeah, I think I can understand the rationale behind removing it.

Warlaan:
That's why it's controversial. I know that many developers will think that they knew C# already and will think that it would be easier for them to use it instead of C#. As a teacher for game programmers I am convinced that in the majority of the cases that will be a bad decision.

If you are interested I can explain why I am convinced of that, but for this thread it's sufficient if we agree that there is room for discussion and that this "recommendation" is unnecessary. If you are experienced it won't affect you and if you aren't you should invest a couple of days to check out GDscript before you invest a couple of months to learn C#.

NathanLovato:
Good catch, thanks!

PR #1803, opened 2018-10-06
Fixed usage of -path flag when debugging

kellams:
When attempting to debug godot, I ran into an error when trying to use the -path flag to open up my project. It seems that the way you supply your project using the -path flag has changed and you need to also add in 'project.godot' to the end or else you will get an error.


merumelu:
It still works without adding project.godot. The fix is to change ``-path`` to ``--path`` because in Godot 3 long parameters were changed to use two dashes.

Issue #1665, opened 2018-08-06
Document softbody and IK


ZX-WT:
Just leaving a record as reminder.

https://godotengine.org/article/soft-body

https://godotengine.org/article/skeleton-inverse-kinematic

ZX-WT:
I will work on this soon, I guess, but there is already a tutorial named Inverse kinematics under tutorials/3d which explain the math and implement IK with GDscript. It is in different nature with the new built-in feature, I think we can keep it but how should we rename these tutorials?

PR #1805, opened 2018-10-06
added softbody tutorial

ZX-WT:
Third PR :) Fix half of #1665 

Again, please review it.

@AndreaCatania Do you have the demo that was in https://godotengine.org/article/soft-body? The file is expired, I would like to include it in this tutorial.


AndreaCatania:
It's Good :)

tutorials/3d/softbody.rst
```diff
@@ -0,0 +1,64 @@
+.. _doc_softbody:
+
+Soft body
+=========
+
+Soft body simulates cloth movement and other deformable bodies.
```
mhilbrunner: Soft bodies (or *soft-body dynamics*) simulate movement, changing shape and other physical properties of deformable objects.
This can for example be used to simulate cloth or to create more realistic characters.

tutorials/3d/softbody.rst
```diff
@@ -0,0 +1,64 @@
...
+
+Basic set-up
+~~~~~~~~~~~
+
+A :ref:`SoftBody <class_SoftBody>` node is used for soft body simulations.
+
+We will create a bouncy cube to demonstrate the set up of a soft body.
```
mhilbrunner: setup

tutorials/3d/softbody.rst
```diff
@@ -0,0 +1,64 @@
...
+
+.. image:: img/softbody_cloak_pinned_attach.png
+
+Last step is to avoid clipping by adding the Kinematic Body `Player` to ``Parent Collision Ignore`` of the ``SoftBody``.
+
+.. image:: img/softbody_cloak_ignore.png
+
+Play the scene and the cloak should simulate correctly.
+
+.. image:: img/softbody_cloak_finish.png
```
mhilbrunner: IMO add a ending sentence. :)

mhilbrunner:
This should also be added to the corresponding index (for an example, see the changes in https://github.com/godotengine/godot-docs/pull/1807)

NathanLovato:
Like with the ragdoll tutorial, how about putting this in Physics? As it's a tutorial about physics simulation

mhilbrunner:
Agree, this should probably be moved to Physics.

tutorials/3d/softbody.rst
```diff
@@ -0,0 +1,64 @@
+.. _doc_softbody:
+
+Soft body
+=========
+
+Soft body simulates cloth movement and other deformable bodies.
+
+Basic set-up
+~~~~~~~~~~~
```
LikeLakers2: You're missing a `~` here.

tutorials/3d/softbody.rst
```diff
@@ -0,0 +1,64 @@
...
+
+A :ref:`SoftBody <class_SoftBody>` node is used for soft body simulations.
+
+We will create a bouncy cube to demonstrate the set up of a soft body.
+
+Open a new scene with a ``Spatial`` node as root. Then, create a ``Softbody`` node. Add a ``CubeMesh`` in the ``mesh`` property of the node in the inspector and increase the subdivision of the mesh for simulation.
```
LikeLakers2: "*Create* a new scene with a Spatial node as root" would probably sound a little better here.

tutorials/3d/softbody.rst
```diff
@@ -0,0 +1,64 @@
...
+
+.. image:: img/softbody_cube.png
+
+Set the parameters to obtain the type of soft body you aim for. Try to keep the ``Simulation Precision`` above 5, otherwise, the soft body may collapse.
+
+.. image:: img/softbody_cube_menu.png
+
+.. note:: Handle some parameters with care, as some value can lead to strange results. For example, if the shape is not completely closed and you set pressure to more than 0.
```
LikeLakers2: You give an example, but it's odd to me that you don't explain what happens if you try that example.

AndreaCatania:
> Like with the ragdoll tutorial, how about putting this in Physics? As it's a tutorial about physics simulation

The node soft body extend mesh instance, for this reason is better don't move it to physics even if it's a physics object.

Initially, when I created it, it was a physics object but reduz told me that was way easier for others if the SoftBody extends a MeshInstance.

So if we keep following this reasoning we should not put it under physics section.

akien-mga:
@AndreaCatania I think this technical detail is maybe not relevant for users learning Godot though. As far as I understand it SoftBody is a physics-related concept, so it should be in the Physics tutorial section IMO.

Or are there non-physics use cases for SoftBody?

AndreaCatania:
Well it's both. It has not a deep interaction with physics thing (e.g. can't be an obstacle) but it's driven by physics engine, but it's first usage is to deform the mesh.

It could be ambiguous in both situation so I think that we should follow the current choise did for the node.

A possible solution to this dilemma is put a link to that page on both section.

mhilbrunner:
I'd expect it under the Physics section, and IMO it belongs there. It's a physics feature. Sure it is used for a 3D game, but then you could also place the 3d spatial audio stuff under 3D, not Audio.

After googling a bit, all defintions I found only talk about Physics too, and both [Unity](https://docs.unity3d.com/560/Documentation/Manual/class-Cloth.html) and [Unreal Engine](https://docs.unrealengine.com/en-us/Engine/Physics/Cloth/Overview) place their respective tutorial under *Physics* too, not *3D*.

AndreaCatania:
I agree with you

akien-mga:
@ZX-WT Could you rebase and move it to the Physics section? Then it should be ready to merge IMO.

akien-mga:
Thanks, and sorry for the delay!

PR #1807, opened 2018-10-06
Fix "Ragdoll system" page

YeldhamDev:
This one was not even in the index...


NathanLovato:
Physics would be a better fit for ragdoll: even if it's only available in 3d, it's all about physics bones and physics simulation.

YeldhamDev:
@NathanLovato I put in the 3D index because the tutorial file was there as well.

PR #1810, opened 2018-10-07
Added a clarification about the Godot-Blender-Exporter

Warlaan:
After I had spent several hours trying to get the physics export to work using the method described in this chapter combined with the "better collada exporter" I finally understood that there is a plugin that is called Godot-Blender-Exporter.
Since it used to be the default to use the "better collada exporter" I didn't understand that the title of the page is a name, not a description. After all the better collada exporter is kind of a godot to blender exporter.

### DIFF
--- a/getting_started/workflow/assets/escn_exporter/index.rst
+++ b/getting_started/workflow/assets/escn_exporter/index.rst
@@ -1,6 +1,9 @@
 Godot-Blender-Exporter
 ======================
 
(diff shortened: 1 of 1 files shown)


Warlaan:
By the way is there a page that explains how tags like .. note:: work? At least the preview is broken, but it's also broken in places where I didn't change anything, so I have no idea if I applied the tag correctly or not.

mhilbrunner:
There is this:
https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html

PR #1808, opened 2018-10-06
Move the ragdoll tutorial from 3d to Physics

NathanLovato:
Opening it as a PR as this is a suggestion, but to me ragdoll better fits Physics rather than 3d - it's all about physics simulation.

Can you please merge #1802 first to avoid conflicts?

PR #1812, opened 2018-10-07
Move images for "Ragdoll system" page to the correct directory

YeldhamDev:
Only the page itself was moved in #1808.


NathanLovato:
Thanks!

PR #1814, opened 2018-10-08
Updated DVector to PoolVector in core_types

### LINKED ISSUE #1678: Outdated reference to DVector in Core Types docs
The docs in:
http://docs.godotengine.org/en/latest/development/cpp/core_types.html#allocating-memory

Is outdated and still mentions `DVector` instead of `PoolVector`. It could also be a chance to better explain the memory model in relation to the recent Copy-On-Write changes

Battlefrog:
fixes #1678

Let me know if any changes are needed.


NathanLovato:
Thank you :)

PR #1820, opened 2018-10-09
Fix bad indentation in some GDScript examples

antoniotorresm:

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_basics.rst
+++ b/getting_started/scripting/gdscript/gdscript_basics.rst
@@ -90,41 +90,41 @@ here's a simple example of how GDScript looks.
     # Function
 
     func some_function(param1, param2):
-        var local_var = 5
+    	var local_var = 5
 
-        if param1 < local_var:
-            print(param1)
-        elif param2 > 5:
-            print(param2)
-        else:
-            print("Fail!")
+    	if param1 < local_var:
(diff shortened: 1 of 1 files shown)


NathanLovato:
You don't want mixed tabs and spaces. Please use the script editor's option to convert everything to spaces

Calinou:
> You don't want mixed tabs and spaces. Please use the script editor's option to convert everything to spaces

The [GDScript style guide](http://docs.godotengine.org/en/latest/getting_started/scripting/gdscript/gdscript_styleguide.html) recommends using tabs, so these should be used in all demo projects.

antoniotorresm:
Changed spaces for tabs on all examples. For some reason tab width is 8 spaces in some parts (see `match` examples).

Calinou:
I just realized this is the documentation repository, not the demo projects repository… I apologize for the confusion.

If I'm not mistaken, spaces should be used since they're used for indenting stuff other than code in rST (otherwise, we would end up with tab and space indentation in the same file).

akien-mga:
Indeed, this is a Python-based documentation generator and uses space for indentation by convention. Tabs are also valid, but we chose spaces to be consistent.

Sorry about the confusion leading to extra work.

akien-mga:
Fixed on the whole repo with a0e32ac0.

PR #1825, opened 2018-10-10
Add move_and_slide_with_snap to K2D doc

cbscribe:
Added description for new method.

I also wanted to add `stop_on_slope` but as there are still a couple of open issues around this one, it's probably best to hold off. See https://github.com/godotengine/godot/issues/21595

### DIFF
--- a/tutorials/physics/using_kinematic_body_2d.rst
+++ b/tutorials/physics/using_kinematic_body_2d.rst
@@ -91,6 +91,15 @@ other parameters allowing you to customize the slide behavior:
 
     This is the maximum angle before a surface is no longer considered a "floor".
 
+``move_and_slide_with_snap``
+~~~~~~~~~~~~~~~~~~
+
+This method adds some additional functionality to ``move_and_slide()`` by adding
+the ``snap`` parameter. As long as this vector is in contact with the ground, the
+body will remain attached to the surface. Note that this means you must disable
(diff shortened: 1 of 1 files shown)


tutorials/physics/using_kinematic_body_2d.rst
```diff
@@ -91,6 +91,15 @@ other parameters allowing you to customize the slide behavior:
...
     This is the maximum angle before a surface is no longer considered a "floor".
 
+``move_and_slide_with_snap``
+~~~~~~~~~~~~~~~~~~
+
+This method adds some additional functionality to ``move_and_slide()`` by adding
+the ``snap`` parameter. As long as this vector is in contact with the ground, the
+body will remain attached to the surface. Note that this means you must disable
+snapping when jumping, for example. You can do this either by setting ``snap``
+to ``Vector2(0, 0)`` or by using ``move_and_slide()`` instead.
```
LikeLakers2: "As long as this vector is in contact with the ground"? That feels like an odd choice of words that might end up confusing the reader.

NathanLovato: It's fine by me: a vector has an origin, a length, and a direction, it can intersect with shapes geometrically speaking

NathanLovato:
Thanks!

PR #1832, opened 2018-10-11
clarified section about adding icons to plugins

Ashafix:
The tutorial was expanded by a few words to make it clearer.
If one follows the tutorial literally the following error message is shown "Base type is not EditorPlugin" which just says "the icon cannot be found".


NathanLovato:
Thanks

PR #1837, opened 2018-10-13
British to American spelling

jcs224:
As per issue #1066

PR #1838, opened 2018-10-13
More British to American spelling changes

jcs224:


LikeLakers2:
You should add these additional changes to #1837 rather than make a PR.

PR #1848, opened 2018-10-15
Update gdscript_advanced.rst

Vikram1323:
# Same example, lua-style support.
    # This syntax is a lot more readable and usable
+ # Like any GDScript identifier, keys written in this form cannot start with a digit.

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_advanced.rst
+++ b/getting_started/scripting/gdscript/gdscript_advanced.rst
@@ -300,6 +300,7 @@ states and quick structs:
 
     # Same example, lua-style support.
     # This syntax is a lot more readable and usable
+    # Like any GDScript identifier, keys written in this form cannot start with a digit.
 
     var d = {
         name = "John",


getting_started/scripting/gdscript/gdscript_advanced.rst
```diff
@@ -300,6 +300,7 @@ states and quick structs:
 
     # Same example, lua-style support.
     # This syntax is a lot more readable and usable
+    # Like any GDScript identifier, keys written in this form cannot start with a digit.
```
LikeLakers2: This isn't actually correct. I've been able to use digits as key names before, see [here](https://github.com/MichiRecRoom/discord.gd/blob/6f6b71292229e1c400954576178532c3bcf3567f/addons/discord.gd/gateway/connection.gd#L19).

NathanLovato:
You can use numbers as keys in dictionaries as shown by LikeLakers, closing

Vikram1323:
You are right but that is not in lua-syntax for the dictionary. Lua-syntax
uses  "=" and python-syntax uses ":"
Check this link:-
http://docs.godotengine.org/en/3.0/getting_started/scripting/gdscript/gdscript_basics.html#dictionary
and this one
https://ibb.co/fFWmi0

LikeLakers2:
Ah... my bad. I forgot there was a reason I used the python-like syntax.

Yeah, actually this PR's okay. Though it might be a good idea to say that the python-like syntax allows for identifiers starting with numbers.

Vikram1323:
Thank you.

NathanLovato:
My bad, missed the fact this was about the lua syntax (I never use it tbh)

Vikram1323:
Thank you.

PR #1850, opened 2018-10-15
Updated openSUSE package requirements

starry-abyss:
A couple of packages I needed to compile master


starry-abyss:
Also had to remove libressl-devel, not sure if it could work drop-in for openssl-devel

mhilbrunner:
Thank you! :)

PR #1853, opened 2018-10-15

Vikram1323:
Not only do they offer a range of features, the design of each engine
+Not only do they offer a range of features, but the design of each engine


NathanLovato:
Thanks! 
You can group your changes, you don't need to open individual branches for small changes - it'll save you some time. Also no need to put the change in the PR's body: Reviewers will always read the commit diffs.

Vikram1323:
Ok.
I'll learn how to do that.😀

PR #1855, opened 2018-10-16
fix markdown

akim8:
image

added spaces between code ticks and words preceding it to fix this


LikeLakers2:
Thanks, but class reference files shouldn't be edited through this repository. Edit [`docs/classes/KinematicBody2D.xml`](https://github.com/godotengine/godot/blob/master/doc/classes/KinematicBody2D.xml#L119) over on the [godotengine/godot](https://github.com/godotengine/godot) repo instead!

NathanLovato:
As LikeLakers said, you need to use the Godot repo. Thanks for your PR, but closing

akim8:
noted, will do

PR #1856, opened 2018-10-16
Fixed typo in "Compiling for Universal Windows Platform"

mega-bit:


NathanLovato:
Thanks for fixing typos! Moving forward, could you please group them in the same PR? You don't need one commit per word either. It would save time for both you and I.

PR #1870, opened 2018-10-17
grammar improvements within import_process.rst

andy5995:

### DIFF
--- a/getting_started/workflow/assets/import_process.rst
+++ b/getting_started/workflow/assets/import_process.rst
@@ -57,7 +57,7 @@ Additionally, extra assets will be preset in the hidden res://.import folder:
 Any assets will be reimported automatically when they, or the
 res://.import folder, is erased. Depending on the size of your project,
 you can decide whether or not to commit that folder to a version
-control system (such as git). After checking out your project on
+control system. After checking out your project on
 another computer, Godot will not need to reimport if the folder exists;
 however, committing that folder will use significantly more space in
 the repository and will increase the transfer time. Essentially, if you
--- a/getting_started/workflow/assets/import_process.rst
+++ b/getting_started/workflow/assets/import_process.rst
@@ -58,12 +58,12 @@ Any assets will be reimported automatically when they, or the
 res://.import folder, is erased. Depending on the size of your project,
 you can decide whether or not to commit that folder to a version
 control system (such as git). After checking out your project on
-another computer, the time it takes for Godot to import will be reduced
-if the folder exists; however, committing that folder will use more
-space in the repository and will increase the transfer time.
-Essentially, if you don't plan on editing your project from another
-location, you will probably find there is no advantage to adding the
-res://.import folder to your remote repository.
+another computer, Godot will not need to reimport if the folder exists;
+however, committing that folder will use significantly more space in
+the repository and will increase the transfer time. Essentially, if you
+don't plan on editing your project from another location, you will
+probably find there is no advantage to adding the res://.import folder
+to your remote repository.
 
 Changing import resource type
 -----------------------------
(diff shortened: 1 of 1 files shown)


getting_started/workflow/assets/import_process.rst
```diff
@@ -54,10 +54,10 @@ Additionally, extra assets will be preset in the hidden res://.import folder:
 
 .. image:: img/asset_workflow5.png
 
-If any of the files present in this folder is erased (or the whole folder), the
+If any of the files present in this folder are erased (or the whole folder), the
 asset or assets will be reimported automatically. As such, committing this folder
-to the version control system is optional. It can save time on
-reimporting time when checking out in another computer, but it takes considerably
+to the version control system is optional. It can save time when
+reimporting or checking out on another computer, but it takes considerably
```
akien-mga: The previous sentence had wrong grammar but the new one no longer has the correct meaning. What this sentence aims to say is:
- You go to a new computer
- You clone the git repository, which does not have the `.import/` folder
- When you open the project the first time, all assets will have to be reimported, which takes time

Or alternatively, if the `.import/` is committed to Git, you no longer need to reimport as the imported files have been cloned to the new computer. But they will take a lot of space in the Git repository.

getting_started/workflow/assets/import_process.rst
```diff
@@ -54,11 +54,16 @@ Additionally, extra assets will be preset in the hidden res://.import folder:
 
 .. image:: img/asset_workflow5.png
 
-If any of the files present in this folder is erased (or the whole folder), the
-asset or assets will be reimported automatically. As such, committing this folder
-to the version control system is optional. It can save time on
-reimporting time when checking out in another computer, but it takes considerably
-more space and transfer time. Pick your poison!
+Any assets will be reimported automatically when they, or the
```
akien-mga: That's confusing, it sounds like the asset will be reimported when it's deleted, but it's only when the corresponding file in the `res://.import/` folder is deleted that the asset will be reimported.

getting_started/workflow/assets/import_process.rst
```diff
@@ -54,11 +54,16 @@ Additionally, extra assets will be preset in the hidden res://.import folder:
...
+res://.import folder, is erased. Depending on the size of your project,
+you can decide whether or not to commit that folder to a version
+control system (such as git). After checking out your project on
```
akien-mga: No need to mention Git specifically here IMO.

akien-mga:
Sorry for the late review. There's one confusing sentence, but otherwise the changes look fine.

getting_started/workflow/assets/import_process.rst
andy5995: I don't understand. Could you or one of the other maintainers edit this so it sounds accurate please?

clayjohn: How about: "Assets will be reimported automatically when they are deleted from the res://.import folder (or if the folder itself is deleted)."

getting_started/workflow/assets/import_process.rst
```diff
@@ -57,7 +57,7 @@ Additionally, extra assets will be preset in the hidden res://.import folder:
 Any assets will be reimported automatically when they, or the
 res://.import folder, is erased. Depending on the size of your project,
 you can decide whether or not to commit that folder to a version
-control system (such as git). After checking out your project on
```
LikeLakers2: Not that I object, but I'm curious why this change was made.

akien-mga:
As per above comments, this does not make things clearer as is, so I won't merge it for now. I still agree that the language of this section could be improved for clarity and grammatical correctness, though.

PR #1871, opened 2018-10-17
fix typo within ui_main_menu.rst (had->have)

andy5995:
This PR is against the 3.0 branch because when I used the "edit this page" link, that's what it defaulted to.


akien-mga:
Thanks!

PR #1872, opened 2018-10-17
Fix spawn location of mob

bliksemman:
This fixes the spawn location of the mob entities.


akien-mga:
Sorry for the delay.

@cbscribe Could you review this change?

akien-mga:
Note that the C# code should be adapted too (provided this is the correct fix).

cbscribe:
Unless something is wrong, `$MobPath.position` should be `(0, 0)`, so this change is unnecessary. If you have a problem with the mob spawning, the fix is to not give your path node an offset.

PR #1881, opened 2018-10-22
Document the range of the return value

Hendrikto:
The documentation contained the type of the return value, but not the range.

I copied the explanation from [here](https://github.com/godotengine/godot/blob/master/doc/classes/InputEventJoypadMotion.xml#L21).


cbscribe:
Changes to the class reference must be done in the godot repo, not godot-docs. Please see the following page for details:
http://docs.godotengine.org/en/latest/community/contributing/updating_the_class_reference.html

PR #1884, opened 2018-10-22
Added line about android studio

melvyniandrag:
I was confused by what SDK meant. The link takes  you to a page offering Android Studio or SDK platform tools and it wasn't immediately clear to me what needed to be downloaded.

### DIFF
--- a/getting_started/workflow/export/exporting_for_android.rst
+++ b/getting_started/workflow/export/exporting_for_android.rst
@@ -9,8 +9,7 @@ following steps detail what is needed to setup the SDK and the engine.
 Download the Android SDK
 ------------------------
 
-Download and install the Android SDK from
-https://developer.android.com/studio/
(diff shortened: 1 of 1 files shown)


Calinou:
The full Android Studio IDE is unnecessary for compiling Godot; the command-line tools are enough. The Gradle build system will automatically download missing SDK components, so you only need to download the SDK and NDK after accepting the license by creating a file. See [this script](https://github.com/Calinou/godot-builds-ci/blob/master/scripts/gitlab-ci/build/android.sh) for an example of automating Android builds.

melvyniandrag:
So maybe the doc should recommend just installing the command line tools then. As it's currently worded, it wasn't immediately clear to me what to install. The doc says to download the sdk, but the download page says "android studio" and "sdk command line tools". Either way, I just figured and extra sentence with more specificity was better.

akien-mga:
@Calinou Would you have a proposal for clearer instructions regarding what needs to be downloaded and installed exactly? Getting the Android buildsystem setup seems to be a major pain point for many users.

akien-mga:
Superseded by #1994.

PR #1886, opened 2018-10-23
Audio buses tutorial: link to proper pages

merumelu:
Link to audio streams tutorial instead of back to the bus tutorial itself.

Also make the caption in one of the images a, well, caption instead of having text in the image.

Fixes #1882

PR #1887, opened 2018-10-24
Audio bus tutorial: update link missed in #1886

merumelu:


mhilbrunner:
Thanks for the fix! :)

PR #1891, opened 2018-10-26
Simplify GDScript mul/div/remainder operator text

aaronfranke:
IMO this is a much better description. It's to the point, and if anyone has further questions, they can look up how the C++ operators work.


akien-mga:
Thanks!

PR #1895, opened 2018-10-29
Add .pck creation tutorial

willnationsdev:
I have translated [this blog post by MrMinimal](https://mrminimal.gitlab.io/2018/10/28/mod-support-in-godot.html) into an introductory guide for the docs. It is not terribly extensive, but it should get the job done.

1. I'm not entirely sure what the right way to credit the blog author would be (is here fine, or should I mention something at the bottom of the file?).
2. Also, I'm not 100% sure if my second bullet point at line 80 has completely accurate content. If someone could double-check that for me, that'd be great (that's the only way I can conceive of implementing mods given what I know).


MrMinimal:
I'm fine this, you don't even have to mention my blog, the content is CC0. But you are free to do so regardless. Thanks for your effort!

getting_started/workflow/export/exporting_pcks.rst
```diff
@@ -0,0 +1,129 @@
...
+-----------------------
+
+In order to pack all resources of a project into a \*.pck file open the project
+and go to Project/Export and click on “Export PCK/Zip”. Also make sure to have
+an export template selected while doing so.
+
+.. image:: img/export_pck.png
+
+Another method would be to `export from the command line
+<https://docs.godotengine.org/en/latest/getting_started/editor/command_line_tutorial.html#exporting>`_.
```
merumelu: This should use `:ref:` instead of a hardcoded link I think. However, to link directly to the "Exporting" section a new reST label in `getting_started/editor/command_line_tutorial.rst` would be needed because apparently you can't use anchors with `:ref:`. :(

Maybe we could add the [Sphinx autosection extension](http://www.sphinx-doc.org/en/master/usage/extensions/autosectionlabel.html) to automate this.

willnationsdev: I just know that other sections written by more experienced contributors had done it that way, but if we can get a more official, Sphinx-like solution in place, that'd be preferred, sure.

getting_started/workflow/export/exporting_pcks.rst
```diff
@@ -0,0 +1,129 @@
...
+ .. code-tab:: gdscript GDScript
+
+    func _your_function():
+        ProjectSettings.load_resource_pack("res://mod.pck")
+        # Now one can use the assets as if they had them in the project from the start
+        var imported_scene = load("res://mod_scene.tscn")
+
+ .. code-tab:: csharp
+
+    public void _YourFunction()
```
neikeq: This should be `private void YourFunction()` instead.

getting_started/workflow/export/exporting_pcks.rst
```diff
@@ -0,0 +1,129 @@
...
+    {
+        ProjectSettings.load_resource_pack("res://mod.pck");
+        // Now one can use the assets as if they had them in the project from the start
+        var imported_scene = ResourceLoader.load("res://mod_scene.tscn") as PackedScene;
```
neikeq: C# API uses the naming case convention from C# where possible. This should be `ProjectSettings.LoadResourcePack` and `ResourceLoader.Load`. Also `importedScene`.

willnationsdev:
@neikeq Just fixed with your suggestions (I just forgot to update that C# code - whoops). Thanks.

getting_started/workflow/export/exporting_pcks.rst
akien-mga: Then more experienced contributors were wrong :)

Internal hyperlinks should indeed use `:ref:`, as do the vast majority of them:
```
$ rg ":ref:" | wc -l
16793
$ rg "http[s]?://docs.godotengine.org/en/" | wc -l
3
```
Actually there's a single occurrence of such a direct link (the other two are just in the layout page and cross-reference between master and stable docs):
```
getting_started/scripting/gdscript/static_typing.rst
107:   types <http://docs.godotengine.org/en/3.0/getting_started/scripting/gdscript/gdscript_basics.html#built-in-types>`__
```

akien-mga: You can reference the whole page `:ref:doc_command_line_tutorial`. To reference the "Exporting" section directly, you'd have to add a new label in that page indeed, with:
```
.. _doc_command_line_tutorial_exporting:
```
(before the "Exporting" title)
which you can then reference with `:ref:doc_command_line_tutorial_exporting`.

akien-mga: I'll merge and fix those two occurrences myself :)

akien-mga:
Thanks!

getting_started/workflow/export/exporting_pcks.rst
akien-mga: To be fair there's two other ones using another URL:
```
$ rg "http[s]?://godot.readthedocs.io/en/"
getting_started/step_by_step/resources.rst
225:If the script's language supports `script classes <https://godot.readthedocs.io/en/latest/getting_started/step_by_step/scripting_continued.html#register-scripts-as-classes>`__,

getting_started/scripting/gdscript/static_typing.rst
111:   `class_name <https://godot.readthedocs.io/en/latest/getting_started/step_by_step/scripting_continued.html#register-scripts-as-classes>`__
```

akien-mga: FYI: 8a49da6d

PR #1905, opened 2018-11-03
Array.sort() clarification

iatenine:
Clarify that NULL entries in an array are not counted in Array.sort()


bojidar-bg:
The class reference is automatically generated using [the files in the docs/classes folder](https://github.com/godotengine/godot/tree/master/doc/classes) of the [godotengine/godot](https://github.com/godotengine/godot) repository, as noted in the comment on top of the file.

Please edit those files instead.

PR #1907, opened 2018-11-04
Fixes unclear working in move_and_slide description

ForLoveOfCats:
The `move_and_slide` velocity arg is not multiplied by the frame time but is multiplied by the physics delta


akien-mga:
Sorry for the delay. Changes to the class reference should be done directly on https://github.com/godotengine/godot, those files are auto-generated from the base XML.

akien-mga:
See http://docs.godotengine.org/en/latest/community/contributing/updating_the_class_reference.html

ForLoveOfCats:
No problem! Thanks for telling me!

PR #1908, opened 2018-11-05
Expanded migration guide with vertex shader information

### LINKED ISSUE #1883: Godot migration tutorial: matrices and the vertex shader
Currently the tutorial focuses on the fragment shader, especially regarding the book of shaders and shadertoy. However, there are some important differences in how Godot handles conversion of models into NDC. Specifically it would be valuable to explain when and how Godot multiples the Vertex by various matrices, and how you can avoid those conversions. 

For example, by default Godot presents `VERTEX` in model space, then after `vertex()` runs it converts from model space into view space. Setting `skip_vertex_transform` avoids the model to view space transform but Godot still applies the view to clip space transform. 

I think the migration tutorial may be the place for this, however it may be better to add a tutorial under the shading section call 'Vertex shaders' and then explain it in detail there, any thoughts?

clayjohn:
This PR adds a section to the Shader Migration guide that explains the use of vertex attributes and writing to gl_Position. 

Closes #1883.

### DIFF
--- a/tutorials/shading/migrating_to_godot_shader_language.rst
+++ b/tutorials/shading/migrating_to_godot_shader_language.rst
@@ -30,6 +30,29 @@ Godot allows uniform variables and functions to be shared by defining the fragme
 shaders in one file. In GLSL the vertex and fragment programs cannot share variables except 
 when varyings are used.
 
+Vertex attributes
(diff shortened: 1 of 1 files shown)


akien-mga:
Looks good, thanks!

PR #1914, opened 2018-11-10
Add NuGet as a requirement for Mono compiling

Cevantime:
Compiling Godot Mono on x11 with the latest source needs NuGet to be intalled which is not a dependency of any current requirement.


akien-mga:
Thanks!

PR #1917, opened 2018-11-13
Improve the "Exporting projects" page

Calinou:
This improves writing style and removes less relevant (and possibly out of date) information on exporting for optical media.


akien-mga:
Thanks, that's much better :)

PR #1923, opened 2018-11-18
Prefabs are a Unity concept.

janvanhoudt:
Since not all readers have Unity experience and thus know what prefabs are, it seems better to clarify it here.


akien-mga:
Thanks!

PR #1926, opened 2018-11-20
Grammar fix + clarity improvements

blurrred:
Added proper article before "background" and put the type of node in parenthesis.


akien-mga:
Thanks!

PR #1928, opened 2018-11-20
Moved vertex_displacement tutorial into shading section

clayjohn:
After re-writing this tutorial the focus became completely on using shaders. Accordingly, it is more appropriately stored in the shading section of the docs.


cbscribe:
Seems reasonable to me. It looks like we have some more shader docs incoming, so it's good to see this section growing.

PR #1931, opened 2018-11-21
Fix Capitalization in Viewport Code Comments

skyace65:
Fixes capitalization in two of the viewport code comments.


cbscribe:
Thanks!

PR #1936, opened 2018-11-26
Grammar fix

blurrred:
Added "to."
This could also be re-written as "This is where we connect nodes to listen to the node we selected."


akien-mga:
Thanks!

PR #1939, opened 2018-11-26
Update scenes_and_nodes.rst

corrigentia:


getting_started/step_by_step/scenes_and_nodes.rst
```diff
@@ -36,7 +36,7 @@ functions. However, any given node always has the following attributes:
 -  It has editable properties.
 -  It can receive a callback to process every frame.
 -  It can be extended (to have more functions).
--  It can be added to other nodes as children.
+-  It can be added to other nodes as a child.
```
cbscribe: This is good - it should be singular. But while you're at it, so should the rest of the sentence, since you can't add a node as a child of more than one other node.

> It can be added to another node as a child.

cbscribe: Nice. Can you squash the two commits into one?

akien-mga:
Thanks!

PR #1944, opened 2018-11-28
Renaming the class gdexample to GDExample

BastiaanOlij:
Renaming the class gdexample to GDExample to be more consistent with Godot


akien-mga:
Thanks!

PR #1947, opened 2018-11-28
added command documentation for multithreaded compilation and how to …

ERUD1T3:
…navigate to the godot binary files


ERUD1T3:
I found out that scons allows you to compile with multiple threads and I added the command for it. I also explicitly gave the command to find and launch the binary files

akien-mga:
Thanks for the contribution, but this is already covered in the Introduction to the Buildsystem: http://docs.godotengine.org/en/3.0/development/compiling/introduction_to_the_buildsystem.html

akien-mga:
Should be more explicit with 5a8fd214.

PR #1948, opened 2018-11-29
Update your_first_game.rst

corrigentia:


akien-mga:
Merged manually as 47dba052, thanks. (Minus the "Now**,** you're ready" change which is controversial in #1949).

corrigentia:
@akien-mga Thanks for sifting through that.

PR #1956, opened 2018-12-03
Update ui_main_menu.rst

corrigentia:


getting_started/step_by_step/ui_main_menu.rst
```diff
@@ -276,7 +276,7 @@ create a Game User Interface with bars and item counters.
 Break down the UI mockup
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
-Responsive User Interface is all about making sure our UIs scale well on
+A Responsive User Interface is all about making sure our UIs scale well on
```
akien-mga: I don't think "Responsive" should be capitalized here.

akien-mga:
Thanks!

getting_started/step_by_step/ui_main_menu.rst
corrigentia: @akien-mga I apologize you've had to clean up after me. You're right; that was wrong. I'm trying to change as little as possible. Sometimes, that means I forget something that is obviously necessary. Thanks for catching that! :smile:

PR #1957, opened 2018-12-03
Update ui_game_user_interface.rst

corrigentia:


getting_started/step_by_step/ui_game_user_interface.rst
```diff
@@ -4,8 +4,8 @@ Design the GUI
 ==============
 
 Now that you've nailed the basics, we're going to see how to build a
-Game User Interface (GUI) with reusable UI components: a life bar, an
-energy bar, and bomb and rupee counters. By the end of this tutorial,
+game Graphical User Interface (GUI) with reusable UI components: a life bar, an
```
bojidar-bg: Uhm, no, the abbreviation GUI means "Graphical User Interface": [TheFreeDictionary](https://acronyms.thefreedictionary.com/GUI), [Wiktionary](https://en.wiktionary.org/wiki/GUI#Noun)
"Game User Interface" is abbreviated to "Game UI" usually, or "game GUI" if one wants to emphasize that it is "Graphical"

getting_started/step_by_step/ui_game_user_interface.rst
```diff
@@ -342,7 +342,7 @@ Inherited Scenes.
 
 Let's save both the ``Counter`` and the ``Bar`` branches as separate
 scenes that we'll reduce to create the ``LifeBar``, the ``EnergyBar``,
-the ``BombCounter``, and the ``RupeeCounter``. Select the ``Bar``
+the ``BombCounter``, and the ``EmeraldCounter``. Select the ``Bar``
```
bojidar-bg: Images should be updated as well, I guess.

getting_started/step_by_step/ui_game_user_interface.rst
```diff
@@ -594,7 +594,7 @@ vertically with the ``LifeBar``.
 
 .. note::
 
-    We could have setup the ``EnergyBar`` this way a few moments
+    We could have set up the ``EnergyBar`` this way a few moments
```
bojidar-bg: Potentially: set <del>up</del> the `EnergyBar` <ins>up</ins>

getting_started/step_by_step/ui_game_user_interface.rst
```diff
@@ -614,7 +614,7 @@ small as possible vertically. Now you can see how the interface looks in
 the context of the game.
 
 Congratulations for getting to the end of this long tutorial. You can
-find final project `here <#>`__.
+find the final project `here <#>`__.
```
bojidar-bg: Someone should link this to `getting_started/step_by_step/files/ui_gui_design.zip`.

getting_started/step_by_step/ui_game_user_interface.rst
akien-mga: Well the change is consistent with what you're saying, it fixes the previous "Game User Interface" to "game GUI" as you mentioned.

bojidar-bg: Oh, my bad, sorry.

merumelu:
Is the Rupee -> Emerald change neccessary? "Rupee" is a Legend of Zelda reference, the counter graphic has the exact same design as LoZ Rupees: https://zelda.gamepedia.com/Rupee (Same for the bombs).

Since Zelda is such a popular and long running series a lot of people probably recognize this reference and in my opinion not knowing about it doesn't make the tutorial harder to understand either. It's a nice easter egg and unless there are trademark/copyright concerns I don't think this needs to be changed.

akien-mga:
I don't have a strong opinion, but it's true that both in German (as mentioned above) and in French, Zelda's "Rupee" is translated the same as the English word "ruby", which is a red gem. So the Zelda reference will likely be lost on most readers of such locale, once translated.

akien-mga:
I reverted the `s/Rupee/Emerald` change to unblock this PR (since, as mentioned above, images need to be fixed accordingly if we're doing this change).

As mentioned above, I do agree that "rupee" is probably not a good name and that the Zelda reference will most likely be lost on many users, and lost in translation. I'll open an issue about it.

akien-mga:
Thanks!

getting_started/step_by_step/ui_game_user_interface.rst
akien-mga: Fixed.

PR #1959, opened 2018-12-03
Fix function name for emit in GDScript in docs

konchunas:
Rename emit() to emit_signal() in documentation about signals since this is the right name.
Calling just emit() produces error


akien-mga:
Thanks!

PR #1964, opened 2018-12-04
Fix some grammar and semantic errors in filesystem.rst

corrigentia:


getting_started/step_by_step/filesystem.rst
```diff
@@ -9,7 +9,7 @@ Introduction
 File systems are yet another hot topic in engine development. The
 file system manages how the assets are stored and how they are accessed.
 A well-designed file system also allows multiple developers to edit the
-same source files and assets while collaborating together.
+same source files and assets while working together.
```
LikeLakers2: Out of curiousity, could I ask why you made this change? It seems a little odd to me.

corrigentia: @LikeLakers2 : Alternative: "while collaborating." (removing "together").

That's all right. :smile:  Here's a short explanation:

to collaborate = to work together
< [from] Latin "collabōrāre": com- + labōrāre

labōrāre = to work
< labor = toil

LikeLakers2: Ah, I see. Thanks for the explanation.

Personally, I like your alternative ("while collaborating.", removing "together") a little more. That said, "while working together" and "while collaborating" effectively mean the same things, so I'll leave it up to you whether you want to make that change or not.

corrigentia: @LikeLakers2 : My pleasure! :wink: And I'll take your advice and change it to what I had initially offered only as an alternative. ~We~ I aim to please. :stuck_out_tongue:

akien-mga:
Thanks!

PR #1967, opened 2018-12-04
"Beneath the surface" more common than "Below the surface"

corrigentia:
beneath = under
below = lower than
Abstract concept, not tangible planar surface of water, for instance.

### DIFF
--- a/getting_started/editor/unity_to_godot.rst
+++ b/getting_started/editor/unity_to_godot.rst
@@ -54,10 +54,10 @@ The pictures below display both editors with colored blocks to indicate common f
 
 Note that Godot editor allows you to dock each panel at the side of the scene editor you wish.
 
-While both editors may seem similar, there are many differences below the surface.
+While both editors may seem similar, there are many differences beneath the surface.
 Both let you organize the project using the filesystem,
 but Godot's approach is simpler with a single configuration file, minimalist text format,
-and no metadata. All this contributes to Godot being much friendlier to VCS systems such as Git, Subversion, or Mercurial.
+and no metadata. All this contributes to Godot being much friendlier to VCS systems, such as Git, Subversion, or Mercurial.
 
 Godot's Scene panel is similar to Unity's Hierarchy panel but, as each node has a specific function,
 the approach used by Godot is more visually descriptive. In other words, it's easier to understand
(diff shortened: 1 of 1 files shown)


bojidar-bg:
[Google Ngrams](https://books.google.com/ngrams/graph?content=beneath+the+surface%2C+below+the+surface&year_start=1800&year_end=2000&corpus=15&smoothing=3&share=&direct_url=t1%3B%2Cbeneath%20the%20surface%3B%2Cc0%3B.t1%3B%2Cbelow%20the%20surface%3B%2Cc0) doesn't think so, unless I'm missing something:

[beneath the surface, below the surface](https://books.google.com/ngrams/graph?content=beneath+the+surface%2C+below+the+surface&year_start=1800&year_end=2000&corpus=15&smoothing=3&share=&direct_url=t1%3B%2Cbeneath%20the%20surface%3B%2Cc0%3B.t1%3B%2Cbelow%20the%20surface%3B%2Cc0)

akien-mga:
Seems to be more or less equivalent in a Google search (when search `"below the surface" -series` to filter out results about the TV series). Since both are pretty common and used seemingly interchangeably, I prefer not to break translations needlessly and keep "below the surface".

akien-mga:
Thanks!

corrigentia:
@bojidar-bg : [The Pitfalls of Using Google Ngram to Study Language | WIRED](https://www.wired.com/2015/10/pitfalls-of-studying-language-with-google-ngram/) excerpts:

<<[...] relying on Google Ngram to study the rise and fall of words and ideas has plenty of pitfalls. A [...] paper [...] outlines some of the major problems with the corpus of scanned books that powers Google Ngram. “It’s so beguiling, so powerful,” says Peter Sheridan Dodds, an applied mathematician at the University of Vermont who co-authored the paper. “But I think there’s a misrepresentation of what people should expect from this corpus right now.” Here are some of the problems.>>

<<is indicative of a shift in society, too.

But the tricky part here is more subtle. If scientific publications are taking up more and more of the the corpus, certain non-scientific terms may appear to fall in relative popularity. For example, are writers less interested in writing about “autumn” or are there just simply more scientific papers totally unrelated to “autumn” crowding the corpus?>>

<<people who are sort of being overly enthusiastic, using data in ways that can’t possibly be justified. And other people try to slam the brakes on it,” he says.

[...] But in a way, it's so easy to use that it lends itself to overuse—and misuse. >>

The point I'm trying to make is that more people use English now, even though they may not be proficient in it.

And yes, you've also missed the semantic context of the word in this particular case. That reminds me of [Freud's distinction](https://en.wikipedia.org/wiki/Subconscious):

"If someone talks of subconsciousness, I cannot tell whether he means the term topographically – to indicate something lying in the mind beneath consciousness – or qualitatively – to indicate another consciousness, a subterranean one, as it were. He is probably not clear about any of it. The only trustworthy antithesis is between conscious and unconscious."

PR #1975, opened 2018-12-05
Fix some grammatical errors in command_line_tutorial.rst

corrigentia:


getting_started/editor/command_line_tutorial.rst
```diff
@@ -7,14 +7,14 @@ Command line tutorial
 
 Some developers like using the command line extensively. Godot is
 designed to be friendly to them, so here are the steps for working
-entirely from the command line. Given the engine relies on little to no
+entirely from the command line. Given the engine relies on few to no
```
akien-mga: Is "few to no <something countable>" really idiomatic? I've never heard that. "little to no" is a common construct, but if it's invalid here, it would be better to rephrase to avoid the awkwardness of "few to no".

akien-mga: It does seem valid (~1M hits in Google, vs ~80M for "little to no"), but still sounds awkward to me. Maybe "almost no" or "practically no"?

corrigentia: > Is "few to no " really idiomatic? I've never heard that. "little to no" is a common construct, but if it's invalid here, it would be better to rephrase to avoid the awkwardness of "few to no".

"Library, libraries" is a countable noun, so the correct form is "many/a few libraries", NOT "~much~/~a little~ libraries". One forms the interrogative "How  **many**  _libraries_?", NOT "How ~much~ _libraries_?".

Contrast that with "milk", which is an uncountable noun, so takes "a lot of/a little milk", NOT "~many~/~a few~ milk". One forms the interrogative "How  **much** _milk_?", NOT "How ~many~ _milk_?".

> It does seem valid (~1M hits in Google, vs ~80M for "little to no"), but still sounds awkward to me. Maybe "almost no" or "practically no"?

I've changed it to "almost no".

akien-mga:
Thanks!

PR #1976, opened 2018-12-05
Fix miscellaneous errors in gdscript_basics.rst

corrigentia:
```GDScript
var dict = {"Z": 3, 8: "middle", "c": [[], []]}
for i in dict:
    print(dict[i])
```
prints
```GDScript
3
middle
[[], []]
```
in both Godot 3.0 stable and 3.1 alpha. ... as expected. I had failed to notice the ``dict`` part in ``print(dict[i])`` because it was just below the ``dict`` in ``for i in dict:``.


akien-mga:
Thanks!

PR #1977, opened 2018-12-05
Break comments to make them readable

mandarinx:
The comments are wider than the viewable area in the docs, which makes them hard to read. On Mac, the scrollbars are hidden so you either need an Apple mouse that can scroll sideways, or fumble with scrolling and catching the scroll handle as it appears, so that you can drag it sideways. It's easier for all of us to wrap the comments.


akien-mga:
Thanks!

PR #1978, opened 2018-12-05
Fix some grammar in gdscript_advanced.rst

corrigentia:
The adjective "strict" has a single syllable, so its comparative form is the regular "-er".

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_advanced.rst
+++ b/getting_started/scripting/gdscript/gdscript_advanced.rst
@@ -368,7 +368,7 @@ The range() function can take 3 arguments:
     range(b, n) # Will go from b to n-1
     range(b, n, s) # Will go from b to n-1, in steps of s
 
-Some staticly typed programming language examples:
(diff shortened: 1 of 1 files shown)


getting_started/scripting/gdscript/gdscript_advanced.rst
```diff
@@ -368,7 +368,7 @@ The range() function can take 3 arguments:
     range(b, n) # Will go from b to n-1
     range(b, n, s) # Will go from b to n-1, in steps of s
 
-Some examples:
+Some staticly typed programming language examples:
```
akien-mga: "statically typed" seems 100 times more common than "staticly typed" according to Google (it's the first time I read the latter).

corrigentia: @akien-mga : Thanks for catching another one! :grin: Definitely wrong! I was hurrying and not paying attention.

akien-mga:
Thanks!

PR #1982, opened 2018-12-06
Fix some grammar in static_typing.rst

corrigentia:
"Coming with" is the most conspicuously awkward construction on this page that I haven't touched because the intent wasn't easy to guess.

### DIFF
--- a/getting_started/scripting/gdscript/static_typing.rst
+++ b/getting_started/scripting/gdscript/static_typing.rst
@@ -85,7 +85,7 @@ How to use static typing
 
 To define the type of a variable or a constant, write a colon after the
 variable’s name, followed by its type. E.g. ``var health: int``. This
-forces the type of variable to always stay the same:
+forces the variable's type to always stay the same:
 
 ::
 
@@ -113,7 +113,7 @@ Currently you can use three types of… types:
 
 .. note::
 
-    You don't need to write type hints for constants as Godot sets it automatically from the assigned value. But you can still do so to make the intent of your code clearer.
+    You don't need to write type hints for constants, as Godot sets it automatically from the assigned value. But you can still do so to make the intent of your code clearer.
 
 Custom variable types
 ~~~~~~~~~~~~~~~~~~~~~
@@ -136,7 +136,7 @@ For the example above, your Rifle.gd would look like this:
     class_name Rifle
 
 If you use ``class_name``, Godot registers the Rifle type globally in
-the editor, and you can use it anywhere without having to preload it
+the editor, and you can use it anywhere, without having to preload it
 into a constant:
 
 ::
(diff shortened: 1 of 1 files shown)


akien-mga:
> Typed GDScript is a powerful tool. Coming with Godot 3.1, it can already help you write more structured code, help you avoid common errors, and create scalable systems.

"Coming with" is used to express that this feature is a new feature in Godot 3.1.

corrigentia:
> "Coming with" is used to express that this feature is a new feature in Godot 3.1.

So something like a feature that is actively being worked on, but has yet to be released to market or for stable, reliable use in "production"?
In that case, maybe something like "Coming up in Godot 3.1, it will help you write more structured code, avoid common errors, and create scalable systems"?

akien-mga:
> So something like a feature that is actively being worked on, but has yet to be released to market or for stable, reliable use in "production"?

Indeed, it has been implemented in Godot's *master* (development) branch, so it's available already in development snapshots (3.1 alpha 2 currently, and nightly builds). But it's not yet available in the stable branch (3.0.x), which many readers might still be using even while browsing the "latest" documentation (instead of "stable").

>  In that case, maybe something like "Coming up in Godot 3.1, it will help you write more structured code, avoid common errors, and create scalable systems"?

Sounds good. Or, since we're getting closer to release, we could drop the version info and just rephrase the sentence assuming that this feature *is* available (since the "latest" docs are meant to target 3.1).

corrigentia:
> [...] Since we're getting closer to release, we could drop the version info and just rephrase the sentence assuming that this feature _is_ available (since the "latest" docs are meant to target 3.1).

@akien-mga : How about "A new feature, it helps you write more structured code, avoid common errors, and create scalable systems."?

akien-mga:
Sounds good to me.

akien-mga:
Well maybe "A new feature since Godot 3.1,", as otherwise this statement will be outdated when we're working on 3.2 :)

corrigentia:
> [...] this statement will be outdated when we're working on 3.2 :)

@akien-mga : You're absolutely right! That was a lack of foresight on my part. :blush:  I've omitted to make the documentation future-proof.

> [...] maybe "A new feature since Godot 3.1," [...]

I suppose you mean it's OK for me to mention the point at which it was introduced, after all, since it accounts for both the case of someone using a version that doesn't provide the feature and the case of someone from the future - when this will have long been available - reading the section, and lets them know what to expect and establishes when this feature was implemented.

Well then, how does "Available as of version 3.1 of Godot, it helps you write more structured code, avoid common errors, and create scalable systems." sound to you?

akien-mga:
> Well then, how does "Available as of version 3.1 of Godot, it helps you write more structured code, avoid common errors, and create scalable systems." sound to you?

Sounds great :)

corrigentia:
> Sounds great :)

@akien-mga : Glad you liked it. I've added the change to the commit.

PR #1985, opened 2018-12-06
Fix some grammar in nodes_purposes.rst

corrigentia:
"Low-level" and "built-in" are compound adjectives, which use a hyphen to "glue" together separate words.


akien-mga:
Thanks!

PR #1988, opened 2018-12-07
Fix some grammar in c_sharp_features.rst

corrigentia:
One gets an "[overview OF](https://www.thefreedictionary.com/overview)" something.

The [past participle](https://en.wikipedia.org/wiki/Participle) form of the verb "[to cast](https://pasttenses.com/cast-past-tense)" is "[cast](https://www.usingenglish.com/reference/irregular-verbs/cast.html)"; [that is the one used](https://grammarist.com/usage/casted/) as an adjective.


akien-mga:
Thanks!

Issue #1986, opened 2018-12-07
Possible error: "to tag scenes which are enemies" is unclear


sudoman:
<https://docs.godotengine.org/en/3.0/getting_started/step_by_step/scripting_continued.html#groups>

> One example would be to tag scenes which are enemies

Do you mean "tag enemies in a scene"?

sudoman:
Based on early documentation before this point, it seemed as if scenes couldn't be nodes, thus the reason why that quote seemed incorrect. However there is some evidence a bit later on in the docs that imply that scenes can be nodes.

bojidar-bg:
A scene is a node plus its children and their properties. Maybe it is not clear enough by that point in the docs, in which case it would be nice to clear it.

I think that the text "to tag scenes which are enemies" can be safely removed or, perheaps, put as a paragraph below the UI and code examples, explaining it in more detail.

clayjohn:
I think an easy fix would be to just change "scenes" to "nodes", the rest of the section refers to the enemies as nodes.

akien-mga:
Fixed by #1991.

PR #1991, opened 2018-12-08
Changed "scenes" to "nodes" for clarity

clayjohn:
In the scripting_continues step_by_step tutorial. This addresses issue #1986 

I figured this was the simplest change, plus it aligns with the language in the rest of the section. 

The user spotted the issue in the 3.0 branch, so it should be cherrypicked there as well.


akien-mga:
Thanks!

Issue #1913, opened 2018-11-09
Minimal jitter?


BillWelden:
The page http://docs.godotengine.org/en/3.0/getting_started/workflow/assets/importing_scenes.html has the phrase "Failing to do this may result in minimal jitter."

Did you mean either "Failing to do this may result in some small amount of jitter." or "Doing this will result i miminal jitter."

"minimal jitter" means "the smallest possible amount of jitter", which is, of course, a good thing.

akien-mga:
Fixed by #1992.

PR #1992, opened 2018-12-08
Clarify "minimal jitter" in importing_scenes tutorial

clayjohn:
This addresses issue #1913

### DIFF
--- a/getting_started/workflow/assets/importing_scenes.rst
+++ b/getting_started/workflow/assets/importing_scenes.rst
@@ -224,7 +224,7 @@ FPS
 
 Most 3D export formats store animation timeline in seconds instead of frames. To ensure
 animations are imported as faithfully as possible, please specify the frames per second
-used to edit them. Failing to do this may result in minimal jitter.
+used to edit them. Failing to do this may result in shaky animations.
 
 Filter Script
 ~~~~~~~~~~~~~


getting_started/workflow/assets/importing_scenes.rst
```diff
@@ -224,7 +224,7 @@ FPS
 
 Most 3D export formats store animation timeline in seconds instead of frames. To ensure
 animations are imported as faithfully as possible, please specify the frames per second
-used to edit them. Failing to do this may result in minimal jitter.
+used to edit them. Failing to do this may result in a small amount of jitter.
```
akien-mga: "Jitter" isn't really countable so it's strange to me to speak about "an amount of jitter". I'd remove the qualifier altogether.

clayjohn: What about "perceptible jittering"?

clayjohn: Actually what about "may result in shaky animations"? I have a feeling that jitter isn't going to translate well, and it's a kinda awkward word in English anyway.

akien-mga: Sounds good.

clayjohn:
@akien-mga Should be good to go now!

akien-mga:
Thanks!

PR #1994, opened 2018-12-08
Improve the Compiling for Android page

Calinou:
This mentions that Gradle will automatically download SDK components (except the NDK, as it is used before Gradle is called).

I removed the Linux build troubleshooting section as I haven't been able to reproduce it on several systems. If someone can still reproduce it, I'll add it back.

Pywin32 should also no longer be necessary now that the SCons parallel build logic was overhauled in Godot.

### DIFF
--- a/development/compiling/compiling_for_android.rst
+++ b/development/compiling/compiling_for_android.rst
@@ -8,27 +8,31 @@ Compiling for Android
 Note
 ----
 
-For most cases, using the built-in deployer and export templates is good
+In most cases, using the built-in deployer and export templates is good
 enough. Compiling the Android APK manually is mostly useful for custom
 builds or custom packages for the deployer.
 
-Also, you still need to do all the steps mentioned in the
-:ref:`doc_exporting_for_android` tutorial before attempting your custom
-export template.
+Also, you still need to follow the steps mentioned in the
+:ref:`doc_exporting_for_android` tutorial before attempting to build
+a custom export template.
 
 Requirements
 ------------
 
 For compiling under Windows, Linux or macOS, the following is required:
 
--  Python 2.7+ or Python 3.5+
--  SCons build system
--  [Windows only] PyWin32 (optional, for parallel compilation)
--  Android SDK version 23.0.3 [Note: Please install all tools and extras of the SDK Manager]
--  Android build tools version 19.1
--  Android NDK r13 or later
+-  `Python 2.7+ or Python 3.5+ <https://www.python.org/downloads/>`_
+-  `SCons <https://scons.org/pages/download.html>`_ build system
+-  `Android SDK <https://developer.android.com/studio/#command-tools>`_ (command-line tools are sufficient)
+
+   -  Required SDK components will be automatically installed by Gradle (except the NDK)
+
+-  `Android NDK <https://developer.android.com/ndk/downloads/>`_ r17 or later
 -  Gradle (will be downloaded and installed automatically if missing)
--  JDK 6 or later (either OpenJDK or Oracle JDK) - JDK 9 & 10 do not work with current Gradle.
+-  JDK 8 (either OpenJDK or Oracle JDK)
+
+   -  JDK 9 or later are not currently supported
+   -  You can download a build from `ojdkbuild <https://github.com/ojdkbuild/ojdkbuild>`_
 
 .. seealso:: For a general overview of SCons usage for Godot, see
              :ref:`doc_introduction_to_the_buildsystem`.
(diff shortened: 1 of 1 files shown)


development/compiling/compiling_for_android.rst
```diff
@@ -8,147 +8,154 @@ Compiling for Android
...
--  Android SDK version 23.0.3 [Note: Please install all tools and extras of the SDK Manager]
--  Android build tools version 19.1
--  Android NDK r13 or later
+-  `Python 2.7+ or Python 3.5+ <https://www.python.org/downloads/>`_
+-  `SCons <https://scons.org/pages/download.html>`_ build system
+-  `Android SDK <https://developer.android.com/studio/#command-tools>`_ (command-line tools are sufficient)
+
+   -  Required SDK components will be automatically installed by Gradle (except the NDK)
+
+-  `Android NDK <https://developer.android.com/ndk/downloads/>`_ r17 or later
```
akien-mga: I think r16 would still work currently. But it doesn't hurt to ask for newer anyway.

development/compiling/compiling_for_android.rst
```diff
@@ -8,147 +8,154 @@ Compiling for Android
...
 
--  Python 2.7+ or Python 3.5+
--  SCons build system
--  [Windows only] PyWin32 (optional, for parallel compilation)
--  Android SDK version 23.0.3 [Note: Please install all tools and extras of the SDK Manager]
--  Android build tools version 19.1
--  Android NDK r13 or later
+-  `Python 2.7+ or Python 3.5+ <https://www.python.org/downloads/>`_
+-  `SCons <https://scons.org/pages/download.html>`_ build system
+-  `Android SDK <https://developer.android.com/studio/#command-tools>`_ (command-line tools are sufficient)
```
akien-mga: What exact component is needed should likely be made clearer, users are often confused about it. In my experience at least the "Android SDK Tools", "Android SDK Platform-tools", latest (or target from manifest?) "Android SDK Build-tools" and "Android Support Repository".

I'll try again from scratch to see what it wants now.

development/compiling/compiling_for_android.rst
```diff
@@ -8,147 +8,154 @@ Compiling for Android
...
 
 ::
 
-    C:\godot> scons platform=android target=release
-    C:\godot> cd platform/android/java
-    C:\godot\platform\android\java> gradlew build
+    scons platform=android target=release
+    cd platform/android/java
+    # On Windows
+    .\gradlew build
```
akien-mga: I think `gradlew build` would work on all three OSes (if `gradlew` is executable on Linux/macOS, which it is). So we can maybe do without the distinction here.

akien-mga: Nevermind, doesn't work on Linux without `./`.

Calinou: It's also interesting to note that `.` is not in the `PATH` by default in PowerShell (but it is in cmd.exe), so `.\gradlew build` will work both in PowerShell and cmd.exe.

development/compiling/compiling_for_android.rst
```diff
@@ -8,147 +8,154 @@ Compiling for Android
...
-
+If you have multiple CPU cores available, you can speed up compilation
+by adding the ``-jX`` argument to the SCons command, where ``X`` is the
+number of cores you want to allocate to building, e.g. ``scons -j4``.
 
 Adding support for x86 devices
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-If you also want to include support for x86 devices, run the scons command
+If you also want to include support for x86 devices, run the SCons command
```
akien-mga: This section (and the first one about the "default" build) should be rewritten as there are now three targets to take into account, and people should build at least `armv7` and `arm64-v8` (and optionally `x86` if they want).

So it should be made clearer from the start that you build Godot for all your target platforms first, and *then* you run `gradlew build` to generate the APK.

development/compiling/compiling_for_android.rst
```diff
@@ -8,147 +8,154 @@ Compiling for Android
...
+-  `Android NDK <https://developer.android.com/ndk/downloads/>`_ r17 or later
 -  Gradle (will be downloaded and installed automatically if missing)
--  JDK 6 or later (either OpenJDK or Oracle JDK) - JDK 9 & 10 do not work with current Gradle.
+-  JDK 8 (either OpenJDK or Oracle JDK)
+
+   -  JDK 9 or later are not currently supported
+   -  You can download a build from `ojdkbuild <https://github.com/ojdkbuild/ojdkbuild>`_
 
-.. seealso:: For a general overview of SCons usage for Godot, see
+.. seealso:: For a general overview of SCons usage in Godot, see
```
akien-mga: This sentence is used in all "Compiling..." pages, so best not change it, or change it everywhere. I think "for" is more accurate here, so no change needed (it's not used "in" (by) Godot per se).

development/compiling/compiling_for_android.rst
```diff
@@ -8,147 +8,154 @@ Compiling for Android
...
 
 Faster compilation
 ~~~~~~~~~~~~~~~~~~
 
-If you are on Unix or installed PyWin32 on Windows and have multiple CPU
-cores available, you can speed up the compilation by adding the ``-jX``
-argument to the SCons command, where ``X`` is the number of cores that you
-want to allocate to the compilation, e.g. ``scons -j4``.
-
+If you have multiple CPU cores available, you can speed up compilation
```
akien-mga: That section can be removed, it's already covered in the intro to the buildsystem linked above in the `.. seealso` section.

development/compiling/compiling_for_android.rst
```diff
@@ -8,147 +8,154 @@ Compiling for Android
...
+    scons platform=android target=release android_arch=x86
+    cd platform/android/java
+    # On Windows
+    .\gradlew build
+    # On Linux/macOS
+    ./gradlew build
 
 
-This will create a fat binary that works in both platforms, but will add
+This will create a fat binary that works on both platforms, but will add
```
akien-mga: That's a bit misleading actually. The generate APK contains the 2 (or 3, with `arm64-v8`) libraries, but when you export the game you will select which platforms you want to target. The size of the resulting APK depends on that choice, and not the original size of the base template.

IMO it would be better to explain this instead.

development/compiling/compiling_for_android.rst
```diff
@@ -8,147 +8,154 @@ Compiling for Android
...
 ~~~~~~~~~~~~~~~
 
 It might be necessary to clean the build cache between two APK compilations,
 as some users have reported issues when building the two export templates
-one after the other.
+one after the other. This can be done using the following command:
+
+::
+
+    scons platform=android -c
```
akien-mga: This will only clean build objects for the default arch (`armv7`). One should specify the exact same arguments as used for compilation to `scons -c`.

This paragraph was added in 2016 and honestly I never saw any mention of it being needed (and as we well know most people don't read docs, or wouldn't know how to clean things here, so if it was indeed needed we'd have heard about it by now), so I'd say best remove it.

development/compiling/compiling_for_android.rst
```diff
@@ -171,80 +178,33 @@ Troubleshooting
...
 
 -  Make sure to use export templates that match your editor version; if
    you use a new Godot version, you *have* to update the templates too.
--  libgodot_android.so is not in ``lib/armeabi-v7a`` or ``lib/armeabi``
--  Device does not support armv7 (try compiling yourself for armv6)
--  Device is Intel, and apk is compiled for ARM.
+-  ``libgodot_android.so`` is not in ``lib/armeabi-v7a`` or ``lib/armeabi``.
+-  The device does not support ARMv7; try compiling for ARMv6 by passing
+   ``android_arch=armv6`` to the SCons command line.
+-  The device is x86, but the APK is only compiled for ARM.
```
akien-mga: Should also mention `arm64-v8`. But same comment as above, it would be better to rephrase and explain that the exported APK can have support for 1 or more platforms depending on 1) what was compiled in the template 2) what was toggled during export.

akien-mga: And as such the "The device does not support ARMv7; try compiling for ARMv6" is also bad advice, as in most cases nowadays, if a device does not support ARMv7 it's likely that it's x86 or ARM64, not obsolete v6. I would simply rephrase it as "The device's architecture does not match the exported one(s). Make sure that your templates were built against that device's architecture, and that the export settings enabled support for that architecture." (feel free to phrase differently)

development/compiling/compiling_for_android.rst
```diff
@@ -8,147 +8,154 @@ Compiling for Android
...
+    scons platform=android -c
 
 Using the export templates
 --------------------------
 
-As export templates for Android, Godot needs release and debug APKs that
-were compiled against the same version/commit as the editor. If you are
-using official binaries for the editor, make sure to install the matching
-export templates, or to build your own from the same version.
+Godot needs release and debug APKs that were compiled against the same
```
akien-mga: It would be good to refer to the "Exporting for Android" docs here. I haven't checked if they are explicit about how to select the architectures that should be included in the exported APK, but if not it should likely be added there and referenced here.

akien-mga:
Thanks, great changes overall :)

development/compiling/compiling_for_android.rst
Calinou: The Exporting for Android page doesn't really explain it, so I'll skip that for now.

development/compiling/compiling_for_android.rst
akien-mga: I tried to set things up anew to document the exact steps needed. Here's what I had to do on Linux:
```
21:26 <Akien> - Download command line tools from https://developer.android.com/studio/#command-tools
21:26 <Akien> - Unzip in the root of the (empty) ANDROID_SDK_HOME, gives the "tools" folder
21:27 <Akien> - Run `./tools/bin/sdkmanager --licenses` to accept licenses
21:27 <Akien> Then gradlew will handle the rest
```
The step for accepting licenses is not straightforward, so IMO it would be worth being explicit about it. Maybe not in this list of requirements as it would quickly get crowded, but maybe a paragraph after the list (and here a "See below for installation details.").

development/compiling/compiling_for_android.rst
```diff
@@ -8,147 +8,145 @@ Compiling for Android
...
+   -  You can download a build from `ojdkbuild <https://github.com/ojdkbuild/ojdkbuild>`_
 
 .. seealso:: For a general overview of SCons usage for Godot, see
              :ref:`doc_introduction_to_the_buildsystem`.
 
 Setting up the buildsystem
 --------------------------
 
-Set the environment variable ANDROID_HOME to point to the Android
+Set the environment variable ``ANDROID_HOME`` to point to the Android
```
akien-mga: Provided that we explain the workflow with command-line tools (see above), might be worth adding this:

"If using the Android command-line tools only, this would be the folder where you placed the downloaded `tools` folder. Later on, `gradlew` will install necessary SDK components in this folder."

development/compiling/compiling_for_android.rst
```diff
@@ -8,147 +8,145 @@ Compiling for Android
...
+-  Release template (used when exporting with **Debugging Enabled** unchecked)
 
 ::
 
-    C:\godot> scons platform=android target=release
-    C:\godot> cd platform/android/java
-    C:\godot\platform\android\java> gradlew build
-
-(on Linux or macOS, execute the `gradlew` script with `./gradlew build`)
+    scons platform=android target=release
```
akien-mga: I would put `android_arch=armv7` here. It's not necessary per se, but explicit is better.

development/compiling/compiling_for_android.rst
```diff
@@ -8,147 +8,145 @@ Compiling for Android
...
 ::
 
-    C:\godot> scons platform=android target=release_debug
-    C:\godot> cd platform/android/java
-    C:\godot\platform\android\java> gradlew build
-
-The resulting APK is in:
-
-::
+    scons platform=android target=release_debug
```
akien-mga: Same here.

development/compiling/compiling_for_android.rst
```diff
@@ -8,147 +8,145 @@ Compiling for Android
...
+with Gradle. For example, for the release template:
 
 ::
 
-    C:\godot> scons platform=android target=release
-    C:\godot> scons platform=android target=release android_arch=x86
-    C:\godot> cd platform/android/java
-    C:\godot\platform\android\java> gradlew build
-
+    scons platform=android target=release
```
akien-mga: And here.

development/compiling/compiling_for_android.rst
```diff
@@ -8,147 +8,145 @@ Compiling for Android
...
-
-Troubleshooting
-~~~~~~~~~~~~~~~
 
-It might be necessary to clean the build cache between two APK compilations,
-as some users have reported issues when building the two export templates
-one after the other.
+This will create a fat binary that works on all platforms.
+The final APK size of exported projects will depend on the platforms you choose
+to support when exporting; in other words, unused platforms will be removed from
```
akien-mga: Nice wording :)

akien-mga:
Perfect, thanks!

PR #2006, opened 2018-12-10
Fix subject-verb agreement in index.rst

corrigentia:
"[...] Makefile and python script [...] are present [...]"


akien-mga:
Thanks!

PR #2008, opened 2018-12-10
Fix some spelling in material.rst

corrigentia:
GSoC = [assumed] Google Summer of Code

e.g. [Latin] = exempli gratia
exempli gratia = for example


akien-mga:
Thanks!

PR #2009, opened 2018-12-10
Add some possessive case apostrophes to physics.rst

corrigentia:


akien-mga:
Added some more (unrelated) formatting fixes.

PR #2026, opened 2018-12-11
Fix some grammar, punctuation, spelling and compliance with the GDScript style guide in custom_drawing_in_2d.rst

corrigentia:


tutorials/2d/custom_drawing_in_2d.rst
```diff
@@ -68,18 +68,18 @@ If re-drawing is required because a state or something else changed,
 simply call :ref:`CanvasItem.update() <class_CanvasItem_update>`
 in that same node and a new _draw() call will happen.
 
-Here is a little more complex example. A texture variable that will be
+Here is a little more complex example - a texture variable that will be
```
bojidar-bg: Perhaps an em dash character will be better here? (—)

tutorials/2d/custom_drawing_in_2d.rst
```diff
@@ -154,17 +154,17 @@ portion of a circle? You will have to code a function to perform this and draw i
...
 
-Basically, drawing a shape on screen requires it to be decomposed into a certain number of points
-linked from one to the following one. As you can imagine, the more points your shape is made of,
+Basically, drawing a shape on the screen requires it to be decomposed into a certain number of points
+linked from one to the next. As you can imagine, the more points your shape is made of,
 the smoother it will appear, but the heavier it will also be in terms of processing cost. In general,
 if your shape is huge (or in 3D, close to the camera), it will require more points to be drawn without
 it being angular-looking. On the contrary, if your shape is small (or in 3D, far from the camera),
-you may reduce its number of points to save processing costs. This is called *Level of Detail (LoD)*.
+you may reduce its number of points to save processing costs; this is called *Level of Detail (LoD)*.
```
bojidar-bg: Suggestion: "is called" → "is known as"

tutorials/2d/custom_drawing_in_2d.rst
```diff
@@ -211,7 +211,7 @@ We first determine the angle of each point, between the starting and ending angl
 The reason why each angle is reduced by 90° is that we will compute 2D positions
 out of each angle using trigonometry (you know, cosine and sine stuff...). However,
 to be simple, cos() and sin() use radians, not degrees. The angle of 0° (0 radian)
-starts at 3 o'clock although we want to start counting at 12 o'clock. So we reduce
+starts at 3 o'clock, although we want to start counting at 12 o'clock. So we reduce
```
bojidar-bg: Suggestion (here and above): "reduce" → "decrease", as reduction is something else in mathematics.
[Wiktionary: reduce](https://en.wiktionary.org/wiki/reduce#Verb): (mathematics) To simplify an equation or formula without changing its value.

tutorials/2d/custom_drawing_in_2d.rst
```diff
@@ -225,17 +225,17 @@ Finally, we insert the point in the PoolVector2Array which was previously define
...
 increase the number of points.
 
-Draw the arc on screen
-^^^^^^^^^^^^^^^^^^^^^^
+Draw the arc on the screen
+^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 We now have a function that draws stuff on the screen:
-It is time to call in the _draw() function.
+It is time to call it inside the _draw() function.
```
bojidar-bg: Suggestion: Swap the period (.) and colon (:) on those two sentences.

tutorials/2d/custom_drawing_in_2d.rst
```diff
@@ -384,7 +384,7 @@ calls _draw(). This way, you can control when you want to refresh the frame.
         _angleFrom += _rotationAngle;
         _angleTo += _rotationAngle;
 
-        // We only wrap angles if both of them are bigger than 360
+        // We only wrap angles if each of them is bigger than 360.
```
bojidar-bg: Here and above: IMHO "both" is better than "each", as it emphasizes that both conditions should be met, and not that something is done for each angle.
Potentially, "if" can be changed to "when".

tutorials/2d/custom_drawing_in_2d.rst
corrigentia: @bojidar-bg Typographically, you are correct (assuming the code sample is interpreted as a clause following the colon). However, I'm afraid the character will not display properly for some readers, as the em dash is not part of the ASCII character set. (Aside: typewriters have disgraced the conventions established by the printing press.) If Unicode is known to be fully supported, then an em dash could be substituted there.

LikeLakers2: Correct me if I'm wrong, but I believe we use the em dash elsewhere in the docs as well. Additionally, Unicode tends to display on most devices that I've seen, with the usual exceptions being *really* old phones (y'know, the ones that were made before Android and iOS). I don't believe there's much to be concerned about when it comes to Unicode support.

tutorials/2d/custom_drawing_in_2d.rst
corrigentia: @bojidar-bg : I disagree with your first assessment; think of "each" as a filter function that takes 2 arguments and checks whether  **each** of them meets a given condition, and then performs a logical conjunction on those intermediary results.
Either way, I've chosen to go the "when" route.
All in all, I think the quality of your remarks has improved considerably since the last time I "saw" you. Keep at it! :sunglasses:

tutorials/2d/custom_drawing_in_2d.rst
bojidar-bg: I just suggested it as some people don't know that em and en dash symbols exist in Unicode, and thus reuse the hyphen for them. Anyway, I rarely use dashes myself, so I like the comma more. :)

akien-mga:
Thanks!

PR #2032, opened 2018-12-12
Remove potentially confusing repetition and fix miscellaneous issues in introduction_to_3d.rst

corrigentia:


tutorials/3d/introduction_to_3d.rst
```diff
@@ -190,22 +190,22 @@ light to your objects.
...
+No matter how many objects are placed in the 3D space, nothing will be
 displayed unless a :ref:`Camera <class_Camera>` is
-also added to the scene. Cameras can either work in orthogonal or
+also added to the scene. Cameras can work in either orthogonal or
 perspective projections:
 
 .. image:: img/tuto_3d10.png
 
-Cameras are associated with and only display to a parent or grandparent
+Cameras are associated with -- and only display to -- a parent or grandparent
```
akien-mga: Shouldn't this use em dashes (`—`)?

Nevertheless, I think commas or parentheses would work just a well here, and are probably better understood by non-native speakers than em dashes (as the latter aren't really idiomatic in most languages (at least, in those I know)).

tutorials/3d/introduction_to_3d.rst
```diff
@@ -190,22 +190,22 @@ light to your objects.
...
 viewport. Since the root of the scene tree is a viewport, cameras will
 display on it by default, but if sub-viewports (either as render target
 or picture-in-picture) are desired, they need their own children cameras
 to display.
 
 .. image:: img/tuto_3d11.png
 
-When dealing with multiple cameras, the following rules are followed for
+When dealing with multiple cameras, the rules about to be mentioned are followed for
```
akien-mga: That's a bit too emphatic IMO. To avoid the repetition, I'd suggest using "the following rules are applied for each viewport" instead. (or "are used")

corrigentia: @akien-mga : How about "[...] the following rules are enforced for each viewport: [...]"?

tutorials/3d/introduction_to_3d.rst
corrigentia: @akien-mga : I've had a similar conversation with @bojidar-bg about em dashes. The concern I raised then was that some readers' devices might not properly render the Unicode em dash character.
At any rate, I suppose I could use parentheses in this case.

tutorials/3d/introduction_to_3d.rst
akien-mga: Sounds good, yes.

tutorials/3d/introduction_to_3d.rst
akien-mga: I think all browsers and OSes currently supported on the market have decent support for Unicode. Emojis made sure of that :)

But yes, I think parentheses are "safer" here.

tutorials/3d/introduction_to_3d.rst
corrigentia: @akien-mga : Cool. So I'm done with this one.

tutorials/3d/introduction_to_3d.rst
corrigentia: @akien-mga : I try not to marginalize Luddites.

PR #2036, opened 2018-12-12
Update class_object.rst

### LINKED ISSUE #1968: Is "new()" method missing in the reference doc?
It's possible to use `Texture.new()` but the `new()` method is not listed anywhere in the doc of `Texture` type nor in the doc of the parent classes (`Resource`, `Reference`, `Object`). Is it intended?
https://docs.godotengine.org/en/3.0/classes/class_object.html#class-object

The only place in the doc where the `new()` method is mentioned is here: https://docs.godotengine.org/en/3.0/classes/class_gdscript.html#class-gdscript-new

davcri:
Add a note on how to instantiate new Objects. Closes  #1968


davcri:
Thanks @LikeLakers2, I'm crafting a new PR sorry for the trouble.

PR #2038, opened 2018-12-12
Updated info regarding c# exported variables

exts:
There was a person on discord that said it wasn't clear regarding why exported variables wasn't showing up in the editor in the your first game example. It was changed where all you needed to do is rebuild your project and it updated in the editor. So I made that clarification to help people not be confused with an old workaround.


akien-mga:
Thanks!

PR #2039, opened 2018-12-13
Docs: Tweak spatial_material

corrigentia:
Fix some grammar, punctuation and spelling mistakes.

### DIFF
--- a/tutorials/3d/spatial_material.rst
+++ b/tutorials/3d/spatial_material.rst
@@ -129,7 +129,7 @@ Use as Albedo
 
 Choosing this option means vertex color is used as albedo color.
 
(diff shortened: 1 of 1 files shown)


akien-mga:
Thanks!

Note that you don't need to prefix PR titles with "Docs:" here, this is the docs repo so it's already obvious ;)

corrigentia:
> Note that you don't need to prefix PR titles with "Docs:" here, this is the docs repo so it's already obvious ;)

@akien-mga : Duh. :blush: I think you've noticed by now that I could be accused of not paying attention to things... :grin: Thanks for pointing out what should have been obvious.

PR #2041, opened 2018-12-13
Fix end-string error in "Custom Drawing in 2D" page

YeldhamDev:


akien-mga:
This actually needs to be
<pre>
``\ s
</pre>
with a space, and there was another occurrence so I fixed both in dc2dfb62. (FYI @corrigentia)

corrigentia:
> This actually needs to be
> 
> ``\ s
> 
> with a space, and there was another occurrence so I fixed both in [dc2dfb6](https://github.com/godotengine/godot-docs/commit/dc2dfb62241283f48ae66548558ef031f3a8696e). (FYI @corrigentia)

@akien-mga : Thank you for the heads-up. I'll keep that in mind.

Issue #1921, opened 2018-11-15
Shading Reference Overhaul


clayjohn:
I plan on overhauling the shader documentation. Currently the shading language reference is fairly comprehensive and well put together. However, there is a lot of information on a [single page](http://docs.godotengine.org/en/latest/tutorials/shading/shading_language.html) and it is becoming increasingly difficult for new users to navigate.

My current idea is to split the Shading language document into a sub-directory containing a few different documents that cover: shaders generally, the Godot shading language, and the different shader types individually. In the process of doing this I will also add a lot more information on shaders and try to make the docs more accessible for new users. 

### Proposed Layout
Shading
- Shading language
  - Shaders
  - Shading language reference
  - Spatial shaders
  - Canvas item shaders
  - Particle shaders
- Shader materials
- Intro to shaders: 2D and 3D water
- ...

#### Shaders
The Shaders doc will introduce the reader to the concept of shaders and to Godot specific shader features (shader_types, render_modes, how a shader fits within a material etc.). It should be a point of first entry for those looking for reference information about shaders in Godot. It is not an introductory tutorial. It contains high-level information about shader_types; render_modes; and vertex, fragment, and light shaders as well as information about shaders generally. 

#### Shading language reference
This will contain the parts of the shading language reference that are not touched by the Shaders doc and the three shader type docs. It will serve as a useful page for more advanced users to read about things such as data types, flow control built-in functions etc. It will remain reference style documentation. I will edit it and expand on it as necessary.

#### Shader types
The three shader_type documents will contain what is in the shading language reference under [Shader types in depth](http://docs.godotengine.org/en/latest/tutorials/shading/shading_language.html#shader-types-in-depth). There will be one document for each shader_type that will contain render_modes, built-ins, and a description of how to use the shader. Most of this information is already contained in the shading language reference, however, I will expand upon it and organize it in an easier to follow fashion.

I decided to organize the documentation by shader type (spatial, canvas_item, particle) instead of processor function (vertex, fragment, light) because I figured different processor functions within the same shader type have more in common than the same processor functions from different shader types. For example, the vertex and fragment functions in spatial shaders have more in common than the vertex shader from a canvas_item shader has with a vertex shader from a spatial shader. Additionally, this provides a place to present information about render_modes. A description of the different processor functions generally will also be provided in the Shader doc, so there is not much of a need for them to have their own pages.

[...]

Issue #1903, opened 2018-11-03
No correspondence table for shader language uniform type and script class type!


D3ZAX:
We need a correspondence table for shader language uniform type and script class type to pass correct script class instance to shader uniform by  function set_shader_param() !

For example, it's Color from GDScript to mat in shader uniform type.

akien-mga:
Fixed by #2046.

Issue #1252, opened 2018-03-22
Shader docs: Visually separate Spatial and CanvasItem built-ins more clearly


JarLowrey:
https://godot.readthedocs.io/en/3.0/tutorials/shading/shading_language.html#fragment-built-ins

[Playing around with shaders for the first time](https://godotengine.org/qa/26127/simple-shader-not-working?show=26129#a26129). Attempting to get COLOR variable's value didn't go as expected, only setting it worked correctly.

bojidar-bg:
You have mistaken the docs for spatial shader with the ones for canvas item shader below.

Spatial (cut):
![image](https://user-images.githubusercontent.com/5276727/37823795-38b8f4d2-2e93-11e8-9d6d-eb30f39f41cb.png)

Canvas Item:
![image](https://user-images.githubusercontent.com/5276727/37823760-194b33c6-2e93-11e8-8536-43927dab6b0b.png)

Perhaps those docs should be cleared a bit by adding "Spatial" or "Canvas Item" to each of the headings, and maybe moving the discussion following some of the tables to separate pages.
Or maybe, moving all the tables to separate pages for spatial/canvas/particles shaders, and keep them as reference material only.

JarLowrey:
At the minimum I think adding to the heading would be good. Thanks!

akien-mga:
Fixed by #2046.

Issue #494, opened 2017-09-29
Revise/edit Shader documentation


cbscribe:
Adding this issue for discussion of revisions and/or additions to the new shader docs here (thanks @reduz ):
https://github.com/godotengine/godot-docs/tree/master/learning/features/shading

Also, now that we have some good documentation to start from, we need to add a shader intro/tutorial to the todo list.

NathanLovato:
Got an intro to shaders with Godot 3 series planned as part of the Kickstarter, but it won't come out before 2018 I think. Got other content to take care of in priority. But it's planned, it's been funded by the backers, so if no one else takes on the task there's at least some guaranteed content eventually.

ZX-WT:
tutorial about the visual shader editor should be added as well
https://godotengine.org/article/visual-shader-editor-back

akien-mga:
Fixed by #2046.

PR #2046, opened 2018-12-13
Overhauled shading reference

clayjohn:
As promised in #1921 I have overhauled the shading reference section. This splits the shading_language.rst doc into five sub documents:
- Shaders (introduces what shaders are in Godot and their structure from a high level)
- Shading language (mostly the same as the old shading language doc, but is limited to shading language stuff, removed "shader types in depth")
- Spatial shaders (contains information relevant to spatial shaders, description + built ins)
- CanvasItem shaders (Same as above, but from CanvasItems)
- Particle shaders (Same as above for Particles)

Overall, this is still in the language of a reference document. I have added a lot of clarification, and high-level description that was not present before. 

I'm hoping this will make shaders a little more accessible for the average Godot user. The reference docs are still pretty long, but I think it is a large improvement over the one mega document from before.

I know this is a huge update, but I think it is very much needed. Please let me know if there is anything I can change!

addresses: #494, #1252, #1903, #1921, and [another](https://github.com/godotengine/godot/issues/17905) from the godot repo


akien-mga:
Thanks!

PR #2054, opened 2018-12-14
Tweak inverse_kinematics

corrigentia:
Fix some grammar, punctuation, word order, word choice and GDScript style guide compliance in inverse_kinematics.

### DIFF
--- a/tutorials/3d/inverse_kinematics.rst
+++ b/tutorials/3d/inverse_kinematics.rst
@@ -27,7 +27,7 @@ IK problem.
 
 .. image:: img/inverse_kinematics.png
 
-For easy calculation in this chapter, we consider the target being a
+For easy calculation in this chapter, we assume the target is a
 child of Skeleton. If this is not the case for your setup, you can always
 reparent it in your script, as you will save on calculations if you
 do so.
--- a/tutorials/3d/inverse_kinematics.rst
+++ b/tutorials/3d/inverse_kinematics.rst
@@ -3,7 +3,7 @@
 Inverse kinematics
 ==================
 
-This tutorial is a follow-up of :ref:`doc_working_with_3d_skeletons`.
+This tutorial is a follow-up to :ref:`doc_working_with_3d_skeletons`.
 
 Previously, we were able to control the rotations of bones in order to manipulate
 where our arm was (forward kinematics). But what if we wanted to solve this problem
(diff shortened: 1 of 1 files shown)


tutorials/3d/inverse_kinematics.rst
```diff
@@ -17,18 +17,18 @@ move each joint in our arm to get to that target.
 Initial problem
 ~~~~~~~~~~~~~~~
 
-Talking in Godot terminology, the task we want to solve here is to position
+In Godot parlance, the task we want to solve here is to position
```
clayjohn: We should remove the "In Godot parlance", it doesn't make sense in context and "parlance" is not a very accessible word for many people. How about instead we just use: "We want to position the 2 angles..."

tutorials/3d/inverse_kinematics.rst
```diff
@@ -17,18 +17,18 @@ move each joint in our arm to get to that target.
...
 lowerarm bone is as close to the target point (which is set by the target Vector3)
 as possible using only rotations. This task is calculation-intensive and never
 resolved by analytical equation solving, as it is an under-constrained
-problem which means that there is more than one solution to an
+problem, which means that there is more than one solution to an
 IK problem.
 
 .. image:: img/inverse_kinematics.png
 
 For easy calculation in this chapter, we consider the target being a
```
clayjohn: I know you didn't touch this line yourself. But we should change it to, "we assume the target is a child of Skeleton".

tutorials/3d/inverse_kinematics.rst
corrigentia: > We should remove the "In Godot parlance", [...] "parlance" is not a very accessible word for many people. How about instead we just use: "We want to position the 2 angles..."

@clayjohn : Amended.

> "In Godot parlance", [...] doesn't make sense in context

I had rephrased the awkward wording intended to point out that that was not an everyday way of looking at such a task, but a specialized, technical, algorithmic manner of breaking down a problem into steps and procedures that can be carried out by a computer agent.

tutorials/3d/inverse_kinematics.rst
clayjohn: My bad, I seem to have commented below the wrong line. This change was intended for line 30. Which currently says "For easy calculation in this chapter, we consider the target being a child of Skeleton."

We should change it to "For easy calculation in this chapter, we assume the target is a child of Skeleton."

corrigentia: Much obliged, @clayjohn.

akien-mga:
Thanks!

PR #2066, opened 2018-12-19
Fix some grammar and punctuation in audio_buses

corrigentia:
Fix some grammar and punctuation in audio_buses.


akien-mga:
Thanks!

PR #2067, opened 2018-12-19
Fix some grammar and punctuation in audio_streams

corrigentia:
Fix some grammar and punctuation in audio_streams.


akien-mga:
Thanks!

PR #1614, opened 2018-07-25
Fixed typos in vertex displacement with shaders tutorial

clayjohn:
A user on Discord noticed the errors while trying to follow the tutorial.

PR #2071, opened 2018-12-21
Fixed a Bug in the Modifying Vertices with Shaders Tutorial's Example Code

nnmsIsaac:
from lines 192 to 202 there was a code snippet that used variables that were referenced before assignment/declaration. I've fixed them and updated the code sample to work.

### DIFF
--- a/tutorials/3d/vertex_displacement_with_shaders.rst
+++ b/tutorials/3d/vertex_displacement_with_shaders.rst
@@ -207,7 +207,7 @@ We can now use this noise function in place of ``cos`` and ``sin`` in the previo
 
 ::
 
-  float height = noise(VERTEX.xz * 4.0);
+  float height = fbm(VERTEX.xz * 4.0);
   VERTEX.y += height * 0.5;
 
 .. image:: img/vertex_displacement_noise1.png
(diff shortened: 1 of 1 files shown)


tutorials/3d/vertex_displacement_with_shaders.rst
```diff
@@ -190,11 +190,11 @@ we increase the frequency each level, decrease the amplitude, and calculate a ne
 ::
 
   float fbm(vec2 x) {
-    float height = 0.0;
-    float amplitude = 0.5;
-    float frequency = 3.0;
+    float h = 0.0;
+    float a = 0.5;
+    float p = 3.0;
```
LikeLakers2: I feel like it'd be better if we used the longer variable names "height", "amplitude" and "frequency", for the sake of readability.

nnmsIsaac:
My correction was not entirely correct, I'm fixing it now.

tutorials/3d/vertex_displacement_with_shaders.rst
nnmsIsaac: Yes, I've updated that as well. Please look at the new commit.

LikeLakers2: 👍

clayjohn:
This has already been fixed in master. If we are going to update the 3.0 branch we should do so by cherrypicking the changes made to master in order to keep things consistent. (For reference: #1614, #1629)

akien-mga:
I cherry-picked the relevant commits in the `3.0` branch, thanks!

PR #2073, opened 2018-12-21
Fix some punctuation and grammar in ray-casting

corrigentia:
Fix some punctuation and grammar in ray-casting.


akien-mga:
Thanks!

PR #2077, opened 2018-12-22
Improve Android export presets reg. passwords.

MelvinWM:
Background: I tried exporting a release APK reg. exporting to Android, but I was stuck for a while on why it did not work, and after some debugging I figured out that the passwords for the keystore and the key has to be the same, since the implementation for exporting to Android currently does not support passing both the keystore password and the key password. The warnings and errors did not make it clear that the passwords has to be the same, and I did not see any good handling reg. that in the engine (one could add a property for setting the key or keystore alias as an additional option, though I am uncertain whether that would be a good idea). This pull request updates the online documentation to make it clear that the passwords have to be the same.

See also https://github.com/godotengine/godot/blob/e3e5a1bb6e3ae3d667eb622003d4c7d0cc713cd0/platform/android/export/export.cpp#L1771 and https://docs.oracle.com/javase/6/docs/technotes/tools/windows/jarsigner.html, section "Keystore Aliases".

Also some other minor improvements and a screenshot.

The documentation changes are for the latest master-branch of Godot as of this writing.


getting_started/workflow/export/exporting_for_android.rst
```diff
@@ -70,17 +70,20 @@ keystore file; such file can be generated like this:
...
-Now fill the following forms in you Android Export Presets:
+Now fill in the following forms in your Android Export Presets:
+
+.. image:: img/editor-export-presets-android.png
 
 - Release: Enter the path to the keystore file you just generated.
-- Release User: Replace with what your key alias.
-- Release Password: Key password.
+- Release User: Replace with the key alias.
+- Release Password: Key password. Note that the keystore password and the key password currently has to be the same.
```
clayjohn: This should read "currently have to be the same".

PR #2089, opened 2019-01-02
Removed slash typo and added dot instead

GlaDos28:


akien-mga:
Files in `classes/` are generated from the XML class reference in https://github.com/godotengine/godot/tree/master/doc/classes, they should therefore be modified there only.

PR #2095, opened 2019-01-03
Fix punctuation in saving_games

corrigentia:
Fix punctuation in saving_games. Clarify possessive, not plural.

### DIFF
--- a/tutorials/io/saving_games.rst
+++ b/tutorials/io/saving_games.rst
@@ -15,18 +15,18 @@ scale as the game grows more complex.
 Identify persistent objects
 ---------------------------
 
-First we should identify what objects we want to keep between game
+Firstly, we should identify what objects we want to keep between game
 sessions and what information we want to keep from those objects. For
-this tutorial, we will use groups to mark and handle objects to be saved
+this tutorial, we will use groups to mark and handle objects to be saved,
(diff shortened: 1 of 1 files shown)


tutorials/io/saving_games.rst
```diff
@@ -112,17 +112,17 @@ like this:
 
 
 This gives us a dictionary with the style
-``{ "variable_name":that_variables_value }`` which will be useful when
+``{ "variable_name":that_variable_s_value }``, which will be useful when
```
LikeLakers2: I can see why you made this change, but the way it looks after your change seems weird to me. Maybe `that_variable_value` (remove the `s`)? Or maybe `variable_value` to match up with `variable_name`.

corrigentia: > I can see why you made this change, but the way it looks after your change seems weird to me. Maybe `that_variable_value` (remove the `s`)? Or maybe `variable_value` to match up with `variable_name`.

@LikeLakers2 : Additional option: ``value_of_variable``.

``variable_value`` almost sounds like "arbitrary/un-deterministic/stochastic/random value" to me, while ``that_variable_value`` merely qualifies which one you mean.

akien-mga:
Thanks!

PR #2100, opened 2019-01-04
Explain the need to set ANDROID_NDK_HOME env var

vnen:
When using custom modules, it's possible that the Gradle build fails if this is not set (or set incorrectly) and the cause it's not explicitly stated in the logs. So documenting this might save some trouble.

### DIFF
--- a/development/compiling/compiling_for_android.rst
+++ b/development/compiling/compiling_for_android.rst
@@ -56,9 +56,9 @@ with ``y``:
 
 Set the environment variable ``ANDROID_NDK_ROOT`` to point to the
 Android NDK. You also might need to set the variable ``ANDROID_NDK_HOME``
-to the same path, in particular if you are using custom Android modules,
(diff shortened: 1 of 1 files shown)


development/compiling/compiling_for_android.rst
```diff
@@ -55,7 +55,10 @@ with ``y``:
 
 
 Set the environment variable ``ANDROID_NDK_ROOT`` to point to the
-Android NDK.
+Android NDK. You also might need to set the variable ``ANDROID_NDK_HOME``
+to the same path, in particular if you are using custom Android modules,
```
clayjohn: Instead of "in particular" can you use "especially"?

akien-mga:
Thanks!

Would it make sense to unify this and have our own buildsystem rely on `ANDROID_NDK_HOME` instead of `ANDROID_NDK_ROOT`? Or is the latter also required by some stuff in the Android ecosystem?

PR #2102, opened 2019-01-04
Write docs for compiling the engine with script encryption key

Xrayez:
Retrieved from old godot fork wiki and adapted according to current source with recently merged PR:

godotengine/godot#24582

https://github.com/ScyDev/godot/wiki/Export-Encrypt-Scripts


akien-mga:
Thanks!

PR #2106, opened 2019-01-05
Fix grammar and punctuation in gui_skinning

corrigentia:
Fix grammar and punctuation in gui_skinning. 

Note: it's "[associate  **with**](https://idioms.thefreedictionary.com/associate+with)", not "associate _to_".

"_Replace_ old  **with**  new." / 
"_Substitute_ new  **for**  old."

Bug: (I think) the image in the "Creating a theme" section was wrong -- the same as the one for "Customizing a control", and confusingly so.
The provided font failed to work for me, even after making copies of it with recognized font extensions.

``mytheme.thm`` no longer matches any of the recognized theme extensions: *.tres, *.theme, *.res; 
the example image shows it saved as a ``.tres`` file.


akien-mga:
Thanks!

PR #2107, opened 2019-01-06
Fix grammar and punctuation in custom_gui_controls

corrigentia:
Fix grammar and punctuation in custom_gui_controls.


akien-mga:
Thanks!

PR #2108, opened 2019-01-06
Fix punctuation in size_and_anchors

corrigentia:
Fix punctuation in size_and_anchors. 
I presume the illustrations are stragglers from an older version of Godot, the look of which I like quite a bit more. 
At any rate, the current version of Godot doesn't present its options in the manner set forth by this page; there seem to have been a number of minor changes done and  ``Grow Direction``  is what it pertains to now.


akien-mga:
Thanks!

The content might need review for 3.x indeed, CC @groud.

PR #2111, opened 2019-01-06
Fix grammar and punctuation in viewports

corrigentia:
Fix grammar and punctuation in viewports.

Misleading information: (I think) the function signature for [Viewport.set_size_override()](https://godot.readthedocs.io/en/latest/classes/class_viewport.html#class-viewport-set-size-override) goes against the example provided.


akien-mga:
Thanks!

> Misleading information: (I think) the function signature for Viewport.set_size_override() goes against the example provided.

Indeed, fixed.

PR #2112, opened 2019-01-07
Use new logging method instead of the deprecated one

YeldhamDev:
`sphinx.util.logging` should be used in the upcoming Sphinx 2.0.0.

### DIFF
--- a/extensions/sphinx_tabs/tabs.py
+++ b/extensions/sphinx_tabs/tabs.py
@@ -7,6 +7,7 @@ import os
 from docutils.parsers.rst import Directive
 from docutils import nodes
 from pygments.lexers import get_all_lexers
+from sphinx.util import logging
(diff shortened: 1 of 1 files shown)


merumelu:
`sphinx_tabs` is a third party extension, we should likely update to a newer version instead of doing edits ourself.

YeldhamDev:
@merumelu I wasn't aware that it was a thirdparty script. Should I close this then?

akien-mga:
Indeed, we should sync with the latest [sphinx-tabs](https://github.com/djungelorm/sphinx-tabs) release.

This was seemingly fixed by djungelorm/sphinx-tabs@4f6cde044c177c41dfb8959b45a2ad789114a797 in v1.1.8.

akien-mga:
Done with 2a0710e8d.

PR #2113, opened 2019-01-08
Tweak using_viewport_as_texture

corrigentia:
Fix punctuation, grammar and spelling in using_viewport_as_texture. 
Add steps for which I've had to go looking elsewhere. 
Prevent ``error(3): Expected ';' after 'shader_type <type>'.``.
Attempt fix in "Coloring the planet" for 
``error(45): Inalid member for float expression: .x`` & 
``error(47): Inalid member for float expression: .x``.
Presuming ``n`` is the one defined by ``float n = noise(unit * 5.0);``, "The Little MLer" would call it nonsense to try to access the ``x`` component of a ``float`` with ``n.x``.

### DIFF
--- a/tutorials/viewports/using_viewport_as_texture.rst
+++ b/tutorials/viewports/using_viewport_as_texture.rst
@@ -7,7 +7,7 @@ Introduction
 ------------
 
 This tutorial will introduce you to using the :ref:`Viewport <class_Viewport>` as a
-texture that can be applied to 3D objects. In order to do so it will walk you through the process
+texture that can be applied to 3D objects. In order to do so, it will walk you through the process
 of making a procedural planet like the one below:
 
 .. image:: img/planet_example.png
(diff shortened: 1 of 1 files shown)


clayjohn:
The variable ``n`` used to be a ``vec4``. In PR #1929 I changed it to use a float and forgot to remove the corresponding ".x"s. Instead of using ``unit`` in its place, all references to ``n.x`` should be changed to ``n``.

akien-mga:
Thanks!

corrigentia:
> The variable `n` used to be a `vec4`. In PR #1929 I changed it to use a float and forgot to remove the corresponding ".x"s. Instead of using `unit` in its place, all references to `n.x` should be changed to `n`.
> 
> Other than that the changes look good! I like the addition of steps in parenthesis (e.g. "ColorRect > CanvasItem > Material > Material > `New ShaderMaterial`").

Cheers for the clarification, @clayjohn! :smile:

PR #2115, opened 2019-01-09
Fix punctuation in custom_postprocessing

corrigentia:
Fix punctuation in custom_postprocessing.
I surmise that the second ``col += texture(TEXTURE, SCREEN_UV + SCREEN_PIXEL_SIZE.xy).xyz;`` line was missing a ``-`` (minus sign) before I went through it. 
If so, I chalk it up to a case of copypasta.

### DIFF
--- a/tutorials/viewports/custom_postprocessing.rst
+++ b/tutorials/viewports/custom_postprocessing.rst
@@ -38,7 +38,7 @@ another :ref:`Control <class_Control>` node.
     ``ViewportContainer`` to render 3D objects in a 2D scene.
 
 For this demo, we will use a :ref:`Node2D <class_Node2D>` with a ``ViewportContainer`` and finally a
-``Viewport``. Your ``Scene`` tab should look like this:
+``Viewport``. Your  **Scene**  tab should look like this:
 
 .. image:: img/post_hierarchy1.png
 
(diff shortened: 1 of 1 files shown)


clayjohn:
Looks good to me! You are absolutely right about the minus sign. 

The one thing I would change is in line 41, i don't think "Scene" should be inline code. I think it should just have quotation marks. I try to reserve inline code for Node names and actual references to functions and variables.

tutorials/viewports/custom_postprocessing.rst
```diff
@@ -33,26 +33,26 @@ another :ref:`Control <class_Control>` node.
...
 
-    Rendering using a ``Viewport`` gives you control over the
+    Rendering using a ``Viewport`` gives you control over
     how the scene render, including the framerate, and you can use the
     ``ViewportContainer`` to render 3D objects in a 2D scene.
 
-For this demo we will use a :ref:`Node2D <class_Node2D>` with a ``ViewportContainer`` and finally a
-``Viewport``. Your Node tab should look like this:
+For this demo, we will use a :ref:`Node2D <class_Node2D>` with a ``ViewportContainer`` and finally a
+``Viewport``. Your  **Scene**  tab should look like this:
```
LikeLakers2: We don't need double-spaces here:
```suggestion
``Viewport``. Your **Scene** tab should look like this:
```

akien-mga:
Thanks!

PR #2119, opened 2019-01-10
Fix punctuation and grammar in shading_language

corrigentia:
Fix punctuation and grammar in shading_language.
I surmise ``vec3 b = a.aaa; // Also valid, creates a vec3 and fills it with "a".`` means that ``b`` will take the ``alpha`` (4th) component of ``vec4 a`` and fill up all 3 of its components with it.
Presume intent to highlight ``b[i] < a [...] ? 0.0 : 1.0``; attempt fix.


tutorials/shading/shading_reference/shading_language.rst
```diff
@@ -514,9 +514,9 @@ When vec_type (float), vec_int_type, vec_uint_type, vec_bool_type nomenclature i
 +-----------------------------------------------------------------------------------------------+------------------------------------------------+
 | vec_type **mix** ( vec_type a, vec_type b, vec_bool_type c )                                  | Linear Interpolate (Bool-Vector Selection)     |
 +-----------------------------------------------------------------------------------------------+------------------------------------------------+
-| vec_type **step** ( vec_type a, vec_type b )                                                  | \` b[i] < a[i] ? 0.0 : 1.0 \`                  |
+| vec_type **step** ( vec_type a, vec_type b )                                                  | \`\`b[i] < a[i] ? 0.0 : 1.0\`\`            |
```
clayjohn: This should be using two back ticks to make this inline code, I don't know why it was escaped initially. Using the backslashes makes it display the back ticks, which is not the desired look.

tutorials/shading/shading_reference/shading_language.rst
```diff
@@ -259,10 +259,10 @@ It is possible to define functions in a Godot shader. They use the following syn
     }
 
 
-You can only use functions that are defined above the function you are calling 
-them from.
+You can only use functions that have been defined above (higher in the editor) the function from which you are calling 
+them.
 
-Function argument can have special qualifiers:
+The function argument can have special qualifiers:
```
clayjohn: This should probably read "Function arguments ..." because we are mentioning functions in general, not a specific function.

akien-mga:
Thanks!

PR #2124, opened 2019-01-10
Fix grammar and punctuation in particle_shader

corrigentia:
Fix grammar and punctuation in particle_shader.


clayjohn:
Looks good!

PR #2130, opened 2019-01-12
Clean instancing page recap

jamesreiati:
Recaps usually highlight what was learned in the previous section. Unity wasn't mentioned in the previous section. The ability to mass edit nodes was mentioned, but wasn't included in the recap. Also, as of Unity 2018.3, nested prefabs are a core concept. I figured removing this line would help remove "informational dependencies" that need to be updated, and keep the section focused.


clayjohn:
I agree, this is a good change.

NathanLovato:
Thanks :)

PR #2134, opened 2019-01-14
Update "Dodge the Creeps" for 3.1

### LINKED ISSUE #1240: Your first game/dodge the creeps: Main script setup needs to be expanded
Here:
http://docs.godotengine.org/en/3.0/getting_started/step_by_step/your_first_game.html

In the section "Main script", the tutorial goes over what should be done really fast:

*Next, click on the Player and connect the hit signal. We want to make a new function named game_over, which will handle what needs to happen when a game ends. Type “game_over” in the “Method In Node” box at the bottom of the “Connecting Signal” window. Add the following code, as well as a new_game function to set everything up for a new game: (...)*

This doesn't explain where `the following code` should be added, the player or the main script. It should be added to `Main`, but that can only be inferred from the section title and the prev. paragraph. This should be clarified, and maybe the signal configuration should be expanded a little and be explained more in-depth.

### LINKED ISSUE #1541: Your First Game/HUD: Confusing section with inaccurate instructions
Very new to Godot, had a lot of trouble understanding this section.

There are references to anchors in this section, but in Godot v3.0.4 the button is labelled "Layout". 
There is then a list of settings, some of which I already know are found in the inspector from earlier in the tutorial (Text, for example). However, 'Layout' is not found there. This may have been a little clearer if the layout button had not been called the 'anchor' button earlier, but I think the list makes the settings seem like they would be found in the same place.

I also found that setting the margins to the listed values did not put things in the same places as shown in the image. I had to do a bit of experimenting to find values which worked.

In other parts of the tutorial, the "Notes" box would come after introducing a new concept, but here it seems to come before which left me feeling like I'd missed something.

I also found that the size of the text did not match the image at all, and spent some time trying to figure out how to resize it. the tutorial actually covers this later, but I felt the image led me to believe that is what I should have nearer that point in the tutorial. Especially so considering the red circle indicating the layout button, though I think I ignored the red circle as the image was presenting a final product rather than what I had at the time.

Further down I am given `$MessageLabel.text = "Dodge the\nCreeps!"` in the `show_game_over()` function, where the text value given earlier did not contain a line break. (I used word-wrap in my attempts to match the provided image). I think some consistency may be needed.

### LINKED ISSUE #1714: Particles section of "Your first game" needs more explanation
I don't seem to get to be able to set the same settings as in the tutorial and with no further explanation it's really hard.

- Visibility rect is by default set to AABB and no instruction on how to set it to Rect3
- Scale curve is by default <null> and no instruction how to set
- The gradient has offset which is set to FloatArray[2] with no instruction how to set to 0
- Same with color that is to be set to 0
Overall no real explanation how any of it works.

### LINKED ISSUE #1730: "Your first game" is way too complicated an example
The "your first game" example : http://docs.godotengine.org/en/3.0/getting_started/step_by_step/your_first_game.html
As a tour of the features of Godot is good, but as a "first game" example it's WAYYY too complicated.

As a developer I was able to track along until the documentation started to make assumptions about hooking up events, and passing flags around, with multiple scenes. As a first game that's super complicated for a first time Godot user.  This should be like game #3 not game #1 .

Also, as a first game there aught to be a few "checkpoints" where the tutorial asks the player to run the game to see if things are working.  I got through all the tutorial and kicked off the game only to have a blank screen -- which of course meant I missed something, but if there were a few checkpoints in the documentation like "try things out now" it would be less of a big task to go back and try to debug everything. This also is another example of how this "Your first game" is too complicated for a first time Godot user. 

Having a player object move, then enemies, then layer in the paths, multiple scenes, etc. would be a better way to ramp up the difficulty without frustrating the user.

### LINKED ISSUE #1996: Unclear where timer signals are getting connected to
https://docs.godotengine.org/en/3.0/getting_started/step_by_step/your_first_game.html#main-script

> Now connect the timeout() signal of each of the Timer nodes.

> In _on_MobTimer_timeout() we will create a mob instance...

It's not perfectly clear where these signals are getting connected to. Is it the Main node? That would make sense given that the section title is "Main Script", but the choice of node is not explicit.

### LINKED ISSUE #1997: Tutorial doesn't say to center or wrap text
https://docs.godotengine.org/en/3.0/getting_started/step_by_step/your_first_game.html#hud

When following the instructions, I noticed that the "Dodge the Creeps!" text floated off of the side of the page and that text was not horizontally centered for that and the other label. I was able to figure out how to fix that in the inspector, but it might be useful to include those instructions in the docs.

### LINKED ISSUE #2000: Missing some instructions for particles setup in tutorial
https://docs.godotengine.org/en/3.0/getting_started/step_by_step/your_first_game.html#particles

The particles demo in the tutorial could use the following information:

* Set the color on the left to be *white*, with 50% alpha
* Setting the scale to 0.5
* How to edit the scale curve

cbscribe:
Updated "Dodge the Creeps", including changes for 3.1:

* New screenshots where necessary
* Update `export` vars to use default values (a common confusion point for new users)
* Remove particles section - this is complex enough to deserve its own document
* Unlike 3.0, in 3.1 setting `CollisionShape2D.disabled = true` generates an error, so this was changed to use `call_deferred()`
* Simplified HUD UI configuration
* Add keyboard event shortcut to Button (now that this works!)

Still TODO: add a couple more test checkpoints to verify everything works along the way.

Fixes #1541 
Closes #1714 
Closes #1730 
Fixes #1996 
Closes #2000 
Closes #1240 
Fixes #1997


NathanLovato:
Great work, as usual 🙂

PR #2136, opened 2019-01-15
Added example flags for visual studio code

Kronkleberry:
Added some example flags for VSC for other users to use without having to dig around on the internet, since it's suggested elsewhere in the docs to use for C#.


akien-mga:
Thanks!

PR #2138, opened 2019-01-15
Fix punctuation and spelling in http_request_class

corrigentia:
Fix punctuation and spelling in http_request_class.


akien-mga:
Thanks!

PR #2144, opened 2019-01-16
Fix punctuation and grammar in vr_primer

corrigentia:
Fix punctuation and grammar in vr_primer.
"then" < When?
"than" < Compared to what?

"acceptable" (adjective) = the nature of an object/thing;
"acceptably" (adverb)     = the nature of an action/process.


akien-mga:
Thanks!

PR #2145, opened 2019-01-16
Tweak vr_starter_tutorial

corrigentia:
Fix punctuation, spelling, grammar and GDScript style guide compliance in vr_starter_tutorial. Remove duplication.
"a" becomes "an" before vowel sound.
Do **NOT** use ["different" after "various"](https://www.thefreedictionary.com/various)!


akien-mga:
Thanks!

PR #2146, opened 2019-01-16
Remove duplicate instructions regarding CenterContainer

JJTriesHisBest:
Change: 
* Remove the first instruction to drag the Characters/Illustration node into the `CenterContainer`
Reason:
* This is repeated later in the paragraph that explains centering UI elements

Change:
* Remove the second instruction to create the `CenterContainer`
Reason:
* The user is instructed to create the `CenterContainer` at the same time they create the `VBoxContainer`. (This makes sense as it is inkeeping with the opening mantra of arranging parent UI elements first)


akien-mga:
CC @NathanLovato

NathanLovato:
Looking good, thanks!

PR #2158, opened 2019-01-25
add note about "Use Snap" at line 627

homchom:
I had this problem when reading the tutorial and I think this clarification would help others too.


cbscribe:
Oops I didn't notice this pr was against 3.0 rather than master. Please make any future changes there. I am working on a separate update to this doc, so I'll wrap this change into that.

PR #2160, opened 2019-01-26
Clarified use of WORLD_MATRIX in CanvasItem Shaders

clayjohn:
This is really a temporary fix, ``WORLD_MATRIX`` should actually be named ``MODELVIEW_MATRIX`` or just ``VIEW_MATRIX`` because it takes vertices in local coordinates and transforms them into view space. Also ``PROJECTION_MATRIX`` goes from view space into clip space. Not from world into view. 

Right now there is no way to get world coordinates without passing in a custom uniform. So it should be clear in the docs.

### DIFF
--- a/tutorials/shading/shading_reference/canvas_item_shader.rst
+++ b/tutorials/shading/shading_reference/canvas_item_shader.rst
@@ -57,6 +57,29 @@ it manually with the following code:
         VERTEX = (EXTRA_MATRIX * (WORLD_MATRIX * vec4(VERTEX, 0.0, 1.0))).xy;
     }
 
+.. note:: ``WORLD_MATRIX`` is actually a modelview matrix. It takes input in local space and transforms it
+          into view space.
+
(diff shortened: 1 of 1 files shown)


tutorials/shading/shading_reference/canvas_item_shader.rst
```diff
@@ -57,6 +57,29 @@ it manually with the following code:
...
     }
 
+.. note:: ``WORLD_MATRIX`` is actually a modelview matrix. It takes input in local space and transforms it
+          into view space.
+
+In order to get the world space coordinates of a vertex, you have to pass in a custom uniform like so:
+
+::
+  
+  material.set_shader_param("global_transform", get_global_transform())
```
akien-mga: Please use four spaces for indentation.

clayjohn: My bad! Fixed now.

akien-mga:
Thanks!

PR #2168, opened 2019-01-31
Fixed a typo in a code example

Bassie-c:
I added a semi-colon to an example, without it the code would crash. Probably a typo.


Bassie-c:
Attempt 2 at adding the semicolon...

NathanLovato:
Thanks :)

PR #2170, opened 2019-02-02
Fixed inout variables in spatial shader docs

clayjohn:
Many variables were marked as ``out`` when in fact they are ``inout``


akien-mga:
Thanks!

PR #2172, opened 2019-02-03
fix typo

AdityaGupta1:
fixed a typo in WebSocketClient doc


vnen:
This is a generated file, should not be changed here. See the file jheader:

```
.. DO NOT EDIT THIS FILE, but the WebSocketClient.xml source instead.
```

AdityaGupta1:
sorry about that, will fix it there instead

PR #2175, opened 2019-02-05
Add a "best practices" section

willnationsdev:
This introduces a new series of Godot best practices to the documentation.

Please let me know what you think. I'd appreciate any feedback you have on whether my suggestions are truly the best practices (they are to the best of my understanding). I also may have made errors in C# implementations (been trying to work on that). I've proofread everything in the Hemingway Editor, but feel free to suggest additional changes.

### DIFF
--- a/getting_started/workflow/best_practices/scene_organization.rst
+++ b/getting_started/workflow/best_practices/scene_organization.rst
@@ -1,0 +1,391 @@
+.. _doc_scene_organization:
+
+Scene organization
+==================
+
+This article covers topics related to the effective organization of
+scene content. Which nodes should one use? Where should one place them?
+How should they interact?
+
+How to build relationships effectively
+--------------------------------------
+
+When Godot users begin crafting their own scenes, they often run into the
+following problem:
+
+They create their first scene and fill it with content before the creeping
+sense that they need to split it up into re-usable pieces haunts them. They
+save branches of their scene into their own scene. However, they then notice
+that the hard references they were able to rely on before are no longer
+possible. Re-using the scene in multiple places creates issues because the
+node paths do not find their targets. Signal connections established in the
+editor break.
+
+To fix these problems, one must instantiate the sub-scenes without them
+requiring details about their environment. One needs to be able to trust
+that the sub-scene will create itself without being picky about how one uses
+it.
+
+One of the biggest things to consider in OOP is maintaining
+focused, singular-purpose classes with
+`loose coupling <https://en.wikipedia.org/wiki/Loose_coupling>`_
+to other parts of the codebase. This keeps the size of objects small (for
+maintainability) and improves their reusability so that re-writing completed
+logic is unnecessary.
+
+These OOP best practices have *several* ramifications for the best practices
+in scene structure and script usage.
+
+**If at all possible, one should design scenes to have no dependencies.**
+That is, one should create scenes that keep everything they need within
+themselves.
+
+If a scene must interact with an external context, experienced developers
+recommend the use of 
+`Dependency Injection <https://en.wikipedia.org/wiki/Dependency_injection>`_.
+This technique involves having a high-level API provide the dependencies of the
+low-level API. Why do this? Because classes which rely on their external
+environment can inadvertantly trigger bugs and unexpected behavior.
+
+To do this, one must expose data and then rely on a parent context to
+initialize it:
+
+1. Connect to a signal. Extremely safe, but should use only to "respond" to
+   behavior, not start it. Note that signal names are usually past-tense verbs
+   like "entered", "skill_activated", or "item_collected".
+
+   ..tabs::
+     ..code-tab:: gdscript GDScript
+     
+       # Parent
+       $Child.connect("signal_name", object_with_method, "method_on_the_object")
+
+       # Child
+       emit_signal("signal_name") # Triggers parent-defined behavior
+
+     ..code-tab:: csharp
+
+       // Parent
+       GetNode("Child").Connect("SignalName", ObjectWithMethod, "MethodOnTheObject");
+
+       // Child
+       EmitSignal("SignalName"); // Triggers parent-defined behavior
+
+2. Call a method. Used to start behavior.
+       
+   ..tabs::
+     ..code-tab:: gdscript GDScript
+     
+       # Parent
+       $Child.method_name = "do"
+
+       # Child, assuming it has String property 'method_name' and method 'do'
+       call(method_name) # Call parent-defined method (which child must own)
+
+     ..code-tab:: csharp
+
+       // Parent
+       GetNode("Child").Set("MethodName", "Do");
+
+       // Child
+       Call(MethodName); // Call parent-defined method (which child must own)
+
+3. Initialize a :ref:`FuncRef <class_FuncRef>` property. Safer than a method
+   as ownership of the method is unnecessary. Used to start behavior.
+       
+   ..tabs::
+     ..code-tab:: gdscript GDScript
+     
+       # Parent
+       $Child.func_property = funcref(object_with_method, "method_on_the_object")
+
+       # Child
+       func_property.call_func() # Call parent-defined method (can come from anywhere)
+
+     ..code-tab:: csharp
+
+       // Parent
+       GetNode("Child").Set("FuncProperty", GD.FuncRef(ObjectWithMethod, "MethodOnTheObject"));
+
+       // Child
+       FuncProperty.CallFunc(); // Call parent-defined method (can come from anywhere)
+
+4. Initialize a Node or other Object reference.
+       
+   ..tabs::
+     ..code-tab:: gdscript GDScript
+     
+       # Parent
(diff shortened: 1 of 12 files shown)


NathanLovato:
Amazing work! Regarding editing and proofing, do you mind if I merge and tweak articles one by one directly? It's much better than last time. I'll gladly compile a few review notes with the things I tweaked and send you the commits, it's just that it'd take a lot more time to leave you individual comments for everything. 🙂

willnationsdev:
@NathanLovato Sure! No problem. :-)

NathanLovato:
In short, there are things like...

## The phrasing could sometimes be more accessible

You still have room to make your content more accessible without changing the meaning or the info you're sharing.

I guess you're an engineer? Or at least you have some computer science background. The hemingway editor will suggest that you shorten sentences, avoid the passive voice, or remove adverbs, but it doesn't help you to make the content accessible to non-native speakers or to change the style based on the document's intent.

## Too many lists

You use lists way too often.

There are 6 heading levels to organize your content hierarchically. To me, 3-levels nested lists make the content feel colder and more technical than it needs to look. You don't get the hyperlinks or checkpoints that headings give you as well.

Sometimes you use nested bullets as notes. Sometimes you use bulleted lists to suggest steps, but it doesn't work as well as individual paragraphs.

NathanLovato:
Then, may I ask you to split your PRs moving forward, but also to put your content up for review as you create it? These are 10 separate articles that could all use their own PR for review and discussion. The expression "best practices" is also up for debate: the articles are a lot more nuanced than the title suggests, and the "best practices" depend a lot on the scope of the project and the problem you're working on. Godot being pretty flexible.

willnationsdev:
> The phrasing could sometimes be more accessible
> I guess you're an engineer? Or at least you have some computer science background.

Yes, I tend to write articles high in detail. Simplifying content is one of my challenges as a writer.

> You use lists way too often.

I believe the last time I submitted content, I was told I didn't use them enough, so I suppose I went too far the other direction this time around.

> Then, may I ask you to split your PRs moving forward, but also to put your content up for review as you create it? These are 10 separate articles that could all use their own PR for review and discussion.

Sure. In hindsight, getting feedback along the way probably would've improved the quality of subsequent articles in the series anyway. XP

> The expression "best practices" is also up for debate: the articles are a lot more nuanced than the title suggests, and the "best practices" depend a lot on the scope of the project and the problem you're working on. Godot being pretty flexible.

I suppose that's true. I tried to isolate topics that I felt in general had more than 1 solution, could apply differently to different situations, and gave recommendations on what each approach to the problems made easier / why one might wish to choose one approach over another (sometimes helping people to determine for themselves what the "best practices" for their project would be, sometimes suggesting which approach is generally the best for Godot specifically).

NathanLovato:
I think it's valuable content overall, and worth merging and refining from there on. Just @akien-mga, could you confirm that I can go ahead, merge, and edit the articles straight? I will commit straight to proof/improve the form without changing the meaning, but open PR to suggest any deeper changes if that's okay.

cbscribe:
I'm also not sure how some of these fit under "workflow" as opposed to the existing subcategories. Isn't "What are Godot classes?" more of a "Getting Started" topic?

"Logic preferences" really didn't tell me what it was about. These are good topics though, they should be more easily found, like the "Scenes vs. Scripts" section.

willnationsdev:
Well, feel free to redistribute them as you see fit if you can find better places for them all.

starry-abyss:
Good read, learned about some new tricks!

>  Use ``_process`` [...] Recurring logic checks and data caching, should execute here.

This part confuses me. Say, I want logic checks for enemy AI - what is the point of making them more often than physics step? :thinking:

> ``_unhandled_input()`` : NOTIICATION_UNHANDLED_INPUT

Missed the letter 'F' in "notification".

> 2. If one has the time/resources on a medium/large game, create a library or
   plugin that can code the management of nodes and resources. If refined
   over time, so as to improve usability and stability, then it could evolve
   into a reliable tool across projects.

I think one shouldn't generally aim to create a "tool across projects" when developing a game, but only 
tool for current project's requirements (improving usability makes sense, esp. if it's going to be used by a teammate, but starting as library/plugin straight away?..). Also, "plugin that can code" - I would buy this one :) 

> If you would like to give a name to a scene, then you can still sorta do this in 3.1 by declaring a script class and giving it a scene as a constant.

"Sorta" is sorta too informal IMHO. :P

YeldhamDev:
Did a quick read, and it's excellent! The docs REALLY needed something like this, it will clarify a lot of things some people who are starting with the engine have questions about.

But I see you're using static typing with a extra space between the variable name and the colon (`var_name : type`). We had a discussion about this some time ago (https://github.com/godotengine/godot/issues/21057), and the "verdict" was more inclined to go without it. It also how the engine itself writes then out when connecting signals.

willnationsdev:
@YeldhamDev Oh, interesting about the spacing. I usually use the Pythonic technique, but I saw some of @NathanLovato's OpenRPG code that was doing the extra space at the time and I just assumed that was the new standardized format. I'll have to make the proper adjustments.

getting_started/workflow/best_practices/what_are_godot_classes.rst
```diff
@@ -0,0 +1,137 @@
...
+    # main.gd
+    extends Node
+    
+    var child # define a variable to store a reference to the child
+
+    func _init():
+        child = Node.new() # Construct the child.
+        child.name = "Child" # Change its name.
+        child.script = preload("child.gd") # Give it custom features.
+        child.owner = self # Serialize this node if self is saved.
```
willnationsdev: I'm not sure if this should really be here since, if one serialized the node with a child, then instantiated the scene, they'd re-execute the _init of the script and add *another* child in addition to deserializing the previously serialized child. Right?

getting_started/workflow/best_practices/scene_organization.rst
```diff
@@ -0,0 +1,391 @@
...
+       # Child
+       emit_signal("signal_name") # Triggers parent-defined behavior
+
+     ..code-tab:: csharp
+
+       // Parent
+       GetNode("Child").Connect("SignalName", ObjectWithMethod, "MethodOnTheObject");
+
+       // Child
+       EmitSignal("SignalName") // Triggers parent-defined behavior
```
willnationsdev: semi-colon

getting_started/workflow/best_practices/scene_organization.rst
```diff
@@ -0,0 +1,391 @@
...
+       Call(MethodName); // Call parent-defined method (which child must own)
+
+3. Initialize a :ref:`FuncRef <class_FuncRef>` property. Safer than a method
+   as ownership of the method is unnecessary. Used to start behavior.
+       
+   ..tabs::
+     ..code-tab:: gdscript GDScript
+     
+       # Parent
+       $Child.func_property = funcref(object_with_method, "MethodOnTheObject")
```
willnationsdev: "MethodOnTheObject" -> "method_on_the_object"

getting_started/workflow/best_practices/scenes_versus_scripts.rst
```diff
@@ -0,0 +1,181 @@
...
+            public const PackedScene MySceneScn= ResourceLoader.load("MyScene.tscn") as PackedScene;
+            public Node ANode;
+            public Node MyNode;
+            public Node MyScene;
+            public Node MyInheritedScene;
+
+            public Game()
+            {
+                ANode = new Node();
+                MyNode = new MyNode(); // Same syntax
```
willnationsdev: Now that I think about it, would this even work? Is the MyNode name exposed to other C# scripts just because they are in the same project? And does `MyNodeScr.new()` do the same thing that `new MyNode()` would do? I'm not good with Godot C#.

willnationsdev:
In the scenes vs. scripts article, there's a C# example that has a `const Script MyNodeScr` and a `public Node MyNode = new MyNode();` I'm starting to think there's an error there. Do C# classes have access to names from other C# classes just because they are in the same project? And would doing `MyNodeScr.new()` accomplish the same thing as `new MyNode()`?

YeldhamDev:
@willnationsdev For inferred types, `var name := value` is correct.

willnationsdev:
@starry-abyss I have updated each of the sections you mentioned. As for the use of "logic checks and data caches" in `_process`, I amended the statement to qualify the needs based on the frequency at one needs the data to update. Let me know if you notice anything else.

bojidar-bg:
Gave the best practices list a cursory check, looking good!

I think they might have a section about modularizing and reusing code, probably somewhere around the section about decoupling scenes. :smiley:

NathanLovato:
Okay let's merge :)

asheraryam:
@bojidar-bg I think that's ok, I imagine the best case scenario would be to have the intro-level notes be in the introductory material while the more in-depth patterns would be in the best practice section.

So the original article would say something like, "For more notes see the [Best Practices] section."

PR #2176, opened 2019-02-06
Fix unexpected indentation in "Exporting for web" page

YeldhamDev:


eska014:
The `tip` admonition should be indented into the `attention` block level so it's contained within

YeldhamDev:
Done.

eska014:
Thanks

PR #2184, opened 2019-02-10
Added tutorial on animating thousands of objects

### LINKED ISSUE #2165: Animating thousands of objects
While working on another tutorial I remember [a talk](https://www.gdcvault.com/play/1024409/Creating-the-Art-of-ABZU) from one of the art guys on ABZU about animating thousands of fish in ABZU using mesh instancing and vertex animations. I thought this might be simple in Godot, and turns out it is! With relatively few lines of code you can render and animate thousands of fish. 

This would make a good tutorial to put in the "creating content" section. 

I am going to write it in 2 parts. First will be combining vertex animation and mesh instancing. The second will be extending the mesh to work in particles so they can take advantage of the GPU to move the fish around as well.

clayjohn:
This tutorial walks users through how to:

1. Animate Meshes using vertex shaders
2. Use MultiMeshes through GDScript
3. Acces per-instance information from MultiMesh in shaders 
4. Use Particles to control movement and animation of objects using Shaders

Note: It makes heavy use of shaders, so it may be better off in the shader section, but it also could be good in the creating 3d content section. Please let me know if anything thinks there is a better location for this.

Note2: I have used 9 gifs. 8 of them are ~600kb, but one is ~6mb. I compressed them as much as I could. If we need to be more careful about space, we can delete the ~6mb gif.

Also, as always, I am very open to comments and critiques!!

Resolves #2165


akien-mga:
Thanks!

PR #2186, opened 2019-02-12
Remove references to Performance singleton.

### LINKED ISSUE #2185: The best_practices/godot_notifications should not recommend using the Performance singleton for getting delta time.
The `best_practices/godot_notifications` page mentions

>One can check for input actions within the input callbacks just the same. If one wants to use delta time, one can fetch it from the Performance singleton as needed.

However the Performance singleton page mentions

>Many of these monitors are not updated in real-time, so there may be a short delay between changes.

As I understand it the Performance singleton is solely meant for monitoring performance, and should not be used for gameplay calculations. Nodes have two functions for getting delta and physics delta time, which should be used instead, `node.get_process_delta_time()` and `node.get_physics_process_delta_time()`

willnationsdev:
Closes #2185.


akien-mga:
Thanks!

PR #2187, opened 2019-02-13
Clarified behavior of get_cell methods for TileMap class.

MidZik:
Clarified behavior of `get_cell` methods for TileMap class. If no tile exists, it will return INVALID_CELL.


cbscribe:
As noted at the top of the file:

```
.. DO NOT EDIT THIS FILE, but the TileMap.xml source instead.
```

Class ref docs must be changed in the main godot repository so that they are propagated to all sources, not just the web docs. See https://docs.godotengine.org/en/latest/community/contributing/updating_the_class_reference.html for details.

MidZik:
Oops, skimmed over that part. I'll submit it to the correct place.

PR #2188, opened 2019-02-14
Fix UI code a life bar documentation

nonameentername:
In order for the following section to still be valid:

```
Play the game to see the bar animate smoothly. But the text displays
decimal number and looks like a mess
```

The variable should be set as a float instead of an int.


akien-mga:
Thanks!

PR #2189, opened 2019-02-14
Fix for your first game documentation

nonameentername:
Godot no longer displays `<null>` text and instead displays `[empty]`


cbscribe:
Thanks!

PR #2193, opened 2019-02-16
typo nome -> name

wbulot:


akien-mga:
As mentioned in the README and each file's header, the files in `classes/` are generated from the class reference XML in https://github.com/godotengine/godot, so that's where they should be edited. Any change here will be overridden next time they're synced.

I fixed this typo upstream in https://github.com/godotengine/godot/commit/1ffd1bc8f3e54e5e40d0543a4b6304221b97d13f.

PR #2196, opened 2019-02-18
Fixed the C# example code in best_practices/logic_preferences

MightyPrinny:
The old code didn't compile.


willnationsdev:
Looks good @akien-mga.

willnationsdev:
Although, the line 75 where the const is removed from Building, I wonder if there is another way to rewrite that. I can't see the context well on my phone, but I recall wanting it to be some sort of constant in the example. Whether that means the code or tutorial text needs to change, idk. Someone who knows the C# syntax better than me (MightyPrinny?) may need to weigh in though.

MightyPrinny:
we can use readonly instead of const, in C# const can only be used for values known at compile time

MightyPrinny:
I updated the code, now it shows how to load a resource to a readonly field

akien-mga:
Thanks!

PR #2200, opened 2019-02-19
Updated c_sharp_basics.rst

MightyPrinny:
Added some extra information about setting up C# for Godot in Windows.


StraToN:
Can you squash these 2 commits into one? Thanks in advance ;)

akien-mga:
.NET Framework 4.5 is actually a requirement for all platforms, so it should probably be mentioned further up and not specifically for Windows.

CC @neikeq

neikeq:
Ideally that dependency would not be needed since we are using Mono's BCL and not .NET Framework's. This is a problem with MSBuild we will have to tackle in the future. If you look at the MSBuild output, it passes the .NET Framework's assemblies to CSC instead of Mono's. It seems to be a Windows-only issue, but I wouldn't be surprised if it could happen on other platforms as well under certain circumstances with .NET Core.

MightyPrinny:
Ok I'll open a new one with one commit

PR #2201, opened 2019-02-19
Fix and improve C# API differences page

neikeq:
Made some fixes, re-structured some sections and added new useful details.


neikeq:
I don't write docs often, so let me know if there's something wrong with my writing.

akien-mga:
Did a couple proofreading edits, looks great overall, thanks!

PR #2206, opened 2019-02-21
Corrected Use Snap shortcut key

godofgrunts:
`S` is the shortcut for Scale Mode. `Shift + S` is the shortcut for Snap Mode


akien-mga:
Thanks!

PR #2207, opened 2019-02-21
Update broken links

hporro:
There is a reference to core/dvector.h file insead of core/pool_vector.h, and a reference to core/string_db.h instead of core/string_name.h


mhilbrunner:
Thanks for this!

Issue #2205, opened 2019-02-21
[TRACKER] GLES2 and GLES3 Differences


clayjohn:
We should track the differences between GLES2 and GLES3 and ideally have something in the docs prior to the release of 3.1 for users to turn to. I have added the difference I can think of off the top of my head, but I am sure that there are more so please add to this issue so we can put a document together soon.

- [x] No Particles in GLES2 (Use CPUParticles instead)
- [x] No MipMaps on ``SCREEN_TEXTURE`` in GLES2. See https://github.com/godotengine/godot/pull/26044
- [x] Color space is different between GLES2 and GLES3. Document usage of ``OUTPUT_IS_SRGB``. See https://github.com/godotengine/godot/issues/25142
- [x] Reduced SpatialMaterial features in GLES2 (Refraction, subsurf, anisotropy, clearcoat, depth)
- [x] Reduced Environment features in GLES2 (auto_exposure, tonemap, ss reflections, ssao, dof blur, glow, adjustment) (only has sky, ambient, and fog)

- [x] GIProbe does not work in GLES2, using lightmaps (BakedLightmap) is recommended.

- [x] Contact shadows also do not work in GLES2

- [x] Reading from depth (DEPTH_TEXTURE) in the shader, does not work on every hardware. On old mobile hardware this is most likely not supported.

- [x] Performance scales poorly in GLES2 with several lights, as each is processed in a separate render pass (in opposition to GLES3 which is all done in a single pass).

- [x] On mobile, GLES2 requires ETC texture compression, while GLES3 requires ETC2. Showing how to enable ETC compression in the project settings if exporting to mobile would be nice.

- [x] Blend Shapes do not work on GLES2 (could be made to work eventually, but it's a lot of work and wont be fast).

- [x] GLES2 besides being linear color, is not HDR capable. This could be noted probably.

- [x] Some GLSL functions are only in GLESS3 (TextureSize for example). Should be documented in the shading doc and mentioned here

StraToN:
Is this list exhaustive ?

clayjohn:
Not necessarily. The goal is to have an exhaustive list, but i may have missed something. If you have anything that should be added, i would be happy to do so.

StraToN:
Unfortunately, I don't :'( That's why I asked. 
If @godotengine/documentation team is ok to keep this as a start, it can indeed be useful to users who wonder what actual differences lie between the 2 versions, so much important that we need to keep both of them in Godot :)

[...]

PR #2209, opened 2019-02-23
Add tutorial on gles2 and gles3 differences

clayjohn:
This relates to #2205 and https://github.com/godotengine/godot/issues/25142 but it closes neither.

### DIFF
--- a/tutorials/misc/gles2_gles3_differences.rst
+++ b/tutorials/misc/gles2_gles3_differences.rst
@@ -6,7 +6,7 @@ Differences between GLES2 and GLES3
 This page documents the differences between GLES2 and GLES3 that are by design and are not the result
 of bugs. There may be differences that are unintentional, but they should be reported as bugs.
 
-.. note:: "GLES2" and "GLES3" are the named used in Godot for the two rendering backends.
+.. note:: "GLES2" and "GLES3" are the names used in Godot for the two OpenGL-based rendering backends.
           In terms of graphics APIs, the GLES2 backend maps to OpenGL 2.1 on desktop, OpenGL ES 2.0 on
           mobile and WebGL 1.0 on the web. The GLES3 backend maps to OpenGL 3.3 on desktop, OpenGL ES
           3.0 on mobile and WebGL 2.0 on the web.
--- a/tutorials/misc/gles2_gles3_differences.rst
+++ b/tutorials/misc/gles2_gles3_differences.rst
@@ -1,0 +1,79 @@
+.. _doc_gles2_gles3_differences:
+
+Differences between GLES2 and GLES3
+===================================
+
+This page documents the differences between GLES2 and GLES3 that are by design and are not the result
+of bugs. There may be differences that are unintentional, but they should be reported as bugs.
+
+Particles
+---------
+
+GLES2 cannot use the :ref:`Particles <class_Particles>` or :ref:`Particles2D <class_Particles2D>` nodes
+as they require advanced GPU features. Instead use CPUParticles, which provides a similar interface to
+a :ref:`ParticlesMaterial <class_ParticlesMaterial>`.
+
+SCREEN_TEXTURE mip-maps
+-----------------------
+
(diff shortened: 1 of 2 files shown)


akien-mga:
Thanks! I did a quick proofreading pass and added some details, squashed into the merge commit.

tutorials/misc/gles2_gles3_differences.rst
```diff
@@ -0,0 +1,79 @@
...
+
+You do so with the following line of code at the end of the fragment shader:
+
+.. code-block:: glsl
+
+		ALBEDO =  mix(pow((ALBEDO + vec3(0.055)) * (1.0 / (1.0 + 0.055)),vec3(2.4)),ALBEDO * (1.0 / 12.92),lessThan(ALBEDO,vec3(0.04045)));
+
+GLES2 does not need this. If your game is going to use both GLES2 and GLES3, wrap it in an ``if`` 
+statement checking to see if the output is in sRGB, using ``OUTPUT_IS_SRGB``. 
+
```
reduz: I think this may be a bit misleading since in most cases it does not really matter and user will be ok with this looking a bit different. Also using mix and  lessThan is very un-optimal in OpenGL ES 2.0 because it's emulated in code (mix does not really accept booleans in GLES2.

I would remove entirely the converting comment and just specify that they are in different colorspaces, which will result in them looking a bit different when light is applied (which is all that really matters).. without proposing anything else.

tutorials/misc/gles2_gles3_differences.rst
```diff
@@ -0,0 +1,79 @@
...
+- Refraction
+- Subsurface scattering
+- Anisotropy
+- Clearcoat
+- Depth mapping
+
+The option to use them may appear in custom :ref:`ShaderMaterials <class_ShaderMaterial>`, however, they 
+will be non-functional. For example, you will still be able to set ``SSS`` (which normally adds 
+subsurface scattering) in your shader, but nothing will happen.
+
```
reduz: These options are non visuble when using GLES2 actually, so they are not non-functional.

tutorials/misc/index.rst
```diff
@@ -11,3 +11,4 @@ Miscellaneous
    binary_serialization_api
    running_code_in_the_editor
    change_scenes_manually
+   gles2_gles3_differences
```
reduz: There are a few more things that are probably worth noting:

-GIProbe does not work in GLES2, using lightmaps (BakedLightmap) is recommended.
-Contact shadows also do not work in GLES2
-Reading from depth (DEPTH_TEXTURE) in the shader, does not work on every hardware. On old mobile hardware this is most likely not supported.
-Performance scales poorly in GLES2 with several lights, as each is processed in a separate render pass (in opposition to GLES3 which is all done in a single pass).
-On mobile, GLES2 requires ETC texture compression, while GLES3 requires ETC2. Showing how to enable ETC compression in the project settings if exporting to mobile would be nice.
-Blend Shapes do not work on GLES2 (could be made to work eventually, but it's a lot of work and wont be fast).
-GLES2 besides being linear color, is not HDR capable. This could be noted probably.

clayjohn: Thanks reduz! I have added these to the tracked and will submit an updated PR soon

tutorials/misc/gles2_gles3_differences.rst
clayjohn: I will clarify that. I was trying trying to let users know that they will still be autocompleted when writing shaders, but that does not mean that they will work.

tutorials/misc/gles2_gles3_differences.rst
clayjohn: This was added as a direct response to [this issue](https://github.com/godotengine/godot/issues/25142). The issue is that they look different even when no light is applied. How about I remove some of the description that could be confusing/misleading. And then remove the color correction code, but still note the use of ``OUTPUT_IS_SRGB``?

PR #2212, opened 2019-02-25
Small update to iOS compilation docs

samgreen:
Wanted to make a few tweaks. I plan to add a guided screenshot tutorial for iOS export in a separate PR.

### DIFF
--- a/development/compiling/compiling_for_ios.rst
+++ b/development/compiling/compiling_for_ios.rst
@@ -37,27 +37,30 @@ Alternatively, you can run
 
 ::
 
-    $ scons p=iphone arch=x86 target=debug
+    $ scons p=iphone arch=x86_64 target=debug
 
 for a Simulator executable.
 
-Additionally since some time Apple requires 64 bit version of application binary when you are uploading to iStore.
-The best way to provide these is to create a bundle in which there are both 32bit and 64 binaries, so every device will be able to run the game.
-It can be done in three steps, first compile 32 bit version, then compile 64 bit version and then use ``lipo`` to bundle them into one fat binary, all those steps can be performed with following commands:
+For recent devices, Apple requires 64-bit versions of application binaries when you are uploading to the Apple Store.
+The best way to provide these is to create a bundle in which there are both 32-bit and 64-bit binaries, so every device will be able to run the game.
+
+It can be done in three steps: first compile the 32-bit version, then compile the 64-bit version and then use ``lipo`` to bundle them into one "universal" binary.
+All those steps can be performed with following commands:
 
 ::
 
-    $ scons p=iphone tools=no target=release arch=arm32
+    $ scons p=iphone tools=no target=release arch=arm
     $ scons p=iphone tools=no target=release arch=arm64
-    $ lipo bin/libgodot.iphone.opt.arm32.a bin/libgodot.iphone.opt.arm64.a -output bin/godot.iphone.opt.universal.a
+    $ lipo bin/libgodot.iphone.opt.arm.a bin/libgodot.iphone.opt.arm64.a -output bin/godot.iphone.opt.universal.a
 
 If you also want to provide a simulator build (reduces the chance of any linker errors with dependencies), you'll need to build and lipo the ``x86_64`` architecture as well.
-:: 
-    $ scons p=iphone tools=no target=release arch=arm32
+
+::
+
+    $ scons p=iphone tools=no target=release arch=arm
     $ scons p=iphone tools=no target=release arch=arm64
     $ scons p=iphone tools=no target=release arch=x86_64
-    $ lipo -create bin/libgodot.iphone.opt.arm32.a bin/libgodot.iphone.opt.arm64.a bin/libgodot.iphone.opt.x86_64.a -output bin/godot.iphone.opt.universal.simulator.a
-
+    $ lipo -create bin/libgodot.iphone.opt.arm.a bin/libgodot.iphone.opt.arm64.a bin/libgodot.iphone.opt.x86_64.a -output bin/godot.iphone.opt.universal.simulator.a
 
 Run
 ---
--- a/development/compiling/compiling_for_ios.rst
+++ b/development/compiling/compiling_for_ios.rst
@@ -10,7 +10,7 @@ Requirements
 
 -  SCons (you can get it from macports, you should be able to run
    ``scons`` in a terminal when installed)
--  Xcode with the iOS SDK and the command line tools.
+-  Xcode 10.0 (or later) with the iOS (10.0) SDK and the command line tools.
 
 .. seealso:: For a general overview of SCons usage for Godot, see
              :ref:`doc_introduction_to_the_buildsystem`.
@@ -47,9 +47,16 @@ It can be done in three steps, first compile 32 bit version, then compile 64 bit
 
 ::
 
-    $ scons p=iphone tools=no bits=32 target=release arch=arm
-    $ scons p=iphone tools=no bits=64 target=release arch=arm64
-    $ lipo -create bin/godot.iphone.opt.32 bin/godot.iphone.opt.64 -output bin/godot.iphone.opt.universal
+    $ scons p=iphone tools=no target=release arch=arm32
+    $ scons p=iphone tools=no target=release arch=arm64
+    $ lipo bin/libgodot.iphone.opt.arm32.a bin/libgodot.iphone.opt.arm64.a -output bin/godot.iphone.opt.universal.a
+
+If you also want to provide a simulator build (reduces the chance of any linker errors with dependencies), you'll need to build and lipo the ``x86_64`` architecture as well.
+:: 
+    $ scons p=iphone tools=no target=release arch=arm32
+    $ scons p=iphone tools=no target=release arch=arm64
+    $ scons p=iphone tools=no target=release arch=x86_64
+    $ lipo -create bin/libgodot.iphone.opt.arm32.a bin/libgodot.iphone.opt.arm64.a bin/libgodot.iphone.opt.x86_64.a -output bin/godot.iphone.opt.universal.simulator.a
 
 
 Run


development/compiling/compiling_for_ios.rst
```diff
@@ -47,9 +47,16 @@ It can be done in three steps, first compile 32 bit version, then compile 64 bit
 
 ::
 
-    $ scons p=iphone tools=no bits=32 target=release arch=arm
-    $ scons p=iphone tools=no bits=64 target=release arch=arm64
-    $ lipo -create bin/godot.iphone.opt.32 bin/godot.iphone.opt.64 -output bin/godot.iphone.opt.universal
+    $ scons p=iphone tools=no target=release arch=arm32
```
akien-mga: `arm` was correct, `arm32` is a working alias, but the actual string testing against in `detect.py` (beyond this alias matching) is `arm`. Could be renamed to `armv7` to match what is passed to the compiler, but that's another topic :)

development/compiling/compiling_for_ios.rst
```diff
@@ -47,9 +47,16 @@ It can be done in three steps, first compile 32 bit version, then compile 64 bit
...
+    $ scons p=iphone tools=no target=release arch=arm64
+    $ lipo bin/libgodot.iphone.opt.arm32.a bin/libgodot.iphone.opt.arm64.a -output bin/godot.iphone.opt.universal.a
```
akien-mga: As per the above, the binaries won't have `arm32` extension a priori (if they do, it's a bug).

development/compiling/compiling_for_ios.rst
```diff
@@ -47,9 +47,16 @@ It can be done in three steps, first compile 32 bit version, then compile 64 bit
...
+
+If you also want to provide a simulator build (reduces the chance of any linker errors with dependencies), you'll need to build and lipo the ``x86_64`` architecture as well.
```
akien-mga: I've seen reports from users that they also need an `x86` build to run the simulator matching their `armv7` build, is this correct? We currently build all four architectures for official binaries.

development/compiling/compiling_for_ios.rst
samgreen: I figured the alias could preserve the clarity of 64 bit vs 32 bit while dropping the verbosity of `bits=`. I don't believe Apple has release a v7 device since ~2013 so we maybe we want to consider only supporting 64 bit moving forward?

development/compiling/compiling_for_ios.rst
samgreen: Doh! Copy and paste error.

```
$ ls bin
libgodot.iphone.debug.arm64.a     
libgodot.iphone.opt.arm64.a
godot.osx.opt.tools.64            
libgodot.iphone.debug.fat.a       
libgodot.iphone.opt.debug.arm64.a
libgodot.iphone.debug.arm.a       
libgodot.iphone.opt.arm.a         
libgodot.iphone.opt.x86_64.a
```

development/compiling/compiling_for_ios.rst
akien-mga: `bits` was unnecessary in the previous example, simply `arch=arm` *is* armv7.

development/compiling/compiling_for_ios.rst
samgreen: If I understand correctly, you're asking whether you need to provide a desktop 32 bit slice for simulator compatibility? I believe 100% of Apple hardware (MacBook and iMac) devices that users would be using have been 64 bit (and thus `x86_64`) since ~2011 or so.

development/compiling/compiling_for_ios.rst
akien-mga: And yeah, in the future we could drop `armv7` support, but not for 3.1. You'd be surprised at how much some users want to support old devices (often because that's the ones they have themselves to do tests on).

samgreen: 64 bit is iPhone 5s and newer. The iPhone 5c was the last 32 bit handheld device from Apple.

development/compiling/compiling_for_ios.rst
akien-mga: See https://github.com/Microsoft/cpprestsdk/issues/1015

samgreen: Thinking about this more, I think the flag we provided historically to support 32 bit intel mac was `i386`. But that was maybe 7-8 years ago.

samgreen: that timing...haha

samgreen: If I'm reading this correctly the last OS X hardware that was 32 bit was 2011.

https://en.wikipedia.org/wiki/List_of_Macintosh_models_grouped_by_CPU_type#Penryn

development/compiling/compiling_for_ios.rst
samgreen: Definitely do not want to rock the 3.1 boat 👍

development/compiling/compiling_for_ios.rst
samgreen: The steam hardware survey has OS X users are 98% 64 bit.

https://store.steampowered.com/hwsurvey/Steam-Hardware-Software-Survey-Welcome-to-Steam?platform=mac

akien-mga: But as per my above link, it doesn't seem linked to the host architecture, only the target one. If I understand correctly, whenever you do a build only for armv7 (e.g. to support those ~1% users running iPhone 5), you can only run it in simulator with a corresponding i386 slice.

samgreen: There are *many other* linker errors and issues that may prevent people from running in the simulator. I am currently building 3 slices and able to run in the simulator with no issues, on a 64 bit mac book from 5 years ago.

Let's clear up something, the simulator is *not* an emulator. I'm not being pedantic on this point, I literally mean that for you to run any iOS app on the simulator, all objects must be compiled for the _host architecture_. No matter what `arm*` slices you build, you must have native objects linked for the simulator to function.

The two are completely orthogonal and much of the simulator falls through to native OS X APIs.

samgreen: In fact, here's the output from `lipo` while inspecting the simulator binary, to show that it provides no capability for any other architectures:

```
lipo -info /Applications/Xcode.app/Contents/Developer/Applications/Simulator.app/Contents/MacOS/Simulator
Non-fat file: /Applications/Xcode.app/Contents/Developer/Applications/Simulator.app/Contents/MacOS/Simulator is architecture: x86_64```

samgreen: The issue you linked to is building for iOS 8 (2014) using an outdated toolchain:

https://github.com/leetal/ios-cmake/blob/master/ios.toolchain.cmake#L234

They should be using the `$(ARCHS_STANDARD)` flag instead of manually specifying architectures.

akien-mga: Alright, thanks :) Sounds we can drop the `x86` build in official iOS templates @hpvb (but keep `x86_64`).

development/compiling/compiling_for_ios.rst
mobileben: I currently support iOS 10 as the minimum. I added more support for the armv7 slices due to this. At least for now, keeping this support really should have no impact on others.

I would imagine once iOS 13 comes out, the desire for iOS 10 support wanes.

development/compiling/compiling_for_ios.rst
mobileben: i386 is required if armv7 slices are included. Even if you are running on a 64-bit Intel. If you build for any simulator where the device is armv7, it will build it as i386. Or if you are building for all architectures and one of the arch is armv7, it will build the i386.

This has broken me. Again, from above. There shouldn't be an impact on people providing support for armv7/i386. It's just "time" and bytes of storage. The time should be amortized on building the lib (provided you're not building it every time with the project, which really shouldn't be the case).

samgreen: Xcode 9.4 was the last version to support i386 compilation. I can build armv7 slices with Xcode 10 and run them on the simulator, without even the capability of building i386.

samgreen: @mobileben Thanks for the reply. Do you have any info about your environment? Just trying to figure out where the disconnect is coming from in my reasoning.

samgreen: I just attempted adding the build setting for `i386` in a new project and received this:

> The i386 architecture is deprecated. You should update your ARCHS build setting to remove the i386 architecture.

samgreen: It looks like the SDK slices (or at least the i386 slice) shipped in the 10.14 SDK are no longer multi-arch and only contain x86_64 sections. This confirms my previous findings via `lipo`.

mobileben: @samgreen I'm using Xcode 10.1. And my min deployment target is 10.0.

32-bit devices are iPhone 5, 5c, and iPad (Oct 2012). So if I build for an iPhone 5 simulator build, at least on my system it most certainly builds with i386

Example snippet

>  /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang -x objective-c++ -arch i386 -fmessage-length=0 -fdiagnostics-show-note-include-stack -fmacro-backtrace-limit=0 -std=c++14 -stdlib=libc++ -fobjc-arc -fmodules 

Moreover, if I lipo the resulting binary

> racerx:ben:JKS.app $ lipo -info JKS
> Non-fat file: JKS is architecture: i386

And if I also go one step further and go into the ~/Library/Developer/CoreSimulator/Devices location to where the device (simulator) is, it also yields that it is i386 arch (same results as above).

And as an FYI, at least for me I ran into this for a couple of reasons.

1. I like to build for different devices (even if it is the simulator as I don't own every device out there)) in order to verify things. In particular, my game works with a backend that vends properly sized assets. It does not use the model info (I use systemctlbyname), rather it uses native bounds values and then computes the best size from there. So I like to ensure that the different devices all work. 
2. I do also run on the simulator different configurations. As you can see from my settings not all of them only build the valid arch. Some build all. And in that case, I hit the i386 issue with cpprestsdk.

I do realize i386 is deprecated and I expect Apple at some point of time to completely remove it. I have a feeling the first step is to further hide the existence of it within Xcode. Realistically ... unless they plan on re-writing how they do things which is doubtful based on it being Apple ... they will need to back support this (32-bit slices) until people no longer support lower iOS SDKs (10 or earlier).

I'm showing my build settings related to Architectures. More or less I use the default except for building active architecture.

screen shot 2019-02-25 at 9 26 56 pm

When you say you can build for armv7 simulator build, are you positive it really isn't building with i386? From my understanding, this is how Xcode works. It implicitly builds with i386 for 32-bit slices.  It builds with x86_64 for 64-bit slices. Or rather this is what I reason it is doing based on observation.

I was the one that added back the i386 slice to cpprestsdk because they upgraded the iOS build process and then it broke me. I had previously provided another method that took into account 32-bit slices but adjusted the build to accommodate the update. By default, it shouldn't include the 32-bit slices.

My build invocation for cpprestsdk is:

`./configure.sh -deployment_target 10.0 -include_32bit -no_bitcode -config_only -clean`

Hope this helps?

development/compiling/compiling_for_ios.rst
BastiaanOlij: 100% agree, drop 32bit, Apple demands us to compile with the latest XCode anyway if you want to publish on the AppStore. Pretty sure the next major release of XCode won't even support compiling 32bit iOS anymore. It's a dead horse.

BastiaanOlij: (3.2+, leave 3.1 as is :) )

akien-mga:
Pushed a proofreading with the changes discussed above and some more language fixes.

akien-mga:
Thanks!

PR #2213, opened 2019-02-25
removed uneccesary comparator

soekmego:
Method `has_animation(param)` returns true if `param` is a valid animation and makes it uneccesary to check  `if boolean == true`


StraToN:
Thanks @soekmego 
@godotengine/documentation Would it be better if this change is followed by its counterpart in the actual source code of the FPS demo?

bojidar-bg:
I think it would be nice to update the demo's source as well, but it is not blocking IMHO.

PR #2214, opened 2019-02-26
Clarify how to use MultiMesh.set_instance_color.

rcorre:
Just calling set_instance_color will do nothing unless you have set
color_format and vertex_color_use_as_albedo. This is really confusing,
and I only discovered my error by finding
https://github.com/godotengine/godot/issues/10217 from another confused
user.

The docs should call out these requirements.


akien-mga:
See that file's header:
```
.. DO NOT EDIT THIS FILE, but the MultiMesh.xml source instead.
```

This should be changed in https://github.com/godotengine/godot

PR #2218, opened 2019-02-27
Updated the ScrollContainer description to be a little less ambiguous

nuke-makes-games:
Made it clear that the scrollbars are automatically created.


akien-mga:
This file is also automatically created, as its header mentions:
```
.. DO NOT EDIT THIS FILE, but the ScrollContainer.xml source instead.
```

You should do these changes at https://github.com/godotengine/godot instead.

PR #2219, opened 2019-02-28
Fix "GDScript scripting guide" link in FAQ

brainmulator:
In section [_What is GDScript and why should I use it?_](https://docs.godotengine.org/en/latest/about/faq.html#what-is-gdscript-and-why-should-i-use-it), in the first paragraph that is a broken link for GDScript basics. [(Broken link)](https://docs.godotengine.org/en/latest/about/gdscript_basics).

This is a pull request to fix the issue.


akien-mga:
Thanks!

PR #2221, opened 2019-03-01
Fixed the example in custom_modules_in_cpp.rst

nbanks:
Judging by this include in other modules, it looks like "core/" must now be specified whereas it wasn't necessary for Godot 3.0.6.

### DIFF
--- a/development/cpp/custom_modules_in_cpp.rst
+++ b/development/cpp/custom_modules_in_cpp.rst
@@ -60,7 +60,7 @@ Inside we will create a simple summator class:
     #ifndef SUMMATOR_H
     #define SUMMATOR_H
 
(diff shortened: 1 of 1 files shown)


akien-mga:
That was fixed in 81e1d6da600a78cd85d600fee5d936cb1b374d28 already. What you're editing here is the `3.0` branch, which is the documentation for 3.0.6, where those are indeed not necessary :)

akien-mga:
I cherry-picked 81e1d6da600a78cd85d600fee5d936cb1b374d28 to `3.0` nevertheless to reduce confusion.

nbanks:
Oops, you're right.   Looks like I googled the wrong documentation.

PR #2224, opened 2019-03-04
Fix "GDScript basics" link text to "GDScript" in tutorial

brainmulator:
The text present in "Step by Step" section, under "Scripting", contains some references to "GDScript" as "GDScript basics". It might be some regression by changing the reference doc text. The doc in 3.0 it's all right.

At first, "overriding" (custom link text) doesn't look like a great way of doing it, because I have to do to 3 of them, but then I saw the pattern in class documentation, like: :ref:`bool<class_bool>`;

This pull request is to fix the regression by "overriding" the text in the reference.


akien-mga:
Thanks!

PR #2234, opened 2019-03-07
Update gdnative-cpp-example.rst

m4gr3d:
Add note to specify when to expect Godot to auto-reload newly added/removed properties.

### DIFF
--- a/tutorials/plugins/gdnative/gdnative-cpp-example.rst
+++ b/tutorials/plugins/gdnative/gdnative-cpp-example.rst
@@ -568,10 +568,11 @@ You can now change this property and when you run your project, you will
 see that our Godot icon travels along a larger figure.
 
 .. note::
-The ``reloadable`` property in the ``gdexample.gdnlib`` must be set to ``true`` for Godot
-to automatically pick up the newly added property.
-However, this setting should be used with care as it may cause instability as the 
-allocated memory may still be from the older version.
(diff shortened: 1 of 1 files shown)


BastiaanOlij:
@m4gr3d cool! thanks :)

@karroffel , what is your current take on the reloadable property? I've always fought with it because memory is allocated based on the class definition when the object is instantiated. If you recompile the DLL and load the new version there is always a risk the memory currently allocated and the data held within does not agree with the new structure.

karroffel:
@BastiaanOlij *in theory* it should not be a problem since no allocations should actually persist the reload.

This should only happen when tool classes are uses, as the editor might hold objects then that have script instances attached to them that are managed by a GDNative library.

So for non-tool usage this should be safe, for tool scripts it should not be used.

BastiaanOlij:
Hmm, I always make that mistake because ProcMesh is a tool class :)  (and there are some other reasons this is dangerous with VR interfaces because they get loaded at boot up, also for the editor). 

@m4gr3d you might want to rewrite your paragraph slightly to indicate the reloadable property can lead to issues with "toolclasses" as karroffel says

m4gr3d:
@BastiaanOlij @karroffel Thanks for the info, I've updated the description accordingly.

akien-mga:
Thanks!

PR #2236, opened 2019-03-09
Fix grammar in introduction

Jayman2000:
In general, there are two ways to refer to Godot in a sentence. You can write either:

1. "Godot", or
2. "The Godot Engine"

This change fixes several occurrences of "Godot Engine" (which is grammatically incorrect if there's no "the" in front of it).

### DIFF
--- a/about/introduction.rst
+++ b/about/introduction.rst
@@ -8,7 +8,7 @@ Introduction
     func _ready():
         $Label.text = "Hello world!"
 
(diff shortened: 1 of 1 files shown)


arkology:
English isn't my native language, but as I know, "the" will be incorrect in this case.

akien-mga:
"Godot Engine" is a proper noun, so it does not need "The".

PR #2237, opened 2019-03-09
Added shading language differences to gles2 doc

clayjohn:
This PR covers the functions that are not available in GLSL in the GLES2 backend. 

addresses #2205 (once this PR is merged I think we can close this issue)


akien-mga:
Thanks!

PR #2251, opened 2019-03-16
Reworded difference between Dict and Resource

puchik:
- Clarified that all objects' variables will be invalidated, not just the "second" object.
- Changed "their memory" to "from memory" to avoid confusion.
- Rephrased some statements to minimize wordiness.
- Replaced "gone" with "deleted".


akien-mga:
Thanks!

akien-mga:
Cherry-picked for 3.1.

PR #2253, opened 2019-03-16
Add Emission Shapes info to Particles2D tutorial

### LINKED ISSUE #1819: Particle2D documentation is lacking
TL;DR
The process of using image masks to generate particles is not documented.

## Problem
After reading several times the Godot 3.0 docs on Particle2D, ParticleSystems and ParticleMaterial, I still don't understand how to use some Emission Shapes, namely
`EMISSION_SHAPE_POINTS = 3` and `EMISSION_SHAPE_DIRECTED_POINTS = 4`.

Thinking I could make some clarifying changes to the docs, I went ahead and downloaded the _2D Particles Demo_ and there they were, three Particle2D nodes emitting in a "**HI!**" shape. Still, I could not figure out how to replicate that; even copying over the same image file from the demo into a new project and trying to recreate it.

There seems to be an undocumented extra step (or at least one that I might have missed) where we need to get an image and generate a .tres file, but I couldn't find how.

References:
Particles 2D docs: http://docs.godotengine.org/en/3.0/classes/class_particles2d.html
Particle Systems  (2D) docs: http://docs.godotengine.org/en/3.0/tutorials/2d/particle_systems_2d.html
Particles Material docs: http://docs.godotengine.org/en/3.0/classes/class_particlesmaterial.html

Thank you!

homer666:
Closes #1819. I also got rid of a strange line that doesn't really make sense.

This is the first time I've modified a tutorial page, so someone should really look over this.


akien-mga:
Thanks, this looks great! :)

akien-mga:
Cherry-picked for 3.1.

PR #2260, opened 2019-03-18
Example of a light shader

ludorverr:
Working with something rather than nothing is always easier when you're new.

### DIFF
--- a/tutorials/shading/shading_reference/spatial_shader.rst
+++ b/tutorials/shading/shading_reference/spatial_shader.rst
@@ -272,7 +272,17 @@ To write a light function, assign something to DIFFUSE_LIGHT or SPECULAR_LIGHT. 
 means no light is processed.
 
 The light function is called for every light in every pixel. It is called within a loop for
-each light type. 
+each light type.
+
+Light shader example:
+
+.. code-block:: glsl
+
+    void light() {
+        DIFFUSE_LIGHT += vec3( dot(NORMAL,LIGHT) * ALBEDO * ATTENUATION );
+    }
+
+Keep in mind DIFFUSE_LIGHT must be added to rather than set, if you don't want one light to overwrite the work of others.
 
 +-----------------------------------+---------------------------------------------+
 | Built-in                          | Description                                 |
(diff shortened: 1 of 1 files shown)


tutorials/shading/shading_reference/spatial_shader.rst
```diff
@@ -272,7 +272,17 @@ To write a light function, assign something to DIFFUSE_LIGHT or SPECULAR_LIGHT.
...
+
+Light shader example:
+
+.. code-block:: glsl
+
+    void light() {
+        DIFFUSE_LIGHT += vec3( dot(NORMAL,LIGHT) * ALBEDO * ATTENUATION );
+    }
+
+Keep in mind DIFFUSE_LIGHT must be added to rather than set, if you don't want one light to overwrite the work of others.
```
clayjohn: "DIFFUSE_LIGHT" should have two back ticks around it ``DIFFUSE_LIGHT``. 

Also, you should rephrase this to be less commanding and more positive. Maybe something like "If you want the lights to add together, add the light contribution to ``DIFFUSE_LIGHT`` using ``+=``, rather than overwriting it."

tutorials/shading/shading_reference/spatial_shader.rst
```diff
@@ -272,7 +272,17 @@ To write a light function, assign something to DIFFUSE_LIGHT or SPECULAR_LIGHT.
 means no light is processed.
 
 The light function is called for every light in every pixel. It is called within a loop for
-each light type. 
+each light type.
+
+Light shader example:
```
clayjohn: This is fine, but in the rest of the tutorial examples are introduced using a sentence. So, something like "Below is an example of a custom light function using a Lambertian lighting model" might be better.

ludorverr:
I've applied your words, feel free to edit anything. If there's a better light shader example you can swap it out too.
I didn't add backticks since previous occurences of DIFFUSE_LIGHT and SPECULAR_LIGHT had no backticks. I've now added backticks for them too.
I also realized the vec3() was pointless so I removed that part.

clayjohn:
Looks good!

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 3.1.

PR #2274, opened 2019-03-21
Make a Japanese Godot's docs

JunShiozawa:
now editing.

### DIFF
--- a/index-ja.rst
+++ b/index-ja.rst
@@ -1,0 +1,126 @@
+Godot(ゴドー)ゲームエンジンの使い方 – *master* branch
+============================
+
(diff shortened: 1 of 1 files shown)


akien-mga:
Thanks for the contribution, but this is not how we handle the localization of the documentation, as it would be very unpractical and impossible to keep up to date.

Instead, translations are done on [Weblate](https://hosted.weblate.org/projects/godot-engine/godot-docs/), and you are very welcome to help translate Godot's documentation in Japanese there. Once it reaches a significant percentage of completion (~10%), I will create an official Japanese instance of the documentation to provide it for users.

Issue #2275, opened 2019-03-21
Emulate Touchscreen is moved to Input Devices section


volzhs:
https://docs.godotengine.org/en/3.1/getting_started/step_by_step/exporting.html

> The first step is to open “Project Settings” and find the Handheld section. Enable the Emulate Touchscreen option.

![Screenshot_20190322_002715](https://user-images.githubusercontent.com/8281454/54764463-30fa4400-4c3b-11e9-9e4a-d08027c94210.png)

cbscribe:
Fixed in #2285

PR #2278, opened 2019-03-22
adds clarity to where the handled section can be found

davidholiday:


volzhs:
that is for Godot 3.0.x, not current master which is for 3.1 or later.
see #2275

davidholiday:
ok. my bad. cheers!

PR #2287, opened 2019-03-24
Add note on using the MODE_KINEMATIC of RigidBody2D

Flonka:
It was hard to understand how to properly use a RigidBody2D in kinematic mode, after some experimentation i figured out what was happening. 

I believe this note should assist anyone with using the kinematic mode.


cbscribe:
I'm not sure this is the right place for this. This article is an introduction to physics nodes, covering all the different types. More in-depth details of working with rigid body nodes  and their modes would be better covered in the rigid body tutorial.

mhilbrunner:
Actually, maybe this should be in the class ref for RigidBody2D?

akien-mga:
> Actually, maybe this should be in the class ref for RigidBody2D?

I agree, this should go directly in the class reference.

Flonka:
Alright, I will take the notes over there at some point : ) I haven't looked at all on how to contribute there. Thanks for the feedback!

PR #2288, opened 2019-03-24
Update Size and Anchors Images

skyace65:
Updates the images on the Size and Anchors page. The old ones were for Godot 2.0, the new ones are for 3.1. All images have been run through png compression. I also changed the word "End" to "1" in the size and anchors page, since the inspector uses 1 and 0 for anchors now instead of End and Begin.


skyace65:
Updated the images so they have a transparent background instead of white.

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 3.1.

PR #2292, opened 2019-03-25
Fix references in best_practices/scenes_versus_scripts.rst

remram44:
This fixes the broken markup for the first reference, and also changes the second reference from an absolute URL to a `:ref:` (which will work locally & after the next version release).

https://docs.godotengine.org/en/latest/getting_started/workflow/best_practices/scenes_versus_scripts.html#named-types

Screenshot_20190325_143020


akien-mga:
Thanks!

akien-mga:
Cherry-picked for 3.1.

PR #2294, opened 2019-03-25
Update to reflect issue #25535 in godot

creikey:

### DIFF
--- a/tutorials/shading/screen-reading_shaders.rst
+++ b/tutorials/shading/screen-reading_shaders.rst
@@ -125,7 +125,9 @@ DEPTH_TEXTURE
 
 For 3D Shaders, it's also possible to access the screen depth buffer. For this,
 the DEPTH_TEXTURE built-in is used. This texture is not linear; it must be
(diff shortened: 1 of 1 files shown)


merumelu:
This should be in a `.. note::` block (or maybe `.. attention::`) for better visiblity. http://docutils.sourceforge.net/docs/ref/rst/directives.html#admonitions

clayjohn:
I dont think its a good idea to add a note to the docs highlighting a bug in the current branch... docs should reflect intended behaviour. Also there is already a PR fixing this bug waiting to be merged.

creikey:
@clayjohn Why is it a bad idea to add a note of a bug? 3.1 is already in feature lock, and so the PR won't be able to be put into effect for some time.

akien-mga:
godotengine/godot#25535 will be fixed eventually, and likely backported to the 3.1 branch, so I don't think it should be documented either.

creikey:
@akien-mga If it will be fixed and then backported to 3.1, then I agree, it should not be documented. Closing the issue.

PR #2303, opened 2019-03-27
Edit Exporting code sample to match updated Dodge the Creeps code

Ursus-Z:
Some minor edits to the Step by Step/Exporting sample code so that it matches changes made to the 3.1 Dodge the Creeps code.

PR #2312, opened 2019-03-29
Change call_deferred to set_deferred (Step by Step)

Ursus-Z:
Changes  **call_deferred** method to **set_deferred**  for CollisionShape2D in  'Your First Game' and 'Exporting' pages, as per @akien-mga's suggestion  ( https://github.com/godotengine/godot-docs/pull/2303#discussion_r269552445 ) :

> It would be better to use `set_deferred("disabled", true)`. But that can be for another PR as I guess the other tutorial needs to be changed too.


cbscribe:
Thanks!

Issue #2298, opened 2019-03-26
Add a note to the documentation that DEPTH is only available in GLES3.


ZainlessBrombie:
[Copied from https://github.com/godotengine/godot/issues/27417]

Writing to the depth buffer in GLES2 completely breaks rendering. Until this is disabled in GLES2, could a note be added to the documentation?

clayjohn:
Fixed by #2313

PR #2313, opened 2019-03-29
Updated gles2 differences with built-in variables

clayjohn:
Address #2298 

This applies to 3.1 and master.

I think anisotropy is supposed to be in GLES2, there is code for it in scene.glsl, but it completely breaks rendering, and reduz has removed it from SpatialMaterial. There is a chance that the implementation is just bugged.

### DIFF
--- a/tutorials/misc/gles2_gles3_differences.rst
+++ b/tutorials/misc/gles2_gles3_differences.rst
@@ -195,3 +195,23 @@ For a complete list of built-in GLSL functions see the :ref:`Shading Language do
 +------------------------------------------------------------------------+--------------------------------------------------+
 | vec_type **fwidth** ( vec_type )                                       |                                                  |
 +------------------------------------------------------------------------+--------------------------------------------------+
+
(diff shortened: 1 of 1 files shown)


tutorials/misc/gles2_gles3_differences.rst
```diff
@@ -195,3 +195,20 @@ For a complete list of built-in GLSL functions see the :ref:`Shading Language do
 +------------------------------------------------------------------------+--------------------------------------------------+
 | vec_type **fwidth** ( vec_type )                                       |                                                  |
 +------------------------------------------------------------------------+--------------------------------------------------+
+
+Godot also provides many built-in variables and render modes. Some cannot be supported in GLES2. Below is a list of
+built-in variables and render modes that, when written to, will have no effect when using the GLES2 backend.
```
mhilbrunner: I'd prefer something like "will have no effect or could even cause issues".

clayjohn: Thanks! Should be good now.

PR #2319, opened 2019-03-30
Fix wrong type on replacing callback's types example

henriiquecampos:
I think the type on the example was wrong, in the callback, and after the example, it is mentioned `CollisionBody2D`, but in the fixed sentence it is `PhysicsBody2D`


mhilbrunner:
Thanks for this!

PR #2320, opened 2019-03-30
Update Audio Buses page for 3.1

skyace65:
Updates the images in the Audio Buses page. The old ones were for Godot 3.0, the new ones are for 3.1. All images have been run through png compression. Additionally I updated the page to mention the new record effect, added EQ to the section for the other EQ effects, and added a section for the Filter effect.


cbscribe:
Thanks!

PR #2322, opened 2019-03-31
update multimesh tutorial

skyace65:
The original tutorial used Node as the root node of the scene, which should not be done for a 3D scene. I've changed it to a spatial node and updated 2 pictures to reflect that. both pictures have gone through png compression.


cbscribe:
I'm fine with this, but I don't see any problem with using a Node if you don't need Spatial properties on your root node, ie it is just a container for several other nodes. Making the root node a Spatial doesn't add any functionality.

skyace65:
The reason I changed it is if the root node is a Node and not Spatial, you can't move it if you link to it in another scene. It doesn't cause problems for the tutorial, but it's a practice I don't think should be encouraged.

cbscribe:
Agreed it's situational. Anyway, I'm fine with it, so let's merge.

PR #2352, opened 2019-04-09
Update unity_to_godot.rst

bitbutter:
Many small changes. Disambiguation, removing redundant language, added a note about nested prefabs in new versions of Unity.

### DIFF
--- a/getting_started/editor/unity_to_godot.rst
+++ b/getting_started/editor/unity_to_godot.rst
@@ -213,7 +213,7 @@ depending on your scene and on what you want to achieve:
 - either add a new node between your target node and its current parent, then add a script to this new node.
 - or, you can split your target node into multiple children and attach one script to each of them.
 
-As you can see, it can be easy to turn a scene tree to a mess. Sonsider splitting any complicated scene into multiple, smaller branches.
+As you can see, it can be easy to turn a scene tree to a mess. Consider splitting any complicated scene into multiple, smaller branches.
 
 Connections: groups and signals
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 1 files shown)


getting_started/editor/unity_to_godot.rst
```diff
@@ -10,6 +10,10 @@ From Unity to Godot Engine
 This guide provides an overview of Godot Engine from the viewpoint of a Unity user,
 and aims to help you migrate your existing Unity experience into the world of Godot.
 
+.. note::
+
+   This article talks about older versions of Unity. Nestable prefabs ('Nested prefabs') were added to Unity 2018.3. Nestable prefabs are analogous to Godot's scenes, and allow a more Godot-like approach to scene organisation.
```
mhilbrunner: IMO we should then either remove all talk about this or if there are notable differences point these out instead.

Not much use using this as an selling point otherwise, as it is just outdated information. Adding this note only makes it slightly less confusing, but not any more correct :)

getting_started/editor/unity_to_godot.rst
```diff
@@ -224,8 +213,7 @@ depending on your scene and on what you want to achieve:
 - either add a new node between your target node and its current parent, then add a script to this new node.
 - or, you can split your target node into multiple children and attach one script to each of them.
 
-As you can see, it can be easy to turn a scene tree to a mess. This is why it is important to have a real reflection
-and consider splitting a complicated scene into multiple, smaller branches.
+As you can see, it can be easy to turn a scene tree to a mess. Sonsider splitting any complicated scene into multiple, smaller branches.
```
mhilbrunner: Sonsider -> Consider

getting_started/editor/unity_to_godot.rst
bitbutter: Agree. I'm thinking an entire re-write of this article might be necessary by someone who's had chance to get familiar with contemporary Unity workflow (with nested prefabs) and Godot.

StraToN: @mhilbrunner The purpose of this article is not to to list the selling points to a Unity user who desires to switch. To me it is primarily oriented to users who already made up their mind.
Thus it is not useless to point out the common points between Godot and Unity, but indeed it is useless to talk about them too much.

akien-mga:
Thanks!

PR #2354, opened 2019-04-09
Split out 'History of GDScript' into new file (didn't belong in 'GDScript basics')

bitbutter:

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_history.rst
+++ b/getting_started/scripting/gdscript/gdscript_history.rst
@@ -1,0 +1,31 @@
+GDScript History
+================
+
+In the early days, the Godot engine used the `Lua <http://www.lua.org>`__
+scripting language. Lua is fast, but creating bindings to an object
+oriented system (by using fallbacks) was complex and slow and took an
+enormous amount of code. After some experiments with
+`Python <https://www.python.org>`__, it also proved difficult to embed.
+
+The last third party scripting language that was used for shipped games
+was `Squirrel <http://squirrel-lang.org>`__, but it was dropped as well.
+At that point, it became evident that a custom scripting language could
+more optimally make use of Godot's particular architecture:
+
+-  Godot embeds scripts in nodes. Most languages are not designed with
+   this in mind.
+-  Godot uses several built-in data types for 2D and 3D math. Script
+   languages do not provide this, and binding them is inefficient.
+-  Godot uses threads heavily for lifting and initializing data from the
+   net or disk. Script interpreters for common languages are not
+   friendly to this.
+-  Godot already has a memory management model for resources, most
+   script languages provide their own, which results in duplicate
+   effort and bugs.
+-  Binding code is always messy and results in several failure points,
+   unexpected bugs and generally low maintainability.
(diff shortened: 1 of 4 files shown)


LikeLakers2:
I am curious why you feel it doesn't belong in "GDScript basics."

I think it belongs there considering it's a relatively short section, and it explains why the engine uses its own programming language rather than simply implementing a more well-known language -- although I do feel it could use some rewording to make it a bit more understandable to people who know nothing about programming. (Seriously though, I doubt most new people know about or even care about "memory management" so long as it works.)

bitbutter:
To me the title 'GDScipt basics' implies that i'm going to learn important/foundational practical information about how the language works. And/or, how to begin using it.

Discussion of how Godot _used_ to work isn't relevant to this aim (I don't care about Lua, Squirrelscript, or the size of the bindings code necessary to use these!).

At the moment this introduction gets in the way of reaching the information I expected to find on this page, which in turn undermines my confidence that the Godot docs are a good place to quickly find information I'm looking for.

Is there any disadvantage to splitting it out into a history page?

mhilbrunner:
I agree that this shouldn't be the first paragraph of the *GDScript basics* page immediately after the introduction.

We are not Wikipedia, our aim should be to teach Godot, not Godot history. Having it as a seperate page is fine - if people want to read about GDScript history, they will look for GDScript history, but someone searching for *GDScript basics* will probably appreciate the shorter text focusing on GDScript.

mhilbrunner:
Also, there is some overlap with these FAQ questions:

https://docs.godotengine.org/en/latest/about/faq.html#what-is-gdscript-and-why-should-i-use-it

https://docs.godotengine.org/en/latest/about/faq.html#what-were-the-motivations-behind-creating-gdscript

Maybe those should link to the new page, which could allow us to shorten the FAQ.

This information does not need to be in two places.

getting_started/scripting/gdscript/gdscript_history.rst
```diff
@@ -0,0 +1,31 @@
+GDScript History
```
akien-mga: Should be lower case 'history'.

getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -13,38 +13,6 @@ create content. It uses a syntax similar to
 to be optimized for and tightly integrated with Godot Engine, allowing great
 flexibility for content creation and integration.
 
-History
```
akien-mga: I would add a `.. seealso::` that references the new page. While I agree that it's not *necessary* to know GDScript history to learn the basics, this page is the one most people will start with when faced with the question "What is GDScript" and the subsequent question many have: "Why the hell does Godot have a DSL instead of using <insert favourite language>?". So it's important that this information is easy to find.

Calinou:
I think reviewing the FAQ question's text and linking to it in a "note" block should be enough. In this case, we don't need to create a new page.

mhilbrunner:
Done in https://github.com/godotengine/godot-docs/pull/3357. Thanks :)

PR #2358, opened 2019-04-09
Added missing `CallDeferred()` to C# Example

ScepticDope:
The note below the example references the call_deferred from the GDScript example but it was missing in the C# one.

### DIFF
--- a/getting_started/step_by_step/your_first_game.rst
+++ b/getting_started/step_by_step/your_first_game.rst
@@ -467,7 +467,7 @@ Add this code to the function:
     {
         Hide(); // Player disappears after being hit.
         EmitSignal("Hit");
-        GetNode<CollisionShape2D>("CollisionShape2D").CallDeferred("SetDisabled", true);
+        GetNode<CollisionShape2D>("CollisionShape2D").SetDeferred("disabled", true);
     }
 
 Each time an enemy hits the player, the signal is going to be emitted. We need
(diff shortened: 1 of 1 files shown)


ScepticDope:
Might also be a good idea to use a `private CollisionShape2D _collisionShape2D;` and set in the `_Ready()` and use it in the last two code blocks?

Does the next code block also need a call_deferred  on the following line? `GetNode<CollisionShape2D>("CollisionShape2D").Disabled = false;` My gut is telling me that enabling a CollisionShape2D during calculations might also throw an error in some cases.

neikeq:
I don't think this won't work. API calls using the PascalCase names don't work unless the user script implements the method (godotengine/godot#16872). It should be `CallDeferred("set_disabled", true)` instead.
I also think it's better to use `set_deferred("disabled", true)`/`SetDeferred("disabled", true)` instead.

getting_started/step_by_step/your_first_game.rst
```diff
@@ -467,7 +467,7 @@ Add this code to the function:
     {
         Hide(); // Player disappears after being hit.
         EmitSignal("Hit");
-        GetNode<CollisionShape2D>("CollisionShape2D").Disabled = true;
+        GetNode<CollisionShape2D>("CollisionShape2D").CallDeferred("SetDisabled", true);
```
ScepticDope:
```suggestion
        GetNode<CollisionShape2D>("CollisionShape2D").SetDeferred("disabled", true);
```

mhilbrunner:
Welcome and thank you! :)

For future PRs: changes should go against the **master** branch by default. Only use a version branch like **3.1** if the change applies to ONLY that version, and not future versions.

mhilbrunner:
Cherrypicked into master: https://github.com/godotengine/godot-docs/commit/4669c36a72dcfedfd84fb7dfbfcb8423a7ed7b7f

PR #2361, opened 2019-04-10
Update audio_streams.rst

bitbutter:
Removing redundant language. Disambiguation.

### DIFF
--- a/tutorials/audio/audio_streams.rst
+++ b/tutorials/audio/audio_streams.rst
@@ -6,10 +6,10 @@ Audio streams
 Introduction
 ------------
 
-As you might have already read in the :ref:`Audio Buses Tutorial<doc_audio-buses>` ,
+As you might have already read in the :ref:`Audio Buses Tutorial<doc_audio-buses>`,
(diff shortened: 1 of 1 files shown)


tutorials/audio/audio_streams.rst
```diff
@@ -9,40 +9,39 @@ Introduction
 As you might have already read in the :ref:`Audio Buses Tutorial<doc_audio-buses>` ,
 sound is sent to each bus via an AudioStreamPlayer.
 
-There are many types of AudioStreamPlayer, which will be explained in detail. Each one loads
+There are several types of AudioStreamPlayer described below. Each one loads
```
mhilbrunner: I'd suggest "different" instead of "several", because the focus should not be on their number, but that they have differences - a different one may be appropriate depending on your use case.

I'd also drop the "described below"/"which will be explained in detail" half-sentence, because it doesn't add much information. I already assume the page will tell me about it, seeing as that is the purpose of this page. :)

bitbutter: I agree, i've committed the changed version to the same patch.

PR #2362, opened 2019-04-10
Escaping double dashes.

DillonFlohr:
Escaping the double dashes in this file to prevent Sphinx from turning them to a long bar "–" in the docs.


mhilbrunner:
Thank you!

PR #2369, opened 2019-04-13
Fix malformed table in external editor page

YeldhamDev:


mhilbrunner:
Ah, missed that in https://github.com/godotengine/godot-docs/pull/2362 - thanks!

PR #2370, opened 2019-04-14
Change case sensitive string to "waiting"

kmcphillips:
Tiny change! 

But I believe keys are case sensitive, so the key added to dict `d` via `d.waiting =` would be `"waiting"` and not `"Waiting"`.


mhilbrunner:
Thats correct. Thank you!

PR #2381, opened 2019-04-16
Mentioned "Scene panel" by name

bitbutter:
The intro to the editor interface didn't mention the 'Scene panel' by name when it's introduced. I've fixed that.

### DIFF
--- a/getting_started/step_by_step/intro_to_the_editor_interface.rst
+++ b/getting_started/step_by_step/intro_to_the_editor_interface.rst
@@ -79,8 +79,8 @@ your project files and assets.
 
 |image11|
 
-On the right side, you’ll find the **Scene dock**, which lists the active
-scene’s content and the **Inspector** in the bottom right corner.
+On the right side, you’ll find the **Scene dock**. This dock contains the **Scene panel** which lists the active
+scene’s content and the **Inspector panel** which allows the properties of the currently selected node to be edited.
 
 |image12|
 


mhilbrunner:
Those panels are AFAIK actually called docks internally. :)

For example see the beginning here:
https://docs.godotengine.org/en/3.1/getting_started/step_by_step/intro_to_the_editor_interface.html

>We’re going to look at the Project Manager, docks, workspaces and everything you need to know to get started with the engine.

Also, Inspector and Scene dock are completely seperate, you can move them wherever you want.

As of today, however, this:
>On the right side

...is actually not true anymore in the default layout. So we should see about changing that.

bitbutter:
I believe that Godot follows the convention of many software editors (e.g. adobe suite) in that a dock is a reference to a place that a panel can be placed (placing the panel is called 'docking' a panel). Panels are the metaphorical 'ships' that can be 'anchored' in a dock :)

https://imgur.com/a/tX0fD2Q

See this image where the left dock (the FileSystem dock) is mentioned as a home for managing both project files (in the Files panel) as well as Assets (in the Assets panel). The dock is the home to two panels.

I'd like to go through this doc and make it clearer exactly which part of the interface is being referred to, by name, when I get chance.

akien-mga:
We actually use "dock" to refer to the panels, so "Scene dock" is the same as "Scene panel". The slots in which you can dock the panels don't have a function-based name, they are called Top-Left dock, Top-Right dock, Bottom-Left dock, Bottom-Right dock, etc. You can dock the Scene panel in any such positional dock, and several places in the documentation will refer to it as the "Scene dock".

NathanLovato:
I wrote this and asked back then: the sections of the interface that you can move around and dock into different places are called docks. You can also call them tabs if they are stacked in some slot into tabs. These names are common to some programs on Linux. On Krita, it's dockers. Anyway, I've said these in tutorials, and over 1+ year nobody told me they had an issue with that or felt confused.

Thanks for the help, but I don't think we need to change the keywords there. I'm closing for now, feel free to discuss it more and/or reopen if need be

PR #2383, opened 2019-04-16
Update services_for_ios.rst

JBacal:

### DIFF
--- a/tutorials/platform/services_for_ios.rst
+++ b/tutorials/platform/services_for_ios.rst
@@ -180,9 +180,9 @@ has 6 methods:
 
 -  ``Error post_score(Variant p_score);``
 -  ``Erroraward_achievement(Variant p_params);``
--  ``Error reset_achievements();``
--  ``Error request_achievements();``
--  ``Error request_achievement_descriptions();``
+-  ``Void reset_achievements();``
+-  ``Void request_achievements();``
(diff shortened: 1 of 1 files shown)


tutorials/platform/services_for_ios.rst
```diff
@@ -180,9 +180,9 @@ has 6 methods:
 
 -  ``Error post_score(Variant p_score);``
 -  ``Erroraward_achievement(Variant p_params);``
```
darthdeus:
```suggestion
 -  ``Error award_achievement(Variant p_params);``
```

Might be worth fixing this as well, which is probably just a typo.

akien-mga:
These changes should be made in the `master` branch, from where they can be cherry-picked to stable branches if relevant.

tutorials/platform/services_for_ios.rst
```diff
@@ -180,9 +180,9 @@ has 6 methods:
...
--  ``Error reset_achievements();``
--  ``Error request_achievements();``
--  ``Error request_achievement_descriptions();``
+-  ``Void reset_achievements();``
```
akien-mga: Why is `Void` uppercase? I'm not very familiar with ObjC++ but AFAIK it also uses `void` with a lowercase.

PR #2385, opened 2019-04-17
Added C# code to Part 1 of the FPS tutorial

Xylord:

### DIFF
--- a/tutorials/3d/fps_tutorial/part_one.rst
+++ b/tutorials/3d/fps_tutorial/part_one.rst
@@ -262,25 +262,18 @@ Add the following code to ``Player.gd``:
         public float MaxSlopeAngle = 40.0f;
         [Export]
         public float MouseSensitivity = 0.05f;
-        [Export]
-        public float MaxSprintSpeed = 30.0f;
-        [Export]
-        public float SprintAccel = 18.0f;
 
-        private Vector3 vel = new Vector3();
-        private Vector3 dir = new Vector3();
-        private bool isSprinting = false;
+        private Vector3 _vel = new Vector3();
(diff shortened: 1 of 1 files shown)


Xylord:
Rémi suggested I make a PR with only Part 1 to receive feedback on style and any other issues there might be before forging on with the other parts. Pretty much everything in this part's GDScripts exists 1:1 in C#, so this was not too difficult. The main thing I'm wondering about is whether anything should be added/modified in the body of the text. All the variables have the same name, except they now use PascalCase/CamelCase as required, so it shouldn't be too confusing, but someone else will know better.

exts:
Update all private fields to start with an underscore per the style guide http://docs.godotengine.org/en/latest/getting_started/scripting/c_sharp/c_sharp_style_guide.html#naming-conventions

mhilbrunner:
Thank you and welcome! Congrats on your first PR :)

I found nothing besides what @exts mentioned above.

Xylord:
@exts Got it, I'll correct this, thank you. Thanks @mhilbrunner, looking forward to this being the first of many.
How do we feel about [Export]ing all the public properties? I believe this replicates the behaviour of the GDScript version better, but I was wondering if that might be unnecessary clutter.

mhilbrunner:
I think thats fine :) Thank you!

mhilbrunner:
Merged. Thanks again!

Xylord:
@mhilbrunner Cool! Then I'll start putting up the other parts. I'm guessing a single PR for all the parts would be better?

mhilbrunner:
Whatever suits you, I don't think it makes much difference :)

PR #2389, opened 2019-04-17
Update class_tween.rst with C# example

MegaMech:
Added C# example for Tween. I tried to follow formatting from another page. So I hope it's correct.

### DIFF
--- a/classes/class_tween.rst
+++ b/classes/class_tween.rst
@@ -198,8 +198,8 @@ Here is a brief usage example that causes a 2D node to move smoothly between two
     
  .. code-tab:: csharp
  
-    var tween = GetNode("Tween") as Tween;
-    tween.InterpolateProperty($Node2D, "position",
+    var tween = (Tween)GetNode("Tween");
+    tween.InterpolateProperty((Node2D)GetNode("Node2D"), "position",
             new Vector2(0, 0), new Vector2(100, 100), 1,
             Tween.TransitionType.Linear, Tween.EaseType.Out);
     tween.Start();
--- a/classes/class_tween.rst
+++ b/classes/class_tween.rst
@@ -187,13 +187,22 @@ Tweens are useful for animations requiring a numerical property to be interpolat
 
 Here is a brief usage example that causes a 2D node to move smoothly between two positions:
 
-::
+.. tabs::
+ .. code-tab:: gdscript GDScript
 
     var tween = get_node("Tween")
     tween.interpolate_property($Node2D, "position",
             Vector2(0, 0), Vector2(100, 100), 1,
             Tween.TRANS_LINEAR, Tween.EASE_IN_OUT)
     tween.start()
+    
+ .. code-tab:: csharp
+ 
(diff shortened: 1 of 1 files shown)


classes/class_tween.rst
```diff
@@ -187,13 +187,22 @@ Tweens are useful for animations requiring a numerical property to be interpolat
...
 
     var tween = get_node("Tween")
     tween.interpolate_property($Node2D, "position",
             Vector2(0, 0), Vector2(100, 100), 1,
             Tween.TRANS_LINEAR, Tween.EASE_IN_OUT)
     tween.start()
+    
+ .. code-tab:: csharp
+ 
+    var tween = GetNode("Tween") as Tween;
```
neikeq: Don't use `as` if you're not going to do a null check. Prefer a cast `(Tween)GetNode("Tween")` or the equivalent `GetNode<Tween>("Tween")`. The reason for this is that a possible `InvalidCastException` in the place of the cast is much better than a `NullReferenceException` somewhere else.

classes/class_tween.rst
```diff
@@ -187,13 +187,22 @@ Tweens are useful for animations requiring a numerical property to be interpolat
...
+    tween.InterpolateProperty($Node2D, "position",
```
neikeq: `$Node2D` --> `GetNode("Node2D")`

LikeLakers2:
Just a friendly reminder that this `.rst` file is auto-generated from <https://github.com/godotengine/godot/blob/master/doc/classes/Tween.xml> over on the engine repo. If you don't want your changes to be overwritten the next time this documentation is updated, you should make your changes there.

MegaMech:
> 
> 
> Just a friendly reminder that these `.rst` files under `classes/` are auto-generated from https://github.com/godotengine/godot/blob/master/doc/classes over on the engine repo. If you don't want your changes to be overwritten the next time this documentation is updated, you should make your changes there.
> 
> Here's the file you should be editing: https://github.com/godotengine/godot/blob/master/doc/classes/Tween.xml

Thanks! I can't find a file with C# in it to know how to do the tab thing in XML

classes/class_tween.rst
MegaMech: Thanks!

neikeq:
Ah, I didn't realize this was the class reference. Right now, we have no way do provide code blocks for different languages in the class reference. Hopefully this can be solved in the future.

MegaMech:
> 
> 
> Ah, I didn't realize this was the class reference. Right now, we have no way do provide code blocks for different languages in the class reference. Hopefully this can be solved in the future.

Does this mean that only the tutorials are... "bilingual"?

neikeq:
For now, yes. We have to figure a way to support multiple languages in the class reference.

akien-mga:
This can't be merged for now as mentioned above, this file is generated from the XML which doesn't support language-specific code blocks.

This needs to be fixed somehow, but it will be via some work way beyond the scope of this PR. Thanks for the contribution nevertheless!

PR #2392, opened 2019-04-20
Add vim/gvim to a list of external editor flags

darthdeus:
Sidenote: Even though the flags work both for vim and gvim, it probably makes more sense to use gvim, because the way Godot launches the editor won't work with normal vim since it doesn't have a terminal to live in.


mhilbrunner:
Thank you!

PR #2396, opened 2019-04-23
Grammar fix: "it's" -> "its"

follower:
Because apparent intended wording is not "...thanks to [it is] high level constructs...".


clayjohn:
Good catch, can you also change "Using Servers" in the title to lowercase. I somehow missed it in my review.

akien-mga:
Thanks!

PR #2397, opened 2019-04-23
Fix inconsistent indentation from mixed tab/spaces

follower:
For some reason, in most cases the indent for the first 2 lines of code examples were space indented while the 3rd line was tab indented which resulted in excessive indentation for the 3rd line when rendered.

Replaced tabs with spaces in these and a couple of other cases.

(I'm not sure if there might be some unintended issue with rendering as a result of this change as I made the edits via the GitHub web editor.)


clayjohn:
Changes look good. 

However, you should always rebuild the docs after making changes to ensure you haven't broken anything.

akien-mga:
Thanks!

PR #2403, opened 2019-04-24
Fix description of get_modified_time

Shewebi:
In Godot 3.1 get_modified_time() returns 0 for nonexistant files and not a String as the current documentation suggests. THis change reflects the current behaviour.


cbscribe:
API docs are autogenerated. Changes to the class reference must be made in [godotengine/godot](https://github.com/godotengine/godot).

See [Updating the Class Reference](https://docs.godotengine.org/en/latest/community/contributing/updating_the_class_reference.html)

PR #2405, opened 2019-04-24
Improve doc & fix grammar

mrwanashraf:
in this commit i have improved the doc & fixed grammar in order for the doc to make more sense and to make it easier for readers to understand.

### DIFF
--- a/tutorials/content/procedural_geometry.rst
+++ b/tutorials/content/procedural_geometry.rst
@@ -4,7 +4,7 @@ Procedural geometry generation
 ==============================
 
 Users often ask how to generate geometry easily from code. This is not very complicated, but it's not obvious.
-Godot provides few classes entirely dedicated to make it this easy. Still, the best tool for the job depends
+Godot provides a few classes entirely dedicated to make it this easy. Still, the best tool for the job depends
 entirely on the use case.
 
 SurfaceTool
@@ -14,7 +14,7 @@ This is the most common helper. :ref:`SurfaceTool<class_SurfaceTool>` is a class
 
 It has a similar API to OpenGL 1.x, and it's meant for static content. This means, the mesh is generated once and then used.
 
-Here is a simple example on how to use it to add a single triangle.
+Here is a simple example of how to use it to add a single triangle.
 
 .. tabs::
  .. code-tab:: gdscript GDScript
--- a/tutorials/content/procedural_geometry.rst
+++ b/tutorials/content/procedural_geometry.rst
@@ -3,7 +3,7 @@
 Procedural geometry generation
 ==============================
 
-Users often ask how to generate geometry easily from code. This is not very complicated, but it's not obvious.
+Users often ask how to generate geometry from code. This is not very complicated, but it's not obvious.
 Godot provides a few classes entirely dedicated to make it this easy. Still, the best tool for the job depends
 entirely on the use case.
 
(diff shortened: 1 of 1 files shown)


tutorials/content/procedural_geometry.rst
```diff
@@ -3,8 +3,8 @@
 Procedural geometry generation
 ==============================
 
-Users often ask how to generate geometry from easily code. This is not very complicated, but it's not obvious.
-Godot provides a few classes entirely dedicated to make it this easy. Still, the best tool for the job depends
+Users often ask how to generate geometry easily from code. This is not very complicated, but it's not obvious.
+Godot provides few classes entirely dedicated to make it this easy. Still, the best tool for the job depends
```
clayjohn: You should not have removed the "a", Without it this sentence is broken. It should read "Godot provides a few classes" as it did previously.

tutorials/content/procedural_geometry.rst
```diff
@@ -14,7 +14,7 @@ This is the most common helper. :ref:`SurfaceTool<class_SurfaceTool>` is a class
 
 It has a similar API to OpenGL 1.x, and it's meant for static content. This means, the mesh is generated once and then used.
 
-Here is a simple example of how to use it to add a single triangle.
+Here is a simple example on how to use it to add a single triangle.
```
clayjohn: "of" is the correct word to use here. It should not have been changed.

tutorials/content/procedural_geometry.rst
mrwanashraf: i will fix it.

tutorials/content/procedural_geometry.rst
mrwanashraf: i will fix this one too.

clayjohn:
Looks good now!

mrwanashraf:
Thanks ^^

cbscribe:
In general, we should avoid using the word "easy" in documentation. At best, it adds nothing to the sentence. "Easy" is a relative term, and we shouldn't assume what *we* find easy will seem so to the reader.

> Users often ask how to generate geometry from code.

contains the exact same information as 

> Users often ask how to generate geometry easily from code.

mrwanashraf:
@cbscribe  how about now?

cbscribe:
My comment was intended to refer to this one as well:

> Godot provides a few classes entirely dedicated to make it this easy.

This sentence is pretty bad. "entirely" is unneeded. "to make it this easy" is just wrong. The more I read this paragraph the less content it has. What is it actually saying? It should be introducing what the rest of the tutorial is going to demonstrate. Here's how I'd fix it:

```
Users often ask how to generate geometry from code. In this tutorial, you'll learn how to use three tools that Godot provides to assist in generating geometry.
```

I think we can probably just leave it there. I hadn't read this doc before, and the rest of it needs a *lot* of work too. "Just explore the APIs and the possibilities." :/

clayjohn:
@cbscribe this is one of the ones reduz has added over the last few days. There are lots of comments like that. It would be worth reviewing all of them.

cbscribe:
Ah, I see. That explains how it got added before being reviewed first. :P

mrwanashraf:
@cbscribe  this is why i made this PR when i read the doc i felt like it needed improvement, well of course it needs much more improvement than what my PR is about but at least a little help wouldn't hurt.

akien-mga:
I pushed the equivalent of this PR's final changes as d420d5d1, thanks for the fix. I agree that the whole tutorial needs a proofreading pass.

PR #2407, opened 2019-04-24
Mesh.PRIMITIVE_TRIANGLES correction in proc geometry tutorial

tree786:
Lines 24 and 68 should be 'Mesh.PRIMITIVE_TRIANGLE**S**' and not 'Mesh.PRIMITIVE_TRIANGLE'


clayjohn:
Good catch!

mhilbrunner:
Thank you!

PR #2412, opened 2019-04-27
Make instances unique to adjust individually

irestone:


akien-mga:
There's a problem with this PR as it contains all the commits of the `3.1` branch which are not in the `master` branch, so it can't be merged.

I agree that there is an issue in the tutorial you're aiming to fix though, but IMO we should replace this section to avoid having to mention an advanced concept like making resources unique this early in the learning curve. See #2405 2459.

PR #2424, opened 2019-05-01
Update instancing.rst with correct information so that instancing works as expected.

prominentdetail:
This isn't mentioned, the local to scene value must be set to ON! It doesn't set the instance value correctly without it set.

### DIFF
--- a/getting_started/step_by_step/instancing.rst
+++ b/getting_started/step_by_step/instancing.rst
@@ -104,8 +104,7 @@ Press "Play" and notice that all of the instanced balls are now
 much more bouncy. Because the instanced balls are based on the saved scene,
 changes to that scene will affect all instances.
 
-You can also adjust individual instances. Set the bounce value back to ``0.5``
-and then in the ``Main`` scene, select one of the instanced balls. Set its
+You can also adjust individual instances. Set the bounce value back to ``0.5``. Open the ``Ball`` scene and select the Physics material. Set your physics material's Resource/Local To Scene value on. Then in the ``Main`` scene, select one of the instanced balls. Set its
 ``Bounce`` to ``1`` and press "Play".
 
 .. image:: img/instancing_property.png


getting_started/step_by_step/instancing.rst
```diff
@@ -104,8 +104,7 @@ Press "Play" and notice that all of the instanced balls are now
 much more bouncy. Because the instanced balls are based on the saved scene,
 changes to that scene will affect all instances.
 
-You can also adjust individual instances. Set the bounce value back to ``0.5``
-and then in the ``Main`` scene, select one of the instanced balls. Set its
+You can also adjust individual instances. Set the bounce value back to ``0.5``. Open the ``Ball`` scene and select the Physics material. Set your physics material's Resource/Local To Scene value on. Then in the ``Main`` scene, select one of the instanced balls. Set its
```
cbscribe: A couple of changes:

- I think using "Make Unique" on the instances might be better here than "Local to Scene". We're very early in the tutorial, and I think that will make more sense to the reader.

- "Physics material" -> PhysicsMaterial

- Note that in the rest of the doc, properties and options are in quotes.

- This might be a good place for a screenshot.

prominentdetail: Yeah.. the local to scene actually gave me some trouble afterwards when I wanted to set default value for all instances without a unique value.. I found out I had to revert the material first and then make the instance unique.. So the make unique seems to be the right way to do it. I'm just going through the tutorials since I'm new, and the instances got me stuck because it wasn't working.

cbscribe: Are you going to revise or should we close? FYI, if you revise, this PR should be made against master, not the 3.1 branch.

akien-mga:
I think the better approach would be to refactor this tutorial to avoid having to explain advanced notions like make resources unique or local to scene. This is one of the first tutorials users go through, it should stick to the basics. See #2459.

PR #2433, opened 2019-05-04
Custom iterators is_done() should check if curr >= end

afable:

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_advanced.rst
+++ b/getting_started/scripting/gdscript/gdscript_advanced.rst
@@ -435,7 +435,7 @@ functions in your script. An example implementation of a forward iterator follow
             self.increment = inc
 
         func is_done():
-            return (curr < end)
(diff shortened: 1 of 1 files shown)


mhilbrunner:
Thanks for pointing this out!

However, while logically you're right that that change would result in `is_done()` correctly returning whether the iterator *is done*, the code is working as is.

That is because the method is named confusingly: `_iter_init()` and `_iter_init()` should actually return `true` as long as the iteration should continue, i.e. as long as the iterator **isn't** *done*.

Exactly the opposite of what the `is_done()` name implies.

I've thus renamed it to `should_continue()`.

I also took the opportunity to improve the surrounding code somewhat, documented in the commit message.

Thanks again for pointing this out, I've added you as a co-author to the commit:
https://github.com/godotengine/godot-docs/commit/aed79c25833cb72172cdb3908f33b420b27b8882

PR #2434, opened 2019-05-04
Use Constant naming conventions as per style guide

afable:
https://docs.godotengine.org/en/3.1/getting_started/scripting/gdscript/gdscript_styleguide.html#naming-conventions


LikeLakers2:
I would like to point out that the style guide makes an exception for loading a class into a constant or variable -- look under "Classes and Nodes" in the section you linked to.

afable:
You are correct! My mistake. Thanks for pointing that out.

PR #2445, opened 2019-05-10
Remove "id" header from column

ttencate:
This cell has to be empty, else the import fails.

### DIFF
--- a/getting_started/workflow/assets/importing_translations.rst
+++ b/getting_started/workflow/assets/importing_translations.rst
@@ -58,7 +58,7 @@ unique and represent a string universally (they are usually in
 uppercase, to differentiate from other strings). Here's an example:
 
 +---------+------------------+----------------+--------------+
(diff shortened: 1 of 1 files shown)


akien-mga:
I think this should be fixed in the engine instead of changing the documentation. The parser should support any random first string that you use as header for the first column IMO.

akien-mga:
I tried in 3.1.1 and the current master branch, and both seem to support having an arbitrary value in the first cell. The code that imports the CSV actually skips it and checks from index 1 for the list of locales:

https://github.com/godotengine/godot/blob/48eaf14030f315a915cece0d33c190c28561ca01/editor/import/resource_importer_csv_translation.cpp#L95-L111

So I'm not sure what import issue you ran into.

I'll change the "id" to "ID" as mentioned above though.

akien-mga:
I added some more details in #2717.

PR #2448, opened 2019-05-11
Fix missing varible "velocity".

Hayden4126:
Unfortunately, I did not fix the missing variable for C#.


cbscribe:
Two issues:

1) PRs should be made against master, not the 3.1 branch.
2) This doc has already been revised in master to move the "shooting" section to a separate tutorial.

PR #2449, opened 2019-05-12
Fixed wrapf() error in the custom drawing in 2d tutorial

thebspin:
The docs had a mistake where it said wrap (And Wrap for c#) where it should be wrap and Wrapf for c#

### DIFF
--- a/tutorials/2d/custom_drawing_in_2d.rst
+++ b/tutorials/2d/custom_drawing_in_2d.rst
@@ -448,8 +448,8 @@ smaller value, which directly depends on the rendering speed.
         // We only wrap angles when both of them are bigger than 360.
         if (_angleFrom > 360 && _angleTo > 360)
         {
-            _angleFrom = Wrap(_angleFrom, 0, 360);
-            _angleTo = Wrap(_angleTo, 0, 360);
+            _angleFrom = Wrapf(_angleFrom, 0, 360);
+            _angleTo = Wrapf(_angleTo, 0, 360);
         }
         Update();
     }
--- a/tutorials/2d/custom_drawing_in_2d.rst
+++ b/tutorials/2d/custom_drawing_in_2d.rst
@@ -448,8 +448,8 @@ smaller value, which directly depends on the rendering speed.
         // We only wrap angles when both of them are bigger than 360.
         if (_angleFrom > 360 && _angleTo > 360)
         {
-            _angleFrom = Wrapf(_angleFrom, 0, 360);
-            _angleTo = Wrapf(_angleTo, 0, 360);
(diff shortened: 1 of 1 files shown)


mhilbrunner:
Thank you! For GDScript this indeed seems right.

For C#, `Wrap` still seems to be correct: https://github.com/godotengine/godot/blob/abc6dd7fa20bf3cb7bda6a555a76c1cf2ba1aff8/modules/mono/glue/Managed/Files/Mathf.cs#L326

Wondering whether that should be `Mathf.Wrap()` in this place in the docs, and whether we should change the C# function to match GDScript. cc @neikeq

thebspin:
Ah maybe i should have not assumed that the notation is the same for C# my bad :)

This being my first time contributing do i now need to update my master with just the GDScript example?

mhilbrunner:
No worries, IMO these two functions *should* really be the same :)

Yeah for now the best is probably to just fix the GDScript example.

You need to update your master branch. (as its the branch you created this PR with)

In general, there are two ways:

You either edit your last commit by amending your changes, or you create a new commit with the fix and then for cleanliness squash those two commits (the original + the fix) into one.

See https://www.burntfen.com/2015-10-30/how-to-amend-a-commit-on-a-github-pull-request or other tutorials around :)

Don't worry, this *is* a little involved when first getting used to it, if you need help or have questions feel free to ask here or on the docs channel on Discord!

neikeq:
It's `Mathf.Wrap` in C#, both for int and float. We named is this way as C# supports overloading, so the "f" suffix was unnecessary.

thebspin:
I tried the amending part, not sure if it went okay? Can you check? :)

bojidar-bg:
@thebspin It almost worked. You should have done a `git push --force` instead of `git pull` though, since the `git pull` generated a merge commit.
Something like this should get the PR to a proper state:
```
git reset --hard HEAD^
git push --force
# And, in case something went wrong:
# git reflog
# git reset --hard <commit SHA from reflog>
```

thebspin:
@bojidar-bg thanks!
I just did ``git reset --hard HEAD^
git push --force``
I am not sure where i can see myself if everything went correct? If i look at my own repo it says last commit 3 days ago so i assume i need to do something else/more?

mhilbrunner:
Were you on the right branch? If you force-pushed the PR branch, that force push would show up here :)

thebspin:
I should be..I didn't change or made any other new branch. I just tried it again, i now see my commit message underneath your comment.

mhilbrunner:
Yay, that looks correct now - besides that the commits should be squashed into one, but for now I'll do that with the merge. Thanks for this, and thanks for your patience with figuring out git :)

Thank you and congrats on your first contribution to Godot!

PR #2454, opened 2019-05-13
Fixed input event when the mouse's button left is pressed in 2d-movement tutorial

Jonasfxh:
While I was completing this tutorial I noticed that **event.is_action_pressed('click')** doesn't work anymore, so I had to change it to **event.is_mouse_button_pressed(BUTTON_LEFT)**. 

I fixed it on GDScript, but it would be a good idea if someone could please check it out on C#.

### DIFF
--- a/tutorials/2d/2d_movement.rst
+++ b/tutorials/2d/2d_movement.rst
@@ -291,7 +291,7 @@ on the screen will cause the player to move to the target location.
     var velocity = Vector2()
 
     func _input(event):
-        if event.is_action_pressed('click'):
+        if event.is_mouse_button_pressed(BUTTON_LEFT):
             target = get_global_mouse_position()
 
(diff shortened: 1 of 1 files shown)


cbscribe:
The "click" action is set up in the first part of the tutorial, which describes what actions the rest of the tutorial will use. The key actions are set up here too: 
https://docs.godotengine.org/en/latest/tutorials/2d/2d_movement.html#setup

Jonasfxh:
My bad. I used "ui_" actions on the whole tutorial. 
@cbscribe we can close this PR and thank you for the answer.

Jonasfxh:
PR closed due to the first step in the tutorial is showing how to add the named ("click") action to point to the LEFT MOUSE click for being able to call it with the function event.is_action_pressed().

PR #2468, opened 2019-05-18
Added self. to position in example

robinator8:
Without this change, the example code did not seem to work.


cbscribe:
This is incorrect. `self` is not necessary when accessing class methods and properties. Note that you didn't even add it to all uses of `position`.

The code in the tutorial has been very thoroughly tested. If you're having trouble making the project work, there are places where you can go for help:
- [QA Forum](https://godotengine.org/qa/)
- [Godot Discord](https://discord.gg/zH7NUgz)

robinator8:
Yes, my mistake. I should have investigated further! Looking back, I'm not sure why the example didn't work for me, as it is now working fine.

PR #2470, opened 2019-05-23
document the ability which allows to export a Curve resource variable…

fogine:
… for godot editor from gdscript

documents:  

```gdscript
export(Curve) var curve
```

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_basics.rst
+++ b/getting_started/scripting/gdscript/gdscript_basics.rst
@@ -1231,6 +1231,9 @@ special export syntax is provided.
 
     export(Texture) var character_face
     export(PackedScene) var scene_file
+    # There are many resource types that can be used this way, try e.g.
+    # the following to list them:
+    export(Resource) var resource
 
     # Integers and strings hint enumerated values.
 


akien-mga:
That's already covered above:
```
# Export can also take a resource type to use as a hint.

export(Texture) var character_face
export(PackedScene) var scene_file
```

`Curve` is also a resource type. If you try `export(Resource)` you'll see that the list of resource types is quite long so we can't list them all in this "basics" page.

fogine:
Thats great! However I think the documentation could be improved so that its more clear for the user who doesnt know godot from inside out.

Something along your lines:
> To see the full list of resource types you can  `export(Resource) var resource`.

Could be added so:

```gdscript
# Export can also take a resource type to use as a hint.
export(Texture) var character_face
export(PackedScene) var scene_file
# To see the full list of supported resource types: 
export(Resource) var resource
```

I believe that would save people some time.

akien-mga:
Sounds good, maybe something like this to make it clearer that it's only an example:
```diff
diff --git a/getting_started/scripting/gdscript/gdscript_basics.rst b/getting_started/scripting/gdscript/gdscript_basics.rst
index 64bb1ac1..c6aead0b 100644
--- a/getting_started/scripting/gdscript/gdscript_basics.rst
+++ b/getting_started/scripting/gdscript/gdscript_basics.rst
@@ -1231,6 +1231,9 @@ special export syntax is provided.
 
     export(Texture) var character_face
     export(PackedScene) var scene_file
+    # There are many resource types that can be used this way, try e.g.
+    # the following to list them:
+    export(Resource) var resource
 
     # Integers and strings hint enumerated values.
```

fogine:
I agree with your proposal, I've made suggested changes to the PR. if you want to take credit for it reauthor the commit or make a separate PR.

akien-mga:
Thanks!

PR #2471, opened 2019-05-23
Update scenes_and_nodes.rst

Unpotable:
I came across an issue while following it, where when running a scene the window would be a ghost, unclickable, with no explanation.  Since many people use high density displays, it's worth noting!

### DIFF
--- a/getting_started/step_by_step/scenes_and_nodes.rst
+++ b/getting_started/step_by_step/scenes_and_nodes.rst
@@ -162,6 +162,8 @@ demo should finally execute:
 
 Success!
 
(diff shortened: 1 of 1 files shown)


getting_started/step_by_step/scenes_and_nodes.rst
```diff
@@ -162,6 +162,8 @@ demo should finally execute:
 
 Success!
 
+(Note: if this doesn't immdiately work and you have a high-DPI display on at least one of your monitors, go to Project -> Project Settings -> Display -> Window then tick "Allow Hidpi" under Dpi!)
```
Calinou: I suggest rewriting it as a note with better formatting:

```suggestion
.. note::

    If this doesn't immediately work and you have a hiDPI display on
    at least one of your monitors, go to
    **Project → Project Settings → Display → Window** then enable
    **Allow Hidpi**.
```

mhilbrunner:
Thank you!

PR #2475, opened 2019-05-26
Introduction: Add "Before you start"

mhilbrunner:
**Preview:**

godot-wheretostart

### DIFF
--- a/about/introduction.rst
+++ b/about/introduction.rst
@@ -17,6 +17,17 @@ This page gives a broad presentation of the engine and of the contents
 of this documentation, so that you know where to start if you are a beginner or
 where to look if you need info on a specific feature.
 
+Before you start
+----------------
+
+The :ref:`Tutorials and resources <doc_community_tutorials>` page lists
+video tutorials contributed by the community. If you prefer video to text,
+those may be worth a look.
+
+In case you have trouble with one of the tutorials or your project,
+you can find help on the various :ref:`Community channels <doc_community_channels>`,
+especially the Godot Discord community, Q&A, and IRC.
+
 About Godot Engine
 ------------------
 


about/introduction.rst
```diff
@@ -17,6 +21,33 @@ This page gives a broad presentation of the engine and of the contents
...
+See the **About the documentation** section below for more details on how the
+Godot documentation is structured, but if you're just wondering
+*Where should I start?*, here are some pointers, depending on your preferences:
+
+**I am a complete beginner to game development.** |br|
+Reading this documentation in order is a good way to learn!
+So read on, finish this page and head on to the first section:
+:ref:`Step by step <toc-learn-step_by_step>`.
+
+**I already have some experience or just want to jump right in!** |br|
```
akien-mga: I'm not sure I would advise "experienced" developers to skip the first parts of the step by step which introduce key concepts of scenes and nodes.

mhilbrunner: Fair point, I guess it depends. If you know enough languages that GDScript isn't that surprising and enough engines that the concepts aren't totally foreign it IMO *is* a good point to start. It is what I did at least. :)

We can either add a sentence along the lines of "If you find you have trouble following the tutorial or want a more solid understanding, look at the content you skipped", or we drop the whole point.

IMO the step-by-step *does* start very slowly (foor good reasons), but maybe advanced people can be assumed to just skip ahead on their own.

I don't have a opinion on this. Any preference? The only thing important to me is directly pointing out the video tutorials on the very first page.

about/introduction.rst
```diff
@@ -17,6 +21,33 @@ This page gives a broad presentation of the engine and of the contents
...
+
+**All this text is boring me, give me something to watch!** |br|
+The :ref:`Tutorials and resources <doc_community_tutorials>` page lists
+video tutorials contributed by the community.
+
+**I need some help, my project isn't working.** |br|
+Take a look at the :ref:`Community channels <doc_community_channels>`,
+especially the Godot Discord community, Q&A and IRC.
+
+Whichever way you choose to start with: have fun and may Godot be with you!
```
cbscribe: Drop the first "with". May sound better: "However you choose to start"

Is "may Godot be with you" a thing now? :P

about/introduction.rst
```diff
@@ -17,6 +21,33 @@ This page gives a broad presentation of the engine and of the contents
...
+Skip some general introduction to the editor and start with
+:ref:`your first game <doc_your_first_game>`.
+
+**All this text is boring me, give me something to watch!** |br|
+The :ref:`Tutorials and resources <doc_community_tutorials>` page lists
+video tutorials contributed by the community.
+
+**I need some help, my project isn't working.** |br|
+Take a look at the :ref:`Community channels <doc_community_channels>`,
+especially the Godot Discord community, Q&A and IRC.
```
cbscribe: Missing Oxford comma.

about/introduction.rst
cbscribe: I agree. A newcomer to the engine, even an experienced one, should see the nodes/scenes stuff. Let them decide if they want to skim/skip.

about/introduction.rst
```diff
@@ -17,6 +21,33 @@ This page gives a broad presentation of the engine and of the contents
 of this documentation, so that you know where to start if you are a beginner or
 where to look if you need info on a specific feature.
 
+Where to start
+--------------
+
+So you're looking to get into Godot Engine - great!
```
cbscribe: Remove "So"

about/introduction.rst
```diff
@@ -17,6 +21,33 @@ This page gives a broad presentation of the engine and of the contents
...
+See the **About the documentation** section below for more details on how the
+Godot documentation is structured, but if you're just wondering
```
cbscribe: I think this should be two separate sentences. Cut at the "but".

about/introduction.rst
```diff
@@ -17,6 +21,33 @@ This page gives a broad presentation of the engine and of the contents
...
+*Where should I start?*, here are some pointers, depending on your preferences:
+
+**I am a complete beginner to game development.** |br|
+Reading this documentation in order is a good way to learn!
+So read on, finish this page and head on to the first section:
```
cbscribe: Add a comma after page. Change "head on" to "continue".

mhilbrunner:
I removed the skipping advice (and the "or read on in order" one, as that only made sense in contrast to this), and only the links to the community tutorials and channels remained.

Two points are too few for a list however, so I simplified the whole text a lot, shortening it to a few sentences pointing those resources out, incorporating the feedback from @cbscribe.

akien-mga:
Thanks!

PR #2478, opened 2019-05-27
Fix "looking around" C# example

CodingInfraviolet:
The example was referencing a variable that did not yet exist, instead of referencing "@event", so it did not compile. This fixes that, leading to correct behaviour.


akien-mga:
Thanks!

PR #2484, opened 2019-05-30
More accurate description of valid Parents

AspenBlack:

### DIFF
--- a/classes/class_control.rst
+++ b/classes/class_control.rst
@@ -566,7 +566,7 @@ Constants
 Description
 -----------
 
-Base class for all User Interface or *UI* related nodes. ``Control`` features a bounding rectangle that defines its extents, an anchor position relative to its parent and margins that represent an offset to the anchor. The margins update automatically when the node, any of its parents, or the screen size change.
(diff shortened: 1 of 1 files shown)


classes/class_control.rst
```diff
@@ -566,7 +566,7 @@ Constants
 Description
 -----------
 
-Base class for all User Interface or *UI* related nodes. ``Control`` features a bounding rectangle that defines its extents, an anchor position relative to its parent and margins that represent an offset to the anchor. The margins update automatically when the node, any of its parents, or the screen size change.
+Base class for all User Interface or *UI* related nodes. ``Control`` features a bounding rectangle that defines its extents, an anchor position relative to its parent control or the curent viewport and margins that represent an offset to the anchor. The margins update automatically when the node, any of its parents, or the screen size change.
```
akien-mga: Changes to the class reference should be done in the engine repo at https://github.com/godotengine/godot/blob/master/doc/classes/Control.xml

This file is automatically generated, as mentioned in its header:
```
.. DO NOT EDIT THIS FILE, but the Control.xml source instead.
```

akien-mga: Also two comments on the style: "current" takes two R-s, and I would add a comma after "viewport".

AspenBlack: Ok well I did not read the header.  Partly because I assumed that the link on the documentation "Edit on GitHub" would like to where it needs to be edited : ).

akien-mga: Yeah sadly the "Edit on GitHub" is inaccurate for the API documentation pages, but can't really be customized :/

PR #2485, opened 2019-05-30
Update tscn.rst

chucklepie:
Path not found. I only presumed 'blob' was in error rather than meaning something and I changed it to 'tree' to match the path to the file.

### DIFF
--- a/development/file_formats/tscn.rst
+++ b/development/file_formats/tscn.rst
@@ -12,7 +12,7 @@ indicate to Godot that the file has been exported from another program and
 should not be edited by the user from within Godot.
 
 For those looking for a complete description, the parsing is handled in the
-file `resource_format_text.cpp <https://github.com/godotengine/godot/blob/master/scene/resources/resource_format_text.cpp>`_
+file `resource_format_text.cpp <https://github.com/godotengine/godot/tree/master/scene/resources/resource_format_text.cpp>`_
 in the class :code:`ResourceFormatLoaderText`
 
 File structure


akien-mga:
That URL was valid: https://github.com/godotengine/godot/blob/master/scene/resources/resource_format_text.cpp

chucklepie:
But when you visit the documentation page:

https://docs.godotengine.org/en/3.1/development/file_formats/tscn.html#entries-inside-the-file

And click on the link to the file 'scene_format_text.cpp' it gives a 404 page not found.

Maybe I updated the wrong file?

akien-mga:
The link is wrong in the 3.1 branch indeed, but was fixed in the master branch since then. By default your are writing the master branch, which is why the link was already correct.

akien-mga:
This commit should be cherry picked to the 3.1 branch by a repo maintainer: 
https://github.com/godotengine/godot-docs/commit/3739074c47cff5d589549803f83102569a543295

I can do it tomorrow if no-one beats me to it.

chucklepie:
Sorry, my fault then. I thought I was in the same place.

akien-mga:
No worry, it's actually because we changed the "Edit on GitHub" link to ensure that it always links to the master branch, to avoid having people PR changes on old branches for things which should be done in master first.

But the drawback is this situation when one wants to fix an issue which is actually specific to an older branch :)

I'll cherry-pick it for `3.1` and `stable` in a bit together with other relevant changes.

PR #2492, opened 2019-06-02
Add C# example to "Removing old creeps"

MessinaBrothers:
The example includes code for GDScript, but not C#


akien-mga:
Thanks!

PR #2497, opened 2019-06-03
Add a Make.config to the Mono for Android build script

neikeq:
Forgot about this part. Otherwise it builds all these sdks instead of just Android.


mhilbrunner:
Thanks!

PR #2511, opened 2019-06-06
'vertex_array' Vector3 correction (proc geom gen tutorial)

tree786:
The `vertex_array` variable is defined as a Vector3, but it only contains 2 elements (I assume this is a mistake). So I have added the third element (being 0) to complete the size (i.e. make it a 3-component vector). These vertex vectors should now define the coordinates of a triangle in 3D space.


NathanLovato:
Thanks :)

PR #2516, opened 2019-06-08
Moved Detect 3D section to more intuitive location

afable:
Docs » Project workflow » Assets workflow » Importing images:
https://docs.godotengine.org/en/3.1/getting_started/workflow/assets/importing_images.html#detect-3d

Moved Detect 3D sub-section at the end of Importing textures main-section as the Detect 3d property is in the very top section of the Import tab (as shown in the Import options picture above).


akien-mga:
Thanks!

PR #2517, opened 2019-06-08
Add a page on localization using gettext

Calinou:
This documents an alternative way to localize Godot projects, which has been supported for a long time (it was already available in the 1.x era).

Please check if I missed anything of note :smiley:


mhilbrunner:
Yay :) thank you

PR #2518, opened 2019-06-09
Add tip about fixing commit mistakes in PR branches + minor grammar fix

puchik:
- Shortened and made minor grammar fixes to the third paragraph under "Updating your branch" in "PR Workflow"
- Added a tip regarding fixing commit mistakes in branches that are in pull requests.

These commands have been useful for me in the past but took some time to figure out how to use correctly. Feel free, however, to close if this is not the best approach to this issue. In any case, it would be useful for the documentation to mention how to proceed after making mistakes in forked repo branches.

### DIFF
--- a/community/contributing/pr_workflow.rst
+++ b/community/contributing/pr_workflow.rst
@@ -193,8 +193,8 @@ current upstream ``master`` branch, you will have to update your branch by
 
 However, if you had local commits, this method will create a so-called "merge
 commit", and you will soon hear from fellow contributors that those are not
-wanted in PRs. Then how to update the branch without creating a merge commit?
-You will have to use the ``--rebase`` option, so that your local commits are
+wanted in PRs. To update the branch without creating a merge commit,
+you will have to use the ``--rebase`` option, so that your local commits are
 replayed on top of the updated upstream ``master`` branch. It will effectively
 modify the Git history of your branch, but that is for the greater good.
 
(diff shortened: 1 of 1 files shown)


Calinou:
It should be noted that `git reset --hard` can be a dangerous operation, especially if some changes are untracked or uncommitted. However, if you've committed changes that you reset using `git reset --hard`, you may still be able to recover them by resetting to a commit ID found with the `git reflog` command.

puchik:
I was thinking about that too, and that is a good explanation. I can add that directly under inside a note box. i.e.
"Note:
git reset --hard can be a dangerous operation, especially if you have untracked or uncommitted changes. However, if you've committed changes that you reset using git reset --hard, you may still be able to recover them by resetting to a commit ID found with the git reflog command."

community/contributing/pr_workflow.rst
```diff
@@ -204,6 +204,20 @@ Therefore, the command that you should (almost) always use is:
 
     $ git pull --rebase upstream master
 
+
+If you have already pushed the merge commit without using ``rebase`` (or
```
NathanLovato: No need to parentheses here, just a comma:
`If you have already pushed the merge commit without using ``rebase``, or have made any mistakes that have resulted in undesired history, ...`

community/contributing/pr_workflow.rst
```diff
@@ -204,6 +204,20 @@ Therefore, the command that you should (almost) always use is:
 
     $ git pull --rebase upstream master
 
+
+If you have already pushed the merge commit without using ``rebase`` (or 
+have made any mistakes that have resulted in undesired history), you may
+use ``reset`` to revert to a specific commit and try again:
```
NathanLovato: Suggestion: you may use a hard reset. Reset itself doesn't change the history. Like @Calinou suggested, it'd be nice to explain what a hard reset is and what it does as it's a destructive operation.

NathanLovato:
Thanks for your contributions! I left you two comments to hopefully help improve it a bit.

puchik:
Thanks! I've pushed the changes.

akien-mga:
That's not the right place for such information IMO, there's a section further down on "Modifying a pull request" which explains how to interactively rebase a branch or amend a commit.

puchik:
This tidbit is supposed to address merging updates from master, whereas "Modifying a pull request" applies more for making small changes to your own work. The difference is that in the former you may be dealing with hundreds of commits, and in the latter it may be one or two which are more manageable to manually pick/fixup.

Let me know what you think.

akien-mga:
I see. Some explanations on how to reset one's branch to upstream/master can indeed be useful. I don't want it to be too verbose though, I'll have a closer look in coming days and see if I like the proposed text or if I would want any amend.

mhilbrunner:
Having looked at the page in question, IMO this is a good addition and the right place for this information.

I'll be daring and merge this, if only to decrease Akiens workload a little, and seeing the timeline on this PR. If we eventually come up with *even further* improvements, we'll chan do those in a new PR.

Thank you, @puchik!

PR #2519, opened 2019-06-10
get_skidinfo() misleading definition

tree786:
It seems (please correct me if you understand it otherwise) that the description for the `get_skidinfo()` function is the inverse of what actually happens. I have run some simple tests, and looks like setting low  `Friction Slip`  (e.g. try 1 or 2) causes more skidding and returns smaller values for `get_skidinfo()`, while when the `Friction Slip` is increased, the car skids less and values printed to the console increase (get closer to 1.0). So it seems that a value of 0 means skidding, while a value of 1.0 means no skidding (the description says the exact opposite to this, from what I understood).

### DIFF
--- a/classes/class_vehiclewheel.rst
+++ b/classes/class_vehiclewheel.rst
@@ -207,7 +207,7 @@ Method Descriptions
 
 - :ref:`float<class_float>` **get_skidinfo** **(** **)** const
 
-Returns a value between 0.0 and 1.0 that indicates whether this wheel is skidding. 0.0 is not skidding, 1.0 means the wheel has lost grip.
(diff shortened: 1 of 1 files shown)


NathanLovato:
Thanks for your contribution, but you need to edit the XML file for the class on the Godot repository: the RST files are generated from that. See https://docs.godotengine.org/en/latest/community/contributing/updating_the_class_reference.html for more info
Here's the file you need to edit, using the [pull request workflow](https://docs.godotengine.org/en/latest/community/contributing/pr_workflow.html): https://github.com/godotengine/godot/blob/master/doc/classes/VehicleWheel.xml

tree786:
Ok, thanks for the info!

PR #2521, opened 2019-06-10
Update the GDScript styleguide

### LINKED ISSUE #2505: GDScript - Spacing between methods
In many GDScript projects, I have seen several ways to separate functions. Sometimes one line:

```
func _ready():
    print("ready")

func _process(delta):
    print("process")
```

But sometimes two:

```
func _ready():
    print("ready")


func _process(delta):
    print("process")
```

I have also seen functions separated by 3 or more lines. I think that we should probably have no more than 2, but what should the policy be between 1 and 2 lines? Situationally? User discretion? Encourage one over the other? Disallow one or the other entirely?

We should probably decide on something and put it in [the style guide](https://docs.godotengine.org/en/3.1/getting_started/scripting/gdscript/gdscript_styleguide.html).

### LINKED ISSUE #2506: GDScript: update the style guide for static type hints
In the [static typing article](https://docs.godotengine.org/en/3.1/getting_started/scripting/gdscript/static_typing.html), and in many projects, I have seen this:

    var number: int = 5

But I have also seen this fairly often, with a space before the colon:

    var number : int = 5

We should probably decide on something and put it in [the style guide](https://docs.godotengine.org/en/3.1/getting_started/scripting/gdscript/gdscript_styleguide.html).

NathanLovato:
Add a note about type hints, closes #2506
Recommend to skip 2 lines between functions to help make the code easier to read, closes #2505

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_styleguide.rst
+++ b/getting_started/scripting/gdscript/gdscript_styleguide.rst
@@ -69,7 +69,20 @@ regular code blocks.
 Blank lines
 ~~~~~~~~~~~
 
-Surround functions and class definitions with a blank line.
+Surround functions and class definitions with two blank lines:
+
+::
+
+    func heal(amount):
+        health += amount
+        health = min(health, max_health)
+        emit_signal("health_changed", health)
+
+
+    func take_damage(amount, effect=null):
+        health -= amount
+        health = max(0, health)
+        emit_signal("health_changed", health)
 
 Use one blank line inside functions to separate logical sections.
 
@@ -194,3 +207,42 @@ Constants
 
 Use CONSTANT\_CASE, all caps, with an underscore (\_) to separate words:
 ``const MAX_SPEED = 200``
+
+Static typing
+-------------
+
+Since Godot 3.1, GDScript supports :ref:`optional static typing<doc_gdscript_static_typing>`.
+
+Type hints
+~~~~~~~~~~
+
+Place the colon right after the variable's name, without a space, and let the GDScript compiler infer the variable's type when possible.
+
+
+**Good**:
+
+::
+
+   onready var health_bar: ProgressBar = get_node("UI/LifeBar")
+
+   var health := 0 # The compiler will use the int type
+
+**Bad**:
+
+::
+
+   # The compiler can't infer the exact type and will use Node 
+   # instead of ProgressBar
+   onready var health_bar := get_node("UI/LifeBar") 
+
+When you let the compiler infer the type hint, write the colon and equal signs together: ``:=``.
+
+::
+
+   var health := 0 # The compiler will use the int type
+
+Add a space on either sides of the return type arrow when defining functions.
+
+::
+
+   func heal(amount: int) -> void:


getting_started/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -69,7 +69,20 @@ regular code blocks.
...
 ~~~~~~~~~~~
 
-Surround functions and class definitions with a blank line.
+Surround functions and class definitions with two blank lines:
+
+::
+
+    func heal(amount):
+        health += amount
+        health = max(health, max_health)
```
clayjohn: should be ``health = min(health, max_health)`` right?

getting_started/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -69,7 +69,20 @@ regular code blocks.
...
+        emit_signal("health_changed", health)
+
+
+    func take_damage(amount, effect=null):
+        health -= amount
+        health = max(0, health)
+        emit_signal("health_changed", health)    func another_function():
```
clayjohn: Looks like you left ``func another_function():`` in by mistake

NathanLovato:
Thanks for the review :)

getting_started/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -194,3 +207,36 @@ Constants
...
+
+Place the colon right after the variable's name, without a space, and let the GDScript compiler infer the variable's type when possible.
+
+**Good**:
+
+::
+
+   onready var health_bar: ProgressBar = get_node("UI/LifeBar")
+
+   var health: = 0 # The compiler will use the int type
```
aaronfranke: Most code I've seen writes this as `name := value`, I suppose to group symbols together and in a similar vain to `name += value`, but `name: = value` would be more consistent with `name: type = value`. I suppose it would be nice to get community feedback on this.

getting_started/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -194,3 +207,36 @@ Constants
...
+   onready var health_bar: = get_node("UI/LifeBar") # The compiler can't infer the exact type and will use Node instead of ProgressBar
+
+   var health : int = 0
+
+
+Add a space on either sides of the return type arrow when defining functions.
+
+::
+
+   func heal(amount:int) -> void:
```
aaronfranke: Note, this is different from [what you said earlier](https://github.com/godotengine/godot-docs/issues/2506#issuecomment-500108259), is this intentional?

getting_started/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -194,3 +207,36 @@ Constants
 
 Use CONSTANT\_CASE, all caps, with an underscore (\_) to separate words:
 ``const MAX_SPEED = 200``
+
+Static Typing
```
akien-mga: Don't capitalize inner words in section titles.

getting_started/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -194,3 +207,36 @@ Constants
...
+-------------
+
+Since Godot 3.1, GDScript supports :ref:`optional static typing<_doc_gdscript_static_typing>`.
```
akien-mga: The leading `_` should not be used in the hyperlink, so it should be `:ref:`optional static typing <doc_gdscript_static_typing>`.

getting_started/scripting/gdscript/gdscript_styleguide.rst
akien-mga: I think the recommended style should be `name := value` indeed.

getting_started/scripting/gdscript/gdscript_styleguide.rst
akien-mga: Indeed, should be `amount: int`.

getting_started/scripting/gdscript/gdscript_styleguide.rst
akien-mga: The style guide doesn't specify whether optional type hints should be used in the documentation when writing tutorials or code examples in the class reference. This is maybe not the right place for it though as this is a general style guide for engine users, but making a note here to make sure that we have some guidelines about this.

cbscribe: Agreed, this is outside the scope of the style guide, which exists to prescribe how GDScript should look, whether using optional type hints or not.

The question of whether type hints belong in tutorials and code snippets throughout the docs is another conversation entirely. For what it's worth, I'm against that, but we can debate that in a separate issue.

NathanLovato: I agree with Chris on this one, I wouldn't have type hints in the docs. Dynamic gdscript is easy on the eye, easy to read, so it helps to keep the manual short. Type hints are an optional feature, the docs should reflect that. Switching at this point would be a lot of work and I would expect it to make beginners more likely to quit reading the docs.

NathanLovato:
Fixed everything you pointed out. I also split a long comment as in the build, the code block would get too long and have a horizontal scrollbar.

getting_started/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -194,3 +207,44 @@ Constants
...
+
+**Bad**:
+
+::
+
+   # The compiler can't infer the exact type and will use Node 
+   # instead of ProgressBar
+   onready var health_bar: = get_node("UI/LifeBar") 
+
+   var health: int = 0
```
akien-mga: I'm not sure I would show this one as "bad". It's ok to recommend using `:=` for primitive types to let the compiler infer the type, but I think using `: int` explicitly should still be valid in the style guide.

NathanLovato: Makes sense, just removed that.

getting_started/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -194,3 +207,42 @@ Constants
...
+
+   var health := 0 # The compiler will use the int type
+
+**Bad**:
+
+::
+
+   # The compiler can't infer the exact type and will use Node 
+   # instead of ProgressBar
+   onready var health_bar: = get_node("UI/LifeBar")
```
aaronfranke: I think the syntax should be `:=` here as well, I know this is a "bad" example but it should demonstrate one bad thing at a time (the inability to infer a type beyond Node)

bojidar-bg: Note that you can do `x := $AnimationPlayer as AnimationPlayer`

NathanLovato:
Fixed all comments

akien-mga:
Thanks!

PR #2524, opened 2019-06-11
Add example code for setting the pause mode

Kaligule:
Add code example for whitelisting a node from being paused. This is intended to enhance the example (which currently only talks about the gui).

The C-equivalent of the code is missing, I am not in a possition to provide that one.

### DIFF
--- a/tutorials/misc/pausing_games.rst
+++ b/tutorials/misc/pausing_games.rst
@@ -50,6 +50,15 @@ property in a node:
 
 .. image:: img/pausemode.png
 
+You can achieve the same result in code:
+
+
+.. tabs::
+ .. code-tab:: gdscript GDScript
+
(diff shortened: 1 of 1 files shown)


tutorials/misc/pausing_games.rst
```diff
@@ -50,6 +50,15 @@ property in a node:
 
 .. image:: img/pausemode.png
 
+You can achieve the same result in code:
+
+
+.. tabs::
+ .. code-tab:: gdscript GDScript
```
StraToN: There's a trimming space at the beginning of line 57. Is that a problem @akien-mga ?

cbscribe: The space is correct here.

akien-mga:
Thanks!

PR #2528, opened 2019-06-12
Added information about shader constants

Chaosus:
Since the constants in shaders has been implemented in https://github.com/godotengine/godot/pull/29316 this information needs to be reflected in docs. I called it "global constants" to differ it from common local 'const' meaning

### DIFF
--- a/tutorials/shading/shading_reference/shading_language.rst
+++ b/tutorials/shading/shading_reference/shading_language.rst
@@ -421,6 +421,24 @@ Uniforms can also be assigned default values:
     uniform vec4 some_vector = vec4(0.0);
     uniform vec4 some_color : hint_color = vec4(1.0);
 
+Global constants
+~~~~~~~~~~~~~~~~
+
+Like uniforms, global constants are shared between all shader stages. However, they are not accessible outside of the shader and must be initialized at declaration.
+Use the **const** keyword to declare a variable as a constant. All basic types, except samplers can be declared as constants. Constants are useful when you want to have access to a value throughout your shader that does not need to be modified. Accessing and using a constant value is slightly faster than using a uniform.
+
+.. code-block:: glsl
+
+   shader_type spatial;
+
+   const float PI = 3.14159265358979323846;
+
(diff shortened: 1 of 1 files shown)


tutorials/shading/shading_reference/shading_language.rst
```diff
@@ -421,6 +421,24 @@ Uniforms can also be assigned default values:
     uniform vec4 some_vector = vec4(0.0);
     uniform vec4 some_color : hint_color = vec4(1.0);
 
+Global constants
+----------------
+
+Like the uniforms global constants are shared between all shader stages. However they are invisible outside of shader and must be initialized at declaration.
```
clayjohn: Instead say: "Like uniforms, global constants..."
and
"However, they are not accessible outside of the shader..."

tutorials/shading/shading_reference/shading_language.rst
```diff
@@ -421,6 +421,24 @@ Uniforms can also be assigned default values:
...
+The **const** keyword is used to make the variable to be a constant. All basic types, except samplers can be declared with that modifier.
```
clayjohn: "Use the **const** keyword to declare a variable as a constant. All basic types, except samplers can be declared as constants. Constants are useful when you want to have access to a value throughout your shader that does not need to be modified. Accessing and using a constant value is slightly faster than using a uniform."

tutorials/shading/shading_reference/shading_language.rst
```diff
@@ -421,6 +421,24 @@ Uniforms can also be assigned default values:
...
+
+.. code-block:: glsl
+
+   shader_type spatial;
+
+   const float PI = 3.14159265358979323846;
+
+They can't be modified, and additionaly cannot have hints, but multiple of them (if they have the same type) can be declared in a single expression e.g
```
clayjohn: "Constants cannot be modified and additionally..."

clayjohn:
Looks great! I just commented with a few suggestions that I think would be beneficial!

Chaosus:
Fixed, check again and merge when you want !

akien-mga:
Thanks!

PR #2541, opened 2019-06-18
index: Localize Weblate badge on i18n instances

akien-mga:
It will now show the translation status for the used language instead
of the overall completion for all languages.
The badge widget's text is also in the used language.

The drawback is that to do this I had to inject the localized image
directive in the `rst_epilog` for all pages, even though we only need
it for the index.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -110,3 +110,11 @@ gettext_compact = False
 # Exclude class reference when marked with tag i18n.
 if is_i18n:
     exclude_patterns = ['classes']
+    # Couldn't find a way to retrieve variables nor do advanced string
+    # concat from reST, so had to hardcode this in the "epilog" added to
+    # all pages. This is used in index.rst to display the Weblate badge.
+    rst_epilog = """
+.. |weblate_widget| image:: https://hosted.weblate.org/widgets/godot-engine/{weblate_locale}/godot-docs/287x66-white.png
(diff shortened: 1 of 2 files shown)


conf.py
```diff
@@ -110,3 +110,16 @@
 # Exclude class reference when marked with tag i18n.
 if is_i18n:
     exclude_patterns = ['classes']
+    # Add remaps here if locale code differs on Weblate and RTD
+    # Used to generate language status badge
+    weblate_remaps = {
+        'zh_CN': "zh_Hans",
```
akien-mga: For the reference, there are more remaps needed in theory, but I only add the one relevant for our current set of localized instances: https://github.com/godotengine/godot-docs-l10n/blob/master/build_langs.txt

akien-mga:
Before:
Screenshot_20190618_144418

After:
Screenshot_20190618_144641

(Don't mind surrounding text in English, I was testing locally without the .po files.)

conf.py
nijel: I've changed Weblate to accept original language codes here as well, so you can remove the remapping now :-). See https://github.com/WeblateOrg/weblate/commit/aa360585647593bcfffd5974c5766be3b63f3108

akien-mga: Thanks!

PR #2543, opened 2019-06-19
Fix #2526: Rewrite GDScript Signals section with gamedev examples

### LINKED ISSUE #2526: Fix signals section of GDScript documentation
**Link to doc: [GDscript signals](http://docs.godotengine.org/en/latest/getting_started/scripting/gdscript/gdscript_basics.html#signals)**

**Problem:**
The signal "data_found" is defined as being able to forward a single parameter variable, while the signal handler is defined as expecting two variables as input from the emit_signal function.

# your_notifier.gd

signal data_found(data)

var your_data = 42

func _process(delta):
   if delta == your_data:
      emit_signal("data_found", your_data)

# your_handler.gd

func your_handler(data, obj):
   print("Your handler was called from: ", obj.get_name(), " with data: ", data)

# your_game.gd

func _ready():
   var notifier = your_notifier.new()
   var handler = your_handler.new()

   notifier.connect("data_found", handler, "your_handler", [notifier])

NathanLovato:
This PR goes beyond the reported issue: the examples in the docs are really abstract in my opinion, and it isn't clear what signals are for. I remember Juan said they're a delegation mechanism, and I think mentioning the observer pattern on top of that would help experienced developers get the hang of them quickly.

To make the explanation clear, I had to add a bit of text compared to the original. Could you tell me if this is the way to go? I tried to keep it short but it may feel half-way between a reference and a tutorial.

In any case, I think the original examples, as they're abstract, don't help to understand how you're meant to use signals.

If it's good, I'll proceed with the rest of the section.

@cbscribe may I ask your feedback on this?


getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1463,49 +1463,88 @@ freed.
...
 
-   func your_handler():
-      print("Your handler was called!")
+    # Here, we define a function to use as a callback when the
+    # character's health_changed signal was emitted
+
+    ...
+    func _on_Character_health_changed(old_value, new_value):
+        if old_value > new_value:
+            progress_bar.color = RED
```
bojidar-bg: Maybe use constants from Color, like `Color.red`, as added by https://github.com/godotengine/godot/pull/14704?

NathanLovato: You're right, done :)

NathanLovato:
@bojidar-bg Thanks for the comment. Any feedback on the writing style and code examples?

getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1463,49 +1463,88 @@ freed.
...
+.. note:: 
 
-Here's an example that creates a custom signal in one script and connects
-the custom signal to a method in a separate script, using the
+   Signals are a `Callback
+          <https://www.wikiwand.com/en/Callback_(computer_programming)>`
+          mechanism. They also fill the role of Observers, a common programming
+          pattern. For more information, read the `Observer
+          tutorial<http://gameprogrammingpatterns.com/observer.html>`_ in the
+          Game Programming Patterns ebook.
```
bojidar-bg: Why is the whole paragraph indented?
Also, I think it would be better to link to Wikipedia instead of Wikiwand.

getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1463,49 +1463,88 @@ freed.
...
+
+You can connect these signals to methods the same way you connect built-in
+signals of nodes like :ref:`class_Button` or :ref:`class_RigidBody`.
+
+Here's an example of when signals can be useful. 
+
+Let's say we have a game where the playable character can take damage from
```
bojidar-bg: Can combine the two sentences:
```
For an example of when signals can be useful, let's consider we have a game where…
```

getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1463,49 +1463,88 @@ freed.
...
+enemies. We want a life bar on screen to react to health changes with an
+animation, but we want to keep the User Interface separate from the player in
```
bojidar-bg: "User Interface" shouldn't be capitalized here?

getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1463,49 +1463,88 @@ freed.
...
+   # Game.gd
+
 
    func _ready():
-      var notifier = your_notifier.new()
-      var handler = your_handler.new()
+      var character_node = get_node('Character')
+      var lifebar_node = get_node('UserInterface/Lifebar')
+
+      character_node.connect("health_changed", lifebar_node, "_on_Character_health_changed")
```
bojidar-bg: Maybe also include an example with connecting through the Node tab, instead of using the `connect` function?

getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1463,49 +1463,88 @@ freed.
...
+    # character's health_changed signal was emitted
+
+    ...
+    func _on_Character_health_changed(old_value, new_value):
+        if old_value > new_value:
+            progress_bar.color = Color.red
+        else:
+            progress_bar.color = Color.green
+        progress_bar.animate(old_value, new_value)
+    ...
```
bojidar-bg: TextureProgess and ProgressBar both don't have a `color` property or an `animate` method. Maybe you could use `modulate` and `value=` instead?

bojidar-bg:
Didn't see many problems with the writing style, just a few random notes :smiley:

getting_started/scripting/gdscript/gdscript_basics.rst
NathanLovato: This is the gdscript code/features reference, so I'm not sure it should be there.

NathanLovato: I mean i feel like this would better fit a tutorial dedicated to signals, as you'd then have to run the user through the node tab, the signals connection interface and its different buttons... plus talk about the advantages of one approach over the other.

getting_started/scripting/gdscript/gdscript_basics.rst
NathanLovato: Yes, you're right, it'll be simpler

NathanLovato: If you use `value = ..` it removes the animation part from the example and it'd make it non-functional. 

My goal is to showcase having to meaningful arguments in the emitted signal that are really used by the callback function.

I did not want to use a tween node as in introduces a new object with a more complicated function call.

NathanLovato:
I just updated the pr. Thank you for the review! I left two comments for the parts I did not change and why. And I will write the rest tomorrow.

NathanLovato:
Just finished writing this!

getting_started/scripting/gdscript/gdscript_basics.rst
bojidar-bg: You are already using the values in the `if` though.

NathanLovato: Yes, so that when the animation starts, the bar's color changes. You can imagine that the animate function would set the bar back to a neutral color after the animation ended.
It's not much but leaving the example the way it is would mean the code would give a weird result on screen: the bar stays red if you lost health, it stays green if you gain health. That's what I meant.

getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1463,104 +1463,137 @@ freed.
...
-These signals may be connected to methods in the same manner as you connect
-built-in signals of nodes such as :ref:`class_Button` or :ref:`class_RigidBody`.
+   # A signal that forwards two arguments
+   signal health_changed(old_value, new_value)
 
-Here's an example that creates a custom signal in one script and connects
-the custom signal to a method in a separate script, using the
-:ref:`Object.connect() <class_Object_method_connect>` method:
+   func took_damage(amount):
+   emit_signal()
```
bojidar-bg: Indentation, also invalid code?

getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1463,104 +1463,137 @@ freed.
...
 
-::
+    func take_damage(amount):
+        var old_health = health
+        health -= amount
 
-   # your_game.gd
+        # We emit the health_changed signal every time the
+        # character took damage
+        emit_signal('health_changed', old_health, health)
```
bojidar-bg: I feel double quotes are used more often than single quotes in GDScript. That being said, I cannot find anything about it in the styleguide (which uses both single and double quotes).

getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1463,104 +1463,137 @@ freed.
...
+        progress_bar.animate(old_value, new_value)
+    ...
 
-::
+.. note::
    
-   func pressed_handler(which):
-      print("A button was pressed! Button's name was:", which.get_name())
+    To use signals, your class has to extend the ``Object`` class or any
+    built-in node type like ``Node``, ``KinematicBody``, ``Control``...
```
bojidar-bg: "built-in node type" -> "type extending it, such as"

bojidar-bg:
Looks good to me overall.

NathanLovato:
Thanks, just fixed everything :)

getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1463,104 +1463,135 @@ freed.
...
+    signal health_changed(old_value, new_value)
 
-::
+    func take_damage(amount):
+        var old_health = health
+        health -= amount
 
-   # your_game.gd
+        # We emit the health_changed signal every time the
+        # character took damage
```
cbscribe: Should be "takes", since we're using present tense everywhere else.

getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1463,104 +1463,135 @@ freed.
...
-   func _ready():
-      var notifier = your_notifier.new()
-      var handler = your_handler.new()
+::
+
+    # Lifebar.gd 
 
-      notifier.connect("data_found", handler, "your_handler")
+    # Here, we define a function to use as a callback when the
+    # character's health_changed signal was emitted
```
cbscribe: Tense: was -> is

getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1463,104 +1463,135 @@ freed.
...
+   func _ready():
+      var character_node = get_node('Character')
+      var battle_log_node = get_node('UserInterface/BattleLog')
 
-   # your_handler.gd
+      character_node.connect("health_changed", battle_log_node, "_on_Character_health_changed", [character_node.name])
 
-   func your_handler(data, obj):
-      print("Your handler was called from: ", obj.get_name(), " with data: ", data)
+Our ``BattleLog`` node will receive each element in the binds array as an extra argument:
```
cbscribe: "will receive" -> "receives"

akien-mga:
See also #30192.

I did some testing and I see potential for being confused by signals even after this update, in particular around the passing of arguments in `emit_signal` and in the `connect` binds array, and how those interact with the callback method and the declaration of the signal.

In my tests, it seems that:
- the declaration of the signal doesn't take arguments into account. Whether it's `signal my_signal` or `signal my_signal(arg1, arg2)` seems to have no impact on the ability to bind methods with any number of parameters.
- arguments passed as varargs in the `emit_signal("name", varargs...)` call will be the first arguments in the callback method from the left.
- arguments passed via the binds array in `connect` will come after the ones from `emit_signal`, if any.

So for example:
```
signal my_signal

func _ready():
	connect("my_signal", self, "_on_my_signal", [ "bind1", "bind2" ])
	emit_signal("my_signal", "emit1", "emit2")

func _on_my_signal(arg1, arg2, bind1, bind2):
	prints(arg1, arg2, bind1, bind2)
```
prints
```
emit1 emit2 bind1 bind2
```

So in this docs `signal health_changed(old_value, new_value)` is the to forward two arguments, but that's not due to this declaration, only because two arguments are actually passed in `emit_signal`.

Which is also why things still work after adding the character name in the binds array, even though there is no mention to make the signal declaration `signal health_changed(old_value, new_value, character_name)`.

asheraryam:
@akien-mga In your example, the signal declaration does not listen any arguments at all, `signal my_signal`, is that intentional?

From the example I understood that the extra arguments in declaration are optional for binded arguments (bind1, bind2) but am not sure if that extends to the base parameters as well (arg1, arg2).

akien-mga:
@asheraryam As I mentioned:

> the declaration of the signal doesn't take arguments into account. Whether it's signal my_signal or signal my_signal(arg1, arg2) seems to have no impact on the ability to bind methods with any number of parameters.

I left out arguments from the signal declaration because they don't do anything. The same code snippet works the same with `signal my_signal`, `signal my_signal(arg1, arg2, arg3)`, or `signal my_signal(arg1, arg2, arg3, arg4, arg5, arg6)`.

asheraryam:
Thanks for the clarification!

> The same code snippet works the same with `signal my_signal`, `signal my_signal(arg1, arg2, arg3)`, or `signal my_signal(arg1, arg2, arg3, arg4, arg5, arg6)`.

This statement should be explicitly added to the docs 😄

NathanLovato:
> I left out arguments from the signal declaration because they don't do anything.

Actually, they're here for the signal connection tool in the node tab. The arguments you add after the signal name will show up in the editor's interface, and Godot will write them for you when it generates a method.

I'm updating the PR to explain this.

NathanLovato:
Just updated. And here's the picture that didn't upload in the previous comment:
 
gdscript_basics_signals_node_tab_1

akien-mga:
Thanks!

PR #2551, opened 2019-06-21
Update examples of proprietary tools

Sajo8:
As of Dec 3, 2018, NVIDIA PhysX is open-source. Thus it should be removed as an example of a proprietary tool from the design philosophy page


LinuxUserGD:
But the Godot FBX importer with assimp is still a proprietary tool?

Calinou:
@LinuxUserGD No, as Assimp doesn't rely on the Autodesk FBX SDK. It uses an unofficial open source FBX implementation.

PR #2552, opened 2019-06-22
Clarify that resources must be made unique

regakakobigman:
Most properties of instances can be edited separately, but resources are shared by default. Unless the PhysicsMaterial of a RigidBody2D is made unique on an instance, editing its properties will affect other instances.

https://www.reddit.com/r/godot/comments/c3k1et/stuck_on_bouncing_ball_tutorial/


NathanLovato:
Thanks :)

PR #2556, opened 2019-06-23
Mention Linux binary compatibility limitations on "Compiling for X11"

Calinou:
See https://github.com/godotengine/godot/issues/29976.

This should probably be cherry-picked to the `3.1` branch as it's quite important for those who build their own templates :slightly_smiling_face:


Houkime:
It might be good to shortly specify the reason, i.e "This is because glibc is forward-compatible but not backward-compatible".

NathanLovato:
Thanks, all good!

PR #2558, opened 2019-06-23
Fix bug in Creeps tutorial where creeps don't animate

giogadi:
In the Dodge the Creeps tutorial, we need to add a call to "Play()" for the creeps' animations to actually play.


cbscribe:
This is not necessary, as in the [Node Setup](https://docs.godotengine.org/en/latest/getting_started/step_by_step/your_first_game.html#node-setup) section this property is set in the Inspector:

giogadi:
Oops, I missed that! Thanks!

PR #2564, opened 2019-06-25
Fix incorrect order in recursive update command

IronicallySerious:
Correct order is `update --init --recursive` instead of `--init update --recursive`


StraToN:
Indeed, examples I could find on the web corroborate this PR. Waiting for some git master validation :)

akien-mga:
Thanks!

PR #2566, opened 2019-06-25
Adding Animation to the Mob Scene

aviraljanveja:
Current Tutorial does not instruct to enable animation in the mob scene. I realized it after noticing the difference between my completed game and the example picture included at the beginning of the Tutorial.
Adding a simple function to the mob scene solved my problem and the enemy mob now had animated motion. 
func _process(delta):
	$AnimatedSprite.play()
This can be verified here on my personal repository. 
Project Link - https://github.com/aviraljanveja/Godot/tree/master/DodgeGame
Please find the directory : Godot/DodgeGame/Play/Dodge Game.exe. alternatively: Godot/DodgeGame/project.godot.

### DIFF
--- a/getting_started/step_by_step/your_first_game.rst
+++ b/getting_started/step_by_step/your_first_game.rst
@@ -608,6 +608,21 @@ choose one of the three animation types:
           the scene. We're going to use ``randomize()`` in our ``Main`` scene,
           so we won't need it here. ``randi() % n`` is the standard way to get
           a random integer between ``0`` and ``n-1``.
+          
+Also we will add a ``_process(delta)`` function below to animate our mob.          
+.. tabs::
+ .. code-tab:: gdscript GDScript
+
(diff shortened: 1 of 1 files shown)


cbscribe:
Yes it does. In the [Node Setup](https://docs.godotengine.org/en/latest/getting_started/step_by_step/your_first_game.html#node-setup) section this property is set in the Inspector:

See also #2558

aviraljanveja:
> Yes it does. In the [Node Setup](https://docs.godotengine.org/en/latest/getting_started/step_by_step/your_first_game.html#node-setup) section this property is set in the Inspector:
> 
> > Set the Playing property in the Inspector to “On”
> 
> See also #2558

Oh I See. Guess the recurrent confusion is due to "ON" marked in the picture next to "LOOP" in the Animations Frame. Wording it a bit differently could vastly improve the clarity.

PR #2568, opened 2019-06-25
Minor Update

aviraljanveja:
A Minor but potentially very helpful update for a recurrent issue. 
1. https://github.com/godotengine/godot-docs/pull/2558
2. https://github.com/godotengine/godot-docs/pull/2566
Simply re-worded and explained clearly. 
That section could sometimes be confusing due to the image right below the text. 
This should help future learners.


cbscribe:
I don't think this fixes the problem.

If the problem is that users don't notice that they should check the "Playing" property - and I agree it is - how is adding "in the Animation panel" going to help that?  They need to be looking in the Inspector. It's probably better cleared up by adding another image to directly illustrate where the property is.

Also, adding "Remember to" is not going to make users more likely to read this. More words muddy the issue; they don't clarify it.

aviraljanveja:
Yeah I agree. No comparison with an additional image.

PR #2569, opened 2019-06-25
Add warning about using spaces in IPA filename.

njt1982:
See: https://github.com/godotengine/godot/issues/30052

### DIFF
--- a/getting_started/workflow/export/exporting_for_ios.rst
+++ b/getting_started/workflow/export/exporting_for_ios.rst
@@ -39,9 +39,8 @@ After you click **Export Project**, there are still two important options left:
           project's name. When you see **exported_xcode_project_name**
           in the following steps, replace it with the name you used instead.
 
-.. note:: Avoid using spaces when you chose your **exported_xcode_project_name** as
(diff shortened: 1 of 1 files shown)


LikeLakers2:
While I'm not an iOS dev, I do question if this is something that could be fixed, rather than having to warn the user about the issue.

njt1982:
@LikeLakers2 I agree about it being fixed in the engine being the ideal outcome, but in the meantime I think users could benefit from a heads up about the consequences of chosing spaces ;)

getting_started/workflow/export/exporting_for_ios.rst
```diff
@@ -39,6 +39,10 @@ After you click **Export Project**, there are still two important options left:
           project's name. When you see **exported_xcode_project_name**
           in the following steps, replace it with the name you used instead.
 
+.. note:: Avoid using spaces when you chose your **exported_xcode_project_name** as
+          this can lead to corruption in your XCode project file.
```
cbscribe: This should be "choose" not "chose".

akien-mga: Fixed.

akien-mga:
Thanks!

PR #2570, opened 2019-06-28
Update 'step_by_step/scripting_continued'

BenjaminRiley:
Removed stray 'do' from first paragraph in Groups section


StraToN:
Thanks !

PR #2575, opened 2019-06-29
Update SConstruct to generate debug symbols on windows

Demiu:
The SConstruct file from the 'GDNative C++ example' was generating debug symbols on OSX and linux, but not on windows.

I've tested this and a breakpoint I put in the gdexample.cpp will trigger with gdb attached to a godot instance run from the editor with this change.


akien-mga:
Thanks!

PR #2577, opened 2019-07-01
Add warning to reflect behavior with tool nodes

creikey:
In response to this issue: https://github.com/godotengine/godot/issues/30104


akien-mga:
I updated the warning so that it doesn't come first before introducing what tool scripts *are*. I also improved the paragraph a bit for clarity.

akien-mga:
Thanks!

PR #2544, opened 2019-06-19
Update changelog to add the "Localization using gettext" page

Calinou:
This also infers page names automatically from the pages.

PR #2578, opened 2019-07-02
Update documentation changelog with new pages added since 3.1

Calinou:
Follow-up to #2544.

This also moves some pages that were said to be added since 3.0, but were actually added after the 3.1 release.

This orders categories in the order they are mentioned in the index.


akien-mga:
Thanks!

PR #1916, opened 2018-11-13
Add instructions on compiling Godot with MinGW.

jtank4:
Adds instructions in development/compiling/compiling_for_windows.rst for how to compile Godot with MinGW.

PR #2579, opened 2019-07-02
Update and shorten "Compiling for Windows"

Calinou:
There's been a lot of build system changes in recent Godot versions. Visual Studio can now be automatically detected, MinGW performance is getting very competitive with MSVC and pywin32 is no longer required (even though SCons will still warn about it).

This also adds a short comparison between MSVC and MinGW, as the former is faster for iterating during development, but the latter is better suited to production builds.

This supersedes #1916.

### DIFF
--- a/development/compiling/compiling_for_windows.rst
+++ b/development/compiling/compiling_for_windows.rst
@@ -10,15 +10,19 @@ Requirements
 
 For compiling under Windows, the following is required:
 
--  Visual C++, `Visual
-   Studio Community <https://www.visualstudio.com/vs/community/>`__
-   (recommended), version 2013 (12.0) or later.
-   **Make sure you read Installing Visual Studio caveats below or you
-   will have to run/download the installer again.**
--  `Python 2.7+ or Python 3.5+ <https://www.python.org/downloads/>`__.
--  `Pywin32 Python Extension <https://github.com/mhammond/pywin32>`__
-   for parallel builds (which increase the build speed by a great factor).
--  `SCons <https://www.scons.org>`__ build system.
+- `Visual Studio Community <https://www.visualstudio.com/vs/community/>`_,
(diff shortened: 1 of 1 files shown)


development/compiling/compiling_for_windows.rst
```diff
@@ -331,152 +130,122 @@ in C++ is needed, for creating modules or extending the engine, working
...
+|               |     dnf install mingw64-gcc-c++ mingw64-winpthreads-static \ |
+|               |                 mingw32-gcc-c++ mingw32-winpthreads-static   |
++---------------+--------------------------------------------------------------+
+| **macOS**     | ::                                                           |
+|               |                                                              |
+|               |     brew install mingw-w64                                   |
++---------------+--------------------------------------------------------------+
+| **Mageia**    | ::                                                           |
+|               |                                                              |
+|               |     urpmi mingw64-gcc-c++ mingw32-gcc-c++                    |
```
akien-mga: Could you add `mingw64-winpthreads-static and mingw32-winpthreads-static`? It turns out they are also needed.

development/compiling/compiling_for_windows.rst
```diff
@@ -331,152 +130,122 @@ in C++ is needed, for creating modules or extending the engine, working
...
-    C:\godot> scons platform=windows tools=no target=release_debug
-
-If you plan on replacing the standard templates, copy these to:
-
-::
-
-    C:\USERS\YOURUSER\AppData\Roaming\Godot\Templates
+If you plan on replacing the standard export templates, copy these to the
+following location, replacing ``<version>`` with the version identifier
+(such as ``3.1.1`` or ``3.2.dev``)::
```
akien-mga: Would be `3.1.1.stable`.

akien-mga:
Thanks!

PR #2580, opened 2019-07-02
Describe the process of creating editor icons globally and per module

Xrayez:
In need for godotengine/godot#30167.

Info compiled from different sources, namely:

https://github.com/godotengine/godot/blob/master/editor/icons/README.md
https://github.com/godotengine/godot-design/blob/master/README.md

Also created a new `Editor development` chapter/section to encompass C++ editor development side specifically which is not necessarily related to pure core development, generic types and patterns (describing a process of creating C++ editor plugins is one example).

### DIFF
--- a/development/cpp/custom_modules_in_cpp.rst
+++ b/development/cpp/custom_modules_in_cpp.rst
@@ -410,6 +410,35 @@ the docs will become accessible within the engine's built-in documentation syste
 In order to keep documentation up-to-date, all you'll have to do is simply modify
 one of the ``ClassName.xml`` files and recompile the engine from now on.
 
+.. _doc_custom_module_icons:
+
+Adding custom editor icons
+--------------------------
+
+Similarly to how you can write self-contained documentation within a module,
+you can also create your own custom icons for classes to appear in the editor.
+
+For the actual process of creating editor icons to be integrated within the engine,
+please refer to :ref:`doc_editor_icons` first.
+
+Once you've created your icon(s), proceed with the following steps:
+
+1. Make a new directory in the root of the module named ``icons``. This is the
+   default path for the engine to look for module's editor icons.
+
+2. Move your newly created ``svg`` icons (optimized or not) into that folder.
+
+3. Recompile the engine and run the editor. Now the icon(s) will appear in
+   editor's interface where appropriate.
+
+If you'd like to store your icons somewhere else within your module,
+add the following code snippet to ``config.py`` to override the default path:
+
+   .. code:: python
+
+       def get_icons_path():
+           return "path/to/icons"
+
 Summing up
 ----------
 
--- a/development/editor/creating_icons.rst
+++ b/development/editor/creating_icons.rst
@@ -1,0 +1,98 @@
+.. _doc_editor_icons:
+
+Editor icons
+============
+
+When a new class is created and exposed to scripting, the editor's interface
+will display it with a default icon representing the base class it inherits
+from. Yet in most cases it is recommended to create icons for new classes
+to improve the user experience.
+
+Creating icons
+~~~~~~~~~~~~~~
+
+In order to create new icons, you first need a vector graphics editor installed.
+For instance, you can use the open-source `Inkscape <https://inkscape.org/>`_ editor.
+
+Clone the ``godot-design`` repository containing all the original editor icons:
+
+   .. code:: bash
+
+       git clone https://github.com/godotengine/godot-design
+
+The icons must be created in a vector graphics editor in ``svg`` format. You
+can use ``engine/icons/inkscape_template.svg`` with default icon properties
+already set up.
+
+Once you're satisfied with the icon's design, save the icon in
+``engine/icons/svg/`` folder. But in order for the engine to automatically
+pick up the icons, each icon's filename:
+
+1. Must be prefixed with ``icon_``.
+
+2. ``PascalCase`` name should be converted to ``snake_case``, so words
+   are separated by ``_`` whenever case changes, and uppercase acronyms must
+   also have all letters, numbers, and special characters separated as distinct
+   words. Some examples:
+
+   +--------------------+----------------------------------+
+   | Name               | Filename                         |
+   +====================+==================================+
+   | ``Polygon2D``      | ``icon_polygon_2_d.svg``         |
+   +--------------------+----------------------------------+
+   | ``CSGPolygon``     | ``icon_c_s_g_polygon.svg``       |
+   +--------------------+----------------------------------+
+   | ``CPUParticles2D`` | ``icon_c_p_u_particles_2_d.svg`` |
+   +--------------------+----------------------------------+
+   | ``C#``             | ``icon_c_#.svg``                 |
+   +--------------------+----------------------------------+
+
+Icon optimization
+~~~~~~~~~~~~~~~~~
+
+Because the editor renders the ``svg``'s at runtime, they need to be small
+in size, so they can be efficiently parsed. Editor icons must be first
+optimized before being added to the engine, to do so:
+
+1. Add them to the ``engine/icons/svg`` folder.
+
+2. Run the ``optimize.py`` script. You must have the ``scour`` package installed:
+
+   .. code:: bash
+
+       pip install scour
+       cd godot-design/engine/icons && ./optimize.py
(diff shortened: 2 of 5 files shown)


development/cpp/custom_modules_in_cpp.rst
```diff
@@ -408,6 +408,35 @@ the docs will become accessible within the engine's built-in documentation syste
...
 
+.. _doc_custom_module_icons:
+
+Adding custom editor icons
+--------------------------
+
+Similarly to how you can write self-contained documentation within a module,
+you can also create your own custom icons for classes to appear in the editor.
+
+For the actual process of creating editor icons to be integrated within engine,
```
akien-mga: within *the* engine

development/editor/creating_icons.rst
```diff
@@ -0,0 +1,83 @@
+.. _doc_editor_icons:
+
+Editor icons
+============
+
+When a new class is created and exposed to scripting, the editor's interface
+will display it with a default icon representing the base class it's inherited
```
akien-mga: "it inherits from" maybe?

development/editor/creating_icons.rst
```diff
@@ -0,0 +1,83 @@
...
+from. Yet in most cases it is recommended to create icons for new classes
+to improve the user experience considerably.
```
akien-mga: "considerably" is likely not needed here.

development/editor/creating_icons.rst
```diff
@@ -0,0 +1,83 @@
...
+
+Creating icons
+~~~~~~~~~~~~~~
+
+In order to create new icons, you first need a vector graphics editor installed.
+For instance, you can use open-source `Inkscape <https://inkscape.org/>`_ editor.
```
akien-mga: *the* open-source

development/editor/creating_icons.rst
```diff
@@ -0,0 +1,83 @@
...
+
+Clone the ``godot-design`` repository containing all the original editor icons:
+
+   .. code:: bash
+
+       git clone https://github.com/godotengine/godot-design
+
+The icons must be created in a vector graphics editor in ``svg`` format. You
+can use ``/engine/icons/inkscape_template.svg`` with default icon properties
```
akien-mga: I'd drop the leading `/` which would bring users to their system root if they copy it.

development/editor/creating_icons.rst
```diff
@@ -0,0 +1,83 @@
...
+already set up.
+
+Once you're satisfied with the icon's design, save the icon in
+``engine/icons/svg/`` folder. But in order for the engine to automatically
+pick up the icons, each icon's filename:
+
+1. Must be prefixed with ``icon_``.
+
+2. Each consecutive capitalized letter in the class name must be converted to
```
akien-mga: That's not really true, as this would lead to `iconc_s_g_p_olygon`. It's correct if you say "prefix" instead of "suffix", but it's still hard to understand IMO.

It would be better to say that the "PascalName" class name should be converted to "snake_case", so words are separated by `_` instead of case changes, and uppercase acronyms also have all letters separated as distinct words. You can give two polygons (one with an acronym and one without -- but maybe with numbers) to make it clearer.

development/editor/creating_icons.rst
```diff
@@ -0,0 +1,83 @@
...
+   lowercase with ``_`` suffix.
+
+So in the case of ``CSGPolygon`` class, the icon's filename should be
+``icon_c_s_g_polygon.svg``.
+
+Icon optimization
+~~~~~~~~~~~~~~~~~
```
akien-mga: Style wise, there should be an empty line after this.

development/editor/creating_icons.rst
```diff
@@ -0,0 +1,83 @@
...
+
+   .. code:: bash
+
+       pip install scour
+       cd godot-design/engine/icons && ./optimize.py
+
+The optimized icons will be generated in ``/engine/icons/optimized`` folder.
+
+Integrating and sharing the icons
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
akien-mga: Same here.

development/editor/creating_icons.rst
```diff
@@ -0,0 +1,83 @@
...
+It's also possible to create custom icons within a module. If you're creating
+your own module and don't plan to integrate it with Godot, you don't need to
+make a separate pull request for your icons to be available within the editor
+as they can be self-contained.
+
+For specific instructions on how to create module icons, refer to
+:ref:`Creating custom module icons<doc_custom_module_icons>`.
+
+Troubleshooting
+~~~~~~~~~~~~~~~
```
akien-mga: And here.

development/editor/creating_icons.rst
```diff
@@ -0,0 +1,83 @@
...
+So in the case of ``CSGPolygon`` class, the icon's filename should be
+``icon_c_s_g_polygon.svg``.
+
+Icon optimization
+~~~~~~~~~~~~~~~~~
+Because the editor renders the ``svg``'s at runtime, they need to be small
+in size, so they can be efficiently parsed. Editor icons must be first
+optimized before being added to the engine, to do so:
+
+1. Add them to the ``/engine/icons/svg`` folder.
```
akien-mga: Remove the leading `/`.

development/editor/creating_icons.rst
```diff
@@ -0,0 +1,83 @@
...
+
+2. Run the ``optimize.py`` script. You must have ``scour`` package installed:
```
akien-mga: *the* `scour` package

development/editor/creating_icons.rst
```diff
@@ -0,0 +1,83 @@
...
+
+   .. code:: bash
+
+       pip install scour
+       cd godot-design/engine/icons && ./optimize.py
+
+The optimized icons will be generated in ``/engine/icons/optimized`` folder.
```
akien-mga: *the* ... folder. Drop the leading `/`.

development/editor/creating_icons.rst
```diff
@@ -0,0 +1,83 @@
...
+
+Integrating and sharing the icons
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+If you're contributing to the engine itself, you should make a pull request
+to ``godot-design`` repository first. Once it's merged, make another pull
```
akien-mga: *the* `godot-design` repository.

akien-mga: In practice the workflow is more the other way around though, we merge first in the main godot repo if the change is wanted, and then the original version can be added to godot-design.

development/editor/introduction.rst
```diff
@@ -0,0 +1,9 @@
+.. _doc_editor_intro:
+
+Introduction
+============
+
+As the engine is constantly evolving and new features are being added, it is
+important to familiarize yourself with the editor and tools development in C++.
+Without the editor, some of the engine's internal functionality would become
+unaccessible, make the development more cumbersome or not intuitive.
```
akien-mga: As an editor-based engine, it's a bit tautological to say that it would be more cumbersome without editor though :P

akien-mga:
Looks good overall! Sorry for the late review.

development/editor/creating_icons.rst
Xrayez: It's interesting how some icons are registered just fine for digits with and without `_`:
`icon_area_2d.svg`
`icon_bone_2_d.svg`

I guess people just use `PascalCase` → `snake_case` converter which leads to this inconsistency.
The documentation should accept the style dictated by these converters then.

development/editor/creating_icons.rst
Xrayez: I don't see a reason why "in theory" workflow should be described at all then, but perhaps "ideally" is the right word for this.

development/editor/introduction.rst
Xrayez: If you think about it "as-if", it makes it more logical. :)

Wonder whether the whole sentence could be dropped, or just remove `cumbersome`.

Xrayez:
Added a table to show more conversion examples (namely the usage of numbers and special characters as in `C#`).

akien-mga:
Thanks!

Issue #2583, opened 2019-07-03
Extend GDScript Basics with a looped Coroutine


HaSa1002:
I just searched over an hour to find a solution, why coroutines don't work with loops. As I now found out you have to save the new state explicitly. What is different from coroutines I know.

I'd suggest to add an example to the reference (I know it is stated there) and something like the following snippet into the GDScript Basics:
Remember to save the new function state when using multiple yields per funtion:
```
func co_func():
    for i in range(1, 5):
        #Do stuff
        yield()

func _ready():
    var co = co_func()
    while co is GDScriptFunctionState && co.is_valid():
        co = co.resume()
```

Xrayez:
Might also be worth adding a warning about yielding to signals with looping.
https://github.com/godotengine/godot/issues/30269#issuecomment-508260002:
> Also, you don't need to use while, the completed signal is only emitted when the function didn't yield anymore.

Namely:
```gdscript
while co is GDScriptFunctionState::
        co = yield(co, 'completed')
```
is redundant, simple `if` statement is enough:
```gdscript
if co is GDScriptFunctionState::
        co = yield(co, 'completed')
```
(I'd use `while` for consistency though, ehehe).

HaSa1002:
I don't get it. I have to `while` there, 'cause there is more than one `yield`?

Xrayez:
In case of "yielding to a signal", the idea is that you *don't* have to use a while loop. Not sure whether the same could be applied to other types of `yield`ing. In your case you could try the same technique (second snippet). So `yield(co, 'completed')` could essentially replace the need for a while loop.

HaSa1002:
@Xrayez is the warning still required? If so what kind of warning should that be? How should I word the warning. I'm not completly getting how to use the concept and since I don't know I don't know what to write. Personally I think a warning might be useful to tell people that this bug exists in versions prior to 3.2.

Xrayez:
@HaSa1002 I felt like this deserves another PR so I wrote it separately in the next section actually (Coroutines & signals): #2644.

HaSa1002:
Thinking about your example a couple of days now, I realize what confused me: The usage of yield to resume a coroutine. That brings me that, I should update the example to be a bit bigger to show how to use a looped coroutine i.e. with the _process function to not block the the games execution.

Xrayez:
[...]

PR #2586, opened 2019-07-03
Added looped Coroutine Example to GDScript Basics to improve understanding of the coroutine concept

HaSa1002:
I added an example for using multiple yields() as I proposed in #2583.


Calinou:
Thanks!

PR #2587, opened 2019-07-04
Updated OS.get_unique_id() platform availability notice

InsightGit:
In Godot 3.0, OS.get_unique_id() only supported iOS and Android, but in Godot 3.1 onwards, OS.get_unique_id() supports all platforms except for UWP and HTML5. This PR updates the docs to reflect that change made in Godot 3.1.


akien-mga:
Thanks!

PR #2591, opened 2019-07-04
Compiling with Mono: Simplify and update Android instructions

neikeq:
Updated Android instructions in `Compiling with Mono`.
Removed section about building the C# API to be bundled with the editor as SCons does that automatically now.


akien-mga:
Thanks!

PR #2594, opened 2019-07-05
Clarify that your game need not be MIT

remram44:
This page made it sound like your game was "the Software" in the license text, and that MIT licensing was required. Whereas it only applies to Godot and not the game.

Legal is hard, hopefully this helps people understand their rights? Could probably be better.

Also note that the two "Copyright (c) ..." lines actually appear on the same line on RTD, maybe `<br>` would help?


Sslaxx:
Having just seen someone comment on a FB post that "All godot games are free license and open source" does rather reinforce the need for this in my view.

tutorials/legal/complying_with_licenses.rst
```diff
@@ -31,6 +31,8 @@ This text reads as follows:
 	THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 
+Note that your game itself does not need to be under the same license, or even an open license.
```
NathanLovato: Would you mind using a `.. note::` block so that it's more visible? I would also emphasize this a little more, e.g.

```
Your games do not need to be under the same license. You are free to release your Godot projects under any license and to create commercial games with the engine.
```

What do you think?

remram44: Sure! This note was supposed to be a clarification for the text of the license, but since I changed that it should be clear enough that it only applies to the engine. Thanks!

NathanLovato:
Thanks!

PR #2595, opened 2019-07-06
[Minor] Fix "it's" that should be "its"

remram44:
I saw some "it's" that should have been "its", so I went over the whole doc. I actually found very few of those, so I guess I'm just overly pedant, but thought I'd send the patch anyway.

Cheers!

### DIFF
--- a/tutorials/shading/your_first_shader/your_second_spatial_shader.rst
+++ b/tutorials/shading/your_first_shader/your_second_spatial_shader.rst
@@ -83,8 +83,8 @@ a low ``METALLIC`` has a more equal representation of sky color and ``ALBEDO`` c
 .. note:: ``METALLIC`` should be close to ``0`` or ``1`` for proper PBR shading. Only set it between
           them for blending between materials.
 
-Water is not a metal, so we will set its ``METALLIC`` property to ``0.0``. But, water is also highly
-reflective, so we will set its ``ROUGHNESS`` property to by quite low as well.
+Water is not a metal, so we will set its ``METALLIC`` property to ``0.0``. Water is also highly
+reflective, so we will set its ``ROUGHNESS`` property to be quite low as well.
 
 .. code-block:: glsl
 
--- a/tutorials/shading/your_first_shader/your_second_spatial_shader.rst
+++ b/tutorials/shading/your_first_shader/your_second_spatial_shader.rst
@@ -83,8 +83,8 @@ a low ``METALLIC`` has a more equal representation of sky color and ``ALBEDO`` c
 .. note:: ``METALLIC`` should be close to ``0`` or ``1`` for proper PBR shading. Only set it between
           them for blending between materials.
 
-Water is not a metal, so we will set it's ``METALLIC`` property to ``0.0``. But, water is also highly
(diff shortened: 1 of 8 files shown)


classes/class_csgshape.rst
```diff
@@ -69,7 +69,7 @@ enum **Operation**:
 
 - **OPERATION_INTERSECTION** = **1** --- Only intersecting geometry remains, the rest is removed.
 
-- **OPERATION_SUBTRACTION** = **2** --- The second shape is susbtracted from the first, leaving a dent with it's shape.
+- **OPERATION_SUBTRACTION** = **2** --- The second shape is susbtracted from the first, leaving a dent with its shape.
```
cbscribe: "susbtracted" is also wrong.

tutorials/shading/your_first_shader/your_second_spatial_shader.rst
```diff
@@ -83,8 +83,8 @@ a low ``METALLIC`` has a more equal representation of sky color and ``ALBEDO`` c
 .. note:: ``METALLIC`` should be close to ``0`` or ``1`` for proper PBR shading. Only set it between
           them for blending between materials.
 
-Water is not a metal, so we will set it's ``METALLIC`` property to ``0.0``. But, water is also highly
-reflective, so we will set it's ``ROUGHNESS`` property to by quite low as well.
+Water is not a metal, so we will set its ``METALLIC`` property to ``0.0``. But, water is also highly
```
cbscribe: Can you also remove that , after "But"?

tutorials/shading/your_first_shader/your_second_spatial_shader.rst
```diff
@@ -83,8 +83,8 @@ a low ``METALLIC`` has a more equal representation of sky color and ``ALBEDO`` c
...
+reflective, so we will set its ``ROUGHNESS`` property to by quite low as well.
```
cbscribe: by -> be

tutorials/shading/your_first_shader/your_second_spatial_shader.rst
remram44: Hmm, why?

cbscribe: Because "But," at the front of a sentence should not be used unless it is followed by a prepositional phrase.

akien-mga:
Files in `classes/` should not be modified, they are generated from the XML in https://github.com/godotengine/godot/tree/master/doc/classes

remram44:
Ok, thank you both! I think I have addressed your comments.

cbscribe:
Can you squash these into a single commit, please?

remram44:
Sure! Though if what you want is a single commit, GitHub has a button for that 😉

NathanLovato:
> Sure! Though if what you want is a single commit, GitHub has a button for that wink

The squash feature on git has some limitations. We can use it here from time to time as the docs' history isn't super important, but e.g. not on the Godot repo as you won't get a merge commit, among other things that make a difference.

NathanLovato:
Thanks for the help!

Issue #2592, opened 2019-07-05
Revise Best Practices


TheDuriel:
@NathanLovato heres your ping.

below is an excert of the discord discussion regarding my issues with the new best practices section. overall its a great addition to the docs, but i think its missing out on explaining some vital principles for tree navigation, and features bad code examples.

unfiltered, raw, copy and paste:

the scene organization page... honestly i'd like it nuked and reworked from scratch using https://cdn.discordapp.com/attachments/319225525052899328/585172151913676800/HowToSceneTree.jpg

especially the funref example in that is terrible, and listing it above node paths is weird.

there are tons of examples of get_parent or ".." in the best practices section. getting a parent node is never a good practice unless you are specifically making something like a tooltip ui element. (where you can get the parent so the tooltip can position itself correctly.)

the godot interfaces page should make it explicit that it is not talking about Nodes, but Reference types in the funref example. and again, its using the scene tree badly with get_child(0) and get_parent() calls.

[12:38 PM] TheDuriel [TTG]: http://docs.godotengine.org/en/latest/getting_started/workflow/best_practices/scene_organization.html
http://docs.godotengine.org/en/latest/getting_started/workflow/best_practices/godot_interfaces.html
relevant pages
[12:42 PM] TheDuriel [TTG]: calling out @willnationsdev since he wrote the page. and @gdquest cuz you're also a big docs contributer :P
[12:52 PM] TheDuriel [TTG]: (just to be clear. the rest of the best practices is fairly good and i'm pretty happy with it)
[2:02 PM] willnationsdev: I just wrote up my own thoughts and no one prompted me to change those things in the PR. Feel free to submit a PR to improve it! That's the whole reason I put the info in the publicly maintained docs rather than just my own blog or something. I'm not gonna think of/know everything. :-)  @TheDuriel [TTG]
[2:03 PM] willnationsdev: Happy to hear your feedback on it
[2:19 PM] willnationsdev: I would recommend that if new docs for those are made, they should highlight all the options available to users (as I tried to do) and then recommend what out of those options is best for different situations, that way we can have comprehensive information.
[2:20 PM] willnationsdev: I would recommend that if new docs for those are made, they should highlight all the options available to users (as I tried to do) and then recommend what out of those options is best for different situations, that way we can have comprehensive information.
[2:21 PM] willnationsdev: Not sure if the existing ones need nuking, as you say, but it if there is faulty info, we should definitely amend or remove it.
[...]

PR #2597, opened 2019-07-06
Simplify best practices intro and "what are Godot classes?"

NathanLovato:
I'm trying to take all the valuable information in the pages and to make it accessible to more people and easier to remember. It looks like I rewrote everything because there was room to simplify most sentences and to clarify the article's structure and focus, but hopefully, the content is still there.

Main changes:

- Add structure to the article, a clear taught point that gets stressed in the intro and the summary at the end, to hopefully get the point across
- Remove extra adverbs, parentheses, and phrasing that lengthen sentences
- Use simpler computer science or programming terms
- Favor direct voice, present tense

I think there'll be more room to improvement for each page in this series, especially as the examples can be quite abstract. If this can work for an experienced developer, in my experience, it makes tutorial unaccessible to most amateur gamedevs. This would be a pity.

### DIFF
--- a/getting_started/workflow/best_practices/introduction_best_practices.rst
+++ b/getting_started/workflow/best_practices/introduction_best_practices.rst
@@ -3,22 +3,27 @@
 Introduction
 ============
 
-This tutorial series aims to illustrate intended Godot workflows, i.e. the
-"best practices" of defining one's game API. Godot allows for a great 
-amount of flexibility in how one designs a project's code and scene
-structure. Each method will have its own advantages and disadvantages,
-so outlining the best practices should help users feel less confused
-as they further explore Godot's depth.
+This series is a collection of best practices to help you work efficiently with
+Godot.
 
-This series is also designed to help get users thinking about the best way
-to reason about best practices. It would be impossible to give a comprehensive
-overview of what one should do for any arbitrary design decision. As such,
-each article breaks problems down into the fundamental questions they pose,
-suggests the available solutions, analyzes the advantages and disadvantages
-of each option, and then highlights the best course of action depending on the
-circumstances.
+Godot allows for a great amount of flexibility in how you structure a project's
+codebase and break it down into scenes. Each approach has its own pros and
+cons, and they can be hard to weigh until you've worked with the engine for long enough.
 
-While the articles in this series are largely topic-directed, it is
-recommended that users begin with the :ref:`doc_what_are_godot_classes` article.
-It is from there that the "best practices" for the rest of the engine
-become more clear, based on established OOP practices.
+There are always many ways to structure your code and solve specific programming
+problems. It would be impossible to cover them all here.
+
+That is why each article starts from a real-world problem. We will break down
+each problem in fundamental questions, suggest solutions, analyze the pros and
+cons of each option, and highlight the best course of action for the problem at hand.
+
+You should start by reading :ref:`doc_what_are_godot_classes`. It explains how
+Godot's nodes and scenes relate to classes and objects in another
+Object-Oriented programming language. It will help you make sense of the rest of the series.
+
+.. note::
+
+   The best practices in Godot rely on Object-Oriented design principles. We
+   use tools like the `single responsibility
+   <https://en.wikipedia.org/wiki/Single_responsibility_principle>`_ principle and
+   `encapsulation <https://en.wikipedia.org/wiki/Encapsulation_(computer_programming)>`_.
--- a/getting_started/workflow/best_practices/what_are_godot_classes.rst
+++ b/getting_started/workflow/best_practices/what_are_godot_classes.rst
@@ -1,80 +1,91 @@
 .. _doc_what_are_godot_classes:
 
-What are Godot classes really?
-==============================
+Godot scenes and scripts are classes
+====================================
 
-Godot offers two main means of creating types: scripts and scenes.
-Both of these represent a "class" since Godot revolves around
-Object-Oriented design. *How* they do this may not be clear to beginner
-or intermediate users though.
+In Godot, scripts and scenes can both be the equivalent of classes in an
+Object-Oriented programming language. The main difference is that scenes are
+`declarative code <https://en.wikipedia.org/wiki/Declarative_programming>`_,
+while scripts can contain `imperative code
+<https://en.wikipedia.org/wiki/Imperative_programming>`_.
 
-Godot Engine provides classes out-of-the-box (like
-:ref:`Node <class_Node>`), but user-created types are not actually classes.
-Instead they are resources that tell the engine a sequence of initializations
-to perform on an engine class.
+As a result, many best practices in Godot boil down to applying Object-Oriented
+design principles to the scenes, nodes, or script that make up your game.
 
-Godot's internal classes have methods that register a class's data with
-a :ref:`ClassDB <class_ClassDB>`. This database provides runtime access to
-class information (also called "reflection"). Things stored in the ClassDB
-include, among other things...
+This guide explains how scripts and scenes work in the engine's core, to help
+you get a sense of how Godot works under the hood, and to help you better
+understand where some of this series' best practices come from.
+
+Making sense of classes in Godot
+--------------------------------
+
+Godot Engine provides built-in classes like :ref:`Node <class_Node>`.
+User-created types are not technically classes. Instead, they are resources that
+tell the engine a sequence of initializations to perform on one of the engine's
+built-in classes.
+
+Godot's internal classes have methods that register a class's data with a
+:ref:`ClassDB <class_ClassDB>`. This database provides runtime access to class
+information. ``ClassDB`` contains information about classes like:
 
 - properties
-
 - methods
-
 - constants
-
 - signals
 
-Furthermore, this ClassDB is what Objects actually check against when
-performing any operation. Access a property? Call a method? Emit a signal?
-It will check the database's records (and the records of the Object's base
-types) to see if the Object supports the operation. Every C++ Object defines
-a static `_bind_methods()` function that describes what C++ content it
-registers to the database and how.
+This ``ClassDB`` is what Objects check against when performing an operation like
+accessing a property or calling a method. ``ClassDB`` checks the database's
+records and the records of the Object's base types to see if the Object supports
+the operation.
 
-So, if the engine provides all of this data at startup, then how does
-a user define their own data? It'd be nice if users could define a custom
-set of data to be appended to an object's data. That way, users could inject
-their own properties and methods into the engine's Object query requests.
+On the engine's side, every class defines a static ``_bind_methods()`` function
+that describes what C++ content it registers to the database and how. When you
+use the engine, you can extend the methods, properties, signals available from
+the ``ClassDB`` by attaching a :ref:`Script <class_Script>` to your node.
 
-*This* is what a :ref:`Script <class_Script>` is. Objects check their attached
-script before the database, so scripts can even override methods.
-If a script defines a `_get_property_list()` method, that data is appended to
-the list of properties the Object fetches from the ClassDB. The same holds
-true for other declarative code.
+Objects check their attached script before the database. This is why scripts can
+override built-in methods. If a script defines a ``_get_property_list()`` method,
+Godot appends that data to the list of properties the Object fetches from the
+ClassDB. The same is true for other declarative code.
 
-This can lead to some users' confusion when they see a script as being
-a class unto itself. In reality, the engine just auto-instantiates the
-base engine class and then adds the script to that object. This then allows
-the Object to defer to the Script's content where the engine logic deems
-appropriate.
+Even scripts that don't inherit from a built-in type, i.e. scripts that don't
+start with the ``extends`` keyword, implicitly inherit from the engine's base
+:ref:`Reference <class_Reference>` class. This allows the Object to defer
+to the script's content where the engine logic deems appropriate.
 
-A problem does present itself though. As the size of Objects increases,
-the scripts' necessary size to create them grows much, much larger.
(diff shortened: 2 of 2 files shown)


akien-mga:
Note for reviewers: I advise to use the "Split Diff" view instead of the default "Unified Diff" view for such changes, so that you can compare the before and after without having their flow broken.

Screenshot_20190717_084112

getting_started/workflow/best_practices/what_are_godot_classes.rst
```diff
@@ -1,80 +1,85 @@
...
-Creating node hierarchies demonstrates this. Each individual Node's logic
-could be several hundred lines of code in length.
-
-let's see a simple example of creating a single Node as a child.
+This ``ClassDB`` is what Objects check against when performing an operation like
+accessing a property or calling a method. ``ClassDB`` checks the database's
+records and the records of the Object's base types to see if the Object supports
+the operation.
+
+On the engine's side, every class defines a static `_bind_methods()` function
```
akien-mga: Should be double backticks in rst.

getting_started/workflow/best_practices/what_are_godot_classes.rst
```diff
@@ -1,80 +1,85 @@
...
+that describes what C++ content it registers to the database and how. When you
+use the engine, you can extend the methods, properties, signals available from
+the ``ClassDB`` by attaching a :ref:`Script <class_Script>` to your node.
+
+Objects check their attached script before the database. This is why scripts can
+override built-in methods. If a script defines a `_get_property_list()` method,
+Godot appends that data to the list of properties the Object fetches from the
+ClassDB. The same is true for other declarative code.
+
+Scripts are not classes in themselves. Even scripts that don't inherit from a
```
akien-mga: I think I understand what the second sentence intends to convey, but it might be a bit confusing. IIRC a script without `extends SomeClass` actually behaves the same as if it had `extends Reference`, which is implicit. So in itself a script cannot not inherit a built-tin type, it's always an `Object` (`Reference` if unspecified, otherwise most likely `Node` or more specific, or a `Resource` type).

akien-mga: Confirmed that a script without `extends` inherits `Reference` implicitly:
Screenshot_20190717_085320
Which also means that it can't be attached to a `Node`, but it can be instanced from code and used as a `Reference`.

getting_started/workflow/best_practices/what_are_godot_classes.rst
```diff
@@ -1,80 +1,85 @@
...
+records and the records of the Object's base types to see if the Object supports
+the operation.
+
+On the engine's side, every class defines a static `_bind_methods()` function
+that describes what C++ content it registers to the database and how. When you
+use the engine, you can extend the methods, properties, signals available from
+the ``ClassDB`` by attaching a :ref:`Script <class_Script>` to your node.
+
+Objects check their attached script before the database. This is why scripts can
+override built-in methods. If a script defines a `_get_property_list()` method,
```
akien-mga: Should be double backticks here too.

getting_started/workflow/best_practices/what_are_godot_classes.rst
```diff
@@ -1,80 +1,85 @@
...
+Let's see a simple example of creating a single ``Node`` as a child. The code
+below creates a new ``Node``, changes its name, assigns a script to it, sets its
+future parent as its owner so it gets saved to disk along with it, and finally
+adds it as a child of the ``Main`` node:
 
 .. tabs::
   .. code-tab:: gdscript GDScript
 
-    # main.gd
+    # Main.gd
```
akien-mga: The code below refers to `"child.gd"`, so this should be made consistent (either both capitalized, or both lower case). Do we have any preference for a given style in the style guide?

getting_started/workflow/best_practices/what_are_godot_classes.rst
```diff
@@ -86,53 +91,43 @@ let's see a simple example of creating a single Node as a child.
...
             public Main()
             {
-                Child = new Node(); // Construct the child.
-                Child.Name = "Child"; // Change its name.
-                Child.Script = (Script)ResourceLoader.Load("child.gd"); // Give it custom features.
-                Child.Owner = this; // Serialize this node if this is saved.
-                AddChild(Child); // Add "Child" as a child of this.
+                Child = new Node();
+                Child.Name = "Child";
+                Child.Script = (Script)ResourceLoader.Load("child.gd");
```
akien-mga: It would be nice to have an example relying on a `.cs` script, CC @neikeq. Would `Child.Script = (Script)ResourceLoader.Load("child.cs");` work in theory?

getting_started/workflow/best_practices/what_are_godot_classes.rst
```diff
@@ -86,53 +91,43 @@ let's see a simple example of creating a single Node as a child.
...
+makes a separate call to the scripting API which leads to many "look-ups" on the
+back-end to find the logic to execute.
 
-Now, why is any of this important to scene organization? Because one must
-understand that scenes *are* objects. One often pairs a scene with
-a scripted root node that makes use of the sub-nodes. This means that the
-scene is often an extension of the script's declarative code.
+Scenes help to avoid this performance issue. :ref:`PackedScene
+<class_PackedScene>`, the base type that scenes inherit from, are resources that
+uses serialized data to create objects. The engine can process scenes in batches
```
akien-mga: are resource**s** that **use** (plural)

getting_started/workflow/best_practices/what_are_godot_classes.rst
```diff
@@ -86,53 +91,43 @@ let's see a simple example of creating a single Node as a child.
...
+objects. One often pairs a scene with a scripted root node that makes use of the
+sub-nodes. This means that the scene is often an extension of the script's
+declarative code.
 
-- how are they organized?
+The content of a scene helps to define:
 
-- how are they initialized?
+- What nodes are available to the script
+- How are they organized
```
akien-mga: Shouldn't it be "How they are organized" if not an interrogation?

getting_started/workflow/best_practices/what_are_godot_classes.rst
```diff
@@ -86,53 +91,43 @@ let's see a simple example of creating a single Node as a child.
...
-- what connections to each other do they have, if any?
+Many Object-Oriented principles which apply to written code *also* apply to
+scenes.
 
-As such, many Object-Oriented principles which apply to "programming", i.e.
-scripts, *also* apply to scenes. Some scripts are designed to only work
-in one scene (which are often bundled into the scene itself). Other scripts
-are meant to be re-used between scenes.
+The scene is *always an extension of the script attached to its root node*. You
+ can see all the nodes it contains as part of a single class.
```
akien-mga: Leading space here.

getting_started/workflow/best_practices/introduction_best_practices.rst
```diff
@@ -3,22 +3,27 @@
...
+each problem in fundamental questions, suggest solutions, analyze the pros and
+cons of each option, and highlight the best course of action for the problem at hand.
+
+You should start by reading :ref:`doc_what_are_godot_classes`. It explains how
+Godot's nodes and scenes relate to classes and objects in another
+object-oriented programming language. It will help you make sense of the rest of the series.
+
+.. note::
+
+   The best practices in Godot rely on object-oriented design principles. We
```
akien-mga: In the other page you always capitalized "Object-Oriented", should it be the same here?

akien-mga:
I left a few comments, but otherwise it's a pretty good improvement, should be good to merge.

getting_started/workflow/best_practices/what_are_godot_classes.rst
neikeq: Assigning a C# script will dispose the existing managed instance:

``` C#
var node = new Node();
ulong nodeId = node.GetInstanceId();

node.Script = ResourceLoader.Load<Script>("MyCustomNode.cs");

// Assert this managed instance was just disposed
Debug.Assert(node.NativeInstance == IntPtr.Zero);

// The native instance now has a new managed instance of type MyCustomNode
node = (MyCustomNode) GD.InstanceFromId(nodeId);

AddChild(node);
```

The reason this works in GDScript is because there's no such managed instance. It's just a `Variant` holding a pointer to the native instance.

If the type of the managed instance is an engine type (no custom derived class) and you're assigning a non-C# script, then it should work (this is what the example above does):

``` C#
var node = new Node();

node.Script = ResourceLoader.Load<Script>("child.gd");

// Assert this managed instance was NOT disposed
Debug.Assert(node.NativeInstance != IntPtr.Zero);

AddChild(node);
```

However, if type of the managed instance is a custom derived class, then the instance is again disposed:

``` C#
var node = new MyCustomNode();
ulong nodeId = node.GetInstanceId();

node.Script = ResourceLoader.Load<Script>("child.gd");

// Assert this managed instance was just disposed
Debug.Assert(node.NativeInstance == IntPtr.Zero);

// The native instance now has a new managed instance of whatever engine type "child.gd" inherits
node = (MyCustomNode) GD.InstanceFromId(nodeId);

// Assuming "child.gd" inherits Node2D
Debug.Assert(node.GetType() == typeof(Node2D));

AddChild(node);
```

I've yet to document this behavior, and maybe I could improve it so it's less verbose than using `InstanceId`.

getting_started/workflow/best_practices/what_are_godot_classes.rst
NathanLovato: My bad, it's a typo. I'd always go for PascalCase for anything that represents a class, especially as Godot generates file names like that for nodes, scenes, and gdscript files. 
But iirc there's no guideline set in stone regarding the file structure, even though we had some discussions about that years ago. Be it naming or project organization.

NathanLovato:
Just fixed everything, except for the c# example. I could use a hand there as I have no experience with c#

getting_started/workflow/best_practices/what_are_godot_classes.rst
akien-mga: Thanks for the detailed explanation. I guess for the purpose of this tutorial explaining those details about the managed instance might be a bit out of scope, so we can probably be lazy and keep things as is under the assumption that we're in the second case (type of the managed instance is an engine type (no custom derived class) and you're assigning a non-C# script).

Once this is documented in more details somewhere else, it might be worth adding a `.. note::` mention here to link to that documentation for details; or alternatively refactor the example to avoid having to discuss this at all here.

akien-mga:
Thanks!

willnationsdev:
Cool! I'm going to read through this while I'm at work later.

getting_started/workflow/best_practices/what_are_godot_classes.rst
```diff
@@ -1,80 +1,91 @@
...
-
-let's see a simple example of creating a single Node as a child.
+This ``ClassDB`` is what Objects check against when performing an operation like
+accessing a property or calling a method. ``ClassDB`` checks the database's
+records and the records of the Object's base types to see if the Object supports
+the operation.
+
+On the engine's side, every class defines a static ``_bind_methods()`` function
+that describes what C++ content it registers to the database and how. When you
+use the engine, you can extend the methods, properties, signals available from
```
willnationsdev: methods, properties, *and* signals

getting_started/workflow/best_practices/what_are_godot_classes.rst
```diff
@@ -86,53 +97,43 @@ let's see a simple example of creating a single Node as a child.
...
+back-end to find the logic to execute.
 
-Now, why is any of this important to scene organization? Because one must
-understand that scenes *are* objects. One often pairs a scene with
-a scripted root node that makes use of the sub-nodes. This means that the
-scene is often an extension of the script's declarative code.
+Scenes help to avoid this performance issue. :ref:`PackedScene
+<class_PackedScene>`, the base type that scenes inherit from, are resources that
+use serialized data to create objects. The engine can process scenes in batches
+on the back-end and provide much better performances than scripts.
```
willnationsdev: much better performance\* than scripts

getting_started/workflow/best_practices/what_are_godot_classes.rst
```diff
@@ -86,53 +97,43 @@ let's see a simple example of creating a single Node as a child.
...
+objects. One often pairs a scene with a scripted root node that makes use of the
+sub-nodes. This means that the scene is often an extension of the script's
+declarative code.
 
-- how are they organized?
+The content of a scene helps to define:
 
-- how are they initialized?
+- What nodes are available to the script
+- How they are organized
```
willnationsdev: This line and the next alternate the "they/are" order, but I'm not sure why and it kinda tripped up my reading a bit. I believe it should be "How they are initialized" in the next line.

getting_started/workflow/best_practices/what_are_godot_classes.rst
```diff
@@ -86,53 +97,43 @@ let's see a simple example of creating a single Node as a child.
...
-in one scene (which are often bundled into the scene itself). Other scripts
-are meant to be re-used between scenes.
+The scene is *always an extension of the script attached to its root node*. You
+can see all the nodes it contains as part of a single class.
 
-**Regardless, the scene is always an extension of the root script, and can
-therefore be interpreted as a part of the class.**
-Most of the points covered in this series will build on this point, so
-keep it in mind.
+Most of the tips and techniques explained in this series will build on his.
```
willnationsdev: build on this*

willnationsdev:
Okay. I know it's already been merged, but there's my feedback. Sorry, I didn't see this PR until after it had been mentioned in #2592 (which I'm subscribed to).

NathanLovato:
Thanks for the feedback and no worries, I have write access to the repo, so I could correct these mistakes already :)

PR #2598, opened 2019-07-06
[Minor] Just some small code corrections

Rudigus:
1º - Missing quotation marks in the first parameter of emit_signal() - the signal's name
2º - Damage calculation was resulting in negative numbers - If the text was printed, it would be, for example: "Player1 took -22 damage.", which is counterintuitive.

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_basics.rst
+++ b/getting_started/scripting/gdscript/gdscript_basics.rst
@@ -1529,7 +1529,7 @@ the :ref:`Object.connect() <class_Object_method_connect>` method:
 
         # We emit the health_changed signal every time the
         # character takes damage
-        emit_signal(health_changed", old_health, health)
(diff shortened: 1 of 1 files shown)


cbscribe:
This PR should be made against the master branch. See 
https://docs.godotengine.org/en/latest/community/contributing/documentation_guidelines.html#the-edit-on-github-link

Rudigus:
Thank you for the notice! I fixed it

NathanLovato:
Now there's an issue with your PR, you need to rebase your commit to master.
If you're having trouble cleaning up your branch, tell me, I can give you a hand. The easiest way to fix it would be to create a new branch from master, cherry-pick your commit with `git cherry-pick commit-id`, and to use this to open a new pull request.

akien-mga:
Merged manually on the master branch as 5067b9b58.

PR #2600, opened 2019-07-07
Update FreeBSD dependencies in "Compiling for X11"

Calinou:
`xineramaproto` is now `xorgproto` in recent FreeBSD versions. `libglapi` no longer seems to be available, but Godot still builds if you omit it.

This also adds `alsa-lib` and `pulseaudio` which are required to build editor and export template binaries.

Tested on FreeBSD 12.0 and 11.2.


akien-mga:
Thanks!

PR #2602, opened 2019-07-08
Made export template name match the default for Linux

Meriipu:
The default export template name for Linux appears to be `Linux/X11` and not `Linux X11`


akien-mga:
Thanks!

PR #2603, opened 2019-07-08
Fix small formatting issue in importing audio samples

frojo:
The bullet points don't render correctly without that blank line


NathanLovato:
thanks!

PR #2611, opened 2019-07-11
[Minor] Getting Started corrections

Rudigus:
1º - For some reason, even with the float type indicator in the export settings, it will display integer sliders for the array elements. Putting ".0" fixes that behavior.

2º - "frequecy" should be "frequency".


cbscribe:
Thanks!

PR #2624, opened 2019-07-15
Update intro to 3D images for Godot 3.1

skyace65:
Updates all the images on the intro to 3D page for Godot 3.1. All images have gone through png compression. One image wasn't updated because it was already made with 3.1. And I added an outline in another image for clarity.


akien-mga:
Thanks!

PR #2627, opened 2019-07-15
Remove outdated information from the HDR page

skyace65:
Currently the HDR page has outdated information on HDR parameters in the Environment resource. This PR removes that information and points to the environment and post processing page which has up to date information on the environment resource.


akien-mga:
Thanks!

PR #2628, opened 2019-07-16
Update environment and post processing page for Godot 3.1

skyace65:
Updates the images for the environment and post processing page for Godot 3.1. All images have gone through png compression.


clayjohn:
I think it determines how it interacts with the ambient occlusion texture assigned in the material. My understanding is that 0 negates the affect of the ao texture while 1 blends them together.

Calinou:
For comparison, this is with an Environment that has **Light Affect** set to 1 and **Ao Channel Affect** set to 1:

image

And **Light Affect** set to 1 but **Ao Channel Affect** set to 0:

image

The material on the left has its **Light Affect** set to 0, whereas the one on the right has its **Light Affect** set to 1.

skyace65:
Thanks to you two and some tests I did I think I've got a good description for what it does now. Thank you!

tutorials/3d/environment_and_post_processing.rst
```diff
@@ -256,6 +256,7 @@ Tweaking SSAO is possible with several parameters:
 - **Radius2/Intensity2:** A Secondary radius/intensity can be used. Combining a large and a small radius AO generally works well.
 - **Bias:** This can be tweaked to solve self occlusion, though the default generally works well enough.
 - **Light Affect:** SSAO only affects ambient light, but increasing this slider can make it also affect direct light. Some artists prefer this effect.
+- **Ao Channel Affect** If a value of zero is used the AO texture of an object will be used and SSAO will not be applied. Values greater than 0 multiply the AO texture by the SSAO to varying degrees. This does not affect objects without an AO texture.
```
Calinou:
```suggestion
- **Ao Channel Affect:** If a value of zero is used, only the AO texture of an object will be used; SSAO will not be applied. Values greater than 0 multiply the AO texture by the SSAO effect to varying degrees. This does not affect objects without an AO texture.
```

skyace65:
Ok, that should be fixed now.

akien-mga:
Thanks!

Issue #2605, opened 2019-07-09
Improve order of built-ins in shader reference docs


clayjohn:
Right now the order of the built-ins in the shader reference docs ([Spatial](http://docs.godotengine.org/en/latest/tutorials/shading/shading_reference/spatial_shader.html), [CanvasItem](http://docs.godotengine.org/en/latest/tutorials/shading/shading_reference/canvas_item_shader.html), [Particles](http://docs.godotengine.org/en/latest/tutorials/shading/shading_reference/particle_shader.html)) is a little random. It would be useful if someone could go through and re-order the entries to put related entries together. 

For example, as pointed out by a user on discord, in the [Spatial shader doc](http://docs.godotengine.org/en/latest/tutorials/shading/shading_reference/spatial_shader.html) It would be useful to put ``ALBEDO`` and ``OUTPUT_IS_SRGB`` together

PR #2631, opened 2019-07-16
Organized and updated spatial shader docs

### LINKED ISSUE #2606: Add "Shadow to Opacity" to Spatial Shader reference
https://github.com/godotengine/godot/pull/29031 Adds a new render mode, ``shadow_to_opacity``, and a new built-in, ``ALPHA``, to the light processor function. Both should be added to the [Spatial Shader doc](http://docs.godotengine.org/en/latest/tutorials/shading/shading_reference/spatial_shader.html)

_edit:_ Oh, and https://github.com/godotengine/godot/pull/24487 added ``UV`` and ``UV2`` to the the light processor function as well. So those two built-ins should be added.

vortexofdoom:
Fixes #2606. Also accomplishes a sub-portion of #2605, but I wanted to submit this first to make sure the organization is actually good. It's definitely more organized and consistent throughout, but there's probably still room for improvement. The grouping is probably more improved than the overall order.

### DIFF
--- a/tutorials/shading/shading_reference/spatial_shader.rst
+++ b/tutorials/shading/shading_reference/spatial_shader.rst
@@ -72,6 +72,10 @@ Render modes
 +---------------------------------+-----------------------------------------------------------------------+
 | **ambient_light_disabled**      | Disable contribution from ambient light and radiance map.             |
 +---------------------------------+-----------------------------------------------------------------------+
+| **shadow_to_opacity**           | Lighting modifies the alpha so shadowed areas are opaque and          | 
+|                                 | non-shadowed areas are transparent. Useful for overlaying shadows onto|
+|                                 | a camera feed in AR.                                                  |
++---------------------------------+-----------------------------------------------------------------------+
 
 Vertex built-ins
 ^^^^^^^^^^^^^^^^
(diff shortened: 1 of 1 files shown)


tutorials/shading/shading_reference/spatial_shader.rst
```diff
@@ -72,6 +72,8 @@ Render modes
 +---------------------------------+-----------------------------------------------------------------------+
 | **ambient_light_disabled**      | Disable contribution from ambient light and radiance map.             |
 +---------------------------------+-----------------------------------------------------------------------+
+| **shadow_to_opacity**           | Allows for overlaying shadows onto a camera feed (good for AR use).   |
```
clayjohn: Im not sure this description properly captures what shadow_to_opacity does. It needs to be more descriptive about how this renders non-shadowed areas transparent with shadowed areas opaque.

vortexofdoom:
"Lighting modifies the alpha so shadowed areas are opaque and non-shadowed areas are transparent. Useful for overlaying shadows onto a camera feed in AR."

Would that be better? A little bit on the nose with regard to your requested changes, but you probably said it as well as it could be said. I think I was trying to avoid getting too wordy, but most of the descriptions are so ridiculously terse I probably shouldn't have used them as a guideline. I was only going to squash changes once so I thought I'd ask first to make sure I didn't need to change it again.

clayjohn:
> Lighting modifies the alpha so shadowed areas are opaque and non-shadowed areas are transparent. Useful for overlaying shadows onto a camera feed in AR

Thats great! Descriptive explanations are definitely preferred. And its great to let people know the intended use case when there really is only one. Don't worry about the length.

akien-mga:
Thanks!

PR #2636, opened 2019-07-17
[WIP, need help] Update baked lightmaps page for Godot 3.1

skyace65:
I've gone through and updated all the images for Godot 3.1, and I've updated the Blender image to be of Blender 2.8. All images have gone through png compression.

There are two things I need help on, I've commented both in this PR. First I have no idea what the documentation is talking about when it says "And import normally. Remember you will need to set the texture size on the mesh after import." and I have no idea what the image below that line is for, or what you're supposed to do with those parameters. I'd like to clarify what it's talking about in this PR so if anyone knows your help would be really appreciated.

Second the bake settings might have been renamed. The docs refer to two settings, bake subdiv and capture subdiv, neither of those exist in 3.1. Instead we have Capture Cell size and Bake Cell size. And I checked Godot 3.0.6, it has the same name as in 3.1 so these must have been changed before 3.0 released. But I'm not sure if this is just a renamed setting because the default value for the subdiv's in the screenshot (256 and 128) is vastly different from the default cell size numbers (.25 and .5). So I don't know if the information on what the subdiv settings do is accurate for the cell size settings. If anyone could clarify this and tell me what still is and isn't accurate that would be very appreciated.


skyace65:
@reduz I was hoping you could answer some questions for me on baked light maps. First, what exactly are the Bake Cell Size and Capture Cell Size settings? Also do you know what the doc is referring to when it says “You will need to set the texture size on the mesh after import?”

reduz:
Oh, this is something that is more or less broken (changing the scale of the lightmaps), because the overall workflow kind of sucks.. the only way to get lightmaps generated is via scene import, I think importing objects does not work. This generates and unwraps UVs on the second channel of the object, which make it usable for lightmapping. It also saves the size the texture used for lightmapping that object should have.

reduz:
I am planning to rewrite most of this workflow for Godot 4.0 because its not very good as it is now.

PR #2640, opened 2019-07-18
Update GI probe page for Godot 3.2

skyace65:
Updates the images and info of the GI probe page for Godot 3.1. All images have gone through png compression.

The part I need help with is the compression setting for the GI probe. I assumed it just compressed the data so it would be smaller, and after some quick tests the compressed lighting data is clearly lower quality, but when saving the data as a .tres file compressed and uncompressed data has the same file size. So does compression work only if you don't save the data as a file?

Update: I updated one of the images so this is good for Godot 3.2. I've written that the compression setting is broken, based on testing I've done it is, currently it provides no benefit. I've opened a bug for it in the main Godot repository.


Calinou:
> but when saving the data as a .tres file compressed and uncompressed data has the same file size

GIProbe data is best saved in a binary `.res` file, as binary data is converted to Base64 when saved into a `.tres` file (or embedded in a `.tscn` scene).

skyace65:
Thanks for the info! I added info under data that if you're saving the data it should be saved as a .res. Unfortunately I did some more saving tests and even with the .res format the compression option doesn't do anything. If anyone knows if that's intentional or not please tell me.

skyace65:
@reduz I also had a question about the GI probe. How exactly does the compress setting work? I can see that quality is reduced but when saving the data, the size remains the same even if the compress option was used. Is it compressing but the benefit gets nullified if you save data? Also what exactly is compression doing?

akien-mga:
Thanks!

PR #2641, opened 2019-07-19
Bug in FPS tutorial (Part 5), if no key 'collider' is found in ray_result, program will terminate in error

chrisphyffer:
Bug in FPS tutorial (Part 5), if no key 'collider' is found in ray_result, program will terminate in error
The Error will be 'Invalid get index 'collider' (on base: Dictionary)


akien-mga:
The bug is actually the `ray_result != null` check. Reading the source code, `intersect_ray()` will always return a Dictionary, which can either be empty or it will contain all documented keys, including `collider`.

So the check should be `not ray_result.empty()`.

chrisphyffer:
Gotcha, I changed the condition from:
`f ray_result != null and "collider" in ray_result:`
`if not ray_result.empty() and "collider" in ray_result:`

Before adding anything to the docs in regards to scripting, i will check the source code first next time.

chrisphyffer:
(archiving - there is a bit more additions to the docs that i would like to add down the line)

PR #2644, opened 2019-07-20
Elaborate on yielding to `completed` signal in GDScript basics

Xrayez:
Also mention that one can get the signal's argument via yield once emitted.

Related to #2583 and in response to https://github.com/godotengine/godot-docs/issues/2583#issuecomment-513420432.

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_basics.rst
+++ b/getting_started/scripting/gdscript/gdscript_basics.rst
@@ -1714,6 +1714,39 @@ into an invalid state, for example:
 
 ``my_func`` will only continue execution once both buttons have been pressed.
 
+You can also get the signal's argument once it's emitted by an object:
+
+::
+
+    # Wait for when any node is added to the scene tree
+    var node = yield(get_tree(), "node_added")
+
+If you're unsure whether a function may yield or not, or whether it may yield
+multiple times, you can yield to the ``completed`` signal conditionally:
+
+::
+
+    func generate():
+        var result = rand_range(-1.0, 1.0)
+
+        if result < 0.0:
+            yield(get_tree(), "idle_frame")
+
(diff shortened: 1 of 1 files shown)


getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1714,6 +1714,27 @@ into an invalid state, for example:
...
+
+    # Wait for when any node is added to the scene tree
+    var node = yield(get_tree(), "node_added")
+
+If you're unsure whether a function may yield or not, or whether it may yield
+multiple times, you can yield to the ``completed`` signal conditionally:
+
+::
+
+    var ret = factory.generate()
```
akien-mga: Doesn't seem good to me in the GDScript basics tutorial as readers don't know what `factory` is, nor its `generate()` method.

Xrayez: Well that's kind of the point I guess, such usage should hide implementation details. I tried to extract some real use case from my own code base and the call trace goes quite deep. Here's some kind of abstract usage of it using built-in methods if that's enough:

```gdscript

func generate():
    var result = rand_range(-1.0, 1.0)
    
    if result < 0.0:
        yield(get_tree(), "idle_frame")
        
    return result
    
    
func make():
    var result = generate()

    if result is GDScriptFunctionState: # still working
        result = yield(result, "completed")
    
    return result
```

Xrayez:
This will be outdated if godotengine/godot#32034 gets merged.

mhilbrunner:
Well, it wasn't merged in the last 8 months. It may be merged tomorrow, or in a month or never - I think we'll go merging this small change and change it if needs be. Thanks! :)

PR #2648, opened 2019-07-22
Fix Main Menu tutorial issue

### LINKED ISSUE #2561: Confusing info in "Design a title screen" tutorial
https://docs.godotengine.org/en/3.1/getting_started/step_by_step/ui_main_menu.html#add-the-ui-sprites

Following the steps from the section above leads to this

image

instead of matching the image in the docs

image

That's what it should look like after setting up the automatic placing (which is in the next step).

Suggest either using the right screenshot or removing it.

skyace65:
Currently the main menu tutorial is inaccurate as it shows a finished main menu for when you've just added images to the TextureRect nodes, and haven't placed them into containers yet. This PR fixes that with an accurate image, the image has gone through png compression.
closes #2561


akien-mga:
Thanks!

PR #2649, opened 2019-07-22
Update GDScript int, float, and bool docs

aaronfranke:
It's useful to explain what C++ type they use in the background so that people with any questions can look up the equivalent type in C++ if they need more specific documentation. I did this previously with https://github.com/godotengine/godot-docs/pull/1891

Fixes https://github.com/godotengine/godot/issues/17224

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_basics.rst
+++ b/getting_started/scripting/gdscript/gdscript_basics.rst
@@ -349,18 +349,21 @@ be assigned any other value.
 bool
 ^^^^
 
-The Boolean data type can only contain ``true`` or ``false``.
+Short for "boolean", it can only contain ``true`` or ``false``.
 
 int
 ^^^
 
-The integer data type can only contain integer numbers (both negative
-and positive).
+Short for "integer", it stores whole numbers (positive and negative).
+It is stored as a 64-bit value, equivalent to "int64_t" in C++.
 
 float
 ^^^^^
 
-Used to contain a floating-point value (real numbers).
+Stores real numbers, including decimals, using floating-point values.
+It is stored as a 64-bit value, equivalent to "double" in C++.
+Note: Currently, data structures such as Vector2, Vector3, and
+PoolRealArray store 32-bit single-precision "float" values.
 
 :ref:`String <class_String>`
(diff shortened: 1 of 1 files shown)


getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -358,18 +358,19 @@ be assigned any other value.
...
-and positive).
+Short for "integer", it stores whole numbers (positive and negative).
+It is stored as a 64-bit value, equivalent to "int64_t" in C++.
 
 float
 ^^^^^
 
-Used to contain a floating point value (real numbers).
+Stores real numbers, including decimals, using floating-point values.
+It is stored as a 64-bit value, equivalent to "double" in C++.
```
akien-mga: Is that true? AFAIK floats are single precision currently.

aaronfranke: Currently, GDScript floats are doubles, but not within vectors or arrays. This does cause issues:

    var n = 0.1
    var v = Vector2(0.1, 0)
    print(n == v.x) # False

https://github.com/godotengine/godot/blob/master/core/variant.h#L144

cbscribe:
Looks like this has a conflict now. Can you please resolve?

getting_started/scripting/gdscript/gdscript_basics.rst
akien-mga: I think this should be mentioned here, as most users will expect that the floating point values they have in `var num : float` and `var vec : Vector2` are of the same type. Yet as you mentioned, that's not the case for vector types:
```gdscript
	var pi : float = 3.14159265358979323846
	print(pi)  # 3.141593
	print(pi * 1e10)  # 31415926535.89793
	print(Vector2(pi, pi))  # (3.141593, 3.141593)
	print(Vector2(pi * 1e10, pi * 1e10))  # (31415926784, 31415926784)
```

aaronfranke: Mentioned.

akien-mga:
Thanks!

PR #2650, opened 2019-07-22
Replace "shaders" with "shadows" in "shadows_disabled"

FlamyAT:
There was a mistyping in the `shadows_disabled` section


akien-mga:
Thanks!

PR #2651, opened 2019-07-24
fix an error about alpha value

guohongci:

### DIFF
--- a/getting_started/step_by_step/ui_code_a_life_bar.rst
+++ b/getting_started/step_by_step/ui_code_a_life_bar.rst
@@ -612,7 +612,7 @@ textures.
 and alpha. If we darken any of the first three channels it darkens the
 interface. If we lower the alpha channel, our interface fades out.
 
-We're going to tween between two color values: from a white with an
+We're going to tween between two color values: from a black with an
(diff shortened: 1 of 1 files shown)


Calinou:
The current text is already correct: `Color(1.0, 1.0, 1.0, 1.0)` is a fully opaque white, not black.

guohongci:
Oh, I thought it's about the alpha color, like in Photoshop's alpha channel view. Cuz from white to white is a little bit confusing.

akien-mga:
It's like in Photoshop or any image editing software, alpha is not a color but the opacity channel. This examples goes from full opacity white to transparent white (opacity 0).

guohongci:
For a coder that makes perfect sense, but for an artist who paints on the alpha channel a lot, it's also a greyscale map.

What bothers me was the wording made me focused on the difference between color WHITE and color PURE WHITE which also fully transparent. Again, I know it all makes sense for coders, I just feel it would be better if it could make sense to everyone including people just started to try coding.

akien-mga:
The "pure" word is superfluous yes, both are "pure" white (1, 1, 1) with different opacity.

PR #2655, opened 2019-07-25
added move_and_collide test_only doc

2DemiGods:


2DemiGods:
this should be integrated in 3.1 docs as well

akien-mga:
As mentioned in the README and header of `class_*.rst` files, those are generated from the class reference in the engine repository, so they should not be modified here but there.
```
.. DO NOT EDIT THIS FILE, but the KinematicBody2D.xml source instead.
```

2DemiGods:
Sorry, I'm doing this on the go. 
I will check there.

PR #2656, opened 2019-07-25
Adds a breadcrumb template to manage hiding 'Edit on Github' link.

StraToN:
To work, requires adding a meta the very beginning of files that have to hide this link:
`:github_url: hide`

This will thus also require a small adding in makerst.py on godotengine/godot.

For  the record, all credits goes to this kind person: https://github.com/sphinx-doc/sphinx/issues/2386#issuecomment-478403897


akien-mga:
Thanks a ton!

PR #2661, opened 2019-07-26
Improve GDNative C Example tutorial + wrap to 80 chars

akien-mga:
Improve language somewhat, and more details where relevant (e.g.
regarding the version of godot_headers).

Add a section on using the GDNativeLibrary editor plugin.

CC @BastiaanOlij @karroffel 

Images optimized with `optipng`.

----

Some comments:

- Should we rename this tutorial to something which would make clearer that it's the first tutorial that should be read, even if people intend to use C++? Right now there is "GDNative C example" and "GDNative C++ example", one might think that they are equivalent.

- Regarding line wrap, I looked into options for enforcing a common style guide on the docs (we have a mix of 72 chars used by `pandoc` when converting the original Textile docs to rST with long lines of unlimited amount of chars used by new contributors, and then some of us using between 70 and 90 chars per line as a loose guideline). Couldn't find any decent tooling for rST, I guess the language is not popular enough, and Sphinx uses a superset of it that rST-specific tools might not support. The best I found was to run `pandoc` to convert from rST to rST, but it adds a number of issues.


BastiaanOlij:
I need to give it a proper read through later but looks good to me

tutorials/plugins/gdnative/gdnative-c-example.rst
```diff
@@ -322,172 +327,232 @@ The variant we return is destroyed automatically by Godot.
...
-    clang -std=c11 -fPIC -c -I/PATH/TO/GODOT/HEADERS simple.c -o simple.os -arch i386 -arch x86_64
+    clang -std=c11 -fPIC -c -I../../godot_headers simple.c -o simple.os -arch i386 -arch x86_64
     clang -dynamiclib simple.os -o ../bin/libsimple.dylib -arch i386 -arch x86_64
 
 On Windows:
 
 .. code-block:: none
 
-    cl /Fosimple.obj /c simple.c /nologo -EHsc -DNDEBUG /MD /I. /IC:\PATH\TO\GODOT\HEADERS
+    cl /Fosimple.obj /c simple.c /nologo -EHsc -DNDEBUG /MD /I. /I..\..\godot_headers
```
akien-mga: I didn't check if that works on Windows, but given that there is `/I.`, I guess relative paths are fine?

tutorials/plugins/gdnative/gdnative-c-example.rst
```diff
@@ -322,172 +327,232 @@ The variant we return is destroyed automatically by Godot.
...
-    clang -shared simple.os -o ../bin/libsimple.so
+    gcc -std=c11 -fPIC -c -I../../godot_headers simple.c -o simple.os
+    gcc -shared simple.os -o ../bin/libsimple.so
 
 On macOS:
 
 .. code-block:: none
 
-    clang -std=c11 -fPIC -c -I/PATH/TO/GODOT/HEADERS simple.c -o simple.os -arch i386 -arch x86_64
+    clang -std=c11 -fPIC -c -I../../godot_headers simple.c -o simple.os -arch i386 -arch x86_64
```
akien-mga: macOS 32-bit support is kind of obsolete now, should we drop the `i386` code? (And thus update my screenshot below where I assigned the dylib to both 32-bit and 64-bit)

tutorials/plugins/gdnative/gdnative-cpp-example.rst
```diff
@@ -663,36 +672,32 @@ methods that have changed so don't remove anything we're omitting:
...
+    ``rpc_mode``, ``usage``, ``hint`` and ``hint_string``. These can be used to
+    further configure how properties are displayed and set on the Godot side.
 
-    Modern C++ compilers are able to infer the class and variable type
-    and allow you to omit the ``<GDExample, float>`` part of our
-    ``register_property`` method. we've had mixed experiences with this
-    however.
+    Modern C++ compilers are able to infer the class and variable type and allow
+    you to omit the ``<GDExample, float>`` part of our ``register_property``
+    method. we've had mixed experiences with this however.
```
karroffel: Shouldn't this sentence start with a capitalized word?

akien-mga: Definitely. I haven't proofread that page yet, only rewrapped it. Full proofreading is my next step :)

tutorials/plugins/gdnative/gdnative-c-example.rst
BastiaanOlij: I've had problems with relative paths when you start placing source files in subfolders. Not important for the example but..

tutorials/plugins/gdnative/gdnative-c-example.rst
BastiaanOlij: Yeah I'm all for it. 10.15 will fully drop support for the architecture, was originally planned for 10.14. But Macs that actually don't have 64bit support haven't been made for a decade.

tutorials/plugins/gdnative/gdnative-c-example.rst
akien-mga: Though when you reach that point you likely have a SCons setup or `Makefile` that takes care of properly setting include paths for the whole codebase, so I guess for the example it's simple to give a copy-paste ready command (assuming that users followed the recommended project setup).

tutorials/plugins/gdnative/gdnative-c-example.rst
akien-mga: Fixed. I just removed the `-arch` part, I assume that clang will infer it from the current host when not specified?

akien-mga: I also change the image accordingly.

Issue #2664, opened 2019-07-28
Fix warnings reported by Sphinx


Calinou:
Sphinx currently emits several warnings when building HTML documentation:

```text
.../classes/class_gdnativelibrary.rst:33: WARNING: Unknown target name: "godot".
.../classes/class_gdnativelibrary.rst:99: WARNING: Unknown target name: "godot".
.../index.rst:63: WARNING: Undefined substitution referenced: "weblate_widget".
.../community/contributing/pr_workflow.rst:384: WARNING: Could not lex literal_block as "shell". Highlighting skipped.
.../community/contributing/pr_workflow.rst:397: WARNING: Could not lex literal_block as "shell". Highlighting skipped.
```

We should aim to fix these, so that we can set up continuous integration using Sphinx's strict mode (`-W`, which turns warnings into errors).

Calinou:
It seems the warnings in `class_gdnativelibrary.rst` can be fixed by replacing `"godot_"` with `"godot\_"`. This should be changed in `makerst.py` somehow.

cur33:
I've been looking into this issue since it seems like no one else is. It's a weird one, since it seems so easy to fix. I noticed that `makerst.py` has a function called `rstize_text` which escapes text, and it seems to work well enough to just call that on the `PropertyDef.default_value.name` field after checking that the field isn't None. Is that a sufficient solution, or should it be more sophisticated?

cur33:
I only tested it on that one XML file, so I'm working on testing it on the whole repo (I have to finish setting it up since this is my first time working on it).

Calinou:
For reference, here are the warnings as of 2020-07-09 we need to fix to enable strict mode on GitHub Actions:

```text
/home/runner/work/godot-docs/godot-docs/classes/class_basematerial3d.rst:26: WARNING: unknown document: ../tutorials/3d/spatial_material
/home/runner/work/godot-docs/godot-docs/tutorials/3d/csg_tools.rst:252: WARNING: undefined label: doc_spatial_material (if the link has no caption the label must precede a section header)
/home/runner/work/godot-docs/godot-docs/tutorials/misc/gles2_gles3_differences.rst:92: WARNING: undefined label: class_bakedlightmap (if the link has no caption the label must precede a section header)
```

NathanLovato:
Is this issue still relevant?

Calinou:
[...]

PR #2666, opened 2019-07-28
Fix Sphinx Weblate widget warning

Calinou:
The `weblate_widget` substitution is now always defined. This also makes it possible to remove some duplication in `index.rst`.

This partially addresses #2664.


akien-mga:
Thanks!

PR #2668, opened 2019-07-28
Add a reference to pow to the description of exp.

Kaligule:
This might be especially usefull since godot script doesn't support ** or ^ as operators, so beginners (like me) might search for the exponential function, when what they really need is the pow function.

I would have liked to add a link to the pow function, but I don't really see how references work in this setup.

### DIFF
--- a/classes/class_@gdscript.rst
+++ b/classes/class_@gdscript.rst
@@ -479,6 +479,8 @@ The natural exponential function. It raises the mathematical constant **e** to t
 
 **e** has an approximate value of 2.71828.
 
+For exponents to other bases use the method pow.
+
 ::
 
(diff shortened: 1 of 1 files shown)


cbscribe:
Class reference docs are auto-generated, so changes to them will be lost here. Changes must be made in the godotengine/godot repo at https://github.com/godotengine/godot/tree/master/doc/classes

See https://docs.godotengine.org/en/latest/community/contributing/updating_the_class_reference.html for detailed instructions on contributing the the docs.

Kaligule:
Thank you. I created a new pull request over there: https://github.com/godotengine/godot/pull/30909

I will therefore close this one.

PR #2669, opened 2019-07-29
Mark Sphinx extensions as parallel-safe

Calinou:
This makes it possible to build the whole documentation in parallel by specifying the `-j <threads>` argument on the `sphinx-build` command line.

On an i7-6700K @ 4.4 GHz, building the documentation can be achieved in 3 minutes instead of 10 minutes with a single thread.

Testing on Windows would be welcome as well. If this works reliably, we could add `-j auto` to the `make html` command to benefit from parallel builds by default :slightly_smiling_face:

### DIFF
--- a/extensions/gdscript.py
+++ b/extensions/gdscript.py
@@ -169,3 +169,8 @@ class GDScriptLexer(RegexLexer):
 
 def setup(sphinx):
     sphinx.add_lexer('gdscript', GDScriptLexer())
+
+    return {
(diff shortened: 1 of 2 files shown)


akien-mga:
Seems good. For the `gdscript` lexer we're upstream, but `sphinx_tabs` come from https://github.com/djungelorm/sphinx-tabs so it might be worth proposing this change there too.

akien-mga:
Thanks!

PR #2671, opened 2019-07-29
Add a notice about the company list on the "Console support in Godot"

Calinou:
It might not be immediately obvious that companies can be added to the list by simply requesting it in an issue (or making a pull request). This should make it more evident.


akien-mga:
Thanks!

PR #2676, opened 2019-08-02
Updated networking tutorial link references.

oldmerkum:
The original website no longer maintained. These links are for the web.archive.org


oldmerkum:
Open suggestions for using a different link, but this is the one I found that works for all the references in the documentation.

oldmerkum:
Updated links to use 'https'.

oldmerkum:
Screwed up and combined two pull requests that are not related.

PR #2678, opened 2019-08-03
Update intro to 2D animation for Godot 3.1

### LINKED ISSUE #2386: Introduction to the 2D animation features. Needs updating for 3.1
From the looks of the animation panel screenshots, this document needs updating to reflect the UI changes to the animation panel in 3.1

Introduction to the 2D animation features
https://docs.godotengine.org/en/3.1/tutorials/animation/introduction_2d.html

skyace65:
Updates the intro to 2D animation page for Godot 3.1. Closes #2386
-Updates all the images, all images have gone through png compression.
-Updates the name of update rate to update mode.
-Add the definition of Capture Mode. This was taken from [here](https://godotengine.org/article/godot-gets-brand-new-animation-editor-cinematic-support)
-Updates the call func track section to be call method track.
-Pointed out in the call method track tutorial that you should normally use an audio track to play audio instead of using call method on a audio stream.
-Removed the references section at the bottom and its images. It doesn't seem necessary since tooltips explain all the buttons well.

The only thing I'm missing is a definition for the new update mode capture. I have no idea what it does and it's the only thing left that needs to be added.


tutorials/animation/introduction_2d.rst
```diff
@@ -66,7 +65,8 @@ Computer animation relies on keyframes
 
 A keyframe defines the value of a property at a certain point in time.
 
-White and blue diamond shapes represent keyframes in the timeline.
+White and blue diamond shapes represent keyframes in the timeline. A
```
merumelu: This might be misleading as there is no further explanation on why white and blue and if there is a difference between the colors.

I assume what the original author meant was just that a keyframe icon turns blue when selected, however Godot 3.1's bezier curve keyframes are actually blue (and call method keyframes were green even before...).

Since they all behave the same maybe drop the "White and blue" part entirely?

skyace65:
That should be fixed now.

akien-mga:
The commit log says "WIP, need help", is this still the case?

skyace65:
No, I figured out the thing I was having trouble with.

akien-mga:
Ok :) It would be good to amend the commit log so that it doesn't appear as a resolved cry for help in the git history.

skyace65:
I Fixed the log

akien-mga:
Thanks!

PR #2684, opened 2019-08-04
Using Area2D tutorial

cbscribe:
To go along with the "Using KinematicBody2D" tutorial, here's one for Area2D.


cbscribe:
Going to go ahead and merge this, since it seems there aren't any more comments. With more readers seeing it, I can address any additional feedback.

PR #2686, opened 2019-08-05
One-letter spelling mistake in Music Sync doc

DeerTears:
In the long run, though, as the sound hardware clock is never exactly in sync with the system clock, **they** timing information will slowly drift away.

Is now:

In the long run, though, as the sound hardware clock is never exactly in sync with the system clock, **the** timing information will slowly drift away.

Feel free to adapt into a bigger pull request with other spelling mistakes if this one gets ignored/closed


clayjohn:
Looks good to me! All fixes are appreciated.

akien-mga:
Thanks.

Issue #2576, opened 2019-06-29
How are _ready and  _process inherited?


aspenforest:
I have a class which extends another class, which has _ready and _process functions.
If the child has also implemented _ready and _process, will they overwrite the parents implementation?
If the child has no _process of its own but inherits it from its parent, will calling self.set_process(true) enable the process which was inherited from the parent? 
Wish there was a decent documentation available about this.

vortexofdoom:
I know they do not override the parent's implementations, it will call both, although I don't know in what order.

aspenforest:
So what you are saying is that Godot doesn't really have actual inheritance.  
Thanks for the clarification.  
I'm really dissapointed that this essential information is not documented in an easily accessible manner.

vortexofdoom:
I believe only certain functions like `_ready()`, `_process()`, and `_physics_process()` are affected by this. User-defined functions should be replaced.

It's possible to get around by calling an empty `_custom_function()` in the parent's `_process()` function and defining that behavior in the inheriting class. In my opinion this is sufficient, it's slightly hacky but only in theory, actual execution efficiency should be identical.

Calinou:
You can call a parent function using the dot syntax: `.method()`

If the parent's virtual method isn't being run, you could likely run it this way.

oldmerkum:
@aspenforest were you able to figure out the answer?
I can test this if not.

aspenforest:
I haven't focused on this issue for a while, cause it hasn't been that
crucial and I have more pressing matters to attend to, but if you
would be able to come to a conclusion on this matter, then perhaps you
could make a pull request to the Godot Docs repo with the necessary
improvements?

Best regards,

oldmerkum:
So, if the class extends Node, and during run-time is added as a child node, then the functions _ready(), process(delta), etc. are not overridden. Instead each implementation of the method is called.
If the class is not added as a child node and instead kept as a member variable, this is not the case.

Is this expected behaviour?

oldmerkum:
Maybe this should be in the scripting section of the godot-docs?

aspenforest:
"So, if the class extends Node, and during run-time is added as a child
node, then the functions _ready(), process(delta), etc. are not overridden."

Does it work the same way if one extends from an user-defined GDScript
file, which itself extends from e.g. Node?

If true, that is the biggest load of turd I have ever witnessed of ANY
programming language.

Thank you for your help on this matter, I appreciate it a lot. :)

[...]

PR #2687, opened 2019-08-05
Improve the GDScript basics page

Calinou:
- Add more information on the `assert` keyword
- Document binary literals
- Use "parent class" everywhere for consistency
- Move information about parent class function overloading to be more visible
- Don't recommend multiline strings for multiline comments to avoid performance issues in tight loops
- Improve grammar and writing style

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_basics.rst
+++ b/getting_started/scripting/gdscript/gdscript_basics.rst
@@ -6,7 +6,7 @@ GDScript basics
 Introduction
 ------------
 
-*GDScript* is a high level, dynamically typed programming language used to
+*GDScript* is a high-level, dynamically typed programming language used to
 create content. It uses a syntax similar to
 `Python <https://en.wikipedia.org/wiki/Python_%28programming_language%29>`_
 (blocks are indent-based and many keywords are similar). Its goal is
@@ -16,7 +16,7 @@ flexibility for content creation and integration.
 History
 ~~~~~~~
 
-In the early days, the engine used the `Lua <http://www.lua.org>`__
+In the early days, the engine used the `Lua <https://www.lua.org>`__
 scripting language. Lua is fast, but creating bindings to an object
 oriented system (by using fallbacks) was complex and slow and took an
 enormous amount of code. After some experiments with
@@ -63,12 +63,12 @@ here's a simple example of how GDScript looks.
 
     class_name MyClass, "res://path/to/optional/icon.svg"
 
-    # Member Variables
+    # Member variables
 
     var a = 5
     var s = "Hello"
     var arr = [1, 2, 3]
-    var dict = {"key": "value", 2:3}
+    var dict = {"key": "value", 2: 3}
     var typed_var: int
     var inferred_type := "String"
 
@@ -82,7 +82,7 @@ here's a simple example of how GDScript looks.
     enum {UNIT_NEUTRAL, UNIT_ENEMY, UNIT_ALLY}
     enum Named {THING_1, THING_2, ANOTHER_THING = -1}
 
-    # Built-in Vector Types
+    # Built-in vector types
 
     var v2 = Vector2(1, 2)
     var v3 = Vector3(1, 2, 3)
@@ -114,7 +114,7 @@ here's a simple example of how GDScript looks.
     func something(p1, p2):
         .something(p1, p2)
 
-    # Inner Class
+    # Inner class
 
     class Something:
         var a = 10
@@ -246,15 +246,15 @@ The following is the list of supported operators and their precedence.
 +---------------------------------------------------------------+-----------------------------------------+
 | **Operator**                                                  | **Description**                         |
 +---------------------------------------------------------------+-----------------------------------------+
-| ``x[index]``                                                  | Subscription, Highest Priority          |
+| ``x[index]``                                                  | Subscription (highest priority)         |
 +---------------------------------------------------------------+-----------------------------------------+
-| ``x.attribute``                                               | Attribute Reference                     |
+| ``x.attribute``                                               | Attribute reference                     |
 +---------------------------------------------------------------+-----------------------------------------+
-| ``is``                                                        | Instance Type Checker                   |
+| ``is``                                                        | Instance type checker                   |
 +---------------------------------------------------------------+-----------------------------------------+
 | ``~``                                                         | Bitwise NOT                             |
 +---------------------------------------------------------------+-----------------------------------------+
-| ``-x``                                                        | Negative / Unary Negation               |
+| ``-x``                                                        | Negative / Unary negation               |
 +---------------------------------------------------------------+-----------------------------------------+
 | ``*`` ``/`` ``%``                                             | Multiplication / Division / Remainder   |
 |                                                               |                                         |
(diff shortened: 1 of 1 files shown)


getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -323,26 +325,28 @@ considered a comment.
...
+::
 
-        """ Everything on these
-        lines is considered
-        a comment. """
+    """
+    Everything on these
+    lines is considered
+    a comment.
+    """
```
cbscribe: Technically, these are not treated as comments, but as string literals.

getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -358,26 +362,26 @@ be assigned any other value.
 bool
 ^^^^
 
-The Boolean data type can only contain ``true`` or ``false``.
+The boolean data type can only contain ``true`` or ``false``.
```
cbscribe: This is typically capitalized, as it's a proper name. Just a nitpick - I have no strong objection either way.

getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -392,7 +396,7 @@ accessed as array.
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 2D Rectangle type containing two vectors fields: ``position`` and ``size``.
-Alternatively contains an ``end`` field which is ``position+size``.
+Alternatively contains an ``end`` field which is ``position + size``.
```
cbscribe: I'd remove the "Alternatively". Seems to imply the field is optional, but it's always there. "Also" might be better.

getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -422,7 +426,7 @@ useful for interpolating rotations.
 
 Axis-aligned bounding box (or 3D box) contains 2 vectors fields: ``position``
 and ``size``. Alternatively contains an ``end`` field which is
-``position+size``.
+``position + size``.
```
cbscribe: As above.

getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -470,7 +474,7 @@ Container built-in types
 
 Generic sequence of arbitrary object types, including other arrays or dictionaries (see below).
 The array can resize dynamically. Arrays are indexed starting from index ``0``.
-Starting with Godot 2.1, indices may be negative like in Python, to count from the end.
+Indices may be negative to count from the end.
```
cbscribe: Negative indices count from the end.

getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -613,12 +617,12 @@ engine will raise an error.
     my_int = Vector2() as int # A Vector2 can't be converted to int, this will cause an error
 
 Casting is also useful to have better type-safe variables when interacting with
-tree::
+the scene tree::
 
-    # will infer the variable to be of type Sprite:
+    # Will infer the variable to be of type Sprite
```
cbscribe: Since you're adding periods to the end of comments, this one needs it too.

getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -613,12 +617,12 @@ engine will raise an error.
...
     var my_sprite := $Character as Sprite
 
-    # will fail if $AnimPlayer is not an AnimationPlayer, even if it has the method 'play()':
+    # Will fail if $AnimPlayer is not an AnimationPlayer, even if it has the method 'play()'
```
cbscribe: Period

getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -634,7 +638,7 @@ expressions and must be assigned on initialization.
     const C = 10 + 20 # Constant expression.
     const D = Vector2(20, 30).x # Constant expression: 20
     const E = [1, 2, 3, 4][0] # Constant expression: 1
-    const F = sin(20) # sin() can be used in constant expressions.
+    const F = sin(20) # 'sin()' can be used in constant expressions.
```
cbscribe: There's a mix of with/without closing punctuation on these comments.

getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -725,14 +729,14 @@ return early with the ``return`` keyword, but they can't return any value.
...
+Referencing functions
 ^^^^^^^^^^^^^^^^^^^^^
 
-Contrary to Python, functions are *not* first class objects in GDScript. This
+Contrary to Python, functions are *not* first-class objects in GDScript. This
 means they cannot be stored in variables, passed as an argument to another
 function or be returned from other functions. This is for performance reasons.
 
-To reference a function by name at runtime, (e.g. to store it in a variable, or
+To reference a function by name at run-time, (e.g. to store it in a variable, or
```
cbscribe: Interchangeable, but "runtime" is by far the more common usage, at least according to google trends.

getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -745,21 +749,12 @@ pass it to another function as an argument) one must use the ``call`` or
...
-
 Static functions
 ^^^^^^^^^^^^^^^^
 
 A function can be declared static. When a function is static, it has no
 access to the instance member variables or ``self``. This is mainly
-useful to make libraries of helper functions:
-
-::
+useful to make libraries of helper functions::
```
cbscribe: Why double colons?

getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1081,40 +1072,41 @@ the ``is`` keyword can be used:
...
+``.some_func``, if the constructor from the inherited class takes arguments,
+they are passed like this::
 
     func _init(args).(parent_args):
        pass
 
-This is better explained through examples. Say we have this scenario:
-
-::
+This is better explained through examples. Say, we have this scenario::
```
cbscribe: The comma is incorrect here. Alternatively, you could say "Consider this scenario"

cbscribe:
After testing, it seems I'm wrong about the :: thing, so ignore those comments :P

getting_started/scripting/gdscript/gdscript_basics.rst
Calinou: Should I remove this paragraph entirely? Since Godot 3.1, multiline strings are no longer highlighted as comments. Besides, this causes a string to be allocated every time the code is run, which may cause performance issues in tight loops.

getting_started/scripting/gdscript/gdscript_basics.rst
Calinou: According to [this Stack Exchange question](https://english.stackexchange.com/questions/4481/should-the-word-boolean-be-capitalized), capitalized "Boolean" is more common, so I'll go for that.

getting_started/scripting/gdscript/gdscript_basics.rst
Calinou: I personally got used to write "run-time" to mean the opposite of "compile-time", and write "runtime" in nouns such as "Java runtime environment". Still, I can replace it with "runtime" if this is desired.

getting_started/scripting/gdscript/gdscript_basics.rst
cbscribe: That might be a better approach, rather than encouraging its use.

Calinou:
As mentioned in https://github.com/godotengine/godot-docs/issues/2576#issuecomment-529754003, it would be good to cherry-pick this if possible :slightly_smiling_face:

PR #2689, opened 2019-08-06
Improve the Data paths page

Calinou:
- Add examples for the editor data paths and mention XDG compliance
- Add information about the self-contained mode

PS: I remember seeing self-contained mode being documented somewhere, but I can't find any references to it in the documentation files. Maybe it was documented on another platform?


cbscribe:
Excellent addition.

PR #2690, opened 2019-08-07
Added notes about local constants for shaders

Chaosus:
Just a notes about new functionality.

### DIFF
--- a/tutorials/shading/shading_reference/shading_language.rst
+++ b/tutorials/shading/shading_reference/shading_language.rst
@@ -246,6 +246,44 @@ Arrays also have a built-in function ``.length()`` (not to be confused with the 
 
 Note: If you use an index below 0 or greater than array size - the shader will crash and break rendering. To prevent this, use ``length()``, ``if``, or ``clamp()`` functions to ensure the index is between 0 and the array's length. Always carefully test and check your code. If you pass a constant expression or a simple number, the editor will check its bounds to prevent this crash.
 
+Constants
+---------
+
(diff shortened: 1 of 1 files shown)


tutorials/shading/shading_reference/shading_language.rst
```diff
@@ -103,6 +103,19 @@ Default integer constants are signed, so casting is always needed to convert to
     uint a = 2; // invalid
     uint a = uint(2); // valid
 
+Local Constants
```
clayjohn: I think we should just have one section titled "Constants" instead of global and local ones. 

### Constants

Use the ``const`` keyword before the variable declaration to make that variable immutable, which means that it cannot be modified. All basic types, except samplers can be declared as constants. Accessing and using a constant value is slightly faster than using a uniform.

```
const vec2 a = vec2(0.0, 1.0);
vec2 b;

a = b; // invalid
b = a; // valid
```

Constants cannot be modified and additionally cannot have hints, but multiple of them (if they have the same type) can be declared in a single expression e.g
```
const vec2 V1 = vec2(1, 1), V2 = vec2(2, 2);
```
Constants can be declared both globally (outside of any function) or locally (inside a function). 
Global constants are useful when you want to have access to a value throughout your shader that does not need to be modified. Like uniforms, global constants are shared between all shader stages. However, they are not accessible outside of the shader and must be initialized at declaration. 

```
shader_type spatial;

const float PI = 3.14159265358979323846;
```

Chaosus: Done. Check again )

akien-mga:
Thanks!

PR #2693, opened 2019-08-11
Update immediategeometry.rst

guohongci:
"
.......
moudules/gdscript/gdscript_tokenizer.cpp:1104 - Condition ' tk_rb[ofs].type != TK_IDENTIFER ' is true. returned: StringName()"
built-in:3 - Parse Error: Unexpected token: void:
modules/gdscript/gdscript.cpp:580 - Method/Function Faild, returning: ERR_PARSE_ERROR
moudules/gdscript/gdscript_tokenizer.cpp:1104 - Condition ' tk_rb[ofs].type != TK_IDENTIFER ' is true. returned: StringName()
.......
"

### DIFF
--- a/tutorials/content/procedural_geometry/immediategeometry.rst
+++ b/tutorials/content/procedural_geometry/immediategeometry.rst
@@ -42,7 +42,7 @@ The example code below draws a single triangle.
 
     extends ImmediateGeometry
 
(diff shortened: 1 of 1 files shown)


cbscribe:
Thanks!

PR #2697, opened 2019-08-11
Document APK size optimization in Exporting for Android

Calinou:
See https://github.com/godotengine/godot/issues/18253.

### DIFF
--- a/getting_started/workflow/export/exporting_for_android.rst
+++ b/getting_started/workflow/export/exporting_for_android.rst
@@ -77,13 +77,35 @@ Now fill in the following forms in your Android Export Presets:
 
 .. image:: img/editor-export-presets-android.png
 
-- Release: Enter the path to the keystore file you just generated.
-- Release User: Replace with the key alias.
-- Release Password: Key password. Note that the keystore password and the key password currently have to be the same.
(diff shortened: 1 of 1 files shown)


mani144:
these are important steps for size optimization

getting_started/workflow/export/exporting_for_android.rst
```diff
@@ -77,13 +77,35 @@ Now fill in the following forms in your Android Export Presets:
...
-Don't forget to disable the ``Export With Debug`` button while choosing the APK's name.
+Don't forget to uncheck the **Export With Debug** checkbox while choosing the APK's name.
 
 .. image:: img/export-with-debug-button.png
 
+Optimizing the APK size
+-----------------------
+
+By default, the APK will contain native libraries for both ARMv7 and ARMv8
+architectures (64-bit). This increases its size significantly. To create a
```
akien-mga: The "(64-bit)" is a bit confusing, as placed here it would apply to both architectures, while only ARMv8 is 64-bit. I think it could simply be removed, otherwise it should be "both ARMv7 (32-bit) and ARMv8 (64-bit) architectures".

akien-mga:
Thanks!

PR #2705, opened 2019-08-13
Mention caveats of using RGTC compression for normal maps with shaders

Calinou:
See discussion in https://github.com/godotengine/godot/issues/31324.

Could this paragraph be made more extensive somehow? I'm not an expert on shaders, so I'd appreciate some feedback on this :smiley:


akien-mga:
Thanks!

PR #2707, opened 2019-08-14
Add Windows snippet for compiling with script encryption key

Xrayez:
Refer to [this QA](https://godotengine.org/qa/50103/help-step-compile-export-template-windows-script-encryption?show=50240#a50240).

### DIFF
--- a/development/compiling/compiling_with_script_encryption_key.rst
+++ b/development/compiling/compiling_with_script_encryption_key.rst
@@ -14,8 +14,8 @@ optimized and without symbols, it would take some effort to find it.
 For this to work, you need to build the export templates from source,
 with that same key.
 
-Step by step (Linux)
---------------------
+Step by step
+------------
 
 1. Generate a 256bit AES key in HEX. You can use the aes-256-cbc variant from
    `this service <https://asecuritysite.com/encryption/keygen>`_.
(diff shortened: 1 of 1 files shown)


development/compiling/compiling_with_script_encryption_key.rst
```diff
@@ -37,14 +37,17 @@ Step by step (Linux)
...
 
-   ::
+   .. tabs::
+    .. code-tab:: bash Linux/macOS
 
        export SCRIPT_AES256_ENCRYPTION_KEY="your_generated_key"
 
+    .. code-tab:: bat Windows
+
+       set SCRIPT_AES256_ENCRYPTION_KEY="your_generated_key"
```
Calinou: Doesn't `set` use this syntax?

```suggestion
       set SCRIPT_AES256_ENCRYPTION_KEY your_generated_key
```

Xrayez: Always trouble with double quotes in examples. 🙂

You are right that the quotes would be included here, yet `=` is still required according to [set syntax](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/set_1#syntax). So I'll just remove the quotes then.

akien-mga:
Thanks!

PR #2710, opened 2019-08-16
State design pattern tutorial

oldmerkum:


oldmerkum:
This is a tutorial is for those wanting to implement the state design pattern in godot. Comments and feedback welcome.

oldmerkum:
My previous pull-request updating the network tutorial links are in these commits as well. I can remove those updates and create a different pull-request for that if it is an issue.

oldmerkum:
After doing some more testing with this tutorial, I realized that the Vector2 class is built-in and was not being passed by reference, but by value. I am going to update the example scripts to access a velocity variable in the 'persistent_state.gd' script.

I also realized that since the 'class_name' keyword is used, it might be better to not use a state_factory and instead have the classes pass a newly instantiated state implementation to 'persistent_state.gd'. The benefit of not doing this would be that one can always know precisely which states are available from 'state_factory.gd', since they are explicitly listed in the dictionary.

Open to suggestions, I will opt to keep the state factory by default.

bojidar-bg:
Note that many people use different ways to implement state machines, including:
* Nodes which contain the script and are called whenever needed.
* `match` statements with a `State` enum. Or `if`-s.
* Swapping scripts.
* `call`-ing functions in the same script for each state.
* Scripts which extend `Object` or `Reference` and contain information about a state.

I'm not sure if "standardizing" a single method in the docs is a good idea, maybe there could be mentions of the other ways, with the in-depth explanation being of the easiest.

oldmerkum:
Good points on the different ways to implement states. I agree that they should be added and that a mention that this is not the only way to implement states should also be included.

I wanted to implement the swapping scripts option at first, but I didn't put enough effort into learning how to do that. I will try to get that working too so that I could have 2 ways in the tutorial.

I do want to make a point on the difference between using the state design pattern and a simple switch statement. I like the switch statement for short tutorials or when there just aren't many states to deal with, but when working a player character I find that using some form of the state design pattern to implement a finite state machine much easier in the long run.

Also, I used this reference heavily my first time ever attempting to implement this pattern:
https://gameprogrammingpatterns.com/state.html
So, I will add that link into the page as well.

oldmerkum:
Mentioned other state implementation options.
Changed the Vector2 velocity variable to be accessed 'persistent_state.gd' member variable directly.
I had some trouble with having a node swap out its own script to change state. Curious if this is possible.

bojidar-bg:
I just tried with the script-swapping approach, but it seems like variables are not automatically preserved when swapping scripts. (Also, I managed to accidentally crash the game while testing the approach.)

I could get it working with something like:
```gdscript
# res://ScriptSwap.gd
extends Node

var shared_variable = 10

func _process(delta):
	prints(get_script().resource_path, shared_variable)

func change_state(to):
	call_deferred("set_script", to)
	for property in get_property_list():
		if property.usage & PROPERTY_USAGE_SCRIPT_VARIABLE:
			set_deferred(property.name, get(property.name))
```
```gdscript
# res://ScriptSwap-A.gd
extends "res://ScriptSwap.gd"

# warning-ignore:unused_argument
func _process(delta):
	shared_variable += 1
	if shared_variable > 15:
		change_state(load("res://ScriptSwap-B.gd"))
```
```gdscript
# res://ScriptSwap-B.gd
extends "res://ScriptSwap.gd"

# warning-ignore:unused_argument
func _process(delta):
	shared_variable -= 1
	if shared_variable < 5:
		change_state(load("res://ScriptSwap-A.gd"))
```

And, reasonably enough, it prints:
```6
res://ScriptSwap-B.gd 15
res://ScriptSwap-B.gd 14
…
res://ScriptSwap-B.gd 5
res://ScriptSwap-B.gd 4
res://ScriptSwap-A.gd 5
res://ScriptSwap-A.gd 6
…
res://ScriptSwap-A.gd 15
res://ScriptSwap-A.gd 16
res://ScriptSwap-B.gd 15
res://ScriptSwap-B.gd 14
```

Notes:
* The check for `property.usage & PROPERTY_USAGE_SCRIPT_VARIABLE` can be replaced by a check for `property.name != "script"`; but that is unnecessary given that non-script properties will be  persisted anyway.
* `# warning-ignore:unused_argument` is used so that the console is not spammed with warnings every time a script is `load`-ed.
* To avoid using `load`, a fourth script holding all states would be necessary. You should not put a list of all states on `ScriptSwap.gd`, since this would immediately cause a circular reference. Likewise, using `preload` instead of `load` would make a circular reference. Only way to avoid one is by putting the states in a (potentially singleton) autoload script, since references to the autoload are not reference-counted.
* You don't need to extend `ScriptSwap.gd`, but the code which changes the script needs to stay somewhere. Maybe in the aforementioned autoload? `StateAutoload.change_state(self, "A")`

oldmerkum:
The problem I encountered is that if you did not define ```_process(delta)``` in ```ScriptSwap.gd```, then the ```change_state(to)``` would not swap scripts.

bojidar-bg:
Interesting... I tried it, but cannot confirm the problem on my end. I'm running 3.2/master from https://github.com/godotengine/godot/commit/ccf294b92f02af6e60206e220dcf4a8474f73f0a.

Calinou:
@oldmerkum That's correct, `master` is what will become 3.2 in the future.

oldmerkum:
I will compile master and test.

oldmerkum:
Okay, I see what I did wrong. I removed the `_process(delta)` from `ScriptSwap.gd` and placed a print in the local `_process(delta)` function of `ScriptSwap-B.gd`. I did the same for `ScriptSwap-A.gd`, but I also added a local variable to `ScriptSwap-B.gd` and I believe that is where the issue arose.
The loop that gets all the properties:
```gdscript
for property in get_property_list():
		if property.usage and PROPERTY_USAGE_SCRIPT_VARIABLE:
			set_deferred(property.name, get(property.name))
```
was also getting the local variables.  I put
```gdscript
		if (property.usage and PROPERTY_USAGE_SCRIPT_VARIABLE
				and property.name == "shared_variable"):
 ```
and it worked properly. Seems that keeping any local items in the class gets one strange interactions. This is a pretty low level strategy which would ultimately help a someone understand how godot works more, but I don't feel comfortable adding that into this tutorial without some more experimentation.

**Comment by oldmerkum** [comment]
Is there more that should be added/edited for this to be merged?

File: tutorials/misc/state_design_pattern.rst [hunk 1]
```diff
@@ -0,0 +1,254 @@
+.. _doc_state_design_pattern:
+
+State Design Pattern
```
akien-mga: Should not use Title Case, but natural case ("State design pattern").

tutorials/misc/state_design_pattern.rst
```diff
@@ -0,0 +1,254 @@
...
+
+Introduction
+------------
+
+Scripting a game can be difficult when there are many states that need to handled, but
+only one script can be attached to a node at a time. Instead of creating a state machine
+within the player's control script, it would make development simpler if the states were
+separated out into different classes.
+
+There are many ways to implement a state machine with godot, and some other methods are below:
```
akien-mga: **G**odot

tutorials/misc/state_design_pattern.rst
```diff
@@ -0,0 +1,254 @@
...
+
+* The player can have a child node for each state, which are called when utilized
```
akien-mga: Add dot at end of line. (Same below.)

tutorials/misc/state_design_pattern.rst
```diff
@@ -0,0 +1,254 @@
...
+* Enums can be used in conjunction with a match statement
+* The state scripts themselves could be swapped out from a node dynamically at run-time 
+
+This tutorial will focus only on adding and removing nodes which have a state script attached. Each state
+script will be an implementation of a different state.
+
+.. note::
+  There is a great resource explaining the concept of the state design pattern here:
```
akien-mga: Use 4 spaces for indentation.

tutorials/misc/state_design_pattern.rst
```diff
@@ -0,0 +1,254 @@
...
+  https://gameprogrammingpatterns.com/state.html
+
+Script-Setup
```
akien-mga: Should be "Script setup"? Not sure why the hyphen.

tutorials/misc/state_design_pattern.rst
```diff
@@ -0,0 +1,254 @@
...
+    
+    class_name State
+
+    var change_state
+    var animated_sprite
+    var persistent_state
+    var velocity
+
+    func _physics_process(delta):
+      persistent_state.move_and_slide(persistent_state.velocity, Vector2.UP)
```
akien-mga: 4 spaces for indentation.

tutorials/misc/state_design_pattern.rst
```diff
@@ -0,0 +1,254 @@
...
+    extends Node2D
+    
+    class_name State
+
+    var change_state
+    var animated_sprite
+    var persistent_state
+    var velocity
+
+    func _physics_process(delta):
```
akien-mga: Use `_delta`, since you're not using `delta` it would generate a warning about unused variable, which a `_` prefix prevents.

tutorials/misc/state_design_pattern.rst
```diff
@@ -0,0 +1,254 @@
...
+      persistent_state.move_and_slide(persistent_state.velocity, Vector2.UP)
+
+    func setup(change_state, animated_sprite, kinematic_body_2d):
```
akien-mga: `kinematic_body_2d` is unused.

tutorials/misc/state_design_pattern.rst
```diff
@@ -0,0 +1,254 @@
...
+	self.change_state = change_state
```
akien-mga: 4 spaces for indent.

tutorials/misc/state_design_pattern.rst
```diff
@@ -0,0 +1,254 @@
...
+
+Finally, this script is actually being designated as a class named ``State``. This makes refactoring the code
+easier, since the file path from using the ``load()`` and ``preload()`` functions in godot will not be needed.
+
+So, now that there is a base state, the two states discussed earlier can be implemented.
+
+.. tabs::
+  .. code-tab:: gdscript GDScript
+
+    # filename: idle_state.gd
```
akien-mga: It should be enough to just use `# idle_state.gd` without specifying `filename: `. That's what is done in most other tutorials.

tutorials/misc/state_design_pattern.rst
```diff
@@ -0,0 +1,254 @@
...
+
+    #filename: state_factory.gd
+
+    class_name StateFactory
+
+    var states
+
+    func _init():
+      states = {
+          "idle":IdleState,
```
akien-mga: Add a space after `:`.

tutorials/misc/state_design_pattern.rst
```diff
@@ -0,0 +1,254 @@
...
+      state = state_factory.get_state(new_state_name).new()
+      state.setup(funcref(self, "change_state"), $AnimatedSprite, self)
+      state.name = "current_state"
+      add_child(state)
+
+.. note:: 
+  The ``persistent_state.gd`` script contains code for detecting input. This was to make the tutorial simple, but it is not usually 
+  best practice to do this.
+
+Project-Setup
```
akien-mga: "Project setup"

tutorials/misc/state_design_pattern.rst
oldmerkum: Should every indent for an rst file be 4 spaces or just the gdscript?

oldmerkum: bump

tutorials/misc/state_design_pattern.rst
oldmerkum: Updated the indents.

tutorials/misc/state_design_pattern.rst
oldmerkum: Updated the indents to 4 spaces.

tutorials/misc/state_design_pattern.rst
oldmerkum: Updated the indents to 4 spaces.

Calinou:
Thanks!

PR #2714, opened 2019-08-19
correct variable reference in setget example

nippysaurus:
Correcting a typo introduced in e173600c79b8b57c951e02e0b1f1bcc8d5e46506.


Calinou:
Since the variable in the getter is referred to as `my_var`, we should correct all instances of `myvar` to use `my_var` instead :wink:

nippysaurus:
@Calinou thanks for the feedback. I'm sure this has come up in discussion previously, but how does the community/team feel about using realistic variable names and variables (like "health") vs "dummy" variables (like "my_var")?

From my perspective this is more easily relatable:

```python
var health setget health_set, health_get

func health_set(new_value):
    health = new_value

func health_get():
    return health_var # Getter must return a value.
```

I'm just not sure if there has been any previous though on this, and if I had a desire to improve this snippet in this way whether this would be the best place to do it?

akien-mga:
Thanks!

PR #2718, opened 2019-08-21
Split GDScript exports documentation into its own page

Calinou:
This makes the GDScript basics page shorter.

Maybe this should be turned into a language-independent page with C# code samples.
(Can all of the GDScript export syntax be represented in C#?)


akien-mga:
> Maybe this should be turned into a language-independent page with C# code samples.
(Can all of the GDScript export syntax be represented in C#?)

That would make sense, WDYT @neikeq?

neikeq:
I agree. But I think not all hints are possible right now in C#. We would need to add some attributes for them.

akien-mga:
Thanks!

PR #2725, opened 2019-08-28
Clarify the location of the "Layout" menu

lamsorsen:
I spent a good 15 minutes looking for the "Layout" menu. Most of that is my own fault but here's a small edit that I think would be slightly more clarifying (using the same terminology from the "Introduction to Godot's editor" page)


akien-mga:
Thanks!

PR #2729, opened 2019-08-29
Updated description for "stepify"

t-karcher:
Mentioned the possibility to use stepify as a rounding function; +examples

### DIFF
--- a/classes/class_@gdscript.rst
+++ b/classes/class_@gdscript.rst
@@ -1149,8 +1149,13 @@ Returns the position of the first non-zero digit, after the decimal point. Note 
 
 - :ref:`float<class_float>` **stepify** **(** :ref:`float<class_float>` s, :ref:`float<class_float>` step **)**
 
-Snaps float value ``s`` to a given ``step``.
+Snaps float value ``s`` to a given ``step``. This can also be used to round a floating point number to an arbitrary number of decimals.
 
+::
+
+    stepify(100, 32) # Returns 96
+    stepify(3.14159, 0.01) # Returns 3.14
+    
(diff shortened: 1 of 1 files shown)


t-karcher:
Oops: I forgot this needs to be updated in the XML class reference of the Godot repo directly. Will close it here and reopen it there.

classes/class_@gdscript.rst
```diff
@@ -1149,8 +1149,13 @@ Returns the position of the first non-zero digit, after the decimal point. Note
 
 - :ref:`float<class_float>` **stepify** **(** :ref:`float<class_float>` s, :ref:`float<class_float>` step **)**
 
-Snaps float value ``s`` to a given ``step``.
+Snaps float value ``s`` to a given ``step``. This can also be used to round a floating point number to an arbitrary number of decimals.
```
akien-mga: This is not "also", you're rephrasing what "Snap float value ``s`` to a given ``step``" means :)

akien-mga: I think it will also truncate and not round. But the sentence can likely be made easier to understand, and adding an example seems like a good idea.

akien-mga: > This is not "also", you're rephrasing what "Snap float value s to a given step" means :)

Sorry, I misread the example and misunderstood the original description it seems. I didn't know this method could be used to the same effect as `fmod` to snap to a step (as described). I always saw it only used for adjusting the number of decimals.
So your addition seems good :)

akien-mga: > I think it will also truncate and not round.

And I tried it and it does round... sorry for commenting before testing :)

PR #2733, opened 2019-08-31
Clarify the behavior of match and continue

xuvatilavv:
I found the description of the fallthrough behavior of 'continue' to be misleading, since it doesn't behave the same way as any language's switch fallthrough that I'm aware of. Added details on how our 'continue' is different.

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_basics.rst
+++ b/getting_started/scripting/gdscript/gdscript_basics.rst
@@ -851,15 +851,15 @@ Basic syntax::
 
 1. Replace ``switch`` with ``match``.
 2. Remove ``case``.
-3. Remove any ``break``\ s. If you don't want to ``break`` by default, you can use ``continue`` for a fallthrough.
+3. Remove any ``break``\ s. If you don't want to ``break`` by default, you can use ``continue`` followed by a Multiple pattern for a fallthrough.
 4. Change ``default`` to a single underscore.
 
 
 **Control flow**:
 
 The patterns are matched from top to bottom.
-If a pattern matches, the corresponding block will be executed. After that, the execution continues below the ``match`` statement.
-If you want to have a fallthrough, you can use ``continue`` to stop execution in the current block and check the ones below it.
+If a pattern matches, the first corresponding block will be executed. After that, the execution continues below the ``match`` statement.
+If you want to have a fallthrough, you can use ``continue`` to stop execution in the current block and check for an additional match in the patterns below it.
 
 There are 6 pattern types:
 


getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -851,15 +851,15 @@ Basic syntax::
 
 1. Replace ``switch`` with ``match``.
 2. Remove ``case``.
-3. Remove any ``break``\ s. If you don't want to ``break`` by default, you can use ``continue`` for a fallthrough.
+3. Remove any ``break``\ s. If you don't want to ``break`` by default, you can use ``continue`` followed by a Multiple pattern for a fallthrough.
```
nippysaurus: Why is "Multiple" capitalized like this? This change does not make this statement any more clear to me.

getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -851,15 +851,15 @@ Basic syntax::
...
 4. Change ``default`` to a single underscore.
 
 
 **Control flow**:
 
 The patterns are matched from top to bottom.
-If a pattern matches, the corresponding block will be executed. After that, the execution continues below the ``match`` statement.
-If you want to have a fallthrough, you can use ``continue`` to stop execution in the current block and check the ones below it.
+If a pattern matches, the first corresponding block will be executed. After that, the execution continues below the ``match`` statement.
+If you want to have a fallthrough, you can use ``continue`` to stop execution in the current block and check for an additional match in the patterns below it.
```
nippysaurus: This looks like a step in the right direction to me. It's definitely more clearly explained.

getting_started/scripting/gdscript/gdscript_basics.rst
officialjacobgray: It's in reference to the "Multiple" pattern type described in the following examples. Perhaps "followed by an additional matching pattern" would be more clear?

nippysaurus: I see the "Multipatterns" section below the specific pattern examples. Is that what you are referring to?

To be honest the wording "continue followed by..." makes it sound like there might be parameters to the `continue` keyword, as if it were a method or something.

officialjacobgray: Interesting... so that is what I'm referring to, but that part is different between the live site and the version here on github that I read while editing. It doesn't match any version of this document, and I don't see any issues or PRs related to the differences. Why would that be?

Considering the live document, this line might more properly be something like "you can use continue and include a multipattern that also matches [the expression?] for a fallthrough", yes?

mhilbrunner: In general, always prefer making changes against the `master` branch, and only care about that branch. Only exceptions are if your changes are only relevant for specific, older versions.

mhilbrunner:
Hey, thanks for this!

Sorry it took a while to get around to this PR :)

I think the first change is a little confusing and we may need to come up with a better explanation, however, the second one is quite an improvement over the current version.

As this PR is a little old, I have created a new one with that change, giving you credit using GitHub `Co-authored-by` feature.

Thanks again, and congrats on your first contribution to the Godot docs! 🎉

mhilbrunner:
Closing as superseded by https://github.com/godotengine/godot-docs/pull/3551 as outlined above.

PR #2734, opened 2019-08-31
updated trailing stop after exclamation mark in...

leela52452:
...text: Always use one space around operators and after commas. Avoid extra spaces in dictionary references and function calls, or to create "columns."

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_styleguide.rst
+++ b/getting_started/scripting/gdscript/gdscript_styleguide.rst
@@ -135,7 +135,7 @@ Whitespace
 ~~~~~~~~~~
 
 Always use one space around operators and after commas. Avoid extra
-spaces in dictionary references and function calls, or to create "columns."
+spaces in dictionary references and function calls, or to create "columns".
 
 **Good**:
 


cbscribe:
In American English, the rule is that the trailing period goes inside the quotes. Since American English is the standard for these docs, this change is incorrect.

Although I would be curious to grep through and see how often this is followed.

akien-mga:
> In American English, the rule is that the trailing period goes inside the quotes. Since American English is the standard for these docs, this change is incorrect.

Is that really the rule even in such a situation though? It does make sense to have the trailing period inside the quotes when the quote is of a sentence ending with a period, but does the rule apply when the quoted text does not end with a period? It does look quite awkward in the above example.

cbscribe:
I would argue it looks awkward because we so often see it done the other way too. It's definitely a minor quibble - in UK English it's acceptable both ways.

Perhaps we should name a designated English style manual so these kinds of things are prescribed? It would at least give us something to point writers to for guidance.

Calinou:
This part was removed from the latest version of the GDScript style guide, making this pull request obsolete. Closing.

Thanks for your contribution anyway :slightly_smiling_face:

PR #2736, opened 2019-09-01
Mention new Firefox security policies in Exporting for the Web

Calinou:
Firefox 68 and later cannot load resources from the local filesystem anymore, so a Web server needs to be used.

Since it will also apply there, this pull request should be cherry-picked into the `3.1` branch to make sure people don't miss it :slightly_smiling_face:

See https://github.com/godotengine/godot/issues/31866.


akien-mga:
Thanks!

PR #2740, opened 2019-09-03
Improve One-click deploy documentation

Calinou:
- Shorten text to be more concise
- Remove references to BlackBerry 10, whose support was removed a long time ago

Note that I didn't mention the current support for one-click deploy with the HTML5 platform, since it may be removed soon until https://github.com/godotengine/godot/issues/16245 is implemented.


akien-mga:
Thanks!

PR #2741, opened 2019-09-04
Fix `res://` appearing as an URL in Scenes and nodes

### LINKED ISSUE #2737: res:// is treated as a link
The [Editor section of the Scenes and nodes tutorial](https://docs.godotengine.org/en/3.1/getting_started/step_by_step/scenes_and_nodes.html#editor) explains what "res://" means. Every time that "res://" is mentioned, it appears as a clickable link that leads to the URL "res://". "res://" is not a valid web address.

Calinou:
This closes #2737.


akien-mga:
I'd use inline code formatting instead.

basdp:
Can we merge this? It has been approved and verified for some time already.

Calinou:
Sure :slightly_smiling_face: Thanks for bringing it to my attention.

PR #2749, opened 2019-09-08
Added C++ flags and 'bin' dir Creation

Melendeze13:
The build process fails when it tries to find 'demo/bin'. When we create
our demo the bin directory is not created by default, so let just make
it here. The build process also fails on OSX because the project uses
c++11 features. Lets add the c++17 flag so this error no longer happens and
because its used in other platform builds.

### DIFF
--- a/tutorials/plugins/gdnative/files/cpp_example/SConstruct
+++ b/tutorials/plugins/gdnative/files/cpp_example/SConstruct
@@ -22,6 +22,9 @@ cpp_library = "libgodot-cpp"
 # only support 64 at this time..
 bits = 64
 
+# Make 'demo/bin' directory
+Execute(Mkdir('demo/bin'))
+
 # Updates the environment with the option variables.
 opts.Update(env)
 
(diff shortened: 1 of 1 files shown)


tutorials/plugins/gdnative/files/cpp_example/SConstruct
```diff
@@ -50,10 +53,10 @@ if env['platform'] == "osx":
     env['target_path'] += 'osx/'
     cpp_library += '.osx'
     if env['target'] in ('debug', 'd'):
-        env.Append(CCFLAGS=['-g', '-O2', '-arch', 'x86_64'])
+        env.Append(CCFLAGS=['-g', '-O2', '-arch', 'x86_64', '-std=c++17'])
```
akien-mga: That's wrong. As you can see for other platforms, this should be appended to `CXXFLAGS` (C++), not `CCFLAGS` (C++ **and** C).

tutorials/plugins/gdnative/files/cpp_example/SConstruct
```diff
@@ -22,6 +22,9 @@ cpp_library = "libgodot-cpp"
 # only support 64 at this time..
 bits = 64
 
+# Make 'demo/bin' directory
+Execute(Mkdir('demo/bin'))
```
akien-mga: Should be `env['target_path']`, since this is configurable as seen above.
But it shouldn't be needed in the first place, SCons should take care of creating the needed path for its build artifacts automatically. Since later on the platform name is appended to `target_path`, and there is no `Mkdir` call, it seems to show that it works fine to create those subdirectories.

akien-mga:
CC @BastiaanOlij @willnationsdev

PR #2753, opened 2019-09-10
Rewrite the entire Matrices and Transforms article

### LINKED ISSUE #1353: a few "Matrices and transforms" issues
- The equations are images, they can be typeset using :math: and this might apply to some other pages maybe
- All matrices are transposed. It's the columns, not rows, are the basis vectors. Transform2D internally stores the basis in a transposed form, which I couldn't convince reduz to fix, but the 3D implementation with Basis is correct such that the columns are basis vectors. (since I think it's totally crazy and Transform2D needs to be corrected, I'm not willing to explain this discrepancy between 2D and 3D).

### LINKED ISSUE #2174: "Matrices and transforms" article offers very little practical uses later on
The [Matrices and transforms](https://docs.godotengine.org/en/latest/tutorials/math/matrices_and_transforms.html#doc-matrices-and-transforms) article starts off nicely with clear explanations and several practical uses. However the subsections later on (affine inverse, multiplications, tips) just throw some math definitions at you (which I had trouble understanding even though I passed several math courses in college), and don't show any practical uses. Same for the 3D part at the end of article where several pictures and examples would be very useful.

### LINKED ISSUE #2726: Improve Transform2D tutorial or class documentation
I'm trying to figure out how to skew a Sprite with Transform 2D, as it was suggested by @bojidar-bg  [here](https://github.com/godotengine/godot/issues/31742#issuecomment-525743140).

Problem is the [ Transform2D class docs](https://docs.godotengine.org/en/3.1/classes/class_transform2d.html#class-transform2d-method-affine-inverse) are in are written in a language I don't speak, and the [tutorial ](https://docs.godotengine.org/en/3.1/tutorials/math/matrices_and_transforms.html)is also not much better:

```
Affine inverse

The affine inverse is a matrix that does the inverse operation of another matrix, 
no matter if the matrix has scale or the axis vectors are not orthogonal. 
The affine inverse is calculated with the affine_inverse() method:

var mi = m.affine_inverse()
pos = m.xform(pos)
pos = mi.xform(pos)
# pos is unchanged

If the matrix is orthonormal, then:

# if m is orthonormal, then
pos = mi.xform(pos)
# is the same is
pos = m.xform_inv(pos)
```

So I still don't know how to skew.
A skew example would be nice. Especially since the only alternative to skew anything apparently is to write Shader code which is not much easier either.

> A matrix is used similarly to a turtle. The turtle most likely had a matrix inside (and you are likely learning this many years after discovering Santa is not real).

???

aaronfranke:
Rewrite of https://docs.godotengine.org/en/latest/tutorials/math/matrices_and_transforms.html (source: https://github.com/godotengine/godot-docs/blob/master/tutorials/math/matrices_and_transforms.rst)

New article source: https://github.com/aaronfranke/godot-docs/blob/matrix-transform/tutorials/math/matrices_and_transforms.rst

Don't bother trying to read the diff, just look at the before and after. I decided it was best to rewrite the entire article from scratch, so almost nothing is remaining. I've removed all old images, all mentions of turtles, and all mentions of Santa Claus not being real.

The old article starts with a transformed object and works to deconstruct it to bring it back into its identity transform, and then throws some definitions, without a clear explanation as to why you'd want to do any of those things. The new article starts with the identity transform, explains how the basis vectors relate to the transformation matrix, and then shows how manipulating them will change the orientation of the object - starting from simple, building up to complex operations. It then offers a few practical applications titled as "the thing you want to do", with the body explaining "how to do it". Fixes #2174, and fixes #2726

The old article had the handedness all over the place, the new article always follows Godot's handedness. Few equations are used, the images primarily showcase relations between the transformation matrix and what it looks like applied to the Godot logo. Fixes #1353

The article focuses on 2D, which is the easiest to explain, and occasionally/briefly mentions how it works in 3D. I've also provided a small project that users can download and play around with, but one TODO is that this might be better to put in the `godot-demo-projects` repo. **EDIT:** See https://github.com/godotengine/godot-demo-projects/pull/366

I also made a PR to the main Godot repo to improve matrix editing, since I noticed and was annoyed by bugs and inconsistencies while writing this article: https://github.com/godotengine/godot/pull/32068

The following users have submitted issues and may wish to give their two cents, so I'll ping them: @cbscribe @tagcup @wingedadventurer @golddotasksquestions


cbscribe:
Thanks for this - I look forward to reading it!

tagcup:
A few comments:

- Scale is applied before rotations

- I disagree with the comment regarding quaternions, why is it even there? The write up is about matrix representation of rotations, not quaternion representations

- It's probably better to cite a standard linear algebra book, maybe oriented for graphics programming, rather than citing YouTube videos or websites by a random dude on the internet.

aaronfranke:
What do you mean by scale is applied before rotations? If a user calculates a rotation using sine and cosine, they need to multiply those outputs by the scale afterwards.

The videos from this "random dude on the internet" are extremely good at explaining linear algebra, provide an animated geometric understanding, and are vastly better than any textbook which requires more time and provides a worse explanation. Collectively they have 6.5 million views.

tagcup:
If you apply scale after rotation, when you rotate a "tall" sprite by 90 degrees, it'd become a rotated fat sprite. This is why Godot and all other programs apply scale first, and rotation afterwards.

So? There are videos on youtube about spirits religions ghosts etc with millions of views as well. Should we throw away science because it attracts less YouTube viewers? Mathematics isn't a social network popularity contest.

tagcup:
No to mention no one can learn about quaternions from those videos linked, sorry to be frank but they're just garbage

tagcup:
No to mention no one can learn about quaternions from those videos linked, sorry to be frank but they're not useful for teaching purposes.

If you already know how quaternions work and relate to 3d rotations, yes, they're nice animations for *visualizing* a quaternion based rotation, which is also different from understanding the math behind it, but you can't realistically expect people can learn quaternions or group theory in a few of 5 minute graphical videos

clayjohn:
@tagcup @aaronfranke I see three differing issues here that should be addressed:
**1) Whether a scale is applied before or after rotation:**
This point ought to be sorted out and corrected if it is incorrect. For factual issues it is important that everything stated in the tutorial be correct. Unlike the below issues which are a matter of pedagogical preference. 

**2) Whether it is worth mentioning Quaternions in a matrices and transformations article:**
The mention of Quaternions is brief and helpful for users to point them to another option. In the future an article on Quaternions would be appropriate and we could link to it here. 

**3) Whether short youtube videos are appropriately included in a linear article article:**
Keep in mind that this is an article on the offical docs for a game engine. This is not a university-level textbook on linear algebra. We are not trying to give readers a fullsome education on mathematics but simply to introduce them to the bare minimum concepts they need to work with Godot. My preference is strongly in favour of explaining things in an accessible way rather than in a comprehensive way.

aaronfranke:
I removed the note about scaling being after rotation. I originally wrote this note in relation to the particular piece of example code, but you're right that conceptually the other way makes sense, and now that I think about it the note only served to confuse most users (who'd just copy the code as-is).

tagcup:
The internet is full of resources like that, I would argue they're one of the main reasons for this state of confusion, perplexion and misunderstanding regarding linear algebra, groups and quaternions among 3D game programmers.

Things can be made simpler but there is a limit to that. You can't skip all the basics and hope to understand a subject from some handwaving arguments and pretty pictures.

When you dumb down, for example, quantum mechanics, you get people writing books on how they found the Secret and new law of gravitation and how positive thinking will create change the universe in their favor because constructive interference.

I wouldn't expect anyone whose attention span is 5 minutes and can't bother reading can understand the use of quaternions, and it's just misleading to suggest otherwise.

But why not just go ahead. It's just silly for me that this can be changed.

aaronfranke:
I'll merge this, hope that's OK (and it has been half a year since @cbscribe said he'd review this). It's likely that this can still be improved further due to the sheer size of this article, but in its current state it is undoubtedly better than the existing text.

PR #2764, opened 2019-09-15
Fixed caption in tutorial

leptoGen:
Caption suggests clicking "Anchor" button, but "Layout" button seems intended.


akien-mga:
Thanks!

PR #2765, opened 2019-09-16
Update Singletons (AutoLoad) tutorial for Godot 3.1

BenjaminRiley:
Changed text to reflect browse button changing from `..` to 📁


getting_started/step_by_step/singletons_autoload.rst
```diff
@@ -116,7 +116,7 @@ it inherits from ``Node``:
 
 The next step is to add this script to the autoLoad list. Open
 ``Project > Project Settings`` from the menu, switch to the "AutoLoad" tab and
-select the script by clicking the ``..`` button or typing its path:
+select the script by clicking the browse button (|Browse|) or typing its path:
```
akien-mga: I wouldn't add icons in the docs, unless this is done in a systematic manner everywhere (which is not the case, this is the first attempt at adding a given icon to the docs).

IMO mentioning the "Browse" button is enough, but it could also be outlined in red in the following screenshot.

BTW that screenshot needs to be fixed, as it has a typo `Gloabl.gd`.

BenjaminRiley: [Introduction to the 2D animation features](https://docs.godotengine.org/en/latest/tutorials/animation/introduction_2d.html) uses icons in a few places, that's what I based my changes on.

akien-mga: Fair enough then :)

BenjaminRiley:
Section looks like this now:

image

akien-mga:
Thanks!

PR #2766, opened 2019-09-17
Document all comments that disable warnings

Kaligule:
These are the 3 types that I found in the code:

- \# warning-ignore:id_of_warning
- \# warning-ignore-all:id_of_warning
- \# warnings-disable

I found them at https://github.com/godotengine/godot/blob/24e1039eb6fe32115e8d1a62a84965e9be19a2ed/modules/gdscript/gdscript_tokenizer.cpp#L558
If there are any more they should also be added.

Having these documented could help people that fight against the warning system - currently it seems quite common to disable warnings alltogether (from what I read on the different support channels).

### DIFF
--- a/getting_started/scripting/gdscript/static_typing.rst
+++ b/getting_started/scripting/gdscript/static_typing.rst
@@ -327,7 +327,7 @@ script editor’s status bar. The example below has 3 warnings:
    warning system example
 
 To ignore specific warnings in one file, insert a special comment of the
-form ``#warning-ignore:warning-id``, or click on the ignore link to the
+form ``# warning-ignore:warning-id``, or click on the ignore link to the
 right of the warning’s description. Godot will add a comment above the
 corresponding line and the code won’t trigger the corresponding warning
 anymore:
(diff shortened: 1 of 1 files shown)


getting_started/scripting/gdscript/static_typing.rst
```diff
@@ -327,7 +327,7 @@ script editor’s status bar. The example below has 3 warnings:
    warning system example
 
 To ignore specific warnings in one file, insert a special comment of the
-form ``#warning-ignore:warning-id``, or click on the ignore link to the
+form ``# warning-ignore:warning-id``, or click on the ignore link to the
```
aaronfranke: I don't know if it's desirable to start warning ignore code with spaces. I prefer to use a space at the beginning of English comments, but not have a space for code, so that I can tell which is which. I would put ignoring warnings into that second category.

Kaligule: I think this is a matter of taste. Personally I prefer the space to be there because it makes the text more readable. I never thought that anyone would like it withour the space better, so I just "fixed" it.

Should I remove the space here?

Kaligule:
I am very new to this repo (and Godot), I would explaining all the warnings in a separate branch. What is the normal workflow here?

akien-mga:
Thanks!

For documenting all warnings, this should be done directly in the ProjectSettings.xml where they are listed (and actually they are already documented there).

PR #2781, opened 2019-09-23
Adds underscore before on_Button_pressed()

wandomPewlin:
so it matches the default name generated by the editor. Otherwise the plugin will not work, if the reader followed the instruction completely.


akien-mga:
Thanks!

PR #2784, opened 2019-09-24
Added null check for GDScript to mirror C# Code

jaggygames:
I also re-worded a few sentences to make the intent a bit clearer.


akien-mga:
Thanks!

PR #2792, opened 2019-09-28
Update debug images for Godot 3.2, add information and reword some de…

skyace65:
Updates the debug page images for Godot 3.2. All images have gone through png compression. 

Adds information on the debug project settings, adds information on the script editor debug tools and options, and adds information on the new network profiler, the description of it was taken from the PR that added it. I also reworded parts of the page that sounded odd or inaccurate.


tutorials/debug/overview_of_debugging_tools.rst
```diff
@@ -51,56 +53,109 @@ Sync Script Changes
 Any script that is saved will be reloaded on the running game.
 When used remotely on a device, this is more efficient with network filesystem.
 
+Script Editor Debug Tools and Options
```
akien-mga: Note that we don't use American Title case for section titles in the Godot docs. This should be "Script editor debug tools and options".

akien-mga: There's an exception for section titles that mirror property names in the engine which are capitalized, which is what is used in previous sections here, so I guess that's what lead you to use this style here.

akien-mga:
Thanks!

PR #2794, opened 2019-09-28
Add note about parameter limit in C++ modules

### LINKED ISSUE #2450: Mention limited number of parameters when creating a C++ module.
I've recently downloaded and compiled Godot, and I've been messing around with the module system, where I had created a subclass of TileMap with this function in:

```cpp
void set_cell_rect(int p_x, int p_y, int p_w, int p_h, int p_tile, bool p_flip_x = false, bool p_flip_y = false, bool p_transpose = false);
```

which has 8 parameters. When compiling, it gave this error (I've only included the first few lines, as it's very long and it only varies by a little bit):

```
1>D:\godot\core/class_db.h(256): error C2672: 'create_method_bind': no matching overloaded function found
1>modules\my_first_module\ExtendedTileMap.cpp(8): note: see reference to function template instantiation 'MethodBind *ClassDB::bind_method<MethodDefinition,void(__cdecl ExtendedTileMap::* )(int,int,int,int,int,bool,bool,bool)>(N,M,const Variant &,const Variant &,const Variant &)' being compiled
1>        with
1>        [
1>            N=MethodDefinition,
1>            M=void (__cdecl ExtendedTileMap::* )(int,int,int,int,int,bool,bool,bool)
1>        ]
1>D:\godot\core/class_db.h(256): error C2784: 'MethodBind *create_method_bind(R (__cdecl T::* )(P1,P2,P3,P4,P5) const)': could not deduce template argument for 'R (__cdecl T::* )(P1,P2,P3,P4,P5) const' from 'M'
1>        with
1>        [
1>            M=void (__cdecl ExtendedTileMap::* )(int,int,int,int,int,bool,bool,bool)
1>        ]
1>d:\godot\core\method_bind.gen.inc(4386): note: see declaration of 'create_method_bind'
```

With the help of the Discord server, I discovered that there is a header file for this exact problem, `core/method_bind_ext.gen.inc`, which increases the limit from 5 to 13.

I think there should be at least a note of this in the C++ Modules section of the documentation, as there will be programmers like me who will want a number of parameters in some functions.

Hope this helps! Thank you for creating one of the best bits of documentation I've ever seen, by the way!

skyace65:
Adds a note in the C++ module page about the parameter limit of 5 and how to raise it. I don't work with C++ normally so if there's a better way to phrase it please tell me. Closes #2450

### DIFF
--- a/development/cpp/custom_modules_in_cpp.rst
+++ b/development/cpp/custom_modules_in_cpp.rst
@@ -224,6 +224,10 @@ You can then zip it and share the module with everyone else. When
 building for every platform (instructions in the previous sections),
 your module will be included.
 
+.. note:: There is a paramter limit of 5 in C++ modules for things such
+          as subclasses. This can be raised to 13 by including the header
+          file ``core/method_bind_ext.gen.inc``.
+
 Using the module
 ----------------
 


development/cpp/custom_modules_in_cpp.rst
```diff
@@ -224,6 +224,10 @@ You can then zip it and share the module with everyone else. When
 building for every platform (instructions in the previous sections),
 your module will be included.
 
+.. note:: There is a paramter limit of 5 in C++ modules for things such
+          as subclasses. This can be raised to 13 with the header file
```
Calinou:
```suggestion
          as subclasses. This can be raised to 13 by including the header file
```

development/cpp/custom_modules_in_cpp.rst
```diff
@@ -224,6 +224,10 @@ You can then zip it and share the module with everyone else. When
...
+          ``core/method_bind_ext.gen.inc``
```
Calinou:
```suggestion
          ``core/method_bind_ext.gen.inc``.
```

skyace65:
That should be fixed now.

akien-mga:
Thanks!

PR #2797, opened 2019-09-29
Update a paragraph in the AssetLib page for clarity

### LINKED ISSUE #2787: Asset Library: Explain clearly the fact that in-project Asset libary gets imported directly into the currently active project
https://docs.godotengine.org/en/3.1/tutorials/assetlib/using_assetlib.html

> Notably, you can only fetch assets for the current version of Godot you are running. Also, you can only download Projects, Demos and Templates from the Project Manager view of the AssetLib, while Addons (tools, scripts, materials etc.) can only be downloaded from the in-project AssetLib.

There needs to be a paragraph next to this mentioning that in-project Asset library gets imported directly into the currently active project. 
This would have saved me enormous amount of time over the last year learning Godot.

skyace65:
Reworded a paragraph on the AssetLib page to make it clear that addons can be downloaded and added to the current project using the in-project AssetLib. Closes #2787

### DIFF
--- a/tutorials/assetlib/using_assetlib.rst
+++ b/tutorials/assetlib/using_assetlib.rst
@@ -124,9 +124,9 @@ Similarly to the web version of the AssetLib, here you can search
 for assets by category, name, and sort them by factors such as name or edit date.
 
 Notably, you can only fetch assets for the current version of Godot you are running.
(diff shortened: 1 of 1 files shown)


akien-mga:
Thanks!

PR #2801, opened 2019-09-29
Replace "use Google" with "search online"

anoadragon453:
Other search engines exist, no need to suggest just one.


Calinou:
Thanks!

PR #2802, opened 2019-09-29
Update WASM statement

anoadragon453:
WASM is supported in Firefox, Chrome, Safari and Edge: https://webassembly.org/ Update the WASM statement to reflect this.


Calinou:
Does the Godot HTML5 export actually work on Safari and Edge though? :slightly_smiling_face:

skyace65:
I just tried exporting a project to test this and I got it working in Edge with only a small error when starting. I'm probably doing something wrong in the workflow since I couldn't get Firefox to play it, but I think it's safe to say it works in Edge.

akien-mga:
Most browsers seem to support WASM indeed: https://caniuse.com/#feat=wasm

akien-mga:
Thanks!

PR #2806, opened 2019-09-30
Fix titles in Vector Math page

anoadragon453:
Fix the bullet point titles on the Vector math page.

Before:

image

After:

image


cbscribe:
Thanks!

PR #2809, opened 2019-10-01
Clarify the description of scaling a matrix

anoadragon453:
Reading this paragraph confused me a bit, so I've tried to reword it so it's a bit clearer.


anoadragon453:
OK, I think that's better. Let me know if any other concepts should be taught in this section.

tutorials/math/matrices_and_transforms.rst
```diff
@@ -308,36 +308,25 @@ as well as :ref:`Spatial.to_local() <class_Spatial_method_to_local>` and :ref:`S
 Scale
 ~~~~~
 
-A matrix can be scaled too. Scaling will multiply the basis vectors by a
-vector (X vector by x component of the scale, Y vector by y component of
-the scale). It will leave the origin alone:
+A matrix can be scaled too. Scaling will multiply the X and Y vectors of the
+matrix by the x and y components of another vector (`X * x`, `Y * y`). It will
```
Calinou: Just a nitpick, should we capitalize "x" and "y" in "x and y components"? If not, we might want to write them as ``x`` and ``y`` instead (as if they're member variables in a type).

anoadragon453: > Just a nitpick, should we capitalize "x" and "y" in "x and y components"? If not, we might want to write them as x and y instead (as if they're member variables in a type).

The capitalization was to differentiate between the member variables and the vectors, but we can use code blocks instead :) Do you think the `X * x` bit be X * `x` instead as well?

> Also, if I recall corrctly, reStructuredText recommends using two backticks instead of one for inline code.

Oop, markdown on the brain. Good catch!

tutorials/math/matrices_and_transforms.rst
```diff
@@ -308,36 +308,25 @@ as well as :ref:`Spatial.to_local() <class_Spatial_method_to_local>` and :ref:`S
...
     m = m.Scaled(new Vector2(2, 2));
 
 .. image:: img/tutomat15.png
 
-These kind of operations in matrices are accumulative. It means every
-one starts relative to the previous one. For those who have been living
-on this planet long enough, a good reference of how transform works is
-this:
-
-.. image:: img/tutomat16.png
```
Calinou: The image should be removed from the repository, so we don't leave an unused image around :slightly_smiling_face:

anoadragon453: Good point.

tutorials/math/matrices_and_transforms.rst
```diff
@@ -308,36 +308,25 @@ as well as :ref:`Spatial.to_local() <class_Spatial_method_to_local>` and :ref:`S
...
 
-    // Make the basis twice its size.
+    // Scale the transform's vectors to twice their size
     var m = Transform2D.Identity;
     m = m.Scaled(new Vector2(2, 2));
 
 .. image:: img/tutomat15.png
 
-These kind of operations in matrices are accumulative. It means every
-one starts relative to the previous one. For those who have been living
```
clayjohn: I am not overly attached to the Santa joke. However, it is important to note the accumulative nature of matrix operations. Also, many people learn programming with a turtle graphics API, so a reference to a turtle API is useful, it doesn't need to carry the same tone if you find it inappropriate.

anoadragon453: > However, it is important to note the accumulative nature of matrix operations.

The language in this sentence was a bit confusing, but I get what it was trying to convey now. I'll add a note about the cumulative properties.

> Also, many people learn programming with a turtle graphics API, so a reference to a turtle API is useful, it doesn't need to carry the same tone if you find it inappropriate.

As a new user I had no idea what this was talking about, or what a turtle API is. Searching online doesn't give me much more of a clue. The image was a star, which didn't have anything to do with turtles.

Either I'm too young to get the reference or it needs to be explained a bit better :)

clayjohn: The image is indeed outdated and only recognizable to older people and should be removed. But turtle graphics APIs are still around. I believe scratch uses one, Processing can use one. https://en.wikipedia.org/wiki/Turtle_graphics

Even if users are not already familiar, playing around with a turtle is a great way to get an intuitive understanding of matrices.

anoadragon453: Ahh ok. Perhaps then the example should be updated to pose a hypothetical scenario where you want to tell a turtle how to move, and you talk to it in terms of matricies.

Then say that this was a way people manipulated matricies in some systems.

clayjohn: I don't think that is necessary. The idea is to capture the subset of readers who already understand turtle graphics and make the connection. I would just reference it like "for those of you that have used turtle graphics APIs before, this may seem familiar as turtle graphics use a matrix internally" 

Of course, use better language. You can also link to the wiki for users that are curious.

anoadragon453: I see, ok. I'll try to make that happen :)

tutorials/math/matrices_and_transforms.rst
```diff
@@ -308,36 +308,25 @@ as well as :ref:`Spatial.to_local() <class_Spatial_method_to_local>` and :ref:`S
...
-the scale). It will leave the origin alone:
+A matrix can be scaled too. Scaling will multiply the X and Y vectors of the
+matrix by the ``x`` and ``y`` components of another vector (``X * x``, ``Y * y``).
+It will leave the origin alone:
 
 .. tabs::
  .. code-tab:: gdscript GDScript
 
-    # Make the basis twice its size.
+    # Scale the transform's vectors to twice their size
```
akien-mga: We actually end comments with a dot on purpose, that's the style we chose.

tutorials/math/matrices_and_transforms.rst
akien-mga: Bump :)

mhilbrunner:
Thank you for this!

In the mean time, this article has already been improved in `master`, see its [history](https://github.com/godotengine/godot-docs/commits/master/tutorials/math/matrices_and_transforms.rst), in particular these commits:
f0ab9e05fc78f4f8da301bbcc2c18e7f4379de92
28a04d587b85f0f18c76ca30abb786c6c3a97a09

Or compare the live `latest` version:
https://docs.godotengine.org/en/latest/tutorials/math/matrices_and_transforms.html

So I'm closing this - still, thank you for your contribution, most of these places were improved :)

PR #2810, opened 2019-10-01
Remove redundant phrase

anoadragon453:
X is often simplified to Y, **which is exactly the same**.

The bolded portion is redundant.


Calinou:
Thanks!

PR #2819, opened 2019-10-02
added exemple for dot notation in class_dictionary

### LINKED ISSUE #2378: Document dictionary keys being case sensitive
https://github.com/godotengine/godot-docs/pull/2370

Keys being sensitive should IMO be mentioned here:
http://docs.godotengine.org/en/3.0/classes/class_dictionary.html

Nutriz:
Closes #2378. Small contribution to add "dot notiation" example in class documentation, like as in GDscript basics

### DIFF
--- a/classes/class_dictionary.rst
+++ b/classes/class_dictionary.rst
@@ -60,6 +60,7 @@ To add a key to an existing dictionary, access it like an existing key and assig
 
 ::
 
+    d.waiting = 14 # Add String "waiting" as a key and assign the value 14 to it.
     d[4] = "hello"  # Add integer 4 as a key and assign the String "hello" as its value.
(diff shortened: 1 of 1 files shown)


akien-mga:
Changes to the class reference should be done in the godotengine/godot repository, see https://docs.godotengine.org/en/latest/community/contributing/updating_the_class_reference.html

This file is generated from the original XML.

classes/class_dictionary.rst
```diff
@@ -60,6 +60,7 @@ To add a key to an existing dictionary, access it like an existing key and assig
 
 ::
 
+    d.waiting = 14 # Add String "waiting" as a key and assign the value 14 to it.
```
akien-mga: BTW there should be two spaces before the `#` to match the style used in that code block.

PR #2820, opened 2019-10-03
fix build errors with `make latexpdf`

drfloob:
This commit provides two small fixes.

Gif images are not supported by latexpdf (see [here](https://github.com/sphinx-doc/sphinx/issues/2852#issuecomment-239778182)). There are a handful of options
to resolve that issue, but the already-available imgconverter extension
takes care of converting them to supported image formats.

The pdf build was also failing due to the deprecated Sphinx `app.warn` used
in the sphinx_tabs extension. The `app.log` deprecation was already
handled in this function, and now `warn` is as well.

I do still need to force latexpdf to continue through errors due to poor support for Japanese unicode characters, but that may be something I can solve locally, I'm not sure. The resulting PDF is totally usable regardless.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -11,7 +11,8 @@ needs_sphinx = '1.3'
 
 # Sphinx extension module names and templates location
 sys.path.append(os.path.abspath('extensions'))
-extensions = ['gdscript', 'sphinx_tabs.tabs', 'sphinx.ext.imgmath']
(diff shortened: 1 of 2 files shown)


akien-mga:
Thanks!

akien-mga:
I reverted the `conf.py` change as it converted GIFs to PNG even for the HTML version, which is definitely not wanted.

PR #2821, opened 2019-10-03
Updated certificate importing instructions for Mono v6

piratesephiroth:
Fixed "Compiling with Mono "

*Bugsquad edit:* Fixes godotengine/godot#32284.

### DIFF
--- a/development/compiling/compiling_with_mono.rst
+++ b/development/compiling/compiling_with_mono.rst
@@ -23,7 +23,7 @@ from the Mono command line prompt, or from the regular command prompt if you add
 
     cert-sync --user cacert.pem
 
-Alternatively, you can do this with the following command, though it won't work for Mono v6 or newer ::
+Alternatively, you can use the following command, though it's deprecated and may only work up to Mono v5 ::
 
     mozroots --import --sync
 
--- a/development/compiling/compiling_with_mono.rst
+++ b/development/compiling/compiling_with_mono.rst
@@ -23,7 +23,7 @@ from the Mono command line prompt, or from the regular command prompt if you add
 
     cert-sync --user cacert.pem
 
-Aternatively, you can do this with the following command, though it won't work for Mono v6 or newer ::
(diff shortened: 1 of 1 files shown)


development/compiling/compiling_with_mono.rst
```diff
@@ -14,8 +14,16 @@ Requirements
...
+requests.
+
+First you have to download **curl**'s CA (Certificate Autorities) certificate bundle: https://curl.haxx.se/ca/cacert.pem
+
+Then you run the following command to import it (on Windows, you can run it
+from the Mono command line prompt, or from the regular command prompt if you added the Mono installation's bin directory to the PATH environemnt variable)::
+
+    cert-sync --user cacert.pem
+
+Aternatively, you can do this with the following command, though it won't work for Mono v6 or newer ::
```
neikeq: Is there any info somewhere to confirm mozroots no longer works with 6.0.0? If not, I would prefer to say `Alternatively, you can use the following command, though it's deprecated and may not work ::`

Also typo `Aternatively` -> `Alternatively`.

piratesephiroth: It seems recent nuget versions requires a new certificate storage format that mozroots doesn't support (see #32284)

image

akien-mga:
Tested the new method, I can confirm that it fixes godotengine/godot#32284 with Mono 6.6.0.156.

IMO we should simply remove the mention of `mozroots` since it apparently no longer works with 6.0+, and has been issuing deprecation warnings for a while. The `cert-sync` method is not a one liner but it seems reliable.

We could also give instructions like:
```
curl -LO https://curl.haxx.se/ca/cacert.pem
cert-sync --user cacert.pem
```
if we want it as an easy to use 2-liner (but it doesn't remove the used `cacert.pem`).

piratesephiroth:
yeah, mozroots works but it's deprecated (because it uses hardcoded urls) and isn't updated as quickly as cert-sync

akien-mga:
Thanks!

PR #2822, opened 2019-10-03
Update images in Scenes and Nodes for Godot 3.1 and add extra informa…

skyace65:
Updates the images in the Scenes and Nodes page for Godot 3.1, all images have gone through png compression. I also added information on choosing which renderer to use when making a project, and explained the root node buttons that appear in the scene tree dock in a empty scene. Also I could use a second opinion on what I added to explain the root node buttons, I just want to make sure I didn't explain too much or too little about choosing root nodes.


getting_started/step_by_step/scenes_and_nodes.rst
```diff
@@ -86,7 +86,11 @@ appears. This helps developers manage their projects.
 .. image:: img/project_manager.png
 
 To create a new project, click the "New Project" option. Choose and create a
-path for the project and specify the project name "New Project":
+path for the project and specify the project name, or use the defualt one. You
```
akien-mga: Typo def**au*lt

akien-mga:
Thanks!

PR #2824, opened 2019-10-04
Update Signals images for Godot 3.2 and tweak tutorial

skyace65:
Updates all images in the signals tutorial for Godot 3.1, all images have gone through png compression. I altered the tutorial so the root node is a Node2D instead of Node, since Node shouldn't be used as a root node. I also clarified the location of the "Connect..." button.


akien-mga:
> I altered the tutorial so the root node is a Node2D instead of Node, since Node shouldn't be used as a root node.

It's not true that Node shouldn't be used as a root node though. In the context of this tutorial it's perfectly fine to use a Node as you don't need any Node2D features like the transform.

That being said, since you redid all the screenshots with Node2D already it doesn't hurt to change it.

akien-mga:
Thanks!

PR #2825, opened 2019-10-05
Update documentation guidelines for new simplified GitHub procedure

Calinou:
The edit URL is now customizable on Read the Docs, which means it's no longer necessary to edit the URL manually. See 24f61b36.


akien-mga:
Thanks!

PR #2828, opened 2019-10-06
Fix a function name in GDNative C tutorial

sheepandshepherd:
The function it referred to as `godot_singleton_init` is actually called [`godot_gdnative_singleton`](https://github.com/godotengine/godot/blob/1d9233c3882afe888b9396f7f2aac917d4dcac4d/modules/gdnative/register_types.cpp#L276).


Calinou:
Thanks!

Calinou:
PS: Should the error message in https://github.com/godotengine/godot/blob/1d9233c3882afe888b9396f7f2aac917d4dcac4d/modules/gdnative/register_types.cpp#L280 be updated?

sheepandshepherd:
Good idea, it should use the library's prefix. I think there are a few other error messages that also leave out the error message.

PR #2829, opened 2019-10-06
Supplement

chesscommands:
The part that was revised this time was added as a supplement because it uses words that have not been explained so far.

Line 168 explains the menu options.
But before that, it doesn't explain what the menu options are, so it might be more helpful to completely rewrite them.

### DIFF
--- a/getting_started/step_by_step/ui_main_menu.rst
+++ b/getting_started/step_by_step/ui_main_menu.rst
@@ -204,7 +204,7 @@ need two containers as children of our ``HBoxContainer``: a
    sitting aside from it
 
 In the node tree, select all the ``TextureRect`` nodes that should go on
-the left side: the logo, the menu options and the version note. Drag and
+the left side: the logo, the menu options (Continue, NewGame, Options) and the version note. Drag and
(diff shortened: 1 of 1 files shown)


getting_started/step_by_step/ui_main_menu.rst
```diff
@@ -204,7 +204,7 @@ need two containers as children of our ``HBoxContainer``: a
    sitting aside from it
 
 In the node tree, select all the ``TextureRect`` nodes that should go on
-the left side: the logo, the menu options and the version note. Drag and
+the left side: the logo, the menu options (Continue, NewGame, Options) and the version note. Drag and
```
cbscribe: This should also have an Oxford comma (as that's the standard we've adopted for the docs).

NathanLovato: What Chris means is we should have a comma before "and", like so:

> menu options (Continue, NewGame, Options), and the

NathanLovato:
According to your branch name, it looks like you edited the file from GitHub, so I'll merge and add the missing comma for you.

Thank you for the patch!

PR #2831, opened 2019-10-06
Fix the style used in an outdated example

sheepandshepherd:
The `If true` example doesn't follow the documentation style (missing comma after `If [code]true[/code]` and it includes the default value). This updates it to that property's current documentation.


NathanLovato:
Thanks :)

PR #2840, opened 2019-10-09
solution for no visual studio compiler found

girng:
Added solution for `warning: No version of Visual Studio compiler found`


Calinou:
Installing SCons is already described at the [top of the page](https://docs.godotengine.org/en/latest/development/compiling/compiling_for_windows.html#setting-up-scons), I don't see a point in mentioning it twice.

akien-mga:
I think the idea was to specify that you need SCons 3.1.1 or later for the latest VS.

PR #2842, opened 2019-10-10
Mention additional MinGW compilation troubleshooting tips

Calinou:
See https://github.com/godotengine/godot/issues/32695.


akien-mga:
Thanks!

PR #2843, opened 2019-10-10
Improvements in "Saving Games" tutorial

pouleyKetchoupp:
Made a pass on this tutorial with a few general improvements to make it easier to use:
- Check for common errors and display error messages (node not instanced, missing save function)
- Proper logic for checking there's a line left when iterating through the save file
- Extra comments and different variable names to make the code more clear


akien-mga:
Thanks!

PR #2844, opened 2019-10-10
Clarify how the "step" value behaves in GDScript exports

Calinou:
See https://github.com/godotengine/godot/issues/32715.

@AzureDVBB Is this considered clear enough, or should I add more detail? I'd like to keep the section relatively short.

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_exports.rst
+++ b/getting_started/scripting/gdscript/gdscript_exports.rst
@@ -92,7 +92,7 @@ Examples
     export(int, 20) var i
     # Allow integer values from -10 to 20.
     export(int, -10, 20) var j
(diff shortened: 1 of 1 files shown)


AzureDVBB:
Did a blind 'taste' test with two of my friends and I think the wording `# Allow floats from -10 to 20 and snap the value to multiples of 0.2` is clearer.
Though adding above this the edge case and the series it produces I believe makes it clearer how it works, like so:
```# Allow integers from 1 to 9 and snap value to multiples of 2 resulting in the series (2..4..6..8..9)
export(int, 1, 9, 2) var i```

Calinou:
@AzureDVBB Thanks for your feedback :slightly_smiling_face: I amended the pull request.

akien-mga:
Thanks!

Issue #2847, opened 2019-10-11
First spatial shader tutorial part 2 confusion


taner1es:
[Your first Spatial shader: part 2 - Advanced effects: waves](https://docs.godotengine.org/en/latest/tutorials/shading/your_first_shader/your_second_spatial_shader.html#advanced-effects-waves)

![resim](https://user-images.githubusercontent.com/19561417/66668095-192cf080-ec5d-11e9-9bcf-31bf3d57dde2.png)

![resim](https://user-images.githubusercontent.com/19561417/66668055-0b776b00-ec5d-11e9-9df1-87d0ece96497.png)

Problem is "wave" function called with 2 parameter but there is no function that we declared in the tutorial wave has 2 parameter

I ended tutorial like below, but still I'm not sure that i did correctly because of the lack of explanations. 

```
float wave(vec2 position, float power){
	position += texture(noise, position / 10.0).x * 2.0 - 1.0;
	vec2 wv = 1.0 - abs(sin(position));
	return pow(1.0 - pow(wv.x * wv.y, 0.65), power);
}

float height(vec2 position, float time) {
  float d = wave((position + time) * 0.4, 8.0) * 0.3;
	d += wave((position - time) * 0.3, 8.0) * 0.3;
	d += wave((position + time) * 0.5, 4.0) * 0.2;
	d += wave((position - time) * 0.6, 4.0) * 0.2;
	return d;
}
```

So I thought it is worth to review the tutorial.

clayjohn:
Great catch. That was a mistake. I got rid of the "power" parameter to simplify the implementation, but forgot to remove the second argument inside the "height" function.

PR #2848, opened 2019-10-12
Fixed spatial shader part2 wave effect redundant parameter usage

taner1es:
Related to #2847 

Inconsistent function calling solved.


clayjohn:
Thanks!

PR #2851, opened 2019-10-12
Gettext pybabel

### LINKED ISSUE #2850: Use Babel for translation?
Hi,

I am used to using gettext for all my translation needs, so I was taken aback by the fact that Godot has poor support for PO files and recommends using CSVs. Being able to accept translations from the community on Transifex has helped a lot with localizing my open-source projects.

As you mention, xgettext will not be able to automatically extract strings to be translated. So I got to work and **I wrote a plugin for Babel to be able to extract strings from scene files** (GDScript is close enough to Python that the Python parser can extract those).

So using a config like this, Godot will extract strings automatically to a POT file:
```
[python: **.gd]
encoding = utf-8

[godot_scene: **.tscn]
encoding = utf-8
```
You just have to run `pybabel extract -F babelrc -k Label/text -k LineEdit/placeholder_text -k tr -o translations.pot .`

The plugin is [here on GitHub](https://github.com/remram44/pybabel-godot) and installable using [`pip install babel babel-godot`](https://pypi.org/project/Babel-Godot/). I also made [an example project](https://github.com/remram44/godot-l10n-example) with a script that extract strings automatically.

Is there interest in adding this to the [Localization using gettext](https://docs.godotengine.org/en/3.1/tutorials/i18n/localization_using_gettext.html) page? I hate to be pushing my own code on other people but this seems like it would be useful. I'm also not sure if the process of setting up Babel is too complex for the target user going through the localization tutorial.

I also welcome comments on babel-godot itself. For example, I wrote it so that which properties should be translated is specified by `-k`, for example `-k text` will translate all "text" properties (found on most Controls) and `-k LineEdit/placeholder_text` will translate the property "placeholder_text" on LineEdit nodes. I'm not sure how friendly that is in practice.

I'll see if I can do something about supporting plural rules next.

remram44:
Fixes #2850?

This mentions the plugin babel-godot I just made, which can extract strings for localization from .tscn files (as well as GDScript via babel's Python parser).

I also tried to make the workflow less scary by mentioning that msginit can usually be replaced by a GUI or web-based app.

### DIFF
--- a/tutorials/i18n/localization_using_gettext.rst
+++ b/tutorials/i18n/localization_using_gettext.rst
@@ -96,9 +96,11 @@ while in the ``locale`` directory:
 
     msginit --no-translator --input=messages.pot --locale=fr
 
-
 The command above will create a file named ``fr.po`` in the same directory
 as the PO template.
+
+Alternatively, you can do that graphically using Poedit, or by uploading the
+POT file to your web platform of choice.
 
 Loading a messages file in Godot
 --------------------------------
@@ -124,7 +126,6 @@ automatically using the ``msgmerge`` tool:
 
     # The order matters: specify the message file *then* the PO template!
     msgmerge --update --backup=none fr.po messages.pot
-
 
 If you want to keep a backup of the original message file (which would be
 saved as ``fr.po~`` in this example), remove the ``--backup=none`` argument.
--- a/tutorials/i18n/localization_using_gettext.rst
+++ b/tutorials/i18n/localization_using_gettext.rst
@@ -58,8 +58,8 @@ install them.
 - **Linux:** On most distributions, install the ``gettext`` package from
   your distribution's package manager.
 
-Creating the PO template
-------------------------
+Creating the PO template (POT) manually
+---------------------------------------
 
 Godot currently doesn't support extracting source strings using ``xgettext``,
 so the ``.pot`` file must be created manually. This file can be placed anywhere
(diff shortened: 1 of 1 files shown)


tutorials/i18n/localization_using_gettext.rst
```diff
@@ -85,6 +85,41 @@ the translated string.
...
+You can then run pybabel like so:
+
+::
+
+    pybabel extract -F babelrc -k text -k LineEdit/placeholder_text -k tr -o godot-l10n.pot .
+
+Use the ``-k`` option to specify what needs to be extracted. In this case,
+arguments to :ref:`tr() <class_Object_method_tr>` will be translated, as well
+as properties named "text" (commonly used by Control nodes) and LineEdit's
+"placeholder_text" property.
```
Calinou: Out of curiosity, does pybabel offer a way for `babelrc` to define the list of functions/properties should be considered as localizable? There are many more properties in Godot which are automatically localized.

tutorials/i18n/localization_using_gettext.rst
```diff
@@ -85,6 +85,41 @@ the translated string.
...
+The Python tool pybabel has support for Godot and can be used to automatically
+create and update the POT file from your scene files and scripts.
+
+After installing ``babel`` and ``babel-godot``, for example using pip:
+
+::
+
+    pip install babel babel-godot
+
+write a mapping file (for example ``babelrc``) which will indicate which files
```
Calinou:
```suggestion
Write a mapping file (for example ``babelrc``) which will indicate which files
```

tutorials/i18n/localization_using_gettext.rst
remram44: pybabel has separate keywords/options, and keywords can only be specified on the command-line. But I can move those to an option instead.

Are you saying there's an official list of the properties of built-in nodes that should be localized? I could put this in the plugin, so there'd be no need for the user to list them.

remram44: There doesn't seem to be :(

Calinou: We can merge this pull request for now, I'll look at documenting the list of automatically-translated properties later.

Calinou:
Thanks!

PR #2855, opened 2019-10-14
Proof best practices: autoloads_versus_internal_nodes

NathanLovato:
I mostly simplified the text, moved some notes around, also used double backticks for inline code. Mostly proofing/editing work.

### DIFF
--- a/getting_started/workflow/best_practices/autoloads_versus_internal_nodes.rst
+++ b/getting_started/workflow/best_practices/autoloads_versus_internal_nodes.rst
@@ -4,9 +4,9 @@ Autoloads versus regular nodes
 ==============================
 
 Godot offers a feature to automatically load nodes at the root of your project,
-making allowing you to access them globally, that can fulfill the role of a
-Singleton: :ref:`doc_singletons_autoload`. These auto-loaded nodes are not freed
-when you change the scene from code with ``SceneTree.change_scene``.
+allowing you to access them globally, that can fulfill the role of a Singleton:
+:ref:`doc_singletons_autoload`. These auto-loaded nodes are not freed when you
+change the scene from code with ``SceneTree.change_scene``.
 
 In this guide, you will learn when to use the Autoload feature, and techniques
 you can use to avoid it.
@@ -20,14 +20,14 @@ many ways to avoid global state thanks to the node tree and signals.
 
 For example, let's say we are building a platformer and want to collect coins
 that play a sound effect. There's a node for that: the :ref:`AudioStreamPlayer
-<class_AudioStreamPlayer>`. But if we if call the ``AudioStreamPlayer`` while it
-is already playing a sound, the new sound interrupts the first.
+<class_AudioStreamPlayer>`. But if we call the ``AudioStreamPlayer`` while it is
+already playing a sound, the new sound interrupts the first.
 
 A solution is to code a global, auto-loaded sound manager class. It generates a
-pool of ``AudioStreamPlayer`` that cycle through as each new request for sound
-effects comes in. Say we call that class ``Sound``, you can use it from anywhere
-in your project by calling ``Sound.play("coin_pickup.ogg")``. This solves the
-problem in the short term but causes more problems:
+pool of ``AudioStreamPlayer`` nodes that cycle through as each new request for
+sound effects comes in. Say we call that class ``Sound``, you can use it from
+anywhere in your project by calling ``Sound.play("coin_pickup.ogg")``. This
+solves the problem in the short term but causes more problems:
 
 1. **Global state**: one object is now responsible for all objects' data. If the
    ``Sound`` class has errors or doesn't have an AudioStreamPlayer available,
@@ -39,6 +39,15 @@ problem in the short term but causes more problems:
 3. **Global resource allocation**: with a pool of ``AudioStreamPlayer`` nodes
    stored from the start, you can either have too few and face bugs, or too many
    and use more memory than you need.
+
+.. note::
+
+   About global access, the problem is that Any code anywhere could pass wrong
+   data to the ``Sound`` autoload in our example. As a result, the domain to
+   explore to fix the bug spans the entire project.
+
+   When you keep code inside a scene, and only one one or two scripts may be
+   involved in audio.
 
 Contrast this with each scene keeping as many ``AudioStreamPlayer`` nodes as it
 needs within itself and all these problems go away:
@@ -63,7 +72,7 @@ that feature for an individual scene using the :ref:`class_name
 
 When it comes to data, you can either:
 
-1. Create a new type of :ref:`Resource <class_Resource>` to share the data
+1. Create a new type of :ref:`Resource <class_Resource>` to share the data.
 
 2. Store the data in an object to which each node has access, for example using
    the ``owner`` property to access the scene's root node.
@@ -77,6 +86,8 @@ Auto-loaded nodes can simplify your code in some cases:
   database, then an autoload can be a good tool. There is no scripting API in
   Godot to create and manage static data otherwise.
 
+- **Static functions**: creating
+
 - **Systems with a wide scope**: If the singleton is managing its own
   information and not invading the data of other objects, then it's a great way to
   create systems that handle broad-scoped tasks. For example, a quest or a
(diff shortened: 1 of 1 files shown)


getting_started/workflow/best_practices/autoloads_versus_internal_nodes.rst
```diff
@@ -1,103 +1,95 @@
...
+effects comes in. Say we call that class ``Sound``, you can use it from anywhere
+in your project by calling ``Sound.play("coin_pickup.ogg")``. This solves the
+problem in the short term but causes more problems:
+
+1. **Global state**: one object is now responsible for all objects' data. If the
+   ``Sound`` class has errors or doesn't have an AudioStreamPlayer available,
+   all the nodes calling it can break.
+
+2. **Global access**: now that any object can call ``Sound.play(sound_path)``
+   from anywhere, there's no longer an easy way to find the source of a bug.
```
willnationsdev: > there's no longer an easy way to find the source of a bug.

I think that we should elaborate here on what we mean. *How* does this make it harder to find bugs? What could potentially happen and what could we have trouble doing in our attempt to solve it?

getting_started/workflow/best_practices/autoloads_versus_internal_nodes.rst
```diff
@@ -1,103 +1,95 @@
...
+Another reason to use an Autoload can be that you want to reuse the same method
+or data across many scenes.
+
+In the case of functions, you can create a new type of ``Node`` that provides
+that feature for an individual scene using the :ref:`class_name
+<doc_scripting_continued_class_name>` keyword in GDScript.
+
+When it comes to data, you can either:
+
+1. Create a new type of :ref:`Resource <class_Resource>` to share the data
```
willnationsdev: Period at the end of this sentence.

getting_started/workflow/best_practices/autoloads_versus_internal_nodes.rst
```diff
@@ -1,103 +1,95 @@
...
+Managing shared functionality or data
+=====================================
+
+Another reason to use an Autoload can be that you want to reuse the same method
+or data across many scenes.
+
+In the case of functions, you can create a new type of ``Node`` that provides
+that feature for an individual scene using the :ref:`class_name
+<doc_scripting_continued_class_name>` keyword in GDScript.
+
```
willnationsdev: Another thing that is important to discuss here for "in the case of functions" is the difference between pure and impure functions and how one can create a library of independently testable static functions within script classes. I elaborate more on this content in [section `1.b.i` of this reddit comment](https://www.reddit.com/r/godot/comments/dhjohq/best_ways_to_implement_systems_with_many/f3om27g/). If you would like, you could add a tidbit here in a second paragraph, before the "when it comes to data" section.

getting_started/workflow/best_practices/autoloads_versus_internal_nodes.rst
```diff
@@ -1,103 +1,95 @@
...
+- **Systems with a wide scope**: If the singleton is managing its own
+  information and not invading the data of other objects, then it's a great way to
+  create systems that handle broad-scoped tasks. For example, a quest or a
+  dialogue system.
+
+Until Godot 3.1, another use was just for convenience: autoloads have a global
+variable for their name generated in GDScript, allowing you to call them from
+any script file in your project. But now, you can use the ``class_name`` keyword
+instead to get autocompletion for a type in your entire project.
```
willnationsdev: I think this section is more useful if we actually include an example of using a singleton/script class as a namespace.

getting_started/workflow/best_practices/autoloads_versus_internal_nodes.rst
```diff
@@ -1,103 +1,95 @@
...
+
+.. note::
+
+   Autoload is not exactly a Singleton. Nothing prevents you from instancing
+   copies of an auto-loaded node. It is only a tool that makes a node load
+   automatically at the root of your scene tree, regardless of your game's node
```
willnationsdev: Correction: "that makes a node load automatically \[as a child of\] the root of your scene tree..." I would recommend explicitly telling people that it will exist at `/root/<autoload_name>`. Clarity is key here.

getting_started/workflow/best_practices/autoloads_versus_internal_nodes.rst
```diff
@@ -1,103 +1,95 @@
...
+   structure or the scene you try.
```
willnationsdev: I was confused by what was meant with "the scene you try" here. Maybe rephrase it? Did you mean the scene one tries to `change_scene` to?

NathanLovato: I meant the autoload will be there even when you press F6 to run an individual node or scene that is not your app's main scene

getting_started/workflow/best_practices/autoloads_versus_internal_nodes.rst
NathanLovato: I took this from the original article. I wouldn't use a singleton for things like these so I don't have examples from experience but I'd guess that:

1. You have to navigate to the Sound class to change the code
2. If you run out of streams and the Sound class cycles and e.g. replaces a long audio sample with a new one, it's not that easy to figure out why that's happening in the middle of a project, especially as you won't get a backtrace and when you try the scene in isolation with F6, everything may work fine.

Does that seem good enough of an example? In my exp. the issues of singletons are not obvious at first, they really start to hurt when your project grows and you have a hard dependency in many places at once. It's difficult to give an example of that.

getting_started/workflow/best_practices/autoloads_versus_internal_nodes.rst
NathanLovato: Yup definitely, that's actually mostly how we use them - function libraries.

NathanLovato:
Thanks, it's getting late here, I'll get to your feedback tomorrow!

getting_started/workflow/best_practices/autoloads_versus_internal_nodes.rst
willnationsdev: Well, those examples work well too. But my primary concern with things like this is that the location of a problem, namely the context in which the method is called, is no longer clear when the functionality can be accessed from anywhere. Rather than dealing with a subset of nodes internal to a scene, your potential domain of trouble-making code spans the entire project. Any code anywhere could have passed erroneous data to the `Sound` singleton. When code is kept inside a scene, there is no realy concern for this problem as only a handful of scripts are usually involved total, let alone dealing with the topic that the singleton handles.

getting_started/workflow/best_practices/autoloads_versus_internal_nodes.rst
```diff
@@ -1,103 +1,95 @@
...
+  dialogue system.
+
+Until Godot 3.1, another use was just for convenience: autoloads have a global
+variable for their name generated in GDScript, allowing you to call them from
+any script file in your project. But now, you can use the ``class_name`` keyword
+instead to get autocompletion for a type in your entire project.
+
+.. note::
+
+   Autoload is not exactly a Singleton. Nothing prevents you from instancing
```
willnationsdev: > instancing

Do we want to say `instantiating` here? I think we should.

getting_started/workflow/best_practices/autoloads_versus_internal_nodes.rst
willnationsdev: Ah, in that case, I would recommend converting `or the scene you try` to `or which scene you run`.

getting_started/workflow/best_practices/autoloads_versus_internal_nodes.rst
NathanLovato: Thinking of this again, I don't think this discussion is a good fit for a guide about "when and how to use auto-loaded nodes in Godot" because you can use a gdscript file that doesn't extend anything to create a library of static functions, i.e. with class_name

getting_started/workflow/best_practices/autoloads_versus_internal_nodes.rst
```diff
@@ -1,103 +1,109 @@
 .. _doc_autoloads_versus_internal_nodes:
 
-Autoloads versus internal nodes
+Autoloads versus regular nodes
+==============================
+
+Godot offers a feature to automatically load nodes at the root of your project,
+allowing you to access them globally, that can fulfill the role of a Singleton:
+:ref:`doc_singletons_autoload`. These auto-loaded nodes are not freed when you
+change the scene from code with ``SceneTree.change_scene``.
```
willnationsdev: Should we replace this with a direct link to the method in the API docs?

    change the scene from code with :ref:`SceneTree.change_scene <class_SceneTree_method_change_scene>`.

akien-mga:
What's the status on this PR?

NathanLovato:
I corrected the typos. The other notes from Will involve writing some new paragraphs, which I would rather do or have someone do in a separate PR.

But I transferred my docs fork some time ago and it seems GitHub lost track of it as a result. I'll open a new PR that addresses the typos Will pointed out.

PR #2857, opened 2019-10-16
Change printf to print (typo)

MintStudios:
I'm guessing printf should be print.


clayjohn:
In this case it is not referring to "print" the GDScript/python it is drawing a contrast between GDScript and C/C++ which use "printf" http://www.cplusplus.com/reference/cstdio/printf/

PR #2858, opened 2019-10-16
printf typo 2 : Change printf to print (typo)

MintStudios:
Again, this should be print. (I think)


getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -371,7 +371,7 @@ PoolRealArray store 32-bit single-precision "float" values.
 A sequence of characters in `Unicode format <https://en.wikipedia.org/wiki/Unicode>`_.
 Strings can contain
 `standard C escape sequences <https://en.wikipedia.org/wiki/Escape_sequences_in_C>`_.
-GDScript also supports :ref:`doc_gdscript_printf`.
```
clayjohn: This is not referring to a specific function. It is a link to the name of another doc. Removing the "f" will break the link

PR #2863, opened 2019-10-19
Change example signal name to animation_finished

e8johan:
The AnimationPlayer signal used to demonstrate the yield function is called "animation_finished".


Calinou:
Thanks!

PR #2866, opened 2019-10-20
Add proper extends / method calls to inspector plugins tutorial

Klowner:
Couple small changes which made the example work for me.
Note that this also required that I expose the `EditorSpinSlider` to GDScript which over here: https://github.com/godotengine/godot/pull/32944


akien-mga:
Woops, looks like this page was merged and never proofread... I'll merge this and do further fixes.

Klowner:
Ahh, sorry :/

akien-mga:
That was not a criticism for you, I just meant that the original page really needs a thorough review. I pushed fixes in 38258519.

PR #2868, opened 2019-10-20
GDScript style guide: Add comments section, and change quotes

aaronfranke:
This PR does two things:

* Add a section about placing spaces at the beginning of comments, but not with disabled code. This helps differentiate text comments from disabled code, and the rest of the style guide already has spaces at the beginning of comments.

* I replaced single-quote strings with double-quote ones to be consistent, a fairly minor change. This doesn't involve any [rules](https://github.com/godotengine/godot-docs/issues/2869) of which should be used, it only makes the guide itself consistent.


NathanLovato:
Thanks!

PR #2871, opened 2019-10-21
Update Particle Systems (2D) page for Godot 3.2

skyace65:
Updates all outdated images for Godot 3.2. All images have gone through png compression. I also added a section in the opening that explains the difference between Particles2D and CPUParticles2D.


akien-mga:
Thanks!

PR #2872, opened 2019-10-21
Improve the Custom builds for Android page

### LINKED ISSUE #2870: Maybe a little mistake in the Docs
I'm not sure if it's a mistake but maybe someone can look at this? 

https://docs.godotengine.org/en/latest/getting_started/workflow/export/android_custom_build.html

"Install Java

Android SDK does not come with Java, so it needs to be installed manually. Instal Java SDK (not runtime or JRE). OpenSDK 8 is recommended, otherwise Oracle’s Java SDK for version 8 will work. Later versions may not work for Android development."

Is OpenSDK 8 right or should it mean OpenJDK?

And it would be nice to include links to the websites where you can download the tools.

Calinou:
This closes #2870.

### DIFF
--- a/getting_started/workflow/export/android_custom_build.rst
+++ b/getting_started/workflow/export/android_custom_build.rst
@@ -3,63 +3,88 @@
 Custom builds for Android
 =========================
 
-Godot provides the option to use custom build Android templates. Instead of using the already pre-built template that ships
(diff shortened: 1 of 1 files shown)


NathanLovato:
Thank you :)

PR #2877, opened 2019-10-24
Script path in C# code is wrong.

Vikram1323:
CustomNode.cs, MyButton.cs and icon.png are in same directory(addons/plugin_name)


akien-mga:
Thanks!

PR #2880, opened 2019-10-25
Typo with the Animation introduction referring to the wrong target node class

20kdc:
Not much to say here. I've got another PR coming up which is probably slightly more interesting, but found this in the meantime


NathanLovato:
Thanks!

PR #2883, opened 2019-10-25
Add Example for custom signals with parameter

Dragoncraft89:
Clarifies the declaration of parameters for custom signals. See godotengine/godot#29501.

I'm not sure if the C# part is correct. That's the best I could find on the internet.

### DIFF
--- a/getting_started/step_by_step/signals.rst
+++ b/getting_started/step_by_step/signals.rst
@@ -167,6 +167,7 @@ You can also declare your own custom signals in Godot:
     extends Node
 
     signal my_signal
+    signal my_signal_with_arguments(x, y)
 
  .. code-tab:: csharp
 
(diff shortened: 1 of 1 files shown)


cbscribe:
While this is good information, I'm not sure it belongs in this doc. I'd prefer to keep the step-by-step section from getting too complex too fast. It may be preferable to have a more comprehensive signals document in the Tutorials section that covers all of the common use cases.

NathanLovato:
I rewrote that part in the signals docs already a while ago, so this functionality is documented in greater details in the gdscript docs: http://docs.godotengine.org/en/latest/getting_started/scripting/gdscript/gdscript_basics.html#signals

That's where a more comprehensive overview of signals belongs. We can always link to it from the getting started guide.

Also:

1. You don't need to declare the arguments to send data along with the signal. These signal arguments are just hints at the moment.
2. I'd recommend using concrete examples instead of programmer-like technical examples ("my_signal", "foobar"). Be it for beginners or for professional developers, concrete examples help to give a sense of feature's purpose.

Thanks for your contribution, but I agree with Chris, to me as well it doesn't belong in the step-by-step guides.

PR #2886, opened 2019-10-27
Show how to pass a value with a signal

johnniehard:
I had some trouble finding how to do this in the documentation, but had seen it used in some youtube tutorial. I think it would be fitting to explain the usage on this page.

I do not know the corresponding C# syntax, so I haven't changed that part.

### DIFF
--- a/getting_started/step_by_step/signals.rst
+++ b/getting_started/step_by_step/signals.rst
@@ -203,6 +203,56 @@ To emit a signal via code, use the ``emit_signal`` function:
             EmitSignal(nameof(MySignal));
         }
     }
+    
+A signal can also optionally declare one or more arguments. Specify the
+argument names between parentheses:
+
+.. tabs::
+ .. code-tab:: gdscript GDScript
+
+    extends Node
+
+    signal my_signal(value, other_value)
+
+ .. code-tab:: csharp
+
+    public class Main : Node
+    {
+        [Signal]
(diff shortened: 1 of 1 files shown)


getting_started/step_by_step/signals.rst
```diff
@@ -203,6 +203,50 @@ To emit a signal via code, use the ``emit_signal`` function:
             EmitSignal(nameof(MySignal));
         }
     }
+    
+A signal can also pass on a value, add it in parent:
```
cbscribe: Did you mean "parentheses" here? Note that a signal can pass any number of arguments, not just "a" value. Also, declaring them is optional. How about using the same wording as in the GDScript doc:

> You can write optional argument names in parentheses in the signal’s definition.

getting_started/step_by_step/signals.rst
```diff
@@ -203,6 +203,50 @@ To emit a signal via code, use the ``emit_signal`` function:
...
+
+ .. code-tab:: csharp
+
+    public class Main : Node
+    {
+        [Signal]
+        public delegate void MySignal();
+    }
+
+To pass the value add it as the second argument to the ``emit_signal`` function:
```
cbscribe: Again, this should not imply that there is only room for one argument.

Calinou:
Thanks!

PR #2891, opened 2019-10-29
Mention the SCons version requirement when using Visual Studio 2019

Calinou:
See #2840.

We could simplify the instructions by saying you need SCons 3.1.1 regardless of the Visual Studio version, but that's technically not the case :slightly_smiling_face:


akien-mga:
Thanks!

Issue #2860, opened 2019-10-17
ANGLE is outdated on Github


sprite-1:
On the [instructions for compiling for UWP](https://docs.godotengine.org/en/3.1/development/compiling/compiling_for_uwp.html), it says to use the `ms-master` branch on [this repo](https://github.com/Microsoft/angle) but that branch is gone, only text files remain.

vnen:
Well, good that upstream ANGLE now officially supports UWP again. But setting up the Google dev build is quite some work, we might not be able to build from ANGLE source anymore (at least not without requiring the dev to do the Google setup first).

sprite-1:
The last commit before they wiped out the repo [is here](https://github.com/microsoft/angle/tree/c61d0488abd9663e0d4d2450db7345baa2c0dfb6) so maybe that can be linked to instead of the main branch in the meantime?

vnen:
Sure, feel free to make the change (and also clarify it's an older commit). I'll have a look at this probably next week and update the build instructions.

skyace65:
Closing this issue as the compiling for UWP page on 4.x/master is completely gone now, and the 3.6 version has the workaround note added in #2892.

PR #2892, opened 2019-10-29
Add note to indicate deprecation of MS ANGLE repo

sprite-1:
Temporary workaround for #2860 until @vnen gets around to writing a more fleshed out build instruction.


akien-mga:
Thanks!

Issue #1809, opened 2018-10-07
Write a list for the demo in godot-demo-project


ZX-WT:
godotengine/godot-demo-projects#280 give me this idea of writing a summary for each demo.

The official demo repo has some simple projects that demonstrate the use of a particular node or feature. But even the user download the demo repo , it is difficult to know what is the focus of a project even a clear name is given to it and find the one he needs.

A list may give them a better idea, the list includes the name of it, the folder, a short summary and most importantly the nodes that it is using. This provide the info for new users what should they open for learning a particular feature.

I even think about a list in order of learning like the rigidbody demo, platformer demo and then pathfinding, so on. But some may not fit in the list.

We can put a link to this list next to the download button for the demo in the official website, this should  help new users.

pwab:
> A list may give them a better idea, the list includes the name of it, the folder, a short summary and most importantly the nodes that it is using. This provide the info for new users what should they open for learning a particular feature.

Great idea. Not sure where to put that list in the docs but there could be a subcategory for demos which could also include explanations (little tutorials) for each demo project.

aaronfranke:
Most of the demo projects now have README files in them, which covers the issue of having a description.

It would still be nice to point to the demo projects from within the docs, and perhaps even from the class reference. For example, KinematicBody would link to the Kinematic Character 3D demo.

Calinou:
@aaronfranke Demos could have a description in the `application/config/description` project setting so it displays when hovering a demo in the project manager.

aaronfranke:
@Calinou Good idea, I created https://github.com/godotengine/godot-demo-projects/pull/449 for that.

PR #2893, opened 2019-10-30
WIP added demo list

ZX-WT:
it's  hacktoberfest, I thought I could not joined due to my health issue, it seems I can still make it :)
This is a PR that I based on my idea #1809. This will take me some time, so I will just make this as a record.


akien-mga:
Fair enough to sneak it in before the end of the month and still take part in hacktoberfest, but in it's current state having a stub saying "here be documentation in the future" isn't really useful :)

PR #2895, opened 2019-10-31
Adding links to multilingual Discord servers

Omicron666:

### DIFF
--- a/community/channels.rst
+++ b/community/channels.rst
@@ -31,7 +31,7 @@ Other chats
 - `Matrix (IRC compatible) <https://matrix.to/#/#godotengine:matrix.org>`_
 - `Discord Server <https://discordapp.com/invite/zH7NUgz>`_
 
-Multilingual Communities
+Multilingual communities
 ------------------------
 
 - `Brazilian Discord Server <https://discord.gg/gWmQKBv>`_
--- a/community/channels.rst
+++ b/community/channels.rst
@@ -29,7 +29,16 @@ Other chats
 -----------
 
 - `Matrix (IRC compatible) <https://matrix.to/#/#godotengine:matrix.org>`_
-- `Discord <https://discordapp.com/invite/zH7NUgz>`_
+- `Discord Server <https://discordapp.com/invite/zH7NUgz>`_
+
+Multilingual Communities
+------------------------
+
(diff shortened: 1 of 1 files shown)


community/channels.rst
```diff
@@ -29,7 +29,16 @@ Other chats
 -----------
 
 - `Matrix (IRC compatible) <https://matrix.to/#/#godotengine:matrix.org>`_
-- `Discord <https://discordapp.com/invite/zH7NUgz>`_
+- `Discord Server <https://discordapp.com/invite/zH7NUgz>`_
+
+Multilingual Communities
```
akien-mga: Should be "communities", lowercase.

community/channels.rst
```diff
@@ -29,7 +29,16 @@ Other chats
...
+------------------------
+
+- `Brazilian Discord Server <https://discord.gg/gWmQKBv>`_
+- `French Discord Server <https://discord.gg/3jvf3XN>`_
+- `French Discord Server #2 <https://discord.gg/gZ3QJ5T>`_
```
akien-mga: Why are there two French Discord communities? How should users know which one to join?

Omicron666: I guess because people are free to create their own server and communities.
I can't really tell people which one to choose if not both.
Maybe, we could put title/name of servers, it may help. (in a way translators won't need to edit by the way)

There are also multilingual channels on main English Godot Discord, we could list available languages there.

By, there are also multiple dev oriented servers that have opened dedicated channels to Godot Engine, and that we could inventorize too.

mhilbrunner: IMO listing multiple Discords is fine, as a Discord user I'm able to decide myself which ones I want to join, if any. If the list becomes overwhelming, we can impose additional requirements (minimum size?)

I'd use the server name instead of numbering however, as the numbering could be seen as implying some kind of ranking, which we probably want to avoid.

Omicron666:
One of French ones is official, the other, bigger, for a youtuber dedicated to Godot tutorials

community/channels.rst
akien-mga: Agreed, I updated the PR accordingly.

akien-mga:
Thanks!

PR #2896, opened 2019-10-31
Adding guide for automatic pull rebase option

Omicron666:

### DIFF
--- a/community/contributing/pr_workflow.rst
+++ b/community/contributing/pr_workflow.rst
@@ -206,7 +206,7 @@ Therefore, the command that you should (almost) always use is:
 
 You may configure git to use ``--rebase`` option automatically for you when you pull, with:
 
-::
+.. tip::
 
     $ git config --global pull.rebase true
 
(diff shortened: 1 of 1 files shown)


community/contributing/pr_workflow.rst
```diff
@@ -204,8 +204,13 @@ Therefore, the command that you should (almost) always use is:
 
     $ git pull --rebase upstream master
 
+You may configure git to use ``--rebase`` option automatically for you when you pull, with:
```
akien-mga: This should be a `.. tip::`.

I'm not sure it's good advice though, it's important that people are aware of when they are rebasing or not. There are situations where `pull --rebase` is NOT the right thing to do.

Omicron666: What are those situations ?
docs doesn't specify them and advocate for rebase option usage without clear distinction.
should we document them ?

Calinou:
As @akien-mga said, I don't think this documentation is the right place to recommend changing `git pull`'s default behavior. While I personally use `pull.rebase true` by default, you need to understand its implications to be able to use it effectively. In other words, you need to know what to do when it asks you to "commit or stash your changes" when switching branches :slightly_smiling_face:

PR #2894, opened 2019-10-30
Added visual shader tut

ZX-WT:
Here's something I find missing in the docs, this is an short introduction to the visual shader editor.
It is mainly based on this article https://godotengine.org/article/major-update-for-visual-shader-in-godot-3-2

It is almost done, I just need to fix some of the text and add some images

PR #2897, opened 2019-10-31
added link to visual shader tutorial

ZX-WT:
This is a follow up PR for #2894. I added a link to the visual shader tutorial in the related part.

### DIFF
--- a/tutorials/shading/shader_materials.rst
+++ b/tutorials/shading/shader_materials.rst
@@ -55,6 +55,10 @@ If you create a visual shader the visual shader editor will open automatically.
 
 .. image:: img/visual_shader_editor.png
 
+.. note::
+
+   To learn more about visual shader, read :ref:`doc_visual_shader`
+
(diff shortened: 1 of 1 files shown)


tutorials/shading/shader_materials.rst
```diff
@@ -55,6 +55,10 @@ If you create a visual shader the visual shader editor will open automatically.
 
 .. image:: img/visual_shader_editor.png
 
+.. note::
+
+   To learn more about visual shader, read :ref:`doc_visual_shader`
```
akien-mga: There should be a dot at the end of the sentence. I'd also say "visual shader**s**"

akien-mga:
Why make this a separate PR instead of adding it to #2894?

ZX-WT:
In case the previous one is not accepted, I realized what I am thinking makes no sense at all... :thinking: 
I will just combine the PRs.

PR #2902, opened 2019-11-01
Some fixes for GLES2/GLES3 differences table

Chaosus:
Changes:

- Removed **outerProduct** - it was implemented in GLES2
- Added **textureGrad**
- Unify some other texture functions to avoid duplication
- Added missed parameter names

changes have been made by using https://truben.no/table/


akien-mga:
Thanks!

PR #2903, opened 2019-11-02
Mobs animation play

hamadmarri:
Mobs animations is not started. I think we need animation.play in _ready function


cbscribe:
This is not necessary, as the instructions have you set it in the Inspector:

cbscribe:
Closing, as there doesn't seem to be any more need for discussion.

PR #2904, opened 2019-11-02
Added "Instancing in code" tutorial to Instancing (continued) page

DeerTears:
The Instancing (continued) page has no instructions on how to instance with code in GDScript. There is no (easy-to-find) reference on doing this in the current docs, so I added one and tested my code to make sure it worked.

I am new to this type of formatting so feel free to give it a quick lookover for formatting and writing standards that I missed. I am glossing over a lot of the concepts here because I really wanted to get the groundwork for instructions out above all else, but I made sure to include some general metaphors to supplement the code. Later I'll look at adding pictures, a formalized tutorial and doing something with the old text.


cbscribe:
My main problem with this addition is that it includes scripting *before* scripting has been introduced (that happens in the next section). This means that readers who are going in order haven't yet learned *how* to even add a script yet, much less all of the things shown here.

Perhaps we should move this until after the "Scripting" section? Also, much of this is demonstrated in "Dodge the Creeps" as well.

akien-mga:
Indeed, this tutorial is one of the first of the Step by Step and is meant as a general introduction to the concept of instantiation. In-depth scripting examples are too early at this stage.

If more documentation about instantiation is needed, it should probably be in a dedicated tutorial, not the Step by Step.

DeerTears:
Understandable. That ethos makes sense, but it only makes me raise more questions about that line "It is also possible to create instances in code by using GDScript, see Instancing (continued)." when Instancing (continued) is the next page in the step-by-step anyways? Oh well, maybe that's another pull request for another time. To make sure there's no scripting lessons until much later on I'll gladly go out and create a new page for this tutorial to sit elsewhere on the docs.

Once I do make that new page though, I'd like to know if it'd still be appropriate to link that new tutorial page back to Instancing (continued) as an optional tutorial for "advanced Godot users". Related links are always nice in my opinion, and I'd also want to make sure it's not mandatory in any way for anyone reading through the step-by-step instructions.

mhilbrunner:
Nothing against related links :)

Thanks for your work on this! I'm going to close this PR as the content doesn't fit in this place as discussed above.

Of course a new PR is welcome, but keep in mind we're thinking about restructuring the docs and tutorials with Godot 4.0, so you may want to follow the discussion there:
https://github.com/godotengine/godot-docs/issues/3390

Thanks again!

PR #2905, opened 2019-11-02
Fixed code tags for true/false/null and a few property references

pouleyKetchoupp:
Changed some inconsistencies in tutorials after this issue was pointed out:
https://github.com/godotengine/godot/issues/33263


akien-mga:
Thanks!

PR #2906, opened 2019-11-02
Fixed http request example that was not working

ankeris:
Previously in the example signal function "_on_HTTPRequest_request_completed" was not registered, therefore had no effect. I added signal (request_completed) registration in the code snippet.

Also, this example helps users understand that you can register multiple different http request signals in one script


Calinou:
Thanks!

PR #2915, opened 2019-11-08
Clarify Particles2D spread parameter

### LINKED ISSUE #2878: Make the documentation on the Spread property of ParticlesMaterial a little more noob friendly
In the [Particle Systems (2D)#Spread](https://docs.godotengine.org/en/3.1/tutorials/2d/particle_systems_2d.html#spread) tutorial and in the [ParticlesMaterial](https://docs.godotengine.org/en/3.1/classes/class_particlesmaterial.html#class-particlesmaterial-property-spread) documentation they both say the Spread property of ParticlesMaterial changes the initial direction of the particle. Neither mention that this only happens if the particle also has a Velocity applied to it (which by default is set to 0.)

Maybe the actual documentation doesn't need to be updated but it couldn't hurt to add it to the tutorial. I mean, the particle is already moving but it turns out that's because of Gravity, not Velocity. And then when you do enable Velocity it shoots out to one side because the direction by default is set on the Y axis, not X.

"Spread doesn't spread and when I increase the Velocity they only spread to the right!"

All these factors are the perfect storm to confuse someone to the new engine

skyace65:
Clarifies that for the spread property to work the initial velocity property must be greater than 0. Closes #2878

### DIFF
--- a/tutorials/2d/particle_systems_2d.rst
+++ b/tutorials/2d/particle_systems_2d.rst
@@ -199,7 +199,8 @@ Spread
 
 This parameter is the angle in degrees which will be randomly added in
 either direction to the base ``Direction``. A spread of ``180`` will emit
-in all directions (+/- 180).
+in all directions (+/- 180). For spread to do anything the "Initial Velocity"
+parameter must be greater than 0.
 
 .. image:: img/paranim3.gif
 


akien-mga:
Thanks!

PR #2917, opened 2019-11-08
Removed inconsitent icon

rhode2nh:
Removed the yellow warning icon in the CollisionShape2D node to be consistent with other images in the tutorial.


cbscribe:
Thanks!

PR #2922, opened 2019-11-10
Add Alpine Linux dependencies oneliner in Compiling for X11

Calinou:
This adds an installation oneliner for [Alpine Linux](https://www.alpinelinux.org/). This distribution is popular in containers, so it can serve as a basis for Dockerfiles or similar.


akien-mga:
Thanks!

PR #2925, opened 2019-11-12
Clarifying hint_albedo converts color space

rainlizard:
Addressing: https://github.com/godotengine/godot/issues/24992 https://github.com/godotengine/godot/issues/11381

https://github.com/godotengine/godot/issues/26196
This guy's minimal reproduction project isn't using hint_albedo and the colors are messed up, I was looking at it to see how TextureArrays are used. (so +1 people who didn't know about this)

I find it funny they're called "hints" while they have special effects like converting color space. But putting that aside...

https://docs.godotengine.org/en/3.1/tutorials/shading/shading_reference/shading_language.html
http://docs.godotengine.org/en/3.0/tutorials/shading/shading_language.html

>As Godot’s 3D engine renders in linear color space, it’s important to understand that textures that are supplied as color (i.e. albedo) need to be specified as such for proper sRGB->linear conversion.

In 3.1 docs (compared to 3.0 docs) this part has been pushed down away from the hints section, thus losing context. Read it on its own and you will not even realize that it's talking about hints.

>Any GLSL type except for void can be a uniform. Additionally, Godot provides optional shader hints to make the compiler understand for what the uniform is used.

I take issue with this. "Optional" shader hints. They're optional so I can skip over these docs, right? And yet they are also _required_ at the same time. Therefore I would move the part about the "sRGB->linear conversion" to the part where it talks about it being optional, in order to clarify to the reader that they're not always optional.

### DIFF
--- a/tutorials/shading/shading_reference/shading_language.rst
+++ b/tutorials/shading/shading_reference/shading_language.rst
@@ -477,7 +477,7 @@ You can set uniforms in the editor in the material. Or you can set them through 
           must match *exactly* to the name of the uniform in the shader or else it will not be recognized.
 
 Any GLSL type except for *void* can be a uniform. Additionally, Godot provides optional shader hints
-to make the compiler understand for what the uniform is used. ``hint_albedo`` is required for proper sRGB->linear conversion, as Godot’s 3D engine renders in linear color space.
+to make the compiler understand for what the uniform is used.
 
 .. code-block:: glsl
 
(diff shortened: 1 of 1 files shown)


tutorials/shading/shading_reference/shading_language.rst
```diff
@@ -477,7 +477,7 @@ You can set uniforms in the editor in the material. Or you can set them through
           must match *exactly* to the name of the uniform in the shader or else it will not be recognized.
 
 Any GLSL type except for *void* can be a uniform. Additionally, Godot provides optional shader hints
-to make the compiler understand for what the uniform is used.
+to make the compiler understand for what the uniform is used. ``hint_albedo`` is required for proper sRGB->linear conversion, as Godot’s 3D engine renders in linear color space.
```
clayjohn: Maybe just add "For example..." Otherwise this sentence feels really out of context.

rainlizard:
You're right starting a sentence with ``hint_albedo`` is a bit weird and abrupt. But does "For example" fit well? Are we providing an "example" of these "optional" shader hints? This isn't really an example, it's an exception. The point is to clarify to the reader that there is an exception to "optional".

However I also just realized that I think putting this sentence here is a bad idea because right below it is an example that requires the above context. This should probably remain as one uninterrupted unit:
Untitled

So I put it below. After some fumbling about I've made the changes, tell me what you think.

akien-mga:
Thanks!

PR #2926, opened 2019-11-13
Add note about shader function performance in GLES2

skyace65:
Adds a note in the GLES2 vs GLES3 page that some shader functions were added to GLES2 with Godot's own shader library. And they may perform worse in GLES2 compared to GLES3.

### DIFF
--- a/tutorials/misc/gles2_gles3_differences.rst
+++ b/tutorials/misc/gles2_gles3_differences.rst
@@ -158,6 +158,8 @@ For a complete list of built-in GLSL functions see the :ref:`Shading Language do
 | vec_type **fwidth** ( vec_type p )                                                          |                                                  |
 +---------------------------------------------------------------------------------------------+--------------------------------------------------+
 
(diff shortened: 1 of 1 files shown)


tutorials/misc/gles2_gles3_differences.rst
```diff
@@ -158,6 +158,9 @@ For a complete list of built-in GLSL functions see the :ref:`Shading Language do
 | vec_type **fwidth** ( vec_type p )                                                          |                                                  |
 +---------------------------------------------------------------------------------------------+--------------------------------------------------+
 
+.. note:: Many functions not in GLES2's GLSL were added with Godots own shader standard library.
+          These functions may performe worse in GLES2 compared to GLES3.
```
clayjohn: I think all the text should be within the note.

skyace65:
Fixed.

clayjohn:
Looks great! Thank you. :)

PR #2928, opened 2019-11-13
Update shader materials page for Godot 3.2

skyace65:
Updates information since in 3.2 you can now convert a visual shader to a text shader. And added an explenation for previewing shader code in the visual shader editor. Updated one of the images for 3.2 and added an image to show the shader code preview. Both images have gone through png compression.


akien-mga:
Thanks!

PR #2929, opened 2019-11-14
Update importing scenes for Godot 3.2 and move some stuff around

skyace65:
-Add FBX to the list of supported formats.
-Move Collada bellow gLTF since with newer versions of Blender people should use gLTF.
-Shorten gLTF description.
-Reword Collada description.
-Remove the section on FBX not being supported since support for FBX was added with the Open Asset Import Library.
-Move Collada Blender export instructions below gLTF since people should be using gLTF in newer versions of Blender.
-Explain that the better Collada export plugin doesn't work for Blender 2.8 or newer.

### DIFF
--- a/getting_started/workflow/assets/importing_scenes.rst
+++ b/getting_started/workflow/assets/importing_scenes.rst
@@ -13,25 +13,16 @@ transferred as close as possible.
 
 Godot supports the following 3D *scene file formats*:
 
-* DAE (COLLADA), which is currently the most mature workflow.
-* glTF 2.0. Both text and binary formats are supported. Godot has full support for it, but the format is new and gaining traction.
+* glTF 2.0. Godot has full support for text and binary formats.
+* DAE (COLLADA), an older format that is fully supported.
 * OBJ (Wavefront) formats. It is also fully supported, but pretty limited (no support for pivots, skeletons, etc).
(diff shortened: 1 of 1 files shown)


getting_started/workflow/assets/importing_scenes.rst
```diff
@@ -13,26 +13,17 @@ transferred as close as possible.
 
 Godot supports the following 3D *scene file formats*:
 
-* DAE (COLLADA), which is currently the most mature workflow.
-* glTF 2.0. Both text and binary formats are supported. Godot has full support for it, but the format is new and gaining traction.
+* glTF 2.0. Godot has full support for text and binary formats.
+* DAE (COLLADA), an older format that is fully supported.
 * OBJ (Wavefront) formats. It is also fully supported, but pretty limited (no support for pivots, skeletons, etc).
 * ESCN, a Godot specific format that Blender can export with a plugin.
+* FBX, supported with the open asset import library. however it should only be used if you can't use another formet.
```
nezvers: Typo - ... if you can't use another formAt

skyace65:
Fixed

akien-mga:
Did some fixes (typos, Collada exporter will be ported to 2.80+ eventually so no need to document that).

akien-mga:
Thanks!

PR #2931, opened 2019-11-14
Add quote style recommendation to the GDScript style guide

### LINKED ISSUE #2869: Add a quote style guideline to the GDScript style guide
The [GDScript style guide](http://docs.godotengine.org/en/latest/getting_started/scripting/gdscript/gdscript_styleguide.html) currently doesn't make any recommendations as for quotes. While the Godot script editor defaults to double quotes, I've seen many people use single quotes in their projects (even though it's still a minority in my experience).

I personally use the following style in my projects, inspired by the Python [black](https://github.com/ambv/black) formatter:

```gdscript
# Normal string
"hello world"
# Use double quotes as usual to avoid escapes
"hello 'world'"
# Use single quotes as an exception to avoid escapes
'hello "world"'
# Both quote styles would require 2 escapes, prefer double quotes if it's a tie
"'hello' \"world\""
```

In this style, double quotes are the "default" way to quote strings, *unless* a single-quoted string makes it possible to escape less characters compared to a double-quoted string. In this case (and only in this case), the string should use single quotes.

What do you think about adopting this style?

Calinou:
This closes #2869.

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_styleguide.rst
+++ b/getting_started/scripting/gdscript/gdscript_styleguide.rst
@@ -185,6 +185,26 @@ spaces in dictionary references and function calls, or to create "columns."
     y        = 100
     velocity = 500
 
+Quotes
+~~~~~~
+
+Use double quotes unless single quotes make it possible to escape less
+characters in a given string. See the examples below:
+
+::
+
+    # Normal string.
+    print("hello world")
+
+    # Use double quotes as usual to avoid escapes.
+    print("hello 'world'")
+
+    # Use single quotes as an exception to the rule to avoid escapes.
+    print('hello "world"')
+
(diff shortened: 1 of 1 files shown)


getting_started/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -185,6 +185,26 @@ spaces in dictionary references and function calls, or to create "columns."
     y        = 100
     velocity = 500
 
+Quotes
+~~~~~~
+
+Use double quotes unless single quotes make it possible to escape less
```
cbscribe: Characters are countable, so it should be "fewer" rather than "less"

NathanLovato:
Thanks, I'm merging and making the change requested by Chris now

PR #2935, opened 2019-11-17
Tweak the theme to better match Godot's visual identity

Calinou:
This also removes the need to load Web fonts, leading to a smoother page loading experience. With this pull request, only 256 KB of data needs to be transferred to load the main page of the `latest` branch on Read the Docs (instead of 940 KB).

These tweaks also improve the visual feedback when hovering or clicking links and navigation items. Font sizes and weights have been tweaked for better consistency.

You can preview this theme by loading [this stylesheet](https://github.com/Calinou/godot-docs/blob/improve-theme/_static/css/custom.css) as an userstyle for docs.godotengine.org. This can be done using the Stylus browser add-on.

If this design is good to go, I can look at implementing a dark theme variant using [`@media (prefers-color-scheme: dark)`](https://css-tricks.com/dark-modes-with-css/) :slightly_smiling_face: 

### Before

godot-docs_theme_old

### After

godot-docs_theme_new_2

<details>
<summary>Old variant</summary>

godot-docs_theme_new
</details>

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -1,0 +1,273 @@
+/**
+ * Various tweaks to the Read the Docs theme to better conform with Godot's visual identity.
+ */
+
+ :root {
+    --navbar-background-color: #333f67;
+    --navbar-background-color-hover: #29355c;
+    --navbar-background-color-active: #212d51;
+    --navbar-current-background-color: #212d51;
+    --navbar-current-background-color-hover: #182343;
+    --navbar-current-background-color-active: #131e3b;
+    --navbar-level-1-color: #c3e3ff;
+    --navbar-level-2-color: #b8d6f0;
+    --navbar-level-3-color: #a3c4e1;
+    --navbar-heading-color: #ff7381;
+    --link-color-active: #105078;
+    --code-literal-color: #d04c60;
+    --highlight-background-color: #f5ffe1;
+    --input-background-color: #fcfcfc;
+    --input-focus-border-color: #5f8cff;
+}
+
+body,
+h1,
+h2,
+h3,
+h4,
+h5,
+h6,
+input[type="text"],
+input[type="button"],
+input[type="reset"],
+input[type="submit"],
+textarea,
+legend,
+.btn,
+.rst-content .toctree-wrapper p.caption,
+.rst-versions {
+    /* Use a system font stack for better performance (no Web fonts required) */
+    font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif;
+}
+
+h1,
+h2,
+h3,
+h4,
+h5,
+h6,
+legend,
+.rst-content .toctree-wrapper p.caption {
+    /* Use a lighter font for headers (Medium instead of Bold) */
+    font-weight: 500;
+}
+
+p,
+article ul,
+article ol,
+.wy-plain-list-disc,
+.wy-plain-list-decimal,
+.rst-content ol.arabic,
+.rst-content .section ul,
+.rst-content .toctree-wrapper ul,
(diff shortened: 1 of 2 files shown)


TheDuriel:
Left my feedback on discord. Suffice to say, I have some big problems with this and I think it definitely needs more eyes on it before a merge is considered.

golddotasksquestions:
Except for the [+] in the tree menu on the left, I find the "before" more clear tbh. (I think the [+] needs to stand out more, as right now you only see that there is a submenu when you hover over one of those items)
But the rest in the "after" seems more grey and lifeless. I'm also sad if the code area would loose the green color, because that's what I'm looking for the most usually and that what my eye is drawn to also immediately in the "before". 

The whole page in the "after" now seems a lot more cluttered, because everything seems to have the same importance. Before it was more structured into different areas. With those blue/darkgrey/white background, it's immediately clear to the eye where to find the searchbar, where the page content is and where the menu area is. In the after this has all the more-less same color coding, or rather no color coding at all, which makes it less clear imho.

Kemeros:
Personally i prefer dark mode in everything so making it lighter is not really what i'd go for. Maybe work on a dark mode theme instead? They are getting quite popular lately. The goal is for the doc to be easy to read for a long period of time, not just look uniform or pretty.

nhydock:
I don't know if I prefer the look of a completely light theme.  Keeping the original readthedocs look but switching it to follow all of the existing Godot colors could give it a nice branded feel while maintaining readability and familiarity.  I do like the changes in fonts and central documentation.  I just know people are sensitive to color intensity of sticky content like navbars so such a big change will always be debated.

image

Calinou:
@Kemeros As stated in OP, I have plans to implement a dark theme if this design is approved. Still, I don't think it should be the default (unless your browser/OS is set to use a dark theme), as dark themes are *less* readable in light environments such as a room with direct sunlight.

Kemeros:
@Calinou I may need glasses. Apologies.

While were talking about the doc page. I have two suggestions for enhancements:

1- Is there a way to smooth scroll to adjust the page when selecting a category on the left? The sudden page jump is a bit jarring.

2- Would it be possible to open the pluses by clicking on the text as well?

Calinou:
> 1- Is there a way to smooth scroll to adjust the page when selecting a category on the left? The sudden page jump is a bit jarring.

This can likely be done using custom JavaScript (in a future PR).

> 2- Would it be possible to open the pluses by clicking on the text as well?

That could be done for first-level items, as clicking on them will only display a table of contents for most of them. There are some exceptions to this (like [Tutorials and resources](http://docs.godotengine.org/en/latest/community/tutorials.html)), so the content may have to be modified slightly to account for this change.

starry-abyss:
Godot's identity is blue and dark gray (before picture), not light gray (after picture). Probably you confused engines :P

Calinou:
I amended this pull request with changes based on @nhydock's mockup (who I added as a co-author). Thanks for your feedback :slightly_smiling_face:

groud:
I like the new theme a lot ! :)

golddotasksquestions:
This looks a lot better now, thanks!

akien-mga:
Let's give this a spin on the `latest` docs and see what users think of it.

akien-mga:
Thanks!

aaronfranke:
My immediate thought is that the blue is too bright. I took a screenshot next to the Godot editor:

Screenshot from 2019-11-29 16-01-42

How does this "better match Godot's visual identity"? The editor is far darker.

nhydock:
it matches the godot website, not the godot editor

Calinou:
> How does this "better match Godot's visual identity"? The editor is far darker.

Here's how it looks with the lightest color in the editor dark theme:

image

I'm not entirely convinced by it, since it clashes with the logo's more saturated blue outlines.

PR #2937, opened 2019-11-18
change debug to release_debug

zaksnet:
Mono build: The docs were advising to use `target=debug` instead of `target=release_debug` for debug templates. This PR is addressing that.

### DIFF
--- a/development/compiling/compiling_with_mono.rst
+++ b/development/compiling/compiling_with_mono.rst
@@ -153,7 +153,7 @@ Export templates
 
 The name of the data directory for an export template differs based on the
 configuration it was built with. The format is
-``data.mono.<platform>.<bits>.<target>``, e.g. ``data.mono.x11.32.debug`` or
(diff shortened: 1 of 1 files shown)


akien-mga:
There's another occurrence: `data.mono.x11.32.debug`

For future commits, please try to be more explicit in the log. Describing what the diff shows is not useful, but saying something like:
```
Mono build: use release_debug target for debug templates

See godotengine/godot#33699.
```

zaksnet:
So

> There's another occurrence: `data.mono.x11.32.debug`
> 
> For future commits, please try to be more explicit in the log. Describing what the diff shows is not useful, but saying something like:
> 
> ```
> Mono build: use release_debug target for debug templates
> 
> See godotengine/godot#33699.
> ```

Sorry, updated. Also, i used the browser to make the changes and i ended up with 2 commits :flushed:. If you can squash them, please.

akien-mga:
Ah, didn't see that it was `3.1`... PRs should be made against the `master` branch...

akien-mga:
Cherry-picked in `master` as b842f77a.

PR #2940, opened 2019-11-18
Improved rotational movement in 2d movement tutorial

rhode2nh:
The current movement system moves the player relatively to the right when press the "up" key. This is not intuitive. The code was changed to move the player relatively up when pressing the "up" key.

### DIFF
--- a/tutorials/2d/2d_movement.rst
+++ b/tutorials/2d/2d_movement.rst
@@ -137,9 +137,9 @@ while up/down moves it forward or backward in whatever direction it's facing.
         if Input.is_action_pressed('left'):
             rotation_dir -= 1
         if Input.is_action_pressed('down'):
-            velocity = Vector2(-speed, 0).rotated(rotation)
+            velocity = Vector2(0, speed).rotated(rotation)
         if Input.is_action_pressed('up'):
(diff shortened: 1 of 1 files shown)


cbscribe:
Up moves forward. In Godot, an angle of zero degrees is along the positive x-axis, ie to the right. When the body has 0 rotation, it's pointing to the right.

You can change the initial orientation of the node so that it starts out rotated 90 degrees to point "up", but the code should not be changed.

rhode2nh:
Thanks for the clarification. It makes more sense why it's like that now. In regards to your edit, do you believe it would be more viable to create a separate page discussing movement/rotations with transform manipulation in the 2d movement document? The current pages are pretty gentle compared to matrix math.

cbscribe:
You wouldn't really have to dig into matrix math. You just highlight that forward movement is

    velocity = transform.x * speed

It's just a thought. It doesn't really make that much difference in 2D - a slight simplification with not having to deal with angles, but an additional thing to explain. However, it does make the transition to 3D smoother, where you're going to need to use the body's transform to move it.

Whether it's here or in a separate doc, it would be a good thing to have.

PR #2946, opened 2019-11-25
Mention that ETC texture compression doesn't support transparency

### LINKED ISSUE #2899: GLES2 vs GLES3 differences: mention that ETC doesn't support transparency
See godotengine/godot#33198.

We have a section about texture compression here: https://docs.godotengine.org/en/latest/tutorials/misc/gles2_gles3_differences.html#texture-compression
It would be worth mentioning notable differences between the texture compression formats available on GLES2 and GLES3, and especially the fact that ETC doesn't support transparency.

CC @clayjohn

Calinou:
I made it appear as a warning so it's relatively prominent.

This closes #2899.


NathanLovato:
Thanks!

PR #2956, opened 2019-11-30
Add new tutorials to doc changelog

merumelu:
Add visual shaders and microphone tutorials and update the version numbers in the first paragraph.

### DIFF
--- a/about/docs_changelog.rst
+++ b/about/docs_changelog.rst
@@ -3,9 +3,10 @@
 Documentation changelog
 =======================
 
-The documentation is continually being improved. The release of version 3.1
+The documentation is continually being improved. The release of version 3.2
 includes many new tutorials, many fixes and updates for old tutorials, and many updates
-to the class reference. Below is a list of new tutorials added since version 3.0.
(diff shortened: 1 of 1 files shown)


clayjohn:
If we are saying that this lists the new tutorials since 3.1 we need to remove all the tutorials that were added before 3.1 was released.

merumelu:
Should we reword it to be version independet instead? Something like this.

> Godot's documentation is continually being improved. With each engine release
> come many new tutorials, fixes and updates for old tutorials, and many updates
> to the :ref:`class reference <toc-class-ref>`. Below is an overview of new tutorials
> and the version they were added.

akien-mga:
It's already version-dependent, the intro paragraph was just wrong. If you scroll down there's the changelog for the 3.1 updates below the latest updates.

mhilbrunner:
Wow, this has been sitting here for a while. :) Sorry!

I'll merge this now, and we should cherrypick this to the 3.2 branch - I think making the initial paragraph more version independent is still a good idea, even if we need to keep the other parts of the document updated manually.

mhilbrunner:
Thank you!

PR #2963, opened 2019-12-03
Add clarification on how to connect C# signals in the editor

theowiik:
The doc does not mention how to connect a C# signal in the editor. Had a lot of trouble figuring out why I could not see my newly added signal.

Copied most of the explanation from the [Your first game](https://docs.godotengine.org/en/3.1/getting_started/step_by_step/your_first_game.html) doc which mentions this problem:

> If you’re using C#, you need to (re)build the project assemblies whenever you want to see new export variables or signals. This build can be manually triggered by clicking the word “Mono” at the bottom of the editor window to reveal the Mono Panel, then clicking the “Build Project” button.

### DIFF
--- a/getting_started/scripting/c_sharp/c_sharp_features.rst
+++ b/getting_started/scripting/c_sharp/c_sharp_features.rst
@@ -94,7 +94,7 @@ Declaring a signal in C# is done with the ``[Signal]`` attribute on a delegate.
     delegate void MySignalWithArguments(string foo, int bar);
 
 These signals can then be connected either in the editor or from code with ``Connect``.
-If you want to connect a signal in the editor, you need to (re)build the project assemblies to see the new signal. This build can be manually triggered by clicking the word “Mono” at the bottom of the editor window to reveal the Mono Panel, then clicking the “Build Project” button. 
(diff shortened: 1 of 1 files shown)


getting_started/scripting/c_sharp/c_sharp_features.rst
```diff
@@ -94,6 +94,7 @@ Declaring a signal in C# is done with the ``[Signal]`` attribute on a delegate.
     delegate void MySignalWithArguments(string foo, int bar);
 
 These signals can then be connected either in the editor or from code with ``Connect``.
+If you want to connect a signal in the editor, you need to (re)build the project assemblies to see the new signal. This build can be manually triggered by clicking the word “Mono” at the bottom of the editor window to reveal the Mono Panel, then clicking the “Build Project” button.
```
neikeq: It's better to recommend the "Build" button at the top right corner. Much easier to access and to explain as well.

theowiik: Changed so that it references the top right build button instead of the mono tab

akien-mga:
Thanks!

PR #2964, opened 2019-12-03
Small consistency fixes for the VR tutorial pages

YeldhamDev:
- Fixed typo: "Starer" -> "Starter"
- Made the titles only capitalized at the beginning, like everywhere else.
- Made the numbers in the titles be actual digits, like in the FPS tutorial.
- Fixed error in a reference due to the use of a plural.


akien-mga:
Thanks!

PR #2965, opened 2019-12-04
Clarify how the method float (String from ) works

Dancovich:
Update the description of this method to be in line with `int int(String from)` in regards to how it deals with a completely invalid parameter and a partially invalid parameter.

The documentation for `float float(String from)` doesn't make it clear what happens if you pass a completely invalid String or a partially invalid String. After some testing it was possible to confirm it works like `int int(String from)` in that regard so this pull request proposes updating the documentation to match.


Calinou:
Please submit this pull request on the [main Godot repository](https://github.com/godotengine/godot) (edit `doc/classes/float.xml`), not here. This is because the class reference in this repository is automatically generated from those XML files.

PR #2966, opened 2019-12-04
[iOS] Add commands for bundling camera and ARKit universal libs.

bruvzg:
Adds commands to accommodate changes in the iOS export made in https://github.com/godotengine/godot/pull/33992, adds missing `-create` flag and changes output file names to match names used in export template.

*Update:* Same commands for cross build added.


akien-mga:
Thanks!

PR #2968, opened 2019-12-05
Add information on profiler and misc debugging tools

skyace65:
Adds information on how to use the profiler and misc debugging tools. I wanted to add more information on the profiler but I couldn't find much explaining it and what I did find didn't explain everything and there's still stuff I don't understand. Regardless this is better than what it was.


akien-mga:
Thanks!

PR #2972, opened 2019-12-10
Replace images and move text for visual shader page

skyace65:
Replaces two images for the visual shader page. Previously it was using images from the shader material page that could be unclear. These two are specifically for this page. Both images have gone through png compression. I also moved one sentence so it's after one of the images, it makes more sense that way.


akien-mga:
Thanks!

PR #2976, opened 2019-12-10
Explain how to add a SpatialMaterial to a mesh

skyace65:
Explains how to add a SpatialMaterial to a mesh and what makes the three material slots different. The image added for this has gone through png compression.

Someone was asking about this on Godot's Facebook group and I realized the documentation doesn't go over this, and probably should.


akien-mga:
Thanks!

PR #2981, opened 2019-12-13
set_script() detailing what happens to variables

rainlizard:
Makes it clear to the reader that this is a clean replacement; no old variables will remain. I didn't know what would happen until I tested it.

It also seems somewhat significant to write this down because I think this is the only other way (other than writing ``var``) to initialize new variables on an existing node. Please correct me if I'm wrong, I've been looking for a way to initialize a new variable.

### DIFF
--- a/classes/class_object.rst
+++ b/classes/class_object.rst
@@ -587,6 +587,8 @@ Adds or changes a given entry in the object's metadata. Metadata are serialized,
 
 Assigns a script to the object. Each object can have a single script assigned to it, which are used to extend its functionality.
 
+The object's old variables will be lost and new variables will be initialized from the new script.
+
 ----
 
 .. _class_Object_method_to_string:


akien-mga:
> Please correct me if I'm wrong, I've been looking for a way to initialize a new variable.

Assuming you mean at runtime (otherwise you can just use `var` in the script), you should be able to do it with `set("name", value)`

Note that this file is generated from the engine repo as mentioned in its header and the README, so it shouldn't be modified here as changes will be lost.

rainlizard:
ah I see the header now, thanks. I'm new to github and didn't even know that text could be hidden like that, I instantly hit ctrl+f to find rather than read. I'll submit pull request under main repo for Object.xml

>void set(property: String, value: Variant)
>
>Assigns a new value to the given property. If the property does not exist, nothing will happen.

nope, "nothing will happen". tested as well.

PR #2991, opened 2019-12-19
Add a list of supported external editors and update C# version info

skyace65:
Added a list of the external editors Godot supports, and updated the C# and mono version info for Godot 3.2.


akien-mga:
Thanks!

PR #2994, opened 2019-12-20
Add a line length guideline to the GDScript style guide

### LINKED ISSUE #2567: Add a line length guideline to the GDScript style guide
Unlike many other languages, the [GDScript style guide](http://docs.godotengine.org/en/latest/getting_started/scripting/gdscript/gdscript_styleguide.html) currently doesn't recommend a maximum line length. We should probably adopt a guideline for this, so that people can write code more consistently across projects.

Values around 80 seem to be quite common in most high-level languages, but in my experience, it tends to feel a bit short when indenting code with tabs (or 4 spaces). This is why some Python formatters such as [Black](https://github.com/ambv/black) default to 88 characters:

> You probably noticed the peculiar default line length.  *Black* defaults to 88 characters per line, which happens to be 10% over 80.  This number was found to produce significantly shorter files than sticking with 80 (the most popular), or even 79 (used by the standard library). In general, [90-ish seems like the wise choice](https://youtu.be/wf-BqAjZb8M?t=260).

If we recommend a line length in the style guide, we can then enable the line length guideline in the script editor by default :slightly_smiling_face:

Calinou:
From consensus in the issue and Discord, a soft limit of 80 characters per line seems to be the way to go.

In the future, we should also document multiline wrapping using parentheses and backslashes and make a recommendation. My vote goes for recommending parentheses whenever possible, as they're less error-prone compared to backslashes :slightly_smiling_face:

This closes #2567.

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_styleguide.rst
+++ b/getting_started/scripting/gdscript/gdscript_styleguide.rst
@@ -86,6 +86,13 @@ Surround functions and class definitions with two blank lines:
 
 Use one blank line inside functions to separate logical sections.
 
+Line length
+~~~~~~~~~~~
+
+Try to keep lines under 80 characters. This ensures greater readability on small
+displays and splitted editors (such as side-by-side diffs). It's OK to go over
+by a few characters, but a line should never exceed 100 characters.
+
 One statement per line
 ~~~~~~~~~~~~~~~~~~~~~~
 


getting_started/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -86,6 +86,13 @@ Surround functions and class definitions with two blank lines:
 
 Use one blank line inside functions to separate logical sections.
 
+Line length
+~~~~~~~~~~~
+
+Try to keep lines under 80 characters. This ensures greater readability on small
```
Scony: how about `at most 79 characters`? Meaning is the same, but the threshold (79) is explicit  - one does not have to think about it.

Calinou: This may sound weird, but I've found 80 characters to be significantly less limiting compared to 79 characters…

I can change the wording to make it more explicit though.

Scony: okay then :)

akien-mga:
I tend to find line length limitation in Python quite limiting and forcing to do a lot of wrapping all the time, which IMO doesn't improve readability much over simply allowing lines of 100-120 chars.

But if the majority wants a hard limit, I don't mind. But then the page should also explain how to wrap lines which would otherwise exceed the line length limitation. Should things be aligned with previous lines, if so following what logic? Or should there be an extra level of indent, or two? How to break lines is also not always straightforward in GDScript (when can you do it simply and when do you need `\`?), so this needs to be covered by the style guide.

NathanLovato:
Same as Akien here, 80 characters is too short. I haven't seen too many python projects actually follow this guideline.

80 characters prevent you from having even just 2 conditions on a given line in input functions or just doesn't work with some statements with the longer parameters/class names and type hints in Godot. 

Here are three examples from our current project:

```
onready var _playback: AnimationNodeStateMachinePlayback = animation_tree["parameters/playback"] # 97 characters
	animation_tree["parameters/move_ground/blend_position"] = direction.length() # 81 characters
	var forwards: Vector3 = player.camera.global_transform.basis.z * input_direction.z # 83 characters
```

I have plenty of them. Type hints make for long lines.

Another thing is that right now we'd have to enforce and review it manually, which is a time sink. The gdscript formatter wasn't ready for that last time I tried it. And yeah we'd need more guidelines for line-wrapping. There, I find black, the Python Software Foundation's formatter, great: [how black wraps lines](https://github.com/psf/black#how-black-wraps-lines).

It uses simple and consistent rules, like it wraps parentheses or brackets the same way, making it easy to read the code.

Scony:
@NathanLovato I'm working on it: https://github.com/Scony/godot-gdscript-toolkit ATM I do linter, but soon I'll start with formatter (similar to black and clang-format)

NathanLovato:
@Scony Nice! Did you see there's a PR open for a formatter in the engine already? It's exposed to the command-line and is meant to follow the gdscript guidelines. How about joining forces there? https://github.com/godotengine/godot/pull/27382

Scony:
@NathanLovato I've got some strong reasons to not implement it in the engine, but haven't seen this PR anyway. I'll try to keep feature-parity with the one from PR ;-) And ofc. participate in discussions. Thanks.

NathanLovato:
@Scony is there something you'd take contributions for? We use python with the team and although I don't have experience with parsers I'd gladly help.

Scony:
@NathanLovato sure! there are plenty of things to do like linter checks (ones commented out in https://github.com/Scony/godot-gdscript-toolkit/blob/master/gdtoolkit/linter/__init__.py), formatter, metrics calculator (like python's radon - https://radon.readthedocs.io/en/latest/) + more linter checks that need to be invented. Whatever you want to do feel free to do - just add an issue and mention you're working on it :)

Calinou:
Based on your feedback, should I move the hard limit to 100 characters and remove the soft limit?

NathanLovato:
In another discussion, people rightly mentioned that if we have guidelines for optional typing, they should stay separate. In the case of line length, you need to increase the limit with the extra type hints. So no need to take that into account.

I think it's good, I'll merge and tweak the phrasing a tiny bit. Thanks!

NathanLovato:
The rationale being that we can change that bit anytime, people can open a discussion or a PR if they think it's unreasonable.

PR #2995, opened 2019-12-20
Add a boolean operators guideline to the GDScript style guide

Calinou:
Follow-up to #2994.

These guidelines are still up to discussion, but I figured those are logical choices in my experience (no pun intended).


NathanLovato:
Thanks. About the writing style: avoid parentheses (as they create sub-sentences and extra, unnecessary propositions), and maybe the informal and/or notation. For instance:

```
Prefer the plain English versions of boolean operators, as they are more accessible:

- Use ``and`` instead of ``&&``.
- Use ``or`` instead of ``||``.
```

PR #2999, opened 2019-12-22
Edit overview_of_debugging_tools

NathanLovato:
Rewrite the intro to explain what the reader will find in this guide and list
debugging tools right away.

Review grammar and improve the text's clarity.

### DIFF
--- a/tutorials/debug/overview_of_debugging_tools.rst
+++ b/tutorials/debug/overview_of_debugging_tools.rst
@@ -3,18 +3,24 @@
 Overview of debugging tools
 ===========================
 
-Introduction
-------------
+This guide will give you an overview of the available debugging tools in the
+engine.
 
-When developing your game, you want to test your game and debug when problems occur. Godot provides several debugging options and
-tools which aid your debugging process. There are the debug dropdown options, Script editor debug options, debug project settings,
-and the debugger.
+Godot comes with a powerful debugger and profilers to track down bugs, inspect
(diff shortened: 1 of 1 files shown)


NathanLovato:
I didn't rewrite the entire page in one go, but hopefully improved quite a bit. I think it does more than giving an overview and intro of the available debugging tools. It goes over most options, but I'd rather split them into different pages after this PR is merged.

skyace65:
After this is merged I'd like to add documentation on the Game Camera Override and I can split it then. How should the pages be split? I was thinking something like:
- Debugger panel
- Other debugging tools
- debugging options

NathanLovato:
I would recommend page titles that are as descriptive as possible. It's okay if a page is a little short. The reason for this is the way people look for information: we scan titles briefly, search for keywords related to our problem. So in the camera's case, I'd call the page like "Overriding the game camera" or "Camera override in game" or "Camera override when testing". Ideally something that puts camera first.

Regarding the split there's some rewriting involved. I'd at least have pages for:

Overview of the debug options <- making it only an overview
Inspecting the game with breakpoints <- you can do a lot with that, like opening nodes at runtime in the inspector
Profiler <- maybe a guide on improving performances
Hot reload <- and options in the debug menu, detailing their use some more, as it's hard to realize their power without more context. Also there are things that won't work, like hot reloading code doesn't reinitialize nodes
Other debugging options

Calinou:
Thanks! (I amended some syntactic changes to the commit before merging it.)

PR #3001, opened 2019-12-22
Explain why we store the return value from move_and_slide

### LINKED ISSUE #2934: Add explanation for 'velocity = move_and_slide(velocity)'
I never understood why move and slide has to be assigned to the variable velocity to make it work even though I went over the [KinematicBody2D tutorial page](https://docs.godotengine.org/en/3.1/tutorials/physics/using_kinematic_body_2d.html) at least 20 to 30 times over the months.
```
var velocity = Vector2(1, 0)
velocity = move_and_slice(velocity)
```

Today user Grimdakka on Reddit made [this comment](https://www.reddit.com/r/godot/comments/dxovz1/may_be_an_stupid_question_2d_movement/f7u8rmk?utm_source=share&utm_medium=web2x) and now it all makes sense. I wish we would have that explanation in the docs as well:

> You need to save the new velocity after move_and_slide does its magic because you're going to pass it to move_and_slide again next frame. Otherwise the velocity would never change between frames except via input.

NathanLovato:
Close #2934

### DIFF
--- a/tutorials/physics/using_kinematic_body_2d.rst
+++ b/tutorials/physics/using_kinematic_body_2d.rst
@@ -138,6 +138,21 @@ Anything you do with ``move_and_slide()`` can also be done with ``move_and_colli
 but it might take a little more code. However, as we'll see in the examples below,
 there are cases where ``move_and_slide()`` doesn't provide the response you want.
 
+In the example above, we assign the velocity that ``move_and_slide()`` returns
+back into the ``velocity`` variable. This is because when the character collides
+with the environment, the function recalculates the speed internally to reflect
+the slowdown.
+
+For example, if your character fell on the floor, you don't want its downward
+speed to keep increasing. Instead, you want its vertical speed to be reset to
+zero.
+
+``move_and_slide()`` may also recalculate the kinematic body's velocity several
+times in a loop as, to produce a smooth motion, it moves the character and
+collides up to 5 times by default. At the end of the process, the function
+returns the character's new velocity that we can store in our ``velocity``
+variable, and use on the next frame.
+
 Examples
 --------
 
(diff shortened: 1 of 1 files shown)


tutorials/physics/using_kinematic_body_2d.rst
```diff
@@ -138,6 +138,21 @@ Anything you do with ``move_and_slide()`` can also be done with ``move_and_colli
...
 there are cases where ``move_and_slide()`` doesn't provide the response you want.
 
+In the example above, we assign the velocity that ``move_and_slide()`` returns
+back into the ``velocity`` variable. This is because when the character collides
+with the environment, the function recalculates the speed internally to reflect
+the slowdown.
+
+For example, if your character fell on the floor, you don't want its downward
+speed to keep increasing. Instead, you want its vertical speed to be reset to
+zero.
```
Feniks-Gaming: using downwards speed and vertical speed for the same thing in one sentence is confusing. I would stick with Vertical for both.

tutorials/physics/using_kinematic_body_2d.rst
```diff
@@ -138,6 +138,21 @@ Anything you do with ``move_and_slide()`` can also be done with ``move_and_colli
...
+
+``move_and_slide()`` may also recalculate the kinematic body's velocity several
+times in a loop as, to produce a smooth motion, it moves the character and
+collides up to 5 times by default. At the end of the process, the function
```
Feniks-Gaming: collides,

NathanLovato: There's no need for a comma here, it's a self-contained proposition with a subject, verb, and complement. "it moves the character and collides up to 5 times by default." 

However, I should use plain English instead of a numeral.

NathanLovato:
Updated the text

Calinou:
Thanks!

PR #3002, opened 2019-12-22
Add robots.txt file to prevent indexing outdated docs

### LINKED ISSUE #2912: Show latest Godot docs in search
I do not know if this is possible or not, but is there any way that the Google search results always yield the latest Godot docs.
3/4 times I end up on a 3.0 docs page and I have to click to go to the latest version of the docs page.

NathanLovato:
Closes #2912

That should disable indexing for version 3.1, 3.0, and 2.0 on the docs, allowing search engines to only index stable and latest for all languages.

I did a test build, the file gets copied over to the build folder.

Once the change is live, you can use the webmaster console to see if it's working: https://support.google.com/webmasters/answer/6062598?hl=en&ref_topic=6061961

### DIFF
--- a/robots.txt
+++ b/robots.txt
@@ -1,0 +1,6 @@
+user-agent: *
+disallow: /*/3.1
+disallow: /*/3.0
+disallow: /*/2.0
+
+sitemap: https://docs.godotengine.org/sitemap.xml
--- a/conf.py
+++ b/conf.py
@@ -91,6 +91,8 @@ html_logo = 'img/docs_logo.png'
 # These folders are copied to the documentation's HTML output
 html_static_path = ['_static']
 
+html_extra_path = ['robots.txt']
+
 # These paths are either relative to html_static_path
 # or fully qualified paths (eg. https://...)
 html_css_files = [


robots.txt
```diff
@@ -0,0 +1,4 @@
+user-agent: *
+disallow: /*/3.1
+disallow: /*/3.0
+disallow: /*/2.0
```
LikeLakers2: The current [robots.txt](http://docs.godotengine.org/robots.txt) includes the following line for the Sitemap:

```
Sitemap: https://docs.godotengine.org/sitemap.xml
```

Do we need to include that as well? The sitemap seems to reference the various translated versions of the site, and I'm not sure whether including that is needed to direct users to the docs of their particular language.

NathanLovato: It's to help the robots find the sitemap, just in case they wouldn't find it. Some search engines support that, others don't. But I'll add it, it doesn't hurt. Thanks!

LikeLakers2:
Hey, just curious if you meant to add the above commit (https://github.com/godotengine/godot-docs/pull/3002/commits/9f7e24de45d8e5f6958c85c14221a4b783a0b8dc) to this PR, or if perhaps that was an accident. It doesn't seem related to the robots.txt stuff.

NathanLovato:
Just fixed it, thanks

Calinou:
Thanks!

PR #3006, opened 2019-12-26
Improve the 'Writing custom documentation' section

Relintai:
I improved the 'Writing custom documentation' section of the Custom modules in C++ page.

I made it a continuation of the tutorial from the previous sections. I also made sure to more explicitly explain, and show how doctool should be used. 

Also this corrects an error, because the current version tells you about `godot/doc/classes/`, but, on godot-master that path should be `godot/modules/summator/doc_classes/`.

### DIFF
--- a/development/cpp/custom_modules_in_cpp.rst
+++ b/development/cpp/custom_modules_in_cpp.rst
@@ -449,54 +449,87 @@ There are several steps in order to setup custom docs for the module:
 1. Make a new directory in the root of the module. The directory name can be
    anything, but we'll be using the ``doc_classes`` name throughout this section.
 
-2. Append the following code snippet to ``config.py``:
+2. Now, we need to edit ``config.py``, add the following snippet:
 
    .. code-block:: python
 
-       def get_doc_classes():
-           return [
-               "ClassName",
-           ]
+        def get_doc_path():
+            return "doc_classes"
 
-       def get_doc_path():
-           return "doc_classes"
+        def get_doc_classes():
+            return [
+                "Summator",
+            ]
 
-The ``get_doc_classes()`` method is necessary for the build system to
-know which documentation classes of the module must be merged, since the module
-may contain several classes. Replace ``ClassName`` with the name of the class
-you want to write documentation for. If you need docs for more than one class,
-append those as well.
-
-The ``get_doc_path()`` method is used by the build system to determine
-the location of the docs. In our case, they will be located in the ``doc_classes``
+The ``get_doc_path()`` function is used by the build system to determine
+the location of the docs. In this case, they will be located in the 
+``modules/summator/doc_classes`` directory. If you don't define this,
+the doc path for your module will fall back to the main ``doc/classes`` 
 directory.
 
-3. Run command:
+The ``get_doc_classes()`` method is necessary for the build system to
+know which registered classes belong to the module. You need to list all of your 
+classes here. The classes that you don't list will end up in the 
+main ``doc/classes`` directory.
 
-   .. code-block:: shell
+.. tip::
 
-      godot --doctool <path>
+    You can use git to check if you have missed some of your classes by checking the 
+    untracked files with ``git status``. For example::
 
-This will dump the engine API reference to the given ``<path>`` in XML format.
-Notice that you'll need to configure your ``PATH`` to locate Godot's executable,
-and make sure that you have write access rights. If not, you might encounter an
-error similar to the following:
+        user@host:~/godot$ git status
+
+    Example output::
+
+        Untracked files:
+            (use "git add <file>..." to include in what will be committed)
+            
+            doc/classes/MyClass2D.xml
+            doc/classes/MyClass4D.xml
+            doc/classes/MyClass5D.xml
+            doc/classes/MyClass6D.xml
+            ...
+    
+
+3. Now we can generate the documentation:
+
+We can do this via running Godot's doctool i.e. ``godot --doctool <path>``,
+which will dump the engine API reference to the given ``<path>`` in XML format.
+
+In our case we'll point it to the root of the cloned repository. You can point it
+to an another folder, and just copy over the files that you need.
+
+Run command:
+
+   ::
+
+      user@host:~/godot/bin$ ./bin/<godot_binary> --doctool .
+
+Now if you go to the ``godot/modules/summator/doc_classes`` folder, you will see 
+that it contains a ``Summator.xml`` file, or any other classes, that you referenced
+in your ``get_doc_classes`` function.
+
+Edit the file(s) following :ref:`doc_updating_the_class_reference` and recompile the engine.
+
+Once the compilation process is finished, the docs will become accessible within 
+the engine's built-in documentation system.
+
+In order to keep documentation up-to-date, all you'll have to do is simply modify
+one of the XML files and recompile the engine from now on.
+
+If you change your module's API, you can also re-extract the docs, they will contain
+the things that you previously added. Of course if you point it to your godot
+folder, make sure you don't lose work by extracting older docs from an older engine build 
+on top of the newer ones.
+
+Note that if you don't have write access rights to your supplied ``<path>``,
+you might encounter an error similar to the following:
 
 .. code-block:: console
 
     ERROR: Can't write doc file: docs/doc/classes/@GDScript.xml
        At: editor/doc/doc_data.cpp:956
 
-4. Get generated doc file from ``godot/doc/classes/ClassName.xml``
-
-5. Copy this file to ``doc_classes``, optionally edit it, then compile the engine.
-
-The build system will fetch the documentation files from the ``doc_classes`` directory
-and merge them with the base types. Once the compilation process is finished,
-the docs will become accessible within the engine's built-in documentation system.
-
-In order to keep documentation up-to-date, all you'll have to do is simply modify
-one of the ``ClassName.xml`` files and recompile the engine from now on.
 
 .. _doc_custom_module_icons:
 


Xrayez:
Knowing about `doc/classes` fallback mechanism, I also want to share a hack with you. 🙂 

If you have lots of classes for which you want to generate docs, having to list them all in `get_doc_classes()` manually can be a lot of work (you may also miss some during the process). I've even made attempts to solve this in godotengine/godot#33037 but then realized it won't work for when docs need to be generated initially.

Knowing this, you can do a hack to get a list of generated classes as reported by `git status`, namely `Untracked` files:

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)

        doc/classes/MyClass2D.xml
        doc/classes/MyClass4D.xml
        doc/classes/MyClass5D.xml
        doc/classes/MyClass6D.xml
 ...  
```

As formatted it is, you can copy the diff list to the `get_doc_classes()`, and then recompile the engine the second time.

Relintai:
Thanks, I'll add these to the pr, when I'll have some time!

Relintai:
I added everything to the pr. Sorry, it took a a lot more time to get to it than I originally expected.

I reworked section 2, because I felt that the fallback mechanic needs to be articulated (it would be too complex to explain as just a note), also I added the `--no-docbase` option to section 3 as a tip.

development/cpp/custom_modules_in_cpp.rst
```diff
@@ -449,54 +449,94 @@ There are several steps in order to setup custom docs for the module:
...
-           return "doc_classes"
+        def get_doc_classes():
+            return [
+                "Summator",
+            ]
+
+The ``get_doc_path()`` function is used by the build system to determine
+the location of the docs. In this case, they will be located in the 
+``modules/summator/doc_classes`` directory. If you don't define this,
+the doc path for you module will fall back to the main ``doc/classes``
```
akien-mga:
```suggestion
the doc path for your module will fall back to the main ``doc/classes`` 
```

development/cpp/custom_modules_in_cpp.rst
```diff
@@ -449,54 +449,94 @@ There are several steps in order to setup custom docs for the module:
...
+classes here. The classes that you don't list will end up in the 
+main ``doc/classes`` directory.
 
-The ``get_doc_path()`` method is used by the build system to determine
-the location of the docs. In our case, they will be located in the ``doc_classes``
-directory.
+.. tip::
+
+    You can use git to check if you have missed some of your classes by checking the 
+    untracked files with ``git status``. For exmample::
```
akien-mga:
```suggestion
    untracked files with ``git status``. For example::
```

development/cpp/custom_modules_in_cpp.rst
```diff
@@ -449,54 +449,94 @@ There are several steps in order to setup custom docs for the module:
...
+            doc/classes/MyClass2D.xml
+            doc/classes/MyClass4D.xml
+            doc/classes/MyClass5D.xml
+            doc/classes/MyClass6D.xml
+            ...
+    
+
+3. Now we can generate the documentation:
+
+We can do this via running godot's doctool e.g. ``godot --doctool <path>``,
```
akien-mga:
```suggestion
We can do this via running Godot's doctool, i.e. ``godot --doctool <path>``,
```

development/cpp/custom_modules_in_cpp.rst
```diff
@@ -449,54 +449,94 @@ There are several steps in order to setup custom docs for the module:
...
+
+In our case we'll point it to the root of the cloned repository. You can point it
+to an another folder, and just copy over the files that you need.
+
+Run command:
 
    ::
 
-      godot --doctool <path>
+      user@host:~/godot$ ./bin/godot* --doctool .
```
akien-mga: That's not going to work if you have more than one `godot*` file in `./bin`.

development/cpp/custom_modules_in_cpp.rst
```diff
@@ -449,54 +449,94 @@ There are several steps in order to setup custom docs for the module:
...
 
-This will dump the engine API reference to the given ``<path>`` in XML format.
-Notice that you'll need to configure your ``PATH`` to locate Godot's executable,
-and make sure that you have write access rights. If not, you might encounter an
-error similar to the following:
+.. note::
 
-.. code-block:: console
+    If you want, you can prevent the core classes to be generated with the
```
akien-mga:
```suggestion
    If you want, you can prevent the core classes from being generated with the 
```

akien-mga: But I'd straight out remove this note, it's not relevant for this use case.
What it does it that it doesn't dump documentation for Variants, but it still includes everything else from the API, so:
- Its actual use case is dubious
- It's not relevant for writing module docs

development/cpp/custom_modules_in_cpp.rst
Xrayez: I'd personally expect the documentation to assume `godot` is at the PATH or so, to simplify this for the sake of example/readability.

development/cpp/custom_modules_in_cpp.rst
```diff
@@ -449,54 +449,94 @@ There are several steps in order to setup custom docs for the module:
...
+    ``--no-docbase`` command line option::
 
-    ERROR: Can't write doc file: docs/doc/classes/@GDScript.xml
-       At: editor/doc/doc_data.cpp:956
+          user@host:~/godot$ ./bin/godot* --doctool . --no-docbase
 
-4. Get generated doc file from ``godot/doc/classes/ClassName.xml``
+Now if you go to the ``godot/modules/summator/doc_classes`` folder, you will see, 
+that now we have a ``Summator.xml`` inside it, along with every class, that 
+you added with your ``get_doc_classes`` function.
```
akien-mga:
```suggestion
Now if you go to the ``godot/modules/summator/doc_classes`` folder, you will see 
that it contains a ``Summator.xml`` file, or any other classes, that you referenced
in your ``get_doc_classes`` function.
```

development/cpp/custom_modules_in_cpp.rst
```diff
@@ -449,54 +449,94 @@ There are several steps in order to setup custom docs for the module:
...
 
-5. Copy this file to ``doc_classes``, optionally edit it, then compile the engine.
+Now you can edit this file, and recompile the engine.
```
akien-mga:
```suggestion
Edit the file(s) following :ref:`doc_updating_the_class_reference` and recompile the engine.
```

development/cpp/custom_modules_in_cpp.rst
```diff
@@ -449,54 +449,94 @@ There are several steps in order to setup custom docs for the module:
...
 
-The build system will fetch the documentation files from the ``doc_classes`` directory
-and merge them with the base types. Once the compilation process is finished,
-the docs will become accessible within the engine's built-in documentation system.
+Once the compilation process is finished, the docs will become accessible within 
+the engine's built-in documentation system.
 
 In order to keep documentation up-to-date, all you'll have to do is simply modify
-one of the ``ClassName.xml`` files and recompile the engine from now on.
+one of the xml files and recompile the engine from now on.
```
akien-mga:
```suggestion
one of the XML files and recompile the engine from now on.
```

development/cpp/custom_modules_in_cpp.rst
Xrayez: Last time I tested `--no-docbase` will also remove all base docs as well, polluting the `git status` with deleted classes, but this something which should be fixed IMO.

development/cpp/custom_modules_in_cpp.rst
Relintai: The problem with this one is the section directly before this one, as we have explanations about building modules as shared libraries (near line 391):

```
...

Once compiled, we should end up with a ``bin`` directory containing both the
``godot*`` binary and our ``libsummator*.so``. However given the .so is not in
a standard directory (like ``/usr/lib``), we have to help our binary find it
during runtime with the ``LD_LIBRARY_PATH`` environ variable:

::

    user@host:~/godot$ export LD_LIBRARY_PATH=`pwd`/bin/
    user@host:~/godot$ ./bin/godot*

... bit more explanation ...

Writing custom documentation
----------------------------
```

I have an idea that *might* be the best of both worlds. I'll push it soon, along with all the other corrections.

Also I personally don't think the executable should be in the PATH in this example, because we build it a section earlier.

Relintai:
Pushed up the new version. 

Apart from the typo fixes, In 'Improving the build system for development' I also touched 2 lines a little bit.

About the `godot*` commands, I opted to change both occurrences to `<godot_binary>`, as the binary's name changes for every platform and target.

I also thought about changing the current directory to `bin`, but I think that would easily lead to confusion when moving  between sections. Al least as a bonus everything can still be easily copy pasted.

Note:
I think that the 'Improving the build system for development' section could be moved to the end of the page, above the 'Summing up' section.
Actually, both it's name, and it's content makes it fit there perfectly (for example it only works on linux). This would makes the document flow a bit better.
(I included this first, but realized that it will make reviewing annoying, so I removed it. I will open a separate pr for it.)

development/cpp/custom_modules_in_cpp.rst
Relintai: I removed this one.

mhilbrunner:
Hey! Sorry it took a bit to get back to this.

Thanks for the PR, it looks good to merge to me - could you fix the conflict? :)

Relintai:
Done

mhilbrunner:
Agreed :)

Thank you for this, and again, sorry it took a while!

PR #3009, opened 2019-12-27
fixed variable name

Enhex:
fixed the variable name - there's no variable named `player_name`, it should be `my_info`.


Calinou:
Thanks! And congratulations for your first documentation contribution :slightly_smiling_face:

PR #3010, opened 2019-12-27
Use sound pressure level rather than power

TaleTN:
See https://en.wikipedia.org/wiki/Decibel#Acoustics (or any audio editor, DAW, analog mixing console, etc.).


Calinou:
Does this represent what Godot currently does with sound levels?

PR #3011, opened 2019-12-28
Clarify touch-input-from-mouse

DesiQuintans:


flofriday:
As someone who just finished the exporting tutorial, this would have saved me 15 minutes.

Maybe we should also change the screenshot to one where both boxes are checked ?

Calinou:
Thanks!

Issue #2997, opened 2019-12-20
Creating a more complete GDScript Style Guide


Scony:
order of class definitions should be agreed and mentioned in the proper docs section.

It may be either fine-grained like e.g.:
```
- tools
- extends
- classnames
- enums
- consts
- signals
- exports
- onreadypubvars
- pubvars
- onreadyprvvars
- prvvars
- others
```

or coarse-grained like e.g.:
```
headers (tool, extend, classname)
definitions (enums, exports, const etc.)
publics
privates
```

but debate is needed

winston-yallow:
I am in favor of the coarse grained order. It would allow code like this:
```gdscript
enum COLOR {BLUE, RED, GREEN}
export(COLOR) var outline_color

enum SHAPE {TRIANGLE, SQUARE}
export(SHAPE) var debug_shape
```

I think when an enum is used only for one (*not* multiple) exports it helps to have it in the line above the export. I feel like it is much more readable to group them by usecase rather than by type.

*(just adding my point I made on discord here for reference)*

TheDuriel:
https://gist.github.com/TheDuriel/e34278a84ea163858d2d5e578b666dbc

NathanLovato:
We've worked on a GDScript style-guide based on the GDScript styleguide from the docs this year with the team. It puts readability and accessibility first, and the information users of a given class need first.
We've been using it and improving throughout the year.

Below is the quick outline. Here is the [complete style guide](https://www.gdquest.com/docs/guidelines/best-practices/godot-gdscript/#code-writing-style). In short, it puts properties at the top, public before private, node dependencies before variables, and initialization functions and callbacks before other built-in callbacks:

```
01. extends and class_name
02. """docstring"""

03. signals
04. `onready` variable
05. exports
06. enums
07. constants
08. public variables
09. private variables

10. optional built-in virtual `_init` method
11. built-in virtual `_ready` method
12. signal callbacks
13. remaining built-in virtual methods
14. public methods
15. private methods
16. static functions
```

We'd like to suggest it as:

1. It's designed to make the code easy to read and understand what's happening, so it works well for teaching.
1. It's the result of teamwork. We've discussed it a lot, we use it with students, got feedback from professional developers as well. We've used it in our recent [beginner platformer Godot course](https://youtu.be/Mc13Z2gboEk) and it doesn't seem to cause issues even for programming beginners.
1. It's designed for teamwork but works well for individual work as well.

[...]

PR #3012, opened 2019-12-28
Add step re. setting main scene

DesiQuintans:
My Windows executable would not run until I figured this out.

### DIFF
--- a/getting_started/step_by_step/exporting.rst
+++ b/getting_started/step_by_step/exporting.rst
@@ -119,6 +119,13 @@ changed:
         emit_signal("hit")
         $CollisionShape2D.set_deferred("disabled", true)
 
+Setting a main scene
+--------------------
+
+The main scene is the one that your game will start in. In *Project -> Project 
+Settings -> Application -> Run*, set *Main Scene* to "Main.tscn" by clicking 
+the folder icon and selecting it.
+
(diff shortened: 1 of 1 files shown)


NathanLovato:
That's correct, thanks!

getting_started/step_by_step/exporting.rst
```diff
@@ -119,6 +119,13 @@ changed:
         emit_signal("hit")
         $CollisionShape2D.set_deferred("disabled", true)
 
+Setting a main scene
+--------------------
+
+The main scene is the one that your game will start in. In *Project -> Project 
+Settings -> Application -> Run*, set *Main Scene* to "Main.tscn" by clicking
```
Sslaxx: Being more explicit about "Main.tscn" might be helpful? Is this a convention that should be followed or is it just an example? Might be worth covering main scene naming conventions (if it's worth it) in https://github.com/godotengine/godot-docs/issues/2997?

NathanLovato: Code architecture (where you place your classes, how your organize your files, separate systems) and naming might be beyond the scope of _GDScript_ programming guidelines. Your main scene could be anything, like the entry point of a python program can be any function, module, or package, really.

PR #3015, opened 2019-12-29
Add a `FAQ` meta keyword to the frequently asked questions page

Calinou:
This makes it possible to find that page by simply searching for `faq` in the documentation search. Previously, no results would have been returned.

It looks a bit strange, maybe there's a way to add invisible text that's still accounted for in the search results.

### DIFF
--- a/about/faq.rst
+++ b/about/faq.rst
@@ -1,3 +1,6 @@
+.. meta::
+    :keywords: FAQ
+
 .. _doc_faq:
 
 Frequently asked questions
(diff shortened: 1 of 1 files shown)


xelivous:
[as per the restructuredtext basics](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#html-metadata), you can just add the following to the top of the faq page and it should work:

```rst
.. meta::
   :keywords: FAQ
```

Calinou:
@xelivous I didn't know about that Sphinx feature, thanks :slightly_smiling_face:

It seems to work using Sphinx's client-side search.

NathanLovato:
Thanks!

PR #3016, opened 2019-12-29
moved explanations of fragment shader passes

bluenote10:
Currently the [Light built-ins](https://docs.godotengine.org/en/3.1/tutorials/shading/shading_reference/canvas_item_shader.html#light-built-ins) section contains information that actually doesn't apply to the `light` shader, but rather the `fragment` shader. This is confusing, for instance the mentioned `AT_LIGHT_PASS` doesn't exist in the light shader, but in the fragment shader.

It would make more sense to me to move the paragraph into the fragment shader section above, serving as a transition to the light shader.

### DIFF
--- a/tutorials/shading/shading_reference/canvas_item_shader.rst
+++ b/tutorials/shading/shading_reference/canvas_item_shader.rst
@@ -191,7 +191,7 @@ When the shader is on a light pass, the ``AT_LIGHT_PASS`` variable will be ``tru
 Light built-ins
 ^^^^^^^^^^^^^^^
 
-The `light` shader runs for each fragment of each light source.
+The `light` processor function runs for each fragment of each light source.
 
 +-------------------------------------+-------------------------------------------------------------------------------+
 | Built-in                            | Description                                                                   |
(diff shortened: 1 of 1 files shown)


bluenote10:
> Further, you are mixing up shaders (e.g. CanvasItem shader/ Spatial shader) with processor functions (e.g. vertex, fragment, light)

I'm coming straight from OpenGL, so yes where `vertex` and `fragment` are really different shaders. So yes, in Godot terms I mean "processor function".

> For CanvasItem shaders the entire shader is run again for each light. The light processor function is only called once per fragment, but both it and the rest of the shader are called once per light (or just once if there is no light). The light processor function just contains the logic for light calculations.

Hm, that was exactly my understanding of it. But it was hard to come to that conclusion because the docs only mention that the shader runs multiple times in the last section, which describes the `light` processor function. This gave the impression that only the `light` functions runs multiple times, not the `fragment` function, which is explained in the section above. That's why it would be clearer to me if this paragraph is mentioned earlier, not just in the `light` section. In particular, the information of having a `AT_LIGHT_PASS` built-in seems to be wrong in the `light` section, and must be moved to the `fragment` section.

clayjohn:
Sorry for the delay. Even with the small change that is left, the documentation is still rendered incorrect. 

the ``light()`` function is just a function called from within the fragment function. It doesn't run or not run based on the light itself. Accordingly, the ``light()`` function is called once for each fragment, for each light. It just has no effect outside the light's radius/texture. 

I am closing this PR for now as the original intended changes are not correct. However, I feel bad closing this and I do not want to discourage you as a first time contributor. If you have more changes you think would clarify the behaviour of the CanvasItemShaders. Please open another PR.

PR #3017, opened 2019-12-29
add a few code snippets and tips for "Your first game" tutorial

HuangJunye:
To make the tutorial more clear

### DIFF
--- a/getting_started/step_by_step/your_first_game.rst
+++ b/getting_started/step_by_step/your_first_game.rst
@@ -746,6 +746,9 @@ function to set everything up for a new game:
 
 .. tabs::
  .. code-tab:: gdscript GDScript
+ 
+    func _on_Player_hit():
+         game_over()
 
     func game_over():
         $ScoreTimer.stop()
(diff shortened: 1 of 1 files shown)


getting_started/step_by_step/your_first_game.rst
```diff
@@ -746,6 +746,9 @@ function to set everything up for a new game:
 
 .. tabs::
  .. code-tab:: gdscript GDScript
+ 
+    func _on_Player_hit():
+         game_over()
```
cbscribe: I don't think this is a good addition. It's extra code for no reason.

It's good for new learners to understand that when they connect a signal, they don't have to create a new function but can connect to an existing one.

HuangJunye: This is exactly the reason why I want to add the line. I want to show new learners what the code should look like after the connecting the signal. I missed this part when I ran through the tutorial and there was an error when running the game. Perhaps we can put these two lines separately to highlight this part. 

In general, I think it would be nice to have a way to show the entire script after each step (I am not sure how), because new learners like me are not sure if we are doing the right thing.

mhilbrunner:
I'll close this now, as it's been open for a while and there doesn't seem to be a clear consensus.
This tutorial is currently being reworked by cbscribe.

Thanks for taking the time to submit a Pull Request and contribute!

PR #3020, opened 2019-12-31
Add yasm to scoop install.

madmiraal:
Windows version of #1849. Helps avoid godotengine/godot#21719 when building on Windows.


Calinou:
Thanks!

madmiraal:
@Calinou Scoop renames it yasm and puts it in the path; so it's picked up by Scons here:
https://github.com/godotengine/godot/blob/27b74f42b4fd6b6e56b7a4924697d8881488bee1/modules/webm/libvpx/SCsub#L285-L288

PR #3022, opened 2020-01-01
GDScript style guide: add suggested code order

NathanLovato:
This PR is a proposal to extend the [GDScript style guide](http://docs.godotengine.org/en/latest/getting_started/scripting/gdscript/gdscript_styleguide.html) with a suggested convention for the order of code in GDScript classes. It follows the discussion in #2997.

The intention behind these new guidelines is to save everyone time:

1. Coming up with a consistent style anyone can use to write code that's easy to read.
2. Helping everyone to get up to speed with GDScript. This would give some good practices to get programming beginners and developers coming from other languages started.
3. Having a reference for code reviews for the official docs and demos, leaving less space to personal taste on code formatting.

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_styleguide.rst
+++ b/getting_started/scripting/gdscript/gdscript_styleguide.rst
@@ -37,8 +37,8 @@ Here is a complete class example based on these guidelines:
     export var initial_state = NodePath()
     var is_active = true setget set_is_active
 
-    onready var state = get_node(initial_state) setget set_state
-    onready var _state_name = state.name
+    onready var _state = get_node(initial_state) setget set_state
+    onready var _state_name = _state.name
 
     func _init():
         add_to_group("state_machine")
@@ -46,15 +46,15 @@ Here is a complete class example based on these guidelines:
 
     func _ready():
         connect("state_changed", self, "_on_state_changed")
-        state.enter()
+        _state.enter()
 
 
     func _unhandled_input(event):
-        state.unhandled_input(event)
+        _state.unhandled_input(event)
 
 
     func _physics_process(delta):
-        state.physics_process(delta)
+        _state.physics_process(delta)
 
 
     func transition_to(target_state_path, msg={}):
@@ -64,10 +64,10 @@ Here is a complete class example based on these guidelines:
         var target_state = get_node(target_state_path)
         assert target_state.is_composite == false
 
-        state.exit()
-        self.state = target_state
-        state.enter(msg)
-        Events.emit_signal("player_state_changed", state.name)
+        _state.exit()
+        self._state = target_state
+        _state.enter(msg)
+        Events.emit_signal("player_state_changed", _state.name)
 
 
     func set_is_active(value):
@@ -78,8 +78,8 @@ Here is a complete class example based on these guidelines:
 
 
     func set_state(value):
-        state = value
-        _state_name = state.name
+        _state = value
+        _state_name = _state.name
 
 
     func _on_state_changed(previous, new):
@@ -408,23 +408,24 @@ We suggest to organize GDScript code this way:
 
 ::
 
-    01. class_name
-    02. extends
-    03. # docstring
+    01. tool
+    02. class_name
+    03. extends
+    04. # docstring
 
-    04. signals
-    05. enums
-    06. constants
-    07. exported variables
-    08. public variables
-    09. private variables
-    10. onready variables
+    05. signals
+    06. enums
+    07. constants
+    08. exported variables
+    09. public variables
+    10. private variables
+    11. onready variables
 
-    11. optional built-in virtual _init method
-    12. built-in virtual _ready method
-    13. remaining built-in virtual methods
-    14. public methods
-    15. private methods
+    12. optional built-in virtual _init method
+    13. built-in virtual _ready method
+    14. remaining built-in virtual methods
+    15. public methods
+    16. private methods
 
 We optimized the order to make it easy to read the code from top to bottom, to
 help developers reading the code for the first time understand how it works, and
@@ -442,7 +443,10 @@ This code order follows four rules of thumb:
 Class declaration
 ~~~~~~~~~~~~~~~~~
 
-Start with the `class_name` if necessary. You can turn a GDScript file into a
+If the code is meant to run in the editor, place the ``tool`` keyword on the
+first line of the script.
+
+Follow with the `class_name` if necessary. You can turn a GDScript file into a
 global type in your project using this feature. For more information, see
 :ref:`doc_gdscript`.
 
@@ -524,11 +528,11 @@ in that order.
 
     func _ready():
         connect("state_changed", self, "_on_state_changed")
-        state.enter()
+        _state.enter()
 
 
     func _unhandled_input(event):
-        state.unhandled_input(event)
+        _state.unhandled_input(event)
 
 
     func transition_to(target_state_path, msg={}):
@@ -538,10 +542,10 @@ in that order.
         var target_state = get_node(target_state_path)
         assert target_state.is_composite == false
 
-        state.exit()
-        self.state = target_state
-        state.enter(msg)
-        Events.emit_signal("player_state_changed", state.name)
+        _state.exit()
+        self._state = target_state
+        _state.enter(msg)
+        Events.emit_signal("player_state_changed", _state.name)
 
 
     func _on_state_changed(previous, new):
--- a/getting_started/scripting/gdscript/gdscript_styleguide.rst
+++ b/getting_started/scripting/gdscript/gdscript_styleguide.rst
@@ -3,20 +3,89 @@
 GDScript style guide
 ====================
 
-Description
------------
-
-This styleguide lists conventions to write elegant GDScript. The goal is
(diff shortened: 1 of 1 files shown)


NathanLovato:
Feedback on the code examples is welcome. The StateMachine class is a complete example that should follow all the guidelines in the guide.

YeldhamDev:
Isn't `"""this"""` a string instead of comment? And thereof not scrapped away when the script is compiled?

NathanLovato:
Is this an issue? It's still treated as a comment, it won't give any compilation errors: http://docs.godotengine.org/en/3.1/getting_started/scripting/gdscript/gdscript_basics.html#comments

NathanLovato:
Ah, I see that multi-line comments have been removed from the latest docs. So we should use `# ` for multi-line comments then?

YeldhamDev:
@NathanLovato As far as I'm aware, multi-line strings do consume space, even if unused, so they should be dropped as comments.

NathanLovato:
This should be fixed

YeldhamDev:
Here's a discussion related to that: https://github.com/godotengine/godot/issues/31060

getting_started/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -241,12 +460,14 @@ characters in a given string. See the examples below:
     # Both quote styles would require 2 escapes; prefer double quotes if it's a tie.
     print("'hello' \"world\"")
 
+.. _naming_conventions:
+
```
aaronfranke: Is this addition supposed to be here?

getting_started/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -3,21 +3,240 @@
...
+Then, add the `extends` keyword if the class extends a built-in type.
+
+Following that, you should have the class's optional docstring as comments. You
+can use that to explain the role of your class to your teammates, how it works,
+and how other developers should use it, for example.
+
+::
+
+   extends Node
+   class_name MyNode
```
aaronfranke: Same problem here.

Feniks-Gaming:
I think we should also discuss if static typing is something required in a style. All example use static typing despite GDScript being primarily dynamic language. I don't think it's good idea to confine users of dynamic language into constraints of typed language. 

I personally preferred typed GDScript but I can see it losing a lot of appeal if we insist that static typing is "default way" of using the language.

If we then insist that static typing is something we want as default we would have to have discussion which style to use `var a := 5` or `var a: int = 5`

I feel that for now all examples should use dynamic version of the language as this is still used as default and statistic typing is referred to as "optional static typic" something that is optional shouldn't make it into examples of official general style guide.

And I say this as someone who prefers static typing

cbscribe:
> I feel that for now all examples should use dynamic version of the language

I agree 100%. Making static type-hinting mandatory removes a great deal of the appeal of GDScript as a dynamic language. Static typing should be in a separate section of the style guide clearly indicating that *if* you're using it, here are the prescribed guidelines.

getting_started/scripting/gdscript/gdscript_styleguide.rst
NathanLovato: Yes it's how you link to sections in rst

NathanLovato:
I just removed the type hints and fixed the typos

Feniks-Gaming:
I don't see any more problems after the changes you implemented now.

aaronfranke:
Shouldn't these additions be at the end instead of the beginning of the document? Code order is a low priority compared to naming conventions and formatting conventions, so I think it should go at the end.

NathanLovato:
My rationale was that it should be a code formatter that formats your code. It's a job for machines. You shouldn't have to worry about spaces, where to insert a line break etc. 

With python you don't need to read pep8 today, you install autopep8 or black and run it automatically.

However, code structure and, yes, naming, which is yet another thing, are something individual developers need to learn.

I'd put naming conventions first, then code structure, and finally, formatting.

Feniks-Gaming:
Is there anything that stops this from being merged I think all issues have been resolved?

NathanLovato:
Yes, the changes haven't been validated by a reviewer yet.

NathanLovato:
I made the changes, left the class example at the top, moved the code order to the bottom of the article, just before the short section on optional static type hints.

Feniks-Gaming:
It doesn't look like there is much dispute here any more I would like to see this merge ASAP as it would make review of all demo projects much easier because it's hard to accept any of those knowing poor guys will have to edit it soon anyway.

getting_started/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -3,21 +3,90 @@
...
+    # Hierarchical State machine for the player.
+    # Initializes states and delegates engine callbacks (_physics_process, _unhandled_input) to the state.
+
+
+    signal state_changed(previous, new)
+
+    export var initial_state = NodePath()
+    var is_active = true setget set_is_active
+
+    onready var state = get_node(initial_state) setget set_state
```
Scony: in the past, I had a discussion with @TheDuriel regarding `public onready variable` and the outcome was that it is always bad practice since it's all about exposing internals to the clients

getting_started/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -326,6 +397,158 @@ are constants:
...
+    02. extends
+    03. # docstring
+
+    04. signals
+    05. enums
+    06. constants
+    07. exported variables
+    08. public variables
+    09. private variables
+    10. onready variables
```
Scony: why not before private variables?

getting_started/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -326,6 +397,158 @@ are constants:
...
+Code order
+----------
+
+This first section focuses on code order. For formatting, see
+:ref:`code_formatting`. For naming conventions, see :ref:`naming_conventions`.
+
+We suggest to organize GDScript code this way:
+
+::
+
```
Scony: what about `tool` keyword?

getting_started/scripting/gdscript/gdscript_styleguide.rst
NathanLovato: As discussed in the issue, several persons want it as close as possible to the _ready function, as onready is evaluated right before _ready and after other properties. But still grouped with the other variables as it's part of the class's properties.

getting_started/scripting/gdscript/gdscript_styleguide.rst
NathanLovato: So what would you do with this case? You'd encapsulate the states in a state machine away from the users? And... add extra methods and wrapper properties to get the info you need from them?

If it's public here, it's because we want to access the active state from outside the state machine. Among other things, we've used the property to display info about the active state on-screen, in a debug console.

In this case, it's quite inconvenient to e.g. bubble the state with signals. And the above code is not inconsistent with some of Godot's API, which exposes some members nodes or objects as properties at times.

getting_started/scripting/gdscript/gdscript_styleguide.rst
NathanLovato: I'd put it at the top: it's important to know that a given piece of code will run in the editor, be it as part of a plugin or not. I don't mind exactly where, but I'd just put it on the first line.

getting_started/scripting/gdscript/gdscript_styleguide.rst
Scony: ok

Scony:
You may also consider inlining some comment-hints in the code example - similarly as TheDuriel does: https://gist.github.com/TheDuriel/e34278a84ea163858d2d5e578b666dbc

getting_started/scripting/gdscript/gdscript_styleguide.rst
NathanLovato: I can make the state variable private no problem. Making it public is not necessarily the best example for e.g. a dev who doesn't have much experience and who may use that in a different context.

NathanLovato: Also if you'd prefer another class example I can come up with one - it's just that this is a complete class that is short and shows most of the guidelines at the same time.

Scony: either way will be fine ^^

NathanLovato:
I added a separate commit so you can review the changes. There's the `tool` keyword and the state property is now private in the class example.

If the general guidelines are okay, we can merge and open an issue or new PRs to refine the document. The docs are like a wiki, they're constantly updated during release cycles, so we can have a more iterative approach compared to the engine repo.

NathanLovato:
Could someone merge the PR now it got three OKs from reviewers? If it needs any improvement later, you can always open an issue later and ping me.

NathanLovato:
@Calinou do you have merge rights? If so, would you mind merging for us? The PR got 3 positive reviews.

Calinou:
Thanks!

NathanLovato:
No, thank you. 🙂

PR #3023, opened 2020-01-01
Add warning to making-plugins reminding about tool keyword

N0hbdy:
I've repeatedly forgotten that other scripts used in editor plugins
*also* need to have the `tool` keyword, otherwise it fails to correctly
run.

View of changes:

Screen Shot 2020-01-01 at 2 46 16 PM

### DIFF
--- a/tutorials/plugins/editor/making_plugins.rst
+++ b/tutorials/plugins/editor/making_plugins.rst
@@ -72,6 +72,11 @@ EditorPlugin script for you. The script has two requirements that you cannot
 change: it must be a ``tool`` script, or else it will not load properly in the
 editor, and it must inherit from :ref:`class_EditorPlugin`.
 
+.. warning::
+    In addition to the EditorPlugin script, any other GDScript that your plugin uses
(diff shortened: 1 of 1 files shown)


Calinou:
Thanks!

tutorials/plugins/editor/making_plugins.rst
```diff
@@ -72,6 +72,11 @@ EditorPlugin script for you. The script has two requirements that you cannot
 change: it must be a ``tool`` script, or else it will not load properly in the
 editor, and it must inherit from :ref:`class_EditorPlugin`.
 
+.. warning::
+    In addition to EditorPlugin script, any other GDScript that your plugin uses
```
Calinou: Here's a small correction, it should be good to go after this change:

```suggestion
    In addition to the EditorPlugin script, any other GDScript that your plugin uses
```

N0hbdy: Thanks, should be fixed!

PR #3024, opened 2020-01-01
Expand Inspector Plugin tutorial

N0hbdy:
As a first-time creator of the inspector plugin (and plugins in general), it took
a while to actually understand exactly what's happening and
what's needed for an inspector plugin, so I've expanded with the information
that I felt would help make the tutorial easier for people doing their first
foray into making one of these plugins.

This adds the "missing" pieces that I feel would have helped me move a bit faster
in getting my plugin up and running a bit faster!

### DIFF
--- a/tutorials/plugins/editor/inspector_plugins.rst
+++ b/tutorials/plugins/editor/inspector_plugins.rst
@@ -8,17 +8,46 @@ editing properties. This tutorial explains how to use the
 :ref:`class_EditorInspectorPlugin` and :ref:`class_EditorProperty` classes to
 write such plugins with the example of creating a custom value editor.
 
-.. note::
+Setup
+-----
 
-    To register these scripts as a new editor plugin, you have to create a
-    ``plugin.cfg`` file as described in :ref:`doc_making_plugins`.
+Just like :ref:`doc_making_plugins`, we start out by making a new plugin,
+getting a ``plugin.cfg`` file created, and start with our
+:ref:`class_EditorPlugin`.  However, instead of using
+``add_custom_node`` or ``add_control_to_dock`` we'll use
+``add_inspector_plugin``.
+
+.. tabs::
+  .. code-tab:: gdscript GDScript
+    # MyEditorPlugin.gd
+
+    tool extends EditorPlugin
+
+    var plugin: EditorInspectorPlugin
+
+    func _enter_tree():
+        # EditorInspectorPlugin is a resource, so we use `new()` instead of `instance()`.
+        plugin = preload("res://addons/MyPlugin/MyInspectorPlugin.gd").new()
+        add_inspector_plugin(plugin)
+
+    func _exit_tree():
+        remove_inspector_plugin(plugin)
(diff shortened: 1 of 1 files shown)


tutorials/plugins/editor/inspector_plugins.rst
```diff
@@ -8,17 +8,48 @@ editing properties. This tutorial explains how to use the
...
+
+.. tabs::
+  .. code-tab:: gdscript GDScript
+    # MyEditorPlugin.gd
+
+    tool extends EditorPlugin
+
+    var plugin: EditorInspectorPlugin
+
+    func _enter_tree() -> void:
```
Feniks-Gaming: Examples shouldn't use static typing as GDScript is primary dynamic language. It helps to avoid confusion for the doc users. Static is only optional way of using GDScript

tutorials/plugins/editor/inspector_plugins.rst
```diff
@@ -8,17 +8,48 @@ editing properties. This tutorial explains how to use the
...
 
 EditorInspectorPlugin
 ---------------------
 
-We start by creating a script extending the :ref:`class_EditorInspectorPlugin`
-class. This is needed to initialize the plugin and add the custom property
-editor that we'll later define.
+To actually connect into the Inspector, we create a
+:ref:`class_EditorInspectorPlugin` class.  This script provides
+the "hooks" into the inspector.  It is used to check if we should enable
```
Feniks-Gaming: What do you mean by "hooks" I don't understand can you expand.

tutorials/plugins/editor/inspector_plugins.rst
N0hbdy: Makes sense, I'll get this updated!

tutorials/plugins/editor/inspector_plugins.rst
N0hbdy: Hmm... callback?  The way that this works is you set it up, and the editor will call the functions within the EditorInspectorPlugin while it goes through the process of building the UI for the inspector.

Generally, I've always referred to this as things that "hook into" an existing flow to modify it.   I'm by no means married to that though, so if we think of a better way of explaining it, then I'd be super happy!  This is probably one of the harder systems in here, compared to most other UI work!

Feniks-Gaming: I think this explanation now makes more sense

So maybe something like 

To actually connect into the Inspector, we create a
:ref:`class_EditorInspectorPlugin` class.  This script the "hooks into" the inspector.  
Thanks to that the editor will call the functions within the EditorInspectorPlugin 
while it goes through the process of building the UI for the inspector. 
It is used to check if we should enable ourselves for any :ref:`class_Object` 
that is currently in the inspector (including any :ref:`class_Resource` that is embedded!).

Or something similar just to explain what you understand by providing hooks

Calinou:
Thanks! I applied the suggestions manually before merging.

PR #3026, opened 2020-01-02
Changed link to GPU intensive website.

vintprox:
There is known issue with heavy page load at www.shadertoy.com, which imposes GPU intensive shaders on main page: https://bugs.chromium.org/p/chromium/issues/detail?id=33620. It leads user's PC to go nuts even on second load. Bug is marked as "WontFix", which is unsettling.

I thought that it's quite exhaustive user experience, that distracts from reading documentation. So, I've placed link with reduced search results to avoid this inconsistency.


Calinou:
Thanks!

PR #3028, opened 2020-01-02
Changled link to Shadertoy (mirror #3026)

vintprox:
#3026
There is known issue with heavy page load at www.shadertoy.com, which imposes GPU intensive shaders on main page: https://bugs.chromium.org/p/chromium/issues/detail?id=33620. It leads user's PC to go nuts even on second load.


Calinou:
Thanks!

PR #3032, opened 2020-01-03
Reinsert clamp in exporting docs

flofriday:
Before this change clamping the players position got removed with the following explanation:

> We don't need to clamp the player's position because you can't click outside the screen.

While it is true that the player cannot click outside the window, the player can still click outside the playable area. Because we set Aspect to "keep" in the Project Settings, Godot will create Black borders if the window (or phone screen) doesn't match the desired aspect ratio. A player can now click on either of the borders and because clamp got removed the Player node will move out of the playable area and hide in the black border.

This is why the clamp should not be removed.

Here is a demo of the bad behaivoir:

ScreenRecording2020-01-03at15110

### DIFF
--- a/getting_started/step_by_step/exporting.rst
+++ b/getting_started/step_by_step/exporting.rst
@@ -104,7 +104,7 @@ changed:
         # We still need to clamp the player's position here because on devices that don't
         # match your game's aspect ratio, Godot will try to maintain it as much as possible 
         # by creating black borders, if necessary.
-        # Without clamp() the player would be able to move under those borders.
+        # Without clamp(), the player would be able to move under those borders.
         position.x = clamp(position.x, 0, screen_size.x)
         position.y = clamp(position.y, 0, screen_size.y)
 
--- a/getting_started/step_by_step/exporting.rst
+++ b/getting_started/step_by_step/exporting.rst
@@ -101,10 +101,8 @@ changed:
             $AnimatedSprite.stop()
 
         position += velocity * delta
-        # We don't need to clamp the player's position
-        # because you can't click outside the screen.
-        # position.x = clamp(position.x, 0, screensize.x)
-        # position.y = clamp(position.y, 0, screensize.y)
+        position.x = clamp(position.x, 0, screen_size.x)
+        position.y = clamp(position.y, 0, screen_size.y)
 
         if velocity.x != 0:
             $AnimatedSprite.animation = "right"
--- a/getting_started/step_by_step/exporting.rst
+++ b/getting_started/step_by_step/exporting.rst
@@ -104,7 +104,7 @@ changed:
         # We still need to clamp the player's position here because on devices that don't
         # match your game's aspect ratio, Godot will try to maintain it as much as possible 
         # by creating black borders, if necessary.
-        # Without clamp the player would be able to move under those borders.
+        # Without clamp() the player would be able to move under those borders.
         position.x = clamp(position.x, 0, screen_size.x)
         position.y = clamp(position.y, 0, screen_size.y)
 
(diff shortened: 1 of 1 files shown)


getting_started/step_by_step/exporting.rst
```diff
@@ -101,10 +101,8 @@ changed:
             $AnimatedSprite.stop()
 
         position += velocity * delta
-        # We don't need to clamp the player's position
```
Sslaxx: While these comments are not relevant, something along the lines of your explanation regarding aspects and screen resizing would not go amiss.

flofriday: Since the code part I modified starts with the following line I thought that no comments were necessary.
> Here is the full script for the player, with comments noting what we've changed:


But maybe you are right, would something like this improve the code?
```
position += velocity * delta
# We still need to clamp the player's position here because Godot tries to 
# keep the aspect ratio by creating black borders if necessary.
# Without clamp the player would be able to move under those borders.
position.x = clamp(position.x, 0, screen_size.x)
position.y = clamp(position.y, 0, screen_size.y)
```

Or do you have something different on your mind?
(Please feel free to improve the text, I am not a native English speaker)

flofriday:
Tell me if you need anything else from me to merge this.
🙂

getting_started/step_by_step/exporting.rst
Feniks-Gaming: I would slightly expand this to 

```
# We still need to clamp the player's position here because on devices that don't
# mach your game's aspect ratio Godot will try to maintain it as much as possible 
#  by creating black borders if necessary.
# Without clamp the player would be able to move under those borders.
```

it gives a little more information as to why this could be a case and why does Godot create black bars out of nowhere

flofriday: I added the comments to explain why the clamp is still needed.

flofriday:
Is there a reason this isn't merged yet ?

Should I change anything about this PR ?

getting_started/step_by_step/exporting.rst
```diff
@@ -101,10 +101,12 @@ changed:
...
-        # because you can't click outside the screen.
-        # position.x = clamp(position.x, 0, screensize.x)
-        # position.y = clamp(position.y, 0, screensize.y)
+        # We still need to clamp the player's position here because on devices that don't
+        # match your game's aspect ratio, Godot will try to maintain it as much as possible 
+        # by creating black borders, if necessary.
+        # Without clamp the player would be able to move under those borders.
```
Calinou:
```suggestion
        # Without clamping, the player would be able to move under those borders.
```

Feniks-Gaming: Maybe rather `# Without using clamp()`?

flofriday: I too would prefer clamp().

I am gonna change it to that in 30 minutes, if that is ok with everyone.

getting_started/step_by_step/exporting.rst
```diff
@@ -101,10 +101,12 @@ changed:
...
 
         position += velocity * delta
-        # We don't need to clamp the player's position
-        # because you can't click outside the screen.
-        # position.x = clamp(position.x, 0, screensize.x)
-        # position.y = clamp(position.y, 0, screensize.y)
+        # We still need to clamp the player's position here because on devices that don't
+        # match your game's aspect ratio, Godot will try to maintain it as much as possible 
+        # by creating black borders, if necessary.
+        # Without clamp() the player would be able to move under those borders.
```
Calinou: Sorry for nitpicking, but there should still be a comma here.

```suggestion
        # Without clamp(), the player would be able to move under those borders.
```

flofriday: Yeah sure :)

Calinou:
Thanks!

PR #3038, opened 2020-01-04
Update docs regarding current Nuget workflow

jasonswearingen:
highlighting proper workflow as per my discovery here: https://github.com/godotengine/godot/issues/34803#issuecomment-570752688

basically it's super easy for c# devs to get the nuget workflow wrong.   this clarifies problems i ran into.


aaronfranke:
Pinging @neikeq

Calinou:
@neikeq Is the workaround above still required in the current `master` branch? If so, I'll merge this.

neikeq:
Yes, this is still requires. It will no longer be needed soon though (I expect to get this solved in no more than two months), so it won't be needed in the 4.0 docs. Older versions will still need it.

Calinou:
Thanks! I applied some formatting changes manually as reStructuredText doesn't support nested formatting (i.e. `**some text ``and code`` and more text**` won't work).

PR #3045, opened 2020-01-08
Added changes about use of Dynamic Typing

Feniks-Gaming:
I have added explanation directing doc contributors to only use Dynamic typed GDScript in their examples unless they are specifically discussing issues related to static typing.

### DIFF
--- a/community/contributing/docs_writing_guidelines.rst
+++ b/community/contributing/docs_writing_guidelines.rst
@@ -274,6 +274,29 @@ element of a list with more than two elements.
 How to write methods and classes
 --------------------------------
 
+Dynamic vs Static typing
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Despite the addition of optional static typing, GDScript remains primarily 
+dynamic language. Unless you are specifily discussing static typing, you
+should only use dynamic typing in your code examples and snipets. 
+This will help reduce confusion to new users by promoting consistent look
+to the code examples they are exposed to. 
+
+**Don't** write using using static typing:
+
+::
+
+    var a := 5
+    var b: String = "My String"
+
+**Do** write code using dynamic typing:
+
+::
+
+    var a = 5
+    var b = "My String"
+
 Give an overview of the node in the brief description
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/community/contributing/docs_writing_guidelines.rst
+++ b/community/contributing/docs_writing_guidelines.rst
@@ -277,25 +277,43 @@ How to write methods and classes
 Dynamic vs Static typing
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
-Despite the addition of optional static typing, GDScript remains primarily 
-dynamic language. Unless you are specifily discussing static typing, you
-should only use dynamic typing in your code examples and snipets. 
-This will help reduce confusion to new users by promoting consistent look
-to the code examples they are exposed to. 
+GDScript aims to be concise and easy to learn. Optional static typing conflicts with this vision
+by making documentation more congested. To improve its accessibility for new users, please use 
+dynamic GDScript code samples where possible. The one exception is topics that explain 
+static typing concepts to users.
 
-**Don't** write using using static typing:
+**Don't** add a type hint with a colon or by casting:
 
 ::
 
     var a := 5
     var b: String = "My String"
+    var my_sprite : Sprite = $Sprite as Sprite
 
-**Do** write code using dynamic typing:
+
+**Do** write varibles with dynamic typing:
 
 ::
 
     var a = 5
     var b = "My String"
+    var my_sprite = $Sprite
+
+
+**Don't** write functions with inferred arguments or return types:
+
+::
+
+    func my_func(arg1 : int, arg2 : String = "") -> void:
+    return
+
+**Do** write functions using dynamic typing:
+
+::
+
+    func my_func(arg1, arg2 = ""):
+    return
+
 
 Give an overview of the node in the brief description
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/community/contributing/docs_writing_guidelines.rst
+++ b/community/contributing/docs_writing_guidelines.rst
@@ -286,35 +286,43 @@ static typing concepts to users.
 
 ::
 
-    const MyNode := preload("preload("res://my_node.gd")
-    var a := 5
-    var b: String = "My String"
-    var my_sprite := $Sprite as Sprite
+    const main_attack := preload("res://fire_attack.gd")
+    var hit_points := 5
+    var name: String = "Bob"
+    var body_sprite := $Sprite as Sprite
 
 
 **Do** write constants variables with dynamic typing:
 
 ::
 
-    const MyNode = preload("preload("res://my_node.gd")
-    var a = 5
-    var b = "My String"
-    var my_sprite = $Sprite
+    const main_attack = preload("res://fire_attack.gd")
+    var hit_points = 5
+    var name = "Bob"
+    var body_sprite = $Sprite
 
 
 **Don't** write functions with inferred arguments or return types:
 
 ::
 
-    func my_func(arg1 : int, arg2 : String = "") -> void:
-        pass
+    func choose(arguments: Array): 
+        #Chooses one of the arguments from array with equal chance
+        randomize()
+        var size := arguments.size()
+        var choice: int = randi() % size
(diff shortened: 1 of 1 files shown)


Feniks-Gaming:
My first ever pull request so I am very open for guidance if I did anything wrong.

community/contributing/docs_writing_guidelines.rst
```diff
@@ -274,6 +274,29 @@ element of a list with more than two elements.
...
+Dynamic vs Static typing
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Despite the addition of optional static typing, GDScript remains primarily 
+dynamic language. Unless you are specifily discussing static typing, you
+should only use dynamic typing in your code examples and snipets. 
+This will help reduce confusion to new users by promoting consistent look
+to the code examples they are exposed to. 
+
+**Don't** write using using static typing:
```
willnationsdev: Assorted grammar/spelling:
- "primarily dynamic" -> "primarily a dynamic"
- "specifily" -> "specifically"
- "snipets" -> "snippets"
- "promoting consistent" -> "promoting a consistent"
- "they are exposed to" -> "to which they are exposed"
- "using using static typing" -> "using static typing"

---
Generally, we will want to make simpler, direct sentences that avoid adverbs and minimize sentence length (at least, in my experience). So, for example:

> GDScript specializes in being concise and easy to learn. Optional static typing conflicts with this vision by making documentation more congested. To improve its accessibility for new users, please use dynamic GDScript code samples where possible. The one exception is topics that explain static typing concepts to users.

community/contributing/docs_writing_guidelines.rst
```diff
@@ -274,6 +274,29 @@ element of a list with more than two elements.
...
+
+::
+
+    var a := 5
+    var b: String = "My String"
```
willnationsdev: Might want to also include an example that uses casting, i.e. `as <Type>`.

community/contributing/docs_writing_guidelines.rst
```diff
@@ -274,6 +274,29 @@ element of a list with more than two elements.
...
+
+**Do** write code using dynamic typing:
+
+::
+
+    var a = 5
+    var b = "My String"
```
willnationsdev: Perhaps also share some code samples that seem like they are "static typing" but which actually aren't. For example:

Script classes...

    extends Node
    class_name MyNode # not actually static typing yet. Just makes a global variable

`is` checks...

    if node is MyNode: # compares an object to a Script at runtime
         pass

preloaded constants without a colon...

    const MyNode = preload("res://my_node.gd")

community/contributing/docs_writing_guidelines.rst
Feniks-Gaming: Thanks. I appreciate your comments it's late now but I will work on this tomorrow and submit all the changes.

community/contributing/docs_writing_guidelines.rst
Feniks-Gaming: I am not sure if class_name is good example here to use it's under static typing in docs so I would assume it's part of static typing way?

willnationsdev: Hmmm. I suppose it sort of fits the bill for "static typing". It's just a global variable, but because it is evaluated at parse-time, it ends up having requirements that must be met in the "static" editor context rather than runtime.

Feniks-Gaming:
@willnationsdev are there any other changes you would like to see or did I miss anything I am happy to add/change

community/contributing/docs_writing_guidelines.rst
```diff
@@ -277,25 +277,43 @@ How to write methods and classes
...
 
 ::
 
     var a := 5
     var b: String = "My String"
+    var my_sprite : Sprite = $Sprite as Sprite
 
-**Do** write code using dynamic typing:
+
+**Do** write varibles with dynamic typing:
```
willnationsdev: "varibles" -> "variables"

community/contributing/docs_writing_guidelines.rst
```diff
@@ -277,25 +277,43 @@ How to write methods and classes
...
+static typing concepts to users.
 
-**Don't** write using using static typing:
+**Don't** add a type hint with a colon or by casting:
 
 ::
 
     var a := 5
     var b: String = "My String"
+    var my_sprite : Sprite = $Sprite as Sprite
```
willnationsdev: Idiomatic static typing would be to use a `:=` for type inference since the cast implies what the type will be inferred as. No need to restate the "Sprite" class twice.

    var my_sprite := $Sprite as Sprite

community/contributing/docs_writing_guidelines.rst
```diff
@@ -277,25 +277,43 @@ How to write methods and classes
...
     var b = "My String"
+    var my_sprite = $Sprite
+
+
+**Don't** write functions with inferred arguments or return types:
+
+::
+
+    func my_func(arg1 : int, arg2 : String = "") -> void:
+    return
```
willnationsdev: Some best practices:
1. Make sure the body of these method calls is indented one level (on the below example too).
1. If the return type is void, then you don't even need to explicitly `return`.
1. If there is no actual content to a method, the keyword you are supposed to use is `pass`. It explicitly communicates to the user that the method is a stub and needs to be implemented (or that it is a pure virtual function that is meant to be left blank). Writing `return` muddles the intent of the script.

community/contributing/docs_writing_guidelines.rst
NathanLovato: By the way, between writing the type and using casting, which would you favor? `var sprite : Sprite = get_node("Body")` or `var sprite := get_node("Body") as Sprite`?

community/contributing/docs_writing_guidelines.rst
NathanLovato: I would also recommend having concrete examples in the docs. While developers or technical writers can make sense of `foobar` examples, I find that real-world examples are more accessible to beginners. Plus you can copy-paste them straight from your own projects.

It's not specific to this page, but there's a lot of `var my_thing = random_stuff` in the docs. That's many missed opportunities to show code examples that people could use.

community/contributing/docs_writing_guidelines.rst
Feniks-Gaming: I am not certain adding this will work well it may be to confusing for a new person so for now I will leave it out unless you feel this is absolutely necessary addition?

Feniks-Gaming:
@NathanLovato would you like me to change the examples of code to something that could be used in real life instead? We could actually make that a requirement of writing good docs? what do you think?

NathanLovato:
> We could actually make that a requirement of writing good docs? what do you think?

Yes, we could.

In my experience so far, contributors don't necessarily check the guidelines so you end up having to tell the same thing over and over.

We could use some tech writer(s) working full-time on the docs for some time, focusing on editing. If the docs has a certain baseline quality, contributors can use the page they're editing as a reference when making changes or adding paragraphs.

NathanLovato:
> would you like me to change the examples of code to something that could be used in real life instead?

If you feel like it sure. I think it's generally beneficial to the reader.

Feniks-Gaming:
I agree 100% about tech writers. I think short term initiatives like this could be dealt with by collecting some money via kickstarter to then give grant to Godot team to do just that. Maybe something worth considering.

What I will do for now is: After work I will edit my functions and variables to reflect real life cases. Once this PR is merged I will open new one with guidance on using real life examples in code snippets so we have some base line for review to sing from the same hymn sheet.

Feniks-Gaming:
I have added real code examples @NathanLovato hope those look better

NathanLovato:
It's good, thanks, I'll remove the remaining typo. Note that for preloaded nodes in constants, as they represent a type, the convention is to use PascalCase:

```gdscript
const HealthBar = preload("res://.../health_bar.gd")
...
HealthBar.new()
```

Also, in the future, could you amend your commits? I'll squash them this time around, but it's best to not have a git history like:

```
Final change
Add another change
Make change
```

To do so, use `git commit --amend` when you made and staged changes, followed by `git push --force` to replace the old commit on your repository. Just note that this updates the content of your commit, it's like updating a save in a game.

Feniks-Gaming:
Can you amend commit directly in github. I have done all of it directly on github online and haven't use any commands. I opened a file with by clicking "edit file" it created clone of repository on my github account and as soon as I saved a file it pushed changes.

NathanLovato:
Ah that's why. No, you can't use most of git's features from GitHub.

Feniks-Gaming:
Well I guess I will have to learn how to do it properly next time

NathanLovato:
If you need a hand to get started, ping me, be it on Discord, on a PR. I don't know how much experience you have with git but it can be a bit complicated to use until it clicks.

PR #3047, opened 2020-01-09
Mention VSync as a fix for Jittering

jasonswearingen:
**Anecdote**:  On my trivial prototype, I spent multiple hours trying to figure out why Multithreaded Rendering caused Jitter, while Single-Threaded rendering did not.   It turns out that Multithreaded causes VSync issues on my machine, and while VSync is enabled by default, it's Windows-Specific counterpart for non-fullscreen apps is not.  (Godot v3.2b5)

Documenting this is important otherwise people will waste a lot of time trying to figure this out and might end up thinking the rendering subsystem is buggy.

### DIFF
--- a/tutorials/misc/jitter_stutter.rst
+++ b/tutorials/misc/jitter_stutter.rst
@@ -49,15 +49,15 @@ framerate performance. Solving this is game specific and will require optimizati
 The second is more complicated, because it is often not associated to the engine or game but the underlying operating system.
 Here is some information regarding stutter on different OSs.
 
-VSync
-^^^^^
+Use VSync
+^^^^^^^^^
 
 A common cause of Jitter is when the game's framerate is out of step with the monitor's refresh rate.  Go to ``` Project Settings -->  Display --> Window``` and ensure that ```Vsync``` is enabled, and if you are on Windows, also ensure that ```Vsync Via Compositor``` is enabled.
 
 .. note::
 
     ```Vsync Via Compositor```, a Windows-specific setting, is not enabled by default.  
-    If you are on Windows and seeing Jitter, enabling this as a first step.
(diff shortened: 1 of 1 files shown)


tutorials/misc/jitter_stutter.rst
```diff
@@ -49,6 +49,16 @@ framerate performance. Solving this is game specific and will require optimizati
 The second is more complicated, because it is often not associated to the engine or game but the underlying operating system.
 Here is some information regarding stutter on different OSs.
 
+VSync
+^^^^^
+
+A common cause of Jitter is when the game's framerate is out of step with the monitor's refresh rate.  Go to ``` Project Settings -->  Display --> Window``` and ensure that ```Vsync``` is enabled, and if you are on Windows, also ensure that ```Vsync Via Compositor``` is enabled.
```
Feniks-Gaming: It's `Use Vsync` not just `Vsync` also can you clarify where use needs to select `Vsync Via Compositor`  I don't know where this setting is. Finally `if you are using Windows` not if `if you are on Windows`

Calinou:
We may enable **Vsync Via Compositor** by default in a 3.2 point release or 4.0. We just need further testing to confirm it works well on various hardware configurations.

starry-abyss:
Anyway, IMHO https://docs.godotengine.org/en/latest/tutorials/misc/jitter_stutter.html should be updated to mention the new Godot 3.2 jitter fixes (Windows compositor, smooth addon, etc).

Calinou:
Due to the regressions related to **Vsync via Compositor** and https://github.com/godotengine/godot/pull/35070#issuecomment-573891667, I'm afraid we can't recommend doing this as this would affect all players. We could still point to the `--enable-vsync-via-compositor` command-line argument as a last resort though.

PR #3048, opened 2020-01-09
Adding information and guidelines about how to properly create a standard CSV translation file

timkrief:
RFC-4180: "If double-quotes are used to enclose fields, then a double-quote appearing inside a field must be escaped by preceding it with another double quote."
I tried to escape it with backslash without knowing the special standards of CSV, it didn't work and I [opened an issue for nothing](https://github.com/godotengine/godot/issues/34949). I think it might be right to explain that directly in the docs so that no ones lack this information and ask for help about that again.


Calinou:
It'd be good to include an example in a code snippet, just so people know what it looks like in practice. Other than that, this looks good to me.

Calinou:
@timkrief Either is fine to me :slightly_smiling_face:

Calinou:
Thanks!

PR #3051, opened 2020-01-10
Fix malformed table in "Inporting translations" page

YeldhamDev:


timkrief:
Japanese characters count for two characters and are not mono-space, so the table had a wrong size and did not displayed at all when I changed it in my last PR without knowing this. This PR fix this issue and must be merged.

PR #3052, opened 2020-01-10
Example table was not displayed

timkrief:
Japanese characters count for two characters and are not mono-space, so the table had a wrong size and did not displayed at all. It should work now.


Feniks-Gaming:
There is already the same issue dealth with ongoing PR https://github.com/godotengine/godot-docs/pull/3051

timkrief:
Already solved by: #3051 , please, merge it

PR #3053, opened 2020-01-10
Expanded guidance about avoiding made up names

Feniks-Gaming:
As we discussed here https://github.com/godotengine/godot-docs/pull/3045#discussion_r365111553 I have expanded guidance to mention that we should avoid using made up names and use real world working code examples where practically possible. And where not possible still use a code that at least resembles something you could see in a project.

### DIFF
--- a/community/contributing/docs_writing_guidelines.rst
+++ b/community/contributing/docs_writing_guidelines.rst
@@ -308,7 +308,7 @@ The exception is topics that explain static typing concepts to users.
 
 ::
 
-    func choose(arguments: Array): 
+    func choose(arguments: Array):
         # Chooses one of the arguments from array with equal chances
         randomize()
         var size := arguments.size()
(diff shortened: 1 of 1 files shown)


community/contributing/docs_writing_guidelines.rst
```diff
@@ -326,6 +326,40 @@ The exception is topics that explain static typing concepts to users.
...
+    func foo():
+        # Do stuff
+
+**Do** write concrete examples:
+
+::
+
+    onready var SFXPlayerGun = preload("res://Assets/Sound/SFXPlayerGun.ogg")
+    onready var AudioPlayer = $Audio/AudioStreamPlayer
+
```
cbscribe: Per styleguide, these variable names should be snake_case.

Feniks-Gaming:
@NathanLovato I could use this help with git thing because I have clearly no idea. I have github desktop and I can't understand how to amend my commits like you wanted me to do. I tried looking for tutorial on it but couldn't find any. 

I can't understand command line, I absolutely hate it and will not use it. I am happy to learn any and all GUI based client if github desktop is not good for it I am happy to try others.

NathanLovato:
> I can't understand command line, 

That's not true. It's much simpler to learn than a programming language, so you can learn to understand it. You may just need a nice shell (especially if you're on windows, cmd is not intuitive at all) and a clear introduction to what shells are and how to work with them. Also, I would recommend any developer to take some time to learn it as it's the kind of tool that can save you a lot of time.

The issue with graphical clients is that they hide a lot of how git works, and git being a complex program, they tend to make you mess things up.

GitHub desktop doesn't seem to allow you to edit the history. It's not available on Linux but it seems to only give you some basic features. Gitkraken is a common client that I heard is much more complete. They also have video guides: https://support.gitkraken.com/working-with-commits/squash/

Feniks-Gaming:
GitKraken did the trick thanks a lot. 

My main issue with command line comes from my dyslexia if I can't see everything laid out nicely it may as well not happen at all. I can follow a bunch of code on one screen I can't follow screen changing with each command line and can't keep the track of what happens.

@cbscribe are the changes satisfactory for you is there anything else that is missing?

NathanLovato:
> My main issue with command line comes from my dyslexia

Ah, that makes sense. It's completely different in this case.

It's good now. Thanks!

PR #3055, opened 2020-01-11
Update compiling_for_osx.rst

clemens-tolboom:
No sure 8 jobs is best for everyone but it boost compiling

### DIFF
--- a/development/compiling/compiling_for_osx.rst
+++ b/development/compiling/compiling_for_osx.rst
@@ -26,7 +26,7 @@ Start a terminal, go to the root dir of the engine source code and type:
 
 ::
 
-    user@host:~/godot$ scons platform=osx --jobs=8
+    user@host:~/godot$ scons platform=osx -j$(sysctl -n hw.logicalcpu)
 
 If all goes well, the resulting binary executable will be placed in the
 "bin" subdirectory. This executable file contains the whole engine and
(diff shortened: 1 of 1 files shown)


Calinou:
It's probably better to use a substitution that automatically detects the number of CPU threads:

```
scons -j$(sysctl -n hw.logicalcpu)
```

clemens-tolboom:
@Calinou that's way better :)

development/compiling/compiling_for_osx.rst
```diff
@@ -26,7 +26,7 @@ Start a terminal, go to the root dir of the engine source code and type:
 
 ::
 
-    user@host:~/godot$ scons platform=osx
+    user@host:~/godot$ scons platform=osx -j$(sysctl -n hw.logicalcpu)
```
Calinou: I also wonder if we should remove the shell placeholder to make it easier to copy-paste:

```suggestion
    scons platform=osx -j$(sysctl -n hw.logicalcpu)
```

This way, the text can be clicked 3 times to select the entire line (without any extraneous text).

clemens-tolboom: Done ... I reintroduced `--jobs` as that helps for scons novices like me.

Should I squash the 3 commits?

Calinou: It's OK, I can squash them myself.

Calinou:
Thanks!

clemens-tolboom:
Thanks too!

PR #3056, opened 2020-01-12
Add a dark theme using the `prefers-color-scheme` CSS media query

### LINKED ISSUE #3041: Add an (optional) dark theme
Please add a button somewhere in the docs to switch to a dark theme.
I am addicted to dark themes, thanks to Sublime Text.
I am currently using a browser plugon for it, but it is not very efficient.

Calinou:
This also includes tweaks to the syntax highlighting theme which were necessary to support switching the theme using CSS variables. Some token colors in the light syntax theme have been swapped around for consistency.

This closes #3041.

## Preview

Index

GDScript basics

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -1,8 +1,16 @@
 /**
- * Various tweaks to the Read the Docs theme to better conform with Godot's visual identity.
+ * Various tweaks to the Read the Docs theme to better conform with Godot's
+ * visual identity. Many colors are also overridden to use CSS variables.
+ * This makes it possible to provide an automatically-used dark theme
+ * based on browser preferences.
  */
 
+ /* Default (light) theme colors */
  :root {
+    --body-color: #404040;
+    --content-wrap-background-color: rgba(0, 0, 0, 0.05);
+    --content-background-color: #fcfcfc;
+    --logo-opacity: 1.0;
     --navbar-background-color: #333f67;
     --navbar-background-color-hover: #29355c;
     --navbar-background-color-active: #212d51;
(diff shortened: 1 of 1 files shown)


nikhilCad:
Thank you very much! Though, I still believe that yo should add a button somewhere.

Calinou:
@nikhilCad It's possible to add such a button, but it's much more involved as it involves adding some custom JavaScript and HTML elements.

NathanLovato:
How can I test this on Linux? I tried to look for a setting in both firefox and chrome, but there's nothing about theme preferences. And changing my system to use a dark theme by default doesn't affect the website.

Calinou:
@NathanLovato This add-on might work: https://addons.mozilla.org/en-US/firefox/addon/dark-mode-website-switcher/?src=external-twitter

NathanLovato:
I tried two addons but couldn't get the theme to switch to dark mode on my browsers. Other websites do use the dark mode, but not the freshly built local html files.

I don't see anything wrong in the css and the website is working fine so I'll merge now. If anybody finds an

Calinou:
> Other websites do use the dark mode, but not the freshly built local html files.

That might be due to a caching issue, you may have to remove the `_build/html` folder entirely after switching branches.

NathanLovato:
I checked that the css was correct and up to date earlier. I deleted the build dir and started from scratch, cleared the cache, but couldn't get it to work. Anyway that's no big deal, I'm sure it'll work better for others, maybe on Windows?

Calinou:
@NathanLovato I was able to test this pull request using Firefox on Linux, so I don't think that's the issue. Either way, I can now see my changes on http://docs.godotengine.org/en/latest/ :tada:

PR #3059, opened 2020-01-13
Expanded class_bool Description with examples

Feniks-Gaming:
I expanded boolean long description with example of when booleans could be use practice because 

Boolean 

Description:

Boolean built-in type.

was pretty vague. As new to programming person it took me longer than it should have to understand what those are and why I would want to use them. Hopefully this will make it clearer to new users reading documentation

### DIFF
--- a/classes/class_bool.rst
+++ b/classes/class_bool.rst
@@ -30,7 +30,66 @@ Methods
 Description
 -----------
 
-Boolean built-in type.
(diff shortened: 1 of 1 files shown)


Calinou:
Contributions to the class reference should be sent on the main Godot repository, not here. See [Contribute to the class reference](https://docs.godotengine.org/en/latest/community/contributing/updating_the_class_reference.html) in the documentation :slightly_smiling_face:

PR #3060, opened 2020-01-13
Update some exporting images and move a line

skyace65:
Updates 3 of the images on the exporting page for Godot 3.2. All images have gone through png compression. I also moved one line to before one of the images so people don't get confused and think there should already be a list of export presets.

The Android images need updating but I'm not familiar with android and I think other changes need to be made when that is done. So I'm leaving it for now. And I don't have time to prepare stuff for the ios export image.


getting_started/step_by_step/exporting.rst
```diff
@@ -153,13 +153,13 @@ version that matches your version of Godot.
 Export presets
 --------------
 
-Next, you can configure the export settings by clicking on *Project -> Export*:
-
-.. image:: img/export_presets_window.png
+Next, you can configure the export settings by clicking on *Project -> Export*
```
Calinou: There should be a period as the colon was removed:
```suggestion
Next, you can configure the export settings by clicking on *Project -> Export*.
```

skyace65:
That should be fixed now.

Calinou:
Thanks!

Issue #3057, opened 2020-01-12
Incorrect syntax highlighting for languages other than GDScript.


asynts:
In the documentation code blocks are created using the `code` directive. But this doesn't work for languages other than GDScript for some reason.

Look at [development/cpp/core_types.rst][1] for example:

~~~rst
.. code:: cpp

    for(List<int>::Element *E=somelist.front();E;E=E->next()) {
        print_line(E->get()); //print the element
    }
~~~

In the [online documentation][2] this is displayed as follows:

![Screenshot from 2020-01-12 11-44-40](https://user-images.githubusercontent.com/31994781/72217636-f5fe2f80-3530-11ea-88a5-419fdd0694c3.png)

At first glance it seems that it's highlighted correctly, but the `print` in the comment is highlighted for some reason. Inspecting the element in the browser reveals the issue:

~~~html
<div class="code cpp highlight-gdscript notranslate">
    <div class="highlight">
        <!-- ... -->
    </div>
</div>
~~~

As the class on the enclosing div suggests, the code block was syntax highlighted as GDScript.

I've never used reStructuredText for anything myself, but after a quick Google search I found this [stack overflow question][3]:

 >  When using `code-block` I always get the correct highlighting. When using `code` I sometimes get
 >  colors and sometimes just literal code blocks. I have yet to figure out what combinations of setting
 >  in conf.py that predictably generates colored output.
 >  
 >  The code directive does have the advantage that the document can be used both in a Sphinx
 >  document tree and at the same time as a stand-alone reStructuredText document.

And indeed, changing the code block to

~~~rst
.. code-block:: cpp

    for(List<int>::Element *E=somelist.front();E;E=E->next()) {
        print_line(E->get()); //print the element
    }
~~~

fixed the issue on my machine:

![Screenshot from 2020-01-12 12-13-37](https://user-images.githubusercontent.com/31994781/72217930-02848700-3535-11ea-8438-a2a155350a30.png)

Is there any reason to use `code` over `code-block`? If not I'd make a pull request to use `code-block` over `code`.

  [1]: https://github.com/godotengine/godot-docs/blob/9302b01152fa0083473086699024dbbe530af0c9/development/cpp/core_types.rst
  [2]: https://docs.godotengine.org/en/3.1/development/cpp/core_types.html#containers
  [3]: https://stackoverflow.com/q/34845889/8746648

akien-mga:
Fixed by #3061.

PR #3061, opened 2020-01-13
Incorrect syntax highlighting for languages other than GDScript.

asynts:
This pull request solves the problem I outlined in #3057.

As already mentioned in the issue the solution is to replace `code` directives with `code-block` directives.

I am still unsure as to why the `code` directive doesn't do the job, but `code-block` seems to be the correct directive since this project uses Sphinx.

I also changed a few minor things that I noticed while making the change, the commits should speak for themselves.

### DIFF
--- a/community/contributing/code_style_guidelines.rst
+++ b/community/contributing/code_style_guidelines.rst
@@ -149,7 +149,7 @@ ones, the following rules should be followed:
 
 Example:
 
-.. code:: cpp
+.. code-block:: cpp
 
     /*************************************************************************/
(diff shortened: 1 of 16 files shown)


akien-mga:
I wanted to look a bit more into it to understand what fundamental difference there is between `.. code::` and `.. code-block::`, and why the latter works fine when the former is iffy, but I can't find much details about that.

If this PR makes things work, let's go with it and enforce usage of `.. code-block::` over `.. code::`.

akien-mga:
Thanks!

PR #3062, opened 2020-01-13
Update 2D Sprite Animation images and tweak tutorial

skyace65:
Updates the two outdated images in the 2D Sprite Animation page, both images have gone through png compression. I also added a specific FPS number the animation should be changed to in the tutorial.


skyace65:
I replaced the images. How is it now?

PR #3063, opened 2020-01-14
Don't reinstall dependencies if they are already installed.

asynts:
Arch Linux's `pacman -S` re-installs packages that are already installed,
unless the option `--needed` is provided.

A user that wants to build Godot probably doesn't want to reinstall their
GCC compiler and all the other stuff listed in the command.


NathanLovato:
Looking good, thanks!

PR #3064, opened 2020-01-14
Fix nonexistent reference in the "GDScript style guide" page

YeldhamDev:
I assume that this is what it was referencing to.


NathanLovato:
Thanks, indeed! I probably broke it when moving paragraphs from the last PR around.

PR #3067, opened 2020-01-16
Fixed errors on Design a title screen tutorial

Darenn:
In "Prepare the main menu scene" : 
- Changed the viewport layout before changing the margins, as it resets the margins.
- Margin right and margin bottom had wrong values.


Calinou:
Are you sure about the MarginContainer custom constants? With positive values only, I get this:

image

With the right and bottom margins set to -120 and -80 respectively, I get this:

image

Darenn:
You're right, don't know how I got that.

PR #3069, opened 2020-01-16
Updates Godot versions in PR Examples

Feniks-Gaming:
Updates PR workflow to use more recent examples of Godot versions. We were still talking about 2.1.5 or Godot 3.0.2  so I updates it to more reflect where we are in development.


Calinou:
Thanks!

PR #3073, opened 2020-01-18
Center the page on wide displays

Calinou:
This prevents the main content from being stuck at the left of the page, which in turn makes it more comfortable to read. The page width is otherwise the same as before.

## Preview

Documentation on 2560×1440 monitor

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -8,7 +8,7 @@
  /* Default (light) theme colors */
  :root {
     --body-color: #404040;
-    --content-wrap-background-color: rgba(0, 0, 0, 0.05);
+    --content-wrap-background-color: #efefef;
     --content-background-color: #fcfcfc;
     --logo-opacity: 1.0;
     --navbar-background-color: #333f67;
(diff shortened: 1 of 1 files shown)


NathanLovato:
It's working well here as well. How about having a darker color in the background (empty area) in the light theme? The value contrast could help emphasize the article:

Screenshot from 2020-01-18 16-26-31-fs8

NathanLovato:
But for now, I'll merge, the change is a welcome improvement already

Calinou:
@NathanLovato That can be done, but I'm not sure which color we should use in the light theme. Maybe a dark blue (slightly darker than the navbar background, but not too much)?

NathanLovato:
That should work. I can't say what would be best, I tried pure black and dark grey earlier, both looked fine.

timothyqiu:
Just noticed that the online version's Read the Docs bar does not have the offset.

demo

Calinou:
@timoschwarzer This is addressed by #3084.

timoschwarzer:
Fix wrong @
cc @timothyqiu

PR #3077, opened 2020-01-19
Mention export template caveats when creating a C++ module

Calinou:
See https://github.com/godotengine/godot/issues/35299.


development/compiling/index.rst
```diff
@@ -1,3 +1,5 @@
+.. _doc_compiling:
```
akien-mga: This should not be needed, you can reference `toc-devel-compiling` directly IINM.

akien-mga:
Thanks!

PR #3078, opened 2020-01-19
Mention the Godot data paths specific to the Flatpak version

Calinou:
Data paths in the Godot Flatpak are different from the non-Flatpak version. This documents the Flatpak-specific paths.


akien-mga:
Thanks!

PR #3082, opened 2020-01-20
Updated style guide to include enums and dir in exception to 2…

Feniks-Gaming:
It appears that general consensus across the docs is that 2 indents for continuous line rule does not apply to enums and dictionaries. This is not mentioned in docs which is confusing to anyone who doesn't know about unofficial exception. This commit makes it official

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_styleguide.rst
+++ b/getting_started/scripting/gdscript/gdscript_styleguide.rst
@@ -140,6 +140,96 @@ regular code blocks.
         sprite.get_scale(), Vector2(2.0, 2.0), 0.3,
         Tween.TRANS_QUAD, Tween.EASE_OUT)
 
+Exceptions to this rule are arrays, dictionaries, and enums. Use a single
+indentation level to distinguish continuation lines:
+
+**Good**:
+
+::
+
+    var party = [
+        "Godot",
+        "Godette",
+        "Steve",
+    ]
+
+    var character_dir = {
+        "Name": "Bob",
+        "Age": 27,
+        "Job": "Mechanic",
+    }
+
+    enum Tiles {
+        TILE_BRICK,
+        TILE_FLOOR,
+        TILE_SPIKE,
+        TILE_TELEPORT,
+    }
+
+**Bad**:
+
+::
+
+    var party = [
+            "Godot",
+            "Godette",
+            "Steve",
+    ]
+
+    var character_dir = {
+            "Name": "Bob",
+            "Age": 27,
+            "Job": "Mechanic",
+    }
+
+    enum Tiles {
+            TILE_BRICK,
+            TILE_FLOOR,
+            TILE_SPIKE,
+            TILE_TELEPORT,
+    }
+
+Trailing comma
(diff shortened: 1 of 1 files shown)


Calinou:
I agree with this change :slightly_smiling_face: 

Does this also apply to arrays? I've always used only 1 indent for them personally.

Feniks-Gaming:
To be honest it really applies to everything but func() from what I have seen across the docs. I can add Array to the list of single indent if we feel like we should. I feel we could do since it's used as that anyway.

Calinou:
I applied some changes manually. Thanks for your contribution :slightly_smiling_face:

getting_started/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -140,6 +140,96 @@ regular code blocks.
...
+            TILE_SPIKE,
+            TILE_TELEPORT,
+    }
+
+Trailing comma
+~~~~~~~~~~~~~~
+
+Use a trailing comma on the last line in arrays, dictionaries, and enums. This
+results in easier refactoring and better diffs in version control as the last
+line doesn't need to be modified when adding new elements.
```
YeldhamDev: Ehhh, I'm very against this one. Makes it look like the dicts and enums are imcomplete. Does other places in the docs does that?

akien-mga: We also use this in the C++ code, it's much cleaner for version control, e.g.:
```diff
diff --git a/core/error_list.h b/core/error_list.h
index b464a93341..604351d9fb 100644
--- a/core/error_list.h
+++ b/core/error_list.h
@@ -88,6 +88,7 @@ enum Error {
        ERR_HELP, ///< user requested help!!
        ERR_BUG, ///< a bug in the software certainly happened, due to a double check failing or unexpected behavior.
        ERR_PRINTER_ON_FIRE, /// the parallel port printer is engulfed in flames
+       ERR_YET_ANOTHER_ERROR,
 };
 
 #endif
```

This is of course only for multiline declaration, it shouldn't be used when it's done inline (`enum SomeEnum { VAR_1, VAR_2 }`).

YeldhamDev: @akien-mga The absolute majority of the enums in the code doesn't have that.

Calinou: Most language style guides seem to have adopted a rule in favor of trailing commas (e.g. [Airbnb JavaScript](https://github.com/airbnb/javascript#commas--dangling), [PEP 8](https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas)).

Also, we probably intend to add trailing commas everywhere in 4.0 (https://github.com/godotengine/godot/issues/16521)).

YeldhamDev: I see. Yep, the diff argument convinced me.

Feniks-Gaming: So we are all good her then yeah?

PR #3084, opened 2020-01-20
Fix the Read the Docs version selector not being centered

### LINKED ISSUE #3081: Alignment of sidebar navigation is broken in the latest branch of the documentation.
As of very recently it seems that the navigation sidebar for the live version of the documentation site has issues with rendering. This appears to only effect the "latest" version branch of the documentation and none of the other versions are effected. This issue applies to the current stable versions of Firefox, Chrome, and Edge on Windows 10.

Should be fairly easy to reproduce, simply switch the documentation version to "latest" and you will see something like the following image. Note the bottom left corner.

Annotation 2020-01-19 235640

Calinou:
This closes #3081.

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -559,6 +559,13 @@ code,
     background-color: var(--navbar-current-background-color);
 }
 
+@media only screen and (min-width: 768px) {
+    .rst-versions {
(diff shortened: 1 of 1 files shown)


akien-mga:
Thanks!

PR #3089, opened 2020-01-21
Fix the "Caution" admonition colors when using the dark theme

Calinou:
This can be seen by browsing [Exporting for Android](https://docs.godotengine.org/en/latest/getting_started/workflow/export/exporting_for_android.html) with the dark theme enabled (via the OS settings).


akien-mga:
Thanks!

PR #3090, opened 2020-01-21
Language improvements in Import Plugin tutorial

Sceluswe:
- Move the first note to the top of the document to stop it from breaking content flow and place it where it's the most relevant.
- Correct grammatical errors.
- Remove unnecessary words.
- Add missing words.


akien-mga:
Thanks!

For the reference, it's probably not necessary to be *that* verbose in your commit messages about the changes. Grammar fixes don't have to be justified unless the fix is not obvious for most English speakers.

Sceluswe:
> Thanks!
> 
> For the reference, it's probably not necessary to be _that_ verbose in your commit messages about the changes. Grammar fixes don't have to be justified unless the fix is not obvious for most English speakers.

Understood! I'll keep em short and sweet in the future. Glad to be of help.

PR #3094, opened 2020-01-22
Add DANGER admonition about freeing nodes in tool scripts

akien-mga:
Closes godotengine/godot#35422 and many similar misunderstandings.


akien-mga:
BTW that page could use a proofreading pass to improve the language (but that's outside the scope of this PR).

PR #3096, opened 2020-01-22
Fix gcc compilation flags for GDNative with C example

vnen:
`-rdynamic` flag is needed to properly export the symbols (otherwise Godot may not find them).

Also changed to use the more common `.o` extension for objects on Linux.


akien-mga:
Thanks!

PR #3099, opened 2020-01-23
Make the search bar fixed, hide the logo when scrolling down

### LINKED ISSUE #2778: Bad visibility of search field after first search
Because of the vast size of the Godot docs when using the online docs (readthedocs) it becomes increasingly difficult to search a second time as the content in the sidebar is so long you end up spending 10+ seconds scrolling with your mouse to reach the search field again. Page up/page down/home/etc keys do nothing as they stop working when you reach top of the content area.

See video

before

If we have the power to do so in readthedocs I would recommend something like "position: fixed" on the search field so that it always stays on top. Something like showcased in the following clip. (although a better implementation would probably be where Godot logo would scroll out of view only leaving the search field visible)

after

Calinou:
This closes #2778.

## Preview

See [this Gfycat](https://gfycat.com/secondfemininehalicore).

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -422,6 +422,26 @@ code,
     background-color: var(--navbar-background-color);
 }
 
+@media only screen and (min-width: 768px) {
+    .wy-side-nav-search {
+        /* Keep the search field visible when scrolling down */
+        position: fixed;
+    }
+
+    /* Simulate a drop shadow that only affects the bottom edge */
+    /* This is used to indicate the search bar is fixed */
+    .wy-side-nav-search::after {
+        content: '';
+        position: absolute;
+        left: 0;
+        bottom: -8px;
+        width: 300px;
+        height: 8px;
+        pointer-events: none;
+        background: linear-gradient(hsla(0, 0%, 0%, 0.2), transparent);
+    }
+}
+
 .wy-side-nav-search > a:hover,
 .wy-side-nav-search .wy-dropdown > a:hover {
     background-color: var(--navbar-background-color-hover);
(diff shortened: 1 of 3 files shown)


YeldhamDev:
I think it would be better if that bit of empty space between the logo and the search bar disappeared as well.

Calinou:
@YeldhamDev Done:

image

akien-mga:
Thanks!

mega-bit:
I'm experiencing some issues.

1.  The search bar doesn't stay at the top of the browser. This is in both Chromium 79 and Firefox 72. See a [gif](https://gfycat.com/recentvioletblackandtancoonhound)

2. When scrolling down very slowly the Godot Logo switches back and forth between collapsed and not-collapsed. This can easiest be reproduced when tapping the arrow-down-key. I only experience this one in Chromium. See a [gif](https://gfycat.com/chiefdapperchameleon)

I'm afraid the gifs don't show the issues quite well due to the low frame rate.

Lastly, I think it would be nicer if instead of jumping to the top immediately the search bar would be fixed once it has naturally reached the top. Right now it looks quite hectic.

Calinou:
@mega-bit I guess this is due to a caching issue (or Read the Docs not copying the updated CSS correctly). Try clearing your browser cache.

mega-bit:
I just did. Still behaves the same. Firefox has further no extensions installed.

YeldhamDev:
@Calinou I cleaned my cache and still could reproduce the issues above.
Also, in my self compiled version, the logo disappears too soon:
Screenshot_20200123_154408

Calinou:
@YeldhamDev Unfortunately, I noticed Read the Docs only applies CSS changes if you do multiple commits (it seems to ignore the first commit that added a file). I guess it's one reason to push changes to the documentation more often… :stuck_out_tongue:

GammaGames:
Thank you for this! I had a PR upstream in the theme that has been sitting ignored, I'm glad you were able to get it done, it works great!

PR #3100, opened 2020-01-23
Improve the fixed search bar

Calinou:
The search bar will now behave in a "sticky" manner. This makes the logo disappear more smoothly instead of being abruptly hidden. Another upside is that the drop shadow will only appear once the search bar is made fixed.

See #3099.

## Preview

See [this Gfycat](https://gfycat.com/serpentinecriminalhochstettersfrog).


akien-mga:
Thanks!

PR #3105, opened 2020-01-25
Keep velocity in _process as in the previous step

Beuc:


Calinou:
@Beuc Can you describe what this fixes or improves? You can reference a GitHub issue ID if this closes a particular issue.

Beuc:
Sure, in the previous tutorial step the variable `velocity` is within the `_process` function; in this next step it is moved to the top-level without indication (while all the changes to be made are explicitly explained by a comment); and, moving `velocity` to the top-level doesn't introduce any advantage as this is a local variable that is re-defined everytime _process is called and not used anywhere else.
So this change just keeps the variable where it was.

cbscribe:
This is likely an artifact of the code being changed in the previous tutorial but not this one. Good change for consistency.

cbscribe:
Thanks - congrats on your first PR!

PR #3107, opened 2020-01-25
Add GDScript reference grammar

vnen:
Add a page with a reference formal grammar definition for GDScript, written in standard EBNF. Note that this grammar may still need some tweaks, but generally I feel it's correct.

Also added casting to the precedence table, since it was missing.


akien-mga:
I feel like the "Getting Started" section is probably not the right place for this. Not sure where to put it though since the Scripting/GDScript reference is there.

Calinou:
There seems to be a GDScript language grammar available here: https://github.com/Scony/godot-gdscript-toolkit

vnen:
> I feel like the "Getting Started" section is probably not the right place for this. Not sure where to put it though since the Scripting/GDScript reference is there.

I'm not sure either. I considered putting it together with the engine development section, but all GDScript is in the "Getting Started". Maybe we can put it in the section with the file formats. 

> There seems to be a GDScript language grammar available here: https://github.com/Scony/godot-gdscript-toolkit

Didn't know this was a thing. Seems it only has a Lark grammar definition that is used by the tool itself. I may take a look to compare and see if I missed something.

> Was it used as a basis for this pull request, or did you make this one from scratch?

This one I wrote from scratch.

Calinou:
I think this would be good to merge after moving it to the File formats section of the documentation.

aaronfranke:
This indeed shouldn't be included in the getting started section. As a Godot user for over 2 years, I don't really know what this is for or what I'm even looking at. This should be in some kind of advanced section which new users wouldn't read, since I'd imagine most people would just be confused by this.

mhilbrunner:
When would this be useful?

Not saying it is never useful, just trying to gauge possible use cases vs. the maintenance effort this requires, as we'd need to update it everytime GDScript changes.

Calinou:
@mhilbrunner I think it's always a good practice to document custom file formats (like we already do with the TSCN format). This makes it easier for third-party tool authors to support GDScript.

mhilbrunner:
Superseded by https://github.com/godotengine/godot-docs/pull/3532 in which I moved the grammar to the custom file formats section

mhilbrunner:
Merged manually, after moving to custom file formats section, mentioned you as co-authored-by, hope thats enough :P 

Thank you!

PR #3109, opened 2020-01-27
Replace x11 with linux in GDNative C++ example

quantmint:
"x11" it's not the correct platform. It works with "linux".


Calinou:
Thanks! (For future reference, this applies only to GDNative's godot-cpp, not the Godot editor.)

PR #3112, opened 2020-01-27
Add a page about finding regressions using `git bisect`

Calinou:
This is a common process during Godot development. Making it more accessible to newcomers should help bugs be resolved faster.

~~@akien-mga Should we have a `bisect needed` label on GitHub issues? It'd make it easier for people to find bugs that need to be bisected before they can be solved. We could then link to the list of issues labeled `bisect needed` on that page.~~

**Edit:** The current `regression` label should be enough for now.

### DIFF
--- a/community/contributing/bisecting_regressions.rst
+++ b/community/contributing/bisecting_regressions.rst
@@ -1,0 +1,143 @@
+.. _doc_bisecting_regressions:
+
+Bisecting regressions
+=====================
+
+.. highlight:: shell
+
+Bisecting is a way to find regressions in software. After reporting a bug on the
+`Godot repository on GitHub <https://github.com/godotengine/godot>`__, you may
+be asked by a contributor to *bisect* the issue. Bisecting makes it possible for
+contributors to fix bugs faster, as they can know in advance which commit caused
+the regression. Your effort will be widely appreciated :)
+
+The guide below explains how to find a regression by bisecting.
+
+What is bisecting?
+------------------
+
+Godot developers use the `Git <https://git-scm.com/>`__ version control system.
+In the context of Git, bisecting is the process of performing a manual
+`binary search <https://en.wikipedia.org/wiki/Binary_search_algorithm>`__
+to determine when a regression appeared. While it's typically used for bugs,
+it can also be used to find other kinds of unexpected changes such as
+performance regressions.
+
+Using official builds to speed up bisecting
+-------------------------------------------
+
+Before using Git's ``bisect`` command, we strongly recommend trying to reproduce
+the bug with an older (or newer) official release. This greatly reduces the
+range of commits that potentially need to be built from source and tested.
+You can find binaries of official releases, as well as alphas, betas,
+and release candidates `here <https://downloads.tuxfamily.org/godotengine/>`__.
+
+For example, if you've reported a bug against Godot 3.2, you should first try to
+reproduce the bug in Godot 3.1 (not a patch release, see below for the reason).
+If the bug doesn't occur there, try to reproduce it in Godot 3.2 *beta 1* (which
+is roughly in the middle of all test builds available). If you can't reproduce
+the bug with Godot 3.2 beta 1, then try newer betas and RC builds. If you do
+manage to reproduce the bug with Godot 3.2 beta 1, then try older alpha builds.
+
+.. warning::
+
+    For bisecting regressions, don't use patch releases such as Godot 3.1.2.
+    Instead, use the minor version's first release like Godot 3.1. This is
+    because patch releases are built from a separate *stable branch*. This kind
+    of branch doesn't follow the rest of Godot's development, which is done in
+    the ``master`` branch.
+
+The Git bisect command
+----------------------
+
+If you've found a build that didn't exhibit the bug in the above testing
+process, you can now start bisecting the regression. The Git version control
+system offers a built-in command for this: ``git bisect``. This makes the
+process semi-automated as you only have to build the engine, run it and try to
+reproduce the bug.
+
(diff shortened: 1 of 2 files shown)


vnen:
> Should we have a `bisect needed` label on GitHub issues? It'd make it easier for people to find bugs that need to be bisected before they can be solved. We could then link to the list of issues labeled `bisect needed` on that page.

Bisect is usually only relevant for regressions and we already have the `regression` label for those.

community/contributing/bisect_workflow.rst
```diff
@@ -0,0 +1,131 @@
+.. _doc_bisect_workflow:
+
+Bisect workflow
+===============
+
+.. highlight:: shell
+
+After reporting a bug on the
+`Godot repository on GitHub <https://github.com/godotengine/godot>`__,
+you may be asked by a contributor to *bisect* the issue. Bisecting makes
```
aaronfranke: I think the first sentence should mention regressions, since that's the main use case for bisecting.

community/contributing/bisect_workflow.rst
```diff
@@ -0,0 +1,131 @@
...
+
+If you're using a stable release as the "good" or "bad" build, use one of the
+following commit hashes depending on the version:
+
+.. code-block:: none
+
+    3.1.2: 0587df4aa5f2977350cc80b1522cdc1e483c4515
+    3.1.1: 66baa3b633fe904ea0d90a9688d602d9f3a0b3bd
+    3.1: 320f49f204cfbf9b480fe62aaa7718afb74920a5
+    3.0.6: 8ac39d886307d76c286e804e027fc39f6b5aaac6
```
aaronfranke: If you are bisecting between 3.0 and 3.1, then it's not useful to have the hash of 3.0.6, since 3.1 and 3.0.6 are on separate branches and their common ancestor is 3.0.

If you are bisecting between 3.0 and 3.0.6, then it's also not useful to have the hash of 3.0.6, since you can just start from the head of the 3.0 branch.

So, I think this article should only mention the hashes of unpatched releases, right now that would mean only 3.0 and 3.1.

community/contributing/bisect_workflow.rst
```diff
@@ -0,0 +1,131 @@
...
+
+    git bisect bad
+
+If the build **does not** exhibit the bug, run the following command:
+
+.. code-block::
+
+    git bisect good
+
+You need to repeat this operation several times. The longer the commit range,
```
aaronfranke: Clarify "this operation" to mean building, testing, and then issuing a new bisect command.

community/contributing/bisect_workflow.rst
```diff
@@ -0,0 +1,136 @@
+.. _doc_bisect_workflow:
+
+Bisect workflow
+===============
+
+.. highlight:: shell
+
+Bisecting is a way to find regressions in software. After reporting a bug on the
+`Godot repository on GitHub <https://github.com/godotengine/godot>`__, you may
```
akien-mga: rST external hyperlinks take only one trailing underscore, not two (same for all links).

community/contributing/bisect_workflow.rst
```diff
@@ -0,0 +1,136 @@
...
+contains the pre-release you downloaded and look for the ``README.txt`` file.
+The commit hash is written inside that file.
+
+If you're using a stable release as the "good" or "bad" build, use one of the
+following commit hashes depending on the version:
+
+.. code-block:: none
+
+    3.1: 320f49f204cfbf9b480fe62aaa7718afb74920a5
+    3.0: 9bd402698cfa299b79b9144f8d7744c308a4e085
```
akien-mga: Encourage them to use tags instead (`3.2-stable`).

community/contributing/bisect_workflow.rst
```diff
@@ -0,0 +1,136 @@
...
+
+To refer to the latest state of the master branch, you can use ``HEAD``
+instead of a commit hash.
```
akien-mga: Actually I think using `master` would be better. `HEAD` is the currently checked out commit, which may not be the `master` one if they're on another branch, or if they already checked out an older commit to try to confirm a `good` before starting a bisect.

community/contributing/bisect_workflow.rst
```diff
@@ -0,0 +1,136 @@
+.. _doc_bisect_workflow:
+
+Bisect workflow
```
akien-mga: I'd call it "Bisecting regressions".

community/contributing/bisect_workflow.rst
```diff
@@ -0,0 +1,136 @@
...
+range of commits that potentially need to be built from source and tested.
+You can find binaries of official releases, as well as alphas, betas,
+and release candidates `here <https://downloads.tuxfamily.org/godotengine/>`__.
+
+For example, if you've reported a bug against Godot 3.2 RC4, you should first
+try to reproduce the bug in Godot 3.1 (not a patch release, see below for the
+reason). If the bug doesn't occur there, try to reproduce it in Godot 3.2 alpha
+1. If the bug doesn't occur there either, the regression appeared between Godot
+3.1 and 3.2 alpha 1. Otherwise, it appeared in a newer pre-release build; you
+should now test newer alphas, betas and RC builds.
```
akien-mga: I'd encourage a more bisect-y approach here, testing intermediate builds first instead of start with alpha 1 and going all the way up to RC 4.
Typically after testing 3.1, I'd test 3.2 beta 1, and then see if I go up or down depending on the result.

Calinou:
@akien-mga Can this be merged now?

akien-mga:
Thanks!

PR #3114, opened 2020-01-28
Optimize images losslessly using `oxipng -o6 --strip --zopfli`

Calinou:
There are significant gains here and there, so I suppose it's worth it. If this increases the size of Git history too much, I can amend it to only include images whose size was reduced by more than 5 KB or so.


akien-mga:
Closing for now as this needs to be redone from scratch.

PR #3119, opened 2020-01-29
Use a dark blue background on wide displays in the light theme

Calinou:
See comments in #3073 for rationale.

### Before

Before

### After

After

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -8,7 +8,7 @@
  /* Default (light) theme colors */
  :root {
     --body-color: #404040;
-    --content-wrap-background-color: #efefef;
+    --content-wrap-background-color: #29355c;
     --content-background-color: #fcfcfc;
     --logo-opacity: 1.0;
     --navbar-background-color: #333f67;


akien-mga:
Hm, I'm not sure. Using a light theme, I expect the screen to be predominantly light, and here it's 50/50. It's not bad per se but I wonder if it's not a bit too much.

ShlomiRex:
Who decides if this is good change I wonder?

groud:
I think I prefer the white background version.

Calinou:
Maybe something in-between like this?

image

I've seen many people complaining about a "broken" appearance on Discord, so I think something needs to be changed here.

groud:
The only broken thing I see is that the number of pixels needed to switch to the wide display mode is a little bit low. I'd rather keep the base mode on wider pages.

Other than that, if we really want to darken the background, I would simply go for a little bit darker white and it should be good. (at least to me ^^)

Calinou:
@groud I think it makes more sense to center the pane as soon as it reaches its maximum width. It's possible to make it only center on wide enough displays using a CSS media query, but I would prefer not to. Otherwise, the behavior would feel inconsistent.

groud:
Well on my screen I have something like 30px of background on each side. So it is kind of useless space that hurts readability. So I think stretching a little bit more outside the default width is OK.

But anyway, it is not that important ^^

FeatherAntennae:
If we put a colored background, it's common to use lighter / pastel colors so the foreground stands out and to avoid tiring the eyes due to small changes in contrasts.

The in between preview looks the best imo, but that's very subjective.

aaronfranke:
Light theme doesn't have to mean that the page is dominantly light, to me it just means that the primary part of the page should have a light background and dark text. I quite like the "After" image in the OP, I think the dark colors on the side look good. Note: I usually use the dark theme, so don't weigh my input too heavily, the light theme should first and foremost look good for light theme users.

pycbouh:
In light of a renewed interest in merging this I have to say that out of those three the current version looks to be a better option.

I agree with akien that I'd expect the most of the page to be lighter in tone if I were to use a light theme. Especially in our case, where the sidebar is really dark and that kind of gives an HDR effect of coming out of a dark tunnel when you switch from the main body to the left side to look for something. Having an overall lighter tone gives a more consistent look to the page, which I'd assume to be easier on the eyes (as easy, as the light theme can be).

The light-blue one can work, but we would need to adjust some colors with it. The sidebar is much more reddish than any other blue color of the page, including now dominant background blue. So we have to either match the sidebar color in tone or make it less red itself.

But personally, I'd go with the option C from #3140:

image

Especially, if we somehow solve the centering problem (#3139).

rodolforg:
Why those side borders, anyway...
Am I the only that prefers to read lines even on wide screens?

Or maybe I don't know wide screens lol

mhilbrunner:
> 
> 
> Why those side borders, anyway...
> Am I the only that prefers to read lines even on wide screens?
> 
> Or maybe I don't know wide screens lol

Because long lines stretching over most of a widescreen are very tiring to read.

mhilbrunner:
Note: we started a poll on the Discord (http://www.anonvote.com/strawpoll/xf29773x) on 12.05.20 and got quite some votes there. As the votes aren't increasing anymore for 24 hours now, and the result remained very stable throughout I think we can conclude:

There isn't a clear preference either way: 54 votes (43%) for dark blue, 37 votes (30%) for light blue and 34 votes (27%) for leave as-is.

As this mirrors this discussion pretty much exactly with no significant consensus for any option, I think we'll best leave it as it currently is.

Thanks for taking the time with the PR, coming up with the options and thanks for all who participated in the discussion.

PR #3121, opened 2020-01-29
Update FBX info in FAQ

skyace65:
Updates the FAQ so it mentions that FBX is supported via the Open Asset Import Library. I just copied the info from the importing 3D scenes page. Also I removed the space between the two lines about using Collada.


akien-mga:
Thanks!

PR #3122, opened 2020-01-29
Fix advanced vector math images for dark mode

### LINKED ISSUE #3120: Diagrams in Advanced Vector Math Unreadable On Dark Background
The first three images on https://docs.godotengine.org/en/3.2/tutorials/math/vectors_advanced.html are unreadable using the new documentation dark theme, which is automatically in use when your OS is set to a dark theme. This is seemingly because the .png files use a transparent background instead of an opaque one, meaning that the black text appears over a dark gray background and the white lines are much too prominent.

I've found that setting a background behind the diagram to #e6e6e6 (start from pure white and set the Value to 90%) is perceptually closest to how the images would normally appear on a white background, so it should be easy to resave them.

The bottom images also use transparency, but they're much more readable.

skyace65:
The old images had a transparent background so it was extremely hard to read them in dark mode. This PR adds a white background to all the images that need one. All images have gone through png compression. closes #3120


Calinou:
Thanks!

PR #3126, opened 2020-01-29
Update grid snap info in your_first_game.rst

beamer159:
Godot 3.2 has a button for Grid Snap:

gridsnap

I got a bit confused when I looked in the specified snapping options (three vertical dots) and this setting was not in there.


Calinou:
Thanks!

PR #3128, opened 2020-01-30
Add C# to exporting page and fix two issues

skyace65:
Adds C# code to the Step by Step exporting page. I also replaced `GetSize()` in your first game with `Size` because `GetSize()` is deprecated. And a C# line of code was out of order.


akien-mga:
Thanks!

PR #3130, opened 2020-01-31
Fix some images being difficult to read when using the dark theme

Calinou:
Follow-up to #3122.

Adding the background color `#fcfcfc` (which matches the light theme's background color) makes images readable again when using the dark theme.

Some images were also optimized losslessly using `oxipng -o6 --strip --zopfli`.


akien-mga:
Thanks!

PR #3134, opened 2020-01-31
Improve image compression recommendations in Importing images

Calinou:
This adds more information about each compression mode and improves formatting a bit.


akien-mga:
Thanks!

PR #3143, opened 2020-02-03
Update intro_to_the_editor_interface.rst

GiorgiBeriashvili:
Visual Studio is usually defined as an IDE, whereas Visual Studio Code is meant to be a code editor.

Even though Visual Studio can work as an external code editor for Godot (especially for Godot Mono), I suppose that Visual Studio Code was intended to be written instead.


Calinou:
Thanks!

PR #3147, opened 2020-02-05
Fix typos in material.rst

Dathide:
Fix two typos in material.rst


getting_started/workflow/assets/escn_exporter/material.rst
```diff
@@ -47,7 +47,7 @@ difficulties in implementation, which are:
 
 .. warning::
   If possible try to use PrincipledBSDF node with GGX distribution as the output shader
```
Calinou: While you're at it, you could add a comma there :slightly_smiling_face: 

```suggestion
  If possible, try to use PrincipledBSDF node with GGX distribution as the output shader
```

akien-mga:
Thanks!

PR #3150, opened 2020-02-06
Remove duplicated code in Sync Audio

clemens-tolboom:
Not sure I got this right but it seems like duplicate code to me.

Furthermore I'm not sure what we can use this **time** for as $player.play has no timing arguments to compensate for. There is no `$player.playDelayed(time)`.

Shouldn't we use `_physics_process` for better timing?


Calinou:
> Shouldn't we use _physics_process for better timing?

`_physics_process()` only runs 60 times per second by default, so it will probably be less accurate than `_process()`.

Calinou:
Thanks!

PR #3152, opened 2020-02-08
Grammar fix for Optimization using Servers page

Jellycream:
I noticed a small grammatical error on the Optimization using Servers page of the 3.2 docs


clayjohn:
Good change! But you've accidentally also included a commit from Akien in there. You will need to rebase your branch and only include your commit in the PR. 

How to do so is described here http://docs.godotengine.org/en/latest/community/contributing/pr_workflow.html

PR #3156, opened 2020-02-10
Possible typo

johnmick:
Tutorial for 3.2 suggests calling 'add_custom_property_editor' which is not a member of EditorInspectorPlugin; however 'add_property_editor' does exist.


Calinou:
Thanks!

PR #3161, opened 2020-02-12
Fix HDRI description in differences between GLES2 and GLES3

### LINKED ISSUE #3160: In GLES 2/3 differences, HDR paragraph does not mention GLES version
[Line 49](https://github.com/godotengine/godot-docs/blame/master/tutorials/misc/gles2_gles3_differences.rst#L49) of GLES 2/3 differences is missing the GLES version.

```
GLES is not capable of using High Dynamic Range (HDR) rendering features.
```

Looking at Issue #2205 it appears that it should state that GLES2 does not support HDR.

skyace65:
The description says "GLES doesn't support HDR" instead of "GLES2 doesn't support HDR." This PR fixes that. Closes #3160


akien-mga:
Thanks!

PR #3168, opened 2020-02-13
Fix inline code formatting that used "`" not "``".

follower:
Restructured text uses double back-ticks to indicate inline monospace code formatting (a.k.a "verbatim"). 

(via <https://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html#inline-markup-and-special-characters-e-g-bold-italic-verbatim>)

Without this change the text is rendered as italic rather than monospaced on the [docs site](https://godot.readthedocs.io/en/3.2/tutorials/networking/high_level_multiplayer.html).

PR #3169, opened 2020-02-13
Fix inline code formatting that used "`" not "``".

follower:
[Replaces #3168]

Restructured text uses double back-ticks to indicate inline monospace code formatting (a.k.a "verbatim"). 

(via <https://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html#inline-markup-and-special-characters-e-g-bold-italic-verbatim>)

Without this change the text is rendered as italic rather than monospaced on the [docs site](https://godot.readthedocs.io/en/3.2/tutorials/networking/high_level_multiplayer.html).

### DIFF
--- a/tutorials/networking/high_level_multiplayer.rst
+++ b/tutorials/networking/high_level_multiplayer.rst
@@ -118,12 +118,12 @@ Managing connections
 --------------------
 
 Some games accept connections at any time, others during the lobby phase. Godot can be requested to no longer accept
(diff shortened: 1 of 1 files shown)


follower:
Nice spotting on the accidental revert. :)

Guess I should've learned my lesson last time I tried to make multiple edits/PRs to a single file via the web interface. :D *sigh*

akien-mga:
Thanks!

PR #3173, opened 2020-02-13
Added parallel compilation to bindings compilation.

quantmint:
This makes the compilation much faster.


tutorials/plugins/gdnative/gdnative-cpp-example.rst
```diff
@@ -131,10 +131,12 @@ below.
 To generate and compile the bindings, use this command (replacing ``<platform>``
 with ``windows``, ``linux`` or ``osx`` depending on your OS):
 
+If your cpu have more that 4 threads, you should replace -j4 with -j<number of threads>.
```
Calinou:
```suggestion
To speed up compilation, add `-jN` at the end of the SCons command line where `N` is the number of CPU threads you have on your system. The example below uses 4 threads.
```

Calinou:
Thanks!

Issue #3104, opened 2020-01-24
Create a style guide for project organization


aaronfranke:
Currently there is no style guide for how to structure Godot projects outside of the contents of the scripts. This should include what is encouraged, what is allowed, and what is discouraged (and most importantly, how should the demo projects be organized).

When created, this should probably be placed in [project workflow](https://docs.godotengine.org/en/latest/getting_started/workflow/index.html).

Some topics that could be included:

Nodes:

* Casing of nodes. PascalCase? Something else? The current demo projects ~~are fairly mixed~~ EDIT: have all been refactored to PascalCase node names due to a general consensus below and also because that is how Godot cases newly created nodes by default.

* How often should nodes be grouped together? If there are 30 boxes in a scene, would it make sense to make all of them a child of "Boxes", or maybe group them like "Boxes/Group1" and include several boxes that are next to each other under the same parent?

* Should `Node` children be added with their own scripts to add behavior, like "components", or is this way of abstracting functionality discouraged?

Files/folders:

* Casing of folders. Most of the demo projects use `snake_case` for this.

* Casing of files. This one's a bit more tricky than the above:

  * What about for C# projects, where it is generally expected that the file name should match the class name? Should MyClass be in `MyClass.cs` or `my_class.cs`? If the former, should C# projects also use PascalCase for folders and other files to be consistent?

  * If the above, should GDScript files also be this way to be consistent with C# files?

* File hierarchy and organization.

  * I propose object-first, which would look like this:

```
    player/
        player.gd
        player.png
        player.tscn
    enemy/
        enemy.gd
        enemy.png
        enemy.tscn
```

* When should folders be used at all? Some demo projects have a completely flat directory structure, others (mostly the larger projects) have many subfolders.

  * How many related files should be allowed to sit next to each other without being in their own folder? If you only have `player.tscn` and `player.gd`, does that deserve a `player` folder?

  * If a project has assets from third-party sources, how should they be indicated to be so?

  * The project I am working on with my brother contains an "Assets" folder for all the stuff we make, and we still have around 30 items in the project's root folder including things like the icon, config files, helper scripts, git stuff, readme/license, C# stuff, translation files, and native libraries, most of which can't be moved to subfolders. Should an "assets" folder containing "as much user-created stuff as possible" be encouraged for large projects?

Calinou:
> Casing of nodes. PascalCase? Something else? The current demo projects are fairly mixed.

[...]

PR #3174, opened 2020-02-14
Fixed the GridMap issue in the Godot FPS Tutorial

### LINKED ISSUE #3151: FPS tutorial maps broken in 3.2
Works in 3.1.2-stable
Broken in 3.2-stable

FPS tutorial at https://docs.godotengine.org/en/3.2/tutorials/3d/fps_tutorial/part_three.html
When launching the scenes for the included levels the character falls through the floor.
The instruction that fails "Space_Level.tscn and Ruins_Level.tscn are complete custom FPS levels created for the purpose of this tutorial. Press F6 to play the open scene, or press the play current scene button, and give each a try."

Editor complains on load that the project is from a old version and upgrades it. Turns out the zip files are two years old.

There are some errors when switching to the scene (in both 3.1.2-stable and 3.2-stable):
 core/math/quick_hull.cpp:401 - Condition "O == E" is true. Continuing.
 core/math/quick_hull.cpp:401 - Condition "O == E" is true. Continuing.

An repeating error in the error log when launching Ruins_level.tscn:
E 0:00:02.308   recover_from_penetration: Condition "shape_idx < 0 || shape_idx >= cs->getNumChildShapes()" is true. Returned: false
  <C++ Source>  modules/bullet/space_bullet.cpp:1241 @ recover_from_penetration()
  <Stack Trace> Player.gd:221 @ process_movement()
                Player.gd:71 @ _physics_process()

The scene switching errors exits in 3.1.2-stable, but the recover_from_penetration is not there, and it seems to work.

Do not have a clue on how to fix this as I only just started with godotengine.

TwistedTwigleg:
The GridMaps should now work with Godot 3.2, making the tutorial doable. I didn't look at the collision shape issues, but I think it is a quick hull issue. If it is, I'm not sure on a great way to fix it without redoing all of the collision shapes.
Closes #3151


aaronfranke:
Since these are binary files, if we are changing them, we should probably fix many things at once. Some notes when I tried out the "Finished" version:

* Tons of unused argument and return value warnings which should be silenced. For example, change `delta` to `_delta`, and add `# warning-ignore:return_value_discarded`.

* The scripts are mostly complaint with the GDScript style guide, but not everywhere. I noticed that arguments with default values are missing spaces around `=`, some method calls have spaces `print (`, `Debug_Display.gd` should have two lines between `_ready` and `_process`, and `Globals.gd` shouldn't have extra newlines at the end of the file (it should have just one). There may be more style violations that I'm not seeing right now.

* The project files are not organized at all, every script and scene is stored in the project root. It would be much better if the files were organized.

* While we don't currently have a solid guide for [project organization](https://github.com/godotengine/godot-docs/issues/3104), one item of consensus so far has been to use `PascalCase` for nodes. This means no underscores in node names (which the FPS tutorial uses everywhere).

* There is no tutorial level, tutorial menu, or controls menu. While I know that this is itself a tutorial that shows users how to build it step-by-step and therefore they would know everything anyway, some users may skip straight to the final version to see what they're building before they build it, and it would also be good practice to encourage game devs to add a tutorial/help system.

akien-mga:
As I mentioned when those archives were first added, I still think it's a bad idea to have ZIP files for project code in this repo. It makes the Git repo bigger (every update increases the size of the repo by the size of the updated ZIP), and it's difficult to keep in sync with the ever changing docs and Godot API.

I believe this should be moved to a dedicated Git repo, which can then be referenced from the docs and downloaded from tagged versions matching the Godot API using the auto-generated GitHub ZIP archives.

TwistedTwigleg:
That is fair. I’m fine with moving the project to a dedicated Git repo. Right now I have it stored [here](https://github.com/TwistedTwigleg/Godot_FPS_Tutorial), though it is a little out of date and would need to be tagged according to Godot versions. We could link to that repo, or we can make a new, official repository for it.

I think that if we link to a dedicated repository, we should limit the downloads to the starter assets and final project, as maintaining the additional 4+ zip files is cumbersome, time consuming, and I’m not sure it is benefiting anyone.

mhilbrunner:
Note: this depends a little on how/if we move forward with https://github.com/godotengine/godot-docs/issues/3390

aaronfranke:
Perhaps the project should be moved to the demo projects repo?

Also, how feasible is it to maintain 6 different versions of the same project? It might make sense to only have starter assets and the final version.

TwistedTwigleg:
Yeah. I’ve glanced through the issue a few times, and it seems the general consensus is to remove tutorials like the FPS tutorial.

> how feasible is it to maintain 6 different versions of the same project? It might make sense to only have starter assets and the final version

I agree. In hindsight, having multiple versions was a bad idea and I do not think it honestly benefits anyone from a learning perspective (or the learning benefits are really minor).

NathanLovato:
I confirm we're moving towards a reference manual and moving away from step-by-step, follow-along tutorials for the official docs. We're only looking to keep dodge the creeps and have a beginner 3D series in the getting started guide.

Would you be OK to move your FPS tutorial series to your own website or another website, or repository? I'll be removing some tutorials and series from the `latest` branch of the docs soon.

And as @aaronfranke said, we're also looking to remove binary files from this repository. It'd be best to have the project in the demos repo, or maybe on a dedicated repo.

Closing this PR for now.

aaronfranke:
Yeah, we can just delete these files from the repo. They still exist in the Git history and in older branches (and in the branch this PR is based on), so it would be easy to salvage it later.

TwistedTwigleg:
> Would you be OK to move your FPS tutorial series to your own website or another website, or repository? I'll be removing some tutorials and series from the `latest` branch of the docs soon.

It's already on my website, RandomMomentania, so I guess that's fine.

> And as @aaronfranke said, we're also looking to remove binary files from this repository. It'd be best to have the project in the demos repo, or maybe on a dedicated repo.

There's a repository for it here: https://github.com/TwistedTwigleg/Godot_FPS_Tutorial

NathanLovato:
Thanks!

PR #3175, opened 2020-02-15
Added current memory issue at build time to README

Rubonnek:
I figured this could be the best place to put this notification, but feedback is welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,9 @@ make SPHINXBUILD=~/.local/bin/sphinx-build html
 ```
 
 The compilation might take some time as the `classes/` folder contains many files to parse.
+
+In case of a `MemoryError` or `EOFError`, you can remove the `classes/` folder and run `make` again. This will drop the class references from the final HTML documentation but will keep the rest intact. Make sure to avoid using `git add .` in this case when working on a pull request, or the whole `classes/` folder will be removed when you make a commit. See [#3157](https://github.com/godotengine/godot-docs/issues/3157) for more details.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -64,6 +64,9 @@ make SPHINXBUILD=~/.local/bin/sphinx-build html
 ```
 
 The compilation might take some time as the `classes/` folder contains many files to parse.
+
+In case of a `MemoryError` or `EOFError`, you can remove the `classes/` folder and run `make` again. This will drop the class references from the final HTML documentation but will keep the rest intact. See [#3157](https://github.com/godotengine/godot-docs/issues/3157) for more details.
```
Calinou:
```suggestion
In case of a `MemoryError` or `EOFError`, you can remove the `classes/` folder and run `make` again. This will drop the class references from the final HTML documentation but will keep the rest intact. Make sure to avoid using `git add .` in this case, or the whole `classes/` folder will be removed when you make a commit. See [#3157](https://github.com/godotengine/godot-docs/issues/3157) for more details.
```

Rubonnek: How about:

"Make sure to avoid using `git add .` in this case when working on a pull request, or the whole `classes/` folder will be removed when you make a commit."

It makes clear what the intent of avoiding `git add` is and doesn't feel out of context in my opinion.

Rubonnek: I've pushed a change with your suggestion and the clarification along with it.

Calinou:
Thanks!

PR #3178, opened 2020-02-16
Added docs for shader struct

Chaosus:
Docs for new feature...

### DIFF
--- a/tutorials/shading/shading_reference/shading_language.rst
+++ b/tutorials/shading/shading_reference/shading_language.rst
@@ -290,6 +290,69 @@ Global constants are useful when you want to have access to a value throughout y
 
     const float PI = 3.14159265358979323846;
 
+
+Structs
+-------
+
+Structs are compound types which can be used for better abstaction of shader code. You can declare them at the global scope like:
+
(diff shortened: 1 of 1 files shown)


tutorials/shading/shading_reference/shading_language.rst
```diff
@@ -290,6 +290,69 @@ Global constants are useful when you want to have access to a value throughout y
...
     const float PI = 3.14159265358979323846;
 
+
+Structs
+-------
+
+Structs are compound types which can be used for better abstaction of shader code. You can declare them at the global scope like:
+
+.. code-block:: glsl
+    struct PointLight {
```
navett52: This document is about Godot's shading language. I don't seem to be able to use structs within it. Is this written with C# in mind? I know structs can be used there.

clayjohn: This document is only about the shader language. Structs were recently added to shaders in master. You will have to compile master yourself to use them.

navett52: Ahh ok, that makes sense. Thanks for the information.

PR #3180, opened 2020-02-16
Fix highlighting of XML tags when using a dark theme

Calinou:
This sets the color of XML tags on pages like [Contribute to the Class Reference](https://docs.godotengine.org/en/3.2/community/contributing/updating_the_class_reference.html).


akien-mga:
Thanks!

PR #3183, opened 2020-02-17
Use more standard terminology. Fix grammar.

follower:
Change to use node relationship terminology ("sibling") that is both more standard[0] & more inclusive.

Fix pluralisation (i.e. "Xs nodes" -> "X nodes").

[0] e.g. https://en.wikipedia.org/wiki/Tree_structure#Terminology_and_properties


akien-mga:
Thanks!

PR #3187, opened 2020-02-18
Fix code samples for Matrices and Transforms

aaronfranke:
Now that [Matrices and Transforms](https://docs.godotengine.org/en/latest/tutorials/math/matrices_and_transforms.html) is live, I can see that I made a mistake. The code tabs expect one blank line for some reason, so right now the first line of each code sample isn't displayed.

I also forgot to use `Transform2D.Identity`  instead of `new Transform2D()` for C#, whoops. Some of the code samples work with `new Transform2D()` but it's not ideal since it initializes everything to zero.


akien-mga:
Thanks!

PR #3196, opened 2020-02-21
Update 2d_sprite_animation.rst

saschell:
previous code was non functional

### DIFF
--- a/tutorials/2d/2d_sprite_animation.rst
+++ b/tutorials/2d/2d_sprite_animation.rst
@@ -159,9 +159,9 @@ released.
     extends KinematicBody2D
 
     func _process(delta):
-        if Input.is_action_pressed("ui_right"):
+        if Input.is_action_just_pressed("ui_right"):
             $AnimationPlayer.play("walk")
-        else:
+        elif Input.is_action_just_released("ui_right"):
             $AnimationPlayer.stop()
 
 


cbscribe:
Code works fine here as written.

saschell:
Im on version 3.0.6.stable on debian 10 (buster). And the code does not work for me. Care to share a git repo of your working version so I can test it?

Maybe I did something else wrong, thanks in advance!

cbscribe:
Tested on current stable (3.2) as well as 3.1.2. I don't have a copy of 3.0.x handy to test on, but as I recall, in older versions, calling `AnimationPlayer.play()` on a playing animation would restart the animation.

This tutorial was written at the time of 3.1's release, so the code should not be changed as it works with the current and previous releases. Docs should reflect current behavior.

akien-mga:
If you're learning Godot, I would strongly suggest to use the latest stable version (which you can download from https://godotengine.org/download).
Might also be worth check if there's already a Debian bug report tracking a request to upgrade to Godot 3.2.

The 3.0 branch is from 2018 and is no longer supported.

As a Linux packager I know the inertia we can have when packaging things, and it's amazing enough that Godot finally ended up in Debian after 4 years, but if it stays outdated it won't serve much the Debian users.

akien-mga:
Note that if you want to learn with 3.0 nevertheless, you should likely use the 3.0 branch of the docs: https://docs.godotengine.org/en/3.0/

It has less content than later branches but at least what is there should be working with 3.0.6 (if not, it can be fixed in the `3.0` branch of this repo).

PR #3197, opened 2020-02-21
Add instant.page to speed up page loading

Calinou:
instant.page prefetches pages when the user hovers links for more than 65 milliseconds. This can be used to make navigation feel significantly faster.

Note that this will have no effect in Firefox when using uBlock Origin, as it forcibly disables prefetching.


mhilbrunner:
Thanks, tested this on a webserver and it did seem to help a little. :) Didn't measure exact times however.

PR #3199, opened 2020-02-23
Minor Documentation Clarifications

mark-spiezio:
Clarify PEM acronym.  
The "public key" goes to the .crt file, not "public pair".


Calinou:
Thanks! (I amended changes to fix rST formatting as the link syntax was invalid.)

PR #3200, opened 2020-02-24
Updated c# function prototype in the example code.

brentbartlett:
It needs to be public however the example shows it as private.


brentbartlett:
Small change to C# documentation for accuracy.

Calinou:
Thanks!

PR #3201, opened 2020-02-25
Unity has an official editor for Linux

matkoniecz:
See https://docs.unity3d.com/Manual/InstallingUnityLinux.html https://itsfoss.com/unity-editor-linux/

### DIFF
--- a/getting_started/editor/unity_to_godot.rst
+++ b/getting_started/editor/unity_to_godot.rst
@@ -22,7 +22,7 @@ Differences
 +===================+====================================================================================+================================================================================================================+
 | License           | Proprietary, closed, free license with revenue caps and usage restrictions         | MIT license, free and fully open source without any restriction                                                |
 +-------------------+------------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------+
-| OS (editor)       | Windows, macOS, Linux | Windows, macOS, X11 (Linux, \*BSD)                                                                             |
+| OS (editor)       | Windows, macOS, Linux                                                              | Windows, macOS, X11 (Linux, \*BSD)                                                                             |
 +-------------------+------------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------+
 | OS (export)       | * **Desktop:** Windows, macOS, Linux                                               | * **Desktop:** Windows, macOS, X11                                                                             |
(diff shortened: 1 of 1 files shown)


matkoniecz:
BTW, UX of the table is poor - see https://docs.godotengine.org/en/3.2/getting_started/editor/unity_to_godot.html - at least for me only the Unity column is visible

Calinou:
> BTW, UX of the table is poor - see https://docs.godotengine.org/en/3.2/getting_started/editor/unity_to_godot.html - at least for me only the Unity column is visible

I'm not sure if anything can be done about this. The table would either have to be collapsed (with very narrow columns) or have horizontal overflow (i.e. scrolling).

Maybe it should be redone entirely so tables aren't used?

getting_started/editor/unity_to_godot.rst
```diff
@@ -22,7 +22,7 @@ Differences
 +===================+====================================================================================+================================================================================================================+
 | License           | Proprietary, closed, free license with revenue caps and usage restrictions         | MIT license, free and fully open source without any restriction                                                |
 +-------------------+------------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------+
-| OS (editor)       | Windows, macOS, Linux (unofficial and unsupported)                                 | Windows, macOS, X11 (Linux, \*BSD)                                                                             |
+| OS (editor)       | Windows, macOS, Linux | Windows, macOS, X11 (Linux, \*BSD)                                                                             |
```
Calinou: Fix column alignment:

```suggestion
| OS (editor)       | Windows, macOS, Linux                                                              | Windows, macOS, X11 (Linux, \*BSD)                                                                             |
```

This doesn't affect the table's final look, but it makes the markup look better.

matkoniecz:
> I'm not sure if much can be done about this. 

This page is displayed in small column, most of my screen is unused. 

Looks like something tested only on mobile devices.

But it is not something fixable by changing documentation text or formatting.

Calinou:
> This page is displayed in small column, most of my screen is unused.

The documentation main page is intentally kept narrow enough so that text is easier to read. (Lines are more readable when they're between 45 and 80 characters.)

Unfortunately, we can't make an exception for a single page due to Read the Docs' limitations.

Calinou:
Thanks!

PR #3207, opened 2020-02-26
Set the recommended distribution for "portable" builds to Ubuntu 18.04

Calinou:
Godot 4.0 requires a C++17-compliant compiler, which is only available in Ubuntu 18.04 and later.

While it's possible to build it on Ubuntu 16.04 with third-party repositories, Ubuntu 18.04 has everything required in the official repositories, making it easier to set up.

This was suggested by Mistery Man on Discord.

### DIFF
--- a/development/compiling/compiling_for_x11.rst
+++ b/development/compiling/compiling_for_x11.rst
@@ -120,7 +120,7 @@ Building export templates
 .. warning:: Linux binaries usually won't run on distributions that are
              older than the distribution they were built on. If you wish to
              distribute binaries that work on most distributions,
-             you should build them on an old distribution such as Ubuntu 16.04.
+             you should build them on an old distribution such as Ubuntu 18.04.
              You can use a virtual machine or a container to set up a suitable
              build environment.
 


development/compiling/compiling_for_x11.rst
```diff
@@ -120,7 +120,7 @@ Building export templates
 .. warning:: Linux binaries usually won't run on distributions that are
              older than the distribution they were built on. If you wish to
              distribute binaries that work on most distributions,
-             you should build them on an old distribution such as Ubuntu 16.04.
+             you should build them on an old distribution such as Ubuntu 18.04.
```
clayjohn: Maybe just say "you should build them on Ubuntu 18.04"

Calinou: Other old distributions such as CentOS can be a valid choice, but I agree Ubuntu LTS is the most relevant choice when it comes to building portable games.

clayjohn: My comment was specifically referring to it as "an old distribution"

aaronfranke: I think it's fine to start writing documentation like this. That statement will become true shortly; in about 2 months, 20.04 LTS will make 18.04 LTS an old Ubuntu LTS.

akien-mga:
18.04 is too recent, that won't be portable enough. It also can't be qualified as "old" when it's the latest LTS version of Ubuntu and less than 2 years old.

SteamOS still uses glibc 2.19 (Ubuntu 14.04 LTS).

Official builds are made on Ubuntu 14.04 with GCC 8: https://github.com/godotengine/build-containers/blob/master/Dockerfile.ubuntu-64

Calinou:
@akien-mga I suppose we should wait before merging this (at least until 4.0 is in beta/RC state). By that time, Ubuntu 20.04 will be released. That is, unless Steam forbids distributing binaries built on Ubuntu 18.04.

mhilbrunner:
This is conflicting now.

Maybe we should remove the mention of any specific distribution and version?

IMO, if you're going to distribute your build, you want to make a deliberate decision about this, in which case we don't really need to suggest any example.

We could point out more directly that you want to make sure you support e.g. the Glibc versions available to your target audience.

Pointing out "use Ubuntu 16.04" (or 18.04) is IMO less useful than the *why*.

Calinou:
> Pointing out "use Ubuntu 16.04" (or 18.04) is IMO less useful than the why.

I agree, but I think we should still recommend a common baseline. Otherwise, people are going to spend a while choosing between Ubuntu, Debian, CentOS and the like for little benefit.

akien-mga:
I still think that anyone aiming to do really portable Linux binaries should use 16.04 or older (we still use 14.04 for the official buildsystem).

aaronfranke:
> @Calinou While it's possible to build it on Ubuntu 16.04 with third-party repositories, Ubuntu 18.04 has everything required in the official repositories, making it easier to set up.

How does one set up a build environment on Ubuntu 14.04 or 16.04? Is this documented somewhere?

Also, Ubuntu 14.04 is going EOL at the beginning of 2022. Even if Godot 4 comes out before then, it won't be by much. It would likely make sense to at least change the recommended version to 16.04 so it's in support.

Calinou:
> Also, Ubuntu 14.04 is going EOL at the beginning of 2022.

Are you referring to some kind of extended support? Ubuntu 14.04 has been EOL since April 2019.

aaronfranke:
> For the general public, Ubuntu 14.04 has been EOL since April 2019.

Yes-ish. Anyone can get the extended updates for free [here](https://ubuntu.com/advantage). Says for personal use though, so I don't know if it applies to Godot. It's possible that Godot could convince Canonical to give a free license. If not it's $25/year for 1 desktop.

PR #3210, opened 2020-02-27
Add information on compiling Godot on Linux using Clang + LLD

### LINKED ISSUE #2614: Add to docs info about clang and LLVM packages
LLVM and clang packages should be added to all os, but for now it is visible only in OpenBSD - https://godot.readthedocs.io/en/latest/development/compiling/compiling_for_x11.html

Calinou:
This also mentions link-time optimization as in the Compiling for Windows page.

This closes #2614.

### DIFF
--- a/development/compiling/compiling_for_x11.rst
+++ b/development/compiling/compiling_for_x11.rst
@@ -114,6 +114,11 @@ manager.
           make the final executable smaller and faster by adding the
           SCons option ``target=release_debug``.
 
+          If you are compiling Godot with GCC, you can make the binary
+          even smaller and faster by adding the SCons option ``use_lto=yes``.
+          As link-time optimization is a memory-intensive process,
+          this will require about 3 GB of available RAM while compiling.
+
 Building export templates
 -------------------------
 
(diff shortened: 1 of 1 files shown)


development/compiling/compiling_for_x11.rst
```diff
@@ -169,3 +174,21 @@ here:
...
+
+Using Clang and LLD for faster development
+------------------------------------------
+
+You can also use Clang and LLD to build Godot. This has two upsides compared to
+the default GCC + GNU ld setup:
+
+- LLD links Godot significantly faster compared to GNU ld or gold. This leads to
+  faster iteration times.
+- Clang tends to give more useful error messages compared to GCC.
```
akien-mga: That's quite political :)

GCC warnings have improved a ton since GCC 6 and in my experience both GCC and Clang provide great warnings, sometimes on different codes. I wouldn't say that any is clearly better than the other in this regard.

Calinou: True, I based myself on how the situation was in 2016 or so. It's probably better now and we can remove that line.

Issue #2453, opened 2019-05-13
Translation app and format


youreperfect:
The translation pages recommend using MS Excel (or at least give it as an example) but at least Excel 2013 has no easy/obvious way to set the export format and defaults to ANSI at least on my system. The translation pages _do_ mention that the format must be UTF-8 but only in passing and this is easily missed leading to cryptic, unicode parse errors when importing.

I recommend making the part about requiring UTF-8 format highly visible. Bonus if you give steps to convert non-UTF-8 files to UTF-8 for different operating systems.

PR #3211, opened 2020-02-27
Clarify CSV encoding requirements when importing translations

Calinou:
Since Excel has no easy way to save files to UTF-8, this adds a warning that recommends using LibreOffice or Google Sheets instead.

This closes https://github.com/godotengine/godot-docs/issues/2453.


akien-mga:
Thanks!

PR #3214, opened 2020-02-28
Small typo fix: 2-bit changed to 1-bit

moscoso:
2-bit images support 4 colors.

Technically, a 1-bit image supports 2 colors: black & white.


Calinou:
Thanks!

PR #3216, opened 2020-02-29
replace 'Linux or other Unix Variants' with macOS

sambler:
The page about building for macOS refers to "Linux and other Unix variants". Change this to macOS.

Two questions :-

The docs use macOS while the build system refers to it as osx and the relevant filenames use osx. Is there any interest in changing the docs to use osx?

Python 2.7 has reached end of life. Is there interest in removing references to py27 from the docs? Use Python 3.5+?

### DIFF
--- a/development/compiling/compiling_for_osx.rst
+++ b/development/compiling/compiling_for_osx.rst
@@ -8,8 +8,7 @@ Compiling for macOS
 Requirements
 ------------
 
-For compiling under Linux or other Unix variants, the following is
-required:
+For compiling under macOS, the following is required:
 
 -  Python 2.7+ or Python 3.5+
 -  SCons build system


neikeq:
I think macOS is the new name, so if we were to change something it would be to that.

Calinou:
> Python 2.7 has reached end of life. Is there interest in removing references to py27 from the docs? Use Python 3.5+?

We still support Python 2.7 in the buildsystem right now but will probably start requiring Python 3.6+ soon.

sambler:
@neikeq Technically macOS was used for from v7.6 to v9. From v10 it was Mac OS X then was OSX and has gone back to mac OS X the last few years. macOS and Mac OSX come from two very different code bases.

While macOS is recognisable, either "OS X" or "mac OS X" would be the correct use. I know it's being picky but it is the way Apple has named its product. Just wondering if there is any interest in changing it.

madmiraal:
According to [Wikepedia](https://en.wikipedia.org/wiki/MacOS#macOS) macOS has been used since 2016.

OS X was the old name used from 2012. Mac OS (with a space and a capital 'M') was the even older name. Mac OS X was simply the 10th version of Mac OS. Since then all their OS versions have been sub versions of 10 with the current version being 10.15 a.k.a. macOS Catalina.

In short, macOS is the current correct name.

Calinou:
Thanks!

akien-mga:
> > Python 2.7 has reached end of life. Is there interest in removing references to py27 from the docs? Use Python 3.5+?
> 
> We still support Python 2.7 in the buildsystem right now. In 4.0, we will probably start requiring Python 3.6+ though.

We'll drop support for Python 2.7 in Godot 4.0, so this could be changed now in the `master` branch yes.

PR #3219, opened 2020-02-29
Add a page on installing plugins

### LINKED ISSUE #2491: Add information about how to use editor plugins
While there are multiple sites that describe how to build a plugin I couldn't find a simple description how to install and use one.

Screen Shot 2019-06-01 at 08 57 05

This guide would probably go like this:

- find the plugin you want
- make sure you have a directory "addons" in your resources
- place the plugin code in that directory
- enable the plugin via the project settings

Calinou:
This closes #2491.


tutorials/plugins/editor/installing_plugins.rst
```diff
@@ -0,0 +1,70 @@
...
+On GitHub, if the plugin doesn't have any *tags* declared, use the **Download ZIP**
+button to download a ZIP of the latest revision:
+
+.. image:: img/installing_plugins_github_download_zip.png
+
+Extract the ZIP archive and move the ``addons/`` folder it contains into your
+project folder. If your project already contains an ``addons/`` folder, move the
+plugin's ``addons/`` folder into your project folder to merge the new folder
+contents with the existing one. Your file manager may ask you whether to write
+into the folder; answer **Yes**. No files will be overwritten in the process.
```
akien-mga: I assume some plugins might not respect the `addons/name/` structure and might be the plugin folder directly at the root of the repo. In this way they're not installable from the asset lib, but it might be good to describe how to install them in this case.

Maybe without going too much into the technical details specific to each file manager, you could describe that the installed plugin in the project folder should be so that `plugin.cfg` is in a subfolder of `addons/`.

akien-mga: I'll merge already though, can be improved in a follow up PR.

akien-mga:
Thanks!

tutorials/plugins/editor/installing_plugins.rst
akien-mga: I'll merge already though, can be improved in a follow up PR.

PR #3220, opened 2020-02-29
C# bullet documentation

LainYC:
Unless I am mistaken, you cannot directly call the hit method, and must use .Call(). Seems to not even be capable of compiling, not sure how much these C# sections are tested?
I also presume there's not much reason to shorten collision to collsion, so I threw that change in as well.


Calinou:
Thanks!

PR #3223, opened 2020-03-01
Refer to dropdown rather than down arrow

vidbina:
The use of the descriptor " down arrow" is somewhat less clear where the "Physics
Material Override" dropdown is arguably less susceptible to misunderstanding.


Calinou:
Thanks!

PR #3224, opened 2020-03-01
Rephrase dock descriptions to be location-agnostic

vidbina:
On my setup the Scene dock was not on the right side but rather on the
left side and when changes are made to the UI, it will be challenging to
keep track of all references in the documentation that may need
updating.

Keeping the documentation loosely coupled to the actual location of
elements in the UI should avoid confusion for new users and minimize
work for maintainers.


Calinou:
Thanks!

PR #3226, opened 2020-03-01
Remove the screenshot of the documentation in the networking tutorial

Jummit:
Also mention some more multiplayer peers, since there are more than only the Enet now.


akien-mga:
Force pushed a rebase.

akien-mga:
Thanks!

PR #3234, opened 2020-03-03
Improve the Compiling for macOS documentation

Calinou:
- Mention Homebrew to install dependencies easily.
- Remove the part about building fat binaries as support for 32-bit macOS was removed in Godot 3.1.
- Improve OSXCross installation instructions.
  - Mention the WSL as an alternative on Windows instead of Cygwin.
  - Change the suggested OSXCross path to something that will work out of the box.


akien-mga:
Thanks!

PR #3007, opened 2019-12-27
Add a scrollbar to the navigation bar

xelivous:
Fixes #1745
Also gives the scrollbar a basic style

*Bugsquad edit (keywords for easier searching): scroll*

PR #3237, opened 2020-03-03
Add scrollbar to the navigation bar (salvaged from xelivous/godot-docs)

### LINKED ISSUE #1745: Enhancement suggestion: scrollbar in left side navigation of the docs website.
Is it possible to show the scrollbar for the left navigation menu? I sometimes find the mouse wheel is unsatisfactory for navigating up and down.

I'd also keep the logo and search input fixed at the top for convenience, but meh.

pycbouh:
Original PR #3007 by @xelivous, fixes #1745.

Rebased against master, as asked by @Calinou, and added support for the dark theme that was originally missing.

Looks like this:
**Light theme**
image

**Dark theme**
image

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -21,6 +21,7 @@
     --navbar-level-2-color: #b8d6f0;
     --navbar-level-3-color: #a3c4e1;
     --navbar-heading-color: #ff7381;
+    --navbar-scrollbar-background: #131e2b;
 
     --link-color: #2980b9;
     --link-color-hover: #3091d1;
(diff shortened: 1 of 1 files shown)


Calinou:
Thanks!

pycbouh:
@Calinou So this needs to be reimplemented then?

Calinou:
@pycbouh Yes, but I don't know how we can make both the scrollbar and the fixed search bar work at the same time.

akien-mga:
Can we make it so that the scrollable area only starts with the table of contents, and not the whole side panel?

pycbouh:
@akien-mga I've made just that in the linked second attempt #3254, but @Calinou has pointed out, that in that case we don't hide the logo as we scroll the main body. I haven't got a chance to look into possible solutions for that.

PR #3240, opened 2020-03-04
Add Asset Library Submission Guidelines

aaronfranke:
This PR contains a handful of requirements and recommendations that I've come up with so far. Formalizing these will help both users and moderators with the submission process.

I've placed them in the "Submitting to the Asset Library" document, above the "Submitting" section, since it's best for users to check if their asset meets the requirements before they begin submitting.

It might be worth linking to this page from the "Submit Asset" form on the asset library.

I also added a little bit to the "About the Asset Library" page. It's not really related to the above, but it's something that should be included as well, so I just put it in here.

### DIFF
--- a/tutorials/assetlib/uploading_to_assetlib.rst
+++ b/tutorials/assetlib/uploading_to_assetlib.rst
@@ -13,6 +13,67 @@ As mentioned in the :ref:`doc_using_assetlib` document, in order to be able to
 submit assets to the AssetLib, you need to have a registered account, and be
 logged in.
 
+Submission guidelines
+---------------------
+
+Before submitting your asset, please ensure it follows all of the
+requirements, and also consider following the recommendations.
+
+Requirements
+~~~~~~~~~~~~
+
+Generally speaking, most assets people submit to the asset library
+are accepted. However, in order for your asset to be accepted, there
+are a few requirements your asset needs to meet to be approved.
+
+* The asset must work. If the asset doesn't run or otherwise doesn't
+  work in the specified Godot version, then it will be rejected.
+
+* No submodules, or any submodules must be non-essential. GitHub
+  does not include submodules in the downloaded ZIP file, so if the
+  asset needs the contents of the submodule, your asset won't work.
+
+* The license needs to be correct. The license listed on the asset
+  library must match the license in the repository.
+
+* Do not set the download commit to "master". The asset library validates
+  all assets with a SHA256 hash, so the version hosted on GitHub needs
+  to be *exactly* the same. Instead of "master", either specify a commit
+  hash, or tag versions and specify a version number.
+
+* Use proper English for the name and description of your asset.
+  This includes using correct capitalization, and using full
+  sentences in the description.
+
+* The icon link must be a direct link. For icons hosted on GitHub, the
+  link must start with "raw.githubusercontent.com", not "github.com".
+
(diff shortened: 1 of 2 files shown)


tutorials/assetlib/what_is_assetlib.rst
```diff
@@ -25,6 +25,20 @@ Please note that the AssetLib is relatively young - it may have various pain poi
...
 where you can submit pull requests and issues, so please do not hesitate to visit it!
 
+Types of assets
+---------------
+
+Be aware that there are, broadly, two different types of assets you can post.
+
+* Assets labeled as "Templates" or "Projects" appear under the "Templates"
+  tab in the Godot project manager. These assets are standalone Godot
+  projects that should be able to run. ("Demos" also show up here, but
```
Calinou: I think anyone can upload an asset in the "Demos" category. Only select assets will have the "Official" support level though.

tutorials/assetlib/uploading_to_assetlib.rst
```diff
@@ -13,6 +13,67 @@ As mentioned in the :ref:`doc_using_assetlib` document, in order to be able to
...
+
+* If your asset is a library for working with other files,
+  consider including example files in the asset.
+
+* The icon should be a square, its aspect ratio should be 1:1. It should
+  also ideally have a minimum resolution of 64x64 pixels.
+
+* While the asset library allows more than just GitHub, consider
+  hosting your asset's source code on GitHub. Other services may not
+  work as well, and a lack of familiarity can be a barrier to contributors.
```
Xrayez:
```suggestion
  work reliably, and a lack of familiarity can be a barrier to contributors.
```

I'm not sure whether "as well" is used as "too" or "as good" here.

pycbouh:
Can we recommend adding a `.gitattributes` file to have cleaner asset installs? Something like
```
/.gitattributes     export-ignore
/.gitignore         export-ignore
/LICENSE            export-ignore
/README.md          export-ignore
/project.godot      export-ignore
```

Calinou:
@pycbouh That makes sense, though we need to tell asset authors to make a second copy of the license in their add-on folder. Otherwise, the asset could become detached from its license if you commit it to version control (which is common to avoid using submodules).

pycbouh:
@Calinou Worth adding this to recommendation as well, then. This can be solved in future, by actually using repository root as `res://addon/addon_name` root, so that everything that installs is already "sandboxed".

aaronfranke:
@pycbouh See https://github.com/godotengine/godot-proposals/issues/554, I think it would be better if assets could be extracted into subfolders. I think it's best to preserve the README/etc files, but we just don't want them in the root of the project folder.

But also, some assets are indeed full projects, so we can't ignore the `project.godot` file for them.

pycbouh:
@aaronfranke Obviously, there are different kinds of addons and this blacklisting thing should be noted in docs with a remark.

Yet, I feel like it should be noted nonetheless, as whitelisting/blacklisting files is a common question when publishing modules, which addons essentially are. And currently this is the only way for plugin developers to avoid polluting end-users' projects. Your proposal obviously supersedes this suggestion, but it's not something we can change short-term.

aaronfranke:
Adding such a note can be done in another PR, if desired. In the meantime, I'll merge this.

PR #3252, opened 2020-03-07
Your first game: play mob animation

bwasty:
The mob animations didn't play for me (using 3.2 on macOS) - this fixed it.


cbscribe:
This change is unnecessary. The directions tell you to:

No need for additional code.

Gr0minet:
Hello @cbscribe, I encountered the same issue as @bwasty. Actually I disabled the "Playing" property because the animation was playing even when the scene wasn't running, and it was annoying. At the time, I didn't understand that it would play the animation during the scene (before I read your message here).
Maybe a line of text could be added to clarify this in the doc? Just a suggestion.

The rest of the tutorial is awesome by the way :)

PR #3254, opened 2020-03-07
Add a scroll to navbar, second attempt

pycbouh:
Another attempt to solve scrollbar in the navigation bar. Previous PR: #3237.

@Calinou I've tried to patch it to keep search bar fixed and to appear normal on mobile. Can you check, if this is better? It's not my finest CSS, but it seems to be working for me.

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -21,6 +21,7 @@
     --navbar-level-2-color: #b8d6f0;
     --navbar-level-3-color: #a3c4e1;
     --navbar-heading-color: #ff7381;
+    --navbar-scrollbar-background: #131e2b;
 
     --link-color: #2980b9;
     --link-color-hover: #3091d1;
@@ -100,6 +101,7 @@
         --navbar-level-2-color: #ccc;
         --navbar-level-3-color: #bbb;
         --navbar-heading-color: #ee7381;
+        --navbar-scrollbar-background: #1c1e21;
 
         --link-color: #8cf;
         --link-color-hover: #9df;
@@ -602,7 +604,7 @@ kbd, .kbd {
 @media only screen and (min-width: 769px) {
     /* Simulate a drop shadow that only affects the bottom edge */
     /* This is used to indicate the search bar is fixed */
-    .wy-side-nav-search.fixed::after {
+    .wy-side-nav-search.fixed-and-scrolled::after {
         content: '';
         position: absolute;
         left: 0;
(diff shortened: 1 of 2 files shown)


Calinou:
I tested this locally. Unfortunately, on desktop, the logo will never hide, which means lots of vertical real estate is wasted. This makes it difficult to browse the navigation bar on smaller displays. Here's a screenshot in 1366×768 for reference:

Screen Shot 2020-03-10 at 23 06 41

akien-mga:
What's the status on this PR?

pycbouh:
@akien-mga Still WIP.

pycbouh:
@Calinou I've looked into the logo situation. For some reason it was hidden based on the sidebar scroll. I changed it to react to the main scroll instead, as it should. The logo now hides, the navigation bar adjusts. Can you give it another look on your end?

pycbouh:
I guess I've realized why the logo is hiding with the sidebar scroll in the live version. Currently the whole left side scrolls. So it works in a way that if you scroll left side we hide the logo to give the navigation more space. As I've changed the scroll on the left to only be present inside the navigation section, it stopped making sense to work like that. It's more common, I think, to fix the sidebar in place as the main scroll is happening.

So this PR changes the current logic a bit, so I figured I'd mention it. Maybe someone would find it less usable now? If an article is short and there is no main scroll, the left side will have the logo visible always. Should I look into other solutions?

Calinou:
> Edit: It also does not feel as smooth. We can refine it more, probably.

We can probably solve this using a CSS transition in a future PR.

pycbouh:
Actually, I think I have a better working solution all together.

pycbouh:
I pushed a second commit, just so it reverts better just in case. The implementation more heavy on JS now, but the logo scrolls smoothly and the sidebar adjusts along with it. Mobile is unharmed as far as I can see, and I clean up after myself if we switch to it mid-page.

If it's better now, I can squash them any time.

pycbouh:
Perfect! Squashed everything together.

akien-mga:
Thanks!

Calinou:
I noticed there's an horizontal scrollbar appearing on the live version:

image

We should probably hide it, but keep the vertical scrollbar. Could you take a look into it if you have some time, @pycbouh? Thanks in advance :slightly_smiling_face:

pycbouh:
Of course. There is also a banner in the live version I didn't account for:
image

PR #3258, opened 2020-03-10
Adjust code sample to use new offset syntax

sambler:
The existing tutorial code uses `set_offset(randi())` - change this to `offset = randi()` to match the 3.2 syntax change.


Calinou:
Thanks!

PR #3267, opened 2020-03-13
Improved CSS media queries to remove conflicts

### LINKED ISSUE #3186: New website is very difficult to use in portrait mode on tablets
I find browsing the docs on my iPad in portrait mode impossible, as the sidebar covers up half of the page content, and I can't scroll horizontally.

Screen Shot 2020-02-17 at 16 42 49

pycbouh:
Fixes #3186.

File `theme.css` uses media queries with `max-width: 768px` to define "mobile" styles. Which makes default styles to apply to every screen and viewport as wide as 769px or wider.

Our `custom.css` implements rules in reverse. Default styles are applied to smaller screens, while media queries override them for wider screens. The query used is `min-width: 768px`. This makes the two files to have conflicting rules at exactly 768px wide mark.

This PR changes queries in `custom.css` and `custom.js` to only start applying from 769px.

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -256,7 +256,7 @@ footer,
     background-color: var(--content-wrap-background-color);
 }
 
-@media only screen and (min-width: 768px) {
+@media only screen and (min-width: 769px) {
     .wy-body-for-nav {
         /* Center the page on wide displays for better readability */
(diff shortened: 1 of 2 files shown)


akien-mga:
Thanks!

PR #3272, opened 2020-03-15
Include x86-64 option in adding support for Android x86 devices.

madmiraal:
This patch adds the instructions for including the x86-64 architecture in the Compiling for Android, Adding Support for x86 devices [section](https://docs.godotengine.org/en/stable/development/compiling/compiling_for_android.html#adding-support-for-x86-devices).


Calinou:
Thanks!

PR #3275, opened 2020-03-16
Use consistant case for Tween

sambler:
Change the first lowercase `tween` to a capitalised `Tween` to match previous code.

The capitalised Tween is a class name and the lowercase tween is a variable name.

While this works, the initial presentation of the call uses the class name for both interpolation options.

This is being picky but is accurate.


Calinou:
Thanks!

PR #3276, opened 2020-03-16
Install sphinx_tabs using pip requirements.txt + rename folder to _extensions

akien-mga:


akien-mga:
I might have merged a bit fast, before I realized that this means users should install sphinx_tabs locally to build the docs. I'll make a docs PR.

akien-mga:
@Calinou It looks like our CSS might need adjustments after this version upgrade.

Before (`stable`):
Screenshot_20200316_141545

After (`latest`):
Screenshot_20200316_141649

PR #3277, opened 2020-03-16
Use new full-page search UI from upstream RTD

akien-mga:
From https://readthedocs-sphinx-search.readthedocs.io,
can also be tested on https://docs.readthedocs.io.

Screenshot_20200316_134641

It only works on RTD so I haven't confirmed yet that it's working, the build resulting from merging this PR will be the test in production.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,5 @@
 # Code tabs for GDScript/C#
 sphinx-tabs
+
+# Full-page search UI for RTD: https://readthedocs-sphinx-search.readthedocs.io
+git+https://github.com/readthedocs/readthedocs-sphinx-search@master


pycbouh:
Probably needs dark theme adjustments, right?

akien-mga:
> Probably needs dark theme adjustments, right?

That's likely yeah. I'll merge so that we can test, and then we'll have to see if we can adjust the theme via our CSS overrides or if we need to actual vendor the extension to edit its source.

akien-mga:
It doesn't think to work yet on https://docs.godotengine.org/en/latest/, not sure if I missed anything.

pycbouh:
Don't you need to enable it?

image

akien-mga:
Ah, I forgot to commit my `conf.py` changes :)

akien-mga:
Should be fixed with c85c68e6945216266135660159b84f4731952fdf.

pycbouh:
image

Yep, seems to be working

PR #3278, opened 2020-03-16
Document installation of requirements.txt with pip

akien-mga:
Follow-up to #3276.


Calinou:
Shouldn't we pin the versions in `requirements.txt`? This way, all contributors will have the same version installed locally.

akien-mga:
> Shouldn't we pin the versions in `requirements.txt`? This way, all contributors will have the same version installed locally.

We could, but I figured that it would force us to bump version every time there's a new sphinx, sphinx_rtd_theme, etc., while in general we should be fine with the latest. The main thing for which pinning the versions might make sense would be for CSS/template contributors I guess.

akien-mga:
Actually RTD doesn't support Sphinx 2 yet so we should probably pin that at least.

akien-mga:
Updated with versions pinned.

PR #3281, opened 2020-03-17
Update Transform and Basis value descriptions for binary serialization

### LINKED ISSUE #3264: Update binary serialization docs
Hi,

It's possible to assign name to Transform2d / Basis / Transform values ?
Like Vector2 / Plane / Quad 

http://docs.godotengine.org/en/latest/tutorials/misc/binary_serialization_api.html#transform2d

Thx :)

aaronfranke:
Closes #3264.

Also fix incorrect documentation about int and float being 32-bit when they are actually 64-bit due to Variant. These docs still need to be updated further for Godot 4.0.


tutorials/misc/binary_serialization_api.rst
```diff
@@ -100,17 +103,17 @@ of packet:
 +----------+-------+-----------+--------------------------+
 | Offset   | Len   | Type      | Description              |
 +==========+=======+===========+==========================+
-| 4        | 4     | Integer   | Signed, 32-Bit Integer   |
```
evandikt: It's possible to be 32 bit if it's godot 32 bit no ?

evandikt: https://github.com/godotengine/godot/blob/master/core/io/marshalls.cpp#L134

aaronfranke: I'm not completely sure, but I know that the current documentation is wrong, there's [this](https://github.com/godotengine/godot/blob/master/core/variant.h#L205), and it will definitely always be 64-bit after https://github.com/godotengine/godot/pull/21922

aaronfranke:
This part of the documentation still needs work, but this PR is an improvement, so I think it's best to merge it now, and let the core devs work on it more later.

PR #3282, opened 2020-03-17
Replace Pool*Arrays with Packed*Arrays in tutorials and references

pycbouh:
Updated all class references and mentions from PoolArrays to PackedArrays. Some documentation around these terms needs to be reworked though.

I have concerns about `tutorials\misc\binary_serialization_api.rst` and `getting_started\scripting\c_sharp\c_sharp_differences.rst`. I've updated them to the best of my understanding, but there may be some mistakes regarding new 64-bit types.

PS. There are other invalid references as well, such as `SpatialMaterial`. I think we should keep the master here in sync with breaking changes in source code. There is going to be a lot of renaming and signature changes for classes and methods in 4.x, so it has to be easier to update the docs along code changes.


tutorials/misc/binary_serialization_api.rst
```diff
@@ -400,19 +404,41 @@ The array data is padded to 4 bytes.
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 +------------------+-------+-----------+---------------------------+
 | Offset           | Len   | Type      | Description               |
 +==================+=======+===========+===========================+
-| 4                | 4     |Integer    | Array Length (Floats)     |
+| 4                | 8     | Integer   | Array Length (Integers)   |
 +------------------+-------+-----------+---------------------------+
-| 8..8+length\*4   | 4     |Integer    | 32 Bits IEE 754 Float     |
+| 8..8+length\*8   | 8     | Integer   | 64 Bits Signed Integer    |
```
aaronfranke: I think the "Array Length" part is always 4.

Also, shouldn't they all be offset by 4 due to this? If so, the existing documentation is wrong.

pycbouh: @aaronfranke Well, it would make sense for it to always be 4, but it uses `encode_uint64` for `PackedInt64Array` specifically. It may be an error in `marshalls.cpp`, though. Int32, Float32 and Float64 all use `encode_uint32` for the length.

aaronfranke:
I think we can all agree that this part of the documentation still needs work, but this PR is an improvement, so I think it's best to merge it now, and let the core devs work on it more later.

Thanks!

PR #3283, opened 2020-03-17
Adjust styling of new instant search popup

pycbouh:
Adjusted light and added dark theme for new instant search popup added in #3277. Changed some default styling to match current static search page better.

Light theme:
https://giant.gfycat.com/JoyousGlisteningBoubou.mp4

Dark theme:
https://giant.gfycat.com/GenuineSaneAiredale.mp4

Edit: I've tested it by manually applying styles to the production page, as this extension does not run on local server due to a built-in limitation:
```
[INFO] Docs are not being served on Read the Docs, readthedocs-sphinx-search will not work.
```

**Technical note:** While extension's styles are added later than `custom.css` it is possible to override them using higher specificity rules, which is what I did.

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -35,6 +35,14 @@
     --input-background-color: #fcfcfc;
     --input-focus-border-color: #5f8cff;
 
+    --search-input-background-color: #e6eef3; /* derived from --input-background-color */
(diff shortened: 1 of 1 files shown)


akien-mga:
Thanks!

PR #2512, opened 2019-06-07
Subjectively improve tutorial on encrypting game saves

crecs:
This is subjective, but I felt the humor in the beginning of this tutorial on encrypting game saves was a bit long winded, out of place, and wouldn't be encouraging to the people who were trying to implement it, namely, mobile developers.

The tutorial was also called "encrypting save games", but "encrypting game saves" sounds more correct.

PR #3292, opened 2020-03-21
remove diatribe

migliaccio:
I understand the intent is to be funny, but it really takes away from the overall quality of the documentation.

### DIFF
--- a/tutorials/io/encrypting_save_games.rst
+++ b/tutorials/io/encrypting_save_games.rst
@@ -3,35 +3,6 @@
 Encrypting save games
 =====================
 
-Why?
-----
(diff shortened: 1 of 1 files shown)


Calinou:
Last time we discussed it in #2512, we decided to keep the paragraph in question.

akien-mga:
We still don't plan to remove this indeed. While a few users like yourself want it removed, the overwhelming response from the community is to laugh out loud and enjoy being able to have some *very rare* light hearted content in an otherwise very dense documentation (thousands of pages).

Stradex:
Get a life nerd

PR #3295, opened 2020-03-22
Clarification on "Identifier" term in docs

kzietek:
When filling iOS export forms, I got a red error that I'm missing an "Identifier". As an iOS developer and I find the term "Identifier" confusing and it should be explicitly told that it's a Bundle Identifier. When I'm looking at the form itself I thought this doesn't need to be filled as there was already a placeholder with reverse dns notation and I thought it was a default value.

It's my first contact with Godot engine and this also should be renamed within the export window itself.


Calinou:
Thanks!

PR #3297, opened 2020-03-22
Add get_slide_collision example to Using Kinematic tutorial

cbscribe:
A lot of users don't understand how to get collision info when using `move_and_slide()`, so I've added a brief example.


akien-mga:
Thanks!

PR #3300, opened 2020-03-24
use csharp syntax on csharp tab

mralext20:
as per the title, this code block was using the GDScript syntax instead of the c# syntax.


pycbouh:
Don't you need semicolons as well?

akien-mga:
Thanks!

PR #3305, opened 2020-03-25
Add a formatting script for local checking and for Travis CI 

aaronfranke:
This is a mirror of https://github.com/godotengine/godot-demo-projects/pull/450 (the code is exactly the same).

This PR adds a simple script which can be ran locally to fix file formatting, or ran on Travis CI in the case of checking pull requests. It enforces UTF-8, LF line endings, no BOM, and newlines at the ends of files. We could possibly add more checks later.

The bottom half of the script was based on the script in the main Godot repo.

Travis CI builds won't run unless they're enabled, so pinging @akien-mga.


Calinou:
Could we also ensure there's no trailing whitespace in files? (We'll probably have to make a pass to remove it from the current documentation.)

akien-mga:
> Could we also ensure there's no trailing whitespace in files? (We'll probably have to make a pass to remove it from the current documentation.)

That would be nice yeah.

I'll merge this as a first step already though.

akien-mga:
Thanks!

PR #3306, opened 2020-03-25
Simplify the "Making main screen plugins" article

aaronfranke:
The existing article was a bit overcomplicated. I don't think an article about main screen plugins should include a side panel, or include complex signals - it should just focus on main screen plugins.

The existing article also had some unnecessary parts, such as the suggestion of manually creating `plugin.cfg` etc files when Godot can do this for you via the "Plugins" menu.

I've also added this plugin to the demo projects repository: https://github.com/godotengine/godot-demo-projects/pull/453

A lot of this article changed, so I'll ping the original author: @StraToN

### DIFF
--- a/tutorials/plugins/editor/making_main_screen_plugins.rst
+++ b/tutorials/plugins/editor/making_main_screen_plugins.rst
@@ -7,43 +7,24 @@ What this tutorial covers
 -------------------------
 
 As seen in the :ref:`doc_making_plugins` page, making a basic plugin that
-extends the editor is fairly easy. This plugin mechanism also allows you to
-create new UIs in the central part of the editor, similarly to the basic 2D, 3D,
-Script and AssetLib views. Such editor plugins are referred as "Main screen
-plugins".
+extends the editor is fairly easy. Main screen plugins allow you to create
+new UIs in the central part of the editor, which appear next to the
+"2D", "3D", "Script", and "AssetLib" buttons. Such editor plugins are
+referred as "Main screen plugins".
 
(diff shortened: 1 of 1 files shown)


tutorials/plugins/editor/making_main_screen_plugins.rst
```diff
@@ -66,176 +47,92 @@ file. In our example, ``main_screen_plugin.gd``.
        pass
 
 
+    func handles(obj):
```
pycbouh: This makes it sound as if `handles` is a required method for main screen plugins, and it is not. This subject is better detailed in a separate section of this article, otherwise you fall into the same issue as with the old article and side panels.

tutorials/plugins/editor/making_main_screen_plugins.rst
```diff
@@ -66,176 +47,92 @@ file. In our example, ``main_screen_plugin.gd``.
...
+       return false
+
+
     func get_plugin_name():
        return "Main Screen Plugin"
 
 
     func get_plugin_icon():
-       return get_editor_interface().get_base_control().get_icon("Node", "EditorIcons")
+       pass
```
pycbouh: This function has a return type, so I think we should leave the example with the editor icon as is.

tutorials/plugins/editor/making_main_screen_plugins.rst
aaronfranke: Good point. I changed the examples to just a mention, but also users can always download the demo project if they want to see an example of how `handles` works.

PR #3318, opened 2020-03-28
Replaced deprecated "get" cs func in fps tutorial

docquantum:
In the 3D fps tutorial, changed the camera global transform
function to use direct property setting instead of getter
since the getter is deprecated in 3.2.1.


Calinou:
The equivalent GDScript code snippet should also be updated to contain the same change.

docquantum:
I'll have to go through the GDScript equivalent of the tutorial and fix things up as well. Been currently trying to make a C# version for part 2, has been a headache with all the typing (best solution so far has been to check for type of weapon instance and downcast)

mhilbrunner:
I'll merge this one for now, feel free to change the GDScript one too!

Thanks and congrats to your first contribution!

PR #3328, opened 2020-03-30
Add missing arguments to C# example of signals

### LINKED ISSUE #3327: signals arg csharp example missing args?
https://github.com/godotengine/godot-docs/blob/master/getting_started/step_by_step/signals.rst

should these last two csharp examples about signals with args have args like the gd ones? thanks

    public class Main : Node
    {
        [Signal]
        public delegate void MySignal();
    }

pycbouh:
Closes #3327.


Calinou:
Thanks!

PR #3329, opened 2020-03-30
Add a list of all default keyboard shortcuts

### LINKED ISSUE #3244: Wrong shortcuts in step-by-step guide?
I'm following the step-by-step guide using 3.2 on macOS and [this page](https://docs.godotengine.org/en/stable/getting_started/step_by_step/intro_to_the_editor_interface.html) mentions the shortcuts F1-F3 for switching workspaces. They don't work - it seems they're actually Alt+1 through Alt+3 (only on macOS perhaps?).
Similar for F4 to 'search the reference' - in the editor settings I found Shift+F1 for 'Search Help' - is that what is meant?

pycbouh:
Closes #3244 and addresses any future inconsistencies.

This page should be useful for people looking up default editor shortcuts and contributors looking for alternative key combinations for their tutorials and guides.

I've made it by hand, but it should be possible to automate the process. All we need is a script that looks for ``ED_SHORTCUT`` calls and extracts the line. Then a regular expression can be used to format that line into a table row, like so:
```
ED_SHORTCUT\("([a-z0-9_/]+)", TTR\("([a-z0-9.\(\)/ ]+)"\), ([A-Z0-9_ +]+)\);

| $2 | :kbd:`$3` | :kbd:`$3` | ``$1`` |
```

Note, that this regex assumes, that every key combination is formatted like ``KEY_ + KEY_``. In engine both this and ``KEY_ | KEY_`` are used interchangeably. Keys themselves should be easy to replace, and section names can be generated from file paths.

I've also found a potential conflicting binding in ``script_text_editor.cpp``, but I cannot confirm if it is in fact causing issues:
```cpp
	ED_SHORTCUT("script_text_editor/goto_previous_bookmark", TTR("Go to Previous Bookmark"), KEY_MASK_CMD | KEY_MASK_SHIFT | KEY_B);
...
#ifdef OSX_ENABLED
	ED_SHORTCUT("script_text_editor/toggle_breakpoint", TTR("Toggle Breakpoint"), KEY_MASK_CMD | KEY_MASK_SHIFT | KEY_B);
#else
	ED_SHORTCUT("script_text_editor/toggle_breakpoint", TTR("Toggle Breakpoint"), KEY_F9);
```


Calinou:
Thanks a lot for making this list :slightly_smiling_face:

We should try to write a shell/Python script to generate the table automatically, but I think this is good to merge for now.

> Note, that this regex assumes, that every key combination is formatted like KEY_ + KEY_. In engine both this and KEY_ | KEY_ are used interchangeably. Keys themselves should be easy to replace, and section names can be generated from file paths.

We should probably settle on either `+` or `|` in the engine. Which one would be the most suited for shortcuts? Are there cases where we *must* use `|` over `+`?

pycbouh:
> We should probably settle on either + or | in the engine. Which one would be the most suited for shortcuts? Are there cases where we must use | over + or vice versa?

I don't see how can there be any difference. Both operators perform an addition. Naturally, for display ``+`` looks better in shortcuts, but from the code point of view I'd stick with ``|``, as it implies a combination of entities rather than just a summation. This is probably better discussed in a context of code contribution guidelines.

> We should try to write a shell/Python script to generate the table automatically.

Shall I create an issue, describing what such script should do? It's going to be tricky combining the generated table with introductory paragraphs, though.

Calinou:
> Shall I create an issue, describing what such script should do? It's going to be tricky combining the generated table with introductory paragraphs, though.

Sure :slightly_smiling_face:

PR #3330, opened 2020-03-31
Use direction_to, remove code example for manually normalizing in vector math

aaronfranke:
Also link to matrices and transforms and 3Blue1Brown from within the advanced vector math article.


Calinou:
Thanks!

PR #3321, opened 2020-03-28
Rename all instances of the `x11` platform to `linuxbsd`

Calinou:
Export template binaries haven't been renamed yet, so their names were left as-is. Let me know if we should rename them from `linux_x11_{arch}_{target}` to `linux_{arch}_{target}` before merging this PR.

This supersedes #3319.

PR #3331, opened 2020-03-31
Update x11 to linuxbsd

quantmint:
Platform x11 is now linuxbsd.


pycbouh:
There is already a more complete PR for this: #3321

quantmint:
Ok, so we can close this

PR #3334, opened 2020-03-31
Fix macOS shortcuts for Delete key

pycbouh:
In #3329 I missed the fact, that editor automatically changes <kbd>Del</kbd> to <kbd>Cmd + BkSp</kbd> on macOS. This fixes all occurrences in the shortcut list.


Calinou:
Thanks!

PR #3336, opened 2020-03-31
Attempting to clean up thread safe language.

mason-bially:
There is some confusing and somewhat conflicting repetition in these docs. I am attempting to clean it up. To be clear I don't know the truth of this, I just know the current documentation reads as a conflicting advice:

> Modifying a unique resource from multiple threads is not supported, but loading them on threads or handling a reference is perfectly supported. Scenes, textures, meshes, etc. Can be loaded and manipulated on threads, then added to the active scene in the main thread.

> Still, this is only really useful if you have *one* thread loading data. Attempting to load or create scene chunks from multiple threads may work, but you risk resources (which are only loaded once in Godot) tweaked by the multiple threads, resulting in unexpected behaviors or crashes.

Notably the first one can be read as "Modifying a unique resource **from multiple threads** is not supported, **but loading** them **on threads** or handling a reference **is perfectly supported**." which implies loading from multiple threads.

I went with the more specific, safer advice after asking briefly on discord.


mason-bially:
Also the whole paragraph seems redundant to the above section.

tutorials/threads/thread_safe_apis.rst
```diff
@@ -57,4 +57,4 @@ In GDScript, reading and writing elements from multiple threads is ok, but anyth
 Resources
 ---------
 
-Modifying a unique resource from multiple threads is not supported, but loading them on threads or handling a reference is perfectly supported. Scenes, textures, meshes, etc. Can be loaded and manipulated on threads, then added to the active scene in the main thread.
+Modifying a unique resource from multiple threads is not supported. However handling references on multiple threads is perfectly supported, hence loading resources on a thread is as well - scenes, textures, meshes, etc - can be loaded and manipulated on a thread and then added to the active scene in the main thread. The limitation here is as described above, one must be careful not to load a the same resource from multiple threads at once, therefore it is easiest to use **one** thread for loading and modifying resources, and then the main thread for adding them.
```
mhilbrunner, line 60: Typo:
`not to load a the same resource`
should be
`not to load the same resource`

mhilbrunner, line 60: IMO I'd remove the `perfectly ` from `perfectly supported`, to me it reads a little weird :) Or replace it with `perfectly fine`.

mhilbrunner, line 60: `then added to the active scene in the main thread`
would replace `in` with `on the main thread` for consistency

mhilbrunner:
Ping @mason-bially  :)

tutorials/threads/thread_safe_apis.rst
mhilbrunner, line 60:
```suggestion
Modifying a unique resource from multiple threads is not supported. However handling references on multiple threads is supported, hence loading resources on a thread is as well - scenes, textures, meshes, etc - can be loaded and manipulated on a thread and then added to the active scene on the main thread. The limitation here is as described above, one must be careful not to load the same resource from multiple threads at once, therefore it is easiest to use **one** thread for loading and modifying resources, and then the main thread for adding them.
```

mhilbrunner:
Applied the few small things I noted and merged.

Thanks for this, and congrats on your first docs contribution! :)

PR #3345, opened 2020-04-05
Make note about atlases 3.2

NHodgesVFX:
Make note that default atlas setting only allows 16 probes. may not apply to godot 4.0

### DIFF
--- a/tutorials/3d/reflection_probes.rst
+++ b/tutorials/3d/reflection_probes.rst
@@ -107,6 +107,6 @@ In the current renderer implementation, all probes are the same size and
 are fit into a Reflection Atlas. The size and amount of probes can be
 customized in Project Settings -> Quality -> Reflections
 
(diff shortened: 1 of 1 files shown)


tutorials/3d/reflection_probes.rst
```diff
@@ -107,4 +107,6 @@ In the current renderer implementation, all probes are the same size and
 are fit into a Reflection Atlas. The size and amount of probes can be
 customized in Project Settings -> Quality -> Reflections
 
+The default settings of Atlas Size: 2048 and Atlas Subdiv: 8 will allow up to 16 reflection probes in a scene. These values need to be increased if you need more.
```
clayjohn, line 110: Atlas size doesnt actually affect how many probes you can have in the scene, it just affects their resolution. 

```suggestion
The default setting of Atlas Subdiv: 8 will allow up to 16 reflection probes in a scene. This value needs to be increased if you need more reflection probes.
```

clayjohn:
Note for maintainers, this PR only applies to 3.2.

akien-mga:
Thanks!

PR #3346, opened 2020-04-05
Fix missing mono suffix in "compiling with mono" guide

MiniDigger:
The section that tries to explain how important the "mono" suffix is, ironically forgets to append said suffix in the example.

ran into this while figuring out why my glue wasnt generated, should have read the whole page before doing stuff ^^


akien-mga:
Thanks!

PR #3347, opened 2020-04-06
Update "Unity to Godot"

asheraryam:
The mentioned disadvantage of godot is now obsolete since https://github.com/godotengine/godot/pull/27742 was merged for 2D and 3D.
i.e. The game camera can be overridden from the editor to explore from multiple angles.

Is it still relevant to mention this point in the bullet points below or should it be removed completely?


Calinou:
Thanks!

> Is it still relevant to mention this point in the bullet points below or should it be removed completely?

Yes, Unity users might wonder how to enable the camera override option since it's not enabled by default.

PR #3353, opened 2020-04-09
Always use parentheses when calling `assert()` in GDScript

Calinou:
This is now required since Godot 3.2.

This also adds an assert message to the Godot interfaces page to match its C# counterpart.


akien-mga:
Thanks!

PR #3354, opened 2020-04-09
Add singular "Signal" meta keyword to signal tutorials

### LINKED ISSUE #2629: Improve integration of documentation on signals
Looking at all the documentation on signals, I'm finding a lot of fragmentation of information on how to use them. Information of varying comprehensiveness and clarity is found in (so far that I've seen):

http://docs.godotengine.org/en/3.1/getting_started/scripting/gdscript/gdscript_basics.html#signals
http://docs.godotengine.org/en/3.1/getting_started/step_by_step/scripting.html#handling-a-signal
https://docs.godotengine.org/en/3.1/getting_started/step_by_step/signals.html
https://docs.godotengine.org/en/3.1/getting_started/scripting/c_sharp/c_sharp_features.html
https://docs.godotengine.org/en/3.1/tutorials/misc/instancing_with_signals.html

Not only that, but many of these pages have information or examples that the others do not contain, even by reference. As signals are so core to what makes godot so flexible and powerful, I began transferring some information and references and as they started to pile up, I decided to open an issue to discuss the structure of some of the changes I had considered making.

My top priority would be this page: https://docs.godotengine.org/en/3.1/getting_started/step_by_step/signals.html
This is the only page titled "signals" and as such is where a lot of people will land when looking for information on how to use signals. However, it contains no examples or even mentions of using arguments with signals, even in the custom signals section.

I had started to update it to contain some of the information from the gdscript basics (which I think are appropriate with a few style considerations) when I came across an idea that I think might be better, namely integrating https://docs.godotengine.org/en/3.1/tutorials/misc/instancing_with_signals.html into the signals file instead.

The signals page says, "Below you can see some examples of how you can use signals in your own projects." But only the Timer portion really qualifies as an example, the custom signals portion is abstract boilerplate that gives very little added insight on the power of signals. The instancing with signals page on the other hand uses custom signals (with arguments!) to great effect. It's somewhat hidden away in the misc section and (in my opinion) perhaps a bit mistitled, as it's not really about instancing using signals, but decoupling using signals.

I think adding that example into the signals page as a demonstration of custom signals would be a great solution. I would try to update it to be a bit more comprehensive, for instance both with and without arguments.

The page could also use some references to the others, as well as some notes on other considerations (for instance, the all lowercase snake_case auto named methods not fitting C# naming conventions)

I'd be happy to do this along with some typo and consistency fixes across all the pages (plus any others anyone else knows about) but I wanted to get some thoughts out there since my idea moved out of adding a couple lines for completeness' sake into something a little bigger.

some other considerations:
"signal" is a mediocre search term compared to "signals", which strikes me as odd since it's the keyword name. Is it possible to add metadata so that one page will be the top result for both "signals" and "signal"?

should we create a page for signals/signal outside of getting started to serve as the real landing page for signals? Probably unnecessary imo but worth discussing all the same.

Calinou:
This makes possible to find them when searching for "signal", instead of just "signals".

This closes #2629.

PS: This appears to work only when using the old-style search, not the new "instant" search. This is probably an issue with the search plugin (try searching for `faq` on the current site).

### DIFF
--- a/getting_started/step_by_step/signals.rst
+++ b/getting_started/step_by_step/signals.rst
@@ -1,3 +1,6 @@
+.. meta::
+    :keywords: Signal
+
 .. _doc_signals:
 
 Signals
@@ -208,7 +211,7 @@ To emit a signal via code, use the ``emit_signal`` function:
             EmitSignal(nameof(MySignal));
         }
     }
-    
+
 A signal can also optionally declare one or more arguments. Specify the
 argument names between parentheses:
 
--- a/tutorials/misc/instancing_with_signals.rst
+++ b/tutorials/misc/instancing_with_signals.rst
@@ -1,3 +1,6 @@
+.. meta::
+    :keywords: Signal
+
 .. _doc_instancing_with_signals:
 
 Instancing with signals


akien-mga:
Thanks!

PR #3355, opened 2020-04-09
Add file name conventions in the GDScript style guide

### LINKED ISSUE #2376: Naming conventions case for files
I would like to adhere to any existing conventions for file name casing. Currently, I can't find any specific file naming conventions in the docs. It feels ambiguous, because
* most tutorials use snake case, but
* I mostly see nodes using Pascal case, and by default, scenes are saved with their root node's name.

Should I use snake case, Pascal case, or does it even matter?

Calinou:
This closes #2376.

**Note:** Before merging this, we may want to modify the Godot editor so it converts automatically-generated file names to `snake_case`. I've been meaning to do this for a while.


mhilbrunner:
As discussed on IRC, we'll merge this as it's correct even if the engine does not yet automatically do this. Thanks!

PR #3358, opened 2020-04-09
Clarify how string enums are created in the Data preferences page

Calinou:
See https://github.com/godotengine/godot-docs/commit/e8289b93f334c533890ae1e19cd345473ba2fac3#r38374868.

### DIFF
--- a/getting_started/workflow/best_practices/data_preferences.rst
+++ b/getting_started/workflow/best_practices/data_preferences.rst
@@ -273,10 +273,10 @@ limited only by their imagination.
 Enumerations: int vs. string
 ----------------------------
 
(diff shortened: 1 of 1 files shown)


mhilbrunner:
IMO this is somewhat better, but still confusing. Maybe remove most of the text, and only add a note that in the special export case strings work too?

Calinou:
@mhilbrunner How would you write the paragraph in question? I'm not sure how to make it less confusing.

mhilbrunner:
Yeah, nevermind, I thought about this since yesterday and didn't come up with anything better. As it's an improvement, I'll merge, and *if* I come up with a further improvement I'll PR that later.

Thanks!

PR #3362, opened 2020-04-10
Add section on platform not appearing in SCons in Compiling for Android

### LINKED ISSUE #3361: Add troubleshooting avenues for building [export templates] for Android
Howdy.

I have spent a long time, much too long, on figuring out why scons would not list `android` as an available platform when attempting to build export templates for Android from a fresh clone of the 3.2 branch from the repo + one small module.

I triple-checked that I had all of the prerequisites, version numbers, the whole shebang. Long story short, the missing link was that the environment values `ANDROID_HOME` and `ANDROID_NDK_ROOT` were unset, and these are mentioned nowhere on the docs or in the forums that I could see.

Thus, I would like to suggest that there is a small troubleshooting section or an aside added to the [Compiling for Android](https://docs.godotengine.org/en/latest/development/compiling/compiling_for_android.html#building-the-export-templates) section that mentions checking these two variables.

Thank you!

Edit: to be clear, [this (rather buried) issue](https://github.com/godotengine/godot/issues/27091) helped me solved my problem, and I came across it by chance after a very specific set and order of keywords entered into Google.

Calinou:
This closes #3361.

### DIFF
--- a/development/compiling/compiling_for_android.rst
+++ b/development/compiling/compiling_for_android.rst
@@ -38,6 +38,8 @@ For compiling under Windows, Linux or macOS, the following is required:
 .. seealso:: For a general overview of SCons usage for Godot, see
              :ref:`doc_introduction_to_the_buildsystem`.
 
+.. _doc_android_setting_up_the_buildsystem:
+
 Setting up the buildsystem
 --------------------------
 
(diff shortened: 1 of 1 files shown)


akien-mga:
Thanks!

PR #3364, opened 2020-04-10
Set up continuous integration using Travis CI

Calinou:
This is used to catch build warnings early on.

[Example job output](https://travis-ci.org/github/Calinou/godot-docs/builds/673323009)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,15 @@
 dist: bionic
+language: python
+
+python:
+  - "3.8"
 
 stages:
   - build
 
 matrix:
   include:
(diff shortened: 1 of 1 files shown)


aaronfranke:
I see that the formatting checks and the Sphinx build are under the same job. It might be better if they were under different jobs, like on the main repo. Thoughts?

Calinou:
@aaronfranke That could be done. However, since the static checks job takes a very low time to complete, there's not much point in offloading it to a separate job (it'd take longer to set up the VM than complete the job).

akien-mga:
Thanks!

PR #3366, opened 2020-04-10
Safe line vs Unsafe line unclear label

FrederickKDP:
Although line 6 is safe, line 7, that actually is what the paragraph talking about, is unsafe and line 8 is safe.
The way it was before, it was in reverse order, making it confusing which is safe and which is not. So I changed the order to be as the picture.
To further assure it is clear, I added the line numbers in parentheses.
The parentheses could be excessive verbosity, but the order importance stands. I believe it was confusing before.
Alternatively, the picture could be changed, having the safe line first. Then it's clearer as the unsafe line it's the only one different and the order would be also correct; 2 safe lines, one unsafe for last. Specifying the lines would not be needed then.

### DIFF
--- a/getting_started/scripting/gdscript/static_typing.rst
+++ b/getting_started/scripting/gdscript/static_typing.rst
@@ -201,9 +201,9 @@ Godot will ensure the type works and if so, the line number will turn
 green at the left of the script editor.
 
 .. figure:: ./img/typed_gdscript_safe_unsafe_line.png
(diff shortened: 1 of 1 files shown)


FrederickKDP:
Request is not ready yet because I would like to hear some thoughts first.
For me, having never heard of the term before, it was hard to understand it.

FrederickKDP:
Thank you!

PR #3368, opened 2020-04-10
Improve navbar scrolling in the live version of the docs

pycbouh:
Further improvements after #3254.

Live version has some differences, not available for local testing. The banner at the bottom of the sidebar introduced some problems. Because I had to spend so much time with `custom.js`, I've refactored it as well. And covered it with updated comments for future contributors.

@Calinou I was unable to reproduce the FF issue you mentioned, but I've disabled horizontal scrolling.

Tested in almost up-to-date Chrome and freshly installed Firefox, including the device emulation mode in the former. Seems to work fine.

Edit:
In case someone want to test the banner locally, I can only suggest temporary putting all the necessary CSS and layout via `custom.*` files. Put the styles at the top of `custom.css`. You need lines 44 through 264 from https://assets.readthedocs.org/static/css/readthedocs-doc-embed.css. Layout can be inserted with JavaScript, just put something like this at the beginning of `$(document).ready(() => {`:

```js
$(document).ready(() => {
  /* DEBUG ONLY *** REMOVE BEFORE COMMIT */

  const ethicalBlock = '<div id="rtd-0534d2vm" class="ethical-rtd ethical-dark-theme">...</div>';
  const sidebarContainer = document.querySelector('.wy-side-scroll');
  sidebarContainer.innerHTML += ethicalBlock;

  /* DEBUG ONLY *** REMOVE BEFORE COMMIT */
```

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -755,6 +755,11 @@ kbd, .kbd {
     margin-right: 0;
 }
 
+/* Banner panel in sidebar */
+.wy-nav-side .ethical-rtd.fixed {
+    position: fixed
+}
+
 /* Version selector (only visible on Read the Docs) */
 
 .rst-versions {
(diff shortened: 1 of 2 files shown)


Calinou:
Thanks!

pycbouh:
That banner... It seems that it gets added in the live version after our script has run. I will look into it.

akien-mga:
Yes, the banner is added by the ReadTheDocs host.

PR #3371, opened 2020-04-11
Make the documentation logo hiDPI-friendly

Calinou:
This also sets the documentation logo to a fixed size in the CSS to prevent reflows during page loading.

## Preview

*Both screenshots were taken at 200% zoom, which is equivalent to Retina display scaling on macOS.*

### Before

godot_docs_before_hidpi

### After

godot_docs_after_hidpi


akien-mga:
Thanks!

PR #3372, opened 2020-04-11
Add a style guide for the Godot shader language

### LINKED ISSUE #3091: Create a Godot Shader Language style guide
We have style guides for GDScript and C#, but there is currently no style guide for GDSL. We should probably create one. I was looking over a pull request in the demo projects repository which adds some shader code, and it had some questionable things in it (such as no space around addition and multiplication), but I found there is no style guide to reference for what *should* be done.

Since GDSL is based on GLSL, and GLSL is based on C's syntax, perhaps we should adopt a style similar to Godot's internal C++ code? It might also be a good idea to write a C++ style guide, but that's a whole other beast by itself.

CC @Chaosus @clayjohn @JFonS @reduz (since y'all have contributed heavily to shaders, at least from what I can gather with a git blame), CC @nathanwfranke (since I know you use shaders a lot), and of course anyone else who uses shaders is welcome to chime in (btw, I have not).

Calinou:
This is partly inspired by Godot's GLSL style guide in addition to the GDScript style guide (mainly for the page structure).

Help is welcome for making the examples more realistic. Some of them might not even be using valid syntax in the first place, so please double-check.

This closes #3091.

### DIFF
--- a/tutorials/shading/godot_shader_language_style_guide.rst
+++ b/tutorials/shading/godot_shader_language_style_guide.rst
@@ -1,0 +1,357 @@
+.. _doc_godot_shader_language_style_guide:
+
+Godot shader language style guide
+=================================
+
+This style guide lists conventions to write elegant shaders. The goal is to
+encourage writing clean, readable code and promote consistency across projects,
+discussions, and tutorials. Hopefully, this will also support the development of
+auto-formatting tools.
+
+Since the Godot shader language is close to C-style languages and GLSL, this
+guide is inspired by Godot's own GLSL formatting. You can view an example of a
+GLSL file in Godot's source code
+`here <https://github.com/godotengine/godot/blob/master/drivers/gles2/shaders/copy.glsl>`__.
+
+Style guides aren't meant as hard rulebooks. At times, you may not be able to
+apply some of the guidelines below. When that happens, use your best judgment,
+and ask fellow developers for insights.
+
+In general, keeping your code consistent in your projects and within your team is
+more important than following this guide to a tee.
+
+.. note:: Godot's built-in shader editor uses a lot of these conventions
+          by default. Let it help you.
+
+Here is a complete shader example based on these guidelines:
+
+.. code-block:: glsl
+
+    shader_type canvas_item;
+    // Screen-space shader to adjust a 2D scene's brightness, contrast
+    // and saturation. Taken from
+    // https://github.com/godotengine/godot-demo-projects/blob/master/2d/screen_space_shaders/shaders/BCS.shader
+
+    uniform float brightness = 0.8;
+    uniform float contrast = 1.5;
+    uniform float saturation = 1.8;
+
+    void fragment() {
+        vec3 c = textureLod(SCREEN_TEXTURE, SCREEN_UV, 0.0).rgb;
+        
+        c.rgb = mix(vec3(0.0), c.rgb, brightness);
+        c.rgb = mix(vec3(0.5), c.rgb, contrast);
+        c.rgb = mix(vec3(dot(vec3(1.0), c.rgb) * 0.33333), c.rgb, saturation);
+        
+        COLOR.rgb = c;
+    }
+
+Formatting
+----------
+
+Encoding and special characters
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+* Use line feed (**LF**) characters to break lines, not CRLF or CR. *(editor default)*
+* Use one line feed character at the end of each file. *(editor default)*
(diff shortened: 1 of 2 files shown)


tutorials/shading/godot_shader_language_style_guide.rst
```diff
@@ -0,0 +1,364 @@
...
+        c.rgb = mix(vec3(0.5), c.rgb, contrast);
+        c.rgb = mix(vec3(dot(vec3(1.0), c.rgb) * 0.33333), c.rgb, saturation);
+        
+        COLOR.rgb = c;
+    }
+
+Formatting
+----------
+
+Encoding and special characters
```
clayjohn, line 52: This should probably be lower down as it has to do more with the text file itself than with the shader contained within.

tutorials/shading/godot_shader_language_style_guide.rst
```diff
@@ -0,0 +1,364 @@
...
+~~~~~~~~~~~
+
+Each indent level should be one tab greater than the block containing it.
+
+**Good**:
+
+.. code-block:: glsl
+
+    void fragment() {
+        COLOR = vec3(1, 1, 1);
```
clayjohn, line 70: vec3 is 3 floating point numbers, so should be:
```suggestion
        COLOR = vec3(1.0, 1.0, 1.0);
```

tutorials/shading/godot_shader_language_style_guide.rst
```diff
@@ -0,0 +1,364 @@
...
+    }
+
+**Bad**:
+
+.. code-block:: glsl
+
+    void fragment() {
+            COLOR = vec3(1, 1, 1);
```
clayjohn, line 78: Same as above:
```suggestion
            COLOR = vec3(1.0, 1.0, 1.0);
```

tutorials/shading/godot_shader_language_style_guide.rst
clayjohn: In GLES2 using an ``int`` in place of a ``float`` can cause a lot of problems. I think we may automatically convert ints to floats in GLES2, but since this is a style guide, it is better to encourage users to be explicit about what types they are using.

tutorials/shading/godot_shader_language_style_guide.rst
```diff
@@ -0,0 +1,364 @@
...
+~~~~~~~~~~~~~~~~~~~~~~
+
+Never combine multiple statements on a single line.
+
+**Good**:
+
+.. code-block:: glsl
+
+    void fragment() {
+        ALBEDO = vec3(1);
```
clayjohn, line 167: Same as above
```suggestion
        ALBEDO = vec3(1.0);
```

tutorials/shading/godot_shader_language_style_guide.rst
```diff
@@ -0,0 +1,364 @@
...
+        EMISSION = vec3(1);
```
clayjohn, line 168:
```suggestion
        EMISSION = vec3(1.0);
```

tutorials/shading/godot_shader_language_style_guide.rst
```diff
@@ -0,0 +1,364 @@
...
+    }
+
+**Bad**:
+
+.. code-block:: glsl
+
+    void fragment() {
+        ALBEDO = vec3(1); EMISSION = vec3(1);
```
clayjohn, line 176:
```suggestion
        ALBEDO = vec3(1.0); EMISSION = vec3(1.0);
```

tutorials/shading/godot_shader_language_style_guide.rst
```diff
@@ -0,0 +1,364 @@
...
+    }
+
+.. _naming_conventions:
+
+Accessing vector members
+------------------------
+
+Use ``r``, ``g``, ``b``, and ``a`` when accessing a vector's members if it
+contains a color. If the vector contains anything else than a color, use ``x``,
+``y``, ``z``, and ``w``.
```
clayjohn, line 287: Consider adding:
```suggestion
``y``, ``z``, and ``w``. This allows those reading your code to better understand what the underlying data represents.
```

tutorials/shading/godot_shader_language_style_guide.rst
Calinou: I put that section in the same place as it is on the GDScript style guide for consistency. I'm not sure if we need to change it, the current location seems fine to me.

clayjohn: Its not a big deal. To me it just feels out of place. But if that is what the GDScript style guide does, then I am fine with it.

akien-mga:
Thanks!

PR #3377, opened 2020-04-12
Small grammatical error fix for readability

MrKovar:
"As one might be able to tell, Dictionaries specialize in tasks that Arrays
aren't." changed to "As one might be able to tell, Dictionaries specialize in tasks that Arrays
do not."


Calinou:
Thanks!

PR #3378, opened 2020-04-13
Fix commands not appearing in "Bisecting regressions" page

YeldhamDev:

### DIFF
--- a/community/contributing/bisecting_regressions.rst
+++ b/community/contributing/bisecting_regressions.rst
@@ -93,19 +93,19 @@ instead of a commit hash.
 is done, in the terminal window, use ``cd`` to reach the Godot repository
 folder and enter the following command:
 
-.. code-block::
+.. code-block:: shell
 
     # <good> is the commit hash of the build that works as expected.
     # <bad> is the commit hash of the build exhibiting the bug.
-    git bisect start <good> <bad>
+    $ git bisect start <good> <bad>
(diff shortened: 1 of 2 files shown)


community/contributing/bisecting_regressions.rst
```diff
@@ -93,19 +93,19 @@ instead of a commit hash.
 is done, in the terminal window, use ``cd`` to reach the Godot repository
 folder and enter the following command:
 
-.. code-block::
+.. code-block:: shell
 
     # <good> is the commit hash of the build that works as expected.
     # <bad> is the commit hash of the build exhibiting the bug.
-    git bisect start <good> <bad>
+    $ git bisect start <good> <bad>
```
Calinou, line 100: I'm not too fond of prefixing commands with `$`, as this makes them harder to select without copying the `$` symbol alongside. (You can triple-click anywhere in a  line to select it.)

This could be fixed by adding `$` symbols with a CSS pseudoselector (whose content can't be selected), but that would be very difficult to do in Sphinx.

YeldhamDev: This is how it's done everywhere else in the docs.

Calinou: I know, I'll still merge this then.

Calinou:
Thanks!

PR #3386, opened 2020-04-15
Add advice about local variables in the GDScript and C# style guides

Calinou:
Follow-up to #3372, where I added such advice in the shader style guide (minus the member variable part).


aaronfranke:
Are the headers intended to be at different levels? `----------------` vs `~~~~~~~~~~~~~~~~`

Calinou:
@aaronfranke Yes, I made the headers siblings of the sections they're related to (rather than children).

PR #3387, opened 2020-04-15
Fix syntax highlighting for various languages

pycbouh:
I've noticed that we were missing some style rules defined by `Highlight.js`, which were not used by GDScript, but were used by others. For example, there is [an incorrect highlighting for JavaScript](https://docs.godotengine.org/en/latest/tutorials/platform/customizing_html5_shell.html#starting-the-project):

image

I've added almost every missing style rule, that was defaulting to `pygments.css`. Only omitted a couple I couldn't find any use for, and those need new, separate colors anyway. I think some of those are used in `diff`s, so if we introduce them in any guide we might need to add those rules.

I've added comments with AST entity names from `pygments.css` into our styles as well. This should help to adjust them in future without a need to go looking for classes' meanings.

Along the way I've checked whether syntax highlighting is working properly for any other code snippets that weren't GDScript and fixed several issues. There were enough of cases where highlighter was not defined or forced to a specific language, causing it to default to, in most of them, GDScript. I've tried to force proper languages wherever I've found an inconsistency. Mostly, this means that some unnecessary highlighting is no more. But there were cases where [highlighting erred and it was visible](https://docs.godotengine.org/en/latest/development/cpp/custom_modules_in_cpp.html#improving-the-build-system-for-development):

image

Sometimes code blocks were used to display a list of files or some arbitrary output from a random source. These cannot be properly highlighted. In some cases a `console` highlighter was forced, which I'm not so sure about, but the main problem was that, once again, style rules were missing [yielding an unreadable text](https://docs.godotengine.org/en/latest/development/cpp/custom_modules_in_cpp.html#writing-custom-documentation):

image

There are a couple of cases where we do not have access to a proper highlighter. One of them is `Resource` format used by GDScript. [This documentation](https://docs.godotengine.org/en/latest/development/file_formats/tscn.html) on `.tscn` files is defaulting to GDScript and for the most part it keeps the text readable, so I left it as is. Same can be said for the article on i18n, which uses snippets of `.po` files.

-----
Couple of additional notes.

[This guide on background loading](https://docs.godotengine.org/en/latest/tutorials/io/background_loading.html) is starting with `cpp` snippets, showcasing method signatures. This is kind of strange, as the article itself seems to be about GDScript approach, and the rest of the code is actually GDScript. I didn't change anything about it, but maybe it is worth a look. There may be other articles like this, and having `cpp` in front of GDScript can be confusing for newcomers.

I'm also a bit confused by [this guide to iOS services](https://docs.godotengine.org/en/latest/tutorials/platform/services_for_ios.html), as it was using Python highlighter for GDScript (which I fixed and assumed to be just some ancient contribution, before GDScript highlighter was available), but there are also some other snippets with code signatures in another language. The problem is, it can be any C-based language from the look of it. So I don't know what highlighter to force there.

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -45,6 +45,7 @@
     --search-credits-link-color: #4392c5; /* derived from --link-color */
 
     --highlight-background-color: #f5ffe1;
+    --highlight-default-color: #404040;
     --highlight-comment-color: #408090;
     --highlight-keyword-color: #007020;
     --highlight-keyword2-color: #902000;
@@ -126,6 +127,7 @@
 
         /* Colors taken from the Godot script editor with the Adaptive theme */
         --highlight-background-color: #202531;
+        --highlight-default-color: rgba(255, 255, 255, 0.85);
         --highlight-comment-color: rgba(204, 206, 211, 0.5);
         --highlight-keyword-color: #ff7085;
         --highlight-keyword2-color: #42ffc2;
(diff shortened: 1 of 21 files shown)


Calinou:
Thanks a lot for taking the time to do this :slightly_smiling_face:

PR #3389, opened 2020-04-15
Add information on detail section of spatial material

### LINKED ISSUE #3184: Explain Spatial Material - Detail Mask
I've had a bit of a difficulty to understand how the Detail property from a Spatial Material works, and the issue was that the Mask property needs a texture, but nowhere in the [3D Spatial Material tutorial](https://docs.godotengine.org/en/3.2/tutorials/3d/spatial_material.html) or in the [Spatial Material Docs](https://docs.godotengine.org/en/3.2/classes/class_spatialmaterial.html#class-spatialmaterial-property-detail-mask) it is explained what this texture will do.

Since there are Albedo and Normal slots for the Detail, it is confusing for a newcomer why and how a mask texture is needed.

I've only come to understand it by reading the first comment on [this issue](https://github.com/godotengine/godot/issues/18377).

I think just one more paragraph with this information is enough:

> The mask texture is interpreted as being the amount of the detail albedo/normal to combine via the chosen blend function. White = full detail, black = no detail. The 'mix' blend function has the mask effectively dictate how much base/detail there should be, so you can use it to make transitions from one albedo/normal to another, while the other blend modes are useful for simply combining the detail texture with the base material's albedo/normal.

And also, something like this in the Inspector hint.

skyace65:
Expands the Detail section of the SpatialMaterial page to explain each setting under Detail. I could use a second opinion on if I was clear enough in the description of the blend modes and mask. Closes #3184

### DIFF
--- a/tutorials/3d/spatial_material.rst
+++ b/tutorials/3d/spatial_material.rst
@@ -462,6 +462,38 @@ UV or triplanar modes, many interesting textures can be achieved.
 
 .. image:: img/spatial_material24.png
 
+There are several settings that control how detail is used.
+
+Mask: The detail mask is a black and white image used to control where the
+blending takes place on a texture. White is for the detail textures, Black
+is for the regular material textures, different shades of gray are for
+partial blending of the material textures and detail textures.
+
+Blend Mode: These four modes control how the textures are blended together.
+
+- Mix: Combines pixel values of both textures. At black, only show the material texture,
+  at white, only show the detail texture. Values of gray create a smooth blend between
+  the two.
+
+- Add: Adds pixel values of one Texture with the other. Unlike mix mode
+  both textures are completely mixed at white parts of a mask and not at gray
+  parts. The original texture is mostly unchanged at black
+
+- Sub: Subtracts pixel values of one texture with the other. The second
+  texture is completely subtracted at white parts of a mask with only a little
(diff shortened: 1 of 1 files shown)


tutorials/3d/spatial_material.rst
```diff
@@ -462,6 +462,38 @@ UV or triplanar modes, many interesting textures can be achieved.
...
+Mask: The detail mask is a black and white image used to control where the
+blending takes place on a texture. White is for the detail textures, Black
+is for the regular material textures, different shades of gray are for
+partial blending of the material textures and detail textures.
+
+Blend Mode: These four modes control how the textures are blended together.
+
+- Mix: Adds pixel values of one texture with the other. Neither texture is
+  completely mixed at white or black, instead being completely mixed in gray
+  parts of a detail mask.
```
clayjohn, line 476: I am having trouble understanding what you mean by "neither texture is completely mixed" maybe something like this?
```suggestion
- Mix: Combines pixel values of both textures. At black, only show the material texture,
  at white, only show the detail texture. Values of gray create a smooth blend between
  the two.
```

skyace65:
@clayjohn That's perfect! I just made the change.

akien-mga:
Thanks!

Saitodepaula:
Great job, thanks everyone!

PR #3402, opened 2020-04-17
Mention self-contained mode in the editor compilation guides

Calinou:
When working on the editor itself, using separate configuration is sometimes desired to avoid conflicts or accidentally losing settings.

This closes https://github.com/godotengine/godot/issues/21567.


akien-mga:
Thanks!

Issue #3390, opened 2020-04-15
Proposal: redesigning the docs for Godot 4


NathanLovato:
# Proposal: redesigning the docs for Godot 4

This high-level proposal drafts a **first** project to review and rework the documentation by Godot 4's release.

It stemmed from a long discussion with @akien-mga where we went over broad issues with the current manual.

**Table of Contents**

- [Proposal for the Godot docs](#proposal-for-the-godot-docs)
  - [Purpose and content of the proposal](#purpose-and-content-of-the-proposal)
    - [The proposal in short](#the-proposal-in-short)
  - [1. Focusing on documentation](#1-focusing-on-documentation)
    - [The state of the documentation](#the-state-of-the-documentation)
    - [What should be in the documentation, and what should not](#what-should-be-in-the-documentation-and-what-should-not)
  - [2. Defining clear content guidelines](#2-defining-clear-content-guidelines)
    - [Examples in manual pages](#examples-in-manual-pages)
  - [3. Reorganizing the documentation to match the guidelines](#3-reorganizing-the-documentation-to-match-the-guidelines)
    - [Improving the landing page and access to content](#improving-the-landing-page-and-access-to-content)
  - [4. Setting up a writing pipeline](#4-setting-up-a-writing-pipeline)
    - [Spell-checker](#spell-checker)
    - [Grammar checker](#grammar-checker)
    - [Prose linter](#prose-linter)
    - [Screenshot automation](#screenshot-automation)
  - [5. Using issues to track content to write systematically](#5-using-issues-to-track-content-to-write-systematically)
  - [6. Redesign the getting started series](#6-redesign-the-getting-started-series)
  - [Moving forward](#moving-forward)

## Purpose and content of the proposal

This proposal reviews the documentation at large, with the following goals:

1.  Improving navigation and content discovery.
2.  Improving the quality and consistency of the content.
3.  Completing the manual.
4.  Making the maintenance load manageable. That is to say, keeping the docs up-to-date.
5.  Solidifying the docs' tooling and pipeline.

As part of this project, we would like to collect the issues users from all horizons have faced with the documentation. We could do so using a form so we can have both qualitative and quantitative feedback. We should have a list of the most common and the most frustrating issues the users face with the current documentation to **set our priorities well**.

### The proposal in short

Here is find an overview of the proposal. **Please read the entire document** before sharing feedback, as the sections below covers the challenges and reasons for each suggestion in greater detail.

Here are the high-level problems this proposal attempts to address:

[...]

PR #3403, opened 2020-04-17
Add a Tween cheatsheet image

Calinou:
Taken from https://github.com/wandomPewlin/godotTweeningCheatSheet with the author's permission. Lossless PNG optimization was applied as usual.

The image isn't referenced in this repository directly, but will be referenced in the Tween class reference.


mhilbrunner:
Yay for having a visual!

But for me these are kind of hard to visualize, or is that just me? Hard to get a feel for the different behaviours if I'm looking for a certain bounce for my game.

lentsius-bark:
I love these, they work very well for me actually. One thing I'd add to the visual is the enum value of the tween.

mhilbrunner:
Probably just me then :) Do we want the Godot version embedded in the image? Seems like an unnecessary maintenance burden for an area where the content hopefully doesn't change as often as Godots version number.

Calinou:
@mhilbrunner I think it's better to display the version number the cheatsheet was last updated at. Think more of it as "this cheatsheet is known to be good for Godot 3.2".

See #3390 where people have complained about outdated documentation not explicitly claiming it's outdated.

mhilbrunner:
Sounds reasonable, yes. :)

mhilbrunner:
Thank you!

PR #3404, opened 2020-04-18
Travis CI: Check for possible typos using codespell

Calinou:
This should hopefully not result in too many false positives due to how codespell works (it only checks for misspellings from its own dictionary). On the other hand, false negatives may slip in.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+os: linux
 dist: bionic
 language: python
 
@@ -21,6 +22,10 @@ matrix:
 
 script:
   - bash ./format.sh
+
+  # Check for possible typos
+  - pip install codespell
+  - codespell -I codespell-ignore.txt {about,community,development,getting_started,tutorials}/**/*.rst
 
   - pip install -r requirements.txt
   # TODO: Add `-W` to turn warnings into errors.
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -22,6 +23,10 @@ matrix:
 script:
   - bash ./format.sh
 
+  # Check for possible typos
+  - pip install codespell
+  - codespell -I codespell-ignore.txt {about,community,development,getting_started,tutorials}/**/*.rst
```
akien-mga, line 28: Does `**/*.rst` also work recursively deeper than 2 levels? (e.g. `tutorials/vr/vr_starter_tutorial/*.rst`)

akien-mga: Just checked, it should :)

Calinou: Ideally, I'd have have used `^classes/**/*.rst`, but it seems to be a zsh-ism :slightly_frowning_face:

akien-mga: You may use `--skip "classes/*.rst"`.

akien-mga: Or `--skip ".git,classes"` seems to work too and also ignored `.git` (locally I'd add `_build`, but on Travis we shouldn't need it).

Calinou: That doesn't seem to work, let's keep it as it currently is then.

akien-mga:
Thanks!

PR #3409, opened 2020-04-19
Correction in object_class.rst

chaserdoe:
corrected _get_property_info -> _get_property_list


Calinou:

Calinou:
Thanks!

Issue #3411, opened 2020-04-19
Possible incorrect reference to tutorial project node


gmoretti:
Hello!! I'm loving GODOT!, so let's see if I can help!

**Your Godot version:**
v3.2.1.stable.official

**Issue description:**
On part 2 of the FPS (line 697) tutorial there is a reference to a Pistol_Aim_Point from which to grab the global transformation. It refers to a line of code in which:

`clone.global_transform = self.global_transform
`

The **self** being the **Pistol_Point** node, this is one of the **Gun_Fire_Point** node, so maybe it should be **Pistol_Point** to actually talk about the self object?

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/latest/tutorials/3d/fps_tutorial/part_two.html

Aim node is not the one at the end of the pistol
![CapturaGodot](https://user-images.githubusercontent.com/7014995/79695163-78effe00-8275-11ea-9fa4-241aa204d4cc.PNG)

gmoretti:
I've made a PR in case I'm right hehhe!
Thanks!

PR #3413, opened 2020-04-19
Fix reference to a node in FPS tutorial part 2

gmoretti:
The element self from which we are grabbing the global transform is Pistol_Point. The Aim Point is further away
Resolves https://github.com/godotengine/godot-docs/issues/3411


Calinou:
cc @TwistedTwigleg, as they wrote the FPS tutorial.

TwistedTwigleg:
I think it is right. I think the other is just for rotation purposes. I’d say it’s good.
(Thanks for the ping!)

Calinou:
Thanks!

PR #3423, opened 2020-04-20
Update gdscript_styleguide.rst

David2753:
- removed duplicate of good example under bad


pycbouh:
That's not a duplicate. That's just two examples of *bad* indentation.

David2753:
I see, didn't catch that!

pycbouh:
Maybe it can be worth it to have whitespaces visible for such cases:

image

PR #3425, opened 2020-04-21
compiling on osx needs to be forced to python3

abakobo:
Since python2 is deprecated, compiling for MacOS needs to be forced to python3.
MacOS has python2 as `python` for it's default env. Changing your python env on OSX is not advised as the system uses it for sereval MacOS features. This simple python3 forcing seems to be the cleanest way to get it working on any MacOS.

### DIFF
--- a/development/compiling/compiling_for_osx.rst
+++ b/development/compiling/compiling_for_osx.rst
@@ -32,7 +32,7 @@ Compiling
 
 Start a terminal, go to the root directory of the engine source code and type::
 
-    scons platform=osx --jobs=$(sysctl -n hw.logicalcpu)
+    python3 $(which scons platform=osx --jobs=$(sysctl -n hw.logicalcpu))
 
 If all goes well, the resulting binary executable will be placed in the
 ``bin/`` subdirectory. This executable file contains the whole engine and


development/compiling/compiling_for_osx.rst
```diff
@@ -32,7 +32,7 @@ Compiling
 
 Start a terminal, go to the root directory of the engine source code and type::
 
-    scons platform=osx --jobs=$(sysctl -n hw.logicalcpu)
+    python3 $(which scons platform=osx --jobs=$(sysctl -n hw.logicalcpu))
```
akien-mga, line 35: Should actually be `python3 $(which scons) ...`.

But IMO it's better to leave it as is and add a `.. note::` about how to enforce running it with Python 3, or directly in the requirements, as done in https://docs.godotengine.org/en/latest/development/compiling/compiling_for_linuxbsd.html#requirements

akien-mga:
Apple is quick to deprecate OpenGL or 32-bits, but they don't manage to ship Python 3 as default in 2020... ;)

abakobo:
Do you mean it should be `python3 $(which scons) platform=osx --jobs=$(sysctl -n hw.logicalcpu` ?
It worked fine for me the other way though I was surprised it was telling me it's auto-detecting osx ;)
I understand your point of view but as a novice godot user I must admit I like to just copy-paste the line. And I suppose a lot(all?) of godot compiling on osx will have to force python3 anyway.

abakobo:
I'll probably post an issue for that before trying to PR again. Thx

PR #3429, opened 2020-04-22
Install dependencies from `requirements.txt` when using virtualenv

Calinou:
The instructions above were already updated to use `requirements.txt`, but not the virtualenv-specific instructions.


mhilbrunner:
Thanks!

PR #3432, opened 2020-04-22
Point getting sources to pull request workflow

omerkling:
Potential contributors can land on the building instructions before the PR instructions.


mhilbrunner:
Thanks, it definitely doesn't hurt to reference those :)

PR #3437, opened 2020-04-24
Enable visible whitespace characters with emphasized lines

### LINKED ISSUE #3426: Make whitespaces visible in GDScript code samples
**Your Godot version:**
Any

**Issue description:**
As evident from #3423, it can be hard to understand different indent sizes in code samples. It may be only limited to the GDScript style guide, because it directly talks about various cases for the indentation, but it may make code more readable for some overall. Which is important, since GDScript is indent-sensitive language.

Unfortunately, docs don't use tabs for indents, which is a problem in itself. I think this limitation is deeply related to how Sphinx works. [Pygments can make both tabs and whitespaces visible](https://pygments.org/docs/filters/#VisibleWhitespaceFilter), if we ever solve that.

For now, we have to work with whitespaces. Ideally, to be as least obstructive as possible, we'd like to have it as an option for `::code-block` directive, but [it has no such functionality](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code-block). So, with a little hack, that is [only going to work in Sphinx prior to version 3.x](https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.add_lexer), we can enforce visible whitespaces on all `gdscript` code samples.

This must be added/changed in `gdscript.py` extension:
```python
from pygments.filters import (
    VisibleWhitespaceFilter,
)
...
def setup(sphinx):
    l = GDScriptLexer()
    l.add_filter(VisibleWhitespaceFilter(spaces=" ",wstokentype=True))
    sphinx.add_lexer("gdscript", l)
```

And styles similar to this should be added in `custom.css`:
```css
.highlight .w {
    position: relative;
}
.highlight .w:before {
    content: "·····································";
    position: absolute;
    left: 0;
    width: 100%;
    overflow: hidden;
    color: #4d566d;
    user-select: none;
}
```

And this is the result:

image

Note, that these characters are ignored when a user is selecting code samples.

However, not everyone likes visible whitespace characters. Forcing it globally may be undesirable. We can try and fix this particular example instead. It would be impossible to add any character in it and not break highlighting and copying. Some may say, that in this trivial example copying is not important. Highlighter errors can also be forcefully disabled.

As a middle ground, there is an option to only make those characters visible when used in a combination with emphasized lines:

image

To achieve this we need to make code blocks like this:
```
.. code-block::
    :emphasize-lines: 2

    for i in range(10):
        print("hello")
```

And adjust CSS rule to be `.highlight .hll .w:before`. Highlighted lines need a small fix as well, because we do not handle them at the moment. We just need to add a proper background color for them, like `background-color: #2d3444;`.

https://docs.godotengine.org/en/stable/getting_started/scripting/gdscript/gdscript_styleguide.html

YuriSizov:
Fixes #3426.

This adds special markup for whitespace characters in GDScript highlighter. While markup is always added, only emphasized lines display actual symbols. The "→" symbol is used, as GDScript must be indented with tabs, and not spaces.

This also automatically replaces 4 spaces with a tab in every GDScript code snippet. This should make it easier to copy code snippets throughout the docs. Note, that it's Sphinx that strips all tabs in `code-block`s.

So far only this particular example highlights tabs, as it has confusing "bad" examples. This also adds two explanatory comments for those examples, so that the distinction is even clearer.

image

image

This also fixes incorrect colors for emphasized lines in the dark theme, which are only found in [this C# basics doc](https://docs.godotengine.org/en/stable/getting_started/scripting/c_sharp/c_sharp_basics.html#project-setup-and-workflow) currently.

### DIFF
--- a/_extensions/gdscript.py
+++ b/_extensions/gdscript.py
@@ -31,6 +31,9 @@ from pygments.token import (
     Number,
     Punctuation,
 )
+from pygments.filters import (
+    VisibleWhitespaceFilter,
+)
 
 __all__ = ["GDScriptLexer"]
 
@@ -341,7 +344,15 @@ class GDScriptLexer(RegexLexer):
 
 
 def setup(sphinx):
-    sphinx.add_lexer("gdscript", GDScriptLexer())
+    """
+    We force the whitespaces filter here because Sphinx takes control away from us
+    further down the line. This hack requires add_lexer to accept an instance of the lexer.
+    This type of argument is not supported starting with Sphinx 3.x.
+    See https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.add_lexer
+    """
+    l = GDScriptLexer()
+    l.add_filter(VisibleWhitespaceFilter(spaces=" ",wstokentype=True))
+    sphinx.add_lexer("gdscript", l)
 
     return {
         "parallel_read_safe": True,
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -48,6 +48,7 @@
     --search-credits-link-color: #4392c5; /* derived from --link-color */
 
     --highlight-background-color: #f5ffe1;
+    --highlight-background-emph-color: #dbe6c3;
     --highlight-default-color: #404040;
     --highlight-comment-color: #408090;
     --highlight-keyword-color: #007020;
@@ -59,6 +60,7 @@
     --highlight-function-color: #06287e;
     --highlight-operator-color: #666666;
     --highlight-string-color: #4070a0;
+    --highlight-whitespace-color: #abb793;
 
     --admonition-note-background-color: #e7f2fa;
     --admonition-note-color: #404040;
@@ -133,6 +135,7 @@
 
         /* Colors taken from the Godot script editor with the Adaptive theme */
         --highlight-background-color: #202531;
+        --highlight-background-emph-color: #2d3444;
         --highlight-default-color: rgba(255, 255, 255, 0.85);
         --highlight-comment-color: rgba(204, 206, 211, 0.5);
         --highlight-keyword-color: #ff7085;
@@ -144,6 +147,7 @@
         --highlight-function-color: #57b3ff;
         --highlight-operator-color: #abc8ff;
         --highlight-string-color: #ffeca1;
+        --highlight-whitespace-color: #636875;
 
         --admonition-note-background-color: #303d4f;
         --admonition-note-color: #bfeeff;
@@ -501,6 +505,30 @@ code,
 
 .highlight {
     background-color: var(--highlight-background-color);
+}
+
+/* Emphasized lines */
+.highlight .hll {
+    background-color: var(--highlight-background-emph-color);
+}
+
+.highlight .w /* Text.Whitespace */ {
+    position: relative;
+    tab-size: 4; /* Defaults to 8, which is too big. */
+    -moz-tab-size: 4;
+}
+/* Whitespace characters only visible on emphasized lines. */
+.highlight .hll .w:before {
+    /* This is a blunt way to add just enough characters to cover most of reasonable indentations. */
+    content: "→   →   →   →   →   →   →   →   →   →   ";
+    color: var(--highlight-whitespace-color);
+    position: absolute;
+    left: 0;
+    width: 100%;
+    overflow: hidden;
+    /* Disable text selection and mouse events, as these elements are display-only. */
+    user-select: none;
+    pointer-events: none;
 }
 
 .highlight .gh /* Generic.Heading */,
--- a/_static/js/custom.js
+++ b/_static/js/custom.js
@@ -201,4 +201,12 @@ $(document).ready(() => {
   registerSidebarObserver(() => {
     registerOnScrollEvent(mediaQuery);
   });
+
+  // GDScript, as per the official style guide, uses tabs for indentation.
+  // Sphinx, however, replaces everything with spaces; additionally, most docs use spaces outright.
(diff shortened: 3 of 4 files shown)


Calinou (approved):
I tested this PR locally. Tab replacement works, but I can't see whitespace (probably because I'm using Sphinx 3.x locally?).

Either way, this looks fine to merge. Nice work :slightly_smiling_face:

pycbouh:
Hmm, from the description of `add_lexer` I'd assume that instances would just stop working, but as far as I understand Python, we were always passing an instance to that function.

> Take a lexer class as an argument. An instance of lexers are still supported until Sphinx-3.x.

Maybe it accepts instances, but ignores them and just reflects their class?

akien-mga:
> So far only this particular example highlights tabs, as it has confusing "bad" examples. This also adds two explanatory comments for those examples, so that the distinction is even clearer.

The added comments, but I don't see the point of highlighting tabs in this example specifically, as IMO it doesn't make things easier to parse. If we want to make it clear that 4-spaces should be used for indentation (that's what the style guide is about), then we should show spaces, not tabs.

There's little use for a style guide on indentation size when using tabs, since any indent should be one tab. By showing `<= 4 spaces` as a tab here, it actually makes things confusing as it looks like the second "bad" example is indented twice (two tabs), when it has actually only one indentation level (of size 6 - but it's valid in GDScript/Python).

pycbouh:
I'm confused. The second "bad" example **is** indented twice, with 8 total spaces, not 6. The guide tells to use tabs, then gives a set of rules to go with that. First one being about only using only one tab for a level of indentation, neither two spaces/half a tab (first example), nor two tabs (second example).

Symbols make the indentation distance more visible, which was the problem for the person who opened the PR which I based the issue on.

akien-mga:
Ok, I thought the second "bad" example used 6 spaces. Then showing two tab symbols is fine if it's 8 spaces / 2 tabs.

But the first "bad" example should then display two spaces, not one size-2 tab.

pycbouh:
> But the first "bad" example should then display two spaces, not one size-2 tab.

I voiced that concern in the issue as well, but it would require a bit of additional JavaScript code to achieve, as far as I can see it. At which point it was suggested to just add a descriptive comment.

I can add the thing, if you want it. But it may be a bit of an overkill for just one example (so far).

akien-mga:
> I can add the thing, if you want it. But it may be a bit of an overkill for just one example (so far).

But I feel that the current state in this PR is also a bit overkill in terms of Javascript code added for two examples, where one of them is not well served by the new code. IMO adding the comments might be enough.

Others might think otherwise though, so I'm happy to rest my case if the consensus is that it's better as done here.

pycbouh:
Well, the JavaScript added by this PR is not necessary for the initial problem at hand. My initial suggestion showed spaces just as well.

But it does affect all GDScript code snippets, not just this one. And I think it's a positive change after all, because it makes all snippets actually formatted in a way that we describe in the guidelines. I'd prefer if we could configure Sphinx to the same extent, though.

Either way, I'm ready to make any requested changes or to accept that this is not a needed PR. 🙂

mhilbrunner:
I'm not that thrilled to add even more functions that depend on a deprecated Sphinx behaviour, or Javascript that runs a string replace on every page load on the user's side.

At least for the latter, I'd hope there is a better way, unfortunately I'm not sure there is.

Still, code samples being copy-paste-able with tabs *is* an improvement IMO.

mhilbrunner:
Okay, after I spent some more time on this:

**Replacing spaces with tabs in the docs**

The right way to do this is IMO server side. We should try to disable the part of Sphinx that converts tabs to spaces when parsing RSTs (although that's what the RST spec says). See [this](https://stackoverflow.com/questions/58132637/preserve-tabs-in-sphinx-makefile-code-blocks) on StackOverflow.

Then, we could either replace spaces with tabs in our code samples manually, or let Sphinx do that on parsing (see [this](https://stackoverflow.com/a/12699677)), or do both.

I haven't fully tested this approach, but it seems workable and is IMO less hacky than running a text-replace on the page via JavaScript after the fact.

**Visible whitespace**

This is IMO useful, but not if it comes at the price of introducing hacks that won't work with higher Sphinx versions. I'm not sure whether there is a better way to do this, if so, I'd support it, if not, this is IMO not worth the technical debt.

mhilbrunner:
(Note that the above is my opinion, after looking into this a little - I'd like to wait for @akien-mga's opinion before you spend any more time on this :))

pycbouh:
> Replacing spaces with tabs in the docs
> The right way to do this is IMO server side.

I completely agree with this. If we can rely on the solution from the SO question you've linked, that would be much better. I do think that we should update all GDScript samples if we do that. And put a requirement for future contributors in our contribution guidelines.

> Visible whitespace
> This is IMO useful, but not if it comes at the price of introducing hacks that won't work with higher Sphinx versions. I'm not sure whether there is a better way to do this, if so, I'd support it, if not, this is IMO not worth the technical debt.

I agree with this too, but I have to say, I don't understand why Sphinx does not support adding filters to custom lexers in any other way. We can try to suggest a change to them, though in what way should they support it (as a lexer option, as a `code-block` directive option, something else?) I do not know. If they agree to implement something to fix the issue, I'd argue that we go with a hacky way until we upgrade to a version that supports it natively. Otherwise... 🤷

mhilbrunner:
I'll close this PR due to the approach being flawed as outlined above. Doing this may still be interesting, but as discussed we need to come up with a better way to do it.

Thanks for taking the time for this PR!

pycbouh:
Alright, I'll make a separate PR to fix highlighted lines, though. This needs to be fixed for those two examples who already use it.

PR #3443, opened 2020-04-25
Fix typo in c# script of bouncing/reflecting part

Nimmda:
Speed variable was with uppercase in declaration but lowercase in the rest of the class. Now it is Uppercase.


Calinou:
Thanks!

PR #3444, opened 2020-04-25
Document GDScript export category and property grouping

Xrayez:
Helps godotengine/godot#4378, godotengine/godot#10303.

I'm not sure how thorough this has to be, but good enough, better than nothing.

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_exports.rst
+++ b/getting_started/scripting/gdscript/gdscript_exports.rst
@@ -222,11 +222,13 @@ example of a built-in category.
 
     func _get_property_list():
         var properties = []
-        properties.append({
+        properties.append(
+            {
                 name = "Debug",
                 type = TYPE_NIL,
                 usage = PROPERTY_USAGE_CATEGORY | PROPERTY_USAGE_SCRIPT_VARIABLE
-        })
+            }
+        )
         return properties
         
 * ``name`` is the name of a category to be added to the inspector;
--- a/getting_started/scripting/gdscript/gdscript_exports.rst
+++ b/getting_started/scripting/gdscript/gdscript_exports.rst
@@ -190,3 +190,75 @@ When changing an exported variable's value from a script in
 automatically. To update it, call
 :ref:`property_list_changed_notify() <class_Object_method_property_list_changed_notify>`
 after setting the exported variable's value.
+
+Advanced exports
+----------------
+
+Not every type of export can be provided on the level of the language itself to
+avoid unnecessary design complexity. The following describes some more or less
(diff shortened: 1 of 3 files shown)


getting_started/scripting/gdscript/gdscript_exports.rst
```diff
@@ -190,3 +190,75 @@ When changing an exported variable's value from a script in
...
+
+::
+
+    func _get_property_list():
+        var properties = []
+        properties.append({
+                name = "Debug",
+                type = TYPE_NIL,
+                usage = PROPERTY_USAGE_CATEGORY | PROPERTY_USAGE_SCRIPT_VARIABLE
+        })
```
Xrayez, line 229: Not sure about GDScript style for this, namely ending `})` on the new line like this.

fire: There's a gdscript autoformatter. I would say follow what that does.

getting_started/scripting/gdscript/gdscript_exports.rst
```diff
@@ -190,3 +190,75 @@ When changing an exported variable's value from a script in
...
```
mhilbrunner, line 229:
```suggestion
    properties.append(
        {
            name = "Debug",
            type = TYPE_NIL,
            usage = PROPERTY_USAGE_CATEGORY | PROPERTY_USAGE_SCRIPT_VARIABLE
        }
    )
```

Seems to be what the GDScript autoformatter prefers to do.

mhilbrunner:
Merged, thanks!

EricEzaM:
I am trying to get this to work but just cant... Can this be expanded on please?

I am adding the code for _get_property_list() and ensuring my script is in tool mode... but nothing changes.

Xrayez:
@EricEzaM yeah this doesn't work well for *existing* properties defined with `export var`. In fact this is mostly useful in cases where you want to add properties programatically, so you have a control over the order of properties being inserted to the inspector. But you could also probably override the behavior of existing properties as in https://github.com/godotengine/godot-proposals/issues/761#issuecomment-620927931.

Also see concrete example I've using myself in https://github.com/godotengine/godot/pull/10303#issuecomment-618979118.

Tbh this needs a more thorough introduction on the `_get_property_list()`, but that also means describing the full functionality of it, could take days to test and document this properly.

> I guess it does achieve the end goal but it is a bit convoluted.

Tbh the export system itself is a bit convoluted: godotengine/godot-proposals#298.

fire:
@ShatReal contact me sometime and I’ll give a run-through. One use is defining what type a variant of type object is.

PR #3445, opened 2020-04-25
Define the Weblate badge size to avoid reflows during page loading

Calinou:
This improves perceived performance slightly.

There are still other sources of reflows during loading (namely the logo and "+" buttons in the navigation bar). I'll try to look at those in the future.


mhilbrunner:
Tested locally, looks good. Thank you! :)

PR #3446, opened 2020-04-25
Changed note-link to https in signals.rst

MomoHunter:


MomoHunter:
Good catch, I fixed that as well. Please take a look

Calinou:
Thanks!

PR #3448, opened 2020-04-26
Autowrap text inside tables by default

Xrayez:
### Before
no-wrap

### After
wrap

---

The CSS is basically taken from this [SO answer](https://stackoverflow.com/a/41596044/2564620), so take this with a grain of salt. But to me this doesn't seem to break anything visually. UPDATE: https://github.com/godotengine/godot-docs/pull/3448#issuecomment-619548077.

To note, it seems like it's possible to manually setup column width on case-to-case basis with:

```rst
.. table::
    :widths: 20 30 50
```

But I guess it's better to enable autowrap by default for most cases, and configure the width when necessary manually, WDYT?

Some code snippets written inside tables may also wrap, but those kind of snippets are not inside `::` code block specifier in the first place which would prevent them to be wrapped, so those cases may need to be fixed if proven critical.

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -463,6 +463,18 @@ footer,
     background-color: var(--table-row-odd-background-color);
 }
 
+/* Override table no-wrap */
+/* The first column cells are not verbose, no need to wrap them */
+.wy-table-responsive table td:not(:nth-child(1)), 
+.wy-table-responsive table th:not(:nth-child(1)) {
+    white-space: normal;
+}
+
+/* Make sure not to wrap keyboard shortcuts */
+.wy-table-responsive table td kbd {
+    white-space: nowrap;
+}
+
 /* Code display tweaks */
 
 code,


Xrayez:
Found critical bug in [Unity to Godot](http://docs.godotengine.org/en/latest/getting_started/editor/unity_to_godot.html), where only Unity is displayed and you have to scroll to see Godot. 😛

Annotation 2020-04-26 154323

This PR fixes it:

Annotation 2020-04-26 154402

Xrayez:
Found some cases where `:kbd:` shortcuts get wrapped which is likely an issue:

Annotation 2020-04-26 150019

For other table usages, this displays fine in most cases (especially with number of columns < 4).

For that I've updated the PR with some changes, namely:
* skip keyboard shortcuts from wrapping;
* do not wrap cells in the first column as they are mostly non-verbose ("Description" columns are).

Result for this case:

Annotation 2020-04-26 160712

Meaning that, yeah you'll still have to scroll to see those editor-specific settings there, but they're not as important as the shortcuts themselves IMO, but as always, choose your own poison!

Some list of pages to see the difference:
http://docs.godotengine.org/en/latest/getting_started/editor/command_line_tutorial.html
http://docs.godotengine.org/en/latest/getting_started/editor/default_key_mapping.html
http://docs.godotengine.org/en/latest/getting_started/editor/unity_to_godot.html

PR #3454, opened 2020-04-27
Update class_color.rst - "contrasted()" desc.

homchom:
Original: "Returns the most contrasting color."
New: "Returns the color's complimentary color."

This change was made because `contrasted()` does not always return the most contrasting color due to its simple code. For example, the most contrasting color to black (0, 0, 0) is white (1, 1, 1), but `Color(0, 0, 0, 1).contrasted() = Color(0.5, 0.5, 0.5, 1)` (light gray).


Calinou:

homchom:
My bad, thanks for letting me know! I'll make a new one on the main repository later today.

PR #3458, opened 2020-04-27
Replace British spelling and fix various typos

pycbouh:
I've been told that we should stick to American English and not British English, so I've checked the docs for various "-sion" in place of "-zion" and "-our" instead of "-or". I've also touched up some typos and miscapitalizations along the way.

There may be more, but I'm not familiar with any tool that can help me here. We should probably look into adding appropriate rules to our CI spellchecker for British spelling occurrences and wrongly capitalized (or not capitalized at all) names (like YouTube, GitHub, PlayStation).


Calinou:
> We should probably look into adding appropriate rules to our CI spellchecker for British spelling occurrences and wrongly capitalized (or not capitalized at all) names (like YouTube, GitHub, PlayStation).

I don't think codespell can do this. Should we just perform a bunch of case-sensitive greps and fail if there is a match?

akien-mga:
CI is too slow... Merging as I need this for my WIP sync of godot-docs-l10n.

akien-mga:
Thanks!

pycbouh:
> Should we just perform a bunch of case-sensitive greps and fail if there is a match?

I fear, there might be a lot of false-positives (negatives?) even if we check against a short list of words that we know to be problematic. Shouldn't codespell check against a dictionary, preferably one language dictionary, like *EN-us*? All of these cases I touched in this PR were highlighted by a spellchecker plugin I use within VS Code, including wrong capitalization.

Calinou:
@pycbouh codespell uses its own dictionaries: https://github.com/codespell-project/codespell/tree/master/codespell_lib/data

I'm saying it might not be possible because I don't know if it can perform case-sensitive checks.

pycbouh:
I see. Well, I wouldn't go with a bunch of greps, because we'd have to make them context aware enough to not fail on URLs and meta markup within docs. This is not critical, after all, and can be reviewed on case by case basis when accepting new docs, until we can find a better solution.

PR #3461, opened 2020-04-27
Fix locale names

mhilbrunner:
This fixes the `hreflang` links for locale names for languages such as `zh_CN` by converting them to ISO 639-1 (`zh-cn`) where needed.

It also prepends "(DEV)" to the page title if built locally and not on ReadTheDocs, which I think @Calinou suggested.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -66,17 +66,6 @@ supported_languages = {
     "zh_CN": "Godot Engine (%s) 简体中文文档",
 }
 
-# Some language codes used on Weblate/RTD and in the docs URL differ,
-# so we need to remap them.
-languages_remap = {
-    "pt_BR": "pt-br",
-    "zh_CN": "zh-cn",
-}
-languages_url = sorted(
-    [l for l in supported_languages.keys() if l not in languages_remap.keys()]
-    + list(languages_remap.values())
-)
-
 language = os.getenv("READTHEDOCS_LANGUAGE", "en")
 if not language in supported_languages.keys():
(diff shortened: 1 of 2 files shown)


conf.py
```diff
@@ -133,11 +121,12 @@
     "github_version": "master",  # Version
     "conf_py_path": "/",  # Path in the checkout to the docs root
     "godot_inject_language_links": True,
-    "godot_docs_supported_languages": languages_url,
+    "godot_docs_supported_languages": list(supported_languages.keys()),
     "godot_docs_basepath": "https://docs.godotengine.org/",
     "godot_docs_suffix": ".html",
     "godot_default_lang": "en",
     "godot_canonical_version": "stable",
+    "godot_title_prefix": "" if on_rtd else "(DEV) "
```
Calinou, line 129: It might be worth adding a comment to explain why this is added:

```suggestion
    # Distinguish local development website from production website.
    # This prevents people from looking for changes on the production website after making local changes :)
    "godot_title_prefix": "" if on_rtd else "(DEV) "
```

akien-mga:
First commit looks good.

For the "DEV" change, it could also be done directly in `conf.py` as a suffix to `html_title`.

mhilbrunner:
That was actually the first version I did, but then its far enough in the title to easily be missed, which misses the point somewhat. This way, the title starts with "(DEV)", making it harder to miss.

The suggestions sound good, will do :)

mhilbrunner:
Done!

PR #3465, opened 2020-04-29
updated 2dmovment.rst

Rudyon:
so it works out of the box in gd3.x

### DIFF
--- a/tutorials/2d/2d_movement.rst
+++ b/tutorials/2d/2d_movement.rst
@@ -47,13 +47,13 @@ Add a script to the kinematic body and add the following code:
 
     func get_input():
         velocity = Vector2()
-        if Input.is_action_pressed('right'):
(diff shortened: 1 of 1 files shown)


cbscribe:
It does work "out of the box" because the first section describes how to set up input actions:
https://docs.godotengine.org/en/3.2/tutorials/2d/2d_movement.html#setup

Calinou:
I agree with @cbscribe here. It's generally advised not to use built-in inputs for gameplay purposes (unless you're just prototyping).

mhilbrunner:
I agree with the others, the tutorial makes sense as-is, and (mis)using the editor keybinds isn't optimal.

I'll close this, but if you want, adding a new PR that adds a line pointing out that the input actions were defined previously as Calinou suggested would be fine and welcome. :)

Still, thanks for taking the time to submit a PR!

PR #3467, opened 2020-04-30
Fix method name

VardenTheOne:
https://docs.godotengine.org/en/3.2/classes/class_editorspatialgizmoplugin.html#class-editorspatialgizmoplugin-method-create-handle-material


mhilbrunner:
Thank you!

PR #3478, opened 2020-05-01
Use `network_peer` as a property instead of using the setter method

Calinou:
Thanks to HaSa on IRC for pointing this out.


akien-mga:
Thanks!

PR #3482, opened 2020-05-02
Add a dedicated page for 3D rendering limitations

### LINKED ISSUE #2808: Document graphics rendering quirks (OpenGL, Vulkan)
As discussed in https://github.com/godotengine/godot/issues/17234

Often we run into "bugs" that are in actual fact just quirks relating to how OpenGL or computer graphics work in general. It would be useful to document some common ones in one place.

cc @zireael07

Calinou:
The "3D performance and limitations" page was renamed to reflect this change. It will specifically focus on 3D performance in the future.

This closes #2808.

### DIFF
--- a/tutorials/3d/3d_rendering_limitations.rst
+++ b/tutorials/3d/3d_rendering_limitations.rst
@@ -1,0 +1,125 @@
+.. _doc_3d_rendering_limitations:
+
+3D rendering limitations
+========================
+
+Introduction
+------------
+
+Due to their focus on performance, real-time rendering engines have many
+limitations. Godot's renderer is no exception. To work effectively with those
+limitations, you need to understand them.
+
+Texture size limits
+-------------------
+
+On desktops and laptops, textures larger than 8192×8192 may not be supported on
+older devices. You can check your target GPU's limitations on
+`GPUinfo.org <https://www.gpuinfo.org/>`__.
+
+Mobile GPUs are typically limited to 4096×4096 textures. Also, some mobile GPUs
+don't support repeating non-power-of-two-sized textures. Therefore, if you want
+your texture to display correctly on all platforms, you should avoid using
+textures larger than 4096×4096 and use a power of two size if the texture needs
+to repeat.
+
+Color banding
+-------------
+
+When using the GLES3 or Vulkan renderers, Godot's 3D engine renders internally
+in HDR. However, the rendering output will be tonemapped to a low dynamic range
+so it can be displayed on the screen. This can result in visible banding,
+especially when using untextured materials. This can also be seen in 2D projects
+when using smooth gradient textures.
+
+There are several ways to alleviate banding. Here are a few examples:
+
+- Bake some noise into your textures. This is mainly effective in 2D, e.g. for
+  vignetting effects.
+- Implement a debanding shader as a :ref:`screen-reading shader <doc_screen-reading_shaders>`.
+  Godot currently doesn't provide a built-in debanding shader, but this may be
+  added in a future release.
+
+.. seealso::
+
+    See `Banding in Games: A Noisy Rant <http://loopit.dk/banding_in_games.pdf>`__
+    for more details about banding and ways to combat it.
+
+Depth buffer precision
+----------------------
+
+To sort objects in 3D space, rendering engines rely on a *depth buffer* (also
+called *Z-buffer*). This buffer has a finite precision: 24-bit on desktop
+platforms, sometimes 16-bit on mobile platforms (for performance reasons). If
+two different objects end up on the same buffer value, then Z-fighting will
+occur. This will materialize as textures flickering back and forth as the camera
+moves or rotates.
+
+To make the depth buffer more precise over the rendered area, you should
+*increase* the Camera node's **Near** property. However, be careful: if you set
+it too high, players will be able to see through nearby geometry. You should
+also *decrease* the Camera node's **Far** property to the lowest permissible value
+for your use case, though keep in mind it won't impact precision as much as the
+**Near** property.
+
+If you only need high precision when the player can see far away, you could
+change it dynamically based on the game conditions. For instance, if the player
+enters an airplane, the **Near** property can be temporarily increased to avoid
+Z-fighting in the distance. It can then be decreased once the player leaves the
+airplane.
+
+Depending on the scene and viewing conditions, you may also be able to move the
+Z-fighting objects further apart without the difference being visible to the
+player.
+
+Transparency sorting
+--------------------
+
+In Godot, transparent materials are drawn after opaque materials. Transparent
+objects are sorted back to front before being drawn based on the Node3D's
+position, not the vertex position in world space. Due to this, overlapping
+objects may often be sorted out of order. To fix improperly sorted objects, tweak
+the material's :ref:`Render Priority <class_Material_property_render_priority>`
+property. This will force specific materials to appear in front or behind of
+other transparent materials. Even then, this may not always be sufficient.
+
+Some rendering engines feature *order-independent transparency* techniques to
+alleviate this, but this is costly on the GPU. Godot currently doesn't provide
+this feature. There are still several ways to avoid this problem:
+
+- Only make materials transparent if you actually need it. If a material only
+  has a small transparent part, consider splitting it into a separate material.
+  This will allow the opaque part to cast shadows and may also improve
+  performance.
+
+- If you want a material to fade with distance, use the StandardMaterial3D
+  distance fade mode **Pixel Dither** or **Object Dither** instead of
+  **PixelAlpha**. This will make the material opaque. This way, it can also
+  cast shadows.
+
+Multi-sample antialiasing
+-------------------------
+
+Multi-sample antialiasing (MSAA) takes multiple *coverage* samples at the edges
+of polygons when rendering objects. It does not increase the number of *color*
+samples used to render a scene. Here's what this means in practice:
+
+- Edges of meshes will be smoothed out nicely (as well as supersampling would).
+- Transparent materials that use *alpha testing* (1-bit transparency) won't be smoothed out.
+- Specular aliasing ("sparkles" that appear on reflective surfaces) won't be reduced.
+
+There are several ways to work around this limitation depending on your performance budget:
+
+- To make specular aliasing less noticeable, open the Project Settings and enable
+  **Rendering > Quality > Screen Space Filters > Screen Space Roughness Limiter**.
+  This filter has a moderate cost on performance. It should be enabled only if
+  you actually need it.
+
+- Enable FXAA in addition to (or instead of) MSAA. Since FXAA is a screen-space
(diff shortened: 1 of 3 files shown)


tutorials/3d/3d_rendering_limitations.rst
```diff
@@ -0,0 +1,125 @@
...
+
+Modern desktop/laptop GPUs can only render textures up to 16384×16384. If you
+need to display a larger texture for some reason, you'll need to split it into
+multiple smaller textures. Old GPUs like the GeForce 8/9 series are limited to
+8192×8192.
+
+Mobile GPUs are typically limited to 4096×4096 textures. Therefore, if you want
+your texture to display correctly on all platforms, you should avoid using
+textures larger than 4096×4096.
+
```
mhilbrunner, line 24: I think *very* recent GPUs like the NVIDIA RTX line may actually up max texture resolution to `32768x32768`. 

Also `Old GPUs like the GeForce 8/9 series are limited to 8192×8192` may be misleading from what I've found?
Geforce 9 would be from 2008. I can't find specs for those, but I can't imagine them supporting anything above `4096x4096`, if that.

NVIDIA Geforce 900, i.e. GTX 950 to 980Ti/TitanX from 2014 still seems to support only up to `4096x4096`.
The Geforce 10 line (2016), i.e. GTX 1030 to 1080 Ti finally seems to have upped this to `16384x16384`.

GPUs around the Geforce 9xx line and below seem to still be fairly popular according to the current [Steam hardware survey data](https://store.steampowered.com/hwsurvey/Steam-Hardware-Software-Survey-Welcome-to-Steam), maybe they even make up the majority (at least for non powerusers).

So maybe we should point out that you want to stay at or below `4096x4096` if you possibly want to support GPUs older than 3 years. (Which is probably the case for the majority of games, besides the next Crysis.)

Also maybe add a note about preferring power-of-two texture sizes for various reasons (like mipmapping/compression, and sometimes speed? Not an expert.), and about adding padding around textures to prevent some problems with texture bleeding.

tutorials/3d/3d_rendering_limitations.rst
```diff
@@ -0,0 +1,125 @@
...
+In OpenGL, 2D line drawing is implemented by the graphics driver. The maximum
+line thickness that can be drawn is implementation-defined. AMD and Intel
+graphics drivers only support 1 pixel-thick lines. NVIDIA graphics drivers allow
+drawing lines up to 5 pixels thick.
+
+Note that this limitation doesn't affect lines drawn by the :ref:`class_Line2D`
+node, since these are drawn using triangle strips instead of relying on OpenGL
+line drawing.
+
```
mhilbrunner, line 36: If this is about 2D line drawing, and `Line2D` isn't affected, where *does* this affect me as a Godot user?

tutorials/3d/3d_rendering_limitations.rst
```diff
@@ -0,0 +1,125 @@
...
+.. seealso::
+
+    See `Banding in Games: A Noisy Rant <http://loopit.dk/banding_in_games.pdf>`__
+    for more details about banding and ways to combat it.
+
+Depth buffer precision
+----------------------
+
+To sort objects in 3D space, rendering engines rely on a *depth buffer*. This
```
mhilbrunner, line 62:
```suggestion
To sort objects in 3D space, rendering engines rely on a *depth buffer* (sometimes called *Z-buffer*). This
```

tutorials/3d/3d_rendering_limitations.rst
```diff
@@ -0,0 +1,125 @@
...
+
+To make the depth buffer more precise over the rendered area, you should
+*increase* the Camera node's **Near** property. However, be careful: if you set
+it too high, players will be able to see through nearby geometry.
+
+If you only need high precision when the player can see far away, you could
+change it dynamically based on the game conditions. For instance, if the player
+enters an airplane, the **Near** property can be temporarily increased to avoid
+Z-fighting in the distance. It can then be decreased once the player leaves the
+airplane.
```
mhilbrunner, line 76:
```suggestion
airplane.

Sometimes, it may also be possible to simply move the Z-fighting objects farther apart without that being visible to the player, depending on the scene.
```

mhilbrunner:
Also, if @clayjohn or someone more versed in rendering than me wants to chime in, that'd be great. :P

tutorials/3d/3d_rendering_limitations.rst
Calinou: This is mainly relevant for CanvasItem's `draw_line()` methods (and `draw_multiline()`, `draw_polyline()`, ...).

tutorials/3d/3d_rendering_limitations.rst
Calinou: I had a GeForce 9600M back in the day. I remember seeing its maximum texture size being set to 8192x8192 according to OpenGL.

Either way, I should make a test project and send it to people with different hardware configurations to confirm this.

mhilbrunner: Yeah, you might be right :)

It seems very hard to find exact specs online, so I could only search for things like `"GTX 950" "Maximum Texture Dimension Size"`, which yields to results like [these](https://github.com/strands-project/strands_3d_mapping/issues/103):
>Device 0: "GeForce GTX 950"
>...
>Maximum Texture Dimension Size (x,y,z) 1D=(65536), 2D=(65536, 65536), 3D=(4096, 4096, 4096)

Or [these](http://cuda-memo.blogspot.com/2015/03/geforce-gtx-970-devicequery.html):
>Device 0: "GeForce GTX 970"
>...
>Maximum Texture Dimension Size (x,y,z)         1D=(65536), 2D=(65536, 65536), 3D=(4096, 4096, 4096)

tutorials/3d/3d_rendering_limitations.rst
mhilbrunner: I suggest adding a note about that then, because I wasn't sure where it applied. :)

Calinou: Actually, I'm not even sure if this is relevant in Godot's context. Godot will automatically use rectangles instead of lines for lines thicker than 1 pixel.

cc @lawnjelly, as he worked on https://github.com/godotengine/godot/pull/38417.

tutorials/3d/3d_rendering_limitations.rst
```diff
@@ -0,0 +1,130 @@
...
+  distance fade mode **Pixel Dither** or **Object Dither** instead of
+  **PixelAlpha**. This will make the material opaque. This way, it can also
+  cast shadows.
+
+Multi-sample antialiasing
+-------------------------
+
+Multi-sample antialiasing (MSAA) works by multiplying the number of *coverage*
+samples. It does not multiply the number of *color* samples used to render a
+scene. Here's what this means in practice:
```
clayjohn, line 110: I wouldn't say that it multiples the samples. Although that is correct in a sense, I think it will confuse users.
```suggestion
Multi-sample antialiasing (MSAA) takes multiple *coverage* samples at the
edges of polygons when rendering objects. It does not increase the number
of *color* samples used to render a scene. Here's what this means in practice:
```

tutorials/3d/3d_rendering_limitations.rst
```diff
@@ -0,0 +1,130 @@
...
+To sort objects in 3D space, rendering engines rely on a *depth buffer* (also
+called *Z-buffer*). This buffer has a finite precision: 24-bit on desktop
+platforms, sometimes 16-bit on mobile platforms (for performance reasons). If
+two different objects end up on the same buffer value, then Z-fighting will
+occur. This will materialize as textures flickering back and forth as the camera
+moves or rotates.
+
+To make the depth buffer more precise over the rendered area, you should
+*increase* the Camera node's **Near** property. However, be careful: if you set
+it too high, players will be able to see through nearby geometry.
```
aaronfranke, line 71: It depends on both the Near and Far properties, so both should be noted. I forget exactly what the relationship is (order of magnitude?), but that should be noted too.

tutorials/3d/3d_rendering_limitations.rst
lawnjelly: Older phones can be limited to 2048x2048 textures. Also on mobile, fill rate and texture bandwidth can be key as it uses lots of power so they do everything to reduce this. So using low res textures is highly recommended where possible. Also consider using mipmap nearest instead of trilinear filtering. And texture compression presumably helps loads too.

tutorials/3d/3d_rendering_limitations.rst
lawnjelly: As far as I can see for 2D TYPE_LINE is drawn with GL_LINES for size 1 and below, above that it uses a combination of rects and an optional 4 GL_LINES around the edges for anti-aliasing.

The option `draw_lines_as_rects` that I added just always uses rects and removes any anti aliasing. So it should run faster on hardware that has poor GL_LINES support, at the cost of not looking as good.

TYPE_POLYLINE I'm not that familiar with and it is hardly used in the editor, but it seems to use a bunch of polys followed by a GL_LINE_LOOP, or a bunch of GL_LINES, depending on whether there are triangles.

It's hard to give that much advice on lines, except to say anything using GL_LINES has the potential to be very slow on some hardware. It is regarded as 'ye olde' OpenGL and some have given up on supporting it. On the whole I've heard it recommended to only use GL_LINES for debugging. For drawing UI lines, and polys with lines around in modern games I get the impression using polys and shaders is a more favoured approach.

It is possible in the future we can modify the TYPE_POLYLINE to do this a bit more sensibly with `draw_lines_as_rects` set. And have a look at line shaders.

So really the whole lines area is subject to change, it might become quite safe to use in the future, but it's could be a little bit risky at the moment. So far the only hardware we've been hearing about with possible slow lines is Mac, but that is still waiting for confirmation.

Also, it may be slow on e.g. phones but we don't know as few people write games with lines.

lawnjelly:
We could probably do with a section on mobile, as the considerations are different because a lot of these use tile renderers. Certain things may have a disproportionate effect on performance on mobile. It's worth doing some testing to ascertain this. I suspect screen reading shaders might be a hit.

Also the limitations due to precision in shaders on mobile.

Generally anything that goes outside the old style fixed function pipelines is pretty dodgy. Although mobile supposedly can do all funky shaders, the support in reality is patchy. This is part of the reason I think we should think about offering some 'fixed function' shaders for mobile, for GLES2, that can be widely tested.

Calinou:
@lawnjelly Some mobile optimizations are already covered in [3D performance and limitations](https://docs.godotengine.org/en/latest/tutorials/3d/3d_performance_and_limitations.html).

tutorials/3d/3d_rendering_limitations.rst
Calinou: I removed the 2D line drawing section since it's not relevant in Godot's context.

mhilbrunner:
Looks good to me, I'll merge. Thanks for this, it's a great addition to the docs :)

PR #3483, opened 2020-05-02
Create Sphinx extension to generate HTML meta description tags

mhilbrunner:
This creates a new Sphinx extension to automatically generate and add HTML meta description tags, primarily to improve search engine result pages for Godot.

----
**Rationale**

Currently, our docs pages come without HTML meta description tags. (`<meta name="description" content="...">`)

Because of this, search engines try to come up with their own descriptions. If they match against only a specific place on the page, they generally show the surrounding context, which works fine.

However, often they just try to autogenerate a description by taking the first text(s) on the pages, which for the Godot docs page leads to bad results, as it hits the navigation first instead of the content for our markup:
godot-description

Sphinx has a way to manually specify meta tags per document, and the ReadTheDocs theme seems to support adding these. However, manually adding a description *for each page* would be a lot of work for not much gain, as the content is already there.

This adds a Sphinx extension that hooks into the generation of each page, tries to sample the first *content* sentences and builds a description from that.

### DIFF
--- a/_extensions/godot_descriptions.py
+++ b/_extensions/godot_descriptions.py
@@ -1,0 +1,117 @@
+# -*- coding: utf-8 -*-
+"""
+    godot_descriptions
+    ~~~~~~~~~~~~~~~~~~
+
+    Sphinx extension to automatically generate HTML meta description tags
+    for all pages. Also comes with some special support for Godot class docs.
+
+    :copyright: Copyright 2020 by The Godot Engine Community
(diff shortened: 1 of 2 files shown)


akien-mga:
Looks good!

Could you format the python file with `black`? https://github.com/psf/black

mhilbrunner:
Done!

akien-mga:
Thanks!

PR #3494, opened 2020-05-04
Improve docs on TIME

RandomShaper:
And improve TIME.

(OMG, after two messed up PRs, I've finally been able to do it.)

### DIFF
--- a/tutorials/shading/shading_reference/canvas_item_shader.rst
+++ b/tutorials/shading/shading_reference/canvas_item_shader.rst
@@ -48,11 +48,12 @@ Global built-ins
 
 Global built-ins are available everywhere, including custom functions.
 
-+-------------------+--------------------------+
-| Built-in          | Description              |
-+===================+==========================+
-| in float **TIME** | Global time, in seconds. |
-+-------------------+--------------------------+
++-------------------+-----------------------------------------------------------------------------+
+| Built-in          | Description                                                                 |
++===================+=============================================================================+
+| in float **TIME** | Global time, in seconds.                                                    |
+|                   | It's subject to the rollover setting (which is 3,600 -1 hour- by default).  |
++-------------------+-----------------------------------------------------------------------------+
 
 Vertex built-ins
 ^^^^^^^^^^^^^^^^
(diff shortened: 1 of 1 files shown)


tutorials/shading/shading_reference/canvas_item_shader.rst
```diff
@@ -48,11 +48,16 @@ Global built-ins
...
-+===================+==========================+
-| in float **TIME** | Global time, in seconds. |
-+-------------------+--------------------------+
++-----------------------+-----------------------------------------------------------------------------+
+| Built-in              | Description                                                                 |
++=======================+=============================================================================+
+| in float **TIME**     | Global time, in seconds.                                                    |
+|                       | It's subject to the rollover setting (which is 3,600 -1 hour- by default).  |
++-----------------------+-----------------------------------------------------------------------------+
+| in vec4 **MODULATE**  | Final modulate color.                                                       |
```
clayjohn, line 57: The way you have defined ``MODULATE`` it is not a global built-in, it is a built in that is available in all three shader functions. Global built-ins are a special type that is accessible everywhere, including custom functions. They are only available in master, and are specified with "global" keyword. see https://github.com/godotengine/godot/pull/37166

RandomShaper: Understood.

But now I'm confused about  _master_ in this repo being for 3.2. Is that no longer true? If that's the case, I guess I can leave the changes to TIME here and move on MODULATE to a separate PR for 3.2.

clayjohn: I think master in this repo is for 4.0 because there is a separate 3.2 branch. So yea, maybe leave this PR with the time changes and then make another PR with the modulate changes for 3.2.

mhilbrunner: Yeah, `master` here is for 4.0.

RandomShaper: PR updated to include only the TIME stuff.

RandomShaper: BTW, the PR template says that _master_ is fo 3.0. Maybe it needs an update.

akien-mga:
Thanks!

PR #3496, opened 2020-05-05
Fix GDNative C++ SConstruct for OSX

vonagam:
Without `-std=c++17` compiling will fail with errors shown [here](https://godotengine.org/qa/65616/problems-compiling-gdnative-c-example-on-osx).

### DIFF
--- a/tutorials/plugins/gdnative/files/cpp_example/SConstruct
+++ b/tutorials/plugins/gdnative/files/cpp_example/SConstruct
@@ -49,22 +49,23 @@ if env['platform'] == '':
 if env['platform'] == "osx":
     env['target_path'] += 'osx/'
     cpp_library += '.osx'
+    env.Append(CCFLAGS=['-arch', 'x86_64'])
+    env.Append(CXXFLAGS=['-std=c++17'])
+    env.Append(LINKFLAGS=['-arch', 'x86_64'])
     if env['target'] in ('debug', 'd'):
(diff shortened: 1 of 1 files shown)


tutorials/plugins/gdnative/files/cpp_example/SConstruct
```diff
@@ -50,10 +50,10 @@ if env['platform'] == "osx":
     env['target_path'] += 'osx/'
     cpp_library += '.osx'
     if env['target'] in ('debug', 'd'):
-        env.Append(CCFLAGS=['-g', '-O2', '-arch', 'x86_64'])
+        env.Append(CCFLAGS=['-g', '-O2', '-arch', 'x86_64', '-std=c++17'])
```
akien-mga, line 53: `CCFLAGS` are the flags for both the C and C++ compilers, so we can't pass `-std=c++17` to the C compiler (Clang would normally raise an error, but I guess Apple Clang is more lenient if it works for you).

This should be:
```
env.Append(CXXFLAGS=['-std=c++17'])
```
to only affect C++.

vonagam: How about now? Moved common flags from conditionals for osx and linux (like it is for windows).

vonagam:
Anything else holding this back?

akien-mga:
Thanks!

PR #3500, opened 2020-05-06
Added BBcode tags for Table and Cell

Vasanthan-:
These tags are inbuilt with Godot but never mentioned in the documentation.


mhilbrunner:
Thank you and congrats on your firstr contribution!

PR #3508, opened 2020-05-08
Mention Object bracket property access and dynamic property indexing

### LINKED ISSUE #3506: Missing documentation on the "bracket" property access on objects
**Your Godot version:**
3.x+

**Issue description:**
There were questions in community channels about accessing various object properties using a syntax like this:
```swift
current_scene[k]
```

While the intention may be obvious to an experienced programmer, this can be confusing for others. It seems the docs don't really mention a lot about the bracket access to various properties throughout the API and GDScript language (see #2952). In this particular case I was only able to find two tangential mentions of bracket access in the [GDScript Basics docs](https://docs.godotengine.org/en/stable/getting_started/scripting/gdscript/gdscript_basics.html), one in the operator precedence table and the other in `Dictionary` property access details.

image
image

There is no mention that you can access any object properties this way, and that you can use a variable to store the key for the property.

https://docs.godotengine.org/en/stable/getting_started/scripting/gdscript/gdscript_basics.html

Calinou:
This closes #3506.

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_basics.rst
+++ b/getting_started/scripting/gdscript/gdscript_basics.rst
@@ -514,6 +514,20 @@ assign to it::
     d[4] = "hello" # Add integer 4 as a key and assign the String "hello" as its value.
     d["Godot"] = 3.01 # Add String "Godot" as a key and assign the value 3.01 to it.
 
+    var test = 4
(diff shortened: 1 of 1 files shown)


mhilbrunner:
Thanks!

PR #3509, opened 2020-05-08
Remove uninformative editor development Introduction page

Calinou:
I think this speaks for itself :slightly_smiling_face:

Maybe we could write something more descriptive there, but I'm not sure what could be added exactly.


mhilbrunner:
It does, thanks :P

PR #3513, opened 2020-05-09
Godot: how to change scene passing parameters

rjack:
A common beginner question is how to pass information between scenes. A common answer is to just use globals, but this could easily lead to problems and obscure global state management bugs.

I updated the Custom Scene Switcher tutorial with a new version of the `goto_scene` function that accepts a dictionary argument, whose values are used to initialize properties in the newly instantiated scene.

I also updated the _autoload.zip_ example accordingly.

Unfortunately, I don't know C# so the relevant snippets need to be updated. Hopefully it should be trivial to update them for anyone with C# knoweledge, if this proposal is accepted.


mhilbrunner:
Hey! Thanks for this.

I'm not really a fan of making the current tutorial more complicated by teaching an additional thing.

IMO it would be good to have this as a second step, building on the first.

cc @cbscribe

cbscribe:
Agreed. This document should be about the autoload feature and how to use it. A scene switcher doesn't really belong in the first place - this should be a separate page.

mhilbrunner:
I'm closing this PR for now, as discussed above it would be a better fit for a standalone tutorial or one building on this page, instead of adding to it.

As we're debating restructuring the docs for Godot 4.0 and trimming down the scope, note that it may be worthwhile to hold off on reworking this tutorial until we're clear on whether that fits into the Godot 4.0 docs - however, you may want to share what you submitted here on a blog or another place with the comunity, as it *is* pretty useful!

You can follow the discussion around the Godot 4.0 docs rework here:
https://github.com/godotengine/godot-docs/issues/3390

Thanks again for taking the time to create this PR!

rjack:
Makes sense, I changed the PR title hoping it can be easily found with search engines by those searching about the topic.

PR #3514, opened 2020-05-09
correct mistake of the document about Animation

wrrwrr111:


cbscribe:
Thanks - nice catch!

akien-mga:
```

.. DO NOT EDIT THIS FILE, but the Animation.xml source instead.
```

PR #3518, opened 2020-05-10
Added note to TileMap.get_used_cells_by_id()

tylerecouture:
Added: Note: The tile ``id`` is referred to as the tile ``index`` in other methods. 
https://github.com/godotengine/godot-proposals/issues/814

### DIFF
--- a/classes/class_tilemap.rst
+++ b/classes/class_tilemap.rst
@@ -556,7 +556,9 @@ Returns a :ref:`Vector2<class_Vector2>` array with the positions of all cells co
 
 - :ref:`Array<class_Array>` **get_used_cells_by_id** **(** :ref:`int<class_int>` id **)** const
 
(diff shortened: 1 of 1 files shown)


akien-mga:
See the comment in the file's header:
```
.. DO NOT EDIT THIS FILE, but the TileMap.xml source instead.
```

There's already a PR on the engine repository to address that proposal: godotengine/godot#38635.

tylerecouture:
That renames the method for a future version.  I'm just trying to clarify the docs for the current version.

akien-mga:
That makes sense indeed for 3.2. This should be changed in the source file here: https://github.com/godotengine/godot/blob/master/doc/classes/TileMap.xml
Otherwise changes made here would be lost next time the class reference is generated.

Calinou:
For reference, this was superseded by https://github.com/godotengine/godot/pull/38686.

PR #3521, opened 2020-05-11
Removed redundant code recalculating velocity

mukul29:
1. 'velocity' was being normalized and multiplied by 'speed' twice making it redundant.
2. The else statement was re-initializing velocity which isn't required.


mhilbrunner:
cc @cbscribe

cbscribe:
Thanks!

PR #3525, opened 2020-05-12
Update the Editor icons optimization guide to follow current usage

Calinou:
We have transitioned from scour to svgcleaner lately due to its better performance and ease of installation.


mhilbrunner:
Thanks!

PR #3535, opened 2020-05-14
Improve Windows compiling instructions

skyace65:
This fixes some issues I ran into when compiling the engine for the first time.

-The information about installing Python is now its own small section
-Added instructions for installing SCons. As someone who hasn't had to install Python stuff before this section gave me issues, I just assumed you needed to double click the install.py file and it would automatically install.


mhilbrunner:
The usual way to install Python dependencies is via the Python package manager `pip`:

`pip install scons`

Indeed, that is what they recommend in their docs:

https://scons.org/doc/production/HTML/scons-user/ch01s02.html

Calinou:
I agree, we should just document the need to run `pip install scons` as @mhilbrunner stated. `scoop install scons` might also work if you have Scoop installed, but I don't know if it can use Python 3 instead of Python 2.

skyace65:
Ok, now it tells you to run the pip command.

mhilbrunner:
Thank you!

PR #3537, opened 2020-05-14
Add a note about case sensitivity in Project organization

Calinou:
See https://github.com/godotengine/godot/issues/19592.

## Proof

The future is now. :upside_down_face:

image


mhilbrunner:
Thank you!

PR #3540, opened 2020-05-15
PR workflow: Improve instructions around updating, amending and rebasing

akien-mga:
I was young and careless when documenting `git rebase -i HEAD~2`,
I since found out that it's much cleaner and safer to rebase on
the upstream branch directly.

### DIFF
--- a/community/contributing/pr_workflow.rst
+++ b/community/contributing/pr_workflow.rst
@@ -60,9 +60,11 @@ The branches on the Git repository are organized as follows:
    As a rule of thumb, the last stable branch is maintained until the next
    major version (e.g. the ``3.0`` branch was maintained until the release of
    Godot 3.1).
-   If you want to make PRs against a maintained stable branch, you will have
-   to check if your changes are also relevant for the ``master`` branch.
--  There might be feature branches at time, usually meant to be merged into
+   If you want to make PRs against a maintained stable branch, please check
+   first if your changes are also relevant for the ``master`` branch, and if so
+   make the PR for the ``master`` branch in priority. Release managers can then
+   cherry-pick the fix to a stable branch if relevant.
+-  There might occasionally be feature branches, usually meant to be merged into
    the ``master`` branch at some time.
 
 Forking and cloning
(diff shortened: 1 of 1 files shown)


community/contributing/pr_workflow.rst
```diff
@@ -187,42 +197,55 @@ To ensure there won't be conflicts between the feature you develop and the
...
+be the only difference with the upstream ``master`` branch.
+
+While rebasing, conflicts may arise if your commits modified code that has been
+changed in the upstream branch in the meantime. If that happens, Git will stop at
+the conflicting commit and will ask you to resolve the conflicts. You can do so
+with any text editor, then stage the changes (more on that later), and proceed with
+``git rebase --continue``. Repeat the operation if later commits have conflicts too,
+until the rebase operation completes.
+
+If you're unsure about what is going on during a rebase and you panick (no worry,
```
mhilbrunner, line 216:
```suggestion
If you're unsure about what is going on during a rebase and you panic (no worry,
```

community/contributing/pr_workflow.rst
```diff
@@ -187,42 +197,55 @@ To ensure there won't be conflicts between the feature you develop and the
...
+          While this is a valid workflow and the default behavior of ``git pull``,
+          merge commits within PRs in our PR workflow. We only use them when merging
+          PRs into the upstream branch.
+
+          The philosophy is that a PR should represent the final stage of the changes
+          made to the codebase, and we are not interested in mistakes and fixes that
+          would have been done in intermediate stages before merging.
+          Git gives us great tools to "rewrite the history" and make it as if we got
+          things right the first time, and we're happy to use it to ensure that
+          changes are easy to review and understand long after they have been merged.
```
mhilbrunner, line 234:
```suggestion
```

I suggest removing this sentence; the previous one does enough to convey the reasoning, IMO. As is, this sentence does encourage "rewriting history" a little too much in my opinion, and would need an added caveat that while that's fine in cases such as this were said history doesn't go back that far and isn't public yet, it needs caution (if done at all) in other cases, because it can really mess things up. But... that then becomes pretty long :)

mhilbrunner:
Thanks :)

PR #3543, opened 2020-05-16
Expand Resource section

mechPenSketch:
Added how you can export classes that inherits from Resource as a convenience.

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_exports.rst
+++ b/getting_started/scripting/gdscript/gdscript_exports.rst
@@ -126,8 +126,8 @@ Examples
     # In the Inspector, you can then drag and drop a resource file
     # from the FileSystem dock into the variable slot.
     
-    # Opening resource from the FileSystem will drop-down an
-    # extremely long lists of new classes to create, however.
+    # Opening the inspector dropdown may result in an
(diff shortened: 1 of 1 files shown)


getting_started/scripting/gdscript/gdscript_exports.rst
```diff
@@ -125,6 +125,13 @@ Examples
     export(Resource) var resource
     # In the Inspector, you can then drag and drop a resource file
     # from the FileSystem dock into the variable slot.
+    
+    # Opening resource from the FileSystem will drop-down an
+    # extremely long lists of new classes to create, however.
+    # Therefore, if you specify an extension of Resource such as:
+    export(AnimationNode) var resource
+    # The drop-down menu will be limited to AnimationNode and all
+    # its inherited classes.
```
mhilbrunner, line 134:
```suggestion
    # Opening the inspector dropdown may result in an
    # extremely long list of possible classes to create, however.
    # Therefore, if you specify an extension of Resource such as:
    export(AnimationNode) var resource
    # The drop-down menu will be limited to AnimationNode and all
    # its inherited classes.
```

mhilbrunner:
Looks good, thank you for your contribution! :)

PR #3544, opened 2020-05-16
Tweak caption styling to be closer to typical captions

Calinou:
### Before

Example caption with text below

### After

Example caption with text below


akien-mga:
Thanks!

PR #3546, opened 2020-05-16
Split the GDScript warning system documentation into its own page

Calinou:
It's not strictly related to static typing, as it can work without it (e.g. to signal unreachable code).

This should also make it a bit easier to find.


mhilbrunner:
Thanks :)

PR #3547, opened 2020-05-16
Center the main pane on wide displays

### LINKED ISSUE #3139: Centering the text pane in docs.godotengine.org
To make the docs more readable on widescreen monitors, the docs were centered. However, this also includes the sidebar, which means the text itself isn't centered. This makes it harder to read on smaller screens.

Centered

It would be easier to read if the text pane was perfectly centered, and the sidebar was directly to the left of it.

Calinou:
Thanks to @ericrybick for the suggestion.

PS: I didn't understand what this part does:

> also add `padding-left: calc(100vw - 100%);` to `:root` to avoid scroll-bar repositioning.

I couldn't notice any difference in Firefox after applying it.

### Before (both panes are centered as a group)

Godot docs homepage

### After (main pane is centered)

Godot docs homepage

This closes #3139.

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -173,6 +173,11 @@
     }
 }
 
+html {
+    /* Hack to prevent taking the scroll bar into account when centering the main pane */
+    padding-left: calc(100vw - 100%);
+}
+
 body,
 h1,
 h2,
@@ -440,8 +445,9 @@ footer,
 
 @media only screen and (min-width: 769px) {
     .wy-body-for-nav {
-        /* Center the page on wide displays for better readability */
-        max-width: 1100px;
+        /* Center the page's main pane on wide displays for better readability */
+        /* `max-width` is (navbar width + main pane width) */
+        max-width: 1400px;
         margin: 0 auto;
     }
 }


ericrybick:
> PS: I didn't understand what this part does:
> 
> > also add `padding-left: calc(100vw - 100%);` to `:root` to avoid scroll-bar repositioning.
> 
> I couldn't notice any difference in Firefox after applying it.

On some browser, if a content overflow happens and the scrollbar appears, the width of the scrollbar is subtracted from the browser window width. With the mentioned "fix/hack", you can avoid that. This will ensure, that a website using it, will be perfectly centered on your screen if the browser window is maximized.

Here I took two screenshots, one with the hack and another without. I placed a perfectly centered vertical line on them. As you can see, in the first one (which does not include the fix), the content is shifted to the left by the width of the scroll-bar.

[The browser support for calc() seems to be pretty good.](https://caniuse.com/#feat=calc)

centeredNo

centeredYes

Calinou:
@ericrybick Thanks for clarifying it! I added the hack to the pull request.

mhilbrunner:
Tested, works fine; however, see discussion in https://github.com/godotengine/godot-docs/issues/3139 - after I've shown this to two people that are more design-oriented than myself, and both have preferred the current layout without further info, I'm hesistant to see this as a clear-cut improvement.

mhilbrunner:
Closing for now, see https://github.com/godotengine/godot-docs/issues/3139#issuecomment-629811756 for reasoning. Thanks for the work! I'd be happy to merge this if we can see some demand, it *does* work, but it's hard to gauge for me whether it's an improvement, or just a *change*.

Thanks for the PR!

PR #3549, opened 2020-05-17
Fixed typo in "logic preferences" best practices file

FeatherAntennae:
Changed "then" into "when" on line 19.


mhilbrunner:
Thank you!

PR #3553, opened 2020-05-17
Remove outdated notice at the end of Background loading

Calinou:
punto is generally not available on IRC for support questions.


mhilbrunner:
Thank you!

PR #3554, opened 2020-05-18
Describe a way to override build options via a file and SCONSFLAGS

Xrayez:
Sources: 
- godotengine/godot#38821, 
- https://github.com/godotengine/godot/pull/36922#issuecomment-599775440

Also modified `optimizing_for_size` by showing an alternative way to pass in a list of disabled modules via `custom.py`, but it may be better to make this the recommended way (there's too many modules to be passed to the command line).


akien-mga:
Thanks!

PR #3557, opened 2020-05-18
Add documentation about exporting for dedicated servers

Calinou:
This also documents how to build dedicated server binaries.

**Note:** The part about compiling will need to be updated for Godot 4.0 once the server platform is usable again.

This closes https://github.com/godotengine/godot/issues/30531.

### DIFF
--- a/development/compiling/compiling_for_linuxbsd.rst
+++ b/development/compiling/compiling_for_linuxbsd.rst
@@ -92,7 +92,7 @@ Start a terminal, go to the root dir of the engine source code and type:
 
 ::
 
-    user@host:~/godot$ scons -j8 platform=linuxbsd
+    scons -j8 platform=linuxbsd
 
 A good rule of thumb for the ``-j`` (*jobs*) flag, is to have at least as many
 threads compiling Godot as you have cores in your CPU, if not one or two more.
@@ -109,7 +109,7 @@ manager.
 
     ::
 
-        user@host:~/godot$ scons platform=linuxbsd use_llvm=yes
+        scons platform=linuxbsd use_llvm=yes
 
     Using Clang appears to be a requirement for OpenBSD, otherwise fonts
     would not build.
@@ -127,6 +127,19 @@ manager.
           and official releases, you can enable
           :ref:`doc_data_paths_self_contained_mode` by creating a file called
           ``._sc_`` or ``_sc_`` in the ``bin/`` folder.
+
+Compiling a headless/server build
+---------------------------------
+
+To compile a *headless* build which provides editor functionality to export
+projects in an automated manner, use::
+
+    scons -j8 platform=server tools=yes target=release_debug
+
+To compile a *server* build which is optimized to run dedicated game servers,
+use::
+
+    scons -j8 platform=server tools=no target=release
 
 Building export templates
 -------------------------
@@ -146,15 +159,15 @@ following parameters:
 
 ::
 
-    user@host:~/godot$ scons platform=linuxbsd tools=no target=release bits=32
-    user@host:~/godot$ scons platform=linuxbsd tools=no target=release_debug bits=32
+    scons platform=linuxbsd tools=no target=release bits=32
+    scons platform=linuxbsd tools=no target=release_debug bits=32
 
 -  (64 bits)
 
 ::
 
-    user@host:~/godot$ scons platform=linuxbsd tools=no target=release bits=64
-    user@host:~/godot$ scons platform=linuxbsd tools=no target=release_debug bits=64
+    scons platform=linuxbsd tools=no target=release bits=64
+    scons platform=linuxbsd tools=no target=release_debug bits=64
 
 Note that cross-compiling for the opposite bits (64/32) as your host
 platform is not always straight-forward and might need a chroot environment.
@@ -197,7 +210,7 @@ the default GCC + GNU ld setup:
 To do so, install Clang and the ``lld`` package from your distribution's package manager
 then use the following SCons command::
 
-    user@host:~/godot$ scons platform=linuxbsd use_llvm=yes use_lld=yes
+    scons platform=linuxbsd use_llvm=yes use_lld=yes
 
 It's still recommended to use GCC for production builds as they can be compiled using
 link-time optimization, making the resulting binaries smaller and faster.
--- a/development/compiling/compiling_for_osx.rst
+++ b/development/compiling/compiling_for_osx.rst
@@ -51,10 +51,23 @@ To create an ``.app`` bundle like in the official builds, you need to use the
 template located in ``misc/dist/osx_tools.app``. Typically, for an optimized
 editor binary built with ``scons p=osx target=release_debug``::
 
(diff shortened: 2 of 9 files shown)


development/compiling/compiling_for_linuxbsd.rst
```diff
@@ -128,6 +128,19 @@ manager.
           :ref:`doc_data_paths_self_contained_mode` by creating a file called
           ``._sc_`` or ``_sc_`` in the ``bin/`` folder.
 
+Compiling a headless/server build
+---------------------------------
+
+To compile a *headless* build which provides editor functionality to export
+projects in an automated manner, use::
+
+    scons -j8 platform=server target=release_debug
```
akien-mga, line 137: I'd make `tools=yes` explicit. Also previous commands use `user@host:~/godot$` as prompt so it would be worth to use it to stay consistent (or remove it everywhere).

development/compiling/compiling_for_osx.rst
```diff
@@ -56,6 +56,19 @@ editor binary built with ``scons p=osx target=release_debug``::
     user@host:~/godot$ cp bin/godot.osx.tools.64 Godot.app/Contents/MacOS/Godot
     user@host:~/godot$ chmod +x Godot.app/Contents/MacOS/Godot
 
+Compiling a headless/server build
+---------------------------------
+
+To compile a *headless* build which provides editor functionality to export
+projects in an automated manner, use::
+
+    scons platform=server target=release_debug --jobs=$(sysctl -n hw.logicalcpu)
```
akien-mga, line 65: Same comments as for Linux.

getting_started/workflow/export/exporting_for_dedicated_servers.rst
```diff
@@ -0,0 +1,83 @@
+.. _doc_exporting_for_dedicated_servers:
+
+Exporting for dedicated servers
+===============================
+
+If you want to run a dedicated server for your project on a machine that doesn't
+have a GPU, you'll need to use a server build of Godot.
```
akien-mga, line 7: It can also be a setup with a GPU but no display server.

getting_started/workflow/export/exporting_for_dedicated_servers.rst
```diff
@@ -0,0 +1,83 @@
...
+``--main-pack`` command-line argument::
+
+    ./godot-server --main-pack my_project.pck
+
+Next steps
+----------
+
+On Linux, to make your dedicated server restart after a crash or system reboot,
+you can
+`create a systemd service <https://medium.com/@benmorel/creating-a-linux-service-with-systemd-611b5c8b91d6>`__.
```
akien-mga, line 76: Isn't there a more official resource for that than a medium post?

Calinou: There's a reference [here](https://www.freedesktop.org/software/systemd/man/systemd.service.html), but I think it's quite verbose and not really approachable for non-specialists.

mhilbrunner:
Looks good to me (besides the things Akien already pointed out), and I'm not sure if we should suggest `-j8` by default. If we do that elsewhere keep it, otherwise I'd drop it I think :)

sesenta:
omg, i almost go crazy trying finding information about this, i spent like 3 hours before found this, you documentation is too good, i made this : [a little bit information i found](https://github.com/sesenta/godot-docs/blob/master/tutorials/using%20Godot%20headless%20and%20Godot%20server)
i'm not speak english and i never before try to help with github, if something that is useful please add, thanks for your effort.

Calinou:
@sesenta I think this should be fixed on the engine side so it can work transparently. The Server platform should probably attempt to load X11 GDNative libraries if no Server library was found.

mhilbrunner:
> This is already done elsewhere on the same page, so I kept it for consistency. We could recommend `-j$(nproc)` on Linux in a future PR (we already use the macOS equivalent on the Compiling for macOS page).

Keep it then, sounds fine :)

development/compiling/compiling_for_linuxbsd.rst
Calinou: I removed the prompt everywhere for consistency. This makes it easier to copy-paste commands as you can now triple-click them to select the whole line (without selecting the prompt at the same time).

tutorials/networking/high_level_multiplayer.rst
```diff
@@ -420,3 +420,17 @@ This may not make much sense for an area-of-effect case like the bomb, but in ot
...
+Exporting for dedicated servers
+-------------------------------
+
+Once you've made a multiplayer game, you may want to export it to run it on
+a dedicated server with no GPU available. See
+:ref:`doc_exporting_for_dedicated_servers` for more information.
+
+.. note::
+
+    The code samples in this page aren't designed to run from a dedicated
```
mhilbrunner, line 433:
```suggestion
    The code samples on this page aren't designed to run on a dedicated
```

mhilbrunner:
Thank you! :)

development/cpp/custom_modules_in_cpp.rst
```diff
@@ -387,8 +387,8 @@ during runtime with the ``LD_LIBRARY_PATH`` environ variable:
 
 .. code-block:: shell
 
-    user@host:~/godot$ export LD_LIBRARY_PATH=`pwd`/bin/
-    user@host:~/godot$ ./bin/godot*
+    export LD_LIBRARY_PATH=`pwd`/bin/
+    ./bin/godot*
```
follower, line 391: FWIW I tend to use the following non-`export` form which I like because it's a one-liner & doesn't "pollute the environment":

```
LD_LIBRARY_PATH=`pwd`/bin/ ./bin/godot*
```

It also has the benefit that the required environment variable configuration is kept with the command.

Calinou: There's a warning just below that recommends using `export` so that the variable is set when running the project from the editor.

That said, I'll change the syntax to make the variable work if the path contains spaces (right now, it won't).

follower:
It might be helpful to update the current descriptions on https://godotengine.org/download/server to be based on this section:

```
To compile a *headless* build which provides editor functionality to export
projects in an automated manner, use::

    scons -j8 platform=server tools=yes target=release_debug

To compile a *server* build which is optimized to run dedicated game servers,
use::

    scons -j8 platform=server tools=no target=release

```

It wasn't until I read the above & saw the `tools=yes` vs `tools=no` distinction that I actually understood what the difference between the two builds was.

This is the existing description:

> *  Headless is a Linux 64-bit build of the editor tools to run on headless environment. It is used from the command line to run tests or export projects.
>
>  *  Server is a Linux 64-bit build of the export template running without graphics and audio, useful to host server instances for Godot games.

(I think it was the language about "the export template" that tripped me up--I mean, I "know" that the export templates are executables but I guess I don't really think of them as "builds" in quite the same way...  )

Perhaps something like:

> *  The *headless* build includes the editor tool functionality that enables it to run tests and export projects in an automated manner. 
>
> * The *server* build is optimized to run dedicated game servers and does not include editor tools, graphics or audio support.

FWIW. :)

PR #3559, opened 2020-05-18
Mention lawnjelly's smoothing add-on in Fixing jitter and stutter

Calinou:
It seems to be working well according to a few people I recommended it to.


mhilbrunner:
Thanks!

PR #3570, opened 2020-05-20
Clarify root node type in your first game

### LINKED ISSUE #3452: Small clarifications on 'create your first game' tutorial
In general I find the Godot docs very precise and easy to follow.

However, in the tutorial to [create your first game](https://docs.godotengine.org/en/latest/getting_started/step_by_step/your_first_game.html) I had two steps that had difficulties with:

- When creating the root node for the Main scene I got a little confused. The explanation says you have to create a Node, but I didn't know which type of node (which is specified in all other places before). Later I understood that the most basic node was meant here, but I think it would help beginners like me if this was specified.
- `Drag Mob.tscn from the “FileSystem” panel and drop it in the Mob property under the Script Variables of the Main node.`
While doing this, I was still modifying the script. To find the script variables, you have select again the main node (not the script editor) and add it there. This is already explained earlier, but the drag and drop made me expect to find a view with this name.

skyace65:
Clarifies that the root node of the main scene in your first game is a Node node. This was the best way I could explain it without going in detail about what a Node node is, I'm open to suggestions if anyone thinks they can improve this. closes #3452 (the rest of that issue was fixed in another PR)

### DIFF
--- a/getting_started/step_by_step/your_first_game.rst
+++ b/getting_started/step_by_step/your_first_game.rst
@@ -666,8 +666,10 @@ This completes the `Mob` scene.
 Main scene
 ----------
 
-Now it's time to bring it all together. Create a new scene and add a
-:ref:`Node <class_Node>` named ``Main``. Click the "Instance" button and select your
+Now it's time to bring it all together. Create a new scene, and create a
+:ref:`Node <class_Node>` as the root of the scene, it should be at the 
+top of the node list and has a white circle as an icon. Name
+the root node ``Main``. Click the "Instance" button and select your
 saved ``Player.tscn``.
 
 .. image:: img/instance_scene.png


getting_started/step_by_step/your_first_game.rst
```diff
@@ -666,8 +666,11 @@ This completes the `Mob` scene.
 Main scene
 ----------
 
-Now it's time to bring it all together. Create a new scene and add a
-:ref:`Node <class_Node>` named ``Main``. Click the "Instance" button and select your
+Now it's time to bring it all together. Create a new scene, click on
+the other node button and create a
```
cbscribe, line 670: >  click on the other node button

This is the third scene we've created so far, so this seems unnecessary. At least title it consistently: "Other Node" button

getting_started/step_by_step/your_first_game.rst
```diff
@@ -666,8 +666,11 @@ This completes the `Mob` scene.
...
+:ref:`Node <class_Node>` node as the root of the scene, it has a
```
cbscribe, line 671: "Node node" is redundant

getting_started/step_by_step/your_first_game.rst
```diff
@@ -666,8 +666,11 @@ This completes the `Mob` scene.
...
+white circle as an icon and should be selected by default. Name
```
cbscribe, line 672: The "," should be the start of a new sentence.

> `Node` is the first node type in the list and has a white circle icon.

Note: it probably *won't* be selected by default if the reader has done any searching of the node list previously. I'd recommend removing that statement.

cbscribe (changes requested):
Is this really a problem? I've been through this tutorial extensively with a large number of students of varying ages and this one has never made the list of pain points. "Node" is highlighted and linked.

That said, if folks think it's necessary, I've made some recommendations for wording changes.

skyace65:
Someone got confused and opened an issue so I figure we might as well clarify it. I've made the changes you suggested.

mhilbrunner:
I think this is okay to merge, but I don't have a strong opinion on this - I'll let @cbscribe decide whether the added complexity of the description is worth it to be more explicit, and to merge or close. :)

Thanks for your work!

mhilbrunner (approved):

mhilbrunner:
Having another looks, I changed my mind - I too don't see this as something that needs fixing, and it *is* very verbose. I'll close this, as there's been no consensus either way. Still, thanks for the contribution! :)

PR #3575, opened 2020-05-21
Fix numbering in the ordering section of Shader language style guide

Calinou:
Thanks to Noshyaar for reporting this one.


mhilbrunner:
Thanks!

PR #3580, opened 2020-05-22
Add VCS documentation

### LINKED ISSUE #935: Docs: How to use Godot with VCS
More a reminder to myself to create a page with some general workflow-y things like what files can/should be `.gitignore`d and so on. (`.import/` and `.mono/`, maybe? Also the usual stuff like `.DS_Store`)

Note that this is in addition to the existing docs on how to **contribute** to Godot using Git :)

### LINKED ISSUE #2838: Document how to use VCS (Version Control System e.g. git) editor functionality
AFAICT the new Version Control functionality available via `Project > Version Control > Set Up Version Control` does not have any documentation on how to set it up.

Out of the box 3.2.alpha1 only displays an error dialog with this message:

    No VCS addons are available.

At a minimum it would be good if the Changelog/Release Notes mentioned that it won't work out of the box. (And that the functionality requires a per-OS compiled addon.)

Related links:

 * https://github.com/godotengine/godot-git-plugin
 * https://github.com/godotengine/godot-git-plugin/releases
 * https://github.com/godotengine/godot/issues/32135 (and https://github.com/godotengine/godot/commit/6df03fe61f8b25f932e0666cfd873e563e8cf1ed)
 * https://github.com/godotengine/godot/blob/ddc42be1dbb66e8ae7e7e13deace0c59fea91abd/editor/plugins/version_control_editor_plugin.cpp#L101
 * https://github.com/godotengine/godot/pull/31461
 * https://gist.github.com/IronicallySerious/1ac3e3ae4a51e6f739c31dd7c46f2566

skyace65:
Adds documentation on how to setup VCS. Given that the documentation for how to use the git plugin is already done in the plugins repository this ended up being relatively small. The one image has gone through png compression.

I could use some help with what files should be ignored by vcs and why. @mhilbrunner you opened up an issue about this a few years ago, any information you have would be very appreciated. 

closes #2838 and closes #935


getting_started/workflow/project_setup/version_control_system.rst
```diff
@@ -0,0 +1,31 @@
+.. _doc_version_contro_system:
+
+Version Control System
+======================
+
+Introduction
+------------
+
+Godot supports the use of version control systems, they can be
+setup or shut down in the editor under Project, Version Control.
```
pycbouh, line 10: Documentation usually uses **Project > Version Control** format, wrapped in double asterisks.

Also, there is no need to write this as a continuous sentence. It's better to expand it a bit, and write as complete sentences. I also think it would be better to combine this part with the later part. So you end up with three sentences before the screenshot:
* "Godot has support".
* "You need a plugin".
* "You can configure it in settings".
* Screenshot
* Next section

getting_started/workflow/project_setup/version_control_system.rst
```diff
@@ -0,0 +1,31 @@
...
+
+.. image:: img/version_control_menu.png
+
+However, version control does not work out of the box and requires
+a plugin for the specific vcs you are using.
```
pycbouh, line 15: "VCS" should always be written in the uppercase.

getting_started/workflow/project_setup/version_control_system.rst
```diff
@@ -0,0 +1,31 @@
...
+
+Official Git plugin
+-------------------
+
+Git is supported with an official plugin, you can find the latest
```
pycbouh, line 20: Once again, no need to concatenate everything into a single sentence.

getting_started/workflow/project_setup/version_control_system.rst
```diff
@@ -0,0 +1,31 @@
...
+releases `here <https://github.com/godotengine/godot-git-plugin/releases>`_ 
+(it is not available in the asset library). Documentation on how
+to use the git plugin can be found `here <https://github.com/godotengine/godot-git-plugin/wiki>`_.
```
pycbouh, line 23: "Git" should be properly capitalized.

getting_started/workflow/project_setup/version_control_system.rst
```diff
@@ -0,0 +1,31 @@
...
+
+Files to exclude from VCS
+-------------------------
+
+There are some folders Godot creates you should have your VCS ignore.
```
pycbouh, line 28: I suggest putting a colon here and formatting the list of folders with `::`, like we do with code.

skyace65:
I've made the changes you suggested. However I think bullet points is more appropriate for the list of stuff to exclude instead of code blocks, since I'd like to add descriptions for why those folders should be excluded.

mhilbrunner:
IMO the first paragraph should be reworded to say that Godot is VCS-friendly/aims to generate mostly readable (and mergeable) files, and that there exist optional plugins to use VCS inside the editor.

As is it sounds a little like the Git plugin is the only recommended approach, while many (dare I say most?) devs use external tooling.

Otherwise looks good to me.

skyace65:
Ok I just reworded it a bit, what do you think? Also what reason should we give for why VCS should ignore the .import and .mono folders?

mhilbrunner:
I'd rename the page to "Version Control System**s**", to make clear that it is not about a single VCS but how to work with them with Godot.

>Also what reason should we give for why VCS should ignore the .import and .mono folders?

To quote Akien from [way back when](https://github.com/github/gitignore/pull/2546#discussion_r156121287):
`.import/, the folder at the root of your project repo (where project.godot is), is where Godot stores all the files it imports automatically based on your source assets and their import flags. This should be git-ignored.`

Also, maybe we should point/link to https://github.com/github/gitignore/blob/master/Godot.gitignore

skyace65:
Okay I've made the changes. I'd rather not link to that repo, it's small enough we should include that info on this page, plus I'd prefer to explain why things are being ignored so people understand how the engine works.

getting_started/workflow/project_setup/version_control_systems.rst
```diff
@@ -0,0 +1,33 @@
...
+systems in the editor itself. However, VCS in the editor requires
+a plugin for the specific VCS you are using. VCS can be setup or
+shut down in the editor under Project > Version Control.
+
+.. image:: img/version_control_menu.png
+
+Official Git plugin
+-------------------
+
+Git is supported with an official plugin. You can find the latest
```
mhilbrunner, line 20:
```suggestion
Using Git from inside the editor is supported with an official plugin. You can find the latest
```

mhilbrunner:
I think this is good to go now. Thanks for taking the time to do this :)

IronicallySerious:
Wow I didn't even know this was already happening. I had this task lined up for me but it's great to see this already done!

PR #3585, opened 2020-05-23
Tweak `LD_LIBRARY_PATH` export to work with paths that contain spaces

Calinou:
This also uses the `$PWD` environment variable which should always be defined. It's also minutely faster than calling `$(pwd)` or using the deprecated backtick syntax.


mhilbrunner:
Thanks!

PR #3587, opened 2020-05-24
Change incorrect MessageLabel to Message in Your first game

t-persson:
The Label created is called 'Message' and not 'MessageLabel'.


t-persson:
This could be changed so that all "Message" instances are called "MessageLabel" instead, if that's better.
Don't know what the preferred name standards are yet as I am just getting started

PR #3594, opened 2020-05-25
Clarify example code for signal

brandonlamb:
Section discussing placing signals above other properties did not actually contain a line for signals (which confused me). Clarifying the style guide by showing a signal on top


mhilbrunner:
Good catch! Thanks and congrats to your first Godot contribution :)

PR #3596, opened 2020-05-26
fixed incorrect animation name ('right' -> 'walk')

Motherships:
I went through the tutorial and noticed inconsistencies

Previously, we called the animation "walk" in https://github.com/godotengine/godot-docs/blob/master/getting_started/step_by_step/your_first_game.rst


cbscribe:
Thanks - missed that on the last review!

PR #3599, opened 2020-05-27
Fix old style -path mention

akien-mga:


Calinou:
(Nice branch name, by the way. :smile:)

PR #3602, opened 2020-05-27
Tweak the theme font stack to follow Bootstrap 5.0.0-alpha1

Calinou:
This most notably removes Oxgyen which was reported to have a poor "œ" character appearance under certain font weights.


akien-mga:
Thanks!

PR #3605, opened 2020-05-28
Add a reference to the Data paths doc in the Step by step guide

pycbouh:
As noted by @Zylann, [this section](https://docs.godotengine.org/en/stable/getting_started/step_by_step/filesystem.html?#user-path) of the Step by Step guide could use a reference to an explanation on how `user://` paths are resolved on various platforms. While the [Data paths](https://docs.godotengine.org/en/stable/tutorials/io/data_paths.html#user-path-persistent-data) doc is not exactly complete without mobile and web platforms, it's the best place to explain this detail. So I've linked them.

Also adjusted formatting in the doc.

### DIFF
--- a/getting_started/step_by_step/filesystem.rst
+++ b/getting_started/step_by_step/filesystem.rst
@@ -21,9 +21,9 @@ included. If a resource has sub-resources that are built-in, the resource is
 saved in a single file together with all the bundled sub-resources. For
 example, a font resource is often bundled together with the font textures.
 
-The Godot file system avoids using metadata files. Existing asset managers and VCSs are better than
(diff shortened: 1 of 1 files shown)


Calinou:
> While the Data paths doc is not exactly complete without mobile and web platforms, it's the best place to explain this detail.

On Web platforms, LocalStorage is used for this. It's not an actual place on the filesystem, but a per-site virtual filesystem provided by the Web browser.

pycbouh:
Yeah, I think I remember that. I just think that the article might've mentioned it in some form. Currently it's completely omitted, which would confuse people exporting for those missing platforms.

PR #3614, opened 2020-05-30
Update docs on input right -> ui_right

Murtagy:

### DIFF
--- a/tutorials/2d/2d_movement.rst
+++ b/tutorials/2d/2d_movement.rst
@@ -47,13 +47,13 @@ Add a script to the kinematic body and add the following code:
 
     func get_input():
         velocity = Vector2()
-        if Input.is_action_pressed('right'):
+        if Input.is_action_pressed('ui_right'):
             velocity.x += 1
-        if Input.is_action_pressed('left'):
+        if Input.is_action_pressed('ui_left'):
             velocity.x -= 1
-        if Input.is_action_pressed('down'):
+        if Input.is_action_pressed('ui_down'):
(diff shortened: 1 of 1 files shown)


Calinou:
You need to set up input actions as described in that tutorial. [Someone already opened a pull request to change the actions to the default hardcoded ones](https://github.com/godotengine/godot-docs/pull/3465), but we'd prefer not encouraging it as it's considered bad practice to use "reserved" inputs for gameplay purposes.

ulvido:
this is confusing for the newcomers. I was in the [2D movement overview](https://docs.godotengine.org/en/stable/tutorials/2d/2d_movement.html) part and I thought the code was not working.

cbscribe:
How is it confusing? If you follow the directions at the beginning of the tutorial, which *tell* you to set up the inputs, then the rest of the tutorial works fine.

ulvido:
great docs btw. you can just understant the document is not working properly at that part if there is this issue on github. You can assume some people searched the problem and saw this topic and sad "ah I found the solution" and leaved.

image

it is not clear which refers which.  this "left" or that "left" or something else?
or can I write Input.is_action_pressed('a'):
how can a newbea bind that information together. the "note" explains some but it's placed at the bottom of the topic. it should be uppper so people can understand before strugle. also according to note check my setup actions and do what? note should be more explaining. my setup actions are correct. the "code" should be fixed acording to actions. 

and a reminder doc picture is outdated for v3.2.3. it is not people see. this is the picture at docs:
image

PR #3615, opened 2020-05-30
Describe the differences between hardware and software mouse cursors

Calinou:
This answers a common question about mouse cursors.


mhilbrunner:
Thanks!

PR #3617, opened 2020-05-30
Add --path option to sample VS Code launch.json

brettle:


pycbouh:
Ah, didn't catch it in time. I feel like the comment should be moved down one line like the one above to avoid unnecessary scrolling.

Calinou:
@pycbouh https://github.com/godotengine/godot-docs/pull/3620

PR #3618, opened 2020-05-30
Add gcc-c++ dependency for Fedora.

brettle:
Probably required for distros but I'm not positive the package name is the same everywhere.


development/compiling/compiling_for_linuxbsd.rst
```diff
@@ -51,7 +51,7 @@ Distro-specific one-liners
 |                  |                                                                                                           |
 |                  |     sudo dnf install scons pkgconfig libX11-devel libXcursor-devel libXrandr-devel libXinerama-devel \    |
 |                  |         libXi-devel mesa-libGL-devel mesa-libGLU-devel alsa-lib-devel pulseaudio-libs-devel \             |
-|                  |         libudev-devel yasm                                                                                |
+|                  |         libudev-devel yasm gcc-c++                                                                               |
```
akien-mga, line 54: That breaks the table as the trailing `|` is no longer aligned.

brettle: Sorry about that. Fixed now.

PR #3620, opened 2020-05-30
Tweak arguments in the VS Code configuration guide

Calinou:
- Use the long name for the `--editor` option to make it more obvious.
- Document how to run a project directly while debugging.
- Put comments above the line to avoid horizontal overflow.


mhilbrunner:
Looks good, thanks!

PR #3626, opened 2020-06-01
Update the naming scheme for the Godot Android library

m4gr3d:
This is based on feedback from https://github.com/godotengine/godot-build-scripts/issues/6#issuecomment-635962906

PR #3629, opened 2020-06-02
Remove `patch` from the version string

m4gr3d:
Remove `patch` from the version string when it's `0`.
Based on https://github.com/godotengine/godot-docs/pull/3626#discussion_r433887279.


mhilbrunner:
Thanks!

PR #3630, opened 2020-06-02
Add autotile tutorial

PeterJSFChaplin:
Modify tutorials > 2D > using tilemaps to include section on
autotiles, covering different bitmask modes and scripted binding

### DIFF
--- a/tutorials/2d/using_tilemaps.rst
+++ b/tutorials/2d/using_tilemaps.rst
@@ -1,7 +1,7 @@
 .. _doc_using_tilemaps:
 
 Using tilemaps
-~~~~~~~~~~~~~~
+==============
 
 Introduction
 ------------
@@ -127,8 +127,10 @@ Atlas tiles
 -----------
 
 Rather than adding individual tiles one at a time, you can define a group of
-tiles all at once using an atlas. Click "New Atlas" and drag to select the
-entire tile sheet.
+tiles all at once using an atlas. This also allows you to randomly generate
+tiles from the group.
+
+Click "New Atlas" and drag to select the entire tile sheet.
 
 .. image:: img/tileset_atlas.png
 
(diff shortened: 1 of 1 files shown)


tutorials/2d/using_tilemaps.rst
```diff
@@ -150,6 +152,108 @@ tiles it contains:
 In addition to saving time when defining the tiles, this can help by grouping
 similar tiles together when you're working with a large number of tiles.
 
+Random tile priorities
+~~~~~~~~~~~~~~~~~~~~~~
+
+When drawing with atlas tiles, enabling the "Use priority" option causes tiles
+to be selected at random. By default, each tile in the tileset has an equal
+likelihood of occuring. You can change the likelihood by setting different
```
mhilbrunner, line 160:
```suggestion
likelihood of occurring. You can change the likelihood by setting different
```

tutorials/2d/using_tilemaps.rst
```diff
@@ -150,6 +152,108 @@ tiles it contains:
...
+---------
+
+Autotiles allow you to define a group of tiles, then add rules to control which
+tile gets used for drawing based on the content of adjacent cells.
+
+Click "New Autotile" and drag to select the tiles you wish to use. You can add
+collisions, occlusion, navigation shapes, tile priorties, and select an icon
+tile in the same manner as for atlas tiles.
+
+Tile selection is controled by bitmasks. Bitmasks can be added by clicking
```
mhilbrunner, line 175:
```suggestion
Tile selection is controlled by bitmasks. Bitmasks can be added by clicking
```

mhilbrunner:
Fixed two small typos. Otherwise looks good to me :)

cbscribe:
I think it would be very helpful to include a mention of how many tiles you need in your tileset for each bitmask mode. This question comes up a lot.

PeterJSFChaplin:
@cbscribe I'm a little wary of how we word that - technically you can create a tileset with only two subtiles for any bitmask mode, if there's some reason you _want_ an autotile with only two subtiles. Especially with the addition of wildcards (aka "ignore" bits).

It depends on what behaviour you want and/or expect.

PeterJSFChaplin:
Ok, I've tried to include some text about the number of supported bitmasks in each mode

cbscribe:
I hear you, but in the support channels autotile is a constant source of confusion. I'm still trying to come up with a better way to unambiguously demonstrate how it works. At least with a fixed, "here are the 47 tiles you need for 3x3 (minimal)" example, people have a place to start.

Thanks - I think this is good to go.

mhilbrunner:
Thanks - and congrats on your first contribution! :)

PR #3634, opened 2020-06-03
i18n: Retrieve localized images from godot-docs-l10n

akien-mga:
The localization for this documentation is handled via the godot-docs-l10n
repository at: https://github.com/godotengine/godot-docs-l10n

Sphinx supports localization of images by automatically looking up files
with a language code suffix (e.g. `myimage.zh_CN.png` as an override for
`myimage.png` when using `zh_CN` locale), but only in the same location as
the original file.

We want to host our localized images together with the PO files in the l10n
repo, so we use a glorious hack to redefine Sphinx's
`sphinx.util.i18n.get_image_filename_for_language` method in a way that fits
our needs.

Fixes godotengine/godot-docs-l10n#5.

Co-authored-by: @binotaliu
Co-authored-by: @groud

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -181,8 +181,41 @@ linkcheck_timeout = 10
 
 # -- I18n settings --------------------------------------------------------
 
+# Godot localization is handled via https://github.com/godotengine/godot-docs-l10n
+# where the main docs repo is a submodule. Therefore the translated material is
+# actually in the parent folder of this conf.py, hence the "../".
+
 locale_dirs = ["../sphinx/po/"]
 gettext_compact = False
+
+# We want to host the localized images in godot-docs-l10n, but Sphinx does not provide
+# the necessary feature to do so. `figure_language_filename` has `{root}` and `{path}`,
+# but they resolve to (host) absolute paths, so we can't use them as is to access "../".
+# However, Python is glorious and lets us redefine Sphinx's internal method that handles
+# `figure_language_filename`, so we do our own post-processing to fix the absolute path
+# and point to the parallel folder structure in godot-docs-l10n.
+# Note: Sphinx's handling of `figure_language_filename` may change in the future, monitor
+# https://github.com/sphinx-doc/sphinx/issues/7768 to see what would be relevant for us.
+figure_language_filename = "{root}.{language}{ext}"
+
+import sphinx
+cwd = os.getcwd()
+
+sphinx_original_get_image_filename_for_language = sphinx.util.i18n.get_image_filename_for_language
+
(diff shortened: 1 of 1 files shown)


akien-mga:
Note: I'm doing some tests and relevant workflow changes over at https://github.com/godotengine/godot-docs-l10n which are necessary to validate this PR, so please don't merge it for now. I'll merge myself once I'm satisfied with the feature (feel free to review though).

binotaliu:
LGTM. BTW, This hack actually has a name, **Monkey Patch.**
I think we also need a instruction on the l10n repo about how to copy all image assets (maybe just run a shell script,) and how to optimize images.

akien-mga:
Yeah I'm working on it:
```
diff --git a/update.sh b/update.sh
index 8b888f4a3..8e4d0655a 100755
--- a/update.sh
+++ b/update.sh
@@ -27,6 +27,7 @@ LANGS="es fr zh_CN"
 
 # Options
 make_templates_list=false
+copy_images=false
 update_sphinx_pot=false
 update_sphinx_po=false
 update_weblate_pot=false
@@ -39,6 +40,9 @@ while [ $# -gt 0 ]; do
     --templates-list|-tl)
       make_templates_list=true
       ;;
+    --copy-images|-ci)
+      copy_images=true
+      ;;
     --sphinx-pot|-st)
       update_sphinx_pot=true
       ;;
@@ -90,6 +94,19 @@ if [ "$make_templates_list" = true ]; then
   echo -e "index.pot\n$templates" > templates_list.txt
 fi
 
+# Copy the images from the docs repo in this repo's parallel structure in `images/`,
+# so that translators can easily find them and author the localized versions.
+if [ "$copy_images" = true ]; then
+  echo "=== Copying images from 'docs/' to 'images/' ==="
+  cd docs
+  # All used images are in `img/` subfolders, so we can simply find and copy them.
+  for folder in $(find -type d -name "img"); do
+    mkdir -p ../images/$folder
+    cp -a $folder/* ../images/$folder/
+  done
+  cd ..
+fi
+
 # Generate/Update Sphinx template from rst files
 if [ "$update_sphinx_pot" = true ]; then
   echo "=== Updating Sphinx templates from source rst files ==="
```

This will copy all the original images so that we can localize them easily. This will use some extra disk space in godot-docs-l10n but IMO it's better this way than requiring that translators find the original image in godot-docs, create the folder structure in godot-docs-l10n and copy it there to modify it. (Especially as the versions might differ between e.g. `3.2` docs on godot-docs-l10n and `4.0` docs on godot-docs.)

But before adding all those images to godot-docs-l10n I want to get rid of the unused ones: #3635.

PR #3636, opened 2020-06-03
Improve CI

mhilbrunner:
Improve CI build times by using a dummy builder for (empty) output, dropping LaTeX, and disabling some Sphinx extensions when building for CI (GDScript lexing, creating a search index, and generating HTML description tags).

In my tests, this improves Travis CI build times: from up to 30 minutes down to 3 minutes.

**Before**
before

**After**
after

**Things that didn't work**

Travis CI caching didn't seem to improve build times.
I tried both enabling the Python PIP cache as well as creating a per-branch cache for the Sphinx doctrees.

**Things that did work**
- Switching to a dummy output writer instead of building HTML files and throwing them away. Build times improved by roughly ~40%.
- Disabling extensions that aren't needed for CI. Build times improved by roughly ~20%.
- Disabling LaTeX. Build times improved by roughly ~20%.

**Caveats**

We're now not really building the HTML anymore, just parsing, then stopping the Sphinx build.
From testing, this still works fine: invalid formatting, spelling errors, and invalid links and references still lead all to warnings.

Some things may or may not be caught anymore however that would previously be detected (in theory), because some extensions/internal Sphinx behaviour may depend on the HTML builder.
Don't ask me for examples; in my testing, this worked fine :) Still, something to be aware of in case we notice anything in the future.

**Things still to investigate, will do after this**

- Moving from Travis to GitHub Actions for more concurrent builds
- Removing the `imgmath` Sphinx extension, it's only used on [this](https://docs.godotengine.org/en/stable/tutorials/math/vector_math.html) page anyway and doesn't work with the dark theme
- Completely removing LaTeX, not only for CI as this does; we don't seem to offer PDF downloads anymore
- Currently, most time is spent reading the `classses/class_*.rst` files. Maybe we can optimize this somehow, they don't really change that often (only when synced with the main repo)
- Theoretically, we could abort the build early on first errors (Sphinx warnings, spelling errors and such) - this would improve CI times at the cost of making CI output less useful, potentially requiring multiple fix-submit-CI cycles from contributors. IMO not worth it.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -13,12 +13,23 @@ needs_sphinx = "1.3"
 # Sphinx extension module names and templates location
 sys.path.append(os.path.abspath("_extensions"))
 extensions = [
-    "gdscript",
-    "godot_descriptions",
-    "sphinx_search.extension",
     "sphinx_tabs.tabs",
     "sphinx.ext.imgmath",
 ]
+
+# Warning when the Sphinx Tabs extension is used with unknown
+# builders (like the dummy builder) - as it doesn't cause errors,
+# we can ignore this so we still can treat other warnings as errors.
+sphinx_tabs_nowarn = True
+
+if not os.getenv("SPHINX_NO_GDSCRIPT"):
+    extensions.append("gdscript")
+
+if not os.getenv("SPHINX_NO_SEARCH"):
+    extensions.append("sphinx_search.extension")
+
+if not os.getenv("SPHINX_NO_DESCRIPTIONS"):
+    extensions.append("godot_descriptions")
 
 templates_path = ["_templates"]
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,12 @@
 os: linux
 dist: bionic
-language: python
+language: python - "3.8"
 
-python:
-  - "3.8"
-
-stages:
-  - build
+env:
+  global:
+    - SPHINX_NO_GDSCRIPT=1
+    - SPHINX_NO_SEARCH=1
+    - SPHINX_NO_DESCRIPTIONS=1
 
 matrix:
   include:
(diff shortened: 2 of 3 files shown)


conf.py
```diff
@@ -13,13 +13,21 @@
...
 sys.path.append(os.path.abspath("_extensions"))
 extensions = [
-    "gdscript",
-    "godot_descriptions",
-    "sphinx_search.extension",
     "sphinx_tabs.tabs",
     "sphinx.ext.imgmath",
 ]
 
+sphinx_tabs_nowarn = True
```
akien-mga, line 20: What does this do?

conf.py
```diff
@@ -13,13 +13,21 @@
...
+
+if not os.getenv("SPHINX_NO_GDSCRIPT"):
```
akien-mga, line 22: Since the use case is only for CI, maybe this can all be grouped in a single env variable like `GODOT_LEAN_BUILD`?

conf.py
mhilbrunner: The code tabs extension produces a warning that it's not compatible with the current builder if that builder is not in a list of "compatible" builders (like the HTML builder is). It then does nothing either way.

It's not a useful warning, as either you use a builder like the HTML builder and it works, or you use our nothing-builder and it does... nothing. :) I'll add a comment.

conf.py
mhilbrunner: I've thought about that, but I think I'd like to be able to disable building the search index which I never use locally, and still be able to have GDScript lexing work and see the pages locally as they would be on RTD.

I don't care strongly however, I can change this to GODOT_LEAN_BUILD too :)

conf.py
mhilbrunner: Done.

akien-mga:
Thanks!

PR #3639, opened 2020-06-04
Update import process images

skyace65:
Updates the images for the import process page for Godot 3.2. All images have gone through png compression.


cbscribe:
I don't think there was ever a consensus, but it has been brought up that using red for highlighting UI elements isn't colorblind-friendly.  See #967 .

I've been using yellow for my own images and/or arrows pointing at the highlighted item.

mhilbrunner:
Thank you!

PR #3642, opened 2020-06-05
Changed wording for matrices and transforms section to be more friendly

Duroxxigar:
This is to just try and change the wording to be a bit more friendly for one section of the matrices and transforms section.

### DIFF
--- a/tutorials/math/matrices_and_transforms.rst
+++ b/tutorials/math/matrices_and_transforms.rst
@@ -211,10 +211,9 @@ Changing the `origin` vector is called a *translating* the transformation
 matrix. Translating is basically a technical term for "moving" the
 object, but it explicitly does not involve any rotation.
 
-This should be fairly common sense, assuming that you read and
-understood the vector tutorial, but let's work through an example.
-Again, we'll start with the identity transform, but this time we'll
-also keep track of the origin vector:
+Let's work through an example to help understand this. We will start
+with the identity transform like last time, except we will keep track
+of the origin vector this time.
 
 .. image:: img/matrices_and_transforms/identity-origin.png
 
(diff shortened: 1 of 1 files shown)


tutorials/math/matrices_and_transforms.rst
```diff
@@ -211,10 +211,8 @@ Changing the `origin` vector is called a *translating* the transformation
 matrix. Translating is basically a technical term for "moving" the
 object, but it explicitly does not involve any rotation.
 
-This should be fairly common sense, assuming that you read and
-understood the vector tutorial, but let's work through an example.
-Again, we'll start with the identity transform, but this time we'll
-also keep track of the origin vector:
+Let's work through an example to help understand this. We will start
+with the identity transform and keep track of the origin vector this time.
```
aaronfranke, line 215: should be "but" keep track

Duroxxigar: And is used when connecting two clauses together. Both sides of the conjunction are independent clauses themselves, thus and is used correctly. The ideas do not contrast each other, hence why but is not being used.

aaronfranke: The point is to highlight what has changed compared to the previous example. "But" allows the first idea to be presented with a modification afterwards. The clauses are in the context of the previous text.

aaronfranke: I suppose it might work if the last example was referenced directly, like "We will start with the identify transform *like last time, except we* will keep track"

Duroxxigar: Fair compromise, fixed.

PR #3643, opened 2020-06-05
Update C# code in your_first_game.rst

nickrodgers42:
Add an `await ToSignal()` call to the C# code to match the `yield` call in the GDScript.


akien-mga:
Thanks!

PR #3651, opened 2020-06-06
Add minor doc improvements

justin-calleja:
I think the examples here: [getting_started/workflow/best_practices/scene_organization.html#how-to-build-relationships-effectively](https://docs.godotengine.org/en/stable/getting_started/workflow/best_practices/scene_organization.html#how-to-build-relationships-effectively) would be more effective if given in a dummy project one could run (and possibly tinker with).

Note: I kept all parts I edited up to (but not including) 80 characters wide in length as that's how the rest of the file seems to be (except for code snippets).


mhilbrunner:
These looks good, thanks :)

PR #3658, opened 2020-06-08
Replace register_method with bind_method

mrushyendra:
ClassDB exposes `bind_method` in order to register functions, and not
`register_method` anymore. Updated the Engine Development section to
mention the correct function.


akien-mga:
I wonder where this `register_method` came from... it has been `bind_method` since before that class was renamed to `ClassDB` from `ObjectTypeDB` (already `bind_method` in `1.0-stable`).

Welp... thanks for the fix :)

PR #3668, opened 2020-06-09
Add a C++ usage guidelines page

Calinou:
This page describes C++ features that are used and allowed in Godot's codebase. It's based on community consensus over the last months, as Godot recently moved to C++17 in the `master` branch.

I've probably missed many things, so please leave comments. That said, the C++ style we expect is now (mostly) set in stone, so most disallowed features written here are not negotiable (such as ``auto``). Please keep the discussion on-topic :slightly_smiling_face:

I know @hpvb intended to write this page at first, but I decided to do it to help speed up the process.

This closes https://github.com/godotengine/godot/issues/9694.


community/contributing/cpp_usage_guidelines.rst
```diff
@@ -0,0 +1,96 @@
...
+"for" range loop
+^^^^^^^^^^^^^^^^
+
+Don't use the ``for`` range loop syntax (``for (String string : some_array)``).
+Use traditional ``for`` loops instead as they're more explicit.
+
+Lambdas
+^^^^^^^
+
+The use of lambdas is generally not allowed, unless they can greatly help with
```
Calinou, line 89: I wrote the part about lambdas this way because we have a few instances of lambdas in the current editor code. These could be rewritten into macros, but it's not really convenient.

community/contributing/cpp_usage_guidelines.rst
```diff
@@ -0,0 +1,96 @@
...
+
+**Any feature not listed below is allowed.** Using features like ``constexpr``
+variables and ``nullptr`` is encouraged when possible. Still, try to keep your
+use of modern C++ features conservative. Their use needs to serve a real
+purpose, such as improving code readability or performance.
+
+Standard template library
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+We doesn't allow using the STL, as Godot provides its own data types (among
```
asmaloney, line 48: "doesn't" -> "don't"

Calinou: Good catch :slightly_smiling_face: Fixed.

FlashDaggerX:
Can we add a short section on the appropriate usage of switch cases? 

I'm bringing it up because of its inconsistent use throughout the code base, specifically on `_notification` functions. Some use switches, others use really long `if` / `if-else` blocks.

Examples:
* Neat `switch` cases [(scene/gui/line_edit.cpp)][1]
* Neat `if` block (one case; no `switch` cases) [(scene/gui/panel.cpp)][2]
* Inappropriate use of `if` blocks [(scene/gui/base_button.cpp)][3]
* Inappropriate use of `switch` cases [(scene/2d/line_2d.cpp)][4]
* Another good example of `switch` cases [(scene/main/node.cpp)][5]
* Another example of bad `if` blocks [(scene/main/http_request.cpp)][6]

[1]: <https://github.com/godotengine/godot/blob/d341181f30b4f2e154748911f5d8a59b43aa6e62/scene/gui/line_edit.cpp#L651>
[2]: <https://github.com/godotengine/godot/blob/d341181f30b4f2e154748911f5d8a59b43aa6e62/scene/gui/panel.cpp#L35>
[3]: <https://github.com/godotengine/godot/blob/d341181f30b4f2e154748911f5d8a59b43aa6e62/scene/gui/base_button.cpp#L81>
[4]: <https://github.com/godotengine/godot/blob/d341181f30b4f2e154748911f5d8a59b43aa6e62/scene/2d/line_2d.cpp#L237>
[5]: <https://github.com/godotengine/godot/blob/d341181f30b4f2e154748911f5d8a59b43aa6e62/scene/main/node.cpp#L53>
[6]: <https://github.com/godotengine/godot/blob/d341181f30b4f2e154748911f5d8a59b43aa6e62/scene/main/http_request.cpp#L390>

Calinou:
@FlashDaggerX Thinking about it, I think we should always use `switch` in `_notification` functions, even if there are only 1 or 2 cases. This is to make refactoring easier; when you add new cases, you don't want to spend time converting the `if` block to a `switch` block.

Ansraer:
Could someone explain the reasoning behind disallowing for-range-loops to me?
The docs only mention that they are not explicit enough, but I don't really feel like that's the case. As long as we are only iterating over not overly complex data it should be quite obvious what is happening.

Plus I would argue that the readability and ease of use make them beneficial for "beginner contributors (who may not be
  professional C++ programmers)". In my experience, it is a LOT easier to understand `for(CustomObject obj : list)` than a 70+ characters long loop statement.

NathanLovato:
What's the PR's status? What's missing? Is there a way we can review and merge what's available and consider extending it later?

Calinou:
@NathanLovato I think this is good to merge as-is, but it would be good to get @akien-mga's approval (especially the part about lambdas).

community/contributing/cpp_usage_guidelines.rst
```diff
@@ -0,0 +1,96 @@
+.. _doc_core_types:
```
aaronfranke, line 1:
```suggestion
.. _doc_cpp_usage_guidelines:
```

aaronfranke (changes requested):

community/contributing/cpp_usage_guidelines.rst
```diff
@@ -0,0 +1,96 @@
...
+**Any feature not listed below is allowed.** Using features like ``constexpr``
+variables and ``nullptr`` is encouraged when possible. Still, try to keep your
+use of modern C++ features conservative. Their use needs to serve a real
+purpose, such as improving code readability or performance.
+
+Standard template library
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+We don't allow using the STL, as Godot provides its own data types (among
+other things). See :ref:`this FAQ question <doc_faq_why_no_stl>` for more
```
aaronfranke, line 49: This causes an error when compiling the docs, not sure what it should be changed to.

about/faq.rst
```diff
@@ -380,7 +380,7 @@ Godot provides its own type-casting system, which can optionally use RTTI
 internally. Disabling RTTI in Godot means considerably smaller binary sizes can
 be achieved, at a little performance cost.
 
-Why does Godot not force users to implement DoD (Data oriented Design)?
+Why does Godot not force users to implement DoD (Data-Oriented Design)?
```
akien-mga, line 383: If it's **DoD**, then it should be capitalized **D**ata-**o**riented **D**esign.

community/contributing/cpp_usage_guidelines.rst
```diff
@@ -0,0 +1,96 @@
...
+**C++17**. While modern C++ brings a lot of opportunities to write faster, more
+readable code, we chose to restrict our usage of C++ to a subset for a few
+reasons:
+
+- It makes it easier to review code in online editors. This is because engine
+  contributors don't always have access to a full-featured IDE while reviewing
+  code.
+- It makes the code easier to grasp for beginner contributors (who may not be
+  professional C++ programmers). Godot's codebase is known to be easy to learn
+  from, and we'd like to keep it that way :)
```
akien-mga, line 19: While both are true, I don't think they're the only reasons why we use a subset and not full-blown over-engineered C++ (hint in the qualifier). This should likely be discussed further with core devs to come with a strong rationale, otherwise we'll get criticism for taking arbitrary technical decisions just for the sake of non-IDE reviews and junior devs.

community/contributing/cpp_usage_guidelines.rst
```diff
@@ -0,0 +1,96 @@
...
+
+To get your pull request merged, it needs to follow the C++ usage guidelines
+outlined here. Of course, you can use features not allowed here in your own C++
+modules or GDNative scripts.
+
+.. note::
+
+    Prior to Godot 4.0, the C++ standard used throughout the codebase was C++03.
```
akien-mga, line 27:
```suggestion
    Prior to Godot 4.0, the C++ standard used throughout the codebase was C++03,
    with a handful of C++11 extensions.
```

community/contributing/cpp_usage_guidelines.rst
akien-mga: @Calinou forgot to add/commit the named reference in the FAQ page.

community/contributing/cpp_usage_guidelines.rst
```diff
@@ -0,0 +1,96 @@
...
+
+Disallowed features
+-------------------
+
+**Any feature not listed below is allowed.** Using features like ``constexpr``
+variables and ``nullptr`` is encouraged when possible. Still, try to keep your
+use of modern C++ features conservative. Their use needs to serve a real
+purpose, such as improving code readability or performance.
+
+Standard template library
```
akien-mga, line 45: For once, this can be capitalized in title case as it's a proper name: https://en.wikipedia.org/wiki/Standard_Template_Library

I'd suggest adding a link to that Wikipedia page in the first sentence too to clarify for everyone what we refer to with STL. Up until recently I thought that STL == `std`, while it's specifically about the template components. (So `std::mutex` or `std::thread` are not part of it, and allowed in Godot (core) code which wraps around them).

community/contributing/cpp_usage_guidelines.rst
```diff
@@ -0,0 +1,96 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+We don't allow using the STL, as Godot provides its own data types (among
+other things). See :ref:`this FAQ question <doc_faq_why_no_stl>` for more
+information.
+
+This means that pull requests should **not** use ``std::string``,
+``std::vector`` and the like. Instead, use Godot's ``String``, ``Vector``
+and so on.
```
akien-mga, line 54: Maybe we should list all Godot templates and their use case (might need help from some `core` gurus) so that contributors are aware of them. Many use the STL because they're not aware that Godot has its own alternative implementations.

community/contributing/cpp_usage_guidelines.rst
```diff
@@ -0,0 +1,96 @@
...
+
+There are some exceptions in the current codebase (such as for threading
+primitives), but please ask before using features from the STL.
+
+"auto" keyword
```
akien-mga, line 59: Could use inline code in headings too (I guess you did not because it might not look great, but we do use it already in some places).

community/contributing/cpp_usage_guidelines.rst
akien-mga: I'd say:
```
Lambdas should be used conservatively when they make code effectively faster or simpler, and do not impede readability.
```

community/contributing/cpp_usage_guidelines.rst
```diff
@@ -0,0 +1,96 @@
...
+^^^^^^^
+
+The use of lambdas is generally not allowed, unless they can greatly help with
+readability. Please ask before using lambdas in a pull request.
+
+"#pragma once"
+^^^^^^^^^^^^^^
+
+To follow the existing style, please use standard ``#ifdef``-based include
+guards instead of ``#pragma once`` in new files.
```
akien-mga, line 96: Add a seealso or link to https://docs.godotengine.org/en/latest/community/contributing/code_style_guidelines.html#header-includes

community/contributing/index.rst
```diff
@@ -36,6 +36,7 @@ about the code style, the contribution workflow, and more.
    :name: toc-community-contributing-to-the-engine
 
    best_practices_for_engine_contributors
+   cpp_usage_guidelines
```
akien-mga, line 39: If this is here, `code_style_guidelines` should be next to it IMO.

community/contributing/cpp_usage_guidelines.rst
HaSa1002: Please. I learned C++ as one unit with the standard library and was a afraid of the non-usage of the stuff I learned are standard.

community/contributing/cpp_usage_guidelines.rst
YeldhamDev: Also, the smiley should probably be dropped. It feels kinda condescending.

fire (approved):
I did a read, and the significant points about autos, STL, and lambdas have been commonly agreed upon by the contributor team.

community/contributing/cpp_usage_guidelines.rst
```diff
@@ -0,0 +1,104 @@
...
+
+.. code-block:: cpp
+
+    // Not so confusing...
+    auto button = memnew(Button);
+
+    // ...but what about this?
+    auto result = EditorNode::get_singleton()->get_complex_result();
+
+Keep in mind hover documentation isn't always readily available for pull request
```
aaronfranke, line 75:
```suggestion
Keep in mind hover documentation isn't often readily available for pull request
```

aaronfranke (changes requested):

aaronfranke (approved):

mhilbrunner (approved):

Calinou:
Should this page be pushed to `3.4` as-is, with a note saying that it only applies to the `master` branch? This would help people find it more easily.

pycbouh:
I think that notable differences and general code compatibility warning should be a part of the page anyway. People still support 3.x actively, so we should mention the caveats in the docs. Then it wouldn't matter in which version we store the page, as it would be applicable to the development in general.

PR #3669, opened 2020-06-09
Add an Editor style guide page

Calinou:
This page describes the writing style used throughout the Godot editor. It complements the already-established Docs writing guidelines.

See https://github.com/godotengine/godot/pull/36023 (and likely other PRs where I've used those conventions over the years).

### DIFF
--- a/development/editor/editor_style_guide.rst
+++ b/development/editor/editor_style_guide.rst
@@ -1,0 +1,98 @@
+.. _doc_editor_style_guide:
+
+Editor style guide
+==================
+
+Introduction
+------------
+
+Thanks for your interest in contributing to the Godot editor!
+
+This page describes the grammar and writing style used throughout the Godot
+editor. Following this style guide will help your contribution get merged faster
+since there will be fewer review steps required.
+
+Writing style
+-------------
+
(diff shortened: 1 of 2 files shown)


mhilbrunner:
Thank you!

development/editor/editor_style_guide.rst
```diff
@@ -0,0 +1,98 @@
...
+Tooltips
+--------
+
+Consider adding tooltips whenever the action performed by a button or menu
+action isn't obvious. You can also provide additional context or highlight
+caveats in the tooltip.
+
+You can do this by calling ``set_tooltip(TTR("Text here."))`` on the
+Control-based node in question. If the tooltip is particularly long (more than
+~80 characters), wrap it over several lines by adding line breaks using ``\n``.
```
akien-mga, line 92: I don't agree with this advice, manual wrapping is not something one should have to do in 2020.

If tooltips don't auto-wrap to a reasonable max width, they have to be fixed in the engine.

Calinou: I think it makes sense to add a custom theme constant in TooltipLabel for this. If it's greater than zero, soft wrapping would be used on the tooltip text. Not sure if it's technically feasible that way though.

PR #3681, opened 2020-06-13
Switch continuous integration to GitHub Actions

Calinou:
This allows freeing up some Travis CI jobs for the main Godot repository, as job parallelism limits are shared across an organization.


akien-mga:
Thanks!

PR #3682, opened 2020-06-13
Improve the tutorial for encrypting save games

Nurali9999:
Make the first few sentences less pessimistic, please.

### DIFF
--- a/tutorials/io/encrypting_save_games.rst
+++ b/tutorials/io/encrypting_save_games.rst
@@ -6,9 +6,9 @@ Encrypting save games
 Why?
 ----
 
-Because the world today is not the world of yesterday. A capitalist
-oligarchy runs the world and forces us to consume in order to keep the
-gears of this rotten society on track. As such, the biggest market for
+Because the world today is not the world of yesterday [citation needed]. A capitalist
+oligarchy runs the world and makes us consume some things in order to keep the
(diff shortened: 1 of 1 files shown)


Calinou:
Sorry, but we intend to keep the joke as it currently is. The paragraph isn't intended to be taken seriously :slightly_smiling_face:

Nurali9999:
Either way, the society is not _rotten_ , and being influenced by rich people isn't _universally_ bad. Also, not all players are miserable people like the paragraph describes. A joke is not funny when it's not at least half-true.

akien-mga:
Well it's funny because it *is* half-true. The adjectives you point out are the ones making it (arguably) half-false :)

Nurali9999:
I apologize for not clearly explaining why I made this commit in the first place. Today, after seeing somebody's reddit post about this document, I wrote the following comment: 

"I get that the author tried to explain that encrypting save games can be used to keep players from cheating in an in-app purchase system, but why call the entire society "rotten"? Why be so pessimistic?

Edit 1: Also, hasn't society always relied on its richest members to function? If so, then the first sentence of the above paragraph is wrong.

Edit 2: To people who say this is a joke, insulting the entire society isn't funny."

This is why I proposed to make the joke slightly more polite.

PR #3683, opened 2020-06-13
Fix shellcheck warnings in `format.sh`

Calinou:
This makes the script behave correctly if the path to the cloned repository contains spaces.


akien-mga:
Thanks!

PR #3684, opened 2020-06-13
Add a note about the introduction in Encrypting save games

Calinou:
Many contributors have opened pull requests to remove it, but the consensus over the years is that most people are actually fine with the paragraph in question.


akien-mga:
I would add "This introduction **is an Easter egg and** is not intended to be taken seriously."

Otherwise some might still wonder what's the point of it.

akien-mga:
Thanks!

PR #3685, opened 2020-06-13
Document that `light()` functions aren't run if vertex shading is forced

Calinou:
See https://github.com/godotengine/godot/issues/39513.


mbrlabs:
This should probably also mention that the same limitation applies if `vertex_lighting` is used as `render_mode` directly in the shader.

Calinou:
@mbrlabs It probably will. Thanks for reminding me about it :slightly_smiling_face:

mbrlabs:
No problem ;)

akien-mga:
Thanks!

PR #3687, opened 2020-06-14
styleguide: clarify "static typing" syntax

nathanleiby:
This change is an edit to: https://docs.godotengine.org/en/stable/getting_started/scripting/gdscript/gdscript_styleguide.html#static-typing

I first explored making an update here because I felt the current docs were confusing:

1. I could not tell if everything below "bad" was actually bad... It seemed like some of it was "good".
2. The examples involving ` var health := 0  # The compiler will use the int type` seemed redundant, and re: (1) it was confusing to see one in "good" and one in "bad".
3. I also wasn't able to understand "Place the colon right after the variable's name, without a space, and let the GDScript compiler infer the variable's type when possible".

First off: do you agree whether a change is useful or did I misunderstand the docs as written?

---

While writing this proposed update, I realize that the write-up is a mixture of style and explaining how types work. Is that what we expect in the style guide, or should it only focus on syntax and code clarity? If the latter, perhaps we should keep this minimal and treat the linked static typing guide as the source of truth: https://docs.godotengine.org/en/stable/getting_started/scripting/gdscript/static_typing.html#doc-gdscript-static-typing.

Thanks for the awesome game engine!

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_styleguide.rst
+++ b/getting_started/scripting/gdscript/gdscript_styleguide.rst
@@ -688,19 +688,37 @@ Static typing
 
 Since Godot 3.1, GDScript supports :ref:`optional static typing<doc_gdscript_static_typing>`.
 
-Type hints
-~~~~~~~~~~
+Declared Types
+~~~~~~~~~~~~~~
 
-Place the colon right after the variable's name, without a space, and let the
-GDScript compiler infer the variable's type when possible.
+To declare a variable's type, use ``<variable>: <type>``:
+
+::
+
+   var health: int = 0
+
+To declare the return type of a function, use ``-> <type>``: 
+
+::
+
+   func heal(amount: int) -> void:
+
+Inferred Types
+~~~~~~~~~~~~~~
+
+In some cases you can let the compiler infer the type, using ``:=``:
+
+::
+   var health := 0  # The compiler will use the int type.
+
+However, in other cases the compiler is not able to infer the exact type, 
+so you should set it explicitly.
 
 **Good**:
 
 ::
 
    onready var health_bar: ProgressBar = get_node("UI/LifeBar")
-
-   var health := 0 # The compiler will use the int type.
 
 **Bad**:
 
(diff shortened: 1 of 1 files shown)


getting_started/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -688,36 +688,42 @@ Static typing
...
 
-   onready var health_bar: ProgressBar = get_node("UI/LifeBar")
+   func heal(amount: int) -> void:
 
-   var health := 0 # The compiler will use the int type.
+Inferred Types
+~~~~~~~~~~~~~~
 
-**Bad**:
+In some cases you can let the compiler infer the type, using ``:=``:
```
NathanLovato, line 709: I'd say the idea is to use type inference whenever you can. It makes your code easier to change (you don't have to change the type when refactoring) and works in many cases. The exception is when it doesn't work, i.e. on get_node() calls, because unless the scene or file of the node is loaded in memory, the compiler lacks context to infer the type.

getting_started/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -688,36 +688,42 @@ Static typing
...
 
 ::
+   var health := 0  # The compiler will use the int type.
 
-   # The compiler can't infer the exact type and will use Node
-   # instead of ProgressBar.
-   onready var health_bar := get_node("UI/LifeBar")
+However, in other cases the compiler is not able to infer the exact type,
```
NathanLovato, line 714: It's not in many cases. It's only when the context is missing, e.g. you're in a GDScript file, you store an arbitrary node in a variable, and the target node isn't loaded in memory. In that case, the compiler falls back on the function's return type. For `get_node`, it's `Node`, but it could be something else if you're using your own wrapper function somewhere.

NathanLovato:
This is looking good to me, it's clearer than what I wrote initially. I just commented on the inferred type part. I'd say it doesn't hurt to mention why it "doesn't work" in the get_node() case, so readers understand why we recommend not to do that.

akien-mga:
@nathanleiby Could you update this PR based on @NathanLovato's review?

getting_started/scripting/gdscript/gdscript_styleguide.rst
nathanleiby: Thanks for the feedback! clarified the `get_node` case: 
512bd80

nathanleiby:
@akien-mga updated! @NathanLovato thanks much for your earlier feedback and apologies for dropping this

NathanLovato:
Thanks much!

PR #3691, opened 2020-06-14
Removed unnecessary files from ZIP files

Jayman2000:
Specifically, I removed stuff like the .import/ folder or .DS_Store files. To do so, I first added the contents of [GitHub's Godot gitignore file](https://github.com/github/gitignore/blob/18aa6d83774d514da479d73a4beb864cd4220231/Godot.gitignore) to the .gitignore file (I didn't commit this change). I then changed to the repository's root directory and ran the following bash script:
```
#!/bin/bash

# Thanks to https://askubuntu.com/a/344418
IFS=$'\n'       # make newlines the only separator

for archive in $(git ls-files *.zip)
do
	echo $archive
	echo ====================================================
	for to_delete in $(zipinfo -1 $archive | git check-ignore --stdin)
	do
		zip -d "$archive" "$to_delete"
	done
	echo
done
```
That script looks through all of the ZIP files that were being tracked by Git. It deletes any files inside those ZIP files that would have been ignored by Git if they weren't compressed.


Calinou:
I would avoid modifying the ZIP files in this repository, as committing new revisions of binary files bloats the repository. We should move each demo to a dedicated repository instead.

See https://github.com/godotengine/godot-docs/issues/3390.

akien-mga:
Indeed, I prefer we don't modify those ZIPs as it adds to the size of the Git repository. Their next edit should be a removal once they have been moved to a more appropriate location.

PR #3693, opened 2020-06-15
Mention the Node2D shearing property in Matrices and transforms

### LINKED ISSUE #3604: Shearing is now possible
**Issue description:**

There is an outdated note about shearing in the docs
> You can’t set the raw values of a Transform2D in the editor, so you must use code if you want to shear the object.

https://docs.godotengine.org/en/latest/tutorials/math/matrices_and_transforms.html

Calinou:
This closes #3604.

### DIFF
--- a/tutorials/math/matrices_and_transforms.rst
+++ b/tutorials/math/matrices_and_transforms.rst
@@ -235,7 +235,7 @@ Putting it all together
 ~~~~~~~~~~~~~~~~~~~~~~~
 
 We're going to apply everything we mentioned so far onto one transform.
-To follow along, create a simple project with a Sprite node and use the 
+To follow along, create a simple project with a Sprite node and use the
 Godot logo for the texture resource.
 
 Let's set the translation to (350, 150), rotate by -0.5 rad, and scale by 3.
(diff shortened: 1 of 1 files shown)


akien-mga:
Note: Don't merge just yet as I'm not 100% we'll actually keep the skew feature cherry-pick in 3.2.2 due to https://github.com/godotengine/godot/issues/38868 - if it messes up version control too much, it's too early to include in a stable build IMO.

@Calinou FYI, I'm currently syncing the `master` and `3.2` branches by merging master into 3.2 and reverting 4.0-specific changes, so no need to put `cherrypick:3.2` on PRs. We can use it once the branches fully diverge and such branch merges are no longer possible/meaningful.

tutorials/math/matrices_and_transforms.rst
```diff
@@ -284,13 +284,15 @@ Shearing the transformation matrix (advanced)
           explores an uncommonly used aspect of transformation matrices
           for the purpose of building an understanding of them.
 
+          Since Godot 3.2.2, Node2D provides a shearing property out of the box.
```
akien-mga, line 287: I reverted it, so it's no longer true. I might add it back in 3.2.3 if we find a fix for the issue it introduced.

aaronfranke:
Technically, skewing and shearing are different things.

* Skewing rotates the Y vector relative to the X vector, but area is not preserved (a 90 degree skew results in a shape with zero area).

* Shearing changes the angle between the axes by adjusting the position of the Y vector in the direction of the X vector (Y's X coordinate, or `t.y.x`, if un-rotated), and it preserves area, but the length of the Y vector changes.

IMO shearing is better for explaining matrices and transforms since it only involves changing one number in the matrix (for an un-rotated transform), so I think the article is best kept as-is with shearing, but we don't want to get it mixed up with skewing. Maybe we could include a section for skewing below the section for shearing, or maybe we can just mention that skewing/shearing are different?

tutorials/math/matrices_and_transforms.rst
akien-mga: Please remove the reference to 3.2.2. It hasn't been cherry-picked back yet, so it might not make it to 3.2.x any time soon.

akien-mga:
Thanks!

PR #3694, opened 2020-06-15
Add a style guide for project organization

### LINKED ISSUE #3104: Create a style guide for project organization

Calinou:
This closes #3104.

### DIFF
--- a/getting_started/workflow/project_setup/project_organization.rst
+++ b/getting_started/workflow/project_setup/project_organization.rst
@@ -44,6 +44,23 @@ They can then use a separate folder to store built levels that use them.
     /characters/npcs/suzanne/suzanne.png
     /levels/riverdale/riverdale.scn
 
+Style guide
+-----------
+
+For consistency across projects, we recommend following these guidelines:
+
+- Use **snake_case** for folder and file names (with the exception of C#
+  scripts). This sidesteps case sensitivity issues that can crop up after
+  exporting a project on Windows. C# scripts are an exception to this rule,
+  as the convention is to name them after the class name which should be
+  in PascalCase.
+- Use **PascalCase** for node names, as this matches built-in node casing.
+- In general, keep third-party resources in a top-level ``addons/`` folder, even
+  if they aren't editor plugins. This makes it easier to track which files are
+  third-party. There are some exceptions to this rule; for instance, if you use
+  third-party game assets for a character, it makes more sense to include them
+  within the same folder as the character scenes and scripts.
+
 Importing
 ---------
 


akien-mga:
Thanks!

Shatur95:
@Calinou, you wrote that only C# scripts are the exception for `snake_case`, does this mean that` .gd` scripts should be named in `snake_case`?

Calinou:
@Shatur95 Yes; the automatically-chosen names currently don't follow this, but this should be changed in the editor code itself.

PR #3695, opened 2020-06-15
Change C# page titles to describe the pages' contents

Calinou:
The page title will be used in the tab name, which can impact search engine results as well as general usability. Using just "Introduction" or "Features" doesn't make it very meaningful.


akien-mga:
Thanks!

PR #3699, opened 2020-06-16
C# Correct how to type cast Dictionaries returned from JSONParseResult

asteinwedel:
Casting with either `(Godot.Collections.Dictionary<string,
object>)parsedResult` or `parsedResult as
Godot.Collections.Dictionary<string, object>` fails with an
InvalidCastException (or silently returning null with the `as` keyword).
Instead, a typed constructor using `new` should be used.

[JsonParseExample.zip](https://github.com/godotengine/godot-docs/files/4786685/JsonParseExample.zip) shows all 3 ways

### DIFF
--- a/tutorials/io/saving_games.rst
+++ b/tutorials/io/saving_games.rst
@@ -267,7 +267,7 @@ load function:
         while (saveGame.GetPosition() < save_game.GetLen())
         {
             // Get the saved dictionary from the next line in the save file
-            var nodeData = (Godot.Collections.Dictionary<string, object>)JSON.Parse(saveGame.GetLine()).Result;
+            var nodeData = new Godot.Collections.Dictionary<string, object>((Godot.Collections.Dictionary)JSON.Parse(saveGame.GetLine()).Result);
 
             // Firstly, we need to create the object and add it to the tree and set its position.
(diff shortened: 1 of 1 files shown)


tutorials/io/saving_games.rst
```diff
@@ -267,7 +267,7 @@ load function:
         while (saveGame.GetPosition() < save_game.GetLen())
         {
             // Get the saved dictionary from the next line in the save file
-            var nodeData = (Godot.Collections.Dictionary<string, object>)JSON.Parse(saveGame.GetLine()).Result;
+            var nodeData = new Godot.Collections.Dictionary<string, object>(JSON.Parse(saveGame.GetLine()).Result as Godot.Collections.Dictionary);
```
neikeq, line 270: It's a good practice to never use the `as` operator unless you will check if the conversion succeeded. It's better to do a normal cast and get a cast exception if it fails instead of a null exception somewhere after.

neikeq, line 270:
```suggestion
            var nodeData = new Godot.Collections.Dictionary<string, object>((Godot.Collections.Dictionary)JSON.Parse(saveGame.GetLine()).Result);
```

asteinwedel: 100% agree good point!

Updated

akien-mga:
Thanks!

PR #3701, opened 2020-06-16
Fix `vformat()` example in Common engine methods and macros

Calinou:
Thanks to Noshyaar on Twitter for reporting it :slightly_smiling_face:


akien-mga:
Thanks!

PR #3702, opened 2020-06-16
Add note about register_types file locations

asmaloney:
I organize my projects with 'include' and 'src' directories. When I put the
register_types.[h,cpp] in them, the module would "compile fine" (no errors),
but it wouldn't work.

This took a long time to figure out, so this PR adds a note to ensure that
these files are in the right place to be picked up by the build.

### DIFF
--- a/development/cpp/binding_to_external_libraries.rst
+++ b/development/cpp/binding_to_external_libraries.rst
@@ -75,7 +75,12 @@ need to be created:
     register_types.h
     register_types.cpp
 
-With the following contents:
+.. note::
+    **Important:** These files must be in the top level folder of your module
+    (next to your ``SCsub`` and ``config.py`` files) in order for the module
+    to be registered properly.
+
+These files should contain the following:
 
 .. code-block:: cpp
 
--- a/development/cpp/binding_to_external_libraries.rst
+++ b/development/cpp/binding_to_external_libraries.rst
@@ -75,10 +75,9 @@ need to be created:
     register_types.h
     register_types.cpp
 
-.. note::
-    **Important:** These files must be in the top level folder of your module
-    (next to your ``SCsub`` and ``config.py`` files) in order for the module
-    to be registered properly.
+.. important::
+    These files must be in the top-level folder of your module (next to your
+    ``SCsub`` and ``config.py`` files) for the module to be registered properly.
 
 These files should contain the following:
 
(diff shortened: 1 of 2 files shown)


development/cpp/binding_to_external_libraries.rst
```diff
@@ -75,7 +75,12 @@ need to be created:
     register_types.h
     register_types.cpp
 
-With the following contents:
+.. note::
+    **Important:** These files must be in the top level folder of your module
```
Calinou, line 79: There's a Sphinx admonition for that™ :wink: 

```suggestion
.. important::
    These files must be in the top-level folder of your module
```

development/cpp/binding_to_external_libraries.rst
```diff
@@ -75,7 +75,12 @@ need to be created:
...
+    (next to your ``SCsub`` and ``config.py`` files) in order for the module
```
Calinou, line 80: Remove "in order" as it's often redundant:

```suggestion
    (next to your ``SCsub`` and ``config.py`` files) for the module
```

development/cpp/custom_modules_in_cpp.rst
```diff
@@ -119,7 +119,12 @@ need to be created:
     register_types.h
     register_types.cpp
 
-With the following contents:
+.. note::
+    **Important:** These files must be in the top level folder of your module
```
Calinou, line 123: Same as above:

```suggestion
.. important::
    These files must be in the top-level folder of your module
```

development/cpp/custom_modules_in_cpp.rst
```diff
@@ -119,7 +119,12 @@ need to be created:
...
+    (next to your ``SCsub`` and ``config.py`` files) in order for the module
```
Calinou, line 124: Same as above:

```suggestion
    (next to your ``SCsub`` and ``config.py`` files) for the module
```

development/cpp/binding_to_external_libraries.rst
asmaloney: Aha! I'm new to Sphinx (Godot seems to use a number of tools I've never used before :-))

I was following what was used below - do you want me to change those as well?

asmaloney: There's also a `**note**:` in `custom_modules_in_cpp.rst` - change that one to `.. note::`?

Calinou: > I was following what was used below - do you want me to change those as well?

Yes, it'd be better to change them as well for consistency.

> There's also a **note**: in custom_modules_in_cpp.rst - change that one to .. note::?

Sure :slightly_smiling_face:

asmaloney:
Thanks @Calinou!

PR #3705, opened 2020-06-17
Install tools required by `format.sh` on GitHub Actions

Calinou:
The script would silently fail otherwise. Thanks to @aaronfranke for catching it :slightly_smiling_face:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,9 @@ jobs:
 
       - name: Lint and build documentation
         run: |
+          # Install tools used by `_tools/format.sh`.
+          sudo apt-get update -qq
+          sudo apt-get install -qq dos2unix recode
           # Append the `PATH` so we can run `codespell`.
           export PATH="$HOME/.local/bin:$PATH"
           pip3 install -r requirements.txt


akien-mga:
`format.sh` should likely be made to error out properly if those tools are missing?

aaronfranke:
Perhaps something like this, above the "Loops" comment:

```bash
if ! type dos2unix; then
    echo "Error: dos2unix is not installed."
    exit 2
fi
if ! type recode; then
    echo "Error: recode is not installed."
    exit 3
fi
```

akien-mga:
But shouldn't the script already exit with a non-zero exit code when trying to run a command that doesn't exist in PATH?

aaronfranke:
Just tested it, when running a command that doesn't exist, the script exits with `0`.

akien-mga:
> Just tested it, when running a command that doesn't exist, the script doesn't error and proceeds to exit with `0`.

There's an issue with the way the script is called then. It should return non `0`:
```
$ cat test.sh 
#!/bin/sh

cmd_that_doesnt_exist

$ sh test.sh 
test.sh: line 3: cmd_that_doesnt_exist: command not found

$ echo $?
127
```

But this isn't blocking for merging this PR anyway.

akien-mga:
Thanks!

PR #3708, opened 2020-06-19
Remove duplicate from GDScript Style Guide in Indentation

Whimfoome:
I noticed in GDScript style guide at "Indentation", the **Bad** showed an example of **Good**, then **Bad**

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_styleguide.rst
+++ b/getting_started/scripting/gdscript/gdscript_styleguide.rst
@@ -118,9 +118,6 @@ Each indent level should be one greater than the block containing it.
 ::
 
     for i in range(10):
-      print("hello")
-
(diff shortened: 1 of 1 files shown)


Calinou:
These are not duplicates. See how they look with visible whitespace:

Whimfoome:
Oh, it's hard to spot the difference. Maybe it should be made with just 1 space instead of 2 to make it more noticeable without visible whitespace?

pycbouh:
It wouldn't make much sense to show it with one space instead, because two spaces are quite common convention and thus make it a more reasonable example.

See also #3437, we considered adding more visible information to this example, but ultimately the implementation was hacky and limited by the tools at our disposal. A more straightforward solution is welcome.

PR #3709, opened 2020-06-19
Added note about global shader arrays

Chaosus:
Since this added by https://github.com/godotengine/godot/pull/32751

### DIFF
--- a/tutorials/shading/shading_reference/shading_language.rst
+++ b/tutorials/shading/shading_reference/shading_language.rst
@@ -195,13 +195,12 @@ Arrays
 ------
 
 Arrays are containers for multiple variables of a similar type.
-Note: As of Godot 3.2, only local and varying arrays have been implemented.
 
 Local arrays
 ~~~~~~~~~~~~
 
 Local arrays are declared in functions. They can use all of the allowed datatypes, except samplers.
-The array declaration follows a C-style syntax: ``typename + identifier + [array size]``.
+The array declaration follows a C-style syntax: ``[const] + [precision] + typename + identifier + [array size]``.
 
 .. code-block:: glsl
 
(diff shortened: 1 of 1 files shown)


tutorials/shading/shading_reference/shading_language.rst
```diff
@@ -252,6 +252,22 @@ Arrays also have a built-in function ``.length()`` (not to be confused with the
...
+.. code-block:: glsl
+    shader_type spatial;
+
+    const lowp vec3 v[1] = lowp vec3[1] ( vec3(0, 0, 1) );
+
+    void fragment() {
+      ALBEDO = v[0]; 
+    }
+
+The only difference in syntax from local shader arrays is the required constness.
```
clayjohn, line 269:
```suggestion
.. note: 
    Global arrays have to be declared as global constants, otherwise they can be declared the same as local arrays.
```

Chaosus: Changed. Note: At global level they cannot be declared without const.

tutorials/shading/shading_reference/shading_language.rst
```diff
@@ -195,13 +195,13 @@ Arrays
 ------
 
 Arrays are containers for multiple variables of a similar type.
-Note: As of Godot 3.2, only local and varying arrays have been implemented.
+Note: As of Godot 4.0, local, global and varying arrays have been implemented.
```
clayjohn, line 198: This is unnecessary now. But we should mention in the uniform section that we still don't support uniform arrays.

Chaosus: Ok, added

akien-mga:
Thanks!

PR #3720, opened 2020-06-22
high-level networking: remove bad practice; bomber demo more explicit

Meriipu:
Removed an instance of a bad practice (letting the clients pass their
own rpc caller id to the server and having the server trust it), as
well as making the server the only one able to unpause the game.

The first of the two is a fairly harmful habit to teach, especially
in a first introduction.

Also made the bomber demo way more explicit (and verbose) by avoiding
as many ambiguous pronouns as possible, instead being quite explicit
about which nodes or methods are being referred to at any time, and
also some intermingled motivations/reminders for why these steps are
being taken.

### DIFF
--- a/tutorials/networking/high_level_multiplayer.rst
+++ b/tutorials/networking/high_level_multiplayer.rst
@@ -291,7 +291,8 @@ every peer and RPC will work great! Here is an example:
             get_node("/root/world/players").add_child(player)
 
         # Tell server (remember, server is always ID=1) that this peer is done pre-configuring.
-        rpc_id(1, "done_preconfiguring", selfPeerID)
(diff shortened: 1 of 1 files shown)


akien-mga:
CC @Faless

Meriipu:
There may be some leftover issues. It turns out that some RPCs get sent (but never accepted; they are rightfully flagged as errors on the other end) despite the master/puppet flags of the client suggesting otherwise (and there is some asymmetry here in that the server does not "wrongly" send the RPC to clients in a similar fashion). 

I assume it is a bug rather than intentional but I might be wrong. Going to be opening an issue for it.

PR #3722, opened 2020-06-23
Remove error that Dictionary starts with 1000 records.

willnationsdev:
When initially writing this, I misread the source code, interpreting a bitwise operation to be applied in base 10 rather than base 2...cause I'm an idiot. Therefore, many people have been likely misinformed, believing Dictionary to be horribly inefficient as you create new instances of them. This is a gross mistake that should never have made it into the "best practices" documentation.

### DIFF
--- a/getting_started/workflow/best_practices/data_preferences.rst
+++ b/getting_started/workflow/best_practices/data_preferences.rst
@@ -103,11 +103,12 @@ Contiguous memory stores imply the following operation performance:
       ordered-aware search algorithm.
 
 Godot implements Dictionary as an ``OrderedHashMap<Variant, Variant>``. The engine
-stores a giant array (initialized to 1000 records) of key-value pairs. When
+stores a small array (initialized to 2^3 or 8 records) of key-value pairs. When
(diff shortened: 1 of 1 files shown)


getting_started/workflow/best_practices/data_preferences.rst
```diff
@@ -121,11 +122,12 @@ the expense of memory and some minor operational efficiency.
...
       Which leads to...
 
-2. Maintaining a huge size for the table.
+2. Maintaining an ever-growing size for the table.
 
-    - The reason it starts with 1000 records, and the reason it forces
-      large gaps of unused memory interspersed in the table is to
-      minimize hash collisions and maintain the speed of the accesses.
+    - HashMaps maintain gaps of unused memory interspersed in the table
+      on purpose in order to reduce hash collisions and maintain the speed of
```
Calinou, line 128: This can be simplified:

```suggestion
      on purpose to reduce hash collisions and maintain the speed of
```

akien-mga:
Thanks!

PR #3728, opened 2020-06-24
Update C# preprocessor defines and add iOS

aaronfranke:
C# is now supported on iOS, so I added it to the list of possible platforms and added the iOS-specific defines.

Also, I corrected a mistake about the texture compression being "one of", while it's possible to have multiple, so I changed it to say "any of" instead.

Also, link to the OS testing demo. Therefore, this PR depends on https://github.com/godotengine/godot-demo-projects/pull/486

This can/should be cherry-picked for 3.2, but only with `s/LINUXBSD/X11`


akien-mga:
Thanks!

PR #3737, opened 2020-06-27
Update exporting_for_android.rst

CyrilBos:
I ran the original command and keytool printed:                                                                                                                                                         ```Warning: 
The JKS keystore uses a proprietary format. It is recommended to migrate to PKCS12 which is an industry standard format using "keytool -importkeystore -srckeystore debug.keystore -destkeystore debug.keystore -deststoretype pkcs12". ```

So I added the `-deststoretype pkcs12` to the command used for the generation. Maybe we should also add a note for old projects to migrate using that command printed by keytool. 

Also, there is a default directory Android/keystores when you install the SDK (tested only on Windows 10 but I think it is the same on Mac & Linux), so maybe we should propose that as the default location for the debug.keystore


akien-mga:
Thanks!

PR #3739, opened 2020-06-27
Removed unneeded code from resource_queue.gd

Rubonnek:
`VisualServer.sync()` is unimplemented in 3.2 at least according to the docs.

I'm not sure why we need the `OS.delay_usec` either.

### DIFF
--- a/tutorials/io/files/resource_queue.gd
+++ b/tutorials/io/files/resource_queue.gd
@@ -81,8 +81,6 @@ func is_ready(path):
 func _wait_for_resource(res, path):
 	_unlock("wait_for_resource")
 	while true:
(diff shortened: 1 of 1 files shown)


bruvzg:
> `VisualServer.sync()` is unimplemented in 3.2 at least according to the docs.

`sync` is implemented in `VisualServerWrapMT` (multi-threaded version of the `VisualServer`).

> I'm not sure why we need the OS.delay_usec either.

It's there to avoid excessive CPU load (while loop doing checks constantly), with the delay it's doing the check only once per frame.

Both lines seem to be reasonable.

Rubonnek:
I see. Thanks for the clarification!

PR #3743, opened 2020-06-29
Add a FAQ item about why Godot can't be used as a library

Calinou:
This is a relatively common question on the forums and Q&A platform, so it makes sense to document it here.

### DIFF
--- a/about/faq.rst
+++ b/about/faq.rst
@@ -289,6 +289,18 @@ developer experiences as a whole.
 Bonus points for bringing screenshots, concrete numbers, test cases, or example
 projects (if applicable).
 
+Is it possible to use Godot as a library?
+-----------------------------------------
+
+Godot is meant to be used with its editor. We recommend you give it a try, as it
+will most likely save you time in the long term. There are no plans to make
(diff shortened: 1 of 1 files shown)


about/faq.rst
```diff
@@ -289,6 +289,22 @@ developer experiences as a whole.
...
+Godot is meant to be used with its editor. We recommend you give it a try, as it
+will most likely save you time in the long term. There are no plans to make
+Godot usable as a library, as it would make the rest of the engine more
+convoluted and difficult to use for casual users.
+
+If you want to use a rendering library, look into using established rendering
+engines such as `Magnum <https://magnum.graphics/>`_,
+`OpenSceneGraph <http://www.openscenegraph.org/>`_ or
+`Urho3D <https://urho3d.github.io/>`_. (These are only recommendations,
+as there are dozens of rendering engines out there.)
```
akien-mga, line 304: I'm not sure it's a good idea to give examples here, it's bound to be very subjective and will change over time.
At most I'd list to an *awesome* list of open source rendering engines.

akien-mga:
Thanks!

PR #3746, opened 2020-06-30
Update oculus quest tutorial image for 3.2.2

skyace65:
The export settings have changed slightly since the tutorial for the quest was first made so this updates that image. The image has gone through png compression.


akien-mga:
Thanks!

PR #3747, opened 2020-07-01
Removed "Constants" paragraph from migrating-to shader page

Chaosus:
Since there is no difference from GLSL already.


akien-mga:
Thanks!

akien-mga:
For the `3.2` branch, I guess we should not remove it but maybe edit the text a bit to mention which types of constants are supported (IIRC we support some but not all)?

Anutrix:
There is also a mention about using uniforms instead of constants in Macros Section below this. Maybe that needs to be changed too.

Chaosus:
In 3.2 branch arrays constants are not supported + I also detected that constants are not supported as function parameters like:
```
void test(const int t) {
}
```
I want to work on this soon...

PR #3748, opened 2020-07-01
Remove references for the now removed "Constants" section in the shad…

YeldhamDev:
The "Constants" section was removed in #3747, but the references to it weren't.


akien-mga:
Thanks!

PR #3749, opened 2020-07-01
Document how to clean generated files using SCons

### LINKED ISSUE #3717: Intro to buildsystem: Add section about `scons --clean`.
**Your Godot version:**
Any.

**Issue description:**
The introduction to the buildsystem page should cover the usage of SCons' `-c` / `--clean` flag to remove build artifacts. In particular, it should mention that the flag has to be used with the same options as the build command line, to let SCons know what objects to remove (e.g. `scons p=x11 tools=yes target=release_debug -c` would remove the X11 editor release build objects (`bin/godot.x11.opt.tools.64` and friends).

It could also mention `git clean -fxd` as a way to remove *all* untracked files from the Git repository, with the caveat that this can also remove files that the user intends to preserve (like IDE configuration files).

https://docs.godotengine.org/en/stable/development/compiling/introduction_to_the_buildsystem.html

Calinou:
This closes #3717.

### DIFF
--- a/development/compiling/introduction_to_the_buildsystem.rst
+++ b/development/compiling/introduction_to_the_buildsystem.rst
@@ -222,6 +222,18 @@ features to include/disable.
 Check the output of ``scons --help`` for details about each option for
 the version you are willing to compile.
 
+Cleaning generated files
+------------------------
+
+Sometimes, you may encounter an error due to generated files being present. You
+can remove them by using ``scons --clean <options>``, where ``<options>`` is the
+list of build options you've used to build Godot previously.
+
+Alternatively, you can use ``git clean -fixd`` which will clean build artifacts
+for all platforms and configurations. Beware, as this will remove all untracked
+and ignored files in the repository. Don't run this command if you have
+uncommitted work!
+
 .. _doc_overriding_build_options:
 
 Overriding the build options


development/compiling/introduction_to_the_buildsystem.rst
```diff
@@ -222,6 +222,18 @@ features to include/disable.
 Check the output of ``scons --help`` for details about each option for
 the version you are willing to compile.
 
+Cleaning generated files
+------------------------
+
+Sometimes, you may encounter an error due to generated files being present. You
+can remove them by using ``scons --clean <options>``, where ``<options>`` is the
```
bruvzg, line 229: Is SCons config changed to remove all `.gen` files? Last time there was issue with `.gen` files, `scons --clean` was removing only `.gen` files that are valid in the current config, and issue was caused by old file that was renamed in the new version of source and was not affected by `--clean`.

development/compiling/introduction_to_the_buildsystem.rst
```diff
@@ -222,6 +222,18 @@ features to include/disable.
...
+list of build options you've used to build Godot previously.
+
+Alternatively, you can use ``git clean -dfxi`` which will clean build artifacts
```
akien-mga, line 232: I use `-fxd`, either to remember ("fixed"). Could be `-fixd` with the `i` switch too.

development/compiling/introduction_to_the_buildsystem.rst
akien-mga: Good point, it's not going to remember generated files that are not/no longer generated by the current config.

It's still a rare occurrence that we change generated files, so IMO it should be documented as a good way to clean build objects to make a clean build. But for the gen files workaround, a full got clean is needed indeed (or `find -name "*.gen.*" -delete` which I often use).

bruvzg: Well, I saw "generated" and was thinking only about `.gen.` file issues, for cleaning up object files `scons --clean` is a good option indeed. Probably both should be mentioned.

bruvzg: > `find -name "*.gen.*" -delete`

At least on macOS find require path as it's first argument as it will fail with `find: illegal option -- n` error, `find . -name "*.gen.*" -delete` will work.

akien-mga:
Thanks!

akien-mga:
This added a bug in the TOC, as the next section was a subsection of the previous one: https://docs.godotengine.org/en/latest/development/compiling/introduction_to_the_buildsystem.html#cleaning-generated-files

PR #3750, opened 2020-07-01
[macOS] Add instructions for Apple Silicon build and MoltenVK library bundling

bruvzg:
- Add instructions for building macOS binaries for Apple Silicon and combining them into Universal binary.
- Add instructions for bundling MoltenVK libraries to the editor `.app` bundle, to use and distribute it without LunarG SDK installation.

### DIFF
--- a/development/compiling/compiling_for_osx.rst
+++ b/development/compiling/compiling_for_osx.rst
@@ -33,9 +33,19 @@ If you are building the ``master`` branch:
 Compiling
 ---------
 
-Start a terminal, go to the root directory of the engine source code and type::
+Start a terminal, go to the root directory of the engine source code.
 
-    scons platform=osx --jobs=$(sysctl -n hw.logicalcpu)
+To compile for Intel (x86-64) powered Macs, use::
+
+    scons platform=osx arch=x86_64 --jobs=$(sysctl -n hw.logicalcpu)
+
+To compile for Apple Silicon (ARM64) powered Macs, use::
+
+    scons platform=osx arch=arm64 --jobs=$(sysctl -n hw.logicalcpu)
+
+To support both architectures in a single "Universal 2" binary, run the above two commands and then use ``lipo`` to bundle them together::
+
+    lipo -create bin/godot.osx.tools.x86_64 bin/godot.osx.tools.arm64 -output bin/godot.osx.tools.universal
 
 If all goes well, the resulting binary executable will be placed in the
 ``bin/`` subdirectory. This executable file contains the whole engine and
@@ -49,12 +59,17 @@ manager.
 
 To create an ``.app`` bundle like in the official builds, you need to use the
 template located in ``misc/dist/osx_tools.app``. Typically, for an optimized
-editor binary built with ``scons p=osx target=release_debug``::
+editor binary built with ``target=release_debug``::
 
     cp -r misc/dist/osx_tools.app ./Godot.app
     mkdir -p Godot.app/Contents/MacOS
-    cp bin/godot.osx.tools.64 Godot.app/Contents/MacOS/Godot
+    cp bin/godot.osx.opt.tools.universal Godot.app/Contents/MacOS/Godot
     chmod +x Godot.app/Contents/MacOS/Godot
+
+If you are building the ``master`` branch, additionally copy the Vulkan library:
+
+    mkdir -p Godot.app/Contents/Frameworks
+    cp <Vulkan SDK path>/macOS/libs/libMoltenVK.dylib Godot.app/Contents/Frameworks/libMoltenVK.dylib
 
 Compiling a headless/server build
 ---------------------------------


development/compiling/compiling_for_osx.rst
```diff
@@ -17,7 +17,7 @@ For compiling under macOS, the following is required:
 - *Optional* - `yasm <https://yasm.tortall.net/>`_ (for WebM SIMD optimizations).
 
 If you are building the ``master`` branch:
-- Download and install the `Vulkan SDK for macOS <https://vulkan.lunarg.com/sdk/home>`__.
+- Download and install (install SDK only if you want to use standalone executable) the `Vulkan SDK for macOS <https://vulkan.lunarg.com/sdk/home>`__.
```
akien-mga, line 20: This addition is not clear to me, and the insert makes the sentence a bit weird. I'd suggest rephrasing.

development/compiling/compiling_for_osx.rst
```diff
@@ -35,7 +35,17 @@ Compiling
 
 Start a terminal, go to the root directory of the engine source code and type::
 
-    scons platform=osx --jobs=$(sysctl -n hw.logicalcpu)
+- for Intel (x86-64) powered Macs
```
akien-mga, line 38: The previous line ends with `::` which starts a code block, I don't think this will work as expected here. I'd suggest rephrasing to whole section:
```
Start a terminal and go to the root directory of the engine source code.

To compile for Intel (x86-64) powered Macs, use::

    scons platform=osx arch=x86_64 --jobs=$(sysctl -n hw.logicalcpu)

To compile for Apple Silicon (ARM64) powered Macs, use::

    scons platform=osx arch=arm64 --jobs=$(sysctl -n hw.logicalcpu)

To support both architectures in a single "Universal 2" binary, run the above two commands and then use ``lipo`` to bundle them together::

    lipo -create bin/godot.osx.tools.x86_64 bin/godot.osx.tools.arm64 -output bin/godot.osx.tools.universal
```

development/compiling/compiling_for_osx.rst
```diff
@@ -53,9 +63,14 @@ editor binary built with ``scons p=osx target=release_debug``::
 
     cp -r misc/dist/osx_tools.app ./Godot.app
     mkdir -p Godot.app/Contents/MacOS
-    cp bin/godot.osx.tools.64 Godot.app/Contents/MacOS/Godot
+    cp bin/godot.osx.tools.universal Godot.app/Contents/MacOS/Godot
```
akien-mga, line 66: In the paragraph above, `scons p=osx target=release_debug` is no longer accurate if we're using the universal version as the reference. I'd suggest to replace it by "built with ``target=release_debug``" only.

development/compiling/compiling_for_osx.rst
```diff
@@ -53,9 +63,14 @@ editor binary built with ``scons p=osx target=release_debug``::
...
     chmod +x Godot.app/Contents/MacOS/Godot
 
+If you are building the ``master`` branch, additionally copy Vulkan library:
```
akien-mga, line 69:
```suggestion
If you are building the ``master`` branch, additionally copy the Vulkan library:
```

development/compiling/compiling_for_osx.rst
```diff
@@ -53,9 +63,14 @@ editor binary built with ``scons p=osx target=release_debug``::
...
+
+    mkdir -p Godot.app/Contents/Frameworks
+    cp {VULKAN_SDK_PATH}/macOS/libs/libMoltenVK.dylib Godot.app/Contents/Frameworks/libMoltenVK.dylib
```
akien-mga, line 72: Is `{VULKAN_SDK_PATH}` an env variable here (and might thus lack `$`), or a placeholder for the user to fill?
If it's a placeholder, I'd suggest using a format that is not so reminiscent of an env varibale, e.g. `<Vulkan SDK path>/macOS/...`.

development/compiling/compiling_for_osx.rst
bruvzg: System-wide (to `/usr/local`) install script is messy, can cause conflicts, and pointless if you want to use `.app` bundle, but can be convenient for stand-alone executable.

But since current SDK documentation "Install the SDK" page say "mount a disk image file and copy the contents into a desired directory", this addition is probably excessive. I'll remove it completely.

development/compiling/compiling_for_osx.rst
bruvzg: and its `godot.osx.opt.tools.universal` in case of `target=release_debug`

development/compiling/compiling_for_osx.rst
bruvzg: placeholder,

akien-mga:
Thanks!

PR #3752, opened 2020-07-03
Add clamp to custom light function example

bmolyneaux:
Without this clamp, negative lighting will be applied to the back side of the object.

### DIFF
--- a/tutorials/shading/shading_reference/spatial_shader.rst
+++ b/tutorials/shading/shading_reference/spatial_shader.rst
@@ -293,7 +293,7 @@ Below is an example of a custom light function using a Lambertian lighting model
 .. code-block:: glsl
 
     void light() {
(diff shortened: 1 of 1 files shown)


tutorials/shading/shading_reference/spatial_shader.rst
```diff
@@ -293,7 +293,7 @@ Below is an example of a custom light function using a Lambertian lighting model
 .. code-block:: glsl
 
     void light() {
-        DIFFUSE_LIGHT += dot(NORMAL, LIGHT) * ATTENUATION * ALBEDO;
+        DIFFUSE_LIGHT += clamp(dot(NORMAL, LIGHT), 0, 1) * ATTENUATION * ALBEDO;
```
clayjohn, line 296: Please change to 0.0 and 1.0. 

Some GLES2 technically cannot convert from ints to floats implicitly. So the change is needed to ensure this works on all hardware.

akien-mga:
Thanks!

PR #3755, opened 2020-07-03
Add OpenXR plugin to list of VR plugins

skyace65:
Adds the OpenXR plugin to the list of official VR plugins.


akien-mga:
Thanks!

PR #3756, opened 2020-07-03
Clarify about "Clear mode"

SimplyCmd:
A small bit of clarification relating to this issue: https://github.com/godotengine/godot/issues/33831
Let me know if this needs to be reworded :)


akien-mga:
The class reference should be edited in the main Godot repository. Files in `classes/` are generated by a script, so any change done here would be lost.

https://github.com/godotengine/godot/blob/master/doc/classes/Viewport.xml

PR #3759, opened 2020-07-04
Update enum usage in "Import plugins"

rgson:
The examples using the `Presets` enum don't work as-is. The enum key must be referenced with the enum name prefixed. According to [GDScript basics](https://docs.godotengine.org/en/stable/getting_started/scripting/gdscript/gdscript_basics.html) this was changed in 3.1.


akien-mga:
Thanks!

PR #3760, opened 2020-07-05
disable smartquotes

binotaliu:
Smartquotes will convert `"..."` to `“...”`, will cause unexpected result when we reference a string without using code block.

also in some languages (zh_TW for example), it will be converted to `「...」`, usually not a desired result.


akien-mga:
Thanks!

PR #3763, opened 2020-07-07
add: zh_TW title

binotaliu:
it's over 30% on Weblate now!

Though still a lot of works need to be done, I think I can open this PR just fine!

I don't want to open another PR whoever merge this please also add `zh_TW` to https://github.com/godotengine/godot-docs-l10n/blob/master/build_langs.txt


akien-mga:
Awesome work! I'll get a zh_TW instance created tomorrow on ReadTheDocs.

PR #3768, opened 2020-07-09
Clarify note on the connect dialogue

ihigg:

### DIFF
--- a/getting_started/step_by_step/scripting.rst
+++ b/getting_started/step_by_step/scripting.rst
@@ -331,7 +331,7 @@ Also, remember that nodes are referenced by name, not by type.
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


getting_started/step_by_step/scripting.rst
```diff
@@ -331,7 +331,7 @@ Also, remember that nodes are referenced by name, not by type.
 
 .. note::
 
-    The right-hand panel of the connect dialogue is for binding specific
+    The 'advanced' panel of the connect dialogue is for binding specific
```
ihigg, line 334: The panel in question only appears once the 'Advanced' slider is toggled. It seems more intuitive to refer to it like this instead.

akien-mga:
Thanks!

Issue #3772, opened 2020-07-10
Class examples needed


monk-e-boy:
**Your Godot version:**
3.2.2

**Issue description:**

Class page is a bit rubbish. It's really hard to figure out how to make a class. Classes are pretty fundimental and there should be a super clear example before the complex documentation

This screenshot from a random YouTube video explains classes in 10 seconds. The links below are useless.

![image](https://user-images.githubusercontent.com/17924807/87201838-519a5180-c2f7-11ea-88d5-edc772f61884.png)

None of us are coming to Godot as noobs, we know how to code. We just want to see how to create a class and move along with our lives. Simple examples are such a time saver. Better if we can copy and paste the code :)

**URL to the documentation page (if already existing):**

https://docs.godotengine.org/en/stable/development/cpp/object_class.html
https://docs.godotengine.org/en/stable/getting_started/workflow/best_practices/what_are_godot_classes.html

akien-mga:
Well you're an expert programmer, but you can also start with the basics: https://docs.godotengine.org/en/latest/getting_started/scripting/gdscript/gdscript_basics.html

monk-e-boy:
You talk down to me like I don't know how to code. I actually teach coding at degree level... so I do know how to code. I'm just saying the documentation needs work.

Maybe split that page up into sections? That make sense? Rather than a huge glob of every idea under the sun.

And it's not linked from the class page, so .... ? ... how would I find it?

I think you need to start with inner classes, then talk about how to move them into separate files and load them. 

![image](https://user-images.githubusercontent.com/17924807/87203273-249b6e00-c2fa-11ea-8485-5d145cede47f.png)

The first thing it says is "inside a class file" ... but doesn't explain what that means.

Calinou:
There should definitely be an example of instancing classes from a file in [this page](https://docs.godotengine.org/en/stable/getting_started/workflow/best_practices/what_are_godot_classes.html). The GDScript basics should remain basic (as its name implies), so it's better to just link to that page instead.

monk-e-boy:
![image](https://user-images.githubusercontent.com/17924807/87203815-4fd28d00-c2fb-11ea-98d9-b4a34d631df8.png)

I think this may need a bit of an edit. The text seems to imply that the code is saved in a file called character.gd, then the code seems to include it's self? Or is the capitalization of Character.gd, and character.gd important?

monk-e-boy:
[...]

PR #3773, opened 2020-07-10
Link to the Object class reference on the Object class page

Calinou:
People who end up on this page from a search engine might be confusedas to where they can find the Object class reference.

See https://github.com/godotengine/godot-docs/issues/3772.


akien-mga:
Thanks!

PR #3778, opened 2020-07-12
Fix missing semicolon in C# code sample in Exporting tutorial

derKosi:
Player.cs code (c#) won't compile because of a missing simicolon.


Calinou:
Thanks!

PR #3781, opened 2020-07-13
Clarify instructions in Exporting for dedicated servers

Calinou:
This also mentions how to export only a PCK file from a headless editor.


akien-mga:
Thanks!

PR #3784, opened 2020-07-13
Contributing: Editor and docs localization

akien-mga:
Took me a GitHub downtime this morning to finally spend the ~4 hours needed to write all this down.

CC @pyacier @binotaliu @Calinou


community/contributing/editor_and_docs_localization.rst
```diff
@@ -0,0 +1,412 @@
...
+    red = ColorN("red", 1)
+    [/codeblock]
+    Supported color names are the same as the constants defined in [Color].
+
+In the above example, ``[code]name[/code]``, ``[code]alpha[/code]``, and
+``[Color]`` should *not* be translated, as they refer respectively to argument
+names and a class of the Godot API. Similarly, the contents of the
+``[codeblock]`` should not be translated, as ``ColorN`` is a function of the
+Godot API and ``"red"`` is one of the named colors it supports. At most, you can
+translate the name of the variable which holds the result (``red = ...``).
```
akien-mga, line 319: I guess this will be useful, I've seen some nice stuff like this (since fixed):
```
Gibt eine Farbe gemäß dem standardisierten [code]Namen[/code] zurück, wobei [code]alpha[/code] im Bereich von 0 bis 1 liegt.
[codeblock]
rot = FarbeN("rot", 1)
[/codeblock]
Die unterstützten Farbnamen sind die gleichen wie die in [Farbe] definierten Konstanten.
```

binotaliu:
Can we also have a community channel for i18n? Like if we can have discord channels for each languages and anyone who has a `translator` role can see these channels?

pyacier:
It's really good, I wish i had this documentation when I started translating the documentation.

akien-mga:
> Can we also have a community channel for i18n? Like if we can have discord channels for each languages and anyone who has a `translator` role can see these channels?

I'd like to have a good communication channel for translators yeah, but I'm not sure what to use.
Adding up to 60 new channels to the Discord server (the number of locales we currently have) seems a bit hard to manage, and especially moderation would be quite tricky as we don't have mods that speak those 60 languages.
But I guess their could be a `#engine-localization` channel as a general place for all translators, and then private chatrooms could be opened for each locale.

I'd prefer having all this handled on Matrix server that we'd administrate, but we haven't set up one (yet).

PR #3785, opened 2020-07-13
Add 2x2 and 3x3 Minimal autotile templates to Using tilemaps

### LINKED ISSUE #3316: Tileset: Provide complete Bitmasks reference textures for 2x2, 3x3 minimal, and 3x3
[The documentation on Tileset and Tilemaps](https://docs.godotengine.org/en/3.2/tutorials/2d/using_tilemaps.html) desperately needs complete example Bitmask textures.
Noone really knows how to set them correctly, which just adds to the gigantic clusterfuck of a UI UX fail that this Tileset resouce is.
https://github.com/godotengine/godot/issues/37360

Calinou:
Thanks to @golddotasksquestions for providing these autotile templates :slightly_smiling_face: 

This closes #3316.


tutorials/2d/using_tilemaps.rst
```diff
@@ -211,25 +211,81 @@ can test for.
...
+This template can be used for sideways or fully top-down perspectives.
+It's designed for a TileMap cell size of 64x64.
+
+Key:
+
+- Red: "on"
+- White: "off"
+
+.. image:: img/autotile_template_2x2.png
+    :target: ../../_images/autotile_template_2x2.png
```
akien-mga, line 225: It's weird to have to hardcode a path like that, and it will likely break for localized versions.

Calinou: I needed to do this to make the image clickable (so it can be saved more easily). Otherwise, I guess we should write a small Sphinx extension for this.

akien-mga: I'd remove the hardcoded path for now, I guess the image can still be saved with right-click?

Calinou: Done. I added an explicit note about it, just in case.

akien-mga:
Thanks!

PR #3790, opened 2020-07-14
fix: validate with is_network_master is correct form.

rdcklinux:
**has_method** returns **true** even if node isn't the **master**
**is_network_master** returns correct validation.

### DIFF
--- a/tutorials/networking/high_level_multiplayer.rst
+++ b/tutorials/networking/high_level_multiplayer.rst
@@ -388,7 +388,7 @@ Example bomb code:
 ::
 
     for p in bodies_in_area:
-        if p.has_method("exploded"):
+        if p.is_network_master():
             p.rpc("exploded", bomb_owner)
 
 Example player code:


Faless:
I believe this is incorrect. `has_method` is used to make sure it's the right kind of object (i.e. not another physics object without that met6hod). Additionally, that function is called by puppets on the master, so it would actually prevent it from working.

rdcklinux:
> I believe this is incorrect. `has_method` is used to make sure it's the right kind of object (i.e. not another physics object without that met6hod). Additionally, that function is called by puppets on the master, so it would actually prevent it from working.
 
the method has_method doesn't work in this scope, when the game instance isn't master is called too.

rdcklinux:
> I believe this is incorrect. `has_method` is used to make sure it's the right kind of object (i.e. not another physics object without that met6hod). Additionally, that function is called by puppets on the master, so it would actually prevent it from working.

it's possible or sense add this method: is_master method() and is_puppet method()

Meriipu:
It is probably a good change.

The actual demo contains this:
```
bomb.gd-# Called from the animation.
bomb.gd:func explode():
bomb.gd-  if not is_network_master():
bomb.gd:    # Explode only on master.
bomb.gd-    return
bomb.gd-  for p in in_area:
bomb.gd:    if p.has_method("exploded"):
bomb.gd:      # Exploded has a master keyword, so it will only be received by the master.
bomb.gd:      p.rpc("exploded", from_player)
bomb.gd-
bomb.gd-
```

But the tutorial leaves the first three lines of explode implicit (which this pull-request is adding in, albeit in a different place). 
It might be a good idea to maybe instead include the whole definition of `func ecplode():` in the tutorial?

This does not change the often-raised issue of authoritativeness, which might still exist (but should probably be addressed separately).

jonbonazza:
perhaps it's just a matter of ANDing the two conditions?

```cpp
if (p.is_network_master() && p.has_method("exploded")) {
    // Do the thing...
}
```

Calinou:
Is this still relevant with the latest documentation as of 4.1?

Piralein:
The Page was updated by #7675 and cherry picked to all 4.x branches.

Closing as the change is no longer relevant.

PR #3793, opened 2020-07-15
Fix reference to the editor/docs localization guide in the "Ways to contribute" page

YeldhamDev:


Calinou:
I've had an issue where an otherwise valid reference was considered invalid by Sphinx. Clearing my local build folder using `make clean` solved the issue, but I don't know if you can do this on Read the Docs.

akien-mga:
Indeed, I could reproduce the issue in my local build from when I wrote the content, but a simple `make html` after poking at one of the files fixed it. It's still fixed even with no changes.

akien-mga:
It's the `.. highlight: none` before the title which breaks it.

akien-mga:
Superseded by b181d3f07fd10152cf71035db9333dfa7395dd0a.

PR #3795, opened 2020-07-15
Fix C# code in Input examples

akien-mga:
And work around two other warnings, making the build warning free.

### DIFF
--- a/classes/class_basematerial3d.rst
+++ b/classes/class_basematerial3d.rst
@@ -23,7 +23,7 @@ This provides a default material with a wide variety of rendering features and p
 Tutorials
 ---------
 
-- :doc:`../tutorials/3d/spatial_material`
+- :doc:`../tutorials/3d/standard_material_3d`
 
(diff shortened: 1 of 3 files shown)


classes/class_basematerial3d.rst
```diff
@@ -23,7 +23,7 @@ This provides a default material with a wide variety of rendering features and p
 Tutorials
 ---------
 
-- :doc:`../tutorials/3d/spatial_material`
+- :doc:`../tutorials/3d/standard_material_3d`
```
akien-mga, line 26: For the reference, this should be fixed in https://github.com/godotengine/godot directly, but for now I'm just hacking it. I don't want to sync the classref with the master branch for the time being as all the node renames will require tons of changes to the docs, which are best done once the refactoring discussed in #3390 has been done. Until then the `master` branch of the docs can stay close to `3.2` and thus make it easy to merge improvements and fixes to the stable branch.

PR #3797, opened 2020-07-15
Mention GDScript escape sequences direclty in GDScript basics

Calinou:
Many C escape sequences aren't valid in GDScript, so it's better to document them directly in a table.

See https://github.com/godotengine/godot/issues/38716.


akien-mga:
Thanks!

PR #3801, opened 2020-07-15
Typo macos compiling

gr4ph0s:
Not sure about the line 66 edit but since at line 48 the output is godot.osx.tools.universal I suppose we should also use it (at least in my case I needed to)

### DIFF
--- a/development/compiling/compiling_for_osx.rst
+++ b/development/compiling/compiling_for_osx.rst
@@ -63,7 +63,7 @@ editor binary built with ``target=release_debug``::
 
     cp -r misc/dist/osx_tools.app ./Godot.app
     mkdir -p Godot.app/Contents/MacOS
-    cp bin/godot.osx.tools.universal Godot.app/Contents/MacOS/Godot
+    cp bin/godot.osx.opt.tools.universal Godot.app/Contents/MacOS/Godot
     chmod +x Godot.app/Contents/MacOS/Godot
 
 If you are building the ``master`` branch, additionally copy the Vulkan library::


development/compiling/compiling_for_osx.rst
```diff
@@ -63,10 +63,10 @@ editor binary built with ``target=release_debug``::
 
     cp -r misc/dist/osx_tools.app ./Godot.app
     mkdir -p Godot.app/Contents/MacOS
-    cp bin/godot.osx.opt.tools.universal Godot.app/Contents/MacOS/Godot
+    cp bin/godot.osx.tools.universal Godot.app/Contents/MacOS/Godot
```
akien-mga, line 66: It's not a typo as if you use `target=release_debug` as mentioned on line 62, that's the name it would have.

But I agree it's confusing. It's a general issue with all compilation instructions which mention `scons p=<name>` as the command to build Godot, but that builds a debug binary (for developers), while most users probably want an optimized one (`scons p=<name> target=release_debug`).

CC @Calinou, maybe we should add `target=release_debug` to all build instructions for the editor?

Calinou: @akien-mga This is probably a good idea. It should be done for all platforms simultaneously, though (in a separate PR).

gr4ph0s: Since I'm new to Godot (my first PR), what's the preferred way?
I remove this change from my PR and create a new issue?

Thanks a lot, both for the quick answers and I agree in any case if the change is larger it should be done in a separate PR.

akien-mga: > I remove this change from my PR and create a new issue?

Yeah that's a good way. The other change in this PR is good to merge.

gr4ph0s:
Sorry I guess I shouldn't close it; I let you merge it when you want or close it if needed.

akien-mga:
Thanks!

PR #3803, opened 2020-07-16
Update gdscript_exports.rst

TheDuriel:
Testing on 3.2.2 reveals that Array sharing does not work as described. And I distinctly remember this not being a feature, but a bug. If the bug is still present in certain scenarios, it should not be advertised as a feature.

The only scenario in which an Array is _incorrectly_ shared. Is this:

`export var test = []`

and this:

`export(Array<, type>) var right: Array = []`

If an Array is exported as such. Empty and with no contents. And is then not assigned a value in the editor. Only then is the reference shared across instances. And only modifications of the array through code will reveal this. (This is a bug similar to how var a: Array, without = [] would create a broken "shared" array in  the past.)

Since the Docs policy is _not to document bugs_, I feel that a complete removal of any mention of shared arrays is in order.

I'm a bit angry right now because there is still so much misinformation spread about this.

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_exports.rst
+++ b/getting_started/scripting/gdscript/gdscript_exports.rst
@@ -165,11 +165,7 @@ If in doubt, use boolean variables instead.
 Exporting arrays
 ----------------
 
-Exporting arrays works, but with an important caveat: while regular
-arrays are created local to every class instance, exported arrays are *shared*
-between all instances. This means that editing them in one instance will
-cause them to change in all other instances. Exported arrays can have
-initializers, but they must be constant expressions.
+Exported arrays can have initializers, but they must be constant expressions.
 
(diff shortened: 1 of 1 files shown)


akien-mga:
Well this was described as a caveat, not a feature. But yes this bug is now fixed so the changes are correct.

> I'm a bit angry right now because there is still so much misinformation spread about this.

You're angry because `3.2` docs haven't been updated yet two weeks after the `3.2.2` to point our that a previous bug has now been fixed? This was not misinformation, but outdated correct information that applies to Godot 3.2.1 and earlier.

TheDuriel:
The wording on this has always been poor, even when it was _kind of_ how it worked. Making it a recurring topic of discussion. Anyways. Now its fixed. <3

PR #3814, opened 2020-07-20
Change platform linuxbsd to x11

e8johan:
As the scons platforms (at least since 3.2.2-stable) has renamed linuxbsd to x11,
all references to the linuxbsd platform have been updated to x11. The platform
is still called LinuxBSD when mentioned as a name in the text, but all scons and
binary references have been changed.

Notice that I did *not* update the tutorials/plugins/gdnative files.

### DIFF
--- a/community/contributing/updating_the_class_reference.rst
+++ b/community/contributing/updating_the_class_reference.rst
@@ -128,7 +128,7 @@ When classes are modified in the source code, the documentation template might b
 
 ::
 
-    ./bin/godot.linuxbsd.tools.64 --doctool .
+    ./bin/godot.x11.tools.64 --doctool .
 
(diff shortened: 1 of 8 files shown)


Calinou:
The platform is now called `linuxbsd` in the `master` branch, but is still `x11` in the `3.2` branch to preserve backwards compatibility. Therefore, the current name is correct.

I think it's better to just add a note that the platform was previously called `x11` in the `3.2` branch.

e8johan:
Thank you for the clarification. Would you like me to add a note to each of the pages I changed here? If so, I will update the MR to do that.

akien-mga:
You should use the `stable` or `3.2` version of the docs if you use Godot 3.2.x: https://docs.godotengine.org/en/latest/development/compiling/compiling_for_linuxbsd.html

I'd only add a note in the Compiling for X11 page, no need to put this detail everywhere when `linuxbsd` will soon become the new normal.

akien-mga:
Follow-up PR: #3815.

PR #3786, opened 2020-07-13
Add a page describing what is GDNative

Calinou:
This closes #1024.

PR #3816, opened 2020-07-21
Getting Started With GDNative C++ 

Gerarama:
This should help those who want to start working with C++ in godot.
note: this is just a draft, there is margin of improvement.


Calinou:
There's already a [GDNative C++ example page](https://docs.godotengine.org/en/latest/tutorials/plugins/gdnative/gdnative-cpp-example.html) (and I'm currently working on a [What is GDNative?](https://github.com/godotengine/godot-docs/pull/3786) page). I think it'd be better to improve that page rather than creating a new one with possibly duplicate information.

Gerarama:
The GDNative C++ example is missing some important informations.
like the order of the code, How to setup SConstruct  for beginners 
and a basic commenting on the c++ code.
I believe my one is a better replacemeant for it.

Jummit:
> I believe my one is a better replacemeant for it.

I like the old better, as yours has some grammar issues, doesn't explain the "why" aspect of things as well and uses non descriptive names like "script1" and "script2". I also don't understand why Visual Studio 2015 is a requirement.

Gerarama:
Yes,There is margin of improvement that can be done, as i said in the description of this commit.

English is not my main language there for i did my best to avoid any grammar issues, please feel free to edit or point them to me.

I used the same code as the orginial  GDNative C++ example page, but edit it to be more explained.
like using the words "script1" and "script2" to tell the reader that its a patten, and you can add "script3" and so long
also my code has comments on each line to explain why this line exist, unlike the GDNative C++ example page that explain what each line does outside the code.

you maybe notist that i missed the SConstruct file, because am waiting for approval before i commit to add my version of it

Calinou:
> you maybe notist that i missed the SConstruct file, because am waiting for approval before i commit to add my version of it

At this point, I think we should stop hosting code files in this repository. They're probably better suited to a template repository that can be forked at will by people.

NathanLovato:
Please don't add duplicate pages tackling the same topic or problem. If the original had issues, we should address them either by updating it if possible. If it's so bad it can't be improved, you may replace the original but I would recommend discussing the direction and teaching points first.

Note that the guide should also work for MacOS and Linux users. IIRC visual studio is only available on Windows.

NathanLovato:
To precise the previous comment: I wouldn't merge this PR as-is. I'm also in favor of improving existing docs pages over rewriting them from scratch, unless it's a necessity (for instance with GDScript 2.0, the syntax changes completely and the corresponding docs need to be re-written).

NathanLovato:
I just learned that a user being marked as ghost means they deleted their GitHub account, meaning nobody's maintaining this. I'll close the PR as a result.

PR #3821, opened 2020-07-22
Style external links differently to distinguish them from internal ones

Calinou:
## Preview

*The first link is external, whereas the second link at the bottom is internal.*

### Light theme

External and internal link comparison

### Dark theme

External and internal link comparison

<details>
<summary>Old version</summary>

Documentation sample
</details>

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -30,6 +30,7 @@
     --link-color-hover: #3091d1;
     --link-color-active: #105078;
     --link-color-visited: #9b59b6;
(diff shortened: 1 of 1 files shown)


akien-mga:
I'm not sure this unicode symbol is the best suited for this.

It's more common to see the "north-east pointing arrow in a box" symbol, which sadly hasn't been accepted into Unicode yet ([last proposal I found](https://www.unicode.org/L2/L2018/18303-external-link.pdf)), but is commonly used as SVG. I guess that should be an option for us too?

Alternatively, the Unicode LINK SYMBOL 🔗 might be fine? https://emojipedia.org/link/

Calinou:
@akien-mga I'll look into using a SVG icon for this. It should be feasible to embed it directly in the CSS using Base64.

Calinou:
The external link icon now uses SVG.

akien-mga:
Thanks!

PR #3828, opened 2020-07-23
Updated Python Code Style Guidelines

Anutrix:
Updated Python Code Style Guidelines to use Black now.

### DIFF
--- a/community/contributing/code_style_guidelines.rst
+++ b/community/contributing/code_style_guidelines.rst
@@ -271,7 +271,8 @@ You then have different possibilities to apply black to your changes:
 
 Manual usage
 ^^^^^^^^^^^^
-You can apply ``black`` manually one or more files with the following
+
+You can apply ``black`` manually to one or more files with the following
 command:
 
 ::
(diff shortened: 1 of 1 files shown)


pycbouh:
As mentioned in https://github.com/godotengine/godot/pull/40627 this should probably refer to pre-commit hooks instead.

Anutrix:
Should I add both pre-commit hooks and the manual way like it's done for clang-format?
If so, the text for that would be almost the same as that for clang-format's pre-commit hook section. Keeping this in mind, should we move pre-commit subsections to separate single section addressing both clang-format and black?

pycbouh:
I think it should be mentioned in some general contribution workflow guide, if not already. But it is still worthwhile mentioning the particular hook for python in the article about python code style. So what if it's similar to CLang one. That's fine. And yes, it doesn't hurt to mention both way to call it, but in my opinion the hooks are more important as they automate all kinds of checks.

community/contributing/code_style_guidelines.rst
```diff
@@ -248,7 +248,16 @@ Python
 Godot's SCons buildsystem is written in Python, and various scripts included
 in the source tree are also using Python.
 
-For those, we follow the `PEP-8 style guide <https://www.python.org/dev/peps/pep-0008/>`__,
-this is however not as strongly enforced as for the C++ code. If you are so
-inclined, you can check and format your Python changes using
-`autopep8 <https://pypi.org/project/autopep8/>`__.
+For those, we follow the `Black style guide <https://github.com/psf/black#the-black-code-style>`__.
+Blacken your Python changes using `Black <https://pypi.org/project/black/>`__.
+You can apply black-format manually one or more files with the following
```
akien-mga, line 253:
```suggestion
You can apply `black` manually one or more files with the following
```

Anutrix:
I forgot to put a newline at the end so CI complained. I fixed it and forced push but it doesn't seem to be detecting that change. It seems github itself is not detecting it as the new line is not showing at https://github.com/godotengine/godot-docs/pull/3828/files.

akien-mga:
Thanks!

PR #3775, opened 2020-07-11
Add a page listing all features supported by Godot

Calinou:
This is probably missing quite a few items, so feel free to chime in.

This closes #3727.

PR #3832, opened 2020-07-24
Add a page listing all features supported by Godot 3.2

Calinou:
Follow-up to #3775. I edited out all features that aren't available in 4.0.


akien-mga:
Thanks!

PR #3838, opened 2020-07-26
Android IAP Custom Permissions

ricardoalcantara:
Hi, I believe that describe how to add the Permission to use IAP is a good addition, I have used IAP before but when I used in my recently project I end up forgetting it, so after I uploaded my APK the console informed me that I need to upload an APK with that BILLING permission to even enable the IAP section, and I needed to google it, but I was no sure about that permission, so to be sure I had to search on my old project to confirm if the permission was really that one 'com.android.vending.BILLING'.


m4gr3d:
cc @timoschwarzer

timoschwarzer:
Hmm, I have to check this later. The billing library should automatically add the `BILLING` permission...

timoschwarzer:
@ricardoalcantara I wasn't able to reproduce the problem. I didn't and don't need to manually add the `BILLING` permission - it works out of the box when using the `GodotGooglePlayBilling` library.

Can anyone verify?

ricardoalcantara:
@timoschwarzer this is the second time I publish a game on Google Play Store, and the same thing happens, you can add the plugin, you can test is, you can publish it but when you try to open the Console to register new products Google inform that you cannot do it until you upload an APK with that Permissions.

image

timoschwarzer:
@ricardoalcantara Hmm, that's weird. I've disassembled my resulting APK using apktool and it has the permission in its `AndroidManifest.xml`. Can you send me your APK? You can also send it to me via email if you don't want to publicly post it (see my GH profile).

ricardoalcantara:
@timoschwarzer Did you upload it to Google store and it recognized the permission out of the box?
I don't have that APK, I added it in the export and published my game.

m4gr3d:
@ricardoalcantara based on your comments, it seems like you used an older version of the IAP plugin (back when it was still integrated within the Godot Engine).
As @timoschwarzer mentioned, the newer version of the plugin should add the permission automatically when the apk is generated.
If that's not happening for you, can you please file a bug in the https://github.com/godotengine/godot-google-play-billing repo.

Thanks!

ricardoalcantara:
@m4gr3d I used this one  https://github.com/godotengine/godot-google-play-billing with Godot 3.2.2-stable. I must have done something wrong then. I will check that and open a issue if it is the case. Thanks.

PR #3839, opened 2020-07-26
Add an exception for "ing" verbs where the subject is not clear

aaronfranke:
See [#38579](https://github.com/godotengine/godot/pull/38579) for a discussion about this.

In this PR, I managed to make the exception itself include an example of what it's describing :)


akien-mga:
Thanks!

PR #3844, opened 2020-07-29
Document the expected window size in Design a title screen

### LINKED ISSUE #3800: Design a Title Screen UI Tutorial - display size note
**Your Godot version:**
3.2.2
**Issue description:**
The Project Setting display size of **1366 x 768** should be noted in the tutorial, or a solution if the result doesn't match should be noted.

Following the tutorial verbatim with a smaller display size tends to result in confusion from the user when the result doesn't fit in the screen (ie:  The main MarginContainter being set to full rect being stuck with a 290 pixel right margin because the contents are too large horizontally).

low_default
On more than one occasion I've had to note that the tutorial uses a **1366 by 768** screen size because the tutorial results did not match the expected end result due to the user's Project Setting display size being smaller.  My own default project size is actually 1024 by 600 which will make this tutorial stretch objects outside of the screen.   This might be due to the fact I am using Godot on a laptop.

The main source of problem would be the Image in the Center Container on the right side.  Due to the 3 factors of the use case:

- CenterContainer doesn't respect size flags
- TextureRect cannot be set to expand because CenterContainer would shrink it to nothing regardless size flags
- The HBoxContainer pushes the CenterContainer to the side based on the images used from the left side and the MarginContainer's left margin.

Simplest solution would be to denote the Size for the tutorial and explain how to reach the setting.  

Optional fix would be to include a path in the tutorial that skips the HBoxContainer and CenterContainer entirely, opting to instead set the TextureRect to SHRINK_CENTER vertically and SHRINK_END horizontally.  Unfortunately I feel that this would fail in displaying the capability of Container usage though, so likely not the intended result here.
no_center

A third Option would be to include a replacement option to the CenterContainer or an optional smaller rect_min_size for the TextureRect so that expand would be able to be checked and the majority of the tutorial could be maintained.

https://docs.godotengine.org/en/stable/getting_started/step_by_step/ui_main_menu.html

Calinou:
This closes #3800.

### DIFF
--- a/getting_started/step_by_step/ui_main_menu.rst
+++ b/getting_started/step_by_step/ui_main_menu.rst
@@ -120,6 +120,16 @@ center the illustration with a ``CenterContainer``.
 Prepare the Main Menu scene
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+.. note::
+
+    This tutorial is based on a window size of 1366×768. To change the project's
+    base window size, open **Project > Project Settings** at the top of the
+    editor then change **Display > Window > Size > Width** to ``1366`` and
+    **Display > Window > Size > Height** to ``768``.
+
+    If you forget to change the window size, anchors and containers may not
+    behave as expected.
+
 Let's create the main menu. We'll build it in a single scene. To create
 an empty scene, click on **Scene > New Scene**.
 


akien-mga:
Thanks!

PR #3850, opened 2020-07-31
Corrected TextEdit reference in Optimizing a build for size

InsightGit:
Optimizing Builds For Size referenced a node type called TextEditor when talking about advanced GUI nodes when it most likely means the TextEdit node instead.


akien-mga:
Thanks!

PR #3853, opened 2020-08-01
Update class_tween.rst

nolikenet:
I suppose position is old naming for a property.
Might be confusing for new users, as example code does not work. Change to translation!

### DIFF
--- a/classes/class_tween.rst
+++ b/classes/class_tween.rst
@@ -25,12 +25,12 @@ Here is a brief usage example that causes a 2D node to move smoothly between two
 ::
 
     var tween = get_node("Tween")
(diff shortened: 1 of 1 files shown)


cbscribe:
Since the example is talking about Node2D, `position` is the correct property.
https://docs.godotengine.org/en/3.2/classes/class_node2d.html

What makes you say the code doesn't work?

Also, please note that class docs must be edited in the godot repo, not here, as they are auto-generated from the source xml.
https://docs.godotengine.org/en/3.2/community/contributing/updating_the_class_reference.html

nolikenet:
@cbscribe sorry, I did not notice that reference was to Node2D. All good then.

PR #3856, opened 2020-08-02
remove rant

Grinshpon:
Remove the `Why` section in the "Encrypting save games" page. It's a politically loaded cringey rant that is completely out of place in the documentation for a game engine.

PR #3857, opened 2020-08-02
Comment out rant

Grinshpon:
Comment out the `Why` section in the "Encrypting save games" page. It's an offensive and politically loaded rant that is completely out of place in the documentation for a game engine. It's basically saying "if you want to use this feature of our engine, you're scum."

Godot is supposed to be welcoming, and this is the opposite of that. I don't like it when the documentation insults the reader.

In order to comply with the "Please don't remove it" comment and make it more of an actual easter egg (which by definition is supposed to be hidden from plain sight), it is not removed but rather hidden as a comment (although I would still prefer if it were removed and not just hidden).

### DIFF
--- a/tutorials/io/encrypting_save_games.rst
+++ b/tutorials/io/encrypting_save_games.rst
@@ -3,34 +3,34 @@
 Encrypting save games
 =====================
 
-Why?
-----
+.. Why?
+.. ----
 
 .. This introduction is an Easter egg and is not intended to be taken seriously.
 .. Please don't remove it :)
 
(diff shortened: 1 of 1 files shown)


pycbouh:
This won't happen either way, so there is no reason to spam PRs. The reasoning is the same as in #3856. The position to keep it has been chosen, and is not going to change because you make multiple PRs. If you want to discuss it, have more points to say, you should use issues or proposals. Issues may get closed as well, though.

Calinou:
Indeed, there are no plans to remove this section from the user-visible documentation as the comment says.

PR #3858, opened 2020-08-02
Message label is not hidden after pressing start button

alagos:
I'm following the first game guide, but isn't hiding the Text label after pressing the Start button.
Before the change:
Screen Shot 2020-08-02 at 10 25 54 PM
After the change:
Screen Shot 2020-08-02 at 10 26 30 PM

### DIFF
--- a/getting_started/step_by_step/your_first_game.rst
+++ b/getting_started/step_by_step/your_first_game.rst
@@ -1139,6 +1139,7 @@ functions:
 
     func _on_StartButton_pressed():
         $StartButton.hide()
+        $Message.hide()
         emit_signal("start_game")
 
     func _on_MessageTimer_timeout():


getting_started/step_by_step/your_first_game.rst
```diff
@@ -1139,6 +1139,7 @@ functions:
 
     func _on_StartButton_pressed():
         $StartButton.hide()
+        $Message.hide()
         emit_signal("start_game")
 
     func _on_MessageTimer_timeout():
```
Calinou, line 1145: The message should be hidden by the message timer below.

alagos: Hmm, I can't check it right now, but it seems I forgot to connect the timeout signal.
Sorry for the troubles.

PR #3866, opened 2020-08-03
Improve the Singletons (AutoLoad) documentation

### LINKED ISSUE #3864: Misleading/ easy to misunderstand Singleton documentation
A singleton is defined in wikipedia as an object of which exists only one instance.
In Godot checking the singleton box in an Autoload object merely allows referencing that object with the "Object.parameter" pattern.

The documentation states it will:
"Act like a singleton, since GDScript does not support global variables by design"

Which it only does in referencing, **not** in that there is only one instance of it. 

So maybe change the check box to "Global Reference" or something more appropriate? Or make it clear in the docu that a godot singleton has little to do with a traditional singleton? 
It's a bit confusing at the moment by saying that it will "act like" a singleton.

Thank you for your time :)

Here is the discussion which prompted this request: https://www.reddit.com/r/godot/comments/i2alxp/exporting_variables_in_singleton_script/

And the link to the docu:
**https://docs.godotengine.org/en/stable/getting_started/step_by_step/singletons_autoload.html**

Calinou:
This closes #3864.

### DIFF
--- a/getting_started/step_by_step/singletons_autoload.rst
+++ b/getting_started/step_by_step/singletons_autoload.rst
@@ -20,33 +20,42 @@ own limitations:
    that require it, but frequently saving and loading data is cumbersome and
    may be slow.
 
-The `Singleton Pattern <https://en.wikipedia.org/wiki/Singleton_pattern>`_ is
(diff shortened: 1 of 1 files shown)


akien-mga:
Thanks!

Issue #2735, opened 2019-08-31
Split up GDScript basics into many smaller sections.


golddotasksquestions:
https://docs.godotengine.org/en/3.1/getting_started/scripting/gdscript/gdscript_basics.html#doc-gdscript
This page is simply overwhelming. I never read it in full, only look up bits and pieces here or there when I need.
Splitting off exports (https://github.com/godotengine/godot-docs/pull/2718) is into one (or more) pages(s) is definitely a good idea, but can we have this for more of it's topics?

While it's called "Basics" of GDScript it actually seems like quite a comprehensive collection of it's features that go way beyond what I would consider basic. It's incredible valuable, overall very well explained information, but if was structured into smaller junks with pages (and more detailed list of the topics to left for each page), I would not get intimidated that easy, plus quickly get an overview about what information I can find here.

asheraryam:
I don't think we _need_ to split up the page, but I think it's a general problem that there's no Table of Contents for large pages like this.

golddotasksquestions:
If there was a only table of content, it would stretch over multiple screens and just make this page larger.

Another problem here is that is you online search for something, like how to instance a scene for example the search result will just be this page and you will have to perform another search within that page to get what you are looking for. Not the best solution either.

Meriipu:
I really liked having it all on one page when I first read it, and I still enjoy it those times where I return to it to look something up.
it is easy for me to get lost when there are tons of nested pages or many pages in a list each with barely any content. 
It leads to a lot of backwards/forwards or a lot of tabs.
It also makes it very hard to ctrl+f to search.

Calinou:
@Meriipu There's a middle ground to find for sure, but I think having complex GDScript concepts on separate pages makes more sense still.

Haledire:
Didn't notice this issue before - so I'll bring this up here
yield is going to be removed in favor of await
yield is a function - await is a keyword
Is the method for coroutines going to be removed from documentation available offline and available strictly from the online docs? it's currently visible in @gdscript because it's a method and not a keyword.

I do feel there's some merit to handling something for offline visibility of the documentation handled in the basics page.

Calinou:
@Haledire Keep in mind you can download offline HTML builds of the documentation from [this page](hugo.pro/projects/godot-builds/).

skyace65:
The GDScript basics page got renamed to GDScript reference at some point. Given that it's a reference page now and not an introduction does it still need to be split up into multiple pages?

[...]

PR #3623, opened 2020-05-31
Update docs for GDScript 2.0

vnen:
GDScript syntax will change a bit after https://github.com/godotengine/godot/pull/39093, so this is an update of the basics and export pages to fit the new way.

I removed the section for exporting arrays because it's still not supported, but I'll add it back once typing is re-added and there is a syntax for typed arrays.

I also removed the groups/categories sections because I plan to add those as annotations.

PR #3870, opened 2020-08-04
Split gdscript intro into two pages

balloonpopper:
Split of the gdscript basics page into two as discussed with @Calinou 

*Bugsquad edit: See discussion in https://github.com/godotengine/godot-docs/issues/2735.*

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_basics.rst
+++ b/getting_started/scripting/gdscript/gdscript_basics.rst
@@ -325,7 +325,7 @@ of characters for example). While you can create your own data types to
 describe your data, Godot has a set of built-in types as described 
 below for you to useful.
 
-Advanced note : Built-in types are stack-allocated. They are passed as values. 
+*Advanced note:* Built-in types are stack-allocated. They are passed as values. 
 This means a copy is created on each assignment or when passing them as 
 arguments to functions. The only exceptions are ``Arrays`` and 
 ``Dictionaries``, which are passed by reference so they are shared. (Pooled 
--- a/getting_started/scripting/gdscript/gdscript_basics.rst
+++ b/getting_started/scripting/gdscript/gdscript_basics.rst
@@ -114,7 +114,7 @@ read this tutorial: :ref:`doc_gdscript_more_efficiently`.
 Language
 --------
 
-In the following, an overview is given to GDScript. Details, such as which
+Below is an overview of GDScript. Details such as which
 methods are available to arrays or other objects, should be looked up in
 the linked class descriptions.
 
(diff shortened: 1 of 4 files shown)


getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -560,9 +571,13 @@ assign to it::
 Data
 ----
 
+.. _doc_gdscript_variables:
+
 Variables
 ~~~~~~~~~
 
+A variable is some data that is stored with an associated label. For example
+you could store a variable "fruit" which contained the contents "apple".
```
Calinou, line 580:
```suggestion
A variable is some data that is stored with an associated label. For example
you could store a variable called ``fruit`` which contains the string ``"apple"``.
```

Calinou (approved):
Looks good to me. The split seems reasonable to me. It's pretty much what I had in mind when I was talking about splitting the page on IRC :slightly_smiling_face: 

I'll let others review it as well.

balloonpopper:
It still has a higher level of assumed knowledge than I'd like it to have, but I thought it wiser to make iterative changes than rewrite the whole thing.

balloonpopper:
@Calinou I made a small change since your review. I realised I'd left a comment in there that the keyword "pass" needed some documentation but hadn't replaced it with content before I made the commit. I've since added it, and fleshed out the text around what a function does while I was at it.

getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -327,7 +327,7 @@ below for you to useful.
 
 Advanced note : Built-in types are stack-allocated. They are passed as values.
```
Calinou, line 328:
```suggestion
*Advanced note:* Built-in types are stack-allocated. They are passed as values. 
```

NathanLovato:
I would split them way more. In particular, I don't like splits into part 1 / part 2 because the titles don't help you find information faster than having it all in one page. 

Instead, I would recommend to create a new sub-section covering GDScript basics where each page covers a specific and clear topic. I'd have to go through all pages to come up with a suggestion for the exact split, but for instance variables, functions, classes (including properties and inheritance), signals, coroutines could already be 5 pages with explicit titles.

And from the original discussion, regarding the user who talked about being able to search one big page: I think for them, a GDScript cheatsheet would likely be better. Many users aren't even aware they can jump around web pages with text search or would jump around when learning the language's fundamentals.

NathanLovato:
Note that we should probably merge #3623 and other PRs for GDScript docs first, even if they're going to change later, to avoid causing many merge conflicts.

balloonpopper:
@NathanLovato I think it's critical for the documentation that searching explicitly for a command brings its results back as the first search.
e.g. if I search for "if", currently the link to "if/then/else" is the 10th result returned whereas it should be number 1. Searching for "var" doesn't even return the basics page at all.  
Having the search function broken, and a lack of examples for each command are the two biggest drawbacks of the documentation I believe. The structure of the documentation is of course important, but I would place it a very distant 3rd compared to a useful search and examples.

It would appear that rather than the suggested GDScript cheat sheet that search ranking could be corrected? https://docs.readthedocs.io/en/stable/config-file/v2.html#search-ranking

NathanLovato:
The search feature isn't relevant to this issue and PR, which is about splitting a page that's too long to process well into multiple.

The docs search is a different task and a job that involves coding on the website's back-end. The person who'd likely know most about that is @Calinou.

Note that page titles seem to have a strong weight in search results and that having pages named GDScript variables or Functions in GDScript would likely be among the top results when searching respectively variable(s) or function(s). You could test that locally by making pages with only a title, building, and doing some test searches.

Also, you won't get good result in about any search engine by searching for `if` or `var`. It's too short for a keyword, too common for the engine to find something relevant.

balloonpopper:
Closing this in light of the work that @NathanLovato is doing on the documentation.

NathanLovato:
Ah thanks, and sorry for that!

PR #3879, opened 2020-08-07
Mention the OpenGL requirement when using the `--no-window` CLI argument

Calinou:
This was requested here: https://godotforums.org/discussion/comment/40833/#Comment_40833


akien-mga:
Thanks!

PR #3881, opened 2020-08-08
Add a preview image of color constants

theowiik:
Add a preview image of all the color constants after the list of constants: https://docs.godotengine.org/en/stable/classes/class_color.html#constants

**Important**: I'm gonna be honest and say that I have not previewed how this looks as I had trouble installing the Sphinx stuff.

A created a simple app that generates this image: https://github.com/theowiik/godolors
And this is the image: 
colors

I can also make the image have less columns if that would make it easier to read.

*Bugsquad edit: This closes https://github.com/godotengine/godot-docs/issues/4060.*


pycbouh:
I don't know if it's the font, the letter spacing, or the black outline, but some of those are very hard to read.

theowiik:
@pycbouh yes I believe I was a bit to eager to get this PR trough. I will continue working on it, and get sphinx installed so I can make sure that it is readable and works.

I found out about the problem with the text being hard to read later, it is when you are viewing the image in "another scale", but I believe I can fix this easily by just making the font larger.

I will continue working on it and mark it as ready for review when I believe I have solved the issues so far

volzhs:
is it possible to sort colors to show similar colors together?
image

theowiik:
@volzhs I am looking into that 👍

theowiik:
@pycbouh I read into [Contributing to the class reference](https://docs.godotengine.org/en/latest/community/contributing/updating_the_class_reference.html), and there does not seems to be any support for having images in the class documention, which makes sense if you are going to be able to view the documentation in the editor.

So I am not sure if this is fitting anywhere else in the docs

pycbouh:
Well, bbcodes are supported. But there doesn't seem to be any record in class reference that uses images. It can be linked as an external resource, I guess.

Calinou:
> Well, bbcodes are supported. But there doesn't seem to be any record in class reference that uses images. It can be linked as an external resource, I guess.

There's an example in the Tween class: https://github.com/godotengine/godot/blob/a32b427f99d39bc6358e8477dc9189bf0f7bdc2e/doc/classes/Tween.xml#L19

theowiik:
@volzhs 

Here is a new version sorted by hue, which is a cheap and easy way to get a approximate order of the colors. The image is not commited yet

colors

Calinou:
I manually merged a losslessly optimized version of this image. Thanks for your contribution!

PR #3888, opened 2020-08-10
Update jitter_stutter.rst

Tooniis:
The human eye can see more than 60F/s. It's monitors that are usually 60 Hz.

### DIFF
--- a/tutorials/misc/jitter_stutter.rst
+++ b/tutorials/misc/jitter_stutter.rst
@@ -32,7 +32,7 @@ Jitter
 There can be many causes of jitter, the most typical one happens when the game *physics frequency* (usually 60 Hz) runs
 at a different resolution than the monitor refresh rate. Check whether your monitor refresh rate is different from 60 Hz.
 
-This is generally not a problem, given that refresh rates higher than 60 Hz are barely visible to the human eye, and
+This is generally not a problem, given that most monitors are 60 Hz, and
 starting with Godot 3.1, a frame timer was introduced that tries to synchronize with refresh as well as possible.
 
 Sometimes only some objects appear to jitter (character or background). This happens when they are processed in different


clayjohn:
These are two very different statements. The original author is saying that the issue is small because the human visual system isn't very capable at those frequencies. 

The author is not saying that the issue doesn't matter because most consumers buy 60Hz monitors. 

Im not sure it's true either. Faster frequency monitors are becoming increasingly common especially among people who play games.

Tooniis:
I don't have much of an opinion about it, I just did it after [this conversation in discord](https://discordapp.com/channels/212250894228652034/212250894228652034/742227185020043405)

clayjohn:
Hmmm. Thanks for linking that. The user is unclear about _why_ the statement is problematic. 

My understanding is that the brain cannot process more than 60hz, but eyes are sensitive to flicker at faster rates than 60hz, so a faster monitor is easier on the eyes and helps you focus on the screen better (hence being better for gaming).

If I'm right, then the original statement is true and is fine. If I'm wrong, the statement is still true because it says humans can _barely_ see above 60hz, it doesn't say that humans cannot detect light flickering at a rate above 60hz. 

Anyway, that was a lot of words to say *shrug* I don't know what to do about this either. That user seems really upset about the line. But I don't think we should change the line just be cause one person made a fuss without explaining themself given that the line changes the substance of the original message.

Calinou:
> My understanding is that the brain cannot process more than 60hz, but eyes are sensitive to flicker at faster rates than 60hz, so a faster monitor is easier on the eyes and helps you focus on the screen better (hence being better for gaming).

Brains *can* process a lot more than 60 Hz. It might be hard to grasp if you don't have access to a high refresh rate monitor, but believe me :slightly_smiling_face:

akien-mga:
*shrug* too, but let's merge, it doesn't hurt.

akien-mga:
Thanks!

PR #3891, opened 2020-08-10
Mention Git for Windows CRLF/LF line ending issues

marechal-p:
Got bitten by this `core.autocrlf` when working with Godot, I think it is good information to have in the documentation.

### DIFF
--- a/getting_started/workflow/project_setup/version_control_systems.rst
+++ b/getting_started/workflow/project_setup/version_control_systems.rst
@@ -43,4 +43,4 @@ Most Git for Windows clients are configured with the ``core.autocrlf`` set to ``
 This can lead to files unnecessarily being marked as modified by Git due to their line endings being converted automatically.
 It is better to set this option as::
 
-    git config core.autocrlf input
+    git config --global core.autocrlf input
--- a/getting_started/workflow/project_setup/version_control_systems.rst
+++ b/getting_started/workflow/project_setup/version_control_systems.rst
@@ -39,7 +39,7 @@ There are some folders Godot creates you should have your VCS ignore:
 Working with Git on Windows
 ---------------------------
 
-Most Git for Windows clients are configured with the `core.autocrlf` set to `true`.
+Most Git for Windows clients are configured with the ``core.autocrlf`` set to ``true``.
 This can lead to files unnecessarily being marked as modified by Git due to their line endings being converted automatically.
 It is better to set this option as::
 
--- a/getting_started/workflow/project_setup/version_control_systems.rst
+++ b/getting_started/workflow/project_setup/version_control_systems.rst
@@ -40,8 +40,8 @@ Working with Git on Windows
 ---------------------------
 
 Most Git for Windows clients are configured with the `core.autocrlf` set to `true`.
-This can lead to weird git status and warnings in the diff output. It is better to set this
-option as:
-::
+This can lead to files unnecessarily being marked as modified by Git due to their line endings being converted automatically.
+It is better to set this option as::
+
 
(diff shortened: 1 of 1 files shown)


getting_started/workflow/project_setup/version_control_systems.rst
```diff
@@ -35,3 +35,13 @@ There are some folders Godot creates you should have your VCS ignore:
...
 - ``.mono/``: This folder stores automatically-generated Mono files. It only exists
   in projects that use the Mono version of Godot.
+
+Working with Git on Windows
+---------------------------
+
+Most Git for Windows clients are configured with the `core.autocrlf` set to `true`.
+This can lead to weird git status and warnings in the diff output. It is better to set this
+option as:
+::
```
Calinou, line 45: It's better to explain practical consequences rather than being vague :slightly_smiling_face: 

```suggestion
Most Git for Windows clients are configured with the `core.autocrlf` set to `true`.
This can lead to files being marked as modified by Git due to their line endings being converted automatically.
It is better to set this option as::

```

marechal-p: Indeed! Thanks

getting_started/workflow/project_setup/version_control_systems.rst
```diff
@@ -35,3 +35,13 @@ There are some folders Godot creates you should have your VCS ignore:
   project, including sensitive information such as Android keystore credentials.
 - ``.mono/``: This folder stores automatically-generated Mono files. It only exists
   in projects that use the Mono version of Godot.
+
+Working with Git on Windows
+---------------------------
+
+Most Git for Windows clients are configured with the `core.autocrlf` set to `true`.
```
Calinou, line 42: Also, the syntax for inline code in reStructuredText uses two backticks instead of one.

```suggestion
Most Git for Windows clients are configured with the ``core.autocrlf`` set to ``true``.
```

getting_started/workflow/project_setup/version_control_systems.rst
```diff
@@ -35,3 +35,13 @@ There are some folders Godot creates you should have your VCS ignore:
...
+This can lead to files unnecessarily being marked as modified by Git due to their line endings being converted automatically.
+It is better to set this option as::
+
+
+  git config core.autocrlf input
```
Calinou, line 47:
```suggestion
    git config core.autocrlf input
```

marechal-p:
Should I squash my commits?

getting_started/workflow/project_setup/version_control_systems.rst
```diff
@@ -35,3 +35,12 @@ There are some folders Godot creates you should have your VCS ignore:
...
   in projects that use the Mono version of Godot.
+
+Working with Git on Windows
+---------------------------
+
+Most Git for Windows clients are configured with the ``core.autocrlf`` set to ``true``.
+This can lead to files unnecessarily being marked as modified by Git due to their line endings being converted automatically.
+It is better to set this option as::
+
+    git config core.autocrlf input
```
Calinou, line 46: Also, it's probably better to use `--global` to ensure it applies to all repositories:

```suggestion
    git config --global core.autocrlf input
```

Otherwise, it would be equivalent to `--local` which makes it only apply to the repository the CWD is currently in.

marechal-p: Hm not so sure about this, I intentionally left `--global` out. My reasoning is that most clients are configured with `true` and I never had any issue but with Godot. Maybe it's fine and I'm just a worrywart.

PR #3894, opened 2020-08-12
Suggest `is_null` over `!is_valid` for examples in error macros

Xrayez:
I see inconsistent usage of those methods for error checking throughout the engine, using `is_null` is much better for interpreting the error condition, both for engine programmers and the way errors are reported in Godot projects, so this should give a good example for new contributors and existing devs. 🙂

### DIFF
--- a/development/cpp/common_engine_methods_and_macros.rst
+++ b/development/cpp/common_engine_methods_and_macros.rst
@@ -200,7 +200,7 @@ Godot features many error macros to make error reporting more convenient.
 
     // Conditionally prints an error message and returns from the function.
     // Use this in methods which don't return a value.
-    ERR_FAIL_COND_MSG(!mesh.is_valid(), vformat("Couldn't load mesh at: %s", path));
+    ERR_FAIL_COND_MSG(mesh.is_null(), vformat("Couldn't load mesh at: %s", path));
 
     // Conditionally prints an error message and returns `0` from the function.
     // Use this in methods which must return a value.


akien-mga:
I don't know, for me "not being valid" is clearer than being null, even if they're equivalent. I'd argue that `is_null()` is not a good description of what is an invalid `Ref`, as it can be confused with a valid but zero-initialized object.

Like `Vector3()` is a null vector, but it's valid, while `Ref<Node> node` is a null reference which is invalid, as opposed to `Ref<Node> node; node.instance()` which is a valid reference to a default initialized Node. `Vector3` is not a type you'd find in `Ref<>`, but still it makes it slightly harder to understand IMO.

One can of course agree that the negation in `!ref.is_null()` also adds cognitive overhead, so I'm not saying a strong no to this proposal :)

akien-mga:
Some numbers for reference (`master` branch):
```
$ rg ERR_FAIL.*is_null | wc -l
298
$ rg ERR_FAIL.*is_valid | wc -l
194
```

Xrayez:
There are classes like `Regex` and `IP_Address` in Godot that have similar `!is_valid()` checks while not being `Ref`, so yeah it doesn't make sense to have `is_null()` for those type of classes (the number of those occurrences are much lower though).

The "problem" comes from Godot using error macros which are the opposite of assertions, so it makes sense to use `is_null()` if you live in a world of "catching error conditions" (C++ in Godot) vs "ensuring things go right" (GDScript). When I see the same error condition in the context of GDScript debugger/editor, it makes it really confusing for me...

So, I'd prefer to have `is_null` logic for `Ref`, so you're instantly aware that those errors pertain to `nullptr` problems in code, and leave `is_valid()` for other stuff (e.g., data validation).

> Like `Vector3()` is a null vector, but it's valid

By the way, `Variant` has the following methods currently:

```c++
	bool is_zero() const;
	bool is_one() const;
	bool is_null() const;
```

But there's no `is_valid()`, strangely. 😮

So, what I'm suggesting exactly may not be a solution, but yeah I'm trying to identify used conventions (if any), or create one.

Perhaps it may be worth introducing `is_not_valid()` method instead? In GDScript we have `is_instance_valid()` which can be passed to `assert()` currently. GDScript has no equivalent of something like `error_cond()`.

Xrayez:
I think this needs a proposal, but it's unlikely to receive support, so closing.

PR #3895, opened 2020-08-12
Update instructions to make subresources unique in inspector

GoOtterGo:
As of 3.2.2 the icon for Object Properties is no longer a down arrow, it is now a tools emoji. It's also not clear from the sentence the tutorial is expecting the user to look for the Object Properties button, just 'the down arrow', so I added clarity there as well.


akien-mga:
Thanks!

PR #3897, opened 2020-08-13
Fix nested list formatting in List of features

Calinou:
See https://github.com/godotengine/godot-docs/pull/3775#issuecomment-673622332.


akien-mga:
Thanks!

PR #3903, opened 2020-08-17
C# setup for linux for nuget packages see issue #41306 in godotengine/godot

Duehok:
Adds what I understood from #41306: https://github.com/godotengine/godot/issues/41306
to use nuget packages, C# on linux needs mono 6.12+ (preview version) and nuget.exe installation

I am guessing that it is also true on MacOS, but I have not tested on this system!

Applies to 3.2
I remember having read that there are changes coming in 4.0 about how the C# assemblies are handled, so these changes might not apply to 4.0 but I have not tested.

Also, link to the impacted page:
https://docs.godotengine.org/en/stable/getting_started/scripting/c_sharp/c_sharp_basics.html

### DIFF
--- a/getting_started/scripting/c_sharp/c_sharp_basics.rst
+++ b/getting_started/scripting/c_sharp/c_sharp_basics.rst
@@ -66,18 +66,24 @@ Make sure to set the following preferences:
    - Set **MSBuild version** to either **Bundled with Rider** or **.NET Core**.
    - Install the **Godot support** plugin.
 
+.. _set-up-csharp-on-linux:
+
 macOS and Linux
 ~~~~~~~~~~~~~~~
 
-Download and install the latest version of the
-`Mono SDK <http://www.mono-project.com/download/>`_. As of Godot 3.1 beta 3,
-the version number doesn't matter since Godot bundles its own Mono 5.18
-installation. We only need the Mono installation for NuGet and MSBuild
-which are required to use C# in Godot.
+Download and install the version 6.12 or above of the
+`Mono SDK <https://www.mono-project.com/download/preview/>`_. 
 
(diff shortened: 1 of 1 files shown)


Duehok:
Hopefully these changes are made obsolete by 
https://github.com/godotengine/godot/pull/41408

I'd like to keep this PR open until there is at least a 3.2.3 RC with the changes included. Then the whole C# setup page will need a good cleanup.

akien-mga:
Can you check with https://godotengine.org/article/release-candidate-godot-3-2-3-rc-6 and adapt the docs accordingly?

BTW, release-specific or time-based references like "as of Godot 3.1 beta 3" (incredibly outdated by now) should really be avoided.

akien-mga:
BTW, pull requests should be made against the `master` branch as required in the PR template. This change is valid for both Godot 4.0 and Godot 3.2.x, so it will be cherry-picked manually for the stable branch once merged in `master`.

neikeq:
You don't need Mono 6.12+ unless there's something I'm missing. You don't need nuget.exe either as we use `msbuild /restore`.
IMO this page should make it clear that you only need MSBuild which can be obtained from different sources. On Linux/macOS those would be:

- Mono (on Linux one must make sure it includes MSBuild, as it may come in a different package)
- .NET Core SDK

On Windows there's also Visual Studio (or its Build Tools subset) and Rider's bundled MSBuild.

NathanLovato:
What's the status on this PR? Is it still relevant or should we close it?

aaronfranke:
This article still needs updating, but most of the changes in this PR aren't good. The repository this PR is based on has been deleted so this PR is invalid, closing.

PR #3905, opened 2020-08-18
Fix first snippet in Change scenes manually

BiRD4:
In the first snippet, a PackedScene is loaded but never instanced, which would throw this error:
`Invalid type in function 'add_child' in base 'Node'. The Object-derived class of argument 1 (PackedScene) is not a subclass of the expected argument class.`


BiRD4:
Thank you! :)

PR #3911, opened 2020-08-19
Viewport Usage enumeration

Yetizone:
Mention that 3D effects are not available when using USAGE_2D

Fixes https://github.com/godotengine/godot/issues/41151

Targeting stable since the Usage enumeration seems to no longer be available in the latest version of the documentation.

### DIFF
--- a/classes/class_viewport.rst
+++ b/classes/class_viewport.rst
@@ -337,7 +337,7 @@ enum **MSAA**:
 
 enum **Usage**:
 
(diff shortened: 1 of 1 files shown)


Calinou:
> Targeting stable since the Usage enumeration seems to no longer be available in the latest version of the documentation.

Indeed, the `master` branch's Vulkan renderer automatically determines the framebuffer allocation to use, making the setting redundant.

Calinou:

PR #3912, opened 2020-08-19
Update your_first_game.rst

ketan-m-patel:
Missing call to start animation in the enemy script.


cbscribe:
You are told to set the "Playing" property in the Inspector. This does not need to be done in code.

ketan-m-patel:
> You are told to set the "Playing" property in the Inspector. This does not need to be done in code.

Ah, I missed that. Thanks.

PR #3916, opened 2020-08-21
Add a page on the different animation track types

skyace65:
I've created a page in the animation section going over the different animation track types, what they do, and how to use them. I've also taken out the call method tracks section from Introduction to the 2D Animation Features, and deleted the images from it. It has been replaced with a section going over call method tracks on the new page.

All images have gone through png compression.

Also while I'm doing this I wanted to know what everyone's opinion is on changing the "into to 2D animation" page into just "intro to animation." The stuff in there still applies to 3D animation, and I think it would be a better place for the 3D transform track information since it's so short.


NathanLovato:
Closing because I rebased and edited in #4242. The first commit in that new PR is yours, from this PR, but rebased on master.

PR #3920, opened 2020-08-21
Remove Unity to Godot page

### LINKED ISSUE #2368: Unity to Godot page is out of date
Currently the Unity to Godot page talks about Unity workflow from before nested prefabs were supported.
getting_started/editor/unity_to_godot.rst

Nested prefabs make a much more Godot-like scene organisation possible in Unity (afaik). I believe this makes much of the document irrelevant, and in need of re-writing from a perspective of highlighting any remaining differences in scene organisation workflow between contemporary versions of unity and godot.

(I don't have experience using the nested prefab version of Unity)

skyace65:
I don't think this is a good page for the documentation and should be removed for two main reasons.

It doesn't make it any easier for people who know Unity to switch over to Godot. It's going over concepts already covered on other pages. It's not giving anyone a head start with learning because they already know Unity.

I would assume most of us here primarily use Godot, which is why are contributing to its documentation. A page like this, going over the workflow of both engines in detail, means that we'd have to keep seeing what Unity is doing to keep it up to date. And we already weren't doing this, there's a note at the top of the page saying it's out of date. Even if we were doing that I think we can agree there's better uses of time for improving the docs.

The supported platform information might be useful. But it's probably better in a different section of the docs than the editor manual.

Closes #2368


Xrayez:
>  A page like this, going over the workflow of both engines in detail, means that we'd have to keep seeing what Unity is doing to keep it up to date. 

Yeah, but the same can be said regarding various [third-party IDEs](https://docs.godotengine.org/en/latest/development/cpp/configuring_an_ide/index.html) written in the Godot's documentation, but nobody is arguing that those instructions must be kept up-to-date as a rationale for removing them (but maybe I haven't stumbled upon this kind of discussion before).

I understand that we're not interested in unintentional "libeling", but I personally think that this kind of information is as important as other comparisons in Godot's documentation. Perhaps this shouldn't be too detailed as you said, but one could just describe the *major* differences between those engines on ideological level, instead of just removing this altogether, in subsequent PRs right after this one as an alternative (yes, I understand that's more work to do, but who likes to work? 😉). But perhaps I missed something? Then again I'm sorry.

I mean, even [the lead developer contrasts Godot with Unity and Unreal Engine](https://www.reddit.com/r/godot/comments/8mhzfo/tynansylvester_of_rimworld_fame_is_evaluating/dzrdyka?utm_source=share&utm_medium=web2x).

That said, @skyace65 I'd be careful removing information (please always link to alternatives if you do decide to remove something), talking about cases such as https://github.com/godotengine/godot/pull/41683#discussion_r482146230. 😛

clayjohn:
@Xrayez it's still up in the stable docs! :)
https://docs.godotengine.org/en/stable/getting_started/editor/unity_to_godot.html

Pro-tip, when sending people docs links, always use stable. Master is unstable and probably won't work with the current released version. Also, we almost never remove docs from past releases.

NathanLovato:
Writing a migration guide may be interesting to some users, sure, but we would need contributors to do the job well and maintain the page, trying new versions of Unity alongside Godot over the years. That was the problem with this page, it was unmaintained and outdated, with nobody to rewrite it.

Another issue is that comparing engines is highly subjective and dependent on your experience. There's not necessarily a good parallel between features; e.g., presenting Godot's nodes as an alternative to Unity's component, something I've read before online, will get the users to use the nodes in unintended ways. You're more supposed to use aggregation and object-oriented programming principles if you want to write code that's easy to maintain in Godot.

As such, I feel that articles comparing engines or migration guides from other engines would better fit a community wiki or external resources, where people can share their experiences and opinion.

Finally, one of the main problems isn't that a migration guide would be work. Rather, it's that the docs are currently both outdated and incomplete. There are hundreds of hours of higher priority work to do at the moment.

Anyway, we're moving towards writing a reference manual for Godot. We'll be removing outdated or low-quality content from the Godot 4.0 docs, along with writing new content; as @clayjohn mentioned though, that does not affect the 3.2 docs or older versions.

Xrayez:
> Pro-tip, when sending people docs links, always use stable. Master is unstable and probably won't work with the current released version. Also, we almost never remove docs from past releases.

Well I'd expect to use the latest version in this case, given the fact that Unity released a new version which may be different to Godot 3.x era. 😛

NathanLovato:
Latest stands for the master branch of the docs, it's the work-in-progress docs that go along with the work-in-progress master branch of the engine. So it's always outdated.

akien-mga:
> Yeah, but the same can be said regarding various third-party IDEs written in the Godot's documentation, but nobody is arguing that those instructions must be kept up-to-date as a rationale for removing them (but maybe I haven't stumbled upon this kind of discussion before).

Is this really an argument made in good faith? Do you consider that a page comparing Unity to Godot and contributor-focused documentation on how to configure popular IDEs to modify Godot code are similar? I have trouble taking this seriously.

If this is just arguing for the sake of it, I think it's a waste of our time.

Let's try to work together towards improving Godot and its documentation, and not try to find an angle to question or criticize all actions that are actually taken.

aliakseikalosha:
I found it very useful for me, found it in a doc, read it, and then tried to improve it not knowing that it already was absolute.

GersonFeDutra:
I think it would be beneficial to newcomers from other technologies to have a reference material to make the move. How biased the transition could be, or how it would enforce bad habits, doesn't matter much because it is already a reality of switching sides - this is a matter of the content quality of these materials.

If these could be made elsewhere can be discussed. But if is made there, would be valid to have such materials linked into the docs. Having a high level not up-to-date material is still better than nothing, while are still none out there.

PR #3921, opened 2020-08-22
Mention licensing requirements in Best practices for engine contributors

Calinou:
This has been an implicit requirement for a long time now, but it doesn't hurt to mention it.

We've technically accepted LGPL-with-static-linking-exception in the past for libwebsockets, but I doubt we'll want to do this again in the future.


NathanLovato:
Great, thanks!

PR #3928, opened 2020-08-24
Fix TextureButton texture number in Design interfaces with the Control nodes

### LINKED ISSUE #3927: Adjective "sixth" does not agree with the previous count of five textures.
**Your Godot version:**
Version 3.2 Docs

**Issue description:**
The documentation says that "TextureButton is like TextureRect, except it has *5* texture slots" and then goes on to list the various textures and what they do. The final one of lists is the clock mask, stating that "The *sixth* image slot, the Click Mask..." The *sixth* in this statement should be changed to *fifth*, as this adjective does not agree with the actual count. There are only five textures to a button, so this is the wrong adjective for this phrase. 

https://docs.godotengine.org/en/stable/getting_started/step_by_step/ui_introduction_to_the_ui_system.html#texturebutton

skyace65:
The TextureButton node has 6 texture slots, not 5. Closes #3927

### DIFF
--- a/getting_started/step_by_step/ui_introduction_to_the_ui_system.rst
+++ b/getting_started/step_by_step/ui_introduction_to_the_ui_system.rst
@@ -97,7 +97,7 @@ the ``Modulate`` property and use the color picker.
 TextureButton
 ~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


Calinou (approved):
Thanks!

PR #3929, opened 2020-08-24
Reorder exporting articles to be more logically ordered

skyace65:
Reorders the exporting articles to make more sense. Changing the icon for windows is now under exporting for PC. Custom builds for Android is now under exporting for Android. and exporting for UWP is under changing the icon for windows so the list now goes desktop, phones, other. I know this is a inconsequential change, I just think it looks better.

I'm not sure if One Click Deploy should be moved or not, since the HTML5 export has a similar feature.


Calinou (approved):
Thanks!

PR #3931, opened 2020-08-25
Add warnings on embedded pck files

### LINKED ISSUE #3093: Mention caveats about exporting for Windows with embedded PCK and/or UPX
*Bugsquad edit:* Moved to a docs issue. See https://github.com/godotengine/godot-docs/issues/3093#issuecomment-577064497

----

**Godot version:**
Godot_v3.2-beta5_x11.64

**OS/device including version:**
Kubuntu/19.10

**Issue description:**
As recommended by reduz in godotengine/godot#27140 I used UPX to compress the executable (and the difference is so dramatic (40mb to 14mb: 65% reduction) it should really be a standard feature in my opinion) but it does not work with the Embed Pck option.

Exporting with Embed Pck for Linux/X11 gives an error but works anyway (the same as godotengine/godot#32513). But after using UPX running the executable only gives the dialog "Error: Could not load game data at path '.'. Is the .pck file missing?"

Doing the same for Windows Desktop has no error on export (and runs fine on my machine using Wine (not affected by godotengine/godot#33466), but when running UPX I get "upx: Articore.exe: CantPackException: superfluous data between sections (try --force)". Using --force leads to the same result as for the Linux/X11 version.

In godotengine/godot#18404 dns says "Godot packing/embed it the wrong way" but closed the issue since Godot 3 no longer had the Embed Pck option at that point, I don't know if it was re-implemented the same way.

**Steps to reproduce:**
New Project, Export, Embed Pck, UPX

skyace65:
Adds a warning that embedded pck files will not work with UPX compression or code signing on windows. #3930 and this PR closes #3093

### DIFF
--- a/development/compiling/optimizing_for_size.rst
+++ b/development/compiling/optimizing_for_size.rst
@@ -159,6 +159,10 @@ If you are targeting desktop platforms, the
 `UPX <https://upx.github.io/>`_ compressor can be used.
 This can reduce binary size considerably.
 
+.. warning::
+
+    You cannot use embedded PCK files with UPX compression.
+
 However, keep in mind that some antivirus programs may detect UPX-packed
 binaries as a virus. Therefore, if you are releasing a commercial game,
 make sure to sign your binaries or use a platform that will distribute them.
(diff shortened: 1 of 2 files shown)


NathanLovato:
Thanks!

PR #3933, opened 2020-08-26
Clarify what UV means in CanvasItem shaders

Beuc:
Describing "UV" by "UV" could be a little more informative ;)


Calinou:
Thanks!

PR #3934, opened 2020-08-26
Fix double scaling

Beuc:
Following the tutorial, we only have one "* height_scale".
It's done in the line below.


Calinou (approved):
Thanks!

PR #3936, opened 2020-08-27
Normalize texture coordinates in [0,1] to avoid seam mark

Beuc:
Currently the vertex XZ position on the mesh plane is normalized from [-1,1] (size==2.0) to [-0.5,0.5] (divided by 2.0) in the varying "vertex_position", and used as texture coordinates.
This creates a seam mark near 0 in X and Y (cross) in the middle of the terrain and doesn't match the screenshots.

What we want is normalize to a texture position within [0,1]. "vertex_position" is hence a bit of a misnomer, and was renamed to "tex_position".
The result matches the screenshots.


Beuc:
Done.

Calinou:
Thanks!

PR #3940, opened 2020-08-28
Use a more explicit `git bisect` command in Bisecting regressions

Calinou:
This command makes it possible to specify the good and bad commits in any order, avoiding issues with specifying commits in the wrong order.


akien-mga:
Thanks!

PR #3942, opened 2020-08-28
Add information on how get_slide_count counts collisions

### LINKED ISSUE #3889: KinematicBody2D collision detection guide improvement
**Your Godot version:**
3.2.2 stable

**Issue description:**
So, in my player kinematicbody2d it always collides with the tilemap when on floor and can collide with another kinematic body (e.g. an enemy figure). However on following the kinematic2d guide, calling get_slide_count() is still only returning 1 item despite another kinematic body hitting it, when it should be 2, if my player is standing still (e.g. not jumping or walking)

Turns out the api has this to say 'Returns the number of times the body collided and *changed direction*'. Maybe it would be good to include in the documentation listed below stating how to use the slide count to include this 'changed direction' (whatever it exactly means by changed) because it's had me stumped for a while now, after following this guide?

https://docs.godotengine.org/en/stable/tutorials/physics/using_kinematic_body_2d.html#detecting-collisions

skyace65:
Someone opened an issue that the Using KinematicBody2D page doesn't explain how the get_slide_count counts collisions. It does not count collisions that don't change the KinematicBody direction. The docs do say under the code example to go to the class reference for more information , but I'm not sure if that would be considered enough. If you guys think that's already enough then close this PR and that issue, if not merge this. Closes #3889

### DIFF
--- a/tutorials/physics/using_kinematic_body_2d.rst
+++ b/tutorials/physics/using_kinematic_body_2d.rst
@@ -130,6 +130,8 @@ and ``get_slide_collision()``:
         var collision = get_slide_collision(i)
         print("I collided with ", collision.collider.name)
 
+.. note:: `get_slide_count()` only counts times the body has collided and changed direction.      
+
 See :ref:`KinematicCollision2D <class_KinematicCollision2D>` for details on what
 collision data is returned.
 


NathanLovato:
Thanks. I think it's good to have some repeated information, especially when its absence can confuse some users.

PR #3943, opened 2020-08-29
Document the GDScript debugger not supporting threads yet

Calinou:
The notice is duplicated to make it harder to miss.

See https://github.com/godotengine/godot/issues/2446.


NathanLovato:
Thanks!

PR #3944, opened 2020-08-29
Document how to start a server in Exporting for dedicated servers

Calinou:
This mainly applies to projects where both the client and server are implemented as part of the same project.


NathanLovato (approved):
I'd like someone else to check this pr as I don't know about the features but it's looking clear to me

akien-mga (approved):

akien-mga:
Thanks!

PR #3945, opened 2020-08-29
Fix Euler angles image being unreadable on a dark background

Calinou:
- Brighten the image's colors to make them easier to see.


NathanLovato:
Thanks!

PR #3951, opened 2020-08-31
Document differences between Bullet and GodotPhysics

### LINKED ISSUE #2161: Document differences between GodotPhysics and Bullet physics backends
Godot has two different backends for 3D physics, the "GodotPhysics" built-in backend and the Bullet backend added as a module.

They don't always produce the same results for the same parameters, so there should be some documentation outlining the main differences between the two.

Some relevant issues:
- godotengine/godot#14650
- godotengine/godot#15744

skyace65:
Adds a page to the physics section going over the known differences between Bullet and GodotPhysics. Closes #2161. These are not all the differences, only the ones that have been brought up in issues.

### DIFF
--- a/tutorials/physics/bullet_and_godotphysics_differences.rst
+++ b/tutorials/physics/bullet_and_godotphysics_differences.rst
@@ -1,0 +1,22 @@
+.. _doc_bullet_and_godotphysics_differences:
+
+Bullet and GodotPhysics Differences
+===================================
+
+Godot includes two different physics engines for 3D physics, Bullet, which
+is the default, and GodotPhysics. There are several important differences
+in how the physics engines function. They will **not** always produce
+identical or even similar results.
+
+RigidBody3D
+-----------
+
+In Bullet physics, the center of mass for RigidBody3D nodes is the center
+of the node. In GodotPhysics, the center of mass is the average of the
+collision shape centers.
+
+Linear and angular damping behave very differently in Bullet compared to
+GodotPhysics. Values over 1 do not make a difference in Bullet, but do make
+a difference in GodoyPhysics. Additionally, in Bullet the effect at 1 is
+significantly stronger compared to GodotPhysics. Conversely, a value of -1
+is significantly less strong in Bullet compared to GodotPhysics.
--- a/tutorials/physics/index.rst
+++ b/tutorials/physics/index.rst
@@ -13,3 +13,4 @@ Physics
    ragdoll_system
    kinematic_character_2d
    soft_body
+   bullet_and_godotphysics_differences


tutorials/physics/bullet_and_godotphysics_differences.rst
```diff
@@ -0,0 +1,22 @@
+.. _doc_bullet_and_godotphysics_differences:
+
+Bullet and GodotPhysics Differences
+===================================
+
+Godot includes two different physics engines for 3D physics, Bullet, which
+is the default, and GodotPhysics. There are several important differences
+in how the physics engines function, and they will NOT always produce
```
Calinou, line 8:
```suggestion
in how the physics engines function, and they will **not** always produce
```

tutorials/physics/bullet_and_godotphysics_differences.rst
```diff
@@ -0,0 +1,22 @@
...
+identical or even similar results.
+
+RigidBody3D
+-----------
+
+In Bullet physics the center of mass for RigidBody3D nodes is the center
```
Calinou, line 14:
```suggestion
In Bullet physics, the center of mass for RigidBody3D nodes is the center
```

tutorials/physics/bullet_and_godotphysics_differences.rst
```diff
@@ -0,0 +1,22 @@
...
+of the node. In GodotPhysics the center of mass is the average of the
```
Calinou, line 15:
```suggestion
of the node. In GodotPhysics, the center of mass is the average of the
```

Calinou (changes requested):

skyace65:
Fixed.

tutorials/physics/bullet_and_godotphysics_differences.rst
```diff
@@ -0,0 +1,22 @@
+.. _doc_bullet_and_godotphysics_differences:
+
+Bullet and GodotPhysics Differences
```
akien-mga, line 3: Don't use Title Case.

tutorials/physics/bullet_and_godotphysics_differences.rst
```diff
@@ -0,0 +1,22 @@
...
+
+In Bullet physics, the center of mass for RigidBody3D nodes is the center
+of the node. In GodotPhysics, the center of mass is the average of the
+collision shape centers.
+
+Linear and angular damping behave very differently in Bullet compared to
+GodotPhysics. Values over 1 do not make a difference in Bullet, but do make
+a difference in GodoyPhysics. Additionally, in Bullet the effect at 1 is
+significantly stronger compared to GodotPhysics. Conversely, a value of -1
+is significantly less strong in Bullet compared to GodotPhysics.
```
AndreaCatania, line 22: This is no more true, lately there was an update to fix it: https://github.com/godotengine/godot/pull/37314. It's also back ported to 3.2.x https://github.com/godotengine/godot/pull/39084

AndreaCatania (changes requested):

madmiraal:
Personally, I think this page is no longer needed. The two issues that drove the creation of #2161, were shown to be a limitation in Bullet physics' damping that, as @AndreaCatania has [pointed out](https://github.com/godotengine/godot-docs/pull/3951#discussion_r485507194), has since been addressed.

In general, I think that if there is a difference between the way Godot and Bullet physics behaves, it's more likely to be a bug that should be addressed. If there are multiple issues, we have tracker issues that are a better way to collate the information than a page in the documentation.

On the other hand, if there were genuine, significant differences between Godot and Bullet physics that a user could use to make an informed decision about which physics engine to use, then I would be in full support of this page.

Regarding the centre of gravity difference. The real issue is not that the centre of gravity is calculated differently, but that the centre of gravity or, more generally, the moment of inertia is not easily definable. See [#29099](https://github.com/godotengine/godot/issues/29099).

In short, I don't think we need an entire page that currently only points out that the centre of gravity may be different if an object has multiple shapes that are not centred. Information that I'm not sure many users would find useful, never mind use to make a decision about which physics engine to use.

skyace65:
@akien-mga do you think this page is still necessary? Since damping is now consistent between Bullet and GodotPhysics there would only be one change to document, and it's already covered in the class reference.  And if not could you close issue #2161?

PR #3952, opened 2020-09-01
Add information on compiling targets

### LINKED ISSUE #2813: EMPHASISE the importance of release_debug target
And by EMPHASISE I mean **E M P H A S I S E**.

See this https://github.com/godotengine/godot/issues/32476 and this https://github.com/godotengine/godot/issues/32195. More issues like these can appear, so docs should clearly say that the default build is slower unless you use `release_debug` target. Also it should mention that different targets have different names (https://github.com/godotengine/godot/issues/32195#issuecomment-533545978).

I myself learned about this, like, last month >_>

skyace65:
Adds information to the Windows, Mac and X11 compiling pages on compiling targets. You cannot miss this information now so this should close #2813

### DIFF
--- a/development/compiling/compiling_for_linuxbsd.rst
+++ b/development/compiling/compiling_for_linuxbsd.rst
@@ -92,11 +92,22 @@ Start a terminal, go to the root dir of the engine source code and type:
 
 ::
 
-    scons -j8 platform=linuxbsd
+    scons -j8 platform=linuxbsd target=release_debug
 
 A good rule of thumb for the ``-j`` (*jobs*) flag, is to have at least as many
 threads compiling Godot as you have cores in your CPU, if not one or two more.
 Feel free to add the ``-j`` option to any SCons command you see below.
+
+You have 3 target options when compiling.
+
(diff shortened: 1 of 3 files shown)


development/compiling/compiling_for_linuxbsd.rst
```diff
@@ -92,12 +92,23 @@ Start a terminal, go to the root dir of the engine source code and type:
...
 Feel free to add the ``-j`` option to any SCons command you see below.
 
+You have 3 target options when compiling.
+
+- ``debug``, Performance will be worse and the executable will be larger.
+  However you can debug Godot and it compiles the fastest.
+- ``release_debug``. Godot will have normal performance and file size, and you
+  can debug it.
+- ``release``. Godot will have normal performance and file size, but Godot
+  cannot be debugged.
```
akien-mga, line 108: The point of `release` is that performance is *higher* that `release_debug`.

akien-mga: IMO this should be covered in the Introduction to the buildsystem (if it isn't already), and each "compiling for" page should have a link to the introduction.

PR #3954, opened 2020-09-02
Improve import process page

### LINKED ISSUE #1465: Document File limitations and when to use ResourceLoader (remappings)
Some File methods (like checking for existence) may not account for remappings like ResourceLoader.

So e.g. checking whether a `.tscn` exists works in editor but not in an exported game, where it is converted to another file instead. Same for files moved to `.import`.

You should use ResourceLoader in those cases.

https://github.com/godotengine/godot/issues/18740

### LINKED ISSUE #2723: Import: Explain the process to reimport a large number of Texturefiles with other than the default settings
https://docs.godotengine.org/en/3.1/getting_started/workflow/assets/import_process.html

I'm trying to reimport all my images with midmaps turned on. 
So I check the "Midmaps" checkbox, go to "Preset..." -> set as default for Texture, select all my images in the filebrowser and hit reimport.
It will reimport, but it won't reimport with midmaps turned on. Reimporting single images works.

Neither would selecting the files I want and changing the settings and then just hitting "reimport"
Which I thought would be the way to do this initially.

It took me a few days an a lot of persistence to finally realize that there are some grey boxes showing up on the left if I select more than one file and that I had to check these to make sure midmaps are on on all files. This is explained nowhere and yet it's anything but intuitive or self-explanatory that setting a preset is not enough to set the reimport options. (Why is that anyway?https://github.com/godotengine/godot/issues/31645)

### LINKED ISSUE #3718: Docs on the resource import system and automatic remapping
**Your Godot version:**
3.0+

**Issue description:**
We have docs on the import process of Godot 3.0+, but it's not very explicit about the automatic remapping that happens between the original resource and the imported one when using specific APIs that support it (`load()`, `ResourceLoader.load()`).

It should also explain what happens when exporting a project (original resources are not included, only the imported version) and what this means for manual loading of resources (again, which should use `ResourceLoader` and not the `File` API).

https://docs.godotengine.org/en/stable/getting_started/workflow/assets/import_process.html

skyace65:
Adds several improvements to the import process page.

-Adds a paragraph explaining that `ResourceLoader` needs to be used instead of `File` when dealing with imported assets.
-Adds a section explaining how to reimport multiple files.
-Change Godot 3.0 to Godot 3.0+
-Reword part of a paragraph to remove the word easy.

Closes #3718, closes #1465, and closes #2723.

Quick question related to this. Is [ResourceLoader](https://docs.godotengine.org/en/stable/classes/class_resourceloader.html) a class? Because the class reference calls it a singleton.


NathanLovato:
Thank you!

PR #3955, opened 2020-09-02
Add yield with multiple arguments example to GDSCript Basics

HaSa1002:
I think this is a really powerful feature, that should be taught somewhere.

However my example is not the best. I thought about using the `HTTPRequest` class and it's `_request_completed` signal, but it requires knowledge about the node and would be a little longer. On the other hand, it would be a real world example.

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_basics.rst
+++ b/getting_started/scripting/gdscript/gdscript_basics.rst
@@ -1591,19 +1591,19 @@ You can also get the signal's argument once it's emitted by an object:
     # Wait for when any node is added to the scene tree.
     var node = yield(get_tree(), "node_added")
 
-If there are more than one argument ``yield`` returns an array containing
-the arguments:
-
-::
+If there is more than one argument, ``yield`` returns an array containing
+the arguments::
 
     signal done(input, processed)
+    
     func process_input(input):
         print("Processing initialized")
         yield(get_tree(), "idle_frame")
         print("Waiting")
         yield(get_tree(), "idle_frame")
         emit_signal(input, "Processed " + input)
-    
+
+
     func _ready():
         process_input("Test") # Prints: Processing initialized
         var data = yield(self, "done") # Prints: waiting
--- a/getting_started/scripting/gdscript/gdscript_basics.rst
+++ b/getting_started/scripting/gdscript/gdscript_basics.rst
@@ -1591,6 +1591,24 @@ You can also get the signal's argument once it's emitted by an object:
     # Wait for when any node is added to the scene tree.
     var node = yield(get_tree(), "node_added")
 
+If there are more than one argument ``yield`` returns an array containing
+the arguments:
+
+::
+
+    signal done(input, processed)
+    func process_input(input):
+        print("Processing initialized")
+        yield(get_tree(), "idle_frame")
+        print("Waiting")
+        yield(get_tree(), "idle_frame")
+        emit_signal(input, "Processed " + input)
+    
+    func _ready():
+        process_input("Test") # Prints: Processing initialized
+        var data = yield(self, "done") # Prints: waiting
+        print(data[1]) # Prints: Processed Test
+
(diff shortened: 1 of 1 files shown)


getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1591,6 +1591,24 @@ You can also get the signal's argument once it's emitted by an object:
     # Wait for when any node is added to the scene tree.
     var node = yield(get_tree(), "node_added")
 
+If there are more than one argument ``yield`` returns an array containing
```
Calinou, line 1594:
```suggestion
If there is more than one argument, ``yield`` returns an array containing
```

getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1591,6 +1591,24 @@ You can also get the signal's argument once it's emitted by an object:
...
+the arguments:
+
+::
+
+    signal done(input, processed)
```
Calinou, line 1599:
```suggestion
the arguments::

    signal done(input, processed)
    
```

getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1591,6 +1591,24 @@ You can also get the signal's argument once it's emitted by an object:
...
+    func process_input(input):
+        print("Processing initialized")
+        yield(get_tree(), "idle_frame")
+        print("Waiting")
+        yield(get_tree(), "idle_frame")
+        emit_signal(input, "Processed " + input)
+    
+    func _ready():
```
Calinou, line 1607:
```suggestion


    func _ready():
```

Calinou:
Note that this pull request will conflict with https://github.com/godotengine/godot-docs/pull/3623.

I'm not sure exactly how we should handle this. @vnen Any ideas? Should we change this PR's base branch so it only applies to the `3.2` branch?

akien-mga:
Yeah, it's probably better to merge in `3.2` only.

getting_started/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1591,6 +1591,24 @@ You can also get the signal's argument once it's emitted by an object:
     # Wait for when any node is added to the scene tree.
     var node = yield(get_tree(), "node_added")
 
+If there is more than one argument, ``yield`` returns an array containing
+the arguments::
```
Sslaxx, line 1595:
```
If there is more than one argument, ``yield`` returns an array containing
the arguments::
```

Might be:

```
If there is more than one argument, ``yield`` returns an array containing
the arguments.

::
```

akien-mga: That's equivalent, the `::` at the end of a sentence will become `:` + a codeblock.
https://www.sphinx-doc.org/en/1.8/usage/restructuredtext/basics.html#literal-blocks

I think we tend to use both forms in the docs currently, we haven't really defined that one should be used over the other (yet).

akien-mga: Ah well not strictly equivalent as your suggestion ends with a dot, not colon. I don't have a strong opinion there.

HaSa1002: It's mixed in the docs. I personally think that using the colon is better when you refer directly to the code (what I do in that passage). Otherwise the dot should be used, but that's just personal preference.

NathanLovato:
Should the PR target `stable` or `3.2`?

@HaSa1002, could you update your PR to not target master? The `master` branch corresponds to the Godot 4.0 docs.

HaSa1002:
stable or 3.2?

HaSa1002:
changed to 3.2, as akien stated above :D

NathanLovato:
Hugo confirmed we should use 3.2. Thanks!

PR #3958, opened 2020-09-02
Edit "What are Godot classes"

### LINKED ISSUE #3910: "Godot scenes and scripts are classes" article is incorrect and confusing
**Your Godot version:** Applies to all versions since this article was added.

**Issue description:**

This article is incorrect, and at a minimum, extremely confusing. A scene is not a class, a scene is a collection of nodes. In object-oriented programming, a class is a type with linear inheritance. You can't have a script extend `MyScene.tscn`, you can't have `MyScene.tscn` extend a type (the closest equivalent is having the root node extend a type). Scenes can be composed of many nodes, but a class always extends a single base type (which can extend other single types, linearly). Scenes are completely different from classes and this documentation article makes no sense.

I'm not sure what the original intention of this article was, CC @willnationsdev @NathanLovato 

There are also other confusing statements in this article, such as:

> User-created types are not technically classes

...but what about in C#, where all user scripts are literally C# classes?

> The scene is *always an extension of the script attached to its root node*. You can see all the nodes it contains as part of a single class.

No, this does not make any sense. Child nodes are not part of their parent's class, they are part of the scene tree. If you take a script attached to a root node of a scene and instance another node with that node's script, the children of that scene do not come with it... because the child nodes aren't part of the parent's class.

In a way, scenes do not even exist at runtime at all. Scenes are just reusable, instanceable, and inheritable groups of nodes. If you run a game and look at the remote scene tree in the editor, you do not see any scenes, you only see the editor.

~~I think this article should just be deleted. At a minimum it's extremely confusing and misleads users.~~

**URL to the documentation page:** https://docs.godotengine.org/en/latest/getting_started/workflow/best_practices/what_are_godot_classes.html

Source: https://github.com/godotengine/godot-docs/blob/master/getting_started/workflow/best_practices/what_are_godot_classes.rst

NathanLovato:
Changes:

- Simplify the article
- Change the title to better reflect the guide's point
- Try to make it clearer that scripts and scenes are not actually
- classes, but rather to be thought of as if they were classes in an OO
- language; OO principles apply to them
- Move section about the performances of scripts vs PackedScene to scenes_versus_scripts.rst

Resolves #3910

*Bugsquad edit:* Preview articles [here](https://github.com/NathanLovato/godot-docs/blob/edit/what-are-godot-classes/getting_started/workflow/best_practices/scenes_versus_scripts.rst) and [here](https://github.com/NathanLovato/godot-docs/blob/edit/what-are-godot-classes/getting_started/workflow/best_practices/what_are_godot_classes.rst).

### DIFF
--- a/getting_started/workflow/best_practices/scenes_versus_scripts.rst
+++ b/getting_started/workflow/best_practices/scenes_versus_scripts.rst
@@ -186,7 +186,7 @@ with it, and finally adds it as a child of the ``Main`` node:
         {
             Child = new Node();
             Child.Name = "Child";
-            Child.Script = (Script)ResourceLoader.Load("child.gd");
+            Child.Script = ResourceLoader.Load<Script>("child.gd");
             Child.Owner = this;
             AddChild(Child);
         }
--- a/getting_started/workflow/best_practices/scenes_versus_scripts.rst
+++ b/getting_started/workflow/best_practices/scenes_versus_scripts.rst
@@ -147,6 +147,60 @@ this way.
 
 On the downside, it also means having to use largely imperative programming.
 
+Performance of Script vs PackedScene
+------------------------------------
+
+One last aspect to consider when choosing scenes and scripts is execution speed.
+
+As the size of objects increases, the scripts' necessary size to create and
+initialize them grows much larger. Creating node hierarchies demonstrates this.
+Each Node's logic could be several hundred lines of code in length.
+
+The code example below creates a new ``Node``, changes its name, assigns a
+script to it, sets its future parent as its owner so it gets saved to disk along
+with it, and finally adds it as a child of the ``Main`` node:
+
+.. tabs::
+  .. code-tab:: gdscript GDScript
+
+    # Main.gd
+    extends Node
+
+    func _init():
+        var child = Node.new()
+        child.name = "Child"
+        child.script = preload("Child.gd")
+        child.owner = self
+        add_child(child)
+
+  .. code-tab:: csharp
+
+    using System;
+    using Godot;
+
+    public class Main : Resource
+    {
+        public Node Child { get; set; }
+
+        public Main()
+        {
+            Child = new Node();
+            Child.Name = "Child";
+            Child.Script = (Script)ResourceLoader.Load("child.gd");
+            Child.Owner = this;
+            AddChild(Child);
+        }
+    }
+
+Script code like this is much slower than engine-side C++ code. Each instruction
+makes a call to the scripting API which leads to many "lookups" on the back-end
+to find the logic to execute.
+
+Scenes help to avoid this performance issue. :ref:`PackedScene
+<class_PackedScene>`, the base type that scenes inherit from, defines resources
+that use serialized data to create objects. The engine can process scenes in
+batches on the back-end and provide much better performance than scripts.
+
 Conclusion
 ----------
 
--- a/getting_started/workflow/best_practices/what_are_godot_classes.rst
+++ b/getting_started/workflow/best_practices/what_are_godot_classes.rst
@@ -1,127 +1,59 @@
 .. _doc_what_are_godot_classes:
 
-Godot scenes and scripts are classes
-====================================
+Applying object-oriented principles in Godot
+============================================
 
-In Godot, scripts and scenes can both be the equivalent of classes in an
-Object-Oriented programming language. The main difference is that scenes are
-`declarative code <https://en.wikipedia.org/wiki/Declarative_programming>`_,
-while scripts can contain `imperative code
-<https://en.wikipedia.org/wiki/Imperative_programming>`_.
+The engine offers two main ways to create reusable objects: scripts and scenes. Neither of these
+technically define classes under the hood.
 
-As a result, many best practices in Godot boil down to applying Object-Oriented
-design principles to the scenes, nodes, or scripts that make up your game.
+Still, many best practices using Godot involve applying object-oriented programming principles to
+the scripts and scenes that compose your game. That is why it's useful to understand how we can
+think of them as classes.
 
-This guide explains how scripts and scenes work in the engine's core, to help
-you get a sense of how Godot works under the hood, and to help you better
-understand where some of this series' best practices come from.
+This guide briefly explains how scripts and scenes work in the engine's core to help you understand
+how they work under the hood.
 
-Making sense of classes in Godot
---------------------------------
+How scripts work in the engine
+------------------------------
 
-Godot Engine provides built-in classes like :ref:`Node <class_Node>`.
-User-created types are not technically classes. Instead, they are resources that
-tell the engine a sequence of initializations to perform on one of the engine's
-built-in classes.
+The engine provides built-in classes like :ref:`Node <class_Node>`. You can extend those to create
+derived types using a script.
 
-Godot's internal classes have methods that register a class's data with a
-:ref:`ClassDB <class_ClassDB>`. This database provides runtime access to class
-information. ``ClassDB`` contains information about classes like:
+These scripts are not technically classes. Instead, they are resources that tell the engine a
+sequence of initializations to perform on one of the engine's built-in classes.
 
-- properties
-- methods
-- constants
-- signals
+Godot's internal classes have methods that register a class's data with a :ref:`ClassDB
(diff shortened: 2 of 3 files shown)


NathanLovato:
Here's a first pass at the article. After shortly talking with @aaronfranke yesterday and re-reading both the current and original versions, I felt it needed a bit of a split and some simplification.

I tried to focus on one point: many best practices in Godot are about applying OO principles to scenes and scripts. This version strips some details and extra notes from the original. The part about performances of scenes vs scripts moved to the scene vs script article.

getting_started/workflow/best_practices/scenes_versus_scripts.rst
```diff
@@ -147,6 +147,63 @@ this way.
 
 On the downside, it also means having to use largely imperative programming.
 
+Performances of scripts vs PackedScenes
```
willnationsdev, line 150: I've personally always thought that the plural form, `Performances`, just sounds *incredibly* awkward in English. You only use it when referring to specific events of evaluating them, not when describing their performance in general.

I'd just drop the 's' to make it more clean. And change `scripts` to `Script`. If that's not enough of a reason, then you could also add more detail to the topic: `Comparing the performance of Script vs PackedScene`.

willnationsdev (commented):
Nice! I like the way things are split up here. :-D

NathanLovato:
Thanks, Will, just addressed your comments.

Feel free to use the suggestion feature and, anyone, if you have merge rights, to use it to edit the PR directly. I'm not pretty about my code or text, you're welcome to modify it anytime.

getting_started/workflow/best_practices/what_are_godot_classes.rst
```diff
@@ -1,127 +1,66 @@
...
-                Child.Name = "Child";
-                Child.Script = (Script)ResourceLoader.Load("child.gd");
-                Child.Owner = this;
-                AddChild(Child);
-            }
-        }
-    }
+    Even scripts that don't use the ``extends`` keyword implicitly inherit from the engine's base
+    :ref:`Reference <class_Reference>` class. This implicit type allows the object to defer to the
+    script's content where the engine logic deems appropriate.
```
willnationsdev, line 50: `This implicit type allows` -> `The native type inheritance allows`

Reason: it isn't just the "implicit" type that enables this behavior, but just the existence of an underlying type at all.

getting_started/workflow/best_practices/what_are_godot_classes.rst
```diff
@@ -1,127 +1,66 @@
...
 
-Script code like this is much slower than engine-side C++ code. Each change
-makes a separate call to the scripting API which leads to many "look-ups" on the
-back-end to find the logic to execute.
+    As a result, you can instance scripts without the ``extends`` keyword from code, but you cannot
+    attach them to a :ref:`Node <class_Node>`.
```
willnationsdev, line 53:
```rst
As a result, you can instantiate scripts without the ``extends`` keyword from code.
Since they extend ``Reference`` though, you cannot attach them to a :ref:`Node <class_Node>`.
```

Reason: more clarity.

getting_started/workflow/best_practices/what_are_godot_classes.rst
```diff
@@ -130,10 +69,11 @@ The content of a scene helps to define:
...
 
-Many Object-Oriented principles which apply to written code *also* apply to
-scenes.
+Why is any of this important to scene organization? Because instances of scenes *are* objects. As a
+result, many Object-Oriented principles that apply to written code also apply to scenes: single
+responsibility, encapsulation, and others.
 
-The scene is *always an extension of the script attached to its root node*. You
-can see all the nodes it contains as part of a single class.
+The scene is *always an extension of the script attached to its root node* so you can interpret it
```
willnationsdev, line 76: `so you can` -> `, so you can`

willnationsdev (changes requested):
After this round of changes, I think it'll be good for merging.

willnationsdev:
I don't have merge rights, so I can't edit the PR directly. :-P

getting_started/workflow/best_practices/what_are_godot_classes.rst
```diff
@@ -1,127 +1,66 @@
...
-    using Godot;
+.. note::
 
-    namespace ExampleProject
-    {
-        public class Main : Resource
-        {
-            public Node Child { get; set; }
+    Objects check their attached script before the database, which is why scripts can override
+    built-in methods. If a script defines a ``_get_property_list()`` method, Godot merges that data
```
aaronfranke, line 44: I think explaining how `_get_property_list()` works is too out-of-scope for an article about applying object-oriented principles. It's also something very rarely used in projects, I've never used it, and I just did a recursive grep for all the projects on my computer and the only instance is in a GDNative project.

getting_started/workflow/best_practices/what_are_godot_classes.rst
```diff
@@ -1,127 +1,66 @@
...
-Objects check their attached script before the database. This is why scripts can
-override built-in methods. If a script defines a ``_get_property_list()`` method,
-Godot appends that data to the list of properties the Object fetches from the
-ClassDB. The same is true for other declarative code.
-
-Even scripts that don't inherit from a built-in type, i.e. scripts that don't
-start with the ``extends`` keyword, implicitly inherit from the engine's base
-:ref:`Reference <class_Reference>` class. This allows the Object to defer
-to the script's content where the engine logic deems appropriate.
+Applying Object-Oriented principles in Godot
```
aaronfranke, line 3: Why are the `O` letters capitalized?

aaronfranke (changes requested):

getting_started/workflow/best_practices/what_are_godot_classes.rst
aaronfranke: Maybe just delete this note? Doesn't everything run by the engine that can have a script attached have the engine checking that script first? Also, isn't it that the script overrides the engine methods, not that the engine defers to them?

getting_started/workflow/best_practices/what_are_godot_classes.rst
willnationsdev: Well, the specific subsection within the article is talking about the relationship between a Script and its contributions to the reflection data collected from ClassDB, so it's *kinda* on topic in the sense that it describes further how Scripts contribute to class information.

getting_started/workflow/best_practices/what_are_godot_classes.rst
willnationsdev: > Also, isn't it that the script overrides the engine methods, not that the engine defers to them?

This is actually true in the general case. I think what I had in mind while writing that statement initially were situations where the engine directly attempts to call a method on a script (like notification callbacks, declarative code methods, etc.). In those cases, the engine directly says, "does it have a script? Does the script have this method? If so, call it."

> Maybe just delete this note?

Yeah. Re-examining it, it doesn't actually add much to the content of the article since it effectively restates existing information in a different way. Seems so anyway.

NathanLovato:
I just addressed all review points.

getting_started/workflow/best_practices/what_are_godot_classes.rst
```diff
@@ -1,127 +1,59 @@
...
-    }
+.. note::
 
-Script code like this is much slower than engine-side C++ code. Each change
-makes a separate call to the scripting API which leads to many "look-ups" on the
-back-end to find the logic to execute.
+    Even scripts that don't use the ``extends`` keyword implicitly inherit from the engine's base
+    :ref:`Reference <class_Reference>` class. As a result, you can instantiate scripts without the
+    ``extends`` keyword from code. Since they extend ``Reference`` though, you cannot attach them to
+    a :ref:`Node <class_Node>`.
```
NathanLovato, line 46: @willnationsdev @aaronfranke I'll just ping you to check this note. I trimmed it with the intention of keeping the article as lean and focused on the topic as possible. The fact GDScript files implicitly extend Reference seems important to repeat in some places around the docs, to me, as it can be a source of confusion.

NathanLovato:
@aaronfranke thanks for the suggestions, I added them to the PR!

aaronfranke (approved):
This is a vast improvement over the old article, and I don't have any further suggestions at the moment.

NathanLovato:
Thanks much for taking the time to review it!

PR #3959, opened 2020-09-02
Fix and improve changing application icon for Windows page

### LINKED ISSUE #1446: An Addition to "Changing application icon for windows"
I was following the instructions on the "Changing application icon for windows" documentation page. It was not working for me properly, so I tried a little bit for myself. I think some of my experiences are worth to be shared with the community to add good looking icons to exported Windows applications:

1. I discovered that **it is important to clear the Windows icon cache to see results**. Maybe this should be mentioned in the documention. You can find the command line commands here: https://superuser.com/questions/499078/refresh-icon-cache-without-rebooting

2. At first I tried GIMP (with the Iconify script) and rcedit as it was mentioned in the article. The problem here was that the icon looked awful at the size the icons have for example when they are on the desktop. I discovered that **your .ico file needs these icon sizes: 256x256, 128x128, 64x64, 48x48, 32x32, 16x16 - probably exactly in this order** (256 as the first icon in the ico file, 16 as the last). I found that out by looking at my exported exe file with the default Godot icon. The **color depth is not important** it seems. It was 24 bits before in the Godot icon, I used 32 bits, because this looks better with my icon.

3. GIMP and the Iconify script isn't the best solution for this in my opinion, because it automatically creates some of the used icon sizes but not all! So **I used the "Greenfish Icon Editor Pro" that is also open source** and very convenient to use for this task.

I hope with these tips I could help some users who had the same problem I had with Windows icons!

skyace65:
There have been some changes since 3.0 released with how Godot handles windows icons that needed to be fixed.

-All images have  been updated and gone through png compression
-There is now a separate setting for the Windows taskbar icon `Windows Native Icon` which this has been updated to use. Also my spelling of `Macos Native Icon` in this section is not a mistake, that's literally how it is spelled in the project settings.
-The old Gimp tutorial no longer works because the script it uses is no longer available. I managed to find a video on youtube and have added a link to that. It's not an ideal tutorial and goes too fast in some parts. But it's the best I could find and actually uses the newer version of Gimp (most stuff on youtube uses 2.8 with the old UI)

There's also a few other things I've changed and added.
-The section on making ICO files no longer makes sense at the bottom of the page since the taskbar and application icon both take ICO files now. So I've moved it to the top of the page and simplified it. The tutorial I've linked and the ImageMagik command both use the correct resolutions for ICO files so I've removed the information going over that. In fact they even use 128x128 which according to the [Microsoft docs on icons](https://docs.microsoft.com/en-us/windows/win32/uxguide/vis-icons) you don't even need.
-At the bottom of the page I've added the commands to clear the windows icon cache. I made a custom Icon and didn't need to do this, But I figure this should be left as a note just in case.

Closes #1446


getting_started/workflow/export/changing_application_icon_for_windows.rst
```diff
@@ -7,6 +7,25 @@ By default, the exported project's icon will be the Godot icon.
 You will most likely want to change that for your project. There are two types
 of icons that can be changed on Windows: the file icon and the taskbar icon.
 
+Creating an ICO file
+--------------------
+
+Windows does not use formats such as png or jpg for application icons, instead
+it uses a windows only format called ICO. You can create your application icon
+in any program, as long as you can bring that file into a program, such as GIMP,
+which can export an ICO file.
```
NathanLovato, line 16:
```suggestion
Windows does not use formats such as png or jpg for application icons. Instead,
it uses a Windows-only format called ICO. You can create your application icon
in any program but you will have to convert it to an ICO file using a program such 
as GIMP.
```

getting_started/workflow/export/changing_application_icon_for_windows.rst
```diff
@@ -7,6 +7,25 @@ By default, the exported project's icon will be the Godot icon.
...
+
+`This tutorial <https://www.youtube.com/watch?v=uqV3UfM-n5Y>`_ goes over how to
+export an ICO file with GIMP.
```
NathanLovato, line 19:
```suggestion
`This video tutorial <https://www.youtube.com/watch?v=uqV3UfM-n5Y>`_ goes over how to
export an ICO file with GIMP.
```

getting_started/workflow/export/changing_application_icon_for_windows.rst
```diff
@@ -16,13 +35,15 @@ is running.
...
-
-.. note:: This is also the icon that gets displayed in the Godot project list.
+**Project → Project Settings → Application → Config → Windows Native Icon**.
+Click on the folder icon and select your ICO file.
 
 .. image:: img/icon_project_settings.png
 
+Doing this will only change the icon for windows. ``Macos Native Icon`` sets
+the icon for Mac. ``Icon`` sets the icon for everything else and the project
+manager.
```
NathanLovato, line 45:
```suggestion
This setting only changes the icon for your exported game on Windows. 
To set the icon for macOS, use ``Macos Native Icon``. And for any other platform,
use the ``Icon`` setting.
```

getting_started/workflow/export/changing_application_icon_for_windows.rst
```diff
@@ -72,28 +89,6 @@ You can now export the project. If it worked correctly, you should see this:
...
-
-It is also possible to convert a PNG image to an hiDPI-friendly ICO file
-using this `ImageMagick <https://www.imagemagick.org/>`_ command:
-
-.. code-block:: none
-
-    magick convert icon.png -define icon:auto-resize=256,128,64,48,32,16 icon.ico
+.. note:: if your icon isn't showing up properly try clearing the icon cache
+          by opening `Run` and using the command ``ie4uinit.exe -ClearIconCache``
+          or ``ie4uinit.exe -show`` on Windows 10
```
NathanLovato, line 94:
```suggestion
.. note:: 
          
          if your icon isn't showing up properly, on Windows 10, try clearing the
          icon cache. To do so, open `Run` and call the command 
          ``ie4uinit.exe -ClearIconCache`` or ``ie4uinit.exe -show``.
```

NathanLovato (approved):

NathanLovato:
I built locally to test, all good (to not have to wait 30 mn for the CI to rebuild)

PR #3960, opened 2020-09-03
Added line to clarify we'll be starting a new project.

androclus:
Going through the tutorial, it can be a little confusing for a beginning user getting to this point moving through the pages of the tutorial, since my Godot Editor is still open with the "instancing" project. Should I wipe and restart? Or am I continuing the project that was started? The docs have not said whether to close down the last project, or whether we are continuing with it.

I looked around a while, and it appears this is a new project (though it is never said). So I think this this simple one-liner should help.

### DIFF
--- a/getting_started/step_by_step/scripting.rst
+++ b/getting_started/step_by_step/scripting.rst
@@ -104,6 +104,8 @@ easier to get an overview of the concepts.
 Scene setup
 ~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


NathanLovato:
Thanks, I'll merge for now but note I'll be rewriting the step-by-step section in the coming months so it may not be worth tweaking until then.

PR #3965, opened 2020-09-03
Add C# example to OpenSimplexNoise reference

Hasnep:
I didn't see any any of these example code snippets on other pages in the API reference section, so I don't know if this change is wanted.

I was a bit confused about the C# interface for `OpenSimplexNoise`, so I wanted to share what I'd learnt.

### DIFF
--- a/classes/class_opensimplexnoise.rst
+++ b/classes/class_opensimplexnoise.rst
@@ -18,7 +18,8 @@ Description
 
 This resource allows you to configure and sample a fractal noise space. Here is a brief usage example that configures an OpenSimplexNoise and gets samples at various positions and dimensions:
 
-::
+.. tabs::
+ .. code-tab:: gdscript GDScript
 
     var noise = OpenSimplexNoise.new()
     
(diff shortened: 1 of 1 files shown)


classes/class_opensimplexnoise.rst
```diff
@@ -34,6 +35,23 @@ This resource allows you to configure and sample a fractal noise space. Here is
     print(noise.get_noise_3d(0.5, 3.0, 15.0))
     print(noise.get_noise_4d(0.5, 1.9, 4.7, 0.0))
 
+ .. code-tab:: csharp
+
+    OpenSimplexNoise noise = new OpenSimplexNoise();
+
+    // Configure
+    System.Random random = new System.Random(); 
+    noise.Seed = random.Next();
```
aaronfranke, line 44: So that the behavior is the same, and for simplicity:

```suggestion
    noise.Seed = GD.Randi(); 
```

Hasnep: Ah, thanks! I tried using `GD.RandI()` and when that didn't work then assumed it didn't exist in C#.

aaronfranke (approved):
This looks good to me.

akien-mga:
This file is part of the class reference and is generated. The source file is https://github.com/godotengine/godot/blob/master/modules/opensimplex/doc_classes/OpenSimplexNoise.xml

PR #3968, opened 2020-09-04
Document multiple language support for the class reference

HaSa1002:
I forgot to document the changes I did in godotengine/godot#40613...

I added a link to the tool that I develop and use myself to translate all the code examples. I suppose that is not ok, but I think it can help some people to directly provide C# code (even though it has to be reviewed, because translation is complicated and sometimes works in unexpected manner :))

Also is there a tool, to format sphinx tables automatically? Doing this by hand *twice* is really cumbersome.


NathanLovato:
Built and tested locally, looking good!

PR #3975, opened 2020-09-06
Fix RichTextEffects not loading correctly in BBCode in RichTextLabel

MrCelticFox:
Added class names to all three of the custom RichText examples, as without these the effects did not load correctly in the "Custom Effect Array" property of the RichTextLabel. (Godot v3.2.2.stable)

### DIFF
--- a/tutorials/gui/bbcode_in_richtextlabel.rst
+++ b/tutorials/gui/bbcode_in_richtextlabel.rst
@@ -232,6 +232,7 @@ Ghost
 
     tool
     extends RichTextEffect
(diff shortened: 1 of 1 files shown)


Calinou:
cc @Eoin-ONeill-Yokai

Eoin-ONeill-Yokai:
@Calinou Yes, this seems correct with the current state of Godot / Rich Text Effects. I'll make sure my own example projects are up to date as well.

Calinou (approved):
Thanks!

PR #3976, opened 2020-09-07
Improve importing images page

skyace65:
Removes the section at the beginning explaining that assets are imported now compared to older Godot versions. I've moved up the list of supported formats to the top of the page, it didn't make sense being in the middle of import settings. and the direct link to that section has been removed. The features page was the only doc page using that link, and its been updated. The svg scale option is now documented.


getting_started/workflow/assets/importing_images.rst
```diff
@@ -212,3 +200,11 @@ Invert Color
...
 
 Reverses the image's color. This is useful, for example, to convert a height map generated by external programs to depth map to use with :ref:`doc_standard_material_3d`.
+
+Svg
+---
+
+Scale
+~~~~~
+
+This option only applies to svg files. It controls the scale of the svg image.
```
Calinou, line 210:
```suggestion
This option only applies to SVG files. It controls the scale of the SVG image. The default scale (1.0) will make the imported SVG match its original design scale.
```

Calinou (changes requested):

skyace65:
Fixed.

akien-mga:
Thanks!

PR #3979, opened 2020-09-07
Qt Creator tutorial: Specify C++ code style settings location

Yetizone:
Hello, I have recently set up QtCreator to work with Godot's source code. 

Towards the end of the tutorial there's a section about how to configure the coding style and it says to go to Options > C++.
Since Options isn't immediately visible when looking at the IDE, I thought it might be a good idea to be a bit more explicit about the location. 

Here's a gif about what I mean: 

Tools_Options


Calinou (approved):
Thanks!

PR #3983, opened 2020-09-09
Fix of a grammatical error.

hubertekvall:
A minor fix but having correct grammar will increase the overall professional appearance of the Godot project. This exact change might not be the best one but it's at least more intelligible than the original text; native English speakers should chime in.

Whether this joke introduction for the entire section should be kept or not is another discussion. Either way it's better for the project if the grammar is correct.


akien-mga:
Thanks!

PR #4000, opened 2020-09-12
Initial port of optimization tutorials to master

clayjohn:
This ports the new optimization tutorials to master. 

Initially we were going to wait until master was more stable and reduz had finished the majority of the rendering optimizations. However, now that @NathanLovato is working on auditing the manual, it makes sense to have a version of these docs here so he can take them into account for the audit. 

I have done the bare minimum to prep these for 4.0. There will be some bits of information that is no longer relevant with 4.0 and there may be omissions.


NathanLovato:
Thanks much, I'll merge as-is and let's plan updates when the time is right for Godot 4.

PR #4001, opened 2020-09-12
Update and simplify android custom build instructions

eddex:
I recently used the [Android Custom Build page](https://docs.godotengine.org/en/latest/getting_started/workflow/export/android_custom_build.html) and noticed some things that could be improved. Additionally, I ran into an issue with the Windows ZIP extraction that somehow screws up the `sdkmanager` (more info on this can be found [here](https://stackoverflow.com/questions/60727326/sdkmanager-error-could-not-find-or-load-main-class-com-android-sdklib-tool-sdk)), so I added instructions on how to prevent those issues to the page as well.


getting_started/workflow/export/android_custom_build.rst
```diff
@@ -28,8 +28,9 @@ Go to the Project menu, and install the *Custom Build* template:
 Make sure export templates are downloaded. If not, this menu will help you
 download them.
 
-This will create an Gradle-based Android project in ``res://android/build`` and
-place a ``.gdignore`` file in ``res://android`` so the Godot filesystem ignores
+A Gradle-based Android project will be created under ``res://android/build``.
+
+Place a ``.gdignore`` file in ``res://android`` so the Godot filesystem ignores
```
m4gr3d, line 33: @eddex The original instruction is actually out of date and this instruction should not be needed as Godot already adds a `.gdignore` file in the `res://android/build` directory. 
You actually want the `res://android` directory to be visible so the user can install Android plugins if desired.

m4gr3d (changes requested):

getting_started/workflow/export/android_custom_build.rst
eddex: alright, gonna change that :)

m4gr3d (approved):
Looks good, thanks for the update!

akien-mga:
Thanks!

PR #4004, opened 2020-09-13
Document a fallback method to change the application icon on Windows

Calinou:
This method isn't as convenient as using rcedit, but it should be much more reliable.

See https://github.com/godotengine/godot-proposals/issues/1097.


akien-mga:
Thanks!

PR #4005, opened 2020-09-13
Remove UPX recommendation in Optimizing for size

Calinou:
The downsides (such as false positives in antivirus programs) are usually not worth the upsides.

Moreover, when distributing a game using a platform like Steam, SteamPipe will compress the data that will be sent to the player downloading the game, even if the original game data is uncompressed.


akien-mga:
Thanks!

PR #4007, opened 2020-09-13
Add a FAQ item about why Godot uses Vulkan/OpenGL instead of Direct3D

Calinou:
This is a relatively common question in various community channels.

### DIFF
--- a/about/faq.rst
+++ b/about/faq.rst
@@ -183,6 +183,24 @@ free and open-source integration, consider starting the integration work yoursel
 Godot is not owned by one person; it belongs to the community, and it grows along
 with ambitious community contributors like you.
 
+Why does Godot use Vulkan or OpenGL instead of Direct3D?
+--------------------------------------------------------
+
+Godot aims for cross-platform compatibility and open standards first and
+foremost. OpenGL and Vulkan are the technologies that are both open and
+available (nearly) on all platforms. Thanks to this design decision, a project
+developed with Godot on Windows will run out of the box on Linux, macOS, and
+more.
+
+Since Godot only has a few people working on its renderer, we would prefer
+having fewer rendering backends to maintain. On top of that, using a single API
+on all platforms allows for greater consistency with fewer platform-specific
+issues.
+
+In the long term, we may develop a Direct3D 12 renderer for Godot (mainly for
+the Xbox's purposes), but Vulkan and OpenGL will remain the default rendering
+backends on all platforms, including Windows.
+
 How should assets be created to handle multiple resolutions and aspect ratios?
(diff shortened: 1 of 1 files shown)


about/faq.rst
```diff
@@ -183,6 +183,30 @@ free and open-source integration, consider starting the integration work yoursel
...
+Why does Godot use Vulkan or OpenGL instead of Direct3D?
+--------------------------------------------------------
+
+This question usually stems from OpenGL having a bad reputation in the 2000s for
+poor driver compatibility and lower performance. However, many popular
+OpenGL-based games have proven to be well-optimized in practice.
+
+OpenGL drivers have gotten much better over the recent years. While AMD graphics
+drivers on Windows have historically suffered from lower OpenGL performance,
+this is being remedied by Vulkan which is a lower-level API with fewer
```
clayjohn, line 195: "this has been remedied by Vulkan"

about/faq.rst
```diff
@@ -183,6 +183,30 @@ free and open-source integration, consider starting the integration work yoursel
...
+driver-specific quirks.
+
+Direct3D is specific to Windows and Xbox, whereas Vulkan and OpenGL can run
+almost anywhere. (Vulkan is supported on macOS and iOS thanks to MoltenVK.)
+
+Since Godot only has a few people working on its renderer, we would prefer
+having fewer rendereing backends to maintain. On top of that, using a single API
```
clayjohn, line 202: "rendering" typo

about/faq.rst
```diff
@@ -183,6 +183,30 @@ free and open-source integration, consider starting the integration work yoursel
...
+--------------------------------------------------------
+
+This question usually stems from OpenGL having a bad reputation in the 2000s for
+poor driver compatibility and lower performance. However, many popular
+OpenGL-based games have proven to be well-optimized in practice.
+
+OpenGL drivers have gotten much better over the recent years. While AMD graphics
+drivers on Windows have historically suffered from lower OpenGL performance,
+this has been remedied by Vulkan which is a lower-level API with fewer
+driver-specific quirks.
```
akien-mga, line 196: I don't think this kind of background on how OpenGL used to be bad is useful here. We don't need to try to justify why users might ask this question, we just have to answer it.

IMO what users need to know is that Godot's aims for cross-platform compatibility and open standards first and foremost. OpenGL and Vulkan are the technologies that are both open and available (nearly) on all platforms, and therefore that's why we give them the priority. A game developed with Godot on Windows will run out of the box on Linux and macOS thanks to this design decision.

akien-mga (approved):

akien-mga:
Thanks!

PR #4008, opened 2020-09-13
Add a FAQ item about "When is the next release of Godot out?"

Calinou:
People still ask me this occasionally, as the [Godot release policy](https://docs.godotengine.org/en/latest/about/release_policy.html) page is usually not read by newcomers.


clayjohn (approved):

akien-mga (approved):
Not today, but surely tomorrow.

PR #4009, opened 2020-09-13
Clarify license requirement in Asset Library submitting instructions

aaronfranke:
One user submitted an asset with an incorrect license, and I had to explain what was wrong so that they could fix it. I think the information of what qualifies as a correct license should be in the documentation, so I added it.

```
* The license needs to be correct. The license listed on the asset
  library must match the license in the repository. The repo MUST
  have a license file, called either "LICENSE" or "LICENSE.md".
  This file must contain the license text itself and a copyright
  statement that includes the year(s) and copyright holder.
```


Calinou (approved):

akien-mga:
Thanks!

PR #2510, opened 2019-06-06
Update exporting_for_web.rst

BaseCrusher:
Added a description of how to use external files after HTML export as well as a simple example of how to do this.
I'm sure the language can be improved. But my English isn't good enough to do so. :(

PR #4015, opened 2020-09-15
Updated custom webexport to be more clear

BaseCrusher:
Added a description of how to use external files after HTML export as well as a simple example of how to do this.

Correlates with this pull request :
https://github.com/godotengine/godot-docs/pull/2510

### DIFF
--- a/getting_started/workflow/export/exporting_for_web.rst
+++ b/getting_started/workflow/export/exporting_for_web.rst
@@ -166,31 +166,6 @@ the ``.pck`` and ``.wasm`` files, which are usually large in size.
 The WebAssembly module compresses particularly well, down to around a quarter
 of its original size with gzip compression.
 
-Trying to access files which aren't compiled with the project but are used by the
-application needs to be loaded with a custom HTML page. The reason for that is 
-that the WebAssembly does not have the permission to access files that aren't
-compiled.
-
-Here is an example of the part of the HTML page that needs to be modified.
-Use the default HTML page `/misc/dist/html/full-size.html <https://github.com/godotengine/godot/blob/master/misc/dist/html/full-size.html>`__. as a refrence 
-and replace the lines 250 to 255 with.
-
-::
-		Engine.load(MAIN_PACK)
-		engine.preloadFile('$GODOT_BASENAME.js')
-		engine.preloadFile('$GODOT_BASENAME.pck')
-		engine.preloadFile('$GODOT_BASENAME.png')
-		engine.preloadFile('$GODOT_BASENAME.wasm')
-		engine.preloadFile('my_text.txt')
-		engine.init()
-		setStatusMode('indeterminate');
-		engine.setCanvas(canvas);
-		engine.start().then(() => {
-			setStatusMode('hidden');
-			initializing = true;
-		}, displayFailureNotice);
-
-
 Export options
 --------------
 
--- a/getting_started/workflow/export/exporting_for_web.rst
+++ b/getting_started/workflow/export/exporting_for_web.rst
@@ -166,6 +166,31 @@ the ``.pck`` and ``.wasm`` files, which are usually large in size.
 The WebAssembly module compresses particularly well, down to around a quarter
 of its original size with gzip compression.
 
+Trying to access files which aren't compiled with the project but are used by the
+application needs to be loaded with a custom HTML page. The reason for that is 
+that the WebAssembly does not have the permission to access files that aren't
+compiled.
+
+Here is an example of the part of the HTML page that needs to be modified.
+Use the default HTML page `/misc/dist/html/full-size.html <https://github.com/godotengine/godot/blob/master/misc/dist/html/full-size.html>`__. as a refrence 
+and replace the lines 250 to 255 with.
+
+::
+		Engine.load(MAIN_PACK)
+		engine.preloadFile('$GODOT_BASENAME.js')
+		engine.preloadFile('$GODOT_BASENAME.pck')
+		engine.preloadFile('$GODOT_BASENAME.png')
+		engine.preloadFile('$GODOT_BASENAME.wasm')
+		engine.preloadFile('my_text.txt')
+		engine.init()
+		setStatusMode('indeterminate');
+		engine.setCanvas(canvas);
+		engine.start().then(() => {
+			setStatusMode('hidden');
+			initializing = true;
+		}, displayFailureNotice);
+
+
 Export options
 --------------
 
--- a/tutorials/platform/customizing_html5_shell.rst
+++ b/tutorials/platform/customizing_html5_shell.rst
@@ -133,6 +133,23 @@ them by calling the :js:meth:`engine.preloadFile` method with a path to a file o
 with an ``ArrayBuffer`` object. In case of the ``ArrayBuffer``, or one of its views, a second argument
 must be specified to define an internal path for the loaded resource.
 
+Here is a short example for loading additionals files. In this case the file "my_text.txt" is loaded in
+when while the engine starts up.
+::
+		Engine.load(MAIN_PACK)
+		engine.preloadFile('$GODOT_BASENAME.js')
+		engine.preloadFile('$GODOT_BASENAME.pck')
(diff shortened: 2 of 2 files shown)


BaseCrusher:
Hmm, there seems to be an issue with my additions. Don't really know what is causing it though.

getting_started/workflow/export/exporting_for_web.rst
```diff
@@ -166,6 +166,31 @@ the ``.pck`` and ``.wasm`` files, which are usually large in size.
...
+		engine.preloadFile('$GODOT_BASENAME.png')
+		engine.preloadFile('$GODOT_BASENAME.wasm')
+		engine.preloadFile('my_text.txt')
+		engine.init()
+		setStatusMode('indeterminate');
+		engine.setCanvas(canvas);
+		engine.start().then(() => {
+			setStatusMode('hidden');
+			initializing = true;
+		}, displayFailureNotice);
```
akien-mga, line 191:
```suggestion
::

    Engine.load(MAIN_PACK)
    engine.preloadFile('$GODOT_BASENAME.js')
    engine.preloadFile('$GODOT_BASENAME.pck')
    engine.preloadFile('$GODOT_BASENAME.png')
    engine.preloadFile('$GODOT_BASENAME.wasm')
    engine.preloadFile('my_text.txt')
    engine.init()
    setStatusMode('indeterminate');
    engine.setCanvas(canvas);
    engine.start().then(() => {
    	setStatusMode('hidden');
    	initializing = true;
    }, displayFailureNotice);
```

getting_started/workflow/export/exporting_for_web.rst
```diff
@@ -166,6 +166,31 @@ the ``.pck`` and ``.wasm`` files, which are usually large in size.
...
 of its original size with gzip compression.
 
+Trying to access files which aren't compiled with the project but are used by the
+application needs to be loaded with a custom HTML page. The reason for that is 
+that the WebAssembly does not have the permission to access files that aren't
+compiled.
+
+Here is an example of the part of the HTML page that needs to be modified.
+Use the default HTML page `/misc/dist/html/full-size.html <https://github.com/godotengine/godot/blob/master/misc/dist/html/full-size.html>`__. as a refrence 
+and replace the lines 250 to 255 with.
```
akien-mga, line 176: Line numbers will change, so hardcoding them in the doc is not a good idea.

getting_started/workflow/export/exporting_for_web.rst
```diff
@@ -166,6 +166,31 @@ the ``.pck`` and ``.wasm`` files, which are usually large in size.
 The WebAssembly module compresses particularly well, down to around a quarter
 of its original size with gzip compression.
 
+Trying to access files which aren't compiled with the project but are used by the
+application needs to be loaded with a custom HTML page. The reason for that is 
+that the WebAssembly does not have the permission to access files that aren't
+compiled.
+
+Here is an example of the part of the HTML page that needs to be modified.
+Use the default HTML page `/misc/dist/html/full-size.html <https://github.com/godotengine/godot/blob/master/misc/dist/html/full-size.html>`__. as a refrence
```
akien-mga, line 175:
```suggestion
Use the default HTML page `misc/dist/html/full-size.html <https://github.com/godotengine/godot/blob/master/misc/dist/html/full-size.html>`__. as a reference 
```

getting_started/workflow/export/exporting_for_web.rst
akien-mga, line 176:
```suggestion
Here is an example of the part of the HTML page that needs to be modified.
Use the default HTML page `misc/dist/html/full-size.html <https://github.com/godotengine/godot/blob/master/misc/dist/html/full-size.html>`__
as a reference and replace the lines 250 to 255 with:
```

BaseCrusher:
Aight, so it's kinda embarassing for me. I've just taken a look into the Godot Docs. There is already a very good documentation for custom HTML export. I will just paste some example code there.

akien-mga:
Well we shouldn't copy paste existing documentation in other pages. If the documentation already exists, it just needs to be referred to where you think the link was missing.

You can use the `.. seealso::` directive to make a custom box that will refer users to extra documentation.

That being said there are already 6 references to `customizing_html5_shell` in that page, so maybe adding one is not necessary.

BaseCrusher:
Right, I've only inserted an example into the custom html export to be a litle bit more clearer. And reverted the changes on the plain html export.

tutorials/platform/customizing_html5_shell.rst
```diff
@@ -133,6 +133,23 @@ them by calling the :js:meth:`engine.preloadFile` method with a path to a file o
 with an ``ArrayBuffer`` object. In case of the ``ArrayBuffer``, or one of its views, a second argument
 must be specified to define an internal path for the loaded resource.
 
+Here is a short example for loading additionals files. In this case the file "my_text.txt" is loaded in
+when while the engine starts up.
+::
```
pycbouh, line 138: Please use `.. code-block:: js` to allow proper syntax highlighting.

pycbouh (changes requested):
If we are adding an example to this section, then in the spirit of the article it should be localized to one particular problem. There seems to be too much code and I'm not sure what it is supposed to demonstrate exactly. Calls to `preloadFile` itself pretty self-explanatory, IMO. But maybe I'm not seeing the problem this example aims to cover.

We should also avoid using code that is not a part of the built-in API and is not defined in the example. Like calls to `setStatusMode` and `displayFailureNotice`.

tutorials/platform/customizing_html5_shell.rst
```diff
@@ -133,6 +133,23 @@ them by calling the :js:meth:`engine.preloadFile` method with a path to a file o
...
+    Engine.load(MAIN_PACK)
```
Faless, line 139: This should be `Engine.load(EXECUTABLE_NAME)`

tutorials/platform/customizing_html5_shell.rst
```diff
@@ -133,6 +133,23 @@ them by calling the :js:meth:`engine.preloadFile` method with a path to a file o
...
+    engine.preloadFile('$GODOT_BASENAME.js')
+    engine.preloadFile('$GODOT_BASENAME.pck')
+    engine.preloadFile('$GODOT_BASENAME.png')
+    engine.preloadFile('$GODOT_BASENAME.wasm')
+    engine.preloadFile('my_text.txt')
```
Faless, line 144: This is wrong, you only need to preload the `pck` and your file (not the PNG, not the wasm, not the js).

Alternatively , you should be able to skip the `Engine.load(EXECUTABLE_NAME)`, `engine.preloadFile(MAIN_PACK)`, and `engine.init()` by using startGame:

```javascript
engine.preloadFile('my_text.txt').then(function() {
    engine.startGame(EXECUTABLE_NAME, MAIN_PACK).then(() => {
        setStatusMode('hidden');
        initializing = false;
    }, displayFailureNotice);
});
```

Faless (changes requested):
I hope I helped clarifying it a bit.

BaseCrusher:
Well, I just stumbled at this again a year after posting it on the original PR. 
Since at the time of posting there wasn't any explanation to how to custumize the loading behaviour for HTML, I wonted to add a small and quick description to how it is done. 
I have to admit that I'm not sure if the example is still needed. Though I think that for newcomers this might be helpful. Now, since I haven't done anything in Godot for a long time I don't know ho anything did change.

If there is no need fo this PR change you can just close it. 
I want to apologize for all this hassle. :(

NathanLovato:
@Faless would you include this example, once corrected, in this document? To know if we should request changes or close the PR.

@BaseCrusher No worries, we appreciate people getting out of their way to contribute to the project, even if a change doesn't make it in the docs. But let's see, I need someone else's input; I'm not familiar with web export.

pycbouh:
The API has since changed, and so have examples. I'll close this as outdated and lacking support, but if the current state of the article is missing any examples, feel free to open a new PR (but take the feedback given here into consideration).

PR #4016, opened 2020-09-15
Update Keywords to include class_name

arccoza:
Includes `class_name` in the keywords list, and updates the definition of `class`.


akien-mga:
Thanks!

PR #4017, opened 2020-09-15
Document unit testing in the engine and modules

Xrayez:
See godotengine/godot#40659.

Adds a new page for C++ unit testing in Godot Engine.

Adds a section to "Custom modules in C++" to document the process of writing module tests as well, reusing the same mechanism.

There are naming conventions which are worth to be considered: godotengine/godot-proposals#1388, but at its current state, this should be enough.

### DIFF
--- a/development/cpp/unit_testing.rst
+++ b/development/cpp/unit_testing.rst
@@ -1,0 +1,290 @@
+.. _doc_unit_testing:
+
+Unit testing
+============
+
+Godot Engine allows to write unit tests directly in C++. The engine integrates
+the `doctest <https://github.com/onqtam/doctest>`_ unit testing framework which
+gives ability to write test suites and test cases next to production code, but
+since the tests in Godot go through a different ``main`` entry point, the tests
+reside in a dedicated ``tests/`` directory instead, which is located at the root
+of the engine source code.
+
+Platform and target support
+---------------------------
+
+C++ unit tests can be run on Linux, macOS, and Windows operating systems.
+
+Tests can only be run with editor ``tools`` enabled, which means that export
+templates cannot be tested currently.
+
+Running tests
+-------------
+
+Before tests can be actually run, the engine must be compiled with the ``tests``
+build option enabled (and any other build option you typically use), as the
+tests are not compiled as part of the engine by default:
+
+.. code-block:: shell
+
+    scons tests=yes
+
+Once the build is done, run the tests with a ``--test`` command-line option:
+
+.. code-block:: shell
+
+    ./bin/<godot_binary> --test
+
+The test run can be configured with the various doctest-specific command-line
+options. To retrieve the full list of supported options, run the ``--test``
+command with the ``--help`` option:
+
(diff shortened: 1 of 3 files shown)


development/cpp/unit_testing.rst
```diff
@@ -0,0 +1,288 @@
...
+the `doctest <https://github.com/onqtam/doctest>`_ unit testing framework which
+gives ability to write test suites and test cases next to production code, but
+since the tests in Godot go through a different ``main`` entry point, the tests
+reside in a dedicated ``tests/`` directory instead, which is located at the root
+of the engine source code.
+
+Platform and target support
+---------------------------
+
+C++ unit tests can be run on Linux, MacOS, and Windows operating systems.
```
Calinou, line 16:
```suggestion
C++ unit tests can be run on Linux, macOS, and Windows operating systems.
```

development/cpp/unit_testing.rst
```diff
@@ -0,0 +1,288 @@
...
+
+Any other options and arguments after the ``--test`` command are treated as
+arguments for doctest.
+
+.. note::
+
+    Tests are compiled automatically if you happen to use ``dev=yes`` option,
+    which is recommended to supply if you plan on contributing to the engine
+    development, as some compilation warnings may be treated as errors in the
+    continuous integration system, which also runs those tests.
```
Calinou, line 54:
```suggestion
    Tests are compiled automatically if you use the ``dev=yes`` SCons option.
    ``dev=yes`` is recommended if you plan on contributing to the engine
    development as it will automatically treat compilation warnings as errors.
    The continuous integration system will fail if any compilation warnings are detected,
    so you should strive to fix all warnings before opening a pull request.
```

development/cpp/unit_testing.rst
```diff
@@ -0,0 +1,288 @@
...
+    ./bin/<godot_binary> --test --source-file="*test_validate*" --success --reporters=xml --out=doctest.txt
+
+.. seealso::
+
+    `doctest: Logging macros <https://github.com/onqtam/doctest/blob/master/doc/markdown/logging.md>`_.
+
+Testing failure paths
+~~~~~~~~~~~~~~~~~~~~~
+
+Sometimes, it's not always feasible to test for *expected* result. With the
```
Calinou, line 216:
```suggestion
Sometimes, it's not always feasible to test for an *expected* result. With the
```

Calinou (changes requested):

Xrayez:
Applied suggested changes.

I was thinking about adding a section on generating test coverage and running tests with sanitizers, but this could be done in a separate PR, and perhaps by someone more knowledgeable in this area or someone who has already done this (@qarmin, @Calinou). 🙂

In fact, if there are some memory leaks in tests, CI won't let you pass those, but this is more or less advanced topic for testing, and sanitizers are not supported on all platforms out of the box.

Calinou (approved):
Thanks!

PR #4023, opened 2020-09-17
qt_creator.rst: Add .glsl to the file extensions step

Yetizone:
Story behind this PR:
I was recently working on an issue which involved tinkering with shaders, but shader files weren't picked up by the IDE and it took me quite a while to figure out what was happening :) 

I've added the .glsl extension first in the list as I thought shader files are like core .cpp files and the tutorial reader should definetly be notified that they are not visible to the IDE by default.


akien-mga (approved):
Yep I do the same myself.

akien-mga:
Thanks!

PR #4025, opened 2020-09-17
Improve audio buses page

skyace65:
Improves the audio buses page based on improvements made to the class reference (https://github.com/godotengine/godot/pull/41987, https://github.com/godotengine/godot/pull/42089). The word "dirty" is removed from the description of distortion. high shelf filter, and low shelf filter now have their own description, which is now accurate. This should be merged after 42089 in merged in the Godot repository.


akien-mga:
Thanks!

PR #4026, opened 2020-09-17
Update encryption key page according to the new PCK encryption feature.

bruvzg:
Update "Script encryption" page to "PCK encryption" to match cnahgest done in https://github.com/godotengine/godot/pull/38308


akien-mga:
> cnahgest

covfefe

akien-mga:
Thanks!

PR #4031, opened 2020-09-18
Fix varying name in Your first spatial shader

CDWimmer:
vertex_position -> tex_position


akien-mga:
The sentence right above it mentions vertrx_position too.

CDWimmer:
You're absolutely right

akien-mga:
Thanks!

PR #4033, opened 2020-09-18
Optimize PCK encryption screenshot

bruvzg:
A little improvement for #4026, new screenshot is:
- taken in loDPI mode (almost 3 times smaller).
- taken with `--single-window` (no macOS rounded window border shadows in the corners).
- have a slightly better composition (less empty space on the image, all tabs visible).


Calinou (approved):
Thanks!

PR #4038, opened 2020-09-18
Document why Godot doesn't support changing the monitor resolution

Calinou:
This also adds advice on using mipmapping to support downsampling 2D textures without aliasing.

This closes https://github.com/godotengine/godot/issues/14542.


akien-mga:
Thanks!

PR #4041, opened 2020-09-19
Add information about external materials not updated on reimport

QbieShay:

### DIFF
--- a/getting_started/workflow/assets/importing_scenes.rst
+++ b/getting_started/workflow/assets/importing_scenes.rst
@@ -173,7 +173,8 @@ When materials are built-in, they will be lost each time the source scene
 is modified and re-imported.
 
 .. note::
-   Godot will not reimport materials that are stored in external files.
+
+    Godot will not reimport materials that are stored in external files.
 
 Keep On Reimport
 ^^^^^^^^^^^^^^^^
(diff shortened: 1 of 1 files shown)


getting_started/workflow/assets/importing_scenes.rst
```diff
@@ -172,6 +172,9 @@ most 3D DCCs don't have the same material options as those present in Godot.
 When materials are built-in, they will be lost each time the source scene
 is modified and re-imported.
 
+.. note::
+   Godot will not reimport materials that are stored in external files.
```
Calinou, line 176:
```suggestion
.. note::

    Godot will not reimport materials that are stored in external files.
```

Calinou (changes requested):
Formatting needs to be fixed so Sphinx accepts it:

getting_started/workflow/assets/importing_scenes.rst
```diff
@@ -172,6 +172,10 @@ most 3D DCCs don't have the same material options as those present in Godot.
 When materials are built-in, they will be lost each time the source scene
 is modified and re-imported.
 
+.. note::
+
+    Godot will not reimport materials that are stored in external files.
```
Calinou, line 177:
```suggestion
    Godot will not reimport materials that are stored in external files
    unless you remove the associated `.material` file before reimporting.
    
    To force reimporting materials every time the 3D scene is reimported,
    change the material storage mode in the 3D scene by selecting it in
    the FileSystem dock, then go the Import dock and set
    **Material > Storage** to **Built-In** instead of **Files**.
```

Calinou (approved):
Thanks!

PR #4043, opened 2020-09-20
Mention limit of 8 lights per mesh

aaronfranke:
The introduction to 3D article mentioned unlimited lights, but it did not mention the limit of 8 lights per mesh. This is a hard-coded limit in Godot. I added a mention of it to the article. This was requested by jcmonkey on Discord.

### DIFF
--- a/tutorials/3d/introduction_to_3d.rst
+++ b/tutorials/3d/introduction_to_3d.rst
@@ -222,5 +222,7 @@ each viewport:
 Lights
 ------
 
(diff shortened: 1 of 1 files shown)


Calinou:
This limit is only relevant in 3.2.x, so the pull request should be switched to be made against the `3.2` branch. (Not sure if I can do this, I can't see a button for it.)

akien-mga:
Thanks!

PR #4046, opened 2020-09-20
Fix typo in Enemy Scene section

JulienDuranleau:
Replace the word *instance* for *instantiate* as it is used as a verb in the Enemy Scene section

### DIFF
--- a/getting_started/step_by_step/your_first_game.rst
+++ b/getting_started/step_by_step/your_first_game.rst
@@ -534,7 +534,7 @@ Now it's time to make the enemies our player will have to dodge. Their
 behavior will not be very complex: mobs will spawn randomly at the edges
 of the screen, choose a random direction, and move in a straight line.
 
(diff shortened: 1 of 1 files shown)


cbscribe:
I prefer "instance" here because it can be used as a verb and it's the term used in the engine, where it's the name of the function we will be using. Semantically, function names are verbs because they take action.

Using a different word here reduces clarity.

akien-mga:
I think the use of italics (*instance*) here make it clear that it's referring to the name of the function used as a verb, instead of the arguably more correct "instantiate" verb. So I'd also err on the side of not changing it.

Note that there's a proposal to rename `instance` to `instantiate` in the Godot API, but that's still up for discussion, as it would be a big compatibility breakage for debatable gain.

akien-mga:
That being said, this use of *instance* is not so localization-friendly, so might be worth reconsidering nevertheless. (In non-English languages, we can't use the `instance` method name as a verb, we're forced to use a verb that actually exists in the language.)

PR #4050, opened 2020-09-21
Add MDN reference for Local server setup

rokonxr:
It's going to help some people.
I don't know if the placement or format is right, please edit as needed.


akien-mga:
Thanks!

PR #4052, opened 2020-09-21
Add a warning about `latest` documentation compatibility

Calinou:
Not everything in the `latest` branch applies to Godot 3.2.x. Read the Docs adds a warning notice for old versions automatically, but it doesn't offer an option to do so for the `latest`/unstable branch so we have to add it ourselves.

**Note:** Right now, clicking the link redirects you to the homepage. Is there a way to get the current page URL/ID so we can
redirect to the same page, but with a different version?

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -230,6 +230,21 @@ def godot_get_image_filename_for_language(filename, env):
 
 sphinx.util.i18n.get_image_filename_for_language = godot_get_image_filename_for_language
 
+# Read the Docs adds a note at the top of the page when reading documentation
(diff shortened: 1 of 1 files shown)


akien-mga:
> **Note:** Right now, clicking the link redirects you to the homepage. Is there a way to get the current page URL/ID so we can
> redirect to the same page, but with a different version?

This could likely be done with JS but I don't think we can do this with `rst_prolog`, it's static content.

akien-mga:
Thanks!

PR #4056, opened 2020-09-22
Prevent the compatibility notice from being the `meta description`

Calinou:
Adding the compatibility notice via JavaScript prevents the `meta description` Sphinx extension from taking the notice as part of the description.


akien-mga (approved):

akien-mga:
Thanks!

PR #4058, opened 2020-09-22
Add a Troubleshooting page with common problems and solutions

Calinou:
Feel free to suggest additional issues to list :slightly_smiling_face:

### DIFF
--- a/about/troubleshooting.rst
+++ b/about/troubleshooting.rst
@@ -1,0 +1,85 @@
+.. _doc_troubleshooting:
+
+Troubleshooting
+===============
+
+This page lists common issues encountered when using Godot and possible solutions.
+
+Everything I do in the editor or project manager appears delayed by one frame.
+------------------------------------------------------------------------------
+
+This is a `known bug <https://github.com/godotengine/godot/issues/23069>`__ on
+Intel graphics drivers on Windows. Updating to the latest graphics driver
+version *provided by Intel* should fix the issue.
+
+You should use the graphics driver provided by Intel rather than the one
+provided by your desktop or laptop's manufacturer because their version is often
+outdated.
+
+The grid disappears and meshes turn black when I rotate the 3D camera in the editor.
+------------------------------------------------------------------------------------
+
+This is a `known bug <https://github.com/godotengine/godot/issues/30330>`__ on
+Intel graphics drivers on Windows.
+
+The only workaround, for now, is to switch to the GLES2 renderer. You can switch
+the renderer in the top-right corner of the editor or the Project Settings.
+
+If you use a computer allowing you to switch your graphics card, like NVIDIA
+Optimus, you can use the dedicated graphics card to run Godot.
+
+The editor or project takes a very long time to start.
+------------------------------------------------------
+
+This is a `known bug <https://github.com/godotengine/godot/issues/20566>`__ on
+Windows when you have specific USB peripherals connected. In particular,
+Corsair's iCUE software seems to cause the bug. Try updating your USB
+peripherals' drivers to their latest version. If the bug persists, you need to
+disconnect the faulty peripherals before opening the editor. You can then
+connect the peripheral again.
+
+The Godot editor appears frozen after clicking the system console.
+------------------------------------------------------------------
+
+When running Godot on Windows with the system console enabled, you can
+accidentally enable *selection mode* by clicking inside the command window. This
+Windows-specific behavior pauses the application to let you select text inside
+the system console. Godot cannot override this system-specific behavior.
+
+To solve this, select the system console window and press Enter to leave
+selection mode.
+
+The project window appears blurry, unlike the editor.
+-----------------------------------------------------
+
+Unlike the editor, the project isn't marked as DPI-aware by default. This is
+done to improve performance, especially on integrated graphics, where rendering
+3D scenes in hiDPI is slow.
+
+To resolve this, open **Project > Project Settings** and enable **Display >
+Window > Dpi > Allow Hidpi**. On top of that, make sure your project is
+configured to support :ref:`multiple resolutions <doc_multiple_resolutions>`.
+
+The project window doesn't appear centered when I run the project.
+------------------------------------------------------------------
+
+This is a `known bug <https://github.com/godotengine/godot/issues/13017>`__. To
+resolve this, open **Project > Project Settings** and enable **Display > Window
+> Dpi > Allow Hidpi**. On top of that, make sure your project is configured to
+support :ref:`multiple resolutions <doc_multiple_resolutions>`.
+
+The project works when run from the editor, but fails to load some files when running from an exported copy.
+------------------------------------------------------------------------------------------------------------
+
+This is usually caused by forgetting to specify a filter for non-resource files
+in the Export dialog. By default, Godot will only include actual *resources*
+into the PCK file. Some files commonly used, such as JSON files, are not
+considered resources. For example, if you load ``test.json`` in the exported
+project, you need to specify ``*.json`` in the non-resource export filter. See
+:ref:`doc_exporting_projects_export_mode` for more information.
+
+On Windows, this can also be due to :ref:`case sensitivity
+<doc_project_organization_case_sensitivity>` issues. If you reference a resource
+in your script with a different case than on the filesystem, loading will fail
+once you export the project. This is because the virtual PCK filesystem is
+case-sensitive, while Windows's filesystem is case-insensitive by default.
--- a/about/index.rst
+++ b/about/index.rst
@@ -7,6 +7,7 @@ About
 
    introduction
    faq
+   troubleshooting
    list_of_features
    docs_changelog
    release_policy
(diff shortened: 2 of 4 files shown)


Yogoda:
This one may seem stupid, but I really thought it was a nasty bug before finding the solution: if you click on the console window, Godot will freeze when saving (just before playing executing the project). To fix, click on the console and press "enter" :) Don't know if it is a common problem though...

Calinou:
@Yogoda I remember seeing a few similar reports, so I'll add it to the list.

nathanfranke:
Should we change the page name to be more specific about the types of issues? These seem to be only know editor display/driver issues. If we ever make a page like "Common bugs in your Godot game" for user errors (For example, one of the issues might be "The wrong scene is opening when I start my project")

We can consider changing the name to "Common Editor Issues" or "Known Bugs with the Editor"

about/troubleshooting.rst
```diff
@@ -0,0 +1,84 @@
...
+Everything I do in the editor/project manager appears delayed by one frame.
+---------------------------------------------------------------------------
+
+This is a `known bug <https://github.com/godotengine/godot/issues/23069>`__
+on Intel graphics drivers on Windows. Updating to the latest graphics driver
+version *provided by Intel* should fix the issue.
+
+You should use the graphics driver provided by Intel rather than the one
+provided by your desktop/laptop's OEM. This is because the OEM-supplied version
+is often outdated.
```
NathanLovato, line 17:
```suggestion
Everything I do in the editor or project manager appears delayed by one frame.
------------------------------------------------------------------------------

This is a `known bug <https://github.com/godotengine/godot/issues/23069>`__ on
Intel graphics drivers on Windows. Updating to the latest graphics driver
version *provided by Intel* should fix the issue.

You should use the graphics driver provided by Intel rather than the one
provided by your desktop or laptop's manufacturer because their version is often
outdated.
```

about/troubleshooting.rst
```diff
@@ -0,0 +1,84 @@
...
+The grid disappears and meshes turn black when I rotate the 3D camera in the editor.
+------------------------------------------------------------------------------------
+
+This is a `known bug <https://github.com/godotengine/godot/issues/30330>`__
+on Intel graphics drivers on Windows.
+
+The only workaround for now is to switch to the GLES2 renderer. You can switch
+the renderer in the top-right corner of the editor or in the Project Settings.
+Alternatively, if you use a laptop with switchable graphics (such as NVIDIA
+Optimus), you can use your laptop's dedicated graphics card to run Godot.
```
NathanLovato, line 28:
```suggestion
The grid disappears and meshes turn black when I rotate the 3D camera in the editor.
------------------------------------------------------------------------------------

This is a `known bug <https://github.com/godotengine/godot/issues/30330>`__ on
Intel graphics drivers on Windows.

The only workaround, for now, is to switch to the GLES2 renderer. You can switch
the renderer in the top-right corner of the editor or the Project Settings.

If you use a computer allowing you to switch your graphics card, like NVIDIA
Optimus, you can use the dedicated graphics card to run Godot.
```

about/troubleshooting.rst
```diff
@@ -0,0 +1,84 @@
...
+
+The editor/project takes a very long time to start.
+----------------------------------------------------
+
+This is a `known bug <https://github.com/godotengine/godot/issues/20566>`__ on
+Windows when having specific USB peripherals connected. This issue has been
+reproduced in particular with Corsair's iCUE software. Try updating your USB
+peripherals' drivers to their latest version. If all else fails, disconnect them
+before starting the project then connect them again.
```
NathanLovato, line 37:
```suggestion
The editor or project takes a very long time to start.
------------------------------------------------------

This is a `known bug <https://github.com/godotengine/godot/issues/20566>`__ on
Windows when you have specific USB peripherals connected. In particular,
Corsair's iCUE software seems to cause the bug. Try updating your USB
peripherals' drivers to their latest version. If the bug persists, you need to
disconnect the faulty peripherals before opening the editor. You can then
connect the peripheral again.
```

about/troubleshooting.rst
```diff
@@ -0,0 +1,84 @@
...
+The Godot editor appears to have frozen after clicking the system console.
+--------------------------------------------------------------------------
+
+When running Godot on Windows with the system console enabled, if you click
+inside the command window, you may have accidentally enabled *selection mode*.
+This Windows-specific behavior will pause the application to let you select text
+inside the system console. The application cannot override this behavior.
+
+To solve this, select the system console window and press Enter to leave
+selection mode.
```
NathanLovato, line 48:
```suggestion
The Godot editor appears frozen after clicking the system console.
------------------------------------------------------------------

When running Godot on Windows with the system console enabled, you can
accidentally enable *selection mode* by clicking inside the command window. This
Windows-specific behavior pauses the application to let you select text inside
the system console. Godot cannot override this system-specific behavior.

To solve this, select the system console window and press Enter to leave
selection mode.
```

about/troubleshooting.rst
```diff
@@ -0,0 +1,84 @@
...
+The project window appears blurry (unlike the editor).
+------------------------------------------------------
+
+Unlike the editor, the project isn't marked as DPI-aware by default. This is
+done to improve performance, especially on integrated graphics where rendering
+3D scenes in hiDPI is slow.
+
+To resolve this, open **Project > Project Settings** and enable
+**Display > Window > Dpi > Allow Hidpi**. On top of that, make sure your project
+is configured to support :ref:`multiple resolutions <doc_multiple_resolutions>`.
```
NathanLovato, line 59:
```suggestion
The project window appears blurry, unlike the editor.
-----------------------------------------------------

Unlike the editor, the project isn't marked as DPI-aware by default. This is
done to improve performance, especially on integrated graphics, where rendering
3D scenes in hiDPI is slow.

To resolve this, open **Project > Project Settings** and enable **Display >
Window > Dpi > Allow Hidpi**. On top of that, make sure your project is
configured to support :ref:`multiple resolutions <doc_multiple_resolutions>`.
```

about/troubleshooting.rst
```diff
@@ -0,0 +1,84 @@
...
+
+The project window doesn't appear centered when I run the project.
+------------------------------------------------------------------
+
+This is a `known bug <https://github.com/godotengine/godot/issues/13017>`__.
+To resolve this, open **Project > Project Settings** and enable
+**Display > Window > Dpi > Allow Hidpi**. On top of that, make sure your project
+is configured to support :ref:`multiple resolutions <doc_multiple_resolutions>`.
```
NathanLovato, line 67:
```suggestion
The project window doesn't appear centered when I run the project.
------------------------------------------------------------------

This is a `known bug <https://github.com/godotengine/godot/issues/13017>`__. To
resolve this, open **Project > Project Settings** and enable **Display > Window
> Dpi > Allow Hidpi**. On top of that, make sure your project is configured to
support :ref:`multiple resolutions <doc_multiple_resolutions>`.
```

about/troubleshooting.rst
```diff
@@ -0,0 +1,84 @@
...
+considered resources. For example, if you load ``test.json`` in the exported
+project, you need to specify ``*.json`` in the non-resource export filter.
+See :ref:`doc_exporting_projects_export_mode` for more information.
+
+On Windows, this can also be due to
+:ref:`case sensitivity <doc_project_organization_case_sensitivity>` issues.
+If you reference a resource in your script with different casing than on the
+filesystem, it will fail once the project is exported. This is because the
+virtual PCK filesystem is case-sensitive, unlike the Windows filesystem which is
+case-insensitive by default.
```
NathanLovato, line 84:
```suggestion
The project works when run from the editor, but fails to load some files when running from an exported copy.
------------------------------------------------------------------------------------------------------------

This is usually caused by forgetting to specify a filter for non-resource files
in the Export dialog. By default, Godot will only include actual *resources*
into the PCK file. Some files commonly used, such as JSON files, are not
considered resources. For example, if you load ``test.json`` in the exported
project, you need to specify ``*.json`` in the non-resource export filter. See
:ref:`doc_exporting_projects_export_mode` for more information.

On Windows, this can also be due to :ref:`case sensitivity
<doc_project_organization_case_sensitivity>` issues. If you reference a resource
in your script with a different case than on the filesystem, loading will fail
once you export the project. This is because the virtual PCK filesystem is
case-sensitive, while Windows's filesystem is case-insensitive by default.
```

NathanLovato (changes requested):
Left some suggestions to follow the writing guidelines more closely. GitHub doesn't catch all changes, it's mostly removing slashes, parens, some passive voice, and making the phrasing a little more explicit. E.g. I replaced the acronym OEM by manufacturer for people who don't know what an OEM is.

daviewales:
I was recently bitten by trying to use Layout -> Full Rect on a ColorRect inside a Node2D.
Node2D has no size, so this doesn't work, as described in #18166. Instead, I should have been using a regular Node.

Is this the kind of issue you are looking for with this list?

NathanLovato:
For issues like these, we probably need to document the UI part well. There would probably be room for some troubleshooting just for the UI system.

NathanLovato:
Thanks for your work, as always!

PR #4062, opened 2020-09-24
Removed GD namespace from RandRange

### LINKED ISSUE #4047: Error in Tutorial "Your first game", type conversion
In Main node script, in event _on_MobTimer_timeout() there are 2 similar errors
1. Code `direction += GD.RandRange(-Mathf.Pi / 4, Mathf.Pi / 4);`
Error: Cannot implicitly convert type double to float. An explicit conversiot exists (are you missing a cast?)
2. Code `mobInstance.LinearVelocity = new Vector2(GD.RandRange(150, 250),0).Rotated(direction);`
Error: Argument 1: cannot convert from double to float

My Godot version is at 3.1.2

greenmoggle:
`GD.RandRange` was returning a double where a float was expected causing a type error. Using the `RandRange` function as described in the tutorial fixes the error.

*Bugsquad edit:* Fixes #4047.

### DIFF
--- a/getting_started/step_by_step/your_first_game.rst
+++ b/getting_started/step_by_step/your_first_game.rst
@@ -898,11 +898,11 @@ Note that a new instance must be added to the scene using ``add_child()``.
         mobInstance.Position = mobSpawnLocation.Position;
 
         // Add some randomness to the direction.
-        direction += GD.RandRange(-Mathf.Pi / 4, Mathf.Pi / 4);
+        direction += RandRange(-Mathf.Pi / 4, Mathf.Pi / 4);
         mobInstance.Rotation = direction;
 
         // Choose the velocity.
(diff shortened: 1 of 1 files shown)


neikeq:
I don't see the difference with the new code. If anything it won't compile unless you're `using static GD;` to not need to specify `GD.`. If I understand the problem correctly, the solution would be to cast the `GD.RandRange` return to float.

greenmoggle:
It's because earlier in the tutorial you create this function
`private float RandRange(float min, float max)
{
return (float)_random.NextDouble() * (max - min) + min;
}`
Though mistakenly the GD.RandRange was used which caused a type error.

neikeq:
Ah! I see now. Makes sense.

neikeq (approved):

akien-mga:
Thanks!

PR #4068, opened 2020-09-25
Escape double quotes for meta descriptions in HTML

Calinou:
If this not done, some descriptions will be closed earlier than intended.


akien-mga:
Thanks!

PR #4069, opened 2020-09-25
[Example] Development philosophy

Xrayez:
Resolves godotengine/godot-proposals#575.

The documentation lacks a thorough description on general principles which govern Godot development, which would act as introductory material for potential contributors that would like to get familiar with the development aspects that start with "Why?" rather than "How?" and "What?", as with the existing "Best practices for engine contributors" page.

This is different to "Design philosophy" as well because this is mainly targeted for people who express interest in contributing as potential engine developers, rather than developers that simply use the engine.

## Why?

1. Defining the scope of a particular project is always considered a good software engineering practice.
2. The scope can more or less clearly determine what features are actually needed for Godot development.
3. Defining goals and non-goals helps to prevent GIP to overflow with proposals which are nice-to-have rather than need-to-have, as described in the GIP rules and guidelines, and instead channel this to other community maintained projects that use Godot for modules and plugin development, where discussions could have a more productive outcomes. Examples of non-productive proposals in GIP: godotengine/godot-proposals#100.

I've moved the previously added "convenience over absolute performance" principle added in #3788 in this PR, so you see what I've been talking about when saying that "Best practices for engine contributors" does not provide answers to these kind of questions.

The information is taken from various sources such as news articles, reddit responses, and even IRC discussions (mostly @reduz opinions as a project founder), and of course likely includes some of my own subjective vision and/or wishful thinking. 😉

Also, this *actually* links the GIP repository directly in the documentation, as a main tool for evaluating what is deemed useful for Godot development, I haven't found any other link to GIP throughout Godot's documentation.

### Disclaimer

This is just an example, feel free to modify and build upon this to better reflect the development philosophy in a way you see it, and add other useful information.

PR #4071, opened 2020-09-26
Clarify "The problem has to exist" in the best practices

Xrayez:
Simply saying that "The problem has to exist" from the start may convey a wrong subconscious message that problems must exist in the engine for the sake of it, especially if taken out of context, so it's better to make this more explicit to mean that "problems can only be solved when they actually exist" up front.

I've also took the opportunity to potentially improve grammar and style as suggested by the http://www.hemingwayapp.com/.

---

<details>
<summary>
To summarize, developers should strive to solve problems, not to create them. 🙂
</summary>

the_general_problem

https://xkcd.com/974/
</details>

### DIFF
--- a/community/contributing/best_practices_for_engine_contributors.rst
+++ b/community/contributing/best_practices_for_engine_contributors.rst
@@ -46,20 +46,19 @@ as source and binary and requires maintenance once it exists. Avoiding the
 addition of anything unnecessary is always considered a good practice in
 software development.
 
-#2: The problem has to exist
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#2: To solve the problem, it has to exist in the first place
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-This is a variation of the previous practice. I believe most developers agree
-that adding anything unnecessary is not a good idea, but what constitutes what
-is necessary and what isn't?
+This is a variation of the previous practice. Adding anything unnecessary is not
+a good idea, but what constitutes what is necessary and what isn't?
 
 .. image:: img/best_practices2.png
 
-The answer to this question is that the problem needs to *exist*. It must not be
-speculation or a belief. The user must be using the software as intended to
-create something they *need*. In this process, the user may stumble into a
-problem that requires a solution in order to continue, or in order to achieve
-greater productivity. In this case, *a solution is needed*.
+The answer to this question is that the problem needs to *exist* before it can
+be actually solved. It must not be speculation or a belief. The user must be
+using the software as intended to create something they *need*. In this process,
+the user may stumble into a problem that requires a solution to proceed, or in
+order to achieve greater productivity. In this case, *a solution is needed*.
 
 Believing that problems may arise in the future and that the software needs to
 be ready to solve them by the time they appear is called *"Future proofing"* and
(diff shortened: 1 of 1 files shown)


community/contributing/best_practices_for_engine_contributors.rst
```diff
@@ -46,20 +46,19 @@ as source and binary and requires maintenance once it exists. Avoiding the
...
-#2: The problem has to exist
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#2: To solve the problem, it has to exist in the first place
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-This is a variation of the previous practice. I believe most developers agree
-that adding anything unnecessary is not a good idea, but what constitutes what
-is necessary and what isn't?
+This is a variation of the previous practice. Adding anything unnecessary is not
+a good idea, but what constitutes what is necessary and what isn't?
```
Xrayez, line 53: This removes personal opinion ("I" vs "We").

Can also be worded as:

```suggestion
This is a variation of the previous practice. It's safe to say that adding
anything unnecessary is not a good idea, but what constitutes what is necessary
and what isn't?
```

Xrayez:
I also believe that problems *always* exist, and it takes a lot of experience and mastery in order to be able to solve problems that don't *appear* in the present moment yet, since the space of all possible problems which can *appear* is not restricted to the present moment, yet it's equally difficult to foresee those kind of problems for most people who would be interested in contributing to the engine development, so as a community of international developers, we're forced to agree upon more or less objective grounds, and the existence of problems that appear in the present moment is deemed quite objective for most people and simplifies the process of reaching the consensus as a community-driven project.

If you think that what I've just wrote makes sense, then I'd be happy to add this here as well.

In fact, these kind of clarifications could be added to the proposed development philosophy in #4069, which would simply *acknowledge* (but not *recommend*) those kind of ideas behind Godot development without outright omitting them, since one of the reason behind writing those best practices is *exactly* inexperience and misunderstanding of those development processes. Lets make this more than a thing which is used as a justification for rejecting proposals.

I hope you understand what I'm saying with this. 🙂

NathanLovato (approved):

NathanLovato:
Thanks for your contribution!

community/contributing/best_practices_for_engine_contributors.rst
```diff
@@ -69,9 +68,9 @@ its characterized by lines of thought such as:
 - I think users will eventually need to...
 
 This is generally considered a bad habit, because trying to solve problems that
-*don't actually exist* in the present will very often lead to code that will be
-written but never used, or to code that is considerably more complex to use and
-maintain than it needs to be.
+*don't actually exist* in the present will very often lead to the code that will
+be written but never used, or the code that is considerably more complex to use
```
akien-mga, line 72: The sentence seems weird to me with the added "the" before code.

"lead to (written, but unused) code", not "lead to the written but unused code" (which code is "the code"? It's generic).

Xrayez: In the previous "to code that is considerably more complex", "code" becomes a verb, so that was quite confusing for me. I'm not particularly strong with those English grammatical articles such as `the` or `a` (they don't exist in my language at all), but according to definition of `the` I've googled:

> denoting one or more people or things **already mentioned or assumed to be common knowledge** [emphasis mine].

Best practices could be considered common knowledge so that's why I added those, "*the* code which is...", rather than "to code", because "to" was already there in "will very often lead to".

akien-mga: > In the previous "to code that is considerably more complex", "code" becomes a verb, so that was quite confusing for me. 

I don't think that's correct.

See https://dictionary.cambridge.org/dictionary/essential-british-english/lead-to-something

If code were used as a verb here, I think the correct phrasing would be "lead to coding".
Here, "lead to code" means "lead to (the presence of) code".

CC @cbscribe

NathanLovato: I missed it, code is still a noun in that case, even without the article. Note the page could still use a whole editing pass to simplify the language, like a fair part of the best practices series. The style is a bit heavy.

PR #4072, opened 2020-09-26
[Docs] Clarify type of Main Node in tutorial.

daviewales:
If a user follows the tutorial, and accidentally [creates the Main scene][1] using a Node2D instead of a basic Node, they will be unable to [add a background][2] with Layout -> Full Rect as instructed later in the tutorial.

This expected (although possibly confusing) behaviour for Node2D is described in [#18166][3].

I encountered this personally while working through the tutorial. Because I am not familiar with the node types available in Godot, and because the node search function scrolls to Node2D when searching for 'Node', I completed the tutorial using a Node2D as my Main node, then encountered the issue where I couldn't use Layout -> Full Rect to set the size of the background colour.

[1]: https://docs.godotengine.org/en/stable/getting_started/step_by_step/your_first_game.html#main-scene
[2]: https://docs.godotengine.org/en/stable/getting_started/step_by_step/your_first_game.html#background
[3]: https://github.com/godotengine/godot/issues/18166

### DIFF
--- a/getting_started/step_by_step/your_first_game.rst
+++ b/getting_started/step_by_step/your_first_game.rst
@@ -669,7 +669,7 @@ Main scene
 ----------
 
 Now it's time to bring it all together. Create a new scene and add a
-:ref:`Node <class_Node>` named ``Main``. Click the "Instance" button and select your
+:ref:`Node <class_Node>` named ``Main``. This should be a basic Node, NOT a Node2D. Click the "Instance" button and select your
 saved ``Player.tscn``.
 
 .. image:: img/instance_scene.png
--- a/getting_started/step_by_step/your_first_game.rst
+++ b/getting_started/step_by_step/your_first_game.rst
@@ -669,7 +669,8 @@ Main scene
 ----------
 
 Now it's time to bring it all together. Create a new scene and add a
-:ref:`Node <class_Node>` named ``Main``. This should be a basic Node, NOT a Node2D. Click the "Instance" button and select your
+:ref:`Node <class_Node>` named ``Main``. Ensure you create a Node, **not** a
+Node2D. Click the "Instance" button and select your
 saved ``Player.tscn``.
 
 .. image:: img/instance_scene.png


Xrayez:
On a side note, do you think having a quick access to `Node` in the "Create Root Node" panel could make this more intuitive for you? See godotengine/godot-proposals#1206.

getting_started/step_by_step/your_first_game.rst
```diff
@@ -669,7 +669,7 @@ Main scene
 ----------
 
 Now it's time to bring it all together. Create a new scene and add a
-:ref:`Node <class_Node>` named ``Main``. Click the "Instance" button and select your
+:ref:`Node <class_Node>` named ``Main``. This should be a basic Node, NOT a Node2D. Click the "Instance" button and select your
```
NathanLovato, line 672:
```suggestion
:ref:`Node <class_Node>` named ``Main``. Ensure you create a Node, **not** a
Node2D. Click the "Instance" button and select your
```

NathanLovato (changes requested):
A small edit to be a bit more consistent with the docs' style.

daviewales:
> > because the node search function scrolls to Node2D when searching for 'Node'
> 
> On a side note, do you think having a quick access to `Node` in the "Create Root Node" panel could make this whole process more intuitive in some way? See [godotengine/godot-proposals#1206](https://github.com/godotengine/godot-proposals/issues/1206).

I'm not sure that the linked proposal would directly address the issue I encountered, as the Create Root Node panel calls the nodes 'scenes' rather than their actual name, which is confusing. e.g. The proposal calls it 'Scene' and '2D scene' rather than 'Node' and 'Node2D', which means I still wouldn't have made the connection from the tutorial.

If the Create Root Node panel used the actual node names it would help.
Also, if adding a custom node and searching 'node' didn't scroll past 'Node' to 'Node2D' that would also help!

NathanLovato:
Just merged, thanks for your contribution

PR #4074, opened 2020-09-27
Remove outdated tutorials from the Godot 4.0 documentation

### LINKED ISSUE #1412: Suggest documenting a gotcha: Godot FPS tutorial framerate is really slow
If you compile Godot with `platform=x11` and omit `target`, Godot's FPS tutorial becomes inexplicably slow when you load the space level in part 3.  This had me wondering what was wrong for awhile, especially because my script profiler was showing zero time.

Eventually someone on Discord helpfully mentioned that I was not compiling an optimized Godot target. Launching this scene on my 2.8GHz laptop had me starting to get 1fps.

I recommend adding a warning to the documentation to ensure that an optimized build is being used when launching this level because a modern highend laptop gets 1fps otherwise!

cc @TwistedTwigleg

### LINKED ISSUE #1599: Explain why FPS camera is rotated
From a note in the [movement section of the FPS tutorial part 1](http://docs.godotengine.org/en/3.0/tutorials/3d/fps_tutorial/part_one.html#making-the-fps-movement-logic):

> Because the camera is rotated by -180 degrees, we have to flip the Z directional vector. Normally forward would be the positive Z axis, so using basis.z.normalized() would work, but we are using -basis.z.normalized() because our camera’s Z axis faces backwards in relation to the rest of the player.

This sounds pretty confusing to a newbie. This very text admits there is a simpler way to do it ("Normally forward would be the positive Z axis"), so why on earth does the tutorial choose this confusing camera rotation?

I'm guessing there is a good reason, so an extra sentence or so explaining **why** would be great.

### LINKED ISSUE #1601: Is current_state needed in the FPS tutorial part 2?
The script in http://docs.godotengine.org/en/3.0/tutorials/3d/fps_tutorial/part_two.html#making-a-system-to-handle-animations manages a `current_state` variable with the name of the current animation.

Is there any reason not to use [`AnimationPlayer.current_animation`](http://docs.godotengine.org/en/3.0/classes/class_animationplayer.html#class-animationplayer-current-animation)?

### LINKED ISSUE #1620: Language in godot-docs/tutorials/3d/fps_tutorial/part_two.rst is unclear
Hi,
I believe the language in parts of this tutorial (which is really really great and helps me a lot), is flawed.
In part 2, the paragraph about the state machines  (function set_animation() ), is the following:

"Lets look at set_animation next.

set_animation sets the animation to the that of the passed in animation state if we can transition to it. In other words, if the animation state we are currently in has the passed in animation state name in states, then we will change to that animation.

To start we check if the passed in animation is the same as the animation state we are currently in. If they are the same, then we write a warning to the console and return true.

Next we see if AnimationPlayer has the passed in animation using has_animation. If it does not, we return false.

Then we check if current_state is set or not. If current_state is not currently set, we set current_state to the passed in animation and tell AnimationPlayer to start playing the animation with a blend time of -1 and at the speed set in animation_speeds and then we return true.

If we have a state in current_state, then we get all of the possible states we can transition to. If the animation name is in the list of possible transitions, we set current_state to the passed in animation, tell AnimationPlayer to play the animation with a blend time of -1 at the speed set in animation_speeds and return true."

" to the that of the passed in animation state" - what is that supposed to mean?
"To start we check if the passed in animation" - same

I would be glad for an explanation and can help with translations if you like.

Best

### LINKED ISSUE #1667: FPS Tutorial part 2 - redundant check in snippets
In the unequip_weapon() methods presented for both the pistol and rifle, "player_node.animation_manager.current_state" is checked in two nested if statements, once that it is equal to a value and next that it is not equal to a different value. If it is equal to the first value, it can not be equal to the second, so the check that follows shouldn't be needed. 

Note: The knife's snippet does not have this double check.

### LINKED ISSUE #1670: FPS Tutorial: Bullet material causes pause on first instancing
On my system, when you first instance a bullet from the pistol, there is a ~1 second pause. Afterwards there is no problem. It confused me at first, because the bullet scene is preloaded, so instancing it the first time shouldn't cause so much delay. So I suspected it had to do with shader compiling.

I figured out that the bullet material "GunMaterial.material" was causing the delay. It has transparency enabled and the depth draw mode is "Opaque Pre-Pass". So, since it is also being used elsewhere, I copied it to a separate material for the bullet, disabled transparency, and set the depth draw mode to "Opaque Only", and this fixed the pausing. Note: I just wasn't sure if other things using the material needed the transparency, so maybe it didn't need to be copied.

Reverting to the old material, I tried it with the shader cache enabled. After a warmup run, the initial pause was much smaller, losing about 1 frame, but it is still noticeable if you are moving.

The easy solution here is just to change the material settings, but I wonder if there is a better way to handle this. Ideally there would be a way to preload the shaders for things without having to show anything to the user.

About my setup:
Graphics type: Integrated
Graphics chip: Intel HD Graphics 5500
Graphics driver: Mesa 18.0.5
Godot version: 3.0.5 64-bit X11
OS: Xubuntu 16.04

### LINKED ISSUE #2346: Control node tutorial does not explain Size Flag options
The documentation on Size Flags needs a bit of love, especially as it's introduced as part of the Godot "Getting Started" tutorial. Whilst there are some image examples, they don't explain what is actually happening.

Taken from: https://github.com/godotengine/godot-docs/blob/3.1/getting_started/step_by_step/ui_introduction_to_the_ui_system.rst

> If you add the "Fill" flag to the Horizontal or Vertical property, the node's bounding box will take all the space it can, but it'll respect its siblings and retain its size.

The wording there is rather confusing. How can something "take all the space it can" but still "retain its size"?

> You'll need some practice to understand the size tags, as their effect can change quite a bit depending on how you set up your interface.

We should probably strive for better documentation than "go try it yourself".

Size Flags are better documented in the Control reference itself (see: https://github.com/godotengine/godot-docs/blob/master/classes/class_control.rst ), though I still think even there it's not quite clear enough. I think a quick fix for that would be to update the tutorial above, and link back to it from the reference.

### LINKED ISSUE #2574: Following GUI tutorial steps not matching results
https://docs.godotengine.org/en/3.1/getting_started/step_by_step/ui_game_user_interface.html#add-the-progress-bar

Adding the TextureProgress node makes the Title and Number nodes to no longer stay centered and don't lead to the result shown in the tutorial:

image

[edit] The next steps also lead to different results.
Some more inconsistencies - https://docs.godotengine.org/en/3.1/_images/ui_gui_step_tutorial_counter_design_4.png

This shows a CenterContainer but nowhere does the tutorial say anything about adding such a node or what it's supposed to do.

The youtube video linked also approaches the whole thing way differently, not sure which one to take as the most recent / accurate way to create GUI's.

### LINKED ISSUE #2875: Unclear instructions in "Control the game’s UI with code"
In the [Update health with a signal when the player takes a hit](https://docs.godotengine.org/en/stable/getting_started/step_by_step/ui_code_a_life_bar.html#update-health-with-a-signal-when-the-player-takes-a-hit) whenever it instructs the reader to add code to the GUI.gd file it provides an example directly below the paragraph of what the code should look like in both GDScript and C#. There's one paragraph that's missing the accompanying code snippet:

> Inside _on_Player_health_changed, let’s call a second function called update_health and pass it the player_health variable.

I wouldn't normally point something like this out but it's a single line buried between a screenshot from another script and a Note section and it's having the reader call a function that hasn't been declared yet. It's easy to miss and the docs have a habit of blurring the line between "we're going to be doing this next" and "I want you to do this right now."

Another issue is a couple paragraphs later the reader is instructed to add the update_health function to _ready to initialize the Number node's text but this causes an error because update_health is called with too few arguments. I'm assuming it's supposed to be called with player_max_health

>Also call update_health at the end of the _ready function to initialize the Number node’s text with the right value at the start of the game. Press F5 to test the game: the life bar updates with every attack!

Some nitpicky stuff coming now: at the top of this same section it tells you to open the Node panel of Player.gd. The screenshot shows there are no connections but if you download and import the project it comes with a connection from the died() signal to the Enemy script. You can see this connection in a later screenshot in the [Fade the bar when the Player dies](https://docs.godotengine.org/en/stable/getting_started/step_by_step/ui_code_a_life_bar.html#fade-the-bar-when-the-player-dies) section.

In [Animate the loss of life with the Tween node](https://docs.godotengine.org/en/stable/getting_started/step_by_step/ui_code_a_life_bar.html#animate-the-loss-of-life-with-the-tween-node) the reader is instructed to add an "animated_health" variable to the top of the script and set its value to 0. The other variables in this script are proceeded by "onready" and this is something the reader wouldn't have encountered before. It'd be nice if the docs mention what this is and whether or not it's needed for the animated_health variable

### LINKED ISSUE #3151: FPS tutorial maps broken in 3.2

### LINKED ISSUE #3217: FPS Tutorial gun mesh broken
image
image

Seems like some faces are wrong in the assets ?

using:
Godot Engine v3.2.stable.mono.official - https://godotengine.org
OpenGL ES 3.0 Renderer: GeForce RTX 2080/PCIe/SSE2

### LINKED ISSUE #3348: Some "step by step" UI images outdated
Some images in the step by step guide are outdated. I would like to recreate these with the current 3.2 version (nothing more, no content changes).
I could check all images and make a PR if this is a wanted update.

### LINKED ISSUE #3751: move_and_slide has incorrect parameter type in 3d fps tutorial
This is about the stable documentation, and applies to at least the 3.2 branch of Godot

In the sample code for moving the player around in the 3d fps tutorial there is the call to move_and_slide:

    vel = move_and_slide(vel, Vector3(0, 1, 0), 0.05, 4, deg2rad(MAX_SLOPE_ANGLE))

The third argument here is a float, but according to the reference (in the same version of the manual) and the sourcecode for the 3.2 branch the third argument should be a boolean.

The code still works (I guess there is an implicit typecast somewhere), but it makes it more confusing to understand move_and_slide

https://docs.godotengine.org/en/stable/tutorials/3d/fps_tutorial/part_one.html#making-the-fps-movement-logic

### LINKED ISSUE #4011: reference sibling objects in _ready function
**Your Godot version:**
v3.2.2.stable.official

**Issue description:**
Hi I'm wondering if this part of the tutorial is correct
```
    When you open a scene in the game, Godot creates nodes one by one,
    following the order in your Scene dock, from top to bottom. GUI and
    Player are not part of the same node branch. To make sure they both
    exist when we access each other, we have to use the _ready function.
    Godot calls _ready right after it loaded all nodes, before the game
    starts. It's the perfect function to set everything up and prepare the
    game session.
```
in the tutorial the function _ready of the GUI node gets a reference of the sibling Player. Furthermore the GUI node is above Player node in the tree.
Using _ready function of a node may not guarantee the existence of its siblings and the call get_node may result in an error.
Shouldn't _ready be used in a common parent?
 
[Set up the Lifebar with the Player's max_health](https://docs.godotengine.org/en/stable/getting_started/step_by_step/ui_code_a_life_bar.html#set-up-the-lifebar-with-the-player-s-max-health)

NathanLovato:
Removed:

- Step-by-step -> UI tutorials
- Step-by-step -> splash screen tutorial
- 3D -> FPS tutorial

Closes #1412
Closes #1599
Closes #1601
Closes #1620
Closes #1667
Closes #1670
Closes #2346
Closes #2574
Closes #2875
Closes #3151
Closes #3217
Closes #3348
Closes #3751
Closes #4011


NathanLovato:
It won't build for now because a page in the class reference links to a page being removed. See godotengine/godot#42372

akien-mga:
I pushed an amend with `classes/class_animationplayer.rst` edited manually. I still don't want to sync the whole classref with `master` yet as it will require changes to most pages to account for node renames, which would make merging changes in `3.2` a hassle.

akien-mga:
Also had to do this change:
```
diff --git a/tutorials/animation/cutout_animation.rst b/tutorials/animation/cutout_animation.rst
index 93c158e6..f14aa42f 100644
--- a/tutorials/animation/cutout_animation.rst
+++ b/tutorials/animation/cutout_animation.rst
@@ -265,7 +265,7 @@ Animation tips
 
 The following section will be a collection of tips for creating animation for
 your cutout rigs. For more information on how the animation system in Godot
-works, see :ref:`doc_animations`.
+works, see :ref:`doc_introduction_2d_animation`.
 
 Setting keyframes and excluding properties
 ------------------------------------------
```

akien-mga:
Thanks!

PR #4095, opened 2020-09-30
Fix markup in the GDScript style guide

hoxu:


Calinou (approved):
Thanks!

getting_started/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -756,6 +756,7 @@ Inferred types
 In most cases you can let the compiler infer the type, using ``:=``:
 
 ::
+
    var health := 0  # The compiler will use the int type.
```
Calinou, line 760:
```suggestion
In most cases you can let the compiler infer the type, using ``:=``::

    var health := 0  # The compiler will use the int type.
```

PR #4099, opened 2020-10-01
Improve grammar in script templates doc

rcorre:
- Remove duplicate `default`
- "Both per-project and the editor" didn't make sense


rcorre:
Many of the RST syntactical elements like `::note` and `:ref:` seem broken in this page too, but I couldn't figure out why.

NathanLovato:
Thank you :)

PR #4100, opened 2020-10-01
Remove needless note about signals and classes.

rcorre:
As far as I know, any user-defined class must extend Object or a
subclass thereof, meaning it would be impossible to define a class that
couldn't use signals. This warning is therefore unnecessary.

Please let me know if I misinterpreted this.
It seemed as easy to open a PR as an issue asking.


NathanLovato:
Yup, that's correct! If I remember correctly the note was because beginners didn't understand that everything is an Object in Godot, and so that signals are available on a plain Object or Reference.

NathanLovato:
Thanks!

PR #4104, opened 2020-10-01
Add options per compile method for .app bundle creation on Mac OSX

remtaine:
In the Compiling for MacOS section, the terminal commands (the directories particularly) listed for creating an .app bundle were only for Universal 2 binaries. I added the commands for the Intel and the ARM64 compile options as well, for smoother compiling.

### DIFF
--- a/development/compiling/compiling_for_osx.rst
+++ b/development/compiling/compiling_for_osx.rst
@@ -59,11 +59,26 @@ manager.
 
 To create an ``.app`` bundle like in the official builds, you need to use the
 template located in ``misc/dist/osx_tools.app``. Typically, for an optimized
-editor binary built with ``target=release_debug``::
+editor binary built with ``target=release_debug``, you may do this::
 
+Start first with::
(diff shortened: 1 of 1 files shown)


efigi:
In the Compiling for MacOS section, the terminal commands (the directories particularly) listed for creating an .app bundle were only for Universal 2 binaries. I added the commands for the Intel and the ARM64 compile options as well, for smoother compiling.

Calinou (changes requested):
We can include a command that works with all kinds of binaries using a glob:

```
cp bin/godot.osx.*tools Godot.app/Contents/MacOS/Godot
```

The above command also doesn't assume that the user is building with `target=debug` instead of `target=release_debug`.

akien-mga:
As it becomes increasingly a requirement to support both architectures, I don't think we need to spell everything out like this. On the other hand using a globbing pattern as suggested by @Calinou also seems problematic if you did build multiple binaries.

Overall I think this page is due a good rewrite for 4.0 as a lot of things are changing.

PR #4106, opened 2020-10-01
Document how to export a clickable action in the Inspector

Calinou:
This trick can be useful for bespoke editor tooling and is found in plugins such as Qodot.

### DIFF
--- a/getting_started/scripting/gdscript/gdscript_exports.rst
+++ b/getting_started/scripting/gdscript/gdscript_exports.rst
@@ -204,6 +204,40 @@ from the FileSystem dock at once.
 
     var c = [a, 2, 3]
 
+Exporting actions as "buttons"
+------------------------------
+
+Godot currently doesn't expose a way to expose a clickable action within the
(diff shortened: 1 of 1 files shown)


NathanLovato:
It's a hack so I'm not in favor of including it in the reference manual. You're teaching people to make a checkbox you can never check and that does non-checkbox stuff. It's also not a supported use of the inspector. I'd leave tricks like these for a community wiki.

NathanLovato:
@akien-mga could you check this PR to see if it should be included in the docs?

akien-mga:
Yeah I'm also not convinced about adding this in the GDScript exports page. It's an interesting tip/hack but not really something that I think should be part of the Getting Started documentation.

We don't really have a place for tips and tricks in the current documentation. Maybe this would be good in the Q&A (possibly a self-answered question) which could have the relevant tag to be linked in the GDScript exports page once the Q&A integration work is deployed?

Calinou:
Closing in favor of https://github.com/godotengine/godot-proposals/issues/2149.

PR #4187, opened 2020-10-02
Fix broken formatting in the beginning of What is GDNative?

Calinou:
The formatting made it appear like this:

image


NathanLovato:
The tooling I use for ReST is a bit broken, I've noticed it doesn't fill correctly in some cases. Thanks

PR #4190, opened 2020-10-02
Your first game: Add "Play Scene" keyboard shortcut for MacOS

bbkane:
Add "Play Scene" keyboard shortcut for MacOS


Calinou (approved):
Thanks!

PR #4199, opened 2020-10-04
Use MinGW-w64 package available in Arch Linux's official repository

Jayanth-L:
Mingw-w64 package is officially available in the arch linux repository and people can install mingw-w64 directly from official repository.


Calinou (approved):
Thanks!

PR #4200, opened 2020-10-04
Fix name of saved main scene

bwDialogs:
Changed main scene name from `Player.tscn` to `Main.tscn`


NathanLovato:
The original text is the correct one: the instruction is to add an instance of Player.tscn as a child of the Main node.

PR #4040, opened 2020-09-19
Add a page about 3D collision shapes

Calinou:

Issue #530, opened 2017-10-15
CollisionObjects Overview


cbscribe:
Overview of the CollisionObject(2D) types: Area(2D), StaticBody(2D), RigidBody(2D), KinematicBody(2D). Goal is to help beginners understand when and why to choose a particular class.

- Definitions
- Use cases for each body type
- Using collision shapes
- Brief code examples

NathanLovato:
Ping @Calinou as you wrote the 3D page, and I believe you mentioned a 2D page was planned too?

Calinou:
@NathanLovato Yes, I'm currently working on a page for 2D collision shapes :slightly_smiling_face:

NathanLovato:
I assigned you then if you're on it.

Calinou:
This was addressed by https://github.com/godotengine/godot-docs/pull/4040 and https://github.com/godotengine/godot-docs/pull/4203, closing.

PR #4203, opened 2020-10-04
Add a page about 2D collision shapes

Calinou:
Follow-up to #4040. See #530.

**Note:** Some of the performance considerations or limitations mentioned may be incorrect for 2D physics, please check. Also, we might want to remove the section about convex collision shapes since there's no way to create them in the 2D editor.


AndreaCatania (approved):
It looks good and the performance considerations are also valid for 2D.

NathanLovato:
Just read it, great job.

PR #4205, opened 2020-10-05
Updated Canvas Layer Doc

rakibulhaq:
passive sentence conveys more clear explanation, may be passive fom was intended, but missed

### DIFF
--- a/tutorials/2d/canvas_layers.rst
+++ b/tutorials/2d/canvas_layers.rst
@@ -40,8 +40,8 @@ CanvasLayers
 
 The answer is :ref:`CanvasLayer <class_CanvasLayer>`,
 which is a node that adds a separate 2D rendering layer for all its
-children and grand-children. Viewport children will draw by default at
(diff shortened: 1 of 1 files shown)


NathanLovato:
Changing the voice doesn't make much difference in meaning here, although I agree it's weird to have the rest of the paragraph in passive and this sentence with the direct voice. But reading the whole paragraph and the page. It needs a rewrite: the explanations are not as clear as they could be.

One of our writing guidelines is to favor the direct voice over passive, so we shouldn't move towards having a page all passive. Would you be willing to do something like that?

rakibulhaq:
understood, if it needs a rewrite, then this minor thing can be omitted as the updated doc would have a better explanation. 

i suggested the change as I thought: that would be more accurate. and the wording is important for me to grasp the full picture more quickly. I may read the writing guidelines to understand better, but "so we shouldn't move towards having a page all passive" for this sake, we would keep one sentence active, that is not right.

anyway, i love the documentation! Please keep up this great work! I want to build a game using Godot.

PR #4208, opened 2020-10-06
Change Dodge the Creeps to be more consistent with the demo projects

aaronfranke:
Over time the docs and demos for Dodge the Creeps have drifted apart, so now I'm making an effort to unify them. With this PR, all the code examples are the same as the demo as of https://github.com/godotengine/godot-demo-projects/pull/531. This includes many bug fixes, enhancements, style changes, usage of best practices, and large C#-specific changes.

I haven't reviewed the images yet (most are probably fine), but I think this is all (or almost all) of the code changes.

### DIFF
--- a/getting_started/step_by_step/your_first_game.rst
+++ b/getting_started/step_by_step/your_first_game.rst
@@ -176,17 +176,17 @@ Start by declaring the member variables this object will need:
 
     extends Area2D
 
-    export var speed = 400  # How fast the player will move (pixels/sec).
-    var screen_size  # Size of the game window.
+    export var speed = 400 # How fast the player will move (pixels/sec).
+    var screen_size # Size of the game window.
 
  .. code-tab:: csharp
 
     public class Player : Area2D
     {
         [Export]
-        public int Speed = 400; // How fast the player will move (pixels/sec).
+        public int speed = 400; // How fast the player will move (pixels/sec).
 
-        private Vector2 _screenSize; // Size of the game window.
+        public Vector2 screenSize; // Size of the game window.
     }
 
 
@@ -213,12 +213,14 @@ which is a good time to find the size of the game window:
 
     func _ready():
         screen_size = get_viewport_rect().size
+        hide() # Start the player hidden.
 
  .. code-tab:: csharp
 
     public override void _Ready()
     {
-        _screenSize = GetViewport().Size;
+        screenSize = GetViewportRect().Size;
+        Hide(); // Start the player hidden.
     }
 
 Now we can use the ``_process()`` function to define what the player will do.
@@ -233,9 +235,9 @@ need to do the following:
 First, we need to check for input - is the player pressing a key? For
 this game, we have 4 direction inputs to check. Input actions are defined
 in the Project Settings under "Input Map". Here, you can define custom events and
-assign different keys, mouse events, or other inputs to them. For this demo,
-we will use the default events that are assigned to the arrow keys on the
-keyboard.
+assign different keys, mouse events, or other inputs to them.
+For this game, we will just use the default events called
+"ui_right" etc that are assigned to the arrow keys on the keyboard.
 
 You can detect whether a key is pressed using
 ``Input.is_action_pressed()``, which returns ``true`` if it's pressed
@@ -245,7 +247,7 @@ or ``false`` if it isn't.
  .. code-tab:: gdscript GDScript
 
     func _process(delta):
-        var velocity = Vector2()  # The player's movement vector.
+        var velocity = Vector2.ZERO # The player's movement vector.
         if Input.is_action_pressed("ui_right"):
             velocity.x += 1
         if Input.is_action_pressed("ui_left"):
@@ -254,6 +256,7 @@ or ``false`` if it isn't.
             velocity.y += 1
         if Input.is_action_pressed("ui_up"):
             velocity.y -= 1
+
         if velocity.length() > 0:
             velocity = velocity.normalized() * speed
             $AnimatedSprite.play()
@@ -264,7 +267,7 @@ or ``false`` if it isn't.
 
     public override void _Process(float delta)
     {
-        var velocity = new Vector2(); // The player's movement vector.
+        var velocity = Vector2.Zero; // The player's movement vector.
 
         if (Input.IsActionPressed("ui_right"))
         {
@@ -290,7 +293,7 @@ or ``false`` if it isn't.
 
         if (velocity.Length() > 0)
         {
-            velocity = velocity.Normalized() * Speed;
+            velocity = velocity.Normalized() * speed;
             animatedSprite.Play();
         }
         else
@@ -339,8 +342,8 @@ the ``_process`` function (make sure it's not indented under the `else`):
 
         Position += velocity * delta;
         Position = new Vector2(
-            x: Mathf.Clamp(Position.x, 0, _screenSize.x),
-            y: Mathf.Clamp(Position.y, 0, _screenSize.y)
+            x: Mathf.Clamp(Position.x, 0, screenSize.x),
+            y: Mathf.Clamp(Position.y, 0, screenSize.y)
         );
 
 
@@ -377,7 +380,7 @@ function:
         if velocity.x != 0:
             $AnimatedSprite.animation = "walk"
             $AnimatedSprite.flip_v = false
-            # See the note below about boolean assignment
+            # See the note below about boolean assignment.
             $AnimatedSprite.flip_h = velocity.x < 0
         elif velocity.y != 0:
             $AnimatedSprite.animation = "up"
@@ -389,7 +392,7 @@ function:
         {
             animatedSprite.Animation = "walk";
             animatedSprite.FlipV = false;
-            // See the note below about boolean assignment
+            // See the note below about boolean assignment.
             animatedSprite.FlipH = velocity.x < 0;
         }
         else if (velocity.y != 0)
@@ -486,8 +489,9 @@ this code to the function:
  .. code-tab:: gdscript GDScript
 
     func _on_Player_body_entered(body):
-        hide()  # Player disappears after being hit.
+        hide() # Player disappears after being hit.
         emit_signal("hit")
+        # Must be deferred as we can't change physics properties on a physics callback.
         $CollisionShape2D.set_deferred("disabled", true)
 
  .. code-tab:: csharp
@@ -495,7 +499,8 @@ this code to the function:
     public void OnPlayerBodyEntered(PhysicsBody2D body)
     {
         Hide(); // Player disappears after being hit.
-        EmitSignal("Hit");
+        EmitSignal(nameof(Hit));
+        // Must be deferred as we can't change physics properties on a physics callback.
         GetNode<CollisionShape2D>("CollisionShape2D").SetDeferred("disabled", true);
     }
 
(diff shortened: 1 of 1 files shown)


NathanLovato (approved):

NathanLovato:
Looking good to me, but I have barely used C# in Godot and would appreciate another pair of eyes for review.

getting_started/step_by_step/your_first_game.rst
```diff
@@ -176,15 +176,20 @@ Start by declaring the member variables this object will need:
 
     extends Area2D
 
-    export var speed = 400  # How fast the player will move (pixels/sec).
-    var screen_size  # Size of the game window.
+    # These only need to be accessed in this script, so we can make them private.
+    # Private variables in GDScript have their name starting with an underscore.
```
akien-mga, line 180: GDScript doesn't have private variables per se, so this might be a bit misleading.

"Your first game" is also supposed to be simple, personally I'm not sure whether introducing users to public/private access at this stage is really useful.

Good practices are good but we should be careful with the simpler demos not to increase the overhead needlessly for beginners (that's why we don't recommend using type hints in simple demos for example).

NathanLovato: In our tuts we call these pseudo-private.

In the docs, I'd assume people have watched a CS intro course or have programming foundations; we'll link to some free resources as part of the getting started rewrite. Other pages already talk about programming patterns or classes, and there's not really a way around it considering Godot is an object-oriented engine, even on the user's side.

Here, the properties could well stay public. I don't know what's best.

cbscribe: I agree with @akien-mga, this is introducing a whole set of concepts that are really of no help in making this program work. It only adds to the cognitive load - if you understand public/private, it's misleading, and if you don't, it's just confusing.

NathanLovato: I imagine the demo in godot-demo-projects was updated to be like that. So it's not you who changed the code, @cbscribe, right? I took it as you had updated the code or someone had consulted you before changing it.

cbscribe: No, I haven't done any updates to the demo project. I did see there were changes to the C# version being made recently, but I didn't review them. Maybe it was slipped in during those?

NathanLovato: The changes were also applied to the GDScript version, yes: https://github.com/godotengine/godot-demo-projects/commit/dd11042d95d0cb996b815818d2766b35ad211e8e

aaronfranke: It's easy to undo the changes to the demo projects, we can change these back to public.

NathanLovato:
Thanks for making the variables public again. @cbscribe could you check the changes and tell us if they're fine to you?

getting_started/step_by_step/your_first_game.rst
```diff
@@ -233,9 +235,11 @@ need to do the following:
...
 this game, we have 4 direction inputs to check. Input actions are defined
 in the Project Settings under "Input Map". Here, you can define custom events and
-assign different keys, mouse events, or other inputs to them. For this demo,
-we will use the default events that are assigned to the arrow keys on the
-keyboard.
+assign different keys, mouse events, or other inputs to them.
+You can define your own custom "move_right" etc input events here, and
+define inputs such as WASD, arrow keys, and/or controller sticks.
+Alternatively, if you want, you can just use the default events called
+"ui_right" etc that are assigned to the arrow keys on the keyboard.
```
cbscribe, line 242: This seems like overexplaining for this game. Let's get them working with default input actions before talking about custom ones. This paragraph just implies you can do it but then leaves the_reader still wondering how.

Using the default actions (and these should be consistently referred to as "input actions") is what we want them doing, not an alternative choice.

getting_started/step_by_step/your_first_game.rst
```diff
@@ -245,15 +249,10 @@ or ``false`` if it isn't.
...
-            velocity.x -= 1
-        if Input.is_action_pressed("ui_down"):
-            velocity.y += 1
-        if Input.is_action_pressed("ui_up"):
-            velocity.y -= 1
+        var velocity = Vector2() # The player's movement vector.
+        velocity.x = Input.get_action_strength("move_right") - Input.get_action_strength("move_left")
+        velocity.y = Input.get_action_strength("move_down") - Input.get_action_strength("move_up")
+
         if velocity.length() > 0:
```
cbscribe, line 256: I don't like this at all. The original code was explicit and clear. Each input is shown to map to a direction. This now gets into something called "action strength" which is not at all apparent what it means.

getting_started/step_by_step/your_first_game.rst
```diff
@@ -516,15 +499,17 @@ starting a new game.
     func start(pos):
         position = pos
         show()
-        $CollisionShape2D.disabled = false
+        # Must be deferred as we can't change physics properties on a physics callback.
+        $CollisionShape2D.set_deferred("disabled", false)
```
cbscribe, line 503: This one doesn't need to be deferred, because it's not called during the physics callback.

getting_started/step_by_step/your_first_game.rst
```diff
@@ -740,7 +723,7 @@ instance.
 
     extends Node
 
-    export (PackedScene) var Mob
+    export(PackedScene) var _mob_scene
```
cbscribe, line 726: This change violates the GDScript style guide:

> Also use PascalCase when loading a class into a constant or a variable

In addition, `export` is not a function - removing the space before the export type hint makes it seem so.

getting_started/step_by_step/your_first_game.rst
```diff
@@ -865,49 +843,60 @@ Note that a new instance must be added to the scene using ``add_child()``.
 
     func _on_MobTimer_timeout():
         # Choose a random location on Path2D.
-        $MobPath/MobSpawnLocation.offset = randi()
+        var mob_spawn_location = get_node("MobPath/MobSpawnLocation");
+        mob_spawn_location.offset = randi()
```
cbscribe, line 847: What is the point of this?

getting_started/step_by_step/your_first_game.rst
```diff
@@ -865,49 +843,60 @@ Note that a new instance must be added to the scene using ``add_child()``.
...
+
         # Create a Mob instance and add it to the scene.
-        var mob = Mob.instance()
-        add_child(mob)
+        var mob_instance = _mob_scene.instance()
+        add_child(mob_instance)
```
cbscribe, line 851: We're instancing a `mob` - why does this variable need to be expanded to `mob_instance`?

getting_started/step_by_step/your_first_game.rst
```diff
@@ -865,49 +843,60 @@ Note that a new instance must be added to the scene using ``add_child()``.
...
+
         # Set the mob's direction perpendicular to the path direction.
-        var direction = $MobPath/MobSpawnLocation.rotation + PI / 2
+        var direction = mob_spawn_location.rotation + TAU / 4
```
cbscribe, line 854: This change is entirely unnecessary. Let's not get into the TAU vs PI argument.

getting_started/step_by_step/your_first_game.rst
```diff
@@ -865,49 +843,60 @@ Note that a new instance must be added to the scene using ``add_child()``.
...
-        direction += rand_range(-PI / 4, PI / 4)
-        mob.rotation = direction
-        # Set the velocity (speed & direction).
-        mob.linear_velocity = Vector2(rand_range(mob.min_speed, mob.max_speed), 0)
-        mob.linear_velocity = mob.linear_velocity.rotated(direction)
+        direction += rand_range(-TAU / 8, TAU / 8)
+        mob_instance.rotation = direction
+
+        # Choose the velocity.
+        mob_instance.linear_velocity = Vector2(rand_range(mob_instance.min_speed, mob_instance.max_speed), 0).rotated(direction)
```
cbscribe, line 864: This results in a much longer and harder to read line. We originally had it written like that in an early version of the game, and this led to additional confusion, which was why it was broken into multiple lines.

getting_started/step_by_step/your_first_game.rst
```diff
@@ -865,49 +843,60 @@ Note that a new instance must be added to the scene using ``add_child()``.
...
-        mob.position = $MobPath/MobSpawnLocation.position
+        mob_instance.position = mob_spawn_location.position
+
         # Add some randomness to the direction.
-        direction += rand_range(-PI / 4, PI / 4)
-        mob.rotation = direction
-        # Set the velocity (speed & direction).
-        mob.linear_velocity = Vector2(rand_range(mob.min_speed, mob.max_speed), 0)
-        mob.linear_velocity = mob.linear_velocity.rotated(direction)
+        direction += rand_range(-TAU / 8, TAU / 8)
```
cbscribe, line 860: As above.

getting_started/step_by_step/your_first_game.rst
```diff
@@ -1231,16 +1219,17 @@ click "Groups" and you can type a new group name and click "Add".
 .. image:: img/group_tab.png
 
 Now all mobs will be in the "mobs" group. We can then add the following line to
-the ``game_over()`` function in ``Main``:
+the ``new_game()`` function in ``Main``:
```
cbscribe, line 1222: Why move this to `new_game()`?

cbscribe (commented):
I've added several review comments. In my opinion, most of these changes are unnecessary, if not outright negative.

getting_started/step_by_step/your_first_game.rst
aaronfranke: Scenes are neither classes nor constants, and in all other instances of the demos exported PackedScenes are using snake_case so it would have to be changed there too.

getting_started/step_by_step/your_first_game.rst
aaronfranke: For clarity's sake, but we can change this back.

getting_started/step_by_step/your_first_game.rst
aaronfranke: This code is actually superior, since it properly supports joysticks, and it's also shorter. It seems clear to me, it says that the X velocity is equivalent to the right button minus the left button.

NathanLovato:
Seeing all the review comments, I should've paid more attention to the PR. They all make sense.

So @cbscribe would you recommend to close the PR and revert changes on the demo? Seems to me like it's the way to go here.

getting_started/step_by_step/your_first_game.rst
cbscribe: Well, we could discuss changing the style guide, but either way this code and the style guide should match.

aaronfranke:
There are still things that can be improved and inconsistencies with the demos, I can amend the PR and change the demos. If you reverted the most recent PR in the demos, they would still not be consistent with the docs (because they weren't before).

getting_started/step_by_step/your_first_game.rst
aaronfranke: This way you can see the enemies after you lose.

getting_started/step_by_step/your_first_game.rst
cbscribe: I get it, but it's is not superior if the user is using default input actions (which they should be in this *beginner* demo).

getting_started/step_by_step/your_first_game.rst
aaronfranke: Actually, there is not a single example of `export(PackedScene)` in the GDScript style guide, so there's nothing that says this should be capitalized.

cbscribe: @NathanLovato has this changed at some point? I believe it used to be included in this section:

>Also use PascalCase when loading a class into a constant or a variable:

> const Weapon = preload("res://weapon.gd")

There's no explicit PackedScene example in the guide now.

I'm fine either way, but this is a debate for the style guide, not for this PR. Whatever is preferred should be added there and followed.

getting_started/step_by_step/your_first_game.rst
aaronfranke: Yes, you're right. This slipped in with a change to the demos and I missed it.

getting_started/step_by_step/your_first_game.rst
aaronfranke: This same node is referenced 2 other times in the script below these lines, so this means that we don't need to get it thrice.

getting_started/step_by_step/your_first_game.rst
aaronfranke: Another interesting thing is that without the `scene` suffix and without the `instance` suffix, we lose the ability to explicitly refer to everything in C#. `mob` would be the instance, `Mob` is the class, so whether we went with a `mob` field or a `Mob` property we would be masking something somewhere.

getting_started/step_by_step/your_first_game.rst
aaronfranke: Good point, I changed this to be split into two lines, but I think it's better to use a local field rather than storing the temporary value inside of `linear_velocity`.

NathanLovato:
Hey @aaronfranke , you should've waited for @cbscribe to check this out. This tutorial is being maintained by Chris, so the least we can do is ask him before making changes. Even it it takes time to get review. When he's too busy or misses the message, you can ping him on Discord.

aaronfranke:
@NathanLovato I addressed his above changes and waited 10 days. If he finds any further issues, he is welcome to make a follow-up PR, or to yell at me and I will make a follow-up PR.

NathanLovato:
It would've taken you 10 seconds to ping the reviewer again. Did you even try to contact him before merging? If not, please do so next time.

PR #4210, opened 2020-10-06
Clarify that procedural geometry can only be created on the CPU

Calinou:
See https://www.reddit.com/r/godot/comments/j5vs8g/creating_a_mesh_using_the_gpu/.


clayjohn (approved):
Looks good. I don't think it is really necessary to document what features aren't. But in this case it is a small change, so it is fine.

NathanLovato:
Thanks!

PR #4216, opened 2020-10-07
Add a tool to move rst files and images

NathanLovato:


NathanLovato:
It's the tool I'm using to move files as in PRs like #4215

PR #4222, opened 2020-10-07
Fix build error in input_examples.rst

NathanLovato:
There's a sphinx build error that prevents the page from building. The C# string interpolation code is correct but it seems the parser or linter can't parse it. I replaced it with string formatting

This also replaces an `as` which people pointed out should be `is`.

Co-authored-by: pycbouh


NathanLovato:
@aaronfranke may I ask for your review on this?

NathanLovato:
The error should be fixed by merging #3703

PR #4236, opened 2020-10-09
Check for updates to GitHub Actions every weekday

jlosito:
Now that dependabot is native to GitHub and this project uses actions, it would be nice to always keep them up to date. This PR will allow dependabot to check for updates to any actions uses and submit PRs with version bumps automatically.

https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/keeping-your-actions-up-to-date-with-github-dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"


NathanLovato:
@akien-mga may I leave this one to you?

neikeq:
I'm not sure if a daily interval is a good idea. Could be too spammy.

jlosito:
> I'm not sure if a daily interval is a good idea. Could be too spammy.

Would monthly be a better choice?

NathanLovato:
Monthly should work well.

Calinou:
Actions being updated isn't *that* common, so I think a daily check is fine.

NathanLovato:
Is this good to merge then? I don't know much about github actions.

aaronfranke:
@jlosito What exactly would update? I assume something in [this file](https://github.com/godotengine/godot-docs/blob/master/.github/workflows/ci.yml)? Is it just `actions/checkout@v2`?

I don't know how often they are planning to bump the major version, but it has stayed at version 2 since 2019, so I can't imagine this would be spammy, maybe one PR every few years. A daily check is fine, I don't think we would get anywhere close to daily PRs.

NathanLovato:
It seems so, according to the github docs.

jlosito:
> @jlosito What exactly would update? I assume something in [this file](https://github.com/godotengine/godot-docs/blob/master/.github/workflows/ci.yml)? Is it just `actions/checkout@v2`?
> 

Yup. It would really only check that single action you're referring to since this is the only one the project uses. The PR with the version bump would look something like the following.

https://github.com/ubuntu/microk8s/pull/1615

NathanLovato:
Alright, looking good, thanks!

PR #4238, opened 2020-10-09
Document that rcedit requires valid metadata to work correctly

Calinou:
This also documents the required version formats.


NathanLovato:
Great thanks

PR #4250, opened 2020-10-11
Create new sections from Project Workflow

### LINKED ISSUE #4249: Split Getting Started -> Project Workflow section
- Make the export sub-section part of the tutorials section
- Turn the assets workflow sub-section into an Asset Pipeline section
- Turn Best Practices into its own section

NathanLovato:
Closes #4249


NathanLovato:
Here is the order of the manual's sections after this PR, new ones are in bold. Suggestions to improve the order are welcome:

-     Scripting
-     2D
-     3D
-     Audio
-     Physics
-     **Assets pipeline**
-     Math
-     Animation
-     Inputs
-     Input and Output (I/O)
-     Internationalization
-     User Interface (UI)
-     Rendering
-     Shaders
-     Networking
-     VR
-     Plugins
-     Platform-specific
-     **Best practices**
-     Performance
-     **Export**

index.rst
```diff
@@ -96,7 +96,9 @@ The main documentation for the site is organized into the following sections:
    tutorials/vr/index
    tutorials/plugins/index
    tutorials/platform/index
+   tutorials/best_practices/index
    tutorials/performance/index
+   tutorials/export/index
```
akien-mga, line 101: The Exporting section might be worth having further up, as exporting a game is not something that one should do *last*, it's actually something that should come up early on during development.

There's actually an "Exporting" tutorial right after Dodge the Creeps, since users will typically want to export their first game to see their finished playable project.

That being said I'm not sure *where* exactly this section should go, and it doesn't impact the actual moving of files so this shouldn't prevent merging.

tutorials/best_practices/index.rst
```diff
@@ -15,3 +15,5 @@ Best practices
    godot_notifications
    data_preferences
    logic_preferences
+   project_organization
+   version_control_systems
```
akien-mga, line 19: I'm not sure if this is the best location for those (both with regard to being in "Best practices" and at the end of that section), since both are topics that users should read early on (at least the project organization one, the VCS is most about the Git plugin which is fully optional).

The rest of the Best practices section seems to be fairly advanced stuff that one would want to read after having used Godot a bit, to understand how to go from "hack your way to make it work" to "best practice guidelines" for complex projects.

Maybe these would be worth having in a section about setting up a project? There could also be documentation about some settings that one needs to define when creating a project (renderer, physics backend for 3D, import formats for VRAM compressed textures depending on the target platforms, etc.).

That being said, I'm fine merging this as is and move things again later on if a better option presents itself.

akien-mga (approved):
Left a couple comments, but I'm fine with this being merged as is and improved later on, if there's a need.

index.rst
NathanLovato: We should probably review and swap all sections. I'd just need to know if it should be by the order in which we'd recommend to read, in order of how likely people are to read it, etc. Alphabetical could also be an option.

tutorials/best_practices/index.rst
NathanLovato: Sure, I put them where I felt could fit with the sections we had. Best practices is advanced at the moment mostly because Will wrote it from an engineer's perspective.

To me, the section name itself, "Best practices", is a good fit for content that answer common beginner questions. What are the best practices is one of the topics people request the most, be it for scripting, how to split or organize scenes, files, etc. I think it's also a section name that people will click. But there's work to do on the section for that to happen.

A project setup section could work. Note that we already have things covered in relevant sections: rendering for the renderers, physics backend in physics... they fit well in there leaving a limited set of topics for a project setup section.

NathanLovato:
Here's a proposal for the section order. It's arbitrary, I'm just trying to figure out a possible "reading order":

- Scripting
- 2D
- 3D
- Audio
- Physics
- Inputs
- Assets pipeline
- Export
- Animation
- User Interface (UI)
- Input and Output (I/O)
- Internationalization
- Math
- Rendering
- Shaders
- Networking
- VR
- Plugins
- Platform-specific
- Best practices
- Performance

Calinou:
For reference, this is what alphabetical order would look like:

- 2D
- 3D
- Animation
- Assets pipeline
- Audio
- Best practices
- Export
- Input and Output (I/O)
- Inputs
- Internationalization
- Math
- Networking
- Performance
- Physics
- Platform-specific
- Plugins
- Rendering
- Scripting
- Shaders
- User Interface (UI)
- VR

NathanLovato:
I also like the alphabetical one. Easy to manage for us, easy to parse for the user. In a conference from the Write the docs project, I heard that after using analytics in some software docs, they noticed that users mostly navigated through inline links. So even if something ends up low in the list, as long as we link to it in relevant pages, users should be able to find it.

akien-mga:
Alphabetical looks pretty good to me. It's quite difficult indeed to decide what should be the "natural" reading order, and letting users parse easily to find topics relevant to them is better IMO.

So let's go with that.

akien-mga:
I'd say we can merge this once sections have been reordered alphabetically (and a rebase is needed anyway).

NathanLovato:
Thanks @akien-mga, done, and I went ahead and merged as the changes were approved!

Issue #4035, opened 2020-09-18
Rewrite of the Getting Started section: plan


NathanLovato:
The task: Rewrite the getting started guide to explain key concepts of Godot and help users get started in a couple of hours. The content should frequently link to other pages in the user manual for more information.

Notes:

_1. By rewrite, I do not mean to throw away everything and restart from zero. But there is a lot of writing and editing work involved with this task.
2. Multiple sub-sections in the current Getting Started section are going to move to other sections. In general, we want Getting started to be one short series, and then link people to other parts of the docs they want to explore. See #3969 for more information._

## Introduction and goals

Godot has different kinds of users with varied skills and needs. The current step-by-step guide contains mixed content, sometimes with language that's too technical for many users. As many persons wrote it over several years, it also lacks coherence.

This task aims to make the series accessible enough so people who are new to game creation, but not entirely new to programming, can follow it. Having some programming foundations is a pre-requisite to learn an extensive API like Godot's.

For people who have no programming experience, we will link to comprehensive free resources to get them started:

1.  Harvard's intro to computer science, CS50, a [free open courseware](https://cs50.harvard.edu/x/2020/), and relies on open-source technologies.
2.  The free ebooks and Udemy courses by Al Sweigart are great for self-paced learning: [Automating the boring stuff with Python](https://automatetheboringstuff.com/), and Invent with Scratch for young adults.

The getting started guide should not cover engine features in-depth. Instead, it will link to pages that cover given topics in greater depth.

The getting started section should help users:

1.  Understand the fundamental concepts of Godot, like nodes, scripting, the scene tree, and more.
2.  Get a sense of available features, how they work, and see if Godot is a good match for their projects.
3.  Know their way around the editor.
4.  Get started creating 2D and 3D games with Godot.
5.  Find more information about specific aspects of the engine.

## The plan

The plan below is a list of topics the Getting Started section should cover. The order and topics are subject to change, of course. This redesign of the getting started series is meant to make it more coherent and to use consistent writing and language.

The Getting Started section page should explain how the guide works and the pre-requisites for each series. In particular, it should link beginners to free resources to learn to code.

I'm also going to work on a landing page for the docs linking to other sections as a separate task.

1.  Introduction
    1.  Introduction to Godot and general overview its features: a guide to help answer questions such as "Is Godot for me? What can it do for me?"
    2.  First look at the editor and its interface.
[...]

PR #2958, opened 2019-12-01
Added Guides and Node Flexibility Tutorial.

willnationsdev:
This creates a supplemental introduction to Godot that should work alongside the step-by-step tutorial in order to quickly get someone familiarized with Godot's systems and link them to relevant parts of the documentation.

The step-by-step tutorial specializes in getting people's feet wet with an actual project and teaches them Godot's basics along the way. This guide focuses on explaining Godot's most important concepts to users at a high level. If it gets too in-depth, then it favors directing the user to other, more detailed parts of the docs. For example, as I was going into detail about node flexibility, there was too much information, so I outsourced it to an entirely new "Node Flexibility" tutorial section.

It has the luxury of not needing to be a tutorial itself. This enables it to move faster, cover more topics, and get people kickstarted more quickly. It is especially useful to people who are already familiar with programming, due to the speed with which it covers important topics. However, it has been organized with the assumption that the reader knows next-to-nothing about programming.

Since I noticed there was a Unity-to-Godot guide in the `getting_started/editor` section as well, I figured it was in the spirit of things to create a dedicated subsection for "guides". We could then conceivably add more guides on getting started with Godot that are tailored to different audiences.

I understand that this is a huge update, but it is ultimately one full-length document, not an entire series like I did before. It was important for me to put it all together in one PR as many parts of this were actually rearranged as I worked on it (even just hours before this submission). I don't expect this to get merged soon, given the amount of content that needs to be reviewed.

A few caveats:

1. I can't compile this at the moment because the `getting_started/animation/introduction_2d` tutorial has some problems with images / image directives.
2. I still need to compress the many images associated with this PR.

Please let me know what you think. Thanks.

Issue #4169, opened 2020-10-02
Getting Started / Introduction / How to learn Godot most efficiently


NathanLovato:
A guide to using the built-in code reference, online documentation, and best practices to get help on available communities.

Part of the getting started session’s rewrite. See #4035 for more information.

PR #4251, opened 2020-10-11
Add 'How to read the Godot API' docs.

willnationsdev:
An introduction to the format of the Godot API class reference pages.

This is a partial rewrite of portions of #2958 

cc @NathanLovato 

Notes:

1. I put it in the getting started section since that's what #4035 seemed to mention. Let me know if it should be moved.
2. For some reason, many of the images I take with Windows Snippet end up blurry even though they look fine in the app. Also tried manually screenshotting and cropping in Paint, but that also lead to a blurry result on the webpage. Any suggestions?


NathanLovato:
The idea is not to provide a reference of the reference but a guide on how to learn new things in Godot using the docs. See  #4169 

Your page is describing the reference page, which I don't find too useful for new users, whose problems are more like "how do I move objects around the screen?", "what can I do with this node?". Multiple parts of the reference pages like tutorials, descriptions, brief description are self-explanatory. I would mention in one or two paragraphs what you can find on a reference page, explain the inheritance, (the current class has all properties and methods from its parents), and most importantly how to search using it.

In short, you don't want to provide a reference describing the reference, but an actionable guide on using and searching the documentation in general, and tips to find the information you're looking for.

I'd also recommend writing what the user will learn at the top of every page you write. Like "This page explains ..." or "In this guide, you will learn...". Doing so

1. Tells the user whether that's the page they're looking for in 10 seconds.
2. Helps you frame your guide in terms of the techniques you teach.

NathanLovato:
Reviewing the page again, I think it's good to have. I'd move it to the scripting section or another one. Not sure which is best, but scripting seems close enough.

Calinou (approved):
Should be good to merge after rebasing and moving the page to the Scripting section.

akien-mga:
> Should be good to merge after rebasing and moving the page to the Scripting section.

Bump :)

willnationsdev:
1. Rebased
2. Moved code to Scripting section.
3. Removed the empty spaces beneath each image (ended up looking really weird in the final version).
4. Changed a reference to Callable to be a link.

willnationsdev:
bump. Should be ready to merge, afaik.

NathanLovato:
Thanks for your work!

PR #4263, opened 2020-10-12
Recommend vertex shading for particles in GPU optimization

Calinou:
Vertex shading is a common way to speed up particle rendering, since shading a quad is much faster on a per-vertex approach compared to a per-pixel approach. (It makes sense for many kinds of "3D" particles too.)

### DIFF
--- a/tutorials/performance/gpu_optimization.rst
+++ b/tutorials/performance/gpu_optimization.rst
@@ -173,6 +173,8 @@ You can increase performance in a fill rate-limited project by reducing the
 amount of work the GPU has to do. You can do this by simplifying the shader
 (perhaps turn off expensive options if you are using a :ref:`StandardMaterial3D
 <class_StandardMaterial3D>`), or reducing the number and size of textures used.
+Also, when using non-unshaded particles, consider forcing vertex shading in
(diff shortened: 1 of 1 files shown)


tutorials/performance/gpu_optimization.rst
```diff
@@ -175,7 +175,8 @@ amount of work the GPU has to do. You can do this by simplifying the shader
 <class_StandardMaterial3D>`), or reducing the number and size of textures used.
 
 **When targeting mobile devices, consider using the simplest possible shaders
-you can reasonably afford to use.**
+you can reasonably afford to use.** Also, when using non-unshaded particles,
```
Calinou, line 178: "Non-unshaded" is a double negative, but since the material flag is called "Unshaded", I'd prefer explicitly mentioning it.

When using an unshaded material, the shading mode doesn't matter.

lawnjelly: Also I'm wondering if moving the mention of vertex shading might be better moved up to join with the paragraph above, I think it might flow better.

NathanLovato:
Thanks, as always!

PR #4264, opened 2020-10-13
Make PerlinNoise3D accept scalars directly

Beuc:
The current code requires creating and plugging a Scalar node just to enter the value "10" (see current screenshot); with this change, the node, like existing Godot nodes, can take the scalar value directly within the PerlinNoise3D Node.

### DIFF
--- a/tutorials/plugins/editor/visual_shader_plugins.rst
+++ b/tutorials/plugins/editor/visual_shader_plugins.rst
@@ -50,6 +50,10 @@ all you need to initialize your plugin.
         return "Classic Perlin-Noise-3D function (by Curly-Brace)"
 
 
+    func _init():
+        set_input_port_default_value(2, 0.0)
+
+
     func _get_return_icon_type():
         return VisualShaderNode.PORT_TYPE_SCALAR
 


Chaosus:
I think the offset param can be rid off too...

Chaosus (changes requested):
I think it's fine but the screenshot in the original article should also be updated accordingly.

Beuc:
With this change, the screenshot will actually look the same with a scalar (or any other) node plugged to 'scale'.
It's different only if the user decides to set the variable within the node.

Beuc:
One issue however: the node doesn't seem to properly save the value for 'scale' if it's set directly within the node. The value is maintained while editing the project (including closing/reopening the scene), but the value is reset to 0 when closing the project and opening it again. Tested with Godot 3.2.3.

Chaosus:
> With this change, the screenshot will actually look the same with a scalar (or any other) node plugged to 'scale'.
> It's different only if the user decides to set the variable within the node.

Yeah, I thought you wanted to reflect it on the screenshot by disconnecting from the node. But if not - I think it's also fine.

> One issue, however: the node doesn't seem to properly save the value for 'scale' if it's set directly within the node. The value is maintained while editing the project (including closing/reopening the scene), but the value is reset to 0 when closing the project and opening it again. Tested with Godot 3.2.3.

Could you report this as an issue? - that would be more correct.

Chaosus (approved):

Beuc:
I can report the issue though I was wondering whether the issue is in Godot or from something missing in my patch.

NathanLovato:
Thanks!

PR #4265, opened 2020-10-13
Fix "Better Collada Exporter" link in FAQ

davtur19:


Calinou (approved):
Thanks!

In the future, this FAQ needs some revision as Better Collada doesn't support Blender 2.80 or later (unless you use one of its forks).
Also, we want to better promote the use of glTF 2.0 as it's the best-supported format right now.

PR #4280, opened 2020-10-16
Add more information to the `has` method

xsellier:
Since someone already opened an issue regarding this function, I though it would be great to have some explanations about those corners cases.
https://github.com/godotengine/godot/issues/39987


Calinou:

PR #4281, opened 2020-10-17
Unscale gravity by delta in `move_and_slide` examples

nwtnni:
The documentation for [move_and_slide][ms] states the following:

> linear_velocity is the velocity vector (typically meters per second).
> Unlike in move_and_collide, you should not multiply it by delta — the
> physics engine handles applying the velocity.

This PR removes the delta scaling and adjusts the `gravity` constant
accordingly, assuming a delta value of 1/60.

[ms]: https://docs.godotengine.org/en/stable/classes/class_kinematicbody.html?#class-kinematicbody-method-move-and-slide

### DIFF
--- a/tutorials/physics/using_kinematic_body_2d.rst
+++ b/tutorials/physics/using_kinematic_body_2d.rst
@@ -130,7 +130,7 @@ and ``get_slide_collision()``:
         var collision = get_slide_collision(i)
         print("I collided with ", collision.collider.name)
 
-.. note:: `get_slide_count()` only counts times the body has collided and changed direction.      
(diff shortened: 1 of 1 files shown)


cbscribe:
This is wrong. Gravity has nothing to do with `move_and_slide()`, which is using the *velocity*, measured in pixels/sec, to calculate the distance (by multiplying by delta). When we calculate *acceleration*, we also need it to be based on the frame time, so multiplying gravity by delta is correct.

nwtnni:
Right, that makes sense. Thanks for clearing up my confusion!

PR #4291, opened 2020-10-19
Improve supported 3d formats faq section

skyace65:
I saw the recent PR for this section and wanted to make further changes.

The section title is better. There is now a list of the supported formats with glTF 2.0 recommended. I've taken out the link to the exporters because that seems better suited for the Import 3D Models page, I think the FAQ section should be short and to the point.

side note related to this PR. In the future we need to add a section to the import page on how to export glTF 2.0 from Maya and Max. It looks like OpenCollada isn't getting any more updates.

### DIFF
--- a/about/faq.rst
+++ b/about/faq.rst
@@ -152,16 +152,22 @@ The main reasons for creating a custom scripting language for Godot were:
 
 GDScript was designed to curtail the issues above, and more.
 
-What type of 3D model formats does Godot support?
--------------------------------------------------
+What 3D model formats does Godot support?
+-----------------------------------------
 
-Godot supports Collada via the `OpenCollada <https://github.com/KhronosGroup/OpenCOLLADA/wiki/OpenCOLLADA-Tools>`_ exporter (Maya, 3DSMax).
-If you are using Blender, take a look at our own `Better Collada Exporter <https://github.com/godotengine/collada-exporter>`_.
+Godot supports the following formats:
 
-As of Godot 3.0, glTF is supported.
+- glTF 2.0 *(recommended)*
+- Collada
+- OBJ
+- FBX (Static meshes only)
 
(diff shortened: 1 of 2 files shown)


about/faq.rst
```diff
@@ -152,17 +152,22 @@ The main reasons for creating a custom scripting language for Godot were:
...
 
-Godot supports Collada via the `OpenCollada <https://github.com/KhronosGroup/OpenCOLLADA/wiki/OpenCOLLADA-Tools>`_ exporter (Maya, 3DSMax).
-If you are using Blender, take a look at our own `Better Collada Exporter <https://github.com/godotengine/collada-exporter>`_.
+Godot supports the following formats:
 
-As of Godot 3.0, glTF is supported.
+- glTF 2.0 *(recommended)*
+- Collada
+- OBJ
+- FBX
```
Calinou, line 163: Current support for animated FBX scenes is really bad, so I recommend we write it off explicitly:

```suggestion
- FBX (static meshes only)
```

The situation should improve a lot in 4.0 thanks to the new FBX importer, but it's too large to be cherry-picked in 3.2.x.

about/faq.rst
```diff
@@ -152,17 +152,22 @@ The main reasons for creating a custom scripting language for Godot were:
...
 
 FBX is supported via the Open Asset Import library. However, FBX is proprietary
 so we recommend using other formats listed above, if suitable for your workflow.
```
Calinou, line 166:
```suggestion
FBX support is the fruit of reverse engineering via the Open Asset Import library.
However, FBX is proprietary so we recommend using other formats listed above,
if suitable for your workflow.
```

Calinou:
Superseded by https://github.com/godotengine/godot-docs/pull/4292.

PR #4293, opened 2020-10-19
Update the IRC logs location to follow migration

Calinou:
eska's logs archive is no longer accessible.


clayjohn (approved):

clayjohn:
We should cherry-pick this for stable as well. And maybe the other point releases.

PR #4296, opened 2020-10-22
Update best_practices_for_engine_contributors

SirQuartz:
Remove the redundant use of the word "actual" in _"The Problem always comes first"_. In _"The Solution must be discussed with others"_ Remove an unnecessary comma as well as the redundant use of the word "own", which is not necessary in this context because "their" is already used. This is a minor improvement to conciseness, clarity, and readability.


NathanLovato:
Thanks

PR #4300, opened 2020-10-22
Make explicitly clear that free() doesn't set references to null

### LINKED ISSUE #4270: Make clear that free() doesn't set any references to it to null (in release builds)
**Your Godot version:**
3.2.2

**Issue description:**
While investigating a segmentation fault in a release build I realized one of the scripts was calling a method on an object that has been freed in the previous frame with `queue_free`. I assumed that all references to freed objects were automatically set to null since that is what seems to be happening in the editor (and debug builds).

That said, I'm not sure how [is_instance_valid](https://docs.godotengine.org/en/stable/classes/class_@gdscript.html?highlight=get_stack#class-gdscript-method-is-instance-valid) is supposed to work then (and it seems it indeed doesn't work at all). Is there something I'm misunderstanding?

https://docs.godotengine.org/en/stable/classes/class_object.html#class-object-method-free

Demindiro:
I believe it isn't made clear enough yet that free() does not set references to an object to `null` in release builds unlike in debug builds.

This pull request adds a note that explicitly states this behaviour shouldn't be relied on.

PS: Am I actually supposed to commit to master? In the master branch, the description of `free()` states:

> Deletes the object from memory. Any pre-existing reference to the freed object will now return null.

Which is evidently incorrect. It also doesn't seem right to me 909 files are changed.

Closes #4270


Calinou:

PR #4307, opened 2020-10-24
Add a symbol next to class reference page links

Calinou:
This makes them easier to distinguish from "user manual" page links, especially in search results. The "C" is for "class" :slightly_smiling_face: 

It's not perfect, but it should be a decent interim solution until we add a way to filter between user manual and class reference results in search.

## Preview

image

image

<details>
<summary>Old version</summary>

### Page (light theme)

image

### Search results (dark theme)

image
</details>

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -31,6 +31,7 @@
     --link-color-active: #105078;
     --link-color-visited: #9b59b6;
     --external-reference-icon: url("data:image/svg+xml;base64,PHN2ZyBoZWlnaHQ9IjEyIiB3aWR0aD0iMTIiIHhtbG5zPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2ZyI+PGcgZmlsbD0ibm9uZSIgc3Ryb2tlPSIjMjk4MGI5Ij48cGF0aCBkPSJtNy41IDcuMXYzLjRoLTZ2LTZoMy40Ii8+PHBhdGggZD0ibTUuNzY1IDFoNS4yMzV2NS4zOWwtMS41NzMgMS41NDctMS4zMS0xLjMxLTIuNzI0IDIuNzIzLTIuNjktMi42ODggMi44MS0yLjgwOC0xLjMxMy0xLjMxeiIvPjwvZz48L3N2Zz4K");
+    --classref-badge-text-color: hsl(0, 0%, 45%);
 
     --hr-color: #e1e4e5;
     --table-row-odd-background-color: #f3f6f6;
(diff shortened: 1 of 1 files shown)


akien-mga:
That's a good idea :+1:

I don't if this `C` in a circle is understandable enough though. It might look like ©, or be interpreted as relating a C API.

One option should be `</>` which seems to be a popular icon for code (at least when you google "code icon"), but since it refers to HTML/XML markup, it might be confusing too.

Could use a script icon similar to what we have in the editor for scripts Screenshot_20201024_161857, though it's maybe a bit too representational.

Or a cog icon like we have for script tabs in the script editor Screenshot_20201024_161928. Quite arbitrary but it's less figurative that a parchment, and should be recognizable by Godot users.


If others have suggestions, please chime in.

pycbouh:
What if it spelled "ref" instead?

NathanLovato:
I think a script icon would be a good fit: it conveys the idea of code or code file. The C or ref don't convey the meaning to me, you'd have to learn what it means first.

pycbouh:
I claim no knowledge of textbooks in English, but I do hope that "ref" conveys "reference" as in "reference material" for people familiar with academic works in any capacity (even if it was Wikipedia). You do kind of have to learn what that refers to here exactly, but that can be said about any sign or symbol or text we may add, I would think.

Maybe a quick investigation into other documentations with class references available online is in order.

Calinou:
Example with `ref` and smaller text:

image

Script icon:

image

Cog icon:

image

Which one do you prefer?

NathanLovato:
The script icon for me, I think it works well with the editor but I don't mind ref either. The cog it too reminiscent of settings buttons on many apps to me.

pycbouh:
Yes, definitely not the gear/cog. Agree on it meaning "settings" for me.

Calinou:
I amended the pull request to use the "ref" badge.

NathanLovato (approved):
Building the branch locally, it works as intended. The only bit that looks a bit off to me is with the JavaScript warning on latest, but it's fine, really, considering it's a work-in-progress version of the docs.

_static/css/custom.css
```diff
@@ -277,6 +279,18 @@ a.btn:hover {
     padding-right: 13px;
 }
 
+/* Distinguish class reference page links from "user manual" page links with a "C" badge. */
```
akien-mga, line 282: Outdated comment :)

akien-mga:
I'm fine with both `ref` and the script icon.

IMO the meaning of this symbol should be explained in an introduction page on how to use the docs, so we make it obvious to users. Similarly, the symbol for external resources could be explicated in the same place.

akien-mga:
Thanks!

Issue #4188, opened 2020-10-02
Include Comments in Code examples


CitrusWire:
Godot Documentation Code examples should have comments

Currently the Godot Documentation, and most of the Godot Engine Demos do not have code comments. Code as a rule should of course be commented, but documentation code doubly so.

Reasons this should be done:
* Code is write-once, read-many.
* Documentation code is write-once read-thousands.
* It gets Godot users used to comments in code and hopefully they'll follow the practice, which will benefit them.
* Documentation has the purpose of teaching people about something. It's really hard to learn about something if that something is not explained.
* Code comments lower the barrier to entry. Not everyone is an UberHack3r who can immediately parse and comprehend any code.
* A documented function (i.e. with a docstring) explains succinctly to the user what the function is doing, what the input expected is, and what the output expected is. All things that are extremely useful to know at a glance for any function, but doubly so for documentation functions.
* Allows users to easily skip over functions that don't do what they already know.

This suggestion incorporates the following points (feel free to split into separate issues):
1) Require all future documentation code to include appropriate comments
2) Retroactively add appropriate code comments to all documentation code

3) Require all future Godot Engine Demos to have appropriate code comments
4) Retroactively add appropriate code comments to Godot Engine Demos

----
Documentation; A prime example:
https://docs.godotengine.org/en/latest/getting_started/scripting/gdscript/gdscript_styleguide.html
Which has "Here is a complete class example based on these guidelines:" - and then a huge wall of code that has only 3 comments at the top. No function comments, no per-line comments, etc.
It's not all terrible, scroll down to "Quotes" and that has good code comments, as does "File names", but that's it for the entire document.

Furthermore, there's nothing on that style guide about commenting at all, beyond the fact you can write them. This despite the style guide purports to be influenced by PEP 8, which itself explicitly includes code comments, including via Documentation Strings and PEP 257.

https://docs.godotengine.org/en/latest/getting_started/scripting/c_sharp/c_sharp_style_guide.html isn't any better.

----
Demo's
The Demo's also require code comments for the same reasons. I've been playing with the translation one today for example. The code is 17 lines, no comments. Simple enough. Yet despite spending 15 minutes fiddling with it, I still have no idea how Godot knows to go from `TranslationServer.setlocale(x)` to "change the image/audio to the right file". I know I can find out from a dozen different places online, but that then obviates the point of this demo.

[...]

PR #4064, opened 2020-09-24
Write content guidelines

NathanLovato:
Some principles and guidelines to help assess what we should include in the docs.

It's inspired by the VueJS principles recommended by @Calinou.

Closes #4063

PR #4308, opened 2020-10-24
feat(multiplayer) networked player movement tut

jonbonazza:

### DIFF
--- a/tutorials/networking/index.rst
+++ b/tutorials/networking/index.rst
@@ -6,6 +6,7 @@ Networking
    :name: toc-learn-features-networking
 
    high_level_multiplayer
+   networked_player_movement
    http_request_class
    http_client_class
    ssl_certificates
--- a/tutorials/networking/networked_player_movement.rst
+++ b/tutorials/networking/networked_player_movement.rst
@@ -1,0 +1,289 @@
+.. _doc_networked_player_movement:
+
+Networked player movement
+=========================
+
+Most modern online multiplayer games operate under what is called a server authoritative client-server model. 
+In this model, clients connect directly to a game server over a socket connection and pass messages back and forth. 
+Being server-authoritative means that the clients are "thin" or "dumb" and make no decisions about the state of the game
+world themselves. Instead, they merely pass player input to the server and handle the display of whatever the server tells it
+that the current state of the game world is. In other words, the server is the arbiter of the game world that all clients must abide by.
+
+Normally, this isn't that big of a deal. A client informs the server of its intent, such as :code:`Player swings sword` and the server 
+receives this intent, modifies its local game world state, and then broadcasts the relevant new state to all clients. When a client receives
+this new state, it updates its own local copy of the game world state and renders it. This asynchronous nature of communication between the 
+server and clients is what allows players spanning geographical regions to interact with a shared game world in real time. There is one facet 
+of online multiplayer games that doesn't work so seamlessly in a sever authoritative model. That being player movement.
+
+If you consider how player movement would work if implemented using the appraoch above, it would go something like this:
+
+1. Player sends intent to move left
+2. Server receives players intent
+3. Server moves player left in its local game world
+4. Server broadcasts player's new transform to all clients
+5. Clients update individual local game worlds with the player's new transform and renders it
+
+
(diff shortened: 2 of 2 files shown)


NathanLovato:
Thanks for your contribution.

I'd recommend you to put this tutorial on a personal blog or repository for now. We're generally removing tutorials from the docs and moving towards a more reference-like style. Your page also doesn't follow our writing guidelines ( https://docs.godotengine.org/en/latest/community/contributing/docs_writing_guidelines.html ), gdscript style guide ( https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_styleguide.html ), and there are things like you can't have comments inside the code for now because they're not translatable. See #4188.

In any case, I'll let someone else chime in, but to me, this doesn't fit what we're working towards with the docs. Although this is still up for discussion. See #3390 and #4064 for more information

jonbonazza:
I understand the writing style mismatches, but can you elaborate on how it doesnt fit the gdscript guidelines? Im following pep8 pretty closely i believe and i dont see anything out of place from whats mentioned in the gdscript guidlines.

No comments is also unfortunate. There are some things that make better sense when you see the explanation next to the exact code it references.

NathanLovato:
Our guidelines aren't pep8. Also, we care about naming and other aspects. Four examples of things you did that don't fit the guidelines:

- You have only one line between functions instead of two for long code snippets
- You use abbreviated variable names, like `curr` for `current`, etc.
- You put pseudo-private properties before public ones
- You put the class_name definition after extends

In any case, this tutorial doesn't follow the docs' intended direction, to me.

jonbonazza:
@NathanLovato understood. This is something that is commonly asked for assistance with in the Godot community channels, such as Discord, so I think having official documentation around this area is still important. I'm open to feedback around how exactly it _should_ be documented, but having documented in some way is something we should definitely do I think.

I've seen the idea thrown around that godot-docs is moving away from "tutorials" (from my understanding, "tutorials" here means "step-by-step" guides?) and we should favor "reference" style docs, but so far, I am unable to understand what exactly is mean by "reference-style" documentation. Having been a software engineer for many yeast, to me "reference" docs means "class reference documentation" which I don't think is what you mean here. Does there exist a clear and concise description of what is meant by "reference-style documentation" in the context of Godot docs? I think this would be helpful to better elaborate on what is desired.

NathanLovato:
Sure. A good example is the vuejs docs: https://v3.vuejs.org/guide/instance.html#lifecycle-diagram

It's not exactly their code reference (describing each entry in their API), it's full of guides, yet the guides describe a vue application instead of teaching you how to code specific types of apps or javascript and html in general. 

In the case of Godot's web docs, I'd start by:

1. Giving an overview of the available apis
2. Giving an overview of each api's broad features and capabilities

At the moment, we need docs that describe how Godot's features work in general over pages that teach how to solve a problem that's not necessarily about Godot.

mhilbrunner:
Content-wise, this is good, but I agree with Nathan that this strays out of the scope of what the Godot docs should (and can) cover.

We can't teach people networking. There are other, better resources for that. The high-level multiplayer tutorial points this out and links to, for example, the GafferOnGames article series.

Likewise, we can document what pathfinding facilities or rendering capabilities are available in Godot, and maybe give examples to use them, but we can't go into different pathfinding algorithms you could implement and show how to do that in Godot, or show how to write your own renderer and switch it out. Although both would be cool :)

We just don't have the resources to do this in a good way and to maintain it.

So closed, but *please* consider posting this somewhere for visibility and longevity, because it is good content. We really should see about getting something more user-oriented like a Wiki going.

mhilbrunner:
Addendum:
> I've seen the idea thrown around that godot-docs is moving away from "tutorials" (from my understanding, "tutorials" here means "step-by-step" guides?) and we should favor "reference" style docs, but so far, I am unable to understand what exactly is meant by "reference-style" documentation.

Reference-style isn't "this is how you do <x> in Godot", but "Godot has these functions, this is how they work".
Godot *doesn't* currently have any built-in ways to do network prediction, reconcilation and such. The users need to build this themselves.

Still, having such "how to do X in Godot" is very valuable and useful, I agree. Thus, a official platform for such would be cool, like a Wiki. It's not that we don't want such content, we just can't maintain it ourselves.

Vennnot:
This would make great course or youtube content Nathan!

PR #4312, opened 2020-10-25
Update class_projectsettings.rst

SirQuartz:
Add missing description to class "ProjectSettings_property_rendering/quality/filters/depth_of_field_bokeh_quality". This is a minor change to add a description to the class.

### DIFF
--- a/classes/class_projectsettings.rst
+++ b/classes/class_projectsettings.rst
@@ -3849,6 +3849,8 @@ The video driver to use ("GLES2" or "Vulkan").
 | *Default* | ``2`` |
 +-----------+-------+
 
+Sets the quality level of the depth of field effect. Higher values will take more samples and be slower, but look smoother.
+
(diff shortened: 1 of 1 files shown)


LikeLakers2:
Hi! You should make these sorts of documentation changes to the [`godotengine/godot` repository](https://github.com/godotengine/godot), specifically the files in the [`doc/classes/`](https://github.com/godotengine/godot/blob/master/doc/classes) folder in that repository. The RST files here get generated from the XML files in that directory -- so your changes would be overwritten soon if you made the change here.

However, it seems that this field has already been filled out in that XML file -- see <https://github.com/godotengine/godot/blob/e16729a8c3664f6ca00665714c7d88acd4ef603e/doc/classes/ProjectSettings.xml#L1010-L1012> -- so I don't know if you need to make that PR.

PR #4324, opened 2020-10-29
GDNative C++: fix links to godotengine/godot-cpp

Naryosha:
The repository was moved from GodotNativeTools to godotengine.


Calinou (approved):
Thanks!

PR #4328, opened 2020-10-31
Explain what fractional means in the Shading language reference

lukostello:


pycbouh:
I think that the mathematical sense should be noted first, and then we can explain it in Layman’s terms in parentheses; and not the other way around.

lukostello:
x - floor(x) is its mathematical sense.. I thought fract() or fractional could mean that there was division happening somewhere

pycbouh:
Fractional is the same thing as decimal, and means [this](https://en.wikipedia.org/wiki/Fractional_part).

What I'm saying is that we should keep the term and if it is confusing explain it in parentheses. We should not replace the term with a layman's explanation. See this approach in the following line of the same article with "Modulo".

lukostello:
ah I see. I'll follow that convention then

lukostello:
I am starting to get a hang of gitKraken

pycbouh:
Good! You need to squash your commits into one though 🙂

Calinou (approved):
Thanks!

(We can squash commits using GitHub on this repository, so I just did it when merging.)

PR #4334, opened 2020-11-02
Clarify "source" mesh in Using MultiMeshInstance

AlexanderPruss:
I found the original sentence confusing because the word "mesh" is used twice, referring to different nodes. 

> One node is used as the target, the mesh that you want to place multiple meshes
on. 

This target mesh is referred to in the tools and later on in the docs as the surface mesh, so I figure we might as well call it that right away - 

> One node is used as the target, the **surface** mesh that you want to place multiple meshes
on.


Calinou (approved):
Thanks!

PR #4337, opened 2020-11-04
make example asteroids spaceship fly up/forward

dubbl:
I noticed that in the physics `RigidBody2D` example game the "asteroids" like spaceship would accelerate downwards when pressing the UP key.

This changes the thrust to the "upwards Vector2" of `Vector2(0, -250)`.

[Source: NASA](https://www.grc.nasa.gov/WWW/K-12/rocket/rktth1.html)


Feniks-Gaming (approved):
Makes sense IMO

NathanLovato:
Thanks!

PR #4340, opened 2020-11-05
Fix the "ref" badge appearing twice in instant search results

### LINKED ISSUE #4338: "ref" badge is duplicated in instant search results
#4307 introduced badges for class reference links, but the badge is displayed twice in the "instant search" results:

image

Calinou:
Tested using the Firefox developer tools on the live website.

Thanks @pycbouh for finding this fix :slightly_smiling_face: 

This closes #4338.

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -291,6 +291,20 @@ a[href*="classes/"]::before {
     margin-right: 0.25rem;
 }
 
+/* Prevent the "ref" badge from appearing twice in the instant search results (not testable locally). */
+.wy-body-for-nav .search__result__single a[href*="classes/"]::before {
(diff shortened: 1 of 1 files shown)


pycbouh:
I hope you don't think that I'm too lazy to make my own PRs 😄 

But since you made one, this should fix [the dedicated results page](https://docs.godotengine.org/en/latest/search.html?q=AnimatedTexture&check_keywords=yes&area=default) as well, I think:
```css
#search-results .context a[href*="classes/"]::before {
    display: none;
}
```

Calinou:
@pycbouh That works, thanks again!

NathanLovato:
Thanks!

Issue #4211, opened 2020-10-06
Make code snippets translatable


CitrusWire:
**Issue description:**
On the back of https://github.com/godotengine/godot-docs/issues/4188 ("Include Comments in Code examples")

Make code snippets in documentation translatable. This way code comments can be translated if desired.

skyace65:
Closing this issue, same issue described in #3284

PR #4343, opened 2020-11-06
Document how to compare floats properly in Shading language

### LINKED ISSUE #4342: Shading reference should warn about comparing floats
Hi, I made an animated sprite character and made an override material that replaces a single color on the shirt of my sprite.

However It works for one sprite, but when I make an animation at the animationplayer using the sprites, they look like no shader has changed them.

Calinou:
This closes #4342.

### DIFF
--- a/tutorials/shaders/shader_reference/shading_language.rst
+++ b/tutorials/shaders/shader_reference/shading_language.rst
@@ -6,11 +6,13 @@ Shading language
 Introduction
 ------------
 
-Godot uses a shading language similar to GLSL ES 3.0. Most datatypes and functions are supported,
-and the few remaining ones will likely be added over time.
+Godot uses a shading language similar to GLSL ES 3.0. Most datatypes and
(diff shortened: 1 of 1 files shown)


clayjohn (commented):
Maybe we should like to this as well https://floating-point-gui.de/

NathanLovato:
Good, but I'd recommend splitting the code example and moving comments out of it; that's because at the moment, they're not translatable and won't be available to users internationally. At least, that's unless you or someone else is planning to tackle #4211 soon-ish.

NathanLovato:
If you're okay with the idea of turning comments into paragraphs so they can be translated, may I go ahead, make the edits, and merge?

clayjohn:
@Calinou ping. 

For the record, I agree with @NathanLovato

Calinou:
@NathanLovato Sure, feel free to do it. Apologies for the late reply.

NathanLovato:
Done, thanks!

PR #4348, opened 2020-11-09
Update Rider prerequisites to match Godot 3.2.3+

van800:


tutorials/scripting/c_sharp/c_sharp_basics.rst
```diff
@@ -38,47 +38,11 @@ page in the Mono documentation.
...
-   - set **Mono Build Tool** to **JetBrains Mono**.
-
-- In Rider:
-
-   - Set **MSBuild version** to either **Bundled with Rider** or **.NET Core**.
-   - Install the **Godot support** plugin.
-
-macOS and Linux
+Prerequisites
 ~~~~~~~~~~~~~~~
```
Calinou, line 42:
```suggestion
~~~~~~~~~~~~~
```

tutorials/scripting/c_sharp/c_sharp_basics.rst
```diff
@@ -38,47 +38,11 @@ page in the Mono documentation.
...
-the version number doesn't matter since Godot bundles its own Mono 5.18
-installation. We only need the Mono installation for NuGet and MSBuild
-which are required to use C# in Godot.
-
-.. note:: To download Mono on macOS, use the "Stable Channel" link
-          from the `Mono Downloads Page <http://www.mono-project.com/download/>`_.
-          The Visual Studio channel is an earlier version of Mono and
-          will not work.
+Install the latest stable version of the `.NET Core`. 
+As of Godot 3.2.3 installing Mono is not a requirement anymore.
```
Calinou, line 45:
```suggestion
As of Godot 3.2.3, installing Mono is not a requirement anymore.
```

tutorials/scripting/c_sharp/c_sharp_basics.rst
```diff
@@ -38,47 +38,11 @@ page in the Mono documentation.
...
-`Mono SDK <http://www.mono-project.com/download/>`_. As of Godot 3.1 beta 3,
-the version number doesn't matter since Godot bundles its own Mono 5.18
-installation. We only need the Mono installation for NuGet and MSBuild
-which are required to use C# in Godot.
-
-.. note:: To download Mono on macOS, use the "Stable Channel" link
-          from the `Mono Downloads Page <http://www.mono-project.com/download/>`_.
-          The Visual Studio channel is an earlier version of Mono and
-          will not work.
+Install the latest stable version of the `.NET Core`.
```
Calinou, line 44:
```suggestion
Install the latest stable version of `.NET Core <https://dotnet.microsoft.com/download/dotnet-core>`__ (3.1 as of writing).
```

tutorials/scripting/c_sharp/c_sharp_basics.rst
```diff
@@ -38,47 +38,11 @@ page in the Mono documentation.
...
-the version number doesn't matter since Godot bundles its own Mono 5.18
-installation. We only need the Mono installation for NuGet and MSBuild
-which are required to use C# in Godot.
-
-.. note:: To download Mono on macOS, use the "Stable Channel" link
-          from the `Mono Downloads Page <http://www.mono-project.com/download/>`_.
-          The Visual Studio channel is an earlier version of Mono and
-          will not work.
+Install the latest stable version of `.NET Core <https://dotnet.microsoft.com/download/dotnet-core>`__ (3.1 as of writing).
+As of Godot 3.2.3, installing Mono is not a requirement anymore.
```
aaronfranke, line 45: Are you sure Mono is not a requirement anymore? I've encountered many situations where C# in Godot wasn't working, then I tell people to install Mono and then it works.

Why mention .NET Core? Godot doesn't use .NET Core, and I can run Godot just fine without .NET Core. Is it just that you can install either Mono or .NET Core and it will work?

Anyway, if kept, the mention of .NET Core should be updated for .NET 5.0. While the "Core" is no longer part of the branding, we may wish to include it anyway so that people know we don't mean .NET Framework.

tutorials/scripting/c_sharp/c_sharp_basics.rst
```diff
@@ -121,6 +85,28 @@ external editors:
     If you are using Visual Studio 2019, you must follow the instructions found
     in the `:ref:doc_c_sharp_configuring_vs_2019_for_debugging` section below.
 
+Windows (Visual Studio)
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Download and install the latest version of
+`Visual Studio <https://visualstudio.microsoft.com/downloads/>`_
+(*not* Visual Studio Code), which contains utilities required to use
+C# in Godot.
```
aaronfranke, line 94: Is this actually required to install Visual Studio? In my experience just Mono and VS Code work. Although I haven't tested on Windows without Visual Studio in awhile, I don't plan to uninstall Visual Studio to test this.

If the tools included in Visual Studio are required, we may want to change this text around so that "JetBrains Rider" is a sub-heading.

van800:
This should likely wait resolution of https://github.com/godotengine/godot/issues/43717

tutorials/scripting/c_sharp/c_sharp_basics.rst
van800: No, visual studio is not a must have. I have just moved it and JetBrains rider down inside 'Configuring an external editor'. Take a look at the whole file, it makes more sense now, I think.

paulloz:
IMHO a lot of this page should be re-written or the whole setup/configuration part should be in a dedicated page.

Things are described in a weird order (this is partially fixed by this PR): it goes from setting up an environment to showing a script example to performances to setting up an environment again. The whole flow seems *really* confusing to new users.
Some notes are quite poorly written (at least I feel like they are) while not giving much information to begin with, e.g.
> Built-in support for Visual Studio will be including in future versions, but in the meantime, the steps below can let you configure VS 2019 for use with Godot C# projects.

Altogether what could be a good page to direct new users to is more often than not... confusing. most of my contributions to the docs are about fixing mistakes and maintaining outdated stuff so I don't really know where to go from here, I just wanted to give my two cents as we see a lot of people struggling with installation/configuration asking for help on Discord. I know there is undergoing work about making the docs more straightforward and intelligible so maybe it'd be the right time to restructure this part too.

P.S. On a minor note, there's a broken anchor ref [here](https://docs.godotengine.org/en/stable/getting_started/scripting/c_sharp/c_sharp_basics.html#configuring-an-external-editor).

tutorials/scripting/c_sharp/c_sharp_basics.rst
van800: Suggesting everybody to use net 5 should wait for resolution of godotengine/godot#43717

tutorials/scripting/c_sharp/c_sharp_basics.rst
aaronfranke: Did you forget to push it?

van800: I don't mean I made any additional changes today, sorry for confusion.

tutorials/scripting/c_sharp/c_sharp_basics.rst
neikeq: Installed Mono isn't used by Godot anymore. It's only needed for building Godot itself.

>Why mention .NET Core? Godot doesn't use .NET Core, and I can run Godot just fine without .NET Core. Is it just that you can install either Mono or .NET Core and it will work?

The docs should say .NET Core SDK, instead of just .NET Core as users could mistakenly download the runtime-only version. The .NET Core SDK is needed for building C# projects with the newer Sdk style (even when using Visual Studio's MSBuild you need to install the .NET Core Visual Studio workload).

neikeq (approved):

tutorials/scripting/c_sharp/c_sharp_basics.rst
van800: `.NET Core` updated to `.NET Core SDK`.
Commits squashed.

aaronfranke:
I'm concerned that the PR in its current state doesn't adequately give instructions for Mac and Linux. It says that you need .NET Core SDK, and that for Windows you need Visual Studio because it "which contains utilities required to use C# in Godot". So where do Mac and Linux users get these utilities? If they're installed with .NET Core or Mono, why mention Visual Studio providing "utilities required" at all? Additionally, because of this same reason, it seems ambiguous as to whether or not the article wants JetBrains Rider users on Windows to also install Visual Studio for those utilities.

van800:
@aaronfranke Well, afaik, installing VS would remove all previously installed dotnetcore sdk and install the latest, so VS installer kinda contains everything needed. 
I don't use VSCode, so I can't add its section along with VS and Rider.

Anyway this state of documentation is a lot better than before.

aaronfranke (approved):
I think we can merge this now and then I will make a follow-up later with instructions for VS Code and adding some more details and clarification where needed.

akien-mga:
Thanks!

PR #4352, opened 2020-11-09
Add a Hosting considerations section to High-level multiplayer

Calinou:
This covers topics such as internal vs public IP addresses and port forwarding.


NathanLovato:
Thanks!

PR #4356, opened 2020-11-11
Add paragraph to specify that IK chains are editor-only

textrivers:
Much needed clarification here that IK chains are not responsive at runtime, as I assumed they would be. I would not have wasted a day trying to make them work if I'd known better. I see lots of questions on forums/Discord about IK solvers that indicate a similar misunderstanding, and I thought it'd be good to save others the trouble. I also added some small proofreading changes.

### DIFF
--- a/tutorials/animation/cutout_animation.rst
+++ b/tutorials/animation/cutout_animation.rst
@@ -236,9 +236,12 @@ position of hands, feet and other extremities of rigs like the one we've made.
 Imagine you want to pose a character's foot in a specific position on the ground.
 Without IK chains, each motion of the foot would require rotating and positioning
 several other bones (the shin and the thigh at least). This would be quite
-complex and lead to imprecise results.
+complex and lead to imprecise results. IK allows us to move the foot directly 
+while the shin and thigh self-adjust.
 
-IK allows us to move directly the foot while the shin and thigh self-adjust.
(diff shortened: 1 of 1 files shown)


tutorials/animation/cutout_animation.rst
```diff
@@ -236,9 +236,12 @@ position of hands, feet and other extremities of rigs like the one we've made.
...
 Without IK chains, each motion of the foot would require rotating and positioning
 several other bones (the shin and the thigh at least). This would be quite
-complex and lead to imprecise results.
+complex and lead to imprecise results. IK allows us to move the foot directly 
+while the shin and thigh self-adjust.
 
-IK allows us to move directly the foot while the shin and thigh self-adjust.
+It should be noted that IK chains in Godot currently work in the editor only, not
+at runtime. They are intended to ease the process of setting keyframes, and are
+not currently useful for techniques like procedural animation.
```
Calinou, line 244:
```suggestion
.. note::

    **IK chains in Godot currently work in the editor only**, not
    at runtime. They are intended to ease the process of setting keyframes, and are
    not currently useful for techniques like procedural animation. 
```

Calinou (approved):
Thanks!

PS: It might be worth checking whether https://github.com/godotengine/godot/pull/39353 makes it possible to use IK at run-time in in 4.0.

PR #4357, opened 2020-11-11
Update GDScript FAQ text to not reference history in link

rodrigosetti:
Since GDScript history was moved from scripting guide to FAQ, we want to remove the link back to guide mentioning "history".


Calinou (approved):
Thanks! Congratulations for your first merged contribution :tada:

PR #4360, opened 2020-11-12
Fix couple of missing `#` and `//` for comments in code examples

Feniks-Gaming:
Couple of code examples were missing # and // for comments after line break. This fixes it.


Calinou (approved):
Thanks!

PR #4382, opened 2020-11-18
Specify that only collada files can use -colonly with empty shapes

### LINKED ISSUE #4374: Description of -colonly import hint for empties is incomplete/outdated
**Your Godot version:** 3.2 and `master`

**Issue description:**

According to the documentation for "Importing 3D scenes" under the "Import hints" section:

> The option `-colonly` can also be used with Blender's empty objects.

It continues with some more information about these empty objects. It is not written that this applies *only* to files imported by the Collada importer which have a `empty_draw_type` section. In turn, [such files](https://github.com/search?l=COLLADA&p=1&q=empty_draw_type&type=Code) are only exported from Blender (2.79?) with the [Better Collada](https://github.com/godotengine/collada-exporter/) exporter (or forks of it). This major limitation should be clarified, or the information should be removed.

In more detail: `_fix_node` (which handles import hints) [checks](https://github.com/godotengine/godot/blob/9e7348f78852b965a628193289075e3a10f1ef85/editor/import/resource_importer_scene.cpp#L417) for the `"empty_draw_type"` meta, which is [only](https://github.com/godotengine/godot/blob/a6916d1f179f760f015f385c01729d205af97abe/editor/import/collada.cpp#L1572) [set](https://github.com/godotengine/godot/blob/00949f0c5fcc6a4f8382a4a97d5591fd9ec380f8/editor/import/editor_import_collada.cpp#L319) when such a section is found by the Collada importer, and this section will only [be there](https://github.com/godotengine/collada-exporter/blob/9a8fae951fe76209a7bb6c6cbc692478585283b8/io_scene_dae/export_dae.py#L1266) for files exported with Better Collada. (In vanilla Blender, the information about the draw type of empties is lost when exporting to Collada or glTF.)

* https://docs.godotengine.org/en/3.2/getting_started/workflow/assets/importing_scenes.html#import-hints
* https://docs.godotengine.org/en/latest/tutorials/assets_pipeline/importing_scenes.html#import-hints

skyace65:
Specifies that only collada files can use the -colonly import hint with empty shapes. Closes #4374. I was tempted to remove it since gltf 2.0 is what everyone should be using now. But as long as collada is still supported I feel like it should remain fully documented.

### DIFF
--- a/tutorials/assets_pipeline/importing_scenes.rst
+++ b/tutorials/assets_pipeline/importing_scenes.rst
@@ -413,7 +413,7 @@ This helps the visual mesh and actual collision to be separated.
 
 The option ``-convcolonly`` works in a similar way, but will create a :ref:`class_convexpolygonshape` instead.
 
-The option ``-colonly`` can also be used with Blender's empty objects.
+With Collada files the option ``-colonly`` can also be used with Blender's empty objects.
 On import, it will create a :ref:`class_staticbody` with
 a collision node as a child. The collision node will have one of a number of predefined shapes,
 depending on Blender's empty draw type:


NathanLovato:
Thanks!

rburing:
Thanks, this is an improvement, but it is still incomplete as explained in https://github.com/godotengine/godot-docs/issues/4374. Namely the correct condition is "With Collada files exported from Blender using the [Better Collada](https://github.com/godotengine/collada-exporter) exporter" (or forks of this exporter). It does not work with Collada files exported from vanilla Blender. Including the draw type of an empty in the `.dae` file (in a custom section) is something that is only done by the Better Collada exporter.

skyace65:
@rburing That page already states that you need to use the better collada exporter because the default is broken. Saying that again for the import hints is needlessly repetitive.

rburing:
Okay, then technically the information is on the page, but I still propose to repeat it for emphasis. A very common case is linking to the [Import hints](https://docs.godotengine.org/en/latest/tutorials/assets_pipeline/importing_scenes.html#import-hints) section of the page, or more specifically [Create collisions](https://docs.godotengine.org/en/latest/tutorials/assets_pipeline/importing_scenes.html#create-collisions-col-convcol-colonly-convcolonly). This skips over the earlier explanation about Collada. An unwitting user has no reason to suspect that "Collada" there does not refer to all Collada files, and will be confused when it does not work. I propose at least adding a link to the [Exporting DAE files from Blender](https://docs.godotengine.org/en/latest/tutorials/assets_pipeline/importing_scenes.html#exporting-dae-files-from-blender) section on the same page, e.g. minimally

> With [Collada](https://docs.godotengine.org/en/latest/tutorials/assets_pipeline/importing_scenes.html#exporting-dae-files-from-blender) files the suffix `-colonly` can also be used with Blender's empty objects.

or more explicitly

> With Collada files [exported from Blender using Godot's plugin](https://docs.godotengine.org/en/latest/tutorials/assets_pipeline/importing_scenes.html#exporting-dae-files-from-blender), the suffix `-colonly` can also be used with empty objects.

skyace65:
@rburing that doesn't fix the issue because not everyone is going to look at just `-colonly`. Or interpret it as the default importer being fine for the rest of the hints. And there's no good place to put that information at the top of that section. I'm going to change the collision shapes link so it doesn't skip over how to export files for Godot.

PR #4387, opened 2020-11-18
Remove Int being same as floor as they are not

### LINKED ISSUE #4358: int(x) is not equal to floor(x) for negative values
**Your Godot version:** 
3.2.3 stable

**Issue description:** 
In documentation it is said that `floor` method returns float and "If you need an integer, you can use int(s) directly". This note (or at least cited part) should be removed because this statement is incorrect for negative values in `floor` method. Casting to an int will truncate value toward zero, and `floor` downward value towards negative infinity.

Test code: 
```
	if int(-3.4) == floor(-3.4):
		print("int(-3.4) == floor(-3.4)")
	else:
		print("int(-3.4) != floor(-3.4)")
```

Output:
`int(-3.4) != floor(-3.4)`

https://docs.godotengine.org/en/stable/classes/class_@gdscript.html#class-gdscript-method-floor

Feniks-Gaming:
Int(x) is not equal to floor(x) for negative values closes #4358

### DIFF
--- a/classes/class_@gdscript.rst
+++ b/classes/class_@gdscript.rst
@@ -532,7 +532,6 @@ Rounds ``s`` to the closest smaller integer and returns it.
     # a is -3.0
     a = floor(-2.99)
 
-**Note:** This method returns a float. If you need an integer, you can use ``int(s)`` directly.
 
 ----
 


Calinou:

Feniks-Gaming:
What is a path to GDScript class refrence I checked class folder and couldn't find one. Searching for "@GDScript" also returned no result. Can you point me in a right direction and I will reopen in a right repository.

pycbouh:
https://github.com/godotengine/godot/blob/master/doc/classes/%40GlobalScope.xml#L307

Feniks-Gaming:
@pycbouh thank you kindly.

Feniks-Gaming:
https://github.com/godotengine/godot/pull/43656

PR #4388, opened 2020-11-18
Clarify that GDScript is not based on Python

### LINKED ISSUE #4378: Reduce or rephrase comparisons between GDScript and Python
*(Sorry for not following the template, it's more of a meta issue)*

There is a striking number of people believing that GDScript is very much related to Python in many of its aspects. This is a misconception that is very often used as an argument when suggesting improvements to GDScript or just general questions about it. While languages do share syntactic similarities, they are very different in practice.

Docs help fuel this confusion, as the language is introduced several times as "similar to Python" or even "based on Python" (paraphrasing, [[1]](https://docs.godotengine.org/en/stable/getting_started/scripting/gdscript/gdscript_basics.html), [[2]](https://docs.godotengine.org/en/stable/about/list_of_features.html#scripting), [[3]](https://docs.godotengine.org/en/stable/about/faq.html#what-were-the-motivations-behind-creating-gdscript), [[4]](https://docs.godotengine.org/en/stable/getting_started/step_by_step/godot_design_philosophy.html#all-inclusive-package)). This needs to be addressed either by reducing the number of such mentions or by rephrasing them as clear as possible. I understand that the intention is probably to give a point of reference for less experienced in programming, but it also invites people more experienced or more confident in it to make unnecessary connections between the languages.

Our Discord server, for one, is full of misguided people asking about those similarities daily, criticizing the language in a vein with "Well, Python can do it, why can't GDScript". Proposals about GDScript 2.0 changes are also a great source of such misconceptions, and vnen has to repeat himself a lot that these are very different languages. So I think we need to put some distance.

Calinou:
This also fixes capitalization where "GDscript" was used instead of "GDScript".

This closes #4378.

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_basics.rst
+++ b/tutorials/scripting/gdscript/gdscript_basics.rst
@@ -7,11 +7,12 @@ Introduction
 ------------
 
 *GDScript* is a high-level, dynamically typed programming language used to
-create content. It uses a syntax similar to
-`Python <https://en.wikipedia.org/wiki/Python_%28programming_language%29>`_
-(blocks are indent-based and many keywords are similar). Its goal is
-to be optimized for and tightly integrated with Godot Engine, allowing great
-flexibility for content creation and integration.
+create content. It uses an indentation-based syntax similar to languages like
+`Python <https://en.wikipedia.org/wiki/Python_%28programming_language%29>`_.
+Its goal is to be optimized for and tightly integrated with Godot Engine,
+allowing great flexibility for content creation and integration.
+
+GDScript is entirely independent from Python and is not based on it.
 
 History
 ~~~~~~~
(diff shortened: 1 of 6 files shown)


tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -13,6 +13,8 @@ create content. It uses a syntax similar to
 to be optimized for and tightly integrated with Godot Engine, allowing great
 flexibility for content creation and integration.
 
+GDScript is entirely independent from Python and is not based on it.
```
pycbouh, line 16: I think this sounds like a denial now 🙂 Maybe the paragraph itself should be rephrased instead, like you did for the Godot design philosophy? If we put "indent-based" to the first place, and move "similar to languages like Python" in the parentheses, and also remove "many keywords are similar", it will be less conflicting.

pycbouh (changes requested):

NathanLovato:
Thanks! I think the mention of Python at all has been a source of confusion over the years. The getting started rewrite will make it clear why there's a custom language, and it doesn't work like Python. It's been the mistake of many of us, me included, to say GDScript was like Python.

PR #4392, opened 2020-11-21
Fix method name in Groups

Ursus-Z555:
Change method 'call_group()' to 'add_to_group()'  under Adding Nodes to a Group section.


Calinou (approved):
Thanks!

PR #4406, opened 2020-11-25
Add a custom role to toggle between snake_case and PascalCase syntax

Calinou:
## Preview

Animation

## TODO

- [ ] See whether we can associate data attributes in Sphinx to avoid reflows during page load and have correct display when JavaScript is disabled.
- [ ] Consider using automatic conversion for inline code literals instead of doing it manually. It would be less work for documentation and class reference contributors, but I'm not sure if we can reasonably cover all use cases this way.
- [ ] Modify [`doc/tools/makerst.py`](https://github.com/godotengine/godot/blob/master/doc/tools/makerst.py) to take this role into account.
- [ ] Modify [`editor/editor_help.cpp`](https://github.com/godotengine/godot/blob/master/editor/editor_help.cpp) to understand the `[snakepascal]snake_case;;PascalCase[/snakepascal]` tag.
- [ ] Test automatic conversion rules in more real-world scenarios. It probably involves testing in production due to the amount of documentation :slightly_smiling_faces:


Feniks-Gaming:
I would suggest renaming it to "GDScript Code Examples" and "C# Code Examples" otherwise some people may be confused that those too can be used interchangeably in thier code.

Calinou:
> I would suggest renaming it to "GDScript Code Examples" and "C# Code Examples" otherwise some people may be confused that those too can be used interchangeably in thier code.

I've discussed naming the options "GDScript" and "C#", but there is no 1:1 mapping between syntax styles and languages. For instance, if you use GDNative or a C++ module, you're going to use snake_case.

Feniks-Gaming:
Maybe then add `snake_case(GDScript)` and `PascalCase`? Likely people who are using something else than GDScript are already experienced enough to figure out what coding style they should follow. This will help brand new who meet the engine for a first time to use the right style while still leaving others to use the Syntax that represents their language

Calinou:
> Maybe then add snake_case(GDScript) and PascalCase?

There's already a tooltip telling you which languages each syntax is used in. I think that's enough :slightly_smiling_face:

Feniks-Gaming:
Okay that's fair.

mhilbrunner:
Now that we updated `sphinx-tabs` in https://github.com/godotengine/godot-docs/pull/4995: the new Sphinx tabs tries to remember the active tab (GDScript vs. C#) too between page loads it seems.

Maybe we can
1. re-use that information
2. toggle this automatically whenever you switch between GDScript/C# tabs
3. maybe make this toggle those tabs automatically too

And while its indeed not a perfect or correct mapping, I'm somewhat leaning towards labelling it 'GDscript/C#' too, may be less confusing to new users while those who know C++ etc. will still likely know what to do/press?

skyace65:
Confirmed with Calinou this can be closed in RocketChat

Calinou:
For context, this needs a more reliable implementation somehow, but I don't know the details. The current approach felt too hacky to be reliable at a large scale.

PR #4407, opened 2020-11-26
Change variable for Android SDK path

remtaine:
The Gradle looks for "ANDROID_SDK_ROOT" and not "ANDROID_HOME" so I changed the path mentioned for clarity

### DIFF
--- a/development/compiling/compiling_for_android.rst
+++ b/development/compiling/compiling_for_android.rst
@@ -202,7 +202,7 @@ Troubleshooting
 Platform doesn't appear in SCons
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Double-check that you've set both the ``ANDROID_HOME`` and ``ANDROID_NDK_ROOT``
+Double-check that you've set both the ``ANDROID_SDK_ROOT`` and ``ANDROID_NDK_ROOT``
 environment variables. This is required for the platform to appear in SCons'
 list of detected platforms.
 See :ref:`Setting up the buildsystem <doc_android_setting_up_the_buildsystem>`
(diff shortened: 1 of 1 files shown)


akien-mga:
> The Gradle looks for "ANDROID_SDK_ROOT" and not "ANDROID_HOME" so I changed the path mentioned for clarity

Are you sure? Even Godot's own custom build workflow sets `ANDROID_HOME` for Gradle: https://github.com/godotengine/godot/blob/master/platform/android/export/export.cpp#L2422

m4gr3d:
@akien-mga `ANDROID_HOME` is [deprecated (but still supported for backward compatibility)](https://developer.android.com/studio/command-line/variables#envar), so it's safe to use either one.

m4gr3d (approved):

m4gr3d (changes requested):
@remtaine There's one more location to update under the `Troubleshooting` section.

m4gr3d:
@akien-mga Is there a `label` to indicate this should be cherry-picked for the `3.2` docs as well?

akien-mga:
> @akien-mga Is there a `label` to indicate this should be cherry-picked for the `3.2` docs as well?

There used to be, but then we stop using it as I was merging the `master` branch in `3.2` regularly (reverting `master`-specific changes).
Now `master` and `3.2` have diverged too much in structure though so we'll have to go back to manually identifying candidates for cherry-picking.

m4gr3d:
The updated file seems to have the same structure in both `master` and `3.2` so it may be possible to cherry-pick here.
Otherwise @remtaine you may need to provide a similar backport for `3.2`.

remtaine:
> @remtaine There's one more location to update under the `Troubleshooting` section.

Okay! I changed that one as well now

akien-mga:

PR #4410, opened 2020-11-27
Rename the Locales pages to Locale codes for better SEO

Calinou:
I had trouble finding this page using the built-in documentation search before. This should make it easier to find, including in third-party search engines.


NathanLovato:
Thanks!

PR #4419, opened 2020-12-01
Add initial integration for questions & answers in relevant pages

Calinou:
[Alpine.js](https://github.com/alpinejs/alpine) is used on the frontend.

The backend is a static REST API generated by https://github.com/Calinou/godot-qa-api-generator.

Pages should have a `.. questions-answers:: tag1 tag2` directive at the bottom when relevant. I've edited some pages to add such tags, but we should probably go over all pages in a future PR.

The API should allow CORS requests (`*`) since the JSON files will be hosted on a different domain (`docs.godotengine.org`/`godot.readthedocs.io` to `godotengine.org`).

Questions are sorted by popularity (highest score first) and the top 10 questions are displayed on relevant pages. The *"(from Godot Q&A)"** link can be clicked to view all questions.

This closes https://github.com/godotengine/godot-proposals/issues/481.

### Light theme

Q&A integration

### Dark theme

Q&A integration


seppoday:
Will we merge this soonish?

Calinou:
> Will we merge this soonish?

reduz wanted to do this differently (with an invisible tag per page), so probably not.

Calinou:
- Superseded by https://github.com/godotengine/godot-docs/pull/7520, which is a different approach we've discussed during Godot Sprint 2023 in Valencia.

Unlike this PR, the approach linked above doesn't rely on the slow Question2Answer backend and doesn't require hosting a dedicated service to expose a REST API. It also goes for a php.net-style "user notes" as opposed to questions & answers. User notes displayed at the bottom of a manual or class reference page are usually more suited to bring clarifications, rather than asking support questions.

PR #4421, opened 2020-12-02
Add a comment style guide for Godot contributions

Calinou:
We have recently (informally) settled on a comment style guide. I think now is a good time to formalize it for newer contributors :slightly_smiling_face:

### DIFF
--- a/community/contributing/code_style_guidelines.rst
+++ b/community/contributing/code_style_guidelines.rst
@@ -303,3 +303,63 @@ Editor integration
 Many IDEs or code editors have beautifier plugins that can be configured to run
 black automatically, for example each time you save a file. For details you can
 check `Black editor integration <https://github.com/psf/black#editor-integration>`__.
+
+Comment style guide
+-------------------
+
+This comment style guide applies to all programming languages used within
+Godot's codebase.
+
+- Begin comments with a space character to distinguish them from disabled code.
+- Use sentence case for comments. Begin comments with an uppercase character and
+  always end them with a period.
+- Reference variable/function names and values using backticks.
+- Wrap comments to ~100 characters.
+- You can use ``TODO:``, ``FIXME:``, ``NOTE:``, or ``HACK:`` as adominitions
+  when needed.
+
+**Example:**
+
+.. code-block:: cpp
+
+    // Compute the first 10,000 decimals of Pi.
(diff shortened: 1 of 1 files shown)


community/contributing/code_style_guidelines.rst
```diff
@@ -301,3 +301,61 @@ Editor integration
...
+
+    /**
+     * Returns the number of nodes in the universe.
+     * This can potentially be a very large number, hence the 64-bit return type.
+     */
+    uint64_t Universe::get_node_count() {
+        // ...
+    }
+
+For member variables, don't use Javasoc-style comments but use single-line comments instead:
```
Calinou, line 352: I think it would be better to use single-line Javadoc-style comments for those (`/** Like this. */`), but since the vast majority of the codebase doesn't, it's better to stay consistent.

What do you think?

community/contributing/code_style_guidelines.rst
```diff
@@ -301,3 +301,62 @@ Editor integration
...
 black automatically, for example each time you save a file. For details you can
 check `Black editor integration <https://github.com/psf/black#editor-integration>`__.
+
+Comment style guide
+-------------------
+
+This comment style guide applies to all programming languages used within
+Godot's codebase.
+
+- Use sentence case for comments. Begin comments with an uppercase character and
```
aaronfranke, line 311: We can add another thing to this list.

```suggestion
- Begin comments with a space character to distinguish them from disabled code.
- Use sentence case for comments. Begin comments with an uppercase character and
```

community/contributing/code_style_guidelines.rst
```diff
@@ -301,3 +301,62 @@ Editor integration
...
+  always end them with a period.
+- Reference variable/function names and values using backticks.
+- Wrap comments to ~100 characters.
```
aaronfranke, line 314: This is just my opinion, I think we should allow up to 100 but if there's going to be wrapping we can do so to 80 characters.

```suggestion
- Wrap comments to 80 characters if the line would exceed 100 characters.
```

Calinou: I'm not sure about it. Wrapping to 80 characters in C++ can make for difficult-to-read comments if they're heavily indented.

Calinou:

Issue #4409, opened 2020-11-27
Document conversion from res:// or user:// to global system paths (ProjectSettings.globalize_path)


PureTryOut:
*Moved by @akien-mga from engine repository, as it's not a bug but a sign of incomplete documentation.*

There should be documentation added to https://docs.godotengine.org/en/stable/tutorials/io/data_paths.html about this (both `globalize_path` and `localize_path`).

The documentation for the actual methods in ProjectSettings could also use some extra information (doesn't mention `user://`, don't give a clear idea of what the result could be). https://docs.godotengine.org/en/stable/classes/class_projectsettings.html#class-projectsettings-method-globalize-path

----

**Godot version:** 3.2.3 from Alpine Linux repositories

**OS/device including version:** Alpine Linux edge

**Issue description:**
I'm messing around with the C#/Mono bindings and I'm currently using a library that requires me to pass it an absolute path. However, it seems there is currently no way to convert a Godot path (for example `cache://` or `user://`) to real paths so non-Godot scripts can work with it. I've seen no way to do it in the documentation and asking around on the Godot Matrix channel indicated it's just not possible.

I expected the string [GetBaseDir()](https://docs.godotengine.org/en/stable/classes/class_string.html#class-string-method-get-base-dir) function to be able to do it, but it still just returns `cache://` rather than the full resolved path.

PureTryOut:
Ah I guess I was confused with some different path then. But yes, I couldn't find it in the docs (and neither could some other people), so docs could use some improvement :sweat_smile: 

Thanks for the pointer, that will do!

Calinou:
Reopening per https://github.com/godotengine/godot/pull/43922#issuecomment-735192282.

fire:
Whoops, why does it do this.

PR #4439, opened 2020-12-06
Mention the ProjectSettings globalize/localize methods in Data paths

Calinou:
This closes https://github.com/godotengine/godot-docs/issues/4409.


Feniks-Gaming (changes requested):
Could we explain what is a benefit in doing so and give example of what will the path look before and after to make it easier to understand. For example I have no idea why I would want to do that.

Calinou:
@Feniks-Gaming Done, I added an example use case for `ProjectSettings.globalize_path()`. I can't think of an use case for `ProjectSettings.localize_path()` though (at least right now).

Feniks-Gaming (approved):

akien-mga (approved):

akien-mga:
Thanks!

PR #4444, opened 2020-12-07
Mention BornCG's 2D tutorials in Tutorials and resources

willnationsdev:
Apparently, BornCG recently started a 2D platformer series on his channel, so I'm adding 2D to the list of stuff he covers. Also, there was some inconsistency in how 2D and 3D was presented for Gonkee, so I adjusted it to be the same as everyone else.


Calinou (approved):
Thanks!

Feniks-Gaming (approved):

PR #4448, opened 2020-12-09
Small GDScript pow fix

### LINKED ISSUE #4445: A small correction for the GDScript.pow()
**Your Godot version:**
3.2.3.official.stable

**Issue description:**
Currently:
> float pow ( float base, float exp )
> Returns the result of x raised to the power of y.

Arguments' names in description ("x" and "y") and in code ("base" and "exp") are different.
Suggestion: replace "x" and "y" in description with "base" and "exp" respectively.

https://docs.godotengine.org/en/stable/classes/class_@gdscript.html?#class-gdscript-method-pow

skyace65:
Replaced `x` and `y` with `base` and `exp` since that's what the example code uses. Closes #4445.

### DIFF
--- a/classes/class_@gdscript.rst
+++ b/classes/class_@gdscript.rst
@@ -961,7 +961,7 @@ Produces:
 
 - :ref:`float<class_float>` **pow** **(** :ref:`float<class_float>` base, :ref:`float<class_float>` exp **)**
 
-Returns the result of ``x`` raised to the power of ``y``.
+Returns the result of ``base`` raised to the power of ``exp``.
 
 ::
 


classes/class_@gdscript.rst
```diff
@@ -961,7 +961,7 @@ Produces:
 
 - :ref:`float<class_float>` **pow** **(** :ref:`float<class_float>` base, :ref:`float<class_float>` exp **)**
 
-Returns the result of ``x`` raised to the power of ``y``.
+Returns the result of ``base`` raised to the power of ``exp``.
```
Feniks-Gaming, line 964: If we are fixing it can we add example code something like 

    Pow(2,3) #2 to power of 3 returns 8

Feniks-Gaming (approved):

Calinou:

skyace65:
@Calinou The file I'm editing doesn't exist in the godot repository, only in the docs repository. And I checked the latest build of the docs and the page is still being built so I'm assuming it wasn't meant to be removed from the docs completely.

Calinou:
@skyace65 It's here: https://github.com/godotengine/godot/blob/master/modules/gdscript/doc_classes/@GDScript.xml#L758-L771

It's part of a module, which means the XML is in `modules/{name}/doc_classes` instead of `doc/classes`.

skyace65:
@Calinou got it, thanks!

PR #4464, opened 2020-12-15
Fix Hide() in _ready() first player example code

MattObs:
Hide() on line 216 and 223 give an unintended result when the user following the tutorial is requested on line 354 to Play the Scene. The Player will not be shown.
Removing Hide() on line 216 and 223 will fix this.

### DIFF
--- a/getting_started/step_by_step/your_first_game.rst
+++ b/getting_started/step_by_step/your_first_game.rst
@@ -213,14 +213,12 @@ which is a good time to find the size of the game window:
 
     func _ready():
         screen_size = get_viewport_rect().size
-        hide() # Start the player hidden.
 
  .. code-tab:: csharp
 
     public override void _Ready()
     {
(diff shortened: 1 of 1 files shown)


NathanLovato:
This isn't a tutorial I worked on, but at a glance you shouldn't only remove this function call; instead, it needs to move to another place later in the tutorial.

Because everything is in one scene in this demo, you need to hide the player at the start of the game.

MattObs:
> This isn't a tutorial I worked on, but at a glance you shouldn't only remove this function call; instead, it needs to move to another place later in the tutorial.
> 
> Because everything is in one scene in this demo, you need to hide the player at the start of the game.

It is mentioned at line 436 to add Hide() to _ready() after the movement has been tested.

NathanLovato:
Great, I just double-checked, indeed! Thanks for the contribution.

PR #4475, opened 2020-12-22
Update class_array.rst

Laguzus:
Updating description for hash()-method. Strictly speaking it does not "Returns a hashed integer value **representing the array contents.**". This information is incomplete and therefore misleading.
It "Returns a hashed integer value representing the array **and** its contents." which also contains the special case, where two different arrays contain exactly the same values but still yield different hash result.

User could be tempted to believe, that hash() yields same values for same array-contents. 
(which I was tempted to. Testing showed **different hashes** for different arrays where values of the arrays where exactly the same)


Calinou:

PR #4477, opened 2020-12-23
Update class_array.rst hash()-description

Laguzus:
Strictly speaking hash() does not return a hashed integer value representing the array contents, rather "array **and** it's contents".

For the original description it would mean that two different arrays A = [0,1] and B = [0,1] produce the same hashed integer value.
This is in fact not true, rather hash() includes not only the array content, but also the array referencial value itself.
This leads to A and B producing different hashes even though they contain the same values.

Hashing A and A yields the same hash though.


Calinou:

PR #4482, opened 2020-12-25
Allow Dependabot to check Python dependencies

jlosito:
This change allows dependabot to check any python dependency which this project uses stated within the requirements.txt file. Dependabot will submit PRs with version bumps.

https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates


Calinou (approved):
Thanks!

PR #4507, opened 2020-12-30
Update import hints documentation

Shatur:
* Add 4 new import hints for shapes from the https://github.com/godotengine/godot/pull/39105.
* Add mising material hints.

Can be merged only after https://github.com/godotengine/godot/pull/39105.


Shatur95:
The proposal was rejected, closing.

Shatur95:
> Add mising material hints.

But not sure what to do with material hints. They are exist in the code, but have an issue with apply order. I asked in the chat what is better to do with them - fix and document or remove from the code, but no one answered me).

PR #4511, opened 2021-01-03
your_first_game: Rust example to define player

Kreyren:
Stripped the example from https://github.com/godot-rust/godot-rust/blob/master/examples/dodge_the_creeps/src/player.rs#L10 to look the same as examples from presented languages.

Should be peer-reviewed by godot-rust for accuracy

Signed-off-by: Jacob Hrbek <kreyren@fsfe.org>


Calinou:
Closing per https://github.com/godotengine/godot-docs/pull/4512#issuecomment-753570079.

PR #4512, opened 2021-01-03
your_first_game: Rust example for screen_size

Kreyren:
Grabbed this from https://github.com/godot-rust/godot-rust/blob/master/examples/dodge_the_creeps/src/player.rs#L34

To fit the examples in other languages this should be used:

```rs
impl Player {
  #[export]
    fn _ready(&mut self, owner: &Area2D) {
      self.screen_size = viewport.size();
    }
}
```

But in rustlang we don't have access to 'viewport' from this scope:

```
error[E0425]: cannot find value `viewport` in this scope
  --> examples/dodge_the_creeps/src/player.rs:36:28
   |
36 |         self.screen_size = viewport.size();
   |                            ^^^^^^^^ not found in this scope
```

Thus it requires use of unsafe code.

Should be peer-reviewed by godot-rust, provided as-is to the best of my
ability

Signed-off-by: Jacob Hrbek <kreyren@fsfe.org>

### DIFF
--- a/getting_started/step_by_step/your_first_game.rst
+++ b/getting_started/step_by_step/your_first_game.rst
@@ -221,6 +221,18 @@ which is a good time to find the size of the game window:
         screenSize = GetViewportRect().Size;
     }
 
+ .. code-tab:: rust
+
+    #[methods]
+    impl Player {
+        #[export]
+        fn _ready(&mut self, owner: &Area2D) {
+            let viewport = unsafe { owner.get_viewport().unwrap().assume_safe() };
+            self.screen_size = viewport.size();
+            owner.hide();
+        }
+    }
+
 Now we can use the ``_process()`` function to define what the player will do.
 ``_process()`` is called every frame, so we'll use it to update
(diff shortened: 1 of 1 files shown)


Calinou:
Thanks for opening a pull request :slightly_smiling_face: 

The godot-rust bindings are not official; they are maintained by the community. Therefore, we don't accept code samples for them in the official documentation. We only accept GDScript and C# code samples since these are officially maintained.

I recommend you open a pull request on the godot-rust documentation to clarify this part of the tutorial. In the long run, it would be better to have a dedicated Rust version of the "Your first game" tutorial in the godot-rust documentation
.

Kreyren:
Care to elaborate why is godot-rust not considered official?

Calinou:
> Care to elaborate why is godot-rust not considered official?

We officially support GDScript, C#, visual scripting and C++ (via modules or GDNative). That makes 4 languages, which is a lot to support already :slightly_smiling_face: 

We haven't even been able to give the C++ bindings proper love yet.

PR #4535, opened 2021-01-12
Add a page referencing the Control gallery image and demo

### LINKED ISSUE #4435: Add a page visually referencing all Control nodes
**Your Godot version:** 3.2.3

**Issue description:**

It's a common source of confusion for new users to be unaware of which Control node they should use to create a given GUI element. Nodes like SpinBox are especially difficult to find for people not used to GUI development.

We should have a page featuring something similar to a "control gallery" where individual GUI elements are represented on an image, outlined and labeled. Something like this (but with nodes outlined and labeled):

image

The project used to generate this control gallery could also be added to https://github.com/godotengine/godot-demo-projects, since it would be a good showcase for GUI node usage in general.

**Edit:** Demo PR opened: https://github.com/godotengine/godot-demo-projects/pull/555

See https://wolthera.info/2020/12/xmpp-client-in-godot/.

Calinou:
This closes #4435.


akien-mga:
Thanks!

PR #4537, opened 2021-01-13
GitHub Actions: Upload ZIPs of compiled HTML documentation for reading

Calinou:
This can be used to read the documentation without an Internet connection.

Builds are made every week and are kept for 15 days. This ensures an artifact is still available in case a scheduled build fails.

**Example:** https://github.com/Calinou/godot-docs/actions/runs/481583437
Note that the size displayed on the artifact download is the *uncompressed* size. The actual download size is much lower (~200 MB) thanks to compression.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,11 @@
 name: Continuous integration
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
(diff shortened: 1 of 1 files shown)


akien-mga (approved):

akien-mga:
Thanks!

PR #4538, opened 2021-01-13
Link to DevDocs on the home page for offline browsing

Calinou:
This is another way to browse the documentation offline. Unlike HTML builds, DevDocs can be used on mobile devices.

See also #4537.


pycbouh (approved):

akien-mga:
Looks like all internal links in DevDocs still point to docs.godotengine.org, is it something that we can help them fix? It's annoying not to be able to use internal links to navigate the docs.

PR #4539, opened 2021-01-13
update constant to match style guidelines

chumoo:
The official style guidelines recommend upper case for constants.

### DIFF
--- a/tutorials/physics/ray-casting.rst
+++ b/tutorials/physics/ray-casting.rst
@@ -245,13 +245,13 @@ To obtain it using a camera, the following code can be used:
 .. tabs::
  .. code-tab:: gdscript GDScript
 
-    const ray_length = 1000
+    const RAY_LENGTH = 1000
 
     func _input(event):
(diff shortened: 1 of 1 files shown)


Calinou:
@aaronfranke Should this change also be made to the C# code sample?

chumoo:
> @aaronfranke Should this change also be made to the C# code sample?

I'm not familiar with C# but the [Godot C# style guide](https://docs.godotengine.org/en/stable/getting_started/scripting/c_sharp/c_sharp_style_guide.html) says that camelCase with an underscore prefix should be used for private fields.

Perhaps "_rayLength" instead of "rayLength" should be used in the C# code sample?

aaronfranke:
The C# style guide says to use PascalCase for all namespaces, type names, methods, properties, constants, events, and camelCase with an underscore for private fields. It doesn't explicitly include an example of a private constant, but it seems like the recommendation is PascalCase. The demos also use PascalCase for private constants.

PR #4551, opened 2021-01-17
Add link in tutorial on how to use C# in Godot similarly to that of GDScript

fcole90:
Add link in the tutorial on how to use C# in Godot similarly to that of GDScript. I think that's useful given that examples in C# are provided.

### DIFF
--- a/getting_started/step_by_step/your_first_game.rst
+++ b/getting_started/step_by_step/your_first_game.rst
@@ -170,7 +170,7 @@ click "Create":
           :ref:`doc_scripting` before continuing.
           
 .. note:: If this is your first time encountering C# in Godot, please read
-          `C# basics <https://docs.godotengine.org/en/stable/getting_started/scripting/c_sharp/c_sharp_basics.html`_. before continuing.
+          `C# basics <https://docs.godotengine.org/en/stable/getting_started/scripting/c_sharp/c_sharp_basics.html>`_. before continuing.
 
 Start by declaring the member variables this object will need:
 
--- a/getting_started/step_by_step/your_first_game.rst
+++ b/getting_started/step_by_step/your_first_game.rst
@@ -13,7 +13,9 @@ a project, and how to build a 2D game.
 .. note:: This project is an introduction to the Godot engine. It
           assumes that you have some programming experience already. If
           you're new to programming entirely, you should start here:
-          :ref:`doc_scripting`.
+          :ref:`doc_scripting`. If you want to use C# in Godot please read also
+          `C# basics <https://docs.godotengine.org/en/stable/getting_started/scripting/c_sharp/c_sharp_basics.html>`_ 
+          before continuing.
 
 The game is called "Dodge the Creeps!". Your character must move and
 avoid the enemies for as long as possible. Here is a preview of the
@@ -165,12 +167,6 @@ click "Create":
             language from the `language` drop down menu before hitting create.
 
 .. image:: img/attach_node_window.png
-
-.. note:: If this is your first time encountering GDScript, please read
-          :ref:`doc_scripting` before continuing.
-          
-.. note:: If this is your first time encountering C# in Godot, please read
-          `C# basics <https://docs.godotengine.org/en/stable/getting_started/scripting/c_sharp/c_sharp_basics.html>`_ before continuing.
 
 Start by declaring the member variables this object will need:
 
--- a/getting_started/step_by_step/your_first_game.rst
+++ b/getting_started/step_by_step/your_first_game.rst
@@ -13,8 +13,8 @@ a project, and how to build a 2D game.
 .. note:: This project is an introduction to the Godot engine. It
           assumes that you have some programming experience already. If
           you're new to programming entirely, you should start here:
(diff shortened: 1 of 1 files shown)


getting_started/step_by_step/your_first_game.rst
```diff
@@ -168,6 +168,9 @@ click "Create":
 
 .. note:: If this is your first time encountering GDScript, please read
           :ref:`doc_scripting` before continuing.
+          
+.. note:: If this is your first time encountering C# in Godot, please read
+          `C# basics <https://docs.godotengine.org/en/stable/getting_started/scripting/c_sharp/c_sharp_basics.html>`_ before continuing.
```
Calinou, line 173: Use an internal reference rather than an external one. This way, the version will match the current documentation version and language automatically:

```suggestion
          :ref:`doc_c_sharp` before continuing.
```

Calinou (changes requested):

skyace65:
I think you should just remove that note completely. There's already a note at the top of the page linking to the scripting page.

If others disagree and want the second note to stay, don't make a separate note for C#, just change the existing one to say "If this is your first time encountering programming" since the scripting page covers C# as well.

pycbouh:
I think I agree that the note here is redundant and we can instead extend the note at the top to include some reference to the C# doc.

fcole90:
The note in the beginning is worded in a way that seems to be directed to people who have no experience in programming at all. The note I'm proposing to add has instead a different purpose, to let people who know how to program already, and possibly know how to use C# as well, how to use it in Godot. This is similar to the note related to GDScript (just before my suggestion), which is directed to people who already know how to program but don't know GDScript specifically, if I understood the aim correctly. 

Perhaps, however, creating a new note entirely could be excessive. Maybe it could be better to just extend the note on GDScript to also include few words and the links

pycbouh:
> This is similar to the note related to GDScript (just before my suggestion), which is directed to people who already know how to program but don't know GDScript specifically, if I understood the aim correctly.

Yes, and there is a note at the beginning linking to the same place as that note about GDScript. So there are two notes, with different context, both linking to the same place. I think we can remove the note about GDScript in favor of the note at the beginning and mention the C# doc there as well.

fcole90:
@pycbouh I updated the proposal as you suggested

getting_started/step_by_step/your_first_game.rst
fcole90: Thanks, I updated as such in the latest commit. I actually wanted to use an internal ref from the beginning but I didn't know what name did I need to use, so resorted to an external one :sweat_smile: Is there any document that lists such internal references?

fcole90: Does ``:ref:`doc_c_sharp` `` reference to [C# basics](https://docs.godotengine.org/en/stable/getting_started/scripting/c_sharp/c_sharp_basics.html) or to the [C# index](https://docs.godotengine.org/en/stable/getting_started/scripting/c_sharp/index.html) page?
My proposal was to have the link pointing to the [C# basics](https://docs.godotengine.org/en/stable/getting_started/scripting/c_sharp/c_sharp_basics.html) as it contains details on how to set up C# in Godot, which I find useful to have a smoother experience going forward

skyace65:
That should say "please also read" instead of "please read also"

fcole90:
Sorry for duplicating the question, but maybe it went unnoticed and it'd be better to check this before anything gets merged :blush: 

As I had modified the link from using the external link syntax to the internal ref, can someone check if  `` :ref:`doc_c_sharp` `` references to the [C# basics](https://docs.godotengine.org/en/stable/getting_started/scripting/c_sharp/c_sharp_basics.html) or to the [C# index](https://docs.godotengine.org/en/stable/getting_started/scripting/c_sharp/index.html) page?
My proposal was to have the link pointing to the [C# basics](https://docs.godotengine.org/en/stable/getting_started/scripting/c_sharp/c_sharp_basics.html) page, as it contains details on how to set up C# in Godot, which I find useful to have a smoother experience going forward

pycbouh:
@fcole90 It references C# Basics. It's not a hidden piece of information either, you can verify it yourself 🙂 Internal IDs are defined at the top of each file: https://raw.githubusercontent.com/godotengine/godot-docs/master/tutorials/scripting/c_sharp/c_sharp_basics.rst

```rst
.. _doc_c_sharp:
```

pycbouh (approved):
I think we can merge it like that, unless there are other suggestions about phrasing or grammar. @Calinou ?

fcole90:
> @fcole90 It references C# Basics. It's not a hidden piece of information either, you can verify it yourself  Internal IDs are defined at the top of each file: https://raw.githubusercontent.com/godotengine/godot-docs/master/tutorials/scripting/c_sharp/c_sharp_basics.rst
> 
> ```
> .. _doc_c_sharp:
> ```

Thanks a lot, I was looking for such a piece of information but could not find it so far :sweat_smile:

fcole90:
Please, let me know if any further edit is needed

pycbouh:
Superseded by https://github.com/godotengine/godot-docs/pull/5259 due to the changes in the master branch.

PR #4552, opened 2021-01-17
Fix `:ref:` link typo in C# basics

fcole90:
The note `If you are using Visual Studio 2019, you must follow the instructions found in the :ref:doc_c_sharp_configuring_vs_2019_for_debugging section below` appears with the link as text instead than a proper `<a href="...">` element because of a misplaced backtick


Calinou (approved):
Thanks!

PR #4553, opened 2021-01-17
Add C# `using` dependencies to Your first game

fcole90:
I was trying to follow the tutorial and when beginning scripting I found that it was failing and not finding the required dependencies. That's because I was removing the script content and replacing it with what was provided, but unfortunately it didn't include such dependencies. I think adding them is important to not get stuck at this point, since it's the first script in the beginner's tutorial, so it might be the first encounter with that.


Calinou (approved):
Thanks!

PR #4559, opened 2021-01-19
Fix `emit_signal()` call in GDScript basics' coroutine example

HiImKarl:
The first argument of emit_signal should be the signal name.


Calinou (approved):
Thanks!

PR #4569, opened 2021-01-23
Add MP3 to the list of supported audio formats

YeldhamDev:


Calinou (approved):
Thanks!

This pull request should also be made for the `3.2` branch since MP3 support was cherry-picked there.

PR #4572, opened 2021-01-24
Update Sphinx to 3.5.2, sphinx_rtd_theme to 0.5.1 and sphinx-tabs to 2.1.0

HaSa1002:
Updating sphinx-tabs gives us the ability to have collapsible code tabs.
See https://github.com/executablebooks/sphinx-tabs/pull/94 and https://github.com/executablebooks/sphinx-tabs/pull/73.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@
 # Sync with readthedocs:
 # https://github.com/readthedocs/readthedocs.org/blob/master/requirements/pip.txt
 # https://github.com/readthedocs/readthedocs.org/blob/master/requirements/local-docs-build.txt
-sphinx==1.8.5  # pyup: <2.0.0
+sphinx==3.4.3  # pyup: <2.0.0
 sphinx_rtd_theme==0.4.3
 
 # Code tabs extension for GDScript/C#
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,11 +3,11 @@
 # Sync with readthedocs:
 # https://github.com/readthedocs/readthedocs.org/blob/master/requirements/pip.txt
 # https://github.com/readthedocs/readthedocs.org/blob/master/requirements/local-docs-build.txt
-sphinx==3.4.3  # pyup: <2.0.0
-sphinx_rtd_theme==0.4.3
+sphinx==3.5.2
+sphinx_rtd_theme==0.5.1
 
 # Code tabs extension for GDScript/C#
-sphinx-tabs==2.0.0
+sphinx-tabs==2.1.0
 
 # Full-page search UI for RTD: https://readthedocs-sphinx-search.readthedocs.io
 readthedocs-sphinx-search==0.1.0rc3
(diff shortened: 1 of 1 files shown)


HaSa1002:
I don't understand what the problem is with CI. It builds fine locally. 🤔

Calinou:
The Install dependencies step fails silently with the following error message:

```
ERROR: sphinx-tabs 2.0.0 has requirement sphinx<4,>=2, but you'll have sphinx 1.8.5 which is incompatible.
```

Can we use a newer Sphinx version on Read the Docs? We've intentionally pinned the Sphinx version to an old version to match the one used on Read the Docs.

HaSa1002:
I don't know how to interpret the info: https://docs.readthedocs.io/en/stable/intro/import-guide.html#building-your-documentation and https://docs.readthedocs.io/en/stable/guides/specifying-dependencies.html#specifying-dependencies. It appears that you can use a different version.

HaSa1002:
I updated sphinx as well as seperate commit.

requirements.txt
```diff
@@ -3,7 +3,7 @@
 # Sync with readthedocs:
 # https://github.com/readthedocs/readthedocs.org/blob/master/requirements/pip.txt
 # https://github.com/readthedocs/readthedocs.org/blob/master/requirements/local-docs-build.txt
-sphinx==1.8.5  # pyup: <2.0.0
+sphinx==3.4.3  # pyup: <2.0.0
```
akien-mga, line 6: The `pyup` comment is no longer relevant, is it?

We should likely also update sphinx_rtd_theme if we update sphinx, as IIRC the reason for pinning it (taken in upstream RTD) was that the theme was not compatible yet with Sphinx 2+.

HaSa1002: I dont know what it wants to state in the first place. ¯\_(ツ)_/¯

akien-mga: Well `pyup` is a tool that updates python dependencies. `pyup: <2.0.0` very likely means "hey pyup, feel free to bump minor/patch version but don't go over `2.0.0`"

requirements.txt
```diff
@@ -3,11 +3,11 @@
...
```
Calinou, line 6:
```suggestion
sphinx==3.4.3
```

requirements.txt
```diff
@@ -3,11 +3,11 @@
...
 sphinx_rtd_theme==0.4.3
 
 # Code tabs extension for GDScript/C#
-sphinx-tabs==1.1.13
+sphinx-tabs==2.0.0
```
Calinou, line 10:
```suggestion
sphinx-tabs==2.1.0
```

requirements.txt
```diff
@@ -3,11 +3,11 @@
 # Sync with readthedocs:
 # https://github.com/readthedocs/readthedocs.org/blob/master/requirements/pip.txt
 # https://github.com/readthedocs/readthedocs.org/blob/master/requirements/local-docs-build.txt
-sphinx==1.8.5  # pyup: <2.0.0
+sphinx==3.4.3  # pyup: <2.0.0
 sphinx_rtd_theme==0.4.3
```
Calinou, line 7:
```suggestion
sphinx_rtd_theme==0.5.1
```

Calinou (approved):
Thanks!

akien-mga:
Cherry-picked for 3.2.

PR #4577, opened 2021-01-25
3D game: Fix translation/rotation values to be the same as in images

viksl:
Quick fix to values for translation and rotation since the images show the values in reverse.
(In the picture translation is 0.2 and rotation 0.1)


pycbouh (approved):

Calinou (approved):
Thanks!

PR #4578, opened 2021-01-25
3D Game: Fix text to reflect the context images for collision masks

viksl:
The image in the instructions has an empty collision mask while the text wasn't reflecting it or in this case even making sense.
Maybe it was just a copy paste typoo or was there supposed to be said something different?

### DIFF
--- a/getting_started/first_3d_game/06.jump_and_squash.rst
+++ b/getting_started/first_3d_game/06.jump_and_squash.rst
@@ -85,8 +85,8 @@ Select the *Player* node and set its *Collision -> Mask* to both "enemies" and
 Then, open the *Mob* scene by double-clicking on ``Mob.tscn`` and select the
 *Mob* node.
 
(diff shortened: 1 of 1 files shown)


getting_started/first_3d_game/06.jump_and_squash.rst
```diff
@@ -85,8 +85,8 @@ Select the *Player* node and set its *Collision -> Mask* to both "enemies" and
 Then, open the *Mob* scene by double-clicking on ``Mob.tscn`` and select the
 *Mob* node.
 
-Set its *Collision -> Layer* to "enemies" and its *Collision -> Mask* to both
-"player".
+Set its *Collision -> Layer* to "enemies" and unset its *Collision -> Mask* 
+leaving the mask empty.
```
Calinou, line 89:
```suggestion
Set its *Collision -> Layer* to "enemies" and unset its *Collision -> Mask*,
leaving the mask empty.
```

Calinou (approved):
Thanks!

PR #4580, opened 2021-01-25
Declaration of HTML Version

MrBrain295:
Declares that the document is HTML5.


Calinou:
For future reference, this is not required as Sphinx (and the theme) already handles this on its own.

image

PR #4581, opened 2021-01-25
4 to four in list_of_features.rst

MrBrain295:
Any number under 10 is spelled instead of putting the number.

### DIFF
--- a/about/list_of_features.rst
+++ b/about/list_of_features.rst
@@ -153,7 +153,7 @@ Vulkan renderer.
 
 **Real-time lighting:**
 
-- Directional lights (sun/moon). Up to 4 per scene.
+- Directional lights (sun/moon). Up to four per scene.
 - Omnidirectional lights.
 - Spot lights with adjustable cone angle and attenuation.
 - Adjustable light "size" for fake area lights (will also make shadows blurrier).


Calinou:
> Any number under 10 is spelled instead of putting the number.

Where did you see this recommendation? I'm not convinced that it helps with readability.

MrBrain295:
I learned that from [https://www.scribbr.com/academic-writing/numbers/](url) but, I do see your point that numbers stand out.

aaronfranke:
Numbers also increase readability for anyone whose native language is not English. For example, I speak some French. While I can read "un, deux, trois, quatre, cinq, six, sept, huit, neuf, dix", it's a lot less readable than "1, 2, 3, 4, 5, 6, 7, 8, 9, 10".

I think this PR is a net negative. However, it would be a good idea to define an explicit style. If @NathanLovato approves, I suggest we make a note of this in the [Docs writing guidelines](https://docs.godotengine.org/en/latest/community/contributing/docs_writing_guidelines.html) page. We could amend this PR with that change.

NathanLovato:
Whichever guideline is fine, and the argument that numeric characters stand out makes sense.

In other languages, like in Japanese, I'd say numeric characters stand out even more because there are no spaces between words.

PR #4598, opened 2021-01-29
Fix collision layer math formula in Physics introduction

Guang1234567:


Calinou:
@Guang1234567 Out of curiosity, why did you make this change? Is the previous code incorrect?

pycbouh:
The commented lines above seem to suggest this formula.

Guang1234567:
@Calinou 

Q:
> @Guang1234567 Out of curiosity, why did you make this change? Is the previous code incorrect?

A:
> The commented lines above seem to suggest this formula.
> by @pyacier 

Thanks.

pyacier:
Maybe there is an error, as far as I know I'm not the author of this part, at most I translated it English to French.

PR #4600, opened 2021-01-30
Document false positives when using Embed Pck on Windows

Calinou:
This closes https://github.com/godotengine/godot/issues/45563.

### DIFF
--- a/tutorials/export/exporting_for_pc.rst
+++ b/tutorials/export/exporting_for_pc.rst
@@ -16,4 +16,9 @@ does not contain the editor and debugger.
 .. warning::
 
     If you export for Windows with embedded PCK files, you will not be able to
-    sign the program, it will break.
+    sign the program as it will break.
+
+    On Windows, PCK embedding is also known to cause false positives in
+    antivirus programs. Therefore, it's recommended to avoid using it unless
+    you're distributing your project via Steam as it bypasses code signing and
+    antivirus checks.


tutorials/export/exporting_for_pc.rst
```diff
@@ -16,4 +16,9 @@ does not contain the editor and debugger.
 .. warning::
 
     If you export for Windows with embedded PCK files, you will not be able to
-    sign the program, it will break.
+    sign the program as it will break.
+
+    On Windows, PCK embedding is also known to cause false positives in
+    antivirus programs. Therefore, it's recommended to avoid using it unless
+    you're distributing your project via Steam as it bypasses code signing and
+    antivirus checks.
```
fmazan, line 24: Would you mind explaining why this should be the case? I see no reason why using Steam would bypass any AV checks let alone PE signature checks.

Calinou: Steam was always given a free pass over code signing on Windows and macOS. This happens because it doesn't set the typical "downloaded" flag that web browsers do when it downloads games.

fmazan: I'd say it's a specialty of Windows, that is does not enforce a binary to have a valid authenticode signature and just happily runs it even with invalid one (unless you have properly configured AppLocker or your PC is in an enterprise environment, where this is usualy enforced by group policies).

Anyway, this has nothing to do with any antivirus checks. Generally speaking, an AV will check every newly created/spawned/dropped executable no matter its digital signature and using Steam won't prevent false positive in this case. Only that specific AV vendor in the original issue could possibly remove a badly crafted detection.

If it was up to me, I'd just mention that: "On Windows, PCK embedding is also known to cause false positives in some antivirus programs."

Calinou:

PR #4603, opened 2021-01-30
Document MP3 support in audio streams page

skyace65:
MP3 support is coming in 3.2.4. Slightly alters the audio stream page and lists MP3 as a supported file type. I'm not familiar enough with audio formats to know what some of the use case differences are. If someone can explain them to me I'll gladly add that to this PR.

### DIFF
--- a/tutorials/audio/audio_streams.rst
+++ b/tutorials/audio/audio_streams.rst
@@ -14,23 +14,20 @@ AudioStream
 -----------
 
 An audio stream is an abstract object that emits sound. The sound can come from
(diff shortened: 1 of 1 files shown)


Calinou:
> I'm not familiar enough with audio formats to know what some of the use case differences are.

Pros of MP3 compared to Ogg Vorbis:

- Lower CPU usage.
- Good for user-supplied music, as MP3 is more common than Ogg Vorbis.

Cons of MP3 compared to Ogg Vorbis:

- Higher file size for the same quality (or lower quality for the same file size).

skyace65:
Superseded by #4605

PR #4606, opened 2021-01-30
Update audio stream and audio import pages, document MP3 support

skyace65:
I removed the audio file information from the audio stream player page and added a link to the importing audio page, we shouldn't be covering the same information twice. Added information on MP3 support to the importing audio page. I've removed the import options screenshot for now, it's out of place where it was and the page doesn't go over the WAV options. Reworded some stuff, the intro about why to import audio seemed unnecessary.

@Calinou thank you for explaining the advantages of MP3 in the old PR. is the information I found for the comparison chart accurate? The website I found it on looked good but I'm not experienced with audio stuff.


tutorials/assets_pipeline/importing_audio_samples.rst
```diff
@@ -3,19 +3,25 @@
...
-* Ogg Vorbis files use a stronger compression that results in much smaller file size, but require significantly more processing power to play back.
+* WAV files use raw data or light compression, make few demands on the
+  CPU to play back (hundreds of simultaneous voices in this format are
+  fine), but take up significant space.
+* Ogg Vorbis files use a stronger compression that results in much
+  smaller file size, but require significantly more processing power to
+  play back.
+* MP3 files are more copmressed than WAV files, but not as compressed as
+  Ogg Vorbis. So they require lower CPU usage than ogg vorbis to play back,
+  but their file size is larger.
```
Calinou, line 22:
```suggestion
* MP3 files use better compression than WAV with IMA-ADPCM, but worse than
  Ogg Vorbis. This means that a MP3 file with roughly equal quality to
  Ogg Vorbis will be significantly larger. On the bright side, MP3 requires
  less CPU usage to play back compared to Ogg Vorbis. This makes MP3 useful
  for mobile and HTML5 projects where CPU resources are limited, especially
  when playing multiple compressed sounds at the same time (such as long
  ambient sounds).
```

tutorials/assets_pipeline/importing_audio_samples.rst
```diff
@@ -3,19 +3,25 @@
...
+Godot provides three options to import your audio data: WAV, Ogg Vorbis
+and MP3.
 
 Each has different advantages.
 
-* WAV files use raw data or light compression, make few demands on the CPU to play back (hundreds of simultaneous voices in this format are fine), but take up significant space.
-* Ogg Vorbis files use a stronger compression that results in much smaller file size, but require significantly more processing power to play back.
+* WAV files use raw data or light compression, make few demands on the
+  CPU to play back (hundreds of simultaneous voices in this format are
+  fine), but take up significant space.
```
Calinou, line 16:
```suggestion
* WAV files use raw data or light compression (IMA-ADPCM). They are lightweight
  on the CPU to play back (hundreds of simultaneous voices in this format are
  fine). The downside is that they take up a lot of disk space.
```

tutorials/assets_pipeline/importing_audio_samples.rst
```diff
@@ -3,19 +3,29 @@
...
+  play back.
+* MP3 files use better compression than WAV with IMA-ADPCM, but worse than
+  Ogg Vorbis. This means that a MP3 file with roughly equal quality to
+  Ogg Vorbis will be significantly larger. On the bright side, MP3 requires
+  less CPU usage to play back compared to Ogg Vorbis. This makes MP3 useful
+  for mobile and HTML5 projects where CPU resources are limited, especially
+  when playing multiple compressed sounds at the same time (such as long
+  ambient sounds).
+
+.. note:: Godot supports MP3 files with the `minimp3 <https://github.com/lieff/minimp3>`__ decoder.
```
Calinou, line 28: This is an implementation detail the user doesn't need to care about, so I'd remove it:

```suggestion
```

skyace65: I wanted something there so if people want to improve the import they know what project to go to. But you're right it doesn't really fit in here.

Calinou (approved):
Thanks!

Calinou:
Cherry-picked to `3.2` (with a notice stating that only Godot 3.2.4 and later support MP3).

PR #4608, opened 2021-01-30
Mention WebXR support in AR/VR primer page

skyace65:
Mentions that Godot supports WebXR out of the box in Godot 3.2.4 and later now. Also I split the intro paragraph into 2 paragraphs for readability.


Calinou (approved):
Thanks!

PR #4609, opened 2021-01-30
Replace instance diagram images for readability on all backgrounds

Maarrk:
Recreated the instance hierarchy diagrams using [diagrams.net](https://www.diagrams.net/) because the existing ones had some ugly white background artifacts (*see screenshot of docs below*)

old diagrams

The colors used in new version are picked from Godot editor UI, and were chosen to be legible against white and dark background that's used in the page.

instancing_diagram_backgrounds


Calinou (approved):
Thanks! (I optimized the images losslessly using `oxipng --strip --zopfli`.)

PR #4620, opened 2021-02-03
Add a page about gamepad/controller support in Godot

Calinou:
This is a first pass at documenting controller support in Godot. Feel free to suggest additional things to document.

@aaronfranke I know that you worked on getting input actions on multiple directions for 3.2.4, and I remember this being useful to implement proper deadzone calculations for 2D top-down movement. Could you explain this a little? I don't really know how to document it, but it would be good to have the documentation page explain how to handle dead zones properly in this situation.

### DIFF
--- a/tutorials/inputs/controllers_gamepads_joysticks.rst
+++ b/tutorials/inputs/controllers_gamepads_joysticks.rst
@@ -1,0 +1,246 @@
+.. _doc_controllers_gamepads_joysticks:
+
+Controllers, gamepads, and joysticks
+====================================
+
+Godot supports hundreds of controller models thanks to the community-sourced
+`SDL game controller database <https://github.com/gabomdq/SDL_GameControllerDB>`__.
+
+Controllers are supported on Windows, macOS, Linux, Android, iOS, and HTML5.
+
+Note that more specialized devices such as steering wheels, rudder pedals and
+`HOTAS <https://en.wikipedia.org/wiki/HOTAS>`__ are less tested and may not
+always work as expected. If you have access to one of those devices, don't hesitate to
+`report bugs on GitHub <https://github.com/godotengine/godot/blob/master/CONTRIBUTING.md#reporting-bugs>`__.
+
+In this guide, you will learn:
+
+- **How to write your input logic to support both keyboard and controller inputs.**
+- **How controllers can behave differently from keyboard/mouse input.**
+- **Troubleshooting issues with controllers in Godot.**
+
+Supporting universal input
+--------------------------
+
+Thanks to Godot's input action system, Godot makes it possible to support both
+keyboard and controller input without having to write separate code paths.
+Instead of hardcoding keys or controller buttons in your scripts, you should
+create *input actions* in the Project Settings which will then refer to
+specified key and controller inputs.
+
+Input actions are explained in detail on the :ref:`doc_inputevent` page.
+
+.. note::
+
+    Unlike keyboard input, supporting both mouse and controller input for an
+    action (such as looking around in a first-person game) will require
+    different code paths since these have to be handled separately.
+
+Which Input singleton method should I use?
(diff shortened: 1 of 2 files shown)


aaronfranke:
This article is missing information about the most important part: What code should people use? It only mentions that the code paths are the same between keyboard input and controller input, but not what those code paths are.

* For joysticks and WASD etc when you have two axes, want both axes to behave as one input, and for input to be in the form of a `Vector2` such as for movement or looking, starting with Godot 4.0, `Input.get_vector` should be used.

* For other types of axis input, such as if you want to handle joystick axes individually, or you have a device where an axis is by itself (such as the throttle on a flight stick), starting with Godot 4.0, `Input.get_axis` should be used.

* For other types of analog input, such as if you are handling a trigger, or only care about one direction on a joystick, or want to handle one direction at a time, starting with Godot 3.1, `Input.get_action_strength` should be used.

* For binary input, if you only want a true or false value, or if the input is inherently binary such as mouse buttons, controller face buttons/start/select/joystick clicks/bumpers, and keyboard keys (except for keys handled as vectors/axes), in pretty much any Godot version, `Input.is_action_pressed` should be used.

tutorials/inputs/controllers_gamepads_joysticks.rst
```diff
@@ -0,0 +1,246 @@
...
+My controller has incorrectly mapped buttons or axes.
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If buttons are incorrectly mapped, this may be due to an erroneous mapping from
+the `SDL game controller database <https://github.com/gabomdq/SDL_GameControllerDB>`__.
+You can contribute an updated mapping to be included in the next Godot version
+by opening a pull request on the linked repository.
+
+There are many ways to create mappings. One option is to use the mapping wizard
+in the `official Joypads demo <https://godotengine.org/asset-library/asset/140>`__.
```
aaronfranke, line 194: Note: This is technically not true just yet, but https://github.com/godotengine/godot-demo-projects/pull/575 was merged, so this sentence will be true the next time I make a release of the demos on the Godot Asset Library.

aaronfranke (approved):
Looks good to me now. I made [a few fixes and tweaks and added C# example code](https://github.com/godotengine/godot-docs/compare/2b8e16799f7fb2d9e522af9d24bc6c4b8e3c8754..65c460d64f50412e7bb291b17a6c2fd4a44910cc), let me know what you think @Calinou.

Calinou:
@aaronfranke Changes look good to me :slightly_smiling_face:

tutorials/inputs/controllers_gamepads_joysticks.rst
```diff
@@ -0,0 +1,246 @@
...
+  axes to behave as a single input, use ``Input.get_vector()``:
+
+.. tabs::
+ .. code-tab:: gdscript GDScript
+
+    # `velocity` will be a Vector2 between `Vector2(-1.0, -1.0)` and `Vector2(1.0, 1.0)`.
+    var velocity = Input.get_vector("move_left", "move_right", "move_forward", "move_back")
+
+    # The line above is a shorter form of:
+    var velocity = Vector2(Input.get_action_strength("move_right") - Input.get_action_strength("move_left"),
```
groud, line 54: This is false in fact. One of the advantages of the using get_vector over this is the handling of the deadzone.
With this solution the deazone end up square instead of circular.

I'm ok with mentionning that they are similar bu it's false to say they are equivalent.

Calinou: PR opened: https://github.com/godotengine/godot-docs/pull/4620#discussion_r575847917

akien-mga:
Cherry-picked for 3.3.

PR #4621, opened 2021-02-04
Add Capture effect to audio buses page (4.0)

skyace65:
The audio effect capture was merged to the master branch of Godot recently, and allows you to access the microphone of your computer, or rather access it properly. You could access it with the record effect but it was apparently not that good. The Record audio page needs to be redone at some point with this. If anyone wants to do it go ahead, I'm going to focus on 3D documentation for right now. The updated image has gone through png compression.


akien-mga (approved):

akien-mga:
Thanks!

PR #4627, opened 2021-02-07
Fix confusing writing in Your first 3D shader

FeralBytes:
The use of "off" does make sense to put a check mark, but in the case of the editor it explicitly says "On" so the use of off or on is best avoided. Check on it's own still makes sense and is clearer.

### DIFF
--- a/tutorials/shaders/your_first_shader/your_first_3d_shader.rst
+++ b/tutorials/shaders/your_first_shader/your_first_3d_shader.rst
@@ -305,7 +305,7 @@ do that by passing in a second noise texture.
   uniform sampler2D normalmap;
 
 Set this second uniform texture to another NoiseTexture with another
-OpenSimplexNoise. But this time, check "As Normalmap".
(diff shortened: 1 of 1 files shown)


tutorials/shaders/your_first_shader/your_first_3d_shader.rst
```diff
@@ -305,7 +305,7 @@ do that by passing in a second noise texture.
   uniform sampler2D normalmap;
 
 Set this second uniform texture to another NoiseTexture with another
-OpenSimplexNoise. But this time, check off "As Normalmap".
+OpenSimplexNoise. But this time, check "As Normalmap".
```
Calinou, line 308:
```suggestion
OpenSimplexNoise. But this time, check **As Normalmap**.
```

Calinou (approved):
Thanks!

Calinou:
Cherry-picked to `3.2`: https://github.com/godotengine/godot-docs/commit/9e486348658207300da13a20a556f56c2857f333

PR #4628, opened 2021-02-07
Document Firefox WebGL being slow on Linux and workaround

Calinou:
There's nothing we can do about it except instruct people to use a Chromium-based browser for their HTML5 gaming needs :slightly_frowning_face:

We could perhaps use `console.warn()` to display a similar message to the console based on the user agent.

### DIFF
--- a/tutorials/export/exporting_for_web.rst
+++ b/tutorials/export/exporting_for_web.rst
@@ -17,6 +17,13 @@ in the user's browser.
                :ref:`iOS' native export functionality <doc_exporting_for_ios>`
                instead, as it will also result in better performance.
 
+.. note::
+
+    If you use Linux, due to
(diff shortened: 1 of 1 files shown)


akien-mga (approved):
Sad but true :(

Calinou:

Anutrix:
https://docs.godotengine.org/en/latest/tutorials/export/exporting_for_web.html states that WebGL performance on Firefox is bad and points to https://bugzilla.mozilla.org/show_bug.cgi?id=1010527 which is now closed since 2 years. 

@Calinou Can we remove that statement now(since it's partially fixed) or at least point to https://bugzilla.mozilla.org/show_bug.cgi?id=1572697 and https://bugzilla.mozilla.org/show_bug.cgi?id=788319 instead?

PR #4629, opened 2021-02-07
Add an Using the web editor page

Calinou:
This page describes the current workflow and limitations of the HTML5 Godot editor.

### DIFF
--- a/tutorials/editor/using_the_web_editor.rst
+++ b/tutorials/editor/using_the_web_editor.rst
@@ -1,0 +1,121 @@
+.. _doc_using_the_web_editor:
+
+Using the Web editor
+====================
+
+Since Godot 3.2.4, there is a `Web editor <https://editor.godotengine.org/>`__
+you can use to work on new or existing projects.
+
+.. note::
+
+    The web editor is in a preliminary stage. While its feature set may be
+    sufficient for educational purposes, it is currently **not recommended for
+    production work**. See :ref:`doc_using_the_web_editor_limitations` below.
+
+Browser support
+---------------
+
+The Web editor requires support for WebAssembly's SharedArrayBuffer. This
+is in turn required to support threading in the browser. The following desktop
+browsers support WebAssembly threading and can therefore run the web editor:
+
+- Chrome 68 or later
+- Firefox 79 or later
+- Edge 79 or later
+
+Opera and Safari are not supported yet. Safari may work in the future once
+proper threading support is added.
+
+**Mobile browsers are currently not supported.**
+
+The web editor supports both the GLES3 and GLES2 renderers, although GLES2 is
+recommended for better performance and compatibility with old/low-end hardware.
+
+.. note::
+
+    If you use Linux, due to
+    `poor Firefox WebGL performance <https://bugzilla.mozilla.org/show_bug.cgi?id=1010527>`__,
+    it's recommended to use a Chromium-based browser instead of Firefox.
+
+.. _doc_using_the_web_editor_limitations:
+
+Limitations
+-----------
+
+Due to limitations on the Godot or Web platform side, the following features
+are currently missing:
+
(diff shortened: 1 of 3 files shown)


tutorials/editor/using_the_web_editor.rst
```diff
@@ -0,0 +1,120 @@
...
+The web editor offers an alternative way to deal with the editor and game
+windows (which are now "tabs"). You can switch between the **Editor** and
+**Game** tabs using the buttons on the top. You can also close the running game
+or editor by clicking the **×** button next to those tabs.
+
+Where are my project files?
+---------------------------
+
+Due to browser security limitations, the editor will save the project files to
+the browser's IndexedDB storage. This storage isn't accessible as a regular folder
```
Calinou, line 114: Does the editor actually use IndexedDB for storage? I don't remember exactly; I may be incorrect there.

Faless: yes, we do use IndexedDB,  the first prototype also had FileSystemAPI support but was pretty messy with an external library so I decided to drop it, since IndexedDB seems to work fine.

tutorials/editor/using_the_web_editor.rst
```diff
@@ -0,0 +1,120 @@
...
+
+Limitations
+-----------
+
+Due to limitations on the Godot or Web platform side, the following features
+are currently missing:
+
+- No C#/Mono support.
+- No GDNative support.
+- No way to download the project files as a ZIP archive.
```
Faless, line 50: This has actually changed, you can go to `Project -> Tools -> Download Project Source`  to download the project files as a ZIP archive.

tutorials/editor/using_the_web_editor.rst
```diff
@@ -0,0 +1,120 @@
...
+**Mobile browsers are currently not supported.**
+
+The web editor supports both the GLES3 and GLES2 renderers, although GLES2 is
+recommended for better performance and compatibility with old/low-end hardware.
+
+.. note::
+
+    If you use Linux, due to
+    `poor Firefox WebGL performance <https://bugzilla.mozilla.org/show_bug.cgi?id=1010527>`__,
+    it's recommended to use a Chromium-based browser instead of Firefox.
```
Faless, line 38: honestly, I don't really like favouring one browser here.
At least in my experiences FF is performing quite okay with the editor.
That said, Chrome do have some nice features that FF is missing, e.g. better clipboard API so I don't mind to much.

tutorials/editor/using_the_web_editor.rst
```diff
@@ -0,0 +1,120 @@
...
+  ``/home/web_user/projects/<project folder>``.
+- Click **Import** and the project will open in the editor.
+
+.. attention::
+
+    It's important to place the project folder in ``/home/web_user/projects``.
+    If your project folder is placed outside ``/home/web_user/projects``, you will
+    lose your project when closing the editor!
+
+This import process will be streamlined in the future.
```
Faless, line 95: note, this will change in RC2 already:
- The zip preload will start the project manager with the `Import Zip` popup.
- Persistence will be across the whole `/home/web_user/` folder.

Faless (changes requested):
Well, thanks a lot, this is really good.
See my comment about downloading project source, and the preloading/import (we could leave that as is for now, and update for RC2)

tutorials/editor/using_the_web_editor.rst
Calinou: > At least in my experiences FF is performing quite okay with the editor.

The web editor is certainly usable on Firefox, but Chromium-based browsers will pretty much always perform better on Linux for anything WebGL. This is especially true at larger resolutions (since the compositing performance largely depends o nresolution).

Calinou:

PR #4634, opened 2021-02-08
inputevent: Fix/update default signal name

Beuc:
See
https://github.com/godotengine/godot/blob/5260b6e046c244ed6376c47c7af5a0f27d934c12/scene/gui/control.cpp#L3038
https://github.com/godotengine/godot/blob/5260b6e046c244ed6376c47c7af5a0f27d934c12/scene/main/viewport.cpp#L2373
https://docs.godotengine.org/fr/stable/classes/class_control.html#signals


Calinou (approved):
Thanks!

PR #4641, opened 2021-02-12
Update link to `awesome-godot` repository

Xrayez:
Links to official [awesome-godot](https://github.com/godotengine/awesome-godot) repo, which seems to be more up-to-date.

PR #4642, opened 2021-02-12
Update link to `awesome-godot` repository

Xrayez:
#4641 version for the `stable` docs, which most devs should use.


Calinou (approved):
Thanks!

PR #4650, opened 2021-02-14
Add a note to set default values for custom resources

RoniPerson:
I had the problem today that I don't specified a default for a value. It took me some time to figure out what was going wrong. Because this is one of the few resources on creating custom resources, I thought it would be good to mention this here.

### DIFF
--- a/tutorials/scripting/resources.rst
+++ b/tutorials/scripting/resources.rst
@@ -266,6 +266,9 @@ Let's see some examples.
             }
         }
 
+Be sure to set a default value for every parameter of the ``_init`` function.
+Otherwise, there will be problems with creating and editing your resource over the inspector.
+
 .. note::
 
     Resource scripts are similar to Unity's ScriptableObjects. The Inspector
--- a/tutorials/scripting/resources.rst
+++ b/tutorials/scripting/resources.rst
@@ -241,8 +241,9 @@ Let's see some examples.
                 [Export]
                 public String[] Strings { get; set; }
 
-                // Make sure that every variable gets a default value. 
-                // Otherwise, there will be problems with creating and editing your resource via the inspector.
+                // Make sure that every parameter has a default value. 
+                // Otherwise, there will be problems with creating and editing
+                // your resource via the inspector.
                 public BotStats(int health = 0, Resource subResource = null, String[] strings = null)
                 {
                     Health = health;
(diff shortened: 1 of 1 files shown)


Calinou (commented):
Good idea, but I think it would be better to add this note as a comment just above the `func _init()` declaration in GDScript (and `public BotStats()` in C#). This way, people will skim over it more naturally as they read the code sample.

tutorials/scripting/resources.rst
```diff
@@ -206,6 +206,7 @@ Let's see some examples.
     export(Resource) var sub_resource
     export(Array, String) var strings
 
+    # Make sure that every variable gets a default value. Otherwise, there will be problems with creating and editing your resource via the inspector.
```
Calinou:
```suggestion
    # Make sure that every variable gets a default value.
    # Otherwise, there will be problems with creating and editing
    # your resource via the inspector.
```

tutorials/scripting/resources.rst
```diff
@@ -238,6 +239,7 @@ Let's see some examples.
                 [Export]
                 public String[] Strings { get; set; }
 
+                // Make sure that every variable gets a default value. Otherwise, there will be problems with creating and editing your resource via the inspector.
```
Calinou:
```suggestion
                // Make sure that every variable gets a default value.
                // Otherwise, there will be problems with creating and editing
                // your resource via the inspector.
```

tutorials/scripting/resources.rst
```diff
@@ -206,6 +206,8 @@ Let's see some examples.
     export(Resource) var sub_resource
     export(Array, String) var strings
 
+    # Make sure that every variable gets a default value. 
+    # Otherwise, there will be problems with creating and editing your resource via the inspector.
```
Calinou, line 210:
```suggestion
    # Make sure that every parameter gets a default value. 
    # Otherwise, there will be problems with creating and editing
    # your resource via the inspector.
```

tutorials/scripting/resources.rst
```diff
@@ -238,6 +240,8 @@ Let's see some examples.
                 [Export]
                 public String[] Strings { get; set; }
 
+                // Make sure that every variable gets a default value. 
+                // Otherwise, there will be problems with creating and editing your resource via the inspector.
```
Calinou, line 244:
```suggestion
                // Make sure that every parameter has a default value. 
                // Otherwise, there will be problems with creating and editing
                // your resource via the inspector.
```

Calinou (approved):
Thanks!

Calinou:

PR #4651, opened 2021-02-14
Clarify deadzone difference for `Input.get_vector()` in Controllers

Calinou:
See https://github.com/godotengine/godot-docs/pull/4620#discussion_r575847917.

### DIFF
--- a/tutorials/inputs/controllers_gamepads_joysticks.rst
+++ b/tutorials/inputs/controllers_gamepads_joysticks.rst
@@ -48,18 +48,26 @@ In Godot 4.0, there are 3 ways to get input in an analog-aware way:
  .. code-tab:: gdscript GDScript
 
     # `velocity` will be a Vector2 between `Vector2(-1.0, -1.0)` and `Vector2(1.0, 1.0)`.
+    # This handles deadzone in a correct way for most use cases.
+    # The resulting deadzone will have a circular shape as it generally should.
     var velocity = Input.get_vector("move_left", "move_right", "move_forward", "move_back")
 
-    # The line above is a shorter form of:
+    # The line below is similar to `get_vector()`, except that it handles
+    # the deadzone in a less optimal way. The resulting deadzone will have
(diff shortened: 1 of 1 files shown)


aaronfranke:
Well, it's not that it's necessarily incorrect, it's just different (and usually sub-par) behavior. I think this should mention that they're very similar except that `get_vector` has a circular deadzone while the manual one has a square deadzone.

groud:
Yep, @aaronfranke is right here. A cross-shaped deadzone is not necessarily incorrect (Rocket league apparently uses it for example), but it's a far less commonly desired behavior.

I would rephrase it a little but otherwise it looks ok.

tutorials/inputs/controllers_gamepads_joysticks.rst
```diff
@@ -48,18 +48,24 @@ In Godot 4.0, there are 3 ways to get input in an analog-aware way:
  .. code-tab:: gdscript GDScript
 
     # `velocity` will be a Vector2 between `Vector2(-1.0, -1.0)` and `Vector2(1.0, 1.0)`.
+    # This handles deadzone in a correct way for most use cases.
+    # The resulting deadzone will have a circular shape as it generally should.
     var velocity = Input.get_vector("move_left", "move_right", "move_forward", "move_back")
 
-    # The line above is a shorter form of:
+    # The line below handles deadzone in a less optimal way. The resulting deadzone
+    # will have a square-ish shape when it should usually have a circular shape.
```
aaronfranke, line 56: This is how I would word it:

```suggestion
    # The line below is similar to `get_vector`, except that it handles the
    # deadzone in a less optimal way. The resulting deadzone will have a
    # square-ish shape when it should ideally have a circular shape.
```

\+ this should be changed in the C# comment as well.

groud (approved):

aaronfranke (approved):

akien-mga:
Cherry-picked for 3.3.

PR #4667, opened 2021-02-19
Added sample code to the bsearch_custom function

### LINKED ISSUE #4564: There is no example of using bsearch_custom in the documentation
**Your Godot version:**
3.2.3.Stable

**Issue description:**
There is no example of using bsearch_custom in the documentation. You need to pass it a custom function, but without an explanation of what parameters and return values it expects I haven't been able to get it to work. I assume it's very similar to sort_custom which does have an example, but what I've tried doesn't work.

https://docs.godotengine.org/en/3.2/classes/class_array.html#class-array-method-bsearch-custom

hungrymonkey:
bsearch_custom arguments

name - type - description
value - Variant - bisection search value
obj - Any object instance - Location of the 2 argument comparison function
func - String - Name of the function declared in obj
before - boolean - first and second resolver

This function is pr against the 3.x series because pr #45698 changed
the call signature to
bsearch_custom, sarray("value", "func", "before"), varray(true));

Closes #4564

### DIFF
--- a/classes/class_array.rst
+++ b/classes/class_array.rst
@@ -216,7 +216,40 @@ Finds the index of an existing value (or the insertion index that maintains sort
 
 - :ref:`int<class_int>` **bsearch_custom** **(** :ref:`Variant<class_Variant>` value, :ref:`Object<class_Object>` obj, :ref:`String<class_String>` func, :ref:`bool<class_bool>` before=true **)**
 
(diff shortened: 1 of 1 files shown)


hungrymonkey:
Wait, I forgot to build it. I will reopen it after i test it

Calinou:

hungrymonkey:
thanks, i will move my commit to the main repo

PR #4668, opened 2021-02-20
Clarification on theme's get_stylebox

asmaloney:
Remove reference to "icon" (must have been a copy-paste error) & clarify where to find out what names & types are valid.


pycbouh:
Hi, and thank you for contributing! Class reference is maintained on the [main repository](https://github.com/godotengine/godot/tree/master/doc/classes) and you would need to make your changes there. Please read about [contributing to the class reference](https://docs.godotengine.org/en/stable/community/contributing/updating_the_class_reference.html).

asmaloney:
Thanks @pycbouh.

I wonder if the CI could automatically reject changes to the class files in this repo and spit out a link to the "contributing to the class reference" docs? Might save people time.

To have the class docs not inline with the actual code (like doxygen) or in the docs repo is... unusual.

pycbouh:
May be a good idea, though I think Akien also uses PRs when syncing online docs (this repo) with the main repo. So it would break merges for him.

PR #4670, opened 2021-02-21
3.2

Yinisiki:
To make it easier for godot users to subscribe to godot tutorial YouTube channels one by one, 
Add all Godot Tutorial channels to The Godot Engine YouTube channel’s Channel page. 
Like in this example: 
https://www.youtube.com/channel/UCRsGzjEgPBjjm_m8GUE-2ag/channels

Instead of opening each channel from the godot documentation tutorials page to new tab and then subscribing to each one separately. 

Moreover, Lot of new godot tutorial channels have come up which I have collected in my youtube channel’s link as given above. 

`<!--`
**Note:** Pull Requests should be made against the `master` by default.

Only make Pull Requests against other branches (e.g. `2.1`) if your changes only apply to that specific version of Godot.

The type of content accepted into the documentation is explained here: https://docs.godotengine.org/en/latest/community/contributing/content_guidelines.html

-->


pycbouh:
It seems like you wanted to [make a new issue](https://github.com/godotengine/godot-docs/issues/new/choose), and not a new PR. If you still want to file an issue, please use one of the templates provided. Though this request is unlikely to be fulfilled.

PR #4671, opened 2021-02-21
Update exporting string enum to include default value

fillup:
In Godot v3.2.3.stable.official when exporting a var as a String enum, a default value must be provided or else the variable remains empty when running the scene. Setting a default value of `""` resolves this issue. 

This documentation appears to be for a newer syntax of GDScript though so I don't know if what I've proposed is exactly right.

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_exports.rst
+++ b/tutorials/scripting/gdscript/gdscript_exports.rst
@@ -57,8 +57,9 @@ Examples
 
     # Editor will enumerate as 0, 1 and 2.
     @export_enum("Warrior", "Magician", "Thief") var character_class
-    # If type is String, editor will enumerate with string names.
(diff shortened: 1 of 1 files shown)


Calinou:
If this only applies to the `3.2` branch, I would recommend opening a **new** pull request against the `3.2` branch.

akien-mga:
I can confirm that there's no problem in the `master` branch with the current example.

And I tested the current code in `3.2` and it works fine too:
Screenshot_20210319_130957
```
export(String, "Rebecca", "Mary", "Leah") var character_name
```

So this change doesn't seem necessary.

akien-mga:
(To clarify, the fact that the value is default initialized to `""` while the inspector shows `Rebecca` is confusing, but this PR doesn't solve that. It's reported as a bug in https://github.com/godotengine/godot/issues/33914.)

Issue #4676, opened 2021-02-23
Illustrate Vector2's angle_to(to) and similar methods


mechPenSketch:
**Your Godot version:**
3.2.3.stable.official

**Issue description:**
It's hard to imagine what is the angle returned from Vector2's method `angle_to(Vector to)` as well as `angle_to_point(Vector to)`.

I suggest adding an image under each of the respective method descriptions. This image contains a graph that illustrate what angle is being returned.

Here is an example of the graph, this one demonstrating `angle_to(Vector to)`:
![godoc_vector_angle_to](https://user-images.githubusercontent.com/15977859/108860012-c4f39800-7628-11eb-80ab-8d84c5a6ac48.png)

Disclaimer: This image is already compressed as instructed in [Image contribution guidelines](https://docs.godotengine.org/en/stable/community/contributing/docs_writing_guidelines.html).

For the graph, I tried following the format similar to the ones in [Vector Math](https://docs.godotengine.org/en/stable/tutorials/math/vector_math.html), but I can't follow through perfectly. I can't get the right arrowhead, and I don't know the font used there. Anyone knows what tools to use?

I would also need advice for the folder placement for the images.

Once this suggestion is approved, I can also create graph for `angle_to_point(Vector to)` as well as `angle()`.

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/stable/classes/class_vector2.html#class-vector2-method-angle-to

Calinou:
The class reference can't have inline images, so these images would have to be hosted in the godot-docs repository and linked using an `[url=]` tag.

Torguen:
Surely it is not "to" who establishes the positive X axis?

Torguen:
I think you are confused, the image is from "angleToPoint"?

mechPenSketch:
> I think you get confused, I think the image is from "angleToPoint"?
> 
> Also the Right axis is set by the vector "to" according to my studies.

The image shown in the first post doesn't show the correct angle, but I have corrected it by the time I made the appropriate pull request.

Torguen:
Could you provide me with the link to the extraction please?

mechPenSketch:
> Could you provide me with the link to the extraction please?

https://github.com/godotengine/godot-docs/pull/4678 is where I've added the images.

KoBeWi:
`angle_to_point()` is now outdated btw, see https://github.com/godotengine/godot/pull/43072#issuecomment-966654422

[...]

PR #4678, opened 2021-02-24
Add images for Vector2's angle methods

mechPenSketch:
The image is to be added under [Vector2's angle_to()](https://docs.godotengine.org/en/stable/classes/class_vector2.html#class-vector2-method-angle-to) method description to illustrate the angle being returned.

As adviced in https://github.com/godotengine/godot-docs/issues/4676#issuecomment-784288436, this is where I'm adding the image.


Calinou (changes requested):
Images that aren't referenced anywhere in the manual should be added to the top-level [`img/` folder](https://github.com/godotengine/godot-docs/tree/master/img).

Don't add images to `classes/` as the contents of that folder may be wiped and regenerated in the future.

Calinou (approved):
Thanks!

Remember to open a pull request against the *main Godot repository* to reference those images in the `doc/classes/Vector2.xml` class reference.

aaronfranke:
For [`vector2_angle_to.png`](https://raw.githubusercontent.com/godotengine/godot-docs/master/img/vector2_angle_to.png) isn't confusing but it could be more clear if the ends were arrows.

For [`vector2_angle.png`](https://raw.githubusercontent.com/godotengine/godot-docs/master/img/vector2_angle.png) why was the same red color used for both the vector line, and the X axis, and where it connects to the X axis? It's not immediately clear that the vector is the diagonal line. An arrow would be good here too.

For [`vector2_angle_to_point.png`](https://raw.githubusercontent.com/godotengine/godot-docs/master/img/vector2_angle_to_point.png), similar to the above. The points and labels are good, but the coloring is misleading. For this one, we should showcase them as points, not arrows, it's only the coloring that could be improved.

mechPenSketch:
> For [`vector2_angle_to.png`](https://raw.githubusercontent.com/godotengine/godot-docs/master/img/vector2_angle_to.png) it isn't confusing but it could be more clear if the ends were arrows.
> 
> For [`vector2_angle.png`](https://raw.githubusercontent.com/godotengine/godot-docs/master/img/vector2_angle.png) why was the same red color used for both the vector line, and the X axis, and where it connects to the X axis? It's not immediately clear that the vector is the diagonal line. An arrow would be good here too.
> 
> For [`vector2_angle_to_point.png`](https://raw.githubusercontent.com/godotengine/godot-docs/master/img/vector2_angle_to_point.png), similar to the above. The points and labels are good, but the coloring is misleading. For this one, we should showcase them as points, not arrows, it's only the coloring that could be improved.

Will these do?

`Vector2.angle()`:
vector2_angle_original

`Vector2.angle_to(to)`:
vector2_angle_to_original

`Vector2.angle_to_point(to)`:
vector2_angle_to_point_original

aaronfranke:
@mechPenSketch Those are great, much better! One more thing: For `Vector2.angle_to_point(to)`, we want to showcase that the points don't have to be on the axes, so ideally nether of the points will be on the dark lines. We can show the line extending back to the X axis line to form the angle though (again though, we should show that it doesn't need to be the origin, so we can maybe move/rotate the line so that the extending backwards part only touches the X axis dark line and not the Y axis).

mechPenSketch:
Will this do?
vector2_angle_to_point_original
The line is still darker, but not at that level so that it instead represents a major grid rather than a minor grid or axis.

aaronfranke:
@mechPenSketch It's not what I would do, but this is still better, and it looks good to me. You can submit a PR with these improved images and ping me for review and I'll give it a :heavy_check_mark: .

PR #4695, opened 2021-02-28
Add libstdc++-static to Fedora pre-compile one-liner

hello-adam:
This was required to compile version 3.2 on Fedora 33.

This might apply to other versions but I'm not sure. I can test the master branch and resubmit there if that's better/easier.


Calinou (approved):
Thanks!

I also applied the change on the `master` branch: https://github.com/godotengine/godot-docs/commit/33887d5c7b0cedec4bbfa090ab3408540693ae10

PR #4700, opened 2021-03-01
Add `.editorconfig` and `.gitattributes` files for automatic settings

Calinou:
This smoothens the contribtor experience by automatically configuring editors that support EditorConfig.

The `.gitattributes` file ensures that all files use LF line endings when committed to Git.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,19 @@
+root = true
+
+[*]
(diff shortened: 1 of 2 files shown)


Calinou:
Cherry-picked to the `3.2` and `3.1` branches.

PR #4702, opened 2021-03-02
Add Godot Tutorials channel and info to 3.2

willnationsdev:


willnationsdev:
Oh, my bad. I just copied it down from 4.0, but I see now that you're alphabetizing the order of them. Whoops.

Calinou (approved):
Thanks!

Issue #4681, opened 2021-02-25
GetAngleTo (Node2D)


Torguen:
**Your Godot version:**
3.2.4 rc3 mono

**Issue description:**
The documentation is not explicit.
The documentation should point out that the difference is not between the position of the node and the vector in parameters, but between the angle between its Rotation and the position of the vector received in parameters.

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/stable/classes/class_node2d.html#class-node2d-method-get-angle-to

Calinou:
@Torguen Does https://github.com/godotengine/godot/pull/46407 address this clearly enough? If not, please comment on that pull request.

mechPenSketch:
Maybe this illustration could help:

![node2_get_angle_to_original](https://user-images.githubusercontent.com/15977859/109281285-197c5a80-7857-11eb-9198-af6fc080297e.png)

Torguen:
But is the positive X axis the top left corner of the screen or is it an axis that sets Node2D in its position in each moment?

Torguen:
It is not like that, according to my tests.
According to my tests, the angle returned depends on "Rotation" of node2D. An angle is not obtained with respect to a horizontal X axis, but to the angle where Node2D points.

mechPenSketch:
Right, I didn't account for Node2D's rotation.

Here is the update:
![node2_get_angle_to_original](https://user-images.githubusercontent.com/15977859/109293812-4a648b80-7867-11eb-8000-0fa5efefeddd.png)

PR #4703, opened 2021-03-02
Add image for Node2D's get_angle_to_point(point)

mechPenSketch:
@Torguen requests the specification of Node2D's method [get_angle_to_point(Vector2 point)](https://docs.godotengine.org/en/stable/classes/class_node2d.html#class-node2d-method-get-angle-to). As discussed in https://github.com/godotengine/godot-docs/issues/4681, I proposed adding a graph to show the angle just like Vector2's angle method as discussed in https://github.com/godotengine/godot-docs/issues/4676.


Calinou (approved):
Thanks!

PR #4704, opened 2021-03-02
Correct LTO RAM consumption

RevolNoom:
Using lto to build godot without sufficient RAM will result in building error, as in:
https://github.com/godotengine/godot/issues/45179#issuecomment-760894732


RevolNoom:
I'm new to github and pull requests. So if there's anything I can do to improve it, please let me know.

Calinou:
I don't think it requires as much as 14 GB of available RAM while liking. It depends on the compiler and OS you're using.

In general, the linking process is a lot slower on Windows than it is on Linux.

RevolNoom:
So, here's an experiment by @bruvzg. He tested the RAM usage. I used Ubuntu directly, unlike him using VM. But still, my 8GB RAM is not enough for building. I didn't know for sure because I didn't follow my RAM usage closely. I'll do a build on ubuntu again tonight to be sure.

>
    For the reference, I have done some testing using Ubuntu 20.10 VM, GCC 10.2.0, LTO builds with -j4:

    with 4 GB of RAM and 2 GB swap → same LTO process crash.
    with 8 GB of RAM and 2 GB swap → same LTO process crash.
    with 12 GB of RAM and 2 GB swap → builds finishes successfully, but it's probably the absolute minimum amount of memory to do GCC LTO build.

    Memory usage peaks at about 10.2 GB.

    Update: Number of threads seems to be less relevant than I expected, it's still peaks at about 9.7 GB with -j2.
 
    LLVM LTO seems to be much less memory hungry, last time I have checked, it was working 4 GB VM.```

>

RevolNoom:
I'm sorry. I admit, I didn't perform a careful checking before creating a pull request. 
The LTO process actually consumes ~6.1GB at peak on my Ubuntu. I'm fixing my commit right away.
GB

Calinou:
@RevolNoom You can amend this pull request instead of creating a new one :slightly_smiling_face:

RevolNoom:
I'm creating a new one because,... well, I forgot the --amend option.

Calinou:
Superseded by https://github.com/godotengine/godot-docs/pull/4710.

PR #4707, opened 2021-03-02
Make example code less misleading in Optimization using Servers

Jayman2000:
The variable used to be called "sprite", but it wasn't a Sprite. It was
a StreamTexture. Also, I added an underscore to the example path so that
it matches the documentation's recommendation for file names:
https://docs.godotengine.org/en/stable/getting_started/workflow/project_setup/project_organization.html#style-guide


clayjohn (approved):

Calinou (approved):
Thanks!

PR #4713, opened 2021-03-03
Fixed a not needed ';' in GDScript sample code

ck90works:
At the end of line 867 there was an unneeded semicolon in the GDScript sample code.
Semicolon is not used in any other lines in the GDScript sample code in this tutorial. 
I removed the semicolon because I think that it might irritate beginners.
I've checked all GDscript sample codes in this file for unnecessary semicolons and didn't find any other than the one in line 867.


Calinou (approved):

PR #4715, opened 2021-03-04
Remove another easily misunderstood joke

dsnopek:
While I know that the Godot community is made of a diverse group of people from all over the world, and that all (I think?) the core maintainers have a native language other than English, people outside the community don't know that.

So, I could easy imagine someone seeing this line:

> The world is full of different markets and cultures and, to maximize profits™, nowadays games are released in several languages.

... and interpreting that to mean that games are _only_ translated to other languages other than English to maximize profits, implying that other languages are somehow lesser.

Again, I know that isn't the intended meaning of the line! I know it's a little joke. :-)

However, in light of people outside the community recently getting the wrong idea about Godot from a joke in the documentation, it might be a good idea to remove this one before that happens again. Also, unlike the previous example which was buried in a deep corner of the documentation where new folks are unlikely to find it, this one is in a tutorial that is much more likely to be seen.

### DIFF
--- a/tutorials/assets_pipeline/importing_translations.rst
+++ b/tutorials/assets_pipeline/importing_translations.rst
@@ -6,10 +6,9 @@ Importing translations
 Games and internationalization
 ------------------------------
 
-The world is full of different markets and cultures and, to maximize
-profits™, nowadays games are released in several languages. To solve
-this, internationalized text must be supported in any modern game
-engine.
+The world is full of different languages and cultures, so nowadays games
+are released in several languages. To handle this, internationalized text
+must be supported in any modern game engine.
 
 In regular desktop or mobile applications, internationalized text is
 usually located in resource files (or .po files for GNU stuff). Games,


tastesliketriangle:
> ... and interpreting that to mean that games are only translated to other languages other than English to maximize profits, implying that other languages are somehow lesser.

it's worth mentioning this interpretation would require a leap in logic considering it doesn't endorse or even mention English.

EmmanuelOga:
:roll_eyes:

tutorials/assets_pipeline/importing_translations.rst
```diff
@@ -6,10 +6,9 @@ Importing translations
 Games and internationalization
 ------------------------------
 
-The world is full of different markets and cultures and, to maximize
-profits™, nowadays games are released in several languages. To solve
-this, internationalized text must be supported in any modern game
-engine.
+The world is full of different languages and cultures, so nowadays games
+are released in several languages. To solve this, internationalized text
+must be supported in any modern game engine.
```
BastiaanOlij, line 11: I would change this to "To support this"
We've not solving the problem of the world having different languages, we're not undoing babylon ;)

BastiaanOlij: Or some change to that effect that actually makes the sentence flow, you know what I mean :)

dsnopek:
@tastesliketriangle:

> it's worth mentioning this interpretation would require a leap in logic considering it doesn't endorse or even mention English.

Well, it's written in English. :-) I can imagine that fact causing some folks to think we're talking about translating from English to something else.

tutorials/assets_pipeline/importing_translations.rst
dsnopek: Yeah, that makes sense! I'll update the PR...

pycbouh:
> Also, unlike the previous example which was buried in a deep corner of the documentation where new folks are unlikely to find it, this one is in a tutorial that is much more likely to be seen.

And yet, that other piece caused controversy on more that one occasion, while this one bothers nobody, apparently. The removal of the encrypting article was not an invitation to start removing all of the character from the docs. Let us deal with the wordings and jokes on the case by case basis, and only if a problem arises.

akien-mga:
I'm OK with removing that "joke" as it's not particularly fun, and might not translate well to all languages. But just removing it makes this intro paragraph pretty bland IMO, and I think we should thrive for making the content better and more inclusive, not just "sanitized" to avoid another time-wasting Twitter controversy :)

We're doing a ton of work to make internationalization as easy as possible (support for gettext plurals, translation context, resource remapping) and to include all languages in the world (with Complex Text Layouts in Godot 4.0, input method engines, better font features, etc.), so our work shows that we're committed to internationalization and making things accessible to all players, regardless of their language. I think that's what this paragraph should aim to convey.

Still, this change in itself is OK. The whole page could use a pass of proofreading to improve the language overall and provide more useful insights in its intro paragraphs.

akien-mga:
Anyway, I'll merge this now as content-wise it's an improvement IMO.

But I would advise against going on a campaign to try to sanitize the whole docs and remove any perceived asperity. These asperities in tutorial-form documentation can be what makes the difference between boring technical reference and a bit of light-hearted fun, as the "Imagine you're a cook" introduction to https://docs.godotengine.org/en/stable/getting_started/step_by_step/scenes_and_nodes.html

If users have specific complaints about sections of the documentation, I'm happy to review their constructive feedback and see if we should change more things, but we don't need to remove all character in hope of being sheltered from any criticism. There will always be _someone_ to dislike something, so like with everything in Godot, we should only act if we actually know that there's an issue, and not future-proof things.

akien-mga:
> The whole page could use a pass of proofreading to improve the language overall and provide more useful insights in its intro paragraphs.

That still stands if anyone feels like improving this page. It could use somewhat better language, and put more emphasis on the nice, inclusive features that we added to the engine to make internationalization as easy as the rest of the game making process.

dsnopek:
> But just removing it makes this intro paragraph pretty bland IMO, and I think we should thrive for making the content better and more inclusive, not just "sanitized" to avoid another time-wasting Twitter controversy :)

Yeah, that version was more bland. I've pushed a new version that completely rewrites the first paragraph to have more personality, and tries to express the diverse nature of the Godot community.

dsnopek:
Ah, I walked away from my computer for a couple minutes because my kids woke up, and when I came back to hit submit, it was already merged!

akien-mga:
Hehe interesting timing, I guess I merged right before your new push, so the follow up might need a new PR.

Calinou:

PR #4722, opened 2021-03-04
Revert "Remove controversial satirical piece 🔥"

cryptoquick:
This reverts commit b872229427dddb9b749f46af597e85e25cf2955a.

I'm submitting this because this not only includes some helpful information, but also, it contains counter-culture sentiment, which is valuable. The Godot community stands up to exploitative practices rampant in the game industry. It's okay to express frustration towards that.

Further, it also includes useful information, but I'm firmly against scrubbing the text of character for the sake of avoiding controversy. To let this slide is to let the forces of corporate blandification win.

### DIFF
--- a/tutorials/io/encrypting_save_games.rst
+++ b/tutorials/io/encrypting_save_games.rst
@@ -1,0 +1,86 @@
+.. _doc_encrypting_save_games:
+
+Encrypting save games
+=====================
+
+Why?
+----
+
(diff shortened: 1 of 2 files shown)


pycbouh:
The information that can be found useful from this article will be added to the related articles in the documentation later on, but hopefully soon. The decision to remove the page will not be reverted. Please, let the community keep its peace finally and refrain from further attempts to revert it.

PR #4731, opened 2021-03-08
Fix GDScript code sample in Matrices and transforms

samlaf:

### DIFF
--- a/tutorials/math/matrices_and_transforms.rst
+++ b/tutorials/math/matrices_and_transforms.rst
@@ -478,8 +478,8 @@ the code we would use:
  .. code-tab:: gdscript GDScript
 
     # Set up transforms just like in the image, except make positions be 100 times bigger.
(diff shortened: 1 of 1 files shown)


Calinou:
@samlaf Out of curiosity, what does this change bring to the table?

samlaf:
Ah, sorry should have mentioned more. I thought this would be obvious to main/core developers.
The code as was didn't run for me. I needed to change it to use Vectors.
I was guessing either the API changed, or that was a mistake made from copying the C# call (which doesn't use vectors it seems).

Calinou (approved):
> Ah, sorry should have mentioned more. I thought this would be obvious to main/core developers.
The code as was didn't run for me. I needed to change it to use Vectors.
I was guessing either the API changed, or that was a mistake made from copying the C# call (which doesn't use vectors it seems).

Indeed, I forgot that there's no GDScript constructor for `Transform2D()` that takes 6 floats.

Thanks!

Calinou:

PR #4733, opened 2021-03-08
Use a separate GitHub Actions workflow to build offline documentation

Calinou:
This is required to get a permalink using https://nightly.link as we need the workflow containing the offline docs ZIP artifact to always be the latest one. Creating a second workflow is a good way to solve this.

This also makes it easier to distinguish between linting checks and offline documentation builds in the repository's Actions tab.

### DIFF
--- a/.github/workflows/build_offline_docs.yml
+++ b/.github/workflows/build_offline_docs.yml
@@ -1,0 +1,31 @@
+name: Build documentation for offline usage
+on:
+  schedule:
+    # Every week on Monday at midnight (UTC).
+    # This keeps the generated HTML documentation fresh.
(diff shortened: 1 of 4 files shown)


Calinou:
I added references to the generated artifact using [nightly.link](https://nightly.link). These links won't work until this pull request has been merged *and* the workflow has been run once using the schedule defined on Monday at midnight (UTC).

Due to this, I suggest we merge this pull request on a Sunday so that the link isn't invalid for too long :slightly_smiling_face:

Calinou:
Cherry-picked to the `3.3`, `3.2` and `stable` branches (with commits modified to have the URLs pertaining to each branch).

PR #4736, opened 2021-03-10
Recommend adding Godot folder to AV exceptions in Compiling for Windows

Calinou:
Antivirus programs will spend precious CPU time scanning the Godot source folder as files are added to it by the compiler.

### DIFF
--- a/development/compiling/compiling_for_windows.rst
+++ b/development/compiling/compiling_for_windows.rst
@@ -91,6 +91,18 @@ Downloading it (cloning) using `Git <https://git-scm.com/>`_ is recommended.
 The tutorial will assume from now on that you placed the source code in
 ``C:\godot``.
 
+.. warning::
+
+    To prevent slowdowns caused by continuous virus scanning during compilation,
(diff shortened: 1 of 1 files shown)


Calinou:

LikeLakers2:
I wouldn't actually recommend this. While it does slowdown the compilation process, that's offset massively by the AV being able to protect the user from infections.

And I know that Godot does not contain malicious code, but I still really wouldn't recommend this.

Calinou:
> I wouldn't actually recommend this. While it does slowdown the compilation process, that's offset massively by the AV being able to protect the user from infections.

The Godot source repository isn't something that arbitrary users can write into without requiring permission from anyone. Also, [antivirus programs are becoming less and less effective by the day](disable-windows-defender.github.io/).

LikeLakers2:
My concern is less about Godot's code (or it's compilation) having some malicious thing, and more about an actual virus or malicious program sitting in the excluded directory, undetected, because the user opted to exclude the folder. That's why I don't think it's worth making the compilation process a tiny bit faster.

PR #4742, opened 2021-03-11
Recommend triangulating and applying transforms in Importing 3D scenes

Calinou:
This is generally good advice for any 3D engine, but Godot also has issues with 3D selection that are more present with misaligned objects.

See https://github.com/godotengine/godot/issues/46852.


Calinou:

PR #4751, opened 2021-03-13
Heading Capitalization in the “about” Folder

MrBrain295:
This capitalizes main headings in the about folder.


Calinou:
Thanks for opening a pull request :slightly_smiling_face: 

Please note that we intentionally use Sentence case for headings rather than Title Case. Instead of changing all headings to use Title Case, we should change the headings that currently use Title Case to use Sentence case.

PR #4754, opened 2021-03-14
"Web" to "web"

MrBrain295:
This changes "Web" to "web" because according to [here](https://www.brookings.edu/blog/techtank/2016/06/01/associated-press-style-change-marks-growth-of-internet-and-web/) it should be "web not "Web.


Faless:
I don't see reasons for this change. Some sources removed the capitalization at some point, others maintains it.
In my opinion the latter is more explicit, so better, but in any case, I don't see compelling reasons to change it.

PR #4756, opened 2021-03-15
Add a section about editor tooltips blinking in the Troubleshooting page

Calinou:
See https://github.com/godotengine/godot/issues/32990.

@cosmoddd @Bargeral Is there a way to disable Fences temporarily, or does the user have to uninstall it entirely to use Godot without suffering from blinking editor tooltips?

### DIFF
--- a/about/troubleshooting.rst
+++ b/about/troubleshooting.rst
@@ -80,6 +80,13 @@ peripherals' drivers to their latest version. If the bug persists, you need to
 disconnect the faulty peripherals before opening the editor. You can then
 connect the peripheral again.
 
+Editor tooltips in the Inspector and Node docks blink when they're displayed.
+-----------------------------------------------------------------------------
+
+This is a `known issue <https://github.com/godotengine/godot/issues/32990>`__
+caused by the third-party Stardock Fences application on Windows.
+The only known workaround is to disable Stardock Fences while using Godot.
+
 The Godot editor appears frozen after clicking the system console.
 ------------------------------------------------------------------
 


akien-mga:
How popular/widely used is this Fences application? I'm not sure we should put prominent mention to issues which might only affect a very small proportion of users in a general Troubleshooting page for the whole engine.

Bargeral:
I don't know that Fences is very popular. I think it is enough to have this out there in some searchable fashion and hopefully people with the issue will find it.  Perhaps it should be worded more generally, such as "inspector tool-tip blinking when using desktop or icon management software such as Fences".  In any case, **thank you all** for your work on this! It was very frustrating and now that I know what's going on it's easy enough to work around. 

It is possible to disable fences temporarily.  

1. Right click on the desktop.
2. Choose "Configure Fences"
3. On the "Home" tab, at the bottom of the tab, un-select "Enable Fences" 

To re-enable
1. Right click on the desktop.
2. Choose "Enable Fences"

mhilbrunner:
> How popular/widely used is this Fences application? I'm not sure we should put prominent mention to issues which might only affect a very small proportion of users in a general Troubleshooting page for the whole engine.

Hard to judge, with Windows not really having a working distribution model for software and all. :P

I think like similar customization options like Rainmeter, if you look at *all* Windows users, the amount is probably miniscule, but in absolute numbers thats still a lot of people, and I have a feeling developers are more likely to customize their setup in this and similar ways and tend to be power users.

At least on a brief look, Fences *has* a ton of downloads on download portals wherever it's offered and often ranked mid to high compared to most other software.

E.g. the last freeware version has ~400k downloads on german website chip.de, with similar numbers on similar sites.

mhilbrunner:
I'm going to merge this for now, as the troubleshooting page *is* a good place to keep a tab on issues like these.

mhilbrunner:
Thank you!

Calinou:

PR #4757, opened 2021-03-16
Rewrite release policy to clarify versioning and include 3.3

akien-mga:
We are often asked if Godot follows SemVer, so I rewrote the relevant section
to clarify that our versioning is based on SemVer but takes some liberties to
accommodate the complexity of versioning so many interconnected components with
a single version number. Examples should help make the rationale clear.

Additionally, with the upcoming 3.3 release we're aiming at increasing the
release pace for future 3.x minor releases, and keep patch releases better
focused on actually being maintenance releases (with 3.2.2 being a significant
exception to that rule).

### DIFF
--- a/about/release_policy.rst
+++ b/about/release_policy.rst
@@ -3,48 +3,107 @@
 Godot release policy
 ====================
 
+Godot's release policy is in constant evolution. What is described below is
+intended to give a general idea of what to expect, but what will actually
+happen depends on the choices of core contributors, and the needs of the
+community at a given time.
+
 Godot versioning
 ----------------
 
-Godot uses ``major.minor.patch`` version numbering. However, it does not
-strictly follow `Semantic Versioning <https://semver.org/>`__. This means that
-releases considered "semver-minor" by that standard (such as 3.1 -> 3.2) will
-most likely introduce breaking changes. Still, there won't be as many breaking
-changes as a "semver-major" version bump such as 3.2 -> 4.0.
+Godot loosely follows `Semantic Versioning <https://semver.org/>`__ with a
+``major.minor.patch`` versioning system, albeit with an interpretation of each
+term adapted to the complexity of a game engine:
 
-In the interest of stability and usability, patch releases may occasionally
-introduce small breaking changes as well. When repackaging Godot projects (e.g.
-in a Flatpak), make sure to always use the same patch version as the one used to
-initially export the project.
+- The ``major`` version is incremented when major compatibility breakages happen
+  which imply significant porting work to move projects from one major version
+  to another.
+
+  For example, porting Godot projects from Godot 2.1 to Godot 3.0 required
+  running the project through a conversion tool, and then performing a number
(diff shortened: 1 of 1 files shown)


about/release_policy.rst
```diff
@@ -3,48 +3,107 @@
...
-introduce small breaking changes as well. When repackaging Godot projects (e.g.
-in a Flatpak), make sure to always use the same patch version as the one used to
-initially export the project.
+- The ``patch`` version is incremented for maintenance releases which focus on
+  fixing bugs and security issues, implementing new requirements for platform
+  support, and backporting safe usability enhancements. Patch releases are
+  backwards compatible.
+
+  Patch versions may include minor new features which do not impact the
+  existing API, and thus have no risk of impacting existing projects.
```
akien-mga, line 49: This is up for debate if we want to follow semver more strictly, but do we really want to prevent ourselves from e.g. adding a new method binding for scripts in stable branches?

Personally I'm of the opinion that this kind of change is safe, and that if it's needed by users *now*, we may as well provide the new API. It does mean that once users start using the new API in a patch release, they can't go back to older releases in that same branch without changing their code back.

about/release_policy.rst
```diff
@@ -3,48 +3,107 @@
...
++-------------+-------------------+--------------------------------------------------------------------------+
+| Godot 3.2   | January 2020      | |partial| Only critical, security and platform support fixes             |
+|             |                   | (last update: 3.2.3).                                                    |
 +-------------+-------------------+--------------------------------------------------------------------------+
-| Godot 3.1   | March 2019        | |partial| Only critical, security and platform support fixes.            |
+| Godot 3.1   | March 2019        | |partial| Only critical, security and platform support fixes             |
+|             |                   | (last update: 3.1.2).                                                    |
 +-------------+-------------------+--------------------------------------------------------------------------+
-| Godot 3.0   | January 2018      | |partial| Only critical, security and platform support fixes.            |
+| Godot 3.0   | January 2018      | |eol| No longer supported (last update: 3.0.6).                          |
```
akien-mga, line 101: Note: This makes 3.0 officially EOL. I haven't updated it in a year and I never hear from users who'd actually need a new release, so I think it's time to officially retire it.

about/release_policy.rst
```diff
@@ -57,30 +116,30 @@ for each Godot version:
...
+releases, so you can expect a new minor release every 3 to 6 months.
 
-.. seealso::
+Maintenance (patch) releases will be released as needed with potentially very
+short development cycles, to provide users of the current stable branch with
+the latest bug fixes for their production needs.
 
-    The `roadmap <https://github.com/godotengine/godot-roadmap>`__ repository
-    documents features that have been agreed upon and may be implemented in future
-    Godot releases.
```
akien-mga, line 86: Removed the roadmap link as the document is outdated. We can add something back once we define a new roadmap and strive to keep it up to date.

clayjohn (approved):

Calinou (approved):

PR #4760, opened 2021-03-18
Use "code editor" instead of "IDE" to refer to Visual Studio Code

bogdasar1985:
Visual Studio Code is NOT an IDE.

### DIFF
--- a/development/cpp/configuring_an_ide/visual_studio_code.rst
+++ b/development/cpp/configuring_an_ide/visual_studio_code.rst
@@ -3,7 +3,7 @@
 Visual Studio Code
 ==================
 
-`Visual Studio Code <https://code.visualstudio.com>`_ is a free cross-platform IDE
+`Visual Studio Code <https://code.visualstudio.com>`_ is a free cross-platform source-code editor
 by `Microsoft <https://microsoft.com>`_ (not to be confused with :ref:`doc_configuring_an_ide_vs`).
 
 Importing the project
--- a/development/cpp/configuring_an_ide/visual_studio_code.rst
+++ b/development/cpp/configuring_an_ide/visual_studio_code.rst
@@ -3,7 +3,7 @@
 Visual Studio Code
 ==================
 
-`Visual Studio Code <https://code.visualstudio.com>`_ is a free cross-platform source-code editor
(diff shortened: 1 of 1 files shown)


pycbouh:
As much as I agree with the sentiment, it's a big and ugly debate what constitutes an integrated development environment, and what is just a code editor buffed up with a lot of plugins. A lot of people call VS Code an IDE, and in the grand scheme of things *it's fine*.

So if you want to push for this change, you'd have to source some resource that would make a distinction for VS Code specifically. How does MS itself call it? What does Wikipedia say? That sort of thing.

LikeLakers2:
Microsoft says this, [here](https://code.visualstudio.com/docs/supporting/faq#_what-is-the-difference-between-visual-studio-code-and-visual-studio-ide):

> **What is the difference between Visual Studio Code and Visual Studio IDE?**
> Visual Studio Code is a streamlined code editor with support for development operations like debugging, task running, and version control. It aims to provide just the tools a developer needs for a quick code-build-debug cycle and leaves more complex workflows to fuller featured IDEs, such as Visual Studio IDE.

Meanwhile, [Wikipedia](https://en.wikipedia.org/wiki/Visual_Studio_Code) also calls it a code editor, in the first paragraph of the article:

> Visual Studio Code is a freeware source-code editor made by Microsoft for Windows, Linux and macOS. [9]

The source, by the way, links to <https://techcrunch.com/2015/04/29/microsoft-shocks-the-world-with-visual-studio-code-a-free-code-editor-for-os-x-linux-and-windows>, which has "code editor" in the title.

So I think it's pretty clear that it's a source-code editor despite its ability to support some of the same functions an IDE has.

pycbouh (approved):

development/cpp/configuring_an_ide/visual_studio_code.rst
```diff
@@ -3,7 +3,7 @@
 Visual Studio Code
 ==================
 
-`Visual Studio Code <https://code.visualstudio.com>`_ is a free cross-platform IDE
+`Visual Studio Code <https://code.visualstudio.com>`_ is a free cross-platform source-code editor
```
Calinou, line 6:
```suggestion
`Visual Studio Code <https://code.visualstudio.com>`_ is a free cross-platform code editor
```

Calinou:

PR #4762, opened 2021-03-18
Add a FAQ item about why Godot aims to keep its core feature set small

Calinou:
This is based on yesterday's conversation in the Godot contributors chat.

### DIFF
--- a/about/faq.rst
+++ b/about/faq.rst
@@ -207,6 +207,48 @@ In the long term, we may develop a Direct3D 12 renderer for Godot (mainly for
 the Xbox's purposes), but Vulkan and OpenGL will remain the default rendering
 backends on all platforms, including Windows.
 
+Why does Godot aim to keep its core feature set small?
+------------------------------------------------------
+
+Godot intentionally does not include features that can be implemented by add-ons
+unless they are used very often. One exemple of this would be advanced
+artificial intelligence functionality.
+
+There are several reasons for this:
+
+- **Code maintenance and surface for bugs.** Every time we accept new code in
+  the Godot repository, existing contributors often take the reponsibility of
+  maintaining it. Some contributors don't always stick around after getting
+  their code merged, which can make it difficult for us to maintain the code in
+  question. This can lead to poorly maintained features with bugs that are never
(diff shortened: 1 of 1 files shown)


akien-mga:
Binary size is one thing but it's not the main reason why we have to be selective with what we include. The problem is not with Godot becoming 200 MB for the editor, it's that all added code requires maintenance, increases compilation time, increases the potential surface area for bugs to be found in the engine, etc. More API to be familiar with to be able to contribute to a given area of the engine, etc.

I think that reducing it to file size concerns for download speeds is not describing the whole picture, even if it's one relevant factor.

Calinou:
> Binary size is one thing but it's not the main reason why we have to be selective with what we include. The problem is not with Godot becoming 200 MB for the editor, it's that all added code requires maintenance, increases compilation time, increases the potential surface area for bugs to be found in the engine, etc. More API to be familiar with to be able to contribute to a given area of the engine, etc.
> 
> I think that reducing it to file size concerns for download speeds is not describing the whole picture, even if it's one relevant factor.

Indeed, I think it makes sense to describe the full picture more. I reworded the FAQ item entirely.

about/faq.rst
```diff
@@ -207,6 +207,48 @@ In the long term, we may develop a Direct3D 12 renderer for Godot (mainly for
 the Xbox's purposes), but Vulkan and OpenGL will remain the default rendering
 backends on all platforms, including Windows.
 
+Why does Godot aim to keep its core feature set small?
+------------------------------------------------------
+
+Godot intentionally does not include features that can be implemented by add-ons
+unless they are used very often. A few examples of this would be artificial
+intelligence or terrain systems.
```
clayjohn, line 215: I wouldn't mention terrain here. The thinking around terrain has shifted and I think the only thing stopping it from being added is the amount of work necessary to do it properly/

clayjohn (approved):

akien-mga (approved):

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 3.2.

PR #4775, opened 2021-03-20
Simplify C# NuGet package doc to match 3.2.3+

31:
Now that 3.2.3+ uses the new csproj system and strongly requires .NET Core SDK 3.1+ (which has a modern MSBuild version), `PackageReference` got simpler for everyone. 🙂

I don't think I've actually used a `PackageReference` in a custom build of Godot Mono `master`, but I expect it's the same as 3.2.3. This change would make sense to cherry-pick back to 3.2 and 3.3, IMO.

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_basics.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_basics.rst
@@ -285,8 +285,8 @@ the ``.csproj`` file located in the project root:
         ...
     </Project>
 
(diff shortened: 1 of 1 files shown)


aaronfranke (approved):
Looks good, but I would appreciate a second review by @neikeq.

neikeq (approved):

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 3.3.

PR #4779, opened 2021-03-22
Fix heading capitalization in class_ysort.rst

MrBrain295:
According to #4751 only the first word needs capitalized in a heading.


pycbouh:
Hi! Class reference is autogenerated from the [main repo](https://github.com/godotengine/godot/tree/master/doc/classes), changes to content must happen there (read more about this [in the docs](https://docs.godotengine.org/en/latest/community/contributing/updating_the_class_reference.html)). That said, this part is generated in [a Python script](https://github.com/godotengine/godot/blob/master/doc/tools/makerst.py) for all the class reference entries, so the script must be adjusted instead. Though I'm not sure it's worth it for a small visual change.

PR #4784, opened 2021-03-28
Document how to avoid light leaks in GI Probes

Calinou:
This advice applies both to the new GIProbe in `master` and old GIProbe in `3.x`.
See discussion on Rocket.Chat: https://chat.godotengine.org/channel/rendering?msg=KoaoQZzvkojiHrdwi

See also https://github.com/godotengine/godot/pull/47428.


akien-mga:
This was cherry-picked already.

PR #4777, opened 2021-03-22
Update C# Connect method: `binds` is a Godot array

31:
It looks like since ~3.0.6, `Connect` takes a `Godot.Collections.Array`, not `object[]`, for `binds`. I think it's because we got `Godot.Collections.Array` around then: https://github.com/godotengine/godot/pull/15880.

(Someone tried to use this code as-is with `object[]` and asked about it in Discord, so I figured I'd follow up here. 🙂)

@neikeq

PR #4785, opened 2021-03-28
Fix C# IntersectRay usage: 'exclude' is a Godot Array

31:
Fix a couple instances in the raycast tutorial where the C# code used `object[]` instead of `Godot.Collections.Array`. This is similar to https://github.com/godotengine/godot-docs/pull/4777, the API change seems to have happened a while back and the tutorial hadn't been updated.

After this PR, a global search for `new object[]` has 0 results, so hopefully this is all for that particular API change. 😄


Calinou (approved):
Thanks!

PR #4683, opened 2021-02-26
Update HTML5 classref with autogenerated jsdoc.

Faless:
See godotengine/godot#46446

PR #4797, opened 2021-03-31
[3.2] Reverts HTML5 changes, fix github links

Faless:
This reverts #4683 from the `3.2` branch.
Changes are not there in `3.2.3`. Clearly a signal we needed `3.3` and that **I shouldn't have marked it as cherry-pickable** :disappointed: 

It also updates the reference to the custom HTML5 shell example to point directly to `3.2`.
This change is not needed in `3.3`/`master` since the link has been/is going to be removed.


akien-mga (approved):

akien-mga:
Thanks!

PR #4748, opened 2021-03-12
Update the "Customizing the HTML5 Shell" page

Faless:
Reflecting new changes in master, and upcoming in `3.2.4`.

See godotengine/godot-proposals#2182 and it's final implementation godotengine/godot#46200 for reference.

PR #4454, opened 2020-12-13
Update documentation about HTML5 exporting and compiling

Faless:
The "Custom HTML page for Web export" and "HTML5 shell class reference" will need some re-write too but probably also more discussion. So I would keep them separated unless you prefer otherwise.

In this PR:
- Bump emscripten version requirements.
- Add informations about Threads and GDNative variant.
- Explain "Vram compression" export option.
- Reorganize and update "Limitiations" section (secure contexts, networking, gamepads).
- Add information to build all the export template variants.

Should I also add a section: "Known browser bugs" to address godotengine/godot#29053 and potentially similar issues?

PR #4798, opened 2021-03-31
[3.3] [HTML5] Backport HTML5 changes from master

Faless:
Backports #4454 and #4748 to `3.3`.


getting_started/workflow/export/exporting_for_web.rst
```diff
@@ -12,27 +12,11 @@ in the user's browser.
...
                (regardless of the browser). We recommend using
                :ref:`iOS' native export functionality <doc_exporting_for_ios>`
                instead, as it will also result in better performance.
 
-.. note::
-
-    If you use Linux, due to
-    `poor Firefox WebGL performance <https://bugzilla.mozilla.org/show_bug.cgi?id=1010527>`__,
-    it's recommended to play the exported project using a Chromium-based browser
-    instead of Firefox.
```
akien-mga, line 34: Did you intend to remove this note?
It physically hurts me to recommend using Chrome on Linux, but I think it's sadly still true (though I resist and still use FF to play Godot web exports :P).

Faless: That was probably a failed merge on my part. I'll re-check the PR myself.
That said, on the specific topic, as we have seen with the recent regression, suggesting using Chrome right now will give a very bad impression if the game has any audio.

Which is why, in general, I don't like we taking the role of evaluating browsers "performances", and I'd rather point user to feature supports (e.g. can I use/mdn docs), and point out mileage might vary with different browsers, pointing maybe to "lead browsers" bug trackers.

akien-mga:
Thanks!

PR #4807, opened 2021-04-03
Validation layers on Android

szymonm-google:
A manual describing the process of enabling Vulkan validation layers on Android. The corresponding PR with the implementation is here: https://github.com/godotengine/godot/pull/47552.


development/cpp/validation_layers_android.rst
```diff
@@ -0,0 +1,49 @@
+.. _doc_validation_layers_android:
+
+Vulkan validation layers on Android
+===================================
+
+Validation layers enable developers to verify their application's correct use
+of the Vulkan API. After enabling validation layers a developer can see errors
+and warning messages in the `logcat's` output.
```
Calinou, line 8:
```suggestion
Validation layers enable developers to verify their application's correct use
of the Vulkan API. After enabling validation layers, a developer can see errors
and warning messages in the ``adb logcat`` output.
```

development/cpp/validation_layers_android.rst
```diff
@@ -0,0 +1,49 @@
...
+
+Enabling validation layers
+--------------------------
+
+Build validation layers from official sources
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To build Android libraries, follow the instructions on `Khronos' repository 
+<https://https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/master/BUILD.md#building-on-android>`__.
+After a successful build, the libraries will be located in ``Vulkan-ValidationLayers/build-android/libs``.
```
Calinou, line 18:
```suggestion
To build Android libraries, follow the instructions on
`Khronos' repository  <https://https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/master/BUILD.md#building-on-android>`__.
After a successful build, the libraries will be located in ``Vulkan-ValidationLayers/build-android/libs``.
```

development/cpp/validation_layers_android.rst
```diff
@@ -0,0 +1,49 @@
...
+        |-- android
+            |-- java
+                |-- app
+                    |-- libs
+                        |-- debug
+                            |-- vulkan_validation_layers
+                                |-- arm64-v8a
+                                |-- armeabi-v7a
+                                |-- x86
+                                |-- x86_64
```
Calinou, line 41:
```suggestion
Your Godot's directory tree should look like on the example below::

    godot
    |-- platform
        |-- android
            |-- java
                |-- app
                    |-- libs
                        |-- debug
                            |-- vulkan_validation_layers
                                |-- arm64-v8a
                                |-- armeabi-v7a
                                |-- x86
                                |-- x86_64
```

development/cpp/validation_layers_android.rst
```diff
@@ -0,0 +1,49 @@
...
+
+If the subdirectory ``libs/debug/vulkan_validation_layers`` doesn't exist, create it.
+
```
Calinou, line 44:
```suggestion
If the subdirectory ``libs/debug/vulkan_validation_layers`` doesn't exist, create it.
```

development/cpp/validation_layers_android.rst
```diff
@@ -0,0 +1,49 @@
...
+
+Compile and run android app
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
Calinou, line 47:
```suggestion
Compile and run the Android app
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

development/cpp/validation_layers_android.rst
```diff
@@ -0,0 +1,49 @@
...
+
+Linked validation layers are automatically loaded and enabled in Android debug builds.
```
Calinou, line 49:
```suggestion
Linked validation layers are automatically loaded and enabled in Android debug builds.
You can use Godot's :ref:`doc_one-click_deploy` feature to quickly test your project with the validation layers enabled.
```

Calinou (commented):
I think this page would be better listed in https://github.com/godotengine/godot-docs/tree/master/tutorials/platform/android rather than C++ development. People who are troubleshooting bugs in their own project may need to enable Vulkan validation layers, even if they don't wish to dive into Godot's C++ code.

development/cpp/validation_layers_android.rst
```diff
@@ -0,0 +1,48 @@
...
+                    |-- libs
+                        |-- debug
+                            |-- vulkan_validation_layers
+                                |-- arm64-v8a
+                                |-- armeabi-v7a
+                                |-- x86
+                                |-- x86_64
+
+If the subdirectory ``libs/debug/vulkan_validation_layers`` doesn't exist, create it.
+
```
Calinou, line 42:
```suggestion
If the subdirectory ``libs/debug/vulkan_validation_layers`` doesn't exist, create it.
```

Calinou, line 42:
```suggestion
If the subdirectory ``libs/debug/vulkan_validation_layers`` doesn't exist, create it.
```

development/cpp/validation_layers_android.rst
```diff
@@ -0,0 +1,48 @@
...
+
+Compile and run the Android app
```
Calinou, line 44:
```suggestion
If the subdirectory ``libs/debug/vulkan_validation_layers`` doesn't exist, create it.

Compile and run the Android app
```

szymonm-google:
> I think this page would be better listed in https://github.com/godotengine/godot-docs/tree/master/tutorials/platform/android rather than C++ development. People who are troubleshooting bugs in their own project may need to enable Vulkan validation layers, even if they don't wish to dive into Godot's C++ code.

Good point! I've moved it to the proper section.

akien-mga:
> I think this page would be better listed in [`master`/tutorials/platform/android](https://github.com/godotengine/godot-docs/tree/master/tutorials/platform/android) rather than C++ development. People who are troubleshooting bugs in their own project may need to enable Vulkan validation layers, even if they don't wish to dive into Godot's C++ code.

I don't think that's the right category for this. Using validation layers is something that will typically be done by engine developers, not engine users. And it's something which requires compiling Godot templates from source anyway, so it should be part of the engine development section.

So the original location was fine IMO (the section folder `cpp` is a misnomer but otherwise it's the correct "Engine development" section). It could also be referenced in the "Compiling for Android" page with a `.. seealso::` directive.

akien-mga:
That being said it's true that it's nice to have it under an Android-specific section, so maybe the change suggested by @Calinou is fine. But this should be done still:

> . It could also be referenced in the "Compiling for Android" page with a .. seealso:: directive.

m4gr3d:
> That being said it's true that it's nice to have it under an Android-specific section, so maybe the change suggested by @Calinou is fine. (WDYT @m4gr3d?)
> But this should be done still:
> 
> > It could also be referenced in the "Compiling for Android" page with a .. seealso:: directive.

I agree that the section should be under `Engine Development` instead especially given that vulkan validation layers are not specific to Android and so this entry is likely to grow with instructions for other platforms.
`Vulkan` itself is a big enough topic that it may make sense to create a `Vulkan` page under `Engine Development` which can be a parent for the Android vulkan validation layers page.

tutorials/platform/android/index.rst
```diff
@@ -7,3 +7,4 @@ Android plugins
 
    android_plugin
    android_in_app_purchases
+   validation_layers_android
```
m4gr3d, line 10: If we decide to keep it here than the page title would definitely have to be updated. But I'm strongly in favor of moving it under `Engine Development`:
`Engine Development` -> `Vulkan` -> `Validation Layers` -> `Android` (the last one could just be a section in the validation layers page).

szymonm-google: @m4gr3d Thanks for the comments. Are you ok with creating this structure with empty chapters for now?

m4gr3d: Yeah that sounds good.

akien-mga: Good idea, we'll likely want to add some instructions on how to set up and use validation layers for other platforms indeed. One page for all platforms should likely be fine, starting with just Android for now.

szymonm-google: I moved the chapter to  `Engine Development` -> `Vulkan` -> `Validation Layers`

development/cpp/vulkan/vulkan_validation_layers.rst
```diff
@@ -0,0 +1,54 @@
...
+warning messages in the ``adb logcat`` output.
+
+Enabling validation layers
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Build validation layers from official sources
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+To build Android libraries, follow the instructions on
+`Khronos' repository  <https://https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/master/BUILD.md#building-on-android>`__.
```
BastiaanOlij, line 24: Note the double https:// here

szymonm-google: great catch!

development/compiling/compiling_for_android.rst
```diff
@@ -113,6 +113,12 @@ The resulting APK will be located at ``bin/android_release.apk``.
 
 The resulting APK will be located at ``bin/android_debug.apk``.
 
+
+.. seealso::
+
+    If you want to enable Vulkan validation layers, see
+    :ref:`Vulkan validation layers on Android<doc_vulkan_validation_layers-android>`.
```
m4gr3d, line 120: @szymonm-google Just double-checking, is this the right path?

m4gr3d: Nevermind, that looks correct.

m4gr3d (approved):
Looks great!

development/cpp/index.rst
```diff
@@ -18,3 +18,4 @@ Engine development
    custom_resource_format_loaders
    custom_audiostreams
    custom_godot_servers
+   vulkan/index
```
akien-mga, line 21: I know we're going back and forth but I'm not sure if we really need a dedicated `vulkan` folder/separate table of content.

Do we intend to have more documentation page relative to Vulkan development? Otherwise this could be `development/cpp/vulkan_validation_layers.rst` directly, without adding an extra level.

If we plan to write more documentation about Vulkan, or about rendering, then making a dedicated Vulkan (or Rendering) sub-section is fine with me. CC @godotengine/rendering

It's not necessarily blocking for this PR though, we can always merge and move the file around later if we change our mind (that just means a 404 on the original link unless we set up a redirection).

m4gr3d: @akien-mga I think having a dedicated vulkan folder makes sense.
For one setting up vulkan on platform like macos (e.g: moltenVk) is non-trivial so some additional setup documentation could be added in this section.

In addition, given that the standard is still evolving and there might be some large difference between desktop and mobile, this section could also be used to highlight that.

szymonm-google: I was under the impression that we all agreed on Fredia's suggestion, that Vulkan is a big topic and it deserves a dedicated category. I think that there will be more docs regarding Vulkan in the future.

akien-mga: Sounds good, let's go with this then :)

akien-mga:
Thanks!

PR #4824, opened 2021-04-09
Add a FAQ item about the UI toolkit used by Godot

Calinou:
The exact wording could be improved, but I feel it's a common enough question to warrant a FAQ item.

### DIFF
--- a/about/faq.rst
+++ b/about/faq.rst
@@ -361,6 +361,8 @@ developer experiences as a whole.
 Bonus points for bringing screenshots, concrete numbers, test cases, or example
 projects (if applicable).
 
+.. _doc_faq_non_game_applications:
+
 Is it possible to use Godot to create non-game applications?
 ------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


about/faq.rst
```diff
@@ -391,6 +395,26 @@ engine instead. Keep in mind rendering engines usually have smaller communities
 compared to Godot. This will make it more difficult to find answers to your
 questions.
 
+What user interface toolkit does Godot use?
```
Calinou, line 398: Is Godot's UI system considered to be *immediate mode*, *retained mode* or something in between? This would be nice to mention while we're at it.

akien-mga: I don't know, but let's merge anyway as the added section is good.

akien-mga (approved):

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 3.3.

PR #4829, opened 2021-04-10
Mention that you can link to the Godot license to comply with it

### LINKED ISSUE #4828: Complying with licenses contradicts godotengine.org/license
**Your Godot version:**
3.2.3
**Issue description:**
"Complying with licenses" says that the dev must "include the license text" while godotengine.org/license states that the dev just has to link to that page.
https://docs.godotengine.org/en/latest/about/complying_with_licenses.html

skyace65:
Adds that you can add a link to the website page https://godotengine.org/license to comply with the Godot license. That page says this is acceptable. Closes #4828

### DIFF
--- a/about/complying_with_licenses.rst
+++ b/about/complying_with_licenses.rst
@@ -78,6 +78,13 @@ Printed manual
 
 If the game includes printed manuals, license text can be included there.
 
+Link to the license
+^^^^^^^^^^^^^^^^^^^
+
+The Godot Engine developers consider that a link to godotengine.org/license
+in your game documentation or credits would be an acceptable way to satisfy
+the license terms. 
+
 Third-party licenses
 --------------------
 


Calinou:
My issue with this is that it's not a great way to achieve compliance. If `godotengine.org/license` is no longer a valid URL in the future, then you start breaching the license.

LikeLakers2:
@Calinou But on <https://godotengine.org/license>, it says:

> The Godot Engine developers consider that a link to this page (godotengine.org/license) in your game documentation or credits would be an acceptable way to satisfy the license terms. 

which seems to imply that the link *won't* stop working in the future (even if it needs to be redirected), since you've already offered linking to that page as a means of complying with the license.

Calinou:
> which I feel puts the onus on the Godot team to prevent that link from becoming invalid, as it implies a promise that the /license URL will never become an invalid URL (even if it becomes a redirect).

We can't guarantee that the Godot website will be up forever, or that it reaches 100% of effective uptime. Who knows if Godot will still exist in 20, 30, 50 years from now :slightly_smiling_face:

LikeLakers2:
Looking through `git blame`, [that bit has been there for four years](https://github.com/godotengine/godot-website/blame/285f8dffa0daaf16daf54eba433f65f00cc87f21/themes/godotengine/pages/license.htm#L32-L33). There's a decent chance that some people have taken advantage of that note as a means to satisfy Godot's MIT license.

Additionally, if the website stops functioning, that means the project has probably been abandoned. At that point, I don't know who would even *care* if the license was breached.

Also, your bit about it not having 100% uptime: I'm *really* not sure that would matter in a copyright court? Websites have downtime all the time, even ones with legal pages, so to say that the website having some downtime could bring about a breach of license (especially since that's not the fault of the one who linked the license page)... it sounds kinda silly.

That said, if you're still concerned about this, then I would recommend removing that bit from the website: https://github.com/godotengine/godot-website/blob/33f5e7486fbc47c07bb5c1cd9774c169c1379467/themes/godotengine/pages/license.htm#L35-L38

But I really don't see why you're concerned, Calinou.

skyace65:
@akien-mga what are your thoughts? Should it still be acceptable to link to the license instead of including it in a game or project?

golddotasksquestions:
> We can't guarantee that the Godot website will be up forever, or that it reaches 100% of effective uptime. Who knows if Godot will still exist in 20, 30, 50 years from now slightly_smiling_face
> 
> I understand that it's slightly less convenient to have to include the license in your project, but it's far more robust.

What you are saying makes a lot of sense to me, however then we can't tell (allow) people to just link to the license. It's one or the other.

mhilbrunner:
I'm obiously no Akien (:P), but as the official website, on the license page no less states
> The Godot Engine developers consider that a link to this page (godotengine.org/license) in your game documentation or credits would be an acceptable way to satisfy the license terms. 
..I think its okay to have that same information reflected in the docs, *especially* since that very page later links to the docs, so not having it reflected there is actively confusing.

I'll merge this, *if* we change our minds ont whether this is acceptable, we need to change both this *and the official website* anyway. :)

Calinou:

PR #4835, opened 2021-04-12
fix the doc for skip_vertex_transform.

mortarroad:
Make documentation consistent with https://github.com/godotengine/godot/blob/f72c91e0b17e6070051dec8fe4b0425dbb02c6d6/drivers/gles3/shaders/scene.glsl#L463


akien-mga:
Just noticed that this was a PR for `3.3`... please make PRs against the development branches (`master` and `3.x`), not `stable`. There's little point in updating documentation only for a previously released version and not future ones.

mortarroad:
I was under the assumption - since 4.0 has Vulkan and stuff - that all of this had been rewritten anyway and therefore did not apply anymore.

akien-mga:
You might be right yes. That documentation is still present but might be obsolete.

PR #4842, opened 2021-04-16
Describe the process of writing integration tests for GDScript

Xrayez:
See godotengine/godot#47701.

Closes godotengine/godot-proposals#1429.

### DIFF
--- a/development/cpp/unit_testing.rst
+++ b/development/cpp/unit_testing.rst
@@ -288,3 +288,62 @@ Here's an example of how GDScript registers test tools in
 
 The custom command-line parsing can be performed by a test tool itself with the
 help of OS :ref:`get_cmdline_args<class_OS_method_get_cmdline_args>` method.
+
+Integration tests for GDScript
+------------------------------
+
+Godot uses doctest to prevent regressions in GDScript during development. There
+are several types of test scripts which can be written:
+
+- tests for expected errors;
+- tests for warnings;
+- tests for features.
+
+Therefore, the process of writing integration tests for GDScript is the following:
+
+1. Pick a type of a test script you'd like to write, and create a new GDScript
+   file under the ``modules/gdscript/tests/scripts`` directory within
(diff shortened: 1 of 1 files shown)


Calinou (approved):

development/cpp/unit_testing.rst
```diff
@@ -288,3 +288,56 @@ Here's an example of how GDScript registers test tools in
...
+            if true # Missing colon here.
+                print("true")
+
+3. Generate ``*.out`` files to update the expected results from the output:
+
+   .. code-block:: shell
+
+       ./bin/<godot_binary> --gdscript-generate-tests godot-source/modules/gdscript/tests/scripts
+
+4. Run GDScript tests with:
```
vnen, line 328: I think the not to check if the `*.out` files are expected should come before this step. If you generate the outputs and immediately run the test, it will always pass.

It should be clear also that if generating change other tests that you didn't add, you should revert it back and not commit.

vnen:
In general looks good to me, apart for that small comment.

development/cpp/unit_testing.rst
Xrayez: Not sure if I understand correctly. Is it possible to run tests for new scripts that don't have `*.out` files generated for them yet? Not sure what would be the point of that, since there will be nothing to compare the output against.

Xrayez: What I mean is that, the process describes how to write *new* tests, not updating existing ones. If you believe it's beneficial to describe those processes separately, then it may be worth it. But I'm afraid this could potentially make the documentation more confusing.

Calinou went through those steps successfully while making #48029, for instance. If someone is going to have a problem, the documentation could be further clarified I guess.

That said, not sure what kind of changes to make.

Xrayez: Added an explicit warning:

image

akien-mga:
Thanks!

PR #4847, opened 2021-04-17
Marking overlapp area (grey) in 3x3minimal wall tilemap templates 

golddotasksquestions:
Currently the area I mapped here in grey is was marked yellow. Yellow is supposed to mark an area that shows the top of the wall. This area marked in grey is never shown as top of the wall, because other tiles are suppose to overlap it. I hope marking it in grey and adding a text in the description of the colors makes it more clear how such a Tileset ought to be drawn.

This PR changes the wall template images and adds a color description line to the tutorial.


Calinou:
Superseded by https://github.com/godotengine/godot-docs/pull/4848.

PR #4850, opened 2021-04-18
Fix entry type in C# example for loading savegames

evroon:
Fixes a small bug in the documentation of loading savegames using C#.

`nodeData` has type `Godot.Collections.Dictionary<string, object>`, so en element of the dictionary has type `KeyValuePair<string, object>`


Calinou (approved):

PR #4854, opened 2021-04-21
Release policy: Add 3.4 beta, mark 3.2 as EOL

akien-mga:
It's a bit unfortunate that we have to mark 3.2 EOL so early (compared to our promise of long-term support), but the 3.3 release blog post should make this clear to users (3.3 is what would have been 3.2.4, so it's the next "maintenance update" for 3.2.3 users - so there's no real possibility nor need to make a different 3.2.4 release).


Calinou (approved):

akien-mga:

PR #4857, opened 2021-04-21
Minor change to section on for loop syntax

saolof:
Modified a section that seemed slightly confusing.

The C-like for loop is specific to languages that wanted to copy C syntax and has nothing to do with static vs dynamic typing. Most new languages do not have it and have a foreach loop over iterators instead, regardless of whether they are statically typed or not.

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_advanced.rst
+++ b/tutorials/scripting/gdscript/gdscript_advanced.rst
@@ -318,7 +318,7 @@ states and quick structs:
 For & while
 -----------
 
-Iterating in some statically typed languages can be quite complex:
+Iterating in C-derived languages can be quite complex:
 
 .. code-block:: cpp
 
@@ -338,7 +338,7 @@ Iterating in some statically typed languages can be quite complex:
         std::cout << *it << std::endl;
     }
 
-This is usually greatly simplified in dynamically typed languages:
+This is usually greatly simplified in modern languages using for loops over iterables:
 
 ::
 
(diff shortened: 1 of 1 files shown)


pycbouh:
I'm not sure why you put C-derived and modern on the opposite ends, but I'm not sure why the original text compares static and dynamic typing here. Those two variants of loops are nothing but syntax sugar.

Also, in the first part it is also a "for" loop. If you want to distinguish the more "modern" syntax, you should refer to it as "for-in" loop.

saolof:
C-derived and modern are not on opposite ends here. But several dynamic languages designed back when C was hyper-ubiquitous have the C for loop, Perl and Javascript for example.  "Modern" just refers to the fact that the for-in loop is a newer invention

pycbouh:
JavaScript has for-in as well, and also a for-of. But my point is that the way your PR is currently written it kind of tells the user that C-derived languages have some outdated format, that "modern" languages fix. I think it's a disingenious concept to convey. But, as I've said, I'm also not sure why there was a comparison between dynamic and static typingm which variations of for loop have nothing to do with.

This needs improvement, but I'm not sure the way you've worded it currently is the correct way to improve it.

saolof:
In that case the best option might be to delete the comparison to a C-like for loop entirely, since it makes the section longer without actually adding much of value for someone learning GDscript.

saolof:
Changed it to just stating that GDScript makes the opinionated decision to have a for-in loop over iterables instead of a C-style loop, and made it read more like a direct translation guide

mhilbrunner:
Thanks and congrats on your first contribution!

PR #4858, opened 2021-04-22
Complying with Licenses: Clarify which fonts FreeType renders

DeeJayLSP:
Although most people only use them already it would be good to point out that FreeType only works with DynamicFonts

### DIFF
--- a/about/complying_with_licenses.rst
+++ b/about/complying_with_licenses.rst
@@ -93,7 +93,7 @@ Here's a list of libraries requiring attribution:
 FreeType
 ^^^^^^^^
 
-Godot uses `FreeType <https://www.freetype.org/>`_ to render fonts. Its license
+Godot uses `FreeType <https://www.freetype.org/>`_ to render :ref:`DynamicFonts <class_DynamicFont>`. Its license
 requires attribution, so the following text must be included together with the
 Godot license:
 


about/complying_with_licenses.rst
```diff
@@ -93,7 +93,7 @@ Here's a list of libraries requiring attribution:
 FreeType
 ^^^^^^^^
 
-Godot uses `FreeType <https://www.freetype.org/>`_ to render fonts. Its license
+Godot uses `FreeType <https://www.freetype.org/>`_ to render :ref:`DynamicFonts <class_DynamicFont>`. Its license
```
akien-mga, line 96: DynamicFont doesn't exist anymore in 4.0 so it's not a good change for `master`. Maybe it should spell "TTF and OTF fonts" instead? @bruvzg

While at it, please try to keep the manual wrapping consistent.

bruvzg: I do not think it's necessary to specify it.
The default font is TTF (https://github.com/godotengine/godot/pull/39603).
And even if it's not used, FreeType is compiled in and attribution is required anyway.

akien-mga: I agree. Users can build Godot without freetype if they want but at this stage they should be aware of what they're doing and this documentation tweak wouldn't matter much.

PR #4863, opened 2021-04-23
Update `latest` warning for godot 3.x

Tadaboody:
Bumping this after the new stable update.
Maybe it should say 3.x.x instead?


akien-mga:
It would be better to use `3.x`.

akien-mga (approved):

akien-mga:
Thanks!

PR #4872, opened 2021-04-26
Document how to support portrait and landscape in Multiple resolutions

Calinou:
This has been a recurring question lately. See https://godotforums.org/discussion/comment/48187 and https://godotforums.org/discussion/26128/customize-fit-text-in-a-label-with-font-size-adjustment-dynamic-font-size.


Calinou:

Issue #4796, opened 2021-03-31
Android Export Fails: 'apksigner' returned with error #1


JohnCWakley:
**Your Godot version:**
v3.3.rc7.official

**Issue description:**
I followed the documentation page (URL below) to setup for Android Exports. When attempting to export to Android, I received the following error: **'apksigner' returned with error #1**

From the command-line, I ran apksigner and received the following error:
```
$ ./apksigner.bat
Error: A JNI error has occurred, please check your installation and try again
Exception in thread "main" java.lang.UnsupportedClassVersionError: com/android/apksigner/ApkSignerTool has been compiled by a more recent version of the Java Runtime (class file version 53.0), this version of the Java Runtime only recognizes class file versions up to 52.0
        at java.lang.ClassLoader.defineClass1(Native Method)
        at java.lang.ClassLoader.defineClass(ClassLoader.java:756)
        at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
        at java.net.URLClassLoader.defineClass(URLClassLoader.java:468)
        at java.net.URLClassLoader.access$100(URLClassLoader.java:74)
        at java.net.URLClassLoader$1.run(URLClassLoader.java:369)
        at java.net.URLClassLoader$1.run(URLClassLoader.java:363)
        at java.security.AccessController.doPrivileged(Native Method)
        at java.net.URLClassLoader.findClass(URLClassLoader.java:362)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:418)
        at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:352)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
        at sun.launcher.LauncherHelper.checkAndLoadMain(LauncherHelper.java:601)
```

During my investigation, I discovered that build-tools;30.0.1 requires JDK 9. For testing, I ran the following:
```
$ sdkmanager.bat --uninstall "build-tools;30.0.1"
$ sdkmanager.bat --install "build-tools;29.0.3"
$ ./apksigner.bat
USAGE: apksigner <command> [options]
       apksigner --version
       apksigner --help
```

**Solution:**
Uninstall build-tools 30.0.1 and install build-tools 29.0.3.
Also, the doc page below should be updated to use build-tools 29.0.3.

**This issue, and it's solution, works on both Windows 10 and Linux (Solus 4.2)**

**URL to the documentation page:**
[...]

PR #4873, opened 2021-04-27
Update build-tools to 30.0.3

allkhor:
Build tools 30.0.1 had issues with compatibility with JDK 8 and 11.
Fixes https://github.com/godotengine/godot-docs/issues/4796

### DIFF
--- a/development/compiling/compiling_for_android.rst
+++ b/development/compiling/compiling_for_android.rst
@@ -61,7 +61,7 @@ Setting up the buildsystem
 
     ::
 
-        tools/bin/sdkmanager --sdk_root=<android_sdk_path> "platform-tools" "build-tools;30.0.1" "platforms;android-29" "cmdline-tools;latest" "cmake;3.10.2.4988404"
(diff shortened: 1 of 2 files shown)


akien-mga:
PRs should be made against the development branch (`master`) and they are then cherry-picked, when relevant, to the stable branches.

m4gr3d (approved):
Thanks for the update!

m4gr3d:
@allkhor Based on the comments in #4796 it seems we can also remove the restriction on JDK 8. Were you able to validate that it also works with the latest JDK version?

akien-mga:
At least on my system (Mageia 9 Linux) I can't get an export to work with OpenJDK 11, neither using 30.0.1 nor 30.0.3.

It fails with:
```
A problem occurred evaluating root project 'build'.
> Could not initialize class sun.security.pkcs.SignerInfo
```

But I suspect this might be an issue with my distro's JDK 11 (derived from Fedora's) and it might not be a problem for all JDK 11 users.

akien-mga:
Thanks!

akien-mga:
Cherry-picked for `master`.

Issue #4866, opened 2021-04-24
Don't name tutorials for a class the same as the class to help with search engines


LightningAA:
**Your Godot version:**
3.3.0
**Issue description:**
When searching on search engine™ for a given class, often enough there are 2 results from the stable godot docs, one is a tutorial and one is the actual api reference, and without reading the synopsis (or I just realized you can look at the url but that's not the most intuitive, most people just look at the titles and then if necessary the descriptions) below each result it's not easy to tell right away which is which slightly annoying.

I think all tutorials for a specific class should be affixed with "How to use `ClassName`s" or something along those lines (some tutorials are prefixed with `Using` but it's not very consistent).

Examples:
- Rigidbody vs Rigidbody
- InputEvent vs InputEvent
- Viewport vs Viewports (the ones with "s" are more clear, but I think "How to use Viewports" would still be more clear)
- Container vs Containers
- VisualShader vs VisualShaders

![image](https://user-images.githubusercontent.com/50304111/115947276-dec42680-a483-11eb-92b6-33f62132881d.png)

**URL to the documentation page (if already existing):**

PR #4880, opened 2021-04-27
Rename documentation pages that have the same name as classref pages

Calinou:
This prevents confusion in search engine results. I kept the existing file names as to not break URLs.

This PR can be cherry-picked to stable branches without conflicts.

This closes https://github.com/godotengine/godot-docs/issues/4866.

### DIFF
--- a/tutorials/2d/using_tilemaps.rst
+++ b/tutorials/2d/using_tilemaps.rst
@@ -1,6 +1,6 @@
 .. _doc_using_tilemaps:
 
-Using tilemaps
+Using TileMaps
 ==============
 
 Introduction
--- a/tutorials/3d/gi_probes.rst
+++ b/tutorials/3d/gi_probes.rst
@@ -1,7 +1,7 @@
 .. _doc_gi_probes:
 
-GI Probes
-=========
+Using GIProbe
+=============
 
 Introduction
 ------------
--- a/tutorials/3d/using_gridmaps.rst
+++ b/tutorials/3d/using_gridmaps.rst
@@ -1,6 +1,6 @@
 .. _doc_using_gridmaps:
 
-Using gridmaps
+Using GridMaps
 ~~~~~~~~~~~~~~
 
 Introduction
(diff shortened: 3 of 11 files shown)


LightningAA:
Still missing some, I only listed some of them in the original issue. Looking through the docs it looks like you're missing:
- SoftBody
- GDNative
- AnimationTree
- GI Probes (this one has a space so maybe it's fine?)

LightningAA:
Also idk if it's really necessary but `Using tilemaps` and `Using gridmaps` should probably have the class name be capitalized accordingly

Calinou:
> GDNative

This one is a difficult case because this is a category (index), not a leaf page.

LightningAA:
Maybe it could be called "GDNative examples" since it looks like that's all it contains

Calinou:
> Maybe it could be called "GDNative examples" since it looks like that's all it contains

It could be expanded to contain more pages that are not necessarily examples. For instance, the *What is GDNative?* page that's already present is not an example.

LightningAA:
Oh, I was looking at the 3.3 documentation

LightningAA:
Nitpicking, but I think it's kind of weird how of the titles end with `s` and some don't, maybe add `s` to the end of all of them?

Calinou:
> Nitpicking, but I think it's kind of weird how of the titles end with `s` and some don't, maybe add `s` to the end of all of them?

I'm not sure about this, as this will turn names like *SoftBody* into *SoftBodies*, which may harm SEO if people are looking for "softbody".

LightningAA:
I do think plural sounds better, but it's probably fine.

LightningAA:
Oh, looks like I also missed `SceneTree`

Calinou:
> Oh, looks like I also missed `SceneTree`

Renamed :slightly_smiling_face:

akien-mga (approved):

akien-mga:
Thanks!

akien-mga:

PR #4882, opened 2021-04-29
Fix a typo

Tobenaii:
Fixed a typo! :>

### DIFF
--- a/classes/class_csgmesh.rst
+++ b/classes/class_csgmesh.rst
@@ -16,7 +16,7 @@ A CSG Mesh shape that uses a mesh resource.
 Description
 -----------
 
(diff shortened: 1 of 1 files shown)


akien-mga:
Thanks for the contribution!

The fix is correct, but there are two issues:

- PRs should be made against the `master` branch so that the issue is fixed first and foremost for all future Godot versions. Then we cherry-pick relevant changes to stable branches (`3.3`, `3.2`, `stable`) ourselves.

- The class reference is generated from XML files in the Godot source repository, so it should not be modified here: https://github.com/godotengine/godot/blob/master/modules/csg/doc_classes/CSGMesh3D.xml

As this is a simple change I've batched it with another doc update that I just pushed to Godot's `master` branch: https://github.com/godotengine/godot/commit/cab5064f2084c49bd7d8064ae941a40da9986da2

PR #4893, opened 2021-05-05
Specify when integers and floats are 32 and when 64 bits

### LINKED ISSUE #4887: Binary serialization API documentation about float inconsistent
**Your Godot version:**
3.3-stable-official

**Issue description:**
In Binary serialization API documentation page for stable [1], float is reported to be 4 bytes.

In same page for latest [2], it's reported to be 8 bytes.

In the push comment that changed length from 8 to 4, [3] it says that it was a mistake for it to be 4 in first place, cause both integers and floats are 8 bytes due to Variants.

That commit changed both float and integer, and it happened before stable was released, however on stable integer is already reported as 8 bytes while float is 4 bytes.

Blaming the file on stable [4] shows that only part of the commit was still present.

I think that:
1. If float is 8 bytes, then:
    1. if it is 8 bytes also in current stable version, an errata should be added to the page
    2. if there was a merge conflict on stable branch that should be fixed cause may happen again
2. If float is 4 bytes, then latest documentation is wrong and should be fixed before next release

[1] https://docs.godotengine.org/en/stable/tutorials/misc/binary_serialization_api.html
[2] https://docs.godotengine.org/en/latest/tutorials/io/binary_serialization_api.html
[3] https://github.com/godotengine/godot-docs/commit/c66b287706eeb33ca73a3d1bbc5722bc6dfb32a1
[4] https://github.com/godotengine/godot-docs/blame/stable/tutorials/misc/binary_serialization_api.rst

SimoneGianni:
My C is rusty, but reading https://github.com/godotengine/godot/blob/master/core/io/marshalls.cpp I figured out the proper format of the binary serialization, and tested on Godot 3.3 stable official that serialization is as described in this PR.

*Bugsquad edit: This closes #4887.*

### DIFF
--- a/tutorials/io/binary_serialization_api.rst
+++ b/tutorials/io/binary_serialization_api.rst
@@ -20,12 +20,10 @@ little-endian-encoded. All packets have a 4-byte header representing an
 integer, specifying the type of data.
 
 The lowest value two bytes are used to determine the type, while the highest value
-two bytes contain flags.
+two bytes contain flags::
 
-```
-base_type = val & 0xFFFF;
-flags = val >> 16;
-```
+    base_type = val & 0xFFFF;
+    flags = val >> 16;
 
 +--------+--------------------------+
 | Type   | Value                    |
@@ -121,8 +119,8 @@ If no flags are set (flags == 0), the integer is sent as a 32 bit integer:
 | 4        | 4     | Integer   | 32-bit signed integer    |
 +----------+-------+-----------+--------------------------+
 
-If flag ENCODE_FLAG_64 is set (flags & 1 == 1), the integer is sent as 
-a 64 bit integer:
+If flag ``ENCODE_FLAG_64`` is set (``flags & 1 == 1``), the integer is sent as 
+a 64-bit integer:
 
 +----------+-------+-----------+--------------------------+
 | Offset   | Len   | Type      | Description              |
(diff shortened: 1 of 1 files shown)


tutorials/io/binary_serialization_api.rst
```diff
@@ -114,11 +133,22 @@ precision.
...
+
 +----------+-------+---------+-----------------------------------+
 | Offset   | Len   | Type    | Description                       |
 +==========+=======+=========+===================================+
-| 4        | 8     | Float   | IEEE 754 double-precision float   |
+| 4        | 4     | Float   | IEEE 754 single-precision float   |
++----------+-------+---------+-----------------------------------+
+
+If flag ENCODE_FLAG_64 is set (flags & 1 == 1), the float is sent as 
+a 64 bit double precision:
```
Calinou, line 145:
```suggestion
If flag ``ENCODE_FLAG_64`` is set (``flags & 1 == 1``), the float is sent as 
a 64-bit double precision number:
```

tutorials/io/binary_serialization_api.rst
```diff
@@ -105,6 +113,17 @@ precision.
...
+If no flags are set (flags == 0), the integer is sent as a 32 bit integer:
+
++----------+-------+-----------+--------------------------+
+| Offset   | Len   | Type      | Description              |
++==========+=======+===========+==========================+
+| 4        | 4     | Integer   | 32-bit signed integer    |
++----------+-------+-----------+--------------------------+
+
+If flag ENCODE_FLAG_64 is set (flags & 1 == 1), the integer is sent as 
+a 64 bit integer:
```
Calinou, line 125:
```suggestion
If flag ``ENCODE_FLAG_64`` is set (``flags & 1 == 1``), the integer is sent as 
a 64-bit integer:
```

tutorials/io/binary_serialization_api.rst
```diff
@@ -17,7 +17,15 @@ Packet specification
...
-integer, specifying the type of data:
+integer, specifying the type of data.
+
+The lowest value two bytes are used to determine the type, while the highest value
+two bytes contain flags.
+
+```
+base_type = val & 0xFFFF;
+flags = val >> 16;
+```
```
Calinou, line 28:
```suggestion
The lowest value two bytes are used to determine the type, while the highest value
two bytes contain flags::

    base_type = val & 0xFFFF;
    flags = val >> 16;
```

tutorials/io/binary_serialization_api.rst
```diff
@@ -114,11 +133,22 @@ precision.
...
++----------+-------+---------+-----------------------------------+
+
+If flag ENCODE_FLAG_64 is set (flags & 1 == 1), the float is sent as 
+a 64 bit double precision:
+
 +----------+-------+---------+-----------------------------------+
+| Offset   | Len   | Type    | Description                       |
++==========+=======+=========+===================================+
+| 4        | 8     | Float   | IEEE 754 double-precision float   |
++----------+-------+-----------+---------------------------------+
```
Calinou, line 151:
```suggestion
+----------+-------+---------+-----------------------------------+
```

SimoneGianni:
@Calinou thanks for the fix, first time didn't follow all conventions properly, sorry.

Calinou (approved):

PR #4895, opened 2021-05-06
Fix GDNative C example

### LINKED ISSUE #2921: GDNative C example case dependancy fault causes exception
"https://docs.godotengine.org/en/3.1/_images/nativescript_library.png" shows the class name as "SIMPLE" whereas the godot_nativescript_register_class call registers it as "Simple" as this is apparently case sensitive the example fails with an 'Attempt to call 'new' in base 'NativeScript' on a null instance. They need to be the same. Either the image should be corrected or both godot_nativescript_register_class and godot_nativescript_register_method should use "SIMPLE" instead of "Simple".

"https://docs.godotengine.org/en/3.1/tutorials/plugins/gdnative/tutorials/plugins/gdnative/img/gdnativelibrary_complete.png" is a missing image in the documentation.

Thank you for all your efforts.

### LINKED ISSUE #4891: Change "Simple" to "SIMPLE" in GDNative C example documentation
**Your Godot version:**
Godot_v3.3-stable_x11.64

**Issue description:**

The GDNative C example documentation has a section which explains to name the NativeScript "Simple", however this conflicts with the actual C source code (gdnative-demos/c/simple/simple.c) which instead uses "SIMPLE" in all caps. You will otherwise get an "Attempt to call function 'new' in base 'NativeScript' on a null instance" error when attempting to call the gdscript example code:

`
onready var data = preload("res://bin/simple.gdns").new()
`

https://github.com/godotengine/godot-docs/blob/master/tutorials/scripting/gdnative/gdnative_c_example.rst

skyace65:
"SIMPLE" is now capitalized where it needs to be, I checked the demo project repository to figure out where it should be changed. The image I replaced went through png compression. Closes #4891, closes #2921


akien-mga (approved):

akien-mga:
Thanks!

akien-mga:

PR #4902, opened 2021-05-10
Function key bindings per platform

tacevee:
The shortcuts F1, F2, F3, F4 work only on Windows.
The mac bindings Alt+1, Alt+2.. were already mentioned in the docs.
Also mention the default bindings on Linux Ctrl+F1, Ctrl+F2.

### DIFF
--- a/getting_started/introduction/first_look_at_the_editor.rst
+++ b/getting_started/introduction/first_look_at_the_editor.rst
@@ -98,13 +98,13 @@ There are four main screen buttons centered at the top of the editor:
 2D, 3D, Script, and AssetLib.
 
 You'll use the **2D screen** for all types of games. In addition to 2D games,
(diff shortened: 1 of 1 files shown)


pycbouh:
> Also mention the default bindings on Linux Ctrl+F1, Ctrl+F2.

Actually it was changed for both Windows and Linux. We don't typically have different shortcuts between those two platforms. Only MacOS gets special treatment.

tacevee:
Ah, so on Windows it is also Ctrl+F1, ... sorry, I can't test on Windows.

Ok, I will try to redo but I am struggling with github...

In a local clone I had edited a few more occurrences, eg also in tutorials/editor/default_key_mapping.rst but I cant push it - so I tried the webinterface but I managed to edit one file only.

Issue #3489, opened 2020-05-03
Inconsistent instructions regarding compiling custom export templates per platform


Xrayez:
**Your Godot version:**
3.0+

**Issue description:**
First, look at the various pages for [compiling Godot](https://docs.godotengine.org/en/latest/development/compiling/index.html) for different platforms, and notice that some pages describe the entire build cycle, while others (mainly macOS and iOS) pretty much lack the instructions on how to compile and/or use custom export templates (how to set them via editor's export presets etc.).

I think it would be nice to have a dedicated page describing the process of compiling export templates for Godot *in general* as an introductory material without relying on the platform-specific instructions. The main takeaway should be that there's a difference between (compiling) an editor and the export templates, and this should be clearly emphasized.

At the very least, each of the platform-specific instructions should likely just point to:
* [Introduction to the buildsystem - Export templates](https://docs.godotengine.org/en/latest/development/compiling/introduction_to_the_buildsystem.html#export-templates).

Some proposals and issues demonstrating the struggle of new users understanding these two processes:
* godotengine/godot#35299:
    > This step should be mentioned in Custom modules in C++ documentation, it should at least link to the export templates documentation as it's really confusing when you follow the tutorial and your custom module doesn't work when exported.
* godotengine/godot-proposals#288.
* Xrayez/godot-imagetools#2

Thoughts?

**URL to the documentation page (if already existing):**
See above, but mostly lack of info for [Compiling for macOS](https://docs.godotengine.org/en/latest/development/compiling/compiling_for_osx.html), and likely some others.

Calinou:
Ideally, Godot should do the export template packaging job for you: https://github.com/godotengine/godot/pull/36332

PR #4904, opened 2021-05-11
Compiling: Add instructions for macOS templates

### LINKED ISSUE #3194: Creating MacOS export templates documentation missing?
Hi there, I've been learning how to compile the engine, and have found that there's instructions on how to create [Windows export templates](https://docs.godotengine.org/en/latest/development/compiling/compiling_for_windows.html#creating-windows-export-templates), but not for [Compiling for MacOS](https://docs.godotengine.org/en/3.2/development/compiling/compiling_for_osx.html), I think adding this would be very helpful to folks using this platform primarily, since they might just miss other platform examples.

akien-mga:
Fixes #3194.

Note: The part about MoltenVK should be removed when cherry-picking for 3.x branches.

Note 2: I didn't add mentions of options like `production=yes` as this needs a more thorough overall of all the Compiling section (see also #3489). For now this PR solves a lack in the macOS documentation to bring it to par with other platforms.

### DIFF
--- a/development/compiling/compiling_for_osx.rst
+++ b/development/compiling/compiling_for_osx.rst
@@ -74,10 +74,15 @@ editor binary built with ``target=release_debug``::
     cp bin/godot.osx.opt.tools.universal Godot.app/Contents/MacOS/Godot
     chmod +x Godot.app/Contents/MacOS/Godot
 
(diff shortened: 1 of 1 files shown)


akien-mga:
Cherry-picked for 3.2 and 3.3.

PR #4912, opened 2021-05-14
Add Windows PowerShell example in Compiling with PCK encryption key

Calinou:
This can be cherry-picked to `3.3` without conflicts.

### DIFF
--- a/development/compiling/compiling_with_script_encryption_key.rst
+++ b/development/compiling/compiling_with_script_encryption_key.rst
@@ -46,9 +46,13 @@ Step by step
 
        export SCRIPT_AES256_ENCRYPTION_KEY="your_generated_key"
 
-    .. code-tab:: bat Windows
+    .. code-tab:: bat Windows (cmd)
 
        set SCRIPT_AES256_ENCRYPTION_KEY=your_generated_key
+
+    .. code-tab:: bat Windows (PowerShell)
+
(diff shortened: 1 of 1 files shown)


akien-mga (approved):

akien-mga:
Thanks!

pycbouh:
To my understanding this will override system or user defined environmental variable. Therefore this can be destructive.

akien-mga:
Does it override for the whole system, or just for the current shell? If it's just the current shell then it's the expected behavior, `export` on Unix and `set` on cmd do the same.

Calinou:
> To my understanding this will override system or user defined environmental variable. Therefore this can be destructive.

In cmd, `setx` must be used to set the environment variable in a persistent manner. I'm not sure what's the powershell equivalent, but I doubt it's `$env`.

pycbouh:
Yep, okay, sorry. Some wording on the usage of `$env` was confusing me, but the official docs are pretty clear that this is not a persistent change:

> When you change environment variables in PowerShell, the change affects only the current session. This behavior resembles the behavior of the `Set` command in the Windows Command Shell and the `Setenv` command in UNIX-based environments. To change values in the Machine or User scopes, you must use the methods of the **System.Environment** class.

I've experimented to make sure and it didn't seem to do anything destructive to the user or machine scope.

akien-mga:
Cherry-picked for 3.2 and 3.3.

PR #4918, opened 2021-05-16
Exporting with Android Tutorial: Debug Keystore Required Options

HaywardMorihara:
This pull request contains documentation updates that remind the developer that they _must_ provide the `Debug`, `Debug User`, and `Debug Password` fields in the Export -> Options -> Keystore menu when exporting their project for Android with debugging. If they do not provide these fields, then they will receive the error: `'apksigner' returned with error #1`.

I am making a pull request with this documentation because it was an error that I myself came across and struggled with and I see that others have as well: https://godotengine.org/qa/101943/error-apksigner-returned-with-error-%231

- [x] Read the [Contributing Guidelines](https://github.com/godotengine/godot-docs#contributing)
- [x] Followed [Image Contributing Guidelines](https://docs.godotengine.org/en/latest/community/contributing/docs_writing_guidelines.html#image-contribution-guidelines) and compressed the example image in this pull request 

One open question I have is whether or not this is the result of a bug. In the [Setting it up in Godot](https://docs.godotengine.org/en/stable/getting_started/workflow/export/exporting_for_android.html#setting-it-up-in-godot) section of the documentation, it already shows that the `Debug Keystore User` and `Debug Keystore Pass` should be set in the `Editor Settings`. I wonder if those values _should_ be propagated to the Export -> Options -> Keystore values as defaults? Currently, I observe that this is NOT the case and no defaults are provided when exporting for Android. If this is a bug, I can create an Issue on https://github.com/godotengine/godot


akien-mga:
The values in the Editor Settings should indeed be used as default, fallback values when you don't fill explicit values in your Android export preset.

The rationale is that you typically have a development specific keystore which you use to sign test exports, and this is shared among all projects. And then you can have specific release and debug keystores for use in *production* when you export your project to Google Play or other stores.

If that's not working, then it's indeed a bug.

HaywardMorihara:
That makes sense. In that case, I'll ensure that I can reproduce and file an issue on https://github.com/godotengine/godot/issues (I see https://github.com/godotengine/godot/issues/48096 but it doesn't look like the exact same issue, though might be related).

Do you think it would be valuable to add this documentation regardless? Or move it to a "troubleshooting" section or something? I don't see one that a relevant troubleshooting section exists, I suppose that's what the Q&A is for. Otherwise, I can close this PR.

HaywardMorihara:
Based on conversation in https://github.com/godotengine/godot/issues/48787, it seems that there is _not_ a bug in the fallback behavior - I was observing the `'apksigner' returned with error #1` error, I believe, because I had accidentally entered the `Debug Keystore` path in the Export options, but not provided the username & password. 

I think that https://godotengine.org/qa/101943/error-apksigner-returned-with-error-%231 should be helpful if anyone else comes across this error, so I'm going to close this pull request, since the user shouldn't necessarily fill out the Export options and likely wants to use the Editor Settings.

PR #4933, opened 2021-05-24
Fix broken links in development/cpp section

ayushk7:
Links directed to master/core were broken and all such links has been modified to the 3.x/core. Now all such links are working.

### DIFF
--- a/development/cpp/common_engine_methods_and_macros.rst
+++ b/development/cpp/common_engine_methods_and_macros.rst
@@ -237,10 +237,10 @@ Godot features many error macros to make error reporting more convenient.
 
 .. seealso::
 
-    See `core/error_macros.h <https://github.com/godotengine/godot/blob/master/core/error_macros.h>`__
+    See `core/error_macros.h <https://github.com/godotengine/godot/blob/3.x/core/error_macros.h>`__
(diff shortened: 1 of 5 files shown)


akien-mga:
Seems good, but for future PRs please make sure changes in the `3.x` branch (if they don't apply to `master`). Changes can then be cherry-picked to stable branches like `3.3`. We usually don't cherry-pick changes from stable branches to development branches, it's the other way around that development happens.

pycbouh:
And I guess for `3.3` it would make sense to set those links to direct to `3.3` and not `3.x`?

ayushk7:
> And I guess for `3.3` it would make sense to set those links to direct to `3.3` and not `3.x`?

Yeah, I will fix it.

akien-mga:
> And I guess for `3.3` it would make sense to set those links to direct to `3.3` and not `3.x`?

I'm not sure about this, these files don't move around in successive 3.x versions, and this is *development* documentation, so it's targeted towards development branches.

The 4.x code did get refactored so an update is needed for `master` and older branches need to track `3.x`, but otherwise I wouldn't make each old branch target its specific branch version here.

akien-mga:
Thanks!

PR #4944, opened 2021-05-29
Add a page on using C++ profilers with Godot

Calinou:
This is useful for troubleshooting performance issues that can't be solved with the GDScript profiler.

The profilers covered aren't the most powerful ones, but they are quick to set up and understand for occasional users.


development/cpp/index.rst
```diff
@@ -7,6 +7,7 @@ Engine development
 
    introduction_to_godot_development
    configuring_an_ide/index
+   using_cpp_profilers
```
akien-mga, line 10: The current ordering of pages in this section is not the best, but this makes it worse IMO. Too many users skip reading documentation about the engine internals, and that's arguably something that one should look for before thinking about using a C++ profiler.

Not sure what would be a better toctree, but this might be worth splitting in several subsections (tooling, engine internals, extending the engine?).

Calinou: I reorganized the table of contents entirely, let me know what you think.

akien-mga (approved):
New toctree looks great :+1:

akien-mga:
Thanks!

PR #4948, opened 2021-06-01
Give input event flow graph a background to make it visible on dark theme

DeeJayLSP:
Black text combined with dark background doesn't fit well, this change adds an outline so it becomes visible.


DeeJayLSP:
Old and new images for comparison

Old image

New image

Calinou:
I think it's better to add a white background to those images instead.

DeeJayLSP:
Image with BG

akien-mga (approved):

akien-mga:
Thanks!

PR #4950, opened 2021-06-03
Backport white background input flow event from master to 3.x

DeeJayLSP:
This applies #4948 to 3.3 branch. It uses the same optimized PNG from that pull request.


DeeJayLSP:
Just a small question: shouldn't the `Control._input_event(ev)` in the graph be changed to `Control._gui_input(ev)`?

Calinou:
> Just a small question: shouldn't `Control._input_event(ev)` in the graph be changed to `Control._gui_input(ev)`?

This graph was likely made in the Godot 2.x days (or even earlier), so it should be updated for Godot 3.0 onwards.

mhilbrunner:
Merging so we at least have the same file again as in master with the fixed background, but yeah, this should be updated to reflect current Godot.

PR #4951, opened 2021-06-03
Add a page on using C++ profilers (3.3)

Calinou:
`3.3` backport of https://github.com/godotengine/godot-docs/pull/4944.

This is useful for troubleshooting performance issues that can't be solved with the GDScript profiler.

The profilers covered aren't the most powerful ones, but they are quick to set up and understand for occasional users.

This PR includes https://github.com/godotengine/godot-docs/pull/4952.

PR #4952, opened 2021-06-03
Add a workaround for VerySleepy freezing when loading symbols

Calinou:
This is fixed in the latest nightly builds of VerySleepy, but nightly builds for VerySleepy are currently unavailable. Also, it'll likely be a while until a new stable release is made.

This is something I encountered when writing the article. Someone else also stumbled upon this bug in https://github.com/godotengine/godot-proposals/issues/2820#issuecomment-853931795, so I figure we have to document this workaround.

No need to cherry-pick this to the `3.3` branch as https://github.com/godotengine/godot-docs/pull/4951 already includes this notice.

### DIFF
--- a/development/cpp/using_cpp_profilers.rst
+++ b/development/cpp/using_cpp_profilers.rst
@@ -19,6 +19,18 @@ Recommended profilers
 These profilers may not be the most powerful or flexible options, but their
 standalone operation and limited feature set tends to make them easier to use.
 
+.. note::
+
+    In its latest stable release (0.90 as of writing), VerySleepy will freeze
+    when loading debug symbols. To solve this, exit VerySleepy then save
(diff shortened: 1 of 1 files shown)


development/cpp/using_cpp_profilers.rst
```diff
@@ -19,6 +19,18 @@ Recommended profilers
 These profilers may not be the most powerful or flexible options, but their
 standalone operation and limited feature set tends to make them easier to use.
 
+.. note::
+
+    In its latest stable release, VerySleepy will freeze when loading debug symbols.
```
akien-mga, line 24: It should likely specify which release, as this will inevitably become outdated at some point (and we might not remove this warning fast enough for this "latest" statement not to become erroneous).

akien-mga (approved):

akien-mga:
Thanks!

PR #4953, opened 2021-06-03
Change text to avoid use of first person in Static typing in GDScript

DeeJayLSP:
Maybe it's better to not use first-person in documentations


Calinou (approved):
Thanks!

PR #4957, opened 2021-06-05
Remove misplaced (?) character in Data preferences

DeeJayLSP:
While I don't know if it was misplaced an isolated "s" there is odd.

### DIFF
--- a/tutorials/best_practices/data_preferences.rst
+++ b/tutorials/best_practices/data_preferences.rst
@@ -43,7 +43,7 @@ Godot stores all variables in the scripting API in the
 `Variant <https://docs.godotengine.org/en/latest/development/cpp/variant_class.html>`_
 class. Variants can store Variant-compatible data structures such as
 :ref:`Array <class_Array>` and :ref:`Dictionary <class_Dictionary>` as well 
-as :ref:`Object <class_Object>`s.
+as :ref:`Object <class_Object>`\ s.
 
 Godot implements Array as a ``Vector<Variant>``. The engine stores the Array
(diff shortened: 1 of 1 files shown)


DeeJayLSP:
If it's supposed to be pluralization then we just need to insert it inside the "Object" that comes before

pycbouh:
Yeah, I think it was supposed to be plural with "s" on the outside.

DeeJayLSP:
From now on I'm considering the misplaced character to be the space before the "s".

tutorials/best_practices/data_preferences.rst
```diff
@@ -42,8 +42,8 @@ Array vs. Dictionary vs. Object
 Godot stores all variables in the scripting API in the
 `Variant <https://docs.godotengine.org/en/latest/development/cpp/variant_class.html>`_
 class. Variants can store Variant-compatible data structures such as
-:ref:`Array <class_Array>` and :ref:`Dictionary <class_Dictionary>` as well as
-:ref:`Object <class_Object>` s.
+:ref:`Array <class_Array>` and :ref:`Dictionary <class_Dictionary>` as well 
+as :ref:`Object <class_Object>`s.
```
mhilbrunner, line 46:
```suggestion
as :ref:`Object <class_Object>`\ s.
```

mhilbrunner:
Looks good, however Sphinx really wants that space after a class reference. Using `\ ` insteaf of a space works in this case.

mhilbrunner:
Thanks!

PR #4959, opened 2021-06-06
Remove duplicate text on Creating iOS plugins

DeeJayLSP:
Godot iOS Plugins is said on the link, no need to say it twice.


Calinou (approved):
Thanks!

PR #4961, opened 2021-06-06
Add third method to "Add the script to the node".

DeeJayLSP:
While it's a complicated way, it would be good to list it too.


mhilbrunner:
This section is documenting named types, not the various ways to attach scripts to a node. Elaborating further doesn't add much to this section.

Still, thanks for the contribution!

PR #4962, opened 2021-06-07
Update class reference link on Scene organization

DeeJayLSP:
doc_autoloads_versus_internal_nodes points to Autoloads versus regular nodes, so name it correctly.


Calinou (approved):
Thanks!

PR #4964, opened 2021-06-07
Remove double space on Exporting for Web

DeeJayLSP:


akien-mga:
Thanks. When doing this kind of formatting fixes I prefer to be thorough and fix all of them, so I made #4965 including this fix.

PR #4965, opened 2021-06-07
Cleanup uses of double spaces between words or after punctuation

akien-mga:


Calinou:

PR #4968, opened 2021-06-07
Remove pleonasm in Localization using gettext

DeeJayLSP:
I think "removing strings that are no longer present in the PO template removed in the PO template." could just be simplified to "removing strings that are no longer present in the PO template".

### DIFF
--- a/tutorials/i18n/localization_using_gettext.rst
+++ b/tutorials/i18n/localization_using_gettext.rst
@@ -156,8 +156,8 @@ Updating message files to follow the PO template
 
 After updating the PO template, you will have to update message files so
 that they contain new strings, while removing strings that are no longer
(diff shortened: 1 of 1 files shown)


tutorials/i18n/localization_using_gettext.rst
```diff
@@ -156,8 +156,8 @@ Updating message files to follow the PO template
 
 After updating the PO template, you will have to update message files so
 that they contain new strings, while removing strings that are no longer
-present in the PO template removed in the PO template. This can be done
-automatically using the ``msgmerge`` tool:
+present in the PO template. This can be doneautomatically using the
```
Calinou, line 159:
```suggestion
present in the PO template. This can be done automatically using the
```

Calinou (approved):
Thanks!

PR #4970, opened 2021-06-08
Design a title screen: change text to avoid use of first person

DeeJayLSP:
I was unable to find this in master, so I'm requesting this in 3.3.


Calinou (approved):
Thanks!

DeeJayLSP:
@Calinou Just a question, is "we" considered as a first person that should be changed or is it OK, since it's plural form?

Calinou:
> @Calinou Just a question, is "we" considered as a first person that should be changed or is it OK, since it's plural form?

"We" is acceptable when used sparingly, but "I" should definitely be avoided at all costs.

PR #4972, opened 2021-06-08
Rework saving_games.rst

ShatReal:
This one follows the "Serialization" demo in the godot demos repo. I hope this is easier follow than the current one. Note that I don't have C# code.


mhilbrunner:
Hey! Thanks for contributing.

Unfortunately, as is I don't see this as much of an improvement over the current one:

1. The current tutorial immediately starts with the content it focuses on, namely: saving/loading information. It breaks this down into steps: Identify what needs to be saved, serialize and save, and load. This new tutorial spends the first half setting up a game scene to save and a UI. The current tutorial would be faster for me to get what I need from if I want to implement saving. *Maybe* this could be improved by just referencing the demo directly, or maybe it should just save a very simple node as an example.
2. It sets up a lot of irrelevant stuff. Like a healtbar. Or a "open user folder" button, which isn't *that* related either.
3. The architecture in the original is better, or at least this one's isn't - the original has a group for things to save, and calls a save method on each thing to serialize. This one has an `enemy` group that gets saved arbitrarily, and also saves the player. I feel the current one encourages good practices more.
4. This one loses some things, like C#. Could be fixed later on, and we could still merge if this was a clear improvement otherwise - but I don't think it is.

The current tutorial clearly could be improved. More time could be spent on loading, serialization could be expanded (like in https://github.com/godotengine/godot-docs/pull/4977), and I'm sure there's more.

I'd still like to thank you for the work that went into this. Don't feel discouraged to continue working on this, although I'd suggest maybe considering improving the current tutorial instead. Also, feel free to talk to the docs team on Rocketchat.

I'm closing this PR for now, but attaching a *salvageable* label as there are some things worth picking for the current tutorial.

Also, I'd be grateful if you could share which things exactly you stumbled upon or found unclear in the current tutorial.

Thanks again!

mhilbrunner (commented):
Hey! Thanks for contributing.

Originally, I closed the PR with a similar comment, but for now I'l leave it open to see if anyone has differing opinions.

1. This new tutorial spends far more time setting up a game, scenes, scripts, a UI and things like healthbars, whereas the current tutorial quickly focuses on the actual *saving*.
2. I prefer the architecture of the original tutorial. Having a way to mark things to save (like the `persist` group) seems cleaner than arbitrarily saving everything in the `enemies` group, and the player. Much clearer how to extend the first one cleanly.
3. This is losing some things, like C#. Not that bad if it would otherwise be an clear improvement.

mhilbrunner:
This now has conflicts after merging other PRs, notably https://github.com/godotengine/godot-docs/pull/4977.

Those could be fixed, but as outlined above I'm not sure this is a clear improvement either way.

The current tutorial *does* have some problems that should be fixed, namely and mostly that JSON should be discouraged and CustomResources and ConfigFiles be recommended and explained instead/in addition to JSON.

However, I think we should rather do those improvements on the current state, rather than reworking the tutorial to include this longer setup, especially as we move away from long form tutorials on the docs.

So I'm going to close this now, parts of this may remain useful as reference. Thanks for your work and contributing, and I hope you understand the outlined reasons for not merging.

PR #4973, opened 2021-06-09
Removed "core/pool_vector.h" reference

BenaceurA:
I couldn't find any reference to it in master


akien-mga:
It doesn't make sense to remove this reference without actually rewriting the relevant section which refers to it. You're right that `PoolVector` was removed and replaced by `Vector<T>` or `PackedTVector` in 4.0, but the whole section needs to be rewritten for it.

PR #4976, opened 2021-06-09
Update the binary serialization API doc to accurately describe how Ob…

levilindsey:
…jects are encoded.

This update is in response to discussion at https://github.com/godotengine/godot/issues/49430.

*Bugsquad edit: Fixes https://github.com/godotengine/godot/issues/49430 by documentation*

### DIFF
--- a/tutorials/io/binary_serialization_api.rst
+++ b/tutorials/io/binary_serialization_api.rst
@@ -12,6 +12,16 @@ in the functions ``get_var`` and ``store_var`` of :ref:`class_File`
 as well as the packet APIs for :ref:`class_PacketPeer`. This format
 is *not* used for binary scenes and resources.
 
+Full Objects vs Object instance IDs
+-----------------------------------
+
+If a variable is serialized with ``full_objects = true``, then any Objects
+contained in the variable will be serialized and included in the result. This
+is recursive.
+
+If ``full_objects = false``, then only the instance IDs will be serialized for
+any Objects contained in the variable.
+
 Packet specification
 --------------------
 
@@ -97,10 +107,10 @@ should be ``(offset - 4) * 2 + 4``. The "float" type itself always uses double
 precision.
 
 0: null
-~~~~~~~
+^^^^^^^
 
 1: :ref:`bool<class_bool>`
-~~~~~~~~~~~~~~~~~~~~~~~~~~
+^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 +----------+-------+-----------+---------------------------+
 | Offset   | Len   | Type      | Description               |
@@ -109,7 +119,7 @@ precision.
 +----------+-------+-----------+---------------------------+
 
 2: :ref:`int<class_int>`
-~~~~~~~~~~~~~~~~~~~~~~~~
+^^^^^^^^^^^^^^^^^^^^^^^^
 
 If no flags are set (flags == 0), the integer is sent as a 32 bit integer:
 
@@ -129,7 +139,7 @@ a 64-bit integer:
 +----------+-------+-----------+--------------------------+
 
 3: :ref:`float<class_float>`
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 If no flags are set (flags == 0), the float is sent as a 32 bit single precision:
 
@@ -149,7 +159,7 @@ a 64-bit double precision number:
 +----------+-------+---------+-----------------------------------+
 
 4: :ref:`String<class_string>`
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 +----------+-------+-----------+----------------------------+
 | Offset   | Len   | Type      | Description                |
@@ -162,7 +172,7 @@ a 64-bit double precision number:
 This field is padded to 4 bytes.
 
 5: :ref:`Vector2<class_vector2>`
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 +----------+-------+---------+----------------+
 | Offset   | Len   | Type    | Description    |
@@ -173,7 +183,7 @@ This field is padded to 4 bytes.
 +----------+-------+---------+----------------+
 
 6: :ref:`Rect2<class_rect2>`
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 +----------+-------+---------+----------------+
 | Offset   | Len   | Type    | Description    |
@@ -188,7 +198,7 @@ This field is padded to 4 bytes.
 +----------+-------+---------+----------------+
 
 7: :ref:`Vector3<class_vector3>`
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 +----------+-------+---------+----------------+
 | Offset   | Len   | Type    | Description    |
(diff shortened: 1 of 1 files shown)


tutorials/io/binary_serialization_api.rst
```diff
@@ -385,8 +385,29 @@ Every name string is padded to 4 bytes.
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++----------+-------+-----------+---------------------------------------------------------------------+
+| Offset   | Len   | Type      | Description                                                         |
++==========+=======+===========+=====================================================================+
+| 4        | 4     | Integer   | val&0x7FFFFFFF = properties, val&0x80000000 = shared (bool)         |
++----------+-------+-----------+---------------------------------------------------------------------+
+
+Then what follows is, for amount of "properties", pairs of name and value,
+one after the other, using this same format.
```
kleonc, line 398: Every case should be described appropriately according to [the source code](https://github.com/godotengine/godot/blob/master/core/io/marshalls.cpp#L1178-L1241). Like what is stored depending on `p_allow_objects` flag and whether object is valid or not.

tutorials/io/binary_serialization_api.rst
```diff
@@ -385,8 +385,29 @@ Every name string is padded to 4 bytes.
...
+
+.. Note::
+
+   Only properties that are configured as storable will be included when
+   serializing an object. By default, script variables are configured as
+   :ref:`PROPERTY_USAGE_SCRIPT_VARIABLE<class_@GlobalScope_constant_PROPERTY_USAGE_SCRIPT_VARIABLE>`,
+   which is **not** included when serializing an object.
```
kleonc, line 405: > only properties that are configured as storable will be included

It's not clear what "configured as storable" means.

> By default, script variables are configured as [constant PROPERTY_USAGE_SCRIPT_VARIABLE], which is [b]not[/b] included.


Note that for example exported property like export var something has more flags set, including PROPERTY_USAGE_STORAGE and thus it would be stored.

I'm not sure what flags are set for what properties but for sure info added in here needs to be correct.

tutorials/io/binary_serialization_api.rst
```diff
@@ -385,8 +385,29 @@ Every name string is padded to 4 bytes.
...
 
-17: :ref:`Object<class_object>` (unsupported)
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+17: :ref:`Object<class_object>`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++----------+-------+-----------+---------------------------------------------------------------------+
+| Offset   | Len   | Type      | Description                                                         |
++==========+=======+===========+=====================================================================+
+| 4        | 4     | Integer   | val&0x7FFFFFFF = properties, val&0x80000000 = shared (bool)         |
```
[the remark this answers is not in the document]
levilindsey: Oh, wow, good point. Not sure what I was looking at when I wrote that, but I definitely wasn't looking at the write part of the source code.

I've replaced this entirely. I'm pretty sure it's accurate now.

tutorials/io/binary_serialization_api.rst
levilindsey: Thanks, good point! Done.

tutorials/io/binary_serialization_api.rst
levilindsey: I've re-worded this to hopefully be a little clearer.

And I re-worded it to say "By default, non-exported script variables are configured as PROPERTY_USAGE_SCRIPT_VARIABLE", which I think is correct, but I could instead re-word this as something more general, like "For example, script variables are often configured as PROPERTY_USAGE_SCRIPT_VARIABLE and are not included."

tutorials/io/binary_serialization_api.rst
```diff
@@ -383,13 +393,69 @@ For each Name and Sub-Name
...
+| 4        | 8     | Integer    | The Object instance ID (64-bit signed integer)  |
++----------+-------+------------+-------------------------------------------------+
+
+``full_objects`` enabled
+""""""""""""""""""""""""
+
++----------+-------+------------+-------------------------------------------------+
+| Offset   | Len   | Type       | Description                                     |
++==========+=======+============+=================================================+
+| 4        | X     | Bytes      | Class name (UTF-8 encoded string)               |
```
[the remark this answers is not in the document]
levilindsey: Good point! I tried writing something that I think makes sense for this.

tutorials/io/binary_serialization_api.rst
```diff
@@ -383,13 +393,69 @@ For each Name and Sub-Name
...
++----------+-------+------------+-------------------------------------------------+
+| X+4      | 4     | Integer    | The number of properties that are serialized    |
++----------+-------+------------+-------------------------------------------------+
+
+For each property:
+
++----------+-------+------------+-------------------------------------------------+
+| Offset   | Len   | Type       | Description                                     |
++==========+=======+============+=================================================+
+| Y        | Z     | Bytes      | Property name (UTF-8 encoded string)            |
```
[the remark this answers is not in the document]
levilindsey: Done.

tutorials/io/binary_serialization_api.rst
levilindsey: I added the same latest fix here as in the other PR.

tutorials/io/binary_serialization_api.rst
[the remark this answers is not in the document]
kleonc: Actually now that I've looked at it I think my suggestion was wrong: position/type 4 listed in this exact file refers to how String is stored as a separate Variant. So there length is being at offset 4 because before it goes (type+flags). Here type and flags aren't stored for this string, it's just length + actual UTF-8 encoded string. So length doesn't have additional 4 bytes offset here and pointing to the type 4 in the list might be misleading. Take a look at how strings are listed in `PackedStringArray` (type 25), here it should be listed similarly. Sorry for that.
So probably something like this:
```
``full_objects`` enabled
""""""""""""""""""""""""

+----------+-------+----------------+----------------------------------------------------------+
| Offset   | Len   | Type           | Description                                              |
+==========+=======+================+==========================================================+
| 4        | 4     | Integer        | Class name (String length)                               |
+----------+-------+----------------+----------------------------------------------------------+
| 8        | X     | Bytes          | Class name (UTF-8 encoded string)                        |
+----------+-------+----------------+----------------------------------------------------------+
| X+8      | 4     | Integer        | The number of properties that are serialized             |
+----------+-------+----------------+----------------------------------------------------------+

For each property:

+----------+-------+----------------+----------------------------------------------------------+
| Offset   | Len   | Type           | Description                                              |
+==========+=======+================+==========================================================+
| Y        | 4     | Integer        | Property name (String length)                            |
+----------+-------+----------------+----------------------------------------------------------+
| Y+4      | Z     | Bytes          | Property name (UTF-8 encoded string)                     |
+----------+-------+----------------+----------------------------------------------------------+
| Y+4+Z    | W     | <variable>     | Property value, using this same format                   |
+----------+-------+----------------+----------------------------------------------------------+
```
Other than this (and [this](https://github.com/godotengine/godot/pull/49470#discussion_r650119429)) great job!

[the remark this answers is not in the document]
levilindsey: Yeah, that makes sense. Ok, done!

tutorials/io/binary_serialization_api.rst
```diff
@@ -383,13 +393,70 @@ For each Name and Sub-Name
...
+""""""""""""""""""""""""
+
++----------+-------+----------------+----------------------------------------------------------+
+| Offset   | Len   | Type           | Description                                              |
++==========+=======+================+==========================================================+
+| 4        | 4     | Integer        | Class name (String length)                               |
++----------+-------+----------------+----------------------------------------------------------+
+| 8        | X     | Bytes          | Class name (UTF-8 encoded string)                        |
++----------+-------+----------------+----------------------------------------------------------+
+| X+12     | 4     | Integer        | The number of properties that are serialized             |
```
kleonc, line 432: Offset should be X+8.

levilindsey: Shouldn't it include the sum of the offsets before it?

kleonc: Offset is a sum of the preceding lengths (here possibly padded to 4 bytes). Or equivalently it's (previous offset) + (previous length).

levilindsey: Oh. Of course. Not sure what I was thinking! Thanks, done.

kleonc: Looks fine now.

mhilbrunner:
Looks good, thank you!

Sorry it took a bit to get to this PR due to the backlog. :)

PR #4977, opened 2021-06-09
Update the saving games doc to describe tradeoffs of JSON vs binary s…

levilindsey:
…jects are encoded.

While updating some docs in response to https://github.com/godotengine/godot/issues/49430, I noticed that this file doesn't mention binary serialization at all, which seems like a significant omission. Maybe I should have created a separate issue to track adding this though.

### DIFF
--- a/tutorials/io/saving_games.rst
+++ b/tutorials/io/saving_games.rst
@@ -310,3 +310,52 @@ Load parent objects first so they are available for the :ref:`add_child()
 call when child objects are loaded. You will also need a way to link
 children to parents as the :ref:`NodePath
 <class_nodepath>` will likely be invalid.
+
+JSON vs binary serialization
+----------------------------
+
+For simple game state, JSON may be the best choice, because it is easy to
+understand, and generates human-readable files that are easy to debug.
+
+But JSON has many limitations. If you need to store more complex game state or
+a lot of game state, :ref:`binary serialization<doc_binary_serialization_api>`
+may be a better approach.
+
+JSON limitations
+~~~~~~~~~~~~~~~~
+
+Here are some important gotchas to know about when using JSON.
+
+Filesize:
+  JSON stores data in text format, which is much larger than binary format.
+
+Data types:
+  JSON only knows about a limited set of data types. If you have data types
+  that JSON doesn't know about, you will need to translate your data to and
+  from types that JSON can handle. For example, some important types that JSON
+  can't parse are: Vector2, Vector3, Color, Rect2, Quat.
+
+Custom logic needed for encoding/decoding:
+  If you have any custom classes that you want to store with JSON, you will
+  need to write your own logic for encoding and decoding those classes.
+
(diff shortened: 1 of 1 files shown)


Calinou:
The saving games tutorial is being reworked here: https://github.com/godotengine/godot-docs/pull/4972

levilindsey:
Oh, nice! It looks like the changes in #4972 are largely independent of my PR here, so this is probably still relevant, and this can wait until that's in.

tutorials/io/saving_games.rst
```diff
@@ -310,3 +310,52 @@ Load parent objects first so they are available for the :ref:`add_child()
...
+
+* Binary serialization should produce smaller files than JSON.
+* Binary serialization can handle most common data types.
+* Binary serialization requires less custom logic for encoding and decoding
+  custom classes.
+
+Note that only properties that are configured as storable will be included
+when serializing an object. By default, script variables are configured as
+:ref:`PROPERTY_USAGE_SCRIPT_VARIABLE<class_@GlobalScope_constant_PROPERTY_USAGE_SCRIPT_VARIABLE>`,
+which is **not** included when serializing an object. You can specify the usage
```
kleonc, line 357: > only properties that are configured as storable will be included

It's not clear what "configured as storable" means.

> By default, script variables are configured as [constant PROPERTY_USAGE_SCRIPT_VARIABLE], which is [b]not[/b] included.

Note that for example exported property like export var something has more flags set, including PROPERTY_USAGE_STORAGE and thus it would be stored.

I'm not sure what flags are set for what properties but for sure info added in here needs to be correct.

levilindsey: I've re-worded this to hopefully be a little clearer.

And I re-worded it to say "By default, non-exported script variables are configured as PROPERTY_USAGE_SCRIPT_VARIABLE", which I think is correct, but I could instead re-word this as something more general, like "For example, script variables are often configured as PROPERTY_USAGE_SCRIPT_VARIABLE and are not included."

kleonc: [Same](https://github.com/godotengine/godot/pull/49470#discussion_r649443225).

kleonc: I think this one seems fine now. :)

levilindsey: Hooray! I assume this should still wait on #4972? Is there anything else to do on my end until then?

kleonc: Absolutely no idea. You can visit [Godot Contributors Chat](https://chat.godotengine.org/home), ask a question like this (probably in [new-contributors channel](https://chat.godotengine.org/channel/new-contributors) or [documentation channel](https://chat.godotengine.org/channel/documentation)) and there's a chance someone will make things clear for you.

mhilbrunner:
Congrats on contributing to the docs and sorry it took a while to get merged :)

Thank you!

PR #4978, opened 2021-06-10
Added a video demonstrating Windows Compile

bmetcalf21:
A video demonstrating the compile in Windows 10.


Calinou:
I appreciate the effort, but I don't think we should refer to a video in the official documentation. The video can end up being outdated quickly.

bmetcalf21:
Thanks for the reply, that makes sense.

PR #4988, opened 2021-06-12
Fixed formatting for "Contributing changes" section

Vlad-Zumer:
Previous format looked wrong and was a hard to understand.

### DIFF
--- a/community/contributing/contributing_to_the_documentation.rst
+++ b/community/contributing/contributing_to_the_documentation.rst
@@ -49,9 +49,9 @@ contribute, you should also read:
 Contributing changes
 --------------------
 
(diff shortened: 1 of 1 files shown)


Vlad-Zumer:
This is what it currently looks like.

image

pycbouh:
I don't think we need a warning block here. I think the idea was to just make the text bold. Similarly we highlight text in the next paragraph (`**Note that the pages in ``classes/`` should not be edited here.**`). The problem here is ambiguity of the RST syntax where it considers `**` at the start of the line as something other than bold tag. This is what needs to be fixed somehow.

pycbouh:
Superseded by https://github.com/godotengine/godot-docs/pull/5073. Thanks for the contribution anyway!

PR #4994, opened 2021-06-13
Getting source: Add --depth 1 arguments

DeeJayLSP:
The whole commit history isn't needed to compile the engine, which is the category of this docs page, so mention the way to avoid cloning all of it.

Also I've changed the second ``--branch`` argument to ``-b`` so the line wouldn't get out of bounds.

### DIFF
--- a/development/compiling/getting_source.rst
+++ b/development/compiling/getting_source.rst
@@ -43,7 +43,7 @@ With ``git``, you can also clone a stable release by specifying its branch or ta
 after the ``--branch`` (or just ``-b``) argument::
 
     # Clone the continuously maintained stable branch (`3.x` as of writing).
-    git clone https://github.com/godotengine/godot.git --branch 3.x
+    git clone https://github.com/godotengine/godot.git -b 3.x
 
     # Clone the `3.2.3-stable` tag. This is a fixed revision that will never change.
     git clone https://github.com/godotengine/godot.git -b 3.2.3-stable
--- a/development/compiling/getting_source.rst
+++ b/development/compiling/getting_source.rst
@@ -30,21 +30,21 @@ If you are using the ``git`` command line client, this is done by entering
 the following in a terminal:
 
 ::
-
-    git clone https://github.com/godotengine/godot.git
+    #Use --depth 1 argument to avoid cloning the whole unneeded commit history.
+    git clone https://github.com/godotengine/godot.git --depth 1
 
 For any stable release, visit the `release page <https://github.com/godotengine/godot/releases>`__
 and click on the link for the release you want.
 You can then download and extract the source from the download link on the page.
 
 With ``git``, you can also clone a stable release by specifying its branch or tag
-after the ``--branch`` argument::
+after the ``--branch`` (or just ``-b``) argument::
 
     # Clone the continuously maintained stable branch (`3.x` as of writing).
-    git clone https://github.com/godotengine/godot.git --branch 3.x
+    git clone https://github.com/godotengine/godot.git --branch 3.x --depth 1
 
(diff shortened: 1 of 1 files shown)


Calinou:
The issue with `--depth=1` is that it prevents specific operations from working (switching branches, checking out an older commit, bisecting) unless you deepen the local history. It needs to be carefully documented.

DeeJayLSP:
Pull request workflow and bisecting regressions are part of the [contributing](https://docs.godotengine.org/en/latest/community/contributing/index.html) section of the docs, while the one in the commit is part of [compiling](https://docs.godotengine.org/en/latest/development/compiling/index.html).

I just don't think frequent ``git`` users would do the mistake of using ``--depth=1`` when willing to contribute to the engine. Better make an advice not to use it when willing so.

development/compiling/getting_source.rst
```diff
@@ -31,20 +31,22 @@ the following in a terminal:
...
 and click on the link for the release you want.
 You can then download and extract the source from the download link on the page.
 
 With ``git``, you can also clone a stable release by specifying its branch or tag
-after the ``--branch`` argument::
+after the ``--branch`` (or just ``-b``) argument::
 
     # Clone the continuously maintained stable branch (`3.x` as of writing).
-    git clone https://github.com/godotengine/godot.git --branch 3.x
+    git clone https://github.com/godotengine/godot.git --branch 3.x --depth 1
```
mhilbrunner, line 46:
```suggestion
    git clone https://github.com/godotengine/godot.git --branch 3.x
```

development/compiling/getting_source.rst
```diff
@@ -31,20 +31,22 @@ the following in a terminal:
...
 
     # Clone the `3.2.3-stable` tag. This is a fixed revision that will never change.
-    git clone https://github.com/godotengine/godot.git --branch 3.2.3-stable
+    git clone https://github.com/godotengine/godot.git -b 3.2.3-stable --depth 1
```
mhilbrunner, line 49:
```suggestion
    git clone https://github.com/godotengine/godot.git -b 3.2.3-stable
```

development/compiling/getting_source.rst
```diff
@@ -31,20 +31,22 @@ the following in a terminal:
 
 ::
 
-    git clone https://github.com/godotengine/godot.git
+    # Use --depth 1 argument to avoid cloning the whole unneeded commit history.
+    # Don't use --depth 1 if you're cloning for engine development purposes.
+    git clone https://github.com/godotengine/godot.git --depth 1
```
mhilbrunner, line 36:
```suggestion
    git clone https://github.com/godotengine/godot.git
    # You can add the --depth 1 argument to omit the commit history.
    # Faster, but not all Git operations (like blame) will work.
```

development/compiling/getting_source.rst
```diff
@@ -32,19 +32,21 @@ the following in a terminal:
...
 For any stable release, visit the `release page <https://github.com/godotengine/godot/releases>`__
 and click on the link for the release you want.
 You can then download and extract the source from the download link on the page.
 
 With ``git``, you can also clone a stable release by specifying its branch or tag
-after the ``--branch`` argument::
+after the ``--branch`` (or just ``-b``) argument::
 
     # Clone the continuously maintained stable branch (`3.x` as of writing).
     git clone https://github.com/godotengine/godot.git --branch 3.x
```
mhilbrunner, line 46:
```suggestion
    git clone https://github.com/godotengine/godot.git -b 3.x
```

mhilbrunner:
I agree that we should avoid making `--depth 1` the copy-paste default.
The hint that it exists is useful however, so lets add that. :)

Thanks!

Calinou:

PR #4996, opened 2021-06-14
C++ profilers: fix SCons options list

DeeJayLSP:
Adds a newline in order to show correctly (check rich diff for comparison)


Calinou (approved):
Thanks!

Issue #4981, opened 2021-06-10
iOS Plugins: inline code text so large it gets out of bounds


DeeJayLSP:
**Your Godot version:**
3.3.2, but I don't think it's applicable to this issue.

**Issue description:**
There are some inline codes in Creating iOS plugins page that get out of bounds like the following:
![image](https://user-images.githubusercontent.com/60024671/121551445-08aabb80-c9e6-11eb-987e-a033cf677977.png)

**URL to the documentation page:**
[Creating iOS plugins](https://docs.godotengine.org/en/stable/tutorials/platform/ios/ios_plugin.html)

PR #5006, opened 2021-06-17
Fix inline code going out of bounds

hpnrep6:
Fixes godotengine/godot-docs#4981. Puts the inline code block in a table so that it stays within the page in [Creating iOS plugins](https://docs.godotengine.org/en/stable/tutorials/platform/ios/ios_plugin.html).

## Before
Screenshot from 2021-06-17 08-39-19

## After
Screenshot from 2021-06-17 08-39-51

### DIFF
--- a/tutorials/platform/ios/ios_plugin.rst
+++ b/tutorials/platform/ios/ios_plugin.rst
@@ -56,19 +56,19 @@ To build an iOS plugin:
 
     - You should use the same header files for iOS plugins and for the iOS export template.
 
-3. In the ``Build Settings`` tab, specify the compilation flags for your static library in ``OTHER_CFLAGS``. The most important ones are ``-fcxx-modules``, ``-fmodules``, and ``-DDEBUG`` if you need debug support. Other flags should be the same you use to compile Godot. For instance:
+3. In the ``Build Settings`` tab, specify the compilation flags for your static library in ``OTHER_CFLAGS``. The most important ones are ``-fcxx-modules``, ``-fmodules``, and ``-DDEBUG`` if you need debug support. Other flags should be the same you use to compile Godot. For instance: 
 
-+--------------------------------------------------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


pycbouh:
Why a table? If this to be a solution, we have a way to define code blocks.

hpnrep6:
> Why a table? If this to be a solution, we have a way to define code blocks.

Wasn't sure if the original format of the line was to be kept. Yes, a code block will also work.

hpnrep6:
> Why a table? If this to be a solution, we have a way to define code blocks.

With code blocks, it looks like:
Screenshot from 2021-06-17 09-37-34
or
Screenshot from 2021-06-17 09-39-13

In my opinion, it might cause some confusion by being inconsistent with the other flags listed in the section. (but it may be a solution if we're staying consistent with the other documentation?)

pycbouh:
Well using a table there looks weird and is a straight up hack, so I wouldn't go that direction either way.

An ideal look would be to wrap the line on whitespace but not on punctuation (so that `[DeviceLibrary].a` doesn't break up, for example). But I'm not sure this is achievable as it would be mostly up to the browser if we allow any sort of wrapping.

mhilbrunner:
Agree a table is not the way to go for this. Code blocks may not be optimal for this, but IMO still better than a table.

hpnrep6:
I have made it so that it uses a code block instead.

**After**
Screenshot from 2021-07-07 23-39-49

mhilbrunner:
Merged, and thank you!

pycbouh:
From your screenshot the list formatting is now a bit broken. Points 3 and 5 have different indent than 4.

hpnrep6:
Sorry, I think I included the wrong screenshot (that was for when there wasn't an empty line between :: and "For instance:". This is what it should look like.

Screenshot from 2021-07-08 11-24-10

pycbouh:
Ah, great! Thank you.

Calinou:
This can't be cherry-picked trivially to the `3.3` branch. Please open a separate PR against the `3.3` branch :slightly_smiling_face:

PR #5008, opened 2021-06-17
Fix C# code label in Multimesh example

hpnrep6:
Adds the C# code label to the example in [Optimization using MultiMeshes](https://docs.godotengine.org/en/stable/tutorials/optimization/using_multimesh.html#multimesh-example).

#### Before
Screenshot from 2021-06-17 14-28-36

#### After
Screenshot from 2021-06-17 14-32-04


pycbouh (approved):

Calinou (approved):
Thanks!

PR #5018, opened 2021-06-19
Replace C# if statement with switch statement to match GDScript

hpnrep6:
The GDScript example uses a match statement with variable **p_what**, so the C# example should ideally also follow the same logic as the GDScript example and use a switch for int **what** instead of an if statement in [Data preferences](https://docs.godotengine.org/en/latest/tutorials/best_practices/data_preferences.html).

A similar situation can be seen in [Godot notifications](https://docs.godotengine.org/en/latest/tutorials/best_practices/godot_notifications.html#process-vs-physics-process-vs-input), where an switch statement is used instead of an if statement when there's only one case. To stay consistent, this example should also use a switch statement.

## Before
Screenshot from 2021-06-19 12-47-02

## After
Screenshot from 2021-06-19 12-46-51

### DIFF
--- a/tutorials/best_practices/data_preferences.rst
+++ b/tutorials/best_practices/data_preferences.rst
@@ -257,7 +257,7 @@ tree structures.
 
         public override void Notification(int what)
         {
-            switch(what)
+            switch (what)
             {
                 case NotificationPredelete:
                     foreach (object child in _children)
--- a/tutorials/best_practices/data_preferences.rst
+++ b/tutorials/best_practices/data_preferences.rst
@@ -257,14 +257,18 @@ tree structures.
 
         public override void Notification(int what)
         {
-            if (what == NotificationPredelete)
+            switch(what)
             {
-                foreach (object child in _children)
-                {
-                    TreeNode node = child as TreeNode;
-                    if (node != null)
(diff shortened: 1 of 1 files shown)


tutorials/best_practices/data_preferences.rst
```diff
@@ -257,14 +257,18 @@ tree structures.
 
         public override void Notification(int what)
         {
-            if (what == NotificationPredelete)
+            switch(what)
```
hpnrep6, line 260:
```suggestion
            switch (what)
```

Calinou:
Is the `default:` case required to be defined?

hpnrep6:
No, it is not required to be defined in this case. I was sticking with the format of the example in [Godot notifications](https://docs.godotengine.org/en/latest/tutorials/best_practices/godot_notifications.html#process-vs-physics-process-vs-input).

One related thing is that it will throw a compilation error if the default: case is removed and _break;_ is removed from the `NotificationPredelete` case, like in 
Screenshot from 2021-06-19 14-30-03
but that's more of a C# thing so it probably isn't that relevant.

Calinou (approved):
Thanks!

PR #5027, opened 2021-06-21
Add common use case scenarios in Multiple resolutions

Calinou:
This is a very common question on community platforms, so adding some example use cases should hopefully help people.

The recommendations favor giving players a good experience over being the easiest to implement for developers. Nonetheless, everything is explained on the page, so it shouldn't be too difficult still. It is possible to take shortcuts, but I don't recommend them :slightly_smiling_face: 

Since this is fairly lengthy, should this be split into a different page?

### DIFF
--- a/tutorials/rendering/multiple_resolutions.rst
+++ b/tutorials/rendering/multiple_resolutions.rst
@@ -263,6 +263,117 @@ To configure stretching at runtime from a script, use the
 ``get_tree().set_screen_stretch()`` method (see
 :ref:`SceneTree.set_screen_stretch() <class_SceneTree_method_set_screen_stretch>`).
 
+Common use case scenarios
+-------------------------
+
+The following settings are recommended to support multiple resolutions and aspect
+ratios well.
+
+Desktop game
+^^^^^^^^^^^^
+
+**Non-pixel art:**
+
+- Set the base window width to ``1920`` and window height to ``1080``. If you have a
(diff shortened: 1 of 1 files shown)


mhilbrunner:
Built locally, seems fine, but *if* the page gets any longer we may indeed want to think about splitting it in the future.

Good work, thank you!

Calinou:

golddotasksquestions:
Thank you, Calinou!

Lielay9:
This is like a year late, but I'll say it anyway. The recommendations for mobile platforms are fine but I'd consider either revising the recommended aspect ratios or adding a separate footnote for supporting tablets. Most still use a ratio of 4:3 and the android developer guide goes as far as recommending testing on [1:1](https://developer.android.com/guide/topics/large-screens/support-different-screen-sizes) devices. 

The recommended default aspect ratio, 16:9, usually works out when the stretch aspect is set to `expand` since tablets have bigger screens. However, on smaller screens (such as iPad Mini; 4:3; 8.3 inches) the "extra" vertical room can mess with the UI elements. For reference going from 16:9 to 4:3, the elements become a third "narrower". Considering this might sometimes be desired, I think adding an extra footnote to remind to also test the program on narrower screens might be the way to go.

Calinou:
@Lielay9 Done in https://github.com/godotengine/godot-docs/pull/6352.

PR #5029, opened 2021-06-21
Add missing C# example semicolon in Using AnimationTree

hpnrep6:
Adds a missing semicolon in the C# code example of [Using AnimationTree](https://docs.godotengine.org/en/latest/tutorials/animation/animation_tree.html#state-machine-travel).

## Before
Screenshot from 2021-06-21 10-40-43

## After
Screenshot from 2021-06-21 10-42-42


Calinou (approved):
Thanks!

PR #5042, opened 2021-06-23
Update OpenXR plugin information

skyace65:
Updates information on the OpenXR plugin, which just had its first release.


Calinou (approved):
Thanks!

Issue #5025, opened 2021-06-21
The latest documentation side menu cannot scroll down on the phone device


Chaosus:
**Your Godot version:**
Don't matter

**Issue description:**
When I open up the latest documentation on my Android phone (XIAOMI MI MAX 1) and open a side menu I cannot scroll down it with my fingers to see the full content of that menu, it stuck:

<details>
<summary>Screenshot from phone</summary>

![image](https://user-images.githubusercontent.com/3036176/122709843-c4ca7880-d267-11eb-9d26-10e0e8ae2f80.png)

</details>

Stable documentation and the PC version of this page works correctly on my phone.

**URL to the documentation page:**
https://docs.godotengine.org/en/latest/

Calinou:
I can confirm this on Chromium 90 on an OnePlus 6. This only happens on the `master` branch, not `stable` or `3.3`. This is likely a regression from https://github.com/godotengine/godot-docs/pull/4884.

cc @leroy

dweremeichik:
Can also confirm in Firefox 81.1.1 on Android.

hpnrep6:
Issue appears to be caused by `.wy-side-scroll { overflow: inherit; }` in algolia.css, since it's no longer present when that line is removed.

PR #5049, opened 2021-06-25
Fix scrolling the navigation bar on mobile

hpnrep6:
Prevents the `overflow-y: scroll;` property from being overridden in the sidebar on mobile. Fixes godotengine/godot-docs#5025

Screenshot from 2021-06-25 18-29-02


Calinou (approved):
Thanks!

PR #5055, opened 2021-06-29
Exporting for Web: Update info about WebGL

akien-mga:
To update when:
- Safari 15 gets released.
- iOS browsers get the support turned on by default
- Once we have a WebGL renderer again in 4.x :)


clayjohn (approved):

akien-mga:

PR #5058, opened 2021-06-29
Update the estimated release date for Godot 4.0

Calinou:
Given how the first alpha isn't even released yet, it is very unlikely that a stable 4.0 release is out in 2021.


akien-mga:
That's a realistic estimate, yet that's not what we *aim for*, so I don't think the change is needed at this point in time. Let's assess further into 2021 if we're really too far from something that looks like a stable release.

mhilbrunner:
Agreed, closing for now for sake of the backlog. Let's see in a few months :)

PR #4594, opened 2021-01-27
Fix Sphinx `add_lexer()` deprecation in the GDScript extension

Calinou:
This closes #4593.

PR #5067, opened 2021-07-06
Fix Sphinx warning about GDScriptLexer

mhilbrunner:
Just fixes a Sphinx warning, because adding lexers was changed to require passing a Class, not an Instance of that Class.

### DIFF
--- a/_extensions/gdscript.py
+++ b/_extensions/gdscript.py
@@ -341,7 +341,7 @@ class GDScriptLexer(RegexLexer):
 
 
 def setup(sphinx):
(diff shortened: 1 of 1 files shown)


pycbouh:
This was already done and then reverted due to issues, see https://github.com/godotengine/godot-docs/pull/4594 and https://github.com/godotengine/godot-docs/pull/4626

akien-mga:
We upgraded Sphinx in the meantime so it should be fine now.

mhilbrunner:
Yeah, tested it locally and on RTD and it seems fine. :)

akien-mga:

PR #5072, opened 2021-07-08
Fix hyperlink formatting in Building the manual

hpnrep6:
**Before**
Screenshot from 2021-07-07 23-42-32

**After**
Screenshot from 2021-07-07 23-44-47


mhilbrunner:
Thanks for the fix!

Calinou:
Cherry-picked to the `3.3` and `stable` branches by backporting the Building the manual with Sphinx page.

PR #5078, opened 2021-07-11
Add GDNative C++ code examples to Dodge the Creeps aka Your first 2D game

aaronfranke:
These code examples are based off of this Dodge the Creeps with GDNative C++ demo I recently made: https://github.com/godotengine/gdnative-demos/tree/master/cpp/dodge_the_creeps

Having a guided example of a working game using GDNative C++ should make it significantly easier for people to start learning how to use GDNative C++. Previously there was very little documentation on GDNative and everything was a struggle.

Two changes I made that affect GDScript and C# versions of the docs:

* In section 5, the Main game scene, I separated the class/member declaration from the `_ready()` declaration with `randomize`. This helps make this part consistent with the other steps and should improve clarity as now it explains why we use `randomize`.

* In `Mob`, I removed the `min_speed` and `max_speed` in favor of hard-coding values in `Main`. These were pretty useless before, because `Mob`s aren't instanced in the editor, and these values are only used once, so it was needlessly complicated. It was also inconsistent with the rotation/direction, which uses hard-coded values for the range of `(-PI / 4, PI / 4)`.

TODO in this PR:

- [x] I don't know if `.. tab::` will be recognized by the docs properly.
- [x] Add a minimal project that users can download which contains the art, headers, SConstruct, GDNative stuff, etc.

TODO in future PRs:

* Add GDNative C++ code examples to Your first 3D game.

* Add GDNative C++ code examples to other parts of the documentation.

* Add GDNative C++ code examples to the snippets in the class reference.


mhilbrunner:
Is there a reason to not simply use `cpp` instead of `gdnativecpp` as the lexer for highlighting? That should fix the build error.

NathanLovato:
> In Mob, I removed the min_speed and max_speed in favor of hard-coding values in Main. These were pretty useless before, because Mobs aren't instanced in the editor, and these values are only used once, so it was needlessly complicated.

Naming two values isn't complicated for students. It gives the values a name, that's part of the point. Here I'd say whatever, it's a small detail and I don't recall anyone not understanding it. You can change it or leave it, doesn't make a difference.

aaronfranke:
@mhilbrunner There is some possible confusion with in-engine modules which are also written in C++, and many users do use this. We can change it to `cpp` if that's desired though.

mhilbrunner:
@aaronfranke Well I guess `cpp GDNative C++` could work instead, but to me personally `cpp` is enough.

aaronfranke:
@mhilbrunner Ok, I changed it to `cpp`. I also added a part to the first page with `.. tab:: GDNative C++` with a link to download a starter project. I think if we write "GDNative C++" on the first page and only "C++" elsewhere it will be fine and not confusing. Let me know if this looks good to you, I haven't attempted to build this locally so I don't know if it works, but at least CI is happy now.

getting_started/first_2d_game/03.coding_the_player.rst
```diff
@@ -156,6 +201,24 @@ which returns ``true`` if it's pressed or ``false`` if it isn't.
         }
     }
 
+ .. code-tab:: cpp
+
+    // This code goes in `player.cpp`.
+    void Player::_process(const double p_delta) {
+        godot::Input *input = godot::Input::get_singleton();
```
Duroxxigar, line 208: Why are you getting this every frame?

mhilbrunner:
The tabs work fine, by the way. Not a fan of re-iterating the requirements for C#/C++, we should have those elsewhere and link there or it becomes harder to maintain when those requirements change. Pages look fine at a first glance, will do a more thorough review now.

aaronfranke:
@mhilbrunner I'm not sure that the C++ version for GDNative is directly related to the engine's C++ version. As far as I know GDNative is a C API so the C++ version shouldn't need to match. I don't know what the point of linking that page would be.

mhilbrunner:
@akien-mga Do we want to add C++ code examples like we do with GDScript and C#? I'm in favor, covering at least the basic getting started tutorials with all 3 languages is a solid starting point for all three.

It is additional maintenance work, however I think the C++ crowd tends to be the one most likely able to help themselves, not too bad IMO. And having C++ documented makes it more official and is valuable, also a good first impression for more advanced users.

Again, it *is* more work on the docs though :) I'm in favor, what do you think, Akien? And maybe @cbscribe  @NathanLovato

akien-mga:
I think having C++ snippets for Dodge the Creeps can be useful, as it helps understand how to get started with GDNative C++.

I'm not sure I would go much beyond that though. Having C++ snippets everywhere would likely be a huge amount of work, it's unclear how it would be impacted by the new extensions system in 4.x, and there's currently a lot of boilerplate to use C++ which can't be replicated everywhere.

IMO if you want to use GDNative C++, you can do the Dodge the Creeps with the C++ snippets, and then you should be able to translate GDScript to C++ fairly easily yourself for more advanced tutorials.

mhilbrunner:
I agree, sounds good to me.

mhilbrunner:
Then this has cooked enough - merged. We'll add it to 4.0 for now, and if there's anymore, we can change it - also, if the opinion changes, but I agree with Akien that having this at least for the 2 initial tutorials is useful.

Thanks for doing the work!

aaronfranke:
Adding to the 3.4 branch of the docs would be nice too.

akien-mga:
IIRC the tutorial has been rewritten for the `master` branch, so it might need a manual backport with significant changes for `3.4`. But yes, that would be nice to have.

aaronfranke:
@akien-mga Why not backport the entire rewrite to 3.4? All the code snippets etc are still for Godot 3.x.

PR #5083, opened 2021-07-12
Note which primitives can currently be batched

lawnjelly:
This is only relevant to 3.x branch, 3.3 onward.

This is my first attempt at a PR to the docs like this so may go horribly wrong...

I'd also ideally like to put a link into:
https://docs.godotengine.org/en/3.3/tutorials/2d/custom_drawing_in_2d.html

But I don't know how to.

Fixes https://github.com/godotengine/godot/issues/50387

### DIFF
--- a/tutorials/optimization/batching.rst
+++ b/tutorials/optimization/batching.rst
@@ -514,11 +514,11 @@ I use a large number of textures, so few items are being batched.
 Appendix
 ~~~~~~~~
 
-Batched Primitives
+Batched primitives
 ^^^^^^^^^^^^^^^^^^
 
-Note that not all primitives can be batched, and batching is not guaranteed,
-especially with primitives using an anti-aliased border. The following
+Not all primitives can be batched. Batching is not guaranteed either,
+especially with primitives using an antialiased border. The following
 primitive types are currently available:
 
 - RECT
@@ -526,8 +526,9 @@ primitive types are currently available:
 - POLY
 - LINE
 
-With non-batched primitives you may be able to get better performance drawing
-them manually with polys in a ``_draw()`` function.
+With non-batched primitives, you may be able to get better performance by
+drawing them manually with polys in a ``_draw()`` function.
+See :ref:`doc_custom_drawing_in_2d` for more information.
 
 .. _doc_batching_light_scissoring_threshold_calculation:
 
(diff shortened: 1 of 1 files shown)


tutorials/optimization/batching.rst
```diff
@@ -514,6 +514,21 @@ I use a large number of textures, so few items are being batched.
...
+especially with primitives using an anti-aliased border. The following
+primitive types are currently available:
+
+- RECT
+- NINEPATCH (depending on wrapping mode)
+- POLY
+- LINE
+
+With non-batched primitives you may be able to get better performance drawing
+them manually with polys in a ``_draw()`` function.
```
Calinou, line 530:
```suggestion
Batched primitives
^^^^^^^^^^^^^^^^^^

Not all primitives can be batched. Batching is not guaranteed either,
especially with primitives using an antialiased border. The following
primitive types are currently available:

- RECT
- NINEPATCH (depending on wrapping mode)
- POLY
- LINE

With non-batched primitives, you may be able to get better performance by
drawing them manually with polys in a ``_draw()`` function.
See :ref:`doc_custom_drawing_in_2d` for more information.
```

lawnjelly:
Thanks for changes @Calinou ! :smile:

akien-mga:
> I'd also ideally like to put a link into:
docs.godotengine.org/en/3.3/tutorials/2d/custom_drawing_in_2d.html
>
> But I don't know how to.

See https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#ref-role

Here you'd put a label before your section, as you can see done for the next section. We use `_doc_<name>_<section>` syntax by convention:

```
.. _doc_batching_batched_primitives

Batched primitives
^^^^^^^^^^^^^^^^^^
```

Then in the page where you want to add a link, you reference it with e.g.:
```
See :ref:`doc_batching_batched_primitives` for details on which primitives benefit from batching.
```

Note that the label takes a leading underscore, but then using it in a `:ref:` doesn't take it (don't ask me why :)).

The link title will be inferred from the section, so it will produce:

> See [Batched primitives](https://docs.godotengine.org/en/3.3/tutorials/optimization/batching.html#batched-primitives) for details on which primitives benefit from batching.

You can also customize the link title if need be with:
```
See :ref:`the batching reference <doc_batching_batched_primitives>` for details on which primitives benefit from batching.
```

mhilbrunner:
Looks good - thank you :)

Issue #5007, opened 2021-06-17
Add configuration examples for macOS/Windows to the "Configuring an IDE" section


pycbouh:
**Your Godot version:**
Any.

**Issue description:**
Specifically, this applies to VS Code. As an example we use a Linux-compatible configuration. And while for the most part the idea for how to configure the editor is given, there are some things that should be set up differently on other operating systems, which often trips people.

We can use code tabs to show examples for all three major OSes. For VS Code that would be samples of `tasks.json` and `launch.json`. Maybe there are similar issues with other editors, but that's the one I see questions about the most.

Here's an adapted example of VS Code's `launch.json` for Windows that I use

```
    {
      "name": "Launch",
      "type": "cppvsdbg",
      "request": "launch",
      "program": "${workspaceFolder}/bin/godot.windows.tools.64.exe",
      // Change the arguments below for the project you want to test with.
      // To run the project instead of editing it, remove the "--editor" argument.
      "args": [ "-e", "--path", "path-to-your-godot-project-folder" ],
      "stopAtEntry": false,
      "cwd": "${workspaceFolder}",
      "environment": [],
      "externalConsole": false,
      "visualizerFile": "${workspaceFolder}/platform/windows/godot.natvis"
    },
```

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/stable/development/cpp/configuring_an_ide/visual_studio_code.html

PR #5087, opened 2021-07-13
Add Windows configuration example for VS Code

pycbouh:
See https://github.com/godotengine/godot-docs/issues/5007. Unfortunately, I don't have a MacOS device so I don't know what the configuration for that can be.

This can be cherrypicked for 3.x branches, but mentions of "linuxbsd" need to be adjusted.

### DIFF
--- a/development/cpp/configuring_an_ide/visual_studio_code.rst
+++ b/development/cpp/configuring_an_ide/visual_studio_code.rst
@@ -30,20 +30,35 @@ Importing the project
 
 - Within the ``tasks.json`` file find the ``"tasks"`` array and add a new section to it:
 
-.. code-block:: js
+.. tabs::
+  .. code-tab:: js LinuxBSD
 
-  {
-    "label": "build",
-    "type": "shell",
-    "command": "scons",
-    "group": "build",
-    "args": [
-      "platform=linuxbsd", // Change to your current platform
-      "target=debug",
-      "-j4"
-    ],
-    "problemMatcher": "$msCompile"
-  }
+    {
+      "label": "build",
+      "group": "build",
+      "type": "shell",
+      "command": "scons",
+      "args": [
+        "-j $(nproc)"
+      ],
+      "problemMatcher": "$msCompile"
+    }
+
+  .. code-tab:: js Windows
+
+    {
+      "label": "build",
+      "group": "build",
+      "type": "shell",
+      "command": "scons",
+      "args": [
+        // Use this when your default shell is Command Prompt (cmd.exe).
+        "-j %NUMBER_OF_PROCESSORS%",
+        // Use this when your default shell is PowerShell.
+        "-j $env:NUMBER_OF_PROCESSORS"
+      ],
+      "problemMatcher": "$msCompile"
+    }
 
 .. figure:: img/vscode_3_tasks.json.png
    :figclass: figure-w480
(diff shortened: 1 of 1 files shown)


development/cpp/configuring_an_ide/visual_studio_code.rst
```diff
@@ -30,20 +30,36 @@ Importing the project
...
+.. tabs::
+  .. code-tab:: js LinuxBSD
+
+    {
+      "label": "build",
+      "group": "build",
+      "type": "shell",
+      "command": "scons",
+      "args": [
+        "platform=linuxbsd",
```
Calinou, line 42: The platform argument can be removed since Godot's SCons setup will detect your target platform automatically:

```suggestion
```

development/cpp/configuring_an_ide/visual_studio_code.rst
```diff
@@ -30,20 +30,36 @@ Importing the project
...
+
+  .. code-tab:: js Windows
+
+    {
+      "label": "build",
+      "group": "build",
+      "type": "shell",
+      "command": "scons",
+      "args": [
+        "platform=windows",
```
Calinou, line 57: The platform argument can be removed since Godot's SCons setup will detect your target platform automatically:

```suggestion
```

development/cpp/configuring_an_ide/visual_studio_code.rst
```diff
@@ -30,20 +30,36 @@ Importing the project
...
+        "target=debug",
+        "-j4" // Adjust the number of threads to use according to your CPU core count.
```
Calinou, line 59: We could use `-j %NUMBER_OF_PROCESSORS%` here, but I don't know if environment variables are expanded.

development/cpp/configuring_an_ide/visual_studio_code.rst
```diff
@@ -30,20 +30,36 @@ Importing the project
...
+
+    {
+      "label": "build",
+      "group": "build",
+      "type": "shell",
+      "command": "scons",
+      "args": [
+        "platform=linuxbsd",
+        "target=debug",
+        "-j4" // Adjust the number of threads to use according to your CPU core count.
```
Calinou, line 44: We could use `-j $(nproc)` here, but I don't know if environment variables are expanded.

development/cpp/configuring_an_ide/visual_studio_code.rst
pycbouh: I guess there is no need for tabs here then, as that's the only difference.

development/cpp/configuring_an_ide/visual_studio_code.rst
pycbouh: They should be, but I'm not sure if it's a good idea to max it out in the example. But I don't have a strong opinion either.

development/cpp/configuring_an_ide/visual_studio_code.rst
pycbouh: Ah, well, with your other suggestion we still need the tabs 🙂

development/cpp/configuring_an_ide/visual_studio_code.rst
pycbouh: Actually it only works if you run this through the cmd.exe. I build with the PowerShell and this errs. I think it should be this for PS:
```
scons platform=windows -j $env:NUMBER_OF_PROCESSORS
```

pycbouh:
Adjusted as suggested, also removed the target argument as redundant.

mhilbrunner:
Thank you. :)

akien-mga:

PR #5088, opened 2021-07-13
Add a `.gitattributes` recommendation to the AssetLib guidelines

pycbouh:
See https://github.com/godotengine/godot-proposals/issues/2981 and comments to https://github.com/godotengine/godot-docs/pull/3240. This question is raised pretty often by plugin developers, so it's well worth mentioning.

### DIFF
--- a/community/asset_library/submitting_to_assetlib.rst
+++ b/community/asset_library/submitting_to_assetlib.rst
@@ -75,6 +75,31 @@ library a better place for all users.
 * If your asset is a library for working with other files,
   consider including example files in the asset.
 
+* Consider adding a **.gitattributes** file to your repo and listing
+  files not required by your asset to function with the ``export-ignore``
+  directive. This directive removes such files from the resulting ZIP file
+  and prevents them from being downloaded by the asset library users.
+  For a typical plugin **.gitattributes** may look like this:
+
+  .. code-block:: none
+
+    /.gitattributes     export-ignore
+    /.gitignore         export-ignore
+    /LICENSE            export-ignore
+    /LICENSE.md         export-ignore
+    /README.md          export-ignore
+    /project.godot      export-ignore
+    /icon.png           export-ignore
+    /icon.svg           export-ignore
+
+  Other types of assets may require a different configuration (e.g.
+  a project template requires **project.godot**).
+
+* If you are submitting a plugin, add a copy of your license and readme
+  to the plugin folder itself. This is the folder that users are guaranteed to
+  keep with their project, so a copy ensures they always have those files handy
+  (and helps them fulfill your licensing terms).
+
 * The icon should be a square, its aspect ratio should be 1:1. It should
   also ideally have a minimum resolution of 64x64 pixels.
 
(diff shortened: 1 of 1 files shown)


community/asset_library/submitting_to_assetlib.rst
```diff
@@ -75,6 +75,28 @@ library a better place for all users.
...
+  and prevents them from being downloaded by the asset library users.
+  For a typical plugin **.gitattributes** may look like this:
+
+  .. code-block:: none
+
+    /.gitattributes     export-ignore
+    /.gitignore         export-ignore
+    /LICENSE            export-ignore
+    /README.md          export-ignore
+    /project.godot      export-ignore
```
Calinou, line 90: If we are excluding `project.godot`, I would also exclude the icon files which are generally `icon.png` or `icon.svg` (it's fine to have nonexistent paths in `.gitattributes`).

I would also exclude the license file if it has a `.md` extension:

```suggestion
    /.gitattributes     export-ignore
    /.gitignore         export-ignore
    /LICENSE            export-ignore
    /LICENSE.md         export-ignore
    /README.md          export-ignore
    /project.godot      export-ignore
    /icon.png           export-ignore
    /icon.svg           export-ignore
```

akien-mga:
I don't think it's a good idea to encourage exporting without license. It means that people downloading a zip of the repo will not get the license terms, while the license should explicitly always go together with the content it governs.

The real problem is the bad workflow that we impose where people should put the whole `addons/<name>/` in their repo, and thus have top-level files that might conflict with the installation destination.

I've been saying this is a terrible idea from the start, and I think time has proven it right repeatedly.

I think we should fix that workflow instead of working it around like this. A proper workflow would be:
- Plugin repo contains plugin files as top-level.
- Godot installs the plugin automatically in `addons/<name>/<plugin files>`

The name could be configured in `plugin.cfg`.

Calinou:
> I think we should fix that workflow instead of working it around like this. A proper workflow would be:
>
> - Plugin repo contains plugin files as top-level.
> -Godot installs the plugin automatically in addons/<name>/<plugin files>
>
> The name could be configured in `plugin.cfg`.

The issue is that a single repository may want to install multiple plugins, especially since we don't have dependency tracking yet. Also, non-plugin assets still need to have some kind of configuration file to define the folder they should install in.

community/asset_library/submitting_to_assetlib.rst
pycbouh: Seems redundant if you don't actually have those files, but can be added I guess.

pycbouh:
> I don't think it's a good idea to encourage exporting without license.

I don't think my additions encourage that? The second point I've added specifically covers that, though. We can work on its phrasing if you don't find it satisfactory.

> * If you are submitting a plugin, add a copy of your license and readme
  to the plugin folder itself. This is the folder that users are guaranteed to
  keep with their project, so a copy ensures they always have those files handy
  (and helps them fulfill your licensing terms).

> I've been saying this is a terrible idea from the start, and I think time has proven it right repeatedly.

You get no arguing from me here, but implementing a better plugin deployment system is a lot of work that we don't have a contributor to do. And this is what people can actually do now to help with the clutter. And what we, plugin developers, actually have to do and encourage others to do.

akien-mga (approved):

mhilbrunner:
Thank you!

akien-mga:
Cherry-picked for 3.3.

PR #5093, opened 2021-07-15
Fixed wording for when _enter_tree and _ready is called.

Duroxxigar:
Previous description could be interpreted as though the `_enter_tree` callback is not called when instantiating a scene and adding it to the scene tree. `_ready` is also not called at all, until it gets added to the scene tree, but the wording can be read as though it does get called.


mhilbrunner:
Thank you!

mhilbrunner:
Cherry-picked into `3.3` and `stable`.

PR #5095, opened 2021-07-16
Fix GLSL hyperlink formatting in List of features

hpnrep6:
https://docs.godotengine.org/en/latest/about/list_of_features.html

## Before
Screenshot from 2021-07-15 22-30-01

## After
Screenshot from 2021-07-15 22-30-13


Calinou (approved):
Thanks!

Calinou:

PR #5100, opened 2021-07-18
Descriptions of additional parameters in ``JSON.print()``

WentBerzerk:
I've added descriptions for the ``indent`` and ``sort_keys`` parameters.
I would appreciate any corrections in case the descriptions of their usage are wrong or if I've used poor wording.


Calinou:

PR #5102, opened 2021-07-19
Add notes about Blender backface culling when exporting to glTF

Calinou:
`master` version of https://github.com/godotengine/godot-docs/pull/5103.

This closes https://github.com/godotengine/godot/issues/50552.

### DIFF
--- a/tutorials/3d/standard_material_3d.rst
+++ b/tutorials/3d/standard_material_3d.rst
@@ -223,6 +223,15 @@ Determines which side of the object is not drawn when backfaces are rendered:
 * **Front:** The front of the object is culled when not visible.
 * **Disabled:** Used for objects that are double-sided (no culling is performed).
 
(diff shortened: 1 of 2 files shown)


timothyqiu:
> By default, Blender will export materials with the cull mode set to **Disabled**

This sounds like "it renders OK in Blender, but Blender has a bad default export option" to me. But in fact it's an option of material which also affects how Blender renders the material.

Calinou:
> This sounds like "it renders OK in Blender, but Blender has a bad default export option" to me. But in fact it's an option of material which also affects how Blender renders the material.

I updated the notes, can you take a look again?

timothyqiu:
I think it's good 🎉

skyace65:
Looks good.

mhilbrunner:
Yep looks good, good work!

PR #5113, opened 2021-07-26
Scene tree: Show functions as code

DeeJayLSP:
These function callbacks should be shown as code for better readability.


mhilbrunner:
Thank you!

PR #5124, opened 2021-07-29
Prevent the "ref" badge from appearing in the sidebar

merumelu:
Before it looked like this:

ref

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -313,6 +313,11 @@ a[href*="classes/"]::before {
     display: none;
 }
 
+/* Prevent the "ref" badge from appearing in the sidebar. */
+.wy-menu-vertical a[href*="classes/"]::before {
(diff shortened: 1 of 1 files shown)


mhilbrunner:
This doesn't seem to appear for me. Which OS/Browser/resolution are you using for that screenshot? On which page?

pycbouh:
@mhilbrunner Check the class reference in the latest branch

image

mhilbrunner:
Thanks, could reproduce it on latest if I *don't* expand the class ref section, funnily enough. Okay. :)

mhilbrunner:
Checkec locally, looks good - thanks for fixing this :)

merumelu:
@mhilbrunner It only happens on pages that are not inside the "classes" folder themself because Sphinx generates relative links.

PR #5125, opened 2021-07-29
Corrected layout instructions for final GUI step

cro-w:
In the final step, original says to use "Center Top" on the GUI node by mistake. "Center Top" causes the GUI to be pinned in the top/center but not scale horizontally with the window. I confirmed this by downloading the finished example project, where the GUI node is properly marked "Top Wide" to produce the intended horizontal sliding of the GUI elements


Calinou:
@cro-w It looks like you meant to only include [`e736c65` (#5125)](https://github.com/godotengine/godot-docs/pull/5125/commits/e736c654792161dba6187cc49ee505bdaba2bd23). You can open a new request to do this :slightly_smiling_face:

PR #5136, opened 2021-07-30
Add extra info from SurfaceTool class page

Dathide:
Added a few sentences about generate_normals() and generate_tangents() taken from the SurfaceTool class page in the docs.


mhilbrunner:
Thank you!

PR #5137, opened 2021-07-31
Replace various links with `ref` roles

merumelu:
Update some links to other documentation pages to use `ref` instead of hardcoded links and do some other small link cleanup that was on the same pages (Point to stable branch instead of 3.2; use a link instead of slapping a URL in the running text).


Calinou (approved):
Thanks!

PR #5138, opened 2021-07-31
GDScript style guide: tint borders of good/bad code examples

### LINKED ISSUE #3962: Tint the background of code blocks red for examples of bad code
**Issue description:**
redish background would help convey information about what not to do for bad code, instead of greenish background,
especially when followed by good code again

https://docs.godotengine.org/en/stable/getting_started/scripting/gdscript/gdscript_styleguide.html#type-hints
there are several spots in doc, of that kind

merumelu:
Add [`class`](https://docutils.sourceforge.io/docs/ref/rst/directives.html#class) directives (aliased by Sphinx as `rst-class`) before code blocks. Thy allow adding arbitrary css classes to the element afer them.

For now it just tints the existing code block border green or red, there might be a better styling.
(Colors are taken from GitHubs diff view in dark mode and are the same for light and dark style since I think they work for both.)

Closes #3962

Screenshot

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -85,6 +85,9 @@
     --kbd-shadow-color: #b0b7bf;
     --kbd-text-color: #444d56;
 
+    --code-example-good-color: #3fb950;
+    --code-example-bad-color: #f85149;
+
     --btn-neutral-background-color: #f3f6f6;
     --btn-neutral-hover-background-color: #e5ebeb;
     --footer-color: #808080;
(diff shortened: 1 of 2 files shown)


pycbouh:
I would probably go for a thick left border instead of a enclosing border, but otherwise nice that they've added a way for it.

merumelu:
> I would probably go for a thick left border instead of a enclosing border, but otherwise nice that they've added a way for it.

:+1: Agreed, it looks better with a thick left border.
Screenshot

pycbouh (approved):

Calinou (approved):
Thanks!

PR #5068, opened 2021-07-06
Remove update block from sphinx-tabs extension

mhilbrunner:
Now fine thanks to Calinou's Algolia work, yay (see https://github.com/godotengine/godot-docs/pull/4884)

Also see https://github.com/godotengine/godot-docs/pull/4995

PR #5140, opened 2021-08-01
Update Sphinx to 4.0.2 to match RTD

pycbouh:
Currently https://github.com/readthedocs/readthedocs.org/blob/master/requirements/pip.txt and https://github.com/readthedocs/readthedocs.org/blob/master/requirements/docs.txt point to Sphinx 4 (4.0.2 to be precise). I've tested it locally and it builds fine even with 4.1.2, so I think it should be safe to update the requirement.

Also updated obsolete comments in the file.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,12 +2,11 @@
 
 # Sync with readthedocs:
 # https://github.com/readthedocs/readthedocs.org/blob/master/requirements/pip.txt
-# https://github.com/readthedocs/readthedocs.org/blob/master/requirements/local-docs-build.txt
-sphinx==3.5.2
+# https://github.com/readthedocs/readthedocs.org/blob/master/requirements/docs.txt
+sphinx==4.0.2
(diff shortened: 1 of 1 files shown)


Calinou (approved):
Thanks!

Calinou:
This doesn't cherry-pick to `3.4` cleanly, and I'm not sure what to do exactly because of the sphinx-tabs regression we had recently.

pycbouh:
If I understand correctly, we need to roll out Algolia on every branch and then we can update tabs and the rest?

Calinou:
> If I understand correctly (from #5068 and #4995), we need to roll out Algolia on every branch and then we can update tabs and the rest?

Maybe, but I'm not sure if Algolia is really an upgrade over what we had previously. What do you think about the new search?

pycbouh:
> What do you think about the new search?

I didn't actually have a good chance to use it, and at this moment it doesn't work for me at all:

image

PR #5142, opened 2021-08-02
Remove comments from gdscript exports page and rephrase exporting a node description

skyace65:
EDIT: updated this PR. Removed the use of comments in the entire page. Changed "is" to "as" in a few examples. Renamed the "examples" section to basic usage. rephrased the export a node section, this messed me up when working on a project because I stopped reading after ` @export var node_path: NodePath` and thought that the node itself was being exported. Yes I should have kept reading, but the statement ` Another node in the scene can be exported as a NodePath.` is not true.

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_exports.rst
+++ b/tutorials/scripting/gdscript/gdscript_exports.rst
@@ -13,7 +13,9 @@ Exporting is done by using the ``@export`` annotation::
 
     extends Button
 
-    @export var number = 5 # Value will be saved and visible in the property editor.
+    @export var number = 5
+    
+In that example the value `5` will be saved and visible in the property editor.
 
 An exported variable must be initialized to a constant expression or have a type specifier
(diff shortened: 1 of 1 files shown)


tutorials/scripting/gdscript/gdscript_exports.rst
```diff
@@ -117,7 +117,7 @@ Examples
 
     # Nodes
 
-    # Another node in the scene can be exported as a NodePath.
+    # Another nodes path in the scene can be exported as a NodePath.
```
mhilbrunner, line 120:
```suggestion
    # Another node's path in the scene can be exported as a NodePath.
```

pycbouh:
I think this may be phrased like that to answer the question "How do you export another node", so I'm not sure this change for technical correctness is purely beneficial.

skyace65:
I've rephrased the exporting a node section again and removed comments from the entire page.

PR #5148, opened 2021-08-04
Add link when internal rooms are first mentioned

timothyqiu:
It's quite a distance from the first mention of internal room to the actual place it's described.


akien-mga:
Thanks!

PR #5150, opened 2021-08-04
[3.4] Optimizing for size: add classes that are disabled by `disable_3d`/`disable_advanced_gui`

merumelu:
Adds a list of classes that are disabled when the explained build options are used.
Generated by parsing `scene/register_scene_types.cpp` and `register_types.cpp` of
all modules with this script:
<details><summary>script</summary>

```python
#!/usr/bin/env python3

import glob
import os
import re
import sys

pattern = re.compile(r'ClassDB::register_(?:virtual_)?class<([A-Z_][\w]+)>\(\);')

def find_classes(filename, build_flag, define):
    excluded_classes = []
    ifdef_count = 0 # Keep track of potential inner #if/#ifdef/#ifndef blocks.

    with open(filename, "r") as f:
        lines = f.readlines()

    for line in lines:
        line = line.split("//", 1)[0]
        line = line.strip()
        if ifdef_count > 0:
            if line.startswith("#endif"):
                ifdef_count -= 1
            elif line.startswith("#if"):
                ifdef_count += 1
            else:
                match = pattern.search(line)
                if match is not None:
                    excluded_classes.append(match.group(1))
        else:
            if line.startswith(f"#ifndef {define}"):
                ifdef_count = 1

    return excluded_classes

def print_excluded_classes(godot_path, docs_path, build_flag, define):
    excluded_classes = []
    warnings = []

    excluded_classes += find_classes(os.path.join(godot_path, "scene", "register_scene_types.cpp"), build_flag, define)
    for module in glob.glob(os.path.join(godot_path, "modules", "**", "register_types.cpp")):
        excluded_classes += find_classes(module, build_flag, define)

    print(f"\nThe following classes are not available when building with ``{build_flag}=yes``:\n")
    for i in sorted(excluded_classes):
        if not os.path.exists(os.path.join(docs_path, "classes", f"class_{i.lower()}.rst")):
            warnings.append(f"WARNING: Added class `{i}` but no rst file `classes/class_{i.lower()}.rst` exists. Sphinx will complain about this.")
        print(f"- :ref:`class_{i}`")

    return warnings

if len(sys.argv) != 3:
    print(f"Usage: {sys.argv[0]} path/to/godot path/to/godot-docs")
    sys.exit(1)

warnings = []
warnings += print_excluded_classes(sys.argv[1], sys.argv[2], "disable_3d", "_3D_DISABLED")
warnings += print_excluded_classes(sys.argv[1], sys.argv[2], "disable_advanced_gui", "ADVANCED_GUI_DISABLED")
if len(warnings) > 0:
    print("\n".join(errors), file=sys.stderr)
```
</details>

Also moves the mention of the online build options generator to the section where
`custom.py` is introduced and do some general reformatting and rewording.

---

Making a PR like this for the `master` branch doesn't make sense at the moment since the generated class reference is out of sync.

### DIFF
--- a/development/compiling/optimizing_for_size.rst
+++ b/development/compiling/optimizing_for_size.rst
@@ -3,7 +3,7 @@
 Optimizing a build for size
 ===========================
 
-.. highlight:: shell
+.. highlight:: none
 
 Rationale
 ---------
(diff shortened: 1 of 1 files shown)


development/compiling/optimizing_for_size.rst
```diff
@@ -3,49 +3,195 @@
...
 a smaller binary. Before continuing, it is recommended to read the previous tutorials
 on compiling Godot for each platform.
 
-.. seealso::
-
-    You can use the online
-    `Godot build options generator <https://godot-build-options-generator.github.io/>`__
-    to generate a ``custom.py`` file containing SCons options.
-    You can then save this file and place it at the root of your Godot source directory.
-
```
akien-mga, line 26: It could be worded differently, but I think it's good to link to this resource early on, and not after users have already copied tons of options without understanding them from this page, which often end up outdated as this page isn't updated as often as the upstream code.

development/compiling/optimizing_for_size.rst
```diff
@@ -63,16 +209,30 @@ a lot of them:
...
+    scons p=<platform> target=release tools=no \
+    module_bmp_enabled=no module_bullet_enabled=no module_camera_enabled=no module_csg_enabled=no \
+    module_dds_enabled=no module_enet_enabled=no module_etc_enabled=no module_gdnative_enabled=no \
+    module_gridmap_enabled=no module_hdr_enabled=no module_jsonrpc_enabled=no module_mbedtls_enabled=no \
+    module_mobile_vr_enabled=no module_opensimplex_enabled=no module_opus_enabled=no module_pvr_enabled=no \
+    module_recast_enabled=no module_regex_enabled=no module_squish_enabled=no module_svg_enabled=no \
+    module_tga_enabled=no module_theora_enabled=no module_tinyexr_enabled=no module_upnp_enabled=no \
+    module_vhacd_enabled=no module_vorbis_enabled=no module_webm_enabled=no module_webrtc_enabled=no \
+    module_websocket_enabled=no module_webxr_enabled=no module_xatlas_unwrap_enabled=no
```
akien-mga, line 220: I don't really like that part of the docs. Many users just copy it without trying to understand what it does, and end up disabling things that they actually need (and then report bugs about it not working).

Overall I think this page should be refactored / rewritten to give more high level explanations of what can be done to reduce binary size, and then redirect users to https://godot-build-options-generator.github.io/ to actually get explanations on what they're enabling or disabling.

development/compiling/optimizing_for_size.rst
```diff
@@ -3,49 +3,195 @@
...
+
+The following classes are not available when building with ``disable_3d=yes``:
+
+- :ref:`class_ARVRAnchor`
+- :ref:`class_ARVRCamera`
+- :ref:`class_ARVRController`
+- :ref:`class_ARVROrigin`
+- :ref:`class_AnimatedSprite3D`
+- :ref:`class_Area`
+- :ref:`class_AudioStreamPlayer3D`
```
akien-mga, line 39: I think maintaining this list might end up being a hassle. The list of modules that can be disabled was already regularly outdated, so the same fate will likely meet this one.

I think it would be better if this was generated on CI for https://godot-build-options-generator.github.io/ so that the information can be displayed there when users want to disable 3D or advanced GUI nodes.

mhilbrunner:
I agree with the feedback above, and am going to close this due to lack of consensus.

Sorry we let it gather dust for so long!

PR #5151, opened 2021-08-04
Don't disable WebP when compiling small binaries in Optimizing for size

Calinou:
WebP decoding is now required in pretty much any project since Godot 3.4 (unless you force PNG for lossless textures in the Project Settings).

Therefore, it's not a good idea to recommend disabling it for most people. Not to mention the added binary size will be outweighted by the smaller lossless WebPs most of the time.

I can update https://godot-build-options-generator.github.io/ accordingly if we agree on this.

### DIFF
--- a/development/compiling/optimizing_for_size.rst
+++ b/development/compiling/optimizing_for_size.rst
@@ -63,7 +63,7 @@ a lot of them:
 
 ::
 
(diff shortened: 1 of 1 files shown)


merumelu:
I already added this in #5150 (but not for `master`).

mhilbrunner (approved):

akien-mga (approved):

akien-mga:
Thanks!

PR #5153, opened 2021-08-06
One phrase fix for idle_and_physics_processing.rst

phantomdesvin:
Both, the _process() and _physics_process() are not called every time the engine draws a frame like its written here, only -process() does. _physics_process() gets called before each physics step like it's explained on this same page.


mhilbrunner:
Good catch! Congrats on your first PR to the Godot docs :)

PR #5165, opened 2021-08-11
Document exporting glTF files from Godot

skyace65:
Documents how to export glTF files from Godot which is coming in 3.4. Most of this information comes from [this blog post](https://godotengine.org/article/introducing-the-godot-gltf-2-0-scene-exporter). The one added image has gone through png compression. Since the location of the asset workflow pages was changed in master do you need me to make another PR for the 3.4 branch?


lyuma:
The following are supported, so these two limitations should be removed.
```
* No support for exporting GridMaps
* No support for bone attachments
```
There was a bugfix for BoneAttachment export which landed yesterday. The bugfix will be backported to 3.4, so those should also be removed there.

Additionally, in Godot master, non-tools builds are supported, so this should be removed.
```
* Only supported in editor builds (tools=yes).
```
I'm not sure if it's been decided yet if this limitation should be removed from 3.4.

skyace65:
I've updated this PR and the 3.4 PR. Leaving the editor build limitation in the 3.4 version until that fix gets backported.

akien-mga:
Thanks!

PR #5174, opened 2021-08-15
Fix set_cell and set_cellv vector2 mention

honix:
Just minor text line swap.


Calinou:

honix:
Got it!

Checked ```godot/doc/classes/TileMap.xml``` on 3.x and 3.3 and 3.2. And it's good and no need to fix!

It's funny I can't find where I saw the problem first time...

PR #5175, opened 2021-08-15
Fix sub resource ID formatting in Object Class

hpnrep6:
https://docs.godotengine.org/en/latest/development/cpp/object_class.html

## Before

Screenshot from 2021-08-15 19-10-09

## After

Screenshot from 2021-08-15 19-08-54

### DIFF
--- a/development/cpp/object_class.rst
+++ b/development/cpp/object_class.rst
@@ -309,7 +309,7 @@ Saving a resource can be done with the resource saver API:
 Instance will be saved. Sub resources that have a path to a file will be
 saved as a reference to that resource. Sub resources without a path will
 be bundled with the saved resource and assigned sub-IDs, like
-"\res://someresource.res::1". This also helps to cache them when loaded.
+``res://someresource.res::1``. This also helps to cache them when loaded.
 
 References:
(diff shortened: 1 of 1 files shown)


development/cpp/object_class.rst
```diff
@@ -309,7 +309,7 @@ Saving a resource can be done with the resource saver API:
 Instance will be saved. Sub resources that have a path to a file will be
 saved as a reference to that resource. Sub resources without a path will
 be bundled with the saved resource and assigned sub-IDs, like
-"res://someresource.res::1". This also helps to cache them when loaded.
+"\res://someresource.res::1". This also helps to cache them when loaded.
```
Calinou, line 312: I'd suggest using an inline code block instead:

```suggestion
``res://someresource.res::1``. This also helps to cache them when loaded.
```

Calinou (changes requested):
Thanks!

hpnrep6:
Applied the changes!

Calinou (approved):
Thanks!

Cherry-picked to the `3.4`, `3.3`, `stable` and `3.2` branches.

PR #5184, opened 2021-08-19
Make pragma once sub-header clearer in C++ usage guidelines

hpnrep6:
https://docs.godotengine.org/en/latest/community/contributing/cpp_usage_guidelines.html#pragma-once

Currently, it looks more like it's misplaced than a sub-header due to inline code blocks remaining a constant font size.

## Before
Screenshot from 2021-08-19 02-26-28

## After
Screenshot from 2021-08-19 02-26-45


Calinou (approved):
Thanks!

PR #5186, opened 2021-08-20
Small navigation page fix (3.4)

skyace65:
I got radius and diameter confused when making the page. This fixes that.

### DIFF
--- a/tutorials/navigation/real_time_navigation_3d.rst
+++ b/tutorials/navigation/real_time_navigation_3d.rst
@@ -39,8 +39,8 @@ navigation mesh above the areas in your scene that can be navigated 
 .. image:: img/baked_navmesh.png
 
 Keep in mind that the navmesh shows where the center of an entity can
-go. For example, if you set the agent radius to 1 then your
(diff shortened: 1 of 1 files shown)


tutorials/navigation/real_time_navigation_3d.rst
```diff
@@ -39,7 +39,7 @@ navigation mesh above the areas in your scene that can be navigated 
 .. image:: img/baked_navmesh.png
 
 Keep in mind that the navmesh shows where the center of an entity can
-go. For example, if you set the agent radius to 1 then your
+go. For example, if you set the agent radius to .5 then your
 navigation mesh will have a distance of .5 from any ledges or walls
```
Calinou, line 43: While we're at it, I'd prefer using explicit leading zero to look less confusing at a distance:

```suggestion
go. For example, if you set the agent radius to 0.5 then your
navigation mesh will have a distance of 0.5 from any ledges or walls
```

Calinou (changes requested):

skyace65:
Fixed

Calinou (approved):
Thanks!

PR #5194, opened 2021-08-24
Add a page on using fonts in Godot

### LINKED ISSUE #4150: Document working with text and fonts in UI
For the User Interface section.

### LINKED ISSUE #4467: No basic information on how to feed BitmapFont
**Your Godot version:**
3.2.3.stable.official

**Issue description:**
There is no notice in the documentation what are the restrictions on the data produced by BMFont. Inside of the code, there is no validation whether the format is correct (like `BMF` which is present in the binary header). It's assumed that the format is text. Why not binary?

Additionally, it seems that the code needs 32-bit channel texture. Why?

There is a setting for using distance fields. This option is not present when exporting from BMFont, so one starts to question when it should be turned on. And if yes, then should the user assume that 8-bits on one channel represents a normalized signed float?

https://docs.godotengine.org/en/stable/classes/class_bitmapfont.html#bitmapfont

### LINKED ISSUE #4743: Document the bitmap font import and usage workflow
### Describe the project you are working on

A 2D game that uses bitmap fonts

### Describe the problem or limitation you are having in your project

I have a bitmap font in Angelcode, .fnt file fomat, created using BMFont (https://www.angelcode.com/products/bmfont/) .
I can import this font format without issue in Godot; it is automatically converted into an appropriate BitmapFont resource.

Here is an example for you to try: 
[bitmap_font_resource_example.zip](https://github.com/godotengine/godot-proposals/files/6125209/bitmap_font_resource_example.zip)

To get the file imported, I have to
- Select a Label node, for example, and target its Custom Font property
- Click "Load" in the font resource slot submenu. The file chooser appears. The .fnt file does not show up inside the explorer.
- Set the file type filter to All files
- Now all file types are visible, I can select my .fnt file

From this point, the bitmap font is correctly imported and can be used. But it is not straightforward to get it done.

My first reflexe was to create a new BitmapFont and add the .fnt file from there, which is not possible. This is confusing. I had to guess the correct procedure with trial and error. No documentation was found either about that possibility, which gives great results for my needs. This deserves to be more visible!

As far as I can remember, there was a way to generate bitmap fonts inside Godot 2. It was removed. Is the feature treated as an easter egg? Seems so! :smile: 

### Describe the feature / enhancement and how it helps to overcome the problem or limitation

- In the Font resource import dialog, add an option for .fnt files and make them visible by default as recognized file type.
- Document the bitmap font import process.

Maybe: rework the overall font import UX so that the process is similar for bitmap and dynamic fonts.

### Describe how your proposal will work, with code, pseudo-code, mock-ups, and/or diagrams

The .fnt file is added as a recognizable type from the Font resource import dialog.

### If this enhancement will not be used often, can it be worked around with a few lines of script?

No, it is a UX issue.

### Is there a reason why this should be core and not an add-on in the asset library?

Font resources are part of the core.

skyace65:
Adds a small page on using fonts in Godot. Both images went through png compression. Closes #4743 closes #4150 and closes #4467


Calinou:
There are some things I'd clarify:

- DynamicFont supports TrueType (`.ttf`) and OpenType (`.otf`) font formats. Additionally, the `master` branch supports Web Open Font Format 1 (`.woff`) is supported, but not WOFF2 (`.woff2`) as it requires a Brotli compression implementation to be linked in the engine. This WOFF1 support may be backportable to `3.x` too (it might be just a matter of exposing it in the DynamicFontData property hint).
- Bitmap font files (`.fnt`) are typically generated with a tool such as BMFont, which should be linked on the page.

skyace65:
I've made adjustments based on your feedback. However, I want to keep that distance field fonts are broken, otherwise we'll end up with confused users and people asking "why isn't this documented?" Which happened in #4467. Additionally should we be encouraging the use of bitmap fonts by linking to BMFont? I was under the impression that they're being used less and less these days, or is there a use case for them I don't know about? And do you think your .woff PR will get merged in time for 3.4? If so I can adjust this PR and wait for that to get merged before this does.

Calinou:
Bitmap fonts are still useful when you need a special appearance (soft shadows, gradients) but don't want to use a shader on every node that displays text. That said, BMFont doesn't have a way to add gradients or soft shadows. libGDX's Hiero does, but it's no longer maintained by the looks of it.

tutorials/ui/gui_using_fonts.rst
```diff
@@ -0,0 +1,56 @@
...
+the inspector settings for themes under **Default Font**.
+
+Regardless of where you are creating the font the process is
+the same. Click on the font setting you want to add a font to
+and a dropdown will appear with two choices: **New Bitmap Font**
+and **New Dynamic Font**.
+
+.. image:: img/new_font.png
+
+Bitmap fonts (``.fnt``) are a format created by `BMFont <http://www.angelcode.com/products/bmfont/h>`__.
```
mhilbrunner, line 21:
```suggestion
Bitmap fonts (``.fnt``) are a format created by `BMFont <http://www.angelcode.com/products/bmfont/>`__ (many BMFont compatible programs exist, like `BMGlyph <http://www.bmglyph.com/>`__).
```

tutorials/ui/gui_using_fonts.rst
```diff
@@ -0,0 +1,56 @@
+.. _doc_gui_using_fonts:
+
+Using Fonts
+===========
+
+Godot allows you to set specific fonts for different UI nodes.
+
+There are 3 places you can setup font usage. The first is the
```
mhilbrunner, line 8:
```suggestion
There are three different places where you can setup font usage. The first is the
```

tutorials/ui/gui_using_fonts.rst
```diff
@@ -0,0 +1,56 @@
...
+Regardless of where you are creating the font the process is
+the same. Click on the font setting you want to add a font to
+and a dropdown will appear with two choices: **New Bitmap Font**
+and **New Dynamic Font**.
+
+.. image:: img/new_font.png
+
+Bitmap fonts (``.fnt``) are a format created by `BMFont <http://www.angelcode.com/products/bmfont/h>`__.
+If you are using a ``.fnt`` file do not select New Bitmap
+Font. Instead, drag the file from the file system panel to the
```
mhilbrunner, line 23:
```suggestion
If you are using a ``.fnt`` file do not select **New Bitmap
Font**. Instead, drag the file from the file system panel to the
```

tutorials/ui/gui_using_fonts.rst
```diff
@@ -0,0 +1,56 @@
...
+font slot. Selecting the font after adding it to the font slot
+gives you 3 settings that can be changed: height, ascent and
+distance field. Distance field fonts were never reimplemented
+for Godot 3.0 and do not work.
```
mhilbrunner, line 27:
```suggestion
gives you three settings that can be changed: **height**, **ascent** and
**distance field**. Distance field fonts were never reimplemented
for Godot 3 and do not work in that version.
```

mhilbrunner (commented):
Some small formatting stuff, otherwise looks very good. :)

skyace65:
Fixed.

mhilbrunner (approved):

mhilbrunner:
Thanks a ton! :)

PR #5196, opened 2021-08-24
Recommend LLDB over GDB for VSCode

ThreeRhinosInAnElephantCostume:
Due to extreme performance issues discussed on the contributor chat, I took the liberty to modify the default configuration for LinuxBSD to encourage the use of LLDB. For less experienced developers, LLDB is indistinguishable from GDB anyway.

### DIFF
--- a/development/cpp/configuring_an_ide/visual_studio_code.rst
+++ b/development/cpp/configuring_an_ide/visual_studio_code.rst
@@ -89,8 +89,9 @@ To run and debug the project you need to create a new configuration in the ``lau
 
     {
       "name": "Launch Project",
-      "type": "cppdbg",
+      "type": "lldb",
       "request": "launch",
+      // Change to godot.linuxbsd.tools.64.llvm for llvm-based builds.
       "program": "${workspaceFolder}/bin/godot.linuxbsd.tools.64",
       // Change the arguments below for the project you want to test with.
       // To run the project instead of editing it, remove the "--editor" argument.
@@ -98,9 +99,26 @@ To run and debug the project you need to create a new configuration in the ``lau
       "stopAtEntry": false,
       "cwd": "${workspaceFolder}",
       "environment": [],
-      "externalConsole": true,
-      "MIMode": "gdb",
-      "setupCommands": [
+      "externalConsole": false,
+      "preLaunchTask": "build"
+    }
+  .. code-tab:: js LinuxBSD_gdb
+
+    {
+      "name": "Launch Project",
+      "type": "cppdbg",
+      "request": "launch",
+      // Change to godot.linuxbsd.tools.64.llvm for llvm-based builds.
+      "program": "${workspaceFolder}/bin/godot.linuxbsd.tools.64", 
+      // Change the arguments below for the project you want to test with.
+      // To run the project instead of editing it, remove the "--editor" argument.
+      "args": [ "--editor", "--path", "path-to-your-godot-project-folder" ],
+      "stopAtEntry": false,
+      "cwd": "${workspaceFolder}",
+      "environment": [],
+      "externalConsole": false,
+      "setupCommands": 
+      [
         {
           "description": "Enable pretty-printing for gdb",
           "text": "-enable-pretty-printing",
@@ -133,6 +151,17 @@ To run and debug the project you need to create a new configuration in the ``lau
    :align: center
 
    An example of a filled out ``launch.json``.
+
+
+.. note::
+
+    Due to sporadic performance issues, it is recommended to use LLDB over GDB on Unix-based systems.
+    Make sure that the `CodeLLDB extension <https://marketplace.visualstudio.com/items?itemName=vadimcn.vscode-lldb>`_
+    is installed.
+
+    If you encounter issues with lldb, you may consider using gdb (see the LinuxBSD_gdb configuration).
+
+    Do note that lldb may work better with llvm-based builds. See :ref:`doc_compiling_for_linuxbsd` for further information.
 
 The name under ``program`` depends on your build configuration,
 e.g. ``godot.linuxbsd.tools.64`` for 64-bit LinuxBSD platform with ``tools`` enabled.


development/cpp/configuring_an_ide/visual_studio_code.rst
```diff
@@ -83,13 +83,13 @@ To run and debug the project you need to create a new configuration in the ``lau
 - Select **C++ (GDB/LLDB)**. There may be another platform specific option here. If selected,
   adjust the configuration example provided accordingly.
 - Within the ``launch.json`` file find the ``"configurations"`` array and add a new section to it:
-
+.. note:: Due to sporadic performance issues, it is recommended to use LLDB over GDB on Unix-based systems. Make sure that the `CodeLLDB extension <https://marketplace.visualstudio.com/items?itemName=vadimcn.vscode-lldb>`_ is installed.
```
pycbouh, line 86: This needs to have an empty line before and after, and we also keep the text lines to a wrapping length.

```suggestion

.. note::

    Due to sporadic performance issues, it is recommended to use LLDB over GDB on Unix-based systems.
    Make sure that the `CodeLLDB extension <https://marketplace.visualstudio.com/items?itemName=vadimcn.vscode-lldb>`_
    is installed.

```

I would also probably move it below the config examples, so that the flow is not broken.

pycbouh (changes requested):

development/cpp/configuring_an_ide/visual_studio_code.rst
ThreeRhinosInAnElephantCostume: Should I squash it?

pycbouh: You didn't move it below yet 🙃 But sure, squash it if you can.

pycbouh (approved):

ThreeRhinosInAnElephantCostume:
Should be ready to merge now.

pycbouh (approved):

groud:
To be noted, while I switched from gdb to lldb too, I figured out that some variables were missing when debugging with lldb. Also, conditional breakpoints aren't working for me (maybe it's due to variable identification not working great though?).

It would be worth checking if others have the same issues. If it's only me then we can go to lldb, but otherwise, it might be at least necessary to warn users.

ThreeRhinosInAnElephantCostume:
> To be noted, while I switched from gdb to lldb too, I figured out that some variables were missing when debugging with lldb. Also, conditional breakpoints aren't working for me (maybe it's due to variable identification not working great though?).
> 
> It would be worth checking if others have the same issues. If it's only me then we can go to lldb, but otherwise, it might be at least necessary to warn users.

Could give me a more specific example? I had similar issues with gdb too, haven't noticed anything like that with lldb, but it's possible that they're both kinda buggy.

groud:
This is a typical example of what is happening, here the `ae` variable is not listed in the debugger:
2021-08-25-095705_1041x371_scrot

It seem that any variable which is not declared for the whole function scope gets ignored, while it used to work with gdb (the only issue was that the first breakpoint would take veeeeery long to trigger, but everything worked fine otherwise).

ThreeRhinosInAnElephantCostume:
image
Doesn't seem to happen on my machine. I am using lldb version version 11.0.1.

groud:
> Doesn't seem to happen on my machine. I am using lldb version version 11.0.1.

I tried with lldb 12 and 11.1.0, still the same issue. That's very annoying for me :/
What's your vscode version?

ThreeRhinosInAnElephantCostume:
> > Doesn't seem to happen on my machine. I am using lldb version version 11.0.1.
> 
> I tried with lldb 12 and 11.1.0, still the same issue. That's very annoying for me :/
> What's your vscode version?

```
> code --version
1.59.1
3866c3553be8b268c8a7f8c0482c0c0177aa8bfa
x64
```
Running on Kubuntu 21.04.

groud:
@ThreeRhinosInAnElephantCostume Which extension did you install to setup lldb ? I use the one from Vadim Chugunov (version 1.6.5) and it comes with a bundled lldb version 13.0.0. 

Anyway, I think I might have found why I don't see those variable. lldb spits me those errors:
```
error: godot.linuxbsd.tools.64 {0x0d27b655}: DIE has DW_AT_ranges(0x530703) attribute, but range extraction failed (missing or invalid range list table), please file a bug and attach the file at the start of this error message
error: godot.linuxbsd.tools.64 {0x0d27b655}: DIE has DW_AT_ranges(0x530703) attribute, but range extraction failed (missing or invalid range list table), please file a bug and attach the file at the start of this error message
error: godot.linuxbsd.tools.64 {0x0527d0ad}: DIE has DW_AT_ranges(0x19a3bc) attribute, but range extraction failed (missing or invalid range list table), please file a bug and attach the file at the start of this error message
error: godot.linuxbsd.tools.64 {0x0527d0ad}: DIE has DW_AT_ranges(0x19a3bc) attribute, but range extraction failed (missing or invalid range list table), please file a bug and attach the file at the start of this error message
...
```
It looks like a bug but I am not sure what can I do for it. I tried changing the llvm library path in the settings but it did not work.

ThreeRhinosInAnElephantCostume:
I am going to test this in a fresh VM later today, should've done that to begin with.

It looks like I also have this extension installed: https://marketplace.visualstudio.com/items?itemName=webfreak.debug .

mhilbrunner:
Would it make sense to document both?

groud:
> Would it make sense to document both? I.e. have the default, and a note on what to change for the other debugger?

I guess it would not hurt, considering it seems that both solution seem to have drawbacks.

To end on my issue, I managed to use lldb correctly (without having some variable missing) by compiling using llvm. It might be worth to mention that lldb is the same set of tool than llvm, and that it might help to compile with llvm when you use lldb.

Fallayn (approved):

mhilbrunner:
Thanks for making the changes we discussed in chat. :)

Calinou:

PR #4465, opened 2020-12-16
Update services_for_ios.rst

Xioor:
Proposing these changes since there is missing documentation on finalizing a purchase for iOS. https://github.com/godotengine/godot-docs/issues/2715

PR #5197, opened 2021-08-25
Always use variables to access iOS-only singletons in Plugins for iOS

Calinou:
Follow-up to https://github.com/godotengine/godot-docs/pull/4465.

This is required to avoid script compilation errors on other platforms.

@naithar Could we expose dummy singletons that do nothing on other platforms to avoid the need to do this?

### DIFF
--- a/tutorials/platform/ios/plugins_for_ios.rst
+++ b/tutorials/platform/ios/plugins_for_ios.rst
@@ -8,7 +8,7 @@ They are using same model of asynchronous calls explained below.
 
 ARKit and Camera access are also provided as plugins.
 
-Latest updates, documentation and source code can be found at `Godot iOS plugins repository <https://github.com/godotengine/godot-ios-plugins>`_ 
+Latest updates, documentation and source code can be found at `Godot iOS plugins repository <https://github.com/godotengine/godot-ios-plugins>`_
 
 Accessing plugin singletons
 ---------------------------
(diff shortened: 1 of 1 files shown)


Difference:
Both stable and latest docs are missing an underscore in product_id in the responce event example 
image

naithar:
> @naithar Could we expose dummy singletons that do nothing on other platforms to avoid the need to do this?

I'm not sure why it's needed, especially since this singletons are coming from non-engine Godot code. Would it be required to create singleton for every possible plugin?

As far as I understand Android plugins are used the same way - singleton are stored in a variable: https://docs.godotengine.org/en/stable/tutorials/platform/android_in_app_purchases.html. So I don't see a reason to handle iOS plugins differently.

Calinou:
> Both stable and latest docs are missing an underscore in product_id in the responce event example

Fixed :slightly_smiling_face: 

(This will be visible in the deployed documentation once this pull request is merged and cherry-picked to the `stable` branch.)

mhilbrunner:
Thanks!

Difference:
It seems the SKPaymentTransactionStateRestored also returns the app receipt in ["receipt"] .
This should be documented, because it is currently the only way to get the receipt.

In my understanding, it is the recommend way to check date of purchase (important if updating app form Paid->Premium) and also to check in app purchases, without calling restore_purchases.

I think there to important functions should be added to the IOS plugin:
RefreshRecipt()  see https://developer.apple.com/documentation/storekit/skreceiptrefreshrequest
GetRecipt() returning the receipt in the same way that SKPaymentTransactionStateRestored gets it.

PR #5214, opened 2021-08-30
Use pip3 instead of pip

ThreeRhinosInAnElephantCostume:
`pip` may refer to the old Python 2 package manager, possibly causing confusion. Missing `pip3` causes a straight-forward that's easily googlable and accounted for by a note.


mhilbrunner:
Thanks!

Calinou:

PR #5216, opened 2021-09-01
Remove possible duplicity of code.

yuvallangerontheroad:
If there is a reason to keep it, could you give an explanation?

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_advanced.rst
+++ b/tutorials/scripting/gdscript/gdscript_advanced.rst
@@ -441,7 +441,6 @@ functions in your script. An example implementation of a forward iterator follow
             return (current < end)
 
         func _iter_init(arg):
-            current = start
             return should_continue()
 
         func _iter_next(arg):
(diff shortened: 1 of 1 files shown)


tutorials/scripting/gdscript/gdscript_advanced.rst
```diff
@@ -441,7 +441,6 @@ functions in your script. An example implementation of a forward iterator follow
             return (current < end)
```
Calinou, line 441: Remove unnecessary parentheses per the GDScript style guide:

```suggestion
            return current < end
```

aaronfranke:
Isn't it necessary to reset the state to the beginning when initializing the iterator? See this doc:

> Make sure to reset the state of the iterator in ``_iter_init``, otherwise nested
for-loops that use custom iterators will not work as expected.

yuvallangerontheroad:
> Isn't it necessary to reset the state to the beginning when initializing the iterator? See this doc:
> 
> > Make sure to reset the state of the iterator in `_iter_init`, otherwise nested
> > for-loops that use custom iterators will not work as expected.

Embarrassing! Sorry. I'll close the pull request now.

PR #5228, opened 2021-09-07
Add note that C# tools plugin requires the mono SDK on Linux

### LINKED ISSUE #5223: Mention that you still need Mono SDK when using VSCode (Linux)
**Your Godot version:**
3.3.3 Mono-enabled (Linux)

**Issue description:**
The docs for setting up C# state that [Mono isn't required anymore](https://docs.godotengine.org/en/stable/getting_started/scripting/c_sharp/c_sharp_basics.html#configuring-an-external-editor):

```
From Godot 3.2.3 onwards, installing Mono SDK is not a requirement anymore, except it is required if you are building the engine from source.
...
In summary, you must have installed .NET Core SDK and the Mono-enabled version of Godot.
```

While technically true, there's a gottcha for developers using VSCode as their editor. The [C# Tools for Godot](https://marketplace.visualstudio.com/items?itemName=neikeq.godot-csharp-vscode) extension, which the docs mention, [requires Mono on Linux](https://github.com/godotengine/godot-csharp-vscode/issues/15). There's no mention of this in Godot's VSCode setup docs.

I read the initial requirements about Godot only needing .NET so I grabbed Microsoft's SDK and wasted several hours figuring out why debugging and code completion didn't work in VSCode. I eventually found the issue on the extension's repo and tried Mono, which got it working.

I propose a note be added to Godot's setup docs for C# about the extension still requiring Mono for Linux users.

https://docs.godotengine.org/en/stable/getting_started/scripting/c_sharp/c_sharp_basics.html#visual-studio-code

skyace65:
What the title says. Closes #5223

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_basics.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_basics.rst
@@ -123,6 +123,10 @@ In Visual Studio Code:
 - Install the `Mono Debug <https://marketplace.visualstudio.com/items?itemName=ms-vscode.mono-debug>`__ extension.
 - Install the `C# Tools for Godot <https://marketplace.visualstudio.com/items?itemName=neikeq.godot-csharp-vscode>`__ extension.
 
+.. note:: if you are using linux you must install the
+          `Mono SDK <https://www.mono-project.com/download/stable/#download-lin>`__
+          for the C# tools plugin to work.
+
 To configure a project for debugging open the Godot project folder in VS Code.
 Go to the Run tab and click on **create a launch.json file**. Select **C# Godot** from the dropdown
 menu. Now, when you start the debugger in VS Code your Godot project will run.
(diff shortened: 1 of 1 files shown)


tutorials/scripting/c_sharp/c_sharp_basics.rst
```diff
@@ -123,6 +123,10 @@ In Visual Studio Code:
 - Install the `Mono Debug <https://marketplace.visualstudio.com/items?itemName=ms-vscode.mono-debug>`__ extension.
 - Install the `C# Tools for Godot <https://marketplace.visualstudio.com/items?itemName=neikeq.godot-csharp-vscode>`__ extension.
 
+.. note:: if you are using linux you must install the
```
mhilbrunner, line 126:
```suggestion
.. note:: If you are using linux you must install the
```

aaronfranke (approved):
I would still be in favor of just always telling users to go install Mono, but @neikeq insisted that it wasn't necessary anymore so we should not do that. Since we don't, we for sure need to tell people to install Mono when it's needed, so this PR makes sense.

neikeq:
> I would still be in favor of just always telling users to go install Mono, but @neikeq insisted that it wasn't necessary anymore so we should not do that. Since we don't, we for sure need to tell people to install Mono when it's needed, so this PR makes sense.

Installing Mono is not needed for Godot, and it will be even less needed in the future where it will need the .NET (6) SDK.

neikeq (approved):

mhilbrunner:
Thanks!

PR #5232, opened 2021-09-09
Updated description for PhysicalBone.

Favkis:
I've finally figured out what causes most problems with wrong ragdoll behavior, and it's that center of mass is located in Physicalbone and you're free to change it. I've never seen anyone say that, haven't seen it anywhere in the docs and no one was able to help me. But here it is, solution was very simple in the end. 
If you create PhysicalBone and select a bone it is attached too, it automatically moves to the HEAD of a selected bone, making it look like it has to be there, and if you change it's position, it'll be oribiting the bone. Turns out that no, you are free to move it to change it's center of mass. And you HAVE to move it to the center of bone if you're making a ragdoll, otherwise your physbox would have center of mass out of collision shape's bounds and ragdoll would do weird things.

(However it creates a bug that if you partially ragdoll only 1 bone that has it's origin (PhysicalBone) offesetted from HEAD of the bone it is attached it, whole physical bone would offset itself to match it's origin to the head of a bone. That can't be intended, so I guess it would be fixed later.)

### DIFF
--- a/classes/class_physicalbone.rst
+++ b/classes/class_physicalbone.rst
@@ -12,6 +12,9 @@ PhysicalBone
 **Inherits:** :ref:`PhysicsBody<class_PhysicsBody>` **<** :ref:`CollisionObject<class_CollisionObject>` **<** :ref:`Spatial<class_Spatial>` **<** :ref:`Node<class_Node>` **<** :ref:`Object<class_Object>`
 
 
+Description
+----------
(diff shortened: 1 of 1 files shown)


mhilbrunner:
Hey! Thanks for your contribution.

The class docs (`.rst` files under `classes/`) are autogenerated from XML files in the main Godot repository:
https://github.com/godotengine/godot/tree/master/doc/classes

If you want to make change to the docs for a class, it needs to be done there.

Thannks for contributing. :)

PR #5233, opened 2021-09-09
Remove Q2 from 3.4 release date

DeeJayLSP:
Q2 ended more than two months ago, update it.


pycbouh:
Fair. Thanks!

aaronfranke:
Q3 doesn't seem very realistic either given that there are only 20 days left and we aren't in the release candidate stage yet.

PR #5234, opened 2021-09-10
Complying with licenses: Explain what the FreeType <year> should be

DeeJayLSP:
According to the FreeType License itself:

```
   """
    Portions of this software are copyright © <year> The FreeType
    Project (www.freetype.org).  All rights reserved.
   """

  Please replace <year> with the value from the FreeType version you
  actually use.

```
While this may be not much a problem, I think end users should know about this.


mhilbrunner:
Thanks!

Calinou:

PR #5235, opened 2021-09-11
Change estimated release date for Godot 3.4 and Godot 4.0

Calinou:
Follow-up to https://github.com/godotengine/godot-docs/pull/5058 and https://github.com/godotengine/godot-docs/pull/5233.

Godot 3.4 will most likely be released in October 2021 since it didn't reach RC stage yet (it's at beta 4 right now).

Godot 4.0 hasn't had an alpha released yet, so it's fair to say that it won't release in 2021 by now. The first half of 2022 is a more likely estimate.

PR #5236, opened 2021-09-11
Document that only the latest patch release in each series receives support

Calinou:
Can be merged independently of https://github.com/godotengine/godot-docs/pull/5235.


pycbouh:
I guess a clarification never hurts, but isn't that kind of a given? Do we have confused users who expect 3.3.1 to be patched without actually going through 3.3.2 and 3.3.3?

mhilbrunner:
No, this isn't so much about "support" as in further updates, but as in "giving support", or stated otherwise: if you run into an issue, update to the latest available _patch_ release for your major.minor version first, _then_ ask for support if it still happens.

E.g. if you're on 3.3.2, update to 3.3.3 before asking for support (which will just ask you to update to 3.3.3 anyway).

Thats how I read it at least. Maybe we should clarify, because I can see why you read it that way.

pycbouh:
I guess I see what you mean, but I'm not sure then how is that related to the release policy. I'm not against this tidbit being added, but it just seems to be a random statement for this article if that's the point of it.

Maybe as a note it could work.

mhilbrunner:
Yeah, I'm also not sure whether people who are likely to just ask for support first thing will instead be stopped by the docs saying this :) But at least it gives of something to point to, I guess?

mhilbrunner:
@Calinou Once you rebased, I'm fine with merging this.

mhilbrunner (approved):

akien-mga:
Thanks!

Calinou:

Issue #5206, opened 2021-08-28
C++ usage guidelines range based for loops


Ansraer:
**Your Godot version:**
bleeding edge

**Issue description:**
The usage guideline page still states that range based for loops are not allowed, which is no longer correct. Instead that paragraph should be changed to something along the lines of: "Most of godot's custom data types support range based for loops, but before using them please consider if it would be more efficient to iterate over the data manually."

**URL to the documentation page:**
https://docs.godotengine.org/en/latest/community/contributing/cpp_usage_guidelines.html#rationale

PR #5237, opened 2021-09-11
Remove outdated guideline about `for` range loops being forbidden

Calinou:
`for` range loops are now implemented in Godot's data containers (`master` branch only), and using them is allowed.

This closes https://github.com/godotengine/godot-docs/issues/5206.


mhilbrunner:
Yay, and thanks!

PR #5241, opened 2021-09-12
use only right-click to remove tile from gridmap

nsrCodes:
The docs suggest to use shift + RMB. Which works too, but only RMB also works. This might confuse as later in the documentation shift + LMB is used for bulk selection.

Edit: This PR also adds a line to explain how to change floors when using gridmap.

### DIFF
--- a/tutorials/3d/using_gridmaps.rst
+++ b/tutorials/3d/using_gridmaps.rst
@@ -85,8 +85,9 @@ The "Cell/Size" property should be set to the size of your meshes. You can leave
 it at the default value for the demo. Set the "Center Y" property to "Off".
 
 Now you can start designing the level by choosing a tile from the palette and
-placing it with Left-Click in the editor window. To remove a tile, hold :kbd:`Shift`
-and use Right-click.
+placing it with Left-Click in the editor window. Use Right-click to remove a tile.
+
+Use the arrows next to the "GridMap" menu to change the floor that you are working on.
 
 Click on the "GridMap" menu to see options and shortcuts. For example, pressing
 :kbd:`S` rotates a tile around the y-axis.


pycbouh:
@groud how is it supposed to work?

nsrCodes:
I did not think of this as a bug. How can I confirm this?

nsrCodes:
Also, can I add a small line explaining how to shift floors? Because they are not mentioned anywhere in this documentation and since I had never used a level editor before it wasn't directly obvious as to how to start drawing on another level.

Calinou:
> Also, can I add a small line explaining how to shift floors? Because they are not mentioned anywhere in this documentation and since I had never used a level editor before it wasn't directly obvious as to how to start drawing on another level.

Sure :slightly_smiling_face:

pycbouh:
> I did not think of this as a bug.

The label means it's a bug in the documentation. 🙃 

> How can I confirm this?

Well, you can look at the code, ask the one who made the tool, or wait for others to check this PR out in case they know.

> Also, can I add a small line explaining how to shift floors?

Sure, it's okay to make other corrections to the article.

nsrCodes:
Confirmed! [The code](https://github.com/godotengine/godot/blob/5f69218edc4f4af574c7a65f077e2e0bf1f28365/modules/gridmap/grid_map_editor_plugin.cpp#L640-L653) works only with LMB.
This PR that worked on improving gridmap clearly states this in its description too - https://github.com/godotengine/godot/pull/25406

> Sure, it's okay to make other corrections to the article.

I meant in the same PR?

pycbouh:
> I meant in the same PR?

Yes, it's okay to do this in the same PR. Squash it into one commit and use a more generic commit message if you do. Something like "Improve GridMaps tutorial and correct editor shortcuts".

mhilbrunner (approved):

mhilbrunner:
Thanks and congrats on your first contribution to the Godot docs!

PR #5238, opened 2021-09-11
Document a workaround for testing scripts without `test()`

ThreeRhinosInAnElephantCostume:
This is a rather obscure feature that's unlikely to ever by used by anything other than the PR that implements it https://github.com/godotengine/godot/pull/52068.

Nevertheless, I think it's worth documenting.

PR #5257, opened 2021-09-15
Fix an incorrect pattern in the unit testing documentation

ThreeRhinosInAnElephantCostume:
Corrects the pattern in the note added by #5238.


mhilbrunner:
Missed that. Thanks for the fix!

PR #5274, opened 2021-09-22
Add example of error handling with load_resource_pack

gelvinp:
The current stable documentation says that ProjectSettings.load_resource_pack() does not provide any error or exception if the import fails, yet the same stable documentation for the function [found here](https://docs.godotengine.org/en/stable/classes/class_projectsettings.html#class-projectsettings-method-load-resource-pack) says that the function returns true on success. I updated the example to reflect this.


pycbouh:
Your PR needs to be made against another branch. `stable` is used for stable documentation, but is not modified directly. If your PR is only relevant to the stable documentation, you need to target 3.4. If the same article and the same error exists in the latest branch of the docs, you need to target `master`. In case the master version is too different and cannot be cherrypicked, it's okay to leave this PR as a 3.x PR and make a dedicated master PR.

gelvinp:
Ah, got it. Looks like I can't just switch this PR to 3.4 either so I'll close it out and make another from the correct branch, sorry.

pycbouh:
You can rebase your commit and force push it. But opening a fresh one is okay as well.

PR #5275, opened 2021-09-22
[3.x] Add example of error handling with load_resource_pack

gelvinp:
The current stable documentation says that ProjectSettings.load_resource_pack() does not provide any error or exception if the import fails, yet the same stable documentation for the function [found here](https://docs.godotengine.org/en/stable/classes/class_projectsettings.html#class-projectsettings-method-load-resource-pack) says that the function returns true on success. I updated the example to reflect this.

Same change as #5274 but now with commits made against correct branch. 
Additionally, the entire "Workflow" category of pages doesn't appear in the latest documentation, thus this change is only relevant for 3.x.

### DIFF
--- a/getting_started/workflow/export/exporting_pcks.rst
+++ b/getting_started/workflow/export/exporting_pcks.rst
@@ -89,27 +89,33 @@ process will build that type of file for the chosen platform.
 Opening PCK files at runtime
 ----------------------------
 
-To import a PCK file, one uses a one-liner. Keep in mind, there is no
-error or exception if the import fails. Instead, one might have to create some
-validation code as a layer on top. The following example expects a “mod.pck”
-file in the directory of the games executable. The PCK file contains a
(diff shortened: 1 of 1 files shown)


pycbouh:
> Additionally, the entire "Workflow" category of pages doesn't appear in the latest documentation, thus this change is only relevant for 3.x.

We had a bit of reorganization for the next version, the document still exists: https://docs.godotengine.org/en/latest/tutorials/export/exporting_pcks.html

But it's probably not cherrypickable due to it being moved, so a dedicated PR for master would be appreciated 🙂

gelvinp:
Good catch, I'll grab that now

mhilbrunner (approved):

mhilbrunner:
Thanks, and congrats on your first contribution to the Godot docs! 🎉

Calinou:

PR #5278, opened 2021-09-22
Add example of error handling with load_resource_pack

gelvinp:
The current documentation says that ProjectSettings.load_resource_pack() does not provide any error or exception if the import fails, yet the documentation for the function says that the function returns true on success. I updated the example to reflect this.

Master branch version of #5275


mhilbrunner (approved):

mhilbrunner:
Thank you!

PR #5284, opened 2021-09-27
Slight clarify in optimization tutorial

albinaask:
I just tested colliding two rigid bodies each having 5000 Collision shapes and 5000 mesh instances with cube meshes , resulting in a total of 20 000 nodes in the tree. Godot does not seem to care the slightest... It takes a solid 10 seconds to spawn everything, but from there I have a solid 83% idle time according to the profiler (running a 4GHz desktop CPU that is).

Therefore this update should give a bit more guidance on when the scene tree may be a bottleneck.

### DIFF
--- a/tutorials/performance/cpu_optimization.rst
+++ b/tutorials/performance/cpu_optimization.rst
@@ -225,8 +225,11 @@ SceneTree
 
 Although Nodes are an incredibly powerful and versatile concept, be aware that
 every node has a cost. Built-in functions such as `_process()` and
-`_physics_process()` propagate through the tree. This housekeeping does start to reduce
-performance when you have very large numbers of nodes (depending on platform, but about 10% of frame time per ten thousand nodes on a desktop CPU).
+`_physics_process()` propagate through the tree. This housekeeping can reduce 
+performance when you have a very large numbers of nodes (how many exactly 
+depends on the target platform and can range from thousands to tens of 
+thousands so ensure that you profile performance on all target platforms 
+during development).
 
 Each node is handled individually in the Godot renderer. Therefore, a smaller
 number of nodes with more in each can lead to better performance.
--- a/tutorials/performance/cpu_optimization.rst
+++ b/tutorials/performance/cpu_optimization.rst
@@ -225,8 +225,8 @@ SceneTree
 
 Although Nodes are an incredibly powerful and versatile concept, be aware that
 every node has a cost. Built-in functions such as `_process()` and
-`_physics_process()` propagate through the tree. This housekeeping can reduce
-performance when you have very large numbers of nodes (usually in the thousands).
+`_physics_process()` propagate through the tree. This housekeeping does start to reduce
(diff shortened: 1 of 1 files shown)


tutorials/performance/cpu_optimization.rst
```diff
@@ -225,8 +225,8 @@ SceneTree
 
 Although Nodes are an incredibly powerful and versatile concept, be aware that
 every node has a cost. Built-in functions such as `_process()` and
-`_physics_process()` propagate through the tree. This housekeeping can reduce
-performance when you have very large numbers of nodes (usually in the thousands).
+`_physics_process()` propagate through the tree. This housekeeping does start to reduce
+performance when you have very large numbers of nodes (depending on platform, but about 10% of frame time per ten thousand nodes on a desktop CPU).
```
Calinou, line 229: I'm not sure if giving an "accurate" figure is a good idea here. People may start extrapolating based on this in weird ways, especially since the CPU model isn't mentioned.

albinaask: Mmmm, agreed. I however have lived for two years worrying about having 1000 nodes in my tree, which as it seems is completely unnecessary. So it may also lead to people trying to over optimize causing slowdown in production time. I do get your point and agree with you when I think about it, perhaps "does start to" is better than "can" but other than that we can close this PR

tutorials/performance/cpu_optimization.rst
```diff
@@ -225,8 +225,11 @@ SceneTree
 
 Although Nodes are an incredibly powerful and versatile concept, be aware that
 every node has a cost. Built-in functions such as `_process()` and
-`_physics_process()` propagate through the tree. This housekeeping can reduce
-performance when you have very large numbers of nodes (usually in the thousands).
+`_physics_process()` propagate through the tree. This housekeeping can reduce 
+performance when you have a very large numbers of nodes (how many exactly 
+depends on the target platform and can range from thousands to tens of 
+thousands so ensure that you profile performance on all target platforms 
+during development).
```
mhilbrunner, line 232:
```suggestion
`_physics_process()` propagate through the tree. This housekeeping can reduce 
performance when you have a very large numbers of nodes (how many exactly 
depends on the target platform and other factors).
```

mhilbrunner (approved):

mhilbrunner:
Thanks for contributing! 🎉

albinaask:
A million thanks for squashing @mhilbrunner , I sat for like four hours trying to do it, I just wound up with error messages and 16 commits instead of one and so on. This is my first PR and it's mesmerizing to be part of something this big!

mhilbrunner:
@albinaask No problem. In the future, you don't need to get frustrated or try to hours - you can join us on https://chat.godotengine.org. It may make some time to get responses as most of us are pretty busy, but we're very willing to help sort out any questions or issues. :) Especially if its to help out new contributors getting set up.

Good work, and thanks again!

PR #5288, opened 2021-09-30
Fix grammar issue in "Ways to contribute" page

katecoded:
In the section "Filing an issue on GitHub", the subsection "How to reproduce the bug" says "Even if you think that the issue is trivial to reproduce, adding a minimal project that lets reproduce it is a big added value." which is missing a word. Changing it to "...lets everyone reproduce it..." fixes the grammatical error.


mhilbrunner (approved):

mhilbrunner:
Thanks for the small fix, and congrats on your first contribution!

PR #5294, opened 2021-10-02
Fix `emit_signal` arity in Signals tutorial

JonathanHarford:
Example shows three arguments, but text suggested there were only two.


Calinou (approved):

PR #5295, opened 2021-10-02
Remove "Setting a main scene" section

JonathanHarford:
This step was covered on [the previous page](https://docs.godotengine.org/en/stable/getting_started/step_by_step/your_first_game.html#testing-the-scene).

### DIFF
--- a/getting_started/step_by_step/exporting.rst
+++ b/getting_started/step_by_step/exporting.rst
@@ -238,14 +238,6 @@ changed:
         }
     }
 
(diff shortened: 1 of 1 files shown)


pycbouh:
People often land on such pages from search engines, and it's not a huge section that just repeats something else. I think it's okay to leave as is.

Calinou:
Closing per @pycbouh's comment. Thanks for the contribution nonetheless!

PR #5297, opened 2021-10-05
classref: Sync with latest 4.0-dev

akien-mga:
This should fail CI, as a lot of classes have been renamed and existing tutorials need to be adapted to fix the renamed hyperlinks.

This can often involve actually porting/rewriting the tutorial if it's no longer accurate for the 4.0 API.

So we need to figure out how we want to handle this. Disable CI checks for invalid hyperlinks and merge as is? Or do all the porting work in this PR or in a dedicated branch, before merging to `master`?


pycbouh:
Branch sounds like a good idea, yeah.

akien-mga:
I pushed this to the `4.0-classref` branch directly: https://github.com/godotengine/godot-docs/tree/4.0-classref

PR #5301, opened 2021-10-06
Fix small typos regarding WebSocketPeer

Infinixius:


pycbouh:
Thanks for your contribution, but class reference is tracked in the main repo (and only synced down here). You need to make PR against files [here](https://github.com/godotengine/godot/tree/master/doc/classes).

See https://docs.godotengine.org/en/stable/community/contributing/updating_the_class_reference.html

Infinixius:
Thanks, went ahead and opened #53474 in the main repository.

Issue #5193, opened 2021-08-23
String format operator missing from the operator precedence list


nobuyukinyuu:
**Your Godot version:**
3.3.3

**Issue description:**
The operator precedence list for GDScript seems to be missing the precedence of the format operator `%` (or is ambiguous about contextual difference between format and modulo).  Strings being formatted in statements like `"out=a:%s" if condition else "b:%s" % val` may be ambiguous based on usage scenarios presented in the Format Strings doc as the operator is never used in compound statements in the examples.

**URL to the documentation page:**
https://docs.godotengine.org/en/stable/getting_started/scripting/gdscript/gdscript_basics.html#operators
https://docs.godotengine.org/en/stable/getting_started/scripting/gdscript/gdscript_format_string.html

Calinou:
cc @vnen

zacryol:
For reference, I just checked, and 
```
var val = "eee"
print("one %s" if true else "two %s" % val)
print("one %s" if false else "two %s" % val)
print(("one %s" if true else "two %s") % val)
``` 
gives the output 
```
one %s
two eee
one eee
```
so the format string does have higher precedence than the ternary

vnen:
It's not missing, it's just the same as the remainder/modulo operator since it uses the same symbol (`%`) which is grouped together with multiplication is division.

PR #5314, opened 2021-10-10
Add Format String to operator/precedence list

zacryol:
resolves https://github.com/godotengine/godot-docs/issues/5193
I verified that it is between `-` (Unary negation) and `*` `/` `%`.

`print(-"Hello %s" % "eee")` causes an error in the editor: `Invalid operands to operator %, Nil and String` meaning that the `-` applied to the string before the format string was evaluated.

`print("Hello %d" % 1 / 2)` also gives an in-editor error: `Invalid operands to operator /, String and int` meaning that the format string was evaluated before dividing.

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_basics.rst
+++ b/tutorials/scripting/gdscript/gdscript_basics.rst
@@ -240,7 +240,9 @@ The following is the list of supported operators and their precedence.
 |                                                                        | as C++. Integer division is truncated   |
 |                                                                        | rather than returning a fractional      |
 |                                                                        | number, and the % operator is only      |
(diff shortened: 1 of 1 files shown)


vnen:
> `print("Hello %d" % 1 / 2)` also gives an in-editor error: `Invalid operands to operator /, String and int` meaning that the format string was evaluated before dividing.

To be clear, it's only evaluated before dividing because operators with the same precedence are evaluated from left to right.

zacryol:
> In this case I would just add a note that `%` is both "remainder" and "string format".

Done.

mhilbrunner:
Thanks!

PR #5316, opened 2021-10-10
Clarify instructions on managing groups

AndreKoscianski:
Hi. I suggested a slight modification to give information right away about creating a new group programmatically, and also how to remove a node (these info were dispersed on forums).

But.. it still lacks a thing:  is it possible/necessary to delete a group? Or, are they automatically disposed of once there are no members left? I fell like the info should be explicitated.

### DIFF
--- a/tutorials/scripting/groups.rst
+++ b/tutorials/scripting/groups.rst
@@ -13,19 +13,16 @@ groups as you want. Then, in code, you can use the SceneTree to:
 This is a useful feature to organize large scenes and decouple code.
 
 
-Creating groups
------------------
+Managing groups
+---------------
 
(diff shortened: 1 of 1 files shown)


mhilbrunner:
Hey! Thank you for contributing. Sorry this took a bit to get a reply!

This looks good. :) To your question, it does indeed seem like removing the last node from a group erases the group automatically:
https://github.com/godotengine/godot/blob/9f7218eb69a3a1c09c652e9064573bffd71bdc02/scene/main/scene_tree.cpp#L145-L147

Update the PR with that info if you want, or let us know if you don't, then I'll go ahead, merge this and add that myself. THanks again!

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 3.4.

PR #5267, opened 2021-09-18
Add more porting providers to Console support in Godot

nightblade9:
Add two more publishers. Sorted alphabetically.
Disclosure: I do not work for or have used either of these publishers myself.

PR #5317, opened 2021-10-11
Revert "Add more porting providers to Console support in Godot"

akien-mga:
Reverts godotengine/godot-docs#5267

These were assumptions, it's not confirmed.

Ratalaika seems to be *working* on a Godot port for *Switch*, no further information is available at this time to confirm that they are actually able to port Godot games currently.

Deck13 is a publisher and there's no compelling information that they do console porting. (Publishing a Godot game on console doesn't mean they did the port.)

### DIFF
--- a/tutorials/platform/consoles.rst
+++ b/tutorials/platform/consoles.rst
@@ -62,16 +62,11 @@ your games to various consoles.
 
 Following is the list of providers:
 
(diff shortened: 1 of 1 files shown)


DeeJayLSP:
According to Deck13 website, one of the services they offer is porting. No clue about doing it with Godot though.

PR #5326, opened 2021-10-15
Show example of casting get_node() for static typing in style guide

zacryol:
~~Before possibly merging:
the part that mentions being `type-safe` should ideally link to the bit on Safe Lines in the docs page on static typing, but I couldn't figure out how to do that.~~

Since the style guide hasn't been updated to use annotations (`@onready` instead of `onready`) yet, this new section can be added to stable docs as-is (though probably not through simple cherry-pick, due to the restructuring).


mhilbrunner (approved):

mhilbrunner:
Thank you!

Calinou:

> (though probably not through simple cherry-pick, due to the restructuring).

I could cherry-pick this without conflicts thanks to Git's file rename detection :slightly_smiling_face:

PR #5329, opened 2021-10-17
Update Complying with Licenses page

nathanfranke:
- Clarify that it is not legal advice.
- Remove outdated third-party attributions, and point them to COPYRIGHT.txt instead.

### DIFF
--- a/about/complying_with_licenses.rst
+++ b/about/complying_with_licenses.rst
@@ -3,6 +3,26 @@
 Complying with licenses
 =======================
 
+.. warning::
+
+    The information on this page is intended to be helpful, but is **not legal advice**.
+    We recommend checking the licensing information in the relevant version of Godot's source code.
+    If you plan to distribute your game, we also recommend consulting with an attorney.
+
+.. note::
+
+    After reviewing this page, ensure you have met all of these requirements:
+
+    - You have an easily accessible license page either within your game or in an external file.
+
+    - The license page includes the full Godot license text (seen below).
+
+    - The license page includes the full attribution of all other third-party libraries.
+    If you automate this with code, Godot's license will be included automatically.
+
+    - The license page attributes all copyrighted assets based on their requirements.
+    For example, CC0 has no requirements, but CC-BY requires you attribute the creator.
+
 What are licenses?
 ------------------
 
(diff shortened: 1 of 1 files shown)


nathanfranke:
Q: I am working on an [addon](https://github.com/nathanfranke/nicense) to simplify the attribution process. Would it be okay to add a note of it to this documentation?

nathanfranke:
Updated and added some sample code on how to generate a licenses text label

about/complying_with_licenses.rst
```diff
@@ -11,20 +17,25 @@ It doesn't have a sole owner either, as every contributor that submits code to
...
-text somewhere in your game or derivative project.
+For Godot itself, the only requirement is to include the license text somewhere in your game or derivative project.
 
 This text reads as follows:
 
-    This game uses Godot Engine, available under the following license:
+    This project uses Godot Engine, available under the following license:
 
     Copyright (c) 2007-2021 Juan Linietsky, Ariel Manzur.
     Copyright (c) 2014-2021 Godot Engine contributors.
```
mhilbrunner, line 41:
```suggestion
    Copyright (c) 2007-2022 Juan Linietsky, Ariel Manzur.
    Copyright (c) 2014-2022 Godot Engine contributors.
```

nathanfranke:
I can't guarantee formatting checks will pass, so I'll need the workflows to be approved :slightly_smiling_face:

mhilbrunner (approved):
Changes look good to me. WDYT, @akien-mga?

akien-mga:
There are some good updates in this page, but it also goes in the general direction of making the requirements very extreme, to a point where personally I'd skip using Godot and go for an engine with EULA just so that I don't have to deal with such requirements.

If the page starts by saying that I can't ship a game without consulting with an attorney, as a small indie dev, I would look for another engine that just asks me to keep a "Made with Unity" splash screen. (And I'm not saying that license compliance is *easier* with proprietary software - it's just that they don't care and nobody respects the license terms of e.g. FreeType which is included in Unity too. It's mentioned in their EULA, but not in any Unity game.)

Likewise, adding code examples on how to dump a ton of license text in people's credits is IMO not the right approach. It looks terrible, and that's not what most licenses require. _Personally_, I'm perfectly fine with folks linking to https://godotengine.org/license and letting _that page_ do the attribution properly.

Eventually, I also intend to add a way to embed the editor's About dialog in the game, in a way that can be shown easily with e.g. a hotkey (for example Alt+L, or whatever wouldn't conflict with typical OS and in-game hotkeys). Then complying with licenses would be as simple as:

```
Made with Godot Engine
https://godotengine.org
(Press Alt+L for license details.)
```

Calinou:
> Eventually, I also intend to add a way to embed the editor's About dialog in the game, in a way that can be shown easily with e.g. a hotkey (for example Alt+L, or whatever wouldn't conflict with typical OS and in-game hotkeys). Then complying with licenses would be as simple as:

I started working on this a while ago, but didn't have time to complete it: https://github.com/Calinou/godot/tree/add-license-notices-gui

nathanfranke:
There is also github.com/nathanfranke/nicense, written in gdscript so not suitable for the engine, but nevertheless Unlicense so totally free to steal

mhilbrunner:
I'm going to close this for now due to the discussion outlined above. It may still be worthwhile to improve this page, but we should discuss what/how first. Thanks for taking the time with this PR :)

PR #5333, opened 2021-10-18
Clarify the meaning of the `height` property of lights.

nabfrew:
It's easy to assume they are the same, but they are quite different for the two types of 2D lights. For myself, it took a bit of confusion and experimentation for me to figure out why this behaviour changed when I changed from point to directional. Hopefully it can save somebody else the trouble.


Calinou:

PR #5339, opened 2021-10-23
Refactor Memory Management section

### LINKED ISSUE #4877: GDScript "Memory management" should be clearer about Node lifecycle
**Your Godot version:**
3.3-stable

**Issue description:**
See https://github.com/godotengine/godot/issues/44226#issuecomment-741883956, the linked documentation makes it sound like all instances are References unless explicitly inheriting Object, but for newcomers they won't necessarily make the connecting that Nodes inherit Object and that they must be freed manually.

It's a bit tangential to GDScript itself (not sure why it's here, maybe this section should be moved to another page), but it should call out that Nodes, like Object, need to be freed, and that they are freed automatically if they're part of the scene tree, when their parent is freed.

https://docs.godotengine.org/en/stable/getting_started/scripting/gdscript/gdscript_basics.html?#memory-management

zacryol:
Resolves #4877 

Makes it more clear that Node is not freed automatically.
The part that states RefCounted as the default was moved to Inheritance
WeakRef example was altered to show a RefCounted object not prevented from freeing.

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_basics.rst
+++ b/tutorials/scripting/gdscript/gdscript_basics.rst
@@ -1306,6 +1306,10 @@ Inheritance uses the ``extends`` keyword::
     # Inherit/extend an inner class in another file.
     extends "somefile.gd".SomeInnerClass
 
+.. note::
+
+    If inheritance is not explicitly defined, the class will default to inheriting
+    :ref:`class_RefCounted`.
 
 To check if a given instance inherits from a given class,
 the ``is`` keyword can be used::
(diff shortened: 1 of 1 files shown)


tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1526,27 +1530,43 @@ See :ref:`doc_running_code_in_the_editor` for more information.
...
 ~~~~~~~~~~~~~~~~~
 
-If a class inherits from :ref:`class_Reference`, then instances will be
-freed when no longer in use. No garbage collector exists, just
-reference counting. By default, all classes that don't define
-inheritance extend **Reference**. If this is not desired, then a class
-must inherit :ref:`class_Object` manually and must call ``instance.free()``. To
-avoid reference cycles that can't be freed, a :ref:`class_WeakRef` function is
-provided for creating weak references. Here is an example:
+GDScript does not use garbage collection. Instead, the :ref:`class_RefCounted`
```
mhilbrunner, line 1533: Well, Godot does not have a garbage collector, that is right in the original. Godot _does_ have garbage collection though - reference counting _is_ a form of it.
So IMO something like "Godot implements reference counting to free resources that are no longer used, instead of a garbage collector." would be better.

mhilbrunner (commented):
Thanks for contributing! A small note below.

mhilbrunner (approved):

mhilbrunner:
Thank you! :)

PR #5345, opened 2021-10-25
Add a page on making a UI controller/keyboard navigable (Master)

### LINKED ISSUE #4151: Document working with focus neighbors in UI
For the User Interface section.

skyace65:
Adds a page on making a UI that can be navigated with a keyboard and controller by using focus. Closes #4151. Once this is merged I can make another PR for the 3.4. The UI index is different so this can't be easily cherry picked.


skyace65:
I've made most of the suggested changes. I've kept the existing section titles because I don't see the suggested ones as clearer or more accurate.

I've also modified the suggested edit by the beginning about the built in UI actions. I've added a warning about not using the UI actions for gameplay code. And replaced the note with what I had written later, which is just part of the paragraph. I think it's important to mention that these can be modified given the defaults don't use joysticks for whatever reason.

pycbouh:
> I've added a warning about not using the UI actions for gameplay code.

This is not a valid warning, though. It's a consideration, sure, but not completely off the table to state that you should never do it.

> I've kept the existing section titles because I don't see the suggested ones as clearer or more accurate.

"Node configuration" is a more appropriate way to describe working with the Inspector, though I understand how it may feel a superficial change. "Necessary code" isn't a particularly good section title, but I don't have other alternatives.

pycbouh (approved):

mhilbrunner:
Well, I think the warning is fine - we don't recommend it, but sure, you may do so anyway. Thats the case for all warnings in our docs :) Depending on the default editor actions isn't a good idea either way.

mhilbrunner (approved):

mhilbrunner:
Thanks for taking the time integrating the feedback!

skyace65:
3.4 version of this PR: #5355

PR #5353, opened 2021-10-30
Fix description of area_shape_exited() signal

Laelaps9:
area_shape_entered() and area_shape_exited() had the exact same description. Changed area_shape_exited()'s description to specify that the signal is emitted when another area's shape exits rather than enter this area.


mhilbrunner:
Hey! Thanks for contributing. The class docs (everything under `/classes/` are autogenerated from XML files in the main repo: https://github.com/godotengine/godot/tree/master/doc/classes

Any edits to these need to be made over there to the source XML files. :)

PR #5357, opened 2021-10-31
Update console support information

skyace65:
Updates the console support page to say that Godot supports Xbox One, Series X and S via UWP export, as well as the upcoming Steam Deck since it's running Linux.


Calinou:
The UWP export is currently broken, so I would prefer not advertising it until it's fixed (unlikely to happen for 4.0).

skyace65:
I checked the main repository issues, it looks like UWP export kind of works but has major issues. I've removed it from this PR and now only the steam deck is mentioned

tutorials/platform/consoles.rst
```diff
@@ -31,7 +31,8 @@ single console often exceeds $1,000 (this is a rough figure).
 Official support
 ----------------
 
-Godot currently does not officially support consoles. The reasons for this are:
+Godot supports the upcoming Steam Deck which runs Linux. The reason other consoles
```
mhilbrunner, line 34:
```suggestion
Godot supports the Linux-based Steam Deck. The reason other consoles
```

I'd drop `upcoming`, that just means we need to update that page _again_ sometime soon.
I'd also prefer to call it `Linux-based` than `Linux`, because while it _is_ based on (Arch)Linux, SteamOS 3 likely has some custom stuff going on.

mhilbrunner:
I would also like @akien-mga to comment. IMO this is fine, but officially stating we _are_ supporting some platform is a decision that needs to be considered up the chain :) Even if its "just" an ArchLinux.

skyace65:
Fixed.

akien-mga (approved):
Makes sense.

akien-mga:
Thanks!

Calinou:

PR #5359, opened 2021-11-01
Change filesystem dock image - first 2d game

SuPythony:
[Here](https://docs.godotengine.org/en/latest/getting_started/first_2d_game/01.project_setup.html) we are told to extract the `art` and `fonts` directory into the project directory and the file explorer screenshot shows how it is structured. But in the last dock image it is shown that the `art` and `fonts` directories are in the `dodge_assets` directory, which is wrong.


Calinou (changes requested):
The screenshot should be made using an editor scale of 100% and the default font size. The screenshot you committed looks like it's using 75% editor scale (or a font size smaller than the default), so it needs to be remade at the correct size.

SuPythony:
I made it using the inbuilt snipping tool with no changes in settings. What are the required dimensions?

Calinou:
> I made it using the inbuilt snipping tool with no changes in settings. What are the required dimensions?

It's not a matter of dimensions, it's a matter of the editor scale factor used. You need to set the **Display Scale** setting to 100% in the Editor Settings. It's likely being set to 75% automatically on your PC because you use a small display (1366×768).

mhilbrunner:
Hey! Thanks for contributing. As outlined above, that picture needs to be taken at the correct settings.

Also, please make sure to use a single commit with a good commit message where it makes sense (commits can be squashed into one in Git) to keep the commit history nice and clean.

I'm going to close this one for now, feel free to open a new fixed PR or to push fixes to this one and ping us/me to reopen!

PR #5362, opened 2021-11-02
Pin docutils to <0.18 to fix CI due to new warning

mhilbrunner:
This pins `docutils` to versions below `0.18`, which should fix the current CI issues in some branches due to the warning `node class 'meta' is already registered, its visitors will be overridden` (with the CI aborting, as we treat warnings as errors). This is due to an update to `docutils`, which `sphinx` is semingly not (yet) compatible with. According to the `sphinx` Github, there are further incompatabilities besides this warning.
Upstream `sphinx` also pins the version to [<0.18](https://github.com/sphinx-doc/sphinx/blob/4.x/setup.py#L26), with work happening recently in the master branch to support `docutils>=18`.

For reference, see also [this docutils issue](https://sourceforge.net/p/docutils/bugs/431/) and [this sphinx issue](https://github.com/sphinx-doc/sphinx/issues/9777).

We should reconsider this pin when updating `sphinx` in the future.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,6 +5,7 @@
 # https://github.com/readthedocs/readthedocs.org/blob/master/requirements/docs.txt
 sphinx==4.0.2
 sphinx_rtd_theme==0.5.2
+docutils<0.18
 
 # Code tabs extension for GDScript/C#
 sphinx-tabs==3.2.0


mhilbrunner:
Cherrypicked to `3.4` (https://github.com/godotengine/godot-docs/commit/14c0aee54f10e2b158a92e24e5da7a1f3307e0ba), `3.3` (https://github.com/godotengine/godot-docs/commit/29d2abfaf29d7aeace7f13d4b588f849d8f38173), `3.2` (fe90ee5244ff56e036f603a730f04b535438d948) and `stable` (https://github.com/godotengine/godot-docs/commit/a3b2611203f99688afcd3eb6916abf53aba0f384).

PR #5368, opened 2021-11-04
Adding Using InputEvent VisualScript snippets

TaeruAlethea:
Created and Included VisualScript snippets.


Calinou:
We currently don't have a defined contribution workflow for visual scripting snippets. I would prefer we define a workflow before accepting a pull request, so that different people can submit snippets that look like they were designed by a single person.

Also, file names should be `snake_case` (only lowercase characters). You also don't need the `Snippits` part (which doesn't match the actual file path).

TaeruAlethea:
Snake_case file names fixed.
Corrected the file names in inputevent.rst

TaeruAlethea:
> We currently don't have a defined contribution workflow for visual scripting snippets. I would prefer we define a VisualScript code sample workflow before merging this pull request, so that different people can submit snippets that look like they were designed by a single person.

@Calinou,
Considering the _overall lack of any_ VScript snippets, I think that having them, even if a bit inconsistent, would be better than non at all.

mhilbrunner:
Thanks for contributing!

We really need to figure out how we're going to maintain VisualScript docs first.
Maintaining a lot of screenhots/images is hard for editor screenshots already and we're struggling to keep those updated, I fear doing the same for VisualScript may be not feasible.

Please don't spend more time on this until we've figured this out.

PR #5371, opened 2021-11-05
Empty method description in ConfigFile

comminux:
Fixes an empty method description in ConfigFile

### DIFF
--- a/classes/class_configfile.rst
+++ b/classes/class_configfile.rst
@@ -90,6 +90,8 @@ Method Descriptions
 
 - void **clear** **(** **)**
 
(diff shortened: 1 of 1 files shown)


pycbouh:
Thanks for your contribution! The class reference is tracked on the main repo [here](https://github.com/godotengine/godot/tree/master/doc/classes). See https://docs.godotengine.org/en/stable/community/contributing/updating_the_class_reference.html.

Keep in mind as well, that the commit message (and the PR title) should explain what you did to fix the issue, not state the issue itself (e.g. "Add a description to clear method of ConfigFile")

PR #5380, opened 2021-11-08
Updated tools to @tools

twarped:
Godot 4.0 complains that tools is an "Unexpected Identifier" so you need to put an @ in front of it, for example @export


Calinou:
There are a lot more occurrences to change, so I'd recommend changing them all in the same pull request: 

image

akien-mga:
I agree, this should be redone to replace all `tool` to `@tool` with a regex.

PR #5386, opened 2021-11-11
Fix missing word in When and how to avoid using nodes for everything

srivastavaarpit121:
Did a minor grammatical correction.


Calinou (approved):

PR #5388, opened 2021-11-11
Typo for loop invariant, i/index, in squash check

berdon:


mhilbrunner (approved):

mhilbrunner:
Well spotted! Thanks for submitting a fix. :)

This likely can't be cherrypicked directly, but we should do so manually as it could be confusing for new users copy/pasting code.

Calinou:
The first 3D game tutorial isn't available in the `3.4` branch, so this can't be cherry-picked.

PR #5395, opened 2021-11-15
Will have to update the y axis not the x

lostpirate:
Earlier version was updating the x axis and it should update y, when moving back and forth.


aaronfranke:
In 2D, the forward direction is typically +X. The current code is consistent with this, and with the GIF in the documentation:

PR #5399, opened 2021-11-16
Fix Record and Capture effect short descriptions in Audio buses

RedMser:
While the individual pages for these effects describe them accurately,
the short descriptions on the audio bus overview page were misleading.


Calinou (approved):

Cherry-picked to the `3.4`, `stable` and `3.3` branches.

PR #5423, opened 2021-11-30
specify correct emscripten version for godot 3.4

dbu:
we need https://github.com/emscripten-core/emscripten/pull/13776 available

i hope this is the right branch to put this to. i assume it also applies to godot 4.

it would be nice if scons would be able to check the emscripten version and complain when it is too low, that would have saved me some time searching for my problem.

### DIFF
--- a/development/compiling/compiling_for_web.rst
+++ b/development/compiling/compiling_for_web.rst
@@ -8,9 +8,9 @@ Compiling for the Web
 Requirements
 ------------
 
-To compile export templates for the Web, the following is required:
+To compile Godot 3.4 export templates for the Web, the following is required:
 
--  `Emscripten 1.39.9+ <https://emscripten.org>`__.
+-  `Emscripten 2.0.17+ <https://emscripten.org>`__.
 -  `Python 3.5+ <https://www.python.org/>`__.
 -  `SCons 3.0+ <https://www.scons.org>`__ build system.
 


development/compiling/compiling_for_web.rst
```diff
@@ -8,9 +8,9 @@ Compiling for the Web
 Requirements
 ------------
 
-To compile export templates for the Web, the following is required:
+To compile Godot 3.4 export templates for the Web, the following is required:
```
dbu, line 11: not too sure, but probably makes sense to give the context here for the required versions.

pycbouh: No need, this is already in Godot 3.4 docs. If this becomes irrelevant in a future 3.x version, then we’ll remove it in a future version.

pycbouh:
> i assume it also applies to godot 4

In that case you need to make a PR against the master branch. I guess you can keep this one and make a new one for master. But in general you need to start with the latest version, and then the changes can be backported

Faless:
The hard requirement of that function was a mistake, and is fixed via: https://github.com/godotengine/godot/pull/55354 .
`1.39.9` is still the minimal supported emscripten version since it's the version required by mono builds.
Only the thread and gdnative builds require more recent versions.

dbu:
ah, bad luck that i tried just a day too early, then. thanks for the fix.

PR #5431, opened 2021-12-06
class_String_method_rsplit correction

meloonics:
In the GDScript example it says it prints the following:
"2"
"Four"
"Three,Two,One"
Which confused me first and isn't true either.


Calinou:

PR #5447, opened 2021-12-12
Update class_label.rst for property visible_characters

belzecue:
Clarify that space characters do not count toward the visible count.  This makes sense but it's not obvious.  In general programming, a space character is never ignored in a string, and a user might easily assume at first (as I did) that the string "Hello World! the musical" with visible_characters set to 12 would produce "Hello World!" when in fact it would produce "Hello World! t".


Calinou:

PR #5449, opened 2021-12-13
Clarify tree order in Using SceneTree

### LINKED ISSUE #4252: Tree Order - Getting started
https://docs.godotengine.org/en/stable/getting_started/step_by_step/scene_tree.html

A few related things:

> Tree order
> 
> Most node operations in Godot, such as drawing 2D, processing, or getting notifications are done in tree order. This means that parents and siblings with a lower rank in the tree order will get notified before the current node.

1) What does this mean? Especially the last sentence. The screenshot below that suggests that things are run top to bottom, but that doesn't seem to be what this paragraph is saying (although, I'm not actually sure what this paragraph is saying as "lower rank" could mean a number of things due to there being various possible ways that "ranking" could happen). Also, what's the "current node" in this context?

2) "notified" = notifications? In which case I'll point to: https://github.com/godotengine/godot-docs/issues/4232 - notifications aren't explained anywhere that I can find, and the doc that does cover in getting-started them basically says "you don't need to know about them" - which is fine, but then you can't use them later). Should use "Signals" instead (That's assuming they're synonyms at this level of abstraction).

The next subsection '"Becoming active" by entering the Scene Tree' does talk about notifications, but it puts them into the context of the GDScript functions they relate to, so it's fine.

skyace65:
Clarifies how tree order works. Closes #4252, Once this is merged I'll make a 3.x branch version since this information is on a different page in that branch.

### DIFF
--- a/tutorials/scripting/scene_tree.rst
+++ b/tutorials/scripting/scene_tree.rst
@@ -117,7 +117,7 @@ in top to bottom order. So the top child node of the top node will get its
 
 This can also be overridden using the ``process_priority`` node property.
 Nodes with a lower number are called first. For example, nodes with the
-priorities "0,1,2,3" would be called in that order.
+priorities "0, 1, 2, 3" would be called in that order (from left to right).
 
 "Becoming active" by entering the *Scene Tree*
 ----------------------------------------------
--- a/tutorials/scripting/scene_tree.rst
+++ b/tutorials/scripting/scene_tree.rst
@@ -106,11 +106,18 @@ Tree order
 ----------
 
 Most node operations in Godot, such as drawing 2D, processing, or getting
-notifications are done in tree order. This means that parents and
-siblings with a lower rank in the tree order will get notified before
-the current node.
+notifications are done in tree order, or top to bottom. For example, the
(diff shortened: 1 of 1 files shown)


Calinou (commented):
Node's `process_priority` property should probably be mentioned here as well :slightly_smiling_face: 

On existing projects, it can override the processing order, so it's important that people know about it.

skyace65:
Added an explanation of that to the documentation.

tutorials/scripting/scene_tree.rst
```diff
@@ -106,12 +106,19 @@ Tree order
...
+node below it has its function called, then the node below that and so
+on. However, children of a node will get called before their parent, also
+in top to bottom order. So the top child node of the top node will get its
+``_ready()`` function called first.
 
 .. image:: img/toptobottom.png
 
+This can also be overridden using the ``process_priority`` node property.
+Nodes with a lower number are called first. For example, nodes with the
+priorities "0,1,2,3" would be called in that order.
```
Calinou, line 120:
```suggestion
priorities "0, 1, 2, 3" would be called in that order (from left to right).
```

Calinou (approved):
Thanks!

PR #5452, opened 2021-12-16
Remove broken link to opengameart

### LINKED ISSUE #5258: Tutorial link to OpenGameArt.org page is landing on access denied page
**Issue description:**
'https://opengameart.org/users/tgfcoder' link on 'https://docs.godotengine.org/en/stable/tutorials/2d/2d_sprite_animation.html' page is landing on an 'Access Denied' page. Assume it's not just me that is getting access denied.

https://github.com/godotengine/godot-docs/blob/master/tutorials/2d/2d_sprite_animation.rst - about the 3rd paragraph in the Note section.

Not sure what artwork we were linking to so can't recommend a fix.

skyace65:
Closes #5258. The person who made the art has since deleted their open game art account. I don't know what license the original art was released as, but as we're still giving credit we should be complying with it whatever it was.

### DIFF
--- a/tutorials/2d/2d_sprite_animation.rst
+++ b/tutorials/2d/2d_sprite_animation.rst
@@ -16,8 +16,7 @@ animate a collection of individual images. Then we will animate a sprite sheet u
 with :ref:`AnimationPlayer <class_AnimationPlayer>` and the *Animation*
 property of :ref:`Sprite2D <class_Sprite2D>`.
 
-.. note:: Art for the following examples by https://opengameart.org/users/ansimuz and by
-                                           https://opengameart.org/users/tgfcoder
+.. note:: Art for the following examples by https://opengameart.org/users/ansimuz and tgfcoder.
 
 Individual images with AnimatedSprite2D
 ---------------------------------------


mhilbrunner (dismissed):

mhilbrunner:
I guess we should eventually redo the tutorial.

akien-mga (approved):

akien-mga:
Well I hadn't seen Max's comment but there are good points. Would be good to identify which assets are from this artist and what the exact license was.

mhilbrunner:
So seems to be the frog image, and the license is CC0 (https://creativecommons.org/publicdomain/zero/1.0/).
Seems fine then, wouldn't even need attribution.

In that case, I'd probably remove the name and the URL, as it seems they want that page removed?

akien-mga:
I would either keep the name, or replace the asset. We can't really presume of their intention with the account deletion, but even if CC0 allows not giving credit, I prefer we always do it anyway - unless specifically required by the author.

mhilbrunner:
I agree.

Then the change is fine as is. :) We still may want to use different assets eventually.

PR #5458, opened 2021-12-21
conf.py: Add support for loading localized class reference

akien-mga:
It will look up a potential `classes/<lang>` folder in the parent repository
of https://github.com/godotengine/godot-docs-l10n and use it if provided.

This folder includes a pre-translated copy of the class reference XML, read
to be used by Sphinx. This turned out to be a simpler approach than generating
PO files for class reference strings in reStructured Text format, given that
their source format is XML and the conversion on a per `msgid` basis would be
quite tricky.

This approach is very hacky, as Sphinx doesn't seem to support any kind of
conditional statements in its `toctree` directives, so we have to outright
replace the content at build time. This should typically only be done on RTD
so it shouldn't end up polluting the Git history.

image

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -269,6 +269,19 @@ def godot_get_image_filename_for_language(filename, env):
 
 sphinx.util.i18n.get_image_filename_for_language = godot_get_image_filename_for_language
 
+# Similar story for the localized class reference, it's out of tree and there doesn't
+# seem to be an easy way for us to tweak the toctree to take this into account.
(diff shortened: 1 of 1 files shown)


akien-mga:
This goes together with https://github.com/godotengine/godot-docs-l10n/pull/19 which will add localized class reference in `zh_CN` and `es` to `godot-docs-l10n`.

akien-mga:
Cherry-picked for 3.4.

PR #5460, opened 2021-12-22
Replace inactive echo service in websocket.rst

snailrhymer:
The recommended websocket echo service (echo.websocket.org) is no longer available, so the minimal example given will not work. I've proposed the first available one I found (piesocket.com/websocket-tester), but it may not be ideal - while it does echo messages sent to it, it will also send other messages to connected clients.

Whether or not this new link is acceptable, the broken link needs replacing.

### DIFF
--- a/tutorials/networking/websocket.rst
+++ b/tutorials/networking/websocket.rst
@@ -37,7 +37,7 @@ This example will show you how to create a WebSocket connection to a remote serv
     extends Node
 
     # The URL we will connect to
-    export var websocket_url = "ws://echo.websocket.org"
+    export var websocket_url = "ws://websocket-echo-service.glitch.me/"
 
     # Our WebSocketClient instance
     var _client = WebSocketClient.new()
@@ -52,8 +52,13 @@ This example will show you how to create a WebSocket connection to a remote serv
         # Alternatively, you could check get_peer(1).get_available_packets() in a loop.
         _client.connect("data_received", self, "_on_data")
 
+        # Set up custom headers needed by this particular echo service
+        var custom_headers = PoolStringArray([
+            "Connection: keep-alive",
+            "User-Agent: Websocket Demo",
+	      ])
         # Initiate connection to the given URL.
-        var err = _client.connect_to_url(websocket_url)
+        var err = _client.connect_to_url(websocket_url, [], false, custom_headers)
         if err != OK:
             print("Unable to connect")
             set_process(false)


tutorials/networking/websocket.rst
```diff
@@ -37,7 +37,7 @@ This example will show you how to create a WebSocket connection to a remote serv
     extends Node
 
     # The URL we will connect to
-    export var websocket_url = "ws://echo.websocket.org"
+    export var websocket_url = "ws://www.piesocket.com/websocket-tester"
```
Faless, line 40: Did you try if it works? The websocket connection URL is supposed to be:
`wss://demo.piesocket.com/v3/channel_1`, plus, it seems it requires an API key?
I think we could alternatively try: https://libwebsockets.org/testserver/ which is a server implementation using the previous library we were using (`libwebsockets`) (again, we should double check the connect URL).

snailrhymer: Ah yes, sorry. I'd forgotten that I copied the address from [https://www.piesocket.com/websocket-tester](https://www.piesocket.com/websocket-tester), so the functioning websocket url for this service would be

`wss://demo.piesocket.com/v3/channel_1?api_key=oCdCMcMPQpbvNjUIzqtvF1d2X2okWpDQj4AwARJuAgtjhzKxVEjQU6IdCjwm&notify_self`

The libwebsockets alternative looks a lot cleaner, but I haven't been able to find a URL for it; what did you use?

snailrhymer:
Oops, didn't realise renaming the branch would effectively delete it and close this request.

snailrhymer:
I've set up an echo service at ws://websocket-echo-service.glitch.me (which can also handle wss connections).

Pros: 
* It exists and works.

Cons:
* Due to some particularities of hosting on glitch, it needs a couple of custom headers to work properly. This doesn't seem ideal for a minimal example (but I've updated the tutorial accordingly).
* It's hosted for free and has some limitations on monthly traffic, but given how long the docs have survived with an in active link, this shouldn't be too problematic.
* I need to not accidentally delete it.

***

If it can be better hosted elsewhere, the code in its entirety is

```
console.log("Server starting...")

const express = require("express");
const app = express();
const http = require('http');
const server = http.createServer(app);
server.listen(3000);

// Respond to http get requests with a simple message
app.get("/", (req, res) => res.send('Connect to ws://websocket-echo-service.glitch.me and send a message!'));

const WebSocket = require('ws');
const wss = new WebSocket.Server({server:server});

wss.on("connection", wss => {
  wss.on("message", message => {
     // Echo back
    wss.send(message)
  });
});
console.log("Server ready")
```

snailrhymer:
Alternatively, wss://libwebsockets.org with dumb-increment-protocol and no custom headers does work, but it provides an increasing sequence of integers instead of echoingmessages. I think echo functionality that lets the user check both sending and receiving data is preferable.

PR #5472, opened 2021-12-26
Add a warning about creating threads at runtime

### LINKED ISSUE #5310: Update Using multiple threads to mention using semaphores to avoid creating threads at run-time
**Your Godot version:** 3.3.4

**Issue description:**

Creating threads at run-time is slow on Windows, so it should be avoided to prevent stuttering. See <https://godotforums.org/discussion/comment/54027/#Comment_54027>.

**URL to the documentation page (if already existing):** https://docs.godotengine.org/en/latest/tutorials/performance/using_multiple_threads.html

skyace65:
Adds a warning to the using multiple threads page about creating threads at runtime. Doing that can cause stuttering on windows and can be avoided with semaphores. Closes #5310


mhilbrunner (approved):

mhilbrunner:
Thanks!

PR #5476, opened 2021-12-27
Improve explenation for Trigger and Capture update modes

### LINKED ISSUE #5127: AnimationPlayer: Better explain "Trigger" and "Capture" Update Modes
**Your Godot version:** 3.3.2 stable, 3.4 beta2

**Issue description:**
I have been using Godot and the AnimationPlayer now for almost three years almost daily and I still don't know what "Trigger" and "Capture" update modes are useful for, or what to do with it, how to test this functionality. I was hoping I would develop an intuitive understanding over time, or come across some 3rd party tutorial explaining it, but it seems like this day will never come.

The [AnimationPlayer Documentation](https://docs.godotengine.org/en/3.4/tutorials/animation/introduction.html#id15) is very sparse on this topic to say the least:

> Trigger: Only update the property on keyframes or triggers

What are "triggers"? This is the only place in the I've heard about this and it's not explained any further. I have no clue what so ever what the difference to "Descrete" mode could be.

> Capture: The current value of a property is remembered, and it will blend with the first animation key found

This sounds like it could be useful, but I have no idea what that means in practice and how it would differ to the "Continous" update mode exactly.

The [API docs of the Animation Resource](https://docs.godotengine.org/en/3.4/classes/class_animation.html#class-animation) unfortunately is not helping either:

> UPDATE_DISCRETE = 1 --- Update at the keyframes and hold the value.
> UPDATE_TRIGGER = 2 --- Update at the keyframes.

Well I know keyframes exist to update property values ... But if Trigger does not hold and not interpolate the value like "discrete" or "continuous" would do, what does it do?

> UPDATE_CAPTURE = 3 --- Same as linear interpolation, but also interpolates from the current value (i.e. dynamically at runtime) if the first key isn't at 0 seconds.

This is more explicit and helpful than the explanation in the AnimationPlayer tutorial, and maybe enough for an API reference, but after reading this I'm still not really sure how to apply "Capture" update mode, how to test it, make use of it. After reading this (which is hard to find for a beginner), I would go to the AnimationPlayer tutorial page, expecting a more detailed explanation, but the opposite is the case.

https://docs.godotengine.org/en/3.4/tutorials/animation/introduction.html#id15

under UPDATE MODE constants:
https://docs.godotengine.org/en/3.4/classes/class_animation.html#class-animation

skyace65:
Gives a clearer explanation for how Capture works and a possible use case, explains what exactly a trigger is and when it's used. Closes #5127.

### DIFF
--- a/tutorials/animation/introduction.rst
+++ b/tutorials/animation/introduction.rst
@@ -231,13 +231,13 @@ mode, track interpolation, and loop mode.
 The update mode of a track tells Godot when to update the property
 values. This can be:
 
--  Continuous: Update the property on each frame
--  Discrete: Only update the property on keyframes
--  Trigger: Only update the property on keyframes or triggers.
+-  **Continuous:** Update the property on each frame
+-  **Discrete:** Only update the property on keyframes
+-  **Trigger:** Only update the property on keyframes or triggers.
    Triggers are a type of keyframe used by the
    ``current_animation`` property of a :ref:`class_AnimationPlayer`,
    and Animation Playback tracks.
--  Capture: if the first keyframe's time is greater than ``0.0``, the
+-  **Capture:** if the first keyframe's time is greater than ``0.0``, the
    current value of the property will be remembered and
    will be blended with the first animation key. For example, you
    could use the Capture mode to move a node that's located anywhere
--- a/tutorials/animation/introduction.rst
+++ b/tutorials/animation/introduction.rst
@@ -233,9 +233,14 @@ values. This can be:
 
 -  Continuous: Update the property on each frame
 -  Discrete: Only update the property on keyframes
--  Trigger: Only update the property on keyframes or triggers
--  Capture: Remember the current value of the property, and blend it with the
-   first animation key
+-  Trigger: Only update the property on keyframes or triggers.
+   Triggers are a type of keyframe used by the
+   ``current_animation`` property of a :ref:`class_AnimationPlayer`,
+   and Animation Playback tracks.
+-  Capture: if the first keyframes time is greater than 0 the
+   current value of the property will be remembered, and it
+   will blend with the first animation key. For example, you
+   could move a node that's anywhere to a specific location.
 
 .. figure:: img/animation_track_rate.png
    :alt: Track mode
--- a/tutorials/animation/introduction.rst
+++ b/tutorials/animation/introduction.rst
@@ -237,10 +237,11 @@ values. This can be:
    Triggers are a type of keyframe used by the
    ``current_animation`` property of a :ref:`class_AnimationPlayer`,
    and Animation Playback tracks.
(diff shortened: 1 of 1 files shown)


tutorials/animation/introduction.rst
```diff
@@ -233,9 +233,14 @@ values. This can be:
...
--  Capture: Remember the current value of the property, and blend it with the
-   first animation key
+-  Trigger: Only update the property on keyframes or triggers.
+   Triggers are a type of keyframe used by the
+   ``current_animation`` property of a :ref:`class_AnimationPlayer`,
+   and Animation Playback tracks.
+-  Capture: if the first keyframes time is greater than 0 the
+   current value of the property will be remembered, and it
+   will blend with the first animation key. For example, you
+   could move a node that's anywhere to a specific location.
```
Calinou, line 243:
```suggestion
-  Capture: if the first keyframe's time is greater than ``0.0``, the
   current value of the property will be remembered and
   will be blended with the first animation key. For example, you
   could move a node that's located anywhere to a specific location.
```

tutorials/animation/introduction.rst
```diff
@@ -233,9 +233,15 @@ values. This can be:
 
 -  Continuous: Update the property on each frame
 -  Discrete: Only update the property on keyframes
```
Calinou, line 235:
```suggestion
-  **Continuous:** Update the property on each frame
-  **Discrete:** Only update the property on keyframes
```

tutorials/animation/introduction.rst
```diff
@@ -233,9 +233,15 @@ values. This can be:
...
-   first animation key
+-  Trigger: Only update the property on keyframes or triggers.
+   Triggers are a type of keyframe used by the
+   ``current_animation`` property of a :ref:`class_AnimationPlayer`,
+   and Animation Playback tracks.
+-  Capture: if the first keyframe's time is greater than ``0.0``, the
+   current value of the property will be remembered and
+   will be blended with the first animation key. For example, you
+   could use the Capture mode to move a node that's located anywhere
+   to a specific location.
```
Calinou, line 244:
```suggestion
-  **Trigger:** Only update the property on keyframes or triggers.
   Triggers are a type of keyframe used by the
   ``current_animation`` property of a :ref:`class_AnimationPlayer`,
   and Animation Playback tracks.
-  **Capture:** if the first keyframe's time is greater than ``0.0``, the
   current value of the property will be remembered and
   will be blended with the first animation key. For example, you
   could use the Capture mode to move a node that's located anywhere
   to a specific location.
```

Calinou (approved):
Thanks!

golddotasksquestions:
Thank you! This helps a lot!

akien-mga:
Cherry-picked for 3.4.

PR #5478, opened 2021-12-27
Fix references to window size project settings in Multiple resolutions

srivastavaarpit121:
Changes done in Multiple Resolutions -> Stretch Settings -> Stretch Mode -> Stretch Mode = 2D, section of the documentation.

The first line refers to base width and height as - ```display/width``` and ```display/height```. So basically, it's trying to refer to them by an address. But when tested with the following Godot releases -
- Godot v3.4.1
- Godot v3.4.2

<br>

The address was this instead -

- ```display/window/size/width```
width

<br>

- ```display/window/size/height```
height

<br>

So the address needs to be corrected. But since the corrected one is quite long, I instead renamed it from -
- ```the size specified in display/width and display/height``` -> to -> ```the base size specified in width and height```

So we are no longer using the obsolete address, and I think it's much easier to understand now.

### DIFF
--- a/tutorials/rendering/multiple_resolutions.rst
+++ b/tutorials/rendering/multiple_resolutions.rst
@@ -135,8 +135,8 @@ demonstrate the effect of different stretch modes. A single sprite, also
 
    .. image:: img/stretch_disabled_expand.gif
 
--  **Stretch Mode = 2D**: In this mode, the size specified in
-   display/width and display/height in the project settings is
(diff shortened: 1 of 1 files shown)


srivastavaarpit121:
What's the status of this pull request?

Calinou (approved):
Thanks!

akien-mga:
Cherry-picked for 3.4.

PR #5483, opened 2021-12-29
Remove references to 'godot-headers' from gdnative cpp example

Ivorforce:
As I understand, this folder is not needed anymore as gdnative-cpp now contains the headers as a submodule.


akien-mga:
Thanks!

PR #5488, opened 2021-12-31
Document how to test translations with the test project setting

### LINKED ISSUE #4471: Mention `Project Settings > Locale > Test` in `Internationalizing games` introduction documentation
**Your Godot version:**
3.2.3

**Issue description:**
I was looking into translations/localization of my project, and the page describing all the basics only lists one way of testing non-standard locales: using the `Command Prompt` and running it through that. Didn't seem improbable to me, but after struggling to get it working I googled further, and found out that the `Project Settings` have a `locale > Test` variable which will do the same.

image

This should really be mentioned on the page, as it is the simplest way to test different locales, which is something any developer working on translations will have to do.

https://docs.godotengine.org/en/stable/tutorials/i18n/internationalizing_games.html

skyace65:
Does what the title says. The image I added has gone through png compression. Closes #4471


tutorials/i18n/internationalizing_games.rst
```diff
@@ -202,10 +202,22 @@ Icons with left and right pointing arrows which may need to be reversed for Arab
...
+do this.
+
+First, in the project settings under **Input Devices > Locale** there is a ``Test`` property.
+Set it to the locale that you want to test and Godot will run the project with that locale
+from the editor.
+
+..image:: img/locale_test.png
+
+Keep in mind that as this is a project setting it will show up in version control, and it
+should be set back to nothing before committing changes.
```
Calinou, line 218:
```suggestion
Testing translations
--------------------

You may want to test a project's translation before releasing it. Godot provides two ways to
do this.

First, in the Project Settings, under **Input Devices > Locale**, there is a **Test** property. Set this property to the locale code of the language you want to test. Godot will run the project with that locale when the project is run (either from the editor or when exported).

..image:: img/locale_test.png

Keep in mind that since this is a project setting, it will show up in version control when it is set to a non-empty value. Therefore, it should be set back to an empty value before committing changes to version control.
```

akien-mga (approved):

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 3.4.

PR #5489, opened 2021-12-31
Add a page on managing editor features

### LINKED ISSUE #5065: Add a page on managing editor features
**Your Godot version:**
Any.

**Issue description:**
Godot editor can be customized to disable parts of its interfaces and features (and even entire hierarchies of classes) to make it more suited to a particular project, but we don't have an article explaining how to do that. A good example of this that many developers likely face is disabling any 3D related functionality when working exclusively on 2D projects.

Such an article would fit nicely into the **Editor manual** section.

### LINKED ISSUE #5114: Document distributing the Godot editor with a custom feature profile
**Your Godot version:** 3.3

**Issue description:**
A common use case for [EditorFeatureProfile](https://docs.godotengine.org/en/stable/classes/class_editorfeatureprofile.html) is allowing educators to reduce the mental load on their students by providing an instance of the engine with reduced features.

The best way to achieve this, currently, is by distributing Godot in self-contained mode, as described in [the related paragraph on the Data Paths tutorial](https://docs.godotengine.org/en/stable/tutorials/io/data_paths.html#doc-data-paths-self-contained-mode), and setting up the intended FeatureProfile beforehand. However, _this usage is not suggested on the EditorFeatureProfile class reference, nor are there are any tutorials describing this use case_.

**Proposed solution:**
I'm willing to create a pull request improving the documentation in that regard, but first I'd like to understand what would be the best approach:
- Adding a new paragraph to the EditorFeatureProfile description, describing this use case and pointing to the relevant documentation (the paragraph linked above). however this would imply linking to the Godot Docs website, which usually occurs only in the "Tutorials" section.
- Adding a link to the Data Paths tutorial in the "Tutorials" section of the EditorFeatureProfile reference. I don't think this would solve the issue, because there would be no context to why it's being linked, i.e. an explanation of the use case mentioned above.

The actual best solution, I believe, would be to write a full tutorial on how to set up this use case, which could then be linked under "Tutorials" from the EditorFeatureProfile. However, I neither have the time nor the qualification to do that right now.

P.S.: I do realize that changes to the class reference should be made to the [other repository](https://github.com/godotengine/godot), however I believe this is a better place to discuss this specific issue.

[EditorFeatureProfile class reference](https://docs.godotengine.org/en/stable/classes/class_editorfeatureprofile.html)
[Data Paths tutorial](https://docs.godotengine.org/en/stable/tutorials/io/data_paths.html)

skyace65:
Does what the title says, I've also copied and slightly modified the instructions on how to enable self contained mode for the editor from the data paths tutorial. Both new images have gone through png compression. Closes #5065 closes #5114.
Edit: Fixed the build issue


pycbouh:
Educators are an interesting angle, but I'd try to incorporate something more game development related. Like working in a team with a UI designer who doesn't need 3d nodes even in a 3d project, or vice versa, environmentalist who doesn't need UI.

It's also important to mention that this feature only hides related objects, properties, and tools, but you can still open a project that relies on them without breaking it. And thus we should also warn that this is not an optimization technic, you don't actually strip those parts from your project (and link to the optimization docs).

skyace65:
Fixed. is there a way to link to an index page with `:ref:` or would it have to be a https:// link?

pycbouh:
> Fixed. is there a way to link to an index page with `:ref:` or would it have to be a https:// link?

You need to give it a TOC identifier, then you can link to it. If you mean https://github.com/godotengine/godot-docs/blob/master/tutorials/performance/index.rst, then add something like `.. _doc_performance:` as the first line.

skyace65:
Done.

tutorials/editor/index.rst
```diff
@@ -48,3 +48,16 @@ Emacs.
 
    command_line_tutorial
    external_editor
+
+Managing Editor Features
```
akien-mga, line 52:
```suggestion
Managing editor features
```

tutorials/editor/managing_editor_features.rst
```diff
@@ -0,0 +1,55 @@
+.. _doc_managing_editor_features:
+
+Managing Editor Features
```
akien-mga, line 3:
```suggestion
Managing editor features
```

akien-mga:
Thanks!

PR #5490, opened 2022-01-01
List advanced GUI Objects (Master)

### LINKED ISSUE #5334: List advanced gui objects
**Your Godot version:**
3.3.3
**Issue description:**
It would be very helpfull to know it when building from source, here's what I have:

## ADVANCED_GUI_NODES (as seen in scene/register_scene_types)

- FileDialog
- PopupMenu
- Tree
- TextEdit
- OptionButton
- SpinBox
- ColorPicker
- ColorPickerButton
- RichTextLabel
- PopupDialog
- WindowDialog
- AcceptDialog
- ConfirmationDialog
- MarginContainer
- ViewportContainer
- SplitContainer
- HSplitContainer (inherits from SplitContainer)
- VSplitContainer (inherits from SplitContainer)
- GraphNode
- GraphEdit

##  Resources (all of them are linked with a node so maybe it could be abstracted?)
- TreeItem (needs Tree)
- RichTextEffect (needs RichTextLabel)
- CharFXTransform (needs RichTextLabel)

https://docs.godotengine.org/en/stable/development/compiling/optimizing_for_size.html#disabling-advanced-gui-nodes

skyace65:
Lists all the advanced GUI objects that will be disabled when building the engine with the disable advanced GUI flag. Closes #5334.

### DIFF
--- a/development/compiling/optimizing_for_size.rst
+++ b/development/compiling/optimizing_for_size.rst
@@ -106,8 +106,8 @@ Tools must be disabled in order to use this flag, as the editor is not designed
 to operate without 3D support. Without it, the binary size can be reduced
 by about 15%.
 
-Disabling advanced GUI nodes
-----------------------------
+Disabling advanced GUI objects
+------------------------------
 
 - **Space savings:** Moderate
 - **Difficulty:** Easy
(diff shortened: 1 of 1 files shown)


mhilbrunner:
I have to say I'm kinda hesistant on this one - is this something we foresee changing a lot, @pycbouh @KoBeWi?
Keeping lists of classes up to date with fast changing code isn't easy for the docs (if its not done automatically, or semi-automatically, like with class docs)

If this doesn't change often, we can merge this. If not, I'm not sure if the maintenance burden vs usefulness tradeoff is worth it.

KoBeWi:
The list only changes when we add some new GUI node, which isn't very often. But yeah, if we list it explicitly, we'd need to remember to keep it updated.

The list is defined here:
https://github.com/godotengine/godot/blob/master/scene/register_scene_types.cpp#L365
Maybe it's ok to just paste this link into the doc? But I guess the code might move around anyways 🤔

pycbouh:
Indeed, this is not a volatile information, we can list it. It is a consideration for maintenance, but not more than anything else that we put in the doc.

mhilbrunner (approved):

mhilbrunner:
Thank you! Merged :)

PR #5493, opened 2022-01-02
Document how export filters work

### LINKED ISSUE #5272: Add Exporting FAQ/Troubleshooting guide
**Your Godot version:**
3.2.1
**Issue description:**
I would like to propose an exporting FAQ/troubleshooting guide.

There is a wealth of information already available in the documentation, but the relevance of it to exporting is not always immediately obvious. A user running into errors might not think to look in the filesystems page, for example, although it contains critical information to the exporting process (in this case, the fact that res:// becomes read only in a build, where it is read/write in the editor).

A quick search through the community Discord revealed many similar questions from the past several months. I propose the following export-related details be specifically clarified on the Exporting page:

1. .json, .txt, and .csv files are blocked from export by default and need to be manually added back in with an export filter (are there any other common file types I'm missing there?)
2. res:// becomes read-only in a build. So not only will calls to write to a file in res:// not work, but simply opening a file with read/write access will crash/softlock/otherwise break the project, even if that write access is not used.

I believe having these and other future details easily accessible in the documentation, either within an existing page or as a page of its own, will decrease confusion and reduce questions in the Discord and other support channels.

https://docs.godotengine.org/en/stable/getting_started/step_by_step/exporting.html#
https://docs.godotengine.org/en/stable/getting_started/step_by_step/filesystem.html#resource-path

skyace65:
Documents how export filters work, and that one needs to be set to allow non resource files to be exported. I would argue this fixes #5272. That issue is about two things people don't know, one is not to access res:// directly. Step by step explains that this shouldn't be done, the [file class reference page](https://docs.godotengine.org/en/latest/classes/class_file.html) explains that this shouldn't be done. I don't think a FAQ section is going to stop anyone. The second issue was that non resource files aren't exported by default, that information wasn't documented anywhere, but now it is. If there's other frequently asked questions that should be documented I'll gladly do it. But I don't think it's necessary after this PR.

### DIFF
--- a/tutorials/export/exporting_projects.rst
+++ b/tutorials/export/exporting_projects.rst
@@ -103,8 +103,8 @@ option in the editor:
 
 .. _doc_exporting_projects_export_mode:
 
-Export mode
-~~~~~~~~~~~
+Resource options
+~~~~~~~~~~~~~~~~
 
 When exporting, Godot makes a list of all the files to export and then
 creates the package. There are 3 different modes for exporting:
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):

mhilbrunner:
Thanks!

akien-mga:
Cherry-picked for 3.4.

PR #5503, opened 2022-01-09
Fix normal map description and improve note

### LINKED ISSUE #4192: Normalmap convention seems wrong or not explicit enough
Godot 3.2.3 and latest at time of posting

There are two copies of an explanation about which normalmap convention Godot is expecting:
https://docs.godotengine.org/en/stable/tutorials/3d/spatial_material.html#normal-map
https://docs.godotengine.org/en/stable/getting_started/workflow/assets/importing_images.html#normal-map

It is said normal maps should be X+, Y-, Z+. However, this is what an OpenGL-style normal map looks like:

image
(Tested in Godot and it checks out)

Other image found by Arnklit on Discord:
image

As you can see, the bright-green area is on top of the image, where normals tend towards the top side, which suggests Y is UP. Which I thought was `Y+`, like this forum post also suggests: https://polycount.com/discussion/196999/difference-betwen-opengl-and-directx

So either the doc is wrong and should say `Y+` (since OpenGL coordinates start from bottom left), or it is actually implying another coordinate system which is top-left Y-down, in which case yes, `Y-` would be up. But this is not explained, so the coordinate precision still doesn't help IMO. At best, it only makes users aware that they *might* have to flip their normalmaps, but it doesnt teach them how to actually recognize one directly.
There should be an example like the second image I posted, explaining the visual difference, to know when the normalmap is correct.

skyace65:
Fixes the normal map description, Godot uses OpenGL style where Y is positive. I've also reworded the note to be a bit better. Closes #4192

### DIFF
--- a/tutorials/3d/standard_material_3d.rst
+++ b/tutorials/3d/standard_material_3d.rst
@@ -375,10 +375,10 @@ and wider compatibility.
 
 .. note::
 
-  Godot requires the normal map to use the X+, Y- and Z+ coordinates. In other
-  words, if you've imported a material made to be used with another engine, you
-  may have to convert the normal map so its Y axis is flipped. Otherwise, the
-  normal map direction may appear to be inverted on the Y axis.
+  Godot requires the normal map to use the X+, Y+ and Z+ coordinates, this is
+  known as OpenGL style. If you've imported a material made to be used with
+  another engine it may be DirectX style, in which case the normal map needs to
+  be converted so its Y axis is flipped.
 
   More information about normal maps (including a coordinate order table for
   popular engines) can be found
--- a/tutorials/assets_pipeline/importing_images.rst
+++ b/tutorials/assets_pipeline/importing_images.rst
@@ -122,10 +122,10 @@ Note that RGTC compression affects the resulting normal map image. You will have
 
 .. note::
 
-  Godot requires the normal map to use the X+, Y- and Z+ coordinates. In other
-  words, if you've imported a material made to be used with another engine, you
-  may have to convert the normal map so its Y axis is flipped. Otherwise, the
-  normal map direction may appear to be inverted on the Y axis.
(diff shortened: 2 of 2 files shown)


Calinou:
The class reference should be updated in a separate PR to match this one:

```
MeshInstance2D.xml
18:			[b]Note:[/b] Godot expects the normal map to use X+, Y-, and Z+ coordinates. See [url=http://wiki.polycount.com/wiki/Normal_Map_Technical_Details#Common_Swizzle_Coordinates]this page[/url] for a comparison of normal map coordinates expected by popular engines.

BaseMaterial3D.xml
161:			[b]Note:[/b] Godot expects the normal map to use X+, Y-, and Z+ coordinates. See [url=http://wiki.polycount.com/wiki/Normal_Map_Technical_Details#Common_Swizzle_Coordinates]this page[/url] for a comparison of normal map coordinates expected by popular engines.
258:			[b]Note:[/b] Godot expects the normal map to use X+, Y-, and Z+ coordinates. See [url=http://wiki.polycount.com/wiki/Normal_Map_Technical_Details#Common_Swizzle_Coordinates]this page[/url] for a comparison of normal map coordinates expected by popular engines.

MultiMeshInstance2D.xml
18:			[b]Note:[/b] Godot expects the normal map to use X+, Y-, and Z+ coordinates. See [url=http://wiki.polycount.com/wiki/Normal_Map_Technical_Details#Common_Swizzle_Coordinates]this page[/url] for a comparison of normal map coordinates expected by popular engines.
```

skyace65:
Done. https://github.com/godotengine/godot/pull/56656

mhilbrunner (approved):

mhilbrunner:
Thanks! Merged the class ref PR too.

akien-mga:
Cherry-picked for 3.4.

Issue #5492, opened 2022-01-02
Document the exact order of all lifetime signals/callbacks/notifications


KoBeWi:
**Your Godot version:**
3.4.1

**Issue description:**
From time to time, I keep seeing people confused about order of lifetime methods. Recently there was a bug report, where it turned out that the user tried to access an `onready` variable inside a callback from signal emitted in `_ready()` of a child node. It's convoluted, but the `onready` variable wasn't initialized in parent, because child nodes will call their `_ready()` first.

We do have some docs about order of some methods, but there isn't any comprehensive guide that lists all callbacks/notifications etc and order of variable initialization.

I made a small test project. For a single node the order is this:
- **variables are initialized**
- **exported variables are initialized**
- `_init()` callback
- `NOTIFICATION_ENTER_TREE`
- `_enter_tree()` callback
- `tree_entered` signal
- `NOTIFICATION_POST_ENTER_TREE`
- **onready variables are initialized**
- `_ready()` callback
- `NOTIFICATION_READY`
- `ready` signal

It gets super-complicated when node has children. My test scene was like this:
![image](https://user-images.githubusercontent.com/2223172/147875711-d3055f10-e073-4f86-bede-e18d8eb4d9c5.png)
Tested with this script, that prints about everything for when node is created and destroyed:
```GDScript
extends Node

func _init() -> void:
	prints(self, "_init() callback")
	
	connect("tree_entered", self, "on_tree_entered")
	connect("ready", self, "on_ready")
	connect("tree_exiting", self, "on_tree_exiting")
	connect("tree_exited", self, "on_tree_exited")

func _enter_tree() -> void:
	prints(name, "_enter_tree() callback")

func _ready() -> void:
	prints(name, "_ready() callback")
	
	if get_parent() == get_tree().root:
		queue_free()

func _exit_tree() -> void:
	prints(name, "_exit_tree() callback")

func _notification(what: int) -> void:
	match what:
		NOTIFICATION_ENTER_TREE:
			prints(name, "NOTIFICATION_ENTER_TREE")
		NOTIFICATION_EXIT_TREE:
			prints(name, "NOTIFICATION_EXIT_TREE")
		NOTIFICATION_READY:
			prints(name, "NOTIFICATION_READY")
		NOTIFICATION_PREDELETE:
			prints(name, "NOTIFICATION_PREDELETE")
		NOTIFICATION_POSTINITIALIZE:
			prints(name, "NOTIFICATION_POSTINITIALIZE")
[...]

PR #5507, opened 2022-01-12
Update class_node.rst

michael-chase:
NOTIFICATION_EXIT_TREE triggers after `_exit_tree()` method executes.


Calinou:

KoBeWi:
The description isn't wrong btw. The notification is received together with `tree_exiting` signal. See https://github.com/godotengine/godot-docs/issues/5492

PR #5510, opened 2022-01-12
Update HUD section in "Your first game" from Custom Fonts to Theme Overrides

DiegoCCMM:
The way fonts were explained in the "Your first game" step was outdated, I've changed the explanation to the up-to-date version using Theme Overrides.


Calinou:
The screenshots should be taken again with the editor scale set to 100% in the Editor Settings (**Interface > Editor > Display Scale**). The screenshots included in this PR use an editor scale of 75%, likely because you are using the editor on a small display (1366×768 laptop).

pycbouh:
The original doc references dedicated font sizes because that's how it is now done in master/4.0. Since this PR is for the master branch of the docs, it's better not to break this existing text. You can keep the old screenshot for it, since all screenshots will require updating anyway due to Godot editor's theme changing.

You can make a dedicated PR for 3.x and remove that mention there.

bruvzg (changes requested):
The only relevant change for the `master` branch is `Custom Fonts` -> `Theme Overrides/Fonts`. The rest is for `3.x` only and is not valid for the `master`, there's no `DynamicFont` and size is set separatly (it's now under `Theme Overrides/Fonts Sizes` insted of `Custom Font Size`).

Screenshot 2022-01-13 at 00 05 40

DiegoCCMM:
I've reverted the changes so the images stays as they should, and maintaned the `Custom Fonts` -> `Theme Overrides/Fonts`. I'll redo this changes in the 3.4 branch

akien-mga:
Superseded by #5542.

PR #5518, opened 2022-01-15
Give example of one liner for natural sort

noidexe:
The documentation for the `sort` method warns the user that it doesn't do natural sort but fails to provide a solution when it's just a one liner thanks to `String.naturalnocasecmp_to()` and lambda support

This suggests exactly the same algorithm as used by the filesystem dock for file sorting.


Calinou:

noidexe:
🤦‍♂️ Thanks!

PR #5521, opened 2022-01-16
Split up and organize gdscript reference

### LINKED ISSUE #2735: Split up GDScript basics into many smaller sections.

skyace65:
Splits up and organizes the gdscript reference page. Still working on it

- All information on creating custom classes is now on its own page in the gdscript section of the scripting documentation
- All information on creating signals has been moved to it's own page in the general scripting section. This page is going to need more work in the future to add C# and go over basic signal creation, right now it only has information on custom signals
- Removed the small History and export sections pointing to other pages. Also the table mentions the export page so there's still a link to it here
- The onready and tool annotation sections have been moved to after the annotations table
- The signal and assert keyword sections have been moved to after the keywords table

Possibly Closes #2735, still working on this, I'm going to see what else is reasonable to split into its own page, possibly the control flow information.

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_basics.rst
+++ b/tutorials/scripting/gdscript/gdscript_basics.rst
@@ -16,16 +16,9 @@ allowing great flexibility for content creation and integration.
 
 GDScript is entirely independent from Python and is not based on it.
 
-History
--------
-
-.. note::
-
-    Documentation about GDScript's history has been moved to the
-    :ref:`Frequently Asked Questions <doc_faq_what_is_gdscript>`.
-
 Example of GDScript
 -------------------
+
 
 Some people can learn better by taking a look at the syntax, so
 here's an example of how GDScript looks.
@@ -136,7 +129,7 @@ as an identifier. Additionally, identifiers must not begin with a digit.
 Identifiers are case-sensitive (``foo`` is different from ``FOO``).
 
 Keywords
-~~~~~~~~
+--------
 
 The following is the list of keywords supported by the language. Since
 keywords are reserved words (tokens), they can't be used as identifiers.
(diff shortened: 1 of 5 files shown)


pycbouh:
> * All information on creating custom classes is now on its own page in the gdscript section of the scripting documentation
> * All information on creating signals has been moved to it's own page in the general scripting section. This page is going to need more work in the future to add C# and go over basic signal creation, right now it only has information on custom signals

You shouldn't completely remove those sections from a page called "GDScript reference". This will make it much harder to look up information. Keep a short one or two paragraph covering how those concepts relate to GDScript and what GDScript-specific language features are there, then link to those new articles. This is especially important if you want to extend the articles to cover more than just GDScript.

Basically, a reference page has to keep all the information that is specific about the language, but general usage guides of the features available in the engine can be moved elsewhere.

-----
A big part that doesn't fit the "reference" style page is the section about "Built-in types". They are universal, and not GDScript specific. They can be moved elsewhere. Incidentally, this is now the only section that explains anything about signals, a concept that is quite unique to GDScript.

PR #5523, opened 2022-01-17
updates have not been reflected in docs

prakhar-s-srivastava:
I am not very well versed but these are the changes


pycbouh:
Thanks for contributing! This information shouldn't be just injected somewhere in the middle without explanation. Instead, the actual examples and all related information must be updated to provide the actual information.

mhilbrunner:
Closed, see above explanation. Thanks for contributing :)

PR #5527, opened 2022-01-18
Update images and instruction for adding a custom font

aleksa-stajcic:
- Updated instructions 
- Uploaded new images
- Deleted old images


aleksa-stajcic:
Closed pr because there is an identical one already open #5510

PR #5538, opened 2022-01-22
Reverse parenting order of Sprite and AnimationPlayer in tutorial

zacryol:
The docs had a sprite as the child of the player, then *immediately* after
says to not do that. Why mislead the reader at all and not just do it
the better way from the start? There nothing really gained from the
sprite being a child here, not even simplicity.


mhilbrunner (approved):

mhilbrunner:
Well spotted! Thanks for contributing :)

akien-mga:
Cherry-picked for 3.4.

PR #5540, opened 2022-01-23
Description of Advanced Exports with examples

rossunger:

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_exports.rst
+++ b/tutorials/scripting/gdscript/gdscript_exports.rst
@@ -236,6 +236,64 @@ described in :ref:`doc_accessing_data_or_logic_from_object`.
 .. warning:: The script must operate in the ``tool`` mode so the above methods
              can work from within the editor.
 
+Using _get_property_list()
+~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


TheYellowArchitect:
What is that error on automated github check? "log not found"?

Calinou:
The class reference that's linked above already includes an example of using `_get_property_list()`, so I would prefer we modify that example instead to include something like dynamically hiding a property depending on the value of another boolean property (likely the most common use case).

AThousandShips:
This targets 3.4, is this still relevant?

tetrapod00:
I looked in https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_exports.html#advanced-exports and https://docs.godotengine.org/en/latest/classes/class_object.html#class-object-private-method-get-property-list, and I think this might cover some things that are not currently covered even in 4.4. I'm not per se opposed to a section on the exports page which covers usage of `_get_property_list()`.

However, since this:
- targets `3.4`,
- hasn't had any response in a long time, 
- needs some work on the content itself, 
- and is opposed by a maintainer in its current form,

I'm going to close this PR and mark this as salvageable. OP, if you're still around, I'd suggest reworking this into a new PR that targets `master`, or perhaps expanding the class reference example as Calinou suggests.

PR #5543, opened 2022-01-24
Update controls name in docs

Songbird0:
Hello!

In the section "2D movement overview", keyboard controls aren't prefixed by `ui_` and Godot 3.4 doesn't like non-prefixed controls.

Is it good for you?

Regards.

### DIFF
--- a/tutorials/2d/2d_movement.rst
+++ b/tutorials/2d/2d_movement.rst
@@ -49,13 +49,13 @@ Add a script to the kinematic body and add the following code:
 
     func get_input():
         velocity = Vector2()
-        if Input.is_action_pressed("right"):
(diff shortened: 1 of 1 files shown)


mhilbrunner:
Hey! If those produce errors, it means you didn't properly set up the custom actions (`right`) etc as described on that page in the previous section, `ui_right` and others work because they're default actions in the editor, but you should set up your own and not rely on these.

Calinou:
> Oh ok, but why should I set up my own if controls are already mapped?


This is mostly about good practices and accessibility. For all in-game actions, you should *always* use custom actions. For user interface actions, you can use the built-in `ui_` actions in your own code.

PR #5544, opened 2022-01-24
[-] Navigation: Replace non-breaking spaces with ASCII spaces

timjklein36:
I noticed that the [Real Time Navigation (3D)](https://docs.godotengine.org/en/stable/tutorials/navigation/real_time_navigation_3d.html) page (from the `3.4` branch) was not displaying images and also showed the reference tag at the top of the page (when, normally, this should not be rendered to HTML by sphinx).

Screenshot 2022-01-24 at 10-22-04 Real Time Navigation (3D)

Upon opening the `rst` file in VSCode, I noticed that the space characters were non-breaking spaces instead of normal ASCII spaces.

Screenshot_20220124_102333

I believe this is what was causing the issue, so this PR has replaced those with normal ASCII space characters.


skyace65:
@mhilbrunner  looks like there's still some link problems even after your fix today, CI is getting hung up on a different file now.

skyace65:
@timjklein36 is there a setting is VSCode that causes non breaking spaces? I was the one who added the page and made it with VSCode.

timjklein36:
@skyace65 I am not sure about why it would happen. Perhaps this post could shed some light: https://stackoverflow.com/questions/57162734/vscode-inserting-non-breaking-spaces-instead-of-regular-spaces.

In any case, it seems my version of VSCode displayed the "wrong" space characters using the yellow block outline to make it obvious. I don't think I enabled any setting to make that happen.

mhilbrunner:
Hey, our CI is fixed now for the 3.4 branch - could you please rebase/push to get a new build? Thanks :)

timjklein36:
@mhilbrunner Thanks! I rebased and pushed.

mhilbrunner:
Thank you!

Issue #5498, opened 2022-01-06
Your first 2D game - 4.0 readying


BryanQuigley:
**Your Godot version:** 4.0

**Issue description:** There are a few places where things have changed in 4.x.  I'm stuck on a few, but appear to be mostly easy fixes?

https://docs.godotengine.org/en/latest/getting_started/first_2d_game/04.creating_the_enemy.html 
Many node names have changed slightly.

https://docs.godotengine.org/en/latest/getting_started/first_2d_game/05.the_main_game_scene.html
export doesn't work here either, this is where I got stuck

TechnicalSoup:
> **URL to the documentation page:** https://docs.godotengine.org/en/latest/getting_started/first_2d_game/03.coding_the_player.html uses export, tried changing to @export but that didn't work

I found that changing **export** to **@export** worked - and I was able to see that variable in the Inspector. You're correct in that it should be shown with the preceding @ symbol in this guide/tutorial.

I'm using the 4.0 build from 20220118 on Windows 11 (64 bit)

BryanQuigley:
Thanks, I think I must have not been inspecting the right thing. That first one works now. The last export I need to find new syntax I guess?

TechnicalSoup:
> The last export I need to find new syntax I guess?

I think the format for the second export will be `@export var mob_scene: PackedScene` although I haven't actually tested that out myself yet. I'll have a go at it in the next few days and let you know if it works.

BryanQuigley:
That worked with one other change:
var mob = mob_scene.instantiate() #was instance

Also had to change some rand_range -> randi_range

acristin:
There are a few renames needed for the 4.0 beta. 

In section 5, "The main game scene":
* [Position2D has been renamed Marker2D](https://github.com/godotengine/godot-proposals/issues/5199)
* [PathFollow2D.origin is now PathFollow2D.progress_ratio](https://github.com/godotengine/godot/commit/bb936b2e273cc3ae1c7d775a0ab7e0e7e2a256a9).

I made an attempt at a fix, but I haven't tested the C# and C++ changes for origin -> progress_ratio. I'll try to verify those soon and update the PR if needed, but the guesses seem reasonable.

TheYellowArchitect:
This is the last one remaining from making the tutorial fully compatible with 4.0
https://github.com/godotengine/godot-docs/issues/6234

YuriSizov:
[...]

PR #5553, opened 2022-01-27
4.0 1st 2d game export,instance, and rand_range fixes

BryanQuigley:
export -> @export (and other changes)
instance -> instantiate

rand_range -> randi_range (if you want integar)

Issues found in #5498.


getting_started/first_2d_game/05.the_main_game_scene.rst
```diff
@@ -337,11 +339,11 @@ Note that a new instance must be added to the scene using ``add_child()``.
         mob.position = mob_spawn_location.position
 
         # Add some randomness to the direction.
-        direction += rand_range(-PI / 4, PI / 4)
+        direction += randi_range(-PI / 4, PI / 4)
```
mhilbrunner, line 342: Did you test this / are you sure this is right? According to https://github.com/godotengine/godot/pull/40718, the 4.0 equivalent of `rand_range()` is `randf_range()`, not `randi_range()`. Notably, `randi_range()` uses `int` instead of `float`, so we may lose a lot due to rounding here compared to the old behaviour.

mhilbrunner (commented):
Hey! Thanks for contributing to Godot's docs.

getting_started/first_2d_game/05.the_main_game_scene.rst
BryanQuigley: I did test, but I didn't know what it should look like... oops.  It is better with randf_range (otherwise it's just 4 directions..).  Updated with changes

getting_started/first_2d_game/05.the_main_game_scene.rst
```diff
@@ -337,11 +339,11 @@ Note that a new instance must be added to the scene using ``add_child()``.
         mob.position = mob_spawn_location.position
 
         # Add some randomness to the direction.
-        direction += rand_range(-PI / 4, PI / 4)
+        direction += randf_range(-PI / 4, PI / 4)
         mob.rotation = direction
 
         # Choose the velocity.
-        var velocity = Vector2(rand_range(150.0, 250.0), 0.0)
+        var velocity = Vector2(randi_range(150, 250), 0.0)
```
vermeeren, line 346: @BryanQuigley As Vector2 works with float as well, I believe this should be `randf_range` also.

BryanQuigley: I did test this one with either option, and I really can't tell the difference.

akien-mga: Changed it back to be consistent with the original code and other languages.

akien-mga (approved):

akien-mga:
Thanks!

PR #5554, opened 2022-01-27
Mention that Godot OpenXR doesn't work with Microsoft MR runtime

skyace65:
VR games made with the OpenXR plugin do not work with the Microsoft MR runtime, as it only supports projects made using DirectX, see https://github.com/GodotVR/godot_openxr/issues/51. This is only for the 3.4 branch as there's no documentation in master yet. I also removed an extra space in a paragraph.

### DIFF
--- a/tutorials/vr/openxr/runtime_selection.rst
+++ b/tutorials/vr/openxr/runtime_selection.rst
@@ -4,7 +4,7 @@ Switching runtimes
 ==================
 
 In OpenXR, it is standard for each runtime to implement a mechanism to make it
-the current runtime.  In Steam, the Oculus application or Windows MR portal,
+the current runtime. In Steam, the Oculus application or Windows MR portal,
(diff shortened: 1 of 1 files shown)


tutorials/vr/openxr/runtime_selection.rst
```diff
@@ -16,6 +16,9 @@ switch the runtime Godot will use when testing:
 
 .. image:: img/switch_runtime.png
 
+The OpenXR plugin will not work with the Microsoft MR runtime. That runtime only supports
+OpenXR applications that use DirectX, and Godot uses OpenGL 3 and 2.
```
Calinou, line 20:
```suggestion
The OpenXR plugin will **not** work with the Microsoft MR runtime.
That runtime only supports OpenXR applications that use DirectX,
but Godot uses OpenGL ES 3.0 or 2.0.
```

Calinou (approved):
Thanks!

PR #5526, opened 2022-01-18
Fix player and mob scene being reversed in Your first game

vgrover21:
Corrected the description of copying animation from Player scene to mob scene which incorrectly mentioned the source and target scenes

PR #5558, opened 2022-01-27
Player and Mob scene are reversed

SkatNMax:
Hi, Line 248 to 250, the named "Mob.tscn" and "Player.tscn" scene are reversed. The animation data must be copied from the "Player" scene. Also in the Mob scene a new AnimationPlayer node must be created to paste it.


Calinou:
Duplicate of https://github.com/godotengine/godot-docs/pull/5526. Thanks for the contribution nonetheless!

PR #5568, opened 2022-01-28
Get rid of openssl warning during PCK encryption setup

### LINKED ISSUE #5567: Warning message when generating an encryption key with openssl V1.1.1 or later
**Your Godot version:** 4.0 Alpha

**Issue description:**

Generating an encryption key with openssl according to the "Compiling with script encryption key" documentation results in a warning message.

```
user@linux:~$ openssl enc -aes-256-cbc -k secret -P -md sha1
*** WARNING : deprecated key derivation used.
Using -iter or -pbkdf2 would be better.
salt=650FCA822BDC94F5
key=FB20BDC7141D3704499F73098A6270F957AC134FBDA1EA99FEF5B3953EC2DF0B
iv =AE94260B8B4E7FA5B1082740E116CAF1
```

There is a detailed explanation of this warning message at: https://unix.stackexchange.com/questions/507131/openssl-1-1-1b-warning-using-iter-or-pbkdf2-would-be-better-while-decrypting 

I would expect that the suggested method does not result in warning messages.

https://docs.godotengine.org/en/stable/development/compiling/compiling_with_script_encryption_key.html

**Proposed Solution:**

I suggest to replace the command by 
```
openssl rand -hex 32 > godot.gdkey
```
This also is a more appropriate way for generating random data for the encryption key.

If this solution finds acceptance, I can create a pull request for updating the documentation.

Sauermann:
Tested with openssl v1.1.1k , v1.1.1m, and v3.0.1 and godotengine/godot@b9a2569be6fcd1127454127d989b504334d2dac8.

resolves #5567

### DIFF
--- a/development/compiling/compiling_with_script_encryption_key.rst
+++ b/development/compiling/compiling_with_script_encryption_key.rst
@@ -25,15 +25,13 @@ Step by step
 
    ::
 
-       openssl enc -aes-256-cbc -k secret -P -md sha1 > godot.gdkey
+       openssl rand -hex 32 > godot.gdkey
 
-   This should output the following to ``godot.gdkey`` file:
+   The output in ``godot.gdkey`` should be similar to:
 
    ::
 
-       salt=5786FE8B91CA048A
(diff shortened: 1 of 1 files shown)


Calinou (approved):

PR #5574, opened 2022-01-30
Path corrections

domschl:


Calinou:
`opt.tools` is correct when you use `target=release_debug` to compile (which you should do unless you're actually working on engine code).

bruvzg:
`lib` is correct change, at least for the current SDK.

The rest of the instructions are a bit outdated, `use_static_mvk` is no longer used, new way to statically link MoltenVK (which is recommended way to go, dylib can cause conflicts with user installed MVK) is `use_volk=no VULKAN_SDK_PATH=<Vulkan SDK path>`.

domschl:
Initially, I removed the `opt.` part from path when building the `.app` directory, because the compile instructions (which defaulted for a debug (non-`opt.`) build didn't fit with the instructions for moving the result into the app dir (which assumes a `target=release_debug` build.

I would suggest the following changes:

* make all instructions for building to use consistent `target=release_debug`.
* add a short summary of relation between `scons` options (`tools=`, `target=`) and their effect on pathnames at the beginning
* Expand 4.0master / Vulkan doc: required are (as far as I understand it) both Vulkan-SDK and Molten-VK. Fix `scons` flags for Vulkan, and change `<Vulkan SDK path>` to `<MoltenVK SDK path>` ?

Does that make sense? Shall I expand this pull-request, or does somebody else with more knowledge can do this?

bruvzg:
> required are (as far as I understand it) both Vulkan-SDK and Molten-VK

[Vulkan SDK](https://vulkan.lunarg.com/sdk/home#mac) for macOS includes MoltenVK, there's no separate MoltenVK SDK.

development/compiling/compiling_for_macos.rst
```diff
@@ -70,7 +70,7 @@ editor binary built with ``target=release_debug``::
 
     cp -r misc/dist/osx_tools.app ./Godot.app
     mkdir -p Godot.app/Contents/MacOS
-    cp bin/godot.osx.opt.tools.universal Godot.app/Contents/MacOS/Godot
+    cp bin/godot.osx.tools.universal Godot.app/Contents/MacOS/Godot
```
akien-mga, line 73:
```suggestion
    cp bin/godot.osx.opt.tools.universal Godot.app/Contents/MacOS/Godot
```

akien-mga:
I re-added the `.opt` part so that this can be merged to fix the `libs` bug.

But I agree that the whole page structure could use a refactor (and generally speaking the `target` usage in each platform specific compiling docs is often confusing users).

akien-mga:
Thanks!

PR #5578, opened 2022-02-03
Fix missing underscore in methods

BeayemX:


Calinou:
Is this needed for both `master` and `3.x`, only `3.x`?

pycbouh:
This is only needed for `master`, where we renamed all virtual methods to start with `_`. In `3.x` they have "public" names. This is one of the things that we need to update in the docs for 4.0.

mhilbrunner (approved):

mhilbrunner:
Thank you!

PR #5582, opened 2022-02-05
Fix first 2D game image

### LINKED ISSUE #5571: Docs: Screenshots for project folder and Godot's FileSystem are not consistent
As you can see at https://docs.godotengine.org/en/stable/getting_started/first_2d_game/01.project_setup.html, dodge_assets shouldn't be in the FileSystem.

image

image

I guess it should show something like this:

image

skyace65:
Fixes a filesystem image. dodge_assets shouldn't show up in the file structure. The image has gone through png compression. Closes #5571.


Calinou (approved):
Thanks!

Issue #5577, opened 2022-02-03
[Godot 4] Step by step tutorial deprecated syntax for signal connecting


SimonImming:
**Godot version:**
Godot Engine v4.0.alpha1.official [31a7ddbf8]

**Issue description:**
The Step by Step tutorial uses a deprecated syntax for connecting signals. This is the code used in the documentation:
```
func _ready():
    var timer = get_node("Timer")
    timer.connect("timeout", self, "_on_Timer_timeout")
```

This code does not work in Godot 4 and produces the following error message:
```
Invalid argument for "connect()" function: argument 2 should be Callable but is res://Sprite2D.gd.
Invalid argument for "connect()" function: argument 3 should be Array but is String.
```

The correct code according to the [Class Reference](https://docs.godotengine.org/en/latest/classes/class_object.html#class-object-method-connect) would be:
```
func _ready():
	var timer = get_node("Timer")
	timer.timeout.connect(_on_Timer_timeout)
```

This code has also been tested in Godot 4 with the tutorial files from the step by step tutorial and works as expected.

**URL to the documentation page:**
[https://docs.godotengine.org/en/latest/getting_started/step_by_step/signals.html](https://docs.godotengine.org/en/latest/getting_started/step_by_step/signals.html)

PR #5585, opened 2022-02-05
fix 'making_plugin' tutorial code snippet

bruno-garcia:
On version: v4.0.alpha1.official [31a7ddbf8]

Following the guide Making Plugins from 'latest' cause: https://docs.godotengine.org/en/latest/tutorials/plugins/editor/making_plugins.html
image

```
Line 6:Invalid argument for "connect()" function: argument 2 should be Callable but is res://addons/sentry/my_button.gd
Line 6:Invalid argument for "connect()" function: argument 3 should be Array but is String.
```

It seems the new API takes a callback directly.

Related to: https://github.com/godotengine/godot-docs/issues/5577

### DIFF
--- a/tutorials/plugins/editor/making_plugins.rst
+++ b/tutorials/plugins/editor/making_plugins.rst
@@ -177,10 +177,10 @@ clicked. For that, we'll need a script that extends from
     {
         public override void _EnterTree()
         {
-            Connect("pressed", clicked);
+            Connect("pressed", Clicked);
         }
 
-        public void clicked()
+        public void Clicked()
         {
             GD.Print("You clicked me!");
         }
--- a/tutorials/plugins/editor/making_plugins.rst
+++ b/tutorials/plugins/editor/making_plugins.rst
@@ -161,7 +161,7 @@ clicked. For that, we'll need a script that extends from
 
 
     func _enter_tree():
-        connect("pressed", clicked)
+        pressed.connect(clicked)
 
 
     func clicked():
--- a/tutorials/plugins/editor/making_plugins.rst
+++ b/tutorials/plugins/editor/making_plugins.rst
@@ -161,7 +161,7 @@ clicked. For that, we'll need a script that extends from
 
 
     func _enter_tree():
-        connect("pressed", self, "clicked")
+        connect("pressed", clicked)
 
 
     func clicked():
@@ -177,7 +177,7 @@ clicked. For that, we'll need a script that extends from
     {
         public override void _EnterTree()
         {
-            Connect("pressed", this, "clicked");
+            Connect("pressed", clicked);
         }
 
         public void clicked()


tutorials/plugins/editor/making_plugins.rst
```diff
@@ -177,7 +177,7 @@ clicked. For that, we'll need a script that extends from
     {
         public override void _EnterTree()
         {
-            Connect("pressed", this, "clicked");
+            Connect("pressed", clicked);
```
bruno-garcia, line 180: Note that I didn't test the C# version.
The GDScript works

tutorials/plugins/editor/making_plugins.rst
```diff
@@ -161,7 +161,7 @@ clicked. For that, we'll need a script that extends from
 
 
     func _enter_tree():
-        connect("pressed", self, "clicked")
+        connect("pressed", clicked)
```
akien-mga, line 164: That works, but it's actually the old style signal connection. The new syntax (and preferred for tutorials) should be:
```suggestion
        pressed.connect(clicked)
```

tutorials/plugins/editor/making_plugins.rst
akien-mga: CC @godotengine/mono

raulsntos: In C# the naming convention for methods is PascalCase so the method `clicked` should be named `Clicked`.

And in Godot 4.0 signals are generated as C# events by the bindings generator so you can connect them like you would any other C# event. The syntax is:

```csharp
Pressed += Clicked;
```

I haven't checked what name is generated for the event because in this case I think it conflicts with the `Pressed` property so it probably would be generated with a suffix like `PressedSignal` or something.

bruno-garcia: Cool, I'll rename the method to follow the C# naming convention. Tks

raulsntos: You still need to change the connect syntax in C#:

```csharp
Pressed += Clicked;
```

mhilbrunner (approved):

tutorials/plugins/editor/making_plugins.rst
```diff
@@ -177,10 +177,10 @@ clicked. For that, we'll need a script that extends from
     {
         public override void _EnterTree()
         {
-            Connect("pressed", this, "clicked");
+            Connect("pressed", Clicked);
```
raulsntos, line 180: The C# syntax is still incorrect

```suggestion
            Pressed += Clicked;
```

raulsntos: Resolved in https://github.com/godotengine/godot-docs/pull/6110

PR #5587, opened 2022-02-06
Update connecting signal code for Godot 4

### LINKED ISSUE #5577: [Godot 4] Step by step tutorial deprecated syntax for signal connecting

skyace65:
Updates the example of connecting a signal via code in the step by step section for Godot 4. Closes #5577.

### DIFF
--- a/getting_started/step_by_step/signals.rst
+++ b/getting_started/step_by_step/signals.rst
@@ -241,12 +241,12 @@ We can now connect the Timer to the Sprite2D in the ``_ready()`` function.
 
     func _ready():
         var timer = get_node("Timer")
-        timer.connect("timeout", self, "_on_Timer_timeout")
+        timer.timeout.connect(_on_Timer_timeout)
 
 The line reads like so: we connect the Timer's "timeout" signal to the node to
-which the script is attached (``self``). When the Timer emits "timeout", we want
-to call the function "_on_Timer_timeout", that we need to define. Let's add it
-at the bottom of our script and use it to toggle our sprite's visibility.
+which the script is attached. When the Timer emits "timeout", we want to call
+the function "_on_Timer_timeout", that we need to define. Let's add it at the
+bottom of our script and use it to toggle our sprite's visibility.
 
 .. tabs::
  .. code-tab:: gdscript GDScript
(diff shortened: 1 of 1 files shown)


getting_started/step_by_step/signals.rst
```diff
@@ -241,12 +241,12 @@ We can now connect the Timer to the Sprite2D in the ``_ready()`` function.
...
         var timer = get_node("Timer")
-        timer.connect("timeout", self, "_on_Timer_timeout")
+        timer.timeout.connect(_on_Timer_timeout)
 
 The line reads like so: we connect the Timer's "timeout" signal to the node to
-which the script is attached (``self``). When the Timer emits "timeout", we want
-to call the function "_on_Timer_timeout", that we need to define. Let's add it
-at the bottom of our script and use it to toggle our sprite's visibility.
+which the script is attached. When the Timer emits "timeout", we want to call
+the function "_on_Timer_timeout", that we need to define. Let's add it at the
```
Calinou, line 248:
```suggestion
which the script is attached. When the Timer emits ``timeout``, we want to call
the function ``_on_Timer_timeout()``, that we need to define. Let's add it at the
```

PR #5589, opened 2022-02-08
Update mentions of "editor_hint"

shmolyneaux:
At some point it seems like `Engine.editor_hint` was removed in favor of `Engine.is_editor_hint`. You can see this in the [latest Engine class reference](https://docs.godotengine.org/en/latest/classes/class_engine.html). This change updates references to `editor_hint` in `running_code_in_the_editor.rst`.

This only updates the gdscript examples. The C# code has references to `Engine.EditorHint` that may need to be updated as well, but I don't have a Mono build I can check against.

### DIFF
--- a/tutorials/plugins/running_code_in_the_editor.rst
+++ b/tutorials/plugins/running_code_in_the_editor.rst
@@ -50,7 +50,7 @@ For example, if you want to execute some code only in the editor, use:
 
  .. code-tab:: csharp
 
-    if (Engine.EditorHint)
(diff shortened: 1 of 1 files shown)


raulsntos:
The C# should be `Engine.IsEditorHint()`

shmolyneaux:
Thanks for finding that change! I've updated my branch to include the C# changes as well.

akien-mga:
Thanks!

PR #5601, opened 2022-02-12
Fix and improve Pyston instructions in Compiling for Linux, *BSD

Calinou:
- Fix SCons path to match Pyston 2.3.2's portable archive.
- Use a symbolic link instead of a wrapper script, which works the same.


akien-mga (approved):

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 3.4.

PR #5609, opened 2022-02-14
Add [br] tag to `GDScript documentation comments`

sent44:
Since `[br]` is not BBCode tag but for some reason it work on GDScript docs and yet still no document about it on this page.


dalexeev:
The mention of `[br]` was added in #5838, but I agree that we need to harmonize BBCodes in `RichTextLabel`, `EditorHelp`, and GDScript doc comments.

sent44:
Closed due #5838 
@dalexeev In usage, I kind of disagree that `Line2` is on difference line from `Line1 [br]`

PR #5611, opened 2022-02-15
Fix a typo in Using the MeshDataTool

denicolas:
### 3D » Procedural geometry » Using the MeshDataTool

"Each face in the face array contains a reference to the three **triangles** and three edges it is composed of."
should be:
"Each face in the face array contains a reference to the three **vertices** and three edges it is composed of."


clayjohn (approved):

Calinou (approved):
Thanks!

PR #5612, opened 2022-02-15
Fix another typo in Using the SurfaceTool

denicolas:
### 3D » Procedural geometry » Using the SurfaceTool

"call commit() to **finished** generating the mesh"
should be:
"call commit() to **finish** generating the mesh"


clayjohn (approved):

Calinou (approved):
Thanks!

PR #5623, opened 2022-02-19
Doc edit suggestion in '05.the_main_game_scene.rst' to avoid confusion 

funbeedev:
Making this suggestion for  `05.the_main_game_scene.rst` because I got a bit confused while reading through this paragraph:

`To have the points snap to the grid, make sure "Use Grid Snap" and "Use Snap" are both selected. These options can be found to the left of the "Lock" button, appearing as a magnet next to some dots and intersecting lines, respectively.`

I suggest that `Use Smart Snap` should replace `Use Snap` . It was a bit unclear to me while going through this section because I couldn't find a `Use Snap` button so I feel it should state `Use Smart Snap` instead.

Also edited the image `grid_snap_button.png` to highlight the `Use Smart Snap` button as the paragraph mentions both buttons should be selected but only highlights the `Use Grid Snap` button in the image.


cbscribe:
Thanks!

PR #5624, opened 2022-02-20
3d game tutorial: fix asset url version

vermeeren:
On the tutorial index the version is already 1.1.0 since commit
a187b111714565cc7a23e97f8209e0f52f17e725. However the reference on the
second page was forgotten. This patch corrects that.


cbscribe:
Thanks!

redactedscribe:
I'm not sure why, but the even after this merge months ago, the live docs for v3.5 _still_ point to tag 1.0.0 on [the second page of the _Your first 3D game_ tutorial](https://docs.godotengine.org/en/3.5/getting_started/first_3d_game/01.game_setup.html).

If the asset links on both pages 1 and 2 were changed to `https://github.com/godotengine/godot-3d-dodge-the-creeps/releases/latest` (rather than linking to specific tag) then no future updates would be needed.

But that still doesn't explain why this merge seems to have been ignored from the live docs...

@cbscribe @Calinou

PR #5637, opened 2022-02-22
SurfaceTool example incorrect

novaplusplus:
Should be add_vertex not set_vertex. I guess the change of all the other params to set_* led to it being inadvertantly changed in the example as well.


Calinou:
> Sorry, I believe I am doing this in the wrong spot. I will check again and re-submit the changes where appropriate...

Indeed, class reference changes should be made on the main Godot repository, not in godot-docs as the `classes/` folder here is automatically generated.

PR #5638, opened 2022-02-22
add legacy ImageMagick command for ICO convert

SylvanSign:


Calinou:
Are there systems out there that do not install the `magick` binary? It's been a thing for several years now (if not a decade).

SylvanSign:
I'm running Debian stable and it doesn't seem to install the `magick` binary :smile:

Calinou (approved):
Thanks!

Linux distribution packaging quirks never cease to amaze me… :no_mouth:

PR #5639, opened 2022-02-23
Document line continuation in GDScript

### LINKED ISSUE #5517: Missing line continuation backslash ("\") documentation in GDScript Basics
**Your Godot version:**
3.4.2
**Issue description:**
The GDScript Basics page contains these sample [lines of code](https://docs.godotengine.org/en/stable/getting_started/scripting/gdscript/gdscript_basics.html?#if-else-elif):
```
    # Alternative syntax with backslashes instead of parentheses (for multi-line expressions).
    # Less lines required, but harder to refactor.
    var fruit_alt = \
            "apple" if count == 2 \
            else "pear" if count == 1 \
            else "banana" if count == 0 \
            else "orange"
    print(fruit_alt)  # banana
```

Until the moment I saw this, I didn't even know this syntax _(inspired by some languages, among which is Python)_ was available. Its usage seems not mentioned anywhere in the documentation,
Not that it should be encouraged. For most cases, enclosing with parenthesis servers the same purpose and is cleaner. Still, I can think of a few, long-winded expressions I may favour it.

The important bit is that, mentions of **"\\"** as a line continuation seems oddly non-existent, unlike the other symbols used in GDScript. Could it be because it's an overwhelming concept to understand to new users? Either way, it should be noted _somewhere_ more substantial than a mere example.

https://docs.godotengine.org/en/stable/getting_started/scripting/gdscript/gdscript_basics.html

skyace65:
Adds a section to the GDScript reference on line continuation. Closes #5517

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_basics.rst
+++ b/tutorials/scripting/gdscript/gdscript_basics.rst
@@ -395,6 +395,27 @@ considered a comment.
 
 .. _doc_gdscript_builtin_types:
 
+Line continuation
+~~~~~~~~~~~~~~~~~
+
+A line of code in GDScript can be continued on the next line by using a backslash
+(``\``). Add one at the end of a line and the code on the next line will act like
+it's where the backslash is. Here is an example:
+
+::
+
+    var a = 1 + \
+    2
+
+A line can be continued multiple times like this:
+
+::
+
+    var a = 1 + \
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):

mhilbrunner:
Thanks!

PR #5643, opened 2022-02-24
requirements: Update to Sphinx 4.4.0, RTD theme 1.0.0, notfound 0.8

akien-mga:
Needed to support Python 3.10.

We're then in sync with https://github.com/readthedocs/readthedocs.org/blob/master/requirements/docs.txt


akien-mga:
Cherry-picked for 3.4.

PR #5646, opened 2022-02-25
changed the rotation script because script didn't work 

sander732:


Calinou:
@sander732 Can you further describe why this change is needed? Which Godot version did you test it on?

akien-mga:
This doesn't look right. You're removing C# code for no reason, and you're replacing the code of the section that's meant to use `look_at` by code of the previous section that doesn't use `look_at`. I think that's a misunderstanding.

If there's an issue with the code in the `look_at` section, please file a bug report so we can look into it.

PR #5648, opened 2022-02-25
Update signals.rst

NobleValerian:
So far, the user has been following along and changing their project as concepts are explained.  If they continue into Custom signals, they learn (for the first time, and by accident, it should be noted) that the "extends" keyword may only be present once per script.  It doesn't appear this section is meant to be followed, as there is no indication of what these scripts should be attached to.


akien-mga:
This makes sense, thanks! I wanted to rephrase the note a bit and since I was reworking these tutorials in #5782, I included it directly in that PR.

PR #5660, opened 2022-03-04
Change estimated release date for Godot 3.5 to Q2 2022

Calinou:
Q1 2022 ends on 2022-03-31, which is too soon considering that only3.5beta1 was released so far. The RC stage wasn't reached yet.


akien-mga:
Thanks!

akien-mga:
Cherry-picked for 3.4.

PR #5662, opened 2022-03-06
Minor addition to help out new users

mohamedAdhamc:
While building the source code on windows. I saw the requirements and proceeded to download what i needed first. it worked out fine but for python, I forgot to  "enable the option to add Python to the PATH in the Python installer" and I then had to redownload it again. I would assume this would happen with many. So I added a "(see below)" to kind of notify that there was information regarding the installation before proceeding to do so.

p.s- thought of explicitly adding "enable the option to add Python to the PATH in the Python installer" there but thought it would be a bit too redundant because it was already mentioned below.


mhilbrunner:
The idea to point out that the whole page should be checked first and there is pertaining information for the installation is good, however it is inconsistent to do this only for Python, I think? There are also sections for Visual Studio and SCons.

Maybe we can add a hint to the top of the list instead.

akien-mga:
I reworked things more in depth to be clearer: #5785.

PR #5663, opened 2022-03-06
Added helpful link.

mohamedAdhamc:
Helpful link to help out beginners.


mhilbrunner:
Good idea, but this should be a internal/Sphinx RST reference so it works across docs versions. :)

akien-mga:
Aside from @mhilbrunner's comment, it's better to link all platform-specific instructions to the page that we already have that describes in depth how to clone the Godot source repository, so I PR'ed #5784 for this.

PR #5670, opened 2022-03-09
Sphinx: Update to 4.4.0, drop incompatible readthedocs-sphinx-search

### LINKED ISSUE #5650: Duplicate results in docs search popup
**Your Godot version:**
stable (3.4)
**Issue description:**
duplicate search results in docs. Seems to always show exactly 2 identical hits for each relevant page
Example showing duplicate results https://docs.godotengine.org/en/stable/classes/class_string.html?rtd_search=pckpacker
image

akien-mga:
The readthedocs-sphinx-search extension is not actively maintained upstream,
and they pin some extremely old versions of Sphinx components which are not
compatible with what we need to use.

It was already dropped in the master branch, this syncs the requirements.txt
from master with 3.4.

Fixes #5650

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -61,9 +61,6 @@ ogp_site_name = "Godot Engine documentation"
 if not os.getenv("SPHINX_NO_GDSCRIPT"):
     extensions.append("gdscript")
 
-if not os.getenv("SPHINX_NO_SEARCH"):
-    extensions.append("sphinx_search.extension")
-
 if not os.getenv("SPHINX_NO_DESCRIPTIONS"):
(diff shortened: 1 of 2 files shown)


akien-mga:
It looks like upstream RTD is using the extension with latest Sphinx and sphinx-tabs: https://github.com/readthedocs/readthedocs.org/blob/72b7a4d9f6ec36006d7652cf7356600257ecec2a/requirements/docs.txt

We could give it another try and see what works and what doesn't. But this should likely be tested in the `master` branch first, so for now in `3.4` I'll merge this to sync it with what we've been using in `master` for months.

PR #5671, opened 2022-03-10
Minor accuracy update to ViewportContainer's description

melalawi:
The original description of this class implies that Viewport scaling occurs automatically. However, `ViewportContainer.stretch` by default is `false`, which means the description of the class is misleading. I should know, it took me way too long to realize why my `ViewportContainer`s weren't behaving as described.


Calinou:

PR #5673, opened 2022-03-11
Add openssl and freetype dependencies to Fedora (3.4)

bfloch:
I am not able to compile godot without `openssl-devel` and `freetype-devel`. This might apply to other distros as well, but I have no way to check.


akien-mga:
What Godot version are you trying to compile?

Godot doesn't use OpenSSL since Godot 3.1... and freetype defaults to using the vendored version, so it's not needed either.

PR #5683, opened 2022-03-12
Fix section header markup

timothyqiu:
The "Multi-threading" [here](https://docs.godotengine.org/en/stable/tutorials/performance/index.html) is not recognized as a section header due to RST syntax restrictions. The underline has to be at least as long as the text.


Calinou (approved):
Thanks!

PR #5684, opened 2022-03-12
Clarify the name of the menu action to add a node in Signals

Zazucki:
In the right-click context menu, it's labeled Add Child Node, not just Add Node or Add New Node.


Calinou (approved):

PR #5681, opened 2022-03-12
Improve pausing games documentation (3.4)

skyace65:
Clarifies that physics are paused for all nodes, even ones that have their pause mode set to `process`. Clarifies that signals still work while the game is paused, and that signal functions for paused nodes will still run. And Clarifies what happens to animation nodes, audio nodes and particle nodes. Cleans up some parts of the page. I removed the image from the example because I don't think it made things any clearer.

This is for issues #4783, #4752 and #3956. I'm going to make a 4.0 PR that closes these and updates the page for pausing changes in 4.0.

@golddotasksquestions I know you mentioned `CONNECT_DEFERRED` in your issue. I tried the code example you made for reddit and from what I'm seeing signals connected with `CONNECT_DEFERRED` do not automatically run when the game is paused, I changed the `wait_time` value to 3 to test this, and if you pause immediately "timeout" doesn't get printed to the console.

PR #5686, opened 2022-03-13
Improve and update pausing games for Godot 4

### LINKED ISSUE #3956: Setting the pause mode to Process doesn't process RigidBody2D
**Your Godot version:**
3.2.stable
**Issue description:**
While making a game I noticed that setting the pause mode of a Rigidbody2d to Process doesn't actually process the movement and physics of it. In my research I've found out that the Physics2DServer is also paused. To specifically fix the mentioned issue I had to do 
```		Physics2DServer.set_active(true)```
. I think this should be added to this tutorial as it is a bit misleading to state that setting the pause mode to process processes all nodes. I haven't tested other nodes but i suspect that this also influences other nodes

https://docs.godotengine.org/en/stable/tutorials/misc/pausing_games.html?highlight=pausing

### LINKED ISSUE #4752: Mention that signals are not "paused" with get_tree().paused = true
**Your Godot version:** Godot 3.2.3, 3.2.4 rc4

**Issue description:**
Signals are not mentioned in Pause tutorial, Pause is not mentioned in the signal tutorial.

The pause tutorial says "This effectively stops the whole game." which is not only misleading but also technically incorrect. I have projects without physics, input events and process, and they would apparently not be affected. (or only partially, since I don't always use CONNECT_DEFFERED)

If I am used to pause in order to "stop the whole game", but then suddenly want to do the same in a project which relies heavily on signals (seems a lot easier than disconnecting and later reconnecting all signals for the same effect), I would be really confused and irritated why it does not stop my game.

https://www.reddit.com/r/godot/comments/m44s0o/til_signal_callbacks_are_still_processed_normally/

"Won't pause signals. It will prevent signal emitting nodes (like Timer nodes, Buttons) from processing and accepting input_events, though. But this will cause functions called by CONNECT_DEFFERED signals to be called the frame after the tree has been paused. Further signals can be emitted from such a function while the tree is paused." 

I posted some example code  for testing [here](https://www.reddit.com/r/godot/comments/m44s0o/til_signal_callbacks_are_still_processed_normally/gqt19z0?utm_source=share&utm_medium=web2x&context=3).

Imho something like that should be mentioned here:
https://docs.godotengine.org/en/3.2/classes/class_scenetree.html#class-scenetree-property-paused
https://docs.godotengine.org/en/3.2/tutorials/misc/pausing_games.html#how-pausing-works
https://docs.godotengine.org/en/3.2/getting_started/step_by_step/signals.html

### LINKED ISSUE #4783: Add physics limitations to 3.x pause_mode
**Your Godot version:**
3.2.3

**Issue description:**
Refer to comments in https://github.com/godotengine/godot/issues/47326

When pause_mode is set to true, _process() and _physics_process() are still called for nodes with pause_mode set to 'process', i.e. they can still run and bodies can still move. However from what I believe they stop colliding and no collision signals are called. This is because the physics world can only be processed or paused as a whole.

It would be good if the pausing games section linked below contained some kind of section detailing this, i.e. the limitations of pause mode on collision handling and how Physics2DServer.set_active(true) can be used as a partial solution. As it stands, this not being mentioned gives a false pretence that a node in pause mode can function normally.

https://docs.godotengine.org/en/stable/tutorials/misc/pausing_games.html

and also,
https://docs.godotengine.org/en/stable/classes/class_scenetree.html#class-scenetree-property-paused

skyace65:
This includes some of the changes I made in #5681 but overhauls the page more. Closes #4783, closes #4752, closes #3956.  The image I've updated has gone through png compression. I removed one image from the example part of the page because I don't think it helps the tutorial at all.


mhilbrunner (approved):

mhilbrunner:
Looks good to me, will leave time for others to review

mhilbrunner:
Alright, in it goes :P Thank you! 🎉

PR #5687, opened 2022-03-13
Your first 2D game: Remove old link

rburing:
The "Your first 2D game" page is one of the first pages seen by new users; we should not confuse users with an outdated link to an older version of the project (which differs from the version described in the current documentation).


skyace65 (approved):

Calinou (approved):
Thanks!

PR #5694, opened 2022-03-16
Add missing `onready` for NodePath usage in GDScript exports

HarmonyHoney:
image

attempting to initialize a variable using get_node() without onready will lead to a compiler error.


Calinou (approved):
Thanks!

PR #5703, opened 2022-03-22
Compiling with Mono: Update path to generated C# solutions

akien-mga:


development/compiling/compiling_with_mono.rst
```diff
@@ -68,8 +68,9 @@ This path must be ``modules/mono/glue`` in the Godot directory::
 
     <godot_binary> --generate-mono-glue modules/mono/glue
 
-This command will tell Godot to generate the file ``modules/mono/glue/mono_glue.gen.cpp``
-and the C# solution for the Godot API at ``modules/mono/glue/Managed/Generated``.
+This command will tell Godot to generate the file ``modules/mono/glue/mono_glue.gen.cpp``,
+the C# solution for the Godot API at ``modules/mono/glue/GodotSharp/GodotSharp/Generated``,
+and the C# solution for the editor tools at ``modules/mono/glue/GodotSharp/GodotSharpEditor/Generated``.
```
raulsntos, line 73: GodotSharpEditor is still Godot API though, but editor-only (not available in exported games), like `EditorInspector`. The wording `editor tools` makes me think of `modules/mono/editor/GodotTools` which contains the C# solution for the Godot plugin, etc.

mhilbrunner (approved):
Seems good. To me its pretty clear, but the wording may indeed be confusing to newer users as pointed out by raulsntos, not sure.

development/compiling/compiling_with_mono.rst
neikeq: Correct. We may want to change to better naming before 4.0.

mhilbrunner:
Merging so the docs are updated. We may still want to improve that naming before 4.0 release proper!

PR #5704, opened 2022-03-22
docs: glue generation, change wrong paths

hannojg:
I found that the path in the documentation wasn't update and confused me as a first time user, trying to generate the glue and not founding the named solution.

I am not sure if "C# solution for GoDot Editor API" is correctly coined?


mhilbrunner:
Thanks for contributing! There was another PR fixing this which I just merged, this should now be improved: https://github.com/godotengine/godot-docs/pull/5703 (as soon as the docs are re-deployed on RTD)

PR #5714, opened 2022-03-25
Add Xbox Series X/S support by Pineapple Works

wkubiak:
Recently we've completed work on a native Xbox GDK port of Godot Engine 3.x so it would be great to reflect that in the docs by extending Pineapple Works' console support to the current-gen Xbox Series X/S consoles.


Calinou (approved):
Thanks!

PR #5722, opened 2022-03-30
Update list of features for Godot 3.5

Calinou:
3.5 version of https://github.com/godotengine/godot-docs/pull/5723. Includes a subset of https://github.com/godotengine/godot-docs/pull/5709 and https://github.com/godotengine/godot-docs/pull/5707.

**Do not merge** until Godot 3.5 is released. Note that this PR will need to have its base branch switched to `3.5` once that branch is created before it should be merged.

PR #5723, opened 2022-03-30
Update list of features for Godot 3.4

Calinou:
`3.4` version of https://github.com/godotengine/godot-docs/pull/5722. Pretty much the same, but with 3.5-only features removed.


mhilbrunner:
🎉 All the features. Thanks for keeping it updated :)

PR #5724, opened 2022-03-31
Document troubleshooting rendering issues on Android and iOS

Calinou:
This is long overdue, but it's finally done now :slightly_smiling_face: 

There is some copy-pasting involved, but I figured this was better than creating a separate page to make sure the information isn't missed by people exporting to mobile for the first time.

See https://github.com/godotengine/godot/pull/59715.

### DIFF
--- a/tutorials/3d/3d_rendering_limitations.rst
+++ b/tutorials/3d/3d_rendering_limitations.rst
@@ -3,6 +3,12 @@
 3D rendering limitations
 ========================
 
+.. seealso::
+
+    In addition to the limitations below, mobile platforms have even more
+    limitations on 3D rendering compared to desktop platforms.
+    See :ref:`doc_mobile_rendering_limitations` for more information.
+
 Introduction
 ------------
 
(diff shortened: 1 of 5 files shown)


skyace65:
Given that this applies to both android and ios I think it might make more sense if this was a single new page in the "rendering" or "platform specific" doc sections. We could then link to it from the exporting pages.

mhilbrunner:
I'm not absolutely against duplicating critical information, but maybe we can have a single page for these and link there? If you disagree, I'm fine with merging this as-is, it's not a huge amount of information.

mhilbrunner (approved):
@Calinou See above note, feel free to either merge this PR or amend it to merge (heh) the pages.

Calinou:
Merged into a single page, please take a look again (just in case).

skyace65:
This looks good, but why is the link to the page on the compiling for ios page instead of the exporting section page like it is for android?

Calinou:
> This looks good, but why is the link to the page on the compiling for ios page instead of the exporting section page like it is for android?

Good catch, thanks!

skyace65 (approved):

mhilbrunner (approved):

mhilbrunner:
Thanks, now this indeed is way cleaner, good idea @skyace65 and thanks for the reviews.

Thanks @Calinou for the work 🎉

PR #5728, opened 2022-04-03
Rename references to Node3D back to Spatial

Calinou:
Spatial was only renamed to Node3D in Godot 4.0, not Godot 3.x.


mhilbrunner (approved):

mhilbrunner:
Good catch!

PR #5732, opened 2022-04-04
Move GDScript text onto tip in 2D Getting started 03 chapter

solar-core:


NathanLovato:
Seems good to me. It seems that prior to this change the text is indented and it shouldn't

Calinou (approved):
Thanks!

Cherry-picked to the `3.4` and `stable` branches.

PR #5756, opened 2022-04-10
Add docs for ignoring return value warning

### LINKED ISSUE #5753: Fix RETURN_VALUE_DISCARDED warning in *Designing the mob scene* in Mob.gd
**Your Godot version:**

3.4.4.stable.official

**Issue description:**

The return value from ```move_and_slide()``` in ```Your first 3D game » Designing the mob scene``` isn't used which causes a code warning that can be confusing to learners. This is addressed in the actual code example (link below) but not in the documentation.

https://github.com/godotengine/godot-docs/blob/4060179184d1059a16ceb91d0a48c765c62784aa/getting_started/first_3d_game/04.mob_scene.rst#L269-L271

image

Where this is addressed in the actual code:

https://github.com/godotengine/godot-3d-dodge-the-creeps/blob/0b7d43cae85580ebdfacdb5ba88f64d3307b9c14/dodge_the_creeps_3d/Mob.gd#L15

Suggestion is to add the ```warning-ignore``` comment to the documentation to bring it inline with the actual example, or to explicitly handle the return value with something like:

```var _disregard = move_and_slide(velocity)```

https://docs.godotengine.org/en/stable/getting_started/first_3d_game/04.mob_scene.html

ohnoimdead:
Sounds like this portion of the docs will change in trunk, so just putting up a PR against 3.4. This is my first PR against this project so let me know if I screwed anything up. :D

Ran the sphinx build locally and verified the ref: works.

Fixes #5753

### DIFF
--- a/getting_started/first_3d_game/04.mob_scene.rst
+++ b/getting_started/first_3d_game/04.mob_scene.rst
@@ -139,6 +139,13 @@ Similarly to the player, we move the mob every frame by calling
 the ``velocity`` every frame: we want the monster to move at a constant speed
 and leave the screen, even if it were to hit an obstacle.
 
+You may see a warning in GDScript that the return value from
+``move_and_slide()`` is unused. This is expected. You can simply ignore the
+warning or, if you want to hide it entirely, add the comment
+``# warning-ignore:return_value_discarded`` just above the
+``move_and_slide(velocity)`` line. To read more about the GDScript warning
+system, see :ref:`doc_gdscript_warning_system`.
+
 We need to define another function to calculate the start velocity. This
 function will turn the monster towards the player and randomize both its angle
 of motion and its velocity.
@@ -267,6 +274,7 @@ Here is the complete ``Mob.gd`` script for reference.
 
 
    func _physics_process(_delta):
+       # warning-ignore:return_value_discarded
        move_and_slide(velocity)
 
    func initialize(start_position, player_position):


mhilbrunner:
cc @NathanLovato Any idea? I have a gut feeling we should avoid having code that emits warnings in the docs in the first place, but not 100% sure what we'd do with the return here.

akien-mga:
The normal usage is `velocity = move_and_slide(velocity)`, but not sure if that's good in the context of this demo.

NathanLovato:
> cc @NathanLovato Any idea? I have a gut feeling we should avoid having code that emits warnings in the docs in the first place, but not 100% sure what we'd do with the return here.

There are several warnings that you get too often to bother. This warning pops up on every call to `connect()` and with other methods where you don't always need the return value, like `move_and_slide()`.

What I would do in this particular lesson is leave the code as-is and add a note telling readers that warnings are not errors, that they're messages that highlight something you may want to look into to help you write better code. Then, we don't need the return value in this case so you can ignore the warning.

ohnoimdead:
@NathanLovato I removed the change to the code but left the warn ignore comment example in the description. I can slim that paragraph down and remove the example comment as well, however, IMO seeing this warning as someone new to Godot makes a great teachable moment and a nice segue to the warn system docs.

Also, probably useful to point out that mob.gd in the git repo does have the warn ignore comment:
https://github.com/godotengine/godot-3d-dodge-the-creeps/blob/4706a1cd9be50653c307790310e1a34f68995177/dodge_the_creeps_3d/Mob.gd#L15

mhilbrunner (approved):

mhilbrunner:
Thanks and congrats on your first contribution to Godot's docs! 🎉

mhilbrunner:
Cherry-picked for 3.5.

PR #5760, opened 2022-04-12
Update the websockets tutorial for Godot 4's new signal connect syntax.

willstott101:
There's loads more of this to do - but I only wanted to change the tutorial which I'm actually following.


tutorials/networking/websocket.rst
```diff
@@ -44,13 +44,13 @@ This example will show you how to create a WebSocket connection to a remote serv
 
     func _ready():
         # Connect base signals to get notified of connection open, close, and errors.
-        _client.connect("connection_closed", self, "_closed")
-        _client.connect("connection_error", self, "_closed")
-        _client.connect("connection_established", self, "_connected")
+        _client.connect("connection_closed", self._closed)
```
mhilbrunner, line 47:
```suggestion
        _client.connection_closed.connect(self._closed)
```
You can avoid strings entirely in GD4 :)

willstott101: right you are, even better

PR #5762, opened 2022-04-12
Fix minor errors on PR workflow page

Dorodri:
Fixes the following mistakes:
    "major" incorrectly used in place of "minor";
    Italic code block, which generates a non code italic text.


Calinou (approved):

PR #5627, opened 2022-02-20
Improve wording and explanation on the console porting process

golddotasksquestions:
Remove the official support section, add explanation on export templates
fixes https://github.com/godotengine/godot-docs/issues/5625

PR #5764, opened 2022-04-13
Console support: Improve explanation and restructure

### LINKED ISSUE #5625: Console Exports: Due to the chosen wording, people mistakenly think out of the box support for consoles is impossible
**Your Godot version:** 3.4.2.stable

**Issue description:**
I've had this conversation now ... I don't know how many times. It's one of these things that come up like clockwork in the community channels:

A lot of people mistakenly seem to think Godot will never have or cannot have out of the box support for consoles like Unity or Unreal does due to Godot being open source.

This is of course completely wrong and merely a matter of incentive and ironing out relevant details.

It seems to me a lot of these people get the idea from this part of the docs page which I am often linked to in these discussions: 

> The reason other consoles are not officially supported are:
> 
>     To develop for consoles, one must be licensed as a company. As an open source project, Godot does not have such a legal figure.
>     Console SDKs are secret and covered by non-disclosure agreements. Even if we could get access to them, we could not publish the platform-specific code under an open source license.
>     Consoles require specialized hardware to develop for, so regular individuals can't create games for them anyway.

While they are not factually wrong, these point makes it apparently sound to many as if there is no legal possible way for Godot users to have the same convenience and cost saving of premade export templates like Unity or Unreal users have.
The text tries to explain this, but phrases like 

>we could not publish the platform-specific code under an open source license.

make is sound as if releasing platform-specific (=proprietary) code under open source license would be a necessity in order to provide Godot users the same console support as Unity or Unreal users. [This is obviously not the case.](https://www.reddit.com/r/godot/comments/svp9ll/comment/hxp3ueu/?utm_source=share&utm_medium=web2x&context=3). 

At no point Godot would ever have to open source proprietary code in order to get the same console support as Unity or Unreal. Of course there are a lot of other questions that need to be answered and probably also legal details to be discussed if console manufactures in fact have the interest to partner with Godot in some shape or form. Such as who pays for the development and maintenance of the export templates and where are they hosted. But it's far from impossible like the doc page seem to suggest for many readers.

https://docs.godotengine.org/en/stable/tutorials/platform/consoles.html#official-support

DeeJayLSP:
This is an unofficial sequel to #5627, which hasn't been updated in almost 2 months. It's supposed to fix #5625.


DeeJayLSP:
After starting the PR, I think I should suggest the Official support section should be fitted into the first one.

DeeJayLSP:
> After starting the PR, I think I should suggest the Official support section should be merged with the porting process one.

Guess I just did that

mhilbrunner (approved):

mhilbrunner:
Thanks! Merged.

PR #5771, opened 2022-04-20
Make the Vulkan SDK note more prominent in Compiling for macOS

Calinou:
`important` in Sphinx is green, but `warning` is yellow, so it's more noticeable.


mhilbrunner (approved):

mhilbrunner:
Thanks!

PR #5776, opened 2022-04-25
Update the websockets tutorial for Godot 4's new signal connect syntax.

willstott101:
There's loads more of this to do - but I only wanted to change the tutorial which I'm actually following.

N.B. replaces: #5760 cause I made a big force-push mess when using a shallow repo.


akien-mga:

PR #5781, opened 2022-04-28
updating setget to godot 4 syntax

HoustonBass:
the @tool annotation was updated, but the godot script itself was not.


akien-mga (approved):

akien-mga:
Thanks!

PR #5782, opened 2022-04-29
Step by step: Improve flow and clarity for code replacement instructions

akien-mga:
Prompted by comments on https://godotforums.org/discussion/29162/improve-godot-docs-getting-started-step-by-step-listening which highlighted potential flow issues in how the Step by Step tutorials instructs users to inject or replace code snippets.

I added more references between the tutorials and the completed script at the end of each section, to make sure that users can follow along if not familiar with programming.

Supersedes #5647 which tried to improve this but did it wrong.
Also supersedes #5648 which was good but I wanted to reword it a bit so I included it here directly while reworking all three docs.

### Notes

- I'm updating this in `master` and will do another PR for `3.4`. The code is still pretty much based on the `3.4` API aside from renaming `Sprite` to `Sprite2D`, more work will be needed to actually port this to use new APIs from Godot 4.0 (for example the signal connection syntax has changed a lot).
- For 4.0, by default I believe the scene and scripts will be given snake case names, so the `Sprite2D.tscn` and `Sprite2D.gd` references will likely have to change. This requires changing screenshots too so I didn't do this here.


getting_started/step_by_step/scripting_first_script.rst
```diff
@@ -229,6 +229,7 @@ them.
  .. code-tab:: gdscript GDScript
 
     var velocity = Vector2.UP.rotated(rotation) * speed
+
```
akien-mga, line 232: Might seem superfluous here, but in the next tutorial we replace the `var velocity = ...` line by a multi-line block and add a separation line - it's best to have it from the start IMO.

NathanLovato:
It's looking good, it's a sound improvement to make.

Now, I checked the forum post too and I think this improvement won't answer this user's needs.

The user seems frustrated by the fact that the docs assume they understand quite a bit about coding. And they do, this series is intended for people who already have decent programming foundations. I mentioned it explicitly in the intro IIRC.

I think that the writing would have to be much more detailed if you want to cater to this kind of user. We have a couple of students in this situation in the absolute beginner course we're making, and they can get confused or thrown off easily.

NathanLovato (approved):

PR #5787, opened 2022-04-29
[Net] Change server in the websocket demo.

Faless:
The old server (echo.websocket.org) is no longer active.
The new server (libwebsockets.org) is the demo server of another popular
websocket library (LWS) which we also previously used.

Closes #5460


akien-mga:
Thanks!

mhilbrunner:
Cherry-picking no longer necessary, was done manually in https://github.com/godotengine/godot-docs/pull/5885

PR #5790, opened 2022-05-01
Document zero padding and left alignment in format string

timothyqiu:
* Zero padding is also supported for real number placeholders `%f`.
* Zero padding is ignored if `-` is present.


Calinou (approved):
Thanks!

PR #4476, opened 2020-12-23
Document using `nproc` instead of hardcoded number of threads

Enhex:
on linux nproc command prints the number of virtual cores available, and putting it inside grave accents will evaluate the command and replace it with its output.

PR #5798, opened 2022-05-03
Open editor with a script

antonWetzel:
[Closed Related Proposal #4476](https://github.com/godotengine/godot-proposals/issues/4476)

Use a scripting language to start the external editor (or anything else).

The script (and shebang behavior) is not tested on linux and macOS.

### DIFF
--- a/tutorials/editor/external_editor.rst
+++ b/tutorials/editor/external_editor.rst
@@ -57,3 +57,130 @@ We have official plugins for the following code editors:
 
 - `Visual Studio Code <https://github.com/godotengine/godot-vscode-plugin>`_
 - `Emacs <https://github.com/godotengine/emacs-gdscript-mode>`_
+
+Open editor with a script
+-------------------------
+
+For advanced behavior a script can be used to open the
+external editor. The location of the script should be in a global directory.
+
+For Unix the script with a `Shebang <https://en.wikipedia.org/wiki/Shebang_(Unix)>`__
+replaces the executable path. For Windows the script interpreter replaces the executable path
+and additionally the command-line flags contain the location of the script.
+
+The following example opens a Visual Studio Code workspace if available.
+
+.. tabs::
+ .. code-tab:: python
+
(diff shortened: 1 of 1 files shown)


Calinou:
I think it'd be better to use a Python script for this, so it doesn't rely on the current Godot version installed on the system (and any bugs it may have).

antonWetzel:
@Calinou I don't know how many users have python avaible; For Windows it is not avaible by default. I could change the code block to have different tabs for Python, GDScript, (Node JS?, Deno TS?), so everyone can choose the preferred langauge.

Calinou:
> @Calinou I don't know how many users have python avaible; For Windows it is not avaible by default. I could change the code block to have different tabs for Python, GDScript, (Node JS?, Deno TS?), so everyone can choose the preferred langauge.

We shouldn't spend too much time maintaining this kind of script, so I'd vouch for including only a Python script. Installing Python on Windows is relatively straightforward by now, as long as the script only uses packages available within the standard library (no `pip` packages).

antonWetzel:
I wrote the scripts for Python, Javascript (Node) and Typescript (Deno); They all only use the standard library.
One of the languages should always be avaible.

dsnopek:
FYI, I made a script similar to @antonWetzel (in Python) back on https://github.com/godotengine/godot-proposals/issues/4476 and I just published a blog about it: https://www.snopekgames.com/tutorial/2022/how-make-godot-open-vs-code-workspace-if-project-has-one

Given how much personal preference is connected with the scripting language used AND the rules for deciding what workspace to open (I use slightly different rules than @antonWetzel because I sometimes put my *.code-workspace files in the .vscode directory to keep things tidy), I'm a little skeptical of trying to put the actual script in the official docs. This might be better in a place where folks can post comments and discuss?

antonWetzel:
Closed because low interrest.
See https://www.snopekgames.com/tutorial/2022/how-make-godot-open-vs-code-workspace-if-project-has-one for solution

PR #5799, opened 2022-05-03
Add FAQ items about portable mode and editor desktop integration

Calinou:
This was discussed on the Godot contributors chat. Note that these two FAQ questions got quite long; maybe a dedicated page should be created instead.


mhilbrunner (approved):
Looks good. I think its right on the border and maybe not _too_ long, lets merge, we can move it later if its becomes an issue. Good work! 🎉

Calinou:
Cherry-picked to the `3.5`, `3.4` and `stable` branches (with `.desktop` file URL changed to point to the `3.x` branch instead of `master`).

PR #5809, opened 2022-05-07
Fix invalid variable name in C# example in Physics introduction

vratajin:
'velocity' does not exist in that scope, does not compile


Calinou (approved):

PR #5815, opened 2022-05-14
Document that autoloads must not be freed to avoid crashes

Calinou:
See https://github.com/godotengine/godot/issues/36310.

We may also want to add a check in debug builds when freeing a node. It would check whether it's an autoload, and if so, print a warning message and prevent freeing the node to avoid a crash.

### DIFF
--- a/tutorials/scripting/singletons_autoload.rst
+++ b/tutorials/scripting/singletons_autoload.rst
@@ -99,6 +99,11 @@ you'll see the autoloaded nodes appear:
 
 .. image:: img/autoload_runtime.png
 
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):

mhilbrunner:
Good addition, thanks :)

mhilbrunner:
Cherry-picked to 3.5 (73fc6f81c7d06f347ab7a9ab074970500a896645) and 3.4 (f4f3e1926ce0943dbfc88fc135ab34a29a9ae120)

PR #5818, opened 2022-05-15
[stable] Fix resource path

okla:
Resource path for "character.png" is "res://character.png", not "res://some_texture.png". This is already fixed in master branch.


Calinou:
Pull requests against the stable branch must be opened against `3.5`. We can then cherry-pick the changes to older stable branches (including `stable`, which currently mirrors `3.4` as 3.5 stable isn't released yet).

Merged manually with https://github.com/godotengine/godot-docs/commit/3b988d8d22f5f408061872e5188f3ba3929200a0. Thanks!

PR #5833, opened 2022-05-22
Add exit condition to prevent thread from hanging in resource_queue.gd

telmotrooper:
The existing `resource_queue.gd` example present in the stable documentation (and used in @BastiaanOlij's video [Godot background loading](https://youtu.be/1oeOwPJ1XHM)) prevents your game from exiting when you call `get_tree().quit()`, because the created thread keeps on hanging after that.

The example provided in the page [Using multiple threads](https://docs.godotengine.org/en/stable/tutorials/performance/threads/using_multiple_threads.html#semaphores) provides an exit condition to the thread, which I have added to the `resource_queue.gd` example (which is the one people trying to figure out background loading will end up looking at).

I also added an `extends Node` to the beginning of the script so it can be used in autoload by default and renamed variable `sem` to `semaphore` for clarity and to keep it consistent with the "Using multiple threads" example.

### DIFF
--- a/tutorials/io/files/resource_queue.gd
+++ b/tutorials/io/files/resource_queue.gd
@@ -3,7 +3,7 @@ extends Node
 var thread
 var mutex
 var semaphore
-var exit_thread := false
+var exit_thread = false
 
 var time_max = 100 # Milliseconds.
 
--- a/tutorials/io/files/resource_queue.gd
+++ b/tutorials/io/files/resource_queue.gd
@@ -1,6 +1,9 @@
+extends Node
+
 var thread
 var mutex
-var sem
+var semaphore
+var exit_thread := false
 
 var time_max = 100 # Milliseconds.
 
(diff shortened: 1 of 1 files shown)


BastiaanOlij:
Looks like a good change to me, won't have time to try it out properly right now but code wise it seems to be a worth while improvement.

tutorials/io/files/resource_queue.gd
```diff
@@ -1,6 +1,9 @@
+extends Node
+
 var thread
 var mutex
-var sem
+var semaphore
+var exit_thread := false
```
Calinou, line 6: We don't use type hints in code samples in the documentation (for consistency with other pages):

```suggestion
var exit_thread = false
```

Calinou (approved):

Cherry-picked to the `3.5`, `3.3`, `stable` and `master` branches.

PR #5834, opened 2022-05-22
Add example for clearcoat parameter (3.5)

### LINKED ISSUE #4992: No example for clearcoat parameter on 3D spatial/standard material
**Your Godot version:**
stable

**Issue description:**
In the 3D spatial (standard) material docs page, there are useful examples for most parameters to visualize their effects. This is not the case of the clearcoat parameter.

https://docs.godotengine.org/en/stable/tutorials/3d/spatial_material.html#clearcoat

skyace65:
Adds an example for the clearcoat material parameter, also explains that it's extremely subtle in 3.x releases but this will change in Godot 4. The image has gone through png compression.


clayjohn:
To me it looks like clearcoat is enabled on the right rather than the left.

skyace65:
You're right, I got that mixed up, fixed.

mhilbrunner (approved):

mhilbrunner:
Thanks!

PR #5837, opened 2022-05-25
Fix outdated comment about GDNative on HTML5

akien-mga:


akien-mga:
Cherry-picked for 3.4 and 3.5.

PR #5838, opened 2022-05-25
Update docs about doc comments (`@desc:` replaced with blank line)

dalexeev:
* Documentation updated to reflect godotengine/godot#55099 changes.
* Added mention of `[br]`.
* Removed extra lines at the top and bottom of the comment (see Calinou's [comment](https://github.com/godotengine/godot/pull/55099#discussion_r881661464)). Since this is a change to the recommended code style conventions, tell me if you don't agree and I'll drop this part.

~~Also, in a separate PR, we need to fix the docs about supported BBCodes: they are not the same as in `RichTextLabel`.~~
I'm going to unify this so that `EditorHelp` will support all the BBCodes that the `RichTextLabel` supports.

https://github.com/godotengine/godot/blob/410b6b5f7d23638997aa7de3faf3b382c95e7fc3/editor/editor_help.cpp#L1475-L1729

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_documentation_comments.rst
+++ b/tutorials/scripting/gdscript/gdscript_documentation_comments.rst
@@ -32,8 +32,8 @@ Tags
 | Brief description | No tag and lives at the very beginning of              |
 |                   | the documentation section.                             |
 +-------------------+--------------------------------------------------------+
-| Description       | ``@desc:``                                             |
-|                   |                                                        |
+| Description       | Use one blank line to separate the description from    |
+|                   | the brief.                                             |
 +-------------------+--------------------------------------------------------+
 | Tutorial          | ``@tutorial[( The Title Here )]:``                     |
 |                   |                                                        |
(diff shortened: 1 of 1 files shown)


tutorials/scripting/gdscript/gdscript_documentation_comments.rst
```diff
@@ -185,6 +177,9 @@ be used in the documentation. Here's the list of available tags:
 +---------------------------+--------------------------------+-----------------------------------+---------------------------------------------------+
 | [signal Class.signalname] | Link to another class's signal | Emit [signal Node.renamed].       | Emit :ref:`renamed <class_node_signal_renamed>`.  |
 +---------------------------+--------------------------------+-----------------------------------+---------------------------------------------------+
+| [br]                      | Line break                     | | Line 1.[br]                     | | Line 1.                                         |
+|                           |                                | | Line 2.                         | | Line 2.                                         |
```
Calinou, line 181: There seems to be extraneous "|` before "Line 1" and "Line 2" on both columns.

dalexeev: These `|` are not an mistake, they are needed to add a line break in a table cell. Compare the difference in the result: [current](https://github.com/godotengine/godot-docs/blob/ed84099d437b1dc63692d4aeff0d188df34c4efd/tutorials/scripting/gdscript/gdscript_documentation_comments.rst) and [without extra pipes](https://github.com/dalexeev/godot-docs/blob/example-for-pipes/tutorials/scripting/gdscript/gdscript_documentation_comments.rst).

Should the line break in the Usage column be removed to make the `[br]` effect more pronounced?

Calinou: That makes sense, I wasn't aware of that rST syntax construct. Looks good to me as it is now :slightly_smiling_face:

Calinou (approved):
Thanks!

PR #5843, opened 2022-05-27
Add optimization tips for mobile in Exporting to HTML5

Calinou:
See https://github.com/godotengine/godot/issues/60914.

cc @mossfrog

### DIFF
--- a/tutorials/export/exporting_for_web.rst
+++ b/tutorials/export/exporting_for_web.rst
@@ -84,6 +84,27 @@ versions supporting WebGL 2.0. Note that Safari has several issues with WebGL
 2.0 support that other browsers don't have, so we recommend using a
 Chromium-based browser or Firefox if possible.
 
+Mobile considerations
+---------------------
+
+The Web export can run on mobile platforms with some caveats. While native
+:ref:`Android <doc_exporting_for_android>` and :ref:`iOS <doc_exporting_for_ios>`
+exports will always perform better by a significant margin, the Web export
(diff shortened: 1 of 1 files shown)


clayjohn:
We should also mention the problem with importing as sRGB in iOS 13 and below Users will need to disable sRGB and "detect 3D" as per https://github.com/godotengine/godot/issues/60914#issuecomment-1127237932

mhilbrunner:
cc @Faless just in case you have anything non-obvious to add :)

mhilbrunner (commented):
Fabio seems busy, @Calinou if you add the above, I'm happy with merging as is. Looks good to me!

Calinou:
> Fabio seems busy, @Calinou if you add the above, I'm happy with merging as is. Looks good to me!

I still need to mention the sRGB workaround for iOS < 13. Note that disabling Detect 3D will also disable mipmap generation, so it'll have to be explicitly enabled by the user to prevent textures from looking grainy in 3D.

mhilbrunner:
@Calinou Friendly bump. :)

Calinou:
Rebased and updated for 4.3.stable. The iOS 13/15 advice is no longer relevant nowadays, so I removed it. It was most likely fixed on Godot or Apple's side with the new Compatibility renderer in 4.0, and iOS 13 is EOL now.

skyace65 (approved):

skyace65:
Thanks!

PR #5844, opened 2022-05-27
Fix class name : change "string" to "String"

dtesniere:
Change case of "s" letter : "string" does not compile but "String" does (for gdScript)


Calinou:

dtesniere:
Thanks for your benevolence and your help
I read quickly the documentation (and the video).
I hope that I've done right this time, but I don't see many rule about minor fixes...

Issue #5845, opened 2022-05-28
State that godot-headers repo requires checkout to specific branch


quantumedbox:
**Issue description:**
`godot-header` prerequisites are written in such way that using `master` branch seems appropriate, while currently it is being rewritten to GDExtension and has no GDNative headers at all, thus confusing new readers.
Saw some people in godot discord server asking "im following the tutorial along, but why compiler couldn't find headers?".

IMO checkout to branch you're targetting should be mandatory for progressing the guide, and not just mentioned.

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/stable/tutorials/scripting/gdnative/gdnative_c_example.html

tetrapod00:
Closed by https://github.com/godotengine/godot-docs/pull/5846

PR #5846, opened 2022-05-28
Point to stable branch of godot-headers in GDNative C example

Calinou:
This closes https://github.com/godotengine/godot-docs/issues/5845.


aaronfranke (approved):
We could also update this for 3.5 in like a month.

PR #5853, opened 2022-06-01
Update class_vector3.rst

CristianoSimao:
Many errors in collisions or vector positions ocur because the Vector3 not store the right values, see this bugs report: https://github.com/godotengine/godot/issues/43476

If some explanation is given in the documentation will help people avoid many time lost trying find the bug, and discovering that Godot like imprecise numbers to calculate position.

If anyone can improve the text it would be great.


Calinou:

PR #5859, opened 2022-06-06
Clarifies that drag preview is made child of node

### LINKED ISSUE #5858: Clarify that 'set_drag_preview' method of Control makes the preview the child of the called Control
**Your Godot version:**
v3.4

**Issue description:**
The `set_drag_preview` method on the Control documentation should probably explicitly say that the preview you hand it will become the child of the control that is being called. The phrase 'The control must not be in the scene tree' misled me to believe that the control is never in the scene tree; perhaps it's obvious to most people, but clarification would be nice!

https://docs.godotengine.org/en/stable/classes/class_control.html#class-control-method-set-drag-preview

linky00:
Closes #5858

### DIFF
--- a/classes/class_control.rst
+++ b/classes/class_control.rst
@@ -2346,7 +2346,7 @@ Forwarding can be implemented in the target object similar to the methods :ref:`
 
 - void **set_drag_preview** **(** :ref:`Control<class_Control>` control **)**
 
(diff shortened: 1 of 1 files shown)


YuriSizov:
This is factually incorrect. The preview is added as a child of the topmost related ancestor (parent, grand parent, etc) control. We can merge your other clarification though, it doesn't hurt.

YuriSizov (changes requested):
See above.

YuriSizov:
Oh, actually, dismiss this. Since you're changing class reference, you must submit a PR against the main repository.

https://github.com/godotengine/godot/blob/master/doc/classes/Control.xml

PR #5876, opened 2022-06-14
Improve side-aligned image positioning

and-rad:
This PR improves how images are displayed when they are left- or right-aligned and are closely following other similarly aligned images.

Images that are aligned to either side have their CSS `float` property set to either `left` or `right`. Text floats around them. A problem may arise if there is not enough text between two such images to fully fill out the space between. In that case, the second image will be placed next to the other image and somewhat below it. See the **Before** example below. This is how `float` behaves.

To prevent this, CSS provides the `clear` property. If set, floating elements will not be placed next to other floating elements depending on which `clear` value is set. Options are `left`, `right`, and `both`. This PR uses the latter to make sure that all floating images appear below each other and never next to each other. See the **After** example below.

## Before:
before

## After:
after

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -217,6 +217,7 @@ legend,
     font-weight: 500;
 }
 
+/* See <https://github.com/godotengine/godot-docs/pull/5876> for context. */
 .rst-content .align-right,
 .rst-content .align-left {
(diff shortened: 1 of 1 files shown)


Calinou (approved):
Thanks!

PR #5882, opened 2022-06-17
Update OS feature tags list with new architecture features and OS names in 4.0

aaronfranke:
This PR updates the list of OS feature tags with the recent changes to architecture tags in https://github.com/godotengine/godot/pull/61732. It also updates the OS name tags to be consistent with 4.0: it removes the `Server` tag, updates the casing to be lowercase, updates the `macos` tag since it was renamed from `OSX` (see https://github.com/godotengine/godot/pull/52291), and it adds `linuxbsd`.

~~I'm not sure where the casing matters for `LinuxBSD`, since `OS.has_feature()` just checks for `"linuxbsd"` after converting to lowercase... @Calinou Can you explain where the casing matters, since it doesn't matter in `OS.has_feature()`? In https://github.com/godotengine/godot/pull/47391 you said "Feature tag names are still case-sensitive".~~

I also documented how GDExtension uses the feature tags.


Calinou:
> Can you explain where the casing matters, since it doesn't matter in OS.has_feature()? In https://github.com/godotengine/godot/pull/47391 you said "Feature tag names are still case-sensitive".

I was referring to the `OS.has_feature()` check still being case-sensitive, which is especially important for projects ported from Godot 3.x, or for projects using custom feature tags on export.

mhilbrunner:
Thank you! 🎉

PR #5888, opened 2022-06-22
Update index.rst

gh0sstt:
Fix a grammatical error, it should be "download a HTM"L copy not "download an html copy"


clayjohn:
"an" is actually the correct word here. While HTML starts with an "H", its pronunciation starts with a vowel-sound, so "an" is the correct word to use.

Calinou:
Closing per @clayjohn's comment. Thanks for the contribution nonetheless!

PR #5903, opened 2022-07-03
correct a little typo on scripting_laguanges

Santiagodval:
the line 27 has a redundand "and": "GDScript, C#,
VisualScript, and C++ and C" it should be "GDScript, C#,
VisualScript, C++ and C"

### DIFF
--- a/getting_started/step_by_step/scripting_languages.rst
+++ b/getting_started/step_by_step/scripting_languages.rst
@@ -24,7 +24,7 @@ Available scripting languages
 -----------------------------
 
 Godot offers **five gameplay programming languages**: GDScript, C#,
-VisualScript, and C++ and C via its GDNative technology. There are more
+VisualScript, C++ and C via its GDNative technology. There are more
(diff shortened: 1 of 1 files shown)


DeeJayLSP:
I believe that's supposed to make clear that both C and C++ are supported through GDNative. If so, another way to clarify should be done, like `"...and, via its GDNative technology, C and C++."`

Santiagodval:
> I believe that's supposed to make clear that both C and C++ are supported through GDNative. If so, another way to clarify should be done, like `"...and, via its GDNative technology, C and C++."`

I agree.

Santiagodval:
I made a commit with the suggestion of DeeJayLSP instead

mhilbrunner (approved):

mhilbrunner:
Seems like an improvement indeed. Thanks and congrats on your first small contribution to the docs :) 🎉
We should also replace 'GDNative' with 'GDExtension' for 4.0/master, but this can be a followup, this one should be cherrypicked for 3.x.

mhilbrunner:
Cherry-picked to 3.4 and 3.5.

PR #5908, opened 2022-07-04
Simplify `.gitattributes` in Submitting to the Asset Library

nathanfranke:
Context: https://github.com/godotengine/godot-proposals/issues/3648

### DIFF
--- a/community/asset_library/submitting_to_assetlib.rst
+++ b/community/asset_library/submitting_to_assetlib.rst
@@ -80,27 +80,30 @@ library a better place for all users.
 * Consider adding a **.gitattributes** file to your repo. This file allows
   giving extra instructions to Git, such as specifying line endings and listing
   files not required for your asset to function with the ``export-ignore``
-  directive. This directive removes such files from the resulting ZIP file
-  and prevents them from being downloaded by the asset library users.
-  For a typical plugin **.gitattributes** may look like this:
+  directive. This directive removes such files from the resulting ZIP file,
+  preventing them from being downloaded by the asset library users.
+  These are common examples of **.gitattributes**:
 
-  .. code-block:: none
+  .. tabs::
 
-    # Normalize EOL for all files that Git considers text files.
-    * text=auto eol=lf
+   .. tab:: Projects / Templates
 
-    # Ignore some files when exporting to a ZIP.
-    /.gitattributes     export-ignore
-    /.gitignore         export-ignore
-    /LICENSE            export-ignore
-    /LICENSE.md         export-ignore
-    /README.md          export-ignore
-    /project.godot      export-ignore
-    /icon.png           export-ignore
-    /icon.svg           export-ignore
+      .. code-block:: shell
 
-  Other types of assets may require a different configuration (e.g.
-  a project template requires **project.godot**).
+        # Normalize line endings for all files that Git considers text files.
+        * text=auto eol=lf
+
+   .. tab:: Addons / Asset Packs
+
+      .. code-block:: shell
+
+        # Normalize line endings for all files that Git considers text files.
+        * text=auto eol=lf
+
+        # Only include the addons folder when downloading from the Asset Library.
+        /**        export-ignore
+        /addons    !export-ignore
+        /addons/** !export-ignore
 
 * If you are submitting a plugin, add a **copy** of your license and readme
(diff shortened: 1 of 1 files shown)


aaronfranke (approved):

YuriSizov:
Asset is not necessarily an addon, so this is not a good change.

aaronfranke:
@YuriSizov The existing `.gitattributes` on the page is already an example for an addon, since it prevents `project.godot` from being exported. Maybe this should be made more clear, or we should do better at explaining that this example is for addons, and for projects they should not include any of the `export-ignore` lines.

YuriSizov:
@aaronfranke Yes, I agree that the preceding text puts this example in the context of a plugin/addon. In that sense, this change is fine. But I also like that the current example gives an extra reminder that you probably don't want to include project files, and screenshots, and git metafiles. An asset can be simply a collection of reusable scenes, like a custom control or a textured/scripted game object. So it's not an addon, and it's not a template project.

But either way, that's just my thoughts. I'm not going to block this PR if no adjustments are made.

gumaciel:
I think this change is valid.

I was thinking of having two types of `.gitattributes` to ignore some files when doing the `.zip`: One for Addons and one for Projects, but I don't know what `.gitattributes` for Projects would be like

aaronfranke:
@gumaciel For projects, the `.gitattributes` file should not export-ignore anything, so the proper file looks like this:

```gitattributes
# Normalize EOL for all files that Git considers text files.
* text=auto eol=lf
```

nathanfranke:
I can update the comment to be more clear. Something like:

# For addons, only export the addons folder. This is not needed for asset packs or projects.

nathanfranke:
Updated to address Yuri's concern. Also changed `EOL` to `line endings` as it doesn't take much more space and is easier for people who don't understand what it means.

community/asset_library/submitting_to_assetlib.rst
```diff
@@ -86,18 +86,13 @@ library a better place for all users.
...
-    # Ignore some files when exporting to a ZIP.
-    /.gitattributes     export-ignore
-    /.gitignore         export-ignore
-    /LICENSE            export-ignore
-    /LICENSE.md         export-ignore
-    /README.md          export-ignore
-    /project.godot      export-ignore
-    /icon.png           export-ignore
-    /icon.svg           export-ignore
+    # For addons, only include the addons folder. This is not needed for asset packs or projects.
```
aaronfranke, line 92:
```suggestion
    # For addons, only include the addons folder when exporting a ZIP.
    # These lines should be deleted for asset packs or projects.
```

nathanfranke: I don't think the average asset maker knows that the asset library uses GitHub ZIPs internally.

aaronfranke: Attempt 2:

```suggestion
    # For addons, only include the addons folder when downloading from the Asset Library.
    # These lines should be deleted for asset packs or projects.
```

aaronfranke (approved):

YuriSizov:
Not sure if this change addresses my concerns, but like I've said, I'm not blocking this in any way (hence why I didn't leave my concerns as a review). But if you are going to add extra instructions, you should do it in text, not in comments of code samples. Comments are not-translatable in our documentation, so this makes them hard to be useful for people reading documentation in languages other than English.

The old file also used comments, but those weren't really aimed at instructing the reader. This is, however:

> These lines should be deleted for asset packs or projects.

nathanfranke:
Updated to use code tabs :slightly_smiling_face: 
image

mhilbrunner:
Maybe @Calinou or someone else who spends more time than me around the Asset Lib can take a glance at this one too :)

nathanfranke:
Should we change the tab to `Projects`?

Calinou:
> Should we change the tab to `Projects`?

I would prefer changing it to "Projects / Templates / Demos" if we have the space, but "Projects" is OK too.

nathanfranke:
> Asset packs should be distributed as add-ons, preferably in the `addons/` subfolder for organization purposes (although there is no hard rule).

Then let's be clear about it :slightly_smiling_face: 
image

Calinou (approved):
Thanks!

PR #5909, opened 2022-07-04
update pull request template to be similar to godotengine/godot

nathanfranke:
\

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,8 +1,7 @@
 <!--
-**Note:** Pull Requests should be made against the `master` by default.
+Please target the `master` branch in priority.
+PRs can target other branches if the same change was done in `master`, or is not relevant there.
 
-Only make Pull Requests against other branches (e.g. `2.1`) if your changes only apply to that specific version of Godot.
-
-The type of content accepted into the documentation is explained here: https://docs.godotengine.org/en/latest/community/contributing/content_guidelines.html
-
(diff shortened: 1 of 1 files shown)


aaronfranke (approved):
See this PR in the main repo: https://github.com/godotengine/godot/pull/62486

YuriSizov (changes requested):
We don't have 3.x here, you must use a specific version branch, the one that corresponds to the latest applicable version. So the text must be adjusted.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,8 +1,7 @@
 <!--
-**Note:** Pull Requests should be made against the `master` by default.
-
-Only make Pull Requests against other branches (e.g. `2.1`) if your changes only apply to that specific version of Godot.
-
-The type of content accepted into the documentation is explained here: https://docs.godotengine.org/en/latest/community/contributing/content_guidelines.html
+Please target the `master` branch in priority.
+PRs can target other branches if the same change was done in `master`, or is not relevant there.
```
YuriSizov, line 3:
```suggestion
PRs can target other branches (e.g. `3.2`, `3.5`) if the same change was done in `master`, or is not relevant there.
PRs must not target `stable`, as that branch is updated manually.
```

I'd still keep an example, and also mention that stable is not to be used.

YuriSizov (changes requested):
Slight adjustment and another note worth pointing out, and then I think this is fine.

YuriSizov (approved):

YuriSizov:
Thanks!

PR #5912, opened 2022-07-04
Update c_sharp_basics.rst

Xerandael:
I spent like an hour finding this.  I think it should be in the docs.  My builds broke mysteriously even though I had the SDK.


skyace65:
When you say your builds broke mysteriously do you mean they were building then stopped? And did you make any changes to your system that might have caused this?

I ask because I just tested this by deleting my editor settings to generate new ones, and dotnet CLI is being selected as the build tool by default on Windows.

Xerandael:
I'd been able to build, then I'd fiddled with Visual Studio for unrelated reasons.  I think installing VS affected this setting automatically.

It took me some time to find the issue because the error message says that .NET isn't available (it was), which sent me on a goose chase.

Maybe this documentation edit isn't the right fix.
________________________________

raulsntos:
Superseded by #5983.

PR #5915, opened 2022-07-05
Update the 05.the_main_game_scene.rst

Santiagodval:
The page is unclear on the way it presents those functions, personally lost an hour trying to solve this problem: the timeouts were not connected. The page should mension it on some form, I listen to suggestions on how and/or where to add it.


cbscribe:
This is not an appropriate place to put this. You can't connect the Timer signals until the script has been added to Main, which happens in the next section. In that section we already have this:

> Now connect the ``timeout()`` signal of each of the Timer nodes (``StartTimer``,
``ScoreTimer`` , and ``MobTimer``) to the main script. ``StartTimer`` will start
the other two timers. ``ScoreTimer`` will increment the score by 1.

PR #5922, opened 2022-07-11
[3.x] Fix click-and-move example

dalexeev:
Added `target` initialization to prevent character from moving to `(0,0)` on startup.


skyace65 (approved):
I've tested the original C# and GDScript code to confirm the issue exists and confirmed the new code fixes it. Good Job!

mhilbrunner:
Thanks for the fix and thanks to @skyace65 for testing and confirming :)

Issue #5928, opened 2022-07-13
Improve "Compiling with script encryption" instructions to make compilation requirement clearer


akien-mga:
**Your Godot version:**
3.x

**Issue description:**
Users who try to use script encryption often miss critical requirements from this page that they need to compile Godot export templates from source.

The page should be improved to make this much more prominent, link to relevant pages on how to compile Godot, explain why the encryption key is set in an environment variable, how to make sure that it has indeed been included in the compiled template, how to use said template, etc.

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/stable/development/compiling/compiling_with_script_encryption_key.html

Calinou:
> how to make sure that it has indeed been included in the compiled template

Does SCons have anything in place for this? If not, I'd make it print a message when compiling (without printing the key's contents, of course).

mhilbrunner:
Reopened, see discussion in https://github.com/godotengine/godot-docs/pull/5930.

PR #5930, opened 2022-07-13
Improve the Compiling with PCK encryption key page

Calinou:
`master` version of https://github.com/godotengine/godot-docs/pull/5931.

This closes https://github.com/godotengine/godot-docs/issues/5928.

### DIFF
--- a/development/compiling/compiling_with_script_encryption_key.rst
+++ b/development/compiling/compiling_with_script_encryption_key.rst
@@ -5,23 +5,31 @@ Compiling with PCK encryption key
 
 .. highlight:: shell
 
-The export dialog gives you the option to encrypt your PCK file with an 256bit
-AES key, when releasing your game. This will make sure your scenes, scripts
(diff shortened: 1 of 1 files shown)


skyace65 (commented):
There's nothing wrong with the changes here, and they are an improvement. But this doesn't fix all of the problems in the issue this is meant to close. I'm not seeing anything for the following. "link to relevant pages on how to compile Godot, explain why the encryption key is set in an environment variable, how to make sure that it has indeed been included in the compiled template, how to use said template, etc."

Calinou:
@skyace65 I forgot about that, apologies. There are probably some things we should improve in the SCons setup before revising this documentation page.

mhilbrunner (approved):

mhilbrunner:
I think these changes are good as they are; they are an improvement. We can let the issue stay open for now as a reminder more changes may be needed. :)

PR #5939, opened 2022-07-17
Update consoles.rst: Added two new porting/publishing companies

panreyes:
Update consoles.rst: Added two new porting/publishing companies: Flynn's Arcades and RAWRLAB games (my own company).
Donut Dodo is the first of them: https://twitter.com/FlynnsArcades/status/1545087739233812481
And there are two more in the way :)

### DIFF
--- a/tutorials/platform/consoles.rst
+++ b/tutorials/platform/consoles.rst
@@ -67,6 +67,10 @@ Following is the list of providers:
   Switch and PS4 porting and publishing of Godot games.
 - `Pineapple Works <https://pineapple.works/>`_ offers
   Switch, Xbox One and Xbox Series X/S porting and publishing of Godot games.
+- `Flynn's Arcade <https://www.flynnsarcades.com/>`_ offers
+  Switch porting and publishing of Godot games.
+- `RAWRLAB games <https://www.rawrlab.com/>`_ offers
+  Switch porting of Godot games.
 
 If your company offers porting and/or publishing services for Godot games,
 feel free to


mhilbrunner (approved):

mhilbrunner:
Welcome, congrats on your first small docs contribution and good luck with the company! :)

wkubiak:
@panreyes just to clarify - RAWRLAB does the porting and Flynn's Arcade publishes games, right?

panreyes:
> @panreyes just to clarify - RAWRLAB does the porting and Flynn's Arcade publishes games, right?

Currently Flynn's Arcades delegates the porting of Godot games to RAWRLAB Games (this may change in the future), but both companies offer this service to their final clients.

wkubiak:
@panreyes thanks for clarifying.

There's a difference in offering a service and actually being capable of offering it.
Any publisher can offer something and then outsource it to some other company.

Because of this, IMHO Flynn's Arcade shouldn't be listed as a company that can offer Godot Engine console ports. By your own admission, they outsource the work to RAWRLAB and that's how it is today.

The alternative to the above would mean that ANY publisher who has published Godot Engine games on console platforms (or plans to do so, like Flynn's Arcade) should be added to the list, even if they don't have the technical capability of producing a Godot Engine port on their own.

panreyes:
I see your point. It may be a good idea to add a "Godot for consoles publishers list".

clayjohn:
The purpose of this doc is to put users in touch with people who can assist them in getting their games ported to consoles. In my opinion this includes publishers who can assist with the full end-to-end process of obtaining a console developer licence, arranging for the game to be ported, and publishing.

wkubiak:
@clayjohn what about the claim that a company is a provider of Godot Engine ports, but in actuality they just use the services of another company from the list - does this still make them a provider of porting services?

If yes than basically anybody can add their name here and say that they offer porting services for Godot, provided that they know anybody who actually can do it, which makes the former's claim kind of farfetched.

wkubiak:
Getting back to the Flynn's Arcade issue, a part of the doc page reads:

> Console ports of Godot are offered by third-party companies (which have ported Godot on their own)

This implies that the list should comprise mainly of companies who are actually able to port Godot to consoles and not entities who only know of or rely on others to do the actual job, which basically repeats the argument I already stated previously.

Adding more weight to what has already been said, we have precedents in past edits to this very doc page where things which proved to be untrue became corrected.

So why should an exception be made now?

Calinou:
Note: Should be cherry-picked at the same time as https://github.com/godotengine/godot-docs/pull/6066.

panreyes:
Flynn's Arcade and RAWRLAB Games are both Nintendo Switch developers, they both have access to the same Godot port and they both have the technical capabilities and knowledge to do the porting in-house. Both of them are offering this service.

I, Pablo Navarro, work for Flynn's Arcade but also have my own company: RAWRLAB Games.

If it helps to solve this issue, we will change the way we work with one of our Godot games to consider it an in-house port.

wkubiak:
@panreyes I think you guys should just decide between yourselves under which name/brand do you want to offer the porting service, because this is quite convoluted as it is.
When things change in the future, you can always make another edit.

Issue #5630, opened 2022-02-22
Add a link to the documentation section of the contributing page under "edit on github" links


skyace65:
This was brought up on Reddit. Anyone who hasn't contributed to the docs before and clicks on the "edit on github" link isn't given any information on how exactly to contribute, just the code. I think a link to [here](https://docs.godotengine.org/en/latest/community/contributing/index.html#writing-documentation) should be good. Title it "Docs writing guide" or something like that.

NaincyKumariKnoldus:
Hey @skyace65 Can I do this ?

skyace65:
> Hey @skyace65 Can I do this ?

Go for it

akien-mga:
To clarify, this is about adding something here:
![image](https://user-images.githubusercontent.com/4701338/157420987-251ff9bb-1eaa-467a-9d31-df23ae0f09d1.png)

That's not easy to do as this is added by the Sphinx theme, and might require editing the theme.

YuriSizov:
> That's not easy to do as this is added by the Sphinx theme, and might require editing the theme.

Actually, this is pretty trivial 👀 RTD & Sphinx have a way to override partial templates, and extend them. And we even override the relevant one already.

![image](https://user-images.githubusercontent.com/11782833/179547956-2eb7538f-e360-4d47-8535-fc869e76eeb0.png)

I'm not sure if this is a bit too much though, so design suggestions are welcome!

PR #5947, opened 2022-07-18
Add a link to doc writing guidelines to the header

YuriSizov:
Closes https://github.com/godotengine/godot-docs/issues/5630 though I'm not sold on the design myself. Suggestions are welcome.

image


_templates/breadcrumbs.html
```diff
@@ -3,5 +3,21 @@
...
+
+<style>
+	.godot-edit-guidelines {
+		float: right;
+		clear: both;
+	}
+
+	@media screen and (max-width: 480px) {
+		.godot-edit-guidelines {
+			display: none;
```
and-rad, line 15: Why hide it on mobile phones?

YuriSizov: Because the edit button is hidden at that resolution as well.

and-rad: Would it make sense to treat them differently? Editing anything on GitHub on mobile sounds like an absolute pain, but the guidelines are just another page in the docs, right?

YuriSizov: I don't think so. This link is an extra for the "Edit on GitHub" button, it's the only reason to add it on every page. And 480px is only a portion of mobile users. Many if not most phones have bigger resolution than that, and so they'll have both the button and the link available.

and-rad: Good point!

YuriSizov:
image

Better?

Calinou:
> Better?

Looks perfect now :slightly_smiling_face:

Calinou (approved):

YuriSizov:
All good now

mhilbrunner:
Cherry-picked to 3.4 and 3.5.

PR #5948, opened 2022-07-18
Add the top navigation bar

YuriSizov:
So I made this thing...

image

_(The learn how to contribute thing is from https://github.com/godotengine/godot-docs/pull/5947, unrelated)._

I got rather annoyed having to scroll through the side bar every time I need to find something. Using the search is slow and unreliable, so browsing the sidebar is usually faster. But it's still a sidebar, tiny and constricted. So I thought that maybe we could have top navigation for sections. And so I made it. It's not even hacky, as editing and extending templates is supported by RTD and Sphinx.

However, I'm not sure it is useful in the end. First of all, most of those links lead nowhere... We just don't have landing pages for sections as there is no normal way to navigate there. If we did have them, we don't really have any content for them, other than tables of content. And putting TOCs there makes them hardly any more useful than the sidebar.

So, maybe we could actually create interesting landings for those sections, add some nice looking blocks for tutorial categories, add descriptions? This PR covers the technical part of adding the navigations, so it's only a matter of coming up with what content we can push there.

Marking this as draft as I'm not certain about this feature yet, but I'd like to use this PR for discussion.

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -1093,3 +1093,29 @@ kbd.compound > .kbd,
     border-left-color: var(--code-example-bad-color);
     border-left-width: 8px;
 }
+
+/* Top-level navigation. */
+.godot-menu-top {
+    border-bottom: 1px solid var(--hr-color);
+    margin: 0 -8px 16px -8px;
+    padding-bottom: 8px;
+}
+
+.godot-menu-top ul {
+    display: flex;
+    justify-content: space-between;
+}
+
+.godot-menu-top li {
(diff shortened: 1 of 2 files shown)


mhilbrunner:
We could make the headers in the sidebar clickable and link to the same landing pages if possible. It would kinda make sense to have a landing page per section instead of linking to the random first page/section in that area?

Would be nice to list general things. Overall, I'm not really sold we need to double the navigation in 2 places because search + current navigation are bad though. It may be better to improve the current navigation and search than to add yet another place.

PR #5955, opened 2022-07-19
Make sidebar categories collapsible

YuriSizov:
An alternative for https://github.com/godotengine/godot-docs/pull/5948. I like it more.

https://user-images.githubusercontent.com/11782833/179815915-9366498c-cabd-43f9-af9d-6c049664a3c1.mp4

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -17,6 +17,7 @@
     --navbar-current-background-color: #212d51;
     --navbar-current-background-color-hover: #182343;
     --navbar-current-background-color-active: #131e3b;
+    --navbar-category-active-color: rgba(255 115 129 / 10%);
     --navbar-level-1-color: #c3e3ff;
     --navbar-level-2-color: #b8d6f0;
     --navbar-level-3-color: #a3c4e1;
@@ -107,6 +108,7 @@
         --navbar-current-background-color: #333639;
         --navbar-current-background-color-hover: #44474a;
         --navbar-current-background-color-active: #222528;
+        --navbar-category-active-color: rgba(238 115 129 / 10%);
         --navbar-level-1-color: #ddd;
         --navbar-level-2-color: #ccc;
         --navbar-level-3-color: #bbb;
@@ -934,6 +936,10 @@ kbd.compound > .kbd,
     background-color: var(--navbar-background-color-active);
 }
 
+.wy-menu-vertical li.toctree-l1 > a {
+    padding: .4045em 1.918em;
+}
+
 .wy-menu-vertical li.toctree-l1.current > a {
     border: none;
 }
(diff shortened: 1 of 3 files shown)


mhilbrunner:
Is the current category always un-collpased? I.e. if I'm landing on a page from Google? Likewise, maybe the first category should be expanded by default when on the start page?

YuriSizov:
> Is the current category always un-collpased? I.e. if I'm landing on a page from Google?

Yes, the category that the page belongs to is expanded by default.

> Likewise, maybe the first category should be expanded by default when on the start page?

Well, sure, I don't mind but it also doesn't really matter IMO.

-----

It does look a bit idle and not clear that you can interact with the titles on the main landing page, that is true. But as several people suggested on Twitter, I should probably add an arrow/chevron/plus character to indicate that they are expandable.

mhilbrunner:
> It does look a bit idle and not clear that you can interact with the titles on the main landing page, that is true. But as several people suggested on Twitter, I should probably add an arrow/chevron/plus character to indicate that they are expandable.

I agree with that, besides this does look cleaner IMO :)

YuriSizov:
Added an arrow to better indicate that sections can be toggled. The video in the OP is updated.

Some people on Twitter complain that auto-hiding other sections when you open another one is not something they'd enjoy. Opinions? I think it's okay to auto-hide, as it keeps the list focused and you don't really need to work with multiple sections at the same time. You need to get to the article you want as fast as possible.

Calinou:
> Some people on Twitter complain that auto-hiding other sections when you open another one is not something they'd enjoy. Opinions? I think it's okay to auto-hide, as it keeps the list focused and you don't really need to work with multiple sections at the same time. You need to get to the article you want as fast as possible.

I'm neutral about this. I much prefer accordion setups that don't automatically fold other sections when you unfold one of them, but our setup involves a really long list. So I think this behavior makes sense here.

Worst of all, we can change it to not automatically fold other sections if there's significant demand.

mhilbrunner:
I'd also like Akien to give this a glance, but I agree with both of you and think this is a good improvement.

_static/css/custom.css
```diff
@@ -17,6 +17,7 @@
     --navbar-current-background-color: #212d51;
     --navbar-current-background-color-hover: #182343;
     --navbar-current-background-color-active: #131e3b;
+    --navbar-category-active-color: rgba(255 115 129 / 10%);;
```
akien-mga, line 20: Double semicolon.

akien-mga (approved):
Seems good to me.

I'm slightly concerned about hiding contents by default as it makes it harder to actually see that there are tutorials that match a given topic of interest (e.g. "Networking", "XR", etc.).

It also highlights issues with existing sections:
- "General" is empty apart from the "About" subsection
- "Development" can be misleading, so should likely be "Engine development"
- "Class Reference" only has "Godot API" subsection, maybe the classes could be toplevel under the section

YuriSizov:
Updated based on suggestions.

1. "General" is now collapsed into "About" (the dedicated "About" page still exists for those who may have stored a direct link to it).
2. "Development" is not "Engine Development". This does collide with its own subsection, though. Can clean it up later, I suppose, as this doesn't affect any links (the link for that subsection is "development/cpp").
3. On the home page the "About" section is unfolded by default. This also acts as a nice hint that you can unfold other sections to see their content.

I've not touched the class reference section, as we already have a different PR that would split those classes into several categories, and that solves the ambiguous "Godot API" problem.

chrome_2022-11-17_20-49-22

akien-mga:
Thanks!

Sauermann:
After this change, it was necessary for me to update Sphinx. Sphinx 4.0.2 didn't have the required functionality and I couldn't open the categories after compiling locally.
A `pip3 install -r requirements.txt` did the trick.

YuriSizov:
This is a purely frontend solution, it shouldn't depend on anything from sphinx, other than the layout of the page. So yeah, a good idea to verify that your local install matches the requirements.

PR #5956, opened 2022-07-19
missing @ hint on export (typo?)

ego-edesse:
Updating to GDScript 2.0 syntax

### DIFF
--- a/getting_started/first_2d_game/03.coding_the_player.rst
+++ b/getting_started/first_2d_game/03.coding_the_player.rst
@@ -30,7 +30,7 @@ Start by declaring the member variables this object will need:
 
     extends Area2D
 
-    export var speed = 400 # How fast the player will move (pixels/sec).
+    @export var speed = 400 # How fast the player will move (pixels/sec).
     var screen_size # Size of the game window.
 
  .. code-tab:: csharp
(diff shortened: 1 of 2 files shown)


YuriSizov:
It would be good to do a pass on the docs and update all cases where annotations should be used now, or at least do it in bulk.

Calinou:
This is not a typo – it's outdated Godot 3.x code. `@export` is an annotation in 4.0, while it was a keyword in 3.X.

ego-edesse:
I thought it was a typo because further in the tutorial it has been updated to GDScript 2.
It was my first PR on Godot project, thats why I didn't make anything big. I'll try to change more then

YuriSizov:
> I thought it was a typo because further in the tutorial it has been updated to GDScript 2.

Well, that's why it would be nice to do a proper pass 🙃 As you can see, if someone changes only some of the outdated syntax, it gets confusing and harder to track afterwards. If you want to take time and look through other related articles, that would be amazing and a great way to contribute!

ego-edesse:
> 

Without being a complete beginner, my knowledge of GDScript is not enough to be able to check pages and be sure I am not leaving anything out. I may come back later when I feel more comfortable but I guess we can already accept these two small commits

mhilbrunner (approved):

mhilbrunner:
Thank you! Sorry this took a bit to get merged :)

PR #5961, opened 2022-07-19
Update the showcase video links

Windvis:
~~This updates the showcase video links to the most recently released version.~~

~~Alternatively we can maybe link to the [showcase playlist instead](https://youtube.com/playlist?list=PLeG_dAglpVo6EpaO9A1nkwJZOwrfiLdQ8) so the links don't need to be updated?~~

Update: This now links to the showcase playlist instead.


Calinou:
> Alternatively we can maybe link to the [showcase playlist instead](https://youtube.com/playlist?list=PLeG_dAglpVo6EpaO9A1nkwJZOwrfiLdQ8) so the links don't need to be updated?

Indeed, I think this is a better idea.

mhilbrunner (approved):

PR #5962, opened 2022-07-19
Increase spacing between list items to improve readability

Calinou:
The spacing may be a bit high for single-line items, but it's definitely welcome for multi-line items.

## Preview

*Part of https://github.com/Calinou/godot-docs/tree/revamp-gi-docs, which uses a lot of lists like those.*

Before | After
-|-
 | image

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -251,6 +251,12 @@ article ol,
     line-height: 25px;
 }
 
+.rst-content section ul li {
+    /* Increase spacing between list items. */
+    margin-top: 8px;
+    margin-bottom: 8px;
+}
+
 body,
 .rst-content table.docutils thead {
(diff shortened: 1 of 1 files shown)


mhilbrunner (commented):
Personally, I think I might prefer something between the two? I kinda agree these could stand some larger spacing, but the change is a little too much for me IMO.

Calinou:
> Personally, I think I might prefer something between the two? I kinda agree these could stand some larger spacing, but the change is a little too much for me IMO.

I reduced the spacing from `10px + 10px` to `8px + 8px`.

skyace65 (approved):

Calinou:
Cherry-picked to the `3.5`, `3.4`, `stable`, `3.3`, `3.2` and `3.1` branches.

PR #5970, opened 2022-07-22
Compiling for Windows: Clarify that VS is Windows only

akien-mga:
See https://github.com/godotengine/godot/issues/63305.

### DIFF
--- a/development/compiling/compiling_for_windows.rst
+++ b/development/compiling/compiling_for_windows.rst
@@ -169,12 +169,13 @@ dependencies. Running it will bring up the Project Manager.
           :ref:`doc_data_paths_self_contained_mode` by creating a file called
           ``._sc_`` or ``_sc_`` in the ``bin/`` folder.
 
-Development in Visual Studio or other IDEs
-------------------------------------------
+Development in Visual Studio
+----------------------------
 
-For most projects, using only scripting is enough but when development
-in C++ is needed, for creating modules or extending the engine, working
-with an IDE is usually desirable.
+Using an IDE is not required to compile Godot as SCons takes care of everything,
+but if you intend to do engine development or debugging of the C++ code on a
+Windows development environment, you may be interested in configuring things
+for use in Visual Studio.
 
 You can create a Visual Studio solution via SCons by running SCons with
 the ``vsproj=yes`` parameter, like this::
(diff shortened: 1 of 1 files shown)


development/compiling/compiling_for_windows.rst
```diff
@@ -184,11 +185,7 @@ the ``vsproj=yes`` parameter, like this::
 You will be able to open Godot's source in a Visual Studio solution now,
 and able to build Godot using Visual Studio's **Build** button.
 
-If you need to edit the build commands, they are located in
-"Godot" project settings, NMAKE sheet. SCons is called at the end of
-the commands. If you make a mistake, copy the command from one of the
-other build configurations (debug, release_debug, release) or
-architectures (Win32/x64); they are equivalent.
```
akien-mga, line 191: I didn't assess how relevant this might still be but it seems to partially overlap with instructions in the dedicated page to configure VS. If some of it is missing there, that's where it should be added.

YuriSizov: Yeah, I think it's fair to offload this information to the main page about configuring VS.

development/compiling/compiling_for_windows.rst
```diff
@@ -169,12 +169,13 @@ dependencies. Running it will bring up the Project Manager.
...
+Development in Visual Studio
+----------------------------
 
-For most projects, using only scripting is enough but when development
-in C++ is needed, for creating modules or extending the engine, working
-with an IDE is usually desirable.
+Using an IDE is not required to compile Godot as SCons takes care of everything,
+but if you intend to do engine development or debugging of the C++ code on a
+Windows development environment, you may be interested in configuring things
+for use in Visual Studio.
```
YuriSizov, line 178:
```suggestion
Using an IDE is not required to compile Godot, as SCons takes care of everything.
But if you intend to do engine development or debugging of the engine's C++ code,
you may be interested in configuring a code editor or an IDE.

Folder-based editors don't require any particular setup to start working with Godot's
codebase. To edit projects with Visual Studio they need to be set up as a solution.
```

I'd try to make it sound less like VS is the only game in town, and try to explain why we dedicate a whole section to it (it has its own project system in the form of solutions).

akien-mga:

PR #5975, opened 2022-07-23
Add editor's support for Web and Android to FAQ

jynus:
Web editor is available since 3.3, and seems to be to the same
"level of support" as the other builds
* https://godotengine.org/download/
* https://editor.godotengine.org/releases/latest/

Android was announced:
https://godotengine.org/article/dev-snapshot-godot-3-5-beta-3

although it isn't on any stable release/doesn't seem 100%
recommended yet, so marking it as experimental.

I am targeting on this patch 3.5 only- my intention is to create
separate patches for master (which currently doesn't have
html support) and older versions.

### DIFF
--- a/about/faq.rst
+++ b/about/faq.rst
@@ -38,6 +38,8 @@ Which platforms are supported by Godot?
 * Windows
 * macOS
 * X11 (Linux, \*BSD)
+* Web (Chrome 68 or later, Firefox 79 or later, Edge 79 or later), see: :ref:`using the web editor <doc_using_the_web_editor>`)
+* Android (experimental)
 
 **For exporting your games:**
 
--- a/about/faq.rst
+++ b/about/faq.rst
@@ -38,7 +38,7 @@ Which platforms are supported by Godot?
 * Windows
 * macOS
 * X11 (Linux, \*BSD)
-* Web (Chrome 68 or later, Firefox 79 or later, Edge 79 or later), see: :ref:`using the web editor <doc_using_the_web_editor>`)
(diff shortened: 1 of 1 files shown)


mhilbrunner:
CC @akien-mga @Calinou I assume we _do_ want to make these officialy supported?

jynus:
Note Web support was mentioned already at https://docs.godotengine.org/en/stable/about/list_of_features.html I only made it compatible with the FAQ.

I though mentioning the android port as "experimental" was reasonable for the FAQ, but obviously, you will know better than me if to "hide it" on docs for now 0:-)

about/faq.rst
```diff
@@ -38,6 +38,8 @@ Which platforms are supported by Godot?
 * Windows
 * macOS
 * X11 (Linux, \*BSD)
+* Web (Chrome 68 or later, Firefox 79 or later, Edge 79 or later), see: :ref:`using the web editor <doc_using_the_web_editor>`)
```
Calinou, line 41: I wouldn't mention browser versions as these are bound to change in the future. Also, Opera should work as it's Chromium-derived:

```suggestion
* Web (Chrome, Firefox, Edge, Opera), see: :ref:`using the web editor <doc_using_the_web_editor>`)
```

I don't remember if Safari works with the web editor.

mhilbrunner (approved):

mhilbrunner:
Merging, Android editor support was just made official by today's Godot 3.5 release (https://godotengine.org/article/godot-3-5-cant-stop-wont-stop).

Thanks!

Issue #3291, opened 2020-03-20
Custom RichTextEffect Label process is only partially documented.


Haledire:
The RichTextLabel documentation about adding custom effects begins to explain about making the effects by creating the script, but ends there.  There is no explanation about the process creating a resource file of the type RichTextEffect, attaching the effect script to the RichTextEffect via the editor, and inserting that resource into the Custom Effect array in the editor.

If the class script is supposed to just work by itself, it seems to not be working that way as of 3.2.  Just making the script extending RichTextEffect was not adding the effect code.

Has the functionality changed? Or are there steps missing in the documentation?  From my experience the documentation is missing steps, but I don't know what the intended process is for this situation.

![not_working](https://user-images.githubusercontent.com/164616/77196057-085c9280-6ab9-11ea-90c7-7c00af871146.png)
with only the script - no resource added
![working](https://user-images.githubusercontent.com/164616/77196059-08f52900-6ab9-11ea-87e5-b87024783223.png)
with resource added to array with script attached

PR #5982, opened 2022-07-24
Update RichTextEffect tutorial

RedMser:
- The code examples have been updated for 4.0 syntax (`@tool` and other API changes)
- The documentation now mentions that `class_name` is a required part for custom effects to show up in the inspector

I am not sure how the Matrix effect is supposed to work, so I simply made it compile by fixing up the API, but it looks wrong. It should probably be properly fixed [by its original creator](https://github.com/Eoin-ONeill-Yokai/Godot-RichTextEffect-SampleProject/blob/master/richtexteffects/matrix.gd).

Please check whether my reference to `install_effects` is correct, I don't have a set up where I can try how the rst looks like in the end.

_Bugsquad edit: closes https://github.com/godotengine/godot-docs/issues/3291_

### DIFF
--- a/tutorials/ui/bbcode_in_richtextlabel.rst
+++ b/tutorials/ui/bbcode_in_richtextlabel.rst
@@ -557,11 +557,13 @@ Custom BBCode tags and text effects
 -----------------------------------
 
 You can extend the :ref:`class_RichTextEffect` resource type to create your own custom
-BBCode tags. You begin by extending the :ref:`class_RichTextEffect` resource type. Add
-the ``tool`` prefix to your GDScript file if you wish to have these custom effects run
-within the editor itself. The RichTextLabel does not need to have a script attached,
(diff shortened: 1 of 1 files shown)


skyace65 (commented):
Ghost works, and Matrix works the same way it did in 3.x so I'd say it's fine. I can't get pulse to do anything though, your version or the 3.x version in 3.4.

RedMser:
@skyace65 Pulse needs a color parameter, otherwise it does not do anything. Try the example usage in the code comment, it should work.

skyace65 (approved):
My bad, it's working perfectly. Also I checked and the reference to install effect should be correct, it's the same way other pages reference methods. Thanks for the PR!

mhilbrunner (approved):

mhilbrunner:
Looks good, thanks for this! Maybe https://github.com/Eoin-ONeill-Yokai can have a look at it too just in case, but I'm going to merge it.

PR #5983, opened 2022-07-24
Add "dotnet CLI" option to all C# editor docs

31:
The JetBrains Rider section contains this line:

> - Set **Mono** -> **Builds** -> **Build Tool** to **dotnet CLI**.

This actually applies just as much to the other editors, so this PR adds the note for them, too.

---

Other Build Tool options tend not to work, and they're only there for advanced edge cases, until Godot 4.1. https://github.com/godotengine/godot-website/pull/341#issuecomment-861342791. We get this Build Tool question in the C# Discord chat very frequently, so I think it's worth at least trying to document a little better.

The error people commonly ask about is this:

> The SDK 'Microsoft.NET.Sdk' specified could not be found.

Usually it seems to be because they had Visual Studio installed at some point earlier and Godot picked that old version of MSBuild that doesn't have (full?) support for the project style used in Godot 3.2.3+.

---

This applies to 3.x.

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_basics.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_basics.rst
@@ -116,6 +116,7 @@ After reading the "Prerequisites" section, you can download and install
 In Godot's **Editor → Editor Settings** menu:
 
 - Set **Mono** -> **Editor** -> **External Editor** to **Visual Studio Code**.
+- Set **Mono** -> **Builds** -> **Build Tool** to **dotnet CLI**.
 
 In Visual Studio Code:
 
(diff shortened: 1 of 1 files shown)


aaronfranke (commented):
I don't recall this being necessary, but I have nothing against this either. @neikeq Can you confirm if it's an improvement to switch this setting? And, maybe this should be made the default setting if it applies to every editor?

31:
From neikeq on an earlier PR: https://github.com/godotengine/godot-website/pull/341#issuecomment-861339067

> `dotnet CLI` has been the default for quite some time. As long as Godot can find it, that is.

My untested hypothesis is that some people have Visual Studio installed (with its own, old MSBuild), install Godot, presumably run Godot, and then continue with the instructions. So even once they go back and install the .NET SDK, it's still on an old detected setting. I haven't tested out that flow, or looked at the code all that closely, though.

But what I do know is that people (especially recently) somewhat frequently come to the C# Discord channel with the `"Microsoft.NET.Sdk" specified could not be found` problem, and after they change this option, their build works. (Or, changing the option fixes their first round of errors and then we can move onto the second layer. 😄)

Another benefit of having this in the doc is so it's easy to find an authoritative source to point at. E.g. if someone *intentionally* chooses the VS MSBuild option because they use Visual Studio, we can point to this to convince them it's the way to go even in that case.

mhilbrunner (approved):

mhilbrunner:
Thank you! 🎉

mhilbrunner:
Cherry-picked to 3.4. and 3.5.

PR #5984, opened 2022-07-24
Remove "Core" from .NET SDK, simplify Mono mention

31:
The .NET Core SDK is simply the .NET SDK as of .NET 5. This is what you see on https://dotnet.microsoft.com/download.

This PR changes the text to remove "Core", and removes the reference to .NET Core 3.1 to avoid confusing people. (Some people see 3.1 and assume they must download *that* version, rather than the latest stable.)

Also removes the reference to Godot 3.2.3 to make the paragraph simpler to read.

---

This applies to 3.x.

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_basics.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_basics.rst
@@ -43,21 +43,19 @@ Setting up C# for Godot
 Prerequisites
 ~~~~~~~~~~~~~
 
-Install the latest stable version of
-`.NET Core SDK <https://dotnet.microsoft.com/download/dotnet-core>`__
-(3.1 as of writing).
+Install the latest stable version of the
+`.NET SDK <https://dotnet.microsoft.com/download>`__, previously known as the
+.NET Core SDK.
 
 As of Godot 3.2.3, installing Mono SDK is not a requirement anymore,
 except it is required if you are building the engine from source.
 
-Godot bundles the parts of Mono needed to run already compiled games,
-however Godot does not include the tools required to build and compile
-games, such as MSBuild. These tools need to be installed separately.
-The required tools are included in the .NET Core SDK. MSBuild is also
-included in the Mono SDK, but it can't build C# projects with the new
-``csproj`` format, therefore .NET Core SDK is required for Godot 3.2.3+.
+Godot bundles the parts of Mono needed to run already compiled games.
+However, Godot does not bundle the tools required to build and compile
+games, such as MSBuild and the C# compiler. These are
+included in the .NET SDK, which needs to be installed separately.
 
-In summary, you must have installed .NET Core SDK
+In summary, you must have installed the .NET SDK
 **and** the Mono-enabled version of Godot.
 
 Additional notes


YuriSizov:
> This PR changes the text to remove "Core", and removes the reference to .NET Core 3.1 to avoid confusing people. (Some people see 3.1 and assume they must download _that_ version, rather than the latest stable.)

I mean, Godot's C# module is indeed built with Core 3.1 specifically, support for 5 and 6 is coming in 4.x.

31:
> I mean, Godot's C# module is indeed built with Core 3.1 specifically, support for 5 and 6 is coming in 4.x.

I don't think it matters what the C# module is built with, rather it matters what devs can use to build games. Godot 3.x currently works with the .NET 6.0 SDK and it's good to be able to use new C# versions (C# 10) because of the newer version of Roslyn included in the .NET 6.0 SDK than the .NET Core 3.1 SDK.

The runtime side is different and unrelated. The Godot 3.x runtime is Mono, not .NET Core 3.1. https://github.com/godotengine/godot-proposals/issues/2333 tracks an update to use .NET 6 as the runtime.

YuriSizov:
> rather it matters what devs can use to build games. Godot 3.x currently works with the .NET 6.0 SDK and it's good to be able to use new C# versions (C# 10)

Sure, as long you're sure the bundled Mono runtime will support those newer language features. I guess we call it `.NET SDK` on the download page anyway, so it's not a big deal.

31:
> Sure, as long you're sure the bundled Mono runtime will support those newer language features.

Yeah, this is true, it's only a subset. There are some very nice features in there, though!

Something nice about the .NET SDK is that it actually [chooses the default C# version it compiles using on the target framework](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version) in the csproj, so using the .NET Core 3.1 SDK actually compiles your code as C# 7.3, and the .NET 6.0 SDK chooses the same C# version. You have to go in and modify the csproj to change the C# version away from that. This means even if someone installs the .NET 6.0 SDK, they can't accidentally use any C# features that won't work in the Mono runtime.

That (and other ways that the .NET SDKs are careful to be backward compatible) ends up making it safe to have people install it rather than an old 3.1 SDK. 🙂

aaronfranke (commented):
The documentation should still state that Mono is required when building the engine from source.

We should specify "(previously known as the .NET Core SDK)" or similar, to avoid confusing users. We don't want a situation where people install an SDK for .NET Framework 4.x and wonder why it isn't working.

31:
> The documentation should still state that Mono is required when building the engine from source.

Hmm, is it really necessary to call this out in this particular section? My impression is that this paragraph is for using Godot rather than building it from source. It seems to me the reader might not even know that that term means. The next section (Additional notes) still mentions "If you are building Godot from source, install the latest stable version of Mono".

Regardless, I'll add it in, and I guess I'll just leave that as my suggestion/opinion.

> We should specify "(previously known as the .NET Core SDK)" or similar, to avoid confusing users.

:+1:

aaronfranke:
@31 That's a good point, it is stated only slightly further down the page. However, I don't want these different parts of the page to seem to contradict each other, if one place says it's not required, and another place tells you to go do it.

31:
> one place says it's not required as a blanket statement

I've removed this and added the original Mono paragraph back in with the push. I don't really mind this part all that much--the main things I was driving for in this PR are removing "Core" and "3.1" and cleaning up the bigger Mono paragraph a bit now that 3.2.3 is fairly old.

aaronfranke (approved):

aaronfranke:
Thanks!

Calinou:

PR #5985, opened 2022-07-24
Add notes for common C# bugs on Windows

31:
Add notes for a few C# bugs that we see often in the C# Discord channel.

---

First, mention the issues with the Visual Studio extension. This has been a problem for a while ([Aug 27, 2020](https://github.com/godotengine/godot-csharp-visualstudio/issues/10#issue-687428489), Godot 3.2.3) and I think it's worth pointing it out here, where it might help inform people's choice of editor, or at least let them know that they aren't alone in having this problem and show how they might be able to work around it.

I also changed "you need to download the Godot Visual Studio extension from github" to "you can download" because it is not required, and some people are happy to use Visual Studio with Godot even without the extension working.

---

Second, a fix for a NuGet behavior that we also see often in the C# Discord where the user's default NuGet.Config file loses its nuget.org source. I have suspicions that it comes from a Visual Studio update issue, but all I know is that I've only seen it happen on Windows and some people who hit it knew they were using Visual Studio.

I'm not totally sure this section is the right place for it, but I haven't noticed a more general troubleshooting section that it might fit into better.

---

This applies to 3.x.

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_basics.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_basics.rst
@@ -152,10 +152,23 @@ In Godot's **Editor → Editor Settings** menu:
 
 - Set **Mono** -> **Editor** -> **External Editor** to **Visual Studio**.
 
-Next, you need to download the Godot Visual Studio extension from github
+Next, you can download the Godot Visual Studio extension from github
 `here <https://github.com/godotengine/godot-csharp-visualstudio/releases>`__.
 Double click on the downloaded file and follow the installation process.
 
+.. note:: The option to debug your game in Visual Studio may not appear after
+          installing the extension. To enable debugging, there is a
+          `workaround for Visual Studio 2019 <https://github.com/godotengine/godot-csharp-visualstudio/issues/10#issuecomment-720153256>`__.
+          There is
+          `a separate issue about this problem in Visual Studio 2022 <https://github.com/godotengine/godot-csharp-visualstudio/issues/28>`__.
+
+.. note:: If you see an error like "Unable to find package Godot.NET.Sdk",
+          your NuGet configuration may be incorrect and need to be fixed.
+
+          A simple way to fix the NuGet configuration file is to regenerate it.
+          In a file explorer window, go to ``%AppData%\NuGet``. Rename or delete
+          the ``NuGet.Config`` file. When you build your Godot project again,
+          the file will be automatically created with default values.
 
 Creating a C# script
 --------------------


tutorials/scripting/c_sharp/c_sharp_basics.rst
```diff
@@ -152,10 +152,22 @@ In Godot's **Editor → Editor Settings** menu:
 
 - Set **Mono** -> **Editor** -> **External Editor** to **Visual Studio**.
 
-Next, you need to download the Godot Visual Studio extension from github
+Next, you can download the Godot Visual Studio extension from github
 `here <https://github.com/godotengine/godot-csharp-visualstudio/releases>`__.
 Double click on the downloaded file and follow the installation process.
 
+.. note:: As of Godot 3.2.3, the Godot Visual Studio extension is not fully
```
aaronfranke, line 159: At this point, if this bug has been around for so long, we should just state that it's not fully compatible without mentioning the Godot version. I don't think it's helpful to suggest users go back to Godot 3.2.1 to fix this.

skyace65:
I'm not a heavy C# user so correct me if I'm wrong or there's an angle I'm missing, but considering the current state of the plugin wouldn't it make more sense to just remove any reference to it at this point? It is functionally broken and from what I'm seeing the workaround causes other issues described [here](https://github.com/godotengine/godot-csharp-visualstudio/issues/26).

skyace65:
Also for future reference, don't make two different PR's for the same page on the same branch. Do it all in one PR. We don't mind that it's covering multiple issues, and since it's split you'll have to rebase one of them after one gets merged.

tutorials/scripting/c_sharp/c_sharp_basics.rst
31: Ah, I somewhat mindlessly used this phrase just because other 3.2.3 stuff on the page is using it, and the correlation to the new csproj format is vaguely interesting to me. 😄 I didn't think about how it comes across as a recommendation to downgrade. Definitely agree on that point, will remove.

31:
I pushed some changes to improve the wording (IMO 🙂) and remove the mention of 3.2.3. I also directly link the comment containing the workaround for VS2019 rather than telling people to read through the whole thread.

Something I'm unclear on is whether the workaround works in VS 2022 or just VS 2019. Someone on the 2022 thread said it does for them, but others in the C# Discord channel and on some gists say it doesn't, or that it seems to at first and then breaks later. That's why the sentence about VS2022 is a little vague--because I don't know if the workaround also works there. If people see this note, maybe they will report their findings on the VS2022 issue. I figure being vague until then is reasonable.

> We don't mind that it's covering multiple issues, and since it's split you'll have to rebase one of them after one gets merged.

I made sure before submitting that the PRs won't conflict as far as Git goes, or is there a more restrictive limit in this repo?

(I actually originally had these two notes split up into two branches because they aren't actually all that related, but I combined them because Git would have considered that particular split to be a conflict. 😄 Avoiding conflict resolution and rereview does make sense to me in general, and I'd be happy to optimize in that direction.)

31:
> I'm not a heavy C# user so correct me if I'm wrong or there's an angle I'm missing, but considering the current state of the visual studio extension wouldn't it make more sense to just remove any reference to it at this point?

I'm not really sure where to set the bar on that. In theory, someone *might* have a strong enough preference for Visual Studio to specifically install VS 2019 and use the workaround in order to get debugging working in their IDE of choice. I don't know how to weigh that vs. the untidiness of the doc pointing to a known-broken tool and not-necessarily-working workarounds.

(Thanks for the link to https://github.com/godotengine/godot-csharp-visualstudio/issues/26, I'd forgotten about that one!)

aaronfranke (approved):

mhilbrunner (approved):

mhilbrunner:
Thank you, great work! 🎉

mhilbrunner:
Cherry-picked to 3.4 and 3.5.

PR #6008, opened 2022-07-30
Added C# and C++ code to match with GDScript and tutorial on Your first 2D game page.

sbalia:
The issue with C# and C++ versions is that they don't have random initialization in the **ready** function as GDScript has.

Link to the documentation:
https://docs.godotengine.org/en/stable/getting_started/first_2d_game/05.the_main_game_scene.html#testing-the-scene

### DIFF
--- a/getting_started/first_2d_game/05.the_main_game_scene.rst
+++ b/getting_started/first_2d_game/05.the_main_game_scene.rst
@@ -432,13 +432,32 @@ call to ``_ready()``:
 
     public override void _Ready()
     {
+        GD.Randomize();
         NewGame();
     }
 
  .. code-tab:: cpp
 
     // This code goes in `main.cpp`.
+    #include "main.hpp"
+
+    #include <SceneTree.hpp>
+
+    #include "mob.hpp"
+
     void Main::_ready() {
+        _hud = get_node<HUD>("HUD");
+        _player = get_node<Player>("Player");
+        _start_position = get_node<godot::Node2D>("StartPosition");
+        _mob_spawn_location = get_node<godot::PathFollow2D>("MobPath/MobSpawnLocation");
+        _mob_timer = get_node<godot::Timer>("MobTimer");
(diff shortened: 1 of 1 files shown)


getting_started/first_2d_game/05.the_main_game_scene.rst
```diff
@@ -432,13 +432,32 @@ call to ``_ready()``:
...
+        _hud = get_node<HUD>("HUD");
+        _player = get_node<Player>("Player");
+        _start_position = get_node<godot::Node2D>("StartPosition");
+        _mob_spawn_location = get_node<godot::PathFollow2D>("MobPath/MobSpawnLocation");
+        _mob_timer = get_node<godot::Timer>("MobTimer");
+        _score_timer = get_node<godot::Timer>("ScoreTimer");
+        _start_timer = get_node<godot::Timer>("StartTimer");
+        // Uncomment these after adding the nodes in the "Sound effects" section of "Finishing up".
+        //_music = get_node<godot::AudioStreamPlayer>("Music");
+        //_death_sound = get_node<godot::AudioStreamPlayer>("DeathSound");
```
raulsntos, line 458: Why does the C++ example need so much more code?

sbalia: Since GDScript contains the full ready function, I copied C++ ready functions from the previous section where `_random->randomize();` was added. That section contains all the code. Maybe it was done by mistake, and only 2 lines are enough for it, but unfortunately, I can't test it.

raulsntos: I see, that makes sense. In that case I think it makes sense to repeat it.

sbalia:
@Calinou , @raulsntos , gentle ping. Should it be merged?

Calinou:
> The issue with C# and C++ versions is that they don't have random initialization in the ready function as GDScript has.

`randomize()` is now implicitly called, so it can be removed for all 3 languages in `master`.

PR #5991, opened 2022-07-26
Add C# examples to Creating your first script

a2937:
All of my examples are built to match the v3.4.4-stable_mono_win64 as that is the only Mono-based version I have. There was something weird with Godot.Sprite not matching the name of Sprite2D but that might just be an engine implementation oddity. The examples I have provided compile and run on my local install though.

PR #6009, opened 2022-07-30
Add C# code examples on Listening to player input

sbalia:
Link to the documentation:
https://docs.godotengine.org/en/stable/getting_started/step_by_step/scripting_player_input.html

### DIFF
--- a/getting_started/step_by_step/scripting_player_input.rst
+++ b/getting_started/step_by_step/scripting_player_input.rst
@@ -42,6 +42,20 @@ code below.
 
     rotation += angular_speed * direction * delta
 
+ .. code-tab:: csharp C#
+
+    var direction = 0;
+    if (Input.IsActionPressed("ui_left"))
+    {
+        direction = -1;
+    }
+    if (Input.IsActionPressed("ui_right"))
+    {
+        direction = 1;
+    }
+
+    Rotation += AngularSpeed * direction * delta;
+
 Our ``direction`` local variable is a multiplier representing the direction in
 which the player wants to turn. A value of ``0`` means the player isn't pressing
 the left or the right arrow key. A value of ``1`` means the player wants to turn
@@ -75,11 +89,19 @@ To only move when pressing a key, we need to modify the code that calculates the
 velocity. Replace the line starting with ``var velocity`` with the code below.
 
 .. tabs::
-   .. code-tab:: gdscript GDScript
+ .. code-tab:: gdscript GDScript
 
     var velocity = Vector2.ZERO
     if Input.is_action_pressed("ui_up"):
         velocity = Vector2.UP.rotated(rotation) * speed
+
+ .. code-tab:: csharp C#
+
+    var velocity = Vector2.Zero;
+    if (Input.IsActionPressed("ui_up"))
+    {
+        velocity = Vector2.Up.Rotated(Rotation) * Speed;
+    }
 
 We initialize the ``velocity`` with a value of ``Vector2.ZERO``, another
 constant of the built-in ``Vector`` type representing a 2D vector of length 0.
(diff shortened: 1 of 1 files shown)


getting_started/step_by_step/scripting_player_input.rst
```diff
@@ -116,6 +138,39 @@ Here is the complete ``Sprite2D.gd`` file for reference.
 
         position += velocity * delta
 
+ .. code-tab:: csharp C#
+
+    using Godot;
+    
+    public class Sprite : Godot.Sprite
```
raulsntos, line 145: The class should be `Sprite2D` since documentation targets master first, for cherry-picking it would have to be replaced by `Sprite`.

getting_started/step_by_step/scripting_player_input.rst
```diff
@@ -116,6 +138,39 @@ Here is the complete ``Sprite2D.gd`` file for reference.
...
+    {
+        private float Speed = 400;
+        private float AngularSpeed = Mathf.Pi;
```
raulsntos, line 148: The C# naming convention for fields is camelCase.

```suggestion
        private float speed = 400;
        private float angularSpeed = Mathf.Pi;
```

In the engine we also prefix private fields with an underscore `_` but I'm not sure if that also applies to the documentation.

sbalia: Yes, I'm aware of C# naming conventions. I just tried to match it with the existing documentation. For example, it was merged recently: https://docs.godotengine.org/en/stable/getting_started/step_by_step/scripting_first_script.html
Related PR: https://github.com/godotengine/godot-docs/pull/5991

I thought if it was reviewed and approved two days ago, it would be safe to follow the same rules. 

I'm for using `_` with private variables.

getting_started/step_by_step/scripting_player_input.rst
sbalia: According to the description of the `Sprite` class, it says a node that displays a 2D texture. I think the name omits `2D`. 
Otherwise, VS Code can't find a reference, and the project is not building when I use the `Sprite2D` class.

Can you please correct/advise?

getting_started/step_by_step/scripting_player_input.rst
sbalia: Updated with a lower case letter.

getting_started/step_by_step/scripting_player_input.rst
raulsntos: You must be using Godot 3.x, in 4.0 the class has been renamed to `Sprite2D`. The documentation targets `master` so it should use the 4.0 name.

getting_started/step_by_step/scripting_player_input.rst
raulsntos: Ah I missed that PR, it's ok match the existing documentation, the naming can be updated in a separate PR but since you already updated it.

getting_started/step_by_step/scripting_player_input.rst
sbalia: done

getting_started/step_by_step/scripting_player_input.rst
sbalia: So, shall I leave it with the lower case letter or revert back to the upperCase? I think it's better to revert it back.

raulsntos: Sure, either way is fine.

sbalia: Reverted back, to match with the current docs.

raulsntos (approved):

Calinou (approved):
Thanks!

Cherry-picked to the `3.5`, `3.4` and `stable` branches (with `Sprite2D` replaced by `Sprite` to match 3.x naming).

PR #6013, opened 2022-07-31
Cheatsheet of ease() with known names

wajrou:
Added known names of curves to be recognized better.
Because curves from -1 to 1 has five different runnings I separated it into middle column and rest of curves (Ease in-out & out-in) are aligned to side columns.


wajrou:
Error in names, fixed in next PR https://github.com/godotengine/godot-docs/pull/6014

PR #6014, opened 2022-07-31
Cheatsheet of ease() with known names of curves

wajrou:
Added known names of curves to be recognized better.

Because curves from -1 to 1 has five different runnings I separated it into middle column and rest of curves (Ease in-out & in) are aligned to side columns.

ease_cheatsheet


wajrou:
Maybe cropped version from -3.2 to 3.2 is even better...

ease_cheatsheet2

wajrou:
Maybe cropped version from -3.2 to 3.2 is even better for doc.

ease_cheatsheet2

mhilbrunner (approved):
Nice improvement! I like the other variant as well, but I guess we may as well go ahead with the first. If you want to discuss changing the range for the pic, a new PR for that would be welcome so we can have a focused discussion.

In the mean time, this will serve nicely.

mhilbrunner:
Thank you, and congrats on your first contribution to Godot's docs! :)

Again, feel free to open a new PR for the version with the changed value range if you want to!

Calinou:
Cherry-picked to the `3.5`, `stable` and `3.4` branches.

PR #6018, opened 2022-08-01
Singletons tutorial has deprecated methods.

Pixelatex:
Replaced deprecated C# methods with their new versions in the singletons/autoload tutorial page.

### DIFF
--- a/tutorials/scripting/singletons_autoload.rst
+++ b/tutorials/scripting/singletons_autoload.rst
@@ -165,7 +165,7 @@ means that the last child of root is always the loaded scene.
 
         public override void _Ready()
         {
-            Viewport root = GetTree().GetRoot();
+            Viewport root = GetTree().Root;
             CurrentScene = root.GetChild(root.GetChildCount() - 1);
         }
     }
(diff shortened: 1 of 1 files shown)


Calinou:
Note that these methods are not deprecated (in the sense that they won't be removed in 4.0). Still, the property syntax is now preferred indeed.

Calinou (approved):
.

Calinou (changes requested):
Before merging, could you apply the same changes to the GDScript code blocks?

Pixelatex:
@Calinou Unless I'm mistaken, GDscript doesn't have a new different syntax for these methods.

I assume they were changed for consistency with other methods in C# because they were using snake case before instead of PascalCase

Calinou:
> Unless I'm mistaken, GDscript doesn't have a new different syntax for these methods.

As of Godot 3.4 (and possibly before), there is a `root` property in SceneTree which can be used as an alternative to `get_root()`. The same likely applies to `current_scene`.

Pixelatex:
> > Unless I'm mistaken, GDscript doesn't have a new different syntax for these methods.
> 
> As of Godot 3.4 (and possibly before), there is a `root` property in SceneTree which can be used as an alternative to `get_root()`. The same likely applies to `current_scene`.

There's no reference to something other than get_root in the doscs and testing it in 3.4.4stable, get_tree().root or .Root is not valid. Are you sure?

Pixelatex:
p.s. @Calinou , unless I'm mistaken, these new methods are nowhere to be found in the docs, want me to add a PR for that as well?

mhilbrunner:
> p.s. @Calinou , unless I'm mistaken, these new methods are nowhere to be found in the docs, want me to add a PR for that as well?

They are documented I believe?

https://docs.godotengine.org/en/latest/classes/class_scenetree.html#class-scenetree-property-root

https://docs.godotengine.org/en/latest/classes/class_scenetree.html#class-scenetree-property-current-scene

mhilbrunner (approved):

mhilbrunner:
Thanks, and congrats on your first contribution to Godot's docs!

Calinou:
Cherry-picked to the `3.5`, `stable` and `3.4` branches.

PR #6020, opened 2022-08-02
Attempt to clarify reliable vs unreliable RPCs

dsnopek:
Per a discussion on https://github.com/godotengine/godot-proposals/issues/2032, the current description for "Reliable" RPC's in the High-level Multiplayer tutorial could give the incorrect impression that reliable RPC's are guaranteed to arrive (based on the "will arrive no matter what" in the original text). This PR attempts to clarify what is actually meant by reliable.

I'm a little worried that talking about acknowledgements is a little too much detail, and it could be perhaps described simpler like:

> - Reliable: if the function call does not arrive, it will be re-transmitted, which adds some overhead and may take a little longer.

... but I think it's useful to understand the nature of the extra overhead.

However, if others think it would be better in the simpler version, I'd be happy to update the PR!

Thanks :-)

### DIFF
--- a/tutorials/networking/high_level_multiplayer.rst
+++ b/tutorials/networking/high_level_multiplayer.rst
@@ -200,8 +200,8 @@ Synchronizing member variables is also possible:
 
 Functions can be called in two fashions:
 
-- Reliable: the function call will arrive no matter what, but may take longer because it will be re-transmitted in case of failure.
-- Unreliable: if the function call does not arrive, it will not be re-transmitted; but if it arrives, it will do it quickly.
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):

mhilbrunner:
I think this is a fair improvement as is, and a good example of how hard it is to not be blind to such when writing docs for a area one has expertise in; anyone in networking _knows_ what reliably means but this is a good thing to expand upon in a tutorial aimed at beginners/non-networking devs. Thank you!

Calinou:

PR #6026, opened 2022-08-04
Replace use of `rotation_degrees` in Running code in the editor

Erveon:
Godot 4 does not expose rotation_degrees anymore, so this documentation would produce errors.
This replaces the code examples to use rotation and converts the degrees used into the radian system.


Calinou (approved):

PR #6029, opened 2022-08-04
Clarify `hint_range` in Shading language

alex-pahdo:
Explain more about what hints do, because otherwise it's unclear what hint_range could mean. Saying that a value is used as a range, implies that it behaves like an iterable.


Calinou (approved):

PR #6030, opened 2022-08-04
Add albedo color configured to white - Update class_multimesh.rst

CristianoSimao:
If the user not set the albedo color to pure white, then the multiplicatiion with a diferent value than 1 will produce color totally wrong.
The pure white color r 255, g 255, b 255 internaly in Godot is 1 then the multiplication work ok.


YuriSizov:
Thanks for your contribution! Please take a note at the top of the file that you've changed. The `.rst` files for the class references are auto-generated based on the files in the main project repo. If you want to make changes to the class reference, you need to make them there:

https://github.com/godotengine/godot/tree/master/doc/classes/MultiMesh.xml

Also note that we typically don't describe colors in the 8-bit style. We describe them with floating point numbers for each aspect.

Issue #4311, opened 2020-10-25
The GDScript Exports docs need to covert `property_can_revert` and `property_get_revert` too.


willnationsdev:
**Your Godot version:**
3.2 stable

**Issue description:**
If you create runtime export variables via the `_get`, `_set`, and `_get_property_list` methods, then there is no clear way to define a default value for reverting any of the generated properties. The Inspector doesn't know that it can revert them.

The Inspector does provide overrides for forcibly determining true/false if a property can be overridden and a Variant response for what the reverted value should be via the `property_can_revert` and `property_get_revert` functions, but these methods are undocumented. I have covered their exclusion from the documentation in godotengine/godot#43078.

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/stable/getting_started/scripting/gdscript/gdscript_exports.html#advanced-exports

akien-mga:
Done for 3.4/3.5 with #6043, but this is still needed for 4.0. See relevant discussion in that PR.

PR #5622, opened 2022-02-18
Make better explanation of advanced exports

Dan-Gamin:
Make better explanation of advanced GDScript exports because It was hard for me the first time I tried to understand it.
This will make it easier for beginners to understand it as the existing explanation does not explain how to add properties to script categories/groups.

PR #4247, opened 2020-10-11
First pass on GDScript 2.0 docs

NathanLovato:
Rebase of #3623 on master to fix merge conflicts. Supersedes the previous PR.

PR #6043, opened 2022-08-07
Add section about default values (`property_can_revert` and `property_get_revert`) under GDScript exports

SimonLarsen:
Add short section documenting default values for advanced exports under Tutorials > Scripting > GDScript > GDScript exports.

Targeting 3.5 since the section "GDScript exports" is mostly missing in latest.

Fixes: #4311.

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_exports.rst
+++ b/tutorials/scripting/gdscript/gdscript_exports.rst
@@ -288,6 +288,34 @@ a variable to to a property also gives you the ability to give it a default stat
         })
         return properties
 
+Adding default values for properties
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To define default values for advanced exports, you need to override the ``property_can_revert()`` and ``property_get_revert()`` methods.
+
+* The ``property_can_revert()`` method takes the name of a property and must return ``true`` if the property can be reverted. This will enable the Revert button next to the property in the inspector.
+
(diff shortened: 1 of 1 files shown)


YuriSizov:
> Targeting 3.5 since the section "GDScript exports" is mostly missing in latest.

This is actually an accumulation of several mistakes. https://github.com/godotengine/godot-docs/pull/3444 added this information to the article and https://github.com/godotengine/godot-docs/pull/5622 expanded on it. But a part of the changes was lost in https://github.com/godotengine/godot-docs/pull/4247 for some reason.

`_get_property_list`, `_get`, `_set`, and the methods from this PR should be documented. In `master` as well, but maybe it's worth splitting this into another article. In `master` we also need to document the new grouping annotations, which are simpler to use than `_get_property_list` (and thus make it even more sensible to split the article).

We can merge this PR as is, but this needs to be cleaned up for either version.

mhilbrunner (approved):

akien-mga (approved):
This looks great for the time being.

I agree with @YuriSizov that this should be further reworked but that can be done later on.
Some changes are also needed in Godot itself to expose these methods as virtual so they appear in the docs, instead of relying on reading the source code or this new documentation section.

And indeed for 4.0 we should move documentation about exports to a dedicated page. Most of this is also relevant for C# or GDExtension so it should likely be outside the GDScript docs, though we'll also need language specific docs for e.g. annotations in GDScript.

akien-mga:
Thanks!

PR #6063, opened 2022-08-10
Default Environment file consumes a lot of RAM

CristianoSimao:
As requested in: https://github.com/godotengine/godot-docs/pull/6060 I am posting here the details:

Some very important things inside Godot have no information or documentation.
Here I am trying explain something I find and help GLES2 run with more speed in low hardware specs in old mobile phones.

The background and ambient light are configured in the Environment resource file.
The settings inside this file can create more draws than necessary.
The default environment file created when you start a new 3D project create a huge
sky texture of 1024 x 512 pixels inside the RAM. If you don't need a sky texture, or
want a custom sky texture with small size, delete the "default_env.tres" in your res:// directory
and create a new WorldEnvironment node so you can create a new environment inside the Inspector panel.

This will delete the default large Sky texture and save 1.5MB of RAM if you not add any other sky texture.

Bellow is some links with images and post about the topic to understand better.

See my report here: godotengine/godot#64183
And here: https://godotforums.org/d/30800-where-do-the-various-textures-in-the-debug-video-ram-monitor-come-from/14

### DIFF
--- a/tutorials/performance/optimizing_3d_performance.rst
+++ b/tutorials/performance/optimizing_3d_performance.rst
@@ -159,3 +159,15 @@ There may also be rendering and physics glitches due to floating point error in
 large worlds. You may be able to use techniques such as orienting the world
 around the player (rather than the other way around), or shifting the origin
 periodically to keep things centred around ``Vector3(0, 0, 0)``.
+
+Environment File
+================
+
+The background and ambient light are configured in the Environment resource file.
+The settings inside this file can create more draws than necessary.
+The default environment file created when you start a new 3D project create a huge
+sky texture of 1024 x 512 pixels inside the RAM. If you don't need a sky texture, or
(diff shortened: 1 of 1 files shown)


mhilbrunner:
I agree with the feedback in the previous PR that this should go under general optimization. I would also simplify this a bit (not sure the exact dimensions of the default sky are important here and may change; also not sure calling a 1024 x 512 sky texture 'huge' is useful).

Besides those, this for sure is important information for people optimizing their game, especially 2D, so good catch. :)

Vennnot:
If this is still true for Godot 4, I will create a new PR so this information is not lost

mhilbrunner:
Yeah its useful information! We should not overstate it though, saving 1.5mb of RAM is usually not _that_ much (maybe outside of mobile devices).

mhilbrunner:
Thanks for contributing! Closing due to the lack of response. A new PR with the changes discussed above would of course be welcome.

PR #6065, opened 2022-08-10
Add C#/Mono console support by Pineapple Works

wkubiak:
We've recently completed tech which enables us to port Godot Mono games to Nintendo Switch and the Xbox family of consoles so we wanted to reflect that in the docs.
Also added GDK next to Xbox consoles to signify that we do native Xbox ports and not just use the deprecated UWP technology.

Tried to make our entry as short and concise as possible, but am open to further modifications if deemed more optimal.

### DIFF
--- a/tutorials/platform/consoles.rst
+++ b/tutorials/platform/consoles.rst
@@ -66,7 +66,7 @@ Following is the list of providers:
 - `Lone Wolf Technology <http://www.lonewolftechnology.com/>`_ offers
   Switch and PS4 porting and publishing of Godot games.
 - `Pineapple Works <https://pineapple.works/>`_ offers
-  Switch, Xbox One and Xbox Series X/S porting and publishing of Godot games.
+  Switch, Xbox One & Xbox Series X/S (GDK) porting and publishing of Godot games (GDScript/C#).
 - `Flynn's Arcade <https://www.flynnsarcades.com/>`_ offers
   Switch porting and publishing of Godot games.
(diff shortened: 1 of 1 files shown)


tutorials/platform/consoles.rst
```diff
@@ -66,7 +66,7 @@ Following is the list of providers:
 - `Lone Wolf Technology <http://www.lonewolftechnology.com/>`_ offers
   Switch and PS4 porting and publishing of Godot games.
 - `Pineapple Works <https://pineapple.works/>`_ offers
-  Switch, Xbox One and Xbox Series X/S porting and publishing of Godot games.
+  Switch, Xbox One & Xbox Series X/S (GDK) porting and publishing of Godot games (GDScript/C#).
```
YuriSizov, line 69:
```suggestion
  Switch, Xbox One and Xbox Series X/S (GDK) porting and publishing of Godot games (GDScript/C#).
```
I'm not sure there is a reason to use an ampersand here. Fine otherwise.

wkubiak: Changed it just for brevity sake and not to have "and (...) and" in one line, but if you think it's better it can be like this.

YuriSizov: I think it's okay to have "and (...) and" here. Can add an Oxford comma to help with reading, or just replace it with a comma if you want.

mhilbrunner (approved):

mhilbrunner:
Thanks, cool tech! :)

akien-mga:

PR #6068, opened 2022-08-12
Add missing vertex normal array in TSCN file format

bitsawer:
Mesh normals array (second item) was missing from the array mesh TSCN resource listing, making it incorrect. The code sample below it seems to be correct.

Should apply to Godot 3.x too.


Calinou (approved):

PR #5830, opened 2022-05-20
Document type variations and update images

skyace65:
Documents the new type variations feature coming in 3.5. This PR should be cherrypickable. If not I can make a second for 3.5. I've also updated images for the theme editor page that needed updates, all images have gone through png compression.

PR #6080, opened 2022-08-16
Document type variations and update images

skyace65:
3.5 version of #5830. The only difference is I copied the paragraph and warning above my small paragraph for the gui skinning page since those are about type variations as well.


Calinou (approved):
Thanks!

PR #6084, opened 2022-08-18
Improved distance_to()

DevBadger10:
I have improved (at least in my eyes, as I still don't 100% understand GDScript, so this may not have been a problem in the first place) the distance_to() documentation in the Vector2 Class.

### DIFF
--- a/classes/class_vector2.rst
+++ b/classes/class_vector2.rst
@@ -400,7 +400,7 @@ This method runs faster than :ref:`distance_to<class_Vector2_method_distance_to>
 
 - :ref:`float<class_float>` **distance_to** **(** :ref:`Vector2<class_Vector2>` to **)** |const|
 
-Returns the distance between this vector and ``to``.
+Returns the number of pixels between this vector and ``to`` halved.
 
 ----
 


YuriSizov:
Hey, thanks for contributing. Please take note of a warning at the top of the file that you've edited. This is a generate file. Changes to the class reference must be made in the XMLs files in the main project repository:
https://github.com/godotengine/godot/tree/master/doc/classes/Vector2.xml

-----

That said, this doesn't look right. First of all, Vector2 itself is unitless. And then there is no reason to assume that it doesn't return the actual distance, and instead only returns half of it.

DevBadger10:
Thanks for the feedback. If I am to want to contribute again, I will be more wary of these warnings. At first, I didn't understand what it returned, so this change may have save the hassle for other noobs like myself, but IDK, there's probably some easy  way of  finding this out in the docs that I just completely missed. Thank you for your time!

YuriSizov:
It returns a distance between two unitless points in 2D space. There is no secret to it, so I'm not sure what's causing your confusion.

DevBadger10:
I'm not sure either.

PR #6094, opened 2022-08-21
Correct has_filter() return type

olibombardier:
Corrected the return type of the has_filter() from string to bool. The master branch already has the right return type.


Calinou:
Superseded by https://github.com/godotengine/godot/pull/64560.

PR #6095, opened 2022-08-21
Fix AnimationNode.has_filter() return type

olibombardier:
Corrected the return type of the has_filter() from string to bool. The master branch already has the right return type.


skyace65:
Hi, we appreciate the pull request but this change needs to be made in the main Godot repository. The class reference pages in the docs repository are generated from what's in the Godot repository. This is where the file you want to edit is https://github.com/godotengine/godot/blob/master/doc/classes/AnimationNode.xml.

PR #6098, opened 2022-08-22
Clarify issue with GLES2 renderer

creikey:


Calinou:
Please open a new pull request targeting the `3.5` branch instead, as the `master` branch does not have a GLES2 renderer. Its OpenGL renderer will use a different approach to lighting (a mixture between single-pass and multipass).

PR #6110, opened 2022-08-26
Update C# signal documentation

raulsntos:
Updates C# signal documentation and code examples to the new API in 4.0.


mhilbrunner (changes requested):
The changes to the `classes/*.rst` files need to be made to the class doc XMLs in the main repo instead and removed from these, as the RSTs are autogenerated and overwritten periodically, so these changes would get lost.

Everything else looks good at first glance.

raulsntos:
Oops, you are right, I used a regex to find all instances and missed that I was editing the class reference :upside_down_face:

mhilbrunner (approved):

mhilbrunner:
Great, thanks for working on 4.0 docs! ❤️

PR #5792, opened 2022-05-02
Remove search highlights per #5665

eruhlinteractive:
Fixes: [https://github.com/godotengine/godot-docs/issues/5665](https://github.com/godotengine/godot-docs/issues/5665)

PR #6116, opened 2022-08-27
Add link to top of article body to remove search highlight if it exists

### LINKED ISSUE #5665: Highlights make reading difficult.
**Your Godot version:**
Godot 3 alpha 3

**Issue description:**
Highlights make reading difficult and there is no easy way to turn off highlights when using the search bar. (Except manually deleting the URL). I have to change this address every time.
Ekran Resmi 2022-03-07 20 30 36

https://docs.godotengine.org/en/latest/classes/class_string.html?highlight=String
Ekran Resmi 2022-03-07 20 15 44

It is very pointless that the search highlighting is still stuck on the redirected page. All browsers already have a search feature. It's ridiculous to impose that.

the-sink:
Fixes #5665
This is an alternate implementation of #5792. I decided to make this a new PR altogether as I felt this version was too different from the other one to just be commented feedback (first PR, apologies if this should be done differently!)

Reasons for this implementation (as opposed to the existing PR) are
- In certain cases highlights can be useful, so the user should have the option to hide them if they so choose
- Sphinx has a built-in `Documentation.hideSearchWords()` method that removes all highlights and updates the URL accordingly, all in one line (much cleaner)

In this version, it just adds a link to the top of the article body that can be clicked to hide the search highlights. It only shows if the `highlight` property exists in the URL, and is removed once clicked. It follows the same setup as described [here](https://sphinx-basic-ng.readthedocs.io/en/latest/usage/components/search-hide/), but in `custom.js` instead of a separate component referenced in the templates.

https://user-images.githubusercontent.com/18225391/187023570-4eee8871-3f01-437f-a1f9-2ce519f40a7e.mp4

### DIFF
--- a/_static/js/custom.js
+++ b/_static/js/custom.js
@@ -198,8 +198,15 @@ const registerSidebarObserver = (function(){
   };
 })();
 
+const hideSearchHighlight = function(){
+  Documentation.hideSearchWords();
+  document.querySelector('.highlight-link').remove();
+};
+
 $(document).ready(() => {
   const mediaQuery = window.matchMedia('only screen and (min-width: 769px)');
+  const articleBody = document.querySelector('div[itemprop="articleBody"]');
+  const url = new URL(location.href);
 
   registerOnScrollEvent(mediaQuery);
   mediaQuery.addListener(registerOnScrollEvent);
@@ -213,7 +220,7 @@ $(document).ready(() => {
     // automatically generated `meta description` tag.
     const strippedUrl = [location.protocol, '//', location.host, location.pathname].join('');
     const updatedUrl = strippedUrl.replace('/latest/', '/stable/');
-    document.querySelector('div[itemprop="articleBody"]').insertAdjacentHTML('afterbegin', `
+    articleBody.insertAdjacentHTML('afterbegin', `
       <div class="admonition attention">
         <p class="first admonition-title">Attention</p>
         <p>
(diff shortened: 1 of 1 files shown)


YuriSizov:
> * In certain cases highlights can be useful, so the user should have the option to hide them if they so choose

I find the button too awkward to justify this case. If it's useful to have the highlight of some sort, even situationally, I'd go with an adjusted style for it. Something more subdued.

the-sink:
> I find the button too awkward to justify this case. If it's useful to have the highlight of some sort, even situationally, I'd go with an adjusted style for it. Something more subdued.

Yeah honestly, after having sat on it for a few days I'd tend to agree with you (and the point of the original issue) - essentially all browsers have a Find/Search in Page feature at this point, so search highlights like this are far less useful now.

> There was a reason why we removed the background, but we kept the border. Or maybe it was added by the Sphinx theme later. But maybe if the colors are not as intense, it could work?

I do have accessibility concerns with desaturating/lowering the contrast between the highlight and the background as it might become more difficult for some to distinguish... but that just leads me back to the original point, nearly all browsers have this and do it better anyways (can be turned on and off on a whim, most have good contrast/are easy to identify, etc).

Perhaps the original PR's idea of simply auto-removing the highlights would be a better choice, but maybe making use of the built in `hideSearchWords()` function so there's less redundant code. I'll leave a mention about it over in #5792.

the-sink:
Closing in favor of #5792!

Issue #6099, opened 2022-08-22
Saving Games page incorrect JSON functions in example


ztc0611:
**Your Godot version:**
4.0 Alpha 14

**Issue description:**
The docs page for Saving Games still uses the `to_json()` and `parse_json()` syntax which is now invalid in Godot 4.0.

**URL to the documentation page:**
https://docs.godotengine.org/en/latest/tutorials/io/saving_games.html?highlight=serialization#saving-and-reading-data

PR #6120, opened 2022-08-29
Update description on the IO to use the JSON class

shinspiegel:
I did see an issue (#6099) regarding the usage of the `to_json` that was removed and updated the docs to use the `JSON` class. This is correct?

Please, keep in mind that English is not my native tongue, and the text could need some better description that I was not able to provide.

Thanks for all the feedback. 

_Bugsquad edit: fixes #6099_

### DIFF
--- a/tutorials/io/saving_games.rst
+++ b/tutorials/io/saving_games.rst
@@ -56,8 +56,8 @@ Serializing
 The next step is to serialize the data. This makes it much easier to
 read from and store to disk. In this case, we're assuming each member of
 group Persist is an instanced node and thus has a path. GDScript
-has helper class :ref:`JSON<class_json>` to convert between dictionary and string, 
-Our node needs to contain a save function that returns this data. 
+has helper class :ref:`JSON<class_json>` to convert between dictionary and string,
+Our node needs to contain a save function that returns this data.
 The save function will look like this:
 
 .. tabs::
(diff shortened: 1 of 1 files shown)


shinspiegel:
Should I update something else on this one?

tutorials/io/saving_games.rst
```diff
@@ -156,8 +153,14 @@ way to pull the data out of the file as well.
             # Call the node's save function.
             var node_data = node.call("save")
 
+            # Creates the helper class to interact with JSON
+            var json = JSON.new()
+
+            # Convert save data into a serialized JSON string
+            var json_string = json.stringify(node_data)
```
Calinou, line 160: This can be achieved with `JSON.stringify()` as a static method now. No need to create a new JSON instance.

tutorials/io/saving_games.rst
```diff
@@ -200,7 +203,7 @@ way to pull the data out of the file as well.
 
 
 Game saved! Now, to load, we'll read each
-line, use ``parse_json()`` to read it back to a dict, and then iterate over
+line, use `parse()<class_json@parse>` method to read it back to a dict, and then iterate over
```
Calinou, line 206:
```suggestion
line. Use the :ref:`parse<class_JSON_method_parse>` method to read the
JSON string back to a dictionary, and then iterate over
```

tutorials/io/saving_games.rst
shinspiegel: Thanks, I didn't know about that.   
Already updated the PR

shinspiegel: Something else that I should update?

mhilbrunner (approved):

mhilbrunner:
Thank you! 🎉

PR #6121, opened 2022-08-30
updated rotation value

RayAsh37:
Corrected the rotation value to get the desired view from the camera.


RayAsh37:
This is regarding the issue #6122

RayAsh37:
Sure

PR #6130, opened 2022-09-01
Data paths: Fix editor paths and improve docs on custom user dir

akien-mga:
I plan to do some refactoring in 4.0 because the current configuration is awkward with both `use_custom_user_dir` and `custom_user_dir_name`, this can be improved further. But I won't change it in `3.x` so this PR should be good to merge first and cherry-pick.

~*Note:* Still compiling locally to test and proofread.~
*Edit:* Tested and proofread.

### DIFF
--- a/tutorials/io/data_paths.rst
+++ b/tutorials/io/data_paths.rst
@@ -4,20 +4,29 @@ File paths in Godot projects
 ============================
 
 This page explains how file paths work inside Godot projects. You will learn how
-to access paths in your projects using the ``res://`` and ``user://`` notations
-and where Godot stores user files on your hard-drive.
+to access paths in your projects using the ``res://`` and ``user://`` notations,
+and where Godot stores project and editor files on your and your users' systems.
 
 Path separators
 ---------------
 
-To make supporting multiple platforms easier, Godot only accepts UNIX-style path
-separators (``/``). These work on all platforms, **including Windows**.
+To make supporting multiple platforms easier, Godot uses **UNIX-style path
+separators** (forward slash ``/``). These work on all platforms, **including
+Windows**.
 
-Instead of writing paths like ``C:\Projects``, in Godot, you should write
-``C:/Projects``.
+Instead of writing paths like ``C:\Projects\Game``, in Godot, you should write
+``C:/Projects/Game``.
 
-Accessing files in the project folder
--------------------------------------
+Windows-style path separators (backward slash ``\``) are also supported in some
+path-related methods, but they need to be doubled (``\\``), as ``\`` is normally
+used as an escape for characters with a special meaning.
+
+This makes it possible to work with paths returned by other Windows
+applications. We still recommend using only forward slashes in your own code to
+guarantee that everything will work as intended.
+
+Accessing files in the project folder (``res://``)
+--------------------------------------------------
 
 Godot considers that a project exists in any folder that contains a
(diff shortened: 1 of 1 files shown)


akien-mga:
The table changes might be a bit difficult to proofread so here are some screenshots.

| Before | After |
| --- | --- |
| image |  image |
| image | image |

tutorials/io/data_paths.rst
```diff
@@ -125,9 +157,9 @@ Self-contained mode
 
 If you create a file called ``._sc_`` or ``_sc_`` in the same directory as the
 editor binary, Godot will enable *self-contained mode*. This mode makes Godot
-write all user data to a directory named ``editor_data/`` in the same directory
-as the editor binary. You can use it to create a portable installation of the
-editor.
+write all editor data, settings, and cache to a directory named ``editor_data/``
+in the same directory as the editor binary.
+You can use it to create a portable installation of the editor.
```
YuriSizov, line 162: Weird line break, should probably move a bit of the sentence to the previous line.

YuriSizov (approved):
Looks nice and correct.

tutorials/io/data_paths.rst
YuriSizov: > This means that future changes to the previous sentence won't necessarily cause a rewrap of the whole paragraph, because there's some whitespace available for growth.

Well, I see your point, but since we don't apply this principle consistently throughout the docs, it feels rather arbitrary. Though that could be interesting if we did. Every sentence from a new line would definitely improve readability of some diffs.

akien-mga: Well I apply this principle wherever it makes sense to me. For example here:
```
- **Cache** contains data generated by the editor, or stored temporarily.
  It can safely be removed when Godot is closed.
```

"temporarily." ends at column 74, if we let ourselves go to 80 the "It" should go on the first line, but it's ugly, so I prefer to wrap earlier.

Overall strict wrapping rules are unenforceable and we have a lot of documentation which doesn't hard wrap at 80 chars, so I don't think consistency is achievable here. Everyone kind of uses what works for them as long as it's not *wildly* inconsistent.

akien-mga: After further edit from bruvzg's input I had to rewrap some more so I went with this:
image

That could indeed be worth exploring as a recommendation for doc writers, though I'm not sure it's _always_ meaningful to wrap new sentences on the next line. Might need a recommendation that depends on the sentence length and whether it fits without wrapping after the previous sentence.

E.g. should this:
```
Hello world. This is another sentence. And now a longer one that
wraps around because it's longer.
```
be:
```
Hello world.
This is another sentence.
And now a longer one that wraps around because it's longer.
```
or
```
Hello world. This is another sentence.
And now a longer one that wraps around because it's longer.
```

tutorials/io/data_paths.rst
```diff
@@ -125,9 +157,9 @@ Self-contained mode
 
 If you create a file called ``._sc_`` or ``_sc_`` in the same directory as the
 editor binary, Godot will enable *self-contained mode*. This mode makes Godot
```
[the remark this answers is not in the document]
bruvzg: Yes, I think it was discussed before and we decided not to change it form 3.x for this reason. But probably should be done for 4.0 release.

akien-mga:

PR #6133, opened 2022-09-02
Add a page documenting Godot's internal rendering architecture

Calinou:
This is intended for new and returning engine contributors.

There are some clarifications I'd like to add – see my review comments and the question below. Also, I may have forgotten talking about certain features of the renderer entirely, so please comment if that's the case.

~~@BastiaanOlij Could you write up a paragraph or two about subpass usage in the Forward Mobile backend, and how it differs from approach in the Clustered Forward backend (and why subpasses are used)? This would be useful to list on this page~~ :slightly_smiling_face:
**Edit:** [Added](https://github.com/godotengine/godot-docs/pull/6133#issuecomment-1281688415)


development/cpp/internal_rendering_architecture.rst
```diff
@@ -0,0 +1,600 @@
...
+
+**Vulkan context creation:**
+
+- `drivers/vulkan/vulkan_context.cpp <https://github.com/godotengine/godot/blob/master/drivers/vulkan/vulkan_context.cpp>`__
+
+Direct3D 12
+^^^^^^^^^^^
+
+Like Vulkan, the Direct3D 12 driver targets modern platforms only.
+**TODO:** Mention required feature level/optional feature levels.
```
Calinou, line 46: @RandomShaper What's the minimum required D3D12 feature level for the renderer, and what's the optimal level (i.e. with all optional features covered)?

Knowing this would also be useful to list supported GPUs on a troubleshooting page. Specifically, there are some AMD GPUs that support Direct3D 12, but only with a limited feature level (Radeon 200/300 series).

development/cpp/internal_rendering_architecture.rst
```diff
@@ -0,0 +1,600 @@
...
+
+Batching and instancing
+^^^^^^^^^^^^^^^^^^^^^^^
+
+In the clustered and mobile backends, Vulkan instancing is used to group
+rendering of identical objects for performance. This is not as fast as static
+mesh merging, but it still allows instances to be culled individually.
+
+In the compatibility backend, a more limited form of 3D instancing is also used.
+**TODO:** Check if this is the case?
```
Calinou, line 297: @clayjohn Does the OpenGL backend in `master` use some form of automatic instancing in 3D? I haven't followed recent developments, so I don't remember if this is different from what's done in 3.x.

development/cpp/internal_rendering_architecture.rst
```diff
@@ -0,0 +1,600 @@
...
+compile-time. To see any changes you've made to those GLSL shaders, you need to
+recompile the editor or export template binary.
+
+Some material features such as height mapping, refraction and proximity fade are
+not part of core shaders, and are performed in the default BaseMaterial3D using
+the Godot shader language instead (not GLSL). This is done by procedurally
+generating the required shader code depending on the features enabled in the
+material.
+
+By convention, shader files with ``_inc`` in their name are included in other
```
paddy-exe, line 211: Maybe also mention that adding include files or more features into the core shaders should only be done if absolutely necessary since they need to compile very quickly? At least that's what I understood from Juan's feedback on my Visual Shader PR: https://github.com/godotengine/godot/pull/64248#issuecomment-1214167428

paddy-exe:
Absolutely amazing work on this! Not sure if this should get its own category but it would be great to have the Visual Shader system and its place in the rendering code as well here.

development/cpp/internal_rendering_architecture.rst
Calinou: Done :slightly_smiling_face:

development/cpp/internal_rendering_architecture.rst
```diff
@@ -0,0 +1,694 @@
...
+accurate additive blending for performance reasons). Lights with shadows are
+rendered in multiple passes. This is required for compatibility with low-end
+mobile devices, but it has a huge cost to performance. As a result, it's
+recommended to only have a handful of lights with shadows in the camera frustum
+at most. The maximum number of lights visible at once can be adjusted in the
+project settings.
+
+In all 3 backends, lights without shadows are much cheaper than lights with shadows.
+Godot currently doesn't separate static shadow rendering from dynamic shadow
+rendering, but this is planned in a future release.
```
clayjohn, line 363: I'm not sure this statement is accurate. Godot doesn't maintain 2 sets of shadow maps (one for static and one for dynamic), but it treats the two differently and only updates shadow maps when dynamic objects change.

development/cpp/internal_rendering_architecture.rst
```diff
@@ -0,0 +1,694 @@
...
+
+Clustering is also used for reflection probe and decal rendering in the
```
clayjohn, line 365:
```suggestion
Clustering is also used for reflection probes and decal rendering in the
```

development/cpp/internal_rendering_architecture.rst
```diff
@@ -0,0 +1,694 @@
...
+Clustered Forward backend.
+
+Shadow mapping
+^^^^^^^^^^^^^^
+
+Both Clustered Forward and Forward Mobile backends use the same shadow mapping
+approach, with the use of a Vogel disk for improved filtering. Dithering is also
+used to improve shadow appearance at the cost of some visible graininess.
```
clayjohn, line 373:
```suggestion
Both Clustered Forward and Forward Mobile backends use percentage closer 
filtering (PCF) to filter shadow maps and create a soft penumbra. Instead of
using a fixed PCF pattern, they use a vogel disk pattern which allows for changing
the number of samples and smoothly changing the quality.
```

development/cpp/internal_rendering_architecture.rst
```diff
@@ -0,0 +1,694 @@
...
+
+Godot also supports percentage-closer soft shadows (PCSS) for more realistic
+shadow penumbra rendering. PCSS shadows are limited to the Clustered Forward
+backend as they're too demanding to be usable in the Forward Mobile backend.
+
+The Compatibility backend doesn't support shadow mapping for any light types yet.
```
clayjohn, line 379:
```suggestion
backend as they're too demanding to be usable in the Forward Mobile backend. PCSS also uses a vogel-disk shaped kernel.

Additionally, both shadow-mapping techniques rotate the kernel on a per-pixel basis
to help soften under-sampling artifacts.

The Compatibility backend doesn't support shadow mapping for any light types yet.
```

development/cpp/internal_rendering_architecture.rst
```diff
@@ -0,0 +1,694 @@
...
+  Compatibility backends.
+
+The Clustered Forward backend supports screen-space ambient occlusion,
+screen-space indirect lighting, screen-space reflections and subsurface scattering.
+
+SSAO uses an implementation derived from Intel's
+`ASSAO <https://www.intel.com/content/www/us/en/developer/articles/technical/adaptive-screen-space-ambient-occlusion.html>`__
+(converted to Vulkan). SSIL is derived from SSAO to provide a high-performance
+implementation of screen-space global illumination (albeit with only indirect
+lighting simulated, hence its name).
```
clayjohn, line 508:
```suggestion
(converted to Vulkan). SSIL is derived from SSAO to provide high-performance
indirect lighting.
```

SSIL is explicitly not a form of global illumination, I would not call it one either as it will create unfair expectations.

development/cpp/internal_rendering_architecture.rst
```diff
@@ -0,0 +1,694 @@
...
+Sky rendering
+^^^^^^^^^^^^^
+
+.. seealso::
+
+    :ref:`doc_sky_shader`
+
+Godot supports using shaders to render the sky background. The radiance and
+irradiance maps (which are used to provide ambient light and reflections for PBR
+materials) are automatically updated based on the sky shader.
```
clayjohn, line 553:
```suggestion
Godot supports using shaders to render the sky background. The radiance map (which is used to provide ambient light and reflections for PBR
materials) is automatically updated based on the sky shader.
```

There is no irradiance map in Godot 4.0

development/cpp/internal_rendering_architecture.rst
```diff
@@ -0,0 +1,694 @@
...
+  Only available in the Clustered Forward backend, not the Forward Mobile or
+  Compatibility backends.
+
+.. seealso::
+
+    :ref:`doc_fog_shader`
+
+Godot supports a frustum-aligned voxel (froxel) approach to volumetric fog
+rendering. As opposed to a post-processingq filter (which is best suited for
+crepuscular rays), this approach is more general-purpose as it can work with any
```
clayjohn, line 585:
```suggestion
rendering. As opposed to a post-processing filter, this approach is more
general-purpose as it can work with any
```

development/cpp/internal_rendering_architecture.rst
```diff
@@ -0,0 +1,694 @@
...
+- `servers/rendering/renderer_rd/shaders/environment/volumetric_fog.glsl <https://github.com/godotengine/godot/blob/master/servers/rendering/renderer_rd/shaders/environment/volumetric_fog.glsl>`__
+- `servers/rendering/renderer_rd/shaders/environment/volumetric_fog_process.glsl <https://github.com/godotengine/godot/blob/master/servers/rendering/renderer_rd/shaders/environment/volumetric_fog_process.glsl>`__
+
+Occlusion culling
+^^^^^^^^^^^^^^^^^
+
+While modern GPUs can handle drawing a lot of triangles, the number of draw
+calls in complex scenes can still be a bottleneck (even with Vulkan and Direct3D
+12). Additionally, mobile GPUs are much more constrained in this regard.
+This makes occlusion culling still relevant today.
```
clayjohn, line 613:
```suggestion
12).
```

We don't need to defend the use of occlusion culling in this document

clayjohn (commented):
We should also include the image from https://github.com/godotengine/godot/pull/44094 somewhere in the document and explain its meaning in more detail. IMO the structure of the RenderingServer is something a lot of new contributors have trouble with

development/cpp/internal_rendering_architecture.rst
Calinou: The "future release" part refers to this proposal: https://github.com/godotengine/godot-proposals/issues/4635

development/cpp/internal_rendering_architecture.rst
```diff
@@ -0,0 +1,694 @@
...
+
+.. note::
+
+    Only available in the Clustered Forward and Forward Mobile backends, not the
+    Compatibility backend. Multiple Viewports can still be used to emulate this
+    when using the Compatibility backend, or to perform 2D resolution scaling.
+
+2D and 3D are rendered to separate buffers to allow for 3D resolution scaling
+without requiring the use of multiple Viewport nodes (which has greater
+performance and memory usage overhead).
```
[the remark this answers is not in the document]
Calinou: Done.

PS: I've wondered if we should say "SDR" instead of "LDR", as I see "SDR" much more often in popular parlance (such as on [monitor review sites](https://rtings.com/)). I rarely see LDR even mentioned in comparison. Do other engines refer to it as "LDR" or "SDR"?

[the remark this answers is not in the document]
clayjohn: I've never heard the term "SSR" before, so I can't really say

BastiaanOlij:
@Calinou first draft of what could be a replacement `Forward Mobile` section. Let me know if its usable:
```
Forward Mobile
^^^^^^^^^^^^^^

This is a forward renderer that uses a traditional single-pass approach to lighting.

Intended for mobile platforms, but can also run on desktop platforms. This
backend is optimized to perform well on mobile GPUs. Mobile GPUs have a very
different architecture compared to desktop GPUs due to their unique constraints
around battery usage, heat and overal bandwidth limitations of reading and writing
data. Compute shaders also have very limited support or aren't supported at all, the
mobile renderer as a result purely uses raster based shaders.

Instead of rendering the whole image as a single unit, the image is divided in
smaller tiles that fit within the faster internal memory of the mobile GPU. Each
tile is rendered and then written out to the destination texture. This all happens
automatically on the graphics driver.

The problem this introduces is that our traditional approach for desktop rendering,
where we render all opaque geometry, then handle the background, then transparent
geometry, then post processing, introduces bottlenecks as each pass will need to
read the current result into tile memory, perform its operations and then write it 
out again. We then wait for all tiles to be completed before moving on to the next
pass.

The first important change in the mobile renderer is that the mobile renderer does
not use the RGBA16F texture formats that the desktop renderer does. Instead it is
using a R10B10G10A2 texture format. This halves the bandwidth required and has
further improvements as mobile hardware often further optimises 32bit formats.
The tradeoff is that the mobile renderer has limited HDR capabilities due to the 
reduced precision in the color data.

The second important change is the use of sub-passes whenever possible. Sub-passes
allows us to perform the rendering steps end-to-end per tile saving on the overhead
introduced by reading from and writing to the tiles inbetween each rendering pass.
The ability to use sub-passes is limited by the inability to read neighboring pixels
as we're constraint to working within a single tile.
This results in not being able to implement features such as glow and dof.
Similarly if there is a requirement to read from the screen texture or depth texture
we must fully write out the rendering result limiting our ability to use sub-passes.
When such features are enabled a mix of sub-passes and normal passes are used, and
these features thus have a notable performance penalty.

On desktop platforms the use of sub-passes won't have any impact on performance 
however this backend can still perform better than Clustered Forward in 
simple scenes thanks to its lower complexity and lower bandwidth usage.
This is especially noticeable on low-end GPUs, integrated graphics or in VR applications.

Given its low-end focus, this backend does not provide high-end rendering
features such as SDFGI and Volumetric fog. Most post-processing effects are also not
available.
```

Calinou:
> @Calinou first draft of what could be a replacement `Forward Mobile` section. Let me know if its usable:

Thanks, I incorporated it :slightly_smiling_face:

Calinou:
I've added a small section on recommended OpenGL/Vulkan learning resources, as it may help with understanding concepts on this page (see https://twitter.com/reduzio/status/1649707259281260550).

contributing/development/core_and_modules/internal_rendering_architecture.rst
```diff
@@ -0,0 +1,762 @@
...
+
+Lightmap baking happens on the GPU using Vulkan compute shaders – CUDA or OpenCL
+are **not** used. Using Vulkan compute allows for greater compatibility and
+performance on all kinds of GPUs. In contrast, CUDA only runs on NVIDIA hardware
+and OpenCL performance is crippled on NVIDIA GPUs.
+
+The GPU-based lightmapper is implemented in the LightmapperRD class, which inherits
+from the Lightmapper class. This allows for implementing additional lightmappers,
+paving the way for a future port of the CPU-based lightmapper present in Godot 3.x.
+This would allow baking lightmaps while using the Compatibility backend.
```
clayjohn, line 502:
```suggestion
Lightmap baking happens on the GPU using Vulkan compute shaders. The
GPU-based lightmapper is implemented in the LightmapperRD class, which inherits
from the Lightmapper class. This allows for implementing additional lightmappers,
paving the way for a future port of the CPU-based lightmapper present in Godot 3.x.
This would allow baking lightmaps while using the Compatibility backend.
```

contributing/development/core_and_modules/internal_rendering_architecture.rst
```diff
@@ -0,0 +1,762 @@
...
+The Forward+ backend supports screen-space ambient occlusion,
+screen-space indirect lighting, screen-space reflections and subsurface scattering.
+
+SSAO uses an implementation derived from Intel's
+`ASSAO <https://www.intel.com/content/www/us/en/developer/articles/technical/adaptive-screen-space-ambient-occlusion.html>`__
+(converted to Vulkan). SSIL is derived from SSAO to provide high-performance
+indirect lighting.
+
+When both SSAO and SSIL are enabled, parts of SSAO and SSIL are run in parallel
+to reduce the performance impact.
```
clayjohn, line 580:
```suggestion
When both SSAO and SSIL are enabled, parts of SSAO and SSIL are shared
to reduce the performance impact.
```

contributing/development/core_and_modules/internal_rendering_architecture.rst
```diff
@@ -0,0 +1,762 @@
...
+Godot's occlusion culling doesn't support dynamic occluders yet, but
+OccluderInstance3D nodes can still have their visibility toggled or be moved.
+However, this will be slow when updating complex occluders this way. Therefore,
+updating occluders at run-time is best done only on simple occluder shapes such
+as quads or cuboids.
+
+This CPU-based approach has a few advantages over other solutions, such as
+portals and rooms or a GPU-based culling solution:
+
+- No manual setup required.
```
clayjohn, line 703:
```suggestion
- No manual setup required (but can be tweaked manually for best performance).
```

clayjohn (commented):
Excellent work! This is a fantastic start to a deep documentation of the rendering architecture. I have left a few comments, and I have tried to make suggestions wherever I had a comment. Please let me know if you have any questions about my suggestions

mhilbrunner (approved):
Fantastic addition to these contributor-facing docs. I have nothing to add besides Clay's feedback.

contributing/development/core_and_modules/internal_rendering_architecture.rst
```diff
@@ -0,0 +1,762 @@
...
+    If you have difficulty understanding concepts on this page, it is
+    recommended to go through an OpenGL tutorial such as
+    `LearnOpenGL <https://learnopengl.com/>`__.
+
+    Modern low-level APIs (Vulkan/Direct3D 12) require intermediate
+    knowledge of higher-level APIs (OpenGL/Direct3D 11) to be used
+    effectively. Many concepts are common to both kinds of graphics APIs.
+    Once you are familiar with modern OpenGL, you may follow a Vulkan tutorial
+    such as `Vulkan Tutorial <https://vulkan-tutorial.com/>`__ or
+    `Vulkan Guide <https://vkguide.dev/>`__.
```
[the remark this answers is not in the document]
Calinou: On the other hand, do we expect people to know about concepts used in modern low-level APIs (barriers, push constants, specialization constants, etc)? RenderingDevice requires knowledge of those to use the API effectively.

contributing/development/core_and_modules/internal_rendering_architecture.rst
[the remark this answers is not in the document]
Calinou: It could be worth adding a brief paragraph on why Vulkan compute is used instead of CUDA/OpenCL (which might seem like the logical choices for doing this at first), but I think shortening this makes sense.

contributing/development/core_and_modules/internal_rendering_architecture.rst
```diff
@@ -0,0 +1,762 @@
...
+
+Rendering drivers
+-----------------
+
+Godot 4 supports the following graphics APIs:
```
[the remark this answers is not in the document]
Calinou: I've put the rendering drivers section after the rendering methods one.

contributing/development/core_and_modules/internal_rendering_architecture.rst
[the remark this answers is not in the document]
clayjohn: Well, the main reason is just that we already have a vulkan context and it is way faster to use Vulkan compute then to take Vulkan GPU resources, upload them to CPU, download them back to GPU for CUDA, process on CUDA, upload back to CPU, put them in vulkan compatible format, download back to GPU for Vulkan.

That being said, in my opinion, a discussion on CUDA vs. Vulkan is not suitable for this document. CUDA isn't a viable choice for games. Our choice to not use it is not unique and does not require an explanation

contributing/development/core_and_modules/internal_rendering_architecture.rst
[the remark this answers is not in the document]
clayjohn: You need to understand many of those concepts to make best use of most APIs (modern OpenGL, WebGPU, Metal, D3D10, D3D11).

The point is, we shouldn't build an expectation that users should know how to use Vulkan in order to contribute to Godot's renderer. RenderingDevice is way easier to use than Vulkan and it requires no Vulkan-specific knowledge. We already have tons of users saying they want to contribute, but don't know Vulkan so they don't.

mhilbrunner:
Thanks for this contribution (and for addressing all the feedback :))! Merged. 🎉 Really good addition.

PR #6134, opened 2022-09-03
Fix smooth follow example in interpolation

aXu-AP:
Old method is not really frame-independent. Formula with explanation found from [here](https://www.gamedeveloper.com/programming/improved-lerp-smoothing-).

### DIFF
--- a/tutorials/math/interpolation.rst
+++ b/tutorials/math/interpolation.rst
@@ -118,8 +118,9 @@ Interpolation can be used to smooth movement, rotation, etc. Here is an example 
 
     func _physics_process(delta):
         var mouse_pos = get_local_mouse_position()
-
-        $Sprite2D.position = $Sprite2D.position.lerp(mouse_pos, delta * FOLLOW_SPEED)
+        var weight = 1 - exp(-FOLLOW_SPEED * delta)
+        
+        $Sprite2D.position = $Sprite2D.position.lerp(mouse_pos, weight)
 
  .. code-tab:: csharp
 
@@ -127,15 +128,15 @@ Interpolation can be used to smooth movement, rotation, etc. Here is an example 
 
     public override void _PhysicsProcess(float delta)
     {
-        Vector2 mousePos = GetLocalMousePosition();
-
         Sprite2D sprite = GetNode<Sprite2D>("Sprite2D");
+        Vector2 mousePos = GetLocalMousePosition();
+        float weight = 1f - Mathf.Exp(-FollowSpeed * delta);
 
-        sprite.Position = sprite.Position.Lerp(mousePos, delta * FollowSpeed);
+        sprite.Position = sprite.Position.Lerp(mousePos, weight);
     }
 
 Here is how it looks:
 
 .. image:: img/interpolation_follow.gif
 
-This useful for smoothing camera movement, allies following you (ensuring they stay within a certain range), and many other common game patterns.
+This is useful for smoothing camera movement, allies following you (ensuring they stay within a certain range), and many other common game patterns. Notice that  making movement framerate-independent is more involved than just multiplying weight by delta value.


aXu-AP:
Notice: I tested the code only in 3.x build, then changed `linear_interpolation` to `lerp`. I think there shouldn't be any other changes between 3.x and 4.0. Also C# code is untested.

TheYellowArchitect:
I tested both versions in 4.0 beta2

The version currently at the docs does interpolate "smoothly" but it jitters. Your version is better as its truly smooth without jitter. I don't like that the code is slightly more complicated (newbies will use it but won't understand it, compared to current version), but the results are worth it imo.

Vennnot:
Who can make a call on wether if we merge this or not?

tutorials/math/interpolation.rst
```diff
@@ -118,24 +118,25 @@ Interpolation can be used to smooth movement, rotation, etc. Here is an example
 
     func _physics_process(delta):
         var mouse_pos = get_local_mouse_position()
-
-        $Sprite2D.position = $Sprite2D.position.lerp(mouse_pos, delta * FOLLOW_SPEED)
```
Flarkk, line 122: I suggest to keep the current version, but to add a comment right after : 
```
// An alternative version with even smoother results with a little more math : 
// $Sprite2D.position = $Sprite2D.position.lerp(mouse_pos, 1 - exp(-FOLLOW_SPEED * delta))
```

tutorials/math/interpolation.rst
```diff
@@ -118,24 +118,25 @@ Interpolation can be used to smooth movement, rotation, etc. Here is an example
...
 
     public override void _PhysicsProcess(float delta)
     {
-        Vector2 mousePos = GetLocalMousePosition();
-
         Sprite2D sprite = GetNode<Sprite2D>("Sprite2D");
+        Vector2 mousePos = GetLocalMousePosition();
+        float weight = 1f - Mathf.Exp(-FollowSpeed * delta);
 
-        sprite.Position = sprite.Position.Lerp(mousePos, delta * FollowSpeed);
```
Flarkk, line 134: Same as above

tutorials/math/interpolation.rst
```diff
@@ -118,24 +118,25 @@ Interpolation can be used to smooth movement, rotation, etc. Here is an example
...
+        sprite.Position = sprite.Position.Lerp(mousePos, weight);
     }
 
 Here is how it looks:
 
 .. image:: img/interpolation_follow.gif
 
-This useful for smoothing camera movement, allies following you (ensuring they stay within a certain range), and many other common game patterns.
+This is useful for smoothing camera movement, allies following you (ensuring they stay within a certain range), and many other common game patterns. Notice that  making movement framerate-independent is more involved than just multiplying weight by delta value.
```
Flarkk, line 142: The final notice not only doesn't bring much information, but may add confusion. I'd rather not add it.

Flarkk:
I feel this trick worth mentioning indeed, but it's true it may be a little bit more involved for beginners.

I suggest keeping the more pedagogical current version, but mention the improved one as a comment for advanced users.

tutorials/math/interpolation.rst
AThousandShips, line 142:
```suggestion
This is useful for smoothing camera movement, allies following you (ensuring they stay within a certain range), and many other common game patterns.
Note that making movement framerate-independent is more involved than just multiplying weight by delta value.
```
"Note" is the usual way to phrase this

aXu-AP: I think I agree with @Flarkk and just not add the whole statement at the end.

tutorials/math/interpolation.rst
aXu-AP: I don't know. Kind of kills the idea of fixing the incorrect method in the first place.

Flarkk: But the original method is not really _wrong_ in itself. It just produces less smooth results, which might not be a problem in the context of a beginners tutorial.

tetrapod00:
I noticed this independently the other day too, and was considering making a PR like this. My take on the solution, with @aXu-AP as co-author, is https://github.com/godotengine/godot-docs/pull/10197.

TheYellowArchitect:
#10197 supersedes this issue, so this can be closed @mhilbrunner

PR #6142, opened 2022-09-06
Fix GDExtension summator code examples

jeffgamedev:
GDExtension documentation is outdated for the example Summator module. This PR fixes the documentation with example code for registration cpp and h files to match those of the other modules.

### DIFF
--- a/development/cpp/custom_modules_in_cpp.rst
+++ b/development/cpp/custom_modules_in_cpp.rst
@@ -119,8 +119,10 @@ These files should contain the following:
 
     /* register_types.h */
 
-    void register_summator_types();
-    void unregister_summator_types();
+    #include "modules/register_module_types.h"
+
+    void initialize_summator_module(ModuleInitializationLevel p_level);
+    void uninitialize_summator_module(ModuleInitializationLevel p_level);
     /* yes, the word in the middle must be the same as the module folder name */
 
 .. code-block:: cpp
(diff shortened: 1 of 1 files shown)


mhilbrunner:
This is in fact not a page documenting GDExtension (no docs for that exist yet outside the [introduction](https://godotengine.org/article/introducing-gd-extensions) blog post and the projects mentioned there).

This page is about the module system. Modules are included (or disabled) at compile time and are good for (optional) core parts of the engine, but cumbersome for custom additions as they require a recompile and thus maintaining your own fork.

There are also pages about the old 3.x extension system called GDNative [here](https://docs.godotengine.org/en/latest/tutorials/scripting/gdnative/what_is_gdnative.html), which is more similar to GDExtension.

For 4.0, we need to replace the docs for GDNative with new ones for GDExtension, and better explain when to use that over a module (nearly always if you're not contributing to Godot).

jeffgamedev:
> This is in fact not a page documenting GDExtension (no docs for that exist yet outside the [introduction](https://godotengine.org/article/introducing-gd-extensions) blog post and the projects mentioned there).
> 
> This page is about the module system. Modules are included (or disabled) at compile time and are good for (optional) core parts of the engine, but cumbersome for custom additions as they require a recompile and thus maintaining your own fork.
> 
> There are also pages about the old 3.x extension system called GDNative [here](https://docs.godotengine.org/en/latest/tutorials/scripting/gdnative/what_is_gdnative.html), which is more similar to GDExtension.
> 
> For 4.0, we need to replace the docs for GDNative with new ones for GDExtension, and better explain when to use that over a module (nearly always if you're not contributing to Godot).

Thank you, Max. Can we rename my PR and reopen it? The module example code is still incorrect.

YuriSizov:
Superseded by https://github.com/godotengine/godot-docs/pull/6147.

PR #6145, opened 2022-09-06
Update "Your first 2d game" UI placement doc

OmarShehata:
I believe the instructions in this section are outdated: https://docs.godotengine.org/en/stable/getting_started/first_2d_game/06.heads_up_display.html#scorelabel

Specifically, this `Layout` property does not seem to exist anymore:

```
Layout : "Top Wide"
...
Layout : "HCenter Wide"
...
etc
```

I had trouble figuring out how to get my UI to look like what was in the picture until I downloaded the demo project and found the correct values there: https://github.com/godotengine/godot-demo-projects/blob/06bdeb97dc245912998d42a8c001d83dcc557e0f/2d/dodge_the_creeps/HUD.tscn#L23

### DIFF
--- a/getting_started/first_2d_game/06.heads_up_display.rst
+++ b/getting_started/first_2d_game/06.heads_up_display.rst
@@ -70,26 +70,46 @@ use the following settings:
 ScoreLabel
 ~~~~~~~~~~
 
--  *Layout* : "Top Wide"
 -  *Text* : ``0``
 -  *Align* : "Center"
+-  *Anchor* :
+
+   -  Right: ``1``
+-  *Margin*: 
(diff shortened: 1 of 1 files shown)


YuriSizov:
> Specifically, this Layout property does not seem to exist anymore

It's not a property, it's a toolbar button, you can see it referenced a bit above the edited part of this article. It is still present as is in 3.x, and in 4.0 it's looking a bit different but is functionally the same.

OmarShehata:
Ahh, my bad! Totally missed that, just assumed it was a property amongst the other list of properties. No changed needed then!

PR #6149, opened 2022-09-06
Removed CollisionObject2D references

Greenfoot5:
As the 2nd step doesn't work with CollisionShape2D, I've removed it to match the demo https://github.com/godotengine/godot-demo-projects/tree/master/2d/dodge_the_creeps


Greenfoot5:
Having worked through the project, the mobs collide with each other. So the demo project will also need updating. I have opened an issue on the demos here: https://github.com/godotengine/godot-demo-projects/issues/773

Greenfoot5:
Ah, I've just re-read the documentation, and was looking at the wrong node type.

PR #6151, opened 2022-09-07
Fix FFmpeg filter command

DeeJayLSP:
You're supposed to use `-vf` to apply filters, not `-f:v`

(at least, in my current version of FFmpeg, `-vf` works fine while `-f:v` doesn't)


YuriSizov:
This seems to be correct, according to the CLI documentation:
```
-vf filter_graph    set video filters
```

The `-f` option is used to force a format, so it's used differently, e.g. `-f mp4`. `-f:v` doesn't seem to exist.

Calinou:
I thought `-vf` was deprecated in favor of `-f:v`.

DeeJayLSP:
Done

Calinou (approved):
Thanks!

PR #6155, opened 2022-09-09
Change estimated Godot 4 release date to H1 2023

aaronfranke:
It's currently September, which leaves three months between the first beta and release, which seems unlikely (well, could be 3.5 months if released near Christmas), so I think it's time to update the estimated release date of Godot 4.

Also, considering the history of 3.x releases, I'm wondering if we should plan for Godot 3.6 to come out in Q1 2023, but in the meantime I have not changed it in this PR.

### DIFF
--- a/about/release_policy.rst
+++ b/about/release_policy.rst
@@ -88,7 +88,7 @@ on GitHub.
 +-------------+----------------------+--------------------------------------------------------------------------+
 | **Version** | **Release date**     | **Support level**                                                        |
 +-------------+----------------------+--------------------------------------------------------------------------+
-| Godot 4.0   | Q4 2022              | |unstable| *Alpha.* Current focus of development (unstable).             |
+| Godot 4.0   | H1 2023              | |unstable| *Alpha.* Current focus of development (unstable).             |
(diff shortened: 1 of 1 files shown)


Calinou (approved):

YuriSizov:
> It's currently September, which leaves three months between the first beta and release, which seems unlikely

The next 3-4 months is still the goal though. This time around we also have some internal deadlines to meet.

akien-mga:
We're still *aiming* for EOY 2022, so I wouldn't change it.

Yes, realistically, we may not make that deadline, but that's the goal. If we slip past, it should be in early 2023. It needs to be before GDC 2023 as that's a big milestone to show Godot 4.0 to the world.

Moving the goal now won't particularly help getting contributors to focus on stabilizing the engine.

If and when it becomes obvious that we couldn't manage to ship 4.0 stable before Christmas, then I'll update the ETA.

aaronfranke:
Superseded by https://github.com/godotengine/godot-docs/pull/6842 since Godot 4.0 was released in 2023 March.

PR #6164, opened 2022-09-10
Added two new isometric autotile templates to Using Tilemaps

ghost:
These templates can be used to make a full 2x2 and 3x3 minimal isometric tileset.


ShatReal:
Since the tileset page has been updated, I don't think my change is compatible anymore. I'll redo it.

PR #6172, opened 2022-09-13
Use official CC BY 3.0 terminology and resources

Jayman2000:
This change makes the docs repo always use the official abbreviation (CC BY 3.0) for its license. Previously, it would sometimes use “CC-BY 3.0” or “CC-BY-3.0”. This change also make the docs repo always point to [the official Commons deed][1] for more information about CC BY 3.0. Previously, it would sometimes link to [an unofficial source][2].

[1]: <https://creativecommons.org/licenses/by/3.0/>
[2]: <https://tldrlegal.com/license/creative-commons-attribution-(cc)>

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -4,7 +4,7 @@ Godot Engine is developed by a community of voluntary contributors who
 contribute on all areas, including writing and maintaining the documentation.
 
 It is impossible to list them all; nevertheless, this file aims at listing
-the writers who contributed significant patches to this CC-BY licensed
+the writers who contributed significant patches to this CC BY licensed
 documentation on the `godot-docs` repository.
 
 GitHub usernames are indicated in parentheses, or as sole entry when no other
--- a/about/faq.rst
+++ b/about/faq.rst
@@ -17,7 +17,7 @@ In short:
 * You are free to modify, distribute, redistribute, and remix Godot to your heart's content, for any reason, both non-commercially and commercially.
 
 All the contents of this accompanying documentation are published under
-the permissive Creative Commons Attribution 3.0 (`CC-BY 3.0 <https://creativecommons.org/licenses/by/3.0/>`_) license, with attribution
+the permissive Creative Commons Attribution 3.0 (`CC BY 3.0 <https://creativecommons.org/licenses/by/3.0/>`_) license, with attribution
 to "Juan Linietsky, Ariel Manzur and the Godot Engine community."
 
 Logos and icons are generally under the same Creative Commons license. Note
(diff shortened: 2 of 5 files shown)


Jayman2000:
> Some reputable third party that explains the licenses without a lot of legalese is very useful, especially on the page about how you can contribute (and therefore become a part of the binding).

Doesn’t the Commons deed already explain the license without a lot of legalese though? [The Commons deed](https://creativecommons.org/licenses/by/3.0/) uses the following terms that I would consider legalese:

- license
- licensor
- technological measures (meaning DRM)
- public domain
- warranties
- publicity, privacy, or moral rights

[The tl;drLegal page](https://tldrlegal.com/license/creative-commons-attribution-(cc)#summary) uses the following terms that I would consider legalese:

- license
- Tivoization
- defamation
- sublicense
- warranty
- liable
- copyright

I would say that the Commons deed uses easier to understand vocabulary. Here’s a more direct comparison. The Commons deed says

> **No additional restrictions** — You may not apply legal terms or technological measures that legally restrict others from doing anything the license permits.

Compare that to how tl;drLegal explains the same thing:

> Doesn’t allow for Tivoization[…]

If you didn’t already know what Tivoization meant, you would have to look it up, and then rephrase the definition in your head because everyone defines Tivoization in terms of software, not in terms of other creative works.

---

Also, I’m not sure how reputable tl;drLegal is. Their page on CC BY 3.0 says

> **You must** give credit to the original author of the work, state their name and the title of the original work, say that you modified the work if you did and **include the attribution logo**, found on their website here.

I’m not aware of anything in CC BY 3.0 that requires the inclusion of a logo.

YuriSizov:
Yes, Common's page itself is quite good to ELI5 their license terms. Still, a third-party source may give a better overview and give an unbiased comparison. This is also useful for all licenses we mention, not just CC. And I wasn't suggesting we should use tldrlegal out of all other platforms.

Jayman2000:
> Yes, Common's page itself is quite good to ELI5 their license terms. Still, a third-party source may give a better overview and give an unbiased comparison. This is also useful for all licenses we mention, not just CC. And I wasn't suggesting we should use tldrlegal out of all other platforms.

What source would you recommend?

mhilbrunner (approved):

mhilbrunner:
Looks good to me. Merged. Thank you! I agree the commons deed is clear enough. If we find a even better source, I'm up for changing it, although there is something to be said about actually linking the original source for the license the content is under.

YuriSizov:
@mhilbrunner I wasn't proposing changing it. I was proposing adding an ELI5 extra. I don't have a prefer website that explains the licenses, but so it can be whatever, but I think there is merit in it.

PR #6174, opened 2022-09-13
Adds a section for books

kubarium:

### DIFF
--- a/community/tutorials.rst
+++ b/community/tutorials.rst
@@ -53,6 +53,21 @@ Text tutorials
 - `Godot Recipes by KidsCanCode <http://godotrecipes.com/>`__
 - `Steincodes <https://steincodes.tumblr.com>`__
 
+Books
+--------------
+
+- `Game Development with Blender and Godot <https://www.amazon.com/dp/1801816026/>`__, Kumsal Obuz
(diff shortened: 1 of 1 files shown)


Calinou:
A lot of those books are quite outdated, so I'm not sure if linking to them is a good idea.

Also, the title underline must match the title's length (this is a reStructuredText requirement).

kubarium:
I can remove the last two that might be outdated. The rest are 1 or 2 years old. 

I'm not sure what you meant by the title's length but I'll go and recheck the text to make sense of it. In the meantime, I'll also wait for your answer about taking the last two books off the list.

YuriSizov:
Out of curiosity, is there a way to link these books without promoting Amazon directly? Are there directories of books which are more ethically friendly?

kubarium:
It's not promoting Amazon. It's promoting the books that happen to be on Amazon as a marketplace.

The concept of directories of books makes me think it's zipped sitting somewhere on a server waiting to be downloaded. How is that ethical?

YuriSizov:
Books can be bought from places other than Amazon, though. And all books have publication identifiers, so there are directories. Google definitely has one, but that's hardly more ethical than Amazon. So I was curious if we could link to a neutral platform.

mhilbrunner:
Thanks for contributing! I'm going to close this one for now due to a lack of consensus to merge.

PR #6178, opened 2022-09-14
Add a page on decals

Calinou:
- See also https://github.com/godotengine/godot-demo-projects/pull/756.

This is only relevant for Godot 4, as Godot 3.x doesn't have decals.


tutorials/3d/using_decals.rst
```diff
@@ -0,0 +1,241 @@
...
+   in the decal's **Cull Mask** property.
+
+Creating decals during gameplay
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Decal nodes can be created from code like any other node. The main part that can
+be challenging is correctly rotating the decal so it can be visible without
+distortion.
+
+TODO
```
Calinou, line 120: I'm having trouble with handling positioning and rotation based on a RayCast's in this section: https://github.com/godotengine/godot-demo-projects/pull/756#issuecomment-1247131472

tutorials/3d/using_decals.rst
```diff
@@ -0,0 +1,282 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Decal nodes can be created from a script like any other node. The main part that can
+be challenging is correctly rotating the decal so it can be visible without
+distortion.
+
+The following code snippet creates a decal at the current mouse cursor location
+when the space bar is pressed:
+
+.. TODO: Add code sample for rotating the decal based on the raycast collision normal.
```
Calinou, line 126: I couldn't figure out this part, but we should merge this documentation before 4.0 is released anyway.

tutorials/3d/using_decals.rst
```diff
@@ -0,0 +1,282 @@
...
+   your dynamic objects' MeshInstance3D nodes to change their visibility layers.
+   For instance, you can move them from layer 1 to layer 2, then disable layer 2
+   in the decal's **Cull Mask** property.
+
+Creating decals during gameplay
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Decal nodes can be created from a script like any other node. The main part that can
+be challenging is correctly rotating the decal so it can be visible without
+distortion.
```
clayjohn, line 121:
```suggestion
Decal nodes can be created from a script like any other node.
```
I wouldn't mention this. It seems out of place for the documentation

tutorials/3d/using_decals.rst
```diff
@@ -0,0 +1,282 @@
...
+
+                add_child(decal)
+
+                decal.position = result["position"]
+                # Rotate the decal to match the camera's orientation.
+                # Depending on your use case, this may make more sense than using the collision normal,
+                # but it can lead to distortion if the camera is nearly parallel to a surface.
+                # To prevent distortion, you'd have to use the collision normal instead to orient the decal
+                # (`result["normal"]`).
+                decal.transform.basis = camera.global_transform.basis
```
clayjohn, line 157: I think it may be more instructive to have a simpler example. This example is pretty complex and isn't really useful enough to be used in a game as it doesnt cover rotation to align with the normal of the surface. 

I would instead do something that is short and works out of the box. Like leaving footprints behind a player

clayjohn (commented):
Great work! This is coming along very nicely

tutorials/3d/using_decals.rst
Calinou: I've simplified the code sample but kept its behavior identical. Hopefully, it should be short enough now.

clayjohn: Its not really a matter of length, its a matter of its use as an educational tool. This example is unnecessarily complex for the thing that it is trying to teach. 

The point this section makes is that decals can be dynamically added through code. But the example we give is more like "how to raycast from the mouse position into the scene". The only decal-specific code is:
```
var decal = Decal.new()
decal.texture_albedo = preload("res://path/to/decal_texture.png")
```

I understand that you are trying to insert a real world example in here (which is a great idea), but the example isn't useful in real world projects as it omits rotating based on surface normal.

IMO code examples should either be instructive, or they should be usable in a real project, as-is this code example is neither.

I suggest removing it for the time being, or replacing it with something that is short and self-contained

mhilbrunner:
@Calinou Rebase ping, sorry :)

tutorials/3d/using_decals.rst
Calinou: I've removed the section entirely for now. I'll create an issue to track its reimplementation.

clayjohn (approved):

akien-mga:
Thanks!

PR #6184, opened 2022-09-16
Update compiling for the web

dzil123:
Update the page to follow the rename of the javascript platform to web for Godot 4.

I am unsure how to mention `production=yes` and `use_closure_compiler=yes`, given that the original doesn't mention LTO.

Source:
https://github.com/godotengine/godot-build-scripts/blob/872c9bfe0ebdc032e6ad48f5bdda759d4e024041/build-release.sh#L274-L287
https://github.com/godotengine/godot-build-scripts/blob/872c9bfe0ebdc032e6ad48f5bdda759d4e024041/build-web/build.sh#L10-L16

### DIFF
--- a/development/compiling/compiling_for_web.rst
+++ b/development/compiling/compiling_for_web.rst
@@ -33,76 +33,67 @@ usually configured by the Emscripten SDK, e.g. when invoking ``emsdk activate``
 and ``source ./emsdk_env.sh``/``emsdk_env.bat``.
 
 Open a terminal and navigate to the root directory of the engine source code.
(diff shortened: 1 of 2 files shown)


mhilbrunner:
For future reference, relevant PRs are https://github.com/godotengine/godot/pull/65023 and https://github.com/godotengine/godot-build-scripts/pull/65.

> I am unsure how to mention production=yes and use_closure_compiler=yes, given that the original doesn't mention LTO.

I think we can merge this one as is, the latter could likely be done with a simple "pass this to minify JS" sentence, the former I'm not 100% sure about. Anyway, in it goes!

Thanks for contributing to Godot's docs 🎉

PR #6186, opened 2022-09-16
Improve Fixing jitter and stutter documentation

Calinou:
- Add section on input lag, containing project and hardware/OS configuration recommendations.
- Remove outdated advice about graphics drivers on Linux.
- Mention that compositing can sometimes be disabled on Linux for better performance.
- Wrap to 80 characters.

See https://www.reddit.com/r/godot/comments/xd6a3z/reduce_input_latency/.

This PR can be redone for `3.6` after we reach an agreement on this one (some setting names have changed).

### DIFF
--- a/tutorials/rendering/jitter_stutter.rst
+++ b/tutorials/rendering/jitter_stutter.rst
@@ -1,17 +1,25 @@
 .. _doc_jitter_stutter:
 
-Fixing jitter and stutter
-=========================
+Fixing jitter, stutter and input lag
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):

Calinou:
I've pushed a small update that adds a recommendation as to mouse DPI settings. This is a hardware-specific tweak that must be done on the player's side, but I figure I'd mention it just in case.

PR #6205, opened 2022-09-20
Fix doctool command

konczg:
This PR fixes the doctool command in the _Writing custom documentation_ section of the _Custom modules in C++_ page. The example incorrectly contains the `bin` folder as its current working directory, while the target path already contains that folder. The `bin` folder should be removed from the cwd path, as `doctool` was designed to be run in the main source folder.

### DIFF
--- a/development/cpp/custom_modules_in_cpp.rst
+++ b/development/cpp/custom_modules_in_cpp.rst
@@ -573,7 +573,7 @@ Run command:
 
    ::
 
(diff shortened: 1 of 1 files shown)


BastiaanOlij (approved):

mhilbrunner (approved):

mhilbrunner:
Thank you, and congrats on your first contribution to Godot's docs! 🎉

PR #6211, opened 2022-09-21
Fix references to the @tool annotation

Evanaellio:
Fix some docs that were still calling it the "tool keyword".


mhilbrunner:
Thank you!

Issue #5618, opened 2022-02-16
Create a complete and concise documentation for GDExtension


nonunknown:
### Describe the project you are working on

A G.O.A.P Implementation for Godotengine via GDExtension

### Describe the problem or limitation you are having in your project

Me and a lot of users has little to none knowledge about GDNative and C++, now that GDNative is "deprecated" in 4.x we have even less knowledge about such powerful tool, some of them:

* Classes we have access to
* Classes which are not exposed to API but we can use in C++ side

### Describe the feature / enhancement and how it helps to overcome the problem or limitation

The documentation will have a "Zero to Hero" object, where a user with basic knowledge of GDExtension will learn:

* Getting the necessary files to work on
* What is each file in there does
* Generate the bindings (for each platform, and cross-platform)
* Creating the First Extension
* `Ref<>` and `*` how and when to use them
* Similarities with the core source code
   - How it works in core, and how to reproduce in GDExtension
   - How to deal with virtuals (overriding a C++ function via GDScript, like `_ready()`)
 * About Classes from GodotEngine core we can use in C++ but are not exposed in the API
 * Creating a simple game
 * Creating a plugin
 * Overriding/Extending Engine features (if possible)
 * Compiling your extensions
 * Advanced Stuff
    - Optimization
    - Debugging
    - File Structure
    - Many More
 

I may have forgot some stuff, so if you're reading this, please comment below suggestions

### Describe how your proposal will work, with code, pseudo-code, mock-ups, and/or diagrams

A Complete documentation in the docs page

### If this enhancement will not be used often, can it be worked around with a few lines of script?

Nope, its a documentation.

### Is there a reason why this should be core and not an add-on in the asset library?

this cant be addon, since is GDExtension.

nonunknown:
yep I got it, but your arfimation "feature is not documented - nobody knows about it" is not true, I know little about C++, also I had little knowledge about GDNative, and even with that I was able to get more or less how GDExtension works, I cant even understand its source code, neither from godot engine. Its all about dedication in the end of the day!

KiwwiPity:
Who is the developer/contributor responsible of GDExtension?

Xrayez:
> @Xrayez mentioned that GDExtensions could throw some issues for addon/plugin developers (recompilation or something iirc).

[...]

Issue #5121, opened 2021-07-29
Documentation pages to add or update for Godot 4.0


Calinou:
## New pages

Pages to write for features that were not present in Godot 3.x.

- [ ] 3D particle systems
  - We don't have a page dedicated to 3D particles yet. This page should mention how to set up attractors, collision and trails. Setting up manual emission in a shader should be done in a dedicated page about particle shaders, since it's an advanced feature.
- [ ] Occlusion culling (using the raster-based approach, different from portals and rooms in `3.x`)
- [ ] Decals
  - See https://youtu.be/CrOI5HUAPCg?t=9760 for inspiration.
- [ ] https://github.com/godotengine/godot-docs/issues/4960

## Pages to rewrite

These pages need rewrites from scratch with new screenshots.

- [ ] 2D lights and shadows
  - Mention specular mapping.
- [ ] 3D lights and shadows
  - Mention the soft shadow properties.
  - Document setting up projectors for both spot and omni lights (omni lights require panorama textures).
  - Mention performance tuning and project settings.
- [ ] 2D particle systems
  - Mention how to set up attractors, collision and trails. Setting up manual emission in a shader should be done in a dedicated page about particle shaders, since it's an advanced feature.
- [ ] Voxel GI (replaces the old GI Probes page)
- [ ] Baked lightmaps (mention the GPU lightmapper and LightmapGI node)

pycbouh:
You can add almost every page on the editor as they all need new screenshots and in some cases updated descriptions too.

Calinou:
> You can add almost every page on the editor as they all need new screenshots and in some cases updated descriptions too.

Indeed, although I think tracking screenshot updates on existing pages should be done in a separate tracker issue.

skyace65:
I already have a PR in progress for lights and shadows. #4570. I'll add in the extra stuff you've mentioned here, though I still need help with fog fade and transmittance bias.

The said does it make sense to work on stuff for 4.0 now? There's no alpha yet so the UI can still change. This PR you made (https://github.com/godotengine/godot/pull/48561) is still waiting to get merged and I wouldn't be surprised to see more UI changes between now and an alpha.

and-rad:
@Calinou What do you want us to do if we stumble across outdated manual pages? Mention them here to get them added to the list or open a new issue, possibly tagging them with Godot 4.0? I'm currently running through the `latest` documentation and am finding a lot of outdated info. I don't want that to go to waste.

[...]

PR #6212, opened 2022-09-22
Addition of GDExtension docs

paddy-exe:
Aims to implement https://github.com/godotengine/godot-docs/issues/5618 in its basic form (more can be added later in additional PRs)

Part of https://github.com/godotengine/godot-docs/issues/5121

What this PR explains:

- [x]  Getting the necessary files to work on
- [x]  What each file in there does
- [x]  Generate the bindings (for each platform, and cross-platform)
- [x]  Creating the First Extension
- [x] Overriding/Extending Engine features (if possible)
- [x] Compiling your extension
- [x] Adding properties to the Node you are extending from
    - [x] Adding a property hint range to show one option for further customizability
- [x] Signals


_tools/redirects/redirects.csv
```diff
@@ -1,8 +1,8 @@
 source,destination
 /classes/_classes.html,/classes/
 /community/tutorials/3d/mesh_generation_with_heightmap_and_shaders.html,/tutorials/3d/mesh_generation_with_heightmap_and_shaders.html
-/community/tutorials/gdnative/gdnative-c-example.html,/tutorials/plugins/gdnative/gdnative-c-example.html
-/community/tutorials/gdnative/index.html,/tutorials/plugins/gdnative/index.html
+/community/tutorials/gdextension/gdextension-c-example.html,/tutorials/plugins/gdextension/gdextension-c-example.html
+/community/tutorials/gdextension/index.html,/tutorials/plugins/gdextension/index.html
```
mhilbrunner, line 5: This should not touch existing redirects, but add new ones at the bottom of the file.
Existing redirects still need to be setup a work for old branches.

development/compiling/compiling_for_web.rst
```diff
@@ -61,10 +61,11 @@ And ``webassembly_debug.zip`` for the debug template::
 
     mv bin/godot.javascript.opt.debug.zip bin/webassembly_debug.zip
 
-Threads and GDNative
+Threads and GDExtension
 --------------------
```
mhilbrunner, line 65:
```suggestion
Threads and GDExtension
--------------------------
```

getting_started/step_by_step/scripting_languages.rst
```diff
@@ -125,21 +125,21 @@ or ClojureCLR. However, C# is the only officially supported .NET option.
           in GDScript, C#, or C++ won't have a significant impact on
           performance.
 
-C and C++ via GDNative
+C and C++ via GDExtension
 ~~~~~~~~~~~~~~~~~~~~~~
```
mhilbrunner, line 129: You need to adapt the headlines/underlines to the changed lengths :)

paddy-exe:
@mhilbrunner Since your last commit, I noticed that the SConstruct file you put in the folder structure doesn't actually run. Also, it is quite complex to understand. Since the build system was revamped the SConstruct doesn't need to that complicated anymore. At least for my template I have it way simpler so my reworked version would be this:
```python
#!/usr/bin/env python
import os
import sys

env = SConscript("godot-cpp/SConstruct")

# For the reference:
# - CCFLAGS are compilation flags shared between C and C++
# - CFLAGS are for C-specific compilation flags
# - CXXFLAGS are for C++-specific compilation flags
# - CPPFLAGS are for pre-processor flags
# - CPPDEFINES are for pre-processor defines
# - LINKFLAGS are for linking flags

# tweak this if you want to use different folders, or more folders, to store your source code in.
env.Append(CPPPATH=["src/"])
sources = Glob("src/*.cpp")

if env["platform"] == "macos":
    library = env.SharedLibrary(
        "demo/bin/osx/libgdexample.{}.framework".format(
            env["platform"],
        ),
        source=sources,
    )
elif env["platform"] == "windows":
    library = env.SharedLibrary(
        "demo/bin/win64/libgdexample.{}.{}{}".format(env["platform"], env["arch"], env["SHLIBSUFFIX"]),
        source=sources,
    )
else:
    library = env.SharedLibrary(
        "demo/bin/linux/libgdexample.{}.{}{}".format(env["platform"], env["arch"], env["SHLIBSUFFIX"]),
        source=sources,
    )

Default(library)

```

And the `.gdextension` file:
```
[configuration]

entry_symbol = "example_library_init"

[libraries]

linux.64="res://bin/linux/libgdexample.linux.64.so"
windows.x86_64="res://bin/win64/libgdexample.windows.x86_64.dll"
macos="res://bin/osx/libgdexample.macos.framework"
```

Imho this would be clearer for new users. What do you think?

tutorials/scripting/gdextension/what_is_gdextension.rst
```diff
@@ -0,0 +1,103 @@
...
+
+Version compatibility
+---------------------
+
+(:ref:`Unlike Godot itself <doc_release_policy>`, GDExtension has stricter version
+compatibility requirements as it relies on low-level *ptrcalls* to function.) -> TODO: check up on this
+
+GDExtension add-ons compiled for a given Godot version are only guaranteed to work
+with the same minor release series. For example, a GDExtension add-on compiled for
+Godot 4.0 will only work with Godot 4.0, 4.0.1, 4.0.2… but not Godot 3.5 or 3.6.
```
Zireael07, line 103: Nitpick: I'd rephrase to make it clearer that 3.5 and 3.6 don't have GDExtension at all (unless I missed something, GDNative was NOT backported to 3.x branch)

Zireael07:
Hmm. If all that GDExtension does is call on a dll, could I have some logic NOT using any Godot specific stuff e.g. in Golang and call on it WITHOUT having any Golang bindings?

paddy-exe:
> Hmm. If all that GDExtension does is call on a dll, could I have some logic NOT using any Godot specific stuff e.g. in Golang and call on it WITHOUT having any Golang bindings?

Good question! Not sure how that applies to other languages but I think in Python used with GDNative it was possible to use some modules assuming they were imported properly.

tutorials/scripting/gdextension/what_is_gdextension.rst
paddy-exe: Good point👍🏻 I made it clearer that GDExtension isn't compatible with GDNative

Zireael07:
Something that just occurred to me, and which IIRC is also missing from GDNative docs in 3.5, what's the debugger/remote tree situation? I expect to have to use the debugger for the language I'm using, but what about remote tree variables if I use a gdns straight on a node, e.g. the player, as is the case with many of the demo projects out there?

paddy-exe:
> Something that just occurred to me, and which IIRC is also missing from GDNative docs in 3.5, what's the debugger/remote tree situation? I expect to have to use the debugger for the language I'm using, but what about remote tree variables if I use a gdns straight on a node, e.g. the player, as is the case with many of the demo projects out there?

I am not sure what you mean with your question. Could you elaborate?

Zireael07:
Literally what I said. Suppose I use a gdns file for my player node. Can I see e.g. the player's position or speed in the Godot remote tree?

Zireael07:
Ah, I was confused (again) by the fact you started from renaming the GDNative docs :P of course I meant Extension

tutorials/scripting/gdextension/what_is_gdextension.rst
```diff
@@ -0,0 +1,104 @@
...
+.. _doc_what_is_gdnative_third_party_bindings:
+
+The bindings below are developed and maintained by the community:
+
+.. Binding developers: Feel free to open a pull request to add your binding if it's well-developed enough to be used in a project.
+.. Please keep languages sorted in alphabetical order.
+
+- `D <https://github.com/godot-d/godot-d>`__
+- `Kotlin <https://github.com/utopia-rise/godot-kotlin-jvm>`__
+- `Nim <https://github.com/pragmagic/godot-nim>`__
```
Zireael07, line 84: Nim bindings are GDNative, not GDExtension :(

paddy-exe: Should be Update now. I left Rust since they already have bindings for 4.0 (even if it's still subject to change)

tutorials/scripting/gdextension/gdextension_cpp_example.rst
```diff
@@ -0,0 +1,634 @@
...
+
+    #ifndef GDEXAMPLE_H
+    #define GDEXAMPLE_H
+
+    #include <sprite2d.hpp>
+
+    namespace godot {
+
+    class GDExample : public Sprite2D {
+        GD_CLASS(GDExample, Sprite2D)
```
yudi-azvd, line 158: Hey, I think you meant `GDCLASS` here

yudi-azvd, line 158:
```suggestion
        GDCLASS(GDExample, Sprite2D)
```

I think you meant `GDCLASS`. I got stuck here for a little while 🤓

paddy-exe: @yudi-azvd Thanks for noticing! It's updated now.

paddy-exe:
Alright, here's a little update what is already done in this PR and works (tested locally by me):

- [x]  Getting the necessary files to work on
- [x]  What is each file in there does
- [x]  Generate the bindings (for each platform, and cross-platform)
- [x]  Creating the First Extension
- [x] Overriding/Extending Engine features (if possible)
- [x] Compiling your extension
- [x] Adding properties to the Node you are extending from

What still needs to be done:
* Signals

yudi-azvd:
Is it OK for me to leave some feedback on what I've read so far? 

This PR is still a draft, maybe it's too soon and that's why I'm asking

yudi-azvd:
## Project structure and what the tutorial is building

I got a little confused on what the tutorial is building: a game with the extension inside or a extension with a little game inside. After reading the first sections a few times I finally understood it is the latter (while reading I was following the tutorial as well)

I considered myself pretty experienced in following tutorials, so I think this might be confusing for beginners too.

**Suggestion** show a directory structure highlighting the folders and files and what they are:

```sh
. gdextension_cpp_example/ # root of the extension project we're building in this tutorial
├── demo/                  # game example/demo to test the extension
├── godot-cpp/             # C++ bindings
├── src/                   # source code of the extension we are building 
└── extension_api.json     # (is it really here? I was confused on this file too)
```

You could show it somewhere in the beginning of the tutorial, it will be clearer for the reader.

## Errors and problems following the tutorial

I'm using Godot 4-beta7 I wasn't getting the effects. I removed the following line

```cpp
void MySimpleGdextension::_bind_methods() {

  ClassDB::bind_method(D_METHOD("_process"), &MySimpleGdextension::_process);
  // Godot was logging this this error:
  // ERROR: Method 'MySimpleGdextension::_process()' already registered as non-virtual.
  //     at: bind_virtual_method (src/core/class_db.cpp:323)
```

```
void MySimpleGdextension::_register_methods() {
    register_method("_process", &MySimpleGdextension::_process);
    register_property<MySimpleGdextension, float>("amplitude", &MySimpleGdextension::amplitude, 10.0);

    // These methods were not getting recognized when building the extension.
```
Maybe these errors do not concern the docs you're writing. If they are beta 7 specific things, what should I do?

## Unrelated doubt

In Godot 3.x the addons could be downloaded from the Godot Assets. Will this be the same for gdextensions? Will they be downloaded on a `addons/` folder for every project?

How will the life cycle of a gdextension look like in production? In the user side?

paddy-exe:
> ## Project structure and what the tutorial is building
> I got a little confused on what the tutorial is building: a game with the extension inside or a extension with a little game inside. After reading the first sections a few times I finally understood it is the latter (while reading I was following the tutorial as well)

Well, technically you can have it both ways. It completely depends on your usecase. If you want to develop an extension for others to use as well it would be a test project for your extension. However, if you only want to use it for your own purpose, you can optimize your work with custom stuff made with GDExtension.

> I consider myself pretty experienced in following tutorials, so I think this might be confusing for beginners too.
> 
> **Suggestion** show a directory structure highlighting the folders and files and what they are:
> 
> ```shell
> . gdextension_cpp_example/ # root of the extension project we're building in this tutorial
> ├── demo/                  # game example/demo to test the extension
> ├── godot-cpp/             # C++ bindings
> ├── src/                   # source code of the extension we are building 
> └── extension_api.json     # (is it really here? I was confused on this file too)
> ```
> 
> You could show it somewhere in the beginning of the tutorial, it will be clearer for the reader.

Good suggestion! However I am not sure if this is isn't better placed before you compile the extension so that the reader can check up on the correct order. At least that would be my personal wishlist when following this tutorial.

> 
> ## Errors and problems following the tutorial
> I'm using Godot 4-beta7 I couldn't get the sprite to move. The following snippet was causing an error:
> 
> ```c++
> void MySimpleGdextension::_bind_methods() {
>   ClassDB::bind_method(D_METHOD("_process"), &MySimpleGdextension::_process);
> 
>   // Godot was logging this this error:
>   // ERROR: Method 'MySimpleGdextension::_process()' already registered as non-virtual.
>   //     at: bind_virtual_method (src/core/class_db.cpp:323)
> ```
> 
> Then I removed the line `ClassDB::bind_method(D_METHOD("_process")...`, reloaded the demo game in Godot and the sprite started moving as expected.

How does your `gdexample.h` file look like? You should add the override keyword before the `;` like this:
`void _process(double delta) override;`

> 
> I was also getting an compilation error from Scons:
> 
> ```c++
> void MySimpleGdextension::_register_methods() {
>     register_method("_process", &MySimpleGdextension::_process);
>     register_property<MySimpleGdextension, float>("amplitude", &MySimpleGdextension::amplitude, 10.0);
> 
>     // These methods were not getting recognized when building the extension.
> ```
> 
> Maybe these errors do not concern the docs you're writing. If they are beta 7 specific problems, what should I do?

`register_method` is not a valid function call in the new API for GDExtension. You use `ClassDB::bind_method` and `ClassDB::add_property` or their Macros like e.g. `ADD_PROPERTY`.

> ## Unrelated doubt
> In Godot 3.x the addons could be downloaded from the Godot AssetsLib tab. Will this be the same for gdextensions? Will they be downloaded to a `addons/` folder for a project?

As far as I know the general workflow would be that you indeed have them in your addons folder (if downloaded from the AssetLib), however this can be adjusted by modifying the GDExtension configuration part with the file paths and moving the library. 
 
> How will the life cycle of a gdextension look like in production? In the user side? Will this be documented?

You mean how they will be developed or used? Not sure what you mean by your question.

yudi-azvd:
Sorry for the late reply. Things are getting crazy at university and writing this much takes a while.

> Good suggestion! However I am not sure if this is isn't better placed before you compile the extension so that the reader can check up on the correct order. At least that would be my personal wishlist when following this tutorial.

I would disagree, but that's ok hahah. You decide.

---

> How does your gdexample.h file look like?

I'm typing the same stuff in the tutorial, except for the class name. Yours is `GDExample` and mine is `MySimpleGdextension`. If you wanna take a look, the repo is [yudi-azvd/my-simple-gdextension](https://github.com/yudi-azvd/my-simple-gdextension).

> You should add the override keyword before the ; like this:
void _process(double delta) override;

Tried and compilation gave me the error error:

```
In file included from src/register_types.cpp:3:0:
src/mysimplegdextension.h:25:14: error: 'void godot::MySimpleGdextension::_process(float)' marked 'override', but does not override
         void _process(float delta) override;
              ^~~~~~~~
```

I'm using g++ (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0 and Python 3.11.

And `override` is not present in the tutorial:

image

See line 177. (I've just downloaded a fresh copy of this PR (not sure). Check the commit)

> `register_method` is not a valid function call in the new API for GDExtension.

Well... I could swear I've read `register_method` before, but then I downloaded a fresh copy of this PR and now it's using `bind_method`.

---

At some point, trying to fix somethings, I forgot to implement `MySimpleGdextension::set_amplitude` and `MySimpleGdextension::get_amplitude` and Scons compiled with no problems. I only noticed this because the Godot logs showed the errors. Is this behavior expected? Shouldn't we get an error like `MySimpleGdextension::get_amplitude is not implemented` from Scons?

---

From `gdextension_cpp_example.rst:512`:
> Now when the project is compiled, we'll see another property called speed.
Changing its value will make the animation go faster or slower.

This suggests that right after the Scons compilation of my gdextension I should be able to see the effects of my changes. But I have to reload the project to actually see the effects. Am i missing something?

--- 

From `gdextension_cpp_example.rst:515`:
> For this example, there is no obvious advantage of using a setter and getter.
A good reason for a setter would be if you wanted to react on the variable being changed.
If you don't need to do something like that, binding the variable is enough.

I dont think this is working. If I don't bind the setter and getter for speed, Godot logs the following error:

```
ERROR: Setter method 'MySimpleGdextension::set_speed()' not found for property 'MySimpleGdextension::speed'.
```
Even thou the methods are declared in the `.h` and implemented in the `.cpp` files. You can try to reproduce commenting out the lines in my demo project ([yudi-azvd/my-simple-gdextension](https://github.com/yudi-azvd/my-simple-gdextension))

tutorials/scripting/gdextension/gdextension_cpp_example.rst
```diff
@@ -0,0 +1,625 @@
...
+        float time_passed;
+        float time_emit;
+        float amplitude;
+    ...
+
+This time, the changes in ``gdexample.cpp`` are more elaborate. First,
+you'll need to set ``time_emit = 0.0;`` in either our ``_init`` method or in our
+constructor. We'll look at the other 2 needed changes one by one.
+
+In our ``_register_methods`` method, we need to declare our signal. This is done
```
yudi-azvd, line 564:
```suggestion
In our ``_bind_methods`` method, we need to declare our signal. This is done
```

I think you meant `_bind_methods`.

tutorials/scripting/gdextension/gdextension_cpp_example.rst
```diff
@@ -0,0 +1,625 @@
...
+
+    void GDExample::_bind_methods() {
+        ClassDB::add_property("GDExample", PropertyInfo(Variant::FLOAT, "amplitude"), "set_amplitude", "get_amplitude");
+        ClassDB::add_property("GDExample", PropertyInfo(Variant::FLOAT, "speed"), "set_speed", "get_speed");
+
+
+        ADD_SIGNAL(MethodInfo("position_changed", PropertyInfo(Variant::OBJECT, "node"), PropertyInfo(Variant::VECTOR2, "new_pos")));
+    }
+
+Here, our ``register_signal`` method can be a single call first taking the
```
yudi-azvd, line 577:
```suggestion
Here, our ``ADD_SIGNAL`` method can be a single call first taking the
```

ADD_SIGNAL?

tutorials/scripting/gdextension/gdextension_cpp_example.rst
paddy-exe: Corrected

tutorials/scripting/gdextension/gdextension_cpp_example.rst
paddy-exe: Corrected

paddy-exe:
> Sorry for the late reply. Things are getting crazy at university and writing this much takes a while.

No worries. It's the same for me 👍 

> As in my previous comment I'm using Godot 4-beta7.
> 
> > Good suggestion! However I am not sure if this is isn't better placed before you compile the extension so that the reader can check up on the correct order. At least that would be my personal wishlist when following this tutorial.
> 
> I would put it right after the `mkdir gdextension_cpp_example` block. But that's ok hahah. You decide it.

You can have a look again. I put two folder structure blocks in there now. The first one close to the point where you suggested with the general structure and later again with the structure + the files
 
> > How does your gdexample.h file look like?
> 
> I'm typing the same stuff in the tutorial, except for the extension name. Yours is `GDExample` and mine is `MySimpleGdextension`. If you wanna take a look, the repo is [yudi-azvd/my-simple-gdextension](https://github.com/yudi-azvd/my-simple-gdextension).
> 
> > You should add the override keyword before the ; like this:
> > void _process(double delta) override;
> 
> Tried it and compilation gave me the error:
> 
> ```
> In file included from src/register_types.cpp:3:0:
> src/mysimplegdextension.h:25:14: error: 'void godot::MySimpleGdextension::_process(float)' marked 'override', but does not override
>          void _process(float delta) override;
>               ^~~~~~~~
> ```
> 
> I'm using g++ (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0 and Python 3.11.
> 
> And `override` **is not** present in the tutorial:
> 
> image
> 
> See line 177. (I've just downloaded a fresh copy of this PR (not sure). Check the commit)
> 
> So I commented out the binding of the `_process` method and the sprite started moving again.

Okay after further investigation it seems like it works for me now as well without the `override` keyword. The next push will update this. 

> 
> > `register_method` is not a valid function call in the new API for GDExtension.
> 
> Well... I could've swear I've read `register_method` before, but then I downloaded a fresh copy of this PR and now it's using `bind_method`. Anyways, `bind_method` works fine as expected.

Yeah I might have forgot to rename an occurence of the `register_method`.

> ## Then I advanced a little more in the tutorial
> At some point, trying to fix somethings, I forgot to implement `MySimpleGdextension::set_amplitude` and `MySimpleGdextension::get_amplitude` and Scons compiled with no problems. I only noticed this because the Godot logs showed the errors.
> 
> Is this behavior expected? Shouldn't we get an error like `MySimpleGdextension::get_amplitude is not implemented` from Scons?

From what I learned up until now in C++ is that you sometimes don't get compilation errors but runtime errors. Seems like this is this type of error. I am unfortunately not advanced in C++ enough to know why this doesn't throw an error before.

> From `gdextension_cpp_example.rst:515`:
> 
> > For this example, there is no obvious advantage of using a setter and getter.
> > A good reason for a setter would be if you wanted to react on the variable being changed.
> > If you don't need to do something like that, binding the variable is enough.
> 
> I dont think this is working. If I don't bind the setter and getter for speed, Godot logs the following error:
> 
> ```
> ERROR: Setter method 'MySimpleGdextension::set_speed()' not found for property 'MySimpleGdextension::speed'.
> ```
> 
> Even thou the methods are declared in the `.h` and implemented in the `.cpp` files. You can try to reproduce commenting out the lines in my demo project ([yudi-azvd/my-simple-gdextension](https://github.com/yudi-azvd/my-simple-gdextension))

You are absolutely right. I noticed this as well when trying to get Properties to work correctly. You either have to use setters and getters or `_set()`, `_get()` and `_get_properties_list()`.  I deleted this paragraph since it was a remnance of the GDNative tutorial, it seems.

> From `gdextension_cpp_example.rst:512`:
> 
> > Now when the project is compiled, we'll see another property called speed.
> > Changing its value will make the animation go faster or slower.
> 
> This suggests that right after the Scons compilation of my gdextension I should be able to see the effects of my changes. In this case, I should be able to see the `speed` property in the inspector, right?. But I have to reload the project to actually see it. Am i missing something?

No, you still have to restart the editor when reloading the extension unfortunately. The Godot team is already aware of this. See here: https://github.com/godotengine/godot-cpp/issues/955

> Note that the Sconstruct that I'm using is a little different compared to the one in [#6212 (comment)](https://github.com/godotengine/godot-docs/pull/6212#issuecomment-1304912829). I had to add the lines [L15-L26](https://github.com/yudi-azvd/my-simple-gdextension/blob/main/SConstruct#L15-L26) because Scons couldn't find `sprite2d.hpp`.

Hmm very weird... 🤔 I used the same SConstruct for my GDExtension template and it compiles for Linux, Windows and Mac without issues. Can you check if you can run this project here: https://github.com/paddy-exe/GDExtensionSummator?
This would be worth having a look at before merging these docs. 

> If you cant reproduce the issues I'm trying present, that's ok. I may have not followed the tutorial to the letter and maybe I've messed up something else. Anyway, thanks for the patience.

No problem. The tutorial is meant to give a succesful result back if it's followed to the letter so thanks for pointing out all the things that still need to be clarified!

tutorials/scripting/gdextension/gdextension_cpp_example.rst
```diff
@@ -0,0 +1,649 @@
...
+To generate and compile the bindings, use this command (replacing ``<platform>``
+with ``windows``, ``linux`` or ``macos`` depending on your OS):
+
+To speed up compilation, add `-jN` at the end of the SCons command line where `N`
+is the number of CPU threads you have on your system. The example below uses 4 threads.
+
+.. code-block:: none
+
+    cd godot-cpp
+    scons platform=<platform> -j4
```
yudi-azvd, line 117:
```suggestion
    scons platform=<platform> -j4 custom_api_file=<PATH_TO_FILE>
```

It makes it clearer for the reader where the `extension_api.json` is used

tutorials/scripting/gdextension/gdextension_cpp_example.rst
```diff
@@ -0,0 +1,649 @@
...
+build files in a subsequent tutorial.
+
+.. note::
+
+    This ``SConstruct`` file was written to be used with the latest ``godot-cpp``
+    master, you may need to make small changes using it with older versions or
+    refer to the ``SConstruct`` file in the Godot 4.0 documentation.
+
+Once you've downloaded the ``SConstruct`` file, place it in your GDExtension module
+folder besides ``godot-cpp``, ``src`` and ``demo``, then run:
```
yudi-azvd, line 341:
```suggestion
folder beside ``godot-cpp``, ``src`` and ``demo``, then run:
```
Maybe you meant "alongside" or "beside". You can use "next to" as well

yudi-azvd:
> You can have a look again. I put two folder structure blocks in there now. The first one close to the point where you suggested with the general structure and later again with the structure + the files

Looks better now 🤓 

> From what I learned up until now in C++ is that you sometimes don't get compilation errors but runtime errors. Seems like this is this type of error. I am unfortunately not advanced in C++ enough to know why this doesn't throw an error before.

I was suspecting as much. Thanks for the clarification

> No, you still have to restart the editor when reloading the extension unfortunately. The Godot team is already aware of this. See here: https://github.com/godotengine/godot-cpp/issues/955

Got it. Will take a better look at this later

> Hmm very weird... thinking I used the same SConstruct for my GDExtension template and it compiles for Linux, Windows and Mac without issues. Can you check if you can run this project here: https://github.com/paddy-exe/GDExtensionSummator?
This would be worth having a look at before merging these docs.

I've tried GDExtensionSummator before and I dont remember any error like the one I got. I downloaded it again and it's compiling now. I'll get back to you about this soon

yudi-azvd:
I tried following the tutorial to the letter now and I still get the compilation error. **Scons can't find sprite2d.hpp**. If you wanna take a look

```sh
git clone --recursive https://github.com/yudi-azvd/gdextension_cpp_example.git
cd gdextension_cpp_example
scons plaltform=<your platform>
```

I stopped following the tutorial right after I compiled the extension for the first time.

paddy-exe:
@yudi-azvd Solved your issue and created a PR for your repo. It was an incomplete path to the `sprite2h.hpp` file

yudi-azvd:
Well now the _process function is does dot run. Still getting the error with `override` keyword. I'll try to troubleshoot on my own.

Thanks for your effort and congrats for writing this document

yudi-azvd:
> Yeah you can delete the `override` keyword. It should work without it

Ok... 😞 I had to delete the MySimpleGdextension add it back again. It's working again

paddy-exe:
@yudi-azvd No need to bind it apparently. I only bound it since I used the override keyword

scriptsengineer:
Yes, I also realized that it is not necessary to bind the process method, see my pr
https://github.com/paddy-exe/godot-docs/pull/1

paddy-exe:
@scriptsengineer You added another function that just encapsulates the `_process` contents. I pushed the update myself now. I will close your PR

scriptsengineer:
Yes, you can embed

paddy-exe:
Hey, sorry for the miscommunication from before. I had a long look at the docs now and decided it was best to leave it as it is (without the extra move function but I did erase that `_process` function notice). Since later in the tutorial there will be functions that are bound like `set_speed()` or `get_speed()` I decided it's not necessary to add the `move()` function just for the sake of it. I do appreciate the feedback though! Thanks for taking the time with the PR as well!

Zireael07:
This PR needs reviewed and merged ASAP, so that the docs are there when 4.0 stable hits.

getting_started/step_by_step/scripting_languages.rst
```diff
@@ -124,10 +124,10 @@ or ClojureCLR. However, C# is the only officially supported .NET option.
           in GDScript, C#, or C++ won't have a significant impact on
           performance.
 
-C and C++ via GDExtension
+C++ via GDExtension
 ~~~~~~~~~~~~~~~~~~~~~~~~~
```
mhilbrunner, line 128: Header underline length needs to be adapted

tutorials/platform/android/android_plugin.rst
```diff
@@ -133,19 +133,19 @@ From your script::
         print(singleton.myPluginFunction("World"))
 
 
-Bundling GDNative resources
+Bundling GDExtension resources
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
mhilbrunner, line 137: Likewise needs the underline to be adapted

tutorials/scripting/gdextension/files/cpp_example/SConstruct
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python
```
mhilbrunner, line 1: I am not sure the docs are the best place to host and maintain this file.

If this doesn't already exist i.e. in the GDExtension repository/example, we should create one, or maybe add it to the demo repository.

tutorials/scripting/gdextension/files/cpp_example/SConstruct
```diff
@@ -0,0 +1,37 @@
...
+import os
+import sys
+
+env = SConscript("godot-cpp/SConstruct")
+
+# For the reference:
```
mhilbrunner, line 7:
```suggestion
# For reference:
```

tutorials/scripting/gdextension/gdextension_cpp_example.rst
```diff
@@ -0,0 +1,643 @@
+.. _doc_gdextension_cpp_example:
+
+GDExtension C++ example
+====================
```
mhilbrunner, line 4: Underline length isn't correct

tutorials/scripting/gdextension/gdextension_cpp_example.rst
```diff
@@ -0,0 +1,643 @@
...
+
+You can download the included example in the test folder of the godot-cpp repository `on
+GitHub <https://github.com/godotengine/godot-cpp>`__.
+
+Setting up the project
+----------------------
+
+There are a few prerequisites you'll need:
+
+- a Godot 4.x executable,
```
mhilbrunner, line 21:
```suggestion
- a Godot 4 executable,
```

tutorials/scripting/gdextension/gdextension_cpp_example.rst
```diff
@@ -0,0 +1,643 @@
...
+- a copy of the `godot-cpp
+  repository <https://github.com/godotengine/godot-cpp>`__.
+
+See also :ref:`Compiling <toc-devel-compiling>` as the build tools are identical
+to the ones you need to compile Godot from source.
+
+You can download these repositories from GitHub or let Git do the work for you.
+Note that these repositories have different branches for different versions
+of Godot. GDExtension modules written will not work in older versions of Godot (3.x) 
+and vice versa, so make sure you download the correct branch.
```
mhilbrunner, line 33: This paragraph refers to "repositories", plural, whereas the instructions immediately above only talk about a single repository.

tutorials/scripting/gdextension/gdextension_cpp_example.rst
```diff
@@ -0,0 +1,643 @@
...
+- SCons as a build tool,
+- a copy of the `godot-cpp
+  repository <https://github.com/godotengine/godot-cpp>`__.
+
+See also :ref:`Compiling <toc-devel-compiling>` as the build tools are identical
+to the ones you need to compile Godot from source.
+
+You can download these repositories from GitHub or let Git do the work for you.
+Note that these repositories have different branches for different versions
+of Godot. GDExtension modules written will not work in older versions of Godot (3.x)
```
mhilbrunner, line 32:
```suggestion
of Godot. GDExtension modules will not work in older versions of Godot (only Godot 4 and up) 
```

tutorials/scripting/gdextension/gdextension_cpp_example.rst
```diff
@@ -0,0 +1,643 @@
...
+    git submodule add -b master https://github.com/godotengine/godot-cpp
+    cd godot-cpp
+    git submodule update --init
+
+If you decide to just download the repositories or clone them into your project
+folder, make sure to keep the folder layout identical to the one described here,
+as much of the code we'll be showcasing here assumes the project follows this
+layout.
+
+Do make sure you clone recursive to pull in both repositories:
```
mhilbrunner, line 60:
```suggestion
Do make sure you clone recursively to pull in both repositories:
```

mhilbrunner (commented):
Some comments

tutorials/scripting/gdextension/gdextension_cpp_example.rst
```diff
@@ -0,0 +1,643 @@
+.. _doc_gdextension_cpp_example:
+
+GDExtension C++ example
+====================
+
+Introduction
+------------
+
+The C++ bindings for GDExtension are built on top of the C GDExtension API
+and provide a nicer way to "extend" nodes in Godot using C++. This new system allows the
```
Gallilus, line 10: "extend" any Class

Gallilus: I do not know if "any" is accurate but it's more then only the nodes.

tutorials/scripting/gdextension/what_is_gdextension.rst
```diff
@@ -0,0 +1,99 @@
...
+    Not all bindings mentioned here may be production-ready. Make sure to
+    research options thoroughly before starting a project with one of those.
+    Also, double-check whether the binding is compatible with the Godot version
+    you're using.
+
+Version compatibility
+---------------------
+
+(:ref:`Unlike Godot itself <doc_release_policy>`, GDExtension has stricter version
+compatibility requirements as it relies on low-level *ptrcalls* to function.) -> TODO: check up on this
```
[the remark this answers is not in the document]
paddy-exe: Yeah, I am not sure if this is correct. I have not enough knowledge about the C++ base system to know if that is correct

tutorials/scripting/gdextension/gdextension_cpp_example.rst
```diff
@@ -0,0 +1,643 @@
...
+    git init
+    git submodule add -b master https://github.com/godotengine/godot-cpp
+    cd godot-cpp
+    git submodule update --init
+
+If you decide to just download the repositories or clone them into your project
+folder, make sure to keep the folder layout identical to the one described here,
+as much of the code we'll be showcasing here assumes the project follows this
+layout.
+
```
Gallilus, line 58: What folder structure?
Maybe insert a graphic.

tutorials/scripting/gdextension/gdextension_cpp_example.rst
```diff
@@ -0,0 +1,643 @@
...
+we'll save it as ``main.tscn``. We'll come back to that later.
+
+Back in the top-level GDExtension module folder, we're also going to create a
+subfolder called ``src`` in which we'll place our source files.
+
+You should now have ``demo``, ``godot-cpp``, and ``src``
+directories in your GDExtension module.
+
+Your folder structure should now look like this:
+
```
Gallilus, line 145: I am not able to see the tree
image

Gallilus: PS: this is what I meant with "Maybe insert a graphic."

tutorials/scripting/gdextension/gdextension_cpp_example.rst
paddy-exe: I also removed the `modules` word as it just confuses the user (modules are a different thing entirely and shouldn't be used as synonyms for extensions)

tutorials/scripting/gdextension/gdextension_cpp_example.rst
```diff
@@ -0,0 +1,643 @@
...
+We're using the namespace ``godot``, since everything in GDExtension is defined
+within this namespace.
+
+Then we have our class definition, which inherits from our Sprite2D through a
+container class. We'll see a few side effects of this later on. The
+``GDCLASS`` macro sets up a few internal things for us.
+
+After that, we declare a single member variable called ``time_passed``.
+
+In the next block we're defining our methods, we obviously have our constructor
```
Gallilus, line 202: obviously?
A lot of people reading this will not be familiar with c++

tutorials/scripting/gdextension/gdextension_cpp_example.rst
```diff
@@ -0,0 +1,648 @@
...
+and destructor defined, but there are two other functions that will likely look
+familiar to some, and one new method.
+
+The first is ``_bind_methods``, which is a static function that Godot will
+call to find out which methods can be called and which properties it exposes. 
+The second is our ``_process`` function, which will work exactly the same 
+as the ``_process`` function you're used to in GDScript. The third is 
+our ``_init`` function which is called after Godot has properly set up
+our object.
+
```
Gallilus, line 212: What is the difference beween extensions  ``_init``  and gdscript  ``_init``

tutorials/scripting/gdextension/gdextension_cpp_example.rst
```diff
@@ -0,0 +1,648 @@
...
+Now when the project is compiled, we'll see another property called speed.
+Changing its value will make the animation go faster or slower.
+
+.. note::
+
+    For simplicity, we've left out the optional parameters in the
+    add_property() method call. These parameters are
+    ``hint``, ``hint_string``, ``usage`` and ``class_name``. These can be used to
+    further configure how properties are displayed and set on the Godot side.
+
```
Gallilus, line 551: add_property() could use a tutorial by itself if it is not already available

Gallilus (commented):
I have been out for a while and just started my GDextension.
So I hope these comments help.
It seems complete and I wish I had found this a week ago.
The only thing I can think of missing is how to document your extension.

tutorials/scripting/gdextension/files/cpp_example/SConstruct
paddy-exe: I suppose you are right. It is mainly based on the test SConstruct file from here: https://github.com/godotengine/godot-cpp/blob/master/test/SConstruct

tutorials/scripting/gdextension/gdextension_cpp_example.rst
paddy-exe: There could be a more detailed tutorial about properties but this PR focuses on having a minimum working GDExtension first.

tutorials/scripting/gdextension/gdextension_cpp_example.rst
paddy-exe: I am not sure to be quite honest. There was this dicussion comment here: https://github.com/godotengine/godot-cpp/issues/348#issuecomment-742084113 saying:
> _init is null because in the Godot API, _init is a """""virtual""""" function. Not exactly in the C++ sense though: this is a function that exists in the API because Godot expects us to be able to implement it optionally, but it doesn't actually define a default implementation itself. The function does not even exists in the engine (or sometimes it does under a different definition), it's purely a script API artifact. I have yet to find a class of the API with a virtual method you can actually call.

tutorials/scripting/gdextension/gdextension_cpp_example.rst
paddy-exe: Good catch! Thx

tutorials/scripting/gdextension/gdextension_cpp_example.rst
paddy-exe: Not sure what you mean. I have code blocks further below that showcase the file structure.

tutorials/scripting/gdextension/gdextension_cpp_example.rst
paddy-exe: I think I fixed it. There was a Line break missing

tutorials/scripting/gdextension/gdextension_cpp_example.rst
paddy-exe: There is no `any`? or do you mean you want it more specific?

tutorials/scripting/gdextension/gdextension_cpp_example.rst
Gallilus: Yes its fixed

tutorials/scripting/gdextension/gdextension_cpp_example.rst
Gallilus: I see all the demo cases extending nodes. Maybe in the tutorial, there should be some reference that other things are also able to extend.

tutorials/scripting/gdextension/gdextension_cpp_example.rst
Gallilus: I would think all `Object`s are able to implement `_init` where the constructor is called when the cpp is created and the _init is called in the godot lifecycle. I would even know how to test the difference.
Let's keep it as is to remove confusion.

tutorials/scripting/gdextension/files/cpp_example/SConstruct
```diff
@@ -0,0 +1,37 @@
...
+elif env["platform"] == "windows":
+    library = env.SharedLibrary(
+        "demo/bin/win64/libgdexample.{}.{}{}".format(env["platform"], env["arch"], env["SHLIBSUFFIX"]),
+        source=sources,
+    )
+else:
+    library = env.SharedLibrary(
+        "demo/bin/linux/libgdexample.{}.{}{}".format(env["platform"], env["arch"], env["SHLIBSUFFIX"]),
+        source=sources,
+    )
```
Faless, line 35: This will put ios and android under linux.

The only special platform is "macos", since it requires the `framework` shenanigans (and some static files which should really be generated by the godot exporter).

If you check the upstream test it does:

```python
if env["platform"] == "macos":
    # macOS magic which also requires manually creating a Info.plist
    library = env.SharedLibrary(
        "demo/bin/libgdexample.{}.{}.framework/libgdexample.{}.{}".format(
            env["platform"], env["target"], env["platform"], env["target"]
        ),
        source=sources,
    )
else:
    # Any other platform
    library = env.SharedLibrary(
        "demo/bin/libgdexample{}{}".format(env["suffix"], env["SHLIBSUFFIX"]),
        source=sources,
    )
```

tutorials/scripting/gdextension/files/cpp_example/SConstruct
BastiaanOlij: I agree with @mhilbrunner, we should host the files used here in a repo of its own or find a way to incorporate the test project from godot-cpp in here.

tutorials/scripting/gdextension/gdextension_cpp_example.rst
```diff
@@ -0,0 +1,652 @@
...
+to the ones you need to compile Godot from source.
+
+You can download this repository from GitHub or let Git do the work for you.
+Note that this repository has different branches for different versions
+of Godot. GDExtensions will not work in older versions of Godot (only Godot 4 and up) and vice versa, so make sure you download the correct branch.
+
+.. note::
+
+    `GDExtension <https://godotengine.org/article/introducing-gd-extensions>`__
+    has been merged in the ``master`` branch of godot-cpp,
```
BastiaanOlij, line 38: Hmm, this note is weird. It makes sense to have in the Godot 3 documentation to refer to needing to use a branch other than master in order to still do things with GDNative.

But I think using the master branch of godot-cpp is expected especially with a view of Godot 4 stable being released any day now.

I'd remove it or only keep `You need to use the ``master`` branch of godot-cpp to use GDExtension and follow this example.`

tutorials/scripting/gdextension/gdextension_cpp_example.rst
```diff
@@ -0,0 +1,652 @@
...
+
+.. code-block:: none
+
+    mkdir gdextension_cpp_example
+    cd gdextension_cpp_example
+    git clone --recursive -b master https://github.com/godotengine/godot-cpp
+
+.. note::
+
+    ``godot-cpp`` now includes ``godot-headers`` as a nested submodule, if you've
```
BastiaanOlij, line 70: Need to double check this, I believe that in GDExtensions we no longer maintain the `godot-headers` repo and just maintain all files in `godot-cpp`.

I think you can thus also remove the mention of needing to recursively clone.

Might want to double check this but I think this is so.

tutorials/scripting/gdextension/gdextension_cpp_example.rst
BastiaanOlij: I'm not 100% sure anymore, it's been awhile, @Faless you may be able to confirm.

In GDNative `_init` had to be defined because the wrapper code always called it. I don't think this is so anymore in GDExternal, I think it's optional.
It should now work the same as in GDScript where you define `_init` and register it in `bind_methods` and it will get called if needed.

tutorials/scripting/gdextension/gdextension_cpp_example.rst
```diff
@@ -0,0 +1,652 @@
...
+    GDExample::GDExample() {
+        // initialize any variables here
+        time_passed = 0.0;
+    }
+
+    GDExample::~GDExample() {
+        // add your cleanup here
+    }
+
+    void GDExample::_process(float delta) {
```
BastiaanOlij, line 237: I think process works differently now. Process is now defined as a virtual in `node.hpp`: `virtual void _process(double delta);`.

So you now simply override it. You also don't need to bind it, it's already bound.

BastiaanOlij: It's also important to call `_process` on the super class or any logic in the super classes will be skipped.

The same applies to a bunch of other functions, just have a look at `node.hpp`.

tutorials/scripting/gdextension/gdextension_cpp_example.rst
```diff
@@ -0,0 +1,652 @@
...
+
+        set_position(new_position);
+    }
+
+This one should be straightforward. We're implementing each method of our class
+that we defined in our header file.
+
+Note our ``_process`` function, which keeps track of how much time has passed
+and calculates a new position for our sprite using a sine and cosine function.
+What stands out is calling ``owner->set_position`` to call one of the built-in
```
BastiaanOlij, line 250: `owner` is no longer applicable I think? That was a GDNative thing.

tutorials/scripting/gdextension/gdextension_cpp_example.rst
```diff
@@ -0,0 +1,652 @@
...
+We're finally ready to run the project:
+
+.. image:: img/gdextension_cpp_animated.gif
+
+Adding properties
+-----------------
+
+GDScript allows you to add properties to your script using the ``export``
+keyword. In GDExtension you have to register the properties with a getter and
+sett function or directly implement the ``_get_property_list``, ``_get`` and
```
BastiaanOlij, line 431: `sett` -> `setter`

tutorials/scripting/gdextension/gdextension_cpp_example.rst
```diff
@@ -0,0 +1,652 @@
...
+        float get_speed();
+    ...
+
+This requires a few more changes to our ``gdexample.cpp`` file, again we're only
+showing the methods that have changed so don't remove anything we're omitting:
+
+.. code-block:: C++
+
+    void GDExample::_bind_methods() {
+        ClassDB::bind_method(D_METHOD("get_speed"), &GDExample::get_speed);
```
BastiaanOlij, line 515: might want to add some `...` to indicate we're not removing the code in the fragment above.

tutorials/scripting/gdextension/gdextension_cpp_example.rst
BastiaanOlij: I don't know, seeing we do amplitude first, we could add additional values to speed to provide a range just to indicate one possible use case. We can then change this note to indicate there are more options.

That said, I'm all for having a separate page that purely handles properties and goes through the various things that are possible with these parameters as we really lack examples of each one.

tutorials/scripting/gdextension/gdextension_cpp_example.rst
```diff
@@ -0,0 +1,652 @@
...
+
+.. code-block:: GDScript
+
+    extends Node
+
+    func _on_Sprite2D_position_changed(node, new_pos):
+        print("The position of " + node.get_class() + " is now " + str(new_pos))
+
+Every second, we output our position to the console.
+
```
BastiaanOlij, line 646: Similar to what was discussed on properties, we really need a follow up page that goes more in depth on signals and specifically how to connect to signals from GDExtensions as the callables syntax is not something one would easily figure out on their own. But that should be a follow up PR

tutorials/scripting/gdextension/what_is_gdextension.rst
```diff
@@ -0,0 +1,99 @@
...
+
+They also both allow you to integrate third-party libraries into Godot. The one
+you should choose depends on your needs.
+
+Advantages of GDExtension
+^^^^^^^^^^^^^^^^^^^^^^
+
+Unlike modules, GDExtension doesn't require compiling the engine's source code,
+making it easier to distribute your work. It gives you access to most of the API
+available to GDScript C#, allowing you to code game logic with full control
```
BastiaanOlij, line 30: to GDscript **and** C#

tutorials/scripting/gdextension/what_is_gdextension.rst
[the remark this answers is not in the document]
BastiaanOlij: Time will tell how well we implemented things but GDExtensions were created as resilient as possible. 
As long as we add to the API in Godot and not remove functions, GDExtensions targeting older versions of Godot will likely work on newer versions. But it is very likely extensions will have to be altered and recompiled when going from say Godot 4.0 to 4.1.

That said, no guarantees here as it's easy to do something in Godot that will break extensions without any checks in place to prevent us from doing so.

BastiaanOlij:
Generally looks good to me, just a few things that I guess were translated from the GDNative version but now work slightly differently in GDExtensions.

tutorials/scripting/gdextension/gdextension_cpp_example.rst
paddy-exe: You are right about this. I will remove this then

tutorials/scripting/gdextension/gdextension_cpp_example.rst
paddy-exe: Updated it with this: 
> nicer way to "extend" nodes and other built-in classes in Godot using C++.

tutorials/scripting/gdextension/gdextension_cpp_example.rst
paddy-exe: @Gallilus @BastiaanOlij 
Alright, I added a property hint for the `speed` property. This should do for a simple example. I might be able to go into further detail in a follow-up PR

paddy-exe:
I worked in the feedback, added a property hint example (with `PROPERTY_HINT_RANGE`), squashed all commits and rebased my changes on master

tutorials/scripting/gdextension/gdextension_cpp_example.rst
BastiaanOlij: Looks good, its enough to peek someones interest and give atleast one example of how to use hints.

BastiaanOlij:
Looking pretty good so far to me.

mhilbrunner (approved):
Looks good to me, gave it another read and spotted nothing else!

Could you convert the .pngs to .webp? We decided to change format for the docs.
Besides that, I'm happy to merge this. :)

paddy-exe:
@mhilbrunner Awesome! I converted the `.png`'s to `.webp` now :D should be ready to merge

_tools/redirects/redirects.csv
```diff
@@ -376,9 +376,9 @@ source,destination
 /tutorials/platform/android_in_app_purchases.html,/tutorials/platform/android/android_in_app_purchases.html
 /tutorials/plugins/android/android_plugin.html,/tutorials/platform/android/android_plugin.html
 /tutorials/plugins/android/index.html,/tutorials/platform/android/index.html
-/tutorials/plugins/gdnative/gdnative-c-example.html,/tutorials/scripting/gdnative/gdnative_c_example.html
-/tutorials/plugins/gdnative/gdnative-cpp-example.html,/tutorials/scripting/gdnative/gdnative_cpp_example.html
-/tutorials/plugins/gdnative/index.html,/tutorials/scripting/gdnative/index.html
+/tutorials/plugins/gdextension/gdextension-c-example.html,/tutorials/scripting/gdextension/gdextension_c_example.html
+/tutorials/plugins/gdextension/gdextension-cpp-example.html,/tutorials/scripting/gdextension/gdextension_cpp_example.html
+/tutorials/plugins/gdextension/index.html,/tutorials/scripting/gdextension/index.html
```
YuriSizov, line 381: This is still wrong, shouldn't be touched.

_tools/redirects/redirects.csv
```diff
@@ -402,3 +402,7 @@ source,destination
 /tutorials/physics/using_kinematic_body_2d.html,/tutorials/physics/using_character_body_2d.html
 /tutorials/plugins/editor/spatial_gizmos.html,/tutorials/plugins/editor/3d_gizmos.html
 /tutorials/3d/procedural_geometry/immediategeometry.html,/tutorials/3d/procedural_geometry/immediatemesh.html
+/tutorials/plugins/gdnative/index.html,/tutorials/plugins/gdextension/index.html
```
YuriSizov, line 405:
```suggestion
/tutorials/scripting/gdnative/index.html,/tutorials/plugins/gdextension/index.html
```

_tools/redirects/redirects.csv
paddy-exe: This is odd. Should I remove these lines or the ones at the bottom of the file? They are essentially doing the same thing (if I would correct the ones you marked)

paddy-exe:
Alright, the `redirect.csv` should now be correctly done (I hope). I rebased everyting as well

_tools/redirects/redirects.csv
YuriSizov: They are not doing the same thing. These are old redirects which should be kept intact. You add new ones on top of them. The ones you've added are fine. Except this one:

```
/tutorials/scripting/gdnative/index.html,/tutorials/plugins/gdextension/index.html
```

It should be

```
/tutorials/scripting/gdnative/index.html,/tutorials/scripting/gdextension/index.html
```

YuriSizov:
Yep, redirects are now fine, good work! :)

paddy-exe:
It is done :O 😅

mhilbrunner (approved):

mhilbrunner:
Merged, finally! 🎉 Amazing work. Thanks for keeping up with all the reviews and change requests for so long.

Amazing docs contribution and nice we have that in now for 4.0 launch. :)

PR #6215, opened 2022-09-23
Update example code for Godot 4 in Using RigidBody

nerdbeard:
``Basis.xform`` no longer exists, replaced with * operator.  I also changed methods to properties like the rest of the Godot 4 examples.


Calinou (approved):
Thanks! 

PR #6221, opened 2022-09-26
Fix missing "s" at the end of a word

HerrSersch:
Change "Do not use it for files in res:// as it will not work in exported project." to "[...] projects."

### DIFF
--- a/classes/class_os.rst
+++ b/classes/class_os.rst
@@ -987,7 +987,7 @@ Kill (terminate) the process identified by the given process ID (``pid``), e.g. 
 
 Moves the file or directory to the system's recycle bin. See also :ref:`Directory.remove<class_Directory_method_remove>`.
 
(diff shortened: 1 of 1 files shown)


skyace65:
Hi, we appreciate the pull request but this change needs to be made in the main Godot repository. The class reference pages in the docs repository are generated from what's in the Godot repository. This is where the file you want to edit is https://github.com/godotengine/godot/blob/master/doc/classes/OS.xml.

You can find more information on contributing to the class reference [here](https://docs.godotengine.org/en/3.5/community/contributing/updating_the_class_reference.html).

HerrSersch:
@skyace65 Sorry for the mistake, and thank you for pointing me in the right direction. Hopefully the [new pull request](https://github.com/godotengine/godot/pull/66603) is correct :)

PR #6204, opened 2022-09-20
Fixes for "Your First 2D Game" tutorial section 5: "The main game scene"

acristin:
[Section 5 of the first 2D game tutorial](https://docs.godotengine.org/en/latest/getting_started/first_2d_game/05.the_main_game_scene.html) needed a few updates to match renames in the 4.0 beta.

Position2D -> Marker2D.

PathFollow2D "offset" with .progress_ratio, etc.

Issues reported in [#5498](https://github.com/godotengine/godot-docs/issues/5498)

Please pay extra attention to the C# and C++ changes, as I haven't verified them yet.

PR #6229, opened 2022-09-27
[2D Game Tutorial] Position2D conversion to Node2D

TheYellowArchitect:
Position2D exists in stable (3.5) https://docs.godotengine.org/en/stable/classes/class_position2d.html?highlight=position2d It does not exist in 4.x https://docs.godotengine.org/en/latest/search.html?q=position2d&check_keywords=yes&area=default https://docs.godotengine.org/en/latest/classes/class_position2d.html?highlight=position2d

So the user is now no longer confused why he cannot create this class and why its not class-highlighted either

### DIFF
--- a/getting_started/first_2d_game/05.the_main_game_scene.rst
+++ b/getting_started/first_2d_game/05.the_main_game_scene.rst
@@ -23,7 +23,7 @@ Now, add the following nodes as children of ``Main``, and name them as shown
   every second
 - :ref:`Timer <class_Timer>` (named ``StartTimer``) - to give a delay before
   starting
-- :ref:`Position2D <class_Position2D>` (named ``StartPosition``) - to indicate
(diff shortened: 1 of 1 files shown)


cbscribe:
The node's name is Marker2D now.

TheYellowArchitect:
Thank you for notifying me of this. Should I close this pull request and update it with Marker2D or keep this as Node2D?
I would argue Node2D is preferred because it fits more use-cases for the user, but I am a newbie with Godot and my opinion is uninformed

cbscribe:
The purpose of a Marker2D is that it has visibility in the editor. Node2D does not.

To be honest, I'm not sure that piecemeal like this is the best way to update this doc. The entire tutorial flow has been worked out and tested over a number of years, and now that we are in beta, it would be best to revise it as a whole.

TheYellowArchitect:
This pull request has the Marker2D conversion https://github.com/godotengine/godot-docs/pull/6204

PR #6239, opened 2022-09-28
[2D Light and Shadows Tutorial] Replaced Light2D with PointLight2D

TheYellowArchitect:
It is impossible to instantiate a Light2D node, and a DirectionalLight2D node has no texture (so this tutorial doesn't work with it at all)

Feel free to deny this commit, because I am a newbie and don't know if it was possible to instantiate a Light2D node in 3.x either (if it was possible, the tutorial should stay as is, and this pull request should be closed)

### DIFF
--- a/tutorials/2d/2d_lights_and_shadows.rst
+++ b/tutorials/2d/2d_lights_and_shadows.rst
@@ -48,7 +48,7 @@ Nodes
 The demo uses four different nodes:
   * :ref:`CanvasModulate <class_CanvasModulate>`
   * :ref:`Sprite2D <class_Sprite2D>`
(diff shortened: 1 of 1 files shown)


Calinou:
Light2D was renamed to PointLight2D in 4.0, so this is correct. There is no DirectionalLight2D in 3.x, so Light2D is not an abstract class there (and only offers point lights).

clayjohn (approved):

mhilbrunner (approved):

mhilbrunner:
Thank you, and congrats on your first merged contribution to Godot's docs! Now onwards to your other PRs... 🎉

PR #6240, opened 2022-09-28
Clean up canvas item shader reference and add new light shader built ins

clayjohn:
Update the CanvasItem shader doc based on how some things work in 4.0 and add new builtins that will be added by https://github.com/godotengine/godot/pull/66565


akien-mga (approved):

akien-mga:
Thanks!

PR #6248, opened 2022-09-29
Remove @ before export

Storms-Engineering:
Typo I assume?  It didn't work with latest godot edition and removing @ fixed the problem.


Calinou:
`@export` is an *annotation* in 4.0.beta. There are no annotations in 3.x, so there is no `@` prefix for the `export` *keyword* there.

Storms-Engineering:
Oh ok roger that thanks @Calinou !

PR #6254, opened 2022-10-02
Remove reference to date and time in OS

xarses:
Since they where moved to Time, lets remove the dangling reference.

Fixes: https://github.com/godotengine/godot/issues/66613

### DIFF
--- a/classes/class_os.rst
+++ b/classes/class_os.rst
@@ -17,7 +17,7 @@ Operating System functions.
 Description
 -----------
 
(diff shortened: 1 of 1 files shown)


classes/class_os.rst
```diff
@@ -17,7 +17,7 @@ Operating System functions.
 Description
 -----------
 
-Operating System functions. OS wraps the most common functionality to communicate with the host operating system, such as the clipboard, video driver, date and time, timers, environment variables, execution of binaries, command line, etc.
```
xarses, line 20: I wasn't sure if it was also appropriate remove the `timers, ` There are still two `dealy_` functions in OS, so I think it's partly appropriate to stay

Calinou:

PR #6263, opened 2022-10-05
fix: invalid label references in `viewports` file

yedpodtrzitko:
- use proper casing in name of various classes, otherwise it's considered an invalid label

- add missing `3D` suffix to Camera, VisualInstance classes


skyace65 (approved):
Looks good, thanks!

PR #6268, opened 2022-10-06
[Instancing with signals] Updated to 4.0

TheYellowArchitect:
https://docs.godotengine.org/en/latest/tutorials/scripting/instancing_with_signals.html
Basically started with changing `instance()` to `instantiate()`, and `MouseButton`, which give error.

Was a good chance to also upgrade the badly named `b` variables into proper intuitive names, and also make the signal emission with 4.0 syntax: `signal.emit(var1, var2)` instead of `emit_signal("signal",var1,var2)`

It's a good tutorial, short and to the point, though kinda abstract so I don't think there is anything more to add; If I put an image above the final codeblock, on how to add a signal via UI, it would ruin the abstraction. The text explaining signals are in node tab of inspector is enough

### DIFF
--- a/tutorials/scripting/instancing_with_signals.rst
+++ b/tutorials/scripting/instancing_with_signals.rst
@@ -63,7 +63,7 @@ You could do this by adding the bullet to the main scene directly:
 .. tabs::
  .. code-tab:: gdscript GDScript
 
-    var bullet_instance = Bullet.instance()
+    var bullet_instance = Bullet.instantiate()
     get_parent().add_child(bullet_instance)
 
  .. code-tab:: csharp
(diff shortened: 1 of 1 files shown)


TheYellowArchitect:
thank you for the praise but it's kinda coincidental, as I'm learning Godot via the docs, so it's natural for me to hit those small bugs, and since I solve them locally to continue, I just replicate my changes and upload (don't even need to use git since it's just text)

After all, a lot of users must have dived into Godot with 4.0 and the quality of the docs is already outstanding (best docs I have seen in a project, especially the in-editor offline one), so a 5-minute edit from me helps a lot of people and that knowledge alone is satisfying :smile:

Issue #6234, opened 2022-09-27
2D Dodge The Creeps Tutorial - Incompatible/Deprecated Code for 4.0


TheYellowArchitect:
**Your Godot version:**
4.0 beta1

**Issue description:**
https://docs.godotengine.org/en/latest/getting_started/first_2d_game/06.heads_up_display.html
Ctrl+f "get ready" for the code snippet^
![dodge-creeps6-code-error](https://user-images.githubusercontent.com/25630803/192607383-afc6b1b1-db63-463e-9e65-7d051eb48ee1.png)

I am a newbie, no idea with what to replace both in order to work as intended

Slightly off-topic (enhancement tag lol): Shouldn't this function be split in different functions? From what I have experienced, having "await X condition" in a codeblock easily creates bugs, each timer should have its own timeout function for modular&clean code

TheYellowArchitect:
The first code error (await) must become (awaitToSignal) and it gives no errors. I still cannot find a replacement for yield (creating a node/timer inside an await/yield is horrible code practice btw)

PR #6243, opened 2022-09-29
Update "Your first 3D game" tutorial for Godot 4

TheYellowArchitect:
- I updated all non-4.0-related images (e.g. showed `Translation` instead of `Position`)
- Added some much-needed images
- Fixed all code to work for 4.0 (except 1 thing, see below)
- Polished many minor things which should have been polished even in the 3.x branch, like hyperlinking/referencing class names and formating node names (instead of _italic_)

Some images are in .webp format because I couldn't get .png or .jpg files in the size of the original images. Otherwise all new images would be 2~3 times bigger like https://github.com/godotengine/godot-docs/pull/6235
So, if docs don't support .webp, don't merge this pull request.

There are so many changes, that I think it's bloated to make a changelog for each (see changed files). 
Edit: Below is fixed/solved

-----------------

~~Instead, I will present the only problem which keeps this from being merged (so I guess it's a draft because of 3 codelines lol)~~

~~Scroll down into the big codeblock of this
https://docs.godotengine.org/en/latest/getting_started/first_3d_game/06.jump_and_squash.html#coding-the-squash-mechanic~~

```python
for index in range(get_slide_count()):
  # We check every collision that occurred this frame.
  var collision = get_slide_collision(index)
  # If we collide with a monster...
  if collision.collider.is_in_group("mob"):
      var mob = collision.collider
      # ...we check that we are hitting it from above.
      if Vector3.UP.dot(collision.normal) > 0.1:
          # If so, we squash it and bounce.
          mob.squash()
          velocity.y = bounce_impulse
```

~~I converted it to this:~~
```python
# Iterate through all collisions that occured this frame
# in C this would be for(int i = 0; i < collisions.Count; i++)
for index in range(get_slide_collision_count()):
    # We get one of the collisions with the player
    var collision = get_slide_collision(index)

    # If the collision is with ground
    if (collision.get_collider(index) == null):
	return

    # If the collider is with a mob
    if collision.get_collider(index).is_in_group("mob"):
	var mob = collision.get_collider(index)
	# we check that we are hitting it from above.
	if Vector3.UP.dot(collision.get_normal(index)) > 0.1:
		# If so, we squash it and bounce.
		mob.squash()
		target_velocity.y = bounce_impulse
```

[...]

PR #6275, opened 2022-10-07
[2D Game Tutorial] Fixing the last codeblock to run on 4.0

TheYellowArchitect:
https://docs.godotengine.org/en/latest/getting_started/first_2d_game/06.heads_up_display.html

The last code which makes the tutorial incompatible for 4.0 is this. I ran it locally on my 2D project to ensure it works. Basically to replace the yield one-shot timer, I used this https://docs.godotengine.org/en/latest/classes/class_scenetreetimer.html

Fixes https://github.com/godotengine/godot-docs/issues/6234 https://github.com/godotengine/godot-docs/issues/5498

------------------------

@Calinou Related to this: If I do an upgrade/polish of the 2D tutorial [like I did with 3.0](https://github.com/godotengine/godot-docs/pull/6243) where I upload more **.webp** images and clarify and ref hyperlink here and there, would it be accepted soon, or delayed until after .webp support? Ofc alongside 4.0.zip for demo projects. I'm asking because it will take a day or two (I already have the images which is the hard part), and if that upgrade is merged in 4+ months, I would rather schedule my priorities in another way.

I know that the focus is on fixing 4.0 bugs and reaching a stable exit out of beta (both are obviously of highest importance), but I ask because there are many users who are hyped for 4.0 and get filtered at the docs, and basically I don't know what is the acceptance requirement for merging a huge pull request like remaking/polishing an entire tutorial.

### DIFF
--- a/getting_started/first_2d_game/06.heads_up_display.rst
+++ b/getting_started/first_2d_game/06.heads_up_display.rst
@@ -211,12 +211,12 @@ We also need to process what happens when the player loses. The code below will 
     func show_game_over():
         show_message("Game Over")
         # Wait until the MessageTimer has counted down.
-        await($MessageTimer, "timeout")
+        await $MessageTimer.timeout
 
         $Message.text = "Dodge the\nCreeps!"
         $Message.show()
         # Make a one-shot timer and wait for it to finish.
-        yield(get_tree().create_timer(1), "timeout")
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):

mhilbrunner:
Thanks for fixing all the code up! 👏 Awesome work.

Issue #5832, opened 2022-05-22
Lambda functions needs to be documented


Chaosus:
**Your Godot version:**
4.0

**Issue description:**
Description of that new [feature](https://github.com/godotengine/godot-proposals/issues/2431) is need to added to GDScript page.

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_basics.html

PR #6276, opened 2022-10-07
Add explanation of lambdas to GDScript basics.

HolonProduction:
Fixes: #5832

Some examples and wordings were copied from the original proposal godotengine/godot-proposals#2431.

This also contains notion of one line functions.

The proposal also has an example of the factory pattern (one function that returns a lambda). I am not sure if this belongs into this section. It could be added though.

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_basics.rst
+++ b/tutorials/scripting/gdscript/gdscript_basics.rst
@@ -892,6 +892,14 @@ argument, unlike Python).
 
 A function can ``return`` at any point. The default return value is ``null``.
 
+If a function contains only one line of code, it can be written on one line::
+
+    func square(a): return a * a
+
+    func hello_world(): print("Hello World")
+
+    func empty_function(): pass
+
 Functions can also have type specification for the arguments and for the return
 value. Types for arguments can be added in a similar way to variables::
 
@@ -928,7 +936,7 @@ return early with the ``return`` keyword, but they can't return any value.
 Referencing functions
 ^^^^^^^^^^^^^^^^^^^^^
 
-Functions are first-class items in terms of the Callable object. Referencing a
+Functions are first-class items in terms of the :ref:`Callable <class_Callable>` object. Referencing a
 function by name without calling it will automatically generate the proper
 callable. This can be used to pass functions as arguments.
 
@@ -952,6 +960,32 @@ callable. This can be used to pass functions as arguments.
           the ``()`` operator directly. This behavior is implemented to avoid
           performance issues on direct function calls.
 
+Lambda functions
+^^^^^^^^^^^^^^^^
+
+Lambda functions allow you to declare functions that do not belong to a class. Instead a :ref:`Callable <class_Callable>` object is created and assigned to a variable directly.
+This can be useful to create Callables to pass around without polluting the class scope.
+
+::
+
+    var lambda = func(x): print(x)
+    lambda.call(42) # Prints "42"
+
+Lambda functions can be named for debugging purposes::
+
+    var lambda = func my_lambda(x):
+        print(x)
+
+Lambda functions capture the local environment. Local variables are passed by value, so they won't be updated in the lambda if changed in the local function::
+
+    var x = 42
+    var my_lambda = func(): print(x)
+    my_lambda.call() # Prints "42"
+    x = "Hello"
+    my_lambda.call() # Prints "42"
+
+.. note:: The values of the outer scope behave like constants. Therefore, if you declare an array or dictionary, it can still be modified afterwards.
+
 Static functions
 ^^^^^^^^^^^^^^^^
 
@@ -961,6 +995,8 @@ useful to make libraries of helper functions::
 
     static func sum2(a, b):
         return a + b
+
+Lambdas cannot be declared static.
 
 
 Statements and control flow


TheYellowArchitect:
Looks good to me. Wish I could build the docs to ensure the code renders properly. Only nitpick is perhaps to change the 2 instances of "lambdas" to "lambda functions" but it doesn't really matter, as only someone who has never heard of lambdas will prefer "lambda functions" lol

HolonProduction:
> The user cannot understand what you mean by "frozen" values, nor "flat". At least I don't, and I have used lambdas many times in the past. Documentation should be as comprehensive and crystal-clear as possible.

I agree with you on "frozen". It is refering to the values that are copied from the outer scope. How about `The values captured from the outer scope behave like...`

The "flat" terminology was copied from the explaination of constants which was introduced before. When changing it here it should be changed there as well. (Line 845)

HolonProduction:
> wish I could build the docs to ensure the code renders properly.

I rendered it with the `make.bat` file. Is there something wrong with this method?
[gdscript_basics.zip](https://github.com/godotengine/godot-docs/files/9737621/gdscript_basics.zip)

TheYellowArchitect:
>wish I could build the docs to ensure the code renders properly.
>I rendered it with the make.bat file. Is there something wrong with this method?
[gdscript_basics.zip](https://github.com/godotengine/godot-docs/files/9737621/gdscript_basics.zip)

I thought that building the docs requires a lot of programs, kinda like building godot itself. I remember I tried to install the programs required to build godot from source and 1 of them conflicted with a program I had, so I didn't continue my installation. I see in https://docs.godotengine.org/en/latest/community/contributing/building_the_manual.html that building the docs is simple.

tl;dr I haven't tried building. I did try that link you uploaded, opened it with my browser (brave) and it didn't render like the docs.

>I agree with you on "frozen". It is refering to the values that are copied from the outer scope. How about The values captured from the outer scope behave like...

Sounds great, as I can understand it and I am certain more users as well. Though perhaps a change on the verb `captured`?
"The values of the outer scope" sounds simpler


>The "flat" terminology was copied from the explaination of constants which was introduced before. When changing it here it should be changed there as well. (Line 845)

I went into that line, and am kinda stunned. Right above is a serious problem, which dwarfs the clarifications here:
>You can also create constants inside a function, which is useful to name local magic values.

local magic values lol
Anyway, outside the scope of this PR, but had to mention.

Back on-topic: On the "flat", I honestly don't know with what phrase to replace. You did good to keep the consistency, but in this case, whoever wrote that "flat" example must have had issues expressing that concept. You can either replace both "flat" sections or just clear up yours at the lambda section, and some1 else cleans up that "flat" section when an appropriate phrasing is found.

tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -952,6 +960,32 @@ callable. This can be used to pass functions as arguments.
           the ``()`` operator directly. This behavior is implemented to avoid
           performance issues on direct function calls.
 
+Lambda functions
+^^^^^^^^^^^^^^^^
+
+Lambda functions allow you to declare functions that do not belong to a class. Instead a Callable object is created and assigned to a variable directly.
```
TheYellowArchitect, line 966: Callable having hyperlink/ref is a small addition but worth mentioning
```
Lambda functions allow you to declare functions that do not belong to a class. Instead a :ref:`Callable <class_Callable>` object is created and assigned to a variable directly.
```

tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -952,6 +960,32 @@ callable. This can be used to pass functions as arguments.
...
+
+    var lambda = func(x): print(x)
+    lambda.call(42) # Prints "42"
+
+Lambdas can be given a name for debugging purposes::
+
+    var lambda = func my_lambda(x):
+        print(x)
+
+They capture the local environment. Local variables are passed by copy, so won't be updated in the lambda if changed in the local function::
```
TheYellowArchitect, line 979: `They` to be explicitly `Lambdas` (or `Lambda functions` if you replace all `Lambdas` with `Lambda functions`)

Passed by value, instead of by copy, anyway final phrase:
```
Lambdas capture the local environment. Local variables are passed by value, so they won't be updated in the lambda if changed in the local function::
```

I haven't heard the term `local enviroment` for any kind of scope, but I think the reader can understand the meaning. Cannot think of something more specific because I am not certain if only current-scope variables are included in `local enviroment` in lambda functions (haven't used them in GDScript)

tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -952,6 +960,32 @@ callable. This can be used to pass functions as arguments.
...
+
+Lambda functions allow you to declare functions that do not belong to a class. Instead a Callable object is created and assigned to a variable directly.
+This can be useful to create Callables to pass around without polluting the class scope.
+
+::
+
+    var lambda = func(x): print(x)
+    lambda.call(42) # Prints "42"
+
+Lambdas can be given a name for debugging purposes::
```
TheYellowArchitect, line 974: Nitpick but its shorter, for the same meaning, feel free to deny this specific comment
```
Lambdas can be named for debugging purposes
```

HolonProduction:
> tl;dr I haven't tried building. I did try that link you uploaded, opened it with my browser (brave) and it didn't render like the docs.

I did not include the style sheets. Feel free to build it yourself. It renders correct on my side.

TheYellowArchitect:
With the change of "Lambdas" to "Lambda functions", perhaps the last line also needs to change to "lambda functions", but anyway not important, especially since lambdas are clear now (e.g. no "flat" or "frozen" values) and I assume a new user who reads "lambdas" can do 1+1=2 and realize "lambdas = lambda functions"
tl;dr: Nitpick on naming consistency

Aside of that nitpick, cannot find any flaws at all :+1:

TheYellowArchitect (approved):

mhilbrunner (approved):

mhilbrunner:
Going without the above nitpick, good addition! Thanks for contributing (and congrats on having your first docs contribution merged!), and thanks TheYellowArchitect for not only contributing so much lately but also all the reviews, thats extremely helpful! 🎉 

Thanks, both of you.

PR #6284, opened 2022-10-09
Update 'velocity' with move_and_slide

videlanicolas:
`move_and_slide` returns the modified velocity, which should be updated in the code. If this is not done, `velocity` will keep increasing due to `velocity.y += delta * gravity`. If one adds a `jump` function to this code it'll not be able to jump, since the vector continues to increase in magnitude and never stops.

By updating the velocity with `move_and_slide` we make sure velocity is 0 when there's no movement, thus reflecting the actual velocity of the object.

### DIFF
--- a/tutorials/physics/kinematic_character_2d.rst
+++ b/tutorials/physics/kinematic_character_2d.rst
@@ -209,7 +209,7 @@ This adds basic support for walking when pressing left and right:
 
         # The second parameter of "move_and_slide" is the normal pointing up.
         # In the case of a 2D platformer, in Godot, upward is negative y, which translates to -1 as a normal.
-        move_and_slide(velocity, Vector2(0, -1))
+        velocity = move_and_slide(velocity, Vector2(0, -1))
 
  .. code-tab:: csharp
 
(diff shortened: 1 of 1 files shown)


TheYellowArchitect:
I haven't done that tutorial yet (am newbie at Godot), but that code does not compile from my side. `move_and_slide` returns boolean, not the velocity itself. This is even confirmed at docs. Mayhaps your changes are for 3.x?

Again, feel free to correct me, I am a newbie and what I wrote above may be wrong

Calinou:
Indeed – this code should be changed to set `velocity` before calling `move_and_slide()` with no arguments.

cbscribe:
I thought the goal was to reduce the number of individual tutorials going forward. Since this is entirely changed for 4.0 CharacterBody, this should probably be removed from master.

videlanicolas:
New PR at #6289.

TheYellowArchitect:
>I thought the goal was to reduce the number of individual tutorials going forward. Since this is entirely changed for 4.0 CharacterBody, this should probably be removed from master.

The "Dodge the Creeps" Tutorial (First 2D Game Tutorial), and "Squash the Creeps" tutorial (First 3D Game Tutorial) indeed use `move_and_slide()` so new users are familiarized with how velocity works. But it's good to have extra content to fully grasp how it works, since velocity direction isn't the simplest thing for a new user, at least from my experience finishing these tutorials. The best would be a remake of this velocity tutorial, so a user doesn't have to do a full tutorial or skim a tutorial just to reach the `move_and_slide()` codeblock. If this tutorial was deleted from `:master:` branch, no one would find it and hence, no one would update it :(

PR #6285, opened 2022-10-09
fix: rename `ViewportContainer` to `SubViewportContainer`

yedpodtrzitko:
This PR renames `ViewportContainer` to `SubViewportContainer`  according to 3to4 converter:

https://github.com/godotengine/godot/blob/bbac8198f89c0cbf0da4293201e54aa847947370/editor/project_converter_3_to_4.cpp#L1533

However if this was just a class rename, then there's another discrepancy. [Viewports tutorial says](https://github.com/godotengine/godot-docs/blob/master/tutorials/rendering/viewports.rst#scale--stretching):

> For **Viewports** which are children of **ViewportContainers** (...)

but the class documentation of **Viewport** says it's inherited only from `Node` and `Object`, see:

https://github.com/godotengine/godot-docs/blob/master/classes/class_viewport.rst#viewport

### DIFF
--- a/tutorials/rendering/multiple_resolutions.rst
+++ b/tutorials/rendering/multiple_resolutions.rst
@@ -427,8 +427,8 @@ performance significantly while keeping the HUD and other 2D elements crisp.
 
 This is done by using the root Viewport node only for 2D elements, then creating
 a Viewport node to display the 3D world and displaying it using a
-ViewportContainer or TextureRect node. There will effectively be two viewports
-in the final project. One upside of using TextureRect over ViewportContainer is
+SubViewportContainer or TextureRect node. There will effectively be two viewports
+in the final project. One upside of using TextureRect over SubViewportContainer is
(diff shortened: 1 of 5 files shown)


clayjohn (approved):

clayjohn:
> However if this was just a class rename, then there's another discrepancy. [Viewports tutorial says](https://github.com/godotengine/godot-docs/blob/master/tutorials/rendering/viewports.rst#scale--stretching):
> 
> > For **Viewports** which are children of **ViewportContainers** (...)
> 
> but the class documentation of **Viewport** says it's inherited only from `Node` and `Object`, see:
> 
> https://github.com/godotengine/godot-docs/blob/master/classes/class_viewport.rst#viewport

In this context "child" does not mean "inherits from" it means that it is a child of another node in Godot's node tree

mhilbrunner (approved):

mhilbrunner:
Thanks for the fixes, merged! :)

PR #6289, opened 2022-10-10
Update 'velocity' with move_and_slide

videlanicolas:
`move_and_slide` returns the modified velocity, which should be updated in the code. If this is not done, `velocity` will keep increasing due to `velocity.y += delta * gravity`. If one adds a `jump` function to this code it'll not be able to jump, since the vector continues to increase in magnitude and never stops.

By updating the velocity with `move_and_slide` we make sure velocity is 0 when there's no movement, thus reflecting the actual velocity of the object.


mhilbrunner (approved):

mhilbrunner:
Good catch, thanks and congrats on your first contribution to Godot's docs!

PR #6299, opened 2022-10-11
Fix Navigation index order to make thematically more sense

smix8:
The original index from all the Navigation pullrequests was ordered from beginner / core topics to more sophisticated / focused / niche topics but this got mixed up due to splitting the pr's and the order the pr's were merged.


mhilbrunner:
Thank you! 🎉

PR #6302, opened 2022-10-12
Update C# in step-by-step Signals example

zacharied:
* Changed code to use new event-style connections
* Changed custom signal example to use the new `SignalName` global
* Use `double` instead of `float` for `_Process` argument

My only reservation here is using `SignalName` without explaining where it comes from, but I wanted to avoid bloating the page with language-specific information. Perhaps an inline comment in the code would be useful?

### DIFF
--- a/getting_started/step_by_step/signals.rst
+++ b/getting_started/step_by_step/signals.rst
@@ -177,11 +177,11 @@ following code, which we saw two lessons ago:
 
  .. code-tab:: csharp C#
 
-    public override void _Process(float delta)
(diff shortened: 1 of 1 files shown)


raulsntos (approved):

mhilbrunner (approved):

mhilbrunner:
Thanks and congrats on your first contribution to these docs! 🎉 
I think its fine as is, but a comment could also be fine; feel free to submit a followup PR if you think it's needed. :)

PR #6306, opened 2022-10-13
Update Csharp code to compute shaders

nonunknown:
the `shaderBytecode` var doesnt return `byte[]` required by `ShaderCreateFromSpirv()`, to solve just call the property `BytecodeCompute`

### DIFF
--- a/tutorials/shaders/compute_shaders.rst
+++ b/tutorials/shaders/compute_shaders.rst
@@ -91,7 +91,7 @@ After that we can load the newly created shader file "compute_example.glsl" and 
     // Load GLSL shader
     var shaderFile = GD.Load<RDShaderFile>("res://compute_example.glsl");
     var shaderBytecode = shaderFile.GetSpirv();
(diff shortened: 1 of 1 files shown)


clayjohn:
``ShaderCreateFromSpirv`` shouldn't expect an array of bytes it accepts a reference to a [RDShaderSPIRV](https://docs.godotengine.org/en/latest/classes/class_rdshaderspirv.html) resource see:
https://github.com/godotengine/godot/blob/926429392a73a0c2261bc4ed4503c99025842d7c/servers/rendering/rendering_device.cpp#L219

I have a feeling this is a problem with the bindings. 

When you run the updated code does it actually work for you?

nonunknown:
well I didnt tested the code yet, but vscode blamed me and solved after I did the change stated in PR, I'm following the tutorial here, when I finish it, I will return to you!

nonunknown:
closing as this is a binding error, not a documentation one!

PR #6309, opened 2022-10-14
Fix usage of NOTIFICATION_WM_CLOSE_REQUEST

Sauermann:
Changes https://docs.godotengine.org/en/latest/tutorials/inputs/handling_quit_requests.html

Propagation now works differently.
Also this notification no longer causes the application to quit.

It would be great, if someone with C# experience would look over the C#-change, since I am unable to test that.

related to godotengine/godot#67379

### DIFF
--- a/tutorials/inputs/handling_quit_requests.rst
+++ b/tutorials/inputs/handling_quit_requests.rst
@@ -14,17 +14,18 @@ to go back otherwise).
 Handling the notification
 -------------------------
 
-On desktop and web platforms, :ref:`Node <class_Node>`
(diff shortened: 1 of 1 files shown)


skyace65:
The change for the OS class reference page needs to be made in the main Godot repository, not the docs. The class reference pages in the docs repository are generated from what's in the engine repository. [This](https://github.com/godotengine/godot/blob/master/doc/classes/OS.xml) is the file you need to change.

mhilbrunner (approved):

mhilbrunner:
Thanks! 🎉

PR #6312, opened 2022-10-15
Draft: Update using tilemaps page

shinspiegel:
Hi, 

I did start to work on the "using tilemaps",    
There are some areas in that I could not find the right words to express. 

Also, I could not understand the underlying difference between the "Match Corners" and "Match Corners and Sides", 
If someone could explain this to me I would be very grateful.

Also, there is something odd on the tilemap, it affects adding new tiles on the adjacent ones, affecting more than it should. This is the intended behavior or this is a bug that should be reported?


mhilbrunner:
Hey! Thanks for taking the time to contribute.

It seems [get-pip.py](https://github.com/godotengine/godot-docs/pull/6312/files#diff-333f925f52b4ce75f65b3e49a8a71acd851ab66561480309d9e34de7a888a452) was included by mistake.

As far as the current draft goes, it seems too unfinished to really review and at the same time not something that needs much discussion from our side, so having it here already doesn't make too much sense IMO. I suggest working on it in your branch and opening a PR once its near to a state where it's reviewable.

If you need help with TileMap functionality, there's tons of amazing Godot [communities](https://godotengine.org/community) around to answer any questions, in my experience both the Discord and the RocketChat are extremely helpful.

Unfortunately, we don't have the capacity on the docs team currently to help with answering those questions.

PR #6318, opened 2022-10-15
Remove old renderer backend comparison and GDNative tutorial pages

mhilbrunner:
This PR:
- Updates the list & description of Github topic labels as it touched those anyway
- Removes the old GDNative docs section and links to it and replaces many occurences of "GDNative" with "GDExtension" where appropriate (some were deliberately left untouched, in places where they refer to technical specifics that may need to be updated, i.e. how to bundle native libs for Android, so those parts remain 'grep'able)
- Removes the page comparing the old renderers (GLES 2 vs 3)

The motivation is to fix build errors and remove outdated information.
As those pages remain in all stable branches (3.x), everything can still be easily accessed (and salvaged if needed).


paddy-exe:
@mhilbrunner I am already working on the GDExtension docs and the rewrite. I have renamed almost every instance of GDNative already👍🏻

clayjohn:
Getting rid of the GLES2/GLES3 document sounds good to me. We will have to rewrite a renderer comparison doc from scratch anyway shortly before 4.0 releases

PR #6314, opened 2022-10-15
Add a page on large world coordinates (double-precision floats)

Calinou:
**Note:** This currently doesn't include the SCons option rename from https://github.com/godotengine/godot/pull/67399.

PR #6324, opened 2022-10-17
Fix text mentioning model vertex snapping in Large world coordinates

clayjohn:
Follow up to https://github.com/godotengine/godot-docs/pull/6314

The doc originally said that model's vertices would snap to the nearest number represented by floating point, however, it is the objects origin position that will snap to the nearest floating point, the individual vertices will be fine. 

While reviewing this addition, I noticed a few oddities arising out of it being in the physics section. I'm not sure this doc is appropriately placed in the physics section as it is equally important for non-physics games as physics game (for example, any 3D game with a large world, 2D idle games with large monetary values, etc.). I think it is worth discussing where this doc may be a better fit. Then the doc can be re-framed to not focus on physics.

### DIFF
--- a/tutorials/physics/large_world_coordinates.rst
+++ b/tutorials/physics/large_world_coordinates.rst
@@ -28,7 +28,7 @@ In practice, this means that as the player moves away from the world origin
 will decrease.
 
 This loss of precision can result in objects appearing to "vibrate" when far
-away from the world origin, as the model's vertices will be snapping to the
+away from the world origin, as the model's position will snap to the
 nearest value that can be represented in a floating-point number. This can also
 result in physics glitches that only occur when the player is far from the world
 origin.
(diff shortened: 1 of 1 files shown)


fire (approved):
Thanks for the technical update.

mhilbrunner:
Yeah, I agree the Physics category doesn't really fit, the other existing categories are IMO worse though.

It may make sense to group a few things in an "Advanced" section because there's some really specific stuff currently strewn about thats only going to be used by quite large or specific projects, but not sure thats worth the effort.

mhilbrunner (approved):

mhilbrunner:
cc @Calinou

Calinou:
> 2D idle games with large monetary values, etc.)

The `float` type in GDScript is always 64-bit, even in single-precision builds.

Calinou (approved):
Thanks!

PR #6327, opened 2022-10-20
Clarify that nodes are structured as trees

stewchan:
As a Godot beginner, the use of the word 'key' is confusing because it sounds like it is an class attribute or property.  Linking it to the concept of nodes as trees makes it more clear.

### DIFF
--- a/getting_started/step_by_step/nodes_and_scenes.rst
+++ b/getting_started/step_by_step/nodes_and_scenes.rst
@@ -26,7 +26,7 @@ All nodes have the following attributes:
 - You can extend them with new properties and functions.
 - You can add them to another node as a child.
 
-The last attribute is key. **Together, nodes form a tree**, which is a powerful
+The last attribute is key - **Together, nodes form a tree**. This is a powerful
 feature to organize projects. Since different nodes have different functions,
 combining them produces more complex behavior. As we saw before, you can build a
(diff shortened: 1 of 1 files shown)


getting_started/step_by_step/nodes_and_scenes.rst
```diff
@@ -26,7 +26,7 @@ All nodes have the following attributes:
 - You can extend them with new properties and functions.
 - You can add them to another node as a child.
 
-The last attribute is key. **Together, nodes form a tree**, which is a powerful
+The last attribute is key - **Together, nodes form a tree**. This is a powerful
```
mhilbrunner, line 29:
```suggestion
The last attribute is key - **together, nodes form a tree**. This is a powerful
```

YuriSizov:
Superseded by https://github.com/godotengine/godot-docs/pull/6827. Thanks for your contribution nonetheless!

PS. Seems like you have seen the review comment left by a maintainer, but you didn't update your PR. The idea is that after we leave review comments, you need to apply them (if you agree), or you need to explain why the change is not needed (if you don't). You can then push additional commits to the PR, or rebase and force-push it all as a single commit, without breaking anything.

The PR is considered ready for merging once we approve it with no withstanding requests for you.

PR #6330, opened 2022-10-21
Position the player in the center of the screen

armorup:
The player is hidden in the top left corner which is confusing to a learner


Calinou:
The player should be positioned by moving it in the 2D editor view, not using a script.

Vennnot:
If the player should be positioned in the editor and not through code then this PR can be closed, no?

PR #6340, opened 2022-10-25
Introduction - Wrong number

denicolas:
[About » Introduction](https://docs.godotengine.org/en/stable/about/introduction.html)

"This documentation is organized in five sections"
should be:
"This documentation is organized in six sections"

### DIFF
--- a/about/introduction.rst
+++ b/about/introduction.rst
@@ -75,7 +75,7 @@ attribution to "Juan Linietsky, Ariel Manzur and the Godot Engine community".
 Organization of the documentation
 ---------------------------------
 
-This documentation is organized in five sections with an impressively
+This documentation is organized in six sections with an impressively
 unbalanced distribution of contents – but the way it is split up should be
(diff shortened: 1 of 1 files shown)


mhilbrunner:
I'd maybe consider removing the number althogether, not sure it's that helpful and its easy to miss to update it if anything changes, what do you think?

denicolas:
> I'd maybe consider removing the number althogether, not sure it's that helpful and its easy to miss to update it if anything changes, what do you think?

Yes, I agree. So maybe the number could be just replaced with "several":
"This documentation is organized in several sections"

mhilbrunner:
That sounds good to me :)

YuriSizov:
Bump :)

YuriSizov:
Nevermind, superseded by https://github.com/godotengine/godot-docs/pull/6572 because I was editing that file anyway. Thanks for your contribution nonetheless :)

PR #6343, opened 2022-10-27
hint_albedo is depreciated

alcomposer:
`hint_albedo` is no longer used, it is `source_color` to convert the source colorspace to linear. Add example of `source_color` for sampler2D

### DIFF
--- a/tutorials/shaders/shader_reference/shading_language.rst
+++ b/tutorials/shaders/shader_reference/shading_language.rst
@@ -748,9 +748,10 @@ used, and how the editor should allow users to modify it.
     uniform vec4 color : source_color;
     uniform float amount : hint_range(0, 1);
     uniform vec4 other_color : source_color = vec4(1.0);
+    uniform sampler2D image : source_color;
 
 It's important to understand that textures that are supplied as color require
-hints for proper sRGB->linear conversion (i.e. ``hint_albedo``), as Godot's 3D
+hints for proper sRGB->linear conversion (i.e. ``source_color``), as Godot's 3D
 engine renders in linear color space.
 
 Full list of hints below:


alcomposer:
I also think that:
```
+----------------------+------------------------------------------------+-----------------------------------------------------------------------------+
| **sampler2D**        | source_color                                   | Used as albedo color.                                                       |
+----------------------+------------------------------------------------+-----------------------------------------------------------------------------+
```
should be changed to (or similar):
```

+----------------------+------------------------------------------------+-----------------------------------------------------------------------------+
| **sampler2D**        | source_color                                   | Use source color space                                                     |
+----------------------+------------------------------------------------+-----------------------------------------------------------------------------+
```

clayjohn (approved):

clayjohn:
> I also think that:
> 
> ```
> +----------------------+------------------------------------------------+-----------------------------------------------------------------------------+
> | **sampler2D**        | source_color                                   | Used as albedo color.                                                       |
> +----------------------+------------------------------------------------+-----------------------------------------------------------------------------+
> ```
> 
> should be changed to (or similar):
> 
> ```
> 
> +----------------------+------------------------------------------------+-----------------------------------------------------------------------------+
> | **sampler2D**        | source_color                                   | Use source color space                                                      |
> +----------------------+------------------------------------------------+-----------------------------------------------------------------------------+
> ```

I'm not sure about this additional change. There is no such thing as source color space. source_color tells the engine that the source of this uniform is a color value and instructs the engine to deal with it appropriately (i.e. convert to linear/sRGB as needed)

alcomposer:
Hm, so with `source_color` the color value is used "as color" otherwise "as data"?

So maybe "Use as color value"?

alcomposer:
AFAIK Godot only deals in sRGB.  However it should still be mentioned that source_color applies the sRGB EOTF, and without will look washed out, or alternatively a game developer needs to implement their own EOTF in fragment shader. 

eg. approximate sRGB to linear: `pow(color, vec3(2.2))`

The choice of source_color as the new keyword is possibly misleading itself, as currently this has nothing to do with color, rather transform function (previously known as Gamma).

I can see that it would be useful if (in future) Godot did it's internal color handling (for example) in linear rec.2020. The image importer would allow the user to select the color space used by the image (or Godot would deduce this from the image metadata).

Then source_color would instruct Godot to map the source (sRGB, rec.2020 etc) to the working linear colorspace.

This sort of solution would need quite a bit of plumbing, but I think it would make sense with the current terminology.

Obviously the benefits of using linear rec.2020 as a working space are huge, allowing saturated highlights to render accurately, as opposed to displaying clipped.

I understand that this is off-topic of the documentation, but we also need to be aware of the implications of how Godot currently deals with color, given that other engines (and the industry as a whole) have begun exploring larger color spaces.

At the end of the day, Vulkan is color space agnostic, so it is Godot that would deal (and does currently to a degree) the transforms, as discussed here: https://docs.godotengine.org/en/3.0/tutorials/3d/high_dynamic_range.html

YuriSizov:
Let's merge the current changes, since they've been approved. If there are further improvements needed, feel free to open a new PR :)

YuriSizov:
Thanks, and congrats on the first merged pull request!

PR #5277, opened 2021-09-22
Fix NDC coordinates for Vulkan in Advanced post-processing

Calinou:
This closes https://github.com/godotengine/godot-docs/issues/5276.

**Note:** Don't cherry-pick to 3.x branches.

PR #6344, opened 2022-10-27
Fix NDC coordinates for Vulkan in Screen-reading shaders

dzil123:
Equivalent of #5276 and #5277 for Screen-reading shaders.


mhilbrunner:
Thanks for contributing!
cc @clayjohn

clayjohn (approved):
Looks good for now. dzil123 and I have been discussing on rocketchat how to make this friendlier for users as right now the behaviour differs between the RD and OpenGL-based renderers

PR #6346, opened 2022-10-28
Remove `print`s which make the program hang in Recording with microphone

Binney:
When I run the code as originally printed, it hangs. I'm pretty sure this is just because it's trying to print KBs of raw WAV binary to the console. I'm in GDscript but I imagine it happens in C# too.

The line `print(recording)` is fine because it just prints `[AudioStreamSample:1234]`.

The [sample project itself](https://github.com/godotengine/godot-demo-projects/blob/master/audio/mic_record/MicRecord.gd) also omits this line so this is really just bringing them in sync


Calinou (approved):
Thanks!

PR #6347, opened 2022-10-29
Update RemoteTransform3D to better explain how it works

nonunknown:
Explains how the node must be setup on tree, aside from using godot for years, this is the first time I'm using it, so I got into lot of headache to understand that it must be child of the target you want to change transform! LOL


skyace65:
Closing this PR as changes made to the class reference need to be made in the engine repository, not the documentation one. The docs repository generates its pages for classes based on what's in the main repository. The file you want to edit is [here](https://github.com/godotengine/godot/blob/master/doc/classes/RemoteTransform3D.xml).

PR #6349, opened 2022-10-29
Fixed outdated scons arguments for export templates.

Gamepro5:
Also added a note to remind the reader to add the `float=64` flag if they want to make export templates for a Godot build that has 64 bit floats.


YuriSizov:
This is now outdated, as the argument and its values has been renamed in https://github.com/godotengine/godot/pull/67399. It should also be added to the introduction article, not to Windows-specific instructions, as it's a universal option.

Feel free to open a new PR with this information!

PR #6354, opened 2022-11-01
Update Big O notation resource link

Skyvastern:
Existing link to Big O notation resource, leads to a non working website. Therefore, a new link needs to be added.

I found [this video](https://www.youtube.com/watch?v=D6xkbGLQesk) to be a helpful guide in understanding Big O Notation. As such, I added this as a new resource for understanding Big O Notation in the docs.

### DIFF
--- a/tutorials/best_practices/data_preferences.rst
+++ b/tutorials/best_practices/data_preferences.rst
@@ -10,7 +10,7 @@ Y or Z? This article covers a variety of topics related to these dilemmas.
 
   This article makes references to "[something]-time" operations. This
   terminology comes from algorithm analysis'
-  `Big O Notation <https://rob-bell.net/2009/06/a-beginners-guide-to-big-o-notation/>`_.
+  `Big O Notation <https://www.youtube.com/watch?v=D6xkbGLQesk>`_.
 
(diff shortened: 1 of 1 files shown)


Calinou:
I'd prefer a text-based article as it's easier to skim over it (and is more friendly to people with slow connections or strict data caps).

Skyvastern:
> I'd prefer a text-based article as it's easier to skim over it (and is more friendly to people with slow connections or strict data caps).

Yeah I can see the benefit of that. Lemme check some useful articles out there on Internet, and update the link with one.

Skyvastern:
@Calinou I've updated the link with the following [article](https://towardsdatascience.com/introduction-to-big-o-notation-820d2e25d3fd). I also came across [this](https://www.freecodecamp.org/news/big-o-notation-why-it-matters-and-why-it-doesnt-1674cfa8a23c/) from freecodecamp, but found the previous one much simpler to understand. So went with that instead. These were the 2 I found most easy to understand, with the needed amount of content.

mhilbrunner:
Seems the original webpage has been restored in the mean time, so I guess we can leave this as is for now. If it breaks again, we still may want to switch to this one.

Thanks for taking the time to submit a PR :)

PR #6367, opened 2022-11-07
Move the debugging with gdb docs into the debugging section

CalebJohn:
I had trouble finding the GDB documentation when I went looking. I initially checked the "Debugging and Profiling" section, and when I didn't see them there, I assumed they didn't exist. I think it makes sense to consolidate all debugger related docs there. Please let me know if I did anything wrong here, this is my first contribution to docs :)

Currently the "debugging with gdb" docs are under "Introduction to to Godot development" (not a bad spot)
image

But it would be more clear to house these tips under "Debugging and Profiling" 

image


clayjohn (approved):
This change makes sense. Although it leaves the "Introduction to Godot development" doc pretty sparse. Let's merge for now, but we should consider reorganizing these docs in the future.

Calinou (changes requested):
The URL should match the file name (which itself should match the page name, but lowercased and with underscores instead of spaces). This means the file should be renamed to `debugging_with_gdb.rst`.

After applying this change, this PR can be merged.

clayjohn (approved):

CalebJohn:
> The URL should match the file name (which itself should match the page name, but lowercased and with underscores instead of spaces). This means the file should be renamed to debugging_with_gdb.rst.

Done. But can I suggest adding this requirement to the [documentation](https://docs.godotengine.org/en/latest/community/contributing/contributing_to_the_documentation.html#adding-new-pages)? Without guidance, I based this filename on a [similar doc](https://github.com/godotengine/godot-docs/blob/master/development/cpp/macos_debug.rst) (I understand that not all existing docs will follow the rules, but currently this is all we have to go on).

PR #6316, opened 2022-10-15
Document the Unification of tools/target build type configuration

NeroBurner:
Introduced in https://github.com/godotengine/godot/pull/66242 the `tools=yes/no` option was removed and merged into the `target` preset.

Replace the `tools=yes/no` calls in the documentation to the new template target presets `editor`, `template_release` and `template_debug`.

---

It's my first PR here, so please tell me if something is missing or done wrong. I just was confused why the `scons target=debug` builds didn't work. After finding out about the change I thought I'd take this opportunity to fix the documentation for everyone and learn quite a bit in the process

PR #6375, opened 2022-11-09
Change build flags in Godot 4 docs to current variants

ZoopOTheGoop:
I was trying to follow the build instructions for production and found that at some point on `master` the flags in `SConstruct` had changed - in particular the `target` values now only include `editor`, `template_release`, and `template_debug`. I made my best guess at the proper ones to use, this probably needs a revision pass by someone who builds Godot from source more often than me.

In particular, I had to remove the documentation for the `headless` section, as it's not entirely clear from `SConstruct` what you'd pass to do it. This probably also needs to be changed on the other platforms, but I haven't had the opportunity to check that yet.

Changelog:
- `target=release_debug` -> `optimize=speed_trace`
- `target=release` -> `optimize=speed` (may also need `debug_symbols=no`?)
- use `production=yes` as default option for production builds
- remove headless compilation as current way to do this is non-obvious with current build targets available, should be added before merging

### DIFF
--- a/development/compiling/compiling_for_linuxbsd.rst
+++ b/development/compiling/compiling_for_linuxbsd.rst
@@ -149,13 +149,16 @@ manager.
 
 .. note:: If you are compiling Godot for production use, then you can
           make the final executable smaller and faster by adding the
-          SCons option ``production=yes``.
+          SCons option `production=yes`.
           
-          If you wish, you can also enable features piecemeal such as 
-          `optimize=speed_trace`, or `use_lto=yes` which will enable
+          If you wish, you can also enable features piecemeal with 
+          `optimize=speed` and/or `lto=full` which will enable
           speed optimizations and link-time optimizations independently.
+          
           In particular, you may not be able to build with LTO on with a weaker computer, 
-          as it can use quite a lot of RAM (up to 7GB).
+          as it can use quite a lot of RAM (up to 7GB). In this case, you may still be able to get
+          some benefits with weaker (but more memory-friendly) settings such as `lto=thin`, or
+          `lto=auto` which should choose a reasonable configuration for your machine.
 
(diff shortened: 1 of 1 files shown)


development/compiling/compiling_for_linuxbsd.rst
```diff
@@ -149,36 +149,22 @@ manager.
...
+          as it can use quite a lot of RAM (up to 7GB). In this case, you may still be able to get
+          some benefits with weaker (but more memory-friendly) settings such as `lto=thin`, or
+          `lto=auto` which should choose a reasonable configuration for your machine.
 
 .. note:: If you want to use separate editor settings for your own Godot builds
           and official releases, you can enable
           :ref:`doc_data_paths_self_contained_mode` by creating a file called
           ``._sc_`` or ``_sc_`` in the ``bin/`` folder.
 
-Compiling a headless/server build
```
Maran23, line 164: Was this functionality removed?
Can't find anything in the original PR: https://github.com/godotengine/godot/pull/66242

ZoopOTheGoop: As I mentioned in the original post, I couldn't find how to do this, so I removed it so as not to give bad instructions. I'll probably revert this because another PR seems to fix the headless build instructions.

YuriSizov:
https://github.com/godotengine/godot-docs/pull/6316 has been merged, so this needs a rebase, if there is still some information worth adding, or we should close this if there isn't anything.

Vennnot:
I think this should be closed. If the creator thinks there is still information to be added we can re-open.

mhilbrunner:
Closing as per the above discussion. Please ping for a reopen if there is still important information to add to the current page, in which case this PR needs to be updated/rebased - or open a new one. Thanks for contributing!

PR #6378, opened 2022-11-12
Improve the Bisecting regressions page

Calinou:
- Update the page to match the new versioning policy.
- Add a Bash function to retrieve alpha/beta/RC commit hashes from the command line (see https://github.com/godotengine/godot/issues/68563 – thanks @akien-mga :slightly_smiling_face:).
- Mention caveats related to project compatibility and output binary names.

### DIFF
--- a/community/contributing/bisecting_regressions.rst
+++ b/community/contributing/bisecting_regressions.rst
@@ -32,20 +32,27 @@ range of commits that potentially need to be built from source and tested.
 You can find binaries of official releases, as well as alphas, betas,
 and release candidates `here <https://downloads.tuxfamily.org/godotengine/>`__.
 
-For example, if you've reported a bug against Godot 3.2, you should first try to
-reproduce the bug in Godot 3.1 (not a patch release, see below for the reason).
-If the bug doesn't occur there, try to reproduce it in Godot 3.2 *beta 1* (which
-is roughly in the middle of all test builds available). If you can't reproduce
-the bug with Godot 3.2 beta 1, then try newer betas and RC builds. If you do
-manage to reproduce the bug with Godot 3.2 beta 1, then try older alpha builds.
+If you have experience with Godot 3.x and can reproduce an issue with Godot 4.0,
+we recommend trying to reproduce the issue in the latest Godot 3.x version (if
+the feature exhibiting the bug is present in 3.x). This can be used to check
+whether the issue is a regression in 4.0 or not.
+
+- If the issue **is present** in 3.x, then you'll need to check whether the issue
+  occurs in older 3.x versions as well.
+- If the issue is **not present** in 3.x, then you can try older 4.0 alphas and
+  betas to determine when the regression started.
 
 .. warning::
 
-    For bisecting regressions, don't use patch releases such as Godot 3.1.2.
-    Instead, use the minor version's first release like Godot 3.1. This is
-    because patch releases are built from a separate *stable branch*. This kind
-    of branch doesn't follow the rest of Godot's development, which is done in
-    the ``master`` branch.
+    Project files may be incompatible between Godot versions.
+    **Make a backup of your project** before starting the bisection process.
+
+    Going from the oldest to the newest build generally reduces the risk of the
+    project not being able to successfully open in the editor, thanks to
+    backwards compatibility. Try to reduce your project to the smallest
+    repeatable example too. The more minimal the project is, the more likely
+    you'll be able to open it without compatibility issues in newer engine
+    versions.
 
 The Git bisect command
 ----------------------
@@ -69,6 +76,9 @@ reproduce the bug.
     there and report the results of your "pre-bisecting" on the GitHub issue so
     another contributor can continue bisecting from there.
 
+Determine the commit hashes
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
 To start bisecting, you must first determine the commit hashes (identifiers) of
 the "bad" and "good" build. "bad" refers to the build that exhibits the bug,
 whereas "good" refers to the version that doesn't exhibit the bug. If you're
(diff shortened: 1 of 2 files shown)


community/contributing/bisecting_regressions.rst
```diff
@@ -105,16 +147,23 @@ Compile Godot. This assumes you've set up a build environment:
...
 .. code-block:: shell
 
-    # <platform> is the platform you're targeting for regression testing,
-    # like "windows", "x11" or "macos".
-    $ scons platform=<platform> -j4
+    $ scons
 
 Since building Godot takes a while, you want to dedicate as many CPU threads as
 possible to the task. This is what the ``-j`` parameter does. Here, the command
 assigns 4 CPU threads to compiling Godot.
```
Sauermann, line 154: Removing the `-j4` parameter in the code block, makes this describing text inconsistent, which refers to the `-j` parameter.

Calinou: Fixed. Godot's SCons setup now detects how many CPU threads should be used depending on the system, but it intentionally leaves 1 thread available on systems with at least 4 threads to make multitasking smoother. There's a fair chance people will be doing something else while waiting for Godot to compile (speaking from experience), so I think this is a good default for bisecting regressions.

community/contributing/bisecting_regressions.rst
```diff
@@ -32,20 +32,27 @@ range of commits that potentially need to be built from source and tested.
...
-For example, if you've reported a bug against Godot 3.2, you should first try to
-reproduce the bug in Godot 3.1 (not a patch release, see below for the reason).
-If the bug doesn't occur there, try to reproduce it in Godot 3.2 *beta 1* (which
-is roughly in the middle of all test builds available). If you can't reproduce
-the bug with Godot 3.2 beta 1, then try newer betas and RC builds. If you do
-manage to reproduce the bug with Godot 3.2 beta 1, then try older alpha builds.
+If you can reproduce an issue on Godot 4.0, we recommend trying to reproduce the
+issue in the latest Godot 3.x version (if the feature exhibiting the bug is
+present in 3.x). This can be used to check whether the issue is a regression in
+4.0 or not.
```
YuriSizov, line 38:
```suggestion
If you can reproduce an issue with Godot 4.0, we recommend trying to reproduce the
issue in the latest Godot 3.x version (if the feature exhibiting the bug is
present in 3.x). This can be used to check whether the issue is a regression in
4.0 or not.
```

I'm also wondering if this is going to remain a good instruction once we release 4.0. It doesn't really cover 4.x releases, and it shouldn't be required, nor recommended IMO, to try 3.x to see if it's a regression. We'll have a lot of new users, hopefully, and making them downgrade just to check doesn't sound reasonable.

community/contributing/bisecting_regressions.rst
```diff
@@ -32,20 +32,27 @@ range of commits that potentially need to be built from source and tested.
...
+
+If the issue **is present** in 3.x, then you'll need to check whether the issue
+occurs in older 3.x versions as well.
+
+If the issue is **not present** in 3.x, then you can try older 4.0 alphas and
+betas to determine when the regression started.
```
YuriSizov, line 44: Maybe this should be a list? It's a bit awkward looking as individual paragraphs made out of single sentences.

For comparison:
image

community/contributing/bisecting_regressions.rst
```diff
@@ -32,20 +32,27 @@ range of commits that potentially need to be built from source and tested.
...
-    of branch doesn't follow the rest of Godot's development, which is done in
-    the ``master`` branch.
+    Project files may be incompatible between Godot versions.
+    **Make a backup of your project** before starting the bisection process.
+
+    Going from the oldest to the newest build generally reduces the risk of the
+    project not being able to successfully open in the editor in subsequent runs
+    with different engine versions. Also, the more minimal the project, the more
+    likely it'll be able to open without compatibility issues in newer engine
+    versions.
```
YuriSizov, line 55:
```suggestion
    Going from the oldest to the newest build generally reduces the risk of the
    project not being able to successfully open in the editor, thanks to backwards
    compatibility. Try to reduce your project to the smallest repeatable example too.
    The more minimal the project is, the more likely you'll be able to open it without
    compatibility issues in newer engine versions.
```

I think the first line is a bit wordy and rich with technical jargon. Also tried to make the transition to the second half smoother.

community/contributing/bisecting_regressions.rst
```diff
@@ -82,12 +92,44 @@ following commit hashes depending on the version:
...
     3.0-stable
 
+You can also use this Bash function to retrieve the Git commit hash of a
+pre-release build (add it to your ``$HOME/.bashrc`` or similar):
+
+::
+
+    gd_snapshot_commit() {
+        curl -s https://downloads.tuxfamily.org/godotengine/$1/$2/README.txt | grep 'from commit' | sed 's/^Built from commit \(.*\)\.$/\1/'
+    }
```
YuriSizov, line 119: Can this be split into multiple lines for readability?

community/contributing/bisecting_regressions.rst
```diff
@@ -105,16 +147,19 @@ Compile Godot. This assumes you've set up a build environment:
...
+Run the engine
+^^^^^^^^^^^^^^
 
 Run the binary located in the ``bin/`` folder and try to reproduce the bug.
 
+.. note::
+
+    Double-check the output file name in ``bin/`` to make sure you're actually
+    running the binary you've just compiled. Different Godot versions will
+    output binaries with different names.
```
YuriSizov, line 161: Can also link to `/development/compiling/introduction_to_the_buildsystem.html#resulting-binary`. It currently doesn't have a reference tag, so that needs to be added as well.

community/contributing/bisecting_regressions.rst
Calinou: Maybe we can reword it like this?

```
If you have experience with Godot 3.x and can reproduce an issue with Godot 4.0, we recommend trying to reproduce the issue in the latest Godot 3.x version (if the feature exhibiting the bug is present in 3.x). This can be used to check whether the issue is a regression in
4.0 or not.
```

We can remove this paragraph once 4.1 is released.

YuriSizov: Sure, that would be an improvement. I guess we can reintroduce a version of what this paragraph used to be when we have at least the 4.1 release, so we can refer to it in the example.

YuriSizov (approved):

YuriSizov:
Thanks!

PR #6379, opened 2022-11-13
Update gdscript_styleguide.rst

FireSquid6:
Adds two changes:
- makes it clear where subclasses should go (at the bottom of the script)
- put the `_enter_tree()` method before `_ready()` but after `_init()` since that's the order in which the methods are executed


mhilbrunner:
cc @Calinou I think this looks good, but if you can give it a glance, esp. the added order :)

Calinou:
> makes it clear where subclasses should go (at the bottom of the script)

Is there consensus on this? I generally put subclasses at the top of a script myself.

Mickeon:
I like to keep them at the bottom to reduce clutter and general "overwhelming"-ness upon first opening a script.

FireSquid6:
Most of the people I've worked with put subclasses at the bottom

tutorials/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -87,6 +91,12 @@ Here is a complete class example based on these guidelines:
     func _on_state_changed(previous, new):
         print("state changed")
         emit_signal("state_changed")
+        
+    class State:
```
YuriSizov, line 95: Whatever the order, this should have two lines before it, otherwise it sticks to the method above for no reason.

YuriSizov:
I also intuitively put subclasses at the bottom in GDScript. That creates a clear separation when they have properties and methods of their own. I think this is a valid style rule.

YuriSizov (approved):

YuriSizov:
Bump :)

YuriSizov:
Thanks!

PR #6384, opened 2022-11-14
Change match typeof() example from REAL to FLOAT

FeralBytes:
TYPE_REAL no longer exists, changed to TYPE_FLOAT


Calinou (approved):

PR #6394, opened 2022-11-19
Update position of the Ethical banner when folding changes

YuriSizov:
This should solve the issue when the EthicalAds banner doesn't update its position as the folding of sections or TOC trees change. For the new foldable sections the solution is trivial, as we control the folding logic. For the TOC trees the issue is that the expand buttons are added by the theme that we use, and are not a part of the initial layout. So I had to devise a trick with the mutation observer, again, to capture the moment they are added and attach to their click events. In local testing all the issues seem to be gone now.

But I'd appreciate a look from @Calinou or @coppolaemilio because my webdev skills are getting rusty.

PS. I also updated some recently added code to follow our codestyle better.

### DIFF
--- a/_static/js/custom.js
+++ b/_static/js/custom.js
@@ -195,6 +195,40 @@ const registerSidebarObserver = (function(){
 
     const observer = new MutationObserver(observerCallback);
     observer.observe(sidebarContainer, observerConfig);
+
+    // Default TOC tree has links that immediately navigate to the selected page. Our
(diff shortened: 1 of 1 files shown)


Calinou (approved):
Looks good to me at a glance (not tested).

mhilbrunner (approved):

mhilbrunner:
> Looks good to me at a glance (not tested).

Same :)

akien-mga:
Thanks!

PR #6402, opened 2022-11-21
Add a page on variable rate shading (VRS)

Calinou:
~~**Marked as draft**, as I haven't been able to use VRS successfully so far (despite https://github.com/godotengine/godot/pull/68710). It doesn't give me the expected visual result, with 2D rendering and geometry edges also being affected when they shouldn't. Getting VRS to work correctly is needed so I can create comparison screenshots and performance figures.~~

**Edit:** I got VRS working on NVIDIA, so I was able to add screenshots and a performance comparison.

Nonetheless, the text portion of the documentation is ready for review.

**Testing project:** [test_vrs.zip](https://github.com/godotengine/godot-docs/files/10471632/test_vrs.zip)


tutorials/3d/variable_rate_shading.rst
```diff
@@ -0,0 +1,116 @@
...
+
+Variable rate shading is only supported on specific desktop and laptop graphics cards:
+
+- NVIDIA Turing and newer (including GTX 1600 series)
+- AMD RDNA2 and newer (both integrated and dedicated GPUs – including Steam Deck)
+- Intel Arc Alchemist and newer **(dedicated GPUs only)**
+
+  - Intel integrated graphics do not support variable rate shading.
+
+As of November 2022, mobile GPUs do not support variable rate shading.
```
BastiaanOlij, line 37: Is this correct? I thought that was one of the important things about the GPUs in Meta Quests, they definitely utilize VRS but maybe they do their own thing here...

Calinou: Indeed, it looks like my statement was wrong. The Snapdragon 888 and newer support variable rate shading: https://www.androidcentral.com/qualcomm-snapdragon-888

Here's a benchmark: https://hothardware.com/news/heres-how-current-phones-perform-variable-rate-shading-games

Calinou:
This PR is now ready for review. Changes since the last revision:

- Add screenshots of textured and untextured scene, with and without VRS.
- Add performance comparison at various viewport resolutions.
- Add tip about mipmap LOD bias to reduce blurriness when enabling VRS.

BastiaanOlij (approved):
This looks great to me, did make me realize we should expose some of the GPU capabilities through the render server, it would be nice to be able to query in code whether things like VRS, Multiview, compute shaders, etc. are supported by the platform so the developer can adjust accordingly.

YuriSizov:
Thanks!

PR #6399, opened 2022-11-20
Adjust the class reference badge to use an icon

YuriSizov:
Closes https://github.com/godotengine/godot-docs/issues/5411. Got an idea on RC to go with a familiar icon from the editor that is associated with the class reference. I think it fits fine here, but am open to alternatives. It is color-adjusted for the dark and light themes, using Godot's own palette. Again, open to suggestions to do something else with the colors, but I think it's okay overall, and doesn't steal too much attention.

Overall, I think this is already way less obnoxious.

**Dark mode:**
[Before](https://user-images.githubusercontent.com/11782833/202878182-dcd6632f-513c-43a7-836a-5190947cab76.png) | [After](https://user-images.githubusercontent.com/11782833/202878185-ff3eb7c4-577a-49a5-99c9-36d7d09055db.png)

**Light mode:**
[Before](https://user-images.githubusercontent.com/11782833/202878198-69040183-6f8b-45b7-a361-01f74b48ca3e.png) | [After](https://user-images.githubusercontent.com/11782833/202878200-f60d0b84-7b01-4c48-90df-26fd64a65aef.png)

PR #6404, opened 2022-11-21
Correct exceptions for the class reference badge

YuriSizov:
Follow-up to https://github.com/godotengine/godot-docs/pull/6399. I forgot to properly update exceptions there, so you could see it in the sections of the sidebar again, for example.

Given the opportunity, I've tried to simplify the CSS rules at play, as we only really need the badge in the body of the page, not in the sidebar. This removes the need for a couple of exceptions (including the one I've just added with https://github.com/godotengine/godot-docs/pull/6395, eh).

Also added a note to the cache buster for future generations. We don't actually need to update it every time, just whenever we want the change to immediately rollout for everyone. Minor style fixes, for example, can still wait a day, as far as I'm concerned. It's still better to increment, but the worst that could happen if you forget to — cache would take up to 86400 seconds to invalidate.


akien-mga (approved):

akien-mga:
Thanks!

Issue #6386, opened 2022-11-16
SurfaceTool tutorial/doc about index() wrong or confusing


pseidemann:
**Your Godot version:**

stable

**Issue description:**

on the page "Using the SurfaceTool", the index() method is briefly mentioned.
what I understand from this page is that index() removes duplicate vertices. how it knows which vertices are duplicate is not mentioned and unknown to me (I suspect identical positions? what about rounding errors?).

"[...] calling index() which shrinks the vertex array to remove duplicate vertices."

there is this short snippet:
```gdscript
# Creates a quad from four corner vertices.
# Add_index does not need to be called before add_vertex.
st.add_index(0)
st.add_index(1)
st.add_index(2)

st.add_index(1)
st.add_index(3)
st.add_index(2)

# Alternatively:
st.index()
```

this implies that if I add 4 vertices and then call index(), I will get a quad. from the documentation of index() it's not clear that this is what will happen. indeed I cannot make index() work to create a quad for me without using add_index().

I suspect the documentation is wrong here or quite confusing. I don't know which information is correct to fix it myself. what is index() really doing?

the documentation in the class for this method is also not clear about it:
"Shrinks the vertex array by creating an index array. This can improve performance by avoiding vertex reuse."

glad for any insights!

**URL to the documentation page:**

https://docs.godotengine.org/en/stable/tutorials/3d/procedural_geometry/surfacetool.html

https://docs.godotengine.org/en/stable/classes/class_surfacetool.html#class-surfacetool-method-index

clayjohn:
More detail could indeed be helpful. ``index()`` essentially runs through the list of vertices and notes which are duplicates (duplicate vertices are vertices with the same position and attribute info (normal, UV, etc.), then it shrinks the vertex array to only contain unique vertices, and generates an index array that maintains the original mesh structure. 

To turn a list of vertices into a quad you would need 6 vertices. ``index()`` would then collapse that into 4 vertices and an index array of length 6.

pseidemann:
> To turn a list of vertices into a quad you would need 6 vertices. `index()` would then collapse that into 4 vertices and an index array of length 6.

this is actually not true. the snippet from the doc is pretty bad, but if you add 4 vertices and then add 6 indices (add_index(), not index()) all in the right order you get a mesh. I got this working after some trial and error.
see also this for reference: https://gamedev.stackexchange.com/a/10741

clayjohn:
[...]

PR #6405, opened 2022-11-21
improve documentation of SurfaceTool

pseidemann:
draft!

early feedback appreciated.

following #6386, I want to improve the documentation of SurfaceTool (tutorial + class reference) regarding index vertices (index()/add_index()).

### DIFF
--- a/tutorials/3d/procedural_geometry/surfacetool.rst
+++ b/tutorials/3d/procedural_geometry/surfacetool.rst
@@ -15,25 +15,33 @@ Attributes are added before each vertex is added:
 .. tabs::
  .. code-tab:: gdscript GDScript
 
+    var st = SurfaceTool.new()
+
+    st.begin(Mesh.PRIMITIVE_TRIANGLES)
+
     st.add_normal() # Overwritten by normal below.
     st.add_normal() # Added to next vertex.
     st.add_color() # Added to next vertex.
     st.add_vertex() # Captures normal and color above.
     st.add_normal() # Normal never added to a vertex.
 
-When finished generating your geometry with the :ref:`SurfaceTool <class_surfacetool>`
+When finished generating your geometry with the :ref:`SurfaceTool <class_surfacetool>`,
 call ``commit()`` to finish generating the mesh. If an :ref:`ArrayMesh <class_ArrayMesh>` is passed
-to ``commit()`` then it appends a new surface to the end of the ArrayMesh. While if nothing is passed
+to ``commit()``, then it appends a new surface to the end of the ArrayMesh. While if nothing is passed
 in, ``commit()`` returns an ArrayMesh.
 
 .. tabs::
  .. code-tab:: gdscript GDScript
 
+    # Add surface to existing ArrayMesh.
     st.commit(mesh)
-    # Or:
+
+    # -- Or Alternatively --
+
+    # Create new ArrayMesh.
     var mesh = st.commit()
 
-Code creates a triangle with indices
+Code creates a triangle with indices.
 
 .. tabs::
  .. code-tab:: gdscript GDScript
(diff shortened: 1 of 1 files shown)


tutorials/3d/procedural_geometry/surfacetool.rst
```diff
@@ -15,25 +15,33 @@ Attributes are added before each vertex is added:
...
     st.commit(mesh)
-    # Or:
+
+    # -- Or Alternatively --
+
+    # Create new ArrayMesh.
     var mesh = st.commit()
 
-Code creates a triangle with indices
+Code creates a triangle with indices.
```
clayjohn, line 44: While we were at it, we should fix this sentence
```suggestion
The below code creates a triangle without indices.
```

tutorials/3d/procedural_geometry/surfacetool.rst
```diff
@@ -76,7 +86,11 @@ to remove duplicate vertices.
     st.add_index(3)
     st.add_index(2)
 
-    # Alternatively:
+    # If don't want to use add_index() but just index(), you need to add six
+    # vertices via add_vertex() (three for each triangle corner).
+    # index() will then take care of generating the correct array index.
+    # it will also remove duplicate vertices where all attributes,
+    # including the position are the same.
```
clayjohn, line 93: Instead of this long paragraph, we should modify the example to be more intuitive. 

One way of doing that is to fill out the example a little more. 

I recommend adding the following just above ``# Creates a quad from four corner vertices.``:

```
# Suppose we have a quad defined by 6 vertices as follows
st.add_vertex(Vector3(-1, 1, 0))
st.add_vertex(Vector3(1, 1, 0))
st.add_vertex(Vector3(-1, -1, 0))

st.add_vertex(Vector3(1, 1, 0))
st.add_vertex(Vector3(1, -1, 0))
st.add_vertex(Vector3(-1, -1, 0))

# We can make the quad more efficient by using an index array and only utilizing 4 vertices

```

Then we can replace the long comment with:

```suggestion
    # Alternatively we can use ``st.index()`` which will create the quad for us and remove the duplicate vertices
```

clayjohn (changes requested):
Great work so far, just a few comments to clear things up a bit more

pseidemann:
hi @clayjohn,
thanks for the help!

is this how you meant your suggestion regarding the code snippet?

I tried to document both ways (duplicate vertices vs "efficient" vertices) but maybe that's too much for one snippet.

clayjohn:
> hi @clayjohn, thanks for the help!
> 
> is this how you meant your suggestion regarding the code snippet?
> 
> I tried to document both ways (duplicate vertices vs "efficient" vertices) but maybe that's too much for one snippet.

Yep, looks good to me

tutorials/3d/procedural_geometry/surfacetool.rst
```diff
@@ -60,14 +68,26 @@ Code creates a triangle with indices
     var mesh = st.commit()
 
 You can optionally add an index array, either by calling ``add_index()`` and adding
-vertices to the index array or by calling ``index()`` which shrinks the vertex array
-to remove duplicate vertices.
+vertices to the index array manually, or by calling ``index()`` once,
+which generates the index array automatically and
+shrinks the vertex array to remove duplicate vertices.
```
stevenctl, line 73: I think the definition of duplicate needs to be called out here. With regards to:

- which vertex properties? Colors? Normals? 
- how is floating point rounding dealt with?

The only way I would have found this info is by reading the linked issue/source.

pseidemann: hi,

unfortunately I don't have such deep insights into godot so I'm not sure how e.g. rounding is handled. it would be good if someone else could take this over

skyace65 (approved):
Did a rebase on this

skyace65:
Thanks!

PR #6409, opened 2022-11-21
Add a page on custom performance monitors

Calinou:
There's probably a bit more information that could be added, but I'm not too sure what to add right now.

See https://github.com/godotengine/godot-docs/issues/5121.

PS: I noticed that negative values *are* supported in the Performance singleton, but the editor doesn't graph them correctly. We should look into fixing this.


mhilbrunner (approved):

tutorials/scripting/debug/custom_performance_monitors.rst
```diff
@@ -0,0 +1,117 @@
...
+appearing as follows:
+
+.. :figure: img/custom_performance_monitors_graph_example.webp
+   :align: center
+   :alt: Example editor graph from a custom performance monitor
+
+   Example editor graph from a custom performance monitor
+
+The performance monitor is only queried if the project is being run from the editor.
+The second parameter is a :ref:`class_Callable`.
```
skyace65, line 97: This sentence seems out of place. I assume it's referring to `Performance.add_custom_monitor()`, but it's several paragraphs after that part of the code, I thought it was referring to something related to what you see in the debugger panel at first.

skyace65 (changes requested):
Looks very good! I only have 1 issue.

skyace65 (approved):

PR #6414, opened 2022-11-22
Document global and per-instance uniforms in Shading language

Calinou:
See https://github.com/godotengine/godot-docs/issues/5121. This takes some parts from https://godotengine.org/article/godot-40-gets-global-and-instance-shader-uniforms (thanks reduz :slightly_smiling_face:).

Some questions I have to further complete this documentation:

- Does using `GeometryInstance3D.get_instance_shader_parameter()` have a performance cost like getting global uniform values through a script? I don't get any warning when using it, but I also can't seem to get any value from it anyway.
- What's the difference between RenderingServer's `global_shader_parameter_set()` and `global_shader_parameter_set_override()`? I'm not sure what use cases each method has in practice. The former fails with an error if the uniform doesn't exist, while the latter fails silently. Only the latter fails early if the uniform's type extends Object.
  - Each uniform variable stores the value and override separately: https://github.com/godotengine/godot/blob/e3a51e53ef055a1fe94351b68173c6dfcd2cb371/servers/rendering/renderer_rd/storage_rd/material_storage.h#L123-L124
- Is a default value assigned to a global uniform ever used? Shader compilation fails if the global uniform isn't defined in the project settings.


mhilbrunner:
cc @clayjohn :) Looks good to me and spotted nothing on a first glance, will give it a more thorough read later.

tutorials/shaders/shader_reference/shading_language.rst
```diff
@@ -839,6 +839,191 @@ The syntax also supports subgroups (it's not mandatory to declare the base group
 
     group_uniforms MyGroup.MySubgroup;
 
+Global uniforms
+^^^^^^^^^^^^^^^
+
+Sometimes, you want to modify a parameter in a lot of instances of a material.
+This takes a lot of work as all these instances need to be tracked and the
+uniform needs to be set for each of them.
```
clayjohn, line 847:
```suggestion
Sometimes, you want to modify a parameter in many different shaders at once.
With a regular uniform this takes a lot of work as all these shaders need to be
tracked and the uniform needs to be set for each of them. Global uniforms allow
you to create and update uniforms that will be available in all shaders.
```

tutorials/shaders/shader_reference/shading_language.rst
```diff
@@ -839,6 +839,191 @@ The syntax also supports subgroups (it's not mandatory to declare the base group
...
+   would be more time-consuming to set this on every material when it starts
+   raining.
+5. **Providing team colors.** In team-based games, you can use global uniforms
+   to provide "friendly" and "enemy" colors that can automatically be used
+   everywhere relevant (player materials, map textures, special effects, GUI
+   elements, ...). This can be used to let players choose team colors freely, for
+   better accessibility. The "friendly" and "enemy" states themselves can be
+   determined by using a :ref:`per-instance uniform
+   <doc_shading_language_per_instance_uniforms>` that indicates the instance's
+   current team.
```
clayjohn, line 886: I would massively simplify this whole block. We don't need to provide a detailed list to the user about where they might find this helpful. Such a list makes sense for a blog post where we are promoting a new feature, but this is an API reference.

```suggestion
Global uniforms are especially useful for environmental effects that affect many objects in a scene like having foliage bend when the player is near or having objects move with the wind. 
```

clayjohn (commented):
Great work! I left a few comments and suggestions

tutorials/shaders/shader_reference/shading_language.rst
```diff
@@ -839,6 +839,154 @@ The syntax also supports subgroups (it's not mandatory to declare the base group
...
+you to create and update uniforms that will be available in all shaders.
+
+Global uniforms are especially useful for environmental effects that affect many
+objects in a scene, like having foliage bend when the player is nearby, or having
+objects move with the wind.
+
+All these use cases are feasible thanks to global uniforms, which can be used in
+every shader type (``canvas_item``, ``spatial``, ``particles``, ``sky`` and
+``fog``).
+
```
clayjohn, line 857:
```suggestion
```
We can move this above

tutorials/shaders/shader_reference/shading_language.rst
```diff
@@ -839,6 +839,154 @@ The syntax also supports subgroups (it's not mandatory to declare the base group
...
     group_uniforms MyGroup.MySubgroup;
 
+Global uniforms
+^^^^^^^^^^^^^^^
+
+Sometimes, you want to modify a parameter in many different shaders at once.
+With a regular uniform this takes a lot of work as all these shaders need to be
+tracked and the uniform needs to be set for each of them. Global uniforms allow
+you to create and update uniforms that will be available in all shaders.
+
```
clayjohn, line 849:
```suggestion
you to create and update uniforms that will be available in all shaders and
in every shader type (``canvas_item``, ``spatial``, ``particles``, ``sky`` and
``fog``)

```

clayjohn (commented):
> Does using GeometryInstance3D.get_instance_shader_parameter() have a performance cost like getting global uniform values through a script? I don't get any warning when using it, but I also can't seem to get any value from it anyway.

Not really, it is just reading from a hashmap

> What's the difference between RenderingServer's global_shader_parameter_set() and global_shader_parameter_set_override()? I'm not sure what use cases each method has in practice. The former fails with an error if the uniform doesn't exist, while the latter fails silently. Only the latter fails early if the uniform's type extends Object.
> - Each uniform variable stores the value and override separately: https://github.com/godotengine/godot/blob/e3a51e53ef055a1fe94351b68173c6dfcd2cb371/servers/rendering/renderer_rd/storage_rd/material_storage.h#L123-L124

Its for use by the [``ShaderGlobalOverrides``](https://docs.godotengine.org/en/latest/classes/class_shaderglobalsoverride.html) node

> Is a default value assigned to a global uniform ever used? Shader compilation fails if the global uniform isn't defined in the project settings.

I don't think so. Imagine if you had multiple shaders that each used a different default value, which one should be picked?

mhilbrunner (approved):

PR #6423, opened 2022-11-24
Updated "2D movement overview"

cbscribe:
Code samples updated for Godot 4, and text updated to match. Also updated the downloadable project.

Please check c# code, as I think I got everything, but couldn't test as I don't use it myself.

Note that code is a lot more concise using `Input.get_axis()` and `Input.get_vector()`. These changes could probably be useful for 3.x docs as well.


tutorials/2d/2d_movement.rst
```diff
@@ -315,10 +245,9 @@ on the screen will cause the player to move to the target location.
 
     public class Movement : CharacterBody2D
     {
-        [Export] public int speed = 200;
+        [Export] public int speed = 400;
 
         public Vector2 target = new Vector2();
```
raulsntos, line 250: In GDScript the `target` field is initialized to `position`, since it's not possible to initialize to a non-static member in C# the closest equivalent would be to set this in the constructor or maybe `_Ready`. Not sure how important this is, it may not be worth adding if it distracts from the main goal of the tutorial.

tutorials/2d/2d_movement.rst
```diff
@@ -241,33 +182,23 @@ is set by the mouse position instead of the keyboard. The character will always
...
-                velocity = new Vector2(-speed, 0).Rotated(Rotation);
-
-            if (Input.IsActionPressed("up"))
-                velocity = new Vector2(speed, 0).Rotated(Rotation);
-
-            velocity = velocity.Normalized() * speed;
+            Velocity = Transform.x * Input.get_axis("down", "up") * speed
         }
 
         public override void _PhysicsProcess(float delta)
```
raulsntos, line 193: In 4.0, the `delta` parameter of `_Process` and `_PhysicsProcess` is now a 64-bit type (like in GDScript).
```suggestion
        public override void _PhysicsProcess(double delta)
```

raulsntos (changes requested):
I reviewed the C# changes, I took the opportunity to suggest a few changes to follow our conventions more closely but I realize there are many parts of the documentation that don't currently follow the C# conventions and it may be a bit out of scope for this PR.

tutorials/2d/2d_movement.rst
cbscribe: That was to prevent the sprite from moving to (0, 0) at first. It's not particularly important, and it wasn't there in the original version.

mhilbrunner (approved):
Looks great!

Can you convert the .png screenshot to a lossless .webp? We're in the process of switching to WebP for newly created images. Programs like GIMP can save to WebP, and there's converters out there too.

tutorials/2d/2d_movement.rst
```diff
@@ -36,77 +36,52 @@ fact that the player can move diagonally by pressing two keys at the same time.
...
-            velocity = MoveAndSlide(velocity);
+            MoveAndSlide();
         }
     }
 
-In the ``get_input()`` function, we check for the four key events and sum them
-up to get the velocity vector. This has the benefit of making two opposite keys
-cancel each other out, but will also result in diagonal movement being faster
-due to the two directions being added together.
+In the ``get_input()`` function, we use `Input.get_vector()` to check for the
```
raulsntos, line 80: I think it may be useful to link to the `Input` documentation here, what do you think?

```suggestion
In the ``get_input()`` function, we use :ref:`Input <class_Input>` ``get_vector()`` to check for the
```

Otherwise, the syntax here should be using two single quotes.

```suggestion
In the ``get_input()`` function, we use ``Input.get_vector()`` to check for the
```

cbscribe:
Thanks for the C# fixes. Yeah, I'll squash all the commits once we're done.

cbscribe:
> Can you convert the .png screenshot to a lossless .webp? 

Sure. Last I saw the webp decision was still under discussion, but I can change that.

raulsntos (commented):
I believe these are the last required C# changes :crossed_fingers:

tutorials/2d/2d_movement.rst
```diff
@@ -305,20 +240,20 @@ on the screen will cause the player to move to the target location.
...
-    public class Movement : CharacterBody2D
+    public partial class Movement : CharacterBody2D
     {
-        [Export] public int speed = 200;
+        [Export]
+        public int Speed { get; set; } = 400;
 
-        public Vector2 target = new Vector2();
-        public Vector2 velocity = new Vector2();
+        private Vector2 _target;
```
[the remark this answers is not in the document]
cbscribe: Done

cbscribe:
OK, last c# changes added and image converted to webp. If there are no more suggestions, I'll squash and merge.

raulsntos (approved):
LGTM

PR #6438, opened 2022-11-29
Fix midword link wrapping issues using Javascript

adamscott:
Adds `<wbr>` (which is suggested linebreaks) to each reference links once the page is ready using RegEx.

Capture d’écran du 2022-11-29 12-49-36

Capture d’écran du 2022-11-29 12-49-52

Similar to the objective of #6436, but more efficient.

### DIFF
--- a/_static/js/custom.js
+++ b/_static/js/custom.js
@@ -250,6 +250,30 @@ $(document).ready(() => {
     registerOnScrollEvent(mediaQuery);
   });
 
+  // Add line-break suggestions to class refernece navigation items.
+  const classReferenceLinks = document.querySelectorAll('.wy-menu-vertical > ul:last-of-type .reference.internal');
+  for (const linkItem of classReferenceLinks) {
+    let textNode = null;
+    linkItem.childNodes.forEach(it => {
+      if (it.nodeType === Node.TEXT_NODE) {
+        // If this is a text node and if it needs to be updated, store a reference.
+        let text = it.textContent;
+        if (!(text.includes(" ") || text.length < 10)) {
+          textNode = it;
+        }
+      }
+    });
+
+    if (textNode != null) {
+        let text = textNode.textContent;
+        // Adds suggested line-breaks, if needed, and replace the original text.
+        text = text.replace(/((?:(?<=[a-z])[A-Z0-9](?!$))|(?<!^)[A-Z](?=[a-z]))/gm, '<wbr>$1');
+
+        linkItem.removeChild(textNode);
+        linkItem.insertAdjacentHTML('beforeend', text);
+    }
+  }
+
   if (inDev) {
     // Add a compatibility notice using JavaScript so it doesn't end up in the
     // automatically generated `meta description` tag.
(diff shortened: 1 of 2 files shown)


adamscott:
Here's a link to the testbed of the regexp used.
https://regex101.com/r/Bn4Fq1

https://github.com/godotengine/godot-docs/blob/ef4ef8eb7d5705cb8add333310eda6f9f344fa1e/_static/js/custom.js#L255-L261

YuriSizov:
This is the best idea that we have at the moment, but I'd love to avoid using JavaScript. Problem is I can't find a template used for the TOC tree, so I'm starting to doubt there is one (and that means we can't easily customize it at the build step).

_static/js/custom.js
```diff
@@ -250,6 +250,16 @@ $(document).ready(() => {
     registerOnScrollEvent(mediaQuery);
   });
 
+  // Add line-break suggestions to code links
+  const referenceInternalList = document.body.querySelectorAll(".reference.internal");
```
YuriSizov, line 254: We use single quotes in this file.

I think we should target specifically the last `<ul>` of `.wy-menu-vertical`, so it only affects the class reference pages. Also using `body` doesn't seem necessary here.

_static/js/custom.js
```diff
@@ -250,6 +250,16 @@ $(document).ready(() => {
...
+  for (const el of referenceInternalList) {
+    if (el.innerText.includes(" ") || el.innerText.length < 10) {
+      continue;
+    }
+
+    el.innerHTML = el.innerText.replace(/((?:(?<!^|[A-Z0-9])[A-Z0-9])|(?:(?<!^)[A-Z0-9](?=[a-z])))/gm, "<wbr>$1");
```
YuriSizov, line 260: I'd ignore numbers and consider them a part of the previous substring. Because while we have `2D` and `3D`, we also have just numbers at the end, and those may end up looking bad:

image

adamscott: I could add a check for the end line.

YuriSizov:
I'm concerned about the sidebar jumping around and general opportunity for a race condition. If this code runs after the expand buttons are added (and they are added by our sphinx theme with JavaScript), then this code seems to remove them. Should be fixable if you don't use `innerText` though.

_static/js/custom.js
YuriSizov: That could work, but this regex is already unnerving 🙃

adamscott:
New regex expression, removes end-of-line orphans.
https://regex101.com/r/t6Vczg

adamscott:
> I'm concerned about the sidebar jumping around and general opportunity for a race condition. If this code runs after the expand buttons are added (and they are added by our sphinx theme with JavaScript), then this code seems to remove them. Should be fixable if you don't use `innerText` though.

I will have to think about this.

adamscott:
@YuriSizov I edited my PR to take account the possible race condition.

YuriSizov:
The approach seems to work fine, but I rewrote it in a safer manner.

YuriSizov (approved):

YuriSizov:
Thanks!

YuriSizov:
This caused a regression, as Safari does not support lookbehind and our custom script just breaks on that line.

PR #6442, opened 2022-11-30
C#: Add table with equivalent string methods

raulsntos:
Adds a table with the list of methods C# methods that are equivalent to the methods in Godot's String (including links to Microsoft's documentation and alternative APIs to consider).

The goal with this is making sure users can find the equivalent C# method to the Godot's String methods, specially when the method is available under a different type/namespace or has a different name in .NET's BCL. Sometimes, providing alternative APIs that are similar and can be useful to consider as an alternative, since sometimes a 1-to-1 replacement does not exist.

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_differences.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_differences.rst
@@ -177,42 +177,203 @@ Example:
 String
 ------
 
-Use ``System.String`` (``string``). Most of Godot's String methods are
-provided by the ``StringExtensions`` class as extension methods.
+Use ``System.String`` (``string``). Most of Godot's String methods have an
+equivalent in ``System.String`` or are provided by the ``StringExtensions``
+class as extension methods.
 
 Example:
 
 .. code-block:: csharp
 
-    string upper = "I LIKE SALAD FORKS";
-    string lower = upper.ToLower();
+    string text = "Bigrams";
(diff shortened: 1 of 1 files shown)


tutorials/scripting/c_sharp/c_sharp_differences.rst
```diff
@@ -177,42 +177,203 @@ Example:
...
+get_string_from_utf16      StringExtensions.GetStringFromUTF16 (Consider using `System.Text.Encoding.UTF16.GetString`_)
+get_string_from_utf32      StringExtensions.GetStringFromUTF32 (Consider using `System.Text.Encoding.UTF32.GetString`_)
+get_string_from_utf8       StringExtensions.GetStringFromUTF8 (Consider using `System.Text.Encoding.UTF8.GetString`_)
+hex_encode                 StringExtensions.HexEncode (Consider using `System.Convert.ToHexString`_)
+=========================  ==============================================================
+
+* Strings are immutable in .NET, so all methods that manipulate a string don't
+  modify the original string and return a newly created string with the
+  modifications applied. To avoid creating multiple string allocations consider
+  using a `StringBuilder`_.
```
neikeq, line 311: I would move this paragraph to the top because the detail about immutability of strings is important.

neikeq (approved):

neikeq:
Thanks!

PR #6444, opened 2022-12-01
Change funcref to Callable in example code.

### LINKED ISSUE #6443: funcref renamed to Callable
**Your Godot version:**
v4.0.beta6.official [7f8ecffa5]

**Issue description:**
Documentation code examples mentions `funcref` which is superseded by `Callable`

[Godot interfaces](https://docs.godotengine.org/en/latest/tutorials/best_practices/godot_interfaces.html)
[Scene organizaton](https://docs.godotengine.org/en/latest/tutorials/best_practices/scene_organization.html)

jomiq:
Small  change. 
Resolves #6443

### DIFF
--- a/tutorials/best_practices/godot_interfaces.rst
+++ b/tutorials/best_practices/godot_interfaces.rst
@@ -472,7 +472,7 @@ accesses:
     @onready var child = $Child
 
     func _ready():
-        child.fn = funcref(self, "print_me")
+        child.fn = Callable(self, "print_me")
         child.my_method()
 
     func print_me():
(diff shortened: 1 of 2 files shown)


YuriSizov:
These samples need to be rewritten, because there is no point constructing Callables like that. Each function is already a Callable, it can be simply referenced.

YuriSizov:
I don't think that the issue that you've linked has any relevance here, it's from 2019. `Callable` is a new concept for Godot 4, and it's not just a rename of `funcref`. Each method of a class is already a `Callable`, so the examples can use them directly, instead of creating `funcref`s. This still needs to be changed, just not the way that you've done it. You can still improve this PR and correct the code samples.

PR #6446, opened 2022-12-01
Replace git checkout

Welteam:
As the command "git checkout" became pretty much deprecated as a top level command because it was deemed too confusing, it is now replaced by "git switch" to navigate branches and "git restore" to change the state of files. I propose we update the page "Pull request workflow" to reflect those changes in git as this tutorial is targeted at beginners.

### DIFF
--- a/community/contributing/pr_workflow.rst
+++ b/community/contributing/pr_workflow.rst
@@ -152,20 +152,20 @@ a feature branch:
     $ git branch better-project-manager
 
     # Change the current branch to the new one
-    $ git checkout better-project-manager
+    $ git switch better-project-manager
 
 This command is equivalent:
 
 ::
 
     # Change the current branch to a new named one, based on the current branch
-    $ git checkout -b better-project-manager
+    $ git switch -c better-project-manager
 
 If you want to go back to the ``master`` branch, you'd use:
 
 ::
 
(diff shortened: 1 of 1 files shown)


skyace65:
I don't closely follow the development of Git. But the [official documentation](https://git-scm.com/docs/git-switch/2.38.0) still considers git switch experimental. So I'm not sure if it makes sense to switch to it yet for the guide.

Calinou:
I agree `git switch` is better to use overall, but the version of Git shipped in some LTS Linux distributions may not contain this command at all (such as Ubuntu 18.04, which will only be EOL in April 2023).

skyace65:
@Calinou I just checked in an ISO, 20.04 ships with git version 25, and checkout was added in 27. Though their is an official git PPA you can add to get the latest version.

Calinou:
> @Calinou I just checked in an ISO, 20.04 ships with git version 25, and checkout was added in 27. Though their is an official git PPA you can add to get the latest version.

I see. I don't think we should force users to bother with a PPA just for a convenience comamnd (as PPAs make distro upgrades more difficult, etc).

Welteam:
I agree so I'm closing this pull request. I've been using a rolling distro for so long, I forgot those issues existed

PR #6450, opened 2022-12-02
Replace all uses of `hint_color`/`hint_albedo` with `source_color`

vmedea:
These have been replaced in Godot 4.
This is my first PR to the Godot docs repository so i hope i got this right.


Calinou (changes requested):
Class reference files should not be modified, as these are generated automatically.

vmedea:
That makes sense. I've removed the changes under `classes/`.

Calinou (approved):
Thanks!

Issue #6452, opened 2022-12-04
Missing Documentation Website


IXLLEGACYIXL:
**Your Godot version:**
GD 4
**Issue description:**
There is a Missing Webpage in the docs
When i click on "See this Page" at the top it throws an "site not found" error

**URL to the documentation page:**
https://docs.godotengine.org/en/latest/development/cpp/unit_testing.html

Calinou:
This is because the page in question does not exist in the stable branch. The feature is not available in 3.x, so the warning isn't relevant here.

We can probably try to do a fetch request on the page URL and see if it exists at all before displaying the link.

YuriSizov:
> We can probably try to do a fetch request on the page URL and see if it exists at all before displaying the link.

We already prefetch it, so I guess doing a `fetch(url, { method: 'HEAD' })` wouldn't be too expensive. Though that may create a jump when the notice is added. I can experiment with it a bit.

PR #6453, opened 2022-12-05
Dynamically adjust the compatibility notice in the latest docs

YuriSizov:
Closes https://github.com/godotengine/godot-docs/issues/6452.

This PR makes the compatibility notice dynamic by checking if the stable version of the document exists using a HEAD request. If it does, the old message is displayed (I've adjusted it a bit to better match the new ones). If it doesn't exist, then this fact is stated instead, with a link to do a search in the stable documentation using this page's title.

Code feels rather hacky, but it's probably no worse than what we already do here.

Here's how it looks in practice. _Note, that it should be faster and less noticeable than that in reality. I've throttled my networking speed to better showcase different states._

https://user-images.githubusercontent.com/11782833/205526824-73deabac-1a04-4818-8032-0f461ac7e972.mp4

https://user-images.githubusercontent.com/11782833/205526826-4ee93eef-bca1-42a1-b4b0-98de1c877d0e.mp4

### DIFF
--- a/_static/js/custom.js
+++ b/_static/js/custom.js
@@ -253,22 +253,63 @@ $(document).ready(() => {
   if (inDev) {
     // Add a compatibility notice using JavaScript so it doesn't end up in the
     // automatically generated `meta description` tag.
-    const strippedUrl = [location.protocol, '//', location.host, location.pathname].join('');
-    const updatedUrl = strippedUrl.replace('/latest/', '/stable/');
+
+    const baseUrl = [location.protocol, '//', location.host, location.pathname].join('');
+    // These lines only work as expected in the production environment, can't test this locally.
+    const fallbackUrl = baseUrl.replace('/latest/', '/stable/');
+    const homeUrl = baseUrl.split('/latest/')[0] + '/stable/';
+    const searchUrl = homeUrl + 'search.html?q=';
+
+    // Insert the base notice with a placeholder to display as we're making a request.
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):
Code LGTM, lets see what Calinou thinks.

How does this interact with localization? Can this message be localized? Can the current one?
I've got no idea :)

_static/js/custom.js
```diff
@@ -253,22 +253,63 @@ $(document).ready(() => {
...
     // Add a compatibility notice using JavaScript so it doesn't end up in the
     // automatically generated `meta description` tag.
-    const strippedUrl = [location.protocol, '//', location.host, location.pathname].join('');
-    const updatedUrl = strippedUrl.replace('/latest/', '/stable/');
+
+    const baseUrl = [location.protocol, '//', location.host, location.pathname].join('');
+    // These lines only work as expected in the production environment, can't test this locally.
+    const fallbackUrl = baseUrl.replace('/latest/', '/stable/');
+    const homeUrl = baseUrl.split('/latest/')[0] + '/stable/';
+    const searchUrl = homeUrl + 'search.html?q='
```
Calinou, line 261: Missing semicolon:

```suggestion
    const searchUrl = homeUrl + 'search.html?q=';
```

Calinou (approved):
Should be good to merge after applying the suggestion.

YuriSizov:
Done!

mhilbrunner:
Merged. Let's see if this works out... thanks! :)

PR #6457, opened 2022-12-06
Improve the new class reference look based on feedback

YuriSizov:
Follow-up to https://github.com/godotengine/godot-docs/pull/6445. Based on the early feedback, I've made some further adjustments to the style of the class reference pages.

- Bring back some contrast on the subdued elements, so they don't get annoyingly hard to read and distracting.
- Make the member signature font use the same size throughout instead of a size gradient, which should reduce a nauseating effect when scrolling.
- Various spacing adjustments to go alongside the size changes, and to improve the feel overall with better visual grouping/separation.
- Add new styling for inline code, making it closer to the look of code blocks in a way, and also less noisy.
- Use JetBrains Mono, same as Godot 4 editor, for all monospace text, to make code easier to read and text thickness to have a pleasant look.

Naturally, with a custom font you have some considerations, as it will take a moment to download and for the page to update. But I think it greatly improves the readability of the document, and helps to elevate the new style. If we go back to use some old font definitions, keep in mind that all those % font sizes need to be increased slightly to look good.

<details>
<summary>Another gallery of samples</summary>

chrome_2022-12-06_13-41-37
chrome_2022-12-06_13-41-49
chrome_2022-12-06_13-44-59
chrome_2022-12-06_13-45-18
chrome_2022-12-06_13-45-32
chrome_2022-12-06_13-45-49
chrome_2022-12-06_13-46-15
chrome_2022-12-06_13-46-39
chrome_2022-12-06_13-47-20
chrome_2022-12-06_13-47-31
chrome_2022-12-06_13-47-41
chrome_2022-12-06_13-54-25
chrome_2022-12-06_13-54-31

And just to show how inline code looks in a random article:

chrome_2022-12-06_13-31-20
chrome_2022-12-06_13-31-36

</details>

PR #6459, opened 2022-12-06
Disable font ligatures globally

YuriSizov:
In https://github.com/godotengine/godot-docs/pull/6457 I've disabled font ligatures in JB Mono for code blocks and inline code tags, but they seem to appear in other places as well. So it's better to just disable them throughout and then enable them wherever it may be needed.

Here's an example of this issue:

chrome_2022-12-06_20-01-40

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -261,6 +261,8 @@ legend,
 .rst-versions {
     /* Use a system font stack for better performance (no Web fonts required) */
     font-family: var(--system-font-family);
+    /* Some fonts that we use (namely JetBrains Mono) can come with ligatures. It's better to opt-in if needed. */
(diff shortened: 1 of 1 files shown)


Calinou (approved):

atirut-w:
Why do people not like them? Ligatures even got disabled for code editing in recent betas.

YuriSizov:
Probably because people are less familiar with them. They are an extra feature, not a natural part of typing on a modern PC. And it's not unreasonable that some have harder time reading them.

Ligatures are still available in the editor, if your font supports it. We tried enabling them by default, and people got confused (without even knowing what these are, which means they couldn't easily look up a way to disable them). In the online documentation we don't have a way to leave it to user's preference, unfortunately.

PR #6460, opened 2022-12-06
Update input_examples.rst

ktyldev:
The C# [`InputKeyEvent`](https://docs.godotengine.org/en/stable/classes/class_inputeventkey.html#class-inputeventkey) doesn't have a `.Keycode` property - it has `.Scancode`.

### DIFF
--- a/tutorials/inputs/input_examples.rst
+++ b/tutorials/inputs/input_examples.rst
@@ -203,7 +203,7 @@ the :kbd:`T`:
     {
         if (inputEvent is InputEventKey keyEvent && keyEvent.Pressed)
         {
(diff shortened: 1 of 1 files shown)


raulsntos:
The `master` branch contains the documentation for 4.0.

The `scancode` property was renamed to `keycode` in  https://github.com/godotengine/godot/pull/18020. As a result of this property being renamed in Core, it is also automatically changed in the C# bindings that are automatically generated.

Calinou:
Closing per @raulsntos' comment. The current code is correct.

The stable documentation for 3.x is here: https://docs.godotengine.org/en/stable/tutorials/inputs/input_examples.html

PR #6462, opened 2022-12-06
add_camera_clarification.rst

Chillstad:
Clarifies that the popup is correctly placed within the viewport, which is not necessarily the current Camera. This removes confusion for users who expect the popup to fit into their Camera when that is not always the case.

### DIFF
--- a/classes/class_popup.rst
+++ b/classes/class_popup.rst
@@ -19,7 +19,7 @@ Base container control for popups and dialogs.
 Description
 -----------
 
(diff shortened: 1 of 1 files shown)


YuriSizov:
https://github.com/godotengine/godot/tree/3.5/doc/classes/Popup.xml

You should also make sure this is not needed in the master version. Or perhaps the master version has a better description already that we could backport to 3.x.

Issue #6411, opened 2022-11-22
NavigationAgent3D with Node3D script is incorrect


esoma:
**Your Godot version:** 4.0.dev

**Issue description:** The documentation for using a `NavigationAgent3D` with avoidance with a `Node3D` passes the velocity in to `NavigationAgent3D.set_velocity` with the units "meters per frame" rather than "meters per second". This causes the avoidance to happen much too late requiring an otherwise massively inflated time horizon.

There is also a typo where the `movement_speed` attribute is later referenced as `move_speed`.

**URL to the documentation page:** https://docs.godotengine.org/en/latest/tutorials/navigation/navigation_using_navigationagents.html#actor-as-node3d

akien-mga:
Fixed by https://github.com/godotengine/godot-docs/pull/6463.

PR #6463, opened 2022-12-07
Corrected Navigation3DAgent script

agomez-aj:
Navigation3DAgent script snippet now shows velocity calculated as 'meters per second' rather than 'meters per frame'. "move_speed" has been changed to "movement_speed" for consistency.


agomez-aj:
resolves #6411

Calinou (approved):
Thanks!

PR #6468, opened 2022-12-09
Update compute_shaders.rst

jgonz-hub:
_Bugsquad edit:_ Fixes https://github.com/godotengine/godot/issues/70772


Calinou:
Can you describe why this change is needed?

clayjohn (approved):
Changes make sense, Metal doesn't support doubles in shaders so this would have failed on MacOS devices

Calinou:
> Changes make sense, Metal doesn't support doubles in shaders so this would have failed on MacOS devices

It's probably a good idea to add a warning in the documentation for this :slightly_smiling_face:

clayjohn:
Merging now as this is a good change

PR #6470, opened 2022-12-10
add mobs group with code

penguinDragonTS:
added mob to mobs group via code b/c only using interface to create mobs group (in Mob scene root node groups tab) + adding the group queue clear line :: prevented my game from starting and had 1 fish on screen... 

<strike>Somone else with more experience could tinker around with the groups or why the interface group editing wasn't doing the trick,</strike> [see next comment] but this is how I [initially] got my 2d game to clear the mobs. 

(Note: only gdscript in code viewer is included.)

### DIFF
--- a/getting_started/first_2d_game/06.heads_up_display.rst
+++ b/getting_started/first_2d_game/06.heads_up_display.rst
@@ -413,6 +413,21 @@ click "Groups" and you can type a new group name and click "Add".
 
 .. image:: img/group_tab.png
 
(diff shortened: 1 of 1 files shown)


penguinDragonTS:
Figured out my real problem; on previous page missed this.

"When you're sure everything is working, remove the call to new_game() from _ready()."

Would encourage re-stating this text on HUD page when hooking up new game button. 
Thanks,
🐧🐉

Calinou:
> Would encourage re-stating this text on HUD page when hooking up new game button.

You can amend this pull request to do that instead :slightly_smiling_face:

PR #6476, opened 2022-12-10
Document the `double` and `single` feature tags

mihe:
This updates the list of available feature tags with `double` and `single`, which are introduced in godotengine/godot#69538.

These feature tags signify whether the engine was built with (what's now called) `precision=double` or (the default) `precision=single`.


aaronfranke (approved):

akien-mga (approved):

aaronfranke:
Thanks!

PR #6478, opened 2022-12-11
Mention increased binary size of exported binaries in Upgrading to Godot 4

Calinou:
See https://github.com/godotengine/godot/issues/68647.


skyace65 (approved):
Good information to know, but if possible I think there should be a mention of why binary sizes are bigger so it doesn't seem like things got worse for no reason.

Calinou:
> Good information to know, but if possible I think there should be a mention of why binary sizes are bigger so it doesn't seem like things got worse for no reason.

Done. I've also added a more accessible link to the changelog, and added a link to a list of functional regression issues (i.e. `regression` issues that are not `bug`s).

mhilbrunner (approved):

mhilbrunner:
Thanks!

PR #6482, opened 2022-12-12
Update 01.game_setup.rst key binding on Mac

3mdistal:
I noticed while building my first game that grid snapping is also the `ctrl` key on macOS.

### DIFF
--- a/getting_started/first_3d_game/01.game_setup.rst
+++ b/getting_started/first_3d_game/01.game_setup.rst
@@ -101,7 +101,7 @@ You should see a wide grey slab that covers the grid and blue and red axes in
 the viewport.
 
 We're going to move the ground down so we can see the floor grid. Select the
-``Ground`` node, hold the :kbd:`Ctrl` key down to turn on grid snapping (:kbd:`Cmd` on macOS),
+``Ground`` node, hold the :kbd:`Ctrl` key down to turn on grid snapping,
 and click and drag down on the Y axis. It's the green arrow in the move gizmo.
 
 .. image:: img/01.game_setup/move_gizmo_y_axis.webp


YuriSizov:
If it's the same, we don't need to put it in parentheses at all. But this may actually be a bug in the editor, if the key binding was intended to be Cmd on macOS.

mhilbrunner:
Relevant RC conversation:

> **mhilbrunner**
> Can someone on Mac confirm whether you use CMD or CTRL for grid snapping? re: https://github.com/godotengine/godot-docs/pull/6482
> 
> **bruvzg**
> It's CTRL (hardcoded to CTRL in various canvas editor places - https://github.com/godotengine/godot/blob/8c729f0f34a92b92a1a8204e1d56b6ad05cfd27b/editor/plugins/canvas_item_editor_plugin.cpp#L352). But probably should be changed to CMD since CTRL+click is right click on macOS, so it's only usable if you click first and ten press CTRL, otherwise it will open popup menu.
> 
> **ajreckof**
> I think this be CMD I had one to change all those problem but I forgot to update since 4.0 I need to remove some rename that breaks compatibility [#71905](https://github.com/godotengine/godot/pull/71905). After removing the rename I will recheck every type ctrl is used to determine whether it should be removed or not.
> I think I was unsure at the time whether to include snapping in it but now that I rethink about it it is obvious it should be changed

mhilbrunner:
Rebased and edited to remove the invalid keybinding mention for now until we fix that.

Thanks for contributing, and congrats on your first merged docs contribution! Sorry this has been sitting for so long, thats unfortunately due to a  mixture of us lacking Mac contributors and the 4.0 backlog. :)

PR #6483, opened 2022-12-12
Fix Introduction to the buildsystem links in CLI documentation

YuriSizov:
Fixes a regression from https://github.com/godotengine/godot-docs/pull/6316. The inner id was changed in the buildsystem documentation article, breaking links from other parts of the documentation.

This doesn't look particularly great this way, but we can't nest links inside of inline code tags, so it is what it is:

image


Calinou (approved):
Thanks!

PR #6489, opened 2022-12-14
Missing word added

Zepher79:


YuriSizov:
You need to open a new PR, targeting either the `master` branch, or one of the version branches, for example `3.6`. You should not open a PR against the stable branch, this branch is synced manually with whatever the current stable version is. Changing the base branch won't help in this case, so you need to create a new PR instead.

Prefer fixing the issue in `master` first, then we can backport it to every other affected version.

PR #6481, opened 2022-12-11
Update GDScript Export Docs

Piralein:
- Update and sort the annotation table in gdscript_basics.
- add section "grouping" in gdscript_export.
- rearrange "enums" to its own section outside of basic usage and add better hints.
- remove some unneeded code and add a link to the gdscript annotation page in classref.

PR #6491, opened 2022-12-15
Fix missing @export_flags_2d_navigation/@export_flags_3d_navigation in annotation list

Cretezy:
Listed on https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_exports.html but missing in GDScripts basics list


tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -347,51 +347,55 @@ Both of these are the same::
...
+| ``@export_flags_2d_navigation`` |                                                                                                   |
+|                                 |                                                                                                   |
+| ``@export_flags_3d_render``     |                                                                                                   |
+|                                 |                                                                                                   |
+| ``@export_flags_3d_physics``    |                                                                                                   |
+|                                 |                                                                                                   |
+| ``@export_flags_3d_navigation`` |                                                                                                   |
 +------------------------------+---------------------------------------------------------------------------------------------------+
```
YuriSizov, line 399: Table formatting needs to be aligned.

YuriSizov:
I think we may want to remove that table and link to the class reference instead. It duplicates the same information already present there, and, for export annotations, in the linked dedicated article.

Cretezy:
@YuriSizov I agree, that would be good too. I do see the value of having a list of the symbols there, especially since it's a reference page, where someone can search any of the annotations.

skyace65:
I agree with Yuri on this, listing out every export annotation is excessive. I don't think taking it out makes it harder to find information, if they're looking for a specific export annotation the docs search would bring up the export page. And if they're searching just the reference page to learn what annotations there are, they would find that section which links to the export page.

I think replacing the table with a bullet point list and links is the way to go. Just note on the link for exports that it includes information on other export annotations such as `export_enum` or something along those lines.

mhilbrunner:
Also in favor of not duplicating information :)

YuriSizov:
I'll see what we can do about this, but in the meantime it seems that there is a bigger PR that also covers this enhancement, so I'll close this one as superseded by https://github.com/godotengine/godot-docs/pull/6481.

Thanks for your contribution nonetheless!

PR #6492, opened 2022-12-16
Designing the mob scene - Typo and simplification

denicolas:
In "[Removing monsters off-screen](https://docs.godotengine.org/en/stable/getting_started/first_3d_game/04.mob_scene.html#removing-monsters-off-screen)" > Note:
"You can learn more about that **here** Memory management." should be:
"You can learn more about that **in** Memory management."

In "[Leaving the screen](https://docs.godotengine.org/en/stable/getting_started/first_3d_game/04.mob_scene.html#leaving-the-screen)":
"Head back to the 3D viewport by clicking on the 3D label at the top of the editor. You can also press Ctrl + F2 (Alt + 2 on macOS)." seems unnecessary.

And then:
"This will take you back to the script editor and add a new function for you," could be:
"This will add a new function for you,"

### DIFF
--- a/getting_started/first_3d_game/04.mob_scene.rst
+++ b/getting_started/first_3d_game/04.mob_scene.rst
@@ -69,7 +69,7 @@ Godot has a node that detects when objects leave the screen,
     reason to use pools is to avoid freezes with garbage-collected languages
     like C# or Lua. GDScript uses a different technique to manage memory,
     reference counting, which doesn't have that caveat. You can learn more
-    about that here :ref:`doc_gdscript_basics_memory_management`.
+    about that in :ref:`doc_gdscript_basics_memory_management`.
 
 Select the *Mob* node and add a *VisibilityNotifier* as a child of it. Another
 box, pink this time, appears. When this box completely leaves the screen, the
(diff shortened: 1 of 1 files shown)


skyace65:
The first issue you addressed was fixed in another PR which added a `:` after the word "here" so that sentence works now. I don't see an issue with mentioning the keyboard shortcuts, this whole section is for people who have never used Godot before and that's exactly the kind of information we want to give them. I also don't see an issue with stating that Godot will take you back to the script editor.

I'm open to hearing reasons why those last two changes should be made, and if you still want to do them this PR needs to be rebased.

denicolas:
I don't see an issue with mentioning the keyboard shortcuts either, what I meant is that the user doesn't need to "Head back to the 3D viewport", so that the whole sentence, "Head back to the 3D viewport by clicking on the 3D label at the top of the editor. You can also press Ctrl + F2 (Alt + 2 on macOS)." seems unnecessary.
And then, if you remove that first sentence, you also need to modify:
"This will take you back to the script editor and add a new function for you"
as the user will not have left the script editor, so the sentence could become:
"This will add a new function for you"

mhilbrunner:
Closing due to lack of consensus. I agree these at first look aren't clear improvements. Still thanks for contributing :)

PR #6498, opened 2022-12-18
Update style guide for new export and setget syntax

vmedea:
- `export` → `@export`
- `setget` → `:set(new_value)`

Also update to use new suggested syntax for emitting and connecting signals.

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_styleguide.rst
+++ b/tutorials/scripting/gdscript/gdscript_styleguide.rst
@@ -35,10 +35,12 @@ Here is a complete class example based on these guidelines:
 
     signal state_changed(previous, new)
 
-    export var initial_state = NodePath()
-    var is_active = true setget set_is_active
+    @export var initial_state: Node
+    var is_active = true:
+        set = set_is_active
 
-    @onready var _state = get_node(initial_state) setget set_state
+    @onready var _state = initial_state:
+        get = set_state
     @onready var _state_name = _state.name
 
 
@@ -777,11 +779,13 @@ variables, in that order.
 
    const MAX_LIVES = 3
 
-   export(Jobs) var job = Jobs.KNIGHT
-   export var max_health = 50
-   export var attack = 5
+   @export var job: Jobs = Jobs.KNIGHT
+   @export var max_health = 50
+   @export var attack = 5
 
-   var health = max_health setget set_health
+   var health = max_health:
+       set(new_health):
+           health = new_health
 
    var _speed = 300.0
 
--- a/tutorials/scripting/gdscript/gdscript_styleguide.rst
+++ b/tutorials/scripting/gdscript/gdscript_styleguide.rst
@@ -53,7 +53,7 @@ Here is a complete class example based on these guidelines:
         
 
     func _ready():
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):

tutorials/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -35,7 +35,7 @@ Here is a complete class example based on these guidelines:
 
     signal state_changed(previous, new)
 
-    export var initial_state = NodePath()
+    @export var initial_state := NodePath()
```
YuriSizov, line 38: We should remove this example and export the node directly.

tutorials/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -35,7 +35,7 @@ Here is a complete class example based on these guidelines:
...
     var is_active = true setget set_is_active
```
YuriSizov, line 39: This is another `setget` example that we need to update. I think we can use a method reference here.

tutorials/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -35,7 +35,7 @@ Here is a complete class example based on these guidelines:
...
 
     @onready var _state = get_node(initial_state) setget set_state
```
YuriSizov, line 41: As mentioned above, should export a node directly here. And the setter needs to be updated to the new syntax.

YuriSizov (changes requested):
This looks good to me, but there are more `setget` instances at the start of the article. We should also remove the NodePath example and export a node directly in there.

tutorials/scripting/gdscript/gdscript_styleguide.rst
vmedea: Updated to use a Node directly

vmedea:
Updated according to the comments.
I've also added a commit that changes the use of signals to the currently suggested ways.

tutorials/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -841,12 +845,12 @@ in that order.
...
         self._state = target_state
         _state.enter(msg)
-        Events.emit_signal("player_state_changed", _state.name)
+        Events.player_state_changed.emit(_state.name)
 
 
     func _on_state_changed(previous, new):
         print("state changed")
-        emit_signal("state_changed")
+        state_changed.emit()
```
vmedea, line 853: I've only changed the syntax here, but functionality-wise this seems strange: the function that's connected to the signal re-emits the signal. I'd think the intent was to emit it in the setter instead.

YuriSizov:
Sorry for the wait. I think some of these changes have been addressed by other PRs in the meantime. Would you mind rebasing this and checking if there is anything left to fix?

vmedea:
Sure, updated.

mhilbrunner (approved):

mhilbrunner:
Sorry we've been letting this one sit for so long! Merged and thanks for contributing :)

PR #6500, opened 2022-12-19
Fix NavigationServers doc due to recent renames

smix8:
Fixes NavigationServers  doc page due to recent function renames.


akien-mga (approved):

akien-mga:
Thanks!

PR #6506, opened 2022-12-21
Specify solution to an error users may encounter

wangaustin:
I downloaded autoload.zip provided and followed the tutorial step by step. Once all scripts are finished, I encountered the error specified in the note. I realized that the scripts may not have been attached to their respective scenes. I attached the scripts to the scenes and reconnected the signals to their respective scripts. This fixed the error.

### DIFF
--- a/tutorials/scripting/singletons_autoload.rst
+++ b/tutorials/scripting/singletons_autoload.rst
@@ -295,6 +295,14 @@ the button.
 
 .. note::
 
+    If you received an error saying "emit_signal: Error calling method from signal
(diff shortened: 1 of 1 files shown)


YuriSizov:
This shouldn't ever be needed. Either the attached project is not correct, or the instructions in the tutorial are unclear. This should be fixed with a note in either case.

YuriSizov:
Please open an issue instead, so we can track it and come up with a proper solution once we figure out the source of your problems.

Issue #4107, opened 2020-10-02
Rename Tutorials section to Manual


NathanLovato:
We’re moving towards a reference manual, we should have a label that better suggests that.

Mickeon:
What's your opinion on this 2 years later?

Calinou:
> What's your opinion on this 2 years later?

I think it's a good idea to rename the section, but it'll break many pull requests.

Mickeon:
Manual is a much more appropriate name for what the current "Tutorials" section are used for. It would be worth it in the long run, I feel.

NathanLovato:
It could be automated, I think it shouldn't be too hard as it's mostly about renaming a directory. It's def. best to do a sprint merging or closing as many PRs as possible before that.

PR #6509, opened 2022-12-24
Rename Tutorials to Manual, move migration guide to a higher level

YuriSizov:
Closes https://github.com/godotengine/godot-docs/issues/4107 without changing URLs. We can still do that, if we want to. But the file structure doesn't necessarily needs to map one-to-one to the display name.

I decided to bring a few options up in the "Manual" section, to give them more visibility due to their general importance. That would be "Best practices", "Editor introduction" (renamed from "Editor manual" to reduce repetition) and "Migrating to a new version" (see below). The rest are kept in their alphabetical order.

Side note: I think we can in fact move those to the "Getting Started" section, so that the "Manual" section is not so huge. They also fit there semantically.

"Migrating to a new version" is a new section where I've extracted the "Upgrading to Godot 4" article by @Calinou. It was previously a part of the "Editor" section, but during a production meeting we agreed it's not the best place for it and it should be moved higher, at least, or to a different/new section all together. For now I just moved it higher in the "Manual" section, but as mentioned above wouldn't mind moving it to "Getting Started".

I wrote a new introduction paragraph for this new section too.

This PR also renames some sections for the editor articles. "The editor's interface", while grammatically correct, stood out to me as we don't normally use articles like that in section titles. So now it's simply "Editor's interface". "For developers" is IMO a strange choice, because it's unclear who is identified there as a developer, and why these features are useful for them specifically. I tried to improve it and its description without any more reorganization, but maybe we shouldn't mark these features as advanced at all.

Since I was making a redirect for "Upgrading to Godot 4" anyway, and we will soon enable errors back for the repository, I also removed the old "About" index page which was no longer in use but was kept to preserve potentially indexed external links. It now has a redirect entry as well. cc @mhilbrunner to update those when we go live with this.

chrome_2022-12-24_21-33-42

image


tutorials/editor/index.rst
```diff
@@ -35,12 +35,12 @@ should be aware of.
...
-For developers
---------------
+Advanced features
+-----------------
 
-The articles below focus on features for developers, like calling Godot from the
-command lines and using an external text editor such as Visual Studio Code or
-Emacs.
+The articles below focus on advanced features useful for experienced developers,
+such as calling Godot from the command lines and using an external text editor
```
Calinou, line 42: While we're at it, we can fix a small grammar error:

```suggestion
such as calling Godot from the command line and using an external text editor
```

clayjohn:
Looks good to me! I agree we should be making these changes now. 

I would also be fine moving the migration guide into the getting started section. We could adapt the text in the getting started index to say something like "Getting started with Godot 4" then have different pathways if you are new to Godot, new to programming, new to Godot 4 etc.

YuriSizov:
> I would also be fine moving the migration guide into the getting started section

I'd be fine moving all three 🙃 

chrome_2022-12-24_19-44-16

They kind of fall out of the pattern for Manual, and do look like a good getting started material.

mhilbrunner (approved):
Looks good to me :)

clayjohn:
Will people search for best practices in "getting started"? I wonder if that section deserves a rename to be more clear that it is more inclusive. 

I'm especially worried since we have folded the sections by default. Alternatively, maybe we should have "getting started" unfolded by default in addition to "about"

YuriSizov:
To me Getting started is about the basics and general information about working with the engine. I know it is currently mostly just beginner's materials in there, but I don't think the title itself is limiting.

But we can change it later, if there are doubts.

clayjohn (approved):
Seems good to me. Lets merge now and then get some feedback from more people

YuriSizov:
Need to remember to update redirects, but if https://github.com/godotengine/godot-docs/pull/6516 is going to be merged, can do it after the both are merged

PR #6510, opened 2022-12-24
Update the structure of the home page

YuriSizov:
While working on https://github.com/godotengine/godot-docs/pull/6509 I noticed that we could hide the entire TOC tree in the body of this article. I don't think there is a good reason to keep it, given that the sidebar with the navigation is right there. I assume the fact that it can be hidden without impacting the tree was never looked into.

So I decided to hide those trees, and that led me to do a bit of an overhaul for the entire page. I adjusted phrasing on some paragraphs, and moved them around too. The text is now broken into easily identifiable subsections, with more important content being placed higher.

Additionally I noticed that in the translation notice we incorrectly say that translations only apply to the latest branch, when in fact we only use translations with the stable branch. We also never mention the "Read the Docs" panel in the internationalized version of the notice, unlike the English one. So I fixed those problems with a slight change of the text.

I also noticed an issue with breadcrumbs that only sometimes manifest, so I improved its styling here as well.

-----
Anyway, here's a before and after of the entire page:

[Before](https://user-images.githubusercontent.com/11782833/209448256-6e9feb09-c805-43c1-a083-d529b31a1c19.png) | [After](https://user-images.githubusercontent.com/11782833/209448258-6d89a40d-3eb7-48d3-ba64-560023af1b31.png)

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -918,12 +918,14 @@ code,
 
 /* Admonition tweaks */
 
+.rst-content .admonition,
 .rst-content .admonition.note,
 .rst-content .admonition.seealso {
     background-color: var(--admonition-note-background-color);
     color: var(--admonition-note-color);
 }
 
+.rst-content .admonition .admonition-title,
 .rst-content .admonition.note .admonition-title,
 .rst-content .admonition.seealso .admonition-title {
     background-color: var(--admonition-note-title-background-color);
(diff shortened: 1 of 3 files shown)


Calinou (approved):

mhilbrunner (approved):
LGTM

YuriSizov:
Well then :)

PR #6512, opened 2022-12-24
Hide excessive elements from the RTD panel, improve the looks

YuriSizov:
The "Read the Docs" panel, that gets injected on the live instance, has a few elements that I'd consider excessive. This makes it more cognitively heavy than it needs to be, which sucks for a panel that is used to switch between versions and languages most of the time.

Unfortunately, the layout of the panel is not very well marked up, so I can only resort to tangential information to hide the unnecessary elements, such as GitHub links (we already have them at the top) and a search field (we already have that as well). But it works, until they change something about it, which is as much as we can hope with these things.

I also made smaller visual changes to make the panel a bit tidier and better fitting into our design. Overall it should be slightly nicer to work with, and it should take noticeably less space.

[Before](https://user-images.githubusercontent.com/11782833/209449012-efe42b84-d666-4d4d-b815-494c12e030c4.png) | [After](https://user-images.githubusercontent.com/11782833/209449014-7dbf5a4e-d86a-436b-82a2-8339545396cb.png)

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -1438,6 +1438,24 @@ p + .classref-constant {
     color: var(--navbar-level-1-color);
 }
 
+.rst-versions .rst-current-version,
+.rst-versions .rst-other-versions {
+    padding: 12px 14px;
+}
+
+.rst-versions .rst-other-versions {
+    color: var(--navbar-heading-color);
+}
+
+.rst-versions .rst-other-versions dl + dl {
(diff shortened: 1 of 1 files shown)


Calinou (approved):

mhilbrunner (approved):

mhilbrunner:
In this goes! ❄️ Wow, you've been really working away on the docs web pages... amazing, really. :)

PR #6515, opened 2022-12-26
Update Shading language to refer to `set_shader_parameter()`

jmdejong:
There is no function `set_shader_uniform`.


Calinou (approved):
Thanks!

PR #6524, opened 2022-12-28
Fix: C# example for custom resource tutorial

Danieo6:
The C# example for a custom resource script is wrong. It is going to produce a `Cannot construct temporary MonoObject because the class does not define a parameterless constructor` error.

```csharp
public BotStats(int health = 0, Resource subResource = null, String[] strings = null)
{
    Health = health;
    SubResource = subResource;
    Strings = strings ?? new String[0];
}
```

Using default values for constructor parameters in C# doesn't consider such constructor as parameterless. The constructor in the example should look as follows:

```csharp
public BotStats()
{
    Health = 0;
    SubResource = null;
    Strings = new String[0];
}
```


YuriSizov:
Seems like it is better solved by https://github.com/godotengine/godot-docs/pull/6525

PR #6525, opened 2022-12-28
Fix Resource scripting blocks and C# examples

31:
Fix the scripting blocks in the Resources page. (At least, I looked over it once and saw these issues. I haven't run 100% of the code.) A few main fixes:

* Fix the code tab text by adding newlines. As of writing, on https://docs.godotengine.org/en/latest/tutorials/scripting/resources.html, the tabs look like this:  
  > image
* Add a missing parameterless constructor for the C# example and update the comment to explain why it's necessary.
* Fix C# style: use nextline braces and use `string` rather than `String`.

(Encountered the broken example with a question on Discord.)

### DIFF
--- a/tutorials/scripting/resources.rst
+++ b/tutorials/scripting/resources.rst
@@ -209,6 +209,7 @@ Attach a script to it named ``bot_stats.gd`` (or just create a new script, and t
 
 .. tabs::
   .. code-tab:: gdscript GDScript
+
     extends Resource
 
     @export var health : int
@@ -229,8 +230,9 @@ Attach a script to it named ``bot_stats.gd`` (or just create a new script, and t
         using System;
         using Godot;
 
-        namespace ExampleProject {
-            public class BotStats : Resource
+        namespace ExampleProject
+        {
+            public partial class BotStats : Resource
             {
                 [Export]
                 public int Health { get; set; }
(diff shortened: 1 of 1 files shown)


YuriSizov:
Seems like it should supersede https://github.com/godotengine/godot-docs/pull/6524.

tutorials/scripting/resources.rst
```diff
@@ -229,7 +230,8 @@ Attach a script to it named ``bot_stats.gd`` (or just create a new script, and t
         using System;
         using Godot;
 
-        namespace ExampleProject {
+        namespace ExampleProject
+        {
             public class BotStats : Resource
```
raulsntos, line 235: These classes are missing the `partial` modifier, while you are updating these examples it'd be nice to fix that too.

raulsntos (approved):
I agree that this PR is better than https://github.com/godotengine/godot-docs/pull/6524.

We likely want to backport this to 3.x too (except for the `partial` modifier).

aaronfranke (approved):

tutorials/scripting/resources.rst
31: Thanks, adding. Yeah, I wound up testing this in 3.x (for the Discord convo) and didn't think about this.

YuriSizov:
Thanks!

PR #6526, opened 2022-12-28
Update `in` description

thiagola92:
This description make it sound like can only be used with `if` and `for` keywords:
> When used with the `if` keyword... When used with the `for` keyword...

But i can use without both keywords (with same behavior as using `if` keyword):
```gdscript
var x = 2 in [1,2,3]

print("yes" in "no yes maybe")

return 7 in [1,2,3,4,5,6,7]
```

EDIT: Add `in` inside keyword table.


tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -264,12 +266,8 @@ The following is the list of supported operators and their precedence.
...
 +--------------------------------------------------------------------------------+-------------------------------------------+
-| ``in``                                                                         | When used with the ``if`` keyword it      |
-|                                                                                | checks if a value is within a string,     |
-|                                                                                | list, range, dictionary, or node.         |
-|                                                                                | When used with the ``for`` keyword it is  |
-|                                                                                | used to iterate though the contents of a  |
-|                                                                                | string, list, range, dictionary or node.  |
+| ``in``                                                                         | Content test / Content iterator           |
+|                                                                                | (when used with ``for``)                  |
```
YuriSizov, line 270: I'd go with 

> Inclusion checker (when used with control flow keywords or in a standalone expression).
> Content iterator (when used with the ``for`` keyword).

thiagola92: Thank you, i was struggling to use the right words about the subject

YuriSizov (approved):

mhilbrunner:
Thanks! Merged.

PR #6527, opened 2022-12-28
Score and replay - Grammar

denicolas:
In "[Keeping track of the score](https://docs.godotengine.org/en/stable/getting_started/first_3d_game/08.score_and_replay.html#keeping-track-of-the-score)":
"which is convenient to output text in labels or **using** the print() function"
should be:
"which is convenient to output text in labels or **to use** the print() function"
or:
"which is convenient to output text in labels or **use** the print() function"

### DIFF
--- a/getting_started/first_3d_game/08.score_and_replay.rst
+++ b/getting_started/first_3d_game/08.score_and_replay.rst
@@ -164,7 +164,7 @@ There, we increment the score and update the displayed text.
 
 The second line uses the value of the ``score`` variable to replace the
 placeholder ``%s``. When using this feature, Godot automatically converts values
-to text, which is convenient to output text in labels or using the ``print()``
+to text, which is convenient to output text in labels or to use the ``print()``
 function.
 
 .. seealso::


YuriSizov:
I think the intended read here is "output text in labels or [output text] using the print function". Maybe it's not well phrased and unclear, but your suggested change doesn't work grammatically either.

getting_started/first_3d_game/08.score_and_replay.rst
```diff
@@ -164,7 +164,7 @@ There, we increment the score and update the displayed text.
 
 The second line uses the value of the ``score`` variable to replace the
 placeholder ``%s``. When using this feature, Godot automatically converts values
-to text, which is convenient to output text in labels or using the ``print()``
+to text, which is convenient to output text in labels or to use the ``print()``
```
skyace65, line 167:
```suggestion
to text, which is convenient when outputting text in labels or the ``print()``
```

skyace65 (changes requested):
What Yuri said, the original isn't good but the wording you suggested isn't grammatically correct.

getting_started/first_3d_game/08.score_and_replay.rst
denicolas: I agree that my initial correction doesn't work grammatically. I think your suggested change doesn't work either. What about this:
```suggestion
to text, which is convenient when outputting text in labels or when using the ``print()``
```

YuriSizov: Yeah, that should be fine.

Vennnot: Does this mean this can now be merged?

YuriSizov: No, the suggested changes weren't made, and by now this PR also needs to be rebased.

mhilbrunner:
Salvaged in https://github.com/godotengine/godot-docs/pull/7369. Thanks for contributing!

PR #6531, opened 2022-12-29
Update scripting_first_script.rst

iqre8:
Couldn't get the code formatting proper, but we need to add some math explanation so people understand what's going on here.


YuriSizov:
I don't think this helps general audience to understand what's going on. We try to ease people into coding with Godot in this article, but frontload it with some vector math. That's totally unnecessary.

mhilbrunner:
Thanks for contributing! However, I agree with Yuri on this. The intro should stay as light as can be.

PR #6532, opened 2022-12-30
Fixed incorrect Texture type in example

AndrewPerson:
In the 4th C# code block, the `texture` variable is loaded as a `Texture` resource, when `AddCustomType` requires it to be a `Texture2D`.


YuriSizov:
Thanks! And congratulations on your first merged pull request!

PR #6534, opened 2022-12-30
Update string literal example

thiagola92:
> `"Hello"`, `"Hi"`

Both examples were about creating string using double quotes and none about single quotes.  
Probably the idea was to be one using double quote and the other using single quote.  

> `"Hello"`, `'Hi'`


skyace65 (approved):
Looks good, thanks!

PR #6535, opened 2022-12-30
Fix example on how to use lld

filiperinaldi:
The "use_lld=yes" was removed in Godot (c8479c0d4d). The new parameter which allows using lld is "linker=lld".


mhilbrunner (approved):

mhilbrunner:
Thanks, and congrats on your first contribution to the docs!

PR #6536, opened 2022-12-30
we should state early on that it requires SCons

thadguidry:
Following the documentation step by step, it was missed that SCons is actually needed on the system to build and not only "uses" it.  Much later the docs clarify the requirements, but it's useful to state this differently at the very beginning.  This simple change would have saved me about 15 mins or so.

### DIFF
--- a/development/compiling/introduction_to_the_buildsystem.rst
+++ b/development/compiling/introduction_to_the_buildsystem.rst
@@ -8,7 +8,7 @@ Introduction to the buildsystem
 SCons
 -----
 
-Godot uses `SCons <https://www.scons.org/>`__ to build. We love it, we are
+Godot requires and uses `SCons <https://www.scons.org/>`__ to build. We love it, we are
 not changing it for anything else. We are not even sure other build
 systems are up to the task of building Godot. We constantly get requests
 to move the build system to CMake, or Visual Studio, but this is not


YuriSizov:
I don't understand the significance of this change. Godot is built with SCons, thus to build Godot you need yo have SCons. What wasn't clear here?

thadguidry:
'uses' doesn't signify enough that something is required.  It's a semantic difference and my change will help.  As I said I wasted about 15 mins thinking that the build system would do something with a thing called SCons and only later after clicking through many next links in doc pages did I see on compiling for Windows that it is something required to be already installed on my system and not something that the build would install for me.  I am making the docs better for very first time users such as myself.

YuriSizov:
> I wasted about 15 mins thinking that the build system would do something with a thing called SCons 

I don't understand what you mean by this.

> only later after clicking through many next links in doc pages did I see on compiling for Windows that it is something required to be already installed on my system 

"Compiling for X" is something you should read immediately after reading the introduction paragraph. The article states as much, linking them in the very next section called "Setup". You are not expected to start building Godot before reading the "Setup" section.

Stating that SCons is required in the first sentence doesn't help anyone, because it's not full requirements, and it lacks any guidance on how to obtain SCons. That's what "Compiling for X" articles are for.

thadguidry:
I understand completely.  The **Introduction to the buildsystem** does indeed have a section for **Setup** with all the links for the "Compiling for X" articles... which unfortunately results in a lot of back and forth between a specific one like **Compiling for Windows**, and the **Introduction to the build;system** but I guess it is unavoidable.  But it would have been a much nicer experience to learn about SCons within the **Compiling for Windows** or others...instead of the other way around as it is now.

I'll close this issue.

PR #6541, opened 2022-12-30
clarify Windows Defender Settings

thadguidry:
...for those that might also use updated Windows 11 and not only Windows 10.


mhilbrunner:
That whole thing is also called "Windows Security" for me, not "Windows Defender Settings".

development/compiling/compiling_for_windows.rst
```diff
@@ -105,7 +105,7 @@ The tutorial will assume from now on that you placed the source code in
 
     For Windows Defender, hit the :kbd:`Windows` key, type
     "Windows Defender Settings" then hit :kbd:`Enter`.
```
skyace65, line 107:
```suggestion
    "Windows Security" then hit :kbd:`Enter`.
```

skyace65 (changes requested):
This actually applies to Windows 10 as well, not just 11. If you wouldn't mind I'd also appreciate you making that defender change I suggested since you're updating this section. Also this PR needs to be rebased.

development/compiling/compiling_for_windows.rst
thadguidry: Agree!

thadguidry:
omg I hate Git sometimes, it's even worst in most IDEs.

skyace65:
Did you mean to get rid of the other part you added? Also if you're having trouble with git we do have a tutorial for it in the docs [here](https://docs.godotengine.org/en/stable/community/contributing/pr_workflow.html) it's not everything in git but does go over making changes.

PR #6542, opened 2022-12-31
rework two funcref code examples

matt08-prog:
Reworked two funcref code blocks as discussed in #6444. I don't have a lot of C# experience, however, so this only fixed the gdscript examples.

### DIFF
--- a/tutorials/best_practices/godot_interfaces.rst
+++ b/tutorials/best_practices/godot_interfaces.rst
@@ -472,7 +472,7 @@ accesses:
     @onready var child = $Child
 
     func _ready():
-        child.fn = print_me()
+        child.fn = print_me
         child.my_method()
 
     func print_me():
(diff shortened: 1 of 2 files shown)


Vennnot:
Can this be merged?

mhilbrunner:
No, because I think its incorrect, when referencing a function in Godot 4, you use its name, but without parentheses as thats a function call then. Will see about fixing this.

tutorials/best_practices/godot_interfaces.rst
```diff
@@ -464,15 +464,15 @@ accesses:
...
+            fn.call()
 
     # parent.gd
     extends Node
 
     @onready var child = $Child
 
     func _ready():
-        child.fn = funcref(self, "print_me")
+        child.fn = print_me()
```
mhilbrunner, line 475:
```suggestion
        child.fn = print_me
```

mhilbrunner:
Fixed and merged. Thanks for contributing!

PR #6543, opened 2022-12-31
Add more notes for Compiling on Windows

thadguidry:
- In summary, recommend to users to use the Python full installer.
- These additional notes can save hours for folks that might not fully know Python packaging in general on Windows and all the issues that can arise later (as I did) if alternatives such as **Python from the Microsoft Store** is used, which is by default encouraged now within Visual Studio Code in places, and can cause issues later on with SCons and module paths during setup.  SCons is not automatically added to PATH with our `pip` commands given if **Python from the Microsoft Store** is used, for instance, the discovery of `scons.exe` itself is a nightmare within a deeply nested default script path that looks like: `C:\Users\me\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.10_qbz5n2kfra8p0\LocalCache\local-packages\Python310\Scripts`
- Official Reference and known issues: https://docs.python.org/3/using/windows.html#the-microsoft-store-package

### DIFF
--- a/development/compiling/compiling_for_windows.rst
+++ b/development/compiling/compiling_for_windows.rst
@@ -25,7 +25,9 @@ For compiling under Windows, the following is required:
 - `Python 3.5+ <https://www.python.org/downloads/windows/>`_.
   **Make sure to enable the option to add Python to the ``PATH`` in the installer.**
 
-  .. note:: There is a Microsoft Store package of Python but it is mainly for interactive use and not full development.
+  .. note::
+  
+           There is a Microsoft Store package of Python but it is mainly for interactive use and not full development.
            Because of restrictions on Microsoft Store apps, Python scripts may not have full write access to shared locations such as TEMP and the registry.
            Instead, it will write to a private copy and likely cause issues with SCons later on.
            We recommend installing Python using their full installer at the link above.
--- a/development/compiling/compiling_for_windows.rst
+++ b/development/compiling/compiling_for_windows.rst
@@ -77,7 +77,9 @@ environment variable after installing it, then check again.
 You can do so by running the Python installer again and enabling the option
 to add Python to the ``PATH``.
 
-.. note:: SCons installed via ``pip`` puts an executable ``scons.exe``
+.. note::
+
+          SCons installed via ``pip`` puts an executable ``scons.exe``
           in the Scripts directory of your Python installation, or in a shadow script directory
           if you did a User Install. For Windows to run ``scons`` as a command, you'll need this in your search path.
           During installation with ``pip`` you might have missed a warning like this:
--- a/development/compiling/compiling_for_windows.rst
+++ b/development/compiling/compiling_for_windows.rst
@@ -24,6 +24,12 @@ For compiling under Windows, the following is required:
   **Important:** When using MinGW to compile the ``master`` branch, you need GCC 9 or later.
 - `Python 3.5+ <https://www.python.org/downloads/windows/>`_.
   **Make sure to enable the option to add Python to the ``PATH`` in the installer.**
+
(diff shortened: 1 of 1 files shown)


YuriSizov:
> These additional notes can save hours for folks that might not fully know Python packaging in general on Windows and all the issues that can arise later (as I did) if alternatives such as **Python from the Microsoft Store** is used

Won't say it doesn't have issues, namely git bash doesn't support it (but that's an issue with git bash, not with this installation of Python), but I've been using it the whole time with no visible problems.

development/compiling/compiling_for_windows.rst
```diff
@@ -24,6 +24,12 @@ For compiling under Windows, the following is required:
   **Important:** When using MinGW to compile the ``master`` branch, you need GCC 9 or later.
 - `Python 3.5+ <https://www.python.org/downloads/windows/>`_.
   **Make sure to enable the option to add Python to the ``PATH`` in the installer.**
+
+  .. note:: There is a Microsoft Store package of Python but it is mainly for interactive use and not full development.
+           Because of restrictions on Microsoft Store apps, Python scripts may not have full write access to shared locations such as TEMP and the registry.
```
YuriSizov, line 29:
```suggestion
  .. note::
  
           There is a Microsoft Store package of Python but it is mainly for interactive use and not full development.
           Because of restrictions on Microsoft Store apps, Python scripts may not have full write access to shared locations such as TEMP and the registry.
```

Though this can work, an empty line is required by convention.

development/compiling/compiling_for_windows.rst
```diff
@@ -69,6 +75,18 @@ environment variable after installing it, then check again.
 You can do so by running the Python installer again and enabling the option
 to add Python to the ``PATH``.
 
+.. note:: SCons installed via ``pip`` puts an executable ``scons.exe``
+          in the Scripts directory of your Python installation, or in a shadow script directory
```
YuriSizov, line 79:
```suggestion
.. note::

          SCons installed via ``pip`` puts an executable ``scons.exe``
          in the Scripts directory of your Python installation, or in a shadow script directory
```

Same here.

development/compiling/compiling_for_windows.rst
```diff
@@ -69,6 +75,18 @@ environment variable after installing it, then check again.
...
+          if you did a User Install. For Windows to run ``scons`` as a command, you'll need this in your search path.
+          During installation with ``pip`` you might have missed a warning like this:
+          .. code-block:: pycon
+             WARNING: The scripts scons-configure-cache.exe, scons.exe and sconsign.exe
+             are installed in 'C:\Users\me\AppData\Roaming\Python\Python310\Scripts'
+             which is not on PATH.
+             Consider adding this directory to PATH or, if you prefer to suppress this warning,
+             use --no-warn-script-location.
+
```
YuriSizov, line 88: Extra line here.

thadguidry: sorry, which line number is extra?

YuriSizov: You have two empty lines there. I think one is enough.

mhilbrunner:
Thanks for submitting these additions! I'd also be interested in which problems you ran into with the Windows Store Python installation. I'm fine with adding a warning if warranted, but at first glance a `%TEMP%` copy and not being able to interact with the registry should hopefully be fine for SCons builds.

thadguidry:
@mhilbrunner After much investigation, and doing a clean reproduction on a VM in Azure, it seems the issue was an extra path statement I had missed in my local environment, which caused the issue.  Indeed the Windows Store Python seems to work fine and SCons has no complaints.

Let's close this issue.

PR #6545, opened 2022-12-31
Add a Nix flake as a reproducible build system

MatthewCroughan:
This adds a Nix flake with a `default` output named "godot-docs-html", that can be built with `nix build`. The main outputs that can be built are:

This will produce a result:

```
❯ nix build
❯ ls result/
404.html     _downloads       objects.inv     _static
about        genindex.html    robots.txt      tutorials
classes      getting_started  search.html
community    _images          searchindex.js
development  index.html       _sources
```

Nix builds are deterministic, and reproducible long-term. The requirements.txt file at the root of this repository doesn't provide enough information to be reproducible, since it doesn't store the sha256 of the dependency when it was originally built, only the version number. This doesn't allow us to verify that it hasn't changed in the meantime. Thankfully the mach-nix library (which you can see in the flake inputs) allowed me to reproduce the godot-docs, since it stores a frozen package database, which is similar to a lockfile.

If we were to update godot-docs to a more recent version of the packages, then it may be possible to reproduce it without the [mach-nix ](https://github.com/DavHau/mach-nix) library, or still using mach-nix and `(builtins.readfile ./requirements.txt)` which would allow you to keep the requirements.txt as a source of truth. It would be better to move to a system like [poetry](https://python-poetry.org/) that creates lockfiles, or to move away from python and use something more reproducible and lockfile based like [mdbook ](https://rust-lang.github.io/mdBook/)(rust) instead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,6 @@ logo.h
 # Output of list-unused-images.sh tool
 tmp-unused-images
 tmp-unused-images-history
+
+# nix build results
+*result*
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,132 @@
+{
+  "nodes": {
+    "flake-parts": {
+      "inputs": {
+        "nixpkgs-lib": "nixpkgs-lib"
+      },
+      "locked": {
+        "lastModified": 1672152762,
+        "narHash": "sha256-U8iWWHgabN07zfbgedogMVWrEP1Zywyf3Yx3OYHSSgE=",
+        "owner": "hercules-ci",
+        "repo": "flake-parts",
+        "rev": "19e0f88324d90509141e192664ded98bb88ef9b2",
+        "type": "github"
+      },
+      "original": {
+        "owner": "hercules-ci",
+        "repo": "flake-parts",
+        "type": "github"
+      }
+    },
+    "flake-utils": {
+      "locked": {
+        "lastModified": 1642700792,
+        "narHash": "sha256-XqHrk7hFb+zBvRg6Ghl+AZDq03ov6OshJLiSWOoX5es=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "846b2ae0fc4cc943637d3d1def4454213e203cba",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "mach-nix": {
+      "inputs": {
+        "flake-utils": "flake-utils",
+        "nixpkgs": "nixpkgs",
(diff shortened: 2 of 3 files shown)


YuriSizov:
Why would we need this though? For local builds it doesn't matter that much, beyond the pinned versions, and for actual builds on RTD — they do the building, we have no control over how they build it.

MatthewCroughan:
@YuriSizov Long term reproducibility. Just making the PR, if you don't want it then you do not need to merge it.

YuriSizov:
I hear what you say and I understand how lock files work with package managers, but what benefits does long term reproducibility give us here? And if we do need it, why Nix flake specifically besides you working on it? I don't think it integrates with Read the Docs, does it?

MatthewCroughan:
What drawbacks does long term reproducibility give you? As I say, I'm just making the PR, there's no requirement that it be merged. Some people care about these things in principle, whilst some do not.

YuriSizov:
Of course there is no requirement for it to be merged, but I would still like to have some rationale that can be discussed. "Why not" is not a good argument, and I've explained our situation where we don't control production builds where locking dependencies would actually be important.

It's great that you're interested in contributing but we prefer not to have contributors do work that ends up being "nice to have", but not actually needed. It's better to put effort into things that are definitely need improving. 🙂

MatthewCroughan:
When someone can't use `pip` to gather the dependencies, because it produces a different result tomorrow than it does today, you will still be able to build this Flake, because it is a reproducible Nix derivation. The proof won't exist until you see this occur in practice, and it might be what you need to understand. I'm not going to spend a lot of time trying to convince you, I've already done that in vain, a lot, regarding reproducibility. If you would like to learn more about Nix, then you can [read the thesis](https://edolstra.github.io/pubs/phd-thesis.pdf)

mhilbrunner:
Thanks for taking the time to contribute to Godot's docs!

At this point, we're not looking to replace the existing SCons build system or to introduce additional build systems along side Scons. If such a move were to happen, it should likely go through the [Godot Proposal process](https://github.com/godotengine/godot-proposals) and clearly outline the benefits of the move and analyse strenght and weaknesses of the new system vs. the status quo.

Of note:
- The docs team is fairly tight on resources and needs to prioritize; currently, those prorities are getting the docs ready for Godot 4.0 RC (updating and completing the class reference, overhauling the old tutorials etc.), and not introducing additional complexity
- The benefits of this move aren't immediately clear via the initial PR and the above discussion, especially considering that we deploy via external hosting/RTD and part of the infrastructure is out of our control anyway

That said, again, thanks for contributing, Nix for sure looks interesting and I'm hoping to have some time soon to look at it more in-depth :)

PR #6547, opened 2023-01-01
fix(01.game_setup.rst): Extents->Size adjustments and new screenshots

nylki:
This MR fixes some terminology changes in the "Setting up the game area" section of "Your first 3D game":

1. name of `CubeMap` -> `BoxMesh`. For consistency I replaced following mentions of "cube" with "box".

2. use correct 3D entity type names: `CollisionShape` -> `CollisionShape3D`  etc.

3. correct usage of `Size` of `CollisionShape3D`, which was `Extents` previously. I removed the explanatory box about difference of Extents vs. Size since it doesn't seem to be necessary anymore.

4. updated relevant illustrations/screenshots


YuriSizov:
This is probably already covered by https://github.com/godotengine/godot-docs/pull/6243, but we can keep this one open as well in case there is something missing from it.

nylki:
@Piralein Thank you for the heads up! I rebased and updated the last commited with the changes you suggested (ref links, webp).

getting_started/first_3d_game/01.game_setup.rst
```diff
@@ -74,33 +74,25 @@ reliable to block even fast-moving objects.
...
 
-.. image:: img/01.game_setup/09.box_extents.webp
+.. image:: img/01.game_setup/09.box_size.webp
 
-.. note::
-
-    In 3D, translation and size units are in meters. The box's total size is
-    twice its extents: ``60`` by ``60`` meters on the ground plane and ``2``
-    units tall. The ground plane is defined by the X and Z axes, while the Y
-    axis represents the height.
```
skyace65, line 88: What's the purpose of removing this note?

getting_started/first_3d_game/01.game_setup.rst
```diff
@@ -74,33 +74,25 @@ reliable to block even fast-moving objects.
...
 
 Collision shapes are invisible. We need to add a visual floor that goes along
 with it. Select the ``Ground`` node and add a :ref:`MeshInstance <class_MeshInstance3D>` as its child.
```
skyace65, line 85:
```suggestion
with it. Select the ``Ground`` node and add a :ref:`MeshInstance3D <class_MeshInstance3D>` as its child.
```

If you wouldn't mind, this should be changed as well

skyace65 (commented):
Changes look good. Just had one question. And if you wouldn't mind there's something else that should be updated.

getting_started/first_3d_game/01.game_setup.rst
nylki: In the previous docs you had to set `extents` in this step of the tutorial instead of size. Therefore both mismatched (size being 30x30, extends being 60) which I suppose warranted this explanatory note to explain the difference.

Since users don't have to deal with extents vs. size at this point, this explanation doesn't not make sense anymore (atleast in this tutorial, not sure if extents are used or explained somewhere else).

skyace65: Thanks!

skyace65 (approved):

skyace65:
Thanks! And congrats on your first merged PR! Just so you know we like to keep everything in 1 commit. I've squashed and merged this time but keep it in mind for future reference. If you don't know how to do that we have a guide on [this page](https://docs.godotengine.org/en/latest/contributing/workflow/pr_workflow.html).

Issue #6539, opened 2022-12-30
Error in Tutorial "Your first 2D game", "coding the player"


jcreek:
**Your Godot version:**
4.0.beta10.mono

**Issue description:**
Following the guide exactly with C# selected gives a `_Process` function of:

```c#
public override void _Process(double delta)
	{
		var velocity = Vector2.Zero; // The player's movement vector.

		if (Input.IsActionPressed("move_right"))
		{
			velocity.x += 1;
		}

		if (Input.IsActionPressed("move_left"))
		{
			velocity.x -= 1;
		}

		if (Input.IsActionPressed("move_down"))
		{
			velocity.y += 1;
		}

		if (Input.IsActionPressed("move_up"))
		{
			velocity.y -= 1;
		}

		var animatedSprite2D = GetNode<AnimatedSprite2D>("AnimatedSprite2D");

		if (velocity.Length() > 0)
		{
			velocity = velocity.Normalized() * Speed;
			animatedSprite2D.Play();
		}
		else
		{
			animatedSprite2D.Stop();
		}
		
		Position += velocity * delta;
		Position = new Vector2(
			x: Mathf.Clamp(Position.x, 0, ScreenSize.x),
			y: Mathf.Clamp(Position.y, 0, ScreenSize.y)
		);
	}
```

This throws an error on the line `Position += velocity * delta;` as `CS0019: Operator '*' cannot be applied to operands of type 'Vector2' and 'double'` - given I've followed the tutorial exactly this seems likely to be either an issue with the documentation or with the beta of Godot 4.0, I'm an experienced C# dev but have never done game development before, so it's also possible I've misunderstood something, but after double-checking I don't think so. 

**URL to the documentation page:**
https://docs.godotengine.org/en/latest/getting_started/first_2d_game/03.coding_the_player.html

raulsntos:
The example is outdated, the `delta` parameter is a `double` in 4.0, in order to multiply a `Vector2` it needs to be converted to `float` first. There are likely many other outdated examples in the documentation.

This should work:

```csharp
Position += velocity * (float)delta;
```

jcreek:
Thanks @raulsntos - I am delighted to confirm that that fixes it! Is there an established process here for creating PRs to solve issues? I'm happy to create the relevant PRs, and raise further issues as I follow the tutorial through :)

raulsntos:
[...]

PR #6548, opened 2023-01-01
fix(#6539) Update C# documentation for Godot 4

jcreek:


raulsntos (commented):
Could you also update the `_Process` method to use `double` instead of `float` for the delta parameter?

Also, C# classes in 4.0 need to be partial so we may want to fix that in this page while we're at it.

raulsntos (approved):
Thanks!

mhilbrunner (approved):

mhilbrunner:
Merged! Thanks for contributing to the docs for Godot Engine :)

PR #6550, opened 2023-01-02
Update navigationserver2d to use a better tutorial

yurivwarmerdam:
(first time attempting to contribute to an open source project like this, so I apologize in advance for any faux-pases I will inevitably commit) 
I have modified the existing asset that uses the deprecated Navigation node so the project now uses Navigaiton2DServer. This makes the tutorial more relevant for this documentation page. 
My new asset is currently pending, so here a link to the github repo that will be referenced in the asset: https://github.com/yurivwarmerdam/navigation 
Obviously this change should only be merged once my asset has been approved.

### DIFF
--- a/classes/class_navigationserver2d.rst
+++ b/classes/class_navigationserver2d.rst
@@ -40,7 +40,7 @@ This server keeps tracks of any call and executes them during the sync phase. Th
 Tutorials
 ---------
 
-- `2D Navigation Demo <https://godotengine.org/asset-library/asset/117>`__
+- `2D Navigation Server Demo <https://godotengine.org/asset-library/asset/6799>`__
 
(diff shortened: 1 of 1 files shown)


Calinou:

There are two issues before this can be merged:

- Class reference tutorial links should only link to official demos, so the demo must be submitted to https://github.com/godotengine/godot-demo-projects' `4.0-dev` branch first.
- Contributions to the class reference should be sent on the main Godot repository, not here. This is because the class reference in this repository is generated automatically based on the XML files in the main repository. See [Contribute to the class reference](https://docs.godotengine.org/en/latest/community/contributing/updating_the_class_reference.html) in the documentation :slightly_smiling_face:

PR #6551, opened 2023-01-02
Update C# code example

c-schembri:
Use the `Root` property instead of the deprecated `GetRoot` method. `PropagateNotification` accepts an `int` and not a `long`, and so `NotificationWmCloseRequest` needs to be cast to `int`.


raulsntos (approved):

mhilbrunner (approved):

mhilbrunner:
Thanks, merged!

PR #6552, opened 2023-01-02
Adding info on how to use Color constants

ego-edesse:
I believe it is worth mentionning on the doc how to access Color constants with C# because it is slighty different from GDScript but not as expected. Besides that, there isn't any info on how to access constants here yet


YuriSizov:
https://github.com/godotengine/godot/tree/3.5/doc/classes/Color.xml

PR #6553, opened 2023-01-03
Clarify NavigationServer created regions

smix8:
Clarify that NavigationServer created navigation regions do not join a navigation map automatically like a NavigationRegion node would. The server API created regions need a manual map assignment with region_set_map().

Related to this issue [here](https://github.com/godotengine/godot/issues/70855).


mhilbrunner (approved):

mhilbrunner:
Thanks!

PR #6555, opened 2023-01-03
First 2d game docs - minor typo fix

bhansconnect:
Super minor typo fixed for first 2d game tutorial

### DIFF
--- a/getting_started/first_2d_game/index.rst
+++ b/getting_started/first_2d_game/index.rst
@@ -34,7 +34,7 @@ recommend you to start with this one, though.
 
 **Why start with 2D?**
 
-3D games are much more complex than 2D ones. It would be best if you stuck to 2D
+3D games are much more complex than 2D ones. It would be best if you stick to 2D
 until you understood the game development process and how to use Godot well.
 
 You can find a completed version of this project at this location:
--- a/getting_started/first_2d_game/index.rst
+++ b/getting_started/first_2d_game/index.rst
@@ -35,7 +35,7 @@ recommend you to start with this one, though.
 **Why start with 2D?**
 
 3D games are much more complex than 2D ones. It would be best if you stick to 2D
-until you understood the game development process and how to use Godot well.
+until you understand the game development process and how to use Godot well.
 
 You can find a completed version of this project at this location:
 
--- a/getting_started/first_2d_game/index.rst
+++ b/getting_started/first_2d_game/index.rst
@@ -34,8 +34,9 @@ recommend you to start with this one, though.
 
 **Why start with 2D?**
 
-3D games are much more complex than 2D ones. It would be best if you stick to 2D
-until you understand the game development process and how to use Godot well.
+If you are new to game development or unfamiliar with Godot, we recommend
+starting with 2D games. This will allow you to become comfortable with both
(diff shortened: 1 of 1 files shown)


YuriSizov:
I think this was correct as is. In the "would + if verb" case, the verb is normally in the past tense.

bhansconnect:
So I think they have two slightly different implied meanings:

"It would be best if you stuck to 2D" really means "It would have been best if you had stuck to 2D". This is hypothetically looking back and suggesting an alternative path.

"It would be best if you stick to 2D" really means "It would be best if you continue to stick to 2D". This is suggesting a current course of action until some criteria is met. So more current and future focused.

I believe they are both grammatically correct, but I think that `stuck` ends up reading worse in this context.
Given the use here is to suggest what a user should currently do until they "understand the game development process and how to use Godot well", I think the current and future looking wording makes more sense.

That said, just realized that also `understood` -> `understand` would be needed for the sentence to fully change here.

bhansconnect:
I decided to ask chatgpt about this and it gave a surprisingly good explanation, so I thought I would add it for extra context as well. I think it explains this better than I did:

-----

It depends on the context and the audience of the tutorial. If the tutorial is written for beginners who are just starting to learn about game development and Godot, it might be more appropriate to use the present tense. For example:

"3D games are much more complex than 2D ones. It would be best if you stick to 2D until you understand the game development process and how to use Godot well."

This wording conveys the idea that the reader should continue to focus on 2D games until they have a good foundation of knowledge and skills.

On the other hand, if the tutorial is written for more experienced game developers who have already learned about game development and Godot, it might be more appropriate to use the past tense. For example:

"3D games are much more complex than 2D ones. It would have been best if you stuck to 2D until you understood the game development process and how to use Godot well."

This wording conveys the idea that the reader should have focused on 2D games before moving on to 3D games.

Ultimately, the choice of tense will depend on the intended audience and the overall tone and style of the tutorial.

YuriSizov:
ChapGPT is not a verifiable source and will confidently tell you a completely false piece of information. It's fine for brainstorming, but it's not a tool that can be referenced like that.

bhansconnect:
I understand that. Maybe I shouldn't have added it and should have just left it to own commentary. I just thought it managed to explain more clearly than I did. The original comment is still completely my own. Just thought a clearer explanation might also be helpful.

mhilbrunner:
To me, present tense reads better. Overall, I'm also not really a fan of the sentence, reads (a tiny) bit condescending. IMO, its best to just say "we recommend to start with 2D if you're either new to game development or Godot Engine until you're comfortable with both, because 3D games are usually more complicated" (or something similar). Makes it more clear its a recommendation (that can be ignored at the readers own risk if they're sure of their skill).

bhansconnect:
A restructuring from the last comment, how does this sound? "If you are new to game development or unfamiliar with Godot, we recommend starting with 2D games. This will allow you to become comfortable with both before tackling 3D games, which tend to be more complicated."

mhilbrunner:
Yeah, that reads way better to me!

mhilbrunner (approved):

mhilbrunner:
Merged, thanks for taking the time to improve this, even if its a single sentence, reads better now ;)

PR #6562, opened 2023-01-06
Update physics_introduction.rst

fractionofadot:
Judging by forum questions and answers, many people do not understand that rigid bodies do not detect collisions with Area2D nodes, but desire that functionality. It would be helpful to include a note in this introduction clarifying that point and providing one way to accomplish that goal.

### DIFF
--- a/tutorials/physics/physics_introduction.rst
+++ b/tutorials/physics/physics_introduction.rst
@@ -320,6 +320,13 @@ Contact monitoring via signals can be enabled via the :ref:`contact_monitor <cla
 property. See :ref:`RigidBody2D <class_RigidBody2D>` for the list of available
 signals.
 
+.. note:: Rigid bodies do not detect collisions with ``Area2D`` nodes. If you need a rigid body to detect collisions with ``Area2D``, consider using the ``body`` argument in the ``Area2D``  ``_on_body_entered(body)`` signal to emit a signal from the rigid body: 
+.. tabs::
+ .. code-tab:: gdscript GDScript
+ 
+    func _on_Area2DName_body_entered(body):
+        body.emit_signal("body_entered", self)
+
 CharacterBody2D
 ---------------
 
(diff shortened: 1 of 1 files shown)


tutorials/physics/physics_introduction.rst
```diff
@@ -320,6 +320,13 @@ Contact monitoring via signals can be enabled via the :ref:`contact_monitor <cla
 property. See :ref:`RigidBody2D <class_RigidBody2D>` for the list of available
 signals.
 
+.. note:: Rigid bodies do not detect collisions with ``Area2D`` nodes. If you need a rigid body to detect collisions with ``Area2D``, consider using the ``body`` argument in the ``Area2D``  ``_on_body_entered(body)`` signal to emit a signal from the rigid body: 
+.. tabs::
```
YuriSizov, line 324: Not sure if this a good suggestion to give in the docs, but either way formatting needs to be improved like so:

```suggestion
.. note::

    Rigid bodies do not detect collisions with ``Area2D`` nodes. If you need a rigid body
    to detect collisions with ``Area2D``, consider using the ``body`` argument in the ``Area2D``
    ``_on_body_entered(body)`` signal to emit a signal from the rigid body: 

.. tabs::
```

cbscribe:
Areas don't collide at all.  It's not correct to say
> Rigid bodies do not detect collisions with ``Area2D`` nodes

Because nothing "detects collisions" with areas - rigid bodies are not anything special here. An object can overlap an area or not. The explanation of how areas work is also in this document. If anything, the "many people" you're referring to are exhibiting a lack of understanding of areas, rather than of rigid bodies.

This document already covers a lot. Adding more just makes it more likely that those who need to read it won't.

Perhaps it would be better to change the existing rigid body text to refer to "bodies" rather than "contacts" if we're concerned with clarity on that front.

skyace65:
Closing due to the lack of a response to cbscribe. I also agree with him, the change isn't correct and I don't think it's worth making the document longer.

PR #6563, opened 2023-01-06
Improve Shader preprocessor documentation

Calinou:
- Add an introduction on why a shader preprocessor can be useful (and how it compares to the preprocessor found in most GLSL codebases).
- Document visual shaders not supporting preprocessing.
- Document `#elif` and `defined()`.
- Add section on performance caveats of `if` statements versus `#if` preprocessor. This is also now referenced on the Shading language page.
- Improve documentation for shader includes.
- Add a shader preprocessor section to the Shader style guide.

### DIFF
--- a/tutorials/shaders/shader_reference/shader_preprocessor.rst
+++ b/tutorials/shaders/shader_reference/shader_preprocessor.rst
@@ -1,24 +1,83 @@
 .. _doc_shader_preprocessor:
 
-Shader Preprocessor
+Shader preprocessor
 ===================
 
-The shader preprocessor is an optional step before shader compilation of text shaders in Godot.
-If you don't need it, you may ignore it, but it contains some useful macros which may speed up your productivity.
+Why use a shader preprocessor?
+------------------------------
+
+In programming languages, a *preprocessor* allows changing the code before the
+compiler reads it. Unlike the compiler, the preprocessor does not care about
+whether the syntax of the preprocessed code is valid. The preprocessor always
+performs what the *directives* tell it to do. A directive is a statement
+starting with a hash symbol (``#``). It is not a *keyword* of the shader
+language (such as ``if`` or ``for``), but a special kind of token within the
+language.
+
+From Godot 4.0 onwards, you can use a shader preprocessor within text-based
+shaders. The syntax is similar to what most GLSL shader compilers support
+(which in turn is similar to the C/C++ preprocessor).
+
+.. note::
+
+    The shader preprocessor is not available in :ref:`visual shaders <doc_visual_shaders>`.
+    If you need to introduce preprocessor statements to a visual shader, you can
+    convert it to a text-based shader using the **Convert to Shader** option in
+    the VisualShader inspector resource dropdown. This conversion is a one-way
+    operation; text shaders cannot be converted back to visual shaders.
+
+Directives
+----------
+
+General syntax
+^^^^^^^^^^^^^^
+
+- Preprocessor directives do not use brackets (``{}``), but can use parentheses.
+- Preprocessor directives **never** end with semicolons.
+- Preprocessor directives can span several lines by ending each line with a
+  blackslash (``\``). The first line break *not* featuring a backslash will end
+  the preprocessor statement.
 
 #define
 ^^^^^^^
-\ **Syntax:** `#define identifier <replacement_code>`.
+\ **Syntax:** ``#define <identifier> [replacement_code]``.
 
-Defines the identifier after that directive as a macro, and replaces all successive occurrence of it with the replacement code given in the shader.
-If the replacement code is not defined, it may only be used within `#ifdef` or `#ifndef` directives.
+Defines the identifier after that directive as a macro, and replaces all
+successive occurrences of it with the replacement code given in the shader.
+Replacement is performed on a "whole words" basis, which means no replacement is
+performed if the string is part of another string (without any spaces separating
+it).
+
+Defines with replacements may also have one or more *arguments*, which can then
+be passed when referencing the define (similar to a function call).
+
+If the replacement code is not defined, the identifier may only be used with
+``#ifdef` or ``#ifndef`` directives.
+
+Compared to constants (``const CONSTANT = value;``), ``#define``s can be used
+anywhere within the shader. ``#define``s can also be used to insert arbitrary
+shader code at any location, while constants can't do that.
 
 .. code-block:: glsl
 
     shader_type spatial;
 
+    // Notice the lack of semicolon at the end of the line, as the replacement text
+    // shouldn't insert a semicolon on its own.
     #define USE_MY_COLOR
     #define MY_COLOR vec3(1, 0, 0)
+
+    // Replacement with arguments.
+    // All arguments are required (no default values can be provided).
+    #define BRIGHTEN_COLOR(r, g, b) vec3(r + 0.5, g + 0.5, b + 0.5)
+
+    // Multiline replacement using backslashes for continuation:
+    #define SAMPLE(param1, param2, param3, param4) long_function_call( \
+            param1, \
+            param2, \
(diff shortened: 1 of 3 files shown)


mhilbrunner:
cc @clayjohn

tutorials/shaders/shader_reference/shader_preprocessor.rst
```diff
@@ -1,56 +1,256 @@
...
+Replacement is performed on a "whole words" basis, which means no replacement is
+performed if the string is part of another string (without any spaces separating
+it).
+
+Defines with replacements may also have one or more *arguments*, which can then
+be passed when referencing the define (similar to a function call).
 
-Defines the identifier after that directive as a macro, and replaces all successive occurrence of it with the replacement code given in the shader.
-If the replacement code is not defined, it may only be used within `#ifdef` or `#ifndef` directives.
+If the replacement code is not defined, it may only be used within ``#ifdef` or
```
clayjohn, line 54:
```suggestion
If the replacement code is not defined, the identifier may only be used with ``#ifdef` or
```

tutorials/shaders/shader_reference/shader_preprocessor.rst
```diff
@@ -1,56 +1,256 @@
...
 #if
 ^^^
-\ **Syntax:** `#if condition`.
 
-The `#if` directive checks the condition and if it evaluates to a non-zero value, the code block is included, otherwise it is skipped.
-In order to evaluate, the condition must be an expression giving a simple floating-point, integer or boolean result. There may be multiple condition blocks connected by `||` or `&&` operators.
-It may be continued by a `#else` block, but must be ended with the `#endif` directive.
+**Syntax:** ``#if <condition>``
+
+The ``#if`` directive checks the ``condition`` passed. If it evaluates to a non-zero
```
clayjohn, line 121:
```suggestion
The ``#if`` directive checks whether the ``condition`` passed. If it evaluates to a non-zero
```

tutorials/shaders/shader_reference/shader_preprocessor.rst
```diff
@@ -1,56 +1,256 @@
...
+`Modern GPUs are quite effective at performing "static" branching <https://medium.com/@jasonbooth_86226/branching-on-a-gpu-18bfc83694f2>`__.
+"Static" branching refers to ``if`` statements where *all* pixels/vertices
+evaluate to ``true`` or ``false`` in a given frame. However, high amounts of
+:abbr:`VGPR (Vector General-Purpose Register)`s (which can be caused by having
+too many branches) can still slow down shader execution significantly.
+
+The downside when using ``#if``s is that it requires compiling more shader
+versions. This can also have a negative impact on performance and cause more
+stuttering due to shader compilation (if not cached). You need to find a balance
+between using ``#if`` and run-time ``if`` statements.
```
clayjohn, line 212: This isn't true for user-space shaders. Regardless of whether ``#if`` conditions are used, each user space shader produces only one set of code which is sent to the renderer.

Our internal shaders have runtime switches to enable/disable defines, which is what causes them to have multiple versions.

tutorials/shaders/shader_reference/shader_preprocessor.rst
```diff
@@ -1,56 +1,256 @@
...
+The :ref:`shading language <doc_shading_language>` supports run-time ``if`` statements:
+
+.. code-block:: glsl
+
+    const bool USE_LIGHT = true;
+
+    if (USE_LIGHT) {
+        // This part is included in the compiled shader, and always run.
+    } else {
+        // This part is included in the compiled shader, but never run.
```
clayjohn, line 183: This isn't true, since USE_LIGHT is a compile time constant this code will get compiled out. Perhaps the example would be better using a uniform?

Also note, your discussion of performance below is only really applicable if USE_LIGHT is a uniform or a run time variable

tutorials/shaders/shader_reference/shading_language.rst
```diff
@@ -516,6 +516,18 @@ Keep in mind that, in modern GPUs, an infinite loop can exist and can freeze
...
+    Branching in shaders can be expensive, especially if every pixel/vertex in a
+    given frame evaluates to different branches. GPUs are massively parallel
+    processors; they're fastest when the data they work with is *predictable*.
+
+    "Static" branching (where every pixel/vertex evaluates to ``true`` or
+    ``false`` in a frame) is less expensive, but it can still have a noticeable
+    cost if used too often. Depending on context, consider using the
+    :ref:`doc_shader_preprocessor`'s ``#if`` directive to reduce the number of
+    branches that are compiled in the final shader.
+
```
clayjohn, line 530: I'm not sure this note makes sense here. It also isn't quite accurate. I would just remove it altogether

clayjohn (commented):
Great work! I left a few comments on style/content

tutorials/shaders/shader_reference/shader_preprocessor.rst
Calinou: I've changed it to use a uniform.

tutorials/shaders/shader_reference/shader_preprocessor.rst
```diff
@@ -1,56 +1,251 @@
...
+
+In programming languages, a *preprocessor* allows changing the code before the
+compiler reads it. Unlike the compiler, the preprocessor does not care about
+whether the syntax of the preprocessed code is valid. The preprocessor always
+performs what the *directives* tell it to do. A directive is a statement
+starting with a hash symbol (``#``). It is not a *keyword* of the shader
+language (such as ``if`` or ``for``), but a special kind of token within the
+language.
+
+From Godot 4.0 onwards, you can now use a shader preprocessor within text-based
```
clayjohn, line 17:
```suggestion
From Godot 4.0 onwards, you can use a shader preprocessor within text-based
```

Looks like you missed this change from my previous suggestion

tutorials/shaders/shader_reference/shader_preprocessor.rst
```diff
@@ -1,56 +1,251 @@
...
+
+    #if defined(USE_LIGHT)
+    // This part is included in the compiled shader, and always run.
+    #else
+    // This part is *not* included in the compiled shader (and therefore never run).
+    #endif
+
+However, the ``#if`` variant can be faster in certain scenarios. This is because
+all run-time branches in a shader are still compiled and take up register space,
+even if they are never run in practice.
```
clayjohn, line 201:
```suggestion
However, the ``#if`` variant can be faster in certain scenarios. This is because
all run-time branches in a shader are still compiled and variables within 
those branches may still take up register space, even if they are never run in
practice.
```

tutorials/shaders/shader_reference/shader_preprocessor.rst
```diff
@@ -1,56 +1,251 @@
...
+
+`Modern GPUs are quite effective at performing "static" branching <https://medium.com/@jasonbooth_86226/branching-on-a-gpu-18bfc83694f2>`__.
+"Static" branching refers to ``if`` statements where *all* pixels/vertices
+evaluate to ``true`` or ``false`` in a given frame. However, high amounts of
+:abbr:`VGPR (Vector General-Purpose Register)`s (which can be caused by having
+too many branches) can still slow down shader execution significantly.
```
clayjohn, line 207:
```suggestion
evaluate to the same result in a given shader invocation. However, high amounts of
:abbr:`VGPR (Vector General-Purpose Register)`s (which can be caused by having
too many branches) can still slow down shader execution significantly.
```

This might be too much information for the manual. It is incorrect to say "frame", but saying wave or warp is likely meaningless to the type of reader that this document is geared towards. I went with shader invocation which is closer to true

clayjohn (approved):
Looks good to me!

geekley:
related: #10035

PR #6569, opened 2023-01-08
fix imulExtended description

jmdejong:
I did not test the actual function but I assume that the name is descriptive and that the gdshader language would follow glsl


mhilbrunner (approved):

mhilbrunner:
Thank you, merged!

PR #6571, opened 2023-01-09
Use 32 bits when converting back from bytes in Using compute shaders

### LINKED ISSUE #6570: Compute tutorial uses 64 instead of 32 bit when converting back from bytes to floats.
**Your Godot version:** 4.0.beta10

**Issue description:** Compute code works but displays wrong output numbers.

**URL to the documentation page:** https://docs.godotengine.org/en/latest/tutorials/shaders/compute_shaders.html

winston-yallow:
This was probably missed when the tutorial was changed to use floats instead of doubles (https://github.com/godotengine/godot-docs/pull/6468)

Fixes #6570


clayjohn (approved):

Calinou (approved):
Thanks!

Issue #6533, opened 2022-12-30
"Organization of the documentation" section has wrong navigable links


thadguidry:
**Your Godot version:**  latest

**Issue description:** _Organization of the documentation_ paragraphs have wrong navigable links, such as [Engine Development](https://docs.godotengine.org/en/latest/index.html#sec-devel)

**URL to the documentation page:**  https://docs.godotengine.org/en/latest/about/introduction.html#organization-of-the-documentation

If you know how to fix the issue you are reporting please
consider opening a pull request. We provide a tutorial on
using git here: https://docs.godotengine.org/en/stable/community/contributing/pr_workflow.html,
writing documentation at https://docs.godotengine.org/en/stable/community/contributing/docs_writing_guidelines.html
and contributing to the class reference here: https://docs.godotengine.org/en/stable/community/contributing/updating_the_class_reference.html

thadguidry:
@YuriSizov Possibly an easy fix for you?

YuriSizov:
It's a regression from https://github.com/godotengine/godot-docs/pull/6510. I guess we should just remove the links and format the names as bold instead.

PR #6572, opened 2023-01-09
Fix links to sections in the Introduction

YuriSizov:
Closes https://github.com/godotengine/godot-docs/issues/6533. Supersedes https://github.com/godotengine/godot-docs/pull/6340 since I was in the neighborhood.

The text is a bit more blind now without accent-colored links, but we can think about a better layout for this section later.

I also fixed some styling related to HTML lists. There's been some ambiguity due to Sphinx building pages differently on production and locally (not sure why, e.g. it uses `<section>` in production, but `<div class="section">` locally, which is the cause of the ambiguity here). It's not major, but makes comparing changes harder.

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -316,14 +316,21 @@ article ol,
 .wy-plain-list-disc,
 .wy-plain-list-decimal,
 .rst-content ol.arabic,
-.rst-content .section ul,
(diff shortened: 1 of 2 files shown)


mhilbrunner (approved):

mhilbrunner:
Thanks!

PR #6573, opened 2023-01-10
Split general debugging and architecture diagram documentation

YuriSizov:
Supersedes https://github.com/godotengine/godot-docs/pull/6367. Unlike the linked PR this one goes a bit further and addresses some of @clayjohn's concerns.

- The debugging hint is placed into the index debugging and profiling article and is rewritten to explain the idea without relying on `gdb` specifically. This bit is not related to `gdb` at all, it simply explains that the PM and the editor are different processes. So `gdb` is only left as an example, and other relevant articles are linked.
- The intro bit from the old introduction article is placed in the index article for "Core and modules", as that's where it fits the most. It's also adjusted slightly.
- The remaining article is renamed to "Godot's architecture diagram", as that's all there left to it. I've added a redirect rule for the old URL to point to the "Core and modules" index page instead.

@CalebJohn is attributed in the commit for starting the work on this issue and doing the initial work that we failed to merge in a timely manner.


mhilbrunner (approved):

mhilbrunner:
Thank you!

PR #6576, opened 2023-01-10
Remove breadcrumb styling adjustments after the base theme update

YuriSizov:
Follow up to https://github.com/godotengine/godot-docs/pull/6369, seems like breadcrumbs were slightly redesigned, and their styling was fixed. So we no longer need these lines, as they create unnecessary offsets.


mhilbrunner (approved):

mhilbrunner:
Wheee, simplification! Thanks.

PR #6584, opened 2023-01-11
Fix NavigationRegion choice paralysis

smix8:
Users were clearly overwhelmed with this choice what to pick.


mhilbrunner (approved):

mhilbrunner:
Way too many choices indeed! Thanks :)

PR #6586, opened 2023-01-11
Update warning ignore tutorial for new annotation-based syntax

gelvinp:
I've updated the text of the warning-ignore tutorial to reflect the new annotation based syntax. However I have not updated the screenshots, as the warning system still does not work correctly and if I did take screenshots, they would show incorrect behavior.

I asked in #5121 what I should do about the screenshots but didn't get a response, so I'm opening a PR with just the text right now, and I can force push something if needed.


YuriSizov:
> what I should do about the screenshots

If there are some issues that prevent us from creating new screenshots presently, I'd keep them all as is and update them all later, when those issues are addressed.

gelvinp:
That's what this PR is, in that case. I've updated the text, but not the screenshots.

gelvinp:
Added new screenshots, not an exact replica but ignoring a warning that *does* work. Also removed the section about file-level ignores. I thought this was true based on reading through the source, but I was not able to make it work in RC 6.

gelvinp:
Updated to resolve conflicts introduced earlier today. Is there a reason this has been sitting unmerged for almost two months?

YuriSizov:
> Is there a reason this has been sitting unmerged for almost two months?

Nothing wrong on your side, don't worry. Jack other priorities for maintainers, and general lack of time to get to everything.

gelvinp:
No worries, I was just unsure if there were additional changes I needed to make that I wasn't aware of. I know things have been busy with 4.0 stable.

mhilbrunner:
Sorry this has been sitting here for a bit due to the volume of PRs; we've just merged https://github.com/godotengine/godot-docs/pull/8250 which targets the same changes and should supersede this. If there is still anything worth adding, a new followup PR is welcome. Thanks again :)

PR #6587, opened 2023-01-12
Fix typo in Mesh level of detail (LOD)

Calinou:
This fixes https://www.reddit.com/r/godot/comments/109b4ep/were_announcing_a_godot_40_documentation_sprint/j3y3fs9/.

### DIFF
--- a/tutorials/3d/mesh_lod.rst
+++ b/tutorials/3d/mesh_lod.rst
@@ -133,10 +133,10 @@ camera moves away, resulting in higher performance but lower quality.
 
 If you need to perform per-object adjustments to mesh LOD, you can adjust how
 aggressive LOD transitions should be by adjusting the **LOD Bias** property on
-any node that inherits from GeometryInstance3D. Positive values will make LOD
-transitions happen sooner than usual (resulting in lower quality but higher
-performance). Negative values will make LOD transitions happen earlier than
-usual (resulting in higher quality but lower performance).
+any node that inherits from GeometryInstance3D. Values *above* ``1.0`` will make
+LOD transitions happen later than usual (resulting in higher quality but lower
+performance). Values *below* ``1.0`` will make LOD transitions happen sooner than
+usual (resulting in lower quality but higher performance).
 
 Additionally, ReflectionProbe nodes have their own **Mesh LOD Threshold** property
 that can be adjusted to improve rendering performance when the reflection probe


skyace65 (changes requested):
This part actually needs more work. Possibly confusing wording aside, the page is still innacurate with this change. There are no negative LOD values, additionally, lower values (such as .001) make the mesh use a lower level of detail for longer, and values greater than 1 make it stay more detailed longer.

Calinou:
> This part actually needs more work. Possibly confusing wording aside, the page is still innacurate with this change. There are no negative LOD values, additionally, lower values (such as .001) make the mesh use a lower level of detail for longer, and values greater than 1 make it stay more detailed longer.

Fixed.

mhilbrunner (approved):

tutorials/3d/mesh_lod.rst
```diff
@@ -133,9 +133,9 @@ camera moves away, resulting in higher performance but lower quality.
 
 If you need to perform per-object adjustments to mesh LOD, you can adjust how
 aggressive LOD transitions should be by adjusting the **LOD Bias** property on
-any node that inherits from GeometryInstance3D. Positive values will make LOD
-transitions happen sooner than usual (resulting in lower quality but higher
-performance). Negative values will make LOD transitions happen earlier than
+any node that inherits from GeometryInstance3D. Values *above* ``1.0`` will make
+LOD transitions happen sooner than usual (resulting in lower quality but higher
```
skyace65, line 137:
```suggestion
LOD transitions happen sooner than usual (resulting in higher quality but lower
```

tutorials/3d/mesh_lod.rst
```diff
@@ -133,9 +133,9 @@ camera moves away, resulting in higher performance but lower quality.
...
+performance). Values *below* ``1.0`` will make LOD transitions happen later than
 usual (resulting in higher quality but lower performance).
```
skyace65, line 139:
```suggestion
usual (resulting in lower quality but higher performance).
```

skyace65 (changes requested):

tutorials/3d/mesh_lod.rst
Calinou: The currently phrasing is already correct. Lower values of LOD Bias makes meshes look *better* at a given distance, not worse.

clayjohn: The current behaviour is that values between 0 and 1 make the LOD transition happen sooner resulting in worse quality for the same distance. 

Values above one delay the LOD transition so you get higher quality at the same distance.

Calinou: > The current behaviour is that values between 0 and 1 make the LOD transition happen sooner resulting in worse quality for the same distance.
> 
> Values above one delay the LOD transition so you get higher quality at the same distance.

So the LOD Bias property works the opposite of the Mesh LOD Threshold project setting? I consider that to be pretty odd :slightly_frowning_face:

clayjohn: > So the GeometryInstance3D LOD Bias property works the opposite of the Mesh LOD Threshold project setting? I consider that to be pretty odd slightly_frowning_face
>
>In general, I expect LOD bias to work in a fashion where positive values use lower-than-needed LODs and negative values use higher-than-needed LODs (similar to mipmap LOD bias).

Yes, it is rather odd. I would expect it to work like textureLod in GLSL as well. i.e. a bias of one bumps you up one level while -1 bumps you down one level

Calinou: Indeed, I've tested this on 4.0.beta11 and what you're saying is correct.

Could we change this behavior before release? It's likely to be a pitfall people will run into.

clayjohn: I don't know. I'm not familiar enough with the LOD system to comment one why it was implemented that way

skyace65 (approved):

clayjohn (approved):

clayjohn:
Looks great, thanks! Let's remember to discuss the behaviour of LOD with other rendering contributors sometime soon

Issue #6265, opened 2022-10-05
setget doesn't exist in 4.0 and isnt part of GDScript Grammar


TheYellowArchitect:
https://docs.godotengine.org/en/latest/development/file_formats/gdscript_grammar.html?highlight=setget#gdscript-grammar

Basically, convert the above `setget` to use `set` and `get`

PR #6594, opened 2023-01-12
Updated GDScript grammar from setget to set/get

solorvox:
Fixes for set/get from older setget for GDScript grammar and tutorials. 
Should fix issue: https://github.com/godotengine/godot-docs/issues/6265

### DIFF
--- a/contributing/development/file_formats/gdscript_grammar.rst
+++ b/contributing/development/file_formats/gdscript_grammar.rst
@@ -37,10 +37,14 @@ for reference purposes.
         | "tool"
         ;
 
-    classVarDecl = [ "onready" ] [ export ] "var" IDENTIFIER [ ":" typeHint ]
-        [ "=" expression ] [ setget ] NEWLINE ;
-    setget = "setget" [ IDENTIFIER ] [ "," IDENTIFIER] ;
-    export = "export" [ "(" [ BUILTINTYPE | IDENTIFIER { "," literal } ] ")" ] ;
+    classVarDecl = [ @onready ] [ @export ] "var" IDENTIFIER [ ":" typeHint ]
+    [ "=" expression ] IDENTIFIER
+    [ ":" [ "set" "(" value ")" ":" { method } ] | "get" ":" { method } ] NEWLINE ;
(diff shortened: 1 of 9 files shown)


contributing/development/file_formats/gdscript_grammar.rst
```diff
@@ -37,10 +37,14 @@ for reference purposes.
         | "tool"
         ;
 
-    classVarDecl = [ "onready" ] [ export ] "var" IDENTIFIER [ ":" typeHint ]
-        [ "=" expression ] [ setget ] NEWLINE ;
-    setget = "setget" [ IDENTIFIER ] [ "," IDENTIFIER] ;
-    export = "export" [ "(" [ BUILTINTYPE | IDENTIFIER { "," literal } ] ")" ] ;
+    classVarDecl = [ @onready ] [ @export ] "var" IDENTIFIER [ ":" typeHint ]
+    [ "=" expression ] IDENTIFIER
```
jmdejong, line 41: I think that `IDENTIFIER` should not be here. I'm still learning the syntax myself, but I don't see it anywhere in your examples

solorvox: Do you mean the first or second occurrence?  I was trying to follow the EBNF wiki.  But also using the previous setget syntax from the manual page.  What I thought it `IDENTIFIER` meant was for example `var foo = bar`.  Where `foo` was the first `IDENTIFIER` and `bar` was the second being an identifier to another variable.  But `bar` could of course be replaced with an expression as well.  Happy to take suggestions as EBNF wasn't completely clear to me either. :)

jmdejong: I think in your example `bar` is part of the expression that is already defined by `[ "=" expression ]`.

The current definition would match `var foo bar` or `var foo = 123 bar` or `@export var foo  : int = 123 bar` (with `bar` as the second identifier).

solorvox:
How does this commit look?

mhilbrunner:
The remaining changes here were merged as part of https://github.com/godotengine/godot-docs/pull/9546. Thanks for contributing!

PR #6595, opened 2023-01-12
Highlight link targets in the theme

Calinou:
This makes it easier to notice where the linked text is on the page, especially on tall monitors when linking at the bottom of a page.

## Preview

https://user-images.githubusercontent.com/180032/212127781-3ebf0b20-b90d-4daa-9af0-10a9e3aba75b.mp4

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -255,6 +255,24 @@
     }
 }
 
+@keyframes target-highlight-fade {
+    from {
+        background-color: hsla(56, 100%, 50%, 0.2);
+    }
+
+    to {
+        background-color: hsla(56, 100%, 50%, 0.05);
+    }
+}
+
+*:target {
+    /* Highlight selected sections, with an animation to make it more noticeable initially. */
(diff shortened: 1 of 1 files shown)


YuriSizov:
I think there is some merit to it, but I'd be cautious adding bright yellow blinking highlights. It's less of a problem with the light theme, but it pretty annoying in the dark theme. Also not sure about the highlight never going fully away. It might've looked better if it was only for the title.

Calinou:
> I think there is some merit to it, but I'd be cautious adding bright yellow blinking highlights. It's less of a problem with the light theme, but it pretty annoying in the dark theme. Also not sure about the highlight never going fully away.

I'd prefer to have a subtle but persistent highlight, so you know what you originally clicked for, even if you alt-tab away and back from the page.

> It might've looked better if it was only for the title.

This might be feasible by using a thick border instead of a background color.

mhilbrunner:
Closing due to lack of consensus on this one. I think the idea has potential, if we can come up with some less intrusive styling.

PR #6603, opened 2023-01-13
Update Importing 3D scenes documentation for Godot 4.0

Calinou:
- Mention `.blend` import.
- Mention `.fbx` import.
- Document OBJ import modes and limitations.
- Mention Blender's built-in Collada usually working OK for simple scenes without animation (from personal experience).
- Remove mentions of ESCN export as the [exporter](https://github.com/godotengine/godot-blender-exporter) is currently not compatible with Godot 4.0. Since contributor interest in this exporter is low, it's unlikely to be fixed for Godot 4.0's release.

This PR doesn't mention the Advanced Import Settings dialog. This is something that can be done in a separate PR, as it's independent from documenting which formats are supported.

I haven't removed the dedicated pages about the ESCN exporter – let me know if I should do that.

See https://github.com/godotengine/godot-docs/issues/5121.

### DIFF
--- a/tutorials/assets_pipeline/importing_scenes.rst
+++ b/tutorials/assets_pipeline/importing_scenes.rst
@@ -13,42 +13,54 @@ transferred as close as possible.
 
 Godot supports the following 3D *scene file formats*:
 
-* glTF 2.0 **(recommended)**. Godot has full support for both text (``.gltf``) and binary (``.glb``) formats.
-* DAE (COLLADA), an older format that is fully supported.
-* OBJ (Wavefront) format + their MTL material files. This is also fully supported, but pretty limited (no support for pivots, skeletons, animations, PBR materials, ...).
-* ESCN, a Godot-specific format that Blender can export with a plugin.
-* FBX, supported via a reverse engineered importer. So we recommend using other formats listed above, if suitable
-  for your workflow.
+- glTF 2.0 **(recommended)**. Godot has full support for both text (``.gltf``) and binary (``.glb``) formats.
+- ``.blend`` (Blender). This works by calling Blender to export to glTF in a
+  transparent manner (requires Blender to be installed).
+- DAE (COLLADA), an older format that is fully supported.
+- OBJ (Wavefront) format + their MTL material files. This is also fully
+  supported, but pretty limited given the format's limitations (no support for
+  pivots, skeletons, animations, UV2, PBR materials, ...).
+- FBX, supported via `FBX2glTF <https://github.com/godotengine/FBX2glTF>`__ integration.
+  This requires installing an external program that links against the proprietary FBX SDK,
+  so we recommend using other formats listed above (if suitable for your workflow).
 
 Just copy the scene file together with the texture to the project repository, and Godot will do a full import.
 
 It is important that the mesh is not deformed by bones when exporting. Make sure that the skeleton is reset to its T-pose
 or default rest pose before exporting with your favorite 3D editor.
 
-Exporting glTF 2.0 files from Blender
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Exporting glTF 2.0 files from Blender (recommended)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-There are three ways to export glTF files from Blender. As a glTF binary (``.glb`` file), glTF embedded (``.gltf`` file),
-and with textures (``gltf`` + ``.bin`` + textures).
+There are 3 ways to export glTF files from Blender:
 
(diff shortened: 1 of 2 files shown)


clayjohn (approved):
Looks good to me Would be nice to have someone from the import team provide input on content

skyace65 (commented):
Everything here looks good, information on using the advanced importer for 3D scenes can be added in a future PR.

Before I give approval, was support for .ESCN dropped in Godot 4? If it was the pages for how to use that importer should be removed as well.

fire:
As far as I know the ESCN importer is .tscn renamed. The lod feature not working, but that's an ordinary bug.

YuriSizov:
> there's usecases for taking an ordinary scn renaming it to go through the import process again.

That doesn't seem like a valid use-case for supporting a custom file extension. Currently format does not work in Godot 4 anyway, so even as a hack it's not usable.

fire:
When I last tested taking a tscn and renaming to escn works, so I'll get a pr for the importer mesh instance 3d conversion in a few hour.

fire:
I pushed a pr https://github.com/godotengine/godot/pull/71409 to modify the importer code that expects importer mesh instance 3d. It will preserve the ability for a Blender escn exporter to be updated and the tscn to escn usecase.

skyace65:
@Calinou the ESCN PR got merged so this can be updated. put the escn info that was deleted back in and it should be good to go.

Calinou:
> @Calinou the ESCN PR got merged so this can be updated. put the escn info that was deleted back in and it should be good to go.

The add-on is still not updated to work with Blender 3.4, so I'm not sure if it's a good idea to list it.

I remember discussing with reduz that ESCN export should no longer be recommended to new users in general.

tutorials/assets_pipeline/importing_scenes.rst
```diff
@@ -63,27 +75,97 @@ either of those glTF binary files are fine.
...
+Blender 3 Path**).
+
+If you keep ``.blend`` files within your project folder but don't want them to
+be imported by Godot, disable **Filesystem > Import > Blender > Enabled** in the
+advanced Project Settings.
+
+.. note::
+
+    When working in a team, keep in mind using ``.blend`` files in your project
+    will require *all* team members to have Blender installed. While Blender is
```
lyuma, line 108:
```suggestion
    will require *all* team members to have the same version of Blender installed. While Blender is
```
Blender pushes bugfixes from time to time that can impact the gltf process, so just an idea to point this out. Blender 3.4 files don't open properly in Blender 3.0, for example

lyuma (commented):
We might also want to add a comment to use the latest version of the tooling because the gltf addons are still actively developed and in the case of Blender, there are major bugfixes in gltf export as recently as blender 3.4

escn import should be put back in but it should be clear that this is "escn for godot 4", and not compatible with escn exporters for blender and elsewhere.
See fire's comment for valid usecases of escn.
- sharing scenes made for the same godot version
- running mesh or animation optimizations on godot content
- 

Also, a side note but shouldn't escn for godot 3 import fine in godot 4? This is the upgrade path so godot should translate it.

YuriSizov:
> sharing scenes made for the same godot version
> running mesh or animation optimizations on godot content

We don't have any tooling for this. If you use it in your projects, you're using a hack that happens to work, but this is not a workflow that Godot is designed to have. Thus these are not valid use cases. These also have nothing to do with importing, so even if we were somehow legitimize those hacks, they are irrelevant to this article.

fire:
Let's remove the documentation until we develop one.

skyace65 (approved):

skyace65:
Given the consensus I'll merge this PR as is.

maksad1git:
i think a should to copy this from the last issue)))

If u want Godot to support .blend import using Blender 3.x for example 3.4.1, u have to:

- Download official Blender 3.0 from here:
https://download.blender.org/release/Blender3.0/blender-3.0.0-windows-x64.zip
- Unzip to some folder where it should be
- Run it once and skip welcome dialog
- Close Blender
- Go to Godot and set Blender path to ur Blender 3.0 like it described here:
https://docs.godotengine.org/en/stable/tutorials/assets_pipeline/importing_scenes.html#importing-blend-files-directly-within-godot
- Import some .blend(default cube for example) made in Blender 3.0
- Close Godot
- Reopen Godot and change Blender path to ur Blender 3.4.1 installation
- enjoy

p.s. Level design is much easier with .blend support, so thanks for that guys. Hope u will add last version support))) i think issue is in little count of lines of code)

fire:
I don't believe anyone backported the blender import to godot 3. 

https://github.com/V-Sekai/godot-blender was the predecessor the godot 4 implementation, but I haven't touched it since Reduz gave the go ahead to port to 4.

PR #6604, opened 2023-01-14
Update particle shader format in Controlling thousands of fish

jmdejong:
The tutorial mentioned using a vertex function in particle shaders, which is not how particle shaders work anymore

### DIFF
--- a/tutorials/performance/vertex_animation/controlling_thousands_of_fish.rst
+++ b/tutorials/performance/vertex_animation/controlling_thousands_of_fish.rst
@@ -55,15 +55,16 @@ A unique thing about particle shaders is that some built-in variables are saved 
 ``TRANSFORM``, ``COLOR``, and ``CUSTOM`` can all be accessed in the shader of the mesh, and
 also in the particle shader the next time it is run.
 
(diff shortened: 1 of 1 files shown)


jmdejong:
I'm not sure what's the best way to style function names (start and process) in this tutorial: with or without backticks.

Calinou:
> I'm not sure what's the best way to style function names (start and process) in this tutorial: with or without backticks.

Backticks are good, but make sure to also add parentheses:

```
``start()``
```

jmdejong:
What about functions that take arguments? The `hash` function was already mentioned by the original text.
Should that be written as `hash()`, `hash(...)` , `hash(x)` or `hash(uint x)`?

Calinou:
> What about functions that take arguments? The `hash` function was already mentioned by the original text. Should that be written as `hash()`, `hash(...)` , `hash(x)` or `hash(uint x)`?

I suggest writing it as `hash()`. We don't need to state the arguments every time, just make sure there are parentheses to make it explicit that it's a function.

clayjohn (approved):

Calinou (approved):
Thanks!

PR #6609, opened 2023-01-14
Fix formatting issues reported by the builder

YuriSizov:
Couple of regressions in formatting from recent PRs, issues were reported by the CI builder.

Of note: we shouldn't add letters at the end of formatting, or we should escape them with "\\" if necessary. In this case I either removed it or moved it inside of the formatting to fix the issue.


mhilbrunner (approved):

mhilbrunner:
Thanks, merged.

PR #6611, opened 2023-01-14
Pin additional build requirements to avoid syntax highlighting issues

YuriSizov:
If you've had your dependencies installed for a long time, you may have an outdated version of Pygments that fails to parse some constructs, such as string interpolation with `$` in C#. Installing `requirements.txt` again didn't update Pygments for me, resolving that my local version, 2.9.0, was enough to meet the dependency requirement in one of the packages.

So to solve this this PR pins Pygments specifically to the same version RTD uses, 2.14.0. This solved all syntax highlighter issues for my local builds. CI builds seems to be unaffected, because they probably install everything anew every time.

There is still one issue with the GLSL snippet here:
https://docs.godotengine.org/en/latest/tutorials/shaders/shader_reference/shader_preprocessor.html#define

It is possible that it's somehow incorrect, so I pinged rendering folks about it on RC. But it can also be fixed with a later PR.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,16 +1,19 @@
-# Base dependencies
-
 # Sync with readthedocs:
 # https://github.com/readthedocs/readthedocs.org/blob/master/requirements/pip.txt
 # https://github.com/readthedocs/readthedocs.org/blob/master/requirements/docs.txt
+
+# Base dependencies
+pygments==2.14.0
+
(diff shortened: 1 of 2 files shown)


mhilbrunner (approved):

mhilbrunner:
Thanks! Makes sense.

PR #6624, opened 2023-01-17
Update Pause Mode documentation

pgpais:
In the Pause Mode documentation, in the Pause Menu example, it was still referring to the old "Process" pause mode. I changed it to refer to "When Paused" and "Always" (as both seemed correct to me)

### DIFF
--- a/tutorials/scripting/pausing_games.rst
+++ b/tutorials/scripting/pausing_games.rst
@@ -93,8 +93,8 @@ Pause Menu Example
 ------------------
 
 Here is an example of a pause menu. Create a popup or panel with controls
-inside, and set its pause mode to "Process" then hide it. By setting the
-root of the pause popup to "Process", all children and grandchildren will
+inside, and set its pause mode to "When Paused" or "Always" then hide it. By setting the
+root of the pause popup to "When Paused" or "Always", all children and grandchildren will
 inherit that state. This way, this branch of the scene tree will continue
 working when paused.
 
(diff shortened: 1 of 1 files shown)


tutorials/scripting/pausing_games.rst
```diff
@@ -93,8 +93,8 @@ Pause Menu Example
 ------------------
 
 Here is an example of a pause menu. Create a popup or panel with controls
-inside, and set its pause mode to "Process" then hide it. By setting the
-root of the pause popup to "Process", all children and grandchildren will
+inside, and set its pause mode to "When Paused" or "Always" then hide it. By setting the
+root of the pause popup to "When Paused" or "Always", all children and grandchildren will
```
skyace65, line 97:
```suggestion
inside, and set its pause mode to "When Paused" then hide it. By setting the
root of the pause popup to "When Paused", all children and grandchildren will
```

skyace65 (changes requested):
Good catch! However I think that part should just use "when paused" for the example, as there's not really a use case for a pause menu processing when the game is running.

pgpais:
Right. I wasn't sure if _Show()_ would happen if the node isn't being processed, hence my suggestion.

skyace65:
It will work as long as the script isn't attached to something with the process mode "when paused", I just confirmed this with the latest beta.

skyace65 (approved):
Thanks! And congrats on your first merged PR.

PR #6634, opened 2023-01-19
Updated tutorial for new set/get method for Godot4

solorvox:
Fixes old setget method for tutorials to new set/get in Godot4. 
Addresses issue: https://github.com/godotengine/godot-docs/issues/6265

### DIFF
--- a/tutorials/2d/custom_drawing_in_2d.rst
+++ b/tutorials/2d/custom_drawing_in_2d.rst
@@ -69,13 +69,11 @@ redrawn if modified:
 
     extends Node2D
 
-    export (Texture) var texture setget _set_texture
-
-    func _set_texture(value):
-        # If the texture variable is modified externally,
-        # this callback is called.
-        texture = value  # Texture was changed.
-        queue_redraw()  # Trigger a redraw of the node.
+    @export var texture: Texture:
+        set(value):
+            # Trigger a redraw of the node when texture is set.
+            texture = value
+            queue_redraw()
 
     func _draw():
         draw_texture(texture, Vector2())
--- a/tutorials/best_practices/data_preferences.rst
+++ b/tutorials/best_practices/data_preferences.rst
@@ -237,7 +237,7 @@ tree structures.
     class_name TreeNode
 
     var _parent : TreeNode = null
-    var _children : = [] setget
+    var _children := []
 
     func _notification(p_what):
         match p_what:
--- a/tutorials/best_practices/godot_interfaces.rst
+++ b/tutorials/best_practices/godot_interfaces.rst
@@ -55,7 +55,9 @@ access.
     tool # Must place at top of file.
 
     # Must configure from the editor, defaults to null.
-    export(Script) var const_script setget set_const_script
+    @export var const_script: Script = null:
+        set = set_const_script
+
     func set_const_script(value):
         if Engine.is_editor_hint():
             const_script = value
--- a/tutorials/best_practices/godot_notifications.rst
+++ b/tutorials/best_practices/godot_notifications.rst
@@ -156,7 +156,8 @@ instantiation:
     # "one" is an "initialized value". These DO NOT trigger the setter.
     # If someone set the value as "two" from the Inspector, this would be an
     # "exported value". These DO trigger the setter.
-    export(String) var test = "one" setget set_test
+    @export var test: String = "one":
+        set = set_test
 
     func _init():
         # "three" is an "init assignment value".
--- a/tutorials/plugins/running_code_in_the_editor.rst
+++ b/tutorials/plugins/running_code_in_the_editor.rst
@@ -181,8 +181,8 @@ run the game, it will spin counter-clockwise.
 Editing variables
 -----------------
 
-Add and export a variable speed to the script. The function set_speed after
-``setget`` is executed with your input to change the variable. Modify
+Add and export a variable speed to the script. Use a custom property ``set``
+method to reset the rotation to zero when speed is assigned a value.  Modify
 ``_process()`` to include the rotation speed.
 
 .. tabs::
--- a/tutorials/scripting/gdscript/gdscript_styleguide.rst
+++ b/tutorials/scripting/gdscript/gdscript_styleguide.rst
@@ -35,10 +35,13 @@ Here is a complete class example based on these guidelines:
 
     signal state_changed(previous, new)
 
-    export var initial_state = NodePath()
-    var is_active = true setget set_is_active
+    @export var initial_state = NodePath()
+    var is_active = true:
+        set = set_is_active
 
-    @onready var _state = get_node(initial_state) setget set_state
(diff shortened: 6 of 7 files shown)


tutorials/2d/custom_drawing_in_2d.rst
```diff
@@ -69,12 +69,14 @@ redrawn if modified:
 
     extends Node2D
 
-    export (Texture) var texture setget _set_texture
+    @export var texture : Texture :
```
dalexeev, line 72:
```suggestion
    @export var texture: Texture:
```

See [GDScript style guide](https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_styleguide.html#declared-types) and [the example](https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_basics.html#properties-setters-and-getters).

tutorials/2d/custom_drawing_in_2d.rst
```diff
@@ -69,12 +69,14 @@ redrawn if modified:
...
+        set(value):
+            texture = value
+            _set_texture(value)
 
     func _set_texture(value):
         # If the texture variable is modified externally,
         # this callback is called.
```
dalexeev, line 79: This comment can be misleading, since in 4.0 setters and getters are always called (both externally and internally).

tutorials/2d/custom_drawing_in_2d.rst
```diff
@@ -69,12 +69,14 @@ redrawn if modified:
 
     extends Node2D
 
-    export (Texture) var texture setget _set_texture
+    @export var texture : Texture :
+        set(value):
+            texture = value
+            _set_texture(value)
```
dalexeev, line 75: Since `_set_texture` is not called anywhere else, I think it makes sense to move its body (`queue_redraw()`) here.

tutorials/best_practices/data_preferences.rst
```diff
@@ -237,7 +237,7 @@ tree structures.
     class_name TreeNode
 
     var _parent : TreeNode = null
-    var _children : = [] setget
+    var _children : = []
```
dalexeev, line 240: See [GDScript style guide - Inferred types](https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_styleguide.html#inferred-types).

solorvox:
I updated with amend commit for those suggestions and merged the conflicted master file.

tutorials/best_practices/godot_interfaces.rst
```diff
@@ -55,7 +55,10 @@ access.
     tool # Must place at top of file.
 
     # Must configure from the editor, defaults to null.
-    export(Script) var const_script setget set_const_script
+    @export var const_script : Script = null :
```
dalexeev, line 58:
```suggestion
    @export var const_script: Script = null:
```

Here and below you have to correct too.

tutorials/2d/custom_drawing_in_2d.rst
```diff
@@ -69,13 +69,12 @@ redrawn if modified:
...
-
-    func _set_texture(value):
-        # If the texture variable is modified externally,
-        # this callback is called.
-        texture = value  # Texture was changed.
-        queue_redraw()  # Trigger a redraw of the node.
+    @export var texture: Texture:
+        set(value):
+            # Trigger a redraw of the node when texture
+            # is set.
```
dalexeev, line 75:
```suggestion
            # Trigger a redraw of the node when texture is set.
```

tutorials/best_practices/godot_notifications.rst
```diff
@@ -156,7 +156,9 @@ instantiation:
     # "one" is an "initialized value". These DO NOT trigger the setter.
     # If someone set the value as "two" from the Inspector, this would be an
     # "exported value". These DO trigger the setter.
-    export(String) var test = "one" setget set_test
+    @export var test : String = "one" :
```
dalexeev, line 159:
```suggestion
    @export var test: String = "one":
```

tutorials/plugins/running_code_in_the_editor.rst
```diff
@@ -182,7 +182,7 @@ Editing variables
 -----------------
 
 Add and export a variable speed to the script. The function set_speed after
-``setget`` is executed with your input to change the variable. Modify
+``set`` is executed with your input to change the variable. Modify
```
dalexeev, line 185: The function `set_speed` does not exist. This sentence needs to be rewritten.

tutorials/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -36,9 +36,14 @@ Here is a complete class example based on these guidelines:
     signal state_changed(previous, new)
 
     export var initial_state = NodePath()
-    var is_active = true setget set_is_active
+    var is_active = true :
```
dalexeev, line 39:
```suggestion
    var is_active = true:
```

tutorials/best_practices/godot_interfaces.rst
```diff
@@ -55,7 +55,10 @@ access.
     tool # Must place at top of file.
 
     # Must configure from the editor, defaults to null.
-    export(Script) var const_script setget set_const_script
+    @export var const_script: Script = null:
+        set(value):
+            set_const_script(value)
```
adamscott, line 60: Shouldn't it be this?
```gdscript
@export var const_script: Script = null:
    set = set_const_script
```

solorvox: One could argue that having the two line function set_const_script() isn't needed anyway I think.  But my thinking was to show how a value is passed via the property setter as omitting it is less clear to me what is happening.

YuriSizov: Well, it's new syntax, it'll take some time for it to become natural. But doing it the way you did comes with an unnecessary overhead, as you create an extra function just to pass the value. We have a way to directly assign a setter function in this syntax, we should use it.

solorvox: If the point is to reduce extra methods, wouldn't it be best to avoid the whole method set_const_script() and just put the conditional `if` inside the setter method altogether?

YuriSizov: That would be equal amount of methods to what Adam suggests.

solorvox: I've made the suggested changes

adamscott (approved):
LGTM

cbscribe:
I notice the style guide hasn't been updated to reflect the new set/get syntax.

Before merging this, we need some consensus on which is the preferred style: at variable declaration, or as a separate function.

mhilbrunner:
The remaining changes here were merged as part of https://github.com/godotengine/godot-docs/pull/9546. Thanks for contributing!

PR #6638, opened 2023-01-20
Update Python version requirements on Compiling pages

Calinou:
The `master` branch has required Python 3.6 to be compiled for a few years now.

This also makes the text more consistent across platforms.


skyace65 (approved):
I just confirmed on my desktop this is accurate and Godot doesn't build without Python 3.6 or later. Changes look good, and I've confirmed the new information about Homebrew and MacPorts

PR #6645, opened 2023-01-21
Update canvas_layers.rst to clarify wording

anderspitman:


Calinou (commented):
I think the current wording is fine as is. The new wording isn't necessarily better, especially due to the "which" which is grammatically out-of-place here.

mhilbrunner:
Thanks for contributing! I agree with Calinou's assessment.

PR #6648, opened 2023-01-22
Document ways to further speed up linking in Compiling for Linux/*BSD

Calinou:
mold and/or system libraries can be used to make iteration times significantly faster when working on the engine.

I've only added a package installation oneliner for Fedora (tested in a fresh Fedora 37 Docker container), but the same approach should be usable on any Linux distribution.

This will need a manual cherry-pick for `3.5`/`3.6` as some `builtin_` options aren't used in `3.x`. mold support is fully available in `3.x` still.


mhilbrunner (approved):
LGTM

PR #6653, opened 2023-01-23
Note on unbalanced sound.

DavidVereb:
When you follow this tutorial, you'll notice the audio stream is quite left-balanced. This note should be handy for a beginner wondering why this is this happening and how to fix it.

Some stylistic advice are always welcome :)


DavidVereb:
True. I didn't notice. Sorry for the unnecesary Pull Request :)

PR #6654, opened 2023-01-23
Fix CameraPivot rotation in Moving the player with code

gjpin:
CameraPivot should be rotated -45 degrees instead of 45 degrees in order to produce the results demonstrated in the screenshots. Otherwise, the camera will be pointing upwards instead of downwards.

Change has already been made in the [master branch](https://github.com/godotengine/godot-docs/blob/master/getting_started/first_3d_game/03.player_movement_code.rst).


Calinou (approved):

PR #6658, opened 2023-01-24
Remove most mentions of SCons `-j` parameter as it's now set automatically

Calinou:
The Introduction to the buildsystem page still mentions this parameter
in case it needs to be changed, but in most situations, the default
value of `(number of CPU threads) - 1` does a good job.


mhilbrunner (approved):

mhilbrunner:
What a good change :) Nice, merged!

PR #6618, opened 2023-01-16
Added documention page about exporting a Godot XR project to Android

BastiaanOlij:
This PR adds a documentation page to the XR section detailing the export process to deploy an XR project on Meta Quest or Pico headsets.

PR #6662, opened 2023-01-25
Add an introduction page for XR tools

BastiaanOlij:
This PR adds an introduction page about installing the XR Tools library and adds content to the basic locomotion page.

Note that I've written this with the assumption that by the time stable is released, the issues around the GDScript files not loading correctly the first time because the `.godot` folder contents hasn't been updated yet will be solved.


Malcolmnixon:
You mention supporting OpenXR and WebXR, but the setting_up_xr.rst file you start from only has the OpenXR example in; however godot-xr-tools now has the StartXR.tscn which will initialize either OpenXR or WebXR depending on which is detected. Perhaps it might be worthwhile offering it to the user when godot-xr-tools is introduced.

Malcolmnixon:
The `godot_xr_tools_root_folder.webp` image has an android folder; but I don't see where you walk the user through installing the android build template.

tutorials/xr/basic_xr_locomotion.rst
```diff
@@ -3,5 +3,65 @@
...
+
+We want to enable this on the right hand controller and we do this by adding a subscene to the right hand :ref:`XRController3D <class_xrcontroller3d>` node. Select ``addons/godot-xr-tools/functions/movement_direct.tscn`` as the scene to add.
+
+This function adds forward and backwards movement to the player by using the joystick on the right hand controller. It has an option to also add left/right strafe but by default this is disabled.
+
+Instead we are going to add the ability for the player to also turn with this joystick so we will add another subscene to our controller node, select ``addons/godot-xr-tools/functions/movement_turn.tscn`` for this.
+
+The turn system by default uses a snap turn approach. This means that turning happens in steps. This may seem jarring however it is a tried and tested method of combatting motion sickness. You can easily switch to a mode that offers smooth turning by changing the ``mode`` property on the turn node.
```
skyace65, line 42:
```suggestion
The turn system by default uses a snap turn approach. This means that turning happens in steps. This may seem jarring however it is a tried and tested method of combating motion sickness. You can easily switch to a mode that offers smooth turning by changing the ``mode`` property on the turn node.
```

tutorials/xr/basic_xr_locomotion.rst
```diff
@@ -3,5 +3,65 @@
...
+
+.. image:: img/godot_xr_tools_floor.webp
+
+Direct movement
+---------------
+
+We're going to start adding some basic direct movement to our setup. This allows the user to move through the virtual world using joystick input.
+
+.. note::
+  It is important to note that moving through the virtual world while the player is standing still in the real world can be nausia inducing especially for players who are new to VR. The default settings on our movement functions are fairly conservative. We advise you to stick to these defaults but offer features in game to enable less comfortable settings for more experienced users who are used to playing VR games.
```
skyace65, line 34:
```suggestion
  It is important to note that moving through the virtual world while the player is standing still in the real world can be nausea inducing especially for players who are new to VR. The default settings on our movement functions are fairly conservative. We advise you to stick to these defaults but offer features in game to enable less comfortable settings for more experienced users who are used to playing VR games.
```

skyace65 (changes requested):
Didn't get to do a full read through yet, but I check both pages for spelling errors and these were the only I saw.

BastiaanOlij:
> The `godot_xr_tools_root_folder.webp` image has an android folder; but I don't see where you walk the user through installing the android build template.

Thats in a different PR :)
https://github.com/godotengine/godot-docs/pull/6618

BastiaanOlij:
> You mention supporting OpenXR and WebXR, but the setting_up_xr.rst file you start from only has the OpenXR example in; however godot-xr-tools now has the StartXR.tscn which will initialize either OpenXR or WebXR depending on which is detected. Perhaps it might be worthwhile offering it to the user when godot-xr-tools is introduced.

Just like there is another page for deploying to Android, I also want to make a separate page about WebXR.

One thing I learned early on when making tutorials is not to try and do too much at once, it just overloads people. Also not everyone is going to want to support both OpenXR and WebXR. That we have things in XR tools that make it easier to set things up if you want both, thats not for everyone to use. Same with our solution for background scene loading.

So the base tutorial is going to stick to straight forward stuff, and then when people want more they should check the XR Tools documentation for learning more about it.

tutorials/xr/basic_xr_locomotion.rst
```diff
@@ -3,5 +3,65 @@
...
+The player can also adjust the orientation by using the left hand controllers joystick.
+
+If you've followed all instructions correctly your scene should now look something like this:
+
+.. image:: img/godot_xr_tools_basic_movement.webp
+
+More advanced movement features
+-------------------------------
+
+Godot XR Tools adds many more movement features such as gliding, a grapple hook implementation (go Spidey!), a jetpack, climbing mechanics, etc.
```
skyace65, line 63:
```suggestion
Godot XR Tools adds many more movement features such as gliding, a grapple hook implementation, a jetpack, climbing mechanics, etc.
```
Given how extremely protective Disney can be of their IP we should probably leave this out. It also looks a little unprofessional.

skyace65 (changes requested):
Outside of the spiderman reference there's a few lines in the file for both pages that go on for way too long and should be broken up into two. Anything where github had to make more lines like this
TooLong

Those two issues aside the content itself is good. My only concern is I feel like this might be going a bit too close to tutorial style documentation instead of manual style, but I'll defer to a second opinion on that, and approve this once the things I mentioned earlier have been fixed.

tutorials/xr/basic_xr_locomotion.rst
BastiaanOlij: Lol, yeah you're right :)

BastiaanOlij:
> Those two issues aside the content itself is good. My only concern is I feel like this might be going a bit too close to tutorial style documentation instead of manual style, but I'll defer to a second opinion on that, and approve this once the things I mentioned earlier have been fixed.

Hmm, I wasn't aware we're changing course there, this is definitely meant as a getting started guide to XR seeing we had that in the Godot 3 documentation and were sorely missing this in Godot 4.

I wouldn't be a fan of loosing that part of our manuals, getting up to speed with a subsystem in Godot feels important.

BastiaanOlij:
> Outside of the spiderman reference there's a few lines in the file for both pages that go on for way too long and should be broken up into two. Anything where github had to make more lines like this
> TooLong

I'll see what I can do here as I can be pretty long winded. However some of these lines are simply long because of the syntax for pointing to other pages is so long, ergo "CollisionShape3D" turns into ":ref:`CollisionShape3D <class_collisionshape3d>`"

skyace65:
> Hmm, I wasn't aware we're changing course there, this is definitely meant as a getting started guide to XR seeing we had that in the Godot 3 documentation and were sorely missing this in Godot 4.
> 
> I wouldn't be a fan of loosing that part of our manuals, getting up to speed with a subsystem in Godot feels important.

Sorry for the late reply.

Showing people how to get started is fine, it's how you do it. My concern with this content is it's coming across as a guide on one big project. What would be fine to me is individual pages going over the content in depth but not being linked together. Obviously the basics of XR in Godot (basic node setup) are a prerequisite for using the movement nodes, but the manual pages don't need to be directly linked from one to the other.

But again, that's just my take and I'll deffer to a second opinion.

mhilbrunner (approved):
Gave it a read, spotted no errors so far. LGTM!

skyace65 (approved):

BastiaanOlij:
> Showing people how to get started is fine, it's how you do it. My concern with this content is it's coming across as a guide on one big project. What would be fine to me is individual pages going over the content in depth but not being linked together. Obviously the basics of XR in Godot (basic node setup) are a prerequisite for using the movement nodes, but the manual pages don't need to be directly linked from one to the other.

Ah I get you, I think the pages I'm hoping to write in the future will mostly stand on their own, but indeed, the basic setup and some form of movement is kind of a prerequisite for any further things you do in XR.

PR #6663, opened 2023-01-25
Switched KinematicBody2D to CharacterBody image in latest docs

### LINKED ISSUE #6660: Learning new features page refers to Godot 3.x node in latest documentation
**Your Godot version:**
4.0

**Issue description:**
On learning new features page, there is an image reference to KinematicBody2D which
does not exist anymore in Godot 4.0 (renamed to CharacterBody). Albeit minor thing,
a new user learning Godot 4 might get confused.

[https://docs.godotengine.org/en/latest/getting_started/introduction/learning_new_features.html#making-the-most-of-this-manual](https://docs.godotengine.org/en/latest/getting_started/introduction/learning_new_features.html#making-the-most-of-this-manual)

moomoo02:
Closes #6660 

Switched out image of KinematicBody2D which does not exist anymore in Godot 4.0 with CharacterBody.

This is what it will look like in the docs:

Screen Shot 2023-01-24 at 8 55 40 PM


skyace65 (changes requested):
Two issues, one we're switching over from png to webp. Two, this image is too big, I'm going to assume you took a screenshot on a 4K monitor, it needs to be scaled down. If you need help with either of those you can find information [here](https://docs.godotengine.org/en/latest/contributing/documentation/docs_image_guidelines.html). Please let me know if anything on that page isn't clear.

skyace65:
Did you use lossy compression on that image? I'm seeing compression artifacts

moomoo02:
I used Preview on Mac to half the dimensions of the image, then converted it webp using an online converter called CloudConvert.  I'm not sure if this method has lossy compression or not, but that may be the case since there seems to be compression artifacts.  Should I try using Krita instead?

skyace65:
Ok the compression artifacts are gone but pixel wise that doesn't look like its been scaled down properly. I took my own screenshot for comparison on my 1080p monitor and the width is around 700 compared to your 1000. It's not going to be an exact match but a difference that big means something isn't right. For the sake of time here's the image I made, it's good to go for this. [Image.zip](https://github.com/godotengine/godot-docs/files/10528957/Image.zip)

Also I'll do a squash and merge this time, but in the future your PR's need to be a single commit, if you don't know how to edit them and keep them as 1 commit we have info [here](https://docs.godotengine.org/en/latest/contributing/workflow/pr_workflow.html)

skyace65 (approved):
Thanks for the PR!

moomoo02:
Thank you for being patient with me on such a simple task T-T

I'm so sorry you had to jump in and create the image.  I'll be sure to keep my PR's in a single commit next time!  Thanks again, I learned a lot from this experience!

skyace65:
Don't worry about it, we all have to start learning somewhere.

PR #6667, opened 2023-01-25
tutorials/scripting: Mention % literal syntax

vmedea:
It was unclear to me what that '%' literal syntax does, add examples in the appropriate places.

`$%` syntax works as well and is the same as `%` but I'm not sure that's worth mentioning.

Might also want to add a mention of % special syntax in `Node.get_node` docs, but those aren't here.


Calinou:
Note that this shorthand is only available in 4.0, not 3.6, so this PR should **not** be cherry-picked.

mhilbrunner (approved):

mhilbrunner:
Thank you! Merged.

PR #6674, opened 2023-01-26
Continue no longer works in a match statement

Mach565:
As of godotengine/godot#71914, continue no longer works in a match statement, so mention of such functionality would be an error.


YuriSizov (approved):

YuriSizov:
Thanks, and congrats on the first merged docs PR!

PR #6675, opened 2023-01-26
Fix typo in Fixing jitter, stutter and input lag

Calinou:
This also renames the Max FPS setting name to match Godot 4.0.


skyace65 (approved):
Looks good!

PR #6681, opened 2023-01-27
Update some of the C# documentation for 4.0

raulsntos:
- Minor updates to the C# basics page.
- Replace `yield` section with `await` section.
- Add mentions of some available BCL types that may be useful to users in relation to existing Godot types.
- Update `Variant` section.

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_basics.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_basics.rst
@@ -82,6 +82,7 @@ external editor of choice. Godot currently supports the following
 external editors:
 
 - Visual Studio 2019
+- Visual Studio 2022
 - Visual Studio Code
 - MonoDevelop
 - Visual Studio for Mac
@@ -187,10 +188,11 @@ Project setup and workflow
 When you create the first C# script, Godot initializes the C# project files
 for your Godot project. This includes generating a C# solution (``.sln``)
 and a project file (``.csproj``), as well as some utility files and folders
-(``.mono`` and ``Properties/AssemblyInfo.cs``).
-All of these but ``.mono`` are important and should be committed to your
-version control system. ``.mono`` can be safely added to the ignore list of your VCS.
-When troubleshooting, it can sometimes help to delete the ``.mono`` folder
+(``.godot/mono``).
+All of these but ``.godot/mono`` are important and should be committed to your
+version control system. Everything under ``.godot`` can be safely added to the
+ignore list of your VCS.
+When troubleshooting, it can sometimes help to delete the ``.godot/mono`` folder
 and let it regenerate.
 
 Example
@@ -248,8 +250,9 @@ For more information, see the :ref:`doc_c_sharp_differences` page.
     You need to (re)build the project assemblies whenever you want to see new
     exported variables or signals in the editor. This build can be manually
     triggered by clicking the word **Build** in the top right corner of the
-    editor. You can also click **Mono** at the bottom of the editor window
-    to reveal the Mono panel, then click the **Build Project** button.
+    editor. You can also click **MSBuild** at the bottom of the editor window
+    to reveal the MSBuild panel, then click the **Build** button to reveal a
+    dropdown, then click the **Build Solution** option.
 
     You will also need to rebuild the project assemblies to apply changes in
     "tool" scripts.
@@ -261,7 +264,7 @@ As C# support is quite new in Godot, there are some growing pains and things
 that need to be ironed out. Below is a list of the most important issues
 you should be aware of when diving into C# in Godot, but if in doubt, also
 take a look over the official
-`issue tracker for Mono issues <https://github.com/godotengine/godot/labels/topic%3Amono>`_.
+`issue tracker for .NET issues <https://github.com/godotengine/godot/labels/topic%3Adotnet>`_.
 
 - Writing editor plugins is possible, but it is currently quite convoluted.
 - State is currently not saved and restored when hot-reloading,
@@ -274,11 +277,13 @@ take a look over the official
   So when using e.g. ``CallDeferred("AddChild")``, ``AddChild`` will not work because
   the API is expecting the original ``snake_case`` version ``add_child``. However, you
   can use any custom properties or methods without this limitation.
+  Prefer using the exposed StringName in the ``PropertyName``, ``MethodName`` and
+  ``SignalName`` to avoid extra StringName allocations and worrying about snake_case naming.
 
 
-As of Godot 3.2.2, exporting Mono projects is supported for desktop platforms
-(Linux, Windows and macOS), Android, HTML5, and iOS. The only platform not
-supported yet is UWP.
+As of Godot 4.0, exporting .NET projects is supported for desktop platforms
+(Linux, Windows and macOS). Other platforms will gain support in future 4.x
+releases.
 
 Performance of C# in Godot
 --------------------------
@@ -299,7 +304,7 @@ a single code location:
 
     using Godot;
 
-    public class YourCustomClass : Node3D
+    public partial class YourCustomClass : Node3D
     {
         private void ExpensiveReposition()
         {
--- a/tutorials/scripting/c_sharp/c_sharp_basics.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_basics.rst
@@ -277,8 +277,8 @@ take a look over the official
   So when using e.g. ``CallDeferred("AddChild")``, ``AddChild`` will not work because
   the API is expecting the original ``snake_case`` version ``add_child``. However, you
   can use any custom properties or methods without this limitation.
-  Prefer using the exposed StringName in the ``PropertyName``, ``MethodName`` and
-  ``SignalName`` to avoid extra StringName allocations and worrying about snake_case naming.
+  Prefer using the exposed ``StringName`` in the ``PropertyName``, ``MethodName`` and
+  ``SignalName`` to avoid extra ``StringName`` allocations and worrying about snake_case naming.
 
 
 As of Godot 4.0, exporting .NET projects is supported for desktop platforms
--- a/tutorials/scripting/c_sharp/c_sharp_differences.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_differences.rst
@@ -494,15 +494,36 @@ Communicating with other scripting languages
 
 This is explained extensively in :ref:`doc_cross_language_scripting`.
 
-.. _doc_c_sharp_differences_yield:
+.. _doc_c_sharp_differences_await:
 
-Yield
------
+Await keyword
+-------------
 
-Something similar to GDScript's ``yield`` with a single parameter can be achieved with
-C#'s `yield keyword <https://docs.microsoft.com/en-US/dotnet/csharp/language-reference/keywords/yield>`_.
+Something similar to GDScript's ``await`` keyword can be achieved with C#'s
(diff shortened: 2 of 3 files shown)


paulloz:
> - There are some methods such as ``Get()``/``Set()``, ``Call()``/``CallDeferred()`` and signal connection method ``Connect()`` that rely on Godot's ``snake_case`` API naming conventions.

Is this still the case with `SignalName` and `Callable`?

raulsntos:
Well, the `StringName`s in SignalName, MethodName and PropertyName contain the snake_case name. When using these methods (or `Callable`) to invoke one of Godot's native methods you still need to use their native name which uses snake_case, and if you use string literals that's how you'd have to use them. The SignalName class is just a convenience that hides that so you don't have to worry about it but that doesn't change how these methods work.

I did add a note about using SignalName and the other classes to recommend users to use it over string literals or nameof, do you think it's not enough? Maybe it's not properly explained?

paulloz:
> Well, the `StringName`s in SignalName, MethodName and PropertyName contain the snake_case name.

Yeah, I meant using string literals directly.

> I did add a note about using SignalName and the other classes to recommend users to use it over string literals or nameof, do you think it's not enough? Maybe it's not properly explained?

Yeah, I saw that. My worry is quite often people just end up using `.Get` and `.Set` for everything. But this is probably a wider issue, IDK.

raulsntos:
I see people using those methods when they really shouldn't, we could hide them with EditorBrowsable Never but that would make it harder for those that actually need them. We could also add an analyzer that warns users using string literals instead of StringName, that may help.

paulloz:
That's a superb idea.

tutorials/scripting/c_sharp/c_sharp_basics.rst
```diff
@@ -274,11 +277,13 @@ take a look over the official
   So when using e.g. ``CallDeferred("AddChild")``, ``AddChild`` will not work because
   the API is expecting the original ``snake_case`` version ``add_child``. However, you
   can use any custom properties or methods without this limitation.
+  Prefer using the exposed StringName in the ``PropertyName``, ``MethodName`` and
+  ``SignalName`` to avoid extra StringName allocations and worrying about snake_case naming.
```
YuriSizov, line 281:
```suggestion
  Prefer using the exposed ``StringName`` in the ``PropertyName``, ``MethodName`` and
  ``SignalName`` to avoid extra ``StringName`` allocations and worrying about snake_case naming.
```

Or is there a reason to avoid formatting it here?

tutorials/scripting/c_sharp/c_sharp_differences.rst
```diff
@@ -46,17 +45,29 @@ Math global functions, like ``abs``, ``acos``, ``asin``, ``atan`` and ``atan2``,
 located under ``Mathf`` as ``Abs``, ``Acos``, ``Asin``, ``Atan`` and ``Atan2``.
 The ``PI`` constant can be found as ``Mathf.Pi``.
 
+C# also provides the static `System.Math`_ and `System.MathF`_ classes that may
```
YuriSizov, line 48:
```suggestion
C# also provides static `System.Math`_ and `System.MathF`_ classes that may
```

tutorials/scripting/c_sharp/c_sharp_differences.rst
```diff
@@ -115,6 +126,8 @@ Example:
         private string _icon;
     }
 
+See also: :ref:`doc_c_sharp_exports`.
+
 Signal keyword
 --------------
```
YuriSizov, line 132:
```suggestion
``signal`` keyword
------------------
```

Should we do it like this for consistency with the annotation formatting?

tutorials/scripting/c_sharp/c_sharp_differences.rst
```diff
@@ -87,10 +98,10 @@ Example:
         }
     }
 
-Export keyword
---------------
+`@export` annotation
+--------------------
```
YuriSizov, line 102:
```suggestion
``@export`` annotation
----------------------
```

tutorials/scripting/c_sharp/c_sharp_differences.rst
```diff
@@ -486,22 +499,52 @@ Use the ``Godot.Collections.Dictionary<T>(Godot.Collections.Dictionary)`` constr
...
 
 This is explained extensively in :ref:`doc_cross_language_scripting`.
 
-.. _doc_c_sharp_differences_yield:
+.. _doc_c_sharp_differences_await:
 
-Yield
------
+Await keyword
+-------------
```
YuriSizov, line 521: Same concern as with the signal.

```suggestion
``await`` keyword
-----------------
```

tutorials/scripting/c_sharp/c_sharp_differences.rst
```diff
@@ -486,22 +499,52 @@ Use the ``Godot.Collections.Dictionary<T>(Godot.Collections.Dictionary)`` constr
 Variant
 -------
 
-``System.Object`` (``object``) is used instead of ``Variant``.
+``Godot.Variant`` is used as the variant type, any variant-compatible type can be converted from/to it.
+Consider avoiding ``Godot.Variant`` and take advantage of C#'s type safety when possible,
+only use ``Godot.Variant`` when necessary to use Godot untyped APIs.
```
YuriSizov, line 504:
```suggestion
``Godot.Variant`` is used to represent the Godot's native :ref:`Variant <doc_variant_class>` type. Any Variant-compatible type can be converted from/to it.
We recommend avoiding ``Godot.Variant`` unless it is necessary to interact with untyped engine APIs.
Take advantage of C#'s type safety when possible.
```

tutorials/scripting/c_sharp/c_sharp_differences.rst
```diff
@@ -486,22 +499,52 @@ Use the ``Godot.Collections.Dictionary<T>(Godot.Collections.Dictionary)`` constr
...
+
+The ``Variant.As{TYPE}`` methods or the generic ``Variant.As<T>`` method can be used to convert
+a ``Godot.Variant`` to a C# type. Since the ``Godot.Variant`` type contains implicit conversions
+defined for all the supported types calling these methods directly is usually not necessary.
```
YuriSizov, line 508:
```suggestion
Any of ``Variant.As{TYPE}`` methods and a generic ``Variant.As<T>`` method can be used to convert
a ``Godot.Variant`` to a C# type. Since the ``Godot.Variant`` type contains implicit conversions
defined for all the supported types calling these methods directly is usually not necessary.
```

tutorials/scripting/c_sharp/c_sharp_differences.rst
```diff
@@ -486,22 +499,52 @@ Use the ``Godot.Collections.Dictionary<T>(Godot.Collections.Dictionary)`` constr
...
+
+Use the ``CreateFrom`` method overloads or the generic ``From<T>`` method to convert a C# type
+to a ``Godot.Variant``.
```
YuriSizov, line 511:
```suggestion
Use ``CreateFrom`` method overloads or a generic ``From<T>`` method to convert a C# type
to a ``Godot.Variant``.
```

tutorials/scripting/c_sharp/c_sharp_differences.rst
```diff
@@ -486,22 +499,52 @@ Use the ``Godot.Collections.Dictionary<T>(Godot.Collections.Dictionary)`` constr
...
-.. _doc_c_sharp_differences_yield:
+.. _doc_c_sharp_differences_await:
 
-Yield
------
+Await keyword
+-------------
+
+Something similar to GDScript's ``await`` keyword can be achieved with C#'s
+`await keyword <https://docs.microsoft.com/en-US/dotnet/csharp/language-reference/keywords/await>`_.
```
YuriSizov, line 524:
```suggestion
Something similar to the GDScript's ``await`` keyword can be achieved with the C#'s
`await keyword <https://docs.microsoft.com/en-US/dotnet/csharp/language-reference/keywords/await>`_.
```

tutorials/scripting/c_sharp/c_sharp_differences.rst
```diff
@@ -486,22 +499,52 @@ Use the ``Godot.Collections.Dictionary<T>(Godot.Collections.Dictionary)`` constr
...
+
+The C# ``await`` keyword can be used with any awaitable expression, it's commonly
```
YuriSizov, line 526:
```suggestion
The ``await`` keyword in C# can be used with any awaitable expression. It's commonly
```

tutorials/scripting/c_sharp/c_sharp_differences.rst
```diff
@@ -486,22 +499,52 @@ Use the ``Godot.Collections.Dictionary<T>(Godot.Collections.Dictionary)`` constr
...
+  * ``A`` has an accessible, readable instance property ``IsCompleted`` of type ``bool``.
+  * ``A`` has an accessible instance method ``GetResult`` with no parameters and no type
+    parameters.
+
+.. _Task: https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task
+.. _Task<TResult>: https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task-1
+.. _ValueTask: https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.valuetask
+.. _ValueTask<TResult>: https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.valuetask-1
+
+The equivalent of await on signal can be achieved with the ``await`` keyword and
```
YuriSizov, line 546:
```suggestion
An equivalent of awaiting a signal in GDScript can be achieved with the ``await`` keyword and
```

YuriSizov (commented):
Various corrections and nitpicks :)

tutorials/scripting/c_sharp/c_sharp_basics.rst
raulsntos: Not really, I probably thought it looked too crowded.

tutorials/scripting/c_sharp/c_sharp_differences.rst
raulsntos: Makes sense to me.

tutorials/scripting/c_sharp/c_sharp_differences.rst
raulsntos: This doesn't sound right to me, there is only one `Variant.As<T>` method so I feel like using `a` is wrong here.

tutorials/scripting/c_sharp/c_sharp_differences.rst
raulsntos: For some reason it sounds weird to me to use the `the` article with `'s`. But if you think it's more correct like this.

YuriSizov: I guess if it's apostrophe'd, then we don't need an article. You can ignore this.

tutorials/scripting/c_sharp/c_sharp_differences.rst
YuriSizov: For some reason it trips me here, but I guess this is because I intuitively read "generic" in its literal sense, not as a CS term :) Sure, you can keep the article for the generic method (and the same for the snippet below). (but the rest of the suggestion stands).

YuriSizov (approved):

mhilbrunner (approved):
Yay for C# updates! Thank you so much 🎉 

Everything looks good to me.

PR #6684, opened 2023-01-27
Add derivative with precision shader functions

Chaosus:
To reflect changes from https://github.com/godotengine/godot/pull/72109.

### DIFF
--- a/tutorials/shaders/shader_reference/shading_language.rst
+++ b/tutorials/shaders/shader_reference/shading_language.rst
@@ -1249,10 +1249,47 @@ is used, it can be scalar or vector.
 | vec4 **textureGather** (samplerCube s, vec3 p [, int comps])                |                                                                     |
 +-----------------------------------------------------------------------------+---------------------------------------------------------------------+
 | vec_type **dFdx** (vec_type p)                                              | Derivative in ``x`` using local differencing.                       |
+|                                                                             | Internally, can use either ``dFdxCoarse`` or ``dFdxFine``, but the  |
+|                                                                             | decision for which to use is made by the GPU driver.                |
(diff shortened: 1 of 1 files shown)


tutorials/shaders/shader_reference/shading_language.rst
```diff
@@ -1249,10 +1249,47 @@ is used, it can be scalar or vector.
 | vec4 **textureGather** (samplerCube s, vec3 p [, int comps])                |                                                                     |
 +-----------------------------------------------------------------------------+---------------------------------------------------------------------+
 | vec_type **dFdx** (vec_type p)                                              | Derivative in ``x`` using local differencing.                       |
+|                                                                             | Internally, it's using either ``dFdxCoarse`` or ``dFdxFine``        |
+|                                                                             | depending on the driver settings.                                   |
```
Calinou, line 1253: What's the default on most desktop/mobile hardware? Is there a way for users to override this in their driver options? (I don't remember seeing an option for derivatives specifically in the NVIDIA Control Panel.)

Calinou (approved):
Looks good to me. Should be good to merge after clarifying the `dFdx` typical expected behavior on most hardware.

tutorials/shaders/shader_reference/shading_language.rst
clayjohn: It can be anything. The specification allows hardware vendors to implement whichever they want. I imagine AMD and NVidia even have game-specific overrides. 

In light of that, we need to highlight for users how they can't rely on it being one or the other. I would go with something like:
```
Internally, can use either ``dFdxCoarse`` or ``dFdxFine``, but the decision for which to use is made by the GPU driver.
```

PR #6695, opened 2023-01-29
fix: replace usage of `File` with `FileAccess`

yedpodtrzitko:
- `File` is now replaced by `FileAccess`, update the examples accordingly

### DIFF
--- a/classes/class_audiostreammp3.rst
+++ b/classes/class_audiostreammp3.rst
@@ -130,11 +130,10 @@ You can load a file without having to import it beforehand using the code snippe
  .. code-tab:: gdscript
 
     func load_mp3(path):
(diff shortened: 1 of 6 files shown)


skyace65:
Two things. One, the class reference changes need to be removed from this PR. The class reference pages are generated based on what's in the main Godot repository, the files you want to change are [here](https://github.com/godotengine/godot/tree/master/doc/classes).

Two, there's already two open PR's making this change for the import plugins page so that needs to be removed. In the future please check open PR's to see if there's already one made for the change you're thinking of doing.

yedpodtrzitko:
ok my bad. I was looking for existing Issue/PR but I somehow missed it the first time.

PR #6696, opened 2023-01-29
Update and add C# RefCounted example in "Scenes versus Scripts"

Piralein:
`Reference` was renamed to `RefCounted`.
Added c# example
(I'm not sure if using `static` is prefered over using an instance.)

### DIFF
--- a/tutorials/best_practices/scenes_versus_scripts.rst
+++ b/tutorials/best_practices/scenes_versus_scripts.rst
@@ -39,19 +39,19 @@ a change in API:
 
     public partial class Game : Node
     {
-        public readonly Script MyNodeScr = (Script)ResourceLoader.Load("MyNode.cs");
-        public readonly PackedScene MySceneScn = (PackedScene)ResourceLoader.Load("MyScene.tscn");
-        public Node ANode;
-        public Node MyNode;
-        public Node MyScene;
-        public Node MyInheritedScene;
+        public static CSharpScript MyNode { get; } = GD.Load<CSharpScript>("MyNode.cs");
+        public static PackedScene MyScene { get; } = GD.Load<PackedScene>("MyScene.tscn");
+        private Node _node;
+        private Node _myNode;
+        private Node _myScene;
+        private Node _myInheritedScene;
 
         public Game()
         {
-            ANode = new Node();
-            MyNode = new MyNode(); // Same syntax
-            MyScene = MySceneScn.Instantiate(); // Different. Instantiated from a PackedScene
-            MyInheritedScene = MySceneScn.Instantiate(PackedScene.GenEditState.Main); // Create scene inheriting from MyScene
+            _node = new Node();
+            _myNode = MyNode.New().As<Node>();
+            _myScene = MyScene.Instantiate(); // Different. Instantiated from a PackedScene
+            _myInheritedScene = MyScene.Instantiate(PackedScene.GenEditState.Main); // Create scene inheriting from MyScene
         }
     }
 
@@ -170,7 +170,7 @@ with it, and finally adds it as a child of the ``Main`` node:
         {
             Child = new Node();
             Child.Name = "Child";
-            Child.Script = ResourceLoader.Load<Script>("child.gd");
+            Child.SetScript(GD.Load<Script>("child.gd"));
             Child.Owner = this;
             AddChild(Child);
         }
@@ -207,8 +207,8 @@ In the end, the best approach is to consider the following:
     .. code-tab:: gdscript GDScript
 
       # game.gd
-      class_name Game # extends Reference, so it won't show up in the node creation dialog
-      extends Reference
+      class_name Game # extends RefCounted, so it won't show up in the node creation dialog
+      extends RefCounted
 
       const MyScene = preload("my_scene.tscn")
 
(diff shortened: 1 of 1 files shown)


tutorials/best_practices/scenes_versus_scripts.rst
```diff
@@ -207,12 +207,29 @@ In the end, the best approach is to consider the following:
...
+      }
+      
+      //Main.cs
+      public partial class Main : Node
+      {
+          public override void _Ready()
+          {
+              AddChild(Game.MyScene.Instantiate(PackedScene.GenEditState.Main));
+          }
+      }
```
Calinou, line 235:
```suggestion
      // Game.cs
      public partial class Game : RefCounted
      {
          public static readonly PackedScene MyScene = (PackedScene)ResourceLoader.Load("MyScene.tscn");
      }
      
      // Main.cs
      public partial class Main : Node
      {
          public override void _Ready()
          {
              AddChild(Game.MyScene.Instantiate(PackedScene.GenEditState.Main));
          }
      }
```

Piralein:
I'm not sure what just happend, I didn't even had the review loaded, but it accepted it.
(squashed it)

Calinou:
> I'm not sure what just happend, I didn't even had the review loaded, but it accepted it. (squashed it)

That's because I applied the suggestion directly 🙂

tutorials/best_practices/scenes_versus_scripts.rst
```diff
@@ -207,12 +207,29 @@ In the end, the best approach is to consider the following:
...
+      {
+          public static readonly PackedScene MyScene = (PackedScene)ResourceLoader.Load("MyScene.tscn");
+      }
+      
+      // Main.cs
+      public partial class Main : Node
+      {
+          public override void _Ready()
+          {
+              AddChild(Game.MyScene.Instantiate(PackedScene.GenEditState.Main));
```
raulsntos, line 233: The GDScript example doesn't give a value to the parameter.

```suggestion
              AddChild(Game.MyScene.Instantiate());
```

tutorials/best_practices/scenes_versus_scripts.rst
```diff
@@ -207,12 +207,29 @@ In the end, the best approach is to consider the following:
...
       extends Node
       func _ready():
           add_child(Game.MyScene.instantiate())
+
+    .. code-tab:: csharp
+
+      // Game.cs
+      public partial class Game : RefCounted
+      {
+          public static readonly PackedScene MyScene = (PackedScene)ResourceLoader.Load("MyScene.tscn");
```
raulsntos, line 225: I think according to the guidelines this should be a property.

Also, we likely prefer `GD.Load` but I'm not so sure about this. I think we use both `ResourceLoader.Load` and `GD.Load` in the documentation, we should probably pick one and be consistent.

```suggestion
          public static PackedScene MyScene { get; } = GD.Load<PackedScene>("MyScene.tscn");
```

raulsntos (changes requested):

tutorials/best_practices/scenes_versus_scripts.rst
Piralein: I used the example above as template, should I change it there too?

raulsntos: Yes, if you could :pray:

Piralein:
Updated the examples above with `GD.Load` and `SetScript`.

tutorials/best_practices/scenes_versus_scripts.rst
```diff
@@ -39,8 +39,8 @@ a change in API:
 
     public partial class Game : Node
     {
-        public readonly Script MyNodeScr = (Script)ResourceLoader.Load("MyNode.cs");
-        public readonly PackedScene MySceneScn = (PackedScene)ResourceLoader.Load("MyScene.tscn");
+        public Script MyNodeScr { get; } = GD.Load<Script>("MyNode.cs");
+        public PackedScene MySceneScn { get; } = GD.Load<PackedScene>("MyScene.tscn");
```
raulsntos, line 43: Considering how these are `const` in GDScript, I would make these properties `static`.

I would also drop the `Scr`/`Scn` suffix, and make the fields below private and renamed them to use `_camelCase`.

```suggestion
        public static Script MyNode { get; } = GD.Load<Script>("MyNode.cs");
        public static PackedScene MyScene { get; } = GD.Load<PackedScene>("MyScene.tscn");
```

tutorials/best_practices/scenes_versus_scripts.rst
```diff
@@ -39,19 +39,19 @@ a change in API:
...
     {
-        public readonly Script MyNodeScr = (Script)ResourceLoader.Load("MyNode.cs");
-        public readonly PackedScene MySceneScn = (PackedScene)ResourceLoader.Load("MyScene.tscn");
-        public Node ANode;
-        public Node MyNode;
-        public Node MyScene;
-        public Node MyInheritedScene;
+        public static CSharpScript MyNode { get; } = GD.Load<CSharpScript>("MyNode.cs");
+        public static PackedScene MyScene { get; } = GD.Load<PackedScene>("MyScene.tscn");
+        private Node _aNode;
```
raulsntos, line 44: I don't know why the `A` prefix used to be there.

```suggestion
        private Node _Node;
```

raulsntos (approved):
LGTM, just a minor nit-pick.

mhilbrunner (approved):

YuriSizov:
Thanks!

PR #6701, opened 2023-01-30
Fix default deadzone value in Controllers, gamepads and joysticks

Calinou:


tutorials/inputs/controllers_gamepads_joysticks.rst
```diff
@@ -165,10 +165,11 @@ enough to ignore the input caused by joystick drifting, but is low enough to not
 ignore actual input from the player.
 
 Godot features a built-in dead zone system to tackle this problem. The default
-value is ``0.2``, but you can increase it or decrease it on a per-action basis
-in the Project Settings' Input Map tab.
-For ``Input.get_vector()``, the deadzone can be specified, or otherwise it
-will calculate the average deadzone value from all of the actions in the vector.
+value is tailored for "digital" actions and is ``0.5``, but you can adjust it on
```
skyace65, line 168: Two issues with this part. The phrase "digital actions" isn't used elsewhere in this page, a few paragraphs earlier mentions digital inputs which I assume is the same thing, and it would be best to be consistent. And two if this is tailored for digital actions is there a downside to lowering it for analog actions/inputs? Because that's the impression I get without more info.

skyace65 (changes requested):

tutorials/inputs/controllers_gamepads_joysticks.rst
Calinou: A digital action refers to using a gamepad axis for an action that is "binary" in nature, such as joystick navigation in menus.

skyace65: I'd say just remove the "is tailored for "digital" actions" bit then, I think that's only going to cause confusion.

mhilbrunner (approved):

PR #6707, opened 2023-01-31
Linking between First 2D Shader and Custom Drawing

runny-yolk:
I'm not new to graphics programming, but I am new to Godot's way of doing things, and I wanted to do some procedural shader graphics, so I used this tutorial as a jumping off point, and I thought it was great, aside from the fact that it doesn't tell you how to draw to the canvas without using a Sprite2D. It does say "you can use any CanvasItem" but not _how_ you can use any CanvasItem, which was fairly frustrating 😅 After an hour of searching I found the page I needed, and I figured since the docs are open source, I should link to it for other people in my situation. An additional benefit of linking those docs here is that users who are looking to increase their skills will find a convenient on-ramp.


clayjohn (approved):

skyace65 (approved):

skyace65:

PR #6710, opened 2023-01-31
Correct CanvasItem shader docs about AT_LIGHT_PASS builtin

clayjohn:
Partially addresses https://github.com/godotengine/godot/issues/72477

I am not sure if we will actually remove ``AT_LIGHT_PASS`` as it would break compatibility. I am updating the docs anyway to ensure they are accurate


mhilbrunner (approved):

mhilbrunner:
Thanks!

PR #6712, opened 2023-02-01
Fix method names in Navigation overview

KekLuck:
The Vector3 to Vector2 variable decleration threw an error and the set_target_location function seems to be called set_target_position now


mhilbrunner (approved):

paulloz:
Can we update C# at the same time?

Lines 160 and 161.
```
            get { return _navigationAgent.TargetPosition; }
            set { _navigationAgent.TargetPosition = value; }
```

YuriSizov (commented):
A couple more renames

YuriSizov:
Thanks, and congrats on the first merged PR!

PR #6371, opened 2022-11-08
Update Saving games documentation for Godot 4.0

xdl:
Updating saving games doc to Godot4 (including a best effort attempt at C# with dotnet6). Overview of changes:

* File → FileAcces methods
  * Removing call to `.close()` in GDScript and using `using` in C#
  * `get_len` → `get_length`
* `res.instance()` to `res.instantiate()`
* `node.filename` → `node.scene_file_path`
* (C#) JSON method update (Print → Stringify)
* (C#) Minor clarification on setting lowercase 'position' - seems to have tripped up people in the past in 3.x too: https://godotengine.org/qa/136145/c%23-problem-with-save-load-node-positions-set-to-0-0
* (GDScript) Adding the omitted `var json_string = save_game.get_line()`
* (C#) object → Variant when declaring a dictionary
  * Was done to get around the compile error: `The generic type argument must be a Variant compatible type: 'object'` - not sure if this is the correct way to go about it, if someone can confirm that'd be great

PR #6713, opened 2023-02-01
Updated for Godot 4 function calls

GinghamJones:
The current documentation doesn't work. This update adds changes to work with Godot 4. Note: I only updated for GDScript as I don't use C#


raulsntos:
Thanks for opening a pull request! There's already a pull request reworking the saving games manual: https://github.com/godotengine/godot-docs/pull/6371. In the future please check open PR's to see if there's already one made for the change you're thinking of doing.

Issue #6716, opened 2023-02-01
EditorGizmo Plugin register methods wrong in documentation (or engine)


Flavelius:
**Your Godot version:**
4.0beta16

**Issue description:**

https://github.com/godotengine/godot-docs/blob/master/tutorials/plugins/editor/3d_gizmos.rst

'add_node3d_gizmo_plugin' is actually 'add_node_3d_gizmo_plugin', which is a horrible name for a function/to type out. It's unclear to me whether the current (second) name really will stay that way (i hope not), but for now alteast the tutorial doesn't reflect the actual method's name.

**URL to the documentation page:**
(https://docs.godotengine.org/en/latest/tutorials/plugins/editor/3d_gizmos.html)

PR #6717, opened 2023-02-01
Fix wrong registration method names for 3D gizmo plugins

Flavelius:
As mentioned here https://github.com/godotengine/godot-docs/issues/6716 add/remove_node3d_gizmo_plugin is the wrong method name. But i'm not sure if it intentionally changed to add/remove_node_3d_gizmo_plugin. Here are the adjustments anway in case it is intentional and going to stay that way.

_Closes https://github.com/godotengine/godot-docs/issues/6716_

### DIFF
--- a/tutorials/plugins/editor/3d_gizmos.rst
+++ b/tutorials/plugins/editor/3d_gizmos.rst
@@ -49,11 +49,11 @@ This would be a basic setup:
 
 
     func _enter_tree():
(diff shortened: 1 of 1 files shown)


YuriSizov (approved):

YuriSizov:
Thanks, and congrats on the first merged documentation PR!

PR #6720, opened 2023-02-01
Clarify ternary operator in the style guide

notPelf:
In the current style guide, the example for ternary operators includes a `not` in the conditional which is the same text color as the preceding `if` giving the impression that the format for a ternary operation is `[some_value] if not [condition] else [other_value]` which is incorrect, and may be confusing to figure out. I'm new to Godot and I've never used python (which I understand GDScript is based on). My only experience is with c# which I don't believe has a `not` keyword so I was quite confused by this.

This change would remove `not` from the condition to eliminate confusion, and rearrange the values to keep the meaning of the example code the same. Any alternative that removes the `not` would also work.

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_styleguide.rst
+++ b/tutorials/scripting/gdscript/gdscript_styleguide.rst
@@ -318,7 +318,7 @@ The only exception to that rule is the ternary operator:
 
 ::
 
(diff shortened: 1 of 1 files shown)


YuriSizov (approved):

YuriSizov:
Thanks, and congrats on the first merged docs PR!

PR #6721, opened 2023-02-02
C#: Add table with equivalent GlobalScope methods

raulsntos:
- Follow up to https://github.com/godotengine/godot/pull/71946.
- Adds a table with the list of C# methods that are equivalent to the methods in Godot's Global Scope and GDScript utility functions to make it easier to users to port from GDScript and find which APIs are available.

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_differences.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_differences.rst
@@ -71,12 +71,14 @@ Example: ``GD.Print`` and ``GD.VarToStr``.
 
 Exceptions:
 
-===========================  =======================================================
-GDScript                     C#
-===========================  =======================================================
-``weakref(obj)``             ``GodotObject.WeakRef(obj)``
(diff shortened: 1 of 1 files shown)


mhilbrunner:
I agree this is useful, but I wonder if there's a better way to maintain this, because this looks like it could get outdated and would be a lot to maintain manually.

Also, not sure the flow "huh, wonder what this method is in C#" -> find this special page is the best. Is there a way we could add that as information to the specific GDScript functions? I.e., "this is <X> in C#"?

Not sure if that'd make sense. Probably also requires further work, so this one probably makes sense anyway. :)

raulsntos:
I see what you mean, and I agree the discoverability of this page is probably not great since it kinda requires users to already know what they are looking for.

We already have C# examples in the class reference of the Core documentation, hopefully that helps. And I expect more experienced users would link to this page to help beginners.

I agree this can be improved, my goal is to have all this information somewhere so that we can link to it in the release post for 4.0.

mhilbrunner (approved):

YuriSizov:
Thanks!

PR #6722, opened 2023-02-02
Document Signal and Callable types in C# differences page

raulsntos:
- Documents the differences between the `Signal` type in Core and the `Signal` struct in C#.
- Documents the differences between the `Callable` type in Core and the `Callable` struct in C#.

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_differences.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_differences.rst
@@ -421,7 +421,7 @@ GDScript              C#
 Currently C# supports ``Callable`` if one of the following holds:
 
 * ``Callable`` was created using the C# ``Callable`` type.
-* ``Callable`` is a basic version of the engine's ``Callable``. Custom ``Callable``s
+* ``Callable`` is a basic version of the engine's ``Callable``. Custom ``Callable``\ s
   are unsupported. A ``Callable`` is custom when any of the following holds:
 
   * ``Callable`` has bound information (``Callable``s created with ``bind``/``unbind`` are unsupported).
--- a/tutorials/scripting/c_sharp/c_sharp_differences.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_differences.rst
@@ -424,7 +424,7 @@ Currently C# supports ``Callable`` if one of the following holds:
 * ``Callable`` is a basic version of the engine's ``Callable``. Custom ``Callable``\ s
   are unsupported. A ``Callable`` is custom when any of the following holds:
 
-  * ``Callable`` has bound information (``Callable``s created with ``bind``/``unbind`` are unsupported).
+  * ``Callable`` has bound information (``Callable``\ s created with ``bind``/``unbind`` are unsupported).
   * ``Callable`` was created from other languages through the GDExtension API.
 
 Some methods such as ``bind`` and ``unbind`` are not implemented, use lambdas instead:
--- a/tutorials/scripting/c_sharp/c_sharp_differences.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_differences.rst
@@ -388,6 +388,59 @@ hex_encode                 StringExtensions.HexEncode (Consider using `System.Co
 .. _System.Uri.EscapeDataString: https://learn.microsoft.com/en-us/dotnet/api/system.uri.escapedatastring
 .. _System.Uri.UnescapeDataString: https://learn.microsoft.com/en-us/dotnet/api/system.uri.unescapedatastring
 
+Signal
+------
+
+The following methods were converted to properties with their respective names changed:
+
+====================  ==============================================================
+GDScript              C#
+====================  ==============================================================
+``get_name()``        ``Name``
+``get_object()``      ``Owner``
+====================  ==============================================================
+
+The ``Signal`` type implements the awaitable pattern which means it can be used with
+the ``await`` keyword. See :ref:`_doc_c_sharp_differences_await`.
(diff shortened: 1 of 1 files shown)


tutorials/scripting/c_sharp/c_sharp_differences.rst
```diff
@@ -388,6 +388,53 @@ hex_encode                 StringExtensions.HexEncode (Consider using `System.Co
...
+The following methods were converted to properties with their respective names changed:
+
+====================  ==============================================================
+GDScript              C#
+====================  ==============================================================
+``get_object()``      ``Target``
+``get_method()``      ``Method``
+====================  ==============================================================
+
+Currently C# doesn't support ``CustomCallable``s which means ``Callable``s coming from
```
paulloz, line 421: Is that so? I've been calling `Callable`s defined in GDScript without issue 🤔

raulsntos: Have you tried a GDScript `Callable` with bound variables?

paulloz: Indeed, doesn't work with bound variables. But this line sounds like it doesn't work, whatever the case is. It's fine if we want to document it as generally not working or "unknown behaviour" to keep the doc clean, though. People who want to try it still can.

paulloz: Yeah, I thought about it for the past hour and to be honest I don't really have a better suggestion.

raulsntos: How about this?

> Currently C# only supports ``Callable``s created in C# as well, ``Callable``s coming from other languages may not work.

Honestly, it doesn't seem much better to me.

paulloz: Yeah, potayto, potahto.

YuriSizov: I'm concerned about the uncertainty in either phrasing. Do we not know what would happen?

paulloz: > Do we not know what would happen?

As far as I understand, "basic" `Callable`s aren't an issue. Only once you start binding arguments to them will it not work.

YuriSizov: Then we should state exactly that, clearly, I believe.

raulsntos: Yes, "basic" `Callable`s work. What doesn't work are `CallableCustom` implementations such as `CallableCustomBind` or `CallableCustomUnbind` (can be created in GDScript using the `bind` and `unbind` methods); or any other implementation such as one provided by a GDExtension (e.g.: a Rust `Callable`).

I wanted to avoid mentioning bounded `Callable`s specifically because those aren't the only type of `Callable` that doesn't work (any `CallableCustom` won't work other than `ManagedCallable`, which is the C# one).

But I've been struggling to think of a better way to express all of this in a concise manner and without scaring users about using `Callable` at all since "basic" ones still work. I'm open to suggestions.

YuriSizov: Then I think the problem with your current phrasing is that "CustomCallable" is never explained in the article, nor linked externally. I think that you should state these things:

1. C# supports its own native implementation of `Callable` called `ManagedCallable`.
2. C# supports the basic version of the engine's `Callable`, i.e. a method without any bound information.
3. C# does not support engine's `Callable`s with bound information, such as created through `bind`/`unbind`.
4. C# is unlikely to work with implementations of `Callable`s from other languages through the GDExtension API.

tutorials/scripting/c_sharp/c_sharp_differences.rst
```diff
@@ -388,6 +388,59 @@ hex_encode                 StringExtensions.HexEncode (Consider using `System.Co
...
+GDScript              C#
+====================  ==============================================================
+``get_object()``      ``Target``
+``get_method()``      ``Method``
+====================  ==============================================================
+
+Currently C# supports ``Callable`` if one of the following holds:
+
+* ``Callable`` was created using the C# ``Callable`` type.
+* ``Callable`` is a basic version of the engine's ``Callable``. Custom ``Callable``s
```
YuriSizov, line 424:
```suggestion
* ``Callable`` is a basic version of the engine's ``Callable``. Custom ``Callable``\ s
```

tutorials/scripting/c_sharp/c_sharp_differences.rst
```diff
@@ -388,6 +388,59 @@ hex_encode                 StringExtensions.HexEncode (Consider using `System.Co
...
+``get_method()``      ``Method``
+====================  ==============================================================
+
+Currently C# supports ``Callable`` if one of the following holds:
+
+* ``Callable`` was created using the C# ``Callable`` type.
+* ``Callable`` is a basic version of the engine's ``Callable``. Custom ``Callable``\ s
+  are unsupported. A ``Callable`` is custom when any of the following holds:
+
+  * ``Callable`` has bound information (``Callable``s created with ``bind``/``unbind`` are unsupported).
```
YuriSizov, line 427:
```suggestion
  * ``Callable`` has bound information (``Callable``\ s created with ``bind``/``unbind`` are unsupported).
```

YuriSizov (approved):

YuriSizov:
Thanks!

PR #6730, opened 2023-02-03
Add a dedicated page about the C# Variant type

raulsntos:
- Clarifies which C# types are compatible with Godot's `Variant`.
- Explains how to use the C# `Variant` struct in a generic context with the `[MustBeVariant]` attribute.
- Adds some notes about some limitations when using the C# `Variant` types (enums, null).

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_differences.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_differences.rst
@@ -541,16 +541,9 @@ operator []              Dictionary[Variant] indexer, Add or TryGetValue
 Variant
 -------
 
-``Godot.Variant`` is used to represent the Godot's native :ref:`Variant <doc_variant_class>` type. Any Variant-compatible type can be converted from/to it.
-We recommend avoiding ``Godot.Variant`` unless it is necessary to interact with untyped engine APIs.
-Take advantage of C#'s type safety when possible.
+``Godot.Variant`` is used to represent Godot's native :ref:`Variant <class_Variant>` type. Any Variant-compatible type can be converted from/to it.
 
-Any of ``Variant.As{TYPE}`` methods or the generic ``Variant.As<T>`` method can be used to convert
-a ``Godot.Variant`` to a C# type. Since the ``Godot.Variant`` type contains implicit conversions
-defined for all the supported types calling these methods directly is usually not necessary.
-
-Use ``CreateFrom`` method overloads or the generic ``From<T>`` method to convert a C# type
-to a ``Godot.Variant``.
+See also: :ref:`doc_c_sharp_variant`.
 
 Communicating with other scripting languages
 --------------------------------------------
--- a/tutorials/scripting/c_sharp/c_sharp_variant.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_variant.rst
@@ -11,7 +11,7 @@ Take advantage of C#'s type safety when possible.
 
 Any of ``Variant.As{TYPE}`` methods or the generic ``Variant.As<T>`` method can be used to convert
 a ``Godot.Variant`` to a C# type. Since the ``Godot.Variant`` type contains implicit conversions
-defined for all the supported types calling these methods directly is usually not necessary.
+defined for all the supported types, calling these methods directly is usually not necessary.
 
 Use ``CreateFrom`` method overloads or the generic ``Variant.From<T>`` method to convert a C# type
 to a ``Godot.Variant``.
@@ -80,7 +80,7 @@ Variant.Type             C# Type
     Godot uses 64-bit integers and floats in Variant. Smaller integer and float types
     such as ``int``, ``short`` and ``float`` are supported since they can fit in the
     bigger type. Be aware that an implicit conversion is performed so using the wrong
-    type will result in potential precission loss.
+    type will result in potential precision loss.
 
 .. warning::
 
--- a/tutorials/scripting/c_sharp/c_sharp_variant.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_variant.rst
@@ -1,0 +1,156 @@
+.. _doc_c_sharp_variant:
+
+C# Variant
+==========
+
+For a detailed explanation of Variant in general, see the :ref:`Variant <class_Variant>` documentation page.
+
+``Godot.Variant`` is used to represent Godot's native :ref:`Variant <class_Variant>` type. Any Variant-compatible type can be converted from/to it.
+We recommend avoiding ``Godot.Variant`` unless it is necessary to interact with untyped engine APIs.
+Take advantage of C#'s type safety when possible.
+
+Any of ``Variant.As{TYPE}`` methods or the generic ``Variant.As<T>`` method can be used to convert
+a ``Godot.Variant`` to a C# type. Since the ``Godot.Variant`` type contains implicit conversions
+defined for all the supported types calling these methods directly is usually not necessary.
+
+Use ``CreateFrom`` method overloads or the generic ``Variant.From<T>`` method to convert a C# type
+to a ``Godot.Variant``.
+
+.. note::
+
+    Since the Variant type in C# is a struct, it can't be null. To create a "null"
+    Variant use the ``default`` keyword or the parameterless constructor.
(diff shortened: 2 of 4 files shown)


tutorials/scripting/c_sharp/c_sharp_variant.rst
```diff
@@ -0,0 +1,156 @@
...
+Use ``CreateFrom`` method overloads or the generic ``Variant.From<T>`` method to convert a C# type
+to a ``Godot.Variant``.
+
+.. note::
+
+    Since the Variant type in C# is a struct, it can't be null. To create a "null"
+    Variant use the ``default`` keyword or the parameterless constructor.
+
+Variant-compatible types
+^^^^^^^^^^^^^^^^^^^^^^^^
```
raulsntos, line 25: This isn't right, I built locally and this uses `h2` but I want it to use `h3`. The next heading `Using Variant in a generic context` should use `h2` but it currently uses `h3`.

tutorials/scripting/c_sharp/c_sharp_variant.rst
```diff
@@ -0,0 +1,156 @@
+.. _doc_c_sharp_variant:
+
+C# Variant
+==========
+
+For a detailed explanation of Variant in general, see the :ref:`doc_variant_class` section in the core and modules tutorial.
+
+``Godot.Variant`` is used to represent the Godot's native :ref:`Variant <doc_variant_class>` type. Any Variant-compatible type can be converted from/to it.
```
mhilbrunner, line 8:
```suggestion
``Godot.Variant`` is used to represent Godot's native :ref:`Variant <doc_variant_class>` type. Any Variant-compatible type can be converted from/to it.
```

tutorials/scripting/c_sharp/c_sharp_variant.rst
```diff
@@ -0,0 +1,156 @@
...
+We recommend avoiding ``Godot.Variant`` unless it is necessary to interact with untyped engine APIs.
+Take advantage of C#'s type safety when possible.
+
+Any of ``Variant.As{TYPE}`` methods or the generic ``Variant.As<T>`` method can be used to convert
+a ``Godot.Variant`` to a C# type. Since the ``Godot.Variant`` type contains implicit conversions
+defined for all the supported types calling these methods directly is usually not necessary.
```
mhilbrunner, line 14:
```suggestion
defined for all the supported types, calling these methods directly is usually not necessary.
```

tutorials/scripting/c_sharp/c_sharp_variant.rst
```diff
@@ -0,0 +1,156 @@
...
+Variant-compatible types
+------------------------
+
+* All the `built-in value types <https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/built-in-types-table>`_,
+  except ``decimal``, ``nint`` and ``nuint``.
+* ``string``.
+* Classes derived from :ref:`GodotObject <class_Object>`.
+* Collections types defined in the ``Godot.Collections`` namespace.
+
+Full list of Variant types and their equivalent C# type:
```
mhilbrunner, line 33: Is there a way to view this list in the class reference? Maintaining such a list manually is prone to being outdated, in my experience.

tutorials/scripting/c_sharp/c_sharp_variant.rst
```diff
@@ -0,0 +1,156 @@
...
+``PackedVector3Array``   ``Godot.Vector3[]``
+``PackedColorArray``     ``Godot.Color[]``
+=======================  ===========================================================
+
+.. warning::
+
+    Godot uses 64-bit integers and floats in Variant. Smaller integer and float types
+    such as ``int``, ``short`` and ``float`` are supported since they can fit in the
+    bigger type. Be aware that an implicit conversion is performed so using the wrong
+    type will result in potential precission loss.
```
mhilbrunner, line 83:
```suggestion
    type will result in potential precision loss.
```

tutorials/scripting/c_sharp/c_sharp_variant.rst
raulsntos: Considering how resistant we are to adding new types to `Variant` I don't think it's likely to get outdated.

I understand your concern though, but I'd rather not separate the list from the rest of the C# Variant documentation. I think it's convenient to have a single page that explains how the `Variant` type works in C#, together with the list of "Variant-compatible types" so that whenever we use that phrase users can tell what we mean.

tutorials/scripting/c_sharp/c_sharp_variant.rst
```diff
@@ -0,0 +1,156 @@
+.. _doc_c_sharp_variant:
+
+C# Variant
+==========
+
+For a detailed explanation of Variant in general, see the :ref:`doc_variant_class` section in the core and modules tutorial.
```
raulsntos, line 6: Is this the right Variant page to link to? I noticed there's https://docs.godotengine.org/en/latest/classes/class_variant.html and https://docs.godotengine.org/en/latest/contributing/development/core_and_modules/variant_class.html.

mhilbrunner: This should link to the latter.

YuriSizov: I think it should link to the former. The class reference page already links to the latter.

YuriSizov: Though I guess, it doesn't matter much. Both pages are kind of bare.

raulsntos: Currently it links to the latter, I think I'm leaning towards linking to the former instead like @YuriSizov suggests.

YuriSizov: Hard to make a meaningful distinction between the two :D But if I have to, former would be for more user-oriented explanation, and latter for more low-level dev explanation. So for general user documentation, class references should be preferred, unless this part specifically wants to point to the implementation detail.

Keep in mind the text should also be adjusted if you change the link.

YuriSizov: The line below still points to the engine doc. Which would be very confusing to the user, since they would appear visually the same :P

raulsntos: I think it makes more sense to link to the class reference then. I also opened https://github.com/godotengine/godot/pull/72721 to update it though because the C# examples are outdated.

tutorials/scripting/c_sharp/c_sharp_differences.rst
```diff
@@ -542,15 +542,8 @@ Variant
 -------
 
 ``Godot.Variant`` is used to represent the Godot's native :ref:`Variant <doc_variant_class>` type. Any Variant-compatible type can be converted from/to it.
```
YuriSizov, line 544: So this needs to have its link updated.

tutorials/scripting/c_sharp/c_sharp_variant.rst
```diff
@@ -0,0 +1,156 @@
+.. _doc_c_sharp_variant:
+
+C# Variant
+==========
+
+For a detailed explanation of Variant in general, see the :ref:`Variant <class_Variant>` documentation page.
+
+``Godot.Variant`` is used to represent Godot's native :ref:`Variant <doc_variant_class>` type. Any Variant-compatible type can be converted from/to it.
```
YuriSizov, line 8: And this as well.

raulsntos: I'll remove the word "native" then, right?

YuriSizov: I think "native" here makes sense as in "native to the engine", but if you think it's confusing in any way, feel free to drop it.

raulsntos: Ah, I can see that. It's probably less confusing **with** "native".

YuriSizov (approved):

YuriSizov:
I think we got there :) Thanks!

PR #6738, opened 2023-02-06
Update paths in preload call in making_plugins.rst

michaeltintiuc:
If not using the the full path Godot will fail to initialize the plugin and show an error message.


YuriSizov:
Which error are you seeing? Preload supports relative paths.

michaeltintiuc:
nvm, I think it might be a bug with latest beta versions, it seems to only happen when clearing the .godot directory. The code seems to run fine though.

PR #6748, opened 2023-02-08
Update Using fonts documentation for 4.0.beta17

Calinou:
- Mention subpixel positioning should be disabled for fonts with a pixel art appearance.
- Update bitmap font import dock image and document image/character margin.
- Fixed character range example to match the font image (it previously had to be reduced by 1 due to an off-by-one error in the importer code).

This closes https://github.com/godotengine/godot/issues/70327.

@bruvzg I've noticed that I need to set the character range to `0-126` and not `0-127` to get the example font to import correctly. Otherwise, I get an "out of range" error (the font has exactly 128 glyphs - 16 columns and 8 rows). Is that expected?

**Edit:** I updated the documentation according to https://github.com/godotengine/godot-docs/pull/6748#issuecomment-1422473126.


mhilbrunner (approved):

bruvzg:
> I've noticed that I need to set the character range to 0-126 and not 0-127 to get the example fontto import correctly. Otherwise, I get an "out of range" error (the font has exactly 128 glyphs - 16 columns and 8 rows). Is that expected?

Bug (https://github.com/godotengine/godot/pull/72887).

bruvzg (approved):

akien-mga:
Thanks!

PR #6753, opened 2023-02-09
Update Mono references to .NET

skyace65:
Replaces mono with .NET, changes the C# version to 10 on the list of features page. Changes supported platforms from "all platforms" to "Windows, Linux and macOS"

@paulloz is this a good way to phrase that we use .NET 6? I know you can use the 7 SDK but it's limited to .NET 6 features

### DIFF
--- a/about/list_of_features.rst
+++ b/about/list_of_features.rst
@@ -412,11 +412,11 @@ Scripting
 :ref:`C#: <toc-learn-scripting-C#>`
 
 - Packaged in a separate binary to keep file sizes and dependencies down.
-- Uses Mono 6.x.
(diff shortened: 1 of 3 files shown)


mhilbrunner (approved):

raulsntos:
Yes, our assemblies target .NET 6. Users are able to use a newer .NET SDK but we can't guarantee it will work properly.

raulsntos (approved):

skyace65:
I've changed it from "supports all platforms" to "supports Windows, Linux and macOS"

mhilbrunner:
Thank you!

PR #6765, opened 2023-02-11
Updated C# example in CharacterBody2D

constantitus:
The "collider" method in KinematicCollision2D has been changed to get_collider() .
Updated the C# example in CharacterBody2D.


mhilbrunner:
Hey! Thanks for helping and contributing :) Changes to the class reference need to made in the [main repository](https://github.com/godotengine/godot/blob/master/doc/classes/CharacterBody2D.xml), everything under `classes/` is autogenerated from there. Only the actual manual and other content of the docs are maintained here, not the class reference.

PR #6766, opened 2023-02-11
Fix errors in navigation_layers bitmask example

smix8:
Fixes errors in navigation_layers example like wrong property names or missing string closure.


mhilbrunner (approved):

mhilbrunner:
Thank you!

PR #6771, opened 2023-02-13
Fix navigation basic setup confusing instructions / images

smix8:
Fixes some confusing images and instructions in the basic setup section as it is not required to place the character below a NavigationRegion node, it can be placed anywhere.


mhilbrunner (approved):

mhilbrunner:
Thanks, merged! Also thanks for taking the chance to update those images to WebP 🎉

PR #6775, opened 2023-02-14
Update xr_action_map.rst

jwmcgettigan:
This PR is essentially proofreading: spelling and grammar fixes along with some rewording.

@BastiaanOlij One other thing that I wasn't sure about:

> You might also want to remove the controllers that you do not wish to setup, more on this later.

It's not clear to me where this is referring to - I think it is the note at the very bottom of the page?

### DIFF
--- a/tutorials/xr/xr_action_map.rst
+++ b/tutorials/xr/xr_action_map.rst
@@ -135,11 +135,10 @@ For instance, we can define an action ``Shoot`` and the input bound to that acti
 You can also poll the current state of an action.
 :ref:`XRController3D <class_xrcontroller3d>` for instance has an ``is_button_pressed`` method.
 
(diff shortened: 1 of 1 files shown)


BastiaanOlij (approved):
Thanks! Looks great!

BastiaanOlij:
> > You might also want to remove the controllers that you do not wish to setup, more on this later.
> 
> It's not clear to me where this is referring to - I think it is the note at the very bottom of the page?

When you remove the action set, it doesn't change the controllers that have been added to the action map. As the default action map currently contains all controller bindings while the advise is to only include controllers you are actually developing with, it is wise to remove them.

I really need to add a CLEAR ALL button to the interface.

skyace65 (approved):
Looks good!

PR #6778, opened 2023-02-14
Add C# Examples for ArrayMesh Tutorial

### LINKED ISSUE #6714: Missing mono documentation for ArrayMesh page
**Your Godot version:**
Godot 4 beta mono edition

**Issue description:**
ArrayMesh documentation is missing C# counterpart

https://docs.godotengine.org/en/latest/tutorials/3d/procedural_geometry/arraymesh.html

markawilson:
Resolves #6714. Although the equivalent C# type to GDScript's PackedVector\*Array is Vector\*[], I've used List<Vector\*> and converted with ToArray(), since the arrays have to resize dynamically as mesh data is added. Happy to revise this if there's any objection to that.

### DIFF
--- a/tutorials/3d/procedural_geometry/arraymesh.rst
+++ b/tutorials/3d/procedural_geometry/arraymesh.rst
@@ -140,7 +140,8 @@ by adding each array to ``surface_array`` and then committing to the mesh.
     surface_array[Mesh.ARRAY_NORMAL] = normals
     surface_array[Mesh.ARRAY_INDEX] = indices
 
-    mesh.add_surface_from_arrays(Mesh.PRIMITIVE_TRIANGLES, surface_array) # No blendshapes, lods, or compression used.
+    # No blendshapes, lods, or compression used.
+    mesh.add_surface_from_arrays(Mesh.PRIMITIVE_TRIANGLES, surface_array)
 
  .. code-tab:: csharp C#
 
@@ -157,7 +158,8 @@ by adding each array to ``surface_array`` and then committing to the mesh.
     var arrMesh = Mesh as ArrayMesh;
     if (arrMesh != null)
     {
-        arrMesh.AddSurfaceFromArrays(Mesh.PrimitiveType.Triangles, surfaceArray); // No blendshapes, lods, or compression used.
+        // No blendshapes, lods, or compression used.
+        arrMesh.AddSurfaceFromArrays(Mesh.PrimitiveType.Triangles, surfaceArray); 
     }
 
 .. note:: In this example, we used ``Mesh.PRIMITIVE_TRIANGLES``, but you can use any primitive type
@@ -191,7 +193,8 @@ Put together, the full code looks like:
         surface_array[Mesh.ARRAY_INDEX] = indices
 
         # Create mesh surface from mesh array.
-        mesh.add_surface_from_arrays(Mesh.PRIMITIVE_TRIANGLES, surface_array) # No blendshapes, lods, or compression used.
+        # No blendshapes, lods, or compression used.
+        mesh.add_surface_from_arrays(Mesh.PRIMITIVE_TRIANGLES, surface_array)
 
  .. code-tab:: csharp C#
 
@@ -228,7 +231,8 @@ Put together, the full code looks like:
             if (arrMesh != null)
             {
                 // Create mesh surface from mesh array
-                arrMesh.AddSurfaceFromArrays(Mesh.PrimitiveType.Triangles, surfaceArray); // No blendshapes, lods, or compression used.
+                // No blendshapes, lods, or compression used.
+                arrMesh.AddSurfaceFromArrays(Mesh.PrimitiveType.Triangles, surfaceArray);
             }
         }
     }
--- a/tutorials/3d/procedural_geometry/arraymesh.rst
+++ b/tutorials/3d/procedural_geometry/arraymesh.rst
@@ -247,7 +247,6 @@ that you find online.
 
     var rings = 50
     var radial_segments = 50
-    var height = 1
     var radius = 1
 
     func _ready():
@@ -310,7 +309,6 @@ that you find online.
 
             var rings = 50;
             var radialSegments = 50;
-            var height = 1;
             var radius = 1;
 
             // Vertex indices.
--- a/tutorials/3d/procedural_geometry/arraymesh.rst
+++ b/tutorials/3d/procedural_geometry/arraymesh.rst
@@ -303,13 +303,13 @@ that you find online.
 
     public partial class MyMeshInstance3D : MeshInstance3D
     {
+        private int _rings = 50;
+        private int _radialSegments = 50;
+        private float _radius = 1;
+
         public override void _Ready()
         {
             // Insert setting up the surface array and lists here.
-
-            var rings = 50;
-            var radialSegments = 50;
-            var radius = 1;
 
             // Vertex indices.
             var thisRow = 0;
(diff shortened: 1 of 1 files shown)


tutorials/3d/procedural_geometry/arraymesh.rst
```diff
@@ -126,6 +142,24 @@ by adding each array to ``surface_array`` and then committing to the mesh.
...
+    
+    surfaceArray[(int)Mesh.ArrayType.Vertex] = verts;
+    surfaceArray[(int)Mesh.ArrayType.TexUV] = uvs;
+    surfaceArray[(int)Mesh.ArrayType.Normal] = normals;
+    surfaceArray[(int)Mesh.ArrayType.Index] = indices;
+
+    var arrMesh = Mesh as ArrayMesh;
+    if (arrMesh != null)
+    {
+        arrMesh.AddSurfaceFromArrays(Mesh.PrimitiveType.Triangles, surfaceArray); // No blendshapes, lods, or compression used.
```
paulloz, line 160: This is creating horizontal scroll which we tend to avoid, maybe put the comment on its own line (and do the same in GDScript).

tutorials/3d/procedural_geometry/arraymesh.rst
```diff
@@ -159,6 +193,46 @@ Put together, the full code looks like:
...
+            surfaceArray[(int)Mesh.ArrayType.Vertex] = verts;
+            surfaceArray[(int)Mesh.ArrayType.TexUV] = uvs;
+            surfaceArray[(int)Mesh.ArrayType.Normal] = normals;
+            surfaceArray[(int)Mesh.ArrayType.Index] = indices;
+
+            var arrMesh = Mesh as ArrayMesh;
+            if (arrMesh != null)
+            {
+                // Create mesh surface from mesh array
+                arrMesh.AddSurfaceFromArrays(Mesh.PrimitiveType.Triangles, surfaceArray); // No blendshapes, lods, or compression used.
```
paulloz, line 231: Ditto horizontal scroll.

tutorials/3d/procedural_geometry/arraymesh.rst
```diff
@@ -126,6 +142,24 @@ by adding each array to ``surface_array`` and then committing to the mesh.
 
     mesh.add_surface_from_arrays(Mesh.PRIMITIVE_TRIANGLES, surface_array) # No blendshapes, lods, or compression used.
 
+ .. code-tab:: csharp C#
+
+    var verts = vertList.ToArray();
```
paulloz, line 147: Do we need to store references to the intermediate arrays here?  

Also, my default would be to use a `Variant[]` until we have to pass it to `AddSurfaceFromArrays`. Would make less engine calls. Maybe @raulsntos can agree/disagree?

```cs
Variant[] surfaceArray = new Variant[(int)Mesh.ArrayType.Max];

surfaceArray[(int)Mesh.ArrayType.Vertex] = vertsList.ToArray();
// ...

arrMesh.AddSurfaceFromArrays(Mesh.PrimitiveType.Triangles, new Godot.Collections.Array(surfaceArray));
```

tutorials/3d/procedural_geometry/arraymesh.rst
markawilson: Thanks, fixed for C# and GDScript both.

tutorials/3d/procedural_geometry/arraymesh.rst
markawilson: Same, fixed in both languages.

markawilson:
While making refinements to address comments above, I noticed that the "height" variable used in the final GDScript example (and initially copied into the corresponding C# example) is not actually used, so I removed it from both.

tutorials/3d/procedural_geometry/arraymesh.rst
markawilson: > Do we need to store references to the intermediate arrays here?

No, I don't think so. Fixed and renamed the List<> variables to match the GDScript examples.

> Also, my default would be to use a Variant[] until we have to pass it to AddSurfaceFromArrays. Would make less engine calls.

That's well beyond my knowledge of the bindings, so I'm learning something here. It might be more consistent with other docs to use the Array, since the [API differences page](https://docs.godotengine.org/en/latest/tutorials/scripting/c_sharp/c_sharp_differences.html) lists Godot.Collections.Array as the equivalent type to PackedVector\*Array, but that might be a secondary concern. I'm out of my depth on this question, so I'm happy to wait for further comment and/or be guided by those who know more than I do.

raulsntos: @paulloz I don't think it would result in less engine calls, the `Godot.Collections.Array` constructor from `Variant[]` ends up doing the same thing that the current example is doing but with an extra `Variant[]` allocation:

https://github.com/godotengine/godot/blob/f2aae8fa5c2e9d9323832fb43c8446c2e518d698/modules/mono/glue/GodotSharp/GodotSharp/Core/Array.cs#L55-L69

Oh wait, I think I get what you mean. So you are saying it would be better to avoid constructing the `Godot.Collections.Array` in case we don't enter the `if (arrMesh != null)`? I agree it would be better, but I'm not sure if such optimizations should be in the documentation where the focus is more on readability.

paulloz: @raulsntos I think I got confused. I thought we were making a more efficient memory copy here, but it's only when dealing with `Packed*Array` on the engine side. Sorry, don't mind me then. It's perfectly fine how it is.

tutorials/3d/procedural_geometry/arraymesh.rst
```diff
@@ -233,6 +299,74 @@ that you find online.
 
       # Insert committing to the ArrayMesh here.
 
+ .. code-tab:: csharp C#
+
+    public partial class ArrayMeshInstance : MeshInstance3D
```
paulloz, line 304: Doc convention would call this class `MyMeshInstance3D`.

tutorials/3d/procedural_geometry/arraymesh.rst
```diff
@@ -233,6 +299,74 @@ that you find online.
...
+    {
+        public override void _Ready()
+        {
+            // Insert setting up the surface array and lists here.
+
+            var rings = 50;
```
paulloz, line 310: If we want to match GDScript, those should be fields in the class instead of local variables.

paulloz (approved):
I finally had time to test this. Can be merged IMO.

tutorials/3d/procedural_geometry/arraymesh.rst
markawilson: Yes, good catch. Updated.

tutorials/3d/procedural_geometry/arraymesh.rst
markawilson: Updated/fixed -- but can I ask, is there a page that documents these conventions? I see that the [C# style guide](https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_style_guide.html) includes several examples with this naming convention, but at the same time, the [documentation writing guidelines](https://docs.godotengine.org/en/latest/contributing/documentation/docs_writing_guidelines.html) urge us not to use names like `my_var` or `my_func()`.

(I mention this not to argue the point! Just to say that I would find it helpful to have a reference that definitively states what conventions to follow, if one exists.)

paulloz: Most of it is following the [style guide](https://docs.godotengine.org/en/latest/tutorials/scripting/c_sharp/c_sharp_style_guide.html). For this particular naming thing, I think we discussed it recently in the rocket.chat. I'll take some time to formalize a C# section in the doc writing guidelines page you mentioned 👍

raulsntos (approved):
LGTM, thanks for contributing to the C# documentation. And thanks to @paulloz too for the review.

mhilbrunner (approved):

mhilbrunner:
Thanks and congrats on your first docs contribution! Merged. Also thanks for the reviews, thats super helpful :)

Issue #6776, opened 2023-02-14
2D Navigation Overview Uses Potentially Problematic Logic In Demo Code


azterior:
**Your Godot version:**
4.0 Latest Docs

**Issue description:**
Having a return statement at the start of a function prevents it from being expanded and would require the user to implement the same change I suggest below in order to expand the script. This removes an unnecessary step a new user may struggle with.

```gdscript
func _physics_process(delta):

    if navigation_agent.is_target_reached():

        return

    ...
```

this logic should be replaced with 

```gdscript
func _physics_process(delta):

    if not navigation_agent.is_target_reached():

        ...
```

Apologies for any issues in this reques, this is my first.

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/latest/tutorials/navigation/navigation_introduction_2d.html

Calinou:
PS: Code blocks should use triple backticks like this (with an optional language name for syntax highlighting):

\`\`\`gdscript
code here
\`\`\`

I edited your post accordingly, but remember to do this in the future :slightly_smiling_face:

smix8:
In fact all the agent script examples should properly have a `is_navigation_finished()` check and return instead of `is_target_reached()` which is unreliable as it only checks the distance to the final target compared to `is_navigation_finished()` that actually checks if the loaded path array was processed to the end.

```gdscript
func _physics_process(delta):
    if navigation_agent.is_navigation_finished():
        return
```

It is a common beginner problem to always process path following even while the navigation ended and this can cause some jitter when agents should stand in place, especially in 2D that uses a large scale and has more precision problems compared to 3D.

skyace65:
Closing this issue per Calinou's explanation on why the scripts are like this, and the changes in #6780.

PR #6780, opened 2023-02-15
Add NavigationAgent.is_navigation_finished() early return to script examples to avoid in-place jitter

smix8:
Adds early return to script examples for `NavigationAgent` path following when `NavigationAgent.is_navigation_finished(`) is `true`.

Related to https://github.com/godotengine/godot-docs/issues/6776.

The old examples had no guards or used `is_target_reached()` which is an unreliable check for this as it only checks the distance to the final target. The `is_navigation_finished()` checks if the actual path following and path array has reached the end.

Without such a check the path following gets processed every frame which can make agent jitter in place when the path is finished due to precision problems, especially in 2D that uses a far larger scale.

### DIFF
--- a/tutorials/navigation/navigation_introduction_2d.rst
+++ b/tutorials/navigation/navigation_introduction_2d.rst
@@ -123,7 +123,7 @@ NavigationServer2D and a NavigationAgent2D for path movement.
         navigation_agent.set_target_location(movement_target)
 
     func _physics_process(delta):
-        if navigation_agent.is_target_reached():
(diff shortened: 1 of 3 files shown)


Calinou (approved):

mhilbrunner (approved):

mhilbrunner:
Thanks for the fix!

smix8:
I guess yes, Calinou already mentioned in that issue the reasoning why we would keep the script examples like this.

PR #6784, opened 2023-02-17
Update class_cubemesh.rst

randompast:
Added a comment to assist people who are writing fragment shaders rather than using the standard shader.

### DIFF
--- a/classes/class_cubemesh.rst
+++ b/classes/class_cubemesh.rst
@@ -19,7 +19,7 @@ Description
 
 Generate an axis-aligned cuboid :ref:`PrimitiveMesh<class_PrimitiveMesh>`.
 
-The cube's UV layout is arranged in a 3×2 layout that allows texturing each face individually. To apply the same texture on all faces, change the material's UV property to ``Vector3(3, 2, 1)``.
+The cube's UV layout is arranged in a 3×2 layout that allows texturing each face individually. To apply the same texture on all faces, change the material's UV property to ``Vector3(3, 2, 1)``.  In a fragment shader, this can be done with ``ALBEDO = texture(tex, UV * vec2(3.0,2.0)).xyz``.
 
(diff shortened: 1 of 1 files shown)


Calinou:

randompast:
Thanks for the help getting started with this process.  I saw this "To modify or create pages in the reference manual, you need to edit .rst files in the [godot-docs GitHub repository](https://github.com/godotengine/godot-docs)" from [here](https://docs.godotengine.org/en/latest/contributing/documentation/contributing_to_the_documentation.html).

"You should not edit .rst files in the classes/ folder of the [documentation repository](https://github.com/godotengine/godot-docs/)." is stated [here](https://docs.godotengine.org/en/latest/contributing/documentation/updating_the_class_reference.html).

The directions I followed [here](https://docs.godotengine.org/en/latest/contributing/documentation/contributing_to_the_documentation.html#editing-pages-online) contradict this. **Maybe the edit link should not go to the godot-docs page for class pages?**

randompast:
I'm seeing something that looks almost right.  Is this the right page?  It doesn't seem to quite match: https://github.com/godotengine/godot/blob/master/doc/classes/BoxMesh.xml

randompast:
Transitioned the comment to BoxMesh: https://github.com/godotengine/godot/pull/73479

YuriSizov:
> I saw this "To modify or create pages in the reference manual, you need to edit .rst files in the godot-docs GitHub repository" from here.

There is a warning next to it, making an exception for the class reference sources. But it's also okay if you make a mistake first time around, we're here to help.

randompast:
> it's also okay if you make a mistake first time around, we're here to help.

@YuriSizov Thanks for understanding.

There's also a naming inconsistency from cubemesh in stable to boxmesh in latest.  Switching to latest (or to an earlier version) yields a "Page not found" error.  How are these situations resolved?

image

YuriSizov:
For tutorials we have redirects, but not for the class reference. We may come up with a solution in future, but for now it is indeed just "Page not found".

PR #6786, opened 2023-02-17
Clarify return value

nobbele:
I didn't look at other documentation to see whether they write the "default" return value or not. While it might be "obvious", I don't think the user should need to assume what the default value should probably be.
And while it might be irrelevant for other event types, IMO this one is useful in cases like

```cs
if (e.IsActionPressed("main_action") && !e.IsEcho())
{ ... }
```

where `main_action` might not be a key and `e` might not be an InputEventKey

### DIFF
--- a/classes/class_inputevent.rst
+++ b/classes/class_inputevent.rst
@@ -215,7 +215,7 @@ Returns ``true`` if this input event's type is one that can be assigned to an in
 
 :ref:`bool<class_bool>` **is_echo** **(** **)** |const|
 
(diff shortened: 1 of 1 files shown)


YuriSizov:
https://github.com/godotengine/godot/tree/master/doc/classes/InputEvent.xml

nobbele:
I see
Also, that link seems dead, stable version exists though

YuriSizov:
Good catch, thanks.

https://docs.godotengine.org/en/latest/contributing/documentation/updating_the_class_reference.html

PR #6792, opened 2023-02-19
Update the `Compiling with Mono` page

RedworkDE:
The `Compiling with Mono` page still contained a whole bunch of information that is no longer relevant. The basically deletes all the bits that are no longer relevant and slightly improves the rest.

The page should probably also be renamed since `mono` is no longer used.

*Bugsquad edit: closes https://github.com/godotengine/godot/issues/65936*


contributing/development/compiling/compiling_with_mono.rst
```diff
@@ -8,119 +8,73 @@ Compiling with Mono
...
+    ./modules/mono/build_scripts/build_assemblies.py --godot-output-dir=./bin
 
-    scons p=<platform> target=template_release module_mono_enabled=yes
+If everything went well, the ``GodotSharp`` directory, containing the managed
+libraries, should have been created in the in the ``bin`` directory.
 
-If everything went well, apart from the normal output, SCons should have created
-the following files in the ``bin`` directory:
+.. note:: Due to NuGet package caching, rebuilding the managed libraries does
+          not reliably update all projects using them. To solve this issue
```
raulsntos, line 67: I don't like the phrasing in here, it makes it sound as if it's a NuGet problem when in reality this behavior is entirely our doing in Godot since we copy the NuGet packages when building (but only if packages with the same version don't already exist in the directory).

We plan to get rid of the `GodotNuGetFallbackFolder` anyway, I think we were going to do that with _editor unification_ but in the end that didn't make it for 4.0.

contributing/development/compiling/compiling_with_mono.rst
```diff
@@ -132,37 +86,32 @@ Example (Windows)
 
     # Build editor binary
     scons p=windows target=editor module_mono_enabled=yes
+    # Build export templates
+    scons p=windows target=template_debug module_mono_enabled=yes
+    scons p=windows target=template_release module_mono_enabled=yes
+    
     # Generate glue sources
-    bin\godot.windows.editor.x86_64.mono --generate-mono-glue modules/mono/glue
+    bin/godot.windows.editor.x86_64.mono --generate-mono-glue modules/mono/glue
```
raulsntos, line 94: This may have been intentional since these are Windows instructions. Even if Windows supports `/`, we may want to use whatever is preferred by Windows, but I don't have a strong opinion about it.

contributing/development/compiling/compiling_with_mono.rst
```diff
@@ -8,119 +8,73 @@ Compiling with Mono
...
+might have similarly named binaries without this suffix. These binaries can't be
+used to generate the Mono glue.
+
+.. note:: The glue sources must be regenerated every time the ClassDB-registered
+          API changes. That is, for example, when a new method is registered to
+          the scripting API or one of the parameters of such a method changes.
+          Godot will print an error at startup if there is an API mismatch
+          between ClassDB and the glue sources.
+
+Building the managed libraries
```
raulsntos, line 55: We may also want to mention the `--push-nupkgs-local` and `--precision` flag in this section. See https://github.com/godotengine/godot/blob/9f68d06ec2d8dbeb237c5aa5c6f9bbcbe26bde5d/modules/mono/README.md.

contributing/development/compiling/compiling_with_mono.rst
```diff
@@ -8,119 +8,73 @@ Compiling with Mono
...
+------------------------------
+
+Once you have generated the Mono glue, you can builds the managed libraries with
```
raulsntos, line 58:
```suggestion
Once you have generated the Mono glue, you can build the managed libraries with
```

contributing/development/compiling/compiling_with_mono.rst
```diff
@@ -172,197 +121,28 @@ The data directory is a dependency for Godot binaries built with the Mono module
 enabled. It contains important files for the correct functioning of Godot. It
 must be distributed together with the Godot executable.
 
-.. note:: The information below doesn't apply for Android, iOS and WASM,
-          as there is no data directory for these platforms.
-
-Export templates
```
raulsntos, line 178: Does this information about export templates in MacOS no longer apply?

raulsntos (changes requested):
Thank you very much for updating this documentation page!

> The page should probably also be renamed since `mono` is no longer used.

That's true, but the module still has a few mentions to `mono` (the directory is still called `mono`). We should update the page to use the naming `.NET module` instead of `Mono module`, even if internally the `mono` name is still used somtimes. Maybe we could clarify in a small note what `mono` is, if needed.

contributing/development/compiling/compiling_with_mono.rst
RedworkDE: I changed it mostly because it messed up the syntax highlighting and that was the only backslash on the page.

contributing/development/compiling/compiling_with_mono.rst
RedworkDE: The data directory for the export templates no longer exists at all and the bit about putting the editor specific `GodotSharp` directory into the .app is mentioned at the end of the editor section (line 132).

contributing/development/compiling/compiling_with_mono.rst
RedworkDE: This is not entirely a godot specific issue, I had issues with other packages as well. TBH I am actually pretty surprised that deleting `GodotNuGetFallbackFolder` even works and that the copy in `~/.nuget/packages` that nuget makes doesn't cause any issues.

raulsntos: I haven't had issues with other packages, but if it's resolved by deleting `GodotNuGetFallbackFolder` then it seems specific to Godot, since that folder should only contain Godot packages.

Keep in mind that the configuration we add to `~/.config/NuGet/config/Godot.Offline.Config` affects non-Godot projects as well, when restoring any NuGet package (including non-Godot packages).

paulloz:
> That's true, but the module still has a few mentions to `mono` (the directory is still called `mono`). We should update the page to use the naming `.NET module` instead of `Mono module`

Agree. IMHO, it's important to have this page up to date, even if it means referencing the word "mono" a few times for now. We can add a note at the top of the page explaining it is used in place of "dotnet" here for legacy reasons. It'll still be possible to update the doc (and module itsel?) step by step.

contributing/development/compiling/compiling_with_mono.rst
RedworkDE: I largely copied that README to document these parameters and also wrote a bit about the `godot-platform` argument.

contributing/development/compiling/compiling_with_mono.rst
RedworkDE: I left this as is for now, I'll have to test some things again and come back to this.

contributing/development/compiling/compiling_with_mono.rst
```diff
@@ -8,119 +8,73 @@ Compiling with Mono
...
----------------------
-
-By default, SCons will try to find Mono in the Windows Registry on Windows or
-via ``pkg-config`` on other platforms. You can specify a different installation
-directory by passing the ``mono_prefix`` command-line option to SCons; e.g.
-``scons [...] mono_prefix=%ProgramFiles%/Mono``.
-
-This is the directory that contains the subdirectories ``include`` and ``lib``.
-Note that as usual, paths including spaces must be wrapped in double quotes.
+You can use ``dotnet --version`` to check your .NET version.
```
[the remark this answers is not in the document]
RedworkDE: Mentioned `--list-sdks` instead, tho am not sure if it is better now, since that latest version is the only one that really matters anyways.

contributing/development/compiling/compiling_with_mono.rst
RedworkDE: [nuget_local_package_cache.zip](https://github.com/godotengine/godot-docs/files/10785701/nuget_local_package_cache.zip)
This demonstrates the issue I was talking about in a fresh docker container that has never seen anything godot:
1. build the docker container
2. start it and run `/test/run.sh`
   - it builds a package and publishes it to a local nuget source
   - runs a project showing the old message
   - updates the package source to a new message
   - rebuilds the package and publishes it again
   - re-runs the project still showing the old message
3. the new message will be shown on re-run if the package is linked as a project reference instead or if `~/ .nuget/packages` is cleared between the runs.

So while the exact parameters of this issue are godot specific, the same symptoms can also happen when no using godot.

Also I found why the nugets cache folder doesn't cause problems: The build is actually setup to delete the package being built from the nuget cache to avoid this issue, but that bit just doesn't do the same for godots cache, and thus not actually solving anything.

contributing/development/compiling/compiling_with_mono.rst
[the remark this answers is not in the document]
raulsntos: I think it's fine to use `--version`, `--list-sdks` or `--info`, all of those should allow you to check which version(s) of the .NET SDK you have installed, which is the point of this line. I don't think it's too important which one we recommend, I just wanted to be clear about what developers should be looking for and what the output of the command means.

If we tell developers to use `--version` we should be clear that this is the version of the latest .NET SDK you have installed, and that this is likely what you care about, but that you may have other SDKs installed and that's fine too.

If, instead, we want to tell developers to use `--list-sdks` or `--info`, we should clarify what part of the output they should be looking for.

Either way, this is something I'd expect a developer familiar with .NET to already know about, so it's probably not a big deal anyway.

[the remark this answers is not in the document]
RedworkDE: I switched to recommending `--info`, since the other options are somewhat broken if no sdk and just a runtime is installed. IMO the output is `--info` is clear enough that that we don't have the clarify things here.

```
root@53c40fc161c2:/# dotnet --version
The command could not be loaded, possibly because:
  * You intended to execute a .NET application:
      The application '--version' does not exist.
  * You intended to execute a .NET SDK command:
      No .NET SDKs were found.

Download a .NET SDK:
https://aka.ms/dotnet-download

Learn about SDK resolution:
https://aka.ms/dotnet/sdk-not-found
root@53c40fc161c2:/# dotnet --list-sdks
root@53c40fc161c2:/# dotnet --info

global.json file:
  Not found

Host:
  Version:      6.0.14
  Architecture: x64
  Commit:       2a90daa2cc

.NET SDKs installed:
  No SDKs were found.

.NET runtimes installed:
  Microsoft.NETCore.App 6.0.14 [/usr/share/dotnet/shared/Microsoft.NETCore.App]

Download .NET:
  https://aka.ms/dotnet-download

Learn about .NET Runtimes and SDKs:
  https://aka.ms/dotnet/runtimes-sdk-info
```

contributing/development/compiling/compiling_with_dotnet.rst
```diff
@@ -0,0 +1,213 @@
...
+If everything went well, the ``GodotSharp`` directory, containing the managed
+libraries, should have been created in the in the ``bin`` directory.
+
+.. note:: By default, all development builds share a version number, which can
+          cause some issues with caching of the NuGet packages. To solve this
+          issue either use ``GODOT_VERSION_STATUS`` to give every build a unique
+          version or delete ``GodotNuGetFallbackFolder`` after every build to
+          clear the package cache.
+
+Unlike "classical" Godot builds, when building with the Mono module enabled
```
raulsntos, line 78:
```suggestion
Unlike "classical" Godot builds, when building with the .NET module enabled
```

raulsntos (commented):
Just a few Mono/.NET renames that were missed.

raulsntos (approved):

RedworkDE:
This can probably close https://github.com/godotengine/godot/issues/65936, which seems a bit unclear but mostly seems to boil down to this documentation being missing.

mhilbrunner (approved):

mhilbrunner:
Thank you! Merged.

PR #6799, opened 2023-02-21
Information about the need to download v. 4.72 of .Net

iophi:
I had a problem with OmniSharp in VS code getting an "Error: The reference assemblies for .NETFramework,Version=v4.7.2 were not found. To resolve this, install the Developer Pack (SDK/Targeting Pack) for this framework version or retarget your application. You can download .NET Framework Developer Packs at...". I see that Godot should download it automatically (https://github.com/godotengine/godot/issues/40546) but It evidently stopped doing that, or sometimes fails to do that. That's why I'm adding this information for anyone, so that when they will get the same error and get flustrated, then they won't quit because of it.

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_basics.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_basics.rst
@@ -50,6 +50,8 @@ version of Godot.
 Download and install the latest stable version of the SDK from the
 `.NET download page <https://dotnet.microsoft.com/download>`__.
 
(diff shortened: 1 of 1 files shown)


RedworkDE:
This is not relevant for 4.0, so I don't think this PR should target `master`.
Also I don't think dates should be added to the docs, unless they are actually relevant.

YuriSizov:
This is indeed not relevant to the master version of the docs. The 3.x version doesn't mention a specific version of .NET at the moment, so it may be relevant there.

PR #6803, opened 2023-02-22
Improve the Shader preprocessor documentation

bitsawer:
Several small fixes and improvements to the shader preprocessor documentation. Some syntax fixes (a few missing \` characters messing up some highlights etc.), a few code logic fixes and one of the `#elif` code examples was identical to the one in `#if` section so I rewrote it, probably  a copy-paste mistake?


Calinou (approved):
Thanks!

Issue #6744, opened 2023-02-07
"Your first 3D game" issues


andreymal:
There's a bunch of weird things I ran into while trying to follow the "Squash the Creeps" tutorial in Godot 4. I'm not a Godot expert and don't know how to fix some of these things, so I don't want to make a pull request, but I hope this list can help other contributors improve the documentation.

### Common

The screenshots are inconsistent:

* some taken in Godot 3, some in Godot 4
* some use names in PascalCase, some in snake_case
* some nodes have the "3D" suffix (Godot 4), some don't (Godot 3), some have the lowercase "3d" suffix (beta1?)
* some screenshots are in PNG format, some are JPG, some are WebP (and I think WebP is worse than PNG), and they vary in quality and UI scale

Perhaps it would be a good idea to recreate all the screenshots after the first release candidate?

### Your first 3D game

* [ ] [`index.rst:34`](https://github.com/godotengine/godot-docs/blob/8a19f445324806e0da2b22bd079bd4470a108887/getting_started/first_3d_game/index.rst?plain=1#L34)
  This link (https://github.com/GDQuest/godot-3d-dodge-the-creeps) redirects to https://github.com/godotengine/godot-3d-dodge-the-creeps — I think it would be nice to fix this unnecessary redirect.
  I also found this link in three other places:
  * [`index.rst:45`](https://github.com/godotengine/godot-docs/blob/8a19f445324806e0da2b22bd079bd4470a108887/getting_started/first_3d_game/index.rst?plain=1#L45)
  * [`01.game_setup.rst:12`](https://github.com/godotengine/godot-docs/blob/8a19f445324806e0da2b22bd079bd4470a108887/getting_started/first_3d_game/01.game_setup.rst?plain=1#L12)
  * [`going_further.rst:20`](https://github.com/godotengine/godot-docs/blob/8a19f445324806e0da2b22bd079bd4470a108887/getting_started/first_3d_game/going_further.rst?plain=1#L20)

### Setting up the game area

* [ ] [`01.game_setup.rst:42`](https://github.com/godotengine/godot-docs/blob/8a19f445324806e0da2b22bd079bd4470a108887/getting_started/first_3d_game/01.game_setup.rst?plain=1#L42)

  > click the *Add Node* button represented by a "+" icon

  This button is actually "Add **Child** Node" even when there are no nodes in the scene

[...]

PR #6811, opened 2023-02-24
Update getting started C# code, remove randomize

raulsntos:
- Sync C# code with GDScript.
- Remove `randomize` calls.
	- This fixes one point of https://github.com/godotengine/godot-docs/issues/6744.
- Rename members that have been renamed in Godot's C# API for 4.0.
	- Fixes https://github.com/godotengine/godot-docs/issues/6878.
- Change `delta` parameter type to `double` in C#.
- Follow our C# code style more closely.
- Other minor code fixes.


getting_started/first_2d_game/03.coding_the_player.rst
```diff
@@ -91,9 +91,9 @@ value written in the script.
 
 .. warning:: If you're using C#, you need to (re)build the project assemblies
              whenever you want to see new export variables or signals. This
-             build can be manually triggered by clicking the word "Mono" at the
-             bottom of the editor window to reveal the Mono Panel, then clicking
-             the "Build Project" button.
+             build can be manually triggered by clicking the word "MSBuild" at the
+             bottom of the editor window to reveal the MSBuild Panel, then clicking
+             the "Build" button.
```
RedworkDE, line 96: Maybe mention the build button in the top right as well? 
image

getting_started/first_3d_game/04.mob_scene.rst
```diff
@@ -198,12 +201,13 @@ We got a random position, now we need a ``random_speed``. ``randi_range()`` will
     {
         // ...
 
-        // We calculate a random speed.
-        float randomSpeed = (float)GD.RandRange(MinSpeed, MaxSpeed);
+        // We calculate a random speed (integer).
+        int randomSpeed = GD.RandRange(MinSpeed, MaxSpeed);
```
RedworkDE, line 205: Why an integer speed? Speed can definitely be continuous, I would change the Min/MaxSpeed definition to float instead, tho that would require changing all the other speeds to float as well.

getting_started/first_3d_game/04.mob_scene.rst
```diff
@@ -164,14 +164,17 @@ between ``-PI / 4`` radians and ``PI / 4`` radians.
...
+
  .. code-tab:: csharp
 
-    // We will call this function from the Main scene
+    // This function will be called from the Main scene.
     public void Initialize(Vector3 startPosition, Vector3 playerPosition)
     {
-        // We position the mob and turn it so that it looks at the player.
+        // We position the mob by placing it at startPosition
+        // and rotate it towards playerPosition, so it looks at the player.
```
RedworkDE, line 174:
```suggestion
        // and rotating it towards playerPosition, so it looks at the player.
```

getting_started/first_3d_game/04.mob_scene.rst
```diff
@@ -164,14 +164,17 @@ between ``-PI / 4`` radians and ``PI / 4`` radians.
...
         LookAtFromPosition(startPosition, playerPosition, Vector3.Up);
-        // And rotate it randomly so it doesn't move exactly toward the player.
+        // In this rotation^, the mob will move directly towards the player
+        // so we rotate it randomly within range of -90 and +90 degrees.
```
RedworkDE, line 177:
```suggestion
        // In this rotation, the mob will move directly towards the player
        // so we rotate it randomly within range of -90 and +90 degrees.
```
Maybe even `With this rotation` (not 100% sure if that is correct).

Or maybe reword this whole comment a bit:
```suggestion
        // Rotate this mob randomly within range of -90 and +90 degrees,
        // so that it doesn't move directly towards the player.
```

getting_started/first_3d_game/04.mob_scene.rst
```diff
@@ -282,39 +286,46 @@ Here is the complete ``Mob.gd`` script for reference.
...
-        // We will call this function from the Main scene
+        // This function will be called from the Main scene.
         public void Initialize(Vector3 startPosition, Vector3 playerPosition)
         {
+            // We position the mob by placing it at startPosition
+            // and rotate it towards playerPosition, so it looks at the player.
             LookAtFromPosition(startPosition, playerPosition, Vector3.Up);
+            // In this rotation^, the mob will move directly towards the player
+            // so we rotate it randomly within range of -90 and +90 degrees.
             RotateY((float)GD.RandRange(-Mathf.Pi / 4.0, Mathf.Pi / 4.0));
```
RedworkDE, line 316: see above

getting_started/first_3d_game/07.killing_player.rst
```diff
@@ -283,35 +279,42 @@ Next is ``Mob.gd``.
...
 
+        // This function will be called from the Main scene.
         public void Initialize(Vector3 startPosition, Vector3 playerPosition)
         {
+            // We position the mob by placing it at startPosition
+            // and rotate it towards playerPosition, so it looks at the player.
             LookAtFromPosition(startPosition, playerPosition, Vector3.Up);
+            // In this rotation^, the mob will move directly towards the player
+            // so we rotate it randomly within range of -90 and +90 degrees.
             RotateY((float)GD.RandRange(-Mathf.Pi / 4.0, Mathf.Pi / 4.0));
```
RedworkDE, line 300: see above

getting_started/first_3d_game/09.adding_animations.rst
```diff
@@ -532,37 +554,44 @@ And the *Mob*'s script.
...
 
+        // This function will be called from the Main scene.
         public void Initialize(Vector3 startPosition, Vector3 playerPosition)
         {
+            // We position the mob by placing it at startPosition
+            // and rotate it towards playerPosition, so it looks at the player.
             LookAtFromPosition(startPosition, playerPosition, Vector3.Up);
+            // In this rotation^, the mob will move directly towards the player
+            // so we rotate it randomly within range of -90 and +90 degrees.
             RotateY((float)GD.RandRange(-Mathf.Pi / 4.0, Mathf.Pi / 4.0));
```
RedworkDE, line 575: see above

getting_started/first_3d_game/04.mob_scene.rst
raulsntos: AFAIK, that's the type used in GDScript so that's what I used.

getting_started/first_3d_game/04.mob_scene.rst
raulsntos: I don't think `rotating` is correct here. The `and` is connecting the sentences "We position the mob [...]" and "rotate it towards [...]".

I feel like if we changed it to `rotating` then it reads more like "We position the mob by placing it [...]" and "We position the mob by rotating it towards [...]", which is not correct.

Maybe we could add a comma after `startPosition`?

getting_started/first_3d_game/04.mob_scene.rst
raulsntos: The `^` is likely an arrow pointing to the previous line. I agree though that `with this rotation` sounds better.

I actually like your reworded comment more than what we currently have, but I'd like to also get the opinion of @TheYellowArchitect who originally wrote this comment in the GDScript version.

getting_started/first_3d_game/04.mob_scene.rst
RedworkDE: Oh, is this supposed to read `(position by placing) and (rotating)` instead of `position by (placing and rotating)`?

Then i would add a comma or maybe even a semicolon and maybe change `and` -> `and then`.

Although this is a comment for just a single function call so my original interpretation of this still seems accurate.

raulsntos: > Oh, is this supposed to read `(position by placing) and (rotate it)` instead of `position by (placing and rotating)`?

Yes, I think that's how it's supposed to be read.

> Although this is a comment for just a single function call so my original interpretation of this still seems accurate.

Well, the method `LookAtFromPosition` does both (positions the node and then rotates it):

https://github.com/godotengine/godot/blob/6296b46008fb8d8e5cb9b60af05fa1ea26b8f600/doc/classes/Node3D.xml#L122-L130

getting_started/first_3d_game/04.mob_scene.rst
akien-mga: Yeah the `^` is weird, can definitely be rephrased.

raulsntos: Changed.

mhilbrunner (approved):

mhilbrunner:
Looks good, merged! Thanks. 🎉

PR #6813, opened 2023-02-24
Enable error reporting on the CI and address the remaining errors

YuriSizov:
Reverts us to the state before https://github.com/godotengine/godot-docs/commit/6952796767163494f24cd2a9269565d30902b5c5.

We had a couple of issues for a long time now, hoping to improve related articles or for related bugs to be fixed, but at this point disabling error reporting for their sake makes us miss obvious bugs in new PRs. So I think it's time.

- Shader syntax highlighter is disabled for those couple samples that have issues with Pygments 2.14;
- Invalid node references removed from the networking documentation (it's pending rework, and removing them doesn't break the flow of the article anyway);
- Invalid internal reference in a recent XR article is fixed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,4 +26,4 @@ jobs:
       # Use dummy builder to improve performance as we don't need the generated HTML in this workflow.
       - name: Sphinx build
         run: |
-          sphinx-build --color -b dummy -d _build/doctrees . _build/html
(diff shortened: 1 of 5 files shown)


mhilbrunner (approved):

mhilbrunner:
LGTM. This has been on my list for a bit, thanks for doing it! All that time spent fixing warnings paying off, nice :)

PR #6823, opened 2023-02-26
Fix doc in "2D Navigation Overview"

urtow:
navigation_agent have method called get_next_path_location(), not get_next_location()

### DIFF
--- a/tutorials/navigation/navigation_introduction_2d.rst
+++ b/tutorials/navigation/navigation_introduction_2d.rst
@@ -127,7 +127,7 @@ NavigationServer2D and a NavigationAgent2D for path movement.
             return
 
         var current_agent_position : Vector2 = global_transform.origin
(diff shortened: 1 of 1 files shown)


mhilbrunner:
cc @smix8

paulloz:
If the method was renamed, it also needs to be changed in the C# example (line 182).

urtow:
@paulloz updated.

P.S. Nice avatar

smix8 (approved):

mhilbrunner (approved):

smix8:
Should also fix the setter function not just the getter.

akien-mga:
Thanks for the contribution! This was missing some places that need to be changed, and actually using the wrong name for the new method (`get_next_path_position`, not `_location`). So I made #6910 to fix it.

PR #6830, opened 2023-02-27
.net windows editor build uses backslashes

dream-young-soul:
Using the slash command in the windows console will report an error
QQ截图20230227120054


dream-young-soul:
> Hey! Thanks for contributing. Other instructions on that page also use forward slashes. I think its fine to either keep as is, add tabs for the commands per platform or to note at the first occurence that it may require backslashes on Windows.
> 
> I think the last one may be the best option.

OK,Your choice is the best.Thanks

mhilbrunner:
Closing as per the discussion above. If someone wants to submit a PR to add that note, thats welcome of course if considered useful. Thanks for contributing!

PR #6835, opened 2023-02-27
Fix minor inconsistencies of gdextension docs

paddy-exe:
* Fixed wrong indentation caused by one tab instead of spaces:
image

* `ADD_SIGNAL` is not a method but a macro which simplifies the usage of the ClassDB method.


YuriSizov (approved):

mhilbrunner:
Thanks!

PR #6838, opened 2023-02-27
Best practices godot notifications

thiagola92:
Update one code block in [Godot Notifications](https://docs.godotengine.org/en/latest/tutorials/best_practices/godot_notifications.html#init-vs-initialization-vs-export)

Setter are always triggered (except in "initialized value").  
Reference: [Properties (setters and getters) note](https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_basics.html#properties-setters-and-getters)  

Tested with Godot 4.0-rc6
```gdscript
extends Node2D


# Setted "two" in the Inspector
@export var test: String = "one":
	set(value):
		test = value
		print("Setting: ", test)

func _init():
	test = "three"
	self.test = "four"

func _ready():
	test = "five"
	self.test = "six"
```
```
Setting: three
Setting: four
Setting: two
Setting: five
Setting: six
```

I don't know if the rest of this section still up to date.

### DIFF
--- a/tutorials/best_practices/godot_notifications.rst
+++ b/tutorials/best_practices/godot_notifications.rst
@@ -158,18 +158,16 @@ instantiation:
     # "one" is an "initialized value". These DO NOT trigger the setter.
     # If someone set the value as "two" from the Inspector, this would be an
     # "exported value". These DO trigger the setter.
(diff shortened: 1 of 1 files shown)


thiagola92:
Superseded, I'm closing.

PR #6839, opened 2023-02-28
Remove duplicate text from GDExtension doc

hakro:
2 changes :
- Make the link to godot-cpp more explicit. The phrase `See also Compiling as the build` breaks the flow, and thus the link should be mentioned again

- Remove duplicate block of text : `If you decide to download the repository ...` which is repeated in the next `Note` block


mhilbrunner (approved):

mhilbrunner:
Thanks for fixing these and congrats on your first merged Godot docs PR!

PR #6841, opened 2023-02-28
Changed admonition-title alignment from 'right' to 'left'.

jwmcgettigan:
Currently, all admonition titles are aligned to the **right**.  This is unintuitive and as all other admonitions [that I'm aware of](https://f5-rtd-howto.readthedocs.io/en/dev/resources/resources_rtd-admonitions.html) align to the **left**, I can only assume that this right alignment is not intended.  This PR simply changes the alignment from the right to the left as seen in the before and after images below.

image

image

The change that caused the right alignment appears to have been committed yesterday in https://github.com/godotengine/godot-docs/commit/ff4f111677f523892817b0c6efe0be1692263dac by @YuriSizov.
Though, I'm not sure if it was already right aligned before that as I hadn't noticed until now.

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -967,7 +967,7 @@ code,
     font-size: 105%;
     line-height: 120%;
     padding: 6px 16px;
-    text-align: right;
+    text-align: left;
 }
 
 .rst-content .admonition .admonition-title:before {


mhilbrunner:
cc @YuriSizov

YuriSizov:
Yes, this is intended. All of the page contents is aligned to the left, naturally, which makes it hard for the titles to stand out. So we are trying it on the right for now :)

YuriSizov:
We've discussed this, and the change has been reverted. I made a direct commit, but credited you for opening this PR and giving early feedback :)

https://github.com/godotengine/godot-docs/commit/f45c014fb1764299297772e58d941cd604ce582f

PR #6855, opened 2023-03-01
Fix: progress_ratio expects a float, not an int

abduelhamit:


mhilbrunner (approved):

mhilbrunner:
Thanks, merged! The change will show up on `stable` docs once we cherrypicked it. Thanks for contributing! 🎉

Issue #4267, opened 2020-10-14
Fragment built-ins "VIEW" is described as being the vector from camera to fragment, when it is actually the vector from fragment to camera.


Arnklit:
**Your Godot version:**
3.2.3

**Issue description:**
Fragment built-ins "VIEW" is described as being the vector from camera to fragment, when it is actually the vector from fragment to camera.

**URL to the documentation page:**
https://docs.godotengine.org/en/stable/tutorials/shading/shading_reference/spatial_shader.html#fragment-built-ins

clayjohn:
Are you sure? Keep in mind, VIEW is in camera space, not in world space. 

I guess either way it needs to be updated. Because it is a normalized direction, rather than a vector from one point to another.

clayjohn:
Thanks for the explanatory image. It's very helpful for me to put my head back on straight. Yes the view vector is relative to the fragment. 

It is the direction to the camera from the current fragment. It is used in conjunction with light direction and surface normal to compute lighting.

PR #6861, opened 2023-03-02
Fixed description for fragment shader VIEW built-in

and-rad:
Current docs say that the VIEW vector goes from the camera to the fragment position, but it actually points in the opposite direction. Screenshot below.

fragment_view

This is the shader code:

```
void fragment() {
	vec3 view_vector_ws = normalize((INV_VIEW_MATRIX * vec4(VIEW, 0.0)).xyz);
	ALBEDO = view_vector_ws;
}
```

If the vector was pointing from the camera to the fragment, the left side of the mesh would be black and the right side red.

_Bugsquad edit: Fixes: https://github.com/godotengine/godot-docs/issues/4267_


mhilbrunner:
cc @clayjohn

clayjohn (approved):

mhilbrunner (approved):

mhilbrunner:
Thank you, merged!

Issue #6866, opened 2023-03-02
Stable version of documention should list a precise version number


novalis:
**Your Godot version:**
3.5.1

**Issue description:**
The docs don't tell me what actual version number stable is.  I assume it's 4.0, but it would be nice to immediately.  Maybe where it says "v: stable" at the bottom, it should say "v: stable (4.0)"?  Or maybe just under the "Godot docs" header where it says "stable"?  Basically, I want to immediately know whether I'm looking at the right version without having to keep track of what's currently considered "stable".

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/stable/classes/class_os.html

YuriSizov:
Yeah, I agree. I always found the "latest/stable" note in the sidebar lacking any context to an average user. I'll look into it.

PR #6867, opened 2023-03-02
Display Godot version more clearly

YuriSizov:
Closes https://github.com/godotengine/godot-docs/issues/6866.

We have evergreen versions, stable and latest, where the version of Godot is not clear. While we only had Godot 3 docs this was mostly okay, but now that we have two stable branches, this leads to a lot of confusion. Personally, I also didn't quite like having that docs branch version written under the logo, as it was lacking context and relevance for users.

So here's an attempt to address these issues:

* Remove the version text from under the logo;
* Add Godot version to the ReadTheDocs panel in the sidebar;
  * Unfortunately, in the unfolded form, in the list of versions, it will still state "stable" and "latest". We have no control over that list, at least not in the layout.
* Add a small title to the top navigation block, similar to the page's title but with Godot version stated instead.
  * I'm stealing the same string used for page titles, so it also states the language, but that's also not bad.

chrome_2023-03-02_20-18-11

And here's how it looks now, for comparison:

chrome_2023-03-02_20-11-07

-----

PS. We should probably mark latest as 4.1 at some point, but perhaps later.

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -1601,6 +1601,7 @@ p + .classref-constant {
     overflow-y: auto;
     overflow-x: hidden;
     max-height: calc(100% - 348px);
+    padding-bottom: 24px;
 }
 @media screen and (max-width: 768px) {
     .wy-nav-side {
--- a/_static/js/custom.js
+++ b/_static/js/custom.js
@@ -13,11 +13,11 @@ const registerOnScrollEvent = (function(){
   // The number of pixels the user must scroll by before the logo is completely hidden.
   const scrollTopPixels = 84;
   // The target margin to be applied to the navigation bar when the logo is hidden.
-  const menuTopMargin = 88;
+  const menuTopMargin = 70;
   // The max-height offset when the logo is completely visible.
-  const menuHeightOffset_default = 180;
+  const menuHeightOffset_default = 162;
   // The max-height offset when the logo is completely hidden.
-  const menuHeightOffset_fixed = 98;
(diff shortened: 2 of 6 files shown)


lyuma:
This is great! I definitely appreciate the version number being easily visible, since right now I've been looking at the address bar each time.

That said, I still have to scan through the whole sentence each time to identify if I'm loading the correct version of the docs.

I'd prefer if it's not too hard if only the version number is put in bold (or highlighted in another way), for example:

Godot Engine **4.0** documentation in English
rather than
**Godot Engine 4.0 documentation in English**

```
<div style="font-size: 105%;">
		{{ godot_docs_title | replace("%s", "<b>" + godot_version + "</b>") }}
</div>
```

lyuma:
Another approach would be to use a different highlight color between 4.x and 3.x (for example a different shade of light blue in older versions, or green or something else)

Calinou:
> Another approach would be to use a different sidebar / highlight color between 4.x and 3.x (for example a different shade of light blue in older versions, or green or something else)

This should be possible by using a slightly different custom CSS file in each branch (i.e. different CSS variable for the highlight color).

mhilbrunner (approved):

YuriSizov:
Different color would be a lot of work and won't help with accessibility. We could make only the version bold, though I'm not sure it would be enough for it to stand out.

mhilbrunner:
Merged! We can discuss and make further improvements in followup PRs. This is already a good improvement IMO. Thanks! :)

PR #6869, opened 2023-03-02
Convert some things from 3.x to 4.x

scrapp08:
Few things are still 3.x for example `setget`, `export`, so have converted those to 4.x. Not everything is converted and fully checked, just a few things that I could see from scanning through


mhilbrunner:
Adressed and fixed by https://github.com/godotengine/godot-docs/pull/6498, closing as duplicate. Thanks for contributing :)

Issue #6872, opened 2023-03-03
GDExtension C++ example: the provided .gdextension file does not work


Jordyfel:
**Your Godot version:**
4.0

**Issue description:**
Copy-pasting the provided text for the .gdextension file does not work, opening the project in Godot fails to load the extension. And why would it, since the names of the libraries listed in it are different than the ones generated from the provided SConstruct file. Copy-pasting the contents of the .gdextension file in the test project in godot-cpp works just fine.

**URL to the documentation page:**
https://docs.godotengine.org/en/stable/tutorials/scripting/gdextension/gdextension_cpp_example.html

Calinou:
cc @paddy-exe

PR #6873, opened 2023-03-03
Fix GDExtension file path for extension registration

paddy-exe:
This PR fixes the issue by adding the necessary `/demo` path part.

Fixes https://github.com/godotengine/godot-docs/issues/6872


mhilbrunner:
Thanks for the quick fix! Merged.

PR #6880, opened 2023-03-03
Fix incorrect images being referenced on Using TileSets

bli00:
The ["Using a collection of scenes"](https://docs.godotengine.org/en/latest/tutorials/2d/using_tilesets.html#using-a-collection-of-scenes) section under tileset tutorial have incorrect image paths.

* fix the image paths


Calinou (approved):
Thanks!

PR #6883, opened 2023-03-04
In XR application V-sync should never be used

BastiaanOlij:
XR applications are the one exception where we should never use v-sync. This adds a line of code to the example that forces v-sync off and adds an explanation to why this is needed.


BastiaanOlij:
Also fixed that the document was using tabs instead of spaces.

Calinou (changes requested):
Text looks good to me, but 4 spaces should be used for indentation, not 2.

BastiaanOlij:
> Text looks good to me, but 4 spaces should be used for indentation, not 2.

Changed that :)

mhilbrunner (approved):

mhilbrunner:
In this goes! Thanks.

PR #6884, opened 2023-03-04
Updated setting_up_xr.rst to include C# example

VenomCodeDev:
Added a c# port for the original GDScript code. This is my first time editing a RST file, so let me know if I did something incorrect with the layout/setup.

### DIFF
--- a/tutorials/xr/setting_up_xr.rst
+++ b/tutorials/xr/setting_up_xr.rst
@@ -74,7 +74,8 @@ Right now all these nodes are on the floor, the will be positioned correctly in 
 
 Next we need to add a script to our root node. Add the following code into this script:
 
-::
+.. tabs::
+ .. code-tab:: gdscript GDScript
 
 	extends Node3D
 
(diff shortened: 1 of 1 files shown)


tutorials/xr/setting_up_xr.rst
```diff
@@ -89,6 +90,24 @@ Next we need to add a script to our root node. Add the following code into this
...
+    public partial class XRInitializer : Node3D
+    {
+        public override void _Ready()
+        {
+            XRInterface xrInterface = XRServer.FindInterface("OpenXR");
+            if (xrInterface != null && xrInterface.IsInitialized())
+            {
+                GetViewport().UseXR = true;
+
+                GD.Print("OpenXR Initialized Successfully!");
```
raulsntos, line 104: The GDScript example has these lines in a different order. We should try to be consistent.

```suggestion
                GD.Print("OpenXR initialized successfully");

                GetViewport().UseXR = true;
```

tutorials/xr/setting_up_xr.rst
```diff
@@ -89,6 +90,24 @@ Next we need to add a script to our root node. Add the following code into this
...
+            }
+            else
+                GD.Print("OpenXR failed init... check headset connection!");
```
raulsntos, line 107: Use the same message as the GDScript example. Also, our preferred style is to use curly brackets even when not necessary.

```suggestion
            else
            {
                GD.Print("OpenXR not initialized, please check if your headset is connected");
            }
```

tutorials/xr/setting_up_xr.rst
```diff
@@ -89,6 +90,24 @@ Next we need to add a script to our root node. Add the following code into this
 		else:
 			print("OpenXR not initialised, please check if your headset is connected")
 
+  .. code-tab:: csharp
+
+    public partial class XRInitializer : Node3D
```
raulsntos, line 95: Since the root node in the scene is called `Main`, I think this class should use the same name.

```suggestion
    public partial class Main : Node3D
```

raulsntos (changes requested):
Thanks for contributing to the C# documentation!

I made a few comments about the C# example, but I also noticed some issues with the existing GDScript example and it would be great if you could fix them as well while you are updating this page:

- The GDScript example uses tabs, I think all our examples in the documentation use 4 space indentation for GDScript code.
- The verb `initialise` should be `initialize` since we use american english in the documentation.

VenomCodeDev:
Dont have the ability to update the GDScript tabs->spaces at the moment. Will gladly do that tomorrow if its still something you'd like done.

mhilbrunner (approved):

PR #6887, opened 2023-03-04
Fixed NotificationWMCloseRequest c# code

Tobi-Mob:
With Godot 4.0 the casing seems to have changed from Wm to WM.
But i don't know, if that was intended 

image


raulsntos (approved):
Thank you for contributing to the C# documentation!

It is indeed intended since [.NET naming guidelines](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions#capitalization-rules-for-identifiers) recommend to use PascalCase for acronyms _**except**_ for 2 letter acronyms, as is this case.

akien-mga:

PR #6890, opened 2023-03-04
Fix Getting State Machine Example for Godot 4

rayanmargham:
Really small fix but I've changed line 315 to work with Godot 4


YuriSizov:
This needs a rebase, and the C# bit also needs updating, if this is indeed a correct change.

AThousandShips:
Closing due to lack of response

PR #6896, opened 2023-03-05
Fix GDScript code style regarding colon (docs)

dalexeev:
[GDScript style guide](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html) recommends the `var name: Type` style (no space before the colon). The same goes for [properties](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_basics.html#properties-setters-and-getters) (it's not in the style guide yet, but vnen confirmed it in Rocket Chat).

Related:

* godotengine/godot#74306
* godotengine/godot-demo-projects#874

### DIFF
--- a/getting_started/first_2d_game/05.the_main_game_scene.rst
+++ b/getting_started/first_2d_game/05.the_main_game_scene.rst
@@ -73,7 +73,7 @@ Your scene should look like this:
 Main script
 ~~~~~~~~~~~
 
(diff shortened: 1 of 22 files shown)


akien-mga (approved):

akien-mga:
Thanks!

PR #6899, opened 2023-03-05
Updates getting_started image with godot 4.0 contents

wirecat:
Node names have changed and the image was out of date with the text documentation. The `AnimatedSprite` and `Sprite` nodes in this image should have `2D` appended to their names.

Unsure if something changed in Godot 4.0 so that the `CanvasItem` is hidden, but I couldn't figure out how to make it visible for the picture. If we can't show `CanvasItem` in the tree anymore, maybe the text should be updated to not mention `CanvasItem`?
**Edit: That was a me issue. I was looking at the Scene dock instead of the Create New Node window. Thanks @cbscribe for pointing it out.**

# Example Of Disconnect
image


cbscribe:
Screen Shot 2023-03-05 at 10 50 29
CanvasItem shows up when I search for Sprite2D...

wirecat:
> Screen Shot 2023-03-05 at 10 50 29 CanvasItem shows up when I search for Sprite2D...

@cbscribe Thank you. I for some reason thought that was a picture of the Scene dock. Updated to be a picture of the Create New Node window.

cbscribe:
Looks good to me. Can you squash your commits, please?

wirecat:
@cbscribe Sure thing. They're squashed.

cbscribe:
Thanks!

Issue #6902, opened 2023-03-05
Using CharacterBody2D/3D uses outdated method in example


gfnekro:
**Your Godot version:**
v4.0.stable.official [92bee43ad]

**Issue description:**
Tutorial code refers to outdated method:

-  `get_slide_count()` should be `get_slide_collision_count()`

In the body of the explanation the method is the right one

**URL to the documentation page:**
https://docs.godotengine.org/en/stable/tutorials/physics/using_character_body_2d.html

PR #6903, opened 2023-03-05
Fix incorrect method name in Using CharacterBody2D

cbscribe:
The text refers to the correct method, but the code sample still had `get_slide_count()`


rm-code:
This looks like it should fix #6902

akien-mga (approved):

PR #6907, opened 2023-03-06
FAQ DOD section: rephrase text and fix some grammar

puchik:
- Rephrased some awkward grammar in the first paragraph (I was going to reword it but realized the performance part is kind of redundant and ended up shortening it, too)
- Restructured and rephrased some of the grammar in the middle and end sections to be less repetitive and avoid run-on sentences.
- Fixed capitalization in the hyphenated word in the section title.

### DIFF
--- a/about/faq.rst
+++ b/about/faq.rst
@@ -532,26 +532,25 @@ removed at run-time to dynamically add and remove behaviors.
 More information about Godot's design choices can be found in
 `this article <https://godotengine.org/article/why-isnt-godot-ecs-based-game-engine>`__.
 
-Why does Godot not force users to implement DoD (Data-oriented Design)?
+Why does Godot not force users to implement DOD (Data-Oriented Design)?
 -----------------------------------------------------------------------
 
-While Godot internally for a lot of the heavy performance tasks attempts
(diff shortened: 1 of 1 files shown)


about/faq.rst
```diff
@@ -532,26 +532,25 @@ removed at run-time to dynamically add and remove behaviors.
 More information about Godot's design choices can be found in
 `this article <https://godotengine.org/article/why-isnt-godot-ecs-based-game-engine>`__.
 
-Why does Godot not force users to implement DoD (Data-oriented Design)?
+Why does Godot not force users to implement DoD (Data-Oriented Design)?
```
akien-mga, line 535: It should match the acronym, so either "DoD (Data-oriented Design)" or "DOD (Data-Oriented Design)".

puchik: Makes sense. A quick search showed DOD was more common than DoD (and is more consistent with OOP etc) so I changed the other ones to DOD. Doesn't seem to be any other references to DoD anywhere else in the docs so that should be it.

mhilbrunner (approved):

mhilbrunner:
Thank you, merged!

PR #6912, opened 2023-03-06
Update 09.adding_animations.rst

nathanjwtx:
I am not sure if `translation` is the correct term for what we are changing but in the editor itself it is called `position` and I found this a bit confusing to begin with. I replaced the `translation` with `position` in the 3 places I found.

If `translation` is the correct term, could an explanation be added to the top of the page?

### DIFF
--- a/getting_started/first_3d_game/09.adding_animations.rst
+++ b/getting_started/first_3d_game/09.adding_animations.rst
@@ -77,7 +77,7 @@ click any of them to create a keyframe, a time and value pair for the
 corresponding property. The keyframe gets inserted where your time cursor is in
 the timeline.
 
-Let's insert our first keys. Here, we will animate both the translation and the
+Let's insert our first keys. Here, we will animate both the position and the
 rotation of the ``Character`` node.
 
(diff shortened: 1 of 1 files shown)


YuriSizov:
Yes, translation is a correct term for 3D transformations, but indeed we renamed the property to position in 3D to match 2D, so your change here is correct.

nathanjwtx:
> Yes, translation is a correct term for 3D transformations, but indeed we renamed the property to position in 3D to match 2D, so your change here is correct.

Thank you for the clarification. It did seem more than a simple typo but I wasn't sure.

Piralein:
This change was already done by #6811.

current file:

- [line 80](https://github.com/godotengine/godot-docs/blame/master/getting_started/first_3d_game/09.adding_animations.rst#L80)
- [line 96](https://github.com/godotengine/godot-docs/blame/master/getting_started/first_3d_game/09.adding_animations.rst#L96)
- [line 166](https://github.com/godotengine/godot-docs/blame/master/getting_started/first_3d_game/09.adding_animations.rst#L166)

YuriSizov:
Ah, indeed. Sorry for the wait, but this was indeed fixed about the same time you made your PR :)

PR #6915, opened 2023-03-06
Fix punctuation in list_of_features.rst

JohnVeness:
Just a tiny punctuation change!

### DIFF
--- a/about/list_of_features.rst
+++ b/about/list_of_features.rst
@@ -450,7 +450,7 @@ Scripting
 
 - Actively developed GDExtension bindings for `D <https://github.com/godot-dlang/godot-dlang>`__,
   `Haxe <https://hxgodot.github.io/>`__, `Python <https://github.com/touilleMan/godot-python>`__, and `Rust <https://github.com/godot-rust/gdextension>`__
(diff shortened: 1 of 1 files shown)


YuriSizov:
When the change is this specific, it would be nice for the commit message to also be more specific. Though this is true in general, even if you do a pass on the entire page. I.e. the title and the commit message should point at what kind of improvement it is, not just to the fact that it has been improved.

JohnVeness:
> When the change is this specific, it would be nice for the commit message to also be more specific. Though this is true in general, even if you do a pass on the entire page. I.e. the title and the commit message should point at what kind of improvement it is, not just to the fact that it has been improved.

You're right, sorry about that. I admit I find it hard to think of commit titles for general proof-reading type stuff, but I will try harder.

YuriSizov:
You can amend the message and force push to this same PR. Just stating that you improve punctuation in such and such article would be enough.

JohnVeness:
I think I have updated (i.e. force pushed) this correctly. Thanks for the help.

mhilbrunner (approved):

mhilbrunner:
Thank you! Merged. (It still contained a stray merge commit though :))

PR #6919, opened 2023-03-07
Updates images to display new Sprite2D

wirecat:
Was displaying the old Sprite name and I thought that could be confusing for people new to the engine.


YuriSizov:
Thanks! Can you run the images through oxipng? They seem to be noticeably bigger in size despite not having significantly more details in the 4.0 version. You can use something like `oxipng -o6 --strip all --zopfli --fix`.

YuriSizov (approved):
Thanks, seems to be better now.

wirecat:
> Thanks, seems to be better now.

Sure thing! I just squashed everything down as well.

mhilbrunner:
@wirecat Thanks for contributing! Terribly sorry to ask this after you've just re-converted them, but could you convert them to WebP? The docs are in the process of switching over to WebP, so all new images should be WebP. (See [image contribution guidelines](https://docs.godotengine.org/en/stable/contributing/documentation/docs_image_guidelines.html).

I think this was missed in the above review as its a pretty recent change.

wirecat:
> @wirecat Thanks for contributing! Terribly sorry to ask this after you've just re-converted them, but could you convert them to WebP? The docs are in the process of switching over to WebP, so all new images should be WebP. (See [image contribution guidelines](https://docs.godotengine.org/en/stable/contributing/documentation/docs_image_guidelines.html)).
> 
> I think this was missed in the above review as its a pretty recent change.

Sure thing! Everything is converted and I believe I have all the references to these images updated.

mhilbrunner (approved):

mhilbrunner:
Merged. 🎉 Thanks for converting them (twice) and taking the time to update images, thats very welcome :)

PR #6922, opened 2023-03-07
Update matrices_and_transforms.rst

Skyvastern:
Removed mention of Godot 3.2 in current version of Godot 4. To me the text also assumed that the current version being used is 3.2.

### DIFF
--- a/tutorials/math/matrices_and_transforms.rst
+++ b/tutorials/math/matrices_and_transforms.rst
@@ -606,6 +606,9 @@ this project which has colored lines and cubes to help visualize the
 :ref:`class_Basis` vectors and the origin in both 2D and 3D:
 https://github.com/godotengine/godot-demo-projects/tree/master/misc/matrix_transform
 
+.. note:: You cannot edit Node2D's transform matrix directly in Godot 4.0's
+          inspector. This may be changed in a future release of Godot.
+
 If you would like additional explanation, you should check out
 3Blue1Brown's excellent video about 3D linear transformations:
 https://www.youtube.com/watch?v=rHLEWRxRGiM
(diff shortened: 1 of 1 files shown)


YuriSizov:
Yes, the text does seem to assume we're still at 3.2. But was this actually changed? If not, it might be worth preserving, but rephrasing instead.

tutorials/math/matrices_and_transforms.rst
```diff
@@ -606,14 +606,6 @@ this project which has colored lines and cubes to help visualize the
 :ref:`class_Basis` vectors and the origin in both 2D and 3D:
 https://github.com/godotengine/godot-demo-projects/tree/master/misc/matrix_transform
 
-.. note:: Spatial's "Matrix" section in Godot 3.2's inspector
-          displays the matrix as transposed, with the columns
-          horizontal and the rows vertical. This may be changed
-          to be less confusing in a future release of Godot.
-
-.. note:: You cannot edit Node2D's transform matrix directly in Godot 3.2's
-          inspector. This may be changed in a future release of Godot.
```
aaronfranke, line 615: The first note is no longer a problem in Godot 4.0, the 3D matrix on Node3D was corrected.

The second note is still relevant for Godot 4.0, the inspector does not let you edit the 2D matrix on Node2D.

aaronfranke:
> But with Godot 4.0, I can't even find matrix options for Node3D and its derived nodes.

Change "Rotation Edit Mode" to "Basis" to view the matrix.

tutorials/math/matrices_and_transforms.rst
Skyvastern: I see. So I think we can remove the first note, and keep the second note. Just have to update version number to 4.0.

aaronfranke:
Yes, the position is the origin of the transformation matrix. They are exactly equivalent. Same as the matrix's last column.

Skyvastern:
> Yes, the position is the origin of the transformation matrix. They are exactly equivalent. Same as the matrix's last column.

Yeah, for this will need to read the docs again to see if there's any place where changes/additions are required. For that will make a separate PR. For this one though, I've readded the 2nd note, and removed the first one. I think that should do it.

aaronfranke (approved):

aaronfranke:
Thanks!

PR #6929, opened 2023-03-08
clarification of "connecting a singal via code"

terminatorbs:
The previous example for "connecting a signal via code" used the editor extensively. 
This version uses only code to spawn a timer and connect the relevant signal, and is a concise, standalone example of how it's done.
I believe that not using the editor at all is more appropriate for this section.

Note: 
This does not include the C# code, and some formatting might be slightly off.

### DIFF
--- a/getting_started/step_by_step/signals.rst
+++ b/getting_started/step_by_step/signals.rst
@@ -265,7 +265,7 @@ instead of ``timeout``.
 If we call the ``timed_exit_game()`` function, a timer of 5 seconds will start, and 
 on timeout emits the ``timeout`` signal which calls the exit function.
 
-.. note:: Alternatively to ``timer.connect(signal_name, function)``, you can use 
+.. note:: Alternatively to ``timer.connect("timeout", function)``, you can use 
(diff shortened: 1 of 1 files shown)


terminatorbs:
Thanks for taking a look! 
And if it's a conscious decision that's fine. In case it stays as is, though, I'd like to mention this:
The initial reason to consider changing this was that both me over a year ago when starting to use godot, and recently my friend too, were confused about how to connect custom signals. 

Specifically, in the line "timer.timeout.connect(_on_Timer_timeout)" someone who knows Godot already obviously understands that the "timeout" part is the signal, and that this can be changed to an arbitrary signal name for custom signals. But for a new user reading about signals the first time, I feel like the connect("signal_name", called_function) way of connecting is much more clear - I'd be happy if we could include at least that part somewhere.

If the current way of connecting is preferred due to performance or editor-specific features (autocomplete? others?), maybe adjust the note at line 265 to say that this signal name can be changed for custom signals, or add a similar note in the custom signals section.

PR #6937, opened 2023-03-09
Fix typo in Wikipedia open-source hyperlink

Repiteo:
Though the previous link works fine if clicked directly, the conversion to the webpage erroneously removes the space character. Because the space character just gets redirected to a hyphen anyway, it avoids the issue altogether to replace it outright


mhilbrunner (approved):

mhilbrunner:
Merged. Thanks for contributing!

PR #6968, opened 2023-03-12
Your first 3D game - Update 09.adding_animations.rst

hans-min:
- Update [09.adding_animations.rst:96](https://github.com/godotengine/godot-docs/blob/8a19f445324806e0da2b22bd079bd4470a108887/getting_started/first_3d_game/09.adding_animations.rst?plain=1#L96) mentioned in #6744 
- Add guide on how to add new keyframes on [09.adding_animations.rst:109](https://github.com/godotengine/godot-docs/blob/8a19f445324806e0da2b22bd079bd4470a108887/getting_started/first_3d_game/09.adding_animations.rst?plain=1#L109) 
- Add `float()` to turn integer division into float division + note on [09.adding_animations.rst:284](https://github.com/godotengine/godot-docs/blob/8a19f445324806e0da2b22bd079bd4470a108887/getting_started/first_3d_game/09.adding_animations.rst?plain=1#L284)

### DIFF
--- a/getting_started/first_3d_game/09.adding_animations.rst
+++ b/getting_started/first_3d_game/09.adding_animations.rst
@@ -93,7 +93,7 @@ Two tracks appear in the editor with a diamond icon representing each keyframe.
 |image11|
 
 You can click and drag on the diamonds to move them in time. Move the
-position key to ``0.2`` seconds and the rotation key to ``0.1`` seconds.
+position key to ``0.3`` seconds and the rotation key to ``0.1`` seconds.
 
 |image12|
 
@@ -106,7 +106,7 @@ In the *Inspector*, set the *Position*'s *Y* axis to ``0.65`` meters and the *Ro
 
 |image13|
 
-Create a keyframe for both properties
+Create a keyframe for both properties by right clicking on the track and choose *Insert Key*
 
 .. image:: img/09.adding_animations/second_keys_both.webp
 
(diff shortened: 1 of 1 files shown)


getting_started/first_3d_game/09.adding_animations.rst
```diff
@@ -271,7 +271,8 @@ following line.
 
    func initialize(start_position, player_position):
        #...
-       $AnimationPlayer.speed_scale = random_speed / min_speed
+       
+       $AnimationPlayer.speed_scale = float(random_speed) / min_speed
```
aaronfranke, line 275: Actually these values should be floats. They are only used in a float context and it is perfectly sensible for a mob to have a speed of `10.5` meters per second. https://github.com/godotengine/godot-docs/pull/6243#discussion_r1133403500

hans-min: right, I think we can change `var random_speed = randf_range(min_speed, max_speed)` instead of `randi-range`. But I also like the to keep it an int and teach the newbies about integer/float division.

aaronfranke: I'm sure we can find some other place to teach this concept. It doesn't seem right to purposefully keep this code wrong (aka using ints) for the sake of showing how to work around a problem.

hans-min: Yep that's better, I agree

PR #6976, opened 2023-03-13
Document the Visual Profiler debugger tab, update some screenshots to 4.0

Calinou:


tutorials/scripting/debug/debugger_panel.rst
```diff
@@ -39,29 +39,118 @@ You can use the buttons in the top-right corner to:
...
+frame) or a percentage of the target frametime (which is currently hardcoded to
+16.67 milliseconds, or 60 FPS).
+
+If framerate spikes occur during profiling, this can cause the graph to be
+poorly scaled. Disable **Fit to Frame** so that the graph will zoom onto the 60
+FPS+ portion.
+
+.. note::
+
+    Remember that Visual Profiler results will vary **heavily** based on
```
clayjohn, line 104:
```suggestion
    Remember that Visual Profiler results can vary **heavily** based on
```

This won't be the case for all users. It depends on where their bottlenecks are. If the game is CPU-bottlenecked window size won't affect performance. Similarly if the GPU bottleneck is a memory load bottleneck in the vertex shader, then changing window size may not affect performance either. 

But, the advice is good in most situations as Godot tends to hit fill-rate bottlenecks pretty quickly. So it is worth keeping this note in place, we should just soften the language a little bit

clayjohn (commented):
The text looks great! I haven't reviewed the images but I trust they are good as well. 

I just left one note with a suggestion to tweak some language

mhilbrunner (approved):

mhilbrunner (approved):

PR #6979, opened 2023-03-14
Add concatenation to shader preprocessor

JohanAR:

### DIFF
--- a/tutorials/shaders/shader_reference/shader_preprocessor.rst
+++ b/tutorials/shaders/shader_reference/shader_preprocessor.rst
@@ -47,14 +47,29 @@ General syntax
 Defines the identifier after that directive as a macro, and replaces all
 successive occurrences of it with the replacement code given in the shader.
 Replacement is performed on a "whole words" basis, which means no replacement is
-performed if the string is part of another string (without any spaces separating
-it).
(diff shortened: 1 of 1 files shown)


JohanAR:
For https://github.com/godotengine/godot/pull/74737 (not merged yet)

bitsawer:
Now that https://github.com/godotengine/godot/pull/74737 has been merged, this PR can be updated too.

Looks good, but I would tweak the shader code example a bit. First, we can change it into a fully functional shader snippet. It's only a few lines more and users can actually copy paste and see that it works. Another is to remove `;` from the SAMPLE macro content so that users must place it everywhere the macro is used for consistency with other code. So the new code snippet would be like this:

```glsl
uniform sampler2D material0;

#define SAMPLE(N) vec4 tex##N = texture(material##N, UV)

void fragment() {
    SAMPLE(0);
    ALBEDO = tex0.rgb;
}
```

JohanAR:
Thanks! Updated as suggested and rebased.

bitsawer (approved):
Tested locally without any problems, looks good to me.

akien-mga (approved):

akien-mga:
Thanks!

Issue #6948, opened 2023-03-10
Misleading example about casting String as int.


hristo-kanchev:
**Your Godot version:**
4.0.stable

**Issue description:**
Casting.

It seems that this part in the docs is wrong:

```gdscript
# For built-in types, they will be forcibly converted if possible, otherwise the engine will raise an error.

var my_int: int
my_int = "123" as int # The string can be converted to int.
```

If you try to do this right now in Godot 4.0.stable, you will get an Error.

Cannot cast a value of type "String" as "int"

**URL to the documentation page:**
https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_basics.html#casting

If you know how to fix the issue you are reporting please
consider opening a pull request. We provide a tutorial on
using git here: https://docs.godotengine.org/en/stable/community/contributing/pr_workflow.html,
writing documentation at https://docs.godotengine.org/en/stable/community/contributing/docs_writing_guidelines.html
and contributing to the class reference here: https://docs.godotengine.org/en/stable/community/contributing/updating_the_class_reference.html

UnhappyTurnip:
If I recall correctly, casting with 'as' is used when casting a base class into a derived class, and since int does not derive from string, it's no surprise it throws an error. I will do some testing and create a PR updating the information.

PR #6982, opened 2023-03-14
Rewrite of casting explanation

UnhappyTurnip:
The previous explanation of casting was, in some cases, confusing, and in some cases, completely wrong, as outlined by issue #6948. This patch should fix all of these issues.


UnhappyTurnip:
The code should be working now, fingers crossed.

PR #6991, opened 2023-03-16
Update and improve BBCode in RichTextlabel documentation for Godot 4.0

Calinou:
- Update screenshots for Godot 4.0.
- Document RichTextLabel having its own vertical scrollbar.
- Document `push_[tag]()` and `pop()` functions as a faster alternative to BBCode.
- Document threading support and when to use it.
- Document how to escape user input to prevent it from having its BBCode tags parsed.
- Document stripping BBCode tags from the text.
- Mention entangled tags not being supported.
- Mention faux bold/italic versus real bold/italic fonts.
- Document `[hint]` and `[dropcap]` BBCode tags.
- Mention animated RichTextEffects pausing when the scene tree is paused and how to control this behavior.
- Modify RichTextEffect examples to use the actual default values.
- Add screenshot showcasing how to add a RichTextEffect to the Custom Effects inspector property.
- Fix and improve `[pulse]` custom RichTextEffect. This adds an easing parameter and makes the target color a multiplier, so that colored text keeps its original color while fading in and out:

https://user-images.githubusercontent.com/180032/225697474-f873a97e-1645-4971-a98c-e52fedecc5d9.mp4

This closes https://github.com/godotengine/godot-docs/issues/6924.

@bruvzg Is there a built-in function to escape user input to prevent it from being parsed as BBCode? I've done this in a script, but this makes me wonder if such a feature should be built-in (as it'll often be needed for multiplayer games with chat).

PR #6993, opened 2023-03-16
Document embolden and font spacing in Using fonts

Calinou:
- Follow-up to https://github.com/godotengine/godot-docs/pull/6991.

This also mentions what the default project font is for reference.

PS: Is there a way we could display subsections in the table of contents on the left? There are a lot of subsections that would benefit from being listed under **Advanced font features**:

image


mhilbrunner (approved):

mhilbrunner:
Looks good to me, merged 🎉 

> PS: Is there a way we could display subsections in the table of contents on the left? There are a lot of subsections that would benefit from being listed under Advanced font features:

Not that I'm aware of, and too many levels of navigation also get hard to manage/navigate. Maybe it may make sense to instead split that section into multiple, not sure.

PR #7021, opened 2023-03-19
Full Rect option in Background section

kastaldi:
Corrected the "Full Rect" option location in Background section because it was wrong in Godot 4.

### DIFF
--- a/getting_started/first_2d_game/07.finishing-up.rst
+++ b/getting_started/first_2d_game/07.finishing-up.rst
@@ -15,7 +15,7 @@ The default gray background is not very appealing, so let's change its color.
 One way to do this is to use a :ref:`ColorRect <class_ColorRect>` node. Make it
 the first node under ``Main`` so that it will be drawn behind the other nodes.
 ``ColorRect`` only has one property: ``Color``. Choose a color you like and
-select "Layout" -> "Full Rect" so that it covers the screen.
+select "Layout" -> "Anchors Preset" -> "Full Rect" so that it covers the screen.
 
 You could also add a background image, if you have one, by using a
 ``TextureRect`` node instead.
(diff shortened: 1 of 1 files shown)


YuriSizov:
This part is actually about the toolbar at the top of the viewport, not the inspector.

getting_started/first_2d_game/07.finishing-up.rst
```diff
@@ -15,7 +15,7 @@ The default gray background is not very appealing, so let's change its color.
 One way to do this is to use a :ref:`ColorRect <class_ColorRect>` node. Make it
 the first node under ``Main`` so that it will be drawn behind the other nodes.
 ``ColorRect`` only has one property: ``Color``. Choose a color you like and
-select "Layout" -> "Full Rect" so that it covers the screen.
+select "Layout" -> "Anchors Preset" -> "Full Rect" so that it covers the screen.
```
mhilbrunner, line 18:
```suggestion
select "Layout" -> "Anchors Preset" -> "Full Rect" either in the toolbar at the top of the viewport or in the inspector so that it covers the screen.
```

mhilbrunner:
I think this phrasing is fine with the top toolbar too. I made it more explicit so users are aware of both ways.

mhilbrunner (approved):

mhilbrunner:
Thanks and congrats on your first merged contribution to Godot's docs! 🎉

YuriSizov:
> I think this phrasing is fine with the top toolbar too. I made it more explicit so users are aware of both ways.

Well, the problem is that in the toolbar it used to be written in text, "Layout" and "Full Rect". Now it's all icons in the toolbar, so this instruction is unclear, as the text, even after your edit, doesn't match the UI. I should've been more explicit in my first comment, but the solution here should've been to either explicitly refer to the inspector, or to describe the toolbar and add a screenshot of the UI.

PR #7024, opened 2023-03-20
Updating the naming in accordance with latest in signals.rts

Glukozavr:
According to the latest Godot version the naming convention for the signals handlers is as follows
_on_node_name_signal_name
, while in the docs stated _on_NodeName_signal_name which appears to be incorrent.


skyace65 (changes requested):
The new image needs to be changed to a .webp file, we're changing everything over to that format as things get updated. You can use [this ](https://squoosh.app/)to convert your image to webp quickly. Also when making a PR please try to keep everything as 1 commit, you can find information on how to do that [here](https://docs.godotengine.org/en/latest/contributing/workflow/pr_workflow.html) if you don't know how.

skyace65:
Superseded by #7092.

PR #7041, opened 2023-03-23
Specify what "changes" means

### LINKED ISSUE #7040: Specify what "changes" means inside Range.value_changed signal
**Issue description:**
It is not clear enough that the `value_changed` signal on a Range-based node will not be called if the new value is equal to the previous value. One might make the assumption that when they set the value of a Range via code, it will always emit this signal, therefore it should be made clear that when the new value equals the previous value, the signal [will not be sent](https://github.com/godotengine/godot/blob/1d14c054a12dacdc193b589e4afb0ef319ee2aae/scene/gui/range.cpp#L81).

[https://github.com/godotengine/godot-docs/blob/33c61914b4207b5864bc9008c225a7f1d7e28465/classes/class_range.rst](https://github.com/godotengine/godot-docs/blob/33c61914b4207b5864bc9008c225a7f1d7e28465/classes/class_range.rst)

shunlog:
fixes  #7040

### DIFF
--- a/classes/class_range.rst
+++ b/classes/class_range.rst
@@ -100,7 +100,7 @@ Emitted when :ref:`min_value<class_Range_property_min_value>`, :ref:`max_value<c
 
 **value_changed** **(** :ref:`float<class_float>` value **)**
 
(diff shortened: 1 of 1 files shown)


YuriSizov:
https://github.com/godotengine/godot/tree/master/doc/classes/Range.xml

You can read more about contributing to the class reference [in our documentation](https://docs.godotengine.org/en/latest/contributing/documentation/updating_the_class_reference.html).

PR #7042, opened 2023-03-23
Corrected virtual function name in "Custom GUI Controls"

Mercerenies:
It is not possible to directly override non-virtual native methods (like `get_minimum_size`), so we have to override `_get_minimum_size` (the virtual variant) instead. Attempting the former produces an error in Godot 4, indicating that the method override will have no effect. I'm not sure if this is new to Godot 4 or applies to 3.x as well.

**Bugsquad note for cherrypicking:** verify whether this applies to `3.x` before cherrypicking.

### DIFF
--- a/tutorials/ui/custom_gui_controls.rst
+++ b/tutorials/ui/custom_gui_controls.rst
@@ -79,13 +79,13 @@ the minimum size will make sure your custom control is not squished by
 the other controls in the container.
 
 To provide this callback, just override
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):

mhilbrunner:
Thanks for contributing to Godot's docs! Looks correct for 4.0. Merged! 🎉 

**Note for cherrypicking:** verify whether this applies to `3.x` first.

PR #7045, opened 2023-03-23
fix paths used in gdextension example

ganymoe:
fixes godotengine#7044

### DIFF
--- a/tutorials/scripting/gdextension/gdextension_cpp_example.rst
+++ b/tutorials/scripting/gdextension/gdextension_cpp_example.rst
@@ -349,9 +349,9 @@ loaded for each platform and the entry function for the module. It is called ``g
 
     [libraries]
 
-    linux.64="res://demo/bin/libgdexample.linux.64.so"
-    windows.x86_64="res://demo/bin/libgdexample.windows.x86_64.dll"
-    macos="res://demo/bin/libgdexample.macos.framework"
+    linux.64="res://bin/libgdexample.linux.64.so"
+    windows.x86_64="res://bin/libgdexample.windows.x86_64.dll"
+    macos="res://bin/libgdexample.macos.framework"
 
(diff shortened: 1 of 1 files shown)


mhilbrunner:
Hm, this was just changed in https://github.com/godotengine/godot-docs/pull/6873 by @paddy-exe 3 weeks ago. @paddy-exe can you maybe see which one is correct? :)

ganymoe:
@paddy-exe

I think the SConstruct file is fine with adding the demo path, as this is used for the actual "godot-project-basefolder" in the GDExtension example, as the structure describes:

```
gdextension_cpp_example/
|
+--demo/                  # game example/demo to test the extension
|   |
|   +--main.tscn
|   |
|   +--bin/
|       |
|       +--gdexample.gdextension
|
+--godot-cpp/             # C++ bindings
|
+--src/                   # source code of the extension we are building
|   |
|   +--register_types.cpp
|   +--register_types.h
|   +--gdexample.cpp
|   +--gdexample.h
|
+--SConstruct
```

for example if you load the main scene like `var main = load("res://main.tscn")` you won't add demo in front of main.tscn as well. right?

paddy-exe (approved):
You are right 👍🏻 seems like I, as we Germans say "verschlimmbesserte" (tried to solve it but made it worse) it. Thanks for noticing. Not sure then though what caused https://github.com/godotengine/godot-docs/issues/6872...
@mhilbrunner should be good to merge

ganymoe:
> as we Germans say "verschlimmbesserte" 

yeah, i know that term in Austria as well ;)  :wave:

mhilbrunner:
Everyone speaking german in this thread :) Anyway thanks for the review paddy-exe, and congrats on your first merged contribution to Godot's docs, ganymoe!

PR #7049, opened 2023-03-24
HTTPRequest GDScript adoption to new API

MartinMajewski:
I reviewed the HTTPRequest introduction examples (https://docs.godotengine.org/en/stable/tutorials/networking/http_request_class.html) and realized that they are not in sync with the current GDScript API of the Signal and HTTPRequest classes.

1. The `connect` signature changed, where you have to provide a callback function directly (1st class citizen) and a flag on how the signal connection shall be maintained.

2. `JSON.parse` is not a static member anymore. Using it would require initializing and storing it inside a variable and a more elaborate error handling which would go beyond the example's simplicity. Using the `.parse_string` function and printing the JSON string to the console is much easier, which perfectly demonstrates the core idea of using HTTPRequest.

Furthermore, I made adjustments to the wording of some sentences.

### DIFF
--- a/tutorials/networking/http_request_class.rst
+++ b/tutorials/networking/http_request_class.rst
@@ -78,7 +78,7 @@ With this, you should see ``{ "hello": "world" }`` printed on the console, with 
 
 For more information on parsing JSON, see the class references for :ref:`JSON <class_JSON>`. Here you'll also find information how to handle JSON parsing errors.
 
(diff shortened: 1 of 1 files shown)


mhilbrunner:
Fixed in https://github.com/godotengine/godot-docs/pull/7183. Thanks for contributing!

PR #7052, opened 2023-03-24
Update high_level_multiplayer.rst

NickCheng0921:
Old documentation used set_network_master instead of new method set_multiplayer_authority. Making this small update since rest of documentation for this page looks correct, but it took me a while to figure out why the old method wasn't working.


TheYellowArchitect:
Half that page needs rework imo. Further down there is:
>The second use is to specify how the function will be called via RPC. There are four different keywords:
> remote
>remotesync
>master
>puppet

There is also MultiplayerSynchronizer&MultiplayerSpawner which are part of High-Level multiplayer, yet are unmentioned.

That said, this change is good.

mhilbrunner:
FYI, a rework of the multiplayer docs is in progress.

PR #7053, opened 2023-03-25
Update/Clarify gdextension docs

paddy-exe:
This PR clarifies the GDExtension docs that you either add `godot-cpp` as a submodule OR clone it to your project folder.
It also updates the branch version being used to `4.0` instead of `master` before since 4.0 is stable and master isn't always compatible with the latest master build of Godot itself.


mhilbrunner (approved):

mhilbrunner:
Looks good! Merged.

Issue #7054, opened 2023-03-25
Web export broken in 4.x, not "significant bugs"


netAction:
**Your Godot version:**

4.0.1

**Issue description:**

The manual says: **There are significant bugs when running HTML5 projects on iOS**. This is wrong in two ways. First, it is broken, not buggy. There are multiple issues open explaining why HTML5 exports do not work on Ios.
https://github.com/godotengine/godot/issues/75281
https://github.com/godotengine/godot/issues/70621
https://github.com/godotengine/godot/issues/74512

Secondly, the text is a link with a reference to a filtered list that doesn't even contain these issues.
https://github.com/godotengine/godot/issues?q=is:issue+is:open+label:platform:html5+ios
The list contains some old bugs from 3.x times when HTML5 exports worked on Ios.

So the docs lack the fundamental information about the missing Ios support. It even contains wrong information about minor bugs that makes users think these were the largest issues.

**URL to the documentation page:**

https://github.com/godotengine/godot-docs/blob/master/tutorials/export/exporting_for_web.rst
https://docs.godotengine.org/en/stable/tutorials/export/exporting_for_web.html

PR #7056, opened 2023-03-25
Document missing macOS/iOS support for HTML5 due to upstream bugs

Calinou:
- This closes https://github.com/godotengine/godot-docs/issues/7054.

### DIFF
--- a/tutorials/export/exporting_for_web.rst
+++ b/tutorials/export/exporting_for_web.rst
@@ -19,11 +19,16 @@ in the user's browser.
                with :kbd:`F12`, to view **debug information** like JavaScript,
                engine, and WebGL errors.
 
-.. attention:: `There are significant bugs when running HTML5 projects on iOS
-               <https://github.com/godotengine/godot/issues?q=is:issue+is:open+label:platform:html5+ios>`__
-               (regardless of the browser). We recommend using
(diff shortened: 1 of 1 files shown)


clayjohn:
I'm not sure this is correct. There are many users facing issues with macOS web export. But I don't think we should say it isn't supported. For many users web exports will work fine on Macs once safari merges their change to use ANGLE over Metal instead of ANGLE over GL

Calinou:
> I'm not sure this is correct. There are many users facing issues with macOS web export. But I don't think we should say it isn't supported. For many users web exports will work fine on Macs once safari merges their change to use ANGLE over Metal instead of ANGLE over GL

The PR isn't saying that it's not supported – just that in practice, HTML5 exports currently won't run on macOS and iOS.

mhilbrunner (approved):

mhilbrunner:
> For many users web exports will work fine on Macs once safari merges their change to use ANGLE over Metal instead of ANGLE over GL

I think pointing out that people shoukd not expect it to work _currently_ (and then removing this note once change is merged and widely deployed) is fine.

skyace65 (approved):

mhilbrunner:
Merged for now. Thanks!

PR #7059, opened 2023-03-25
[contributing/development/compiling/compiling_for_linuxbsd] update style; use directives

umarcor:
This PR updates the style of compiling guidelines for Linux. A clarification is moved to an admonition and tabs are used to show distro-specific one-liners.

### DIFF
--- a/contributing/development/compiling/compiling_for_linuxbsd.rst
+++ b/contributing/development/compiling/compiling_for_linuxbsd.rst
@@ -17,24 +17,33 @@ For compiling under Linux or other Unix variants, the following is
 required:
 
 - GCC 7+ or Clang 6+.
+
 - `Python 3.6+ <https://www.python.org/downloads/>`_.
-- `SCons 3.0+ <https://scons.org/pages/download.html>`_ build system. If your distribution uses Python 2 by default,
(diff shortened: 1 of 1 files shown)


YuriSizov:
Any specific reason you changed the format of the first list? Does the dashed list break when you add a note?

umarcor:
@YuriSizov it's the recommended syntax for unordered/bulleted lists in Sphinx: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#lists-and-quote-like-blocks.

YuriSizov:
Be that as it may, the existing syntax here is also valid. RST has a problem with competing/conflicting syntaxes for a few things, but unless we plan to enforce some specific style across the board (and we don't, as far as I can tell), I would prefer we kept changes in PRs to the minimum and avoid those which are unnecessary. Without a style guide adopted by the project this is just arbitrary, even if otherwise justified.

umarcor:
@YuriSizov now changed to use `-` instead of `*`.

mhilbrunner (approved):

mhilbrunner:
Thanks for cleaning this page up, and congrats on your first merged contribution to Godot's docs! 🎉

PR #7062, opened 2023-03-26
Clarifying some common pitfalls for Visual Studio debugging

armmah:
Two minor changes made to this page, to help other people avoid wasting their time:
1. The page says at the end that it should be possible to place a breakpoint and if the steps were followed properly it should just work, however the dev_buil paramter is not mentioned anywhere in the same page and I would not be able to complete the setup without help from @anvilfolk.
2. I spent a couple of minutes trying to understand why the debugger would not attach and instead would exit with code 255 - no errors or hints shown (like path not found, or could not parse, etc). Ended up being a simple mistake of having white spaces in the path. I think its worth mentioning in the docs.

P.S. I'd actually like to break the lines between vsproj and dev_build descriptions (in markup they are each on separate line but end up on same line), but I have no idea how to do it :p


contributing/development/configuring_an_ide/visual_studio.rst
```diff
@@ -47,7 +49,7 @@ Project Manager opens a project, the initial process is terminated and the debug
 - Open the **Debugging** section and under **Command Arguments** add two new arguments:
   the ``-e`` flag opens the editor instead of the Project Manager, and the ``--path`` argument
   tells the executable to open the specified project (must be provided as an *absolute* path
-  to the project root, not the ``project.godot`` file).
+  to the project root, not the ``project.godot`` file, also must not contain whitespaces in the path).
```
kleonc, line 52: You can pass path with spaces though (not sure about other whitespaces) by using double quotes. Like: `--path "some/path/name with spaces"`.

kleonc: > I could replace it to suggest always using the quotes, which should be the safer option.


Fine to me. But a note in here could also be added:
```suggestion
  to the project root, not the ``project.godot`` file; if the path contains spaces be sure to pass it inside double quotation marks).
```
or something like that.

armmah: Done

mhilbrunner:
The changes look good to me at first glance.

Please convert the PNG file to WebP while you're at it (we're in the process of moving the docs repository over to WebP, see [image creation guidelines](https://docs.godotengine.org/en/stable/contributing/documentation/docs_image_guidelines.html)).

contributing/development/configuring_an_ide/visual_studio.rst
```diff
@@ -15,7 +15,9 @@ Visual Studio requires a solution file to work on a project. While Godot does no
 with the solution file, it can be generated using SCons.
 
 - Navigate to the Godot root folder and open a Command Prompt or PowerShell window.
-- Run ``scons platform=windows vsproj=yes`` to generate the solution.
+- Run ``scons platform=windows vsproj=yes dev_build=yes`` to generate the solution with debug symbols.
+        The vsproj parameter signals that you want Visual Studio solution generated.
+        The dev_build parameter makes sure the debug symbols are included, to place breakpoints and step through code.
```
kleonc, line 20: I'd reword this as you can always "place breakpoints", they just wouldn't work without the debug symbols. :upside_down_face:
```suggestion
        The ``vsproj`` parameter signals that you want Visual Studio solution generated.
        The ``dev_build`` parameter makes sure the debug symbols are included, allowing to e.g. step through code using breakpoints.
```

I'm not sure about formatting/tabs, someone from the docs team should take care of that.

armmah:
> Please convert the PNG file to WebP while you're at it

Hmm, in that directory there are 68 PNGs currently. Will it make sense to convert only one image that I'm submitting even tho rest of the page will remain as PNG?

IMO it would make sense to do the conversion in a separate PR.

mhilbrunner:
> Hmm, in that directory there are 68 PNGs currently. Will it make sense to convert only one image that I'm submitting even tho rest of the page will remain as PNG?
> IMO it would make sense to do the conversion in a separate PR.

There were too many pictures on the docs to do the conversion at once, and its also needlessly introducing binary data if the images remain otherwise identical, so re-encoding just for the sake of it was deemed not worth it. Instead, we chose the policy that any newly introduced (or modified) images need to be in WebP.

A PR re-creating outdated images to update them for Godot 4 and converting them to WebP would be welcome. A PR just running a batch conversion over all images just for the sake of converting them to WebP wouldn't be.

> Is this meant only for full screen images to make sure they always fit in 1080p? Otherwise what's the point of downscaling if I captured a region of 500 pixels in height on my 1440p monitor it will be the same 500 pixels on 1080p monitor.

Yeah, this only applies to screenshots of content that is scaled by resolution; if you have a piece of content thats the same pixel size on both resolutions, downscaling is not needed. That paragraph could be a bit more clear maybe, agreed :)

armmah:
Alright, I will make those two changes as suggested (the webp format and the different wording on breakpoint).
Any tips on the line break? Or we just leave it as is and move forward with just those two modifications?

contributing/development/configuring_an_ide/visual_studio.rst
armmah: done

kleonc (approved):
LGTM (again: not sure about the formatting).

contributing/development/configuring_an_ide/visual_studio.rst
```diff
@@ -15,7 +15,9 @@ Visual Studio requires a solution file to work on a project. While Godot does no
 with the solution file, it can be generated using SCons.
 
 - Navigate to the Godot root folder and open a Command Prompt or PowerShell window.
-- Run ``scons platform=windows vsproj=yes`` to generate the solution.
+- Run ``scons platform=windows vsproj=yes dev_build=yes`` to generate the solution with debug symbols.
+        The ``vsproj`` parameter signals that you want Visual Studio solution generated.
+        The ``dev_build`` parameter makes sure the debug symbols are included, allowing to e.g. step through code using breakpoints.
```
mhilbrunner, line 20:
```suggestion
- | Run ``scons platform=windows vsproj=yes dev_build=yes`` to generate the solution with debug symbols.
  | The ``vsproj`` parameter signals that you want Visual Studio solution generated.
  | The ``dev_build`` parameter makes sure the debug symbols are included, allowing to e.g. step through code using breakpoints.
```

mhilbrunner (approved):

mhilbrunner:
Turns out you can use `|`s for literal formatting to keep linebreaks :) Anyway, thanks for contributing and congrats on your first merged PR for Godot's docs! Great work.

anvilfolk:
@armmah congrats on your first contribution!!!!!!

PR #7066, opened 2023-03-26
[ci] push to gh-pages on forks

umarcor:
In order to allow contributors to test their modifications without installing and running the tools locally, this PR modifies the CI build on forks. Instead of doing the `dummy` build, the default HTML is built and it's pushed to gh-pages. As a result, when contributors push their changes to any branch, they will have their `username.github.io/godot-docs` updated.

No history will be saved in branch gh-pages. So, each time CI is run, the content will be overwritten.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,8 @@ on:
 jobs:
   build:
     runs-on: ubuntu-20.04
+    permissions:
+      contents: write
     steps:
       - name: Checkout
         uses: actions/checkout@v3
@@ -23,6 +25,25 @@ jobs:
           bash _tools/format.sh
           codespell -I _tools/codespell-ignore.txt -x _tools/codespell-ignore-lines.txt {about,community,development,getting_started,tutorials}/**/*.rst
 
-      # Use dummy builder to improve performance as we don't need the generated HTML in this workflow.
       - name: Sphinx build
-        run: make SPHINXOPTS='--color -W' dummy
+        run: |
+          # On godotengine, use dummy builder to improve performance as we don't need the generated HTML in this workflow.
+          # On forks, build the HTML site, to push it to gh-pages.
+          case '${{ github.repository }}' in
+            godotengine/godot-docs) TARGET='dummy';;
+            *) TARGET='html';;
+          esac
+          make SPHINXOPTS='--color -W' "$TARGET"
+
+      - name: 🚀 Publish site to GitHub Pages
+        if: github.repository != 'godotengine/godot-docs'
+        run: |
+          cd _build/html
+          touch .nojekyll
+          git init
+          cp ../../.git/config ./.git/config
+          git add .
+          git config --local user.email "gotdot-docs@GitHubActions"
+          git config --local user.name "GitHub Actions"
+          git commit -a -m "update ${{ github.sha }}"
+          git push -u origin +HEAD:gh-pages


mhilbrunner:
One thing I just thought about: we'd need to make sure this wouldn't negatively impact search engines for the official docs, both for SEO (official docs bein downranked due to duplicate content) and just to avoid confusion as users may end up finding these instead of the official docs (its already a problem for some of the translations).

umarcor:
@mhilbrunner you can see https://umarcor.github.io/godot-docs/. Apart from the URL being signifficantly different (because the official docs are *not* hosted on GitHub), there is no favicon and the title is prefixed with `(DEV)`. Actually, I believe it would be desirable to add a favicon to the official docs (currently the RTD one is shown), and disable it on forks.

We might add an admonition to a `rst_prolog` which would be shown at the top of each page, similar to how RTD warns about PR versions. However, it seems that this repo does already have two of those admonitions but they are not added through the prolog feature. Do you know where are those defined?

Moreover, see https://github.com/umarcor/godot-docs/commit/b879b60b3a36ce31ef291edd34786f5ea68435e8. That's not part of this PR, but it is meant as an enhacement. Changing that, along with using an envvar for `github_user` as well will make the 'Edit on GitHub' links point to the forks, instead of this repo.

I also tried to change the "*master* branch" that is shown in the breadcrumbs. See https://github.com/umarcor/godot-docs/commit/a5febb249b2d8802c8cc61249cc37137228d2c9f. Unfortunately, that didn't work.

With regard to the RTD overlay, that seems to be enabled unconditionally. So, even when the docs are built on GitHub Actions, the overlay is always shown. I believe it would be interesting to disable that, because it would make it more obvious that forks are a single version build.

With regard to SEO and search engines, I don't know whether that would impact negatively or actually benefit. I believe it's a matter of faith in any case, because AFAIAA the algorithms they use are not public.

mhilbrunner:
> With regard to SEO and search engines, I don't know whether that would impact negatively or actually benefit. I believe it's a matter of faith in any case, because AFAIAA the algorithms they use are not public.

Not really, "Google may downrank your page if there are other pages with mostly or exactly the same content" is pretty well-documented, even by Google itself going as far back as [2006](https://developers.google.com/search/blog/2006/12/deftly-dealing-with-duplicate-content).

And we already have had the problem with some translations taking up search results and annoying users. If you aren't speaking spanish, its obvious a result is wrong if its in spanish, yet it still reduces usability if part of your Godot Google results is taken up by spanish docs pages (or in this case, clearly third-party Github pages with the same, but probably outdated content).

The solution for both of these is to make sure search engines don't index these pages, i.e., they have `<meta name="robots" content="noindex">` if not on Read The Docs.

mhilbrunner:
That seems to be the right place to me, but maybe @Calinou or @YuriSizov can confirm. We should also make sure to only add that for non-ReadTheDocs builds.

YuriSizov:
Yeah, you can modify that template. To check if we're on RTD and not elsewhere use `{% if READTHEDOCS %}`. It will affect local builds as well, of course, but that should be fine in this case.

umarcor:
I modified the template to include `<meta name="robots" content="noindex">`  when `not READTHEDOCS`.

mhilbrunner:
Hey! After quite some consideration, I think this approach has some challenges:

- Our build HTML is currently around 1.7 GB, Github says Pages sites and repositores should stay below 1 GB:
https://docs.github.com/en/pages/getting-started-with-github-pages/about-github-pages
- I don't think the typical docs contributor will take the time (or understand how) to look at the built page with their changes
- Where it would be interesting is for us reviewers to check the pages site for a contributor and be able to review without building locally, however, as it only randomly supports the most recent PR, thats going to be useless for this case the majority of the time
- The latter is also confusing for users: You submit a PR, it builds, submit another and try to review the first and your changes aren't there anymore
- Fixing the above by i.e. having one subfolder per open PR would multiply the size further, which seems already prohibitive

I would really love a workflow of "user submits docs PR, changes get built and published *somewhere* to review", but this implementation IMO isn't optimal. We should investigate salvaging this part of the PR to avoid forks/mirrors polluting Google results:
> I modified the template to include <meta name="robots" content="noindex"> when not READTHEDOCS.

PR #7068, opened 2023-03-28
Makefile: use sphinx-build -M

umarcor:
It seems that the structure of the Makefile was not updated since it was created. Sphinx introduced an alternative build solution some years ago, that prevents having dozens of targets in the Makefile. See https://www.sphinx-doc.org/en/master/man/sphinx-quickstart.html#cmdoption-sphinx-quickstart-use-make-mode and https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-M.

Run `make` in this PR (without arguments) to get the list of built-in "Modes":

```
# make
Sphinx v6.1.3
Please use `make target' where target is one of
  html        to make standalone HTML files
  dirhtml     to make HTML files named index.html in directories
  singlehtml  to make a single large HTML file
  pickle      to make pickle files
  json        to make JSON files
  htmlhelp    to make HTML files and an HTML help project
  qthelp      to make HTML files and a qthelp project
  devhelp     to make HTML files and a Devhelp project
  epub        to make an epub
  latex       to make LaTeX files, you can set PAPER=a4 or PAPER=letter
  text        to make text files
  man         to make manual pages
  texinfo     to make Texinfo files
  gettext     to make PO message catalogs
  changes     to make an overview of all changed/added/deprecated items
  xml         to make Docutils-native XML files
  pseudoxml   to make pseudoxml-XML files for display purposes
  linkcheck   to check all external links for integrity
  doctest     to run all doctests embedded in the documentation (if enabled)
  coverage    to run coverage check of the documentation (if enabled)
  clean       to remove everything in the build directory
```

I kept the `gettext` target, because that seems to be customised to generate the content outside of this repo.

As a side note, it seems that the Makefile is not used in CI at all. Is it used somewhere else or is it untested?

PR #7073, opened 2023-03-28
[ci] use Makefile

umarcor:
~This PR is based on #7068 and it will be kept as a draft until that's merged.~

Instead of hardcoding the sphinx-build commands in CI workflows, this PR uses `make dummy` and `make html`. That allows to test that the procedure recommended to users/contributors does work.

### DIFF
--- a/.github/workflows/build_offline_docs.yml
+++ b/.github/workflows/build_offline_docs.yml
@@ -15,8 +15,7 @@ jobs:
           - master
           - stable
     steps:
-      - name: Checkout
-        uses: actions/checkout@v3
+      - uses: actions/checkout@v3
         with:
           ref: ${{ matrix.branch }}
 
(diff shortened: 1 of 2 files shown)


mhilbrunner:
> B. Revert the change in https://github.com/godotengine/godot-docs/pull/7068 but let PAPER ?= so that it can be changed from the command-line.

Seems to be the best tradeoff between not breaking things for people unexpectedly and while improving this (but I also don't know how commonly LaTeX is used by people with Godot's docs, or how common it is for them to want to change the paper size, as we don't have numbers :) My gut feeling says not many, but this is mostly a getting-rid-of-cruft change, so we should try to not break use cases if we can, I guess.)

May have to ponder this a bit though. Any thoughts?

YuriSizov:
For the record, this PR doesn't need to depend on any other, changes in the workflow configs are isolated and can be merged as is.

umarcor:
@YuriSizov this PR now contains a single commit.

YuriSizov:
Ah, right, my bad. Current syntax relies on the environment variable support that you've added. To work with the existing format it needs to be passed as an argument instead.

umarcor:
It's ok 😆

I had done that way at the beginning because I felt intuitively it would work after my review but not before. Yet, I did not think about the specific change that motivated the feeling.

I changed this to pass arguments intead of envvars now.

YuriSizov (approved):
LGTM now!

mhilbrunner (approved):

mhilbrunner:
Merged, thanks!

PR #7074, opened 2023-03-28
[contributing/documentation/building_the_manual] use tabs and admonitions

umarcor:
This PR updates https://docs.godotengine.org/en/stable/contributing/documentation/building_the_manual.html.

- Group tabs are used to show the steps with variants for Windows or other platforms.
- Some clarifications/comments are moved into admonitions (hint, tip or important).
- The example providing a filelist is changed to be based on the make call (see https://github.com/godotengine/godot-docs/commit/e7c7cf1510600a5295f573eaa481e60b0d3b2eb9) instead of explaining it with the explicit call to sphinx-build.

### DIFF
--- a/contributing/documentation/building_the_manual.rst
+++ b/contributing/documentation/building_the_manual.rst
@@ -7,11 +7,12 @@ This page explains how to build a local copy of the Godot manual using the
 Sphinx docs engine. This allows you to have local HTML files and build the
 documentation as a PDF, EPUB, or LaTeX file, for example.
 
-Before you get started, make sure that you have:
-
-- `Git <https://git-scm.com/>`_
-- `make <https://www.gnu.org/software/make/>`_ (unless you’re using Windows)
-- `Python <https://www.python.org/>`_ 3
+.. important::
+    Before you get started, make sure that you have
+    `Git <https://git-scm.com/>`_,
+    `Python 3 <https://www.python.org/>`_,
+    and, unless you are using Windows,
+    `make <https://www.gnu.org/software/make/>`_.
 
 .. note:: Python 3 should come with the ``pip3`` command. You may need to write
     ``python3 -m pip`` (Unix) or  ``py -m pip`` (Windows) instead of ``pip3``.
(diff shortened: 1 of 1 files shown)


YuriSizov:
I'd have to see it live, but generally speaking we shouldn't overdo the notes. A large quantity of them, especially in a row, makes the article harder to read and follow. And from the looks of it, this PR adds a few spans of several notes in a row (2-3). I'd keep some of them as plain text at least.

If there are many disjointed hints and tips in the article, we can put them in a dedicated section as a list or an FAQ-like block.

umarcor:
@YuriSizov I added a couple of subsections, to move two of the groups of 2-3 notes in a row. See https://umarcor.github.io/godot-docs/contributing/documentation/building_the_manual.html

contributing/documentation/building_the_manual.rst
```diff
@@ -61,52 +95,71 @@ Before you get started, make sure that you have:
...
+        .. code:: sh
+
+            make html SPHINXOPTS=-j2
 
-   sphinx-build -b html ./ _build/html
+Specifying a list of files
+--------------------------
 
-You can also specify a list of files to build, which can greatly speed up compilation:
+You can specify a list of files to build, which can geatly speed up compilation:
```
mhilbrunner, line 161:
```suggestion
You can specify a list of files to build, which can greatly speed up compilation:
```

mhilbrunner (approved):
Overall, I think this is an improvement. I like the tabs.

I don't think the PR makes the page heavier and harder to parse, but it _is_ a heavy page (regardless of this PR). Not sure how to improve, but there may be some future work here.

umarcor:
@YuriSizov @mhilbrunner anything pending here?

contributing/documentation/building_the_manual.rst
```diff
@@ -1,17 +1,18 @@
 .. _doc_building_the_manual:
 
 Building the manual with Sphinx
-===============================
+###############################
```
mhilbrunner, line 4:
```suggestion
===============================
```

Not sure on the reason for this change, as other files use `=`. RST primer says it doesn't make a difference.

contributing/documentation/building_the_manual.rst
```diff
@@ -1,17 +1,18 @@
...
-
-- `Git <https://git-scm.com/>`_
-- `make <https://www.gnu.org/software/make/>`_ (unless you’re using Windows)
-- `Python <https://www.python.org/>`_ 3
+.. important::
+    Before you get started, make sure that you have
+    `Git <https://git-scm.com/>`_,
+    `Python 3 <https://www.python.org/>`_,
+    and, unless you are using Windows,
+    `make <https://www.gnu.org/software/make/>`_.
```
mhilbrunner, line 15:
```suggestion
Before you get started, make sure that you have:

- `Git <https://git-scm.com/>`_
- `make <https://www.gnu.org/software/make/>`_ (unless you’re using Windows)
- `Python <https://www.python.org/>`_ 3
```

Reverting this, as the 2 boxes immediately one after the other _is_ a bit happy, having built and checked the page again.

mhilbrunner (approved):

mhilbrunner:
Merged. Nope, just too much backlog :)

Thanks again for contributing!

PR #7083, opened 2023-03-29
Update example to use correct arguments for connect()

Skyvastern:
In Godot 4, the current expression for connecting a method to a signal won't work. As such I've updated it with an updated set of arguments.


mhilbrunner:
Fixed in https://github.com/godotengine/godot-docs/pull/7183. Thanks for contributing!

PR #7089, opened 2023-03-30
Update Making Main Screen Plugins page

seafiish:
Edited [Making Main Screen Plugins](https://docs.godotengine.org/en/latest/tutorials/plugins/editor/making_main_screen_plugins.html) page:
- Added a note in the beginning, that it assumes you know how to make generic plugins.
- Edited the first paragraph in [Main screen scene](https://docs.godotengine.org/en/latest/tutorials/plugins/editor/making_main_screen_plugins.html#main-screen-scene) to make more sense
- Fixed the button script code (in the same section)
- Changed the sentience after the button code.

I don't know C#, so I'm not sure if the button script will work, I apologize. 

This should make the page totally Godot 4 compatible, but I'm not positive?

Also, the repo that holds the [example project](https://github.com/godotengine/godot-demo-projects/blob/master/plugins/addons/main_screen/print_hello.gd) will need to be updated. 

This is my first ever PR, so please forgive any mistakes :)

### DIFF
--- a/tutorials/plugins/editor/making_main_screen_plugins.rst
+++ b/tutorials/plugins/editor/making_main_screen_plugins.rst
@@ -5,6 +5,9 @@
 Making main screen plugins
 ==========================
 
+.. note:: This tutorial assumes you already know how to make generic plugins. If
+          in doubt, refer to the :ref:`doc_making_plugins` page.
+
 What this tutorial covers
 -------------------------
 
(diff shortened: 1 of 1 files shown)


FivelFox:
hi!

uhh, I see C# was added properly to this page. great! but how do I resolve this conflict? sorry, this is my first time dealing with this

YuriSizov:
You would need to do a rebase against the current master branch. I would advise to not use your own master but create feature branches for each PR instead. That makes everything easier to manage.

Though at this point you have deleted your entire forked repository, not sure if intentional or not. You would need to restore it and submit a new PR.

PR #7106, opened 2023-04-01
Update size and anchors tutorial

jnettl:
Updated the size and anchors tutorial to reflect the changes made in Godot 4.


mhilbrunner:
Hey! Thanks for contributing. Please convert the PNG file(s) to WebP while you're at it (we're in the process of moving the docs repository over to WebP, see [image creation guidelines](https://docs.godotengine.org/en/stable/contributing/documentation/docs_image_guidelines.html)).

jnettl:
No worries, glad to be here!

YuriSizov:
This is probably a good fix-up, but just for the record, there is a plan to rewrite these articles (this one, and the one about containers). Here's the current short plan:

<details>
The task is to replace https://docs.godotengine.org/en/latest/tutorials/ui/size_and_anchors.html and parts of https://docs.godotengine.org/en/latest/tutorials/ui/gui_containers.html with a new article about creating layouts.

## Creating UI layouts

- Introduction into scalable UIs for games (kind of covered already, but a fresh pass would be good to tie the rest of the article).

#### Layout modes in Godot

This comment is a great reference IMO:
https://www.reddit.com/r/godot/comments/zkdqm5/comment/izzik20/

- Godot has 2 complementing systems for creating layouts; when is each useful.
- Precise positioning; How anchors, offsets, and grow direction work; Comparison of Position vs Anchors.
- Behavior-driven positioning; How containers, size flags, container specific properties work.

It may be worth it to port the entire containers article here, except for the custom containers part (custom containers deserve a dedicated article that fully explores the subject, not a small paragraph at the end with an obscure code sample).

#### Editor tools for layouts

- How to use the inspector, when certain layout and sizing options are available.
- How to use viewport widgets.
</details>

jnettl:
Ah cheers, it's good to have a reference for a proper rewrite. I was planning on raising it once these were merged anyway. I'll put something together soon but it'll be good to not have outdated info for now at least!

Vennnot:
Even though it will be rewritten, can this be merged?

mhilbrunner:
cc @YuriSizov - would be cool if you could review this so we can merge it as an improvement until this is rewritten.

tutorials/ui/size_and_anchors.rst
```diff
@@ -17,37 +15,53 @@ and mobile phones have different resolutions and aspect ratios.
 There are several ways to handle this, but for now, let's just imagine
 that the screen resolution has changed and the controls need to be
 re-positioned. Some will need to follow the bottom of the screen, others
-the top of the screen, or maybe the right or left margins.
+the top of the screen, or maybe the right or left borders.
+
+.. image:: img/ui_anchor_and_margins.webp
```
YuriSizov, line 20: This image has a bunch of incorrect terminology. First of all, it refers to a parent container. While it's a generic word, a container is a term in Godot that carries special meaning. One of the characteristic of containers is that you don't use anchors with them. It should be "parent control", as it was in the original image.

Terms "anchor pin" and "anchor ratio" aren't great either as they don't exist in this form in Godot. Point and value would be more appropriate.

Furthermore, here and below you use the term "margin", but in Godot 4 this property was renamed to avoid confusion with MarginContainers. It's offsets now, which it is correctly called way below in the article, but not directly below this image, and not in the image itself.

tutorials/ui/size_and_anchors.rst
```diff
@@ -17,37 +15,53 @@ and mobile phones have different resolutions and aspect ratios.
...
+(in case there is no parent control) the viewport. Anchors adjust where the
+margin distances are relative *to*. We can think of them as a percentage of the
+parent's size, represented by a value between 0 and 1.0.
 
-.. image:: img/anchors.png
+:ref:`Containers <doc_gui_containers>` are used to compose the layout of control
+scenes, offering powerful layout and sizing functionality. The MarginContainer
+is where we typically set margins for its child controls. When using containers,
+controls will inherit the layout properties of their parent, removing the need
+to modify the anchors of a control.
```
YuriSizov, line 32: This is just wrong and misleading.

tutorials/ui/size_and_anchors.rst
```diff
@@ -17,37 +15,53 @@ and mobile phones have different resolutions and aspect ratios.
...
-the top of the screen, or maybe the right or left margins.
+the top of the screen, or maybe the right or left borders.
+
+.. image:: img/ui_anchor_and_margins.webp
+
+This is done using the *margin* and *anchor* properties. Margins represent a
+distance in pixels relative to the top-left corner of the parent control or
+(in case there is no parent control) the viewport. Anchors adjust where the
+margin distances are relative *to*. We can think of them as a percentage of the
+parent's size, represented by a value between 0 and 1.0.
```
YuriSizov, line 26: Anchors should probably be explained before offsets. Also offsets are not relative to the top-left corner, that's only true by default. They are relative to anchor points, and anchor points in turn default to the top-left corner. I think this explanation here needs more work.

YuriSizov: Also you should be consistent when formatting numbers. It should be `0.0` throughout the article.

tutorials/ui/size_and_anchors.rst
```diff
@@ -17,37 +15,53 @@ and mobile phones have different resolutions and aspect ratios.
...
-This is done by editing the *margin* properties of controls. Each
-control has four margins: left, right, bottom, and top, which correspond
-to the respective edges of the control. By default, all of
-them represent a distance in pixels relative to the top-left corner of
-the parent control or (in case there is no parent control) the viewport.
+Each control has an individual anchor that can be adjusted from the beginning
+to the end of the parent. So the vertical (top, bottom) anchors adjust from 0
+(top of parent) to 1.0 (bottom of parent) with 0.5 being the center. The
+horizontal (left, right) anchors similarly adjust from left to right of the
+parent.
```
YuriSizov, line 41: Why is this mentioned in the section about presets? It should be explained above, in the section about the basics.

tutorials/ui/size_and_anchors.rst
```diff
@@ -17,37 +15,53 @@ and mobile phones have different resolutions and aspect ratios.
...
 
-.. image:: img/margin.png
+Instead of manually setting the *anchor* properties of a control, you can use
+the toolbar's Layout menu, above the viewport. Besides centering, it gives you
+many options to align and resize control nodes. These will automatically set
+the *anchor* properties.
```
YuriSizov, line 46: They set anchors, offsets, and grow direction (this last one is new to 4.0).

tutorials/ui/size_and_anchors.rst
```diff
@@ -17,37 +15,53 @@ and mobile phones have different resolutions and aspect ratios.
...
 
-Note that when you wish the edge of a control to be above or left of the
-anchor point, you must change the margin value to be negative.
+Custom Anchors
+--------------
 
-For example: when horizontal anchors are changed to 1, the margin values
-become relative to the top-right corner of the parent control or viewport.
+The *anchor* properties can be set manually using the custom anchors preset
+option. Here we can change the exact placement and shape of the control
```
YuriSizov, line 58: It's strange that only at this point in the article we explain that you can set anchors manually, when we already explained that, hey, you don't have to. I don't think this flows well.

We also skip the entire Layout section of the Inspector, which is likely the main tool everyone would use, and not the custom anchors or the toolbar. If we go from the Inspector side of things, we can gradually explain that you can position the node directly, then by using layout presets and anchor presets, and then manually. And then mention the toolbar.

YuriSizov (changes requested):
Sorry for the delay. I looked further into this PR, and I don't think it's good to merge. It doesn't improve the content much, misses the important information from Godot 4, and also has misleading statements. It needs to be improved a lot, so I left some pointers to get started.

skyace65:
Closing this PR due to lack of activity for over two months after Yuri's review.

PR #7112, opened 2023-04-03
Add a page on using GCC/Clang/MSVC sanitizers

Calinou:
This also covers Web-specific Emscripten options to enable its own sanitizers.


vnen:
It seems MSVC is also supported, it was added as an option in godotengine/godot#46941

Calinou:
It looks like I read Windows' `detect.py` wrong – we only have asan support when using Visual Studio, and no ~~satanizers~~ sanitizers are currently supported when using MinGW.

mhilbrunner:
In this goes :) Thank you for writing this! 🎉

PR #7113, opened 2023-04-04
GDExtension example: `_process` delta is a double

akien-mga:
It's `Variant::FLOAT` in the bindings, which is always a double.

Changed other member variables for consistency so there's no implicit conversions between float and double.

Fixes https://github.com/godotengine/godot-cpp/issues/1073.

---

Note: Didn't test the actual code, but `Node::_process` does get generated as:
```
gen/include/godot_cpp/classes/node.hpp
232:    virtual void _process(double delta);
```

Not sure if this was untested or just an oversight when writing the doc. CC @paddy-exe


mhilbrunner (approved):

paddy-exe:
@akien-mga I tried out the example myself: https://github.com/paddy-exe/gdextension_cpp_example
It worked flawlessly so I didn't notice a necessity to use doubles instead of floats

paddy-exe (approved):

mhilbrunner:
Thanks Mr. Maintainer! 🎉

akien-mga:
I'll give it a try to confirm, it's weird that it would work for some but not for others.

PR #7115, opened 2023-04-05
Add NavigationLinks to navigation overviews

smix8:
Adds NavigationLinks to navigation overviews.

### DIFF
--- a/tutorials/navigation/navigation_introduction_3d.rst
+++ b/tutorials/navigation/navigation_introduction_3d.rst
@@ -34,6 +34,8 @@ Godot provides the following objects and classes for 3D navigation:
         - NavRegion RID
             Reference to a specific navigation region that can hold navigation mesh data.
             The region can be enabled / disabled or the use restricted with a navigationlayer bitmask.
+        - NavLink RID
+            Reference to a specific navigation link that connects two navigation mesh positions over arbitrary distances.
         - NavAgent RID
             Reference to a specific avoidance agent with a radius value use solely in avoidance.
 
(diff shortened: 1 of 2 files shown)


tutorials/navigation/navigation_introduction_3d.rst
```diff
@@ -34,16 +34,28 @@ Godot provides the following objects and classes for 3D navigation:
...
+    - Regions can join their navigation meshes by proximity for a combined navigation mesh.
+
+- :ref:`NavigationLink3D<class_NavigationLink3D>` Node
+    A Node that connects two positions on navigation mesh over arbitrary distances for pathfinding.
+
+    - The link can be enabled / disabled.
+    - The link can be made one-way or bidirectional.
+    - The use in pathfinding can be further restricted through the navigationlayers bitmask.
+
+    Links tell the pathfinding that a connection exists and at what cost. The actual agent handling and movement needs to happen in custom scripts.
```
skyace65, line 58: I'm a bit confused why the last sentence is here on the 3D and 2D page. Is it implying that there's extra stuff that needs to be done in a pathfinding script for navigation links to work? Or does it mean moving agents around in general? Cause if it's the latter then it should be under the NavigationAgent node description.

smix8: You need to use a script to properly handle movement through a link because it is just a virtual connection between two points. There is nothing specific to NavigationAgent, it is the same for whatever custom setup users might use.

skyace65: Got it, thanks for clarifying!

skyace65 (approved):

Issue #7121, opened 2023-04-05
"Your first 3D game" Bug: Mobs should not rotate on Y axis following player jumping on spawn


Levelleor:
**Your Godot version:**
4.0.1.stable

**Issue description:**
Mobs also follow player on the vertical axis in the 3D game tutorial. Mobs work in, basically, 2D plane and should only get the direction of the player in X/Y planes. If player is above or below the mobs shouldn't matter since their rotation is only defined initially on instantiation. If player, say, falls below the ground we would experience this kind of bug:

![image](https://user-images.githubusercontent.com/11915185/230107764-77f4e5ed-22ab-431d-92c3-18d5f2ef75b7.png)

Normally, this wouldn't be an issue since player should be unable to leave boundaries of the level, but, in this game Player is able to jump altering the Y position, thus mobs using it to position themselves on spawn. You can see the issue present on the game's GIF advertised on the Godot's tutorial where one of the last mobs coming from top right is a bit too tilted:

![image2](https://docs.godotengine.org/en/stable/_images/squash-the-creeps-final.gif)

Specifically you can see one of the latest mobs is a bit too tilted due to player jumping and it being incorrectly set:
![image](https://user-images.githubusercontent.com/11915185/230108419-0a431401-fed9-4bda-8e4d-e2f9ed083c98.png)

Looks like the issue is in this line in the Mob's script:
`look_at_from_position(start_position, player_position, Vector3.UP)`
It is supposed to be something like the following to not reference the player's Y position):
`look_at_from_position(start_position, Vector3(player_position.x, 0, player_position.z), Vector3.UP)`

I am new to godot which is why I haven't opened a PR, although I could if this solution seems viable enough for the documentation, please let me know. 

**URL to the documentation page:**
https://docs.godotengine.org/en/stable/getting_started/first_3d_game/04.mob_scene.html

Levelleor:
It's actually supposed to be the X axis to rotate it up and down but it is skewed here for some reason and I have no idea why. Y works as expected. 

I assume 
 look_at_from_position(start_position, Vector3(player_position.x, 0, player_position.z), Vector3.UP)

makes more sense although it doesn't do what's expected of it

PR #7123, opened 2023-04-05
Fix Mob's initial rotation remain untilted

Levelleor:
Fixes the following issue: https://github.com/godotengine/godot-docs/issues/7121


Levelleor:
Is this still considered for merge? Not sure if this has already been fixed though.

PR #7130, opened 2023-04-07
Fix incorrect `:abbr:` usage for abbreviation display

Calinou:
Previously, some abbreviations would display their full form in the text and the abbreviation in the tooltip. It should be the other way around.


mhilbrunner (approved):

mhilbrunner:
Thanks!

PR #7134, opened 2023-04-08
clarify importance of naming files

abaines:
godot does not require file and class names to match exactly, but this is different than many other languages

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_styleguide.rst
+++ b/tutorials/scripting/gdscript/gdscript_styleguide.rst
@@ -629,6 +629,13 @@ This is consistent with how C++ files are named in Godot's source code. This
 also avoids case sensitivity issues that can crop up when exporting a project
 from Windows to other platforms.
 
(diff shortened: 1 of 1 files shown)


abaines:
Source: [A conversation I had with Balthamel [3.5, 2D] on on Discord](https://discord.com/channels/212250894228652034/477544613511692358/1094288501106823278). 

Thanks Balthamel! 🧡

abaines:
If there needs to be any edits, please consider keeping the error messages. My hope is having those error messages listed will help search engines to come to this page when other devs are having these types of issues. 🤞

Thank you 🚀

YuriSizov:
Thanks for opening a PR. But I don't quite understand what's the advice here. Also, these errors are very general and can happen for a lot of different reasons, we should not list them here.

mhilbrunner:
Agree with Yuri's comment above, I am not sure what issues you were running into, but yeah there are no limitations on these and things should work in any case, at least for GDScript.

If we missed something, please let us know and provide further context :) Thanks for contributing!

PR #7139, opened 2023-04-09
Fix small typos in the particle shader doc

nake90:
A word was missing a letter and a description had a wrongly capitalized word. Now it matches the rest of the document.

I thought that maybe that capitalization was made on purpose (like talking about "Particle" as an object), but in other parts of the document it does not do it like that. For example in `Process built-ins`, so I think it is a typo after all and fixed it.


clayjohn (approved):

mhilbrunner (approved):

mhilbrunner:
Thanks and congrats on your first merged docs contribution!

PR #7141, opened 2023-04-09
add missing line to json parsing example

Simon-Laux:
```gdscript
var json = JSON.new()
```
was missing, It was quite confusing for me until I found https://godotengine.org/qa/118343/godot-4-0-how-to-get-a-dictionary-from-a-json-type

### DIFF
--- a/classes/class_json.rst
+++ b/classes/class_json.rst
@@ -34,6 +34,7 @@ The **JSON** enables all data types to be converted to and from a JSON string. T
     # Save data
     # ...
     # Retrieve data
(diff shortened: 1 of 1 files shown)


mhilbrunner (dismissed):

mhilbrunner:
Thanks for contributing! The change looks good, but the class-reference RST files are autogenerated from XML files in the main repository, so the change needs to be made there instead: https://github.com/godotengine/godot/blob/master/doc/classes/JSON.xml

PR #7161, opened 2023-04-14
Add a CI action to sync up the class reference

YuriSizov:
Let's reduce manual labor :)

This adds a workflow that can be triggered on demand, and also runs on a schedule once a week. It checks out the target version of the main repo, generates new RSTs, and submits a PR for us to approve and merge. I opted to go through a PR to stay on a safer side, though we can also commit directly if we want to.

Here's an example of a PR it creates: https://github.com/YuriSizov/godot-docs/pull/38 _(contains extra commits that temporary appeared in the master, look for the last one)_
And here's an example of a workflow run: https://github.com/YuriSizov/godot-docs/actions/runs/4700847275/jobs/8336065111

-----

I only tested this with the master branch. When you want to run a workflow, you can select any branch you want, but it will not work correctly. We need to map versions of docs to versions of the engine for it to work. Initially I wanted to commit a modified version of this workflow to each branch, but I don't think it will work. I think workflows that are triggered manually need to exist on the master branch and are triggered based on the script from that branch. I will test further what happens with other branches once we merge this.

Another point to consider is automatic runs. I set this up to run once a week, on Saturday night European time. But this is only relevant for the in-dev versions, such as `master` and `3.6`, as of now. For stable branches I want to only keep the manual trigger. So I also need to see how to achieve that with other branches.

### DIFF
--- a/.github/workflows/sync_class_ref.yml
+++ b/.github/workflows/sync_class_ref.yml
@@ -1,0 +1,68 @@
+name: Sync Class Reference
+
+on:
+  workflow_dispatch:
+  # Scheduled updates are only enabled for in-dev branches.
+  schedule:
+    # Run it at night (European time) every Saturday.
+    # The offset is there to try and avoid the high load times.
+    - cron: '15 3 * * 6'
+
+# Make sure jobs cannot overlap.
+concurrency:
+  group: classref-sync-ci
+  cancel-in-progress: true
+
+jobs:
+  build:
+    name: Update class reference files based on the engine revision
+    runs-on: ubuntu-latest
+    env:
+      engine_rev: 'master'
+
+    steps:
+      - name: Checkout the documentation repository
+        uses: actions/checkout@v3
+
+      - name: Checkout the engine repository
+        uses: actions/checkout@v3
+        with:
+          repository: 'godotengine/godot'
+          # Use the appropriate branch for the documentation version.
+          ref: ${{ env.engine_rev }}
+          path: './.engine-src'
+
+      - name: Store the engine revision
+        id: 'engine'
+        run: |
+          cd ./.engine-src
+          hash=$(git rev-parse HEAD)
+          hash_short=$(git rev-parse --short HEAD)
+          echo "Checked out godotengine/godot at $hash"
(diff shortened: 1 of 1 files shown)


mhilbrunner:
Guess we could consider whether to bother Rémi for a review.
Besides that, I need to check what happens if we push this to the 4.0 branch (as we currently do regularly push master to 4.0).

YuriSizov:
> Besides that, I need to check what happens if we push this to the 4.0 branch (as we currently do regularly push master to 4.0).

Nothing, really. It currently cannot work with non-master branches. I'd look into it more, but it's tricky to test because it needs to be merged into master first to become available on GitHub. So I'd prefer we merge it as is to automate the master, and then I'll see how to make it work with other branches.

As I mentioned, you can actually trigger it for any branch from the master branch version, so we may not even need it in other branches.

mhilbrunner:
@YuriSizov After consulting the [docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows), it may not be a problem at all:
> Scheduled workflows run on the latest commit on the default or base branch.

So if I'm understanding that correctly, we won't need to do anything and this will still be safe to push to other branches and simply won't have any effect there for the scheduled run.

I guess lets merge and test :)

mhilbrunner:
YOLO. Thanks 🎉

YuriSizov:
Ah. If that's true, then it's a bummer, we'd need to implement some sort of a workaround for 3.6 then (or sync 3.6 manually). But that also means there is no reason to push it to other branches at all, because it won't do anything in them.

YuriSizov:
Pushed a version of this workflow to 3.6: https://github.com/godotengine/godot-docs/commit/dfecc71e6ddc975ac8588218bf5e4071220e0ca1

Can now be manually triggered (syncs with the `3.x` branch of the main repo).

YuriSizov:
Pushed a version of this workflow to 3.5: https://github.com/godotengine/godot-docs/commit/f846dd9eb4d8f65699d5ab3e1a8d306d27df4f8d

Can now be manually triggered (syncs with the 3.5 branch of the main repo).

PR #7164, opened 2023-04-15
Update link to set of predefined color mappings in `creating_icons.rst`

andrejp88:
The existing link appears to have pointed to a May 2020 version of `editor_themes.cpp`. For the Godot 4.0 docs, I think it makes sense to point to the tag of the most recent release (4.0.2).


mhilbrunner (approved):

mhilbrunner:
Makes sense. Thanks, and congrats on your first merged godot-docs contribution!

PR #7168, opened 2023-04-16
Mention Windows graphics driver TDR issues on Using compute shaders

Calinou:
This is a frequently encountered issue when working with complex compute shaders. ~~The advice to increase TDR isn't ideal as it requires system intervention, but it's sometimes the only way to deal with the issue.~~

**Edit:** Removed part about increasing TDR, since it requires system intervention and isn't suited for games as a result.

### DIFF
--- a/tutorials/shaders/compute_shaders.rst
+++ b/tutorials/shaders/compute_shaders.rst
@@ -252,6 +252,7 @@ assign it to a uniform set which we can pass to our shader later.
 
 Defining a compute pipeline
 ---------------------------
+
 The next step is to create a set of instructions our GPU can execute.
 We need a pipeline and a compute list for that.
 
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):
Meh, not really a good solution for a game you want to distribute. We may want to point that out more clearly, that you need to plan how to deal with this if you want to distribute. And we also may want to put a similar disclaimer to the Adobe one that editing the registry is *Here Be Dragons* territory.

clayjohn:
Yes, I think we need to tell users that if they are running into TDRs with compute shaders they need to split the computer shader into multiple dispatches. Editing the registry file isnt a solution they can pass on to players

Calinou:
> Yes, I think we need to tell users that if they are running into TDRs with compute shaders they need to split the computer shader into multiple dispatches. Editing the registry file isnt a solution they can pass on to players

I don't know how this could be done, as I have little experience with compute shaders myself.

mhilbrunner:
cc @clayjohn see above :)

tutorials/shaders/compute_shaders.rst
```diff
@@ -321,6 +322,19 @@ example, we synchronize right away because we want our data available for readin
...
+    Long computations can cause Windows graphics drivers to "crash" due to
+    :abbr:`TDR (Timeout Detection and Recovery)` being triggered by Windows. This
+    is a mechanism that reinitializes the graphics driver after a certain amount
+    of time has passed without any activity from the graphics driver.
+
+    Depending on the duration your compute shader takes to execute, you may have to increase
+    the TDR timeout duration for it to execute successfully (especially on slower GPUs).
+    To increase TDR timeout duration, follow the instructions on
+    `this page <https://substance3d.adobe.com/documentation/spdoc/gpu-drivers-crash-with-long-computations-128745489.html>`__.
+    Note that increasing TDR timeout duration requires administrator privileges.
```
clayjohn, line 336:
```suggestion
    Depending on the duration your compute shader takes to execute, you may need to split it into multiple dispatches to reduce the time each dispatch takes and reduce the chances of triggering a TDR.
```

mhilbrunner:
Thanks you two, merged. :)

PR #7171, opened 2023-04-16
step_by_step/instancing.rst: Add macOS shortcuts

royzwambag:
There were 2 instances where the default windows shortcut was used, but not the macOS shortcut, so I added them. I also let Godot convert the example `instancing.zip` project to be Godot 4 compatible (it prompted to do this when opening the project in Godot 4), then rezipped it.

### DIFF
--- a/getting_started/step_by_step/instancing.rst
+++ b/getting_started/step_by_step/instancing.rst
@@ -45,7 +45,7 @@ you to download the ball's sample project we prepared for you:
 :download:`instancing.zip <files/instancing.zip>`.
 
 Extract the archive on your computer. To import it, you need the Project Manager.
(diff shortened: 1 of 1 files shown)


YuriSizov:
Thanks for opening a PR. We have other contributions updating the `instancing.zip` file, but the current plan is to get rid of it, because it's hard to check for changes and maintain. Instead it should be uploaded as a project on GitHub and tracked normally.

So could you please remove that change from your PR?

mhilbrunner (approved):

mhilbrunner:
Thanks for updating the PR, and congrats on your first merged godot-docs contribution! 🎉

PR #7173, opened 2023-04-17
Fix grammer in logic_preferences.rst

BWPanda:
The [documentation](https://docs.godotengine.org/en/stable/tutorials/best_practices/logic_preferences.html) has this sentence:

> Some **properties setters** have code to update other corresponding values...

Which doesn't sound right to me.

This PR changes it to:

> Some **property setters** have code to update other corresponding values...

But it could also possibly be:

> Some **property's setters** have code to update other corresponding values...

### DIFF
--- a/tutorials/best_practices/logic_preferences.rst
+++ b/tutorials/best_practices/logic_preferences.rst
@@ -16,7 +16,7 @@ properties such as the node's name or position. A common dilemma is, when
 should you change those values?
 
 It is the best practice to change values on a node before adding it to the
-scene tree. Some property setters have code to update other
+scene tree. Some property's setters have code to update other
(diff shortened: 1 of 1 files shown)


clayjohn:
I think ``property's setters`` is the correct permutation here as I don't think ``property setter`` is a common phrase to talk about a setter (it seems to imply that we are talking about a setter that operates on a property as opposed to a setter that operates on something else?).

BWPanda:
Thanks @clayjohn, PR updated.

clayjohn (approved):

mhilbrunner (approved):

mhilbrunner:
Thanks! Merged.

PR #7186, opened 2023-04-18
[contributing/documentation/building_the_manual] fix top-level heading

umarcor:
Currently, the pages has multiple top-level headings. As a result, "subsections" are shown as siblings, instead of children. See https://docs.godotengine.org/en/latest/contributing/documentation/building_the_manual.html.

It should have a single top-level heading, and others should be at least h2. This PR fixes it.

Ref: #7074


YuriSizov:
Yes, that works. We don't use `^` consistently, half the time it's `~`. Either is allowed, but we should probably do a pass an unify those, and leave a note in the contributors docs.

YuriSizov:
Thanks!

Issue #6497, opened 2022-12-18
Add separate system requirements page


lostminds:
**Your Godot version:**
4.0beta8

**Issue description:**
The minimum requirements needed to run exported games on different platforms is currently placed under the "Platforms" heading in the "List of features" page (not yet updated for 4.0). I think this makes it hard to find for users looking for this information as you can't find it searching for "requirements". And even though I'd consider supporting a platform a feature I wouldn't call a minimum system requirement a feature, so to me it feels like an unexpected place for the information. 

So, as discussed in https://github.com/godotengine/godot-proposals/discussions/5946#discussioncomment-4438455 I'd suggest moving this minimum system requirements list for the different platforms to a new separate page with the name "System requirements" under "About". And then link to this from other pages in the docs (like the list of features or export instructions) where users might be interested in this information. All to help make this information easier to find and help developers in the process of publishing games for multiple platforms.

**URL to the documentation page (if already existing):**
https://github.com/godotengine/godot-docs/blob/master/about/list_of_features.rst#platforms

Calinou:
Relevant discussion: https://github.com/godotengine/godot-proposals/discussions/5946

PR #7198, opened 2023-04-20
Add a page on Godot system requirements

Calinou:
Let me know if the format is good and I'll complete the TODO items :slightly_smiling_face:

I'm not sure if we should feature two separate sets of minimum/recommended specs for exported projects (one for simple projects, one for more complex projects), as a project's complexity is a subjective topic.

I also don't know if we should feature a 3rd tier such as "optimal" (e.g. for a smooth 4K 60 FPS experience). This something a lot of AAA games have been doing recently ([example](https://user-images.githubusercontent.com/180032/233505125-c7eb3053-ca9b-4d3e-9d2f-97373c527eeb.png)), but I feel it would introduce too many combinations to check and maintain.

This also adds a note about Windows 7/8/8.1 support being done on a best-effort basis. This closes https://github.com/godotengine/godot/issues/77304.

Once we agree on a final design, I can work on a 3.x version of this page.

- This closes https://github.com/godotengine/godot-docs/issues/6497.

## TODO

- [x] Add editor system requirements.
- [x] Add recommended requirements for exported projects.

### DIFF
--- a/about/system_requirements.rst
+++ b/about/system_requirements.rst
@@ -1,0 +1,369 @@
+.. _doc_system_requirements:
+
+System requirements
+===================
+
+This page contains system requirements for the editor and exported projects.
+These specifications are given for informative purposes only, but they can be
+referred to if you're looking to build or upgrade a system to use Godot on.
+
+Godot editor
+------------
+
+These are the **minimum** specifications required to run the Godot editor and work
+on a simple 2D or 3D project:
+
+Desktop or laptop PC - Minimum
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. When adjusting specifications, make sure to only mention hardware that can run the required OS version.
+.. For example, the x86 CPU requirement for macOS is set after the MacBook Air 11" (late 2010 model),
+.. which can run up to macOS 10.13.
+
++----------------------+-----------------------------------------------------------------------------------------+
+| **CPU**              | - **Windows:** x86_32 CPU with SSE2 instructions, or any x86_64 CPU                     |
+|                      |                                                                                         |
+|                      |   - *Example: Intel Core 2 Duo E8200, AMD Athlon XE BE-2300*                            |
+|                      |                                                                                         |
+|                      | - **macOS:** x86_64 or ARM CPU (Apple Silicon)                                          |
+|                      |                                                                                         |
+|                      |   - *Example: Intel Core 2 Duo SU9400, Apple M1*                                        |
+|                      |                                                                                         |
+|                      | - **Linux:** x86_32 CPU with SSE2 instructions, x86_64 CPU, ARMv7 or ARMv8 CPU          |
+|                      |                                                                                         |
+|                      |   - *Example: Intel Core 2 Duo E8200, AMD Athlon XE BE-2300, Raspberry Pi 4*            |
++----------------------+-----------------------------------------------------------------------------------------+
+| **GPU**              | - **Forward+ rendering method:** Integrated graphics with full Vulkan 1.0 support       |
+|                      |                                                                                         |
+|                      |   - *Example: Intel HD Graphics 5500 (Broadwell), AMD Radeon R5 Graphics (Kaveri)*      |
+|                      |                                                                                         |
+|                      | - **Mobile rendering method:** Integrated graphics with full Vulkan 1.0 support         |
+|                      |                                                                                         |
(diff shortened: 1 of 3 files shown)


about/system_requirements.rst
```diff
@@ -0,0 +1,116 @@
...
+|                      |   - *Example: Intel GD Graphics 5500 (Broadwell), AMD Radeon R5 Graphics (Kaveri)*      |
+|                      |                                                                                         |
+|                      | - **Compatibility rendering method:** Integrated graphics with full OpenGL 3.3 support  |
+|                      |                                                                                         |
+|                      |   - *Example: Intel GD Graphics 2500 (Ivy Bridge), AMD Radeon R5 Graphics (Kaveri)*     |
++----------------------+-----------------------------------------------------------------------------------------+
+| **RAM**              | - **For native exports:** 2 GB                                                          |
+|                      | - **For web exports:** 4 GB                                                             |
++----------------------+-----------------------------------------------------------------------------------------+
+| **Storage**          | 100 MB (used for the executable, project files and cache)                               |
```
bruvzg, line 73: macOS release export executable is 130 MB (since it includes 2 architectures).

about/system_requirements.rst
```diff
@@ -0,0 +1,116 @@
...
++----------------------+-----------------------------------------------------------------------------------------+
+| **Operating system** | - **For native exports:** Windows 7, or Linux distribution released after 2016          |
```
bruvzg, line 75: macOS 10.12+ (for Compatibility renderer, theoretically I do not think anyone tested it), 10.15+ (for Vulkan, some limited functionality on 10.14 is possible).

about/system_requirements.rst
```diff
@@ -0,0 +1,116 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
++----------------------+-----------------------------------------------------------------------------------------+
+| **CPU**              | - **Android:** SoC with any 32-bit or 64-bit ARM or x86 CPU                             |
+|                      |                                                                                         |
+|                      |   - *Example: Qualcomm Snapdragon 430, Samsung Exynos 5 Octa 5430*                      |
+|                      |                                                                                         |
+|                      | - **iOS:** SoC with any 64-bit ARM CPU                                                  |
+|                      |                                                                                         |
+|                      |   - *Example: Apple A7 (iPhone 5S)*                                                     |
```
bruvzg, line 89: For Vulkan renderer it's likely A9+ (iPhone 6S / SE 1.gen.), A7 have Metal support, but not all features used by Godot are supported.

and-rad:
I think "optimal" is unneccessary. When going with minimum and recommended, it's probably a good idea to come to an agreement on what kinds of features these include. It might also be worth it to have minimum and recommended specs separated by 2D/3D because a simple 3D game can easily become way more demanding than a complex 2D game.

By "agreement" I'm thinking something like, what's the required spec to get a game running at 30 (60?) FPS with MSAA, SDFGI, SSAO enabled, a particle system with X amount of particles in the scene and between Y and Z vertices coming from N meshes, drawn with K number of draw calls? That would then count as the baseline for the "recommended" settings.

How to reach this kind of agreement, I have no idea 🤷‍♂️

lostminds:
There's a nice pr https://github.com/godotengine/godot/pull/77740 that will hopefully reduce iOS device requirements, so once that is merged I don't think an A9 processor will be required anymore for vulkan renderers on iOS.

Regarding minimum/recommended my perspective would be to give the minimum requirements the primary focus. As in what will be required to run the game at all (not get a black screen or crash on startup). And for that purpose also perhaps add some practical hints on how this can be configured on the different platforms to avoid distribution to users/devices that don't support your game. 

For example even though the actual requirement is typically specific graphics hardware, this can not be specified as a required capability in the mobile app stores (Apple App Store and Google Play), so it's not practically that useful. Instead you need to specify a higher OS version requirement that indirectly means the device will have the required hardware capabilities to support that OS. And getting a hint about this would I think be useful to users looking to distribute mobile  games.

Calinou:
I've added specifications for the editor and updated the page to mention ARM Linux support (with suggested ARM boards to run Godot on). This should be ready to merge now.

mhilbrunner (approved):

mhilbrunner:
Thank you! 🙏 Merged.

mhilbrunner:
Cherry-picked to 4.2.

aaronfranke:
> Once we agree on a final design, I can work on a 3.x version of this page.

That would be very nice to have documented.

PR #7200, opened 2023-04-21
Fix Keyboard modifiers example

MisoMosiSpy:
1. Fixed keyboard modifier example to use correct modifier identifier
2. Updated event output as per 4.0 implementation
3. Updated screenshot for InputMap as per 4.0 
4. Added note about "Show Built-in Actions" button added to InputMap settings.


mhilbrunner (approved):

mhilbrunner:
Thank you! Sorry this got no response for a bit due to the backlog, but I found time to review and test this finally :)

PR #7202, opened 2023-04-21
Clarified hiDPI setting in docs

AlexKalopsia:
Small fix. The "Enable hiDPI" setting can't be found, unless Advanced Settings is on (which is not the default case)


skyace65 (changes requested):
That note should actually be removed completely. It was there because HiDPI was not enabled by default in Godot 3. All new projects made in Godot 4 have it enabled by default.

mhilbrunner:
Closing as superseded by https://github.com/godotengine/godot-docs/pull/7293. Thanks for contributing!

PR #7203, opened 2023-04-21
Fix link to the 3.x version of Compiling for Linux

kashiw1n:
Since there is no "3.x " branch of the docs, I'm not sure which branch the link should point to instead. I went with the latest 3.x, but maybe 3.0 is better?


skyace65:
I would change it to say 3.6 since that is planned to be the last 3.x release (and of course the bugfix releases that go with it).

Piralein (approved):

mhilbrunner (approved):

mhilbrunner:
I think its good as is. Thanks and congrats on your first merged contribution to Godot's docs!

PR #7206, opened 2023-04-22

regardt-nel:
I am suggesting a small addition to the sentence as it was unclear to me at first whether the external program needs a plugin or the plugin needs to available to Godot.  I take it the latter is true?


mhilbrunner (approved):

mhilbrunner:
Yeah indeed, this is talking about Godot plugins. Thanks for contributing, always useful to have fresh eyes on stuff like this :)

PR #7207, opened 2023-04-22

regardt-nel:
It was not clear to me whether or not other 3d engines also uses scenes.  The update aims to make a more clear distinction between prefab behaviour in other engines vs the behaviour of scenes in Godot.


regardt-nel:
Very good point, I agree!

PR #7210, opened 2023-04-22
Update code to 4.0+ for screenshotting the viewport tutorial

### LINKED ISSUE #7162: Outdated Viewport Capture Example Code
**Your Godot version:**
4.0.2.stable

**Issue description:**
The following code snippet included in the documentation causes an Invalid Call error in Godot v4.0.2.stable:
`var img = get_viewport().get_texture().get_data()`

https://docs.godotengine.org/en/stable/tutorials/rendering/viewports.html

jynus:
The viewport tutorial:
https://docs.godotengine.org/en/latest/tutorials/rendering/viewports.html

did not work for Godot 4- as instead of get_data() from Texture one should use get_image() from Texture2D.

Also setting a texture didn't work as documented, fixed too. Tested by using the code on a simple scene with a sprite.

This should be backported to 4.0 branch, but not previous ones.

Fixes #7162

### DIFF
--- a/tutorials/rendering/viewports.rst
+++ b/tutorials/rendering/viewports.rst
@@ -138,14 +138,13 @@ following code:
 
 ::
 
-   # Retrieve the captured Image using get_data().
-   var img = get_viewport().get_texture().get_data()
+   # Retrieve the captured Image using get_image().
+   var img = get_viewport().get_texture().get_image()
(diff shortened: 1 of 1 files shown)


skyace65 (approved):

skyace65:
Just tested the new code and it works, thanks!

Issue #6201, opened 2022-09-19
Rewrite Importing 3D scenes page


FreshlyBrewedCode:
**Your Godot version:**  4.0 beta1

**Issue description:** The current documentation about importing 3D assets from external tools such as Blender is outdated. It does not cover the new workflow present in the latest 4.0 beta release. It also seems to be missing some notable 3.x features.

**URL to the documentation page (if already existing):** [Importing 3D scenes](https://docs.godotengine.org/en/latest/tutorials/assets_pipeline/importing_scenes.html#import-hints)

#### Context 
I've recently started experimenting with Godot and was struggling to find a good workflow for importing 3D assets. This was partially because the latest documentation is not up-to-date with the latest beta build of Godot 4.0. However, I found it generally difficult to figure out a reliable 3D workflow from Blender to Godot using just the official docs.

### Missing/Outdated Documentation
 Related to #5121. 
- Document new 4.0 import workflows
    - Presets do not seem to exist anymore instead the advanced import dialog has to be used to extract meshes and animations or use external materials
    ![image](https://user-images.githubusercontent.com/11772241/191074304-e40df11c-e90a-4bfd-a9b5-a309c470b131.png)
- Document the .blend import feature (https://github.com/godotengine/godot/pull/54886)
    - Mention related editor settings:
       ![image](https://user-images.githubusercontent.com/11772241/191072983-8cbdfc9b-f0d4-4769-8188-0f4ab1e394b8.png)
    - Also relevant for 3.x?
- Possibly mention the limitation/bug of imported scene instances not getting updated and possible workarounds (if godotengine/godot/pull/57606 does not get merged in the meantime).
- Add more information about how animations are imported
    - `AnimationPlayer` is created for an imported scene
    - Recommended workflow
    - Animation retargeting (Godot 4.0) (https://github.com/godotengine/godot-proposals/issues/4510) 

### Improvements
Here are some general ideas for improving the clarity of the page:
- Instead of having several *Exporting \<format> from Blender* sections, create sections for different formats/tools (e.g. "Importing from Blender", "Importing from Maya", "Importing FBX").
    - Cover the available options for getting 3D assets from a given tool into Godot and mention the recommended approach (if applicable)
        - Blender
        - Maya
        - 3DS Max
        - Substance
    - Cover specific limitations/features/considerations for a given file format
        - glTF
        - FBX
        - OBJ
        - DAE 

[...]

PR #7221, opened 2023-04-25
Configuring node import options fix

Vennnot:
Changed NavMesh to Occluder as meant by the documentation and the image above it.

Here is the original page: https://docs.godotengine.org/en/latest/tutorials/assets_pipeline/importing_scenes.html

Here is the issue where this was mentioned: https://github.com/godotengine/godot-docs/issues/6201#issuecomment-1371629067

Additionally, I would break the parent issue above into check boxes to make them more easily trackable. That, and separate the proposals for improvements from actual issues where things are not documented. So the actual issue can be worked on more easily and closed.

 Irrelevant to this PR but, I've noticed a lot of Github issues where the issue is no longer valid since either the documentation has changed so much or a fix was already merged. Can I send you a list @YuriSizov ? We could stream line things and make it easier for first time contributors to actually find something valid to work on.

### DIFF
--- a/tutorials/assets_pipeline/importing_scenes.rst
+++ b/tutorials/assets_pipeline/importing_scenes.rst
@@ -380,7 +380,7 @@ This exposes several per-node import options:
   will keep the original mesh visible, while **NavMesh Only** will only import
   the navigation mesh (without a visual representation). **NavMesh Only** is
   meant to be used when you've manually authored a simplified mesh for navigation.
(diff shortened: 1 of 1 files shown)


skyace65 (approved):
Thanks!

YuriSizov:
> Can I send you a list @YuriSizov ? We could stream line things and make it easier for first time contributors to actually find something valid to work on.

Just comment on the relevant issue that you think is outdated, and one of the maintainers will check.

PR #7223, opened 2023-04-25
Fix C# issues in inspector-plugin tutorial

wlsnmrk:
C#'s `EditorInspectorPlugin::_CanHandle()` method takes an argument of type `GodotObject`, but the tutorial page for inspector plugins gave an override example taking type `Variant`.

Additionally, the same page had one C# line that was causing horizontal scroll in the code panel, and I've included a fix for that.

### DIFF
--- a/tutorials/plugins/editor/inspector_plugins.rst
+++ b/tutorials/plugins/editor/inspector_plugins.rst
@@ -143,7 +143,7 @@ specifically add :ref:`class_EditorProperty`-based controls.
 
     public partial class MyInspectorPlugin : EditorInspectorPlugin
     {
-        public override bool _CanHandle(Variant @object)
+        public override bool _CanHandle(GodotObject @object)
         {
             // We support all objects in this example.
             return true;
--- a/tutorials/plugins/editor/inspector_plugins.rst
+++ b/tutorials/plugins/editor/inspector_plugins.rst
@@ -285,7 +285,7 @@ followed by ``set_bottom_editor()`` to position it below the name.
             EmitChanged(GetEditedProperty(), _currentValue);
         }
 
-        public override void UpdateProperty()
+        public override void _UpdateProperty()
         {
             // Read the current value from the property.
             var newValue = (int)GetEditedObject().Get(GetEditedProperty());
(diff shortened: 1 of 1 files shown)


wlsnmrk:
Converting to draft as I've found some additional issues with the code on this page:
* `_ParseProperty()` signature is also incorrect
* There's an undeclared variable used in `_ParseProperty()`
* `UpdateProperty()` is not exposed as an override-able method for `EditorProperty`

I'm happy to fix the first two, but I'm searching for a solution to the third.

wlsnmrk:
Should be good to go now, with example code on that page compiling/working.

tutorials/plugins/editor/inspector_plugins.rst
```diff
@@ -143,20 +143,22 @@ specifically add :ref:`class_EditorProperty`-based controls.
...
+        public override bool _CanHandle(GodotObject @object)
         {
             // We support all objects in this example.
             return true;
         }
 
-        public override bool _ParseProperty(GodotObject @object, int type, string name, int hintType, string hintString, int usageFlags, bool wide)
+        public override bool _ParseProperty(GodotObject @object, Variant.Type type, 
+            string name, PropertyHint hintType, string hintString, 
+            PropertyUsageFlags usageFlags, bool wide)
```
raulsntos, line 154: I don't think we usually split long method signature lines in the documentation, but I think it's fine to start doing it. I'm not sure what should be the maximum character length though.

raulsntos (approved):
The changes are correct, at some point the API changed before the 4.0 release and I forgot to update this documentation.

tutorials/plugins/editor/inspector_plugins.rst
wlsnmrk: No worries either way - I was asked to break lines that were causing horizontal scroll (in both C# and GDScript) on an earlier PR, so I assumed that was the preferred approach.

The [C# style guide](https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_style_guide.html#general-guidelines) suggests 100 characters, but I had to break this at around 90 to avoid scroll.

mhilbrunner (approved):

mhilbrunner:
Thanks! And thanks for the review. Merged!

mhilbrunner:
Pushed to `4.0` and `stable`.

PR #7229, opened 2023-04-26
Updated Object's set_meta documentation to indicate metadata's name must be a valid identifier

damdalf:
Fixes: https://github.com/godotengine/godot/issues/70141

It looks like the original cause of the issue has been fixed by https://github.com/godotengine/godot/pull/70386, but no one had added this information to the method's documentation and the issue is still open. Thus, I figured this change would be an excellent first issue to tackle. This change simply adds another **Note:** under the documentation of Object's set_meta(String name, Variant value) method that indicates the "name" argument must be a valid identifier (i.e., the String cannot begin with a number).

### DIFF
--- a/classes/class_object.rst
+++ b/classes/class_object.rst
@@ -1190,6 +1190,8 @@ Returns ``true`` if a metadata entry is found with the given ``name``. See also 
 
 \ **Note:** Metadata that has a ``name`` starting with an underscore (``_``) is considered editor-only. Editor-only metadata is not displayed in the Inspector and should not be edited, although it can still be found by this method.
 
+\ **Note:** A metadata’s ``name`` must be a valid identifier. A valid identifier cannot start with a number.
+
 .. rst-class:: classref-item-separator
 
 ----


YuriSizov:
https://github.com/godotengine/godot/tree/master/doc/classes/Object.xml

You can read more about contributing to the class reference [in our documentation](https://docs.godotengine.org/en/latest/contributing/documentation/updating_the_class_reference.html).

damdalf:
@YuriSizov Apologies for my mistake! I completely failed to realize this. This commit is part of an assignment for my university's software architecture class where we have been tasked to study an open source project for the duration of the semester. Our last assignment requires us to attempt to resolve a series of open issues in the project's repository. Is there a way for this issue to be transferred over to the [https://github.com/godotengine/godot/] repository so that I can properly resolve this issue? Please let me know! Thank you so much.

YuriSizov:
The issue that you are trying to resolve is in the main repository as it is. And PRs can't be transferred, because these are totally different projects with different files in them, you need to make a new one. Same for the other PR.

For the record, it doesn't matter in which repository the issue is. A PR made for one repo can close an issue in another repo, if the person merging it has the write access. But that's for maintainers to worry, not for you. In this case some issues can be created in one Godot repo, and fixed in another, and it will just work from your perspective.

PR #7268, opened 2023-05-02
docs: add script templates example

russmatney:
First of all, thanks much for all the work on such an amazing engine and docs site! :tada: 

I ran into the new script template style today - the docs are solid, but could use an actual example to confirm what they imply. Using a node name as a directory name is kind of odd - this would have helped me figure it out at a glance.


mhilbrunner (approved):

mhilbrunner:
Good call adding an example here. Thanks and congrats on your first merged contribution to Godot's docs!

russmatney:
Woo! Thanks for the quick merge!

PR #7274, opened 2023-05-04
Build ePubs of the documentation for offline reading

Calinou:
This format is more suited to mobile devices or e-readers compared to HTML. Note that the generated ePubs are quite large (156 MB for `master`) and may struggle on low-end devices.

PDF builds could also be added, but there isn't as much of an use when you have both HTML and ePub already.

### Outstanding issues

- Table of contents generation appears to be broken (there's only one chapter). I remember it working at some point while working on this branch, but I don't know what broke it.
- The class reference is missing. Like the above issue, it also worked at some point, but I don't know why it's missing now.
  - The lack of a built-in class reference may be desired to make the ePub easier to open though. With the class reference, it's 5,500 pages instead of 2,900 in `master`. You can already read the class reference offline in the editor anyway.
- Formatting could be improved somewhat to make it more "book-like", but I don't know how to customize the ePub output. For example:
  - Give admonitions their usual colored background.
  - Give headings a colored background depending on their level.
  - Add a line below H1 headings.
  - Force `figure` images to be center-aligned and on a new line. Right now, they ignore the `:align: center` option and may appear on the same line as text.
  - Give `figure` captions a different text style (italic, slightly different color).
  - Fix visible blank lines between lines in multiline code blocks.
- The margin on the page's edges is very low, but this may not be a bad idea as it should help with readability on mobile devices.

**Note:** This PR changes the stable branch that is built to `3.6`, so that we have both Godot 3.x and 4.x docs available for download. Stable branches will need to have their URLs updated accordingly. There isn't much use in having both `stable` (4.0) and `latest` (4.1) built right now, since these barely have any differences. In the long term, I'd suggest restoring `stable` builds, but this could be done after 4.1 is released.

___

Example outputs: <sub>(links expire in September 2023)</sub>

- `master`: https://0x0.st/HZs0.zip
- `3.6`: https://0x0.st/HZsU.6.zip

### DIFF
--- a/.github/workflows/build_offline_docs.yml
+++ b/.github/workflows/build_offline_docs.yml
@@ -8,12 +8,12 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     strategy:
       matrix:
         branch:
           - master
-          - stable
+          - 3.6
     steps:
       - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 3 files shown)


mhilbrunner (approved):

mhilbrunner:
Thanks, lets try this out! 🎉

PR #7276, opened 2023-05-04
Document setting up Vulkan validation layers on desktop platforms

Calinou:
@bruvzg Do you know the steps required to set up validation layers on iOS?

### DIFF
--- a/contributing/development/debugging/vulkan/vulkan_validation_layers.rst
+++ b/contributing/development/debugging/vulkan/vulkan_validation_layers.rst
@@ -3,16 +3,150 @@
 Validation layers
 =================
 
-Validation layers enable developers to verify their application's correct use
-of the Vulkan API.
+Validation layers enable developers to verify their application's correct use of
+the Vulkan API. Validation layers can be enabled in both debug and release
+builds, including in exported projects.
 
-.. _doc_vulkan_validation_layers-android:
+.. note::
+
+    Enabling validation layers has a performance impact, so only enable them
+    when you actually need the output to debug the application.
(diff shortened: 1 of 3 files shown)


bruvzg:
> Do you know the steps required to set up validation layers on iOS?

I do not think it is supported, MoltenVK can't load layers directly - https://github.com/KhronosGroup/MoltenVK/blob/main/Docs/MoltenVK_Runtime_UserGuide.md#limitations, so the loaded is required, on macOS you can use volk + `libvulkan.1.dylib` and load layers the same way as Linux (dylib and pre-built layer files are part of the LunarG SDK). But there's no loader for iOS.

contributing/development/debugging/vulkan/vulkan_validation_layers.rst
```diff
@@ -3,10 +3,87 @@
...
+    when you actually need the output to debug the application.
+
+Windows/macOS
+-------------
+
+Install the Vulkan SDK `<https://vulkan.lunarg.com/sdk/home>`__, which contains
+validation layers as part of its default installation. No need to enable any
+optional features in the installer; installing the core Vulkan SDK suffices. You
+don't need to reboot after installing the SDK, but you may need to close and
+reopen your current terminal.
```
bruvzg, line 22: In case of macOS, layers will work only when using Godot built with `use_volk=yes`.

mhilbrunner (approved):

mhilbrunner:
Thanks, merged! And thanks for reviewing, bruvzg.

PR #7299, opened 2023-05-06
Update scenes_versus_scripts.rst minor correction

Ubiguchi:
Swap mention of scene and script that mistakes scripts as declarative.

### DIFF
--- a/tutorials/best_practices/scenes_versus_scripts.rst
+++ b/tutorials/best_practices/scenes_versus_scripts.rst
@@ -11,8 +11,8 @@ declarative code.
 
 Each system's capabilities are different as a result.
 Scenes can define how an extended class initializes, but not what its
-behavior actually is. Scenes are often used in conjunction with a script so
-that the scene acts as an extension of the scripts declarative code.
+behavior actually is. Scenes are often used in conjunction with a script, 
+the scene declaring a composition of nodes, and the script adding behaviour with imperative code.
 
 Anonymous types
 ---------------
--- a/tutorials/best_practices/what_are_godot_classes.rst
+++ b/tutorials/best_practices/what_are_godot_classes.rst
@@ -54,8 +54,8 @@ The behavior of scenes has many similarities to classes, so it can make sense to
 a class. Scenes are reusable, instantiable, and inheritable groups of nodes. Creating a scene is
 similar to having a script that creates nodes and adds them as children using ``add_child()``.
 
-We often pair a scene with a scripted root node that makes use of the scene's nodes. As such, the
-scene is often an extension of the script's declarative code.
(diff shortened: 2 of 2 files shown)


mhilbrunner:
I agree that senctence is a bit weird, just swapping the mentions won't work though, Scenes don't really have "declarative code".

Ubiguchi:
Fair point, however the docs mentions scene's as having declarative code or being declarative in numerous places, so this PR improves consistency even if the docs' concept of declarative code is not quite correct.

I'll do some further investigation and see if I can isolate incorrect references to declarative code.

Ubiguchi:
So I think I can explain the issue clearly now, so let me explain it, and if you agree I'll close the PR and raise another to address the issue across the docs. 

As clearly stated in  [scene organization](https://docs.godotengine.org/en/stable/tutorials/best_practices/scene_organization.html#choosing-a-node-tree-structure), scenes are declarative and scripts are imperative. This means scenes are _defined_ declaratively, which you can see by opening their tscn file that contains declarative code. In contrast, scripts _contain_ or are _defined_ by imperative code, which you can see by opening their file.

The problem with the PR is therefore the word "has" which implies ownership.

The [scenes_versus_scripts.rst](https://docs.godotengine.org/en/stable/tutorials/best_practices/scenes_versus_scripts.html#when-to-use-scenes-versus-scripts) content could therefore be updated as follows:

---
**When to use scenes versus scripts**

We've already covered how scenes and scripts are different. Scripts define an engine class extension with imperative code, scenes with declarative code.

Each system's capabilities are different as a result. Scenes can define how an extended class initializes, but not what its behavior actually is. Scenes are often used in conjunction with a script ~so that the scene acts as an extension of the scripts declarative code.~, **the scene declaring a composition of nodes, and the script adding behaviour with imperative code.** 

--- 

The other place in the docs that needs similar treatment is [Applying Object-Oreinted principles in Godot](https://docs.godotengine.org/en/stable/tutorials/best_practices/what_are_godot_classes.html#scenes), which could be corrected as follows:

---
The behavior of scenes has many similarities to classes, so it can make sense to think of a scene as a class. Scenes are reusable, instantiable, and inheritable groups of nodes. Creating a scene is similar to having a script that creates nodes and adds them as children using add_child().

We often pair a scene with a scripted root node that makes use of the scene's nodes. ~As such, the scene is often an extension of the script's declarative code.~**As such, the script extends the scene by adding behavior through imperative code.**

Ubiguchi:
Okay, changes made to PR as per previous comments.

mhilbrunner (approved):

mhilbrunner:
Thank you! Congrats on your first merged contribution to Godot's documentation. 🎉

PR #7308, opened 2023-05-08
Error in gdexample.cpp

Yunusemre-Oeztuerk:
When adding the getter and setter functions to the example project a void has been added to the constructor  which was not there during the creation of the gdexample.cpp file and is also causing the error C2533: constructors not allowed a return type. Simply removing it again fixes it.

Using Windows 10 and Godot 4.0.2


paddy-exe (approved):

paddy-exe:
Good catch! Thanks!

mhilbrunner (approved):

mhilbrunner:
Thank you! Merged.

PR #7322, opened 2023-05-09
Remove obsolete and confusing tip re: move_and_slide()

cbscribe:
Since we no longer pass a velocity to `move_and_slide()`, the mention of `delta` here is no longer relevant, and probably misleading. It was originally needed to distinguish between the very similar-looking calls to `move_and_collide()`.


mhilbrunner (approved):

mhilbrunner:
Cherry-picked to `4.0`/`stable` docs branch.

PR #7339, opened 2023-05-11
Update submitting_to_assetlib.rst

toafloast:
Added asset file structure (addons/AssetName) to reccomendations. Assets on the Asset Library are inconsistent in file structure. This is often problematic, as the LICENSE file is typically placed inside of res://, which will usually either get overwritten or ignored.  While the submission guidelines heavily imply this is how you should structure your assets, it was never explicitly stated.

### DIFF
--- a/community/asset_library/submitting_to_assetlib.rst
+++ b/community/asset_library/submitting_to_assetlib.rst
@@ -60,7 +60,7 @@ if you follow these recommendations, you can help make the asset
 library a better place for all users.
 
 * When creating non-project assets, it is common practice to place your files
-  inside of an **addons/AssetName/** folder. Do this to avoid having your files 
+  inside of an **addons/asset_name/** folder. Do this to avoid having your files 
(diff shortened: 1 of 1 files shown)


community/asset_library/submitting_to_assetlib.rst
```diff
@@ -59,6 +59,11 @@ These things are not required for your asset to be approved, but
 if you follow these recommendations, you can help make the asset
 library a better place for all users.
 
+* When creating non-project assets, it is common practice to place your files
+  inside of an **addons/AssetName/** folder. Do this to avoid having your files
```
aaronfranke, line 63:
```suggestion
  inside of an **addons/asset_name/** folder. Do this to avoid having your files 
```

We want to encourage lowercase file names to avoid issues with case sensitivity.

mhilbrunner (approved):

mhilbrunner:
Thanks and congrats on your first merged contribution to Godot's docs!

PR #7340, opened 2023-05-12
Document using NoiseTexture3D with FogMaterial in Volumetric fog

Calinou:
This also adds a tip about using a NoiseTexture3D as particle attractor vector field texture.

Don't cherry-pick as [NoiseTexture3D is a new feature in 4.1](https://github.com/godotengine/godot/pull/76486).

PR #7341, opened 2023-05-12
Document faking volumetric fog using QuadMeshes

Calinou:
- Follow-up to https://github.com/godotengine/godot-docs/pull/7340.

In some cases, it may be better to use specially configured QuadMeshes
as an alternative to volumetric fog.

- Quads work with any rendering method, including Forward Mobile and Compatibility.
- Quads do not require temporal reprojection to look smooth, which makes them suited to fast-moving dynamic effects such as lasers. They can also represent small details which volumetric fog cannot do efficiently.
- Quads generally have a lower performance cost than volumetric fog (with some exceptions).


mhilbrunner (approved):

mhilbrunner:
Thank you!

mhilbrunner:
Cherry-picked to `4.0`/`stable` docs branch.

PR #7345, opened 2023-05-14
Add Loop Break To Doc Code

MrDDaye:
Issue in collision count loop. 

If multiple collisions occur with mob over one `move_and_slide()`, score is incremented multiple times. Only tested in GDScript, but likely the same issue in C# too.

Found in Godot 4.0.2


AThousandShips:
There's another PR that does this more completely:
* #7767

AThousandShips:

PR #7351, opened 2023-05-15
Change cpu_optimization.rst languages to link to gdextension implementations

Lunarmagpie:
These links linked to the godot 3 language libraries but the docs are for godot 4. I replaced rust with the new project and removed JS because I don't know of a godot 4 library for it.

### DIFF
--- a/tutorials/performance/cpu_optimization.rst
+++ b/tutorials/performance/cpu_optimization.rst
@@ -191,8 +191,7 @@ Other languages
 ~~~~~~~~~~~~~~~
 
 Third parties provide support for several other languages, including `Rust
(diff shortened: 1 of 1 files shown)


mhilbrunner:
I think the other one can stay, it seems to have a Godot 4 branch that is actively maintained: https://github.com/Geequlim/ECMAScript/tree/gd4

Maybe we should link directly to that branch though.

mhilbrunner (approved):

mhilbrunner:
After more in-depth review, that branch isn't fully ready for Godot 4 yet, so its probably fine to remove now and re-add this later.

Thanks for contributing and congrats on your first merged contribution! 🎉

PR #7361, opened 2023-05-17
Update class_navigationagent2d.rst to fix minor typo

DevBadger10:
Changed "cause" to "because" to add some more formality into this area of the docs.


mhilbrunner:
Hey! Thanks for contributing. Everything under `classes/` is autogenerated from XML files in the main repository, so the change needs to be made here instead: https://github.com/godotengine/godot/blob/master/doc/classes/NavigationAgent2D.xml
Also, the 3D version has the same issue, so the change should be made there as well: https://github.com/godotengine/godot/blob/master/doc/classes/NavigationAgent3D.xml

PR #7376, opened 2023-05-18
Fix wrong navigation agent example code in

smix8:
Fixes wrong navigation agent example code.

Kinda silly that the NavigationAgent example code had multiple C++ examples in the GDScript section for months that I copied from the source code when drafting the documentation the first time. Now while updating I drafted again with C++ source code and forgot to change.

Not a single user complain and now it was even changed again with the pr today without fixing it and it passed the pr review.

Every single person involved should pour some ash on their head, including me.

### DIFF
--- a/tutorials/navigation/navigation_using_navigationagents.rst
+++ b/tutorials/navigation/navigation_using_navigationagents.rst
@@ -145,14 +145,14 @@ toggle avoidance on agents, create or delete avoidance callbacks or switch avoid
 
     var agent: RID = get_rid()
     # Enable avoidance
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):

mhilbrunner:
Thanks! Does this need a separate PR for 4.0?

smix8:
@mhilbrunner
Yes, cause in Godot 4.0 the `agent_set_avoidance_callback()` function is named `agent_set_callback()` and `agent_set_use_3d_avoidance()` does not exist.

PR #7378, opened 2023-05-18
Update GDExtension tutorial for entry functions

paddy-exe:
Docs part of https://github.com/godotengine/godot-cpp/pull/1095 
Same changes as in https://github.com/godotengine/godot-cpp/pull/1115

### DIFF
--- a/tutorials/scripting/gdextension/gdextension_cpp_example.rst
+++ b/tutorials/scripting/gdextension/gdextension_cpp_example.rst
@@ -268,8 +268,8 @@ GDExtension plugin.
 
     extern "C" {
     // Initialization.
-    GDExtensionBool GDE_EXPORT example_library_init(const GDExtensionInterface *p_interface, const GDExtensionClassLibraryPtr p_library, GDExtensionInitialization *r_initialization) {
-        godot::GDExtensionBinding::InitObject init_obj(p_interface, p_library, r_initialization);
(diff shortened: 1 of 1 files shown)


tutorials/scripting/gdextension/gdextension_cpp_example.rst
```diff
@@ -369,6 +370,8 @@ loaded for each platform and the entry function for the module. It is called ``g
     android.release.arm64 = "res://bin/libgdexample.android.template_release.arm64.so"
 
 This file contains a ``configuration`` section that controls the entry function of the module.
+You are also able to set a minimum compatible version with ``compatability_minimum`` of Godot
+which prevents older version of Godot to load your extension.
```
dsnopek, line 374: IMO we should be a little more assertive about this. Maybe:

```suggestion
You should also set the minimum compatible Godot version with ``compatability_minimum``,
which prevents older version of Godot from trying to load your extension.
```

I also snuck in a couple unrelated wording changes that sound more natural to me, but you don't have take those. :-)

My main suggestion is switching from "You are also able" to "You should" which is a little more assertive.

mhilbrunner:
Is this relevant for 4.0 or 4.1+ only?

paddy-exe:
> Is this relevant for 4.0 or 4.1+ only?

This will be 4.1+ only.

tutorials/scripting/gdextension/gdextension_cpp_example.rst
paddy-exe: Thanks for the feedback! I put it in because I think it does sound better than my version

mhilbrunner (approved):

mhilbrunner:
Thank you! Merged.

PR #7380, opened 2023-05-19
mention that  is not an allowed operator for strings and stringnames

you-win:
Mention in the docs for `String` and `StringName` that the `not` operator cannot be used on `String`s and `StringName`s.

[Related resolved issue](https://github.com/godotengine/godot/issues/58579)


you-win:
It appears I made this edit in the wrong place, will submit a PR for the main godot repo editing the class xml docs instead.

PR #7393, opened 2023-05-20
Fix vector math tutorial

### LINKED ISSUE #7388: Documentation about vector math explaining reflection is outdated
Godot 4.0.2

The page about vector mathematics in the Reflection section contains an outdated code example
GDScript example generates an error "invalid get index 'normal' (on base KinematicCollision2D)"

move_and_collide() function returns KinematicCollision2D that no longer gives direct access to the the normal and the remainder attributes

Looks like "collision.normal" should be replaced with "collision.get_normal( )" and
"collision.remainder"  should be replaced with "collision.get_remainder()"

https://docs.godotengine.org/en/stable/tutorials/math/vector_math.html#reflection

yashmistri:
fixes mistakes in the "reflection" code example in both gdscript and c#

cast delta to float in C# example because delta is typed as double and Vector2 cant be multiplied by double as per godotengine/godot#70314
tested in 4.0.2.stable.mono

fixes #7388

### DIFF
--- a/tutorials/math/vector_math.rst
+++ b/tutorials/math/vector_math.rst
@@ -232,18 +232,18 @@ Here is a GDScript example of the diagram above using a :ref:`CharacterBody2D
     # object "collision" contains information about the collision
     var collision = move_and_collide(velocity * delta)
     if collision:
-        var reflect = collision.remainder.bounce(collision.normal)
-        velocity = velocity.bounce(collision.normal)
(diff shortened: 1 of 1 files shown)


raulsntos (approved):
LGTM. Thanks for contributing to the C# documentation.

mhilbrunner (approved):

mhilbrunner:
Thank you! Merged.

PR #7406, opened 2023-05-24
Add information about 3D asset direction conventions to "Importing 3D scenes"

aaronfranke:
This PR adds a note to the documentation to explain the different conventions and specifications about orienting 3D assets in Godot. It explains Godot/OpenGL camera directions, Godot/Blender/glTF asset directions, and Godot/Blender cardinal directions. It links to the glTF specification for +Z being codified as asset front. I tried to be careful where I use the words "forward" vs "front" to make things clear.

Why not open this PR earlier: Before we were not certain what would end up in the engine, but now it is fairly certain so now seems like a good time. For example initially there was some push from @TokageItLab to add `GLTF_`* constants and document how glTF works in the core engine. I fully agree that this information must be documented, I just disagreed on it being in the core engine's constants and documentation, so now it can live in the docs repo :)

I'm fully open to feedback on any ways we can make this documentation more clear, concise, and comprehensive. I think we can all agree that coordinate systems are a confusing topic.

The third paragraph in this PR refers to the `use_model_front` option and the `MODEL_`* constants, so it expects that https://github.com/godotengine/godot/pull/75875 and/or https://github.com/godotengine/godot/pull/72842 is merged into the engine.

### DIFF
--- a/tutorials/assets_pipeline/importing_scenes.rst
+++ b/tutorials/assets_pipeline/importing_scenes.rst
@@ -29,6 +29,33 @@ Copy the scene file together with the textures and mesh data (if separate) to
 the project repository, then Godot will do a full import when focusing the
 editor window.
 
+3D asset direction conventions
+------------------------------
+
+Godot uses a right-handed, Y-is-up coordinate system, with the -Z axis as
+the camera's forward direction. This is the same as OpenGL. This implies
+that +Z is back, +X is right, and -X is left for a camera.
+
+The convention for 3D assets is to face the opposite direction as the camera,
+so that characters and other assets are facing the camera by default.
+This convention is extremely common in 3D modeling applications, and is
+`codified in glTF as part of the glTF 2.0 specification <https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#coordinate-system-and-units>`.
+This means that for oriented 3D assets (such as characters),
+the +Z axis is the direction of the front, so -Z is the rear,
++X is the left side, and -X is the right side for a 3D asset.
+In Blender, this means that +Y is rear and -Y is front for an asset.
+
+When rotating an oriented 3D asset in Godot, use the ``use_model_front``
+option on the ``look_at`` functions, and use the ``Vector3.MODEL_*``
+constants to perform calculations in the oriented asset's local space.
+
+For assets without an intrinsic front side or forward direction, such as
+a game map or terrain, take note of the cardinal directions instead.
+The convention in Godot and the vast majority of other applications is
+that +X is east and -X is west. Due to Godot's right-handed Y-is-up
+coordinate system, this implies that +Z is south and -Z is north.
+In Blender, this means that +Y is north and -Y is south.
+
 Exporting glTF 2.0 files from Blender (recommended)
 ---------------------------------------------------
 


tutorials/assets_pipeline/importing_scenes.rst
```diff
@@ -29,6 +29,33 @@ Copy the scene file together with the textures and mesh data (if separate) to
...
+that +Z is back, +X is right, and -X is left for a camera.
+
+The convention for 3D assets is to face the opposite direction as the camera,
+so that characters and other assets are facing the camera by default.
+This convention is extremely common in 3D modeling applications, and is
+`codified in glTF as part of the glTF 2.0 specification <https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#coordinate-system-and-units>`.
+This means that for oriented 3D assets (such as characters),
+the +Z axis is the direction of the front, so -Z is the rear,
++X is the left side, and -X is the right side for a 3D asset.
+In Blender, this means that +Y is rear and +Y is front for an asset.
```
TokageItLab, line 46: The asset's front in Blender is -Y and world space forward is +Y. However, it is complicated but for cameras and lights, the front is -Z even if it is Blender.

aaronfranke: Ah yeah that's a typo, I accidentally wrote +Y twice. Good catch.

I think we only need to talk about the asset forward to Blender since it's weird with cameras/etc.

TokageItLab (approved):
I thinks it's fine. However, as mentioned in https://github.com/godotengine/godot/issues/75228, there is often confusion in the handling of coordinates in Blender, so I would like to add another troubleshooting page separate from this.

tutorials/assets_pipeline/importing_scenes.rst
```diff
@@ -29,6 +29,33 @@ Copy the scene file together with the textures and mesh data (if separate) to
...
 editor window.
 
+3D asset direction conventions
+------------------------------
+
+Godot uses a right-handed, Y-is-up coordinate system, with the -Z axis as
+the camera's forward direction. This is the same as OpenGL. This implies
+that +Z is back, +X is right, and -X is left for a camera.
+
+The convention for 3D assets is to face the opposite direction as the camera,
```
TokageItLab, line 39: It is said as a convention in here, but my understanding is that the definition of the right-hand coordinate system is that the camera and the asset are facing each other as the assumption. But I am not certain that this is the truth, so it may be best to defer to reduz for judgment.

aaronfranke: The handedness of the coordinate system itself does not make any assumptions of what things exist in it. The handedness only determines how the axes relate to each other. That is, if you know X and Y, where is Z? In right-handed, +Y up and +X right implies +Z back, alternatively +Y up and +X left implies +Z forward, etc.

TokageItLab: What I am saying is that in "any" coordinate system, the +X axis of the camera is consistent with being on the right.

aaronfranke: That's not how handedness is defined mathematically, though. Math doesn't know about assets or cameras.

+X camera right is not consistent in all apps thanks to Unreal throwing a monkey wrench into things. Unreal uses +X forward, +Y right, and +Z up (this is left-handed).

tutorials/assets_pipeline/importing_scenes.rst
```diff
@@ -29,6 +29,33 @@ Copy the scene file together with the textures and mesh data (if separate) to
...
+so that characters and other assets are facing the camera by default.
```
skyace65, line 40: This part is confusing me (and making me realize my 3D teachers in college taught even less than I thought, which wasn't much). What exactly is "the camera" that you're referring to here? I assume it's not one you would attach to a player character

aaronfranke: @skyace65 Does this image help you understand?

image

TokageItLab: It may be paraphrased as a viewpoint of the viewport.

skyace65: So the camera in this case would be the viewport camera when you a open a blank scene?

TokageItLab: In most cases, yes.

mhilbrunner (approved):

mhilbrunner:
@aaronfranke Any part of this that still needs to be discussed (i.e., the `discussion` label) - it seems like a good improvement from my POV

PR #7409, opened 2023-05-24
Make it possible to read abbreviations by tapping on mobile

Calinou:
## Preview

Screenshot_20230524_120313_Bromite


YuriSizov:
Shouldn't the browser handle this already?

mhilbrunner:
I agree they should, but apparently mobile browsers by default don't for whatever reason (lack of demand?) - see e.g. https://bitsofco.de/making-abbr-work-for-touchscreen-keyboard-mouse/, and indeed in my testing at least latest mobile Safari doesn't seem to do anything for `abbr` by default.

mhilbrunner:
Thanks!

PR #7417, opened 2023-05-25
Remove outdated warning from Importing translations tutorial

### LINKED ISSUE #7413: Warning on tutorial page for importing translations is out of date
Your Godot version: 4.0

Issue description:

The tutorials documentation page for importing translations warns that there is no native way to save an Excel file as a UTF-8 encoded CSV. However, when I use Save As in Excel I do have the option in the drop-down to choose a UTF-8 encoded CSV file format. The link on the documentation page that confirms this as an issue and offers a workaround was last updated about 6 years ago. I'm thinking Excel has updated since then and this warning is likely no longer needed. 

URL to the documentation page:

https://docs.godotengine.org/en/stable/tutorials/assets_pipeline/importing_translations.html

MisoMosiSpy:
Fixes #7413


skyace65 (approved):
Confirmed the option is there in myself, and it's the first CSV option in the dropdown list so I don't think we need a note at all.

PR #7422, opened 2023-05-26
Remove agent avoidance page

smix8:
Removes agent avoidance page.

While reworking the pages for the avoidance rework pr I realized that this page now has mostly old or even wrong information, duplicated information from other pages, or information that should be part directly of other pages like the navigation agent or navigation obstacles pages. With all that change the page would be left empty.

I originally had plans to have a page about the NavigationAvoidance as a system. The current agent avoidance page was made at a point where the old avoidance was hard limited on just agents but now we have working obstacles and all kinds of other avoidance features. It does not make sense with the current page name and url so I think it is better to just do some house cleaning and remove it.


mhilbrunner:
I'm taking from the branch name that this should _not_ be cherrypicked for 4.0?

YuriSizov:
Well, the rework is only 4.1+, so 4.0 docs should remain as is. We can perhaps add a note if there are some issues with the article that we are not planning to fix.

smix8:
This should only be for 4.1.
The old 4.0 version is correct for that version. The old rvo avoidance had many problems that is why that old page also focused so much on those. They are fixed in 4.1 or mentioned in better places where they make contextual more sense.

mhilbrunner (approved):

mhilbrunner:
Thanks! Merged.

PR #7426, opened 2023-05-27
GDScript: Add mention that `@onready @export` doesn't work as expected

dalexeev:
Closes https://github.com/godotengine/godot-docs/issues/7428

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_basics.rst
+++ b/tutorials/scripting/gdscript/gdscript_basics.rst
@@ -404,11 +404,11 @@ between parentheses and separated by commas.
 
 Both of these are the same::
 
-    @onready
-    @export_node_path("TextEdit", "LineEdit")
-    var input_field
+    @annotation_a
+    @annotation_b
+    var variable
 
-    @onready @export_node_path("TextEdit", "LineEdit") var input_field
+    @annotation_a @annotation_b var variable
 
 .. _doc_gdscript_onready_annotation:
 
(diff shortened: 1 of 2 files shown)


tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -435,6 +438,31 @@ can replace the above code with a single line::
 
     @onready var my_label = get_node("MyLabel")
 
+.. warning::
+
+    Applying ``@onready`` and any ``@export`` annotation to the same variable
+    doesn't work as you might expect. The ``@onready`` annotation will make
```
mhilbrunner, line 444:
```suggestion
    doesn't work as you might expect. The ``@onready`` annotation will cause
```

dalexeev: I copied the text from here:

YuriSizov:
I don't think that examples should be peppered with warning ignore annotations. It makes them messier and the need for a warning is not explained. I don't think they add to the point being demonstrated, even if they make this code runnable without errors.

dalexeev:
> I don't think that examples should be peppered with warning ignore annotations.

In the first case, this is an example of using multiple annotations on the same variable.  Since `@onready` doesn't work with `@export`, I changed it to `@warning_ignore`. We can use abstract examples (`@annotation_a`, `@annotation_b`) instead of real examples, if you wish.

In the second case, the code demonstrates what happens when both `@onready` and `@export` are used at the same time (if the warning is disabled or ignored).

YuriSizov:
Yes, I think the first example could be something neutral. Having a warning ignore annotation in there can be considered at face value.

For the second one having an annotation conflicts with the text where you say we don't recommend ignoring it. And people are going to come by and copy the sample as is, they always do, so giving it in this unsafe way can propagate bad coding practices. You can instead link to whatever documentation we have for suppressing warnings in the text and let the example fail if they try to use it.

mhilbrunner (approved):

mhilbrunner:
Thank you! Merged.

PR #7436, opened 2023-05-29
Update pygments to 2.15.1 and restore fixed shader syntax highlighting

bitsawer:
Updates pygments to 2.15.1 now that PR's https://github.com/pygments/pygments/pull/2350 and https://github.com/pygments/pygments/pull/2357 have been merged and tested in the wild for a while. Also restores the GLSL syntax highlighting to the previously buggy codeblocks that those PR's fixed.

I did a clean build and everything seems to work with no warnings or errors. If  you are testing locally, remember to upgrade your packages (`pip install -U -r requirements.txt` or pip3).

code

PR #7289, opened 2023-05-05
RTD: Change OS image to Ubuntu 22.04

mhilbrunner:
This should hopefully fix the recent build errors on ReadTheDocs due to a recent OpenSSL change in urllib3

See:
- https://github.com/urllib3/urllib3/issues/2168
- https://github.com/readthedocs/readthedocs.org/issues/10290

PR #7443, opened 2023-05-30
Update Sphinx and requirements

mhilbrunner:
In the vein of https://github.com/godotengine/godot-docs/pull/7436, this is a start on updating the version of Sphinx and its requirements we're using as everything is getting kinda old (we use Sphinx 4.x, current is 7.x; this PR updates it to 5.x at least).

Updating to Sphinx 6.x currently seems to break some things (like the sidebar menu toggling) which we need to investigate.

Checked in a local build:
- General build and layout
- Sidebar menu
- Images
- Code tabs
- Code highlighting

Things that seem iffy/may need testing:
- Search

Supersedes https://github.com/godotengine/godot-docs/pull/6575, https://github.com/godotengine/godot-docs/pull/6755, https://github.com/godotengine/godot-docs/pull/6749
Sphinx changelog: https://www.sphinx-doc.org/en/master/changes.html#release-5-0-2-released-jun-17-2022

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,16 +6,16 @@
 pygments==2.15.1
 
 # Sphinx base and RTD theme.
-sphinx==4.4.0
-sphinx_rtd_theme==1.1.1
+sphinx==5.0.2
+sphinx_rtd_theme==1.2.1
 
 # Sphinx extensions.
 
 # Code tabs extension to display codeblocks in different languages as tabs.
-sphinx-tabs==3.4.0
+sphinx-tabs==3.4.1
 # Adds a 'copy' button to the right of codeblocks.
-sphinx-copybutton==0.5.1
+sphinx-copybutton==0.5.2
 # Custom 404 error page (more useful than the default).
 sphinx-notfound-page==0.8.3
 # Adds Open Graph tags in the HTML `<head>` tag.
-sphinxext-opengraph==0.7.5
+sphinxext-opengraph==0.8.2


mhilbrunner:
> but since it works right now and it is not a security concern, we could just keep it as is.

My general feeling is that the longer we wait, the more effort updating is going to take eventually (because _everything_ is then broken, instead of a small breakage here or there that is easier to locate and fix), and eventually we _will_ need to update.

Something like https://github.com/godotengine/godot-docs/pull/7289: we are forced to update a single component (like the OS), which may then require us to live with a newer Python, which requires a newer Sphinx, or which requires newer dependencies etc.
Or a security issue is discovered, we now need to update a component urgently, but doing that now takes a lot of time as everything grew so outdated.

With that said, I agree this is a consideration, and currently there is nothing pressing forcing us to do an update. I'd just prefer to do it eventually on our terms, instead of it kinda coming back to haunt us :)

YuriSizov:
Yeah, that's fair. If we don't update regularly, we may be in a bad place to do that by the time it's actually required. Though I don't see anything particularly critical that might arise in the future. It's just a PITA to adjust our stuff all the time. Perhaps we should implement it in a less hacky way. For example, we can actually fork the theme and make our adjustments a native part of our forked theme.

Piralein:
- the sphinx_rtd_theme is not compatible with sphinx 7 at the moment, the highest version we can choose is 6.2.1
- The sidebar toggling is broken because sphinx 6.0 removed jquery, adding it manually restores the functionality. (https://www.sphinx-doc.org/en/master/changes.html#id91)

I did a complete build with 6.2.1 and got a lot of svg errors, not sure what this is about.

YuriSizov:
We also extend templates, both core and from the RTD theme. So we need to double-check that our overrides don't erase anything added by newer versions.

Piralein:
I tested locally with:
````
sphinx==6.2.1
sphinx_rtd_theme==1.3.0
sphinx-tabs==3.4.1
sphinx-copybutton==0.5.2
sphinx-notfound-page==1.0.0
sphinxext-opengraph==0.8.2
````

All modules seem to be compatible. There are only 2 issues:

1. The warnings with social cards - added by sphinxext-opengraph
This can be disabled, see: https://github.com/wpilibsuite/sphinxext-opengraph/blob/main/docs/source/socialcards.md
````
ogp_social_cards = {
    "enable": False
}
```` 

2. our config prevents the `sphinx_rtd_theme` to load the `sphinxcontrib.jquery` module.
https://github.com/readthedocs/sphinx_rtd_theme/issues/1452
It works by removing `html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]`.
This seems to be some outdated configuration no longer used. The documentation of  https://sphinx-rtd-theme.readthedocs.io/en/stable/installing.html doesn't mention it.

we could also add it to the extension ourself:
````
extensions = [
    "sphinxcontrib.jquery",
]
````

**Comment by Piralein** [comment]
```
pygments==2.18.0

sphinx==6.2.1
sphinx_rtd_theme==1.3.0

sphinx-tabs==3.4.5
sphinx-copybutton==0.5.2
sphinx-notfound-page==1.0.2
sphinxext-opengraph==0.9.1

sphinxcontrib-applehelp==1.0.8
sphinxcontrib-htmlhelp==2.0.5
sphinxcontrib-qthelp==1.0.7
sphinxcontrib-serializinghtml==1.1.10
sphinxcontrib-devhelp==1.0.6

sphinxcontrib-video==0.2.1
```
My current setup, I haven't discovered any problems yet. (with the changes above)

mhilbrunner:
Superseded by https://github.com/godotengine/godot-docs/pull/10158.

PR #7447, opened 2023-05-31
Fix NavigationLink doc using old property names

smix8:
Fixes NavigationLink doc using old property names.

The properties that use `location` were renamed to `position` a while ago.
The term `navmesh` mixed with mentions of navigation meshes confused some users so it gets streamlined to just `navigation mesh`.

Could be backported to Godot 4.0 docs.


mhilbrunner (approved):

mhilbrunner:
Thank you! Merged.

PR #7448, opened 2023-05-31
NavigationAgent2D, NavigationAgent3D, Navigation Overview typo fixes

JustinDodemaide:
For both NavigationAgent2D and NavigationAgent3D, in the description of property 'target_desired_distance': Changed "to low" to "too low"
Changed "cause" to "because"

For 2D  Navigation Overview, under NavigationServer2D: Changed "an navmesh" to "a navmesh"

### DIFF
--- a/classes/class_navigationagent2d.rst.bak
+++ b/classes/class_navigationagent2d.rst.bak
@@ -582,7 +582,7 @@ Does not affect normal pathfinding. To change an actor's pathfinding radius bake
 - void **set_target_desired_distance** **(** :ref:`float<class_float>` value **)**
 - :ref:`float<class_float>` **get_target_desired_distance** **(** **)**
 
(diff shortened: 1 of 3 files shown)


skyace65:
Two things. One, it looks like you've committed some .bak files you didn't mean to.

Two, the class reference is not updated in the docs repository. The pages in the docs repository are generated from the main Godot repository. The files you're looking for are [here](https://github.com/godotengine/godot/blob/master/doc/classes/NavigationAgent2D.xml) and [here](https://github.com/godotengine/godot/blob/master/doc/classes/NavigationAgent3D.xml).

You can find information on updating the class reference in the docs [here](https://docs.godotengine.org/en/stable/contributing/documentation/updating_the_class_reference.html). Additionally when you update this PR please keep everything as 1 commit, if you don't know how to do it there's information [here](https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html#modifying-a-pull-request).

skyace65:
You don't need to close this one because the change you made for the navmesh 2D page is still valid, you would just need to remove the other pages that aren't updated in this repository, and open a separate PR in the engine repository for those.

PR #7474, opened 2023-06-05
Document `-j auto` in Building the manual

Calinou:
I was considering defaulting to `-j2` or even `-j3`, but Windows systems with 8 GB of RAM would likely struggle with this.

In the meantime, it's worth documenting Sphinx's `-j auto` for people with lots of RAM. On a system with 32 threads, a full clean `make html` build can peak to more than 20 GB of RAM for Sphinx itself:

image

With `-j auto` on an Intel Core i9-13900K, the documentation builds in 1m16s, whereas `-j4` requires 4m15s.


mhilbrunner (approved):

mhilbrunner:
Thanks!

PR #7481, opened 2023-06-07
Document EMJC bitmap compression not being supported in Using fonts

Calinou:
- Document which system fonts are used on Android when using aliases.
- Mention font metrics varying across platforms when using system fonts.
- Fix notice about which platforms support system fonts.


mhilbrunner (approved):

mhilbrunner:
Merged, thank you! 🎉

PR #7483, opened 2023-06-07
Update threading code examples to use new syntax

bitsawer:
I noticed that some threading code examples still use old Godot 3.x syntax which doesn't work in 4.x anymore. Updated a few code sample so that they compile and run correctly.

Also removed the `article_outdated` tag, seems like the article is now pretty well up-to-date. I assume that tag is manually handled like this and not done by some automatic process?

### DIFF
--- a/tutorials/performance/using_multiple_threads.rst
+++ b/tutorials/performance/using_multiple_threads.rst
@@ -1,5 +1,3 @@
-:article_outdated: True
-
 .. _doc_using_multiple_threads:
 
 Using multiple threads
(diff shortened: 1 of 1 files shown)


akien-mga (approved):
LGTM. I don't remember what's the consensus regarding the use of type hints in examples but assuming it's recommended / allowed, this should be good to go.

mhilbrunner (approved):

mhilbrunner:
re: type hints, see [GDScript style guide](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html#static-typing). I think in this case it's okay to have them :)

> Also removed the article_outdated tag, seems like the article is now pretty well up-to-date. I assume that tag is manually handled like this and not done by some automatic process?

Yes exactly. Thanks for overhauling this page!

PR #7495, opened 2023-06-10
Update NavigationObstacle API doc

smix8:
Updates NavigationObstacle API doc.

Godot 4.1 only.

Update for when PR https://github.com/godotengine/godot/pull/78081 gets merged.

### DIFF
--- a/tutorials/navigation/navigation_using_navigationobstacles.rst
+++ b/tutorials/navigation/navigation_using_navigationobstacles.rst
@@ -61,7 +61,8 @@ For static use an array of ``vertices`` is required.
 .. tabs::
  .. code-tab:: gdscript GDScript
 
-    extends Node2D
+    # For 2D
+
     # create a new "obstacle" and place it on the default navigation map.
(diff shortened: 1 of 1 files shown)


mhilbrunner:
One consideration is that everything in code blocks is not translated for docs translations.

It may make sense to have "For 2D"/"For 3D" formatted above the code blocks as small headers or even just bolded text, what do you think?

Likewise, prefer keeping descriptions of what the code is doing next to the code instead of as comment _in_ the code, although thats a case-by-case (or comment-by-comment) decision, short comments are generally fine :)

smix8:
I would preferred "extends Node2D/3D" in the first line to signal that the code snippet is for 2D or 3D but there are users regularly do not understand. They think the code snippet is broken when they do a direct 1:1 copy of the code in their projects.

Not sure, most comments, when placed next to the code, become hard to read due to line length/wrap in my browser.

skyace65 (approved):

skyace65:
Thanks!

PR #7505, opened 2023-06-13
C#: Replace FuncRef with Callable in outdated example

raulsntos:
- `FuncRef` was removed in 4.0.
- The GDScript examples were updated in https://github.com/godotengine/godot-docs/pull/6542.

### DIFF
--- a/tutorials/best_practices/scene_organization.rst
+++ b/tutorials/best_practices/scene_organization.rst
@@ -106,10 +106,10 @@ initialize it:
      .. code-tab:: csharp
 
        // Parent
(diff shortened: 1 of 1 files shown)


raulsntos:
> Note: it might be better to use `ChildType.PropertyNames.FuncProperty` instead of `"FuncProperty"`, but it's probably not that big of a deal.

Agreed, but this whole page uses strings everywhere already. To be honest, I think the C# examples in this page should be rewritten to avoid strings and using `Get`/`Set`/`Call`/`Connect` when possible. But I will leave that for a future PR.

mhilbrunner (approved):

mhilbrunner:
Thanks! Merged.

PR #7513, opened 2023-06-14
Update multiple_resolutions.rst

atomikf:
Small typo correction.
There is a line "Set the base window width to 720 and window height to 1080."
It seems the '1080' should be changed to '1280'.


skyace65 (approved):

skyace65:
Thanks!

PR #7518, opened 2023-06-14
Add a compatibility policy and common questions to Release policy

Calinou:
- Add answers to common questions:
  - Which version should I use for a new project?
  - Should I upgrade my project to use new engine versions?

### DIFF
--- a/about/release_policy.rst
+++ b/about/release_policy.rst
@@ -135,6 +135,60 @@ provided for testing purposes only.
     See :ref:`doc_upgrading_to_godot_4` for instructions on migrating a project
     from Godot 3.x to 4.x.
 
+.. _doc_release_policy_which_version_should_i_use:
+
+Which version should I use for a new project?
+---------------------------------------------
+
+We recommend using Godot 4.x for new projects, as the Godot 4.x series will be
+supported long after 3.x stops receiving updates in the future. One caveat is
+that a lot of third-party documentation hasn't been updated for Godot 4.x yet.
+If you have to follow a tutorial designed for Godot 3.x, we recommend keeping
+:ref:`doc_upgrading_to_godot_4` open in a separate tab to check which methods
+have been renamed (if you get a script error while trying to use a specific node
+or method that was renamed in Godot 4.x).
+
+If your project requires a feature that is missing in 4.x (such as GLES2/WebGL
+1.0), you should use Godot 3.x for a new project instead.
+
+.. _doc_release_policy_should_i_upgrade_my_project:
+
+Should I upgrade my project to use new engine versions?
+-------------------------------------------------------
+
+.. note::
+
+    Upgrading software while working on a project is inherently risky, so
+    consider whether it's a good idea for your project before attempting an
+    upgrade. Also, make backups of your project or use version control to
+    prevent losing data in case the upgrade goes wrong.
+
+    That said, we do our best to keep minor and especially patch releases
+    compatible with existing projects.
+
+The general recommendation is to upgrade your project to follow new *patch*
+releases, such as upgrading from 4.0.2 to 4.0.3. This ensures you get bug fixes,
+security updates and platform support updates (which is especially important for
+mobile platforms). You also get continued support, as only the last patch
+release receives support on official community platforms.
+
+For *minor* releases, you should determine whether it's a good idea to upgrade
+on a case-by-case basis. We've made a lot of effort in making the upgrade
+process as seamless as possible, but some breaking changes may be present in
+minor releases, along with a greater risk of regressions. Some fixes included in
+minor releases may also change a class' expected behavior as required to fix
+some bugs. This is especially the case in classes marked as *experimental* in
+the documentation.
+
+*Major* releases bring a lot of new functionality, but they also remove
+previously existing functionality and may raise hardware requirements. They also
+require much more work to upgrade to compared to minor releases. As a result, we
+recommend sticking with the major release you've started your project with if
+you are happy with how your project currently works. For example, if your
+project was started with 3.5, we recommend upgrading to 3.5.2 and possibly 3.6
+in the future, but not to 4.0+, unless your project really needs the new
+features that come with 4.0+.
+
 .. _doc_release_policy_when_is_next_release_out:
 
 When is the next release out?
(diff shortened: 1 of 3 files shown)


about/release_policy.rst
```diff
@@ -132,6 +132,43 @@ provided for testing purposes only.
...
+minor releases, along with a greater risk of regressions. This is especially the
+case in classes marked as *experimental* in the documentation.
+
+*Major* releases bring a lot of new functionality, but they also remove
+previously existing functionality and may raise hardware requirements. They also
+require much more work to upgrade to compared to minor releases. As a result, we
+recommend sticking with the major release you've started your project with if
+you are happy with how your project currently works. That said, you may have to
+upgrade if your project *needs* a feature that won't be backported to a previous
+major release.
```
clayjohn, line 170: I think it would be helpful to add an example of what a major release is in case the reader is not familiar with semvar e.g. "For example, if your project was started with 3.5.0, we recommend upgrading to 3.5.2 and possible 3.6 in the future, but not to 4.0+, unless your project really needs the new features that come with 4.0+"

about/release_policy.rst
```diff
@@ -158,3 +195,55 @@ The 3.6 release is still planned and should be the last stable branch of Godot
...
+    changes are safe for a given release. The list is not exhaustive; it only
+    outlines the most common situations encountered during Godot's development.
+
+The following changes are acceptable in patch releases:
+
+- Fixing a bug that has no major negative impact on most projects, such as a
+  visual or physics bug. Godot's physics engine is not deterministic, so physics
+  bug fixes are not considered to break compatibility. If fixing a bug has a
+  negative impact that could impact a lot of projects, it should be made
+  optional (e.g. using a project setting or separate method).
```
clayjohn, line 214: On my first read I thought this was referring to bugs which have no major negative impact. But it seems you are actually referring to the PR having "no major negative impact"

It might help to reword it a bit

```suggestion
- Fixing a bug in a way that has no major negative impact on most projects, such as a
  visual or physics bug. Godot's physics engine is not deterministic, so physics
  bug fixes are not considered to break compatibility. If fixing a bug has a
  negative impact that could impact a lot of projects, it should be made
  optional (e.g. using a project setting or separate method).
```

clayjohn (commented):
Looks great! I left a couple of suggestions

about/release_policy.rst
```diff
@@ -132,6 +132,43 @@ provided for testing purposes only.
     See :ref:`doc_upgrading_to_godot_4` for instructions on migrating a project
     from Godot 3.x to 4.x.
 
+Which version should I use for a new project?
+---------------------------------------------
+
+We recommend using Godot 4.x for new projects, as the Godot 4.x series will be
+supported long after 3.x stops receiving support in the future. One caveat is
```
YuriSizov, line 139:
```suggestion
supported long after 3.x stops receiving updates in the future. One caveat is
```

about/release_policy.rst
```diff
@@ -132,6 +132,43 @@ provided for testing purposes only.
...
+have been renamed (if you get a script error while trying to use a specific node
+or method that was renamed in Godot 4.x).
+
+If your project requires a feature that is missing in 4.x (such as GLES2/WebGL
+1.0), you should use Godot 3.x for a new project instead.
+
+Should I upgrade my project to use new engine versions?
+-------------------------------------------------------
+
+The general recommendation is to upgrade your project to follow new *patch*
```
YuriSizov, line 152: I would suggest that we start with a word of caution, that upgrades are inherently risky and it's better to evaluate the changes yourself. Then follow up with a notion that we do our best to keep minor and especially patch releases compatible. Then the rest of what you wrote.

about/release_policy.rst
```diff
@@ -158,3 +195,55 @@ The 3.6 release is still planned and should be the last stable branch of Godot
...
+
+- Fixing a bug that has no major negative impact on most projects, such as a
+  visual or physics bug. Godot's physics engine is not deterministic, so physics
+  bug fixes are not considered to break compatibility. If fixing a bug has a
+  negative impact that could impact a lot of projects, it should be made
+  optional (e.g. using a project setting or separate method).
+- Adding a new optional parameter to a method.
+- Small-scale editor usability tweaks.
+
+The following changes are acceptable in minor releases, but not patch releases:
```
YuriSizov, line 218: I would add here bugfixes which significantly change the behavior or the output, with the aim to meet user expectations better. Compared to patch releases where we may favor keeping a buggy behavior so we don't break existing projects which likely rely on the bug or implement a workaround. (E.g. https://github.com/godotengine/godot/pull/78727)

about/release_policy.rst
```diff
@@ -158,3 +195,55 @@ The 3.6 release is still planned and should be the last stable branch of Godot
...
+  deprecated flag to its class reference, which will show up in the editor. When
+  a method is marked as deprecated, it's slated to be removed in the next
+  *major* release.
+- Changes that affect the default project theme's visuals.
+- Performance optimizations that result in visual changes.
+
+The following changes are considered **compatibility-breaking** and can only be
+performed in a new major release:
+
+- Renaming or removing a method, member variable or a class.
```
YuriSizov, line 233:
```suggestion
- Renaming or removing a method, member variable, or class.
```

about/release_policy.rst
```diff
@@ -158,3 +195,55 @@ The 3.6 release is still planned and should be the last stable branch of Godot
...
+- Modifying a node's inheritance tree by making it inherit from a different class.
```
YuriSizov, line 234:
```suggestion
- Modifying node's inheritance tree by making it inherit from a different class.
```

about/release_policy.rst
```diff
@@ -158,3 +195,55 @@ The 3.6 release is still planned and should be the last stable branch of Godot
...
+- Changing a project setting's default value in a way that affects existing
```
YuriSizov, line 235:
```suggestion
- Changing the default value of a project setting in a way that affects existing
```

about/release_policy.rst
```diff
@@ -158,3 +195,55 @@ The 3.6 release is still planned and should be the last stable branch of Godot
...
+What are the criteria for compatibility across engine versions?
+---------------------------------------------------------------
+
+.. note::
+
+    This section is intended to be used by contributors to determine which
+    changes are safe for a given release. The list is not exhaustive; it only
+    outlines the most common situations encountered during Godot's development.
+
+The following changes are acceptable in patch releases:
```
YuriSizov, line 208: Worth mentioning in some form that we tend to be more conservative with the fixes we allow in each subsequent patch release. So e.g. 4.0.1 may receive more impactful fixes than 4.0.4 would.

about/release_policy.rst
Calinou: Isn't "node" a countable noun?

Calinou:
I've pushed suggestions and also mentioned the new questions in the FAQ.

about/release_policy.rst
YuriSizov: It is. I'm uncertain about the rules of using articles around possessive nouns, but after a bit of searching I think you've got it correct originally.

mhilbrunner (approved):

mhilbrunner:
Gave it a read and this seems good to go now. @clayjohn @YuriSizov As you you've made suggestions, feel free to five it a quick glance and approve (or complain :P), otherwise I'm going to merge this soon.

YuriSizov (approved):
Merge away :)

clayjohn (approved):

mhilbrunner:
Thanks for writing this up! Merged.

PR #7519, opened 2023-06-14
tell codespell to ignore the locale code page

skyace65:
EDIT: Originally this was just going to add the latest things to break CI, but Akien suggested it would just be simpler to ignore the page in the contributors chat. And given that almost the entire ignore lines file was from that page I'm inclined to agree, so this PR makes codespell ignore that entire page now.


_tools/codespell-ignore-lines.txt
```diff
@@ -17,4 +17,5 @@
 | MM, BU       | Myanmar                                    |
 | RO           | Romania                                    |
 | VN           | Viet Nam                                   |
+| vai           | Vai                           |
```
Piralein, line 20:
```suggestion
| Vaii        | Vai                         |
```
this is the line in current master.

skyace65:
I Changed this PR to ignore the entire locale codes file. Updated the original description.

akien-mga (approved):

_tools/codespell-ignore-lines.txt
```diff
@@ -1,20 +1 @@
...
-| te            | Telugu                        |
-| tem           | Timne                         |
-| Beng        | Bengali                     |
-| Maka        | Makasar                     |
-| BA           | Bosnia and Herzegovina                     |
-| FO           | Faroe Islands                              |
-| MM, BU       | Myanmar                                    |
-| RO           | Romania                                    |
-| VN           | Viet Nam                                   |
 .. image:: img/expres.png
```
akien-mga, line 1: We could probably rename that file to something not truncated and get rid of the ignore lines file altogether, but maybe we'll need it in the future for some other false positives.

akien-mga:
Thanks!

PR #7520, opened 2023-06-15
Add support for user-provided comments with Giscus

Calinou:
In a fashion similar to <https://php.net>'s comments, this allows users to leave comments on pages that don't have `:allow_comments: False` somewhere in the page's source. Both manual and class reference pages can receive comments. Index pages cannot have comments, as discussion should occur on "leaf" pages.

GitHub Discussions is used as a backend on the same repository.[^1] **This means that Discussions *must* be enabled on godotengine/godot-docs before this commit is merged to `master`.** Users can choose to use the "Custom" watch mode if they don't want to get notifications for discussion updates, but still get notifications for issue and pull request updates.

[^1]: We can choose to use a dedicated repository containing just discussions if we wish to do so.

User comments are intended to be used for the following purposes:

- Add a clarification or correct something in the documentation, without having to open a pull request. Contributors are encouraged to take a look at discussions from time to time, and see if there's information worth incorporating in the pages themselves. Don't forget to reply to the comment when doing so :slightly_smiling_face:
- Mention a workaround for a common issue.
- Link to useful third-party resources that are relevant to the current page, such as tutorials or add-ons.

User comments should **not** be used for technical support. Other community platforms should be used for that.

More details can be found in the [User-contributed notes policy](https://github.com/Calinou/godot-docs/discussions/46) I've written.

Page-to-discussion matching is done using the `pagename` Sphinx variable, which is independent of the Godot version and documentation language. Being independent of the Godot version allows keeping old comments when the Godot version changes, while also allowing users from `/stable` and `/4.0` to "see" each other in discussions.

See https://giscus.app for more information.

- See https://github.com/godotengine/godot-proposals/issues/481.

## TODO

- [x] Add `:allow_comments: False` on all index/listing pages.
- [x] Modify `makerst.py` in the main Godot repository to write `:allow_comments: False` at the top of `classes/index.rst`, as was done manually here. https://github.com/godotengine/godot/pull/85006
- [x] Set up Discussions on ~~godotengine/godot-docs~~ [godotengine/godot-docs-user-notes](https://github.com/godotengine/godot-docs-user-notes) and use the Giscus setup form again, so that the discussion category unique ID is updated. [Use the following options, with "Calinou/godot-docs" replaced by "godotengine/godot-docs".](https://github.com/godotengine/godot-docs/assets/180032/9e9412fb-3760-4dbe-9209-00d957b9e018)
  - Rename the default Announcements category to *Page comments* while keeping its type on Announcements. Change its emoji to :left_speech_bubble:. Remove all other categories, so that users cannot create discussions manually (the Giscus app will do it for them).
  - Create a second category called *Rules* with the Announcements type and :straight_ruler: emoji.
  - Create a post called "User-contributed notes policy" in the *Rules* category with the contents from https://github.com/Calinou/godot-docs/discussions/46, then pin the post globally (not in a single category) and lock it. Make sure reactions are **not** allowed when locking the post. Use the Markdown code below.
- [x] Replace URL to the *User-contributed notes* policy in `layout.html` to point to godot-docs' version, rather than the Calinou fork.

## Preview

https://github.com/godotengine/godot-docs-user-notes/discussions

### Example at the bottom of the *3D text* page

*The theme follows the system theme, which means it'll adapt to the documentation's theme automatically.*

image

### DIFF
--- a/_templates/layout.html
+++ b/_templates/layout.html
@@ -66,6 +66,35 @@
   {% endif %}
 
   {% block body %}{% endblock %}
+
+{% if (not meta or meta.get('allow_comments') != 'False') and godot_show_article_comments %}
+<hr>
+<h2>User-contributed notes</h2>
+<p>
+  <em>Please read the <a href="https://github.com/godotengine/godot-docs-user-notes/discussions/1">User-contributed notes policy</a> before submitting a comment.</em>
+</p>
+{# Use https://giscus.app/ to regenerate the script tag if needed. #}
+{# data-term is set to be language-independent and version-independent, so that comments can be centralized for each page. #}
+{# This increases the likelihood that users will encounter comments on less frequently visited pages. #}
+<script src="https://giscus.app/client.js"
+  data-repo="godotengine/godot-docs-user-notes"
+  data-repo-id="R_kgDOKuNx0w"
+  data-category="User-contributed notes"
+  data-category-id="DIC_kwDOKuNx084CbANb"
+  data-mapping="specific"
+  data-term="{{ pagename }}"
+  data-strict="1"
+  data-reactions-enabled="0"
+  data-emit-metadata="0"
+  data-input-position="bottom"
+  data-theme="preferred_color_scheme"
+  data-lang="en"
+  data-loading="lazy"
+  crossorigin="anonymous"
+  async></script>
+<br>
+{% endif %}
+
 {%- if self.comments()|trim %}
   <div class="articleComments">
     {%- block comments %}{% endblock %}
--- a/about/complying_with_licenses.rst
+++ b/about/complying_with_licenses.rst
@@ -1,3 +1,5 @@
+:allow_comments: False
+
 .. _doc_complying_with_licenses:
 
 Complying with licenses
--- a/about/docs_changelog.rst
+++ b/about/docs_changelog.rst
@@ -1,3 +1,5 @@
+:allow_comments: False
+
 .. _doc_docs_changelog:
 
 Documentation changelog
--- a/about/faq.rst
+++ b/about/faq.rst
@@ -1,3 +1,5 @@
+:allow_comments: False
+
 .. meta::
     :keywords: FAQ
 
--- a/about/introduction.rst
+++ b/about/introduction.rst
@@ -1,3 +1,5 @@
+:allow_comments: False
+
 .. _doc_about_intro:
 
 Introduction
--- a/about/list_of_features.rst
+++ b/about/list_of_features.rst
@@ -1,3 +1,5 @@
+:allow_comments: False
+
 .. _doc_list_of_features:
 
 List of features
--- a/about/release_policy.rst
+++ b/about/release_policy.rst
@@ -1,3 +1,5 @@
+:allow_comments: False
+
 .. _doc_release_policy:
 
 Godot release policy
--- a/community/asset_library/index.rst
+++ b/community/asset_library/index.rst
@@ -1,3 +1,5 @@
+:allow_comments: False
+
 Asset Library
 =============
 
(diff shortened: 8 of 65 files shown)


mhilbrunner:
The User-contributed notes policy seems good. I'd maybe change the bit about licensing to maybe say in the docs *and* the Godot project/engine, and try to link to existing license pages for both instead of mentioning any specific license, but not sure.

We should also consider making sure this does not work in local dev builds and in forks by default unless you toggle something, otherwise if another random docs website mirror appears, I could see that lead to confusion.

Also need to check how reporting/moderation would work.

Besides those small need-to-check TODOs, this looks good to me and it has my stamp of approval.

Calinou:
> I'd maybe change the bit about licensing to maybe say in the docs and the Godot project/engine, and try to link to existing license pages for both instead of mentioning any specific license, but not sure.

We don't have the documentation license listed on [godotengine.org/license](https://godotengine.org/license), but I've changed the wording to also mention the engine itself and its license.

> We should also consider making sure this does not work in local dev builds and in forks by default unless you toggle something, otherwise if another random docs website mirror appears, I could see that lead to confusion. Same for translations - if a page in a non-english translation has no comments yet, it may be too easy to just leave a comment in whatever native language, becoming confusing for other users.

I've changed it to disable comments integration if not building on Read the Docs, and if the language isn't English.

> I'm also less sure whether we should bother adding this to 3.x docs at all. Everything worth discussing in 3.x has probably already been discussed ad nauseam elsewhere, and we don't get many 3.x docs PRs these days, so I assume most things at this point in time are mostly complete and in a good shape, mature as 3.x is. So I could see the confusion between comments between 2 major versions outweigh the benefit.

I agree with making it 4.x-only.

_templates/layout.html
```diff
@@ -67,6 +67,35 @@
...
+<hr>
+<h2>User-contributed notes</h2>
+<p>
+  <em>Please read the <a href="https://github.com/Calinou/godot-docs/discussions/46">User-contributed notes policy</a> before submitting a comment.</em>
+</p>
+{# Use https://giscus.app/ to regenerate the script tag if needed. #}
+{# data-term is set to be language-independent and version-independent, so that comments can be centralized for each page. #}
+{# This increases the likelihood that users will encounter comments on less frequently visited pages. #}
+<script src="https://giscus.app/client.js"
+  data-repo="Calinou/godot-docs"
```
mhilbrunner, line 81:
```suggestion
  data-repo="godotengine/godot-docs"
```
Something worth investigating for future followups would be to see if we could automatically attach a small markup indicating the branch/version the comment was left with, i.e. adding "Godot 4.1" in a subtle font next to the comment's username for a comment left via a 4.1 branch page. (`stable` would of course complicate this)

mhilbrunner:
> I'm not sure what would be the best way to disable it in forks automatically, preferably at build-time in conf.py.

Yeah, a build var similar to `godot_show_article_status` in `conf.py`, but default to `false` and overridable via an environment var would be easiest.

_templates/layout.html
Calinou: I need to regenerate the HTML tag using https://giscus.app (so that `data-repo-id` is also changed), but I need the Discussions feature on godotengine/godot-docs to be enabled first:

image

I don't have permission to do this.

Calinou:
> Yeah, a build var similar to godot_show_article_status in conf.py, but default to false and overridable via an environment var would be easiest (we can set custom env vars in the RTD dashboard).

I've added `show_article_comments` which defaults to `on_rtd and not is_i18n` in `conf.py`. This means it's enabled automatically when building on RTD, and only for English.

_templates/layout.html
```diff
@@ -1,4 +1,4 @@
-{% extends "!layout.html" -%}
+<{% extends "!layout.html" -%}
```
YuriSizov, line 1: Is this a typo?

classes/index.rst
YuriSizov, line 1: This is a generated file, there is no point editing it in this PR. We should simply sync the class reference (I think we've merged your main repo PR already).

mhilbrunner:
Thanks for working on this! Let's see how this works out.

PR #7523, opened 2023-06-15
Recommend designing lighting in the Godot editor in Importing 3D scenes

Calinou:
This is a general recommendation that gets asked about regularly on community platforms.

This also mentions that newer Blender versions will result in better `.blend` import, especially 3.5 and later.

See https://www.reddit.com/r/godot/comments/149rpqe/exporting_lights_with_model_best_practice/.


SocialGoo (approved):

skyace65:
Thanks!

PR #7525, opened 2023-06-15
C#: Document new version defines and remove mentions of old defines

RedworkDE:
The version defines were added in https://github.com/godotengine/godot/pull/78249 and https://github.com/godotengine/godot/pull/78270

The feature defines haven't existed in a long time, and even if https://github.com/godotengine/godot/pull/53920 finally gets anywhere, they will be in a different format, The architecture defines no longer exist either and there doesn't seem to be any plan / demand to reintroduce them.

For the platforms that have multiple defines we should probably just pick one to mention, but I am not sure which of them is preferred.

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_features.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_features.rst
@@ -141,6 +141,20 @@ Or you can detect which engine your code is in, useful for making cross-engine l
     #endif
         }
 
+Or you can write scripts that target multiple Godot versions and can take
+advantage that are only available on some of those versions:
+
+.. code-block:: csharp
+
+        public void UseCoolFeature()
+        {
+    #if GODOT4_3_OR_GREATER || GODOT4_2_2_OR_GREATER
+            // Use CoolFeature, that was added to Godot in 4.3 and cherry-picked into 4.2.2, here.
+    #else
+            // Use a workaround for the absence of CoolFeature here.
+    #endif
+        }
+
 Full list of defines
 ~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


raulsntos:
> For the platforms that have multiple defines we should probably just pick one to mention, but I am not sure which of them is preferred.

I wonder why some platforms have multiple defines, probably a leftover from an older version.

For example, the Web platform used to be called JavaScript[^1], and since Godot adds OS features for each platform it's likely that C# was just emulating this by adding the `GODOT_JAVASCRIPT` define, then when it was renamed we may have added `GODOT_WEB` but didn't remove the old one.

Also, at some point there was a `html5` OS feature, but that doesn't seem to exist anymore either.

I guess the situation with (`GODOT_OSX` and `GODOT_MACOS`) and (`GODOT_IPHONE` and `GODOT_IOS`) is similar. These platforms were also renamed (OSX to MacOS and iPhone to iOS)[^2].

We could probably remove the duplicate defines, although I guess it breaks compatibility. But I agree that we should only mention the could mention only one in the documentation so I'd pick the ones that match the platform name.

[^1]: https://github.com/godotengine/godot/commit/d20b32186fc192f5e527a1211291b0cb293f4e66
[^2]: https://github.com/godotengine/godot/commit/8823eae328547991def3b13ee2919291d29a278b

RedworkDE:
Ok, I removed the old apple platform defines and used `GODOT_WEB` as the define for the web platform, and let it do double duty as both the platform type and platform define.

raulsntos (approved):
LGTM

mhilbrunner (approved):

Issue #6779, opened 2023-02-14
Change indentation from spaces to tabs for codeblocks


Repiteo:
**Your Godot version:**
4.0-rc1

**Issue description:**
The GDScript style guide specifies use of "Tabs instead of spaces for indentation". However, the entirety of the documentation's code examples format with spaces instead of tabs. Handling the codeblocks with tabs instead like [the main site appears capable of](https://godotengine.org/article/introducing-gd-extensions/) would remedy this inconsistency (though it'd need css adjustments to indent 4 blocks instead of 8)

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html
(Applicable to entirety of documentation where codeblocks are present)

YuriSizov:
I think we looked into this a few years ago, and there were issues with how RTD generates HTML. But worth checking again.

Calinou:
We could convert sets of 4 spaces to tabs at load-time using JavaScript, but this may be slow on large pages.

ronyeh:
What about changing the editor? If we paste a large block of code, perhaps we can ask the developer if they would like the indentation converted from spaces => tabs?

ronyeh:
Or when someone clicks the floating COPY button, we run some JS to convert spaces to tabs! It's not a perfect solution, but it'll handle a common case, where someone clicks the copy button on the docs.

ronyeh:
Here's a hack I threw together:

```js
function copyTextToClipboard(text) {
  var textArea = document.createElement('textarea');
  textArea.value = text;
  document.body.appendChild(textArea);
  textArea.select();
  document.execCommand('copy');
  document.body.removeChild(textArea);
}

function convertSpacesToTabs(inputString) {

  const fourSpaces = '    ';
  const lines = inputString.split('\n');
  const convertedLines = [];

  for (let line of lines) {
    let numTabs = 0;

    while (line.startsWith(fourSpaces)) {
      line = line.substring(4);
      numTabs++;
    }

    const tabs = '\t'.repeat(numTabs);

    const convertedLine = tabs + line;
    convertedLines.push(convertedLine);
  }

  const output = convertedLines.join('\n');
  copyTextToClipboard(output);
  return output;
}

document.querySelectorAll(".copybtn").forEach((button) => {
  button.addEventListener('click', (event) => {
    setTimeout(()=>{
      convertSpacesToTabs(event.target.parentNode.innerText);
[...]

PR #7529, opened 2023-06-16
Convert spaces to tabs via new extension

Repiteo:
Closes godotengine/godot-docs#6779

### DIFF
--- a/_extensions/spaces_to_tabs.py
+++ b/_extensions/spaces_to_tabs.py
@@ -1,0 +1,39 @@
+# -*- coding: utf-8 -*-
+"""
+    spaces_to_tabs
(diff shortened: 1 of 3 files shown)


Repiteo:
There's probably a better way of doing this, as ideally there'd be some callback right after an html file is made. But Sphinx documentation is a bit daunting to put it lightly, so this method feels like the next-best thing: parsing through all the html files after they're built. It only targets the section of the html file that would have codeblocks & only applies them if preceeded with a linebreak

mhilbrunner:
Hey! Thanks for working on this, this has been on my to do list for a long while. I'll try to test it ASAP :)

Repiteo:
Closing in favor of #8153

PR #7536, opened 2023-06-18
Updates to visual_shaders.rst to show new way to create visual shaders

rametta:
The explanation for creating visual shaders seems to be outdated. There is no more option for creating a visual shader right below a normal shader, now you must pick visual shader by picking the correct "type" from the create shader dialog. This PR updates the docs and images to reflect this new workflow


skyace65:
Two issues with those images. One they're way too big, I'm assuming you took screenshots on a 4K screen, those need to be scaled down. Additionally, we're changing over from png to webp. You can find information on doing both of those [here](https://docs.godotengine.org/en/stable/contributing/documentation/docs_image_guidelines.html).

paddy-exe:
Great work @rametta! Could you also please squash the commits into one?

skyace65:
What did you use when scaling the images down and converting to webp? There's a noticeable amount of compression artifacts in those images.

rametta:
I used Sqoosh, that was suggested in the docs

rametta:
I used Squoosh. I think I downsized too far before. I've redid it, and re-pushed. Should be good now.

skyace65:
Sorry to ask for another change but could you do the images again and use the lossless compression for Squoosh. There's still very slight artifacts in the images. I talked with another maintainer and going forward we'll probably ask for lossless compression unless it's a over 300KB file. I'm making a PR for the image guide now so that's clearer in the future.

rametta:
@skyace65 No worries. Re-pushed with lossless 🙂

skyace65:
Thanks for the fixes! And congrats on your first merged PR!

PR #7538, opened 2023-06-19
Change outdated and confuse statements from C# performance

AndreSacilotto:
The _preliminary benchmarks_ were done with a tool that was for version 3.X and is no longer in active development. So it doesn't make sense to keep outdated information on a tutorial page.

And even if it made sense, stating that one language can be 4x faster than the other in specific cases, and saying nothing about what those cases are doesn't help much.

The reason for the edit is the frustration of having multiple people come over time on the forum and discord thinking that by simply using C# they would be writing a more performatic game.

The new statements given are based in the community responses given for people that came for the:
> roughly ~4x speed boost

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_basics.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_basics.rst
@@ -326,11 +326,16 @@ You can read more about this error on the `C# language reference <https://learn.
 Performance of C# in Godot
 --------------------------
 
-According to some preliminary `benchmarks <https://github.com/cart/godot3-bunnymark>`_,
-the performance of C# in Godot — while generally in the same order of magnitude
-— is roughly **~4×** that of GDScript in some naive cases. C++ is still
-a little faster; the specifics are going to vary according to your use case.
-GDScript is likely fast enough for most general scripting workloads.
+Comparing GDScript to C# is tricky since C# can be faster in some specific cases; generally,
+those cases are when not iterating with the engine. So using C# can bring you performance benefits,
+but it is not guaranteed.
+
+In theory, if C++ (GDNative) is used right it will always be faster than C#, but for most projects,
+the difference is negligible, so use C# or C++ if you like them and know what you are doing; otherwise,
+stick with GDScript, which is likely fast enough for most general scripting workloads.
+
+The project don't need to stick to a single language; it can use multiple of them at the same time.
+Be aware that doing so will bring its own pitfalls.
 
 Most properties of Godot C# objects that are based on ``GodotObject``
 (e.g. any ``Node`` like ``Control`` or ``Node3D`` like ``Camera3D``) require native (interop) calls as they talk to


tutorials/scripting/c_sharp/c_sharp_basics.rst
```diff
@@ -326,11 +326,16 @@ You can read more about this error on the `C# language reference <https://learn.
...
+Comparing GDScript to C# is tricky since C# can be faster in some specific cases; generally,
+those cases are when not iterating with the engine. So using C# can bring you performance benefits,
+but it is not guaranteed.
+
+In theory, if C++ (GDNative) is used right it will always be faster than C#, but for most projects,
+the difference is negligible, so use C# or C++ if you like them and know what you are doing; otherwise,
+stick with GDScript, which is likely fast enough for most general scripting workloads.
+
+The project don't need to stick to a single language; it can use multiple of them at the same time.
+Be aware that doing so will bring its own pitfalls.
```
Calinou, line 338:
```suggestion
Comparing GDScript performance to C# is tricky, since C# can be faster in some specific cases. However, C#'s performance can also be brought down by garbage collection which occurs at random and unpredictable moments. This can result in difficult-to-diagnose stuttering issues in complex projects, and is not exclusive to Godot. As a result, C# may result in a performance benefit for some number crunching scenarios, but it is not guaranteed.

For most projects that don't need to perform intensive computations, the difference is negligible. Use C# if you like it and are familiar with the language; otherwise,
stick with GDScript, which is likely fast enough for most general scripting workloads.
Also, remember that using a faster scripting language will not help improve performance
in GPU-bottlenecked scenarios, nor in scenarios where the bottleneck is within the engine (such as physics or navigation).

Your project doesn't need to stick to a single language. It can use multiple of them at the same time.
Be aware that doing so will bring its own pitfalls, as covered in :ref:`doc_cross_language_scripting`.
```

AndreSacilotto: I don't see how C++ is out of scope, the topic is "Performance of C# in Godot", GDNative is integrated in the engine so...

Also I fear that don't adding C++ in the equation can make people come asking "Is C++ faster than C#? The docs don't say anything about it".

I agree that C++ comes out of nowhere, but I couldn't find a better way of writing "C++ _can_ be faster than C#, but I doubt you would need that".

mhilbrunner:
I'd probably just remove the specific 4x figure, as that seems indeed a pretty specific assumption, if that helps.

raulsntos:
I agree that just removing the 4x figure is good. I'm not sure we want to get into comparing C# with other languages in the documentation, I understand it's a frequent question but it feels out of place in this documentation page.

I think it's weird to mention other languages such as GDScript or C++ in the C# documentation. I feel like this section of the documentation should focus more on comparing _C#_ to _C# in Godot_. That is to say, the differences between the Godot C# API and the regular C# API outside of Godot. For example, the cost of interop and marshalling that is unique to the Godot C# bindings.

tetrapod00:
In the interest of resolving stalled issues, I have opened a new PR which *only* removes the 4x faster claim: https://github.com/godotengine/godot-docs/pull/10286.

For the record I quite like the revised comparison written by Calinou and AndreSacilotto, but it does feel like too long and too much of a general performance comparison to exist in this section of the docs. I think somewhere like the FAQ might be a good place (and this question *is* asked frequently).

tetrapod00:
Closing this in favor of https://github.com/godotengine/godot-docs/pull/10286 and https://github.com/godotengine/godot-docs/pull/10287 (which builds off of this work and credits you as a co-author on the commit). Thanks for your contribution!

PR #7540, opened 2023-06-21
Fix example in JSON class.

BWPanda:
I tried the first block of example code from https://docs.godotengine.org/en/stable/classes/class_json.html and got an error:

> Identifier "json" not declared in the current scope.

I tried changing it to `var error = JSON.parse(json_string)` but then got this error:

> Cannot call non-static function "parse()" on the class "JSON" directly. Make an instance instead.

So I think it needs this line added: `var json = JSON.new()`

### DIFF
--- a/classes/class_json.rst
+++ b/classes/class_json.rst
@@ -34,6 +34,7 @@ The **JSON** enables all data types to be converted to and from a JSON string. T
     # Save data
     # ...
     # Retrieve data
+    var json = JSON.new()
     var error = json.parse(json_string)
     if error == OK:
         var data_received = json.data


BWPanda:
I just found the 'load' example here: https://docs.godotengine.org/en/stable/tutorials/io/saving_games.html#saving-and-reading-data

This verifies the above:

```
# Creates the helper class to interact with JSON
var json = JSON.new()
```

dalexeev:
https://github.com/godotengine/godot-docs/blob/774c9a82db6c488a6042c16f2672eb3982def1ac/classes/class_json.rst?plain=1#L3-L6

Calinou:

BWPanda:
Oops, sorry. Will make a PR to the proper repo then.

I'm using the latest version (4).

PR #7542, opened 2023-06-21
Wrap table columns on mobile

### LINKED ISSUE #7504: Table on mobile squished to one character wide
**Your Godot version:** 4.1 beta

**Issue description:** Some tables on mobile appear messed up.

image

**URL to the documentation page:** https://docs.godotengine.org/en/stable/classes/class_tileset.html

Piralein:
closes #7504.

wrap

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -791,6 +791,32 @@ html.writer-html5 .rst-content .wy-table-responsive > table td > p code.literal 
     padding: 2px 5px;
 }
 
+/* Wrap table columns on mobile */
+@media screen and (max-width: 1024px) {
+    .rst-content table.docutils {
+        border-top-width: 1px;
+    }
+
+    .rst-content table.docutils tr {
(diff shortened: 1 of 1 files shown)


mhilbrunner:
cc @yurchor @Calinou CSS magic :) This seems to be a good fix, but maybe you want to review in case there is something I'm missing or you know a better way

YuriSizov:
A better way would be to have a header before each corresponding cell. Without that some tables won't become more usable than they currently are. But at least that makes tables fit the screen nicely, and somewhat improves some of the tables, especially tiny ones.

The problem with the proper solution is that it can't be done with just CSS. We'd need to change the layout somehow. Ideally, we'd want two separate layouts, with one visible at desktop resolution and the other at mobile resolution. But in the RST format we'd still want one table so maintenance doesn't suffer. This means the change must come from the generator, and I'm not sure we have that kind of control.

Issue #7537, opened 2023-06-19
Missing line in tutorial about navigation with different actors


1998marcom:
**Godot version:** 4.0.2

**Issue description:** Supposedly missing line in tutorial about navigation with different actors (and typo)

**URL to the documentation page:** [docs.godotengine.org/en/4.0/tutorials/navigation/navigation_different_actor_types.html](https://docs.godotengine.org/en/4.0/tutorials/navigation/navigation_different_actor_types.html)

The page has some good code for bootstrapping one's own code, but I had issues with regions not merging until I manually activated the navigation maps with `NavigationServer3D.map_set_active(map_rid, true)`. If the maps are inactive by default and require manual activation, and this is their intended behaviour, I would add the previous line to the tutorial.

Also, the penultimate line should be:
`var path_small_agent = NavigationServer3D.map_get_path(navigation_map_small, start_pos, end_pos, true)`
rather than:
`var path_small_agent = NavigationServer3D.map_get_path(navigation_mesh_small_size, start_pos, end_pos, true)`
I believe it is just a clear typo.

PR #7557, opened 2023-06-25
Update navigation different actor types for new SourceGeometryData

smix8:
Updates navigation different actor types for new SourceGeometryData.

Fixes https://github.com/godotengine/godot-docs/issues/7537.

Godot 4.1 only.


skyace65 (approved):

skyace65:
Thanks!

GeorgeS2019:
@smix8 
Great job. 
Is there a planned demo OR closest demo on how to deal with the different actor types

smix8:
No plans for a demo. As seen in the script example it is about querying paths from different maps for each distinct agent size. I think this is for the most part self-explanatory and not very different from existing demos that query navigation paths from the default map.

PR #7560, opened 2023-06-26
Fix OBJECT_POSITION description to match engine behavior.

cridenour:
While doing some custom SDF work in a fog shader, I noticed the documentation for `OBJECT_POSITION` did not match engine behavior.

You can see an example in the engine's fog shader where they need both `WORLD_POSITION` and `OBJECT_POSITION` to get a position in local space here: https://github.com/godotengine/godot/blob/3710f069293f2fde8afe33fea898c4b36fa5e943/scene/resources/fog_material.cpp#L160


clayjohn (approved):

mhilbrunner (approved):

mhilbrunner:
Thanks for taking the time to contribute and fix this for all Godot users! Merged. 🎉

PR #7566, opened 2023-06-26

dahliakyu:
change _on_body_entered(body) to _on_body_entered(_body) to avoid warnings

### DIFF
--- a/getting_started/first_2d_game/03.coding_the_player.rst
+++ b/getting_started/first_2d_game/03.coding_the_player.rst
@@ -489,7 +489,7 @@ this code to the function:
 .. tabs::
  .. code-tab:: gdscript GDScript
 
-    func _on_body_entered(body):
+    func _on_body_entered(_body):
         hide() # Player disappears after being hit.
         hit.emit()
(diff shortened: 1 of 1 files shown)


skyace65:
Personally I think it would make more sense to just explain the warnings as part of the tutorial.

NathanLovato:
Definitely agreed with @skyace65, if users get thrown off by warnings, the tutorial should instead explain that warning are not errors, but rather hints to help improve the code, but that in this case, you can safely ignore it.

dahliakyu:
> Definitely agreed with @skyace65, if users get thrown off by warnings, the tutorial should instead explain that warning are not errors, but rather hints to help improve the code, but that in this case, you can safely ignore it.
> 
> Or if it's decided to show this convention, using a leading underscore, it should be explained briefly but clearly how it works and where it comes from.

Would definitely agree as well, just that the later case was adopted in the [completed project](https://github.com/godotengine/godot-demo-projects/tree/master/2d/dodge_the_creeps) here. There are also other differences (some naming conventions) between the completed project codes and the tutorial itself, I do think it would be nice to have some brief explanations in the tutorial.

PR #7571, opened 2023-06-28
Update OpenXR passthrough documentation for transparent background

BastiaanOlij:
Once upon a time we had automated some of the background settings but that was changed to give the user more control. The documentation never reflected this.


skyace65 (approved):

skyace65:
Thanks!

PR #7574, opened 2023-06-28
Update the information about upcoming and currently supported releases

YuriSizov:
Updated 4.1 and added 4.2 with their current estimates, also adjusted the estimate for 3.6. Changed some wording too, as minor versions only receive patch releases which do not add new usability features, generally.

### DIFF
--- a/about/release_policy.rst
+++ b/about/release_policy.rst
@@ -77,41 +77,44 @@ If you experience an issue using an older patch release, please upgrade to the
 latest patch release of that series and test again before reporting an issue
 on GitHub.
 
-+-------------+----------------------+--------------------------------------------------------------------------+
-| **Version** | **Release date**     | **Support level**                                                        |
-+-------------+----------------------+--------------------------------------------------------------------------+
-| Godot 4.1   | Q2-Q3 2023 (estimate)| |unstable| *Development.* Receives new features as well as bug fixes     |
-| (`master`)  |                      | while under development.                                                 |
-+-------------+----------------------+--------------------------------------------------------------------------+
-| Godot 4.0   | March 2023           | |supported| Receives fixes for bugs, security and platform support       |
-|             |                      | issues, as well as backwards-compatible usability enhancements.          |
-+-------------+----------------------+--------------------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


Calinou (approved):

Piralein (approved):

about/release_policy.rst
```diff
@@ -77,41 +77,44 @@ If you experience an issue using an older patch release, please upgrade to the
...
+| (`3.x`, LTS) |                      | improvements, as well as bug fixes, while under development.             |
++--------------+----------------------+--------------------------------------------------------------------------+
+| Godot 3.5    | August 2022          | |supported| Receives fixes for bugs and security issues, as well as      |
+|              |                      | patches that enable platform support.                                    |
++--------------+----------------------+--------------------------------------------------------------------------+
+| Godot 3.4    | November 2021        | |eol| No longer supported, as fully superseded by the compatible 3.5     |
+|              |                      | release (last update: 3.4.5).                                            |
++--------------+----------------------+--------------------------------------------------------------------------+
+| Godot 3.3    | April 2021           | |eol| No longer supported, as fully superseded by the compatible 3.4     |
+|              |                      | release (last update: 3.3.4).                                            |
```
akien-mga, line 102: I wonder if we want to simplify those by now? Not sure why we wrote it this way compared to previous 3.x releases.
```suggestion
| Godot 3.4    | November 2021        | |eol| No longer supported (last update: 3.4.5).                          |
+--------------+----------------------+--------------------------------------------------------------------------+
| Godot 3.3    | April 2021           | |eol| No longer supported (last update: 3.3.4).                          |
```

akien-mga (approved):

about/release_policy.rst
YuriSizov: I think we may want to move deprecated versions to a separate table that is a bit more compact. With the upcoming pace of 3 minor releases of 4.x per year this table will grow exponentially.

akien-mga: Yeah it makes sense to have the upcoming and supported releases in one table, and the EOL ones in another.

YuriSizov:
I'll look into separating the table in a follow-up.

PR #7587, opened 2023-07-01
Adding the "Make Unique" Step to instancing in getting started

rboddy:
You must make an instance of a scene unique to edit it's physic properties in Godot 4.0.3. Screenshot included for clarity.


rboddy:
I missed the note that exists with the caveat about Make Unique for Physics Properties. This isn't a necessary change given that note exists, but I think a screenshot could still help.

getting_started/step_by_step/instancing.rst
```diff
@@ -137,8 +137,11 @@ on the corresponding tab above the viewport.
 
 .. image:: img/instancing_scene_tabs.png
 
-Select one of the instanced Ball nodes and, in the Inspector, set its Gravity
-Scale value to ``10``.
+Select one of the instanced Ball nodes and, in the Inspector, click the arrow next to the physics material property. Now click "make unique" to enable unique physics properties for this instance.
```
Calinou, line 140:
```suggestion
Select one of the instanced Ball nodes. In the Inspector, click the arrow next
to the  **Physics Material** property then choose **Make Unique**
to allow physics properties to be modified separately for this instance.
```

getting_started/step_by_step/instancing.rst
```diff
@@ -137,8 +137,11 @@ on the corresponding tab above the viewport.
...
+
+.. image:: img/instancing_make_unique.png
+
+Now set its Gravity Scale value to ``10``.
```
Calinou, line 144:
```suggestion
Set its **Gravity Scale** value to ``10``.
```

Piralein:
you can read more about it in the [Doc image guidelines](https://docs.godotengine.org/en/stable/contributing/documentation/docs_image_guidelines.html#format-conversion)

rboddy:
> Please convert the image to the `.webp` format. you can read more about it in the [Doc image guidelines](https://docs.godotengine.org/en/stable/contributing/documentation/docs_image_guidelines.html#format-conversion)

Will do, converting shortly

skyace65:
Is this PR still something you can work on?

Piralein:
I'm going to close this PR, because the change isn't necessarily needed and was not updated as requested.
While the screenshot could be helpful, the change to the resource should be an additional step explaining and showing the workflow instead of replacing the current one. The note should in this case also be updated.

PR #7591, opened 2023-07-01
GDScript: Add warning about overriding non-virtual native methods

dalexeev:
See godotengine/godot#78911. I think this is an important enough aspect of the engine to be addressed, despite the fact that there is a warning in GDScript.


tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1535,6 +1535,22 @@ the function name with the attribute operator::
     func dont_override():
         return super.overriding() # This calls the method as defined in the base class.
 
+.. warning::
+
+    One of the common misconceptions is trying to override *non-virtual* engine methods
+    such as ``get_class()``, ``queue_free()``, etc. This is not supported for technical reasons.
+
+    In Godot 3 you can *shadow* engine methods in GDScript, and it will work if you call this method in GDScript.
+    But the engine will not execute your code if the method is called inside the engine on some event.
```
Calinou, line 1544:
```suggestion
    In Godot 3, you can *shadow* engine methods in GDScript, and it will work if you call this method in GDScript.
    But the engine will not execute your code if the method is called inside the engine on some event.
```

Calinou (approved):
Should be good to merge after applying the suggestion.

mhilbrunner:
Thank you! Merged. I think this is well worded and a nice explanation :)

PR #7594, opened 2023-07-01
Link to Using C++ profilers page in CPU optimization

Calinou:
This page was created after CPU optimization, so it wasn't linked beforehand.

### DIFF
--- a/tutorials/performance/cpu_optimization.rst
+++ b/tutorials/performance/cpu_optimization.rst
@@ -56,19 +56,8 @@ External profilers
 Although the Godot IDE profiler is very convenient and useful, sometimes you
 need more power, and the ability to profile the Godot engine source code itself.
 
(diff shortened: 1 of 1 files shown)


skyace65 (commented):
Some of the C++ profilers on that list aren't on the using cpp profilers page. Do we not recommend the ones not listed (Visutal Studio, VTune, Valgrind) anymore?

Calinou:
> Some of the C++ profilers on that list aren't on the using cpp profilers page. Do we not recommend the ones not listed (Visutal Studio, VTune, Valgrind) anymore?

I think all 3 profilers should be usable with Godot, but we need someone to document their use. I don't use these profilers myself.

skyace65 (approved):

skyace65:
Thanks!

PR #7612, opened 2023-07-05
Update C# differences list for 4.1

raulsntos:
- This PR updates the lists of equivalent APIs with the new methods added in 4.1.
- I also noticed some methods in the lists were using the wrong capitalization.
	- The method names haven't changed in 4.1, it's just the documentation that was wrong.
	- I made a separate PR to backport this fix to the 4.0 documentation to make cherry-picking easier.
		- https://github.com/godotengine/godot-docs/pull/7613.

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_collections.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_collections.rst
@@ -121,6 +121,7 @@ all                      `System.Linq.Enumerable.All`_
 any                      `System.Linq.Enumerable.Any`_
 append                   Add
 append_array             AddRange
+assign                   Clear and AddRange
 back                     ``Array[^1]`` or `System.Linq.Enumerable.Last`_ or `System.Linq.Enumerable.LastOrDefault`_
 bsearch                  BinarySearch
 bsearch_custom           N/A
@@ -140,6 +141,7 @@ hash                     GD.Hash
 insert                   Insert
 is_empty                 Use ``Count == 0``
 is_read_only             IsReadOnly
+is_same_typed            N/A
 is_typed                 N/A
 make_read_only           MakeReadOnly
 map                      `System.Linq.Enumerable.Select`_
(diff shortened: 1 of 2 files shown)


RedworkDE (commented):
LGTM, but one things that this page doesn't mention and we probably should add: In C# strings are utf16 while godot strings are utf32, which can changes indices etc when working with string that contain emoji / other non-BMP chars.

raulsntos:
That is indeed a very important difference, I agree it should probably be mentioned in this page. However, this is usually something that users don't have to think about. Do you have any situation in mind that would cause surprising behavior?

You mentioned indices, but I can't think of any API right now where string indices are used outside of BCL APIs (which would be compatible with C# string indices anyway).

RedworkDE:
```
print("😬".length()) # 1
Console.WriteLine("😬".Length); // 2
```

Another example is `get_word_breaks`, which looks like it works correctly until a `𝔲𝔰𝔢𝔯` comes along and everything breaks, because the returned indices are now wrong.

raulsntos:
I'm aware that GDScript's `string.length()` and C#'s `String.Length` will return different values. But in which cases can that become an issue for users? I wouldn't think this would ever be noticed unless you are doing interop and even then, it's probably rare.

Let me clarify that I'm not against adding this note to the documentation, in fact I'm very much in favor. I'm just trying to understand how severe the issue is.

As for `get_word_breaks`, I can see how that one would be problematic. I agree that in this case we should probably fix the bindings, since I'd expect users to use the returned value on a C# string. Although, I'm not sure how high the usage of this method is, since servers are usually an advanced feature and you wouldn't use them directly.

RedworkDE:
Especially when only working with english text it is pretty much a non-issue, but once you internationalize unicode edge cases are one of those things that will occasionally cause weird issues and dealing with both utf16 and utf32 just introduces more opportunities for errors.

mhilbrunner (approved):

mhilbrunner:
Merged for now as it looks good to me :) Thanks for providing both this and the 4.0 one! 🎉

PR #7613, opened 2023-07-05
Fix C# differences list for 4.0

raulsntos:
Fixes the list of equivalent APIs for 4.0.

The methods haven't changed, it's just the documentation that was wrong. Most likely API that was modified after these lists were added to the documentation, but before 4.0 was released.

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_collections.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_collections.rst
@@ -121,6 +121,7 @@ all                      `System.Linq.Enumerable.All`_
 any                      `System.Linq.Enumerable.Any`_
 append                   Add
 append_array             AddRange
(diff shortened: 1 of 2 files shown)


RedworkDE (approved):
https://github.com/godotengine/godot/pull/69547 was created two days after the string table was created, with the now incorrect casing.

Changes look good, didn't check if there are any other places that are not correct.

mhilbrunner (approved):

mhilbrunner:
@raulsntos I assume this should be in 4.1, and `master` for 4.2+ as well? If so, we need to either cherry-pick or do new PRs against those branches :)

raulsntos:
For 4.1+ there's a different PR that also adds other differences that don't apply to 4.0:
- https://github.com/godotengine/godot-docs/pull/7612

PR #7620, opened 2023-07-06
Update mouse_and_input_coordinates.rst

cuppajoeman:
outdated method that no longer exists in 4.1

### DIFF
--- a/tutorials/inputs/mouse_and_input_coordinates.rst
+++ b/tutorials/inputs/mouse_and_input_coordinates.rst
@@ -36,7 +36,7 @@ for example:
            print("Mouse Motion at: ", event.position)
 
        # Print the size of the viewport.
-       print("Viewport Resolution is: ", get_viewport_rect().size)
+       print("Viewport Resolution is: ", get_viewport().get_visible_rect().size)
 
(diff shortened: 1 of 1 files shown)


raulsntos:
The method [`get_viewport_rect`](https://github.com/godotengine/godot/blob/4.1/doc/classes/CanvasItem.xml#L458-L463) still exists in 4.1. It would break compatibility to remove it and we're committed to avoid breaking compatibility.

cuppajoeman:
You're right, but the docs are actually a little misleading as the call to `get_viewport_rect` would only work if the script is extending a `CanvasItem`. I changed my original modification to better suit this.

Where can I find the docs formatting guide?

raulsntos:
> Where can I find the docs formatting guide?

Probably here: https://docs.godotengine.org/en/stable/contributing/documentation/index.html

mhilbrunner:
I think this can be closed, now that https://github.com/godotengine/godot-docs/pull/8149/files is merged?

mhilbrunner:
Followup PR welcome then :) Going to close this one s superseded. Thanks for taking the time to contribute!

PR #7623, opened 2023-07-07
Fix Advanced language docs

a2937:
I fixed the misspelling of "opinionated". Then I corrected the blocks of code.
I removed an extra star in the array initializer as well as the nearby constant. Following that, I duplicated a variable for the standard lib iteration so that both examples will be valid. Then I made sure both the index and value of each character in the array were seen. 

 
Edit: Though I did notice that the character arrays were never actually initialized. Should I update the example?


tutorials/scripting/gdscript/gdscript_advanced.rst
```diff
@@ -322,23 +322,22 @@ Iterating using the C-style for loop in C-derived languages can be quite complex
...
+    const char* stringA = new char[50];
+    const std::string stringsB = new char[50];
 
     [..]
 
     for (int i = 0; i < 50; i++) {
-
-        printf("Value: %s\n", i, strings[i]);
-    }
+		printf("Value: %c Index: %d\n", stringsA[i], i);
```
AThousandShips, line 331:
```suggestion
		printf("Value: %c Index: %d\n", stringA[i], i);
```

tutorials/scripting/gdscript/gdscript_advanced.rst
```diff
@@ -322,23 +322,22 @@ Iterating using the C-style for loop in C-derived languages can be quite complex
 
 .. code-block:: cpp
 
-    const char* strings = new const char*[50];
+    const char* stringA = new char[50];
+    const std::string stringsB = new char[50];
```
AThousandShips, line 326:
```suggestion
    const std::string stringB = new char[50];
```

Also this will cause a memory leak, so better to not give an example like this, but do
```suggestion
    const std::string stringB = std::string(50, ' ');
```
Or something

tutorials/scripting/gdscript/gdscript_advanced.rst
```diff
@@ -322,23 +322,22 @@ Iterating using the C-style for loop in C-derived languages can be quite complex
...
+		printf("Value: %c Index: %d\n", stringsA[i], i);
+	}
 
     // Even in STL:
 
-    for (std::list<std::string>::const_iterator it = strings.begin(); it != strings.end(); it++) {
-
-        std::cout << *it << std::endl;
-    }
+	for (std::string::const_iterator it = stringsB.begin(); it != stringsB.end(); it++) {
```
AThousandShips, line 336:
```suggestion
	for (std::string::const_iterator it = stringB.begin(); it != stringB.end(); it++) {
```

AThousandShips:
I think though that you've misinterpreted how the code is intended to work, it's intended to be an array of strings, not a string, so should be correct, and these changes makes it not match the GDScript code

tutorials/scripting/gdscript/gdscript_advanced.rst
```diff
@@ -322,23 +322,22 @@ Iterating using the C-style for loop in C-derived languages can be quite complex
...
     [..]
 
     for (int i = 0; i < 50; i++) {
-
-        printf("Value: %s\n", i, strings[i]);
-    }
+		printf("Value: %c Index: %d\n", stringsA[i], i);
+	}
 
     // Even in STL:
```
bruvzg, line 334:
```suggestion
    // Even in STL:
    std::list<std::string> strings;
    
    [..]
```

bruvzg (commented):
> Though I did notice that the character arrays were never actually initialized.

It's not intended to be usable code, just a demonstration of the loop structure, so initialization is represented by `[..]` block, and there's no need to rename variables. I guess it can be made in two independent code blocks to make clear it's two examples, not one.

tutorials/scripting/gdscript/gdscript_advanced.rst
```diff
@@ -322,23 +322,24 @@ Iterating using the C-style for loop in C-derived languages can be quite complex
...
+    std::list<std::string> strings;
+    
+    [..]
 
-    for (std::list<std::string>::const_iterator it = strings.begin(); it != strings.end(); it++) {
-
-        std::cout << *it << std::endl;
-    }
+	for (std::string::const_iterator it = stringsB.begin(); it != stringsB.end(); it++) {
```
AThousandShips, line 338:
```suggestion
	for (std::string::const_iterator it = strings.begin(); it != strings.end(); it++) {
```

tutorials/scripting/gdscript/gdscript_advanced.rst
```diff
@@ -322,23 +322,24 @@ Iterating using the C-style for loop in C-derived languages can be quite complex
...
-    const char* strings = new const char*[50];
+    const char** strings = new const char*[50];
 
     [..]
 
     for (int i = 0; i < 50; i++) {
-
-        printf("Value: %s\n", i, strings[i]);
-    }
+		printf("Value: %c Index: %d\n", stringsA[i], i);
```
AThousandShips, line 330:
```suggestion
		printf("Value: %c Index: %d\n", strings[i], i);
```

a2937:
Let me try again

PR #7625, opened 2023-07-08
macOS: Move Vulkan SDK from a note to a requirement

aaronfranke:
This is a requirement, so it should go with the other requirements.

I also changed the link to point directly to the DMG - this should be a stable link and I think it's better since it will save users time not having to navigate the LunarG site.


skyace65 (approved):

skyace65:
Thanks!

PR #7630, opened 2023-07-08
Fix mention of Godot3's or_lesser for export_range

CactiChameleon9:
In this page: https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_exports.html#limiting-editor-input-ranges

The docs say to use the keyword "or_lesser" for ranges that can go below what is specified. This was changed in Godot4 and therefore should be "or_less"

I have also updated the C# version (I am assuming they were both changed, but I have not tested this)

This might need to be backported to previous doc versions?

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_exports.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_exports.rst
@@ -207,13 +207,13 @@ Allow floats from -10 to 20 and snap the value to multiples of 0.2.
     [Export(PropertyHint.Range, "-10,20,0.2")]
     public float Number { get; set; }
 
-If you add the hints "or_greater" and/or "or_lesser" you can go above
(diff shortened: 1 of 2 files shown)


raulsntos:
`or_lesser` was renamed to `or_less` in 4.0:
- https://github.com/godotengine/godot/pull/59589.

> I have also updated the C# version (I am assuming they were both changed, but I have not tested this)

Yes, it also applies to C#. The C# export attributes use the same hint strings than whatever Core defines. Thank you for also updating the C# documentation!

raulsntos (approved):

skyace65 (approved):

skyace65:

CactiChameleon9:
Thank you for merging it - that was surprisingly painless :)

PR #7631, opened 2023-07-08
Fix wording for scalar multiplication of vectors

Sauermann:
Also add an example for multiplication with a negative scalar


clayjohn (approved):

AThousandShips:
Maybe add a note how a negative scalar changes the magnitude but in the opposite direction

mhilbrunner (approved):

clayjohn:
Thanks!

PR #7636, opened 2023-07-10
Fix some articles being incorrectly marked as outdated

Calinou:
These articles were updated for 4.0 and the API hasn't changed in 4.1.

For future reference, I went through all articles currently marked as outdated (100 of them) and these are the only 3 that were true false positives. Many other pages were up-to-date in terms of text content, but screenshots still showed the Godot 3 editor theme (sometimes even 3.0 instead of 3.2+).

### DIFF
--- a/tutorials/2d/using_tilemaps.rst
+++ b/tutorials/2d/using_tilemaps.rst
@@ -1,5 +1,3 @@
-:article_outdated: True
-
 .. _doc_using_tilemaps:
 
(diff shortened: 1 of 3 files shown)


mhilbrunner (approved):

mhilbrunner:
Thanks for doing the work of going through all of these, that was somewhere on my list but... seemed a lot of work, so thanks 🙏 Merged.

PR #7637, opened 2023-07-10
Update Command line tutorial for Godot 4.1

Calinou:
- Update list of command line arguments and descriptions to match Godot 4.1.
- Add a legend to clearly denote which CLI arguments are supported in each build type.
  - This is something we could add to `--help`'s output as part of https://github.com/godotengine/godot/pull/36252, so that you know which arguments are **not** present in less full-featured build types. (We wouldn't display non-functional arguments in export template builds though.)
- Add instructions on making Godot easily available in the `PATH` on Windows and macOS.
- Fix invalid backtick usage in some pages.

___

- See https://github.com/godotengine/godot-docs/issues/5121.

## Preview

Screenshot_20230710_191455

Screenshot_20230710_191505


mhilbrunner:
I wonder if there is a way to auto-generate the list of arguments and descriptions by calling the Godot executable and getting it's help output. Would help them stay in sync.

Probably does not change often enough to do that work though, and may make translations more difficult, not sure.

Anyway, thats kinda unrelated to this PR besides it fixing that outdated content, so thanks! Merged.

PR #7646, opened 2023-07-11
Fix bug in collision masking bits example

rametta:
In the example for setting collision mask bits by script, it is explaining how to set the bits for layers 1, 3, and 4 but the code example is also enabling layer 32, which it shouldn't. Layer 32 should also be `0` instead of `1` to following what was explained above it and below it.


skyace65 (approved):

skyace65:
Thanks!

Issue #7648, opened 2023-07-12
Documentation for GDExtension in 4.1 still refers to 4.0


dcuny:
**4.1**

**Page says it's up to date for 4.1, but still refers to version 4.0**

**[Godot Engine 4.1 documentation in English](https://docs.godotengine.org/en/stable/tutorials/scripting/gdextension/gdextension_cpp_example.html)**

The GDExtension tutorial page says that it's currently up to date with 4.1.

However, the "Note" says that "_you need to use the 4.0 branch of godot-cpp, which is only compatible with Godot 4.0 and being used here as an example._"

The code references the 4.0 version as well:

`git submodule add -b 4.0 https://github.com/godotengine/godot-cpp`

and

`git clone -b 4.0 https://github.com/godotengine/godot-cpp`

Additionally, the release notes state there is a binary incompatibility between GDExtensions built in 4.0 vs. 4.1, which requires a small change. I don't see that information referenced on the page, and this "small change" would be helpful to know about _before_ I start trying to port my 4.0 extensions to 4.1, so I can make that change.

dcuny:
In the [discussion about what changed between Godot 4.0 and 4.1](https://github.com/godotengine/godot-docs/pull/7611), there's mention of a "tiny change to the source code":

> GDExtensions compiled for Godot 4.0 aren't compatible with Godot 4.1 - a tiny change to the source code is required, and they need to be recompiled. As soon as this PR is merged, I'm planning to make a follow-up PR that adds detailed information on what folks need to do to get their GDExtensions working in 4.1.

It appears the change is in the `example_library_init()`. This is the 4.0 version:

```
extern "C" {
// Initialization.
GDExtensionBool GDE_EXPORT example_library_init(const GDExtensionInterface *p_interface, const GDExtensionClassLibraryPtr p_library, GDExtensionInitialization *r_initialization) {
    godot::GDExtensionBinding::InitObject init_obj(p_interface, p_library, r_initialization);

    init_obj.register_initializer(initialize_example_module);
    init_obj.register_terminator(uninitialize_example_module);
    init_obj.set_minimum_library_initialization_level(MODULE_INITIALIZATION_LEVEL_SCENE);

    return init_obj.init();
}
```

And here's what appears to be the 4.1 version:

```
extern "C" {
// Initialization.
[...]

PR #7653, opened 2023-07-12
Update GDExtension docs with 4.1 branch info

paddy-exe:
Fixes https://github.com/godotengine/godot-docs/issues/7648

### DIFF
--- a/tutorials/scripting/gdextension/gdextension_cpp_example.rst
+++ b/tutorials/scripting/gdextension/gdextension_cpp_example.rst
@@ -35,8 +35,8 @@ of Godot. GDExtensions will not work in older versions of Godot (only Godot 4 an
 .. note::
 
     To use `GDExtension <https://godotengine.org/article/introducing-gd-extensions>`__
-    you need to use the ``4.0`` branch of godot-cpp,
-    which is only compatible with Godot 4.0 and being used here as an example.
+    you need to use the ``4.0`` branch or later of godot-cpp.
+    Godot 4.1 is being used here as an example.
     The ``master`` branch is the development branch and is being updated
     regularly to work with godot's ``master`` branch.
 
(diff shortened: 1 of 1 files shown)


skyace65 (approved):

skyace65:
Thanks!

dcuny:
I don't understand how this bug process works. It's been over a week, the original bug has been closed, this ticket merged, and the page _still_ is incorrect.

Additionally, this bullet point is _also_ incorrect:

> a Godot 4 executable,

With the changes to GDExtension, this needs to be the Godot 4.1. Godot 4.0 is not binary compatible with GDExtensions compiled under 4.0.

One of the problems with writing:

> Godot 4.1 is being used here as an example.

is that it glosses over the _other_ difference between Godot 4.0 and 4.1 (which is mentioned in #7648) - the change to the code in the class initializer `example_library_init`.

It would be better to state at the top of the article something along the lines of:

* GDExtension in Godot 4.1 breaks binary compatibility with Godot 4.0;
* The instructions here only apply to Godot 4.1; and
* Instructions for building GDExtension in Godot 4.0 can be found [[INSERT LINK HERE]]

The text can then proceed with the assumption that the user is using Godot 4.1 (or later).

dsnopek:
@dcuny I just posted PR https://github.com/godotengine/godot-docs/pull/7699 in attempt to address your concerns! Please let me know what think.

I didn't follow your suggestions exactly, as I wanted to still attempt to make the page as "future proof" as possible, and not clutter it up with all the specifics related to the GDExtension breakage between Godot 4.0 and 4.1.

dcuny:
The most important thing is that the instructions provided work with the version that it says it work with. 😄 

I'm not sure how future-proof you can make it. You can be optimistic and _hope_ that `GDExtention` remains stable, but honestly, I'd expect it to continue to be problematic in the foreseeable future.

Thanks!

PR #7666, opened 2023-07-15
Fix linear velocity assignment

KristianBerg:
The current tutorial tries to assign a linear velocity to the base Node2D (which has no such attribute) instead of the child RigidBody2D.

I tested the GdScript code but not C#.

This is my first contribution, open to any feedback!


AThousandShips:
The `mob_scene` is a `RigidBody2D` though based on everything in the tutorial, and the example in the demo projects, see [here](https://docs.godotengine.org/en/stable/getting_started/first_2d_game/04.creating_the_enemy.html#node-setup)

KristianBerg:
Ah I see what I got wrong

PR #7671, opened 2023-07-16
Add link to JSON class reference in http_request_class.rst

Skyvastern:
Added link to JSON class reference in one of the lines.


AThousandShips:
I'm not sure if this needs to be repeated again, it's in the line just above

mhilbrunner:
Yeah I think in this case, the link in the line above is enough :) Thanks for taking the time to contribute though!

Issue #7142, opened 2023-04-10
Hi. I'm going nuts trying to find the right documentation for high level multiplayer for godot 4. 


DocMine:
**Your Godot version:**
4.0.3
**Issue description:**
The website seems to have old documentation; 
**URL to the documentation page:**
https://docs.godotengine.org/en/stable/tutorials/networking/high_level_multiplayer.html
If you know how to fix the issue you are reporting please
consider opening a pull request. We provide a tutorial on
using git here: https://docs.godotengine.org/en/stable/community/contributing/pr_workflow.html,
writing documentation at https://docs.godotengine.org/en/stable/community/contributing/docs_writing_guidelines.html
and contributing to the class reference here: https://docs.godotengine.org/en/stable/community/contributing/updating_the_class_reference.html

YuriSizov:
There isn't a new multiplayer article yet, but it is in works.

Geams:
The multiplayer property (type MultiplayerAPI) is missing from the property table with the SceneTree docs.
[Doc](https://docs.godotengine.org/en/latest/classes/class_scenetree.html)

mhilbrunner:
Fixed by https://github.com/godotengine/godot-docs/pull/7675

Issue #5754, opened 2022-04-10
Update High-level multiplayer documentation for Godot 4.0


GamesOnAcidzz:
**Godot Alpha 6**

**RPC changes still aren't documented:**

**https://docs.godotengine.org/en/latest/tutorials/networking/high_level_multiplayer.html?highlight=high%20level**

it still uses the remote, master... instead of @rpc(any_peer, call_local)....

benbot:
@GamesOnAcidzz do you know which PR added the new multiplayer api?

DeadlyEssence01:
I've started some work on this.

RadiantUwU:
RPC still not documented

mhilbrunner:
Fixed by https://github.com/godotengine/godot-docs/pull/7675

botbrayden:
Check moustache in the house 2 review  https://steamcommunity.com/profiles/76561198251297736/recommended/1877820/
Cannot find the fourth frozen button!

PR #7675, opened 2023-07-16
Update high level multiplayer manual for Godot 4.

Jordyfel:
Updates the high level multiplayer manual for Godot 4.

I have used the class reference, blog posts and my experience as a user of the API to get the information. I expect to have missed some details as I don't know the internals of `MultiplayerAPI`.

### DIFF
--- a/tutorials/networking/high_level_multiplayer.rst
+++ b/tutorials/networking/high_level_multiplayer.rst
@@ -1,5 +1,3 @@
-:article_outdated: True
-
 .. _doc_high_level_multiplayer:
 
 High-level multiplayer
@@ -103,47 +101,42 @@ for full IPv6 support.
 Initializing the network
 ------------------------
 
-The object that controls networking in Godot is the same one that controls everything tree-related: :ref:`SceneTree <class_SceneTree>`.
+High level networking in Godot is managed by the :ref:`SceneTree <class_SceneTree>`.
 
-To initialize high-level networking, the SceneTree must be provided a NetworkedMultiplayerPeer object.
+Each node has a ``multiplayer`` property, which is a reference to the ``MultiplayerAPI`` instance configured for it
+by the scene tree. Initially, every node is configured with the same default ``MultiplayerAPI`` object.
 
-To create that object, it first has to be initialized as a server or client.
-
-Initializing as a server, listening on the given port, with a given maximum number of peers:
-
-::
-
-    var peer = NetworkedMultiplayerENet.new()
-    peer.create_server(SERVER_PORT, MAX_PLAYERS)
-    get_tree().network_peer = peer
-
-Initializing as a client, connecting to a given IP and port:
+It is possible to create a new ``MultiplayerAPI`` object and assign it to a ``NodePath`` in the the scene tree,
+which will override ``multiplayer`` for the node at that path and all of its descendants.
+This allows sibling nodes to be configured with different peers, which makes it possible to run a server
+and a client simultaneously in one instance of Godot.
 
 ::
 
-    var peer = NetworkedMultiplayerENet.new()
-    peer.create_client(SERVER_IP, SERVER_PORT)
-    get_tree().network_peer = peer
+    # By default, these expressions are interchangeable.
+    multiplayer # Get the MultiplayerAPI object configured for this node.
+    get_tree().get_multiplayer() # Get the default MultiplayerAPI object.
 
-Get the previously set network peer:
-
-::
-
-    get_tree().get_network_peer()
-
-Checking whether the tree is initialized as a server or client:
+To initialize networking, a ``MultiplayerPeer`` object must be created, initialized as a server or client,
+and passed to the ``MultiplayerAPI``.
 
 ::
 
-    get_tree().is_network_server()
+    # Create client.
+    var peer = ENetMultiplayerPeer.new()
+    peer.create_client(IP_ADDRESS, PORT)
+    multiplayer.multiplayer_peer = peer
 
-Terminating the networking feature:
+    # Create server.
+    var peer = ENetMultiplayerPeer.new()
+    peer.create_server(PORT, MAX_CLIENTS)
+    multiplayer.multiplayer_peer = peer
+
+To terminate networking:
 
 ::
 
-    get_tree().network_peer = null
-
-(Although it may make sense to send a message first to let the other peers know you're going away instead of letting the connection close or timeout, depending on your game.)
+    multiplayer.multiplayer_peer = null
 
 .. warning::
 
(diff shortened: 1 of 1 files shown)


tutorials/networking/high_level_multiplayer.rst
```diff
@@ -103,388 +101,280 @@ for full IPv6 support.
 Initializing the network
 ------------------------
 
-The object that controls networking in Godot is the same one that controls everything tree-related: :ref:`SceneTree <class_SceneTree>`.
-
-To initialize high-level networking, the SceneTree must be provided a NetworkedMultiplayerPeer object.
+High level networking in Godot is managed by the :ref:`SceneTree <class_SceneTree>`.
 
-To create that object, it first has to be initialized as a server or client.
+Every node has a ``multiplayer`` property, which is a reference to a default ``MultiplayerAPI`` instance.
```
AThousandShips, line 106: Misleading, unless defined for the sub-tree it defaults to the default one, but this implies it's always a default one

tutorials/networking/high_level_multiplayer.rst
```diff
@@ -103,388 +101,280 @@ for full IPv6 support.
...
+and a client simultaneously in one instance of Godot.
 
 ::
 
-    var peer = NetworkedMultiplayerENet.new()
-    peer.create_server(SERVER_PORT, MAX_PLAYERS)
-    get_tree().network_peer = peer
+    # By default, these expressions are equivalent.
+    multiplayer
+    get_tree().get_multiplayer()
```
AThousandShips, line 117: Again, only if none has been configured, this is going to confuse people

tutorials/networking/high_level_multiplayer.rst
```diff
@@ -103,388 +101,280 @@ for full IPv6 support.
...
-
-To initialize high-level networking, the SceneTree must be provided a NetworkedMultiplayerPeer object.
+High level networking in Godot is managed by the :ref:`SceneTree <class_SceneTree>`.
 
-To create that object, it first has to be initialized as a server or client.
+Every node has a ``multiplayer`` property, which is a reference to a default ``MultiplayerAPI`` instance.
 
-Initializing as a server, listening on the given port, with a given maximum number of peers:
+It is possible to create a new ``MultiplayerAPI`` object and assign it to a ``NodePath`` in the the scene tree,
+which will override ``multiplayer`` for the node at that path and all of it's children.
```
AThousandShips, line 109:
```suggestion
which will override ``multiplayer`` for the node at that path and all of it's descendants.
```
Not limited to just children

Jordyfel: What is the difference? I don't believe I've come across this term in Godot before.

AThousandShips: "Children" would imply just the direct children, and not grandchildren, great-grandchildren, etc.

Jordyfel: Makes sense.

tutorials/networking/high_level_multiplayer.rst
Jordyfel: That's what I put the "by default" for, but I can see it not being clear enough. I'm open to suggestions for better language since I'm not too sure.

AThousandShips: Why is this relevant? It's only true for a very specific case, so it's unhelpful information

Jordyfel: I wanted to specifically mention that `Node.multiplayer` is a reference to a "global" object, and not one unique per node. I've seen people ask why you cannot set this property even though you can get it. When learning the API I was myself confused about why one can just get multiplayer from every node and receive the same object.

AThousandShips: Then the correct code is `get_tree().get_multiplayer(get_path())`

Jordyfel: I believe this should stay mentioned, otherwise the example I placed below is confusing. Why is `multiplayer` getting set in a singleton, but works for every node? I understand it's documented in the class reference, but I think it's worth for new users to see.

AThousandShips: Sure, but the current wording is far too confusing and will cause people to assume this is the correct way of getting the `MultiplayerAPI` of a `Node`, which is bad

Jordyfel: I did not understand which is the correct way. I assume it's through the `multiplayer` property, because why would it exist otherwise.

It makes sense to me that the preferred way should be mentioned in the docs.

AThousandShips: The correct way is `multiplayer`, it ensures you are in the tree

However calling `get_tree().get_multiplayer()` returns the default one, not the one associated with any one node

Jordyfel: Edited these to be more clear.

AThousandShips: This is still going to confuse people, why not use the proper equal expression as an example?

Jordyfel: I'm not willing to die on this hill... for now I want to see if a networking maintainer approves of the content generally before I put more effort into polishing it.

AThousandShips: I'm sorry if this was unpleasant, I didn't mean for it to become an argument, my apologies

mhilbrunner:
@Jordyfel Thanks for working on this, I've work in progress on expanding and reworking the networking sections, but as that turned out to take longer than anticipated I had planned to do a quicker update of the existing ones - so this is very welcome.

I'm currently building this page locally and will be reviewing it, thanks for contributing to Godot, especially to networking docs ❤️

mhilbrunner (approved):
Great work! I've finally found time to review this in depth. This is a great update to bring this page up to date for Godot 4.

I've found a few things, but these may as well be worth doing in followup PRs (by you, me or someone else):

1. The warning about the Android INTERNET permission was removed in this PR. Was that intentional?
2. Where it explains the signals, the signal names should probably be links to their class docs.
3. I think the flow of the page could be improved (that has nothing to do with this PR though, it was already the case before): It starts with a lot of very high level general networking talk, then gets nitty gritty, and kinda goes back and forth and thats hard to follow. We should probably split the general networking overview, security considerations, comparing Godot's networking technologies etc. from the actual "this is how you use MultiplayerAPI".

It also needs advanced tutorials for the synchronizer tech etc, but that is out of scope for this PR.

All in all, I'll likely merge this sooner rather than later and iterate based on it, as I have further changes in my own fork and will work on the above (further PRs are of course welcome!)

Because as of right now, the current state of the page is badly outdated, so I'm in favor of improving it and then improving it further, rather than spending more time before merging. :)

Jordyfel:
I agree with everything you mentioned, this page probably has to be split/reorganized, though I'm not very confident in doing that since my networking knowledge is rather surface level. I just wanted to have all the basics in one place sooner rather than later because this API, more than most other, is a pain to figure out from the class reference as a new user (not so much the case for the scene replication nodes, which is why I didn't include them, though it would be better if they had tutorials as well).

I didn't notice the android permission warning, will put it back as it was.
I noticed a lot of class/method names could use links to the class reference, I just hadn't figured out how to do it when I was writing. Will try to get it done soon, or if you decide to merge before that, in a separate PR.

tutorials/networking/high_level_multiplayer.rst
FrederickKDP: I find  `get_tree().get_multiplayer()` to be overly verbose. `get_multiplayer()` works on its own and makes more sense in my opinion. This is important for a tutorial.

AThousandShips: Those are not equivalent, as is the entire point of this discussion here

FrederickKDP: > Those are not equivalent, as is the entire point of this discussion here

I'm not saying that they are. I'm saying I'm not fond of using `get_tree().get_multiplayer()` for the reasons explained. If people really believe it's a must, I believe further explanation is necessary. Because otherwise, people will have the same doubts that we are having in this discussion.
Reasons to use `get_tree().get_multiplayer()` should be clear if used in my opinion.

FrederickKDP:
> I agree with everything you mentioned, this page probably has to be split/reorganized, though I'm not very confident in doing that since my networking knowledge is rather surface level. I just wanted to have all the basics in one place sooner rather than later because this API, more than most other, is a pain to figure out from the class reference as a new user (not so much the case for the scene replication nodes, which is why I didn't include them, though it would be better if they had tutorials as well).
> 
> I didn't notice the android permission warning, will put it back as it was. I noticed a lot of class/method names could use links to the class reference, I just hadn't figured out how to do it when I was writing. Will try to get it done soon, or if you decide to merge before that, in a separate PR.

Agreed. I was myself in this situation. It's very painful to figure it out as a novice.
I would argue that a point to have here is how in-depth should this page be. Is it for novices, or for pros. If is for newcomers, how pros will find clarification? If it tries to be for both, how to explain that without overly complicate for newbies?

tutorials/networking/high_level_multiplayer.rst
AThousandShips: I'm sorry your original comment made it seem that way based on the suggestion of replacing one with the other

FrederickKDP: I see, no worries. It was confusing, the way I put it. My fault as well.
Always please feel free to argue my posts. I'm new here, so I appreciate a lot the feedback. Hopefully I may contribute to the discussion.

mhilbrunner:
Merged! Thanks and congrats @Jordyfel, this is amazing, doubly so for a first contribution :)

> I would argue that a point to have here is how in-depth should this page be

Absolutely. We _strive_ to keep things as welcoming as reasonably possible, but we absolutely do not try to teach people game development (or network engineering), that is way out of scope of the docs and our resources. The Godot's are meant to be a technical manual for this specific game engine; creating in-depth networking/multiplayer tutorials is work for the greater community, we can only manage to document the tech.

Jordyfel:
Other issues I found which can seemingly also be closed by this: #5754, #7142. @mhilbrunner

Some guy on reddit complained that the issue he was tracking wasn't closed.

mhilbrunner:
Thanks for the ping, I closed those issues now :)

PR #7676, opened 2023-07-16

Varusal:
Edit:
Not an issue with the code.

~~Changed clamp code since with the GDScript the player was only able to move diagonal.~~

~~With the change the player is now able to move in all directions (code is the same as in the demo sample)~~

### DIFF
--- a/getting_started/first_2d_game/03.coding_the_player.rst
+++ b/getting_started/first_2d_game/03.coding_the_player.rst
@@ -297,8 +297,7 @@ the ``_process`` function (make sure it's not indented under the `else`):
  .. code-tab:: gdscript GDScript
 
         position += velocity * delta
-        position.x = clamp(position.x, 0, screen_size.x)
-        position.y = clamp(position.y, 0, screen_size.y)
+        position = position.clamp(Vector2.ZERO, screen_size)
 
  .. code-tab:: csharp
 


AThousandShips:
This works though, are you sure you copied the code correctly?

Varusal:
Hi @AThousandShips 
just tested it you are right. I missed y with x and vice versa.

Sorry about the miss information!

AThousandShips:
No worries, still worth adding, but please update the commit message to not say that it fixes non-functioning behavior :)

Piralein (approved):

mhilbrunner (approved):

mhilbrunner:
Thanks, merged! Congrats on your first merged contribution to Godot's docs! 🎉

dedetheprogrammer:
This is not a issue as such, but I think there is a improvement on the clamp code:
```GDScript
position += velocity * delta
position = position.clamp(Vector2.ZERO, screen_size)
```

Maybe there are some features that I don't know in Godot, but there is a small question that is probably interesting to mention, as we see, with clamp we are clamping the position between both declared vectors, but there is an issue with the pivot:

image

As we can see, the pivot is at the center of the Player node, but if we use the given code, half of the sprite is out of the screen because of the definition of the pivot itself:

image

Then, because the pivot defines the position of the node, when we clamping the position, the position is truly inside the screen size but half of the sprite is outside because the pivot it's at the center of the node. 

After some research and investigation, I've found a solution for the people is interested (in Godot 4.1.1.) to avoid the sprite going outside the screen:

```GDScript
# Get the SpriteFrames component of your AnimatedSprite2D node:
var sprite_frames = $AnimatedSprite2D.sprite_frames
# Get the first texture of the wanted animation (in this case, walk, you can also get the size
# in differents cases)
# If your animation frames has different sizes, use $AnimatedSprite2D.frame instead of 0
var texture       = sprite_frames.get_frame_texture("walk", 0)
# Get frame size:
var texture_size  = texture.get_size()
# This is not the end, you will get the texture size, not the node real size, then you need to
# multiply the texture size with the node scale
var as2d_size     = texture_size * $AnimatedSprite2D.get_scale()
```

I don't know the method to edit a page so I put this here

PR #7673, opened 2023-07-16
Fix typo in `06.heads_up_display.rst`

vmarnauza:
Fix typo: to -> too

PR #7686, opened 2023-07-18
Little typo fix 06.heads_up_display.rst

ViktorVoloshko:
"too" instead of "to" as should be.

This was fixed in master branch #7673 but not in 4.1 branch.


skyace65:
That PR has the "cherrypick:4.1" label which means it will get applied to the 4.1 branch in the future.

PR #7687, opened 2023-07-18
Update Triaging guidelines

Sauermann:
- Remove unavailable labels
- Add newly created labels
- Update descriptions, where applicable
- Sort content alphabetically

### DIFF
--- a/contributing/workflow/bug_triage_guidelines.rst
+++ b/contributing/workflow/bug_triage_guidelines.rst
@@ -45,8 +45,6 @@ The following labels are currently defined in the Godot repository:
 -  *Bug*: describes something that is not working properly.
 -  *Cherrypick*: describes something that can be backported to a stable branch
    after being merged in the ``master`` branch.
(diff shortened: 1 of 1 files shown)


contributing/workflow/bug_triage_guidelines.rst
```diff
@@ -75,17 +75,15 @@ The following labels are currently defined in the Godot repository:
    the code base.
```
AThousandShips, line 75: Suggest adding something about removing the label if an open PR is associated with it, to prevent duplicates as new contributors might not be able to recognize that a PR is linked, as suggested in chat

mhilbrunner (approved):
Looks good to me. I'm happy to have a note as suggested by AThousandShips added, but otherwise this is good to merge for me.

contributing/workflow/bug_triage_guidelines.rst
Sauermann: That is a good idea. I have updated the PR.

Issue #7616, opened 2023-07-05
Incorrect code sample in GDScript advanced language documentation


bezeltorpe:
**Your Godot version:** 4.0

**Issue description:** Error in documentation for GDScript
The printf in the C++ code sample under "For & while" has one format specifier, but two arguments. (Issue also exists in the 3.1 documentation for what it's worth.) I'm not sure how it's supposed to look, so I opened an issue instead of a PR...

Additionally, there is a typo on the following line of text after the code sample.

> Because of this, GDScript makes the opinonated decision to have a for-in loop over iterables instead:

"opinonated" should be "opinionated"

**URL to the documentation page:** [link](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_advanced.html#for-while)

PR #7691, opened 2023-07-19
Fix Advanced language docs

a2937:
I fixed the misspelling of "opinionated". Then I corrected the blocks of code.
I removed an extra star in the array initializer as well as the nearby constant. Following that, I duplicated a variable for the standard lib iteration so that both examples will be valid. Then I made sure both the index and value of each character in the array were seen.

_Bugsquad edit: fixes https://github.com/godotengine/godot-docs/issues/7616_


mhilbrunner (approved):

mhilbrunner:
Thanks! Merged.

PR #7692, opened 2023-07-20
Update index.rst

MrOrangeTan:
Fixed the issue on line 35. The issue being the perfect opportunity for a pun not taken.

### DIFF
--- a/.github/workflows/sync_class_ref.yml
+++ b/.github/workflows/sync_class_ref.yml
@@ -2,15 +2,10 @@ name: Sync Class Reference
 
 on:
   workflow_dispatch:
-  # Scheduled updates only run on the default/master branch.
-  schedule:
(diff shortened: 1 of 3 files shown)


AThousandShips:

AThousandShips:
You will also have to rebase your PR on master, with `git rebase -i master`, see [here](https://docs.godotengine.org/en/latest/contributing/workflow/pr_workflow.html#the-interactive-rebase)

Calinou:
I don't think we should use this pun, as it won't translate well to other languages. In its original meaning, "Git" is an insult after all.

Also, the `master` branch should remain called the `master` branch to match the GitHub repository.

YuriSizov:
So thanks for lightening the mood, but we are going to reject this ;)

PR #6848, opened 2023-03-01
Update using Skeleton2D documentation

ShatReal:
See #6388

I have renamed the cutout animation page to "2D Skeletons" and the 2d skeletons page to "2d Skeletal Deform" and updated the contents of each page accordingly. I cut out the stuff about making bones with sprites because that's not possible anymore.

PR #7694, opened 2023-07-20
Update skel2d14.png to skel2d14_1.webp

zhangwb1996:
Update skel2d14.png to skel2d14_1.webp


skyace65:
A more comprehensive update of that page is already being done in #6848. I'll leave this open until that PR gets merged but in the future please check to see if someone is already working on what you want to update.

In addition please use the color yellow for outlines, specifically hex fffb44 (fffb44ff if there's transparency). we use this instead of red for the benefit of color blind people. And please keep PRs to a single commit. We have guidelines on all of this and how to do it in the docs [here](https://docs.godotengine.org/en/stable/contributing/documentation/index.html#guidelines).

PR #7697, opened 2023-07-20
Update signals.rst

MohsenHNSJ:
Fixed a typo at line 309 - parenthesis not present after calling the function in C#


AThousandShips:
Another instance on line 386

MohsenHNSJ:
Sorry, I was wrong.
This is a call for action, so no Parenthesis is required.
The current text is fine and correct.

If you put Parathesis, You will get (CS0029: Cannot implicitly convert type 'void' to 'System.Action' ) and the build process would fail.

PR #7699, opened 2023-07-20
Update GDExtension C++ tutorial to explain compatibility and point out Godot 4.1 breakage

dsnopek:
These changes attempt to address a critique of this page from [here](https://github.com/godotengine/godot-docs/pull/7653#issuecomment-1639121638).

The issues that were called out there revolve around the compatibility breakage between GDExtensions for Godot 4.0 and 4.1.

My changes here attempt to explain generally how compatibility is meant to work with GDExtension, as well as its current status as experimental, and a link to more detailed information around the specific 4.0 -> 4.1 breakage.

Please let me know what you think!

### DIFF
--- a/tutorials/scripting/gdextension/gdextension_cpp_example.rst
+++ b/tutorials/scripting/gdextension/gdextension_cpp_example.rst
@@ -33,12 +33,23 @@ Note that this repository has different branches for different versions
 of Godot. GDExtensions will not work in older versions of Godot (only Godot 4 and up) and vice versa, so make sure you download the correct branch.
 
 .. note::
-
     To use `GDExtension <https://godotengine.org/article/introducing-gd-extensions>`__
-    you need to use the ``4.0`` branch or later of godot-cpp.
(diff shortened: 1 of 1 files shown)


dsnopek:
> I do wonder if we have to change the branch info for every minor version after the system will be stable?

I'm hoping that we won't, because that'll be annoying to maintain :-)

For now, the page is still saying it's using `4.1` as an example, but if this continues to confuse people, I guess we may have to keep updating it every minor version. We'll see if any new issues get opened about it in 4.2!

mhilbrunner (approved):

mhilbrunner:
@dsnopek Great work, merged. Any reason this should not be also cherry-picked to 4.0? Even if it uses 4.1 as an example, all this should be relevant to all 4.x versions when creating GDExtensions, right?

dsnopek:
Thanks!

Well, it links to the "Updating to 4.1" page which I don't think is present in the 4.0 docs. But if that bit were remove, yes, I guess this is relevant for 4.0 as well!

PR #7700, opened 2023-07-20
Update player_scene_nodes.webp

MohsenHNSJ:
The screenshot now correctly shows the state it should show.

At line 31, it is requested to select the "Group selected nodes" option, that will show an icon next to the "Player" node. This icon is not present at the current screenshot.


skyace65:

Issue #7685, opened 2023-07-18
Improve GDExtension documentation 


tarragonfly:
**Your Godot version:**
4.1

**Issue description:**

GDExtension documentation is very sparse for people who are not deeply familiar with Godot. It's unclear what exactly is the difference between going the engine module route vs GDExtension besides not having to re-compile the engine. All the docs mention is that engine modules offer deeper integration, but in what ways and for which systems, it's not very clear. It's also unclear if any special considerations have to be taken into an account for mobile and web platforms. There's no information on platform support in general.

**URL to the documentation page (if already existing):**

https://docs.godotengine.org/en/stable/tutorials/scripting/gdextension/what_is_gdextension.html

dsnopek:
Thanks! I agree that GDExtension could use lots more documentation :-)

To address some specific points:

> clear information on where exactly GDExtension does not go as deep as engine modules

The _goal_ of GDExtension and godot-cpp is that you should be able to write C++ code that looks just like module code (and with some `#ifdef`s even be able to compile as a module or GDExtension), however, there are still some missing pieces. Those pieces aren't missing on purpose, they just haven't been done yet, and we may not even realize it (it's hard to notice things that _aren't_ there ;-)).

So, rather than trying to document the missing pieces, I'd personally prefer if folks opened issues when they encounter something that's possible from a module that isn't possible in a GDExtension, so we can start the discussion around how to expose that functionality to GDExtension.

I guess this too should be put in the documentation :-)

> In this case, AnimationPlayer is the relevant discussion point. It's accessible via engine module route, but can it also be extended via GDExtension?

Yes!

> Since the docs mention engine modules having deeper engine integration, it gets chosen by default, which is not ideal.

Well, since you should be able to write the source code such that it can be compiled either as a module or GDExtension (there's an example of this in Godot itself: the core 'text_server_adv' module can optionally be compiled as a GDExtension), you could start by attempting to make it as GDExtension, but if you hit roadblock, open some issues and switch to compiling as a module until those issues are resolved? This way you're still able to get work done, and GDExtension becomes more capable at the same time.

tarragonfly:
> Thanks! I agree that GDExtension could use lots more documentation :-)
> 
> To address some specific points:
> 
> > clear information on where exactly GDExtension does not go as deep as engine modules
> 
[...]

PR #7706, opened 2023-07-21
Add request to report missing exposed GDExtension functionality

paddy-exe:
* adds note to the "What is GDExtension" docs to call for issue creation when people can't do something in GDExtension that is possible with modules and should be exposed to GDExtension

Created in mind of https://github.com/godotengine/godot-docs/issues/7685#issuecomment-1644017193


dsnopek (approved):
Thanks for making this PR! Looks great to me :-)

paddy-exe:
> Thanks for making this PR! Looks great to me :-)

No problem :D should this get backported to all 4.0 versions then?

mhilbrunner (approved):

mhilbrunner:
Makes sense to me. Thanks! Merged.

PR #7716, opened 2023-07-23
Update list of Godot version tags in Bisecting regressions

Calinou:


mhilbrunner (approved):

mhilbrunner:
Thanks! Long-term, I am not sure whether we need to maintain a complete list here, but updating it doesn't hurt I guess. And if we do want to maintain it, we should probably start to maintain a check list of places to update when we release a new version.

PR #7720, opened 2023-07-24
Mark UWP as non-functional in its respective export and compilation pages

Calinou:
The associated text was removed as it's not relevant for Godot 4.

It may be better to keep page stubs rather than removing it entirely to avoid breaking URLs, so I went for that route.


skyace65 (approved):

mhilbrunner (approved):

mhilbrunner:
Makes sense, merged.

Issue #7681, opened 2023-07-18
GDExtension C++ example page inconsistency


VincentSinel:
**Your Godot version:**
4.1
**Issue description:**

In the GDExtension C++ example page, for downloading godot-cpp repository, line of code provided use godot version 4.0 in the line :
`git submodule add -b 4.0 https://github.com/godotengine/godot-cpp`

but  after that for the code provide for `register_types.cpp`, it use class `GDExtensionInterfaceGetProcAddress` that doesn't exist for Godot 4.0 but only with version 4.1 of godot-cpp.

**URL to the documentation page:**

https://docs.godotengine.org/en/stable/tutorials/scripting/gdextension/gdextension_cpp_example.html

PR #7737, opened 2023-07-27
Fix version number in GDExtension page

KuKetto:
The [documentation](https://docs.godotengine.org/en/stable/tutorials/scripting/gdextension/gdextension_cpp_example.html) stated Godot 4.0 instead of 4.1

Cloning repository 4.0 caused a missing type specifier in **Creating a simple plugin** section in the register_types.cpp implementation. As in godot-cpp 4.0 [docs](https://github.com/godotengine/godot-cpp/tree/4.0#getting-started) `const GDExtensionInterface *p_interface` is used instead of `GDExtensionInterfaceGetProcAddress p_get_proc_address` [(here)](https://github.com/godotengine/godot-cpp/tree/4.1#getting-started) 
gdexample.gdextension also stated `compatibility_minimum = 4.1` so that could cause issues as well.

This correct versioning should fix this version issue, and the project shall compile without any errors.

### DIFF
--- a/tutorials/scripting/gdextension/gdextension_cpp_example.rst
+++ b/tutorials/scripting/gdextension/gdextension_cpp_example.rst
@@ -35,8 +35,8 @@ of Godot. GDExtensions will not work in older versions of Godot (only Godot 4 an
 .. note::
 
     To use `GDExtension <https://godotengine.org/article/introducing-gd-extensions>`__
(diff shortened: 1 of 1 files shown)


KuKetto:
https://github.com/godotengine/godot-docs/issues/7681#issuecomment-1640279189

PR #7745, opened 2023-07-28
Update 06.heads_up_display.rst

DestroCore:
Changes a simple spelling error in 06.heads_up_display.rst (https://docs.godotengine.org/en/stable/getting_started/first_2d_game/06.heads_up_display.html#)

- 'to' is changed to 'too' in the context
 - ('still TOO small' rather than 'still to small')


AThousandShips:

PR #7752, opened 2023-07-30
Made scrollbar follow user's color theme

Ymanawat:
This pull request makes a small fix to the documentation website. The problem is that the scrollbar on the website doesn't follow the user's theme settings. I tested this in different web browsers. We had similar issue on the Godot website as well.
light scrollbar (1)


Piralein (approved):

mhilbrunner:
Cherry-picked to 4.1

Issue #7592, opened 2023-07-01
Squash the Creeps 3D Score Bug


J-John1:
**Your Godot version:**
4.0.2.stable

**Issue description:**
While making the "Squash The Creeps 3D" tutorial, I had an issue with the following piece of code, when trying to use it in the ScoreLabel to update the current score:

	for index in range(get_slide_collision_count()):
		# We get one of the collisions with the player
		var collision = get_slide_collision(index)
		
		# If the collision is with ground
		if (collision.get_collider() == null):
			continue
		
		# If the collider is with a mob
		if collision.get_collider().is_in_group("mob"):
			var mob = collision.get_collider()
			# we check that we are hitting it from above.
			if Vector3.UP.dot(collision.get_normal()) > 0.1:
				# If so, we squash it and bounce.
				mob.squash()
				target_velocity.y = bounce_impulse

When playing the game, each squashed creep increases the score by a random integer between 1-3 points each time. I believe this happens because the squashed signal is being emitted for each collision count. I tried putting "return" at the end of the "if collision.get_collider().is_in_group("mob"):" statement and this seemed to fix the problem. Due to the limitation of my programming experience, I believe it would be for the better if someone more knowledgeable could check the cause for this problem.

**URL to the documentation page:**
https://docs.godotengine.org/en/stable/getting_started/first_3d_game/06.jump_and_squash.html

PR #7767, opened 2023-08-02
Fix duplicate mob squash logic in 3D tutorial (Your first 3D game) 

jjaxu:
fixes https://github.com/godotengine/godot-docs/issues/7592

Hi there! 👋 

I was following the [Godot 3D tutorial (Your first 3D game)](https://docs.godotengine.org/en/stable/getting_started/first_3d_game/index.html) and believe to have discovered a bug regarding squashing the mobs.

The [collision detection](https://docs.godotengine.org/en/stable/getting_started/first_3d_game/06.jump_and_squash.html#coding-the-squash-mechanic) section mentions `"Godot makes the body move sometimes multiple times in a row to smooth out the character's motion. So we have to loop over all collisions that may have happened."` and thus checks for collisions in a loop.

This current logic however will sometimes result in duplicate calls to `mob.squash()`, which will double or triple count the score for one mob kill. The simple fix is to `break` out of the loop after handling the first squash occurrence to prevent further duplicate calls.

### Demo
Here's the demo, notice how the first two kills increased the score by two, then the last two by one. I've seen anywhere between 1-3 
demo

I've updated all occurrences of this code snippet that appear for both GDScript and C# to include the `break` statement, as well as an explanation of the logic. The files changed are:
- 06.jump_and_squash
- 07.killing_player 
- 09.adding_animations

### Preview
GDScript
image
C#
image
Explanation
image

Please help verify whether this is the correct solution. If it is, I'm also more than happy to apply the fix to the [godot-3d-dodge-the-creeps](https://github.com/godotengine/godot-3d-dodge-the-creeps/blob/master/dodge_the_creeps_3d/Player.gd#L60) repo as well.

Thanks and have a great day!

* *Bugsquad edit, alternative to: #8333*

### DIFF
--- a/getting_started/first_3d_game/06.jump_and_squash.rst
+++ b/getting_started/first_3d_game/06.jump_and_squash.rst
@@ -255,6 +255,8 @@ With this code, if no collisions occurred on a given frame, the loop won't run.
                     # If so, we squash it and bounce.
                     mob.squash()
                     target_velocity.y = bounce_impulse
+                    # Prevent further duplicate calls.
+                    break
 
  .. code-tab:: csharp
 
@@ -279,6 +281,8 @@ With this code, if no collisions occurred on a given frame, the loop won't run.
                     // If so, we squash it and bounce.
                     mob.Squash();
                     _targetVelocity.Y = BounceImpulse;
+                    // Prevent further duplicate calls.
+                    break;
                 }
             }
         }
@@ -308,6 +312,10 @@ allows us to compare it to the up direction.
 With dot products, when the result is greater than ``0``, the two vectors are at
 an angle of fewer than 90 degrees. A value higher than ``0.1`` tells us that we
 are roughly above the monster.
+
+After handling the squash and bounce logic, we terminate the loop early via the ``break`` statement
+to prevent further duplicate calls to ``mob.squash()``, which may otherwise result in unintended bugs
+such as counting the score multiple times for one kill.
 
 We are calling one undefined function, ``mob.squash()``, so we have to add it to
 the Mob class.
--- a/getting_started/first_3d_game/07.killing_player.rst
+++ b/getting_started/first_3d_game/07.killing_player.rst
@@ -395,6 +395,8 @@ Finally, the longest script, ``Player.gd``:
                     # If so, we squash it and bounce.
                     mob.squash()
                     target_velocity.y = bounce_impulse
+                    # Prevent further duplicate calls.
+                    break
 
         # Moving the Character
         velocity = target_velocity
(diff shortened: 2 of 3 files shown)


getting_started/first_3d_game/06.jump_and_squash.rst
```diff
@@ -255,6 +255,8 @@ With this code, if no collisions occurred on a given frame, the loop won't run.
                    # If so, we squash it and bounce.
                    mob.squash()
                    target_velocity.y = bounce_impulse
+                   # Prevent further duplicate calls
```
AThousandShips, line 258:
```suggestion
                   # Prevent further duplicate calls.
```
End comments with a period, same elsewhere

jjaxu:
I believe this fixes: https://github.com/godotengine/godot-docs/issues/7592

AThousandShips:
To link it add the line "Fixes: #7592" to your first message (the one at the very top)

jjaxu:
Linked the issue. Glad to see using `break` was the intended solution. What about the standalone [godot-3d-dodge-the-creeps](https://github.com/godotengine/godot-3d-dodge-the-creeps/blob/master/dodge_the_creeps_3d/Player.gd#L60) repo? Seems to me that it's no longer maintained and the demos have since been migrated to [godot-demo-projects](https://github.com/godotengine/godot-demo-projects/blob/master/3d/squash_the_creeps)

AThousandShips:
That does seem to be the case yes

jjaxu:
@AThousandShips Is there anything else that needs to be done before this can be approved and merged? There is nothing else from my side.

AThousandShips:
You'll have to see what further reviewers say

sdlllllll:
I'm a beginner and also noticed this bug in the tutorial. Is that mean if I use collision detection in any calculation, a "break" or "return" is necessary in order to prevent a duplicate calculation.

george-lim:
This would be a very helpful for onboarding new people, I also ran into this bug.

@AThousandShips any chance we could get another pair of eyes from the team to review? It's definitely a nice to have and the changes seem harmless / trivial :)

AThousandShips (approved):
Looks good to me, haven't verified the behavior but is the same logic as the 2D case

jjaxu:
Thanks @AThousandShips, I have squashed my commits and rebased the changes onto the latest `master`.

jjaxu:
@AThousandShips  Commit https://github.com/godotengine/godot-docs/commit/426c83b added some whitespace changes which caused merge conflicts. I've rebased my changes again to resolve. Will require approval again. Thanks!

Calinou (approved):
Thanks!

J-John1:
Thanks for your help everyone!

PR #7769, opened 2023-08-03
Update class_editorplugin

### LINKED ISSUE #7757: Scene autoload at Making plugings: Registering autoloads/singletons in plugins
**Your Godot version:**

4.1

**Issue description:**

At the end of the page, there is a code block with a comment:

> # The autoload can be a scene or script file.

If you look at `EditorPlugin.add_autoload_singleton` documentation:

> Adds a script at path to the Autoload list as name.

It should be added that a scene can also be passed.

https://docs.godotengine.org/en/stable/tutorials/plugins/editor/making_plugins.html

Ymanawat:
Changes script to scene or script file

This resolves #7757


AThousandShips:

PR #7780, opened 2023-08-06
Update gdscript_advanced.rst

j981:
the translation between C-style reverse loop and GDScript shows a buggy code in the latter


AThousandShips:

PR #7781, opened 2023-08-06
Document C# diagnostics

raulsntos:
Adds documentation pages for every C# diagnostic reported by Godot analyzers so they can be linked from the IDE.

These pages attempt to follow the same format as Microsoft's .NET documentation (example: [CA1001](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1001)).

- Counterpart in the main repository: https://github.com/godotengine/godot/pull/80489.

### DIFF
--- a/tutorials/scripting/c_sharp/diagnostics/GD0001.rst
+++ b/tutorials/scripting/c_sharp/diagnostics/GD0001.rst
@@ -1,0 +1,44 @@
+GD0001: Missing partial modifier on declaration of type which is a subclass of GodotObject
+==========================================================================================
+
+====================================  ======================================
+                                      Value
+====================================  ======================================
+**Rule ID**                           GD0001
+**Category**                          Usage
+**Fix is breaking or non-breaking**   Non-breaking
+**Enabled by default**                Yes
+====================================  ======================================
+
+Cause
+-----
+
+A type that derives from ``GodotObject`` is not declared partial.
+
+Rule description
+----------------
+
+Godot source generators add generated code to user-defined types to implement
+the integration with the engine. Source generators can't add generated code to
+types that aren't declared partial.
+
+.. code-block:: csharp
+
+    // The source generators can't enhance this type to work with Godot.
+    public class InvalidNode : Node { }
+
+    // The source generators can enhance this type to work with Godot.
+    public partial class ValidNode { }
+
+How to fix violations
+---------------------
+
+To fix a violation of this rule, add the ``partial`` keyword to the type
+declaration.
+
+When to suppress warnings
+-------------------------
+
+Do not suppress a warning from this rule. Types that derive from ``GodotObject``
+but aren't partial can't be enhanced by the source generators, resulting in
+unexpected runtime errors.
--- a/tutorials/scripting/c_sharp/diagnostics/GD0002.rst
+++ b/tutorials/scripting/c_sharp/diagnostics/GD0002.rst
@@ -1,0 +1,52 @@
+GD0002: Missing partial modifier on declaration of type which contains one or more subclasses of GodotObject
+============================================================================================================
+
+====================================  ======================================
+                                      Value
+====================================  ======================================
+**Rule ID**                           GD0002
+**Category**                          Usage
+**Fix is breaking or non-breaking**   Non-breaking
+**Enabled by default**                Yes
+====================================  ======================================
+
+Cause
+-----
+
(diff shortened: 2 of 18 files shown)


tutorials/scripting/c_sharp/diagnostics/GD0001.rst
```diff
@@ -0,0 +1,66 @@
...
+    // The source generators can enhance this type to work with Godot.
+    public partial class ValidNode { }
+
+How to fix violations
+---------------------
+
+To fix a violation of this rule, add the ``partial`` keyword to the type
+declaration.
+
+.. tip::
```
RedworkDE, line 39: IMO this is not really helpful for this error as it you are going to have the boilerplate code collection of the century because you write all that generated code manually (so unlikely to ever happen) or you implement your own source generator (vastly more likely) and still have to mark that member as partial.

Is this already documented elsewhere or should this be moved to another page.

tutorials/scripting/c_sharp/diagnostics/GD0002.rst
```diff
@@ -0,0 +1,74 @@
+GD0002: Missing partial modifier on declaration of type which contains one or more subclasses of GodotObject
```
RedworkDE, line 1:
```suggestion
GD0002: Missing partial modifier on declaration of type which contains one or more subclass of GodotObject
```

(I think) But this is probably because the actual diagnostic.

tutorials/scripting/c_sharp/diagnostics/GD0203.rst
```diff
@@ -0,0 +1,56 @@
...
+Cause
+-----
+
+A delegate annotated with the ``[Signal]`` attribute has a return type when
+``void`` was expected.
+
+Rule description
+----------------
+
+Every signal must return ``void``. There can be multiple callbacks registered
```
RedworkDE, line 24: (Not really a change suggestion) This is a also a thing with normal C# events, tho I am not sure how this is resolved there.

RedworkDE (approved):

tutorials/scripting/c_sharp/diagnostics/GD0203.rst
raulsntos: I believe when invoking C# events with a return value it just takes the value from the last callback, but I don't think it's very useful so probably nobody uses it.

tutorials/scripting/c_sharp/diagnostics/GD0001.rst
raulsntos: I consider disabling source generators to be an advanced use case, so I agree we could remove the tip or maybe make it less prominent. I just though it would make sense to mention all the ways in which you can "fix" the error.

> Is this already documented elsewhere

It's not documented anywhere else, maybe it should be, but I don't really want to promote it too much since it's likely not what most users would want to do.

RedworkDE: I feel like this belongs on a page `Pointers for advanced usage scenarios, but you have to go source diving to really realize them` or alternatively it could just stay a thing that is mentions on a few Github issues, but otherwise can only be found in the source. tho mentioning it on the docs would make it clear that this is not just an internal thing for development.

tutorials/scripting/c_sharp/diagnostics/GD0002.rst
RedworkDE: It just looked wrong to me, with some more thought, both versions are probably fine, but if we want to change it I would also go to a different wording like `contains nested classes that are derived from GodotObject`

tutorials/scripting/c_sharp/diagnostics/GD0001.rst
raulsntos: I decided to remove the tip and we can decide how to surface this feature in the future.

tutorials/scripting/c_sharp/diagnostics/GD0002.rst
raulsntos: I like that wording. This change needs to be done in the main repository as well, is there any other diagnostic that you think could be worded better? It'd be nice to make all the changes in a single PR while we're at it.

raulsntos: - I've opened https://github.com/godotengine/godot/pull/80489.

mhilbrunner (approved):

mhilbrunner:
Merged.

PR #7786, opened 2023-08-07
Update script-based workaround for window focus with gamepads

mubinulhaque:
There was an error on the "Controllers, gamepads, and joysticks" page where the workaround for Window focus contained the method 'event_is_action_pressed()' and the method 'input_is_action_pressed()', both of which did the same thing and thus made the former redundant. I've fixed the first method to do what it was intended to do, reduced the number of lines needed for both methods and changed their names to be more readable for beginners.


AThousandShips:
You need to rebase your PR to `master` you created a branch based on the wrong branch

AThousandShips:
Superseeded by: #7795

PR #7790, opened 2023-08-09
Update class_nodepath.rst

lotuspar:
Fixes a code example with a diff. style to the other ones on the page


AThousandShips:

PR #7815, opened 2023-08-17
Point to GlobalScope from Evaluating Expressions tut

aXu-AP:
Update to 4.x since functions were moved from GDScript to GlobalScope.
https://docs.godotengine.org/en/4.1/tutorials/scripting/evaluating_expressions.html#built-in-functions


tutorials/scripting/evaluating_expressions.rst
```diff
@@ -198,7 +198,7 @@ The output from the script will be::
 Built-in functions
 ------------------
 
-Most methods available in the :ref:`class_@GDScript` scope are available in the
+All methods in the :ref:`class_@GlobalScope` are available in the
```
AThousandShips, line 201:
```suggestion
All methods in :ref:`class_@GlobalScope` are available in the
```
I'd that without the "scope" after it is more correct to just say "in X" as it is a name itself

aXu-AP: True, thanks for the tip. I like the second version better.

mhilbrunner (approved):

mhilbrunner:
Merged. Congrats on your first merged contribution! 🎉 Sorry this one fell through the cracks and was sitting here :) Thanks for contributing!

aXu-AP:
No worries, it happens when you have a lot on your plate 👍

mhilbrunner:
Cherry-picked to 4.1.

PR #7820, opened 2023-08-19
Custom editor icon section's command wrong or out of date

gogojokid:
4.1.1.stable

https://docs.godotengine.org/en/latest/community/contributing/content_guidelines.html

In the "*.gdextension" code block, the [Icon] tag does not work in 4.1.1.stable version, and it does not match the naming conventions of other tags. I test it, [icons] tag is worked well. In addition, I added a blank lines between the tag and the content to make it more consistent with other "*.gdextension" code block formatting.

Bugsquad edit: Fixes https://github.com/godotengine/godot-docs/issues/7885

### DIFF
--- a/tutorials/scripting/gdextension/gdextension_cpp_example.rst
+++ b/tutorials/scripting/gdextension/gdextension_cpp_example.rst
@@ -442,8 +442,9 @@ For example:
 
 .. code-block:: none
 
-    [Icon]
-    GDExample = "res://icons/GDExample.svg"
+    [icons]
+
+    GDExample = "res://icons/gd_example.svg"
 
 The path should point to a 16 by 16 pixel SVG image. Read the guide for :ref:`creating icons <doc_editor_icons>`
 for more information.


capnm (changes requested):
Maybe change the name of the icon, it's not obvious that it doesn't have to be the same, e.g. "res://icon_gdexample.svg".

paddy-exe (approved):
Looks good. Had a look at Yuri's implementation PR details and `[icons]` is correct. Thanks for the correction!

capnm:
> The standard recommended procedure is to have icons stored in such a directory, see all the modules in the engine

These icons should be somewhere in the external extension directory? It isn't a internal module or part of the project.
res://??/icons/gd_example.svg ?

(It took me about an hour to figure out why it didn't work ;)

AThousandShips:
Can you please update your commit message, it says "out of data"

YuriSizov:
Commit message looks correct to me 👀

AThousandShips:
"Custom editor icon section's command wrong or out of data." not "date"

YuriSizov:
Ah, I thought you were seeing "out of data" as the entire commit message for some reason. My bad.

AThousandShips:
The message still says "out of data"

gogojokid:
> The message still says "out of data"

I am sorry, I spent some time looking at Godot's PR process and didn't notice it, it will change soon.

capnm (approved):

capnm:
Update2: I suggest leaving a comment there (correct me, but this doesn't seem to be documented anywhere) and using a relative path:
```
[icons]

; No res: and leading slash means it is relative to this gdextension file.
GDExample = "icons/gd_example.svg"
```

AThousandShips:
I think this is outside the scope of this PR, which is fixing a simple out of date description, needs to be discussed and looked into

capnm:
I'm now 'out of data' :smile: and hope that  someone someday hits the commit button.

gogojokid:
I think we should use "res://" here to reduce confusion, because on this page and elsewhere in the documentation, "res://" is always explicitly used for relative paths. Since it is a relative path, it means that the user can specify it freely, and technically can also use the path on the upper layer of the addon without any restrictions. For example, if you only make GDExtensions for your own use, these extensions can all read icons from the recommended global path "res://icons" for sharing. But if publishing them, it would be better to encapsulate the icons in the "res://addon/your_extension" directory.

To make this clearer, some more explanation could be added, and an example added, or stated in the Plugins section, since there seems to be no other mention of addons in this section. However, this PR is indeed just a simple fixing (sorry I didn’t use “Fixes” to start the title because I was not familiar with the PR process before). I agree with AtousandShips, we can merge it as soon as possible so that new users will no longer get the wrong information. Perhaps capnm can submit an Issue or PR as a discussion of Documentation Enhancement?

dsnopek (approved):

mhilbrunner (approved):

mhilbrunner:
Merges. Thanks and congrats on your first merged contribution to Godot's docs!

mhilbrunner:
Cherry-picked to 4.1

PR #7821, opened 2023-08-19
Better explaining how mirroring works.

lvella:
I wrote everything I wish was written when I tried to make an infinite scrolling background, addressing every mistake I made.


AThousandShips:

PR #7825, opened 2023-08-20
Update singletons_autoload.rst

Hugo-Trentesaux:
fix scene name to match zip template

### DIFF
--- a/tutorials/scripting/singletons_autoload.rst
+++ b/tutorials/scripting/singletons_autoload.rst
@@ -262,7 +262,7 @@ Finally, we need to fill the empty callback functions in the two scenes:
     private void OnButtonPressed()
     {
         var global = GetNode<Global>("/root/Global");
-        global.GotoScene("res://scene_2.tscn");
+        global.GotoScene("res://Scene2.tscn");
     }
 
 and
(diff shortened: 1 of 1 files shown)


raulsntos:
The template seems to be a GDScript project, in which case I believe the [naming convention](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html#file-names) is to use snake_case so maybe the [template](https://github.com/godotengine/godot-docs-project-starters/tree/master/src/singleton_autoload_starter) should be updated instead.

That said, the naming convention in C# is to use PascalCase for all files and this PR only updates the C# code, so maybe we should make this change regardless of how the files are named in the template.

Piralein:
Both should be changed, the files in the template project should be all `snake_case` and the C# version in the docs should be `PascalCase`.
(In my opinion)

Piralein (approved):

mhilbrunner (approved):

mhilbrunner:
Sorry this has been sitting for a bit. I agree with the discussion above, we should fix this regardless of the GDScript template (but we should likely fix that too :))

Merged. Thanks for contributing, and congrats on your first merged contribution to Godot's docs!

mhilbrunner:
Cherry-picked to 4.1

PR #7826, opened 2023-08-21
Patch lobby example implementation.

Jordyfel:
Most importantly, fix line 350 where I was erasing from the wrong dictionary. Whoops.

Also multiline some comments so annoying scrollbars don't appear in the webpage, and use `String.is_empty()` instead of comparing to an empty string, since it should be faster.


mhilbrunner:
Thanks! Merged. Will try to cherry-pick soonish if nothing crops up. Thanks for keeping up with maintainig this :)

mhilbrunner:
Cherry-picked to 4.1

Issue #7763, opened 2023-08-02
Incorrect/misleading explanation of duct typing and dynamic typing in the static typing page.


hilfazer:
**Your Godot version:**
4.1.1 stable
doc page is up to date with Godot 4.1

**Issue description:**

> GDScript supports [duck-typing](https://stackoverflow.com/a/4205163/8125343), so even if your timer is of type Timer, it is also a Node and an Object, two classes it extends.

That's not what duct typing is about, that's what inheritance does - it is the IS-A relationship.
https://en.wikipedia.org/wiki/Is-a

> With dynamic GDScript, you also don't care about the node's type as long as it has the methods you need to call.

Now that is what duck typing (not the dynamic typing) is about.

**URL to the documentation page:**
https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/static_typing.html

PR #7845, opened 2023-08-27
Fix incorrect/misleading explanation of duck typing and dynamic typing in the static typing page

mateuseap:
## What I did 

- Adjusted a incorrect/misleading explanation of duck typing and dynamic typing in the static typing page. I based myself on what is written on 'Accessing data or logic from an object' section of [`godot_interfaces.rst`]([/tutorials/vest_practices/godot_interfaces.rst] ).

#### Fixes: #7763

### DIFF
--- a/tutorials/scripting/gdscript/static_typing.rst
+++ b/tutorials/scripting/gdscript/static_typing.rst
@@ -188,10 +188,10 @@ an error or not at runtime.
 This happens when you get a child node. Let's take a timer for example:
 with dynamic code, you can get the node with ``$Timer``. GDScript
 supports `duck-typing <https://stackoverflow.com/a/4205163/8125343>`__,
-so even if your timer is of type ``Timer``, it is also a ``Node`` and an
-``Object``, two classes it extends. With dynamic GDScript, you also
-don't care about the node's type as long as it has the methods you need
-to call.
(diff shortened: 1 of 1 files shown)


tutorials/scripting/gdscript/static_typing.rst
```diff
@@ -188,10 +188,10 @@ an error or not at runtime.
 This happens when you get a child node. Let's take a timer for example:
 with dynamic code, you can get the node with ``$Timer``. GDScript
 supports `duck-typing <https://stackoverflow.com/a/4205163/8125343>`__,
-so even if your timer is of type ``Timer``, it is also a ``Node`` and an
-``Object``, two classes it extends. With dynamic GDScript, you also
-don't care about the node's type as long as it has the methods you need
-to call.
+so even if your timer is of type ``Timer``, Godot doesn't validate wheter it
```
YuriSizov, line 191:
```suggestion
so even if your timer is of type ``Timer``, Godot doesn't validate whether it
```

YuriSizov:
I think this needs to be rephrased even further. Also the currently proposed change makes it pointless to use `Timer` or any other specific type as an example. I think instead of removing the mention of base types they should be incorporated in a different way that supports the point about duck typing.

PR #7850, opened 2023-08-28
Update exporting to Android XR devices for new loader

BastiaanOlij:
In Godot 4.2 the way we include the loader has changed. This updates the documentation to reflect this change.

Note that as we aren't able to include the new version in the asset library until 4.2 is released that part of the docs needed clarification. I didn't want to remove the section on obtaining it from the asset library as this will become valid once 4.2 is released.


mhilbrunner (approved):

mhilbrunner:
Merged, thank you!

Issue #7852, opened 2023-08-28
MeshDataTool.get_face_vertex() has a unclear description.


kravohi:
**Your Godot version:**
Godot v4.0.2

**Issue description:**
The function description for `MeshDataTool.get_face_vertex()` is unclear as to the data type that is returned.

> Returns the specified vertex of the given face.
> Vertex argument must be either 0, 1, or 2 because faces contain three vertices.

This is a bit confusing since the function returns the vertex index.
May I suggest something like this?

> Returns the specified vertex index of the given face.
> Vertex argument must be either 0, 1, or 2 because faces contain three vertices.

As a side note, it would make more sense if this function returned a vertex (Vector3) instead of a vertex index (Int).
Maybe that's just me.

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/4.0/classes/class_meshdatatool.html

clayjohn:
I think your suggestion is good. A code example would work well here too:
```
var index = mesh_data_tool.get_face_vertex(0, 1) # Gets the index of the second vertex of the first face.
var position = mesh_data_tool.get_vertex(index)
var normal = mesh_data_tool.get_vertex_normal(index)
```

> As a side note, it would make more sense if this function returned a vertex (Vector3) instead of a vertex index (Int).
Maybe that's just me.

I don't think so. As you need the index to access all other vertex attributes as well (normal, uv, etc.)

mateuseap:
Done! I've added the new description suggested by @kravohi and the code snippet suggested by @clayjohn, now I think the method reference is quite good.

PR #7854, opened 2023-08-28
Improve `MeshDataTool.get_face_vertex` method description

mateuseap:
## What I did 

- Improved the `MeshDataTool.get_face_vertex` method description by making it more clear about what is the data type that is returned and adding a code example.

#### Fixes: #7852


AThousandShips:

Issue #7796, opened 2023-08-10
Alert about the risks of using SceneTree.create_tween() instead of Node.create_tween()


fguillen:
**Your Godot version:**
4.1.1

**Issue description:**
When create a tween by `SceneTree.create_tween()` the Tween will be created outside the Node. So when the Node is freed the Tween remains alive. As when creating a Tween the more usual is to manipulate the Node when it is created there are many chances that using `SceneTree.create_tween()` we will end up in an error situation:

![Screenshot 2023-08-10 at 22 38 06](https://github.com/godotengine/godot-docs/assets/23043/9766f74f-f435-4618-a706-bbb49604bdac)

I suggest adding a warning note:

> A Tween can be created by using either SceneTree.create_tween or Node.create_tween. Tweens created manually (i.e. by using Tween.new()) are invalid and can't be used for tweening values. 

Add: 

Beware that when creating the Tween by using SceneTree.create_tween the Tween will be detached from the actual Node and it won't be killed when the Node is freed. Maybe you will prefer to use Node.create_tween if you want the Tween to be automatically killed when the Node is freed.

Or something like this. If this would be there it would have saved me[ so many headaches](https://godotforums.org/d/35636-using-scenetreecreate-tween-is-very-danger) :)

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/stable/classes/class_tween.html#description

mateuseap:
@fguillen I've added your alert suggestion to `Tween` class description, I found it quite useful ;)

PR #7855, opened 2023-08-28
Add an alert in `Tween` class description

mateuseap:
## What I did 

- I've added an alert in `Tween` class description section talking about the risks of using `SceneTree.create_tween` method instead of `Node.create_tween` when creating a `Tween`, documentation page: https://docs.godotengine.org/en/stable/classes/class_tween.html#description.

#### Fixes: #7796


AThousandShips:

mateuseap:
Thanks @AThousandShips, I'll check it out the contributing guide for class reference.

PR #7806, opened 2023-08-14
Update key_concepts_overview.rst

VictorDev200:
Changing 'Character' to 'Player'

PR #7856, opened 2023-08-28
Update key_concepts_overview.rst

abrahamprz:
Changed how the node is called. The documentation said "Character" but the image showed "Player".


Piralein:
This change was already done by #7806 in master, it will be cherry-picked to the 4.1 branch in some time.
No need to do it individual for the 4.1 branch, please always target master if the change is not related to a specfic version.

Thank you for contribution, nevertheless!
Closing as duplicate.

PR #7863, opened 2023-08-30
Use relative font size for `code` elements

dalexeev:
Fixes too small font size in headers ([example](https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_documentation_comments.html)). The size in the text hasn't changed (at least for the font on my OS).


mhilbrunner (approved):

mhilbrunner:
Makes sense, thanks. Merged.

PR #7866, opened 2023-08-31
fixed typo

Luke9389:
I'm pretty sure this was a typo. 

Would love another 👀 to verify.


YuriSizov:
Hey, thanks for opening a PR! The sample is correct and it is explained in this very section why it is correct. `assert` makes sure that the statement in the first argument is true. If it's not it fails. This examples shows how to add an options message _if the assertion fails_.

So we want to make sure that `enemy_power` is below `256`, and we print `Enemy is too powerful!` is it's not.

Luke9389:
I see! Thanks for taking the time to check this out.  🙇

PR #7877, opened 2023-09-04
Update introducing_xr_tools.rst

rhiskey:
Added one step to reload UserSettings and reenable plug-in

### DIFF
--- a/tutorials/xr/introducing_xr_tools.rst
+++ b/tutorials/xr/introducing_xr_tools.rst
@@ -42,6 +42,11 @@ Now go to the ``Plugins`` tab and enable the plugin.
 
 .. image:: img/godot_xr_tools_enable.webp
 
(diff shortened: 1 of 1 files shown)


BastiaanOlij:
Once the plugin is enabled, the autoload entry is added automatically. The screenshot is actually wrong, if you add `UserSettings` a second time you end up duplicating things. Only the `XRToolsUserSettings` entry created by the plugin should be there.

What is often required however is a restart, with the plugin not being automatically enabled when installed, scripts aren't loaded properly and Godot doesn't recover well from this. This has however improved in 4.1 and is improving further as more things get tightened up.

All in all the only thing that may be needed in the documentation is a simple remark that after enabling the plugin, you should restart Godot.

AThousandShips:

PR #7861, opened 2023-08-30
Update key_concepts_overview.rst

thomasdye:
Character should be Player, according to the picture. Reordered Camera2D, Sprite2D, and CollisionShape2D to be the same order as the picture.

PR #7878, opened 2023-09-05
Fix doc 

carlhester:
The text describing the Node screenshot is incorrect.  The screenshot shows a node named Player, not Character.


AThousandShips:
Note that this is already worked on:
* https://github.com/godotengine/godot-docs/pull/7861

PR #7880, opened 2023-09-05
Compiling for x11: use `linker=lld`

MJacred:
`use_lld` boolean option has been replaced by `linker=<default|bfd|gold|lld|mold>`

Stable is already up-to-date and mentions `linker`: https://docs.godotengine.org/en/stable/contributing/development/compiling/compiling_for_linuxbsd.html#using-clang-and-lld-for-faster-development

### DIFF
--- a/development/compiling/compiling_for_x11.rst
+++ b/development/compiling/compiling_for_x11.rst
@@ -225,7 +225,7 @@ the default GCC + GNU ld setup:
 To do so, install Clang and the ``lld`` package from your distribution's package manager
 then use the following SCons command::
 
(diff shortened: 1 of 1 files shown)


akien-mga (approved):
That's correct. Should also be cherry-picked to `3.5`, but not earlier versions. 4.0+ already correct.

mhilbrunner (approved):

mhilbrunner:
Thanks for noticing and contributing, merged. Will cherry-pick to 3.5 with the next batch of cherry-picks.

PR #7882, opened 2023-09-05
Update webrtc.rst

peterlindgren:
Updated gdscript examples to match the godot-demo-projects repository code: https://github.com/godotengine/godot-demo-projects/tree/4.0/networking/webrtc_minimal

PR #7887, opened 2023-09-07
Update WebRTC docs to 4.x

greenfox1505:
I don't know if other issues exist with this doc, but at minimum, this helps these two examples run in Godot4.


AThousandShips:
Work is already being done in:
* https://github.com/godotengine/godot-docs/pull/7882

PR #7889, opened 2023-09-07
Remove UWP mentions as the platform port was removed in 4.2

akien-mga:
Draft for now, as we should discuss what to do with the "Compiling for UWP" and "Exporting for UWP" pages.

They were already stubs clarifying that this is not supported yet. We may want to keep them to clarify that the platform support was fully dropped for 4.x, and avoid having 404 errors.
At the same time, how long do we want to keep mentions of removed platform support in the docs?

Shouldn't be cherry-picked.


mhilbrunner:
> At the same time, how long do we want to keep mentions of removed platform support in the docs?

If it is on a single "Supported platforms" section, I'm undecided whether it hurts to list discontinued platforms with the versions where we first dropped support or that last supported them.

Theoretically, we could redirect the "Compiling for UWP" and "Exporting for UWP" pages there then as an explanation.
I'm also fine with keeping them around as stubs for now just saying UWP was dropped, but eventually we should maybe just remove those, seeing as UWP was never supported in 4.x in the first place.

mhilbrunner (approved):
Besides the above discussion the changes look good to me once we come to a decision. :)

akien-mga:
Added a note in the FAQ about the UWP removal.

We can maybe add redirects that point to that section of the FAQ? Can you handle this @mhilbrunner?

Calinou (approved):

mhilbrunner (approved):

mhilbrunner:
Will do. Thanks!

PR #7891, opened 2023-09-07
Grammar fix

Dacromir:


AThousandShips:

AThousandShips:
What in your opinion is wrong with this? It's a grammatically correct sentence? It says that if the first statement isn't true then the second is

mhilbrunner:
Hey! Thanks for taking the time to contribute. I agree with the above that this is currently fine as is though :)

PR #7907, opened 2023-09-13
Update physics_introduction.rst

### LINKED ISSUE #7906: Add `look_at` alternative documentation for RigidBody2D
**Your Godot version:** 4.1.1

**Issue description:**

RigidBody has helpful documentation on how to implement a workaround for the `look_at` method since this method does not work for RigidBody nodes. This documentation is linked below.

https://docs.godotengine.org/en/stable/tutorials/physics/rigid_body.html#the-look-at-method

Unfortunately, I needed a solution to get `look_at` functionality for RigidBody2D, but could not find a similar workaround in the docs. It would be helpful to have an alternative (probably simpler) implementation documented for 2D space.

**URL to the documentation page (if already existing):** https://docs.godotengine.org/en/stable/tutorials/physics/physics_introduction.html#using-rigidbody2d

beamer159:
Add `look_at` workaround method documentation for RigidBody2D.

This PR resolves #7906. It is helpful and consistent to have a corresponding documented workaround for implementing `look_at` for RigidBody2D nodes, since there is currently documentation of the same kind for RigidBody.


nlupugla:
It would be nice to add an explanation for why this is necessary, even if it's just a link to or repeat of the explanation for 3D.

beamer159:
I thought I added an explanation for why this is needed, which is essentially a repeat of the explanation for 3D.

> Using the Node2D's ``look_at()`` method can't be used by a RigidBody2D each frame to follow a target. Here is a custom ``look_at()`` method that will work reliably with rigid bodies

PR #7920, opened 2023-09-15
Update set/get for pose position/rotation/scale on Skeleton3D Class Reference

nathanjf:
Wrote docs for:

get_bone_pose_position
get_bone_pose_rotation
get_bone_pose_scale
set_bone_pose_position
set_bone_pose_rotation
set_bone_pose_scale


AThousandShips:

PR #7925, opened 2023-09-15
Rigidbody2D to Mob cast type error  when using 'mob' as variable name

JScavello:
In C# using mob as the variable name for the Mob causes a cast type error going Rigidbody2D to Mob. Changing the variable to something other than 'mob' solves this issue.

Godot 4.1.1
.NET SDK 6.0.414


AThousandShips:

raulsntos:
> In C# using mob as the variable name for the Mob causes a cast type error going Rigidbody2D to Mob. Changing the variable to something other than 'mob' solves this issue.

The name of the variable can be anything and doesn't affect casting. What's the error you are getting?

JScavello:
Reverted changes in my build back to 'mob' variable to gather accurate error messages, and everything builds now and monsters are spawning and decaying correctly with no more error messages.

PR #7944, opened 2023-09-17
Changed Input. to event. in the event_is_action_pressed() example function

FaintLocket424:
The function event_is_action_pressed() within the Windows focus handling section was using the same code from the above function input_is_action_pressed and was not using the event parameter which is was made for.

The function is designed to replace event.is_action_pressed() to handle a defocused window and controller inputs.

I believe this is a simple typo.


AThousandShips (approved):
Correct both functionally and in line with the description

Sauermann (approved):
Using `event` is more appropriate here.

mhilbrunner (approved):

mhilbrunner:
Merged, thanks and congrats on your first merged contribution to Godot's docs! 🎉

mhilbrunner:
Cherry-picked to 4.1

PR #7950, opened 2023-09-18
updating scripting_first_script.rst

yarnthen:
updated c# script for Hello World Example

### DIFF
--- a/getting_started/step_by_step/scripting_first_script.rst
+++ b/getting_started/step_by_step/scripting_first_script.rst
@@ -143,10 +143,10 @@ Add the following code to your script:
 
  .. code-tab:: csharp C#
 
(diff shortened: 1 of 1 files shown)


AThousandShips (changes requested):
This is not correct, the code is the same as the GDScript code, and there's no `_ready` method in C#

The C# equivalent of `_init` is the constructor of a class

yarnthen:
please see example in 
https://github.com/godotengine/godot-demo-projects/blob/master/mono/pong/Logic/Ball.cs

_ready in c#. Specifically this is done in Godot 4.1.1 stable version

AThousandShips:
But the GDScript version has `_init` which is the equivalent of the constructor in C#

Issue #6911, opened 2023-03-06
Visual Studio integration is unclear and outdated


Trider12:
**Godot version: 4.0**

**[URL to the documentation page](https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_basics.html#configuring-an-external-editor)**

**Issue description:**
1. Now that Godot targets .Net 6.0, VS 2019 is not supported and should be removed from the available editors list.
2. The docs say to install `Mobile development with .NET` and `.NET Core cross-platform development`, but both of them are now out of support and, I believe, are not required due to the transition to .Net 6.0 (I may be wrong here).
![image](https://user-images.githubusercontent.com/47952308/223108483-fc4f93fd-41d0-4249-9eaf-9b3a7f90cf9d.png)

4. The docs mention the VS [extension](https://github.com/godotengine/godot-csharp-visualstudio) which (has been barely working for 3.x) is [obsolete](https://github.com/godotengine/godot-csharp-visualstudio/issues/42#issuecomment-1250411467) for 4.x.
5. There is no information on how to actually debug the game with VS. There's this [hack](https://www.reddit.com/r/GodotCSharp/comments/xgpqfh/oc_rundebug_godot4_c_projects_from_visual_studio/), but it is a hack, has its issues and doesn't look like a proper solution. If it is the proper solution, then it needs to be documented.

silenuz:
In Visual Studio 2022 it's pretty easy.  Right click the project, choose properties.

Under properties select Debug, and then click the link to open the Debug launch profiles dialog.

Delete the current profile, and add an executable profile.  Set the executable path to point to the main Godot exe.

Set arguments to be --path . (or --path "/absolute/path/to/project/folder")  Note the quotes.

Enable native debugging and thats it.  Should be able to now launch a debug session with Visual Studio.

If you want to also be able to debug editor tools add another profile and add --editor to the argument line.

![0-Visual-studio-debug](https://user-images.githubusercontent.com/13521146/223510644-d395f966-90aa-4361-ad36-4c8cdef9ca17.gif)

Trider12:
@silenuz Yeah, I did exactly that (even in VS2019 using this [hack](https://learn.microsoft.com/en-us/answers/questions/620284/how-to-target-net-6-0-with-visual-studio-2019)), but my point is that this needs to be in the documentation, not on Reddit/GitHub.

silenuz:
[...]

PR #7962, opened 2023-09-19
Remove outdated VS setup and 2019 support info

31:
Removes VS 2019 from the list of supported IDEs. It doesn't support .NET SDK 6.0 and can't build Godot projects. https://learn.microsoft.com/en-us/visualstudio/releases/2019/compatibility#-visual-studio-2019-support-for-net-development

Removes mention of the VS extension. It only applies to 3.x projects and debugging is done with ordinary debugging tools after the move to .NET in 4.x.

* Related: https://github.com/godotengine/godot-docs/issues/6911 (but I only saw this issue after preparing this change)

IMO this page should also include debugger setup. It is more normal than it was before, but it's still more involved than *just* hitting F5. I don't have VS and don't have the time to write this up though.

This should be applied to the 4.x branches.

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_basics.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_basics.rst
@@ -77,7 +77,6 @@ click on **Editor → Editor Settings** and scroll down to
 external editor of choice. Godot currently supports the following
 external editors:
 
-- Visual Studio 2019
 - Visual Studio 2022
 - Visual Studio Code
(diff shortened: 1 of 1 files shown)


skyace65 (approved):

mhilbrunner (approved):

mhilbrunner:
Thank you! Merged.

mhilbrunner:
Cherry-picked to 4.1

PR #7969, opened 2023-09-20
Recommend using fullscreen on Linux in Fixing jitter and stutter

Calinou:
Stuttering is more common on Linux when playing in windowed mode, regardless of how powerful the hardware is. This can occur even if compositing is disabled in the window manager (as tested on KWin).

- See https://github.com/godotengine/godot/issues/81908.


mhilbrunner (approved):

mhilbrunner:
Thanks!

mhilbrunner:
Cherry-picked to 4.1

PR #7976, opened 2023-09-21
 Update mouse_and_input_coordinates.rst

profande:
This page currently refers to ``event.speed`` which is a field that no longer exists on InputEventMouseMotion and has since been renamed to ``event.velocity``


AThousandShips:

profande:
Gotcha, will open new request on ``master``.

PR #7984, opened 2023-09-21
Update complying_with_licenses.rst

JohnVeness:
Moved the tip regarding how to get license information in the Engine singleton to after the list of inclusion methods, as it read oddly saying "These are the most common approaches" and then the reader having to scroll quite a bit to see the list of approaches. Also removed engine contributor methods, as seemed unnecessary and potentially confusing.

On the other hand, moved the tip about custom builds up higher, to put it next to, and gave it equal prominence with, the info about default templates.

Finally, some minor wording changes, for grammar, for clarity, or to remove redundancy.

### DIFF
--- a/about/complying_with_licenses.rst
+++ b/about/complying_with_licenses.rst
@@ -8,7 +8,7 @@ What are licenses?
 
 Godot is created and distributed under the `MIT License <https://opensource.org/licenses/MIT>`_.
 It doesn't have a sole owner either, as every contributor that submits code to
-the project does it under this same license and keeps ownership of the
+the project does it under this same license and keeps ownership of their
 contribution.
 
 The license is the legal requirement for you (or your company) to use and
@@ -24,7 +24,7 @@ with the original one.
 
 .. warning::
 
-    In your project's credits screen, remember to also list third-party notices
+    Alongside the Godot license text, remember to also list third-party notices
     for assets you're using, such as textures, models, sounds, music and fonts.
 
     Free assets in particular often come with licenses that require attribution.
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):
These seem like nice small improvements to that page, mostly in structure and wording.

I agree with most of them; removing the info how to `.get_author_info()` and `.get_donor_info()` I understand from the point of view of this specific page, I think it was mostly mentioned here for completeness as it may not be documented elsewhere, and you may want to build (some) of these into your game's credits screen or something. I can see reasons to remove that info or leave it on the page, not sure.

Overall seems good. I'd like a second pair of eyes though, considering its a pretty legally important page, cc @YuriSizov or @akien-mga

about/complying_with_licenses.rst
```diff
@@ -58,31 +58,9 @@ This text reads as follows:
 Inclusion
 ---------
 
-The license does not specify how it has to be included, so anything is valid as
+The license does not specify how it has to be included, as
```
clayjohn, line 61: The meaning of this sentence has changed. Previously it said that "anything was valid" if the condition is met. Now it says that "the licence does not specify how it has to be included" if the condition is met. Which is clearly incorrect. 

The licence doesn't specify how it has to be included, therefore users are free to do anything they choose as long as the licence can be displayed under some condition

JohnVeness: Ah yes, I see what you mean. How about:

The license text must be made available to the user. The license doesn't specify how it has to be included, but here are the most common approaches (you only need to implement one of them, not all).

akien-mga:
This looks pretty good to me. I've been reworking this page today before noticing this PR, so I went back and worked these changes in my copy. Please re-review :)

Superseded by #8513.

PR #7998, opened 2023-09-22
Fix Player Node Type in C# Script of 3D Tutorial (05.spawning_mobs.rst)

chrlee:
Currently, the C# code example of the [spawning mobs page in the "Your First 3D Game"](https://docs.godotengine.org/en/stable/getting_started/first_3d_game/05.spawning_mobs.html) tutorial does not build due to trying to get a node of type `Player` on a `CharacterBody3D` causing the error: `The type or namespace name 'Player' could not be found`

### DIFF
--- a/getting_started/first_3d_game/05.spawning_mobs.rst
+++ b/getting_started/first_3d_game/05.spawning_mobs.rst
@@ -256,7 +256,7 @@ Let's code the mob spawning logic. We're going to:
         // And give it a random offset.
         mobSpawnLocation.ProgressRatio = GD.Randf();
 
-        Vector3 playerPosition = GetNode<Player>("Player").Position;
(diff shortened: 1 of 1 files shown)


AThousandShips:

AThousandShips:
Are you sure you aren't having some errors in your compilation? Because if it works for `Mob` then it must work for `Player` right?

raulsntos:
The Player class is created in [_Moving the player with code_](https://docs.godotengine.org/en/stable/getting_started/first_3d_game/03.player_movement_code.html). Did you perhaps skip this step? Make sure the class name is `Player` with an uppercase `P` and that the file that declares it is named `Player.cs`.

chrlee:
Learning pains, my mistake :)

PR #8000, opened 2023-09-22
Fix minor typo in 04.mob_scene.rst

gmlewis:
As far as I can tell, when working through the `Getting Started > Your first 3D game > Contents > Designing the mob scene > Leaving the screen` [section](https://docs.godotengine.org/en/stable/getting_started/first_3d_game/04.mob_scene.html#leaving-the-screen), the only place I can find the `VisibleOnScreenNotifier3D node` is in the main panel on the left-hand side of the user interface. Therefore, I believe this is a typo that can be fixed. Thanks.


AThousandShips (changes requested):
This is incorrect, it does not refer to the scene tree, but the Node dock, the signals:

image

Issue #6596, opened 2023-01-12
Can we have C# Api Docs at docs.godotengine.org?


JupiterRider:
There are Tools to generate HTML from C# Code. One of these tools is for example DocFX:
https://github.com/dotnet/docfx

Can we have this at the Godot Page to Provide C# Api Docs as well? Currently there is GDScript only:
https://docs.godotengine.org/en/stable/classes/index.html

paulloz:
> I don't think we should make a dedicated build of the class reference only for C#. Instead, I'd perform live conversion of snake_case syntax to PascalCase using JavaScript: #4406

That solution won't cover 100% of the C# API tho. Parts of it aren't just case conversion.  
FWIW I spent a decent amount of time speaking to people here and there (whether on GH, Twitter, Discord and Rocket) about how we could bring a decent C# doc to the userbase, and it never went anywhere. Probably partly on me for getting tired of it not going anywhere and stopping to push for it.

YuriSizov:
> That solution won't cover 100% of the C# API tho. Parts of it aren't just case conversion.

I think this speaks to the idea of having a separate document, if that's needed. Wherever 2nd and 3rd party languages don't map to the engine API it would be very hard to provide the same level of coverage and quality without duplicating everything.

That said, the class reference is not language specific, it's a mirror of the engine API itself. GDScript happens to use that API without any transformations, and C# happens to have different conventions than our ClassDB. We already provide examples in both languages in the docs. For the purpose of lookup, we can rely on the conversion and maybe on the keyword system that we've been discussing for the class reference.

paulloz:
> Wherever 2nd and 3rd party languages don't map to the engine API it would be very hard to provide the same level of coverage and quality without duplicating everything.

Well, that's then the same discussion as ever. Where do we put C# on the "this is unofficial" to "this is official" line. It's not an easy thing to explain to most newcomers who get in because C# is part of the official communication and at the same time they can't find a basic classdoc reflecting the API.

YuriSizov:
> Well, that's then the same discussion as ever. Where do we put C# on the "this is unofficial" to "this is official" line. It's not an easy thing to explain to most newcomers who get in because C# is part of the official communication and at the same time they can't find a basic classdoc reflecting the API.

[...]

PR #8024, opened 2023-09-23
C#: Document potentially confusing built-in float vs C# float/double

### LINKED ISSUE #7975: Place notes about Doubles for C# users
**Your Godot version:**4.1-Mono

**Issue description:**
In the Method declaration, `float` is the type listed for the type of delta for both the `_process` and `_physics_process` method. I discovered the issue when writing C# code for the `_Process` C# binding. It uses the type `double` for delta in the C# binding. I assumed they are of the same type at first glance. It wouldn't occur normally for me to click on `float` because I know what a float is. There should have been a note about this for C# users that wouldn't think about checking the documentation for the `float` type. Yes, while it does list in the documentation about `float`s; it's inconvenient. 

**URL to the documentation page:** https://docs.godotengine.org/en/stable/classes/class_node.html#class-node-method-physics-process

If you know how to fix the issue you are reporting please
consider opening a pull request. We provide a tutorial on
using git here: https://docs.godotengine.org/en/stable/community/contributing/pr_workflow.html,
writing documentation at https://docs.godotengine.org/en/stable/community/contributing/docs_writing_guidelines.html
and contributing to the class reference here: https://docs.godotengine.org/en/stable/community/contributing/updating_the_class_reference.html

jynus:
Current Godot documentation includes numerous usages of the built-in `float` data type on its reference manual. While it is perfectly documented, it would be great to add an additional reference on the "C# API differences to GDScript" section.

Fixes #7975

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_differences.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_differences.rst
@@ -342,6 +342,29 @@ Example:
 
     Input.Singleton.JoyConnectionChanged += Input_JoyConnectionChanged;
 
+Float
+-----
+
+Some engine functions return or have as input parameters values of the built-in
+:ref:`float <class_float>` type. This type is a 64-bit double-precision floating-point
+number, equivalent to ``double`` in C# (`System.Double` for .NET).
+
+This can cause confusion when working with C# scripts because the functions will appear
+on documentation as using or returning a Godot's ``float``, but that will require an 8-byte
+C# ``double`` or a conversion to or from another type, such as a 4-byte C# ``float``.
+
+Example:
+
+:ref:`Node._process() <class_Node_method__process>` virtual method receives as an argument
+the ``delta`` time since the last frame, in seconds. This is a built-in :ref:`float <class_float>`
+type, but on C# you should work with the following method signature:
+
+.. code-block:: csharp
+
+    public override void _Process(double delta)
+    {
+    }
+
 String
 ------
 


raulsntos:
The class reference uses GDScript types so you get `int` and `float` everywhere and those are 64-bit in GDScript but the 64-bit types have a different name in other languages (in C# the 32-bit types are `int` and `float`, and the 64-bit types are `long` and `double`).

But this doesn't mean that every time a `float` appears in the class reference, it means `double` in C#. There is metadata in some methods that indicates if the type should be 32-bit vs 64-bit or signed vs unsigned.

For example, take a look at this issue: https://github.com/godotengine/godot/issues/74452. The class reference uses `int` because GDScript doesn't have unsigned types, but in C# the type used is `uint` (32-bit unsigned integer).

Sorry for the long wall of text, my point is that a blanket statement where we say `GDScript float = C# double` is not always true, it's a bit more complicated. Also, it doesn't only apply to floating point numbers, but also integers. I think if we can convey all that nuance, it's not a bad thing to have in the documentation, although I personally think users would still miss it.

tutorials/scripting/c_sharp/c_sharp_differences.rst
```diff
@@ -342,6 +342,29 @@ Example:
 
     Input.Singleton.JoyConnectionChanged += Input_JoyConnectionChanged;
 
+Float
+-----
+
+Some engine functions return or have as input parameters values of the built-in
+:ref:`float <class_float>` type. This type is a 64-bit double-precision floating-point
+number, equivalent to ``double`` in C# (`System.Double` for .Net).
```
raulsntos, line 350: The [.NET brand guidelines](https://github.com/dotnet/brand/blob/main/dotnet-brand-guidelines.pdf) say the name is all uppercase.

```suggestion
number, equivalent to ``double`` in C# (`System.Double` for .NET).
```

jynus:
> Sorry for the long wall of text, my point is that a blanket statement where we say `GDScript float = C# double` is not always true, it's a bit more complicated. Also, it doesn't only apply to floating point numbers, but also integers. I think if we can convey all that nuance, it's not a bad thing to have in the documentation, although I personally think users would still miss it.

Thank you, you are completely right. In your opinion, do you think it is going to be extremely hard to transmit all of this nuance in a relatively short text, or should I still try to amend and give it a try? Given this no longer only applies to floats, it may need a section in the api differences doc for "Engine/C++ vs GDScript vs C# data types".

A proper (fully complete) fix may only happen if in the future there is C# -only reference- but my intention was at least to briefly describe this issue. I am not super invested in it happening right now, though. :-D

jynus:
I think I am going to give up on this- the type issue is somewhat explained on https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_variant.html (including Float's variant as double), and there is no good, short explanation for this.

raulsntos:
> A proper (fully complete) fix may only happen if in the future there is C#-only reference

There is a discussion about this in https://github.com/godotengine/godot-docs/issues/6596 if you are interested.

> Do most IDEs provide useful hints about this?

C# IDEs show the type information, and using autocomplete to add a method override will use the right type.

> the type issue is somewhat explained on [docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_variant.html](https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_variant.html) (including Float's variant as double), and there is no good, short explanation for this.

I think the problem is that users see `int` and `float` in the class reference and assume they refer to the 32-bit C# types because they have the same name, so they'll miss any notes we have in the _C# Variant_ page or any notes we add to the _C# API differences to GDScript_ page.

Thanks for the contribution nonetheless!

PR #8029, opened 2023-09-24
A few small corrections, mostly typos

ChilyVanily:


contributing/workflow/pr_workflow.rst
```diff
@@ -436,7 +436,7 @@ The interactive rebase
 
 If you didn't follow the above steps closely to *amend* changes into a commit
 instead of creating fixup commits, or if you authored your changes without being
-aware of our workflow and Git usage tips, reviewers might request of your to
+aware of our workflow and Git usage tips, reviewers might request of you to
```
AThousandShips, line 439:
```suggestion
aware of our workflow and Git usage tips, reviewers might request that you
```

classes/class_variant.rst
AThousandShips, line 1: These changes needs to be done over [here](https://github.com/godotengine/godot), changing the generated files is not allowed, see [here](https://docs.godotengine.org/en/latest/contributing/documentation/updating_the_class_reference.html)

contributing/workflow/pr_workflow.rst
AThousandShips: This is still valid, only the `classes/` files are limited

PR #8046, opened 2023-09-25
Remove the documentation of an unimplemented feature of the tilemap selection tool

ershn:
The documentation states that right clicking with the tilemap selection tool removes the selection.
However, when you do so in Godot 4.1.1 nothing happens. A right click behaves exactly the same as a left click.

I initially opened an issue on the engine repository thinking it was a bug, but after talking with contributors it seems more likely that the documentation is wrong.
Removing tiles with the selection tool is not very intuitive and does indeed seem a bit weird.

https://github.com/godotengine/godot/issues/81933


AThousandShips:
Should of course only be merged when the actual feature is implemented, as the PR has not been merged

mhilbrunner:
Thanks! Merged.

mhilbrunner:
Cherry-picked to 4.1

PR #8069, opened 2023-09-27
Add instructions on sky shader optimization

kb173:
We had performance issues with our sky shader and did some research and testing. In the [announcement of sky shaders](https://godotengine.org/article/custom-sky-shaders-godot-4-0/), we found this note:

> If the sky is especially smooth, you can even render the entire thing to the radiance cubemap and then read from the cubemap when drawing to the screen.

We figured out that this optimization is really simple, but extremely valuable: it gives much finer control over the trade-off between visual fidelity and performance of the sky shader. With static skies, it can even completely eliminate the need to re-render the sky.

This optimization is probably a necessity for complex sky shaders, and since it was mentioned in their announcement, it seems to be an intended usage and not some optimization hack. I hope this addition to the docs helps others who struggle with optimizing their sky shader.


mhilbrunner (approved):

mhilbrunner:

PR #8072, opened 2023-09-27
Add C# code to random_number_generation and fix shuffle bag example.

ssBandit:
Changes made to shuffle bag example:
- Replace .empty() with .is_empty()
- More descriptive comment
- Changed code-block into a tabbed code block


tutorials/math/random_number_generation.rst
```diff
@@ -409,17 +440,54 @@ ends up empty. When that happens, you reinitialize it to its default value::
...
+            _fruits.Shuffle();
+        }
+
+        // Get a random fruit, since we shuffled the array,
+        var randomFruit = _fruits[0];
+        // and remove it from the `_fruits` array.
+        _fruits.RemoveAt(0);
+        // Returns "apple", "orange", "pear", or "banana" every time the code runs, removing it from the array.
+        // When all fruit are removed, it refills the array.
+        return randomFruit;
```
paulloz, line 488: If we're using `Godot.Collections.Array`, `randomFruit` is a `Variant` here. We either need to return `randomFruit.AsString()`, or use a generic collection `Godot.Collections.Array<string>` to begin with.

tutorials/math/random_number_generation.rst
```diff
@@ -315,6 +320,29 @@ We can apply similar logic from arrays to dictionaries as well:
         # The same metal may be selected multiple times in succession.
         return random_metal
 
+ .. code-tab:: csharp
+
+    private var metals = new Godot.Collections.Dictionary
```
paulloz, line 325: `var` can't be used to define a field.

tutorials/math/random_number_generation.rst
```diff
@@ -315,6 +320,29 @@ We can apply similar logic from arrays to dictionaries as well:
...
+        GD.Randomize();
+        for (int i = 0; i < 20; i++)
+        {
+            GD.Print(GetMetal());
+        }
+    }
+
+    public Godot.Collections.Dictionary GetMetal()
+    {
+        var randomMetal = metals.Values()[GD.Randi() % metals.Length];
```
paulloz, line 343: This method doesn't return.
This line wouldn't compile for several reasons:
- `.Values` is a property, not a method
- I don't think the return of `.Values` can be `[]`-indexed
- Godot collections use `.Count` and not `.Length`

paulloz (changes requested):
Hello 👋 Thanks for the contribution.  
I made some comment in-line. In general, I'm not sure if we should use Godot collections here, instead of system ones. I don't really see an obvious reason for the former, and we establish [here](https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_collections.html#choose-a-collection) to avoid it if possible. Maybe someone else can confirm/infirm that (cc @raulsntos?).

ssBandit:
Thanks for taking a look and noticing the mistakes.  

  The reason for using Godot collections is that in the shuffle bag example it uses .Shuffle(), and the metal example uses .Duplicate() which would be bit more difficult to do with pure system collections. 
(And the documentation kinda implies the use of Godot collections for [Array](https://docs.godotengine.org/en/stable/classes/class_array.html) and [Dictionary](https://docs.godotengine.org/en/stable/classes/class_dictionary.html). Maybe there needs to be a disclaimer about it there.)

tutorials/math/random_number_generation.rst
```diff
@@ -409,17 +443,54 @@ ends up empty. When that happens, you reinitialize it to its default value::
...
+        }
+    }
+
+    public string GetFruit()
+    {
+        if(_fruits.Count == 0)
+        {
+            // Fill the fruits array again and shuffle it.
+            _fruits = _fruitsFull.Duplicate();
+            _fruits.Shuffle();
```
raulsntos, line 482: If we use System collections, shuffling can be implemented as a custom method:

```csharp
private void Shuffle(Metal[] metals)
{
	for (int i = 0; i < metals.Length - 1; i++)
	{
		int j = GD.RandRange(i, metals.Length);
		if (j != i)
		{
			(metals[i], metals[j]) = (metals[j], metals[i]);
		}
	}
}
```

@paulloz Do you think this is good enough or would it be better to keep using Godot collections for the convenience of the `Shuffle` method?

raulsntos (commented):
In general I think we should prefer System collections, as recommended in the documentation page that @paulloz linked.

I can see how using the `Shuffle` method is more convenient than using a custom solution. But I don't think it's too much code.

I think if this documentation page is meant to be a general guide to randomness, using System collections is preferable. I don't think the goal of this page is to show you how to use `Shuffle` specifically, but rather to explain the _shuffle bag_ pattern (which can also be implemented without Godot's `Shuffle` method).

I don't think the text necessarily implies using Godot collections, the word _array_ can be understood as C# arrays, and _dictionary_ is also the name of the System collection. But I see your point, it's true that sometimes the text in the documentation doesn't align perfectly with the C# examples, but that's already the case in many documentation pages.

tutorials/math/random_number_generation.rst
```diff
@@ -310,11 +315,37 @@ We can apply similar logic from arrays to dictionaries as well:
...
+        GD.Randomize();
+        for (int i = 0; i < 20; i++)
+        {
+            GD.Print(GetMetal());
+        }
+    }
+
+    public Godot.Collections.Dictionary GetMetal()
+    {
+        var randomMetal = _metals.KeyValuePair(GD.Randi() % metals.Count);
```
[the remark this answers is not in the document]
paulloz: Wouldn't work either, `Godot.Collections.Dictionary.Values` gives an `ICollection` (doesn't implement `this[int]`).

tutorials/math/random_number_generation.rst
```diff
@@ -310,11 +315,37 @@ We can apply similar logic from arrays to dictionaries as well:
...
+    private Godot.Collections.Dictionary _metals = new Godot.Collections.Dictionary
+    {
+        {"copper", new Godot.Collections.Dictionary{{"quantity", 50}, {"price", 50}}},
+        {"silver", new Godot.Collections.Dictionary{{"quantity", 20}, {"price", 150}}},
+        {"gold", new Godot.Collections.Dictionary{{"quantity", 3}, {"price", 500}}}
+    };
+
+    public override void _Ready()
+    {
+        GD.Randomize();
```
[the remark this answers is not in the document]
paulloz: TIL 😲

tutorials/math/random_number_generation.rst
paulloz: It really depends on what we want to emphasize on this page. Unless we explicitly want to emphasize usage of `Godot.Collections`, I'd rather use system ones.  

On a somewhat unrelated note, we could implement a `GD.Shuffle()` (or somewhere else in the API) that uses the Godot random, and list it on the [API differences](https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_differences.html) page at some point.

tutorials/math/random_number_generation.rst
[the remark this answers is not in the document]
raulsntos: Oh, you are right. I missed that.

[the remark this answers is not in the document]
ssBandit: Oh shoot, I meant [GetKeyValuePair().Value](https://github.com/godotengine/godot/blob/19890614c6a78ec36030ce65c7da05f07fcdb9ed/modules/mono/glue/GodotSharp/GodotSharp/Core/Dictionary.cs#L428)

(Should stop doing quick edits half awake)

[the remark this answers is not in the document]
paulloz: `GetKeyValuePair()` is private. I don't think there's really a straightforward way to oneline this using `Godot.Collections.Dictionary`.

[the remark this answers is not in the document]
raulsntos: I guess the best we could do here is use Linq's [`ElementAt`](https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.elementat), if we want to keep it short.

```csharp
var (_, randomMetal) = _metals.ElementAt(GD.Randi() % metals.Count);
return randomMetal;
```

paulloz:
@raulsntos should we finally get this merged?

tutorials/math/random_number_generation.rst
```diff
@@ -409,17 +420,53 @@ ends up empty. When that happens, you reinitialize it to its default value::
...
-        # Prints "apple", "orange", "pear", or "banana" every time the code runs.
+        # Returns "apple", "orange", "pear", or "banana" every time the code runs, removing it from the array.
+        # When all fruit are removed, it refills the array.
         return random_fruit
 
+ .. code-tab:: csharp
+
+    private Godot.Collections.Array<string> _fruits = new() { "apple", "orange", "pear", "banana" };
+    // A copy of the fruits array so we can restore the original value into `fruits`.
+    private Godot.Collections.Array _fruitsFull;
```
raulsntos, line 439: Since this is a copy of the `_fruits` array it should have the same type.

```suggestion
    private Godot.Collections.Array<string> _fruitsFull;
```

tutorials/math/random_number_generation.rst
```diff
@@ -409,17 +420,53 @@ ends up empty. When that happens, you reinitialize it to its default value::
...
+            _fruits.Shuffle();
+        }
+
+        // Get a random fruit, since we shuffled the array,
+        var randomFruit = _fruits[0];
+        // and remove it from the `_fruits` array.
+        _fruits.RemoveAt(0);
+        // Returns "apple", "orange", "pear", or "banana" every time the code runs, removing it from the array.
+        // When all fruit are removed, it refills the array.
+        return randomFruit.AsString();
```
raulsntos, line 467: `randomFruit` is already a string, not a Variant, because the array is typed.

```suggestion
        return randomFruit;
```

tutorials/math/random_number_generation.rst
```diff
@@ -409,17 +420,53 @@ ends up empty. When that happens, you reinitialize it to its default value::
...
+    {
+        if(_fruits.Count == 0)
+        {
+            // Fill the fruits array again and shuffle it.
+            _fruits = _fruitsFull.Duplicate();
+            _fruits.Shuffle();
+        }
+
+        // Get a random fruit, since we shuffled the array,
+        var randomFruit = _fruits[0];
```
raulsntos, line 462: We should probably avoid using `var` in the documentation to make the type more clear.

```suggestion
        string randomFruit = _fruits[0];
```

raulsntos (commented):
Thanks for the ping, I totally forgot about this. I took another look and found a couple things that I missed in my previous review.

tutorials/math/random_number_generation.rst
raulsntos: > I also think that any valid version of this is better than what we currently have

Agreed, so let's not block the PR on this. I'll resolve.

tutorials/math/random_number_generation.rst
```diff
@@ -295,26 +285,49 @@ We can apply similar logic from arrays to dictionaries as well:
...
         # The same metal may be selected multiple times in succession.
         return random_metal
 
+ .. code-tab:: csharp
+
+    private Godot.Collections.Dictionary<string, Godot.Collections.Dictionary<string, int>> _metals = new()
+    {
+        {"copper", new Godot.Collections.Dictionary{{"quantity", 50}, {"price", 50}}},
+        {"silver", new Godot.Collections.Dictionary{{"quantity", 20}, {"price", 150}}},
+        {"gold", new Godot.Collections.Dictionary{{"quantity", 3}, {"price", 500}}}
```
raulsntos, line 312:
```suggestion
        {"copper", new Godot.Collections.Dictionary<string, int>{{"quantity", 50}, {"price", 50}}},
        {"silver", new Godot.Collections.Dictionary<string, int>{{"quantity", 20}, {"price", 150}}},
        {"gold", new Godot.Collections.Dictionary<string, int>{{"quantity", 3}, {"price", 500}}},
```

tutorials/math/random_number_generation.rst
```diff
@@ -295,26 +285,49 @@ We can apply similar logic from arrays to dictionaries as well:
...
+    {
+        for (int i = 0; i < 20; i++)
+        {
+            GD.Print(GetMetal());
+        }
+    }
+
+    public Godot.Collections.Dictionary GetMetal()
+    {
+        var (_, randomMetal) = _metals.ElementAt(GD.Randi() % metals.Count);
```
raulsntos, line 325:
```suggestion
        var (_, randomMetal) = _metals.ElementAt((int)(GD.Randi() % _metals.Count));
```

tutorials/math/random_number_generation.rst
```diff
@@ -295,26 +285,49 @@ We can apply similar logic from arrays to dictionaries as well:
...
+
+    public override void _Ready()
+    {
+        for (int i = 0; i < 20; i++)
+        {
+            GD.Print(GetMetal());
+        }
+    }
+
+    public Godot.Collections.Dictionary GetMetal()
```
raulsntos, line 323:
```suggestion
    public Godot.Collections.Dictionary<string, int> GetMetal()
```

tutorials/math/random_number_generation.rst
```diff
@@ -461,7 +507,6 @@ terrain. Godot provides :ref:`class_fastnoiselite` for this, which supports
 
     public override void _Ready()
     {
-        GD.Randomize();
         // Configure the FastNoiseLite instance.
         _noise.NoiseType = NoiseTypeEnum.SimplexSmooth;
```
raulsntos, line 511:
```suggestion
        _noise.NoiseType = FastNoiseLitle.NoiseTypeEnum.SimplexSmooth;
```

tutorials/math/random_number_generation.rst
```diff
@@ -131,16 +131,12 @@ varying by the deviation (1.0 by default):
...
 
     # Prints a random floating-point number from a normal distribution with a mean 0.0 and deviation 1.0.
-    var random = RandomNumberGenerator.new()
-    random.randomize()
-    print(random.randfn())
+    print(randfn(0.0, 1.0))
 
  .. code-tab:: csharp
 
     // Prints a normally distributed floating-point number between 0.0 and 1.0.
```
raulsntos, line 138:
```suggestion
    // Prints a random floating-point number from a normal distribution with a mean 0.0 and deviation 1.0.
```

raulsntos (approved):
Looks good to me, thanks! And sorry it took me so long to review.

ssBandit:
No problem. Thanks for the diligent review.

tetrapod00:
Thanks, and thanks for your patience with the long review process! Congrats on your first merged docs contribution

ssBandit:
Thank you.  
Even though the one thing I was meaning to correct in the first place (empty() -> is_empty()) was fixed already by someone else :smiling_face_with_tear: 
I should split commits into separate pull requests in the future.

PR #8079, opened 2023-09-28
Update gui_using_fonts.rst

NiftyHat:
Following the instructions to copy the glyph from the fontello sites resulted in shows a missing glyph symbol in game and not the intended symbol. I think that under certain conditions fontello doesn't pack the glyphs at the same address as the characters rendered on the website. 

Anyway, copying the character from the generated html document produces consistent results, so that's what I changed it to in the docs.


AThousandShips:

AThousandShips:
Updated your branch for you

PR #8080, opened 2023-09-28
Marking CustomMainLoop sample C# class as GlobalClass

TinyMatt:
The [sample code](https://docs.godotengine.org/en/latest/classes/class_mainloop.html#tab-0-QyM=) for custom C# `MainLoop` is missing the `[GlobalClass]` export. Without this attribute the sample code will cause an error on startup.

### DIFF
--- a/classes/class_mainloop.rst
+++ b/classes/class_mainloop.rst
@@ -54,6 +54,7 @@ Here is an example script implementing a simple **MainLoop**:
 
     using Godot;
     
+    [GlobalClass]
     public partial class CustomMainLoop : MainLoop
     {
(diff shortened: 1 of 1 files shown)


skyace65:
Hey, we appreciate the help but any changes to the class reference pages of the documentation need to be made in the main engine repository. The files in this repository are generated from what's in the engine repository, you can find more info [here](https://docs.godotengine.org/en/stable/contributing/documentation/updating_the_class_reference.html). [This](https://github.com/godotengine/godot/blob/master/doc/classes/MainLoop.xml) is the file you want to edit.

PR #8083, opened 2023-09-28
Update ways_to_contribute.rst

MIfoodie:
Updated the amount of developers that have contributed to the engine (the number was pretty out of date).

### DIFF
--- a/contributing/ways_to_contribute.rst
+++ b/contributing/ways_to_contribute.rst
@@ -77,7 +77,7 @@ Contributing code changes upstream has two big advantages:
 
 -  The whole community will benefit from your work, and other contributors will
    behave the same way, contributing code that will be beneficial to you. At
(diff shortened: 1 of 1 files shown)


skyace65 (commented):
Apologies if this comes as nitpicky but I think it would be better to leave it at 2,000 since it's a nice round number. 2,220 comes off as oddly specific, and I don't think we want to update this every time another 10 people contribute.

MIfoodie:
> Apologies if this comes as nitpicky but I think it would be better to leave it at 2,000 since it's a nice round number. 2,220 comes off as oddly specific, and I don't think we want to update this every time another 10 people contribute.

Hahaha while I was writing/typing the number I was thinking about this same issue and I just couldn't make up my mind. I'll make the changes.

mhilbrunner (approved):

mhilbrunner:
Thanks, merged. :)

PR #8086, opened 2023-09-28
Add note on fn key to class reference tool doc

rtflynn:
The integrated class reference tool requires a F1 keypress, which on windows laptops with a 'function' key actually opens a microsoft edge browser window and makes a useless bing query. This behavior can't be re-binded to another key. The way to get around this is to use 'fn + F1'.


skyace65:
I need you to please fix all these extra commits that got added to your PR. If you can't fix it in this PR feel free to open a second. I noticed on your docs fork that you have a master and 4.1 branch, that's not an issue but if you fork from 4.1 when making a PR and try to merge it into master you'll get issues like this.

rtflynn:
Thanks, I had a feeling I'd done some funny business with the forking. I'll just remove this PR and make another :D

PR #7990, opened 2023-09-21
Correct mouse_and_input_coordinates.rst

profande:
This page currently refers to ``event.speed`` which is a field that no longer exists on ``InputEventMouseMotion`` and has since been renamed to ``event.velocity``.

PR #8095, opened 2023-09-29
Correct invalid property for InputEventMouseMotion

CPritch:
Fix invalid property for InputEventMouseMotion. Should be 'velocity' instead of 'speed'


AThousandShips:
Thank your for your contribution!

This was already done in:
* https://github.com/godotengine/godot-docs/pull/7990

And just needs to be cherry-picked, which does not need a separate PR

PR #8104, opened 2023-09-29
Fix minor documentation bug

### LINKED ISSUE #8103: `RefCounted.unreference()` returns true only if the object has died, not if it succeeded decrementing.
**Your Godot version:** 4.1.1-stable

**Issue description:** `RefCounted.unreference()` returns true only if the object has died, not if it succeeded decrementing.

**URL to the documentation page:** https://docs.godotengine.org/en/stable/classes/class_refcounted.html#class-refcounted-method-unreference

[Source code](https://github.com/godotengine/godot/blob/4.1/core/object/ref_counted.cpp#L78)

RadiantUwU:
fixes #8103 
will work for 4.x too

### DIFF
--- a/classes/class_refcounted.rst
+++ b/classes/class_refcounted.rst
@@ -111,7 +111,7 @@ Returns ``true`` if the increment was successful, ``false`` otherwise.
 
 Decrements the internal reference counter. Use this only if you really know what you are doing.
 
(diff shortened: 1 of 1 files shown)


RadiantUwU:
image
Will work with 3.x too

Issue #8116, opened 2023-10-01
Encourage users to use @export, not get_node()


PickleJesus123:
**Your Godot version:**
4.1.1

**Issue description:**
The Godot tutorials should encourage developers to use `@export` to reference nodes, instead of get_node(). `@export` is good practice, since using get_node(str) creates "brittle code". If a developer wants to refactor their nodes/scripts (studios do this quite often in the industry, especially as projects get larger), it becomes a long, tedious, and error-prone task with get_node(). However, if they use `@export`, nodes can be renamed/moved freely without worry. The Unity equivalent to get_node(str) is transform.Find(str), and it is _highly discouraged_.

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/stable/tutorials/scripting/nodes_and_scene_instances.html

PickleJesus123:
Corresponding pull-request: #8122

Mickeon:
It's worth keeping in mind `@export` nodes have a noticeable flaw, currently. That is, references to the Nodes are not transferred to new Nodes when duplicated. See https://github.com/godotengine/godot/issues/78060

AThousandShips:
Please put \` around your mentions of `@export` or you will keep tagging a GitHub user

PR #8122, opened 2023-10-01
Fixed 'Getting Nodes' to adhere to official Godot best practices

PickleJesus123:
Brings the "Getting Nodes" section of [this tutorial page](https://docs.godotengine.org/en/stable/tutorials/scripting/nodes_and_scene_instances.html) up to date with [current best practices](https://docs.godotengine.org/en/stable/tutorials/best_practices/godot_interfaces.html#acquiring-object-references) of using `@export` to reference nodes instead of  get_node(str).

From the changes: "... Relying on the get_node(path) method can lead to brittle project structures, as it's tied to hard-coded node paths which can easily break with renaming or scene restructuring. Since Godot 4.0, it's recommended to reference your nodes by using `@export` in your code"

The frequent use of get_node(str) is one of the most heavily criticized aspects of Godot from industry professionals. The Unity equivalent is transform.Find(str), which is _seriosuly discouraged_ in most cases due to how fragile it makes projects. There seems to be consensus on the [corresponding issue page](https://github.com/godotengine/godot-docs/issues/8116) that this change needs to be made so that new Godot developers develop good habits from the get-go.

You can preview the changed tutorial page [here](https://github.com/PickleJesus123/godot-docs/blob/master/tutorials/scripting/nodes_and_scene_instances.rst)


NorPhi:
I don't see ```@export```being touted as the current best practice.  ```@export``` is already over-used, a change like this would just encourage new users to clutter their inspector even more and once you copy a node all the inspector references will break iirc.

Stronkkey:
Please rebase your commit history using `git rebase -i`. It's hard to keep track of 24 commits.

shaunhegarty:
> I don't see `@export` as being touted as the current best practice. `@export` is already over-used, a change like this would just encourage new users to clutter their inspector even more. Also, if you copy a node all the inspector references will break iirc.

This would be the reference here: https://github.com/godotengine/godot-docs/blob/228c10e1e647bac17e351ce0b7418b4a712d1c4c/tutorials/best_practices/godot_interfaces.rst?plain=1#L138-L160

The implication here is that it's best practice as fastest, not as "best". 

I've been using export a lot myself, but I'm new and my projects aren't complicated. It feels a lot better than get_node or $NodeName because I move stuff all the time. Maybe what's really needed is some notes or discussion on what to use and when to be included as info in the best practices with some advice directing the user to best practices in the tutorial areas?

PR #8128, opened 2023-10-02
fix minor issues

FelixBurkhard:
Hey everyone,

I just started exploring Godot and went through the Getting Started part. I noticed a few thing that I think should be changed:

1. The `Dodge the Creeps!` string was not consistent through the page.
2. The `\n` should be removed from the Label Text because the Label should do the wrapping.
3. There were two reminders to remove the `new_game()` call with only a few lines in between.
4. The `async` method in C# should probably return a `Task` instead of `void`. I am not 100% sure about this, as I didn't dig into how the method is actually called. Also I think this is fine if done intentionally to not confuse users who are new not just to Godot but to coding in general.

As someone who is new to game development I must say I really liked the Getting Started docs! A big thanks to all the contributors!


FelixBurkhard:
closing because I used the wrong base branch -.-

PR #8142, opened 2023-10-03
Update the state of C#/.NET support as of 4.2

YuriSizov:
Add more details to the documentation to avoid confusion, and also remove a scary message that we don't really need anymore. See also https://github.com/godotengine/godot-website/pull/671.


about/list_of_features.rst
```diff
@@ -429,12 +435,16 @@ Scripting
 :ref:`C#: <toc-learn-scripting-C#>`
 
 - Packaged in a separate binary to keep file sizes and dependencies down.
-- Uses .NET 6.
+- Supports .NET 6 and 7.
```
raulsntos, line 438: Should we just say .NET 6 and greater? Otherwise we'll have to modify it again when .NET 8.0 is released on November.

tutorials/scripting/c_sharp/c_sharp_basics.rst
```diff
@@ -29,9 +18,8 @@ it is implemented with .NET 6.0.
 
 .. attention::
 
-    Projects written in C# using Godot 4.x currently cannot be exported to
-    Android, iOS and web platforms. To use C# on those platforms, use Godot 3
-    instead.
+    Projects written in C# using Godot 4.x currently cannot be exported to iOS
```
raulsntos, line 21: The experimental Android export only supports architectures arm64 and x64. Should this mentioned here?

getting_started/step_by_step/scripting_languages.rst
```diff
@@ -128,9 +128,8 @@ officially supported .NET option.
 
 .. attention::
 
-    Projects written in C# using Godot 4.x currently cannot be exported to
-    Android, iOS and web platforms. To use C# on those platforms, use Godot 3
-    instead.
+    Projects written in C# using Godot 4 currently cannot be exported to iOS
```
raulsntos, line 131: Same.

raulsntos (commented):
I feel like "experimental Android support" doesn't say much from the user's perspective. What does _experimental_ mean to users? Maybe we should list the limitations of Android exports somewhere.

about/list_of_features.rst
YuriSizov: We can, yes. As long as it's clear and easy to find, I think.

YuriSizov:
> Maybe we should list the limitations of Android exports somewhere.

Sure, I'm not aware of any specificities, but I can add whatever makes sense to the notes. As for "experimental", I think it makes it pretty clear that it's in a working state but in need of public testing. But we can also link something relevant to its status if we have any external resources, or add some details as you suggest.

tutorials/scripting/c_sharp/c_sharp_basics.rst
YuriSizov: I would probably add it to the list of features page and then reference it from here.

raulsntos:
By the way, I'm not against using the word "experimental" and I think it's the right word to use for the current status. My concern was only that it's not descriptive enough to users about the limitations, so adding some more details somewhere and linking to it sounds good to me.

Off the top of my head, the current limitations for Android exports are:
- Only some architectures are supported: `arm64` and `x64`.
- The user project needs to target .NET 7.0 (which requires them modifying the `.csproj`).

YuriSizov:
@raulsntos Okay, I've added more details into notes, and also improved the index page for C# with a bit of structure, some renames for clarity and more details about platform support (which I link to from all those notes). Let me know what you think.

chrome_2023-10-03_17-50-27
chrome_2023-10-03_17-50-36

tutorials/scripting/c_sharp/index.rst
```diff
@@ -1,17 +1,54 @@
...
    c_sharp_variant
    c_sharp_signals
    c_sharp_exports
    c_sharp_global_classes
-   c_sharp_style_guide
-   diagnostics/index
+
+.. _doc_c_sharp_platforms:
+
+C# platform support
```
raulsntos, line 41: I love the new index page, but I'm not sure if this is the best place for this section. We have similar sections in `c_sharp_basics` (_Current gotchas and known issues_ or _Common pitfalls_) but I'm not sure if that's better, to be honest I think the `c_sharp_basics` page should probably be split into multiple pages.

YuriSizov: I know what you mean, and I was thinking removing some stuff from the basics page that can be now found on the index page... but ultimately decided that a bit of duplication is fine. These are likely to be the two landing pages, so it makes sense to have a bit of important information on both of them.

I do support reorganizing the docs, though. I'm not entirely sure why we have articles framed as introduction to C#. That's probably unnecessary and we could go straight into Godot-specific stuff (with some notes here and there that link Godot API to C# concepts).

I could give it a go as time allows, but it would be harder for me to find a good angle here since I'm neither a C# user nor maintainer 🙃 But I can help, if you're interested.

raulsntos: I would love that, maybe we can do that after the 4.2 release. For now I think this is good.

raulsntos (approved):
Looks great, thanks for all the improvements to the C#/.NET documentation.

mhilbrunner (approved):

mhilbrunner:
Thanks! Merged.

PR #8144, opened 2023-10-03
Update C# differences for 4.2

raulsntos:
Since we're entering feature freeze and it's unlikely new API will be added, I think it's now safe to update this page for 4.2.


mhilbrunner (approved):

mhilbrunner:
Thanks as always for caring about the docs :) Merged.

PR #7792, opened 2023-08-10
Update signals.rst and instancing.rst

Alex2782:
**Update instancing.rst:**
> For combinations, use the + symbol with a space on either side of the symbol.
https://docs.godotengine.org/en/latest/contributing/documentation/docs_writing_guidelines.html#keyboard-shortcut-guidelines

**Update signals.rst** 
MacOS Shortcut for "Open 2D Editor" is `Ctrl + Command + 1`

Bildschirmfoto 2023-08-06 um 16 50 01

PR #8147, opened 2023-10-03
Update signals.rst to reflect proper macOS shortcut for viewport-switching

squishypixel:
Typo: Correct the shortcut for macOS


mhilbrunner:
hey! Thanks for taking the time to fix this. As noted above, we already got a similar PR earlier and now merged it: https://github.com/godotengine/godot-docs/pull/7792. The page should be updated to reflect that change soon. Thanks again :)

PR #8148, opened 2023-10-04
Update script filename in `scripting_first_script.rst`

nskins:
By default, the script is named `Sprite2D.gd` (not `sprite_2d.gd`) as shown in the relevant image. This PR updates the guide to reflect this. 

**EDIT**: The following two lessons build off the first, so I updated the script name in those as well.

### DIFF
--- a/getting_started/step_by_step/scripting_first_script.rst
+++ b/getting_started/step_by_step/scripting_first_script.rst
@@ -89,7 +89,7 @@ other options by default and click the Create button to create the script.
 
 .. image:: img/scripting_first_script_attach_node_script.webp
 
-The Script workspace should appear with your new ``sprite_2d.gd`` file open and
+The Script workspace should appear with your new ``Sprite2D.gd`` file open and
 the following line of code:
 
 .. tabs::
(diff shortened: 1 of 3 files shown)


mhilbrunner:
Snake case (as it is) is correct for file names according to the [GDScript style guide](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html#file-names).

It is the images that need to be updated in this case :) PR welcome. Note that images now also should be in WebP instead of PNG, according to our [image contribution guidelines](https://docs.godotengine.org/en/stable/contributing/documentation/docs_image_guidelines.html).

nskins:
@mhilbrunner Thanks for the feedback! I just want to mention that, if that is the case, the filename is initialized incorrectly (unless this was fixed recently...). Below is a snapshot of the "Attach Node Script" window after following the exact directions in the tutorial (`v4.1.1-stable`). Note that the tutorial does not instruct you to change the filename. Might it then be a bug with the editor? The script filenames should instead be initialized in snake case?

Screenshot from 2023-10-06 15-57-00

Piralein:
The behaviour can be changed in your project settings. `Editor/Naming - Scene Name Casing` (Advanced Settings enabled)
It should default to `snake_case` in a new project.

PR #8152, opened 2023-10-04
Clarified that the default template uses type hints.

### LINKED ISSUE #8151: The default script template does not match Godot editor
**Your Godot version:**
v4.1.2.stable.official [399c9dc39]

**Issue description:**

The [default script template](https://docs.godotengine.org/en/latest/tutorials/scripting/creating_script_templates.html#default-template) does not match what the Godot editor produces with default settings. Instead, it matches what the editor will produce when "Add Type Hints" has been turned on in the Editor settings.

Default Godot editor:
```
extends Node


# Called when the node enters the scene tree for the first time.
func _ready():
	pass # Replace with function body.


# Called every frame. 'delta' is the elapsed time since the previous frame.
func _process(delta):
	pass
```

Godot editor with "Add Type Hints" turned on:
```
extends Node


# Called when the node enters the scene tree for the first time.
func _ready() -> void:
	pass # Replace with function body.


# Called every frame. 'delta' is the elapsed time since the previous frame.
func _process(delta: float) -> void:
	pass
```

It appears that the template used to use `func _ready()%VOID_RETURN%:` and `func _process(delta%FLOAT_TYPE%)%VOID_RETURN%:`, but when the template was [updated ](https://github.com/godotengine/godot-docs/issues/5887) to use the new `_XXX_` format, those were replaced with explicit type hints.

https://docs.godotengine.org/en/latest/tutorials/scripting/creating_script_templates.html#default-template

allenwp:
Fixes #8151. The default template did not match what Godot produces with default settings. This change clarifies that this default template includes type hints.

### DIFF
--- a/tutorials/scripting/creating_script_templates.rst
+++ b/tutorials/scripting/creating_script_templates.rst
@@ -140,7 +140,7 @@ the file with the ``meta-default: true`` header.
 
 Only one template can be set as default at the same time for the same node type.
 
-The ``Default`` templates for basic Nodes, for both GDScript and C#, are shown here so you can
+The ``Default`` templates for basic Nodes, for both GDScript with type hints and C#, are shown here so you can
 use these as the base for creating other templates:
 
 .. tabs::


AThousandShips:
This isn't correct, the template [is](https://github.com/godotengine/godot/blob/bfd78bb917887cfc1fd842ba23570394cad8bedb/modules/gdscript/editor/script_templates/Node/default.gd) what the actual template is, it doesn't say that it is the *output* of the template, but *is* the template

PR #8153, opened 2023-10-04
Automatically convert spaces to tabs in CodeBlocks

### LINKED ISSUE #6779: Change indentation from spaces to tabs for codeblocks

Repiteo:
Closes #6779
Supersedes #7529

---

A significant improvement from the previous design! This is no longer a Sphinx extention; consequently, it doesn't need to integrate with the build system to function. Instead, it's incorporated in `custom.js` utilizing a single regex instead of relying on multiple passes. Thanks to pulling from the node system, it's also able to apply to specific codeblocks, fixing an oversight from the previous implementation that added tab characters to C#/Python scripts as well

Measurements on my local machine showed a total process time to take between ~4.5ms and ~0.5ms, depending on the amount of codeblocks on the page. I don't know the best means of measuring this for other environments, so the best I can do is compare it to the `linkItem` loop directly above it, which will clock in at ~23ms regardless of the current page

Like the previous PR, this serves to make the displayed GDScript code consistent with its [Style Guide](https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_styleguide.html) (as well as other codeblocks that'd reasonably expect tabs, like the [cpp example](https://docs.godotengine.org/en/latest/tutorials/scripting/gdextension/gdextension_cpp_example.html)). This will make any copy/pasted content immediately suitable for a default workspace, instead of needing to rely on autoformatting and/or the editor throwing an error for inconsistent indentation. Also, as part of making this consistent, all codeblocks now have the `tab-size: 4` css property

### DIFF
--- a/_static/js/custom.js
+++ b/_static/js/custom.js
@@ -291,6 +291,19 @@ $(document).ready(() => {
     }
   }
 
+  // Change indentation from spaces to tabs for codeblocks.
+  const codeBlocks = document.querySelectorAll('.rst-content div[class^="highlight"] pre');
+  for (const codeBlock of codeBlocks) {
+    const classList = codeBlock.parentNode.parentNode.classList;
+    if (!classList.contains('highlight-gdscript') && !classList.contains('highlight-cpp')) {
+      // Only change indentation for GDScript and C++.
+      continue;
+    }
+    let html = codeBlock.innerHTML;
+    html = html.replace(/(?<=^(<span class="w">)?( {4})*)( {4})/gm, '\t');
+    codeBlock.innerHTML = html;
+  }
+
   // See `godot_is_latest` in conf.py
   const isLatest = document.querySelector('meta[name=doc_is_latest]').content.toLowerCase() === 'true';
   if (isLatest) {
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -846,6 +846,7 @@ code,
 
 .highlight {
     background-color: var(--highlight-background-color);
+    tab-size: 4;
 }
 
 /* Emphasized lines */
--- a/tutorials/migrating/upgrading_to_godot_4.1.rst
+++ b/tutorials/migrating/upgrading_to_godot_4.1.rst
@@ -253,7 +253,7 @@ This means that GDExtensions made for Godot 4.0 will need to be recompiled for G
 
 In Godot 4.0, your "entry_symbol" function looks something like this:
 
-.. code-block:: C++
+.. code-block:: cpp
 
   GDExtensionBool GDE_EXPORT example_library_init(const GDExtensionInterface *p_interface, const GDExtensionClassLibraryPtr p_library, GDExtensionInitialization *r_initialization) {
       godot::GDExtensionBinding::InitObject init_obj(p_interface, p_library, r_initialization);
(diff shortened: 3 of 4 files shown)


_static/js/custom.js
```diff
@@ -291,6 +291,19 @@ $(document).ready(() => {
     }
   }
 
+  // Change indentation from spaces to tabs for codeblocks.
+  const codeBlocks = document.querySelectorAll('.rst-content div[class^="highlight"] pre');
+  for (const codeBlock of codeBlocks) {
+    const classList = codeBlock.parentNode.parentNode.classList;
+    if (classList.contains('highlight-csharp') || classList.contains('highlight-python')) {
+      // CSharp and Python use spaces for indentation, ignore.
```
mhilbrunner, line 299: I wonder if we should check for GDScript instead. This seems like a possible footgun when adding/changing languages, and I'm not sure spaces in random codeblocks should always be converted (shaders? Stuff that isn't code but formatted as code due to reasons?)

Besides that concern, this looks really good to me. :) Thanks for improving upon it still while we didn't even get to merging the first one.

Repiteo: The main two I had in mind was GDScript and, to a lesser extent, C++. It should be simple enough to change the checks to opt-in for those exclusively, instead of the current opt-out format

Repiteo: Updated to opt-in format!

Also, changed code-block formatting in a few areas from "C++" to "cpp" so that the filter would properly recognize them

mhilbrunner (approved):

_static/js/custom.js
mhilbrunner: That was quick! Thanks, looks good 🙏

mhilbrunner:
Alright, lets go ahead with this in `master` and see if anything comes up, otherwise we can cherry-pick to the other branches soon.

Thanks again for working on this :)

mhilbrunner:
Cherry-picked to 4.1

YuriSizov:
So this PR breaks in Safari 16.3 and before, because those versions do not support lookbehind in RegExp: https://caniuse.com/js-regexp-lookbehind.

Repiteo:
How recent does a js feature need to be before it's safe to assume people will have upgraded? I wouldn't have thought that would be considered if later Safari versions do support it (though Safari *is* unique in that it was only added this year). Either way, it shouldn't be too hard to make a PR with an alternative implementation

YuriSizov:
@Repiteo This feature was only added in Safari this year. It's too recent.

Repiteo:
Alright, fair enough

In that case, it might be worth revisiting the original implementation style after all; something that could bypass the need for js shenanigans entirely because it's all pre-formatted. That'd be a longer endeavor, so for now I'll work on a slightly bulkier regex pattern

YuriSizov:
By the way, why do we need to check for `<span class="w">` in the first place?

Repiteo:
Ah! That's a quirk of the pygments system. Every code language has defined logic for handling whitespace, but our gdscript implementation is legacy in certain respects from the version used in the pygments repo, so it lacks the whitespace span wrappers. Because the space→tab conversion applies to cpp as well, which uses that system, it requires that prefix check. Though to clarify, that part isn't the only thing that uses the lookbehind; it's still used for `( {4})*`

YuriSizov:
> it's still used for `( {4})*`

And there it's used so we can replace every 4 spaces which are preceded by 4N spaces?

Repiteo:
Yep! It also checks if those 4 spaces are at the beginning of a string. The lookbehind is what allows a single regex check to handle all of that

Repiteo:
Fix PR is up!
https://github.com/godotengine/godot-docs/pull/8334

PR #8159, opened 2023-10-05
Ogg to Ogg Vorbis

adotcodes:
I changed "Ogg" to "Ogg Vorbis" to make it clear that it's referring to the Vorbis codec used in Ogg files.

### DIFF
--- a/classes/class_audiostream.rst
+++ b/classes/class_audiostream.rst
@@ -21,7 +21,7 @@ Base class for audio streams.
 Description
 -----------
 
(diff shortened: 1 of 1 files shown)


mhilbrunner (dismissed):

mhilbrunner:
The class reference files (everything under `classes/`) are generated from the main repository, so the source files need to be edited there, and PRs need to be made against the [main repository](https://github.com/godotengine/godot), not the docs, otherwise any changes are going to be lost during the next update.

See [Contributing to the class reference](https://docs.godotengine.org/en/latest/contributing/documentation/updating_the_class_reference.html).

PR #8160, opened 2023-10-05
Update custom audiostream docs to work with Godot v4.1.1

### LINKED ISSUE #7492: Custom AudioStreams documentation refers to function which does not exist
**Your Godot version:**
4.0.3

**Issue description:**
I'm attempting to use GDExtension to add a custom audio stream, which is well documented for a module at this page. However, the page references AudioServer::get_singleton()->audio_data_alloc, which I cannot find in any Godot source or in the C++ binding.

I've done a full grep of every cpp file in both, with no results.

I suspect it was removed or replaced at some point, but I'm not sure by what. Worse, I've only got the name to tell me what it's doing; nothing to tell me how this is different from any other allocation. I'll have to go through old code, maybe, to figure it out.

https://docs.godotengine.org/en/stable/contributing/development/core_and_modules/custom_audiostreams.html

I unfortunately do not know what the correct information would be, so I am unable to open a pull request to fix it, myself, at this time. All help is much appreciated.

nonameentername:
This fixes #7492


Calinou (commented):
Looks good to me on principle, but I don't have a GDExtension setup to test this.

Could you upload a reproduction project to make sure this works as intended?

contributing/development/core_and_modules/custom_audiostreams.rst
```diff
@@ -49,44 +49,55 @@ Therefore, playback state must be self-contained in AudioStreamPlayback.
...
 	class AudioStreamMyTone : public AudioStream {
 		GDCLASS(AudioStreamMyTone, AudioStream)
 
 	private:
 		friend class AudioStreamPlaybackMyTone;
-		uint64_t pos;
+		float pos;
 		int mix_rate;
 		bool stereo;
 		int hz;
```
Calinou, line 68: It may be worth assigning default values here instead of the constructor to avoid issues with undefined behavior if one of those is accidentally unset.

nonameentername:
@Calinou [here](https://github.com/nonameentername/godot-custom-audiostream) is an example project using the same code as the one in the documentation.

PR #8167, opened 2023-10-05
GDScript: Update `match` patterns and literal docs

dalexeev:
1\. Fix nested lists:

2\. Fix constant and variable patterns, they are actually [literal and expression patterns](https://github.com/godotengine/godot/blob/c2b916751d7a52e7f002eb7361fe2b12c1cbc488/modules/gdscript/gdscript_analyzer.cpp#L2266-L2291).

3\. Use level 5 headers for `match` subsections, it looks better and creates anchors. I wanted to add level 6 headers for patterns, but they are almost the same as regular text, which in my opinion looks worse.

4\. Fix the funny sentence about 6 types of patterns before a list of 7 items (due to multiple patterns).

5\. Add `null`, `false` and `true` to literal table.

6\. The shorthands `$NodePath` and `%UniqueNode` are not literals.

7\. Move warning about comparison since it is not related to the basic syntax

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_basics.rst
+++ b/tutorials/scripting/gdscript/gdscript_basics.rst
@@ -341,7 +341,11 @@ Literals
 ~~~~~~~~
 
 +---------------------------------+-------------------------------------------+
-| **Literal**                     | **Type**                                  |
+| **Example(s)**                  | **Description**                           |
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):

mhilbrunner:
LGTM, thanks :) Merged.

PR #8171, opened 2023-10-06
Anticipate the target argument

DeeJayLSP:
For a bit of consistence with the rest of the argument lists in the page.

Better for translation, as the `target` word didn't appear as an argument before.


mhilbrunner (approved):

mhilbrunner:
Thanks! Looks good to me. Merged. :)

Issue #8163, opened 2023-10-05
Damping for 2D and 3D is described incorrectly.


thygrrr:
**Your Godot version:**
4.1.2.stable but also 4.2.dev.custombuild

**Issue description:**
The documentation is grossly(!) incorrect in describing the linear and angular damping techniques used. It insinuates a **critical spring damper function** (sometimes known as smoothdamp), in particular in the language where it says:

> Note: Good values are in the range 0 to 1. At value 0 objects will keep moving with the same velocity. Values greater than 1 will aim to reduce the velocity to 0 in less than a second e.g. a value of 2 will aim to reduce the velocity to 0 in half a second. 

This is wrong because:
- it's actually implemented as a **negative exponential damper**, which behaves very differently (and poorly, because this implementation is fully tickrate dependent), this warrants special care to be taken to educate the user of pitfalls
- a value of 1 reduces the magnitude to 0 in **more than** the described time of 1 second, never less. In fact, it will barely have removed 65% of the speed at that time, and the rate of deceleration is shrinking)
- it's not trying "to stop in D seconds", that may have been the intent of the programmer but it's not what the function does: especially at the 0.1 default value; which will very definitely not have stopped after anywhere near 10 seconds (the sleep threshold of 0.1m/s is crossed after 23 seconds for a low starting speed of 1m/s, and 46 seconds for a speed fo 10m/s, and a whole minute at 40m/s [90 mph / 145 kph])
- 1 is not a good upper bound for this value, it can easily go as high as 10 or 20 and still make perfect sense. In no way is a value betweeen 0 and 1 "good" with this damper, in fact, it's too low for most intentes and leads to floaty physics
- technically, the 2D physics engine supports values as low as -1 and only below this value will start throwing errors.
- the tickrate dependence means that a value of d=65 will stop an object dead immediately (after a distance of 0) at a tickrate of 60 tps, but at 200 tps, it will keep moving for a timespan of a good 20 frames (speed=1) or much more (speed>1, e.g. 10), and after a definitely nonzero distance that grows significantly with higher speed. (too tired to solve the integral right now)

The source code applies the following operation:
```cpp
real_t damp = 1.0 - p_step * total_linear_damp;
//...
linear_velocity *= damp;
```

This is, after `n` physics ticks, equivalent of 
`linear_velocity = linear_velocity_start * pow(1-p_step*total_linear_damp, n)`

[...]

PR #8176, opened 2023-10-06
Angular and Linear Damp 

thygrrr:
The values are explained in more detail, with correct maths and important caveats highlighted.
The ranges are corrected, as 0..1 was not a good range at all. (in fact, 1 is already the default for 2D). The values are changed as "suggested" instead of "good", because they are very subjective.

Resolves https://github.com/godotengine/godot-docs/issues/8163

### DIFF
--- a/classes/class_projectsettings.rst
+++ b/classes/class_projectsettings.rst
@@ -8544,7 +8544,7 @@ Suggested values are in the range ``0`` to ``30``. At value ``0`` objects will k
 
 During each physics tick, Godot will multiply the angular velocity of RigidBodies by ``1.0 - combined_damp / physics_ticks_per_second``, where ``combined_damp`` is the sum of the angular damp of the body and this value, or the area's value the body is in if the body defaults to combine damp values. (see :ref:`DampMode<enum_RigidBody2D_DampMode>`)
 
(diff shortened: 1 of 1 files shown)


AThousandShips:

thygrrr:
Oh crap sorry about that.

thygrrr:
```xml
		<member name="physics/2d/default_angular_damp" type="float" setter="" getter="" default="1.0">
			The default angular damp in 2D.
			[b]Note:[/b] Good values are in the range [code]0[/code] to [code]1[/code]. At value [code]0[/code] objects will keep moving with the same velocity. Values greater than [code]1[/code] will aim to reduce the velocity to [code]0[/code] in less than a second e.g. a value of [code]2[/code] will aim to reduce the velocity to [code]0[/code] in half a second. A value equal to or greater than the physics frame rate ([member ProjectSettings.physics/common/physics_ticks_per_second], [code]60[/code] by default) will bring the object to a stop in one iteration.
		</member>
```

How do I use proper sphinx admonishments from here?

AThousandShips:
It says it is generated at the very top of the document :)

See the instructions [here](https://docs.godotengine.org/en/latest/contributing/documentation/class_reference_primer.html)

thygrrr:
Thank you for your kind advice, I've submitted PR https://github.com/godotengine/godot/pull/82950 instead. :)

Issue #6622, opened 2023-01-16
How do we avoid falling into line by line translation from GDScript to C#?


paulloz:
Hey 🙂 

I was reading through the recent updates in the documentation and stumbled on a C# example that IMHO warrants a bit of discussion. I noticed it due to the recent PR on the CharacterBody2D page (#6614), but the part that bugs me was already there in the 3.x documentation.

So, on this page, we find the following snippet of code.

```cs
public override void _PhysicsProcess(double delta)
{
    var collision = MoveAndCollide(Velocity * (float)delta);
    if (collision != null)
    {
        Velocity = Velocity.Bounce(collision.GetNormal());
        if (collision.GetCollider().HasMethod("Hit"))
        {
            collision.GetCollider().Call("Hit");
        }
    }
}
```

It is (as almost always) a direct translation of the GDScript example. More often than not, converting examples this way is fine. However, here, we are taking a very "duck typing-y" approach of what was returned by `collision.GetCollider()`... And I'd argue presenting this as a good approach in C# could be harmful (I lack a better word here) to the reader. In this particular instance, I guess most people would write this using an interface. But we don't really elaborate on what the object containing that `Hit()` method would be on this page, so I don't know if there's a point.

Anyway, I'm of course not saying we need to tailor every single C# code block completely differently than their GDScript counterpart. But I was curious of other people's opinion about this, if we should try to "correct" this kind of thing when we see them or if I'm just entirely delusional.

raulsntos:
The problem is that in a tutorial/manual, while you are able to provide code in multiple languages, **the text stays the same** and that makes it challenging if you need to describe completely different approaches depending on the language.

There are plenty of examples of this in the documentation, see the [_Saving Games_](https://docs.godotengine.org/en/latest/tutorials/io/saving_games.html) page, which uses duck typing and Godot's JSON serializer while it may better to use interfaces and one of .NET's more powerful serializers.

I've been thinking about adding notes throughout the documentation that mention related .NET APIs that users may want to consider so that, at the very least, they are aware that there are more options out there. Little notes in the documentation will hopefully not be too in your face for non-C# users, while still providing additional information for C# users. I don't know if this would be very helpful though.

[...]

PR #8177, opened 2023-10-06
Standardize/Improve C# GD.Load() examples

wlsnmrk:
Following on from #7193, I've searched the docs for examples of `GD.Load()` that might not illustrate its requirement for absolute paths, fixed them to include a `res://` prefix, and made a few other changes:
* To standardize on the C# style, filenames and directories have been converted to PascalCase, except when loading GDScript files, and in one case a GLSL file (see below).
* I've changed to the typed version of `GD.Load()` (e.g., `GD.Load<PackedScene>()`) when examples were casting the result of an untyped call.
* I've used a `res://Path/To/FileName.extension` convention when the corresponding GDScript did not have an absolute path specified, to illustrate that the file might live in an arbitrary directory. When the GDScript example did have a `res://`, I've used the same absolute path.

Specific items I'd like to point up for review:
* `tutorials/shaders/compute_shaders.rst`: 
    * [I left a GLSL filename snake_cased](https://github.com/wlsnmrk/godot-docs/blob/45c6a1a4e4a5a1efddc6772c501e4f9411146f71/tutorials/shaders/compute_shaders.rst?plain=1#L187). I'm not a GLSL expert, but from some quick searching it seems like snake_case is kind of a de-facto standard. Happy to be corrected.
* `tutorials/best_practices/scenes_versus_scripts.rst`:
    * Required some significant changes in [the example for loading a script and assigning it to a `Node`](https://github.com/godotengine/godot-docs/compare/master...wlsnmrk:godot-docs:gd-load-fixes#diff-10a5f1b0589fa7ad1b08151c1dcd11bb5228644a9fbc421804f65bbece39b2d1R169-R183). I've verified that the updated version runs correctly in a test project, but wouldn't mind if someone double-checked my work.
        * Corresponding GDScript example was a `Node`, so I've changed C# from a `Resource` to a `Node`
        * `SetScript()` disposes the C# wrapper object, and the call comes before the node is in the tree, so I've added the object-ID workaround from https://github.com/godotengine/godot/issues/31994 to obtain a new reference
        * Moved the `AddChild()` call to before setting `Owner` on the child, per [documentation on `Owner`](https://docs.godotengine.org/en/stable/classes/class_node.html#class-node-property-owner)
    * [Changed a comment to hopefully make its meaning a little more clear](https://github.com/godotengine/godot-docs/compare/master...wlsnmrk:godot-docs:gd-load-fixes#diff-10a5f1b0589fa7ad1b08151c1dcd11bb5228644a9fbc421804f65bbece39b2d1R55), following the GDScript example.
* `tutorials/best_practices/godot_interfaces.rst`:
    * ~~I've left a field loading a snake_case GDScript rather than changing it to load a `.cs` script [in one example](https://github.com/wlsnmrk/godot-docs/blob/45c6a1a4e4a5a1efddc6772c501e4f9411146f71/tutorials/best_practices/godot_interfaces.rst?plain=1#L90), since it's loading from a (notional) add-on and those tend to be GDScript rather than C#. However, there doesn't seem to be a corresponding value in the GDSCript example, and I'm not sure if this example really needs to illustrate add-on directory layouts or cross-language scripting. Would be happy to change this to something else.~~ Changed below.
* `tutorials/scripting/resources.rst`:
    * [Changed a comment](https://github.com/godotengine/godot-docs/compare/master...wlsnmrk:godot-docs:gd-load-fixes#diff-9543c0c477dd21ce6356f57c792659ee49439568e90442bc16cf7df93be83461R90) to describe Godot loading the resource when the `GD.Load()` line is _executed_ rather than _read_. I'm not sure what the semantics of this are on the GDScript side, but I believe this to be more correct for C#; happy to be corrected.
    * [I've changed the filename "robi.png" to PascalCase](https://github.com/godotengine/godot-docs/compare/master...wlsnmrk:godot-docs:gd-load-fixes#diff-9543c0c477dd21ce6356f57c792659ee49439568e90442bc16cf7df93be83461R91) (i.e., "Robi.png"), in keeping with all the other changes. However, since it's referred to as "robi.png" in the body of the page, I'm not sure if we should make an exception to the naming convention here to avoid confusion.
* `tutorials/scripting/cross_language_scripting.rst`:
    * [Changed the path for loading a C# script from a GDScript to use PascalCase](https://github.com/godotengine/godot-docs/compare/master...wlsnmrk:godot-docs:gd-load-fixes#diff-7875f3ee292a53a639d2a0c22b902d99c0c356f9ad9339b1abee1f9209d0e2d8R77), to match the corresponding example for loading a GDScript from a C# script, which uses snake_case.

### DIFF
--- a/tutorials/best_practices/scenes_versus_scripts.rst
+++ b/tutorials/best_practices/scenes_versus_scripts.rst
@@ -29,9 +29,9 @@ a change in API:
     const MyNode = preload("my_node.gd")
     const MyScene = preload("my_scene.tscn")
     var node = Node.new()
-    var my_node = MyNode.new() # Same method call
-    var my_scene = MyScene.instantiate() # Different method call
-    var my_inherited_scene = MyScene.instantiate(PackedScene.GEN_EDIT_STATE_MAIN) # Create scene inheriting from MyScene
+    var my_node = MyNode.new() # Same method call.
+    var my_scene = MyScene.instantiate() # Different method call.
+    var my_inherited_scene = MyScene.instantiate(PackedScene.GEN_EDIT_STATE_MAIN) # Create scene inheriting from MyScene.
 
   .. code-tab:: csharp
 
@@ -39,8 +39,10 @@ a change in API:
 
     public partial class Game : Node
     {
-        public static CSharpScript MyNode { get; } = GD.Load<CSharpScript>("MyNode.cs");
-        public static PackedScene MyScene { get; } = GD.Load<PackedScene>("MyScene.tscn");
+        public static CSharpScript MyNode { get; } =
+            GD.Load<CSharpScript>("res://Path/To/MyNode.cs");
+        public static PackedScene MyScene { get; } =
+            GD.Load<PackedScene>("res://Path/To/MyScene.tscn");
         private Node _node;
         private Node _myNode;
         private Node _myScene;
@@ -50,8 +52,10 @@ a change in API:
         {
             _node = new Node();
             _myNode = MyNode.New().As<Node>();
-            _myScene = MyScene.Instantiate(); // Different. Instantiated from a PackedScene
-            _myInheritedScene = MyScene.Instantiate(PackedScene.GenEditState.Main); // Create scene inheriting from MyScene
+            // Different than calling new() or MyNode.New(). Instantiated from a PackedScene.
+            _myScene = MyScene.Instantiate();
+            // Create scene inheriting from MyScene.
+            _myInheritedScene = MyScene.Instantiate(PackedScene.GenEditState.Main);
         }
     }
 
@@ -155,14 +159,14 @@ with it, and finally adds it as a child of the ``Main`` node:
         var child = Node.new()
         child.name = "Child"
         child.script = preload("child.gd")
-        child.owner = self
         add_child(child)
+        child.owner = self
 
   .. code-tab:: csharp
 
     using Godot;
 
-    public partial class Main : Resource
+    public partial class Main : Node
     {
         public Node Child { get; set; }
 
@@ -170,9 +174,13 @@ with it, and finally adds it as a child of the ``Main`` node:
         {
             Child = new Node();
             Child.Name = "Child";
-            Child.SetScript(GD.Load<Script>("child.gd"));
-            Child.Owner = this;
+            var childID = Child.GetInstanceId();
+            Child.SetScript(GD.Load<Script>("res://Path/To/Child.cs"));
+            // SetScript() causes the C# wrapper object to be disposed, so obtain a new
+            // wrapper for the Child node using its instance ID before proceeding.
+            Child = (Node)GodotObject.InstanceFromId(childID);
             AddChild(Child);
+            Child.Owner = this;
         }
     }
 
@@ -207,7 +215,7 @@ In the end, the best approach is to consider the following:
     .. code-tab:: gdscript GDScript
 
       # game.gd
-      class_name Game # extends RefCounted, so it won't show up in the node creation dialog
+      class_name Game # extends RefCounted, so it won't show up in the node creation dialog.
       extends RefCounted
 
       const MyScene = preload("my_scene.tscn")
@@ -222,7 +230,8 @@ In the end, the best approach is to consider the following:
       // Game.cs
       public partial class Game : RefCounted
       {
-          public static PackedScene MyScene { get; } = GD.Load<PackedScene>("MyScene.tscn");
+          public static PackedScene MyScene { get; } =
+              GD.Load<PackedScene>("res://Path/To/MyScene.tscn");
       }
 
       // Main.cs
--- a/tutorials/best_practices/godot_interfaces.rst
+++ b/tutorials/best_practices/godot_interfaces.rst
@@ -80,14 +80,14 @@ access.
         // No "preload" loads during scene load exists in C#.
 
         // Initialize with a value. Editable at runtime.
-        public Script MyScript = GD.Load<Script>("MyScript.cs");
+        public Script MyScript = GD.Load<Script>("res://Path/To/MyScript.cs");
 
         // Initialize with same value. Value cannot be changed.
-        public readonly Script MyConstScript = GD.Load<Script>("MyScript.cs");
+        public readonly Script MyConstScript = GD.Load<Script>("res://Path/To/MyScript.cs");
 
         // Like 'readonly' due to inaccessible setter.
         // But, value can be set during constructor, i.e. MyType().
-        public Script Library { get; } = GD.Load<Script>("res://addons/plugin/library.gd");
+        public Script MyNoSetScript { get; } = GD.Load<Script>("res://Path/To/MyScript.cs");
 
         // If need a "const [Export]" (which doesn't exist), use a
         // conditional setter for a tool script that checks if it's executing
--- a/tutorials/best_practices/logic_preferences.rst
+++ b/tutorials/best_practices/logic_preferences.rst
@@ -96,7 +96,7 @@ either? Let's see an example:
         public override void _Ready()
         {
             // Can assign the value during initialization.
-            ABuilding = GD.Load<PackedScene>("res://office.tscn");
+            ABuilding = GD.Load<PackedScene>("res://Office.tscn");
         }
     }
 
--- a/tutorials/physics/using_character_body_2d.rst
+++ b/tutorials/physics/using_character_body_2d.rst
@@ -337,7 +337,7 @@ uses the mouse pointer. Here is the code for the Player, using ``move_and_slide(
 
     public partial class MyCharacterBody2D : CharacterBody2D
     {
-        private PackedScene _bullet = (PackedScene)GD.Load("res://bullet.tscn");
+        private PackedScene _bullet = GD.Load<PackedScene>("res://Bullet.tscn");
         private int _speed = 200;
 
         public void GetInput()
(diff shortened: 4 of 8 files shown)


tutorials/best_practices/scenes_versus_scripts.rst
```diff
@@ -39,8 +39,10 @@ a change in API:
 
     public partial class Game : Node
     {
-        public static CSharpScript MyNode { get; } = GD.Load<CSharpScript>("MyNode.cs");
-        public static PackedScene MyScene { get; } = GD.Load<PackedScene>("MyScene.tscn");
+        public static CSharpScript MyNode { get; } = 
+            GD.Load<CSharpScript>("res://Path/To/MyNode.cs");
+        public static PackedScene MyScene { get; } = 
+            GD.Load<PackedScene>("res://Path/To/MyScene.tscn");
```
raulsntos, line 45: Trailing whitespace.

```suggestion
        public static CSharpScript MyNode { get; } =
            GD.Load<CSharpScript>("res://Path/To/MyNode.cs");
        public static PackedScene MyScene { get; } =
            GD.Load<PackedScene>("res://Path/To/MyScene.tscn");
```

tutorials/best_practices/scenes_versus_scripts.rst
```diff
@@ -222,7 +228,8 @@ In the end, the best approach is to consider the following:
       // Game.cs
       public partial class Game : RefCounted
       {
-          public static PackedScene MyScene { get; } = GD.Load<PackedScene>("MyScene.tscn");
+          public static PackedScene MyScene { get; } = 
+              GD.Load<PackedScene>("res://Path/To/MyScene.tscn");
```
raulsntos, line 232: Trailing whitespace.

```suggestion
          public static PackedScene MyScene { get; } =
              GD.Load<PackedScene>("res://Path/To/MyScene.tscn");
```

raulsntos:
> I left a GLSL filename snake_cased. [...] it seems like snake_case is kind of a de-facto standard.

I don't know but I feel like a C# user would probably use PascalCase for those files too. But it's fine to leave it out of the PR for now.

> I've left a field loading a snake_case GDScript rather than changing it to load a `.cs` script in [one example](https://github.com/wlsnmrk/godot-docs/blob/45c6a1a4e4a5a1efddc6772c501e4f9411146f71/tutorials/best_practices/godot_interfaces.rst?plain=1#L90), since it's loading from a (notional) add-on and those tend to be GDScript rather than C#. However, there doesn't seem to be a corresponding value in the GDSCript example, and I'm not sure if this example really needs to illustrate add-on directory layouts or cross-language scripting.

I don't think this example is meant to show cross-language scripting, could probably be replaced with `res://Path/To/MyScript.cs` like the other members.

> I've changed the filename "robi.png" to PascalCase (i.e., "Robi.png"), in keeping with all the other changes. However, since it's referred to as "robi.png" in the body of the page, I'm not sure if we should make an exception to the naming convention here to avoid confusion.

I think there are plenty of C# examples in the documentation that do this already. For example, in [_Your first 2D game: Coding the player_](https://docs.godotengine.org/en/stable/getting_started/first_2d_game/03.coding_the_player.html) there are mentions to methods and signals by their GDScript name in the text, the screenshots also show a GDScript project.

It's true that this is confusing to users. We try to address it in the C# documentation, explaining that there are different naming conventions in C# and that they may have to do some "translation" when reading the class reference and tutorials. It's not ideal, but it's a complicated problem to solve (somewhat related issues about it: https://github.com/godotengine/godot-docs/issues/6622 https://github.com/godotengine/godot-docs/issues/6596).

tutorials/best_practices/scenes_versus_scripts.rst
```diff
@@ -162,17 +166,19 @@ with it, and finally adds it as a child of the ``Main`` node:
...
 
         public Main()
         {
             Child = new Node();
             Child.Name = "Child";
-            Child.SetScript(GD.Load<Script>("child.gd"));
-            Child.Owner = this;
+            var childID = Child.GetInstanceId();
+            Child.SetScript(GD.Load<Script>("res://Path/To/Child.cs"));
+            Child = (Node)GodotObject.InstanceFromId(childID);
```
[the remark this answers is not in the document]
wlsnmrk: Done; hopefully this is sufficient?

tutorials/best_practices/scenes_versus_scripts.rst
```diff
@@ -162,17 +166,19 @@ with it, and finally adds it as a child of the ``Main`` node:
...
             AddChild(Child);
+            Child.Owner = this;
```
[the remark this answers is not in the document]
wlsnmrk: Good catch, thanks!

wlsnmrk:
> I don't know but I feel like a C# user would probably use PascalCase for those files too. But it's fine to leave it out of the PR for now.

Okay -- I've left it in snake_case for the time being.

> I don't think this example is meant to show cross-language scripting, could probably be replaced with `res://Path/To/MyScript.cs` like the other members.

Done, and changed the name of the field since `Library` didn't really make sense anymore.

> It's true that this is confusing to users. We try to address it in the C# documentation, explaining that there are different naming conventions in C# and that they may have to do some "translation" when reading the class reference and tutorials. It's not ideal, but it's a complicated problem to solve (somewhat related issues about it: #6622 #6596).

Fair enough; I'll leave it with the capital letter for the moment. It does seem like a difficult problem to solve generally, though I'm in agreement that a separate class reference at least would be very helpful.

raulsntos (approved):
Looks good to me, thanks!

tutorials/best_practices/scenes_versus_scripts.rst
```diff
@@ -50,8 +52,10 @@ a change in API:
         {
             _node = new Node();
             _myNode = MyNode.New().As<Node>();
-            _myScene = MyScene.Instantiate(); // Different. Instantiated from a PackedScene
-            _myInheritedScene = MyScene.Instantiate(PackedScene.GenEditState.Main); // Create scene inheriting from MyScene
+            // Different than calling new() or MyNode.New(). Instantiated from a PackedScene
```
AThousandShips, line 55:
```suggestion
            // Different than calling new() or MyNode.New(). Instantiated from a PackedScene.
```
Per the [comment style](https://docs.godotengine.org/en/latest/contributing/development/code_style_guidelines.html#comment-style-guide).

tutorials/best_practices/scenes_versus_scripts.rst
```diff
@@ -50,8 +52,10 @@ a change in API:
...
+            _myScene = MyScene.Instantiate();
+            // Create scene inheriting from MyScene
```
AThousandShips, line 57:
```suggestion
            // Create scene inheriting from MyScene.
```

tutorials/best_practices/scenes_versus_scripts.rst
```diff
@@ -155,24 +159,28 @@ with it, and finally adds it as a child of the ``Main`` node:
...
         public Main()
         {
             Child = new Node();
             Child.Name = "Child";
-            Child.SetScript(GD.Load<Script>("child.gd"));
-            Child.Owner = this;
+            var childID = Child.GetInstanceId();
+            Child.SetScript(GD.Load<Script>("res://Path/To/Child.cs"));
+            // SetScript() causes the C# wrapper object to be disposed, so obtain a new
+            // wrapper for the Child node using its instance ID before proceeding
```
AThousandShips, line 180:
```suggestion
            // wrapper for the Child node using its instance ID before proceeding.
```

tutorials/best_practices/scenes_versus_scripts.rst
wlsnmrk: Should I also fix the corresponding GDScript comments where applicable, or just stick to the updated C# comments?

AThousandShips: Can do them too where applicable :)

wlsnmrk: I started looking over the files I'd modified for comment formatting, and spotted a chance to get rid of some foo/bar examples in the cross-language scripting tutorial. However, those go hand-in-hand with several spots where a comment is used to indicate the expected output of a print line (the raw string, no quotes). ([Here's one example](https://github.com/godotengine/godot-docs/blob/9ce9593ee38b008ca325ec277118318e49c37bbc/tutorials/scripting/cross_language_scripting.rst?plain=1#L121).) 

Two questions:
* Should those instances be updated, and if so, what would be the appropriate format for that? Something like `Expected output is "My Value 1".`?
* If they should be updated, should I open another PR for that? Seems like this issue is drifting away from the original purpose here.

AThousandShips: I'd say to deal with it separately indeed

Generally error prints are supposed to also end with a period, but it's far from uniform, and don't think it's relevant here, at least not to justify doing a sweeping fix

wlsnmrk:
Additional changes beyond the suggestions:
* Comment fixes in the same vein as the reviewed lines, in `singletons_autoload.rst` and `cross_language_scripting.rst`
* Realized the C# file name I introduced in `cross_langugage_scripting.rst` didn't match the type name, so fixed that
* Removed a GDScript line in `cross_language_scripting.rst` that referred to a nonexistent C# field.

I think there are more changes that could be made to bring the `cross_language_scripting.rst` C# and GDScript examples more in line with each other and their respective style guides, but I'll leave that for a different PR.

mhilbrunner (approved):

mhilbrunner:
Thanks, merged :)

PR #8184, opened 2023-10-07
Fix Rotation-Pivot if Camera is Behind Player

GigglingGalaxy:
JUST one NOTE in the page of tutorial for Player Movement--this is my first attempt at Pull Req so please help me fix/do better next time!!

Added note for PIVOT line in GD Script. Why: In 4.1 CB3D is now rotated 180...so when rotating CharacterBody3D in direction of movement, if you want 3rd person view from behind the player, you need to subtract (not add) here: $Pivot.look_at(position - direction, Vector3.UP) 
This way, your player will correctly rotate with movement. 

I think a lot of people following this tutorial will be adapting it to have their camera behind the player (3rd Person View) and will needlessly become frustrated trying to rotate actual game objects instead of one simple code fix.

### DIFF
--- a/getting_started/first_3d_game/03.player_movement_code.rst
+++ b/getting_started/first_3d_game/03.player_movement_code.rst
@@ -441,6 +441,21 @@ the ground should fill the background.
 
 Test your scene and you should be able to move in all 8 directions and not glitch through the floor!
 
+.. note::
+
+    If you want your camera behind your CharacterBody3D (you don't want the player 
+    facing the camera), and you want your Character to rotate in the direction of 
+    movement, you can change the GD Player Script *Pivot* code to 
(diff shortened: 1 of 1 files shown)


AThousandShips:
Your PR is based on the wrong version, you need to have the PR based on `master` and target `master

AThousandShips:
Updated your branch

skyace65:
Closing this PR as the note is a bit out of scope of the getting started tutorial. The purpose of getting started is to show people how the engine works. This feels like we'd be telling people "here's how you can use this code in your own project", and that's not something we necessarily want to be worried about when keeping this tutorial up to date. I would advise people to check youtube tutorials for this kind of stuff instead, there's tons of great Godot content creators.

PR #8197, opened 2023-10-09
Updated 3D tutorial to Godot 4.1.1

adambiser:
The way to copy animations appears to be different now.

I have followed the 3D tutorial using Godot 4.1.1 and it appears to be up-to-date except for the change in this pull request.

I notice that "v3.2.4.beta2.official" appears in the Project Manager image under "Setting up the game area".


skyace65 (commented):
While this does copy the animation, I think it would be better to explain how to use the animation libraries instead (where you get to by clicking "manage animations")

adambiser:
My intention was to update the existing tutorial text to Godot 4.1.1.
Adding to or expanding the tutorial might be better as another pull request and by someone with more experience with Godot than I have.

skyace65:
Superseded by #8955.

PR #8202, opened 2023-10-09
Improving unit test documentation page

k0T0z:
I think it makes more sense to mention about compiler version inside the unit testing page.

### DIFF
--- a/contributing/development/core_and_modules/unit_testing.rst
+++ b/contributing/development/core_and_modules/unit_testing.rst
@@ -55,6 +55,12 @@ arguments for doctest.
     detected, so you should strive to fix all warnings before opening a pull
     request.
 
+.. note::
+
+    If you are using a 64-bit compiler architecture, ``<godot_binary>`` will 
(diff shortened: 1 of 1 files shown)


Chaosus:
Commits needs to be squashed (see https://docs.godotengine.org/en/latest/contributing/workflow/pr_workflow.html#modifying-a-pull-request)

Calinou:
I'm not sure this comment is very useful, as "run the correct binary" is rather obvious. Also, the amount of people compiling both 64-bit and 32-bit binaries during development is pretty low (you generally need to build 32-bit only for distribution, and only for Windows).

k0T0z:
@Calinou Fair enough. Yeah I kept running the wrong binary and wondered why the tests are not running and @Chaosus helped me with it. It is intended to maybe save some time but I don't if it's the right way of writing this note so. The final decision is yours.

mhilbrunner:
Closing as per the above discussion. If someone else runs into this, we may want to reconsider. :) Thanks for contributing!

k0T0z:
That's fine. Thank you.

PR #8206, opened 2023-10-09
Fix minor mistake

mongolsteppe:
Fixed reference to a non-existing CapsuleShape2D, which might confuse beginners following the tutorial


mongolsteppe:
I misunderstood the tutorial, my bad.

Issue #7498, opened 2023-06-11
Epub documentation fails to render, throws errors


kevzettler:
**Your Godot version:**
Tried to download epub documentation from: https://nightly.link/godotengine/godot-docs/workflows/build_offline_docs/master/godot-docs-epub-master.zip

**Issue description:**
Every page of the epub renders an error:

```
“This page contains the following errors:error on line 45 at column 10: Opening and ending tag mismatch: meta line 0 and head
Below is a rendering of the page up to the first error.”

Excerpt From
Godot Engine
Juan Linietsky, Ariel Manzur and the Godot community
This material may be protected by copyright.
```

Fails to render on Apple Books and Kindle

![image](https://github.com/godotengine/godot-docs/assets/126209/df5f188a-0a23-498a-961e-32601176deff)

eimfach:
I would like to look into the issue and fix it, where should I start ? Is it just the meta and head tags ? Looks like a small issue on the surface ..

skyace65:
This was fixed by #8207.

PR #8207, opened 2023-10-09
Fix epub-related failures

dylannorthrup:
# Motivation
I ran into the same problems as described in #7498 and looked into the issue. The underlying errors primarily stemmed from three repeated issues:
* Having multiple instances of `<meta name="description" ....>` in the same page (which is [against W3.org standards](https://www.w3.org/TR/2014/REC-html5-20141028/document-metadata.html#meta-description))
* Having unclosed `<meta>` tags
* Having an extra `</div>` tag

EDIT: Removed all module upgrades to reduce the scope of this PR.
~~In addition, the versions of modules in the current `requirements.txt` did not have the correct mime-type for `.webp` images and thus could not handle them properly.  And, after updating the modules, running `make epub` encountered a `ThemeError` due to some included CSS and JS file paths having parameters.~~

# Changes
This PR makes the following updates:
* Alter `_extensions/godot_descriptions.py` to skip adding a new `<meta name="description"...>` tag if one already exists in `context["metatags"]`
* Alter `_extensions/godot_descriptions.py` and `_templates/layout.html` to modify `<meta ...>` tags to `<meta ... />` (i.e. having the meta element close itself)
* Alter `_templates/layout.html` to remove the extra `</div>` element on line 69 (which was chosen over the `</div>` on line 75 because the indentation on the comments block appeared to indicate it should be inside the `document` block)
* ~~Update the `css/custom.css` and `js/custom.js` paths to remove appended parameters (which the comments indicated were intended to be used for cache busting after updates).~~
* ~~Update all versions in `requirements.txt` to match the versions in the [pip.txt](https://github.com/readthedocs/readthedocs.org/blob/main/requirements/pip.txt) and [docs.txt](https://github.com/readthedocs/readthedocs.org~~/blob/main/requirements/docs.txt) in the main branch of the RTD repo~~
  * ~~[pygments](https://github.com/readthedocs/readthedocs.org/blob/7440c7bdac0e80f872188bccae7b020409dcd580/requirements/pip.txt#L192)~~
  * ~~[sphinx](https://github.com/readthedocs/readthedocs.org/blob/7440c7bdac0e80f872188bccae7b020409dcd580/requirements/docs.txt#L94)~~
  * ~~[sphinx_rtd_theme](https://github.com/readthedocs/readthedocs.org/blob/7440c7bdac0e80f872188bccae7b020409dcd580/requirements/docs.txt#L132)~~
  * ~~[sphinx-copybutton](https://github.com/readthedocs/readthedocs.org/blob/7440c7bdac0e80f872188bccae7b020409dcd580/requirements/docs.txt#L118C1-L118C1)~~
  * ~~[sphinx-notfound-page](https://github.com/readthedocs/readthedocs.org/blob/7440c7bdac0e80f872188bccae7b020409dcd580/requirements/docs.txt#L128)~~
  * ~~[sphinxext-opengraph](https://github.com/readthedocs/readthedocs.org/blob/7440c7bdac0e80f872188bccae7b020409dcd580/requirements/docs.txt#L158)~~

# Results
I have been able to generate a valid EPUB file that renders properly on my Mac in `Books.app`. I have not tested generation of any other documentation. 

EDIT: Because the earlier version of the `sphinx` module does not have a valid mimetype for `*.webp` images, building an EPUB file will generate several of the following errors:

```
WARNING: unknown mimetype for _images/IMAGE_FILENAME.webp, ignoring
```

Those images will not be included in the resulting EPUB file, but I was unable to find a way to update the `mime_suffixes` list [configured in the sphinx module](https://github.com/sphinx-doc/sphinx/blob/bb74aec2b6aa1179868d83134013450c9ff9d4d6/sphinx/util/images.py#L19-L26).

~~I continue to get a `WARNING: [Social card] image cannot be an SVG image, skipping...` error when doing EPUB generation, but I was unable to identify where this was coming from and it does not appear to have caused any errors in the resulting EPUB file.~~

### DIFF
--- a/_extensions/godot_descriptions.py
+++ b/_extensions/godot_descriptions.py
@@ -103,10 +103,11 @@ def generate_description(app, pagename, templatename, context, doctree):
     doctree.walkabout(generator)
 
     description = (
-        '<meta name="description" content="' + generator.create_description() + '">\n'
+        '<meta name="description" content="' + generator.create_description() + '" />\n'
     )
 
-    context["metatags"] += description
+    if not re.search('<meta name="description"', context["metatags"]):
+        context["metatags"] += description
 
 
 def setup(app):
--- a/_extensions/godot_descriptions.py
+++ b/_extensions/godot_descriptions.py
@@ -106,7 +106,7 @@ def generate_description(app, pagename, templatename, context, doctree):
         '<meta name="description" content="' + generator.create_description() + '" />\n'
     )
 
-    if not re.search('<meta name="description"', context["metatags"]):
+    if not '<meta name="description"' in context["metatags"]:
         context["metatags"] += description
 
 
--- a/_templates/layout.html
+++ b/_templates/layout.html
@@ -6,12 +6,12 @@
 {% endblock -%}
 
 {% block extrahead -%}
-<meta name="doc_version" content="{{ version }}">
-<meta name="doc_is_latest" content="{{ godot_is_latest }}">
+<meta name="doc_version" content="{{ version }}" />
+<meta name="doc_is_latest" content="{{ godot_is_latest }}" />
 {% endblock -%}
 
 {% block linktags -%}
-  <meta name="theme-color" content="#3d8fcc">
+  <meta name="theme-color" content="#3d8fcc" />
   {% if godot_inject_language_links -%}
   {% for alternate_lang in godot_docs_supported_languages -%}
   {# Convert to ISO 639-1 format, e.g. zh_CN -> zh-cn -#}
@@ -66,7 +66,6 @@
   {% endif %}
 
   {% block body %}{% endblock %}
-</div>
 {%- if self.comments()|trim %}
   <div class="articleComments">
     {%- block comments %}{% endblock %}
--- a/conf.py
+++ b/conf.py
@@ -199,14 +199,14 @@ html_extra_path = ["robots.txt"]
 html_css_files = [
     'css/algolia.css',
     'https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.css',
-    "css/custom.css?10", # Increment the number at the end when the file changes to bust the cache.
+    "css/custom.css",
 ]
 
 if not on_rtd:
     html_css_files.append("css/dev.css")
 
 html_js_files = [
-    "js/custom.js?6", # Increment the number at the end when the file changes to bust the cache.
+    "js/custom.js",
     ('https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.js', {'defer': 'defer'}),
     ('js/algolia.js', {'defer': 'defer'})
 ]
--- a/conf.py
+++ b/conf.py
@@ -199,14 +199,14 @@ html_extra_path = ["robots.txt"]
 html_css_files = [
     'css/algolia.css',
     'https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.css',
-    "css/custom.css",
+    "css/custom.css?10", # Increment the number at the end when the file changes to bust the cache.
 ]
 
 if not on_rtd:
     html_css_files.append("css/dev.css")
 
 html_js_files = [
-    "js/custom.js",
+    "js/custom.js?6", # Increment the number at the end when the file changes to bust the cache.
     ('https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.js', {'defer': 'defer'}),
     ('js/algolia.js', {'defer': 'defer'})
 ]
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,7 @@
 # Sync with readthedocs:
 # https://github.com/readthedocs/readthedocs.org/blob/master/requirements/pip.txt
 # https://github.com/readthedocs/readthedocs.org/blob/master/requirements/docs.txt
+# Noop
 
 # Base dependencies
 pygments==2.16.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,22 +1,21 @@
 # Sync with readthedocs:
 # https://github.com/readthedocs/readthedocs.org/blob/master/requirements/pip.txt
 # https://github.com/readthedocs/readthedocs.org/blob/master/requirements/docs.txt
-# Noop
 
 # Base dependencies
-pygments==2.16.1
+pygments==2.15.1
 
 # Sphinx base and RTD theme.
-sphinx==7.2.6
-sphinx_rtd_theme==2.0.0rc2
+sphinx==4.4.0
+sphinx_rtd_theme==1.1.1
 
 # Sphinx extensions.
 
 # Code tabs extension to display codeblocks in different languages as tabs.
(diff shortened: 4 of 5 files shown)


conf.py
```diff
@@ -199,14 +199,14 @@
 html_css_files = [
     'css/algolia.css',
     'https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.css',
-    "css/custom.css?10", # Increment the number at the end when the file changes to bust the cache.
```
mhilbrunner, line 202: I think these cache busters are still used when deploying changes to the website CSS to get around long caching by the CDN/RTD. I don't think we can just remove them for the sake of EPUB, cc @YuriSizov

mhilbrunner:
Thanks for contributing! No time to test this yet, but the changes look good besides the one that I noted above at first glance. CI also fails on the requirements.txt.

_extensions/godot_descriptions.py
```diff
@@ -103,10 +103,11 @@ def generate_description(app, pagename, templatename, context, doctree):
     doctree.walkabout(generator)
 
     description = (
-        '<meta name="description" content="' + generator.create_description() + '">\n'
+        '<meta name="description" content="' + generator.create_description() + '" />\n'
```
mhilbrunner, line 106: Void elements like `meta` should only specify a start tage, but not self-close: https://html.spec.whatwg.org/#void-elements

The common recommendation everywhere seems to be to just use `<meta>`, and outside of XHTML that always was valid. The majority of the web content out there seems to do this. Does this specific thing really break EPUB? That seems weird.

See also this discussion: https://stackoverflow.com/questions/4693939/self-closing-tags-void-elements-in-html5

_templates/layout.html
```diff
@@ -6,12 +6,12 @@
 {% endblock -%}
 
 {% block extrahead -%}
-<meta name="doc_version" content="{{ version }}">
-<meta name="doc_is_latest" content="{{ godot_is_latest }}">
+<meta name="doc_version" content="{{ version }}" />
+<meta name="doc_is_latest" content="{{ godot_is_latest }}" />
```
mhilbrunner, line 10: Same re: self-closing tags here.

_templates/layout.html
```diff
@@ -6,12 +6,12 @@
...
 {% endblock -%}
 
 {% block linktags -%}
-  <meta name="theme-color" content="#3d8fcc">
+  <meta name="theme-color" content="#3d8fcc" />
```
mhilbrunner, line 14: Same re: self-closing tags here.

conf.py
mhilbrunner: Confirmed we still need and use these, so removing it won't work. Does this really break EPUB? It is a perfectly valid URL after all.

conf.py
```diff
@@ -199,14 +199,14 @@
...
+    "css/custom.css",
 ]
 
 if not on_rtd:
     html_css_files.append("css/dev.css")
 
 html_js_files = [
-    "js/custom.js?6", # Increment the number at the end when the file changes to bust the cache.
+    "js/custom.js",
```
mhilbrunner, line 209: Same re: cache busters as discussed above.

requirements.txt
```diff
@@ -1,21 +1,22 @@
 # Sync with readthedocs:
 # https://github.com/readthedocs/readthedocs.org/blob/master/requirements/pip.txt
 # https://github.com/readthedocs/readthedocs.org/blob/master/requirements/docs.txt
+# Noop
```
mhilbrunner, line 4: I think this was included accidentally :) Needs to be removed.

mhilbrunner (changes requested):
I had some more time to have another look and leave a few comments.

I agree upgrading our dependencies is worthwhile (and planned for some time), though doing it in this PR would mean this is likely going to need a lot of time for review and proper testing, seeing as it jumps multiple major versions for core things like Sphinx.

_extensions/godot_descriptions.py
dylannorthrup: Per the [W3C Recommendations on HTML5](https://www.w3.org/TR/2014/REC-html5-20141028/syntax.html) section on [Start tags](https://www.w3.org/TR/2014/REC-html5-20141028/syntax.html#start-tags), void elements can be self-closing (which appears to be different than having an end-tag). Also, the [examples for `meta` on developer.mozilla.org](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#examples) are self-closing.

> The common recommendation everywhere seems to be to just use <meta>, and outside of XHTML that always was valid. The majority of the web content out there seems to do this. Does this specific thing really break EPUB? That seems weird.

Per the w3c epub docs, the [epub format is XHTML](https://idpf.org/epub/30/spec/epub30-contentdocs.html#sec-overview-relations-html) that is ["a profile of HTML5"](https://idpf.org/epub/30/spec/epub30-contentdocs.html#sec-xhtml).

_templates/layout.html
dylannorthrup: Same re: epub is XHTML

_extensions/godot_descriptions.py
mhilbrunner: Thanks for the detailed reponse, that makes sense. :)

Calinou (approved):
Tested locally by viewing the generated ePub in Okular 23.08.1, it works as expected.

_extensions/godot_descriptions.py
```diff
@@ -103,10 +103,11 @@ def generate_description(app, pagename, templatename, context, doctree):
...
     )
 
-    context["metatags"] += description
+    if not re.search('<meta name="description"', context["metatags"]):
```
Calinou, line 109: Why not use `'<meta name="description"' in context["metatags"]` to check for presence within the string?

dylannorthrup: Mostly because I'm more familiar with regex (from using `perl`, `ruby` and other languages) and don't know the various string comparison functions in `python`.  If this is a blocker, I'm happy to update the code to whatever is suggested.

mhilbrunner: @dylannorthrup If you are around to make that change that'd be welcome, as thats a bit more pythonic. Besides that, I'll do a final test later today and then merge if nothing else comes up. :) Sorry this has been sitting a bit.

mhilbrunner, line 109:
```suggestion
    if not '<meta name="description"' in context["metatags"]:
```

mhilbrunner: I've gone ahead and made the change.

dylannorthrup: My apologies for the delayed response. I had some personal issues that intervened. Thank you for making the change!

dylannorthrup:
@mhilbrunner Please let me know if there's anything else needed for this PR. I should be much more responsive to requests going forward.

mhilbrunner:
@dylannorthrup No, unless we notice something down the line - because I've just gone ahead and squashed and merged the PR :) Congrats, and thanks for keeping up with all the requests.

FrederickKDP:
Hi, I have been trying to fix 3.x epub version #9485 . I tried to build locally and same problem. Could someone tell me if the same issues happened here?
Maybe I can try read how this developed and port the solutions back to 3.x

Thank you

Issue #8208, opened 2023-10-09
Error in C# Switch Case Statement in OS Finding Example


imaperson:
Godot 4.1

The C# example code is missing a `:` after `case "OpenBSD"` in the documentation.

https://docs.godotengine.org/en/stable/classes/class_os.html#class-os-method-get-name

akien-mga:
Fixed by https://github.com/godotengine/godot/pull/83066.

PR #8209, opened 2023-10-09
Adding missing semicolon.

imaperson:
PR to fix Issue #8208


AThousandShips:

Closing because this is not the way to update this (see at the top of the file)

PR #8211, opened 2023-10-09
inspector_plugins.rst: add new "wide" parameter to _parse_property

MannyC:
Godot 4.0 added a new parameter to EditorInspectorPlugin._parse_property.  This was updated in the C# version of the tutorial, but not the gdscript version

### DIFF
--- a/tutorials/plugins/editor/inspector_plugins.rst
+++ b/tutorials/plugins/editor/inspector_plugins.rst
@@ -123,12 +123,12 @@ specifically add :ref:`class_EditorProperty`-based controls.
         return true
 
 
-    func _parse_property(object, type, path, hint, hint_text, usage):
+    func _parse_property(object, type, name, hint_type, hint_string, usage_flags, wide):
         # We handle properties of type integer.
         if type == TYPE_INT:
             # Create an instance of the custom property editor and register
             # it to a specific property path.
-            add_property_editor(path, RandomIntEditor.new())
+            add_property_editor(name, RandomIntEditor.new())
             # Inform the editor to remove the default property editor for
             # this property type.
             return true


tutorials/plugins/editor/inspector_plugins.rst
```diff
@@ -123,7 +123,7 @@ specifically add :ref:`class_EditorProperty`-based controls.
         return true
 
 
-    func _parse_property(object, type, path, hint, hint_text, usage):
+    func _parse_property(object, type, path, hint, hint_text, usage, wide):
```
AThousandShips, line 126:
```suggestion
    func _parse_property(object, type, name, hint_type, hint_string, usage_flags, wide):
```
To match the default, and matches the C#

MannyC: Looking at the [underlying engine code](https://github.com/godotengine/godot/blob/4b7cc99bc195387c784efced7fdd360501d9a5cf/editor/editor_inspector.cpp#L1098), the engine names are being remapped from the the internal names with the original names and type being

```c++
(Object *p_object, const Variant::Type p_type, const String &p_path, const PropertyHint p_hint, const String &p_hint_text, const BitField<PropertyUsageFlags> p_usage, const bool p_wide)
...
GDVIRTUAL_CALL(_parse_property, p_object, p_type, p_path, p_hint, p_hint_text, p_usage, p_wide, ret);
```
and the binding
```c++
GDVIRTUAL_BIND(_parse_property, "object", "type", "name", "hint_type", "hint_string", "usage_flags", "wide")
```

Is there a standard for names used in gdscript?  Looking around I don't see other instances of type information being added to the name (`hint` -> `hint_type`, `usage` -> `usage_flags`) or rephrasing (`hint_text` -> `hint_string`).  

Also `path` was renamed to `name`.   It does appear to be the local property name rather than a path though.

AThousandShips: The binding should be used here, as it is what you get if you add this method in script, IMO

AThousandShips:
You will need to update the use of `path` below

tutorials/plugins/editor/inspector_plugins.rst
MannyC: I was more questioning whether there was also an issue with the binding or not. 

I just need to change the call in the method to match now.

MannyC:
> You will need to update the use of `path` below

Yeah, I had seen that but I couldn't see how to include it in a batch commit using the on-site UI.  

Do you want these squashed of does it not matter?

MannyC:
Hmm, give me another second on this one

MannyC:
Okay, should be good to go

mhilbrunner (approved):

mhilbrunner:

mhilbrunner:
Cherry-picked to 4.1.

PR #8227, opened 2023-10-12
Update the state of C#/.NET support as of 4.2 beta 1

raulsntos:
- Follow-up to https://github.com/godotengine/godot-docs/pull/8142
- Follow-up to https://github.com/godotengine/godot/pull/82729


tutorials/scripting/c_sharp/index.rst
```diff
@@ -42,13 +42,23 @@ C# platform support
...
 - Only some Android architectures are supported: ``arm64`` and ``x64``.
 
-Currently, projects written in C# cannot be exported to iOS and web platforms. To use C#
-on those platforms, consider Godot 3 instead.
+iOS support is currently experimental and has a few limitations.
+
+- Projects targeting iOS need to target .NET 8.0 or higher. This requires
+  modifying the `.csproj` file generated by Godot for existing projects.
+
+- Only some iOS architectures are supported: ``arm64``.
```
raulsntos, line 59: I'm not sure this is the only architecture supported, also didn't mention anything about the simulator but I assumed we support the same architecture.

shana: There aren't any architecture limitations for iOS, on the C# export side - it's arm64 for the device builds, and both x64 and arm64 for simulator builds.

The official template package doesn't include the simulator arm64 template, though, so users won't be able to create final packages for that specific combo out of the box. They can build the templates themselves and that would work.

tutorials/scripting/c_sharp/index.rst
```diff
@@ -42,13 +42,23 @@ C# platform support
...
+  modifying the `.csproj` file generated by Godot for existing projects.
 
 - Only some Android architectures are supported: ``arm64`` and ``x64``.
 
-Currently, projects written in C# cannot be exported to iOS and web platforms. To use C#
-on those platforms, consider Godot 3 instead.
+iOS support is currently experimental and has a few limitations.
+
+- Projects targeting iOS need to target .NET 8.0 or higher. This requires
+  modifying the `.csproj` file generated by Godot for existing projects.
```
shana, line 57: Should we add the line that we inject into new projects here, as an example?

```
<TargetFramework Condition=" '$(GodotTargetPlatform)' == 'ios' ">net8.0</TargetFramework>
```

tutorials/scripting/c_sharp/index.rst
raulsntos: What about `arm32` and `x86_64`? Does iOS not support those architectures? For the simulator we can mention that we don't provide official templates for `arm64` (so it's only `x64` right?).

shana: No, iOS doesn't support anything other than arm64, not since iPhone 5. The simulator supports arm64 and x64 only because Apple still supports Intel macs, they haven't deprecated everything that's not silicon (yet).

shana: Yup, `x64` for the simulator!

raulsntos: OK, I removed the part about iOS architectures and only added a note about the iOS simulator templates limitations, let me know if it looks good now.

mhilbrunner (approved):

mhilbrunner:
@raulsntos I assume this is good to go? Let me know and I'll merge it :)

raulsntos:
I made the changes that @shana suggested, so I think the limitations are correct now.

mhilbrunner:
Great, thanks!

PR #8231, opened 2023-10-12
Remove reference to a script that was designed for godot v3 and crashed in v4

TimCabbage:
Instead instructed the user that the desired scaling option is available since version 4.2

### DIFF
--- a/tutorials/rendering/multiple_resolutions.rst
+++ b/tutorials/rendering/multiple_resolutions.rst
@@ -324,10 +324,8 @@ Desktop game
     viewport, you should use the ``canvas_items`` stretch mode instead of the ``viewport``
     stretch mode.
 
-    Godot currently doesn't have a way to enforce integer scaling when using the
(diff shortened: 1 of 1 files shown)


skyace65 (changes requested):
Good change but a few things should be adjusted. I do not think saying that a subviewport was used for this in older Godot versions is necessary. Anyone looking up info for an older Godot version should be using the older version of the docs.

Additionally I don't think this info should be in a note since it's as necessary as what's in the bullet points above the note. I'd make it a bullet point under the one for "set the stretch aspect to keep..." Something like "Set the scale mode to integer"

When editing a doc PR please keep everything to one commit, if you don't know how to do that we have a guide [here](https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html#modifying-a-pull-request).

mhilbrunner:
We just merged a PR that supersedes this one: https://github.com/godotengine/godot-docs/pull/8298, so I'm going to close this one. Thanks for pointing this out and taking the time to contribute :)

PR #8236, opened 2023-10-13
bugfix: call groups on wrong object

CrisHigham:
It seems that the call_group is on the wrong element. Very new to Godot (hence doing the tutorial)
Perhaps fix is not best practice, or perhaps I even missed something.


CrisHigham:
I am not convinced that my fix is actually working, it just stopped throwing errors. Will convert to issue.

PR #8251, opened 2023-10-15
Navlink Description Fix

AdamJStankiewicz:
The statement that navigation links do not allow automatic traversal is false, I have updated that part of the description.


AdamJStankiewicz:
idk what all checks have failed means, however I can provide proof that my new description is correct

AdamJStankiewicz:
haha all checks passed lets go

skyace65:
So what do you mean by automated movement? Does an agent get teleported from one link to another?

AdamJStankiewicz:
hey skyace, the agent will automatically jump / walk between 2 linked navmeshes

AdamJStankiewicz:
it seems to be some sort of interpolation, but it is all automatic

skyace65:
CC @smix8

smix8:
A NavigationLink does no provide any movement on its own.
Not even the NavigationAgent does. The user custom script does the entire movement.

The note / warning in the documentation exists because just too many users expected the NavigationLink or NavigationAgent to handle links automatically, e.g. handle platforms, gaps or small jumps, and complained when their agent did get stuck or tipped down a cliff. Since the warning(s) those wrong expections stopped.

All the NavigationLinks does is add a virtual navigation map connection between two navigation mesh polygons. The links has not even a real, query-able navigation mesh polygon as its own connection. It is a purely virtual two-point connection.

AdamJStankiewicz:
Yes, how ever saying there is code required is misleading

AdamJStankiewicz:
There is no additional code required for a navigation agent to traverse a navigation link.

AdamJStankiewicz:
And if some people have issues, the note needs to not be ABSOLUTE, it says outright that it never works, however that is simply not the case

AdamJStankiewicz:
There needs to be some form of fix to the note and description, as automatic traversal is possible, however if the begin and end points are place incorrectly, then it may not work. This is exactly what I said in me rewrite

AdamJStankiewicz:
Your suggested fix to tell people that it will never work is misleading and can do more harm then good

YuriSizov:
@AdamJStankiewicz Please use the edit button on your message instead of multiposting. Every comment you add notifies dozens of people subscribed to this repository.

AdamJStankiewicz:
sorry kind of new to github, will do

skyace65:
@smix8 I just created a test project using the basic navigation script for agents from the docs, and agents will act as if the link is a physical link and float across it, attached the project below. I'm going to assume this is because something isn't setup properly.

[DocTest.zip](https://github.com/godotengine/godot-docs/files/12968053/DocTest.zip)

AdamJStankiewicz:
hey any update on this?

timothyqiu:
The movement is produced by your script, not the navigation link.

A navigation link only connects its entry and exit points logically. So the exit point is right after the entry point in the navigation path. When the object reaches the entry point of a navigation link, calling `get_next_path_position()` returns the corresponding exit point (as they are considered connected).

If your movement code does not have any concept of "link", then it'll then make the object move to the exit point in the same fashion as moving between regular waypoints. But again, this is not navigation link moving your object, it's a result of your script deciding it should simply move from entry point to exit point regularly.

The correct way is to connect to the `link_reached` signal. The object should change to a separate movement logic when reaching the entry point, and then change back to regular movement after reaching the exit point.

skyace65:
Closing per timothyqiu's comment.

PR #8267, opened 2023-10-18
Fixing small wrong info

rhogeranacleto:
The docs says to includes the `get_tree().call_group("mobs", "queue_free")` to the `new_game()` when is actually to the `game_over()`

### DIFF
--- a/getting_started/first_2d_game/06.heads_up_display.rst
+++ b/getting_started/first_2d_game/06.heads_up_display.rst
@@ -411,7 +411,7 @@ click "Groups" and you can type a new group name and click "Add".
 .. image:: img/group_tab.webp
 
 Now all mobs will be in the "mobs" group. We can then add the following line to
-the ``new_game()`` function in ``Main``:
+the ``game_over()`` function in ``Main``:
 
 .. tabs::
  .. code-tab:: gdscript GDScript


getting_started/first_2d_game/06.heads_up_display.rst
```diff
@@ -411,7 +411,7 @@ click "Groups" and you can type a new group name and click "Add".
 .. image:: img/group_tab.webp
 
 Now all mobs will be in the "mobs" group. We can then add the following line to
-the ``new_game()`` function in ``Main``:
+the ``gamer_over()`` function in ``Main``:
```
AThousandShips, line 414:
```suggestion
the ``game_over()`` function in ``Main``:
```

AThousandShips (commented):
Please make your changes with `git commit --amend` to keep it a single commit

skyace65 (changes requested):
The 2D dodge the creeps project in the demo repository does have that line of code in ``new_game()`` and not ``game_over()`` and I don't see anything else in the docs suggesting that line of code should be in the game over function. What are you looking at that implies this is the wrong place to put it?

AThousandShips:
Could be confusion over that `new_game` comes after `game_over`, when I skimmed the file I thought the change was correct for that reason I believe

rhogeranacleto:
It could be my beginner's feeling about how this works.
When I put that line on the new game the Mob's stays on screen until I click on new_game button, which sometimes the 3-second timeout is not enough to destroy all remaining enemies, and it feels like the game isn't actually finished.
But I understand your point. You can close this PR then.

PR #8269, opened 2023-10-18
Update GDScript reference, 'in'-keyword explanation

bziemons:
Very small change.
The explanation for the keyword 'in' states it checks, whether an element is in a list (among others). This is the only occurrence of the word 'list' in this context and I assume it actually means array. To avoid future confusion, it makes sense to replace the word list with array in this explanation.


mhilbrunner (approved):

mhilbrunner:
That indeed makes sense. Thanks and congrats on your first merged contribution to Godot's docs!

PR #8270, opened 2023-10-18
Importing Audio Samples: loop points are in samples, not seconds

mariomadproductions:
Importing Audio Samples incorrectly said that loop points are in seconds. I've corrected that to say samples (which is the case as of 4.1.2)

### DIFF
--- a/tutorials/assets_pipeline/importing_audio_samples.rst
+++ b/tutorials/assets_pipeline/importing_audio_samples.rst
@@ -150,8 +150,8 @@ metadata, but you can choose to apply a specific loop mode:
 
 When choosing one of the **Forward**, **Ping-Pong** or **Backward** loop modes,
 loop points can also be defined to make only a specific part of the sound loop.
-**Loop Begin** is set in seconds after the beginning of the audio file. **Loop
-End** is also set in seconds after the beginning of the audio file, but will use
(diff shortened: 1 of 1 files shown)


ellenhp (approved):
This is right for WAV files, but for oggs and mp3s the loop points are in seconds, which might be where this came from. Still, oggs and mp3s don't have a loop end import option or selectable loop modes so this is more accurate in context than what was there before.

mariomadproductions:
> This is right for WAV files, but for oggs and mp3s the loop points are in seconds, which might be where this came from. Still, oggs and mp3s don't have a loop end import option or selectable loop modes so this is more accurate in context than what was there before.

Ah I see. Thanks.

mhilbrunner (approved):

mhilbrunner:
Merged. Thanks and congrats on your first merged contribution to Godot's docs, and thanks Ellen for the review :)

mariomadproductions:
> Merged. Thanks and congrats on your first merged contribution to Godot's docs, and thanks Ellen for the review :)
> 
> Note for future PRs: its best practice to create a branch for each PR, instead of submitting them from your main branch.

Thanks, and I'll keep that in mind.

mhilbrunner:
Cherry-picked to 4.1.

Issue #8256, opened 2023-10-16
"Pull Request Workflow" should point out that force push is required after rebase


BrianMacIntosh:
**Your Godot version:**
v4.2.beta.custom_build [e0cb60a8a]

**Issue description:**
After locally rebasing a branch for a pull request, `git push --force` is required. Attempting to push normally unhelpfully tells the user to `git pull` again. I tripped on this in a PR (https://github.com/godotengine/godot/pull/82707). In general, the Pull Request Workflow page is very thorough and helpful for git newcomers, so I think the **Updating your branch** section should say something about this.

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html#updating-your-branch

AThousandShips:
It does, [here](https://docs.godotengine.org/en/latest/contributing/workflow/pr_workflow.html#modifying-a-pull-request)

BrianMacIntosh:
I had read the whole thing (days ago, of course), but the way I'm using it (and the way I think the structure suggests) is to return to it when I need to do a particular task, find the section for that task, and walk through it. Because I didn't need to amend or squash anything here, I didn't think to check the amending/squashing sections.

From reading those sections previously, I didn't connect a simple `git pull --rebase` with the need to force-push - I thought it was only necessary because those two tasks are doing 'weirder' things with modifying and rewriting existing commits.

BrianMacIntosh:
> I'm also not sure this is what caused your issue, but an incorrect rebase
I think you're right, but the thing that led to that was that after I _correctly_ rebased, I tried to push and was told by git I still needed to pull. So I attempted to do that, and that's where the trouble started.

AThousandShips:
Note that the section you're looking at isn't for updating a created feature branch, but for updating your own master branch before creating a feature, so this isn't relevant to that issue

AThousandShips:
Did you use `git pull --rebase` to update your branch or `git rebase -i`?

BrianMacIntosh:
I see what you mean about that section being intended for pulling to `master`. The directions there do work for feature branches (with the addition of `git push --force`), but if you would prefer people rebase interactively, a different possible fix could be to slightly reword the **The interactive rebase** section to point out that it's good for that situation. Currently it's framed very specifically as "this is what you do if you need to squash commits".

I used `git pull --rebase` in this case.

AThousandShips:
[...]

PR #8274, opened 2023-10-19
Noting that the rebase in 'Updating your branch' requires a force push.

BrianMacIntosh:
The reasons for this are described in detail in the 'The interactive rebase' section, but other sections such as 'Modifying a pull request' also include a brief note to this effect. This addresses #8256. While the 'interactive rebase' section is more relevant to rebasing a PR, 'Updating your branch' does contemplate that you may have local commits to master, and a force push is necessary in that case.

### DIFF
--- a/contributing/workflow/pr_workflow.rst
+++ b/contributing/workflow/pr_workflow.rst
@@ -248,6 +248,10 @@ section <doc_pr_workflow_rebase>` for instructions.
          to find the commit ID of the previous state that you would like to restore, and
          use it as argument of ``git reset --hard`` to go back to that state.
 
+If you had local changes before rebasing, the rebase rewrites your local git history.
+In that case, you will need to force push using ``git push --force origin master``
+instead of pushing normally.
+
 Making changes
 --------------
 


contributing/workflow/pr_workflow.rst
```diff
@@ -248,6 +248,10 @@ section <doc_pr_workflow_rebase>` for instructions.
          to find the commit ID of the previous state that you would like to restore, and
          use it as argument of ``git reset --hard`` to go back to that state.
 
+If you had local changes before rebasing, the rebase rewrites your local git history.
+In that case, you will need to force push using ``git push --force origin master``
```
AThousandShips, line 252: This is wrong, unless you mean that the user should update their master branch with changes, we do not recommend making PR work on that branch, so should be worded differently

BrianMacIntosh: That section already has language that assumes the user has their own commits to master, so it seems like there might be a larger structural issue here. I'm just going to back out of this rather than trying to work around that.

AThousandShips: That's because it's about updating your `master` branch, but it's vague, the general thing needs to be clarified that it's about preparing for a PR, and separate it out IMO

PR #8275, opened 2023-10-19
Fix syntax in csharp example inside class_object.html page

### LINKED ISSUE #8220: Object Class Reference page uses incorrect C# example
**Your Godot version:**
4.2dev6
**Issue description:**
The [page ](https://docs.godotengine.org/en/latest/classes/class_object.html#class-object) includes this line:
> You can create new instances, using `Object.new()` in GDScript, or `new Object` in C#.

As of 4.2dev6, he C# implementation of Godot's `Object` class is called `GodotObject`, not `Object`.

The example code should be replaced with something like `new GodotObject()`.

This is likely not the only place where a mistake like this exists, I have not done a thorough search.

https://docs.godotengine.org/en/latest/classes/class_object.html#class-object

k0T0z:
Fixes #8220


AThousandShips:

k0T0z:
@AThousandShips Ohh, I am pretty sure I have read this page about the generated files but I didn't look closer, my bad, thank you.

AThousandShips:
It also says so at the top of the file :)

PR #8284, opened 2023-10-19
compiling-for-windows.rst: Mention SCons option dev_build=yes for debugging Godot

MyPingO:
Reasoning:
Clarity and discoverability

Context:
When I looked up "compile Godot source code", this is the page I read to lead me through it. It failed to mention this crucial command line argument. Spent a few days trying to figure out why I couldn't place breakpoints when debugging Godot source code.

### DIFF
--- a/contributing/development/compiling/compiling_for_windows.rst
+++ b/contributing/development/compiling/compiling_for_windows.rst
@@ -148,7 +148,11 @@ your CPU architecture, but this can be overridden using ``arch=x86_64`` or
 This executable file contains the whole engine and runs without any
 dependencies. Running it will bring up the Project Manager.
 
-.. note:: If you are compiling Godot for production use, then you can
+.. note:: If you want the ability to debug Godot, include the SCons option
+          ``dev_build=yes``. Refer to :ref:`doc_introduction_to_the_buildsystem`
+          for more details.
+
+          If you are compiling Godot for production use, then you can
           make the final executable smaller and faster by adding the
           SCons option ``target=template_release``.
 


MyPingO:
3rd time is the charm?

anvilfolk (approved):
I am currently a big proponent of the strategy of adding a reference to `dev_build=true` anytime anyone anywhere whispers the word `scons` until we stop getting 10 asks about this per week on RocketChat! 😅 

In seriousness, I do think more mentions of `dev_build` is good. The primary reason folks compile Godot for themselves is to consider contributing to the engine, and we do get far too many folks trying to work on their first issues with this problem.

It's ultimately the documentation/website folks' choice, and I can't tell whether the `rst` works, but am generally for more references like this to decrease the chance it's missed :)

Calinou:
If we go for the route of repeating what the introduction page says, this should really be done on all platform pages rather than only being mentioned on a single platform page.

MyPingO:
I think a better approach would be to "copy-paste" in this case.
On the page I'm editing there is a link to the page that talks about other options for SCons, and dev_build=yes is one of the first things you see

> Talking about this line: For a general overview of SCons usage for Godot, see Introduction to the build system
> (https://docs.godotengine.org/en/latest/contributing/development/compiling/introduction_to_the_buildsystem.html#doc-introduction-to-the-buildsystem).

But the problem is that this line is out of sight from most readers because it's placed in a part of the page where I wasn't looking (it's on a note that's under two unrelated notes)

My attention was drawn to the Compiling section on this page and that's where I went to follow instructions on what to do. This is what 90% of people do too because it's the section that talks about what we want to accomplish: Compiling Godot.

Picking the right spot to place a reference link is crucial to it being seen. Just having a link at the top of each related page will make it the reader's fault if they don't catch it, but I think the goal is to make it as visible as possible to avoid confusion as much as possible.

anvilfolk:
Yeah, the software person in me revolts a little bit at copying this little info nugget everywhere...

But! I believe a big part of the issue with folks not noticing they need `dev_build` is that they likely already have several tabs open (at least one of their OS and one for their IDE, probably the PR workflow one and other generic pages), and it becomes extremely easy to accomplish a step from one of them and close it, or switch to another.

Unless this info is front and center almost everywhere (and it _is_ actually replicated in a lot of places already!), it's probably going to keep being missed here and there. I would let the software human in me be more opposed to this if it wasn't such a common stumbling block  :)

AThousandShips:
People likely won't go through levels of links to find necessary information, there's already quite a bit of information overload risk in these pages

PR #8285, opened 2023-10-19
Document glow in 2D in Environment and post-processing

Calinou:
This also documents using glow as a blurring solution (e.g. for menu background blur), which is a common question on support platforms.

The reference to an outdated quality setting in the depth of field properties was also updated.

Don't cherry-pick, as HDR in 2D is only in 4.2 and later.

- See https://github.com/godotengine/godot/pull/80215.


mhilbrunner (approved):

mhilbrunner:
Thank you! 🎉

PR #8289, opened 2023-10-20
Remove duplicate reminders that are very close to each other

hidalgob:
Was reading through the tutorial and saw this information popping up at least 3 times. There's another reminder a couple lines below saying the same thing. So I'm proposing to remove this one.

### DIFF
--- a/getting_started/first_2d_game/06.heads_up_display.rst
+++ b/getting_started/first_2d_game/06.heads_up_display.rst
@@ -337,9 +337,6 @@ window and selecting the ``new_game()`` method or type "new_game" below "Receive
 in the window. Verify that the green connection icon now appears next to
 ``func new_game()`` in the script.
 
-Remember to remove the call to ``new_game()`` from
-``_ready()``.
-
 In ``new_game()``, update the score display and show the "Get Ready" message:
 
 .. tabs::


AThousandShips:

AThousandShips:
Updated your branch

skyace65:
Important information. We added the other note on this page recently knowing it was a duplicate, and the author of that PR missed this note when they read it. I'm in favor of removing it because this seems like excessive repetition to me, and if people don't fully read the page that's on them. Though I do remember we made an exception for something else because people kept reporting an issue.

@mhilbrunner @cbscribe you guys approved the PR that added the second note. Any change in opinion or do you still want two notes?

Piralein:
#8213 is doing the same change.

skyace65:
@Piralein I'd be in favor of removing both notes and adding a warning as you suggested in #8213.

mhilbrunner:
> I'd be in favor of removing both notes and adding a warning as you suggested in #8213.

I'm fine with this too. :)

mhilbrunner:
Thanks for contributing! I've merged https://github.com/godotengine/godot-docs/pull/8437, which aims to fix this and supersedes this PR. Thanks for bringing this to our attention and taking the time to contribute :)

Issue #6790, opened 2023-02-19
(HELP WANTED) C# basics page lacks crucial information about async code usage


tvardero:
C# basics page does not reflect that scene's methods `_Ready` and `_Process` have single threaded synchronization context and should wait for async methods completition only if those have `ConfigureAwait(false)`.

I don't feel confident enough to make documentation changes, nor confident enought in knowing async / threading model of C# in Godot engine. 
True root of the problem could be not a single threaded Scene, but something else. 
Sorry for that.

**Your Godot version:**
Godot 3.4.4
Godot 4 Beta 8 and probably higher

**Issue description:**
See https://github.com/godotengine/godot/issues/63725

**URL to the documentation page:**
https://github.com/godotengine/godot-docs/blob/master/tutorials/scripting/c_sharp/c_sharp_basics.rst

PR #8296, opened 2023-10-21
Add async/await guidelines to C# basics

carrollj:
- Fixes https://github.com/godotengine/godot-docs/issues/6790

The documentation differs a bit from what the issue mentioned, but I think it's for good reason. I got the impression that the thread that spawned the issue (https://github.com/godotengine/godot/issues/63725) spurred a misunderstanding: That _Process executing before awaits inside of _Ready completed was a problem. In reality, it's expected behavior and matches the behavior of GDScript.

This documentation provides a way to asynchronously execute code from top level functions such as _Ready without needing to mark those functions as async. I believe that is important, since many people familiar with async/await in C# have been warned against it very heavily.

I also included code where I marked the top level functions as async. I tested that code pretty heavily for all the common pitfalls people bring up when saying "avoid async void", and it does not appear to suffer from any of them. I could have missed something, but overall it seems to perform even marginally better than the non async void approach. 

Finally, I included an approach using Task.Run(), since that seems to be the most commonly recommended approach to executing async methods in Godot. Personally, I think it robs you of all the value of having async methods with only additional risks to gain (switching threads). But since it's out there I thought it important to document the approach while describing the pitfalls and workarounds for those pitfalls, especially since I rarely see anyone mention them when recommending using Task.Run.

Another thing I excluded was the recommendation to use `ConfigureAwait(false)`. Using it should be avoided in application level code since it bypasses the SynchronizationContext provided by the application (``GodotSynchronizationContext``) for the called method and every async method in that call chain. The ``GodotSynchronizationContext`` does a great job of keeping continuations on the same thread as its caller, and bypassing that has the potential to create problems where a continuation is unexpectedly run on a different thread than the caller, causing thread safety issues that are almost invisible to the average user. Besides, there aren't any issues in the provided examples that using `ConfigureAwait(false)` would solve, except some minor performance gains in some scenarios, so I figured it was best not to mention it.

One thing to note is that my ``SampleAsyncNode`` for avoiding race conditions does not work in the latest release of 4.2 beta 3 due to a bug introduced in 4.2 dev 4 (https://github.com/godotengine/godot/issues/82279). 

Sorry for the wall of text

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_basics.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_basics.rst
@@ -388,3 +388,234 @@ The following tools may be used for performance and memory profiling of your man
 - Visual Studio.
 
 Profiling managed and unmanaged code at once is possible with both JetBrains tools and Visual Studio, but limited to Windows.
+
+Using ``async``/``await``
+-------------------------
+
+You might face a scenario where you must ``await`` a method call.
+You will notice that when you use ``await``, you are required to mark the method you use it in as ``async``, 
+and change the return type to an awaitable type, such as ``Task`` or ``Task<T>``. 
+Consequently, you must call your now ``async`` method using ``await``, 
+which propagates the problem all the way up the call chain.
+This is why many people compare ``async``/``await`` to a "zombie virus", 
+because it tends to spread once introduced.
+
+In Godot, the conclusion to this spread is the entry point methods of a node, such as ``_Ready()`` or ``_Process()``.
+You will notice that the return types of these methods are ``void`` rather than ``Task``.
+It is considered conventional wisdom in C# to avoid ``async void`` at all times, with the exception of event handlers.
+The problem is that it is impossible to change the signatures of these methods since they are defined by the classes they inherit.
+
+There are a couple options to address this problem, but each option comes with its own caveats and considerations. 
+To compare these options, we will work with the following script:
+
+.. code-block:: csharp
+
+    using Godot;
+    using System;
+    using System.Threading.Tasks;
+
+    public partial class AsyncTestNode : Node
+    {
+        private int _taskCount = 0;
+        private DateTime start;
+        public override void _Ready()
+        {
+            start = DateTime.Now;
+        }
+    
+        public override void _Process(double delta)
+        {
+
+        }
+
+        // Prints the amount of time since _Ready started, the current thread, and the name of the calling method
+        // It prints this once when DoStuffAsync is first called, then once again after `duration` in seconds
+        private async Task DoStuffAsync(double duration, string methodName)
+        {
+            var taskId = ++_taskCount;
+            PrintCurrentThread($"Task {taskId} started from {methodName}");
+            await Task.Delay(TimeSpan.FromSeconds(duration));
+            PrintCurrentThread($"Task {taskId} completed");
+        }
+    
+        private void PrintCurrentThread(string info)
+        {
+            var timeStamp = DateTime.Now - start;
+            GD.PrintT(timeStamp.ToString(@"mm\:ss\.ff"), $"Thread: {System.Environment.CurrentManagedThreadId}", info);
+        }
+    }
+
+The first option is to start the task through the Task factory.
+
+.. code-block:: csharp
+
+    // This function can be put in a global static class for convenience
+    public static Task StartTaskFromFactory(Func<Task> newTask)
+    {
+        return Task.Factory.StartNew(newTask,
+                                    CancellationToken.None, 
+                                    TaskCreationOptions.None, 
+                                    TaskScheduler.FromCurrentSynchronizationContext());
+    }
+
+    public override void _Ready()
+    {
+        start = DateTime.Now;
+
+        StartTaskFromFactory(async () => await DoStuffAsync(.5, nameof(_Ready)));
+    }
+
+    public override void _Process(double delta)
+    {
+        if (_taskCount < 3)
+            StartTaskFromFactory(async () => await DoStuffAsync(.5, nameof(_Process)));
+    }
+
+The second option is to mark the entry point method as async anyway.
+
+.. code-block:: csharp
+    
+    public override async void _Ready()
+    {
+        start = DateTime.Now;
+        await DoStuffAsync(.5, nameof(_Ready));
+    }
+
+
+    public override async void _Process(double delta)
+    {
+        if (_taskCount < 3)
+            await DoStuffAsync(.5, nameof(_Process));
+    }
+
+Both the manual task starting method and the ``async void`` method 
+behave identically to an equivalent script written in GDScript 
+that uses its version of the ``await`` keyword; 
+the method pauses once it reaches the ``await``-ed method call.
+The game loop will run until the task completes, at which point execution will continue on the main thread.
+
+Let's look at the output from the above code:
+
+.. code-block::
+
+    00:00.00	Thread: 1	Task 1 started from _Ready
+    00:00.02	Thread: 1	Task 2 started from _Process
+    00:00.03	Thread: 1	Task 3 started from _Process
+    00:00.50	Thread: 1	Task 1 completed
+    00:00.53	Thread: 1	Task 2 completed
+    00:00.53	Thread: 1	Task 3 completed
+
+The first observation from the output is that the game loop continues 
+without waiting for the completion of the ``_Ready()`` method. 
+This continuation can introduce issues, especially if methods like ``_Process()`` 
+rely on variables or objects that get initialized only after the ``await`` call in ``_Ready()``.
+Such asynchronous timing problems are termed `Race Condition <https://en.wikipedia.org/wiki/Race_condition#In_software/>`_, 
+which is one of the main hazards when working with asynchronous code.
+To avoid errors due to race conditions, be sure to check that values are initialized before you use them, 
+and use ``IsInstanceValid()`` after you ``await`` a function.
+
+Here is a pattern you can adopt to avoid race conditions:
+
+.. code-block:: csharp
+
+    public partial class SampleAsyncNode : Node
+    {
+        [Signal] public delegate void InitializedEventHandler();
+        [Export] public int EntityID { get; set; } = 1;
+    
+        readonly SomeCustomRepository _db = new();
+        private int _health;
+        private bool _init;
+    
+        // We will check IsInvalid after we await async methods
+        // Otherwise we risk the continuation running in a disposed context
+        private bool IsInvalid => !IsInstanceValid(this) || this.IsQueuedForDeletion();
+    
+        public override async void _Ready()
+        {
+            var entity = await _db.FindAsync(EntityID);
+    
+            // Even though we are still in _Ready(), we need to check IsInvalid
+            // It's possible that this node was freed by its parent, or some other source while awaiting
+            if (IsInvalid)
+                return;
+    
+            _health = entity.Health;
+            _init = true;
+            EmitSignal(SignalName.Initialized);
+        }
+    
+        public async Task DealDamage(int damage)
+        {
+            // DealDamage depends on _health being inititalized
+            // Awaiting Initialized will cause all calls to DealDamage to queue up
+            // Once Initialized is emitted, all queued DealDamage calls will continue at once
+            await ToSignal(this, SignalName.Initialized);
+    
+            // If you don't want to queue calls while waiting for initialization, just return if not initialized
+            // if (!_init)
+            //     return;
+    
+            if (IsInvalid)
+                return;
+    
+            _health -= damage;
+    
+            // If the number of queued calls to DealDamage is greater than the initial value of _health...
+            // This line will free the node before all calls to DealDamage are continued
+            // That is why it is important to check IsInvalid after awaiting
+            if (_health < 0)
+                QueueFree();
+        }
+    
+        public override void _ExitTree()
+        {
+            // If this unit was freed before initialization completed...
+            // Emit the Initialized signal so that everything awaiting it will be released
+            if (!_init)
+                EmitSignal(SignalName.Initialized);
+    
+            _db.Dispose();
+        }
+    }
+
+The third option is to execute the ``async`` method synchronously. 
+This is most commonly done when you need to use an asynchronous 
+method from a third party library that has no synchronous equivalent, 
+and it is not feasible to convert everything upstream to ``async``.
+
+.. code-block:: csharp
+
+    public override void _Ready()
+    {
+        start = DateTime.Now;
+    
+        Task.Run(async () => await DoStuffAsync(.5, nameof(_Ready))).GetAwaiter().GetResult();
+    }
+    
+    public override void _Process(double delta)
+    {
+        if (_taskCount < 3)
+            Task.Run(async () => await DoStuffAsync(.5, nameof(_Process))).GetAwaiter().GetResult();
+    }
+
+Let's look at the output from the above code:
+
+.. code-block::
+
+    00:00.00	Thread: 4	Task 1 started from _Ready
+    00:00.50	Thread: 4	Task 1 completed
+    00:00.52	Thread: 4	Task 2 started from _Process
+    00:01.02	Thread: 4	Task 2 completed
+    00:01.03	Thread: 4	Task 3 started from _Process
+    00:01.53	Thread: 4	Task 3 completed
+
+The output from running the tasks synchronously shows that 
+the tasks executed in the expected order for synchronous operations. 
+The output also shows that the code was executed on Thread 4, 
+rather than Thread 1 like in the first two options.
+This is important to keep in mind, because any code that is not 
+executed on the main thread (Thread 1) cannot interact with the scene tree, as it is not thread safe.
+You should use ``CallDeferred``/``SetDeferred``, ``CallThreadSafe``/``SetThreadSafe``, 
+or ``CallDeferredThreadGroup``/``SetDeferredThreadGroup`` to interact with thread 
+safe objects or APIs from threads other than the main thread.


tutorials/scripting/c_sharp/c_sharp_basics.rst
```diff
@@ -388,3 +388,217 @@ The following tools may be used for performance and memory profiling of your man
 - Visual Studio.
 
 Profiling managed and unmanaged code at once is possible with both JetBrains tools and Visual Studio, but limited to Windows.
+
+Using ``async``/``await``
+-----------------
```
AThousandShips, line 393:
```suggestion
-------------------------
```

tutorials/scripting/c_sharp/c_sharp_basics.rst
```diff
@@ -388,3 +388,217 @@ The following tools may be used for performance and memory profiling of your man
...
+
+You might face a scenario where you must ``await`` a method call.
+You will notice that when you use ``await``, you are required to mark the method you use it in as ``async``, and change the return type to ``Task`` or ``Task<T>``. 
+Consequently, you must call your now ``async`` method using ``await``, which propogates the problem all the way up the call chain.
```
AThousandShips, line 397:
```suggestion
Consequently, you must call your now ``async`` method using ``await``, which propagates the problem all the way up the call chain.
```

tutorials/scripting/c_sharp/c_sharp_basics.rst
```diff
@@ -388,3 +388,217 @@ The following tools may be used for performance and memory profiling of your man
...
+
+.. code-block:: csharp
+
+    using Godot;
+    using System;
+    using System.Threading.Tasks;
+
+    public partial class AsyncTestNode : Node
+    {
+        int _taskCount = 0;
```
paulloz, line 419: We always explicitly type access modifiers. Ditto for all implicitly private fields.

tutorials/scripting/c_sharp/c_sharp_basics.rst
```diff
@@ -388,3 +388,217 @@ The following tools may be used for performance and memory profiling of your man
...
+    }
+
+
+    public override async void _Process(double delta)
+    {
+        if (_taskCount < 3)
+            await DoStuffAsync(.5, nameof(_Process));
+    }
+
+Both the manual task starting method and the ``async void`` method behave identically to an equivalent script written in GDScript that uses its version of the ``await`` keyword; the method pauses once it reaches the ``await``ed method call.
```
AThousandShips, line 487:
```suggestion
Both the manual task starting method and the ``async void`` method behave identically to an equivalent script written in GDScript that uses its version of the ``await`` keyword; the method pauses once it reaches the ``await``-ed method call.
```
Can't do these like that, there needs to be a space or hyphen between

tutorials/scripting/c_sharp/c_sharp_basics.rst
```diff
@@ -388,3 +388,217 @@ The following tools may be used for performance and memory profiling of your man
...
+    00:00.02	Thread: 1	Task 2 started from _Process
+    00:00.03	Thread: 1	Task 3 started from _Process
+    00:00.50	Thread: 1	Task 1 completed
+    00:00.53	Thread: 1	Task 2 completed
+    00:00.53	Thread: 1	Task 3 completed
+
+The first observation from the output is that the game loop continues without waiting for the completion of the ``_Ready()`` method. 
+This continuation can introduce issues, especially if methods like ``_Process()`` rely on variables or objects that get initialized only after the ``await`` call in ``_Ready()``.
+Such asynchronous timing problems are termed `Race Condition <https://en.wikipedia.org/wiki/Race_condition#In_software/>`_, which is one of the main hazards when working with asynchronous code.
+To avoid errors due to race conditions, be sure to check that values are initialized before you use them, and use ``IsInstanceValid()`` after ``await``ing a function.
```
AThousandShips, line 504:
```suggestion
To avoid errors due to race conditions, be sure to check that values are initialized before you use them, and use ``IsInstanceValid()`` after ``await``-ing a function.
```

tutorials/scripting/c_sharp/c_sharp_basics.rst
AThousandShips, line 1: A lot of the lines here are probably far too wide, please match the maximum length of the rest of the document

tutorials/scripting/c_sharp/c_sharp_basics.rst
```diff
@@ -388,3 +388,217 @@ The following tools may be used for performance and memory profiling of your man
 - Visual Studio.
 
 Profiling managed and unmanaged code at once is possible with both JetBrains tools and Visual Studio, but limited to Windows.
+
+Using ``async``/``await``
+-----------------
+
+You might face a scenario where you must ``await`` a method call.
+You will notice that when you use ``await``, you are required to mark the method you use it in as ``async``, and change the return type to ``Task`` or ``Task<T>``.
```
paulloz: Probably or `ValueTask`/`ValueTask<T>` (or any generalized async type).

tutorials/scripting/c_sharp/c_sharp_basics.rst
```diff
@@ -388,3 +388,217 @@ The following tools may be used for performance and memory profiling of your man
...
+-----------------
+
+You might face a scenario where you must ``await`` a method call.
+You will notice that when you use ``await``, you are required to mark the method you use it in as ``async``, and change the return type to ``Task`` or ``Task<T>``. 
+Consequently, you must call your now ``async`` method using ``await``, which propogates the problem all the way up the call chain.
+This is why many people compare ``async``/``await`` to a "zombie virus", because it tends to spread once introduced.
+
+In Godot, the conclusion to this spread is the entry point methods of a node, such as ``_Ready()`` or ``_Process()``.
+You will notice that the return types of these methods are ``void`` rather than ``Task``.
+It is considered conventional wisdom in C# to avoid ``async void`` at all times, with the exception of event handlers.
```
paulloz, line 405: I'm personally not a big fan of the "because wisdom". I'd rather have that explained, or at least linked to an explanation on the Microsoft docs.

tutorials/scripting/c_sharp/c_sharp_basics.rst
carrollj: Changed this one to "after you ``await`` a function" to get rid of the -ing

paulloz (commented):
I have a hard time seeing this on the **C# basics** page. The ins and outs of `async` in C# aren't that basic IMHO. Especially when we introduce async code running on different threads at the end.

carrollj:
> I have a hard time seeing this on the **C# basics** page. The ins and outs of `async` in C# aren't that basic IMHO. Especially when we introduce async code running on different threads at the end.

I'm inclined to agree, I only put it here since it's where the issue requested it. If it belongs somewhere else, I'd be happy to movie it. I skimmed over a lot of details because I felt it was getting far to long for the C# basics page.

tutorials/scripting/c_sharp/c_sharp_basics.rst
carrollj: That is one of the things I ended up trimming due to length. I would prefer to explain the reasoning behind it, then explain why doing so in Godot does not cause any issues, but that got too long and not basic for what I felt belonged in the page.

I can add a link to an explanation of why to avoid it, but I would also like to include a paragraph explaining why it doesn't apply in Godot.

tutorials/scripting/c_sharp/c_sharp_basics.rst
carrollj: I can reword this "change the return type to any ``await``-able return type, such as ``Task`` or ``Task<T>``". Sound good?

mhilbrunner:
cc @raulsntos :) Thoughts? Where should this go?

raulsntos:
There's definetely a need for a documentation page that explains asynchronous programming in the context of Godot. I feel like it's a complicated topic though, so I think it deserves its own dedicated page.

The documentation should likely explain how Godot's `SynchronizationContext` works, and how it affects `async`/`await`. Some of this is just how C# works, unrelated to Godot, but it's complicated and not a lot of developers know about it. I wouldn't consider myself to be very knowledgeable about it either, and often find myself re-reading Stephen Toub's articles about it.

- [How Async/Await Really Works in C#](https://devblogs.microsoft.com/dotnet/how-async-await-really-works/)
- [Async/Await FAQ](https://devblogs.microsoft.com/pfxteam/asyncawait-faq/)

I haven't had a chance to review this PR in depth yet, but from a quick look it seems to be too superficial, it explains basic concepts that are already covered by [Microsoft's documentation](https://learn.microsoft.com/en-us/dotnet/csharp/asynchronous-programming/) so we could probably just link to them.

In general, I prefer to avoid duplicating Microsoft's documentation, otherwise we risk our documentation getting outdated as new concepts get introduced in C#. Instead, Godot's documentation should focus on how the C# concepts (such as `async` and `await`) apply in the context of Godot. Do these work the same way as in a normal C# application? If not, how are they different? In this case, we have a custom `SynchronizationContext` implementation. What does that mean for users and how does it affect their `async`/`await` code?

So to summarize:
- I think this should be a dedicated page. Definitely a welcomed addition, thank you for contributing.
- The documentation should probably skip some basic C# concepts that are already explained in Microsoft's documentation and just link it.
- The documentation should at least mention that Godot implements a custom `SynchronizationContext` and try to explain how that affects game developers.

carrollj:
Thanks for the feedback. I'll be happy to make the requested changes; I felt I had to omit too many things to keep the section in line with being in the C# **Basics** page. I wasn't sure about the policy towards linking to external sources. I know some sites, like Stackoverflow, prefer having the relevant concepts explained on page in case the external source changes, is moved, or goes down, so I erred on that side. I'll make sure to include links to documentation instead of explaining the fundamentals.

Stephen Toub's articles are excellent resources, is it okay to link to those as well?

It'll probably take me a few days to make the revisions due to Halloween and other obligations. In the meantime, where do you suggest this new page go? I'm guessing something like `tutorials/scripting/c_sharp/c_sharp_async_await.rst` with a link to it added in the toctree under the "Godot API for C#" section in `/c_sharp/index.rst`.

Additionally, there is a section on the await keyword in `/c_sharp/c_sharp_differences.html`. Should that be rolled into this new page? Or perhaps link to it?

tutorials/scripting/c_sharp/c_sharp_basics.rst
```diff
@@ -388,3 +388,234 @@ The following tools may be used for performance and memory profiling of your man
...
+
+    using Godot;
+    using System;
+    using System.Threading.Tasks;
+
+    public partial class AsyncTestNode : Node
+    {
+        private int _taskCount = 0;
+        private DateTime start;
+        public override void _Ready()
```
AThousandShips, line 421:
```suggestion
 
        public override void _Ready()
```

tutorials/scripting/c_sharp/c_sharp_basics.rst
```diff
@@ -388,3 +388,234 @@ The following tools may be used for performance and memory profiling of your man
...
+
+.. code-block:: csharp
+    
+    public override async void _Ready()
+    {
+        start = DateTime.Now;
+        await DoStuffAsync(.5, nameof(_Ready));
+    }
+
+
```
AThousandShips, line 484:
```suggestion
```

carrollj:
Sorry @AThousandShips, I didn't mean to trigger this to be re-reviewed, I was just squashing and rebasing this branch before making the requested changes

AThousandShips:
No worries, I realized I would wait until the restructuring

Perustaja:
I think there needs to just be a summarization with a separate page here. I also think some stuff could be rephrased/rearranged. For rephrasing, the term zombie virus seems a bit informal and out of place.

I also think that the first explanation essentially says "just use async void" but then starts off with an explanation using TaskFactory. I think start with the async void explanation and then show the rest.

It's a good explanation of how it works though, because using async code here is quite a bit different than, say, ASP.NET. If you want I'd be happy to give a draft with it all on the current page. It is a lot though, and the main basics page should probably tersely show how to make it work as expected, with a branch off explaining the nuances as you've done.

tvardero:
_open for criticism_

You should block the thread that enters the `_Ready` function by synchronously waiting for all your async tasks (via `.Result` or `.Wait()`), whether they are long-running or short-running. This is necessary if awaiting them is critical for the integrity of the parent scene.

In the case of the thread entering the `_Process` function, it should only be blocked for really fast async tasks. For short-running tasks, I recommend considering making their code synchronous where possible. For long-running tasks, avoid awaiting them and instead run them in the background. One approach could be to enqueue work (`private readonly Queue<YourWorkObject> _queue`) and regularly check it in the background for incoming work. If you seek a different method, you can search the web for "how to enqueue work in the background" or "how to run tasks in the background without awaiting them."

Any network/database call, heavy mathematical calculation, or I/O operations (such as working with files) should be considered as long-running tasks in the context of the `_Process` function and should be left asynchronous. Regarding heavy mathematical calculations, they could indeed be made synchronous. However, if their performance impact is considerable and/or they could be parallelized for better performance, it's better to leave them as `Task.Run`. (By the way, it's nice to know about methods like `Parallel.ForEach` for your mathematical operations, terrain generations, etc.)

The use of `.ConfigureAwait(false)` is justified only for code used as part of a generic library that is agnostic to the environment in which it's being used. Examples include ORMs (EntityFramework), message bus connectors (MassTransit), background workers (Hangfire), and CS algorithms—they do not care whether they run in a web application, desktop app, or Godot game; hence, they do not care about the threading model there.

Using `ConfigureAwait(false)` in code that interacts with Godot could lead to synchronization issues or access violations. For instance, in WinForms, you are not allowed to interact with your canvas from any thread except the UI thread. Trying to paint a button red or move an input field to the right from such threads results in an access violation. `ConfigureAwait(false)` instructs your program that your async task could finish on any thread other than the one that started the task. If you start your falsely-configured awaitable task from the UI thread, you might end up with it finishing its job on another thread and having no access to the canvas in a WinForm application.

(Please note that if Godot does not have such considerations about accessing UI from different threads, then what I've written on the `ConfigureAwait` subject might not be very relevant.)

tutorials/scripting/c_sharp/c_sharp_basics.rst
```diff
@@ -388,3 +388,234 @@ The following tools may be used for performance and memory profiling of your man
...
+            GD.PrintT(timeStamp.ToString(@"mm\:ss\.ff"), $"Thread: {System.Environment.CurrentManagedThreadId}", info);
+        }
+    }
+
+The first option is to start the task through the Task factory.
+
+.. code-block:: csharp
+
+    // This function can be put in a global static class for convenience
+    public static Task StartTaskFromFactory(Func<Task> newTask)
```
tvardero, line 453: In my current understanding just calling the task `newTask()` will capture current synchronization context.
If that is correct, then using TaskFactory is unnecessary.

tutorials/scripting/c_sharp/c_sharp_basics.rst
```diff
@@ -388,3 +388,234 @@ The following tools may be used for performance and memory profiling of your man
...
+                                    CancellationToken.None, 
+                                    TaskCreationOptions.None, 
+                                    TaskScheduler.FromCurrentSynchronizationContext());
+    }
+
+    public override void _Ready()
+    {
+        start = DateTime.Now;
+
+        StartTaskFromFactory(async () => await DoStuffAsync(.5, nameof(_Ready)));
```
tvardero, line 465: Making lambda async is not necessary and also results in state machine generation.
Here you can return the task itself. `() => DoStuffAsync(.5, nameof(_Ready))`

tutorials/scripting/c_sharp/c_sharp_basics.rst
```diff
@@ -388,3 +388,234 @@ The following tools may be used for performance and memory profiling of your man
...
+The game loop will run until the task completes, at which point execution will continue on the main thread.
+
+Let's look at the output from the above code:
+
+.. code-block::
+
+    00:00.00	Thread: 1	Task 1 started from _Ready
+    00:00.02	Thread: 1	Task 2 started from _Process
+    00:00.03	Thread: 1	Task 3 started from _Process
+    00:00.50	Thread: 1	Task 1 completed
```
tvardero, line 504: Task 1 should complete before _Process starts.
Not waiting for Task 1 to complete may result in incomlete scene initalization (for example if we were loading something from internet in _Ready function)

tvardero: Fo now solution is to block _Ready thread:
```csharp
public override _Ready() {
  DoStuffAsync().GetAwaiter().GetResult();
  // or
  DoStuffAsync().ConfigureAwait(false).Wait();
}
```

I don't thing it is good to exit _Ready() if the node is not yet ready de-facto. Telling users, who knows programming weakly, to implement `IsNodeInitialized` or `IsInstanceValid` might confuse them and make their games bugged.

tutorials/scripting/c_sharp/c_sharp_basics.rst
```diff
@@ -388,3 +388,234 @@ The following tools may be used for performance and memory profiling of your man
...
+
+Using ``async``/``await``
+-------------------------
+
+You might face a scenario where you must ``await`` a method call.
+You will notice that when you use ``await``, you are required to mark the method you use it in as ``async``, 
+and change the return type to an awaitable type, such as ``Task`` or ``Task<T>``. 
+Consequently, you must call your now ``async`` method using ``await``, 
+which propagates the problem all the way up the call chain.
+This is why many people compare ``async``/``await`` to a "zombie virus",
```
hawkerm, line 400: If you use this analogy, could be good to quote the source articles here with other best practices: (first searchable reference these days here: https://learn.microsoft.com/archive/msdn-magazine/2013/march/async-await-best-practices-in-asynchronous-programming - though this later article has an origin attributing a C# team member: https://learn.microsoft.com/archive/msdn-magazine/2015/november/asynchronous-programming-async-from-the-start)

skyace65:
Closing this due to the lack of changes to address feedback to Raulsntos.

PR #8308, opened 2023-10-22
Fix animation bug in Your first 2D game documentation

weirddan455:
I found a bug while going through the "Your first 2D game" tutorial.  If the player moves in a diagonal direction, the animation does not play.  This is because the tutorial code changes the animation twice in that case which resets the animation every frame.

To fix, create a temporary variable containing the animation to be set and only set it once in the AnimatedSprite2D node.  This way, you get the "up" animation when the player moves in a diagonal.


AThousandShips:
How does it do that? Where does it set it the second time?

AThousandShips (changes requested):
These changes do not make sense, the code will now assign `animation` once, and then check it, it does not prevent anything and adds an unnecessary change, it also now stops the animation twice if the velocity is small

AThousandShips:
The current code works correctly and the changes do not add anything to the code, chances are you made a mistake when copying the code causing your issues

PR #8309, opened 2023-10-22
Document alternative dotfile creation method for Windows

LazyWizard:
There's a much simpler method of creating a dotfile on Windows directly in Explorer that the documentation doesn't mention.

According to the comments of [this StackOverflow answer](https://superuser.com/questions/64471/create-rename-a-file-folder-that-begins-with-a-dot-in-windows/406758#406758) this method appears to have been added in Windows 7. Since [Godot only supports Windows 7 and later](https://docs.godotengine.org/en/stable/about/list_of_features.html#platforms), a warning about compatibility shouldn't be necessary.

### DIFF
--- a/tutorials/best_practices/project_organization.rst
+++ b/tutorials/best_practices/project_organization.rst
@@ -80,9 +80,12 @@ This can be useful to speed up the initial project importing.
 
 .. note::
 
-    To create a file whose name starts with a dot on Windows, you can use a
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):
TIL. That is useful. Thanks, and congrats on your first merged contribution to Godot's docs! 🎉

PR #8321, opened 2023-10-24
Use Node2D instead of PhysicsBody2D for C# collision example

dkgv:
Hi 👋

While reading the ["Getting started"](https://docs.godotengine.org/en/stable/getting_started/introduction/index.html) tutorial, I noticed that the C# method `OnBodyEntered` currently accepts a `PhysicsBody2D` object, rather than a `Node2D` object. This seems to prevent Godot from recognizing it as a valid handler for the `body_entered` signal.

I never used Godot before so it's possible I overlooked something, if that's the case please just close this PR.

### DIFF
--- a/getting_started/first_2d_game/03.coding_the_player.rst
+++ b/getting_started/first_2d_game/03.coding_the_player.rst
@@ -497,7 +497,7 @@ this code to the function:
 
  .. code-tab:: csharp
 
-    private void OnBodyEntered(PhysicsBody2D body)
+    private void OnBodyEntered(Node2D body)
     {
         Hide(); // Player disappears after being hit.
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):

raulsntos (approved):

The PR is correct. This code excerpt was quite outdated, the signal parameter was changed from `PhysicsBody2D` to `Node` in 3.2 (see https://github.com/godotengine/godot/pull/30043) and then to `Node2D` in 4.0 (see https://github.com/godotengine/godot/pull/46364).

mhilbrunner:

mhilbrunner:
Cherry-picked to 4.1.

PR #8322, opened 2023-10-24
Mention support for per-tile rotation in Using TileSets

Calinou:
This also updates screenshots to display the rotation/flip buttons.

Don't cherry-pick, as this feature isn't present in 4.1.

- See https://github.com/godotengine/godot/pull/80144.


tutorials/2d/using_tilesets.rst
```diff
@@ -605,6 +605,13 @@ atlas), but configured in different ways. For example, you may want to use the
 same tile image, but rotated, flipped, or modulated with a different color. This
 can be done using *alternative tiles*.
 
+.. tip::
+
+      Since Godot 4.2, you don't *have* to create alternative tiles to rotate or
```
skyace65, line 610: So even though it's not required is it recommended? I just ask because of the emphasis on *have* and if so that's  not clear

Calinou: It's neither required nor recommended. It's just a case of different use cases :slightly_smiling_face:

skyace65: I'd remove the * from "have" then. It just seems weird to have emphasis on it like this

mhilbrunner: @Calinou I agree we could remove the emphasis here, other than that, this PR LGTM.

skyace65 (approved):

skyace65:
Thanks!

PR #8357, opened 2023-10-29
Update README.md - one mistake.

ayushrakesh:
I have noticed one grammatical mistake in **README** inside section "**License**"

"**Ariel Manzur and the Godot**"  should be "**Ariel Manzur, and the Godot**"

Screenshot-

Screenshot (122)

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ Here are some quick links to the areas you might be interested in:
 
 ## License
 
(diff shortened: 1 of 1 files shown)


synalice (approved):

paddy-exe:
According to this article here: https://www.grammarly.com/blog/comma-before-and/ it's not a grammatical error but a choice to use it or not. As long as it's consistent with the grammar elsewhere I don't see a reason to change it now in the README

AThousandShips:
We do enforce Oxford comma in the docs, but unsure what applies here

mhilbrunner:
We use the copyright notice in a lot of places without Oxford comma, it seems (for example the official website, and in engine), even if the docs tend to use the Oxford comma. For consistency and to not create a large amount of busywork for the sake of it, I'll close this. Technically, this would be in line with the docs though :) Thanks!

PR #8358, opened 2023-10-29
Fix navigationpath example script.

rcorre:
- `move_speed` does not exist, the field is named `movement_speed`
- `get_world_3d()` will return null in a regular field assignment, accessing nodes should be done in @onready


Calinou (approved):
Thanks!

mhilbrunner:
Cherry-picked to 4.1.

PR #8361, opened 2023-10-30
Update Frequently asked questions for Godot 4.2

Calinou:
- Update type names in question about GDScript and other languages.
- Move question about plugins to be located just below the question about SDKs.
- Mention that using C++ exceptions would increase compile times, as evidenced by Godot's SCons setup recently disabling exception handling by default.
- Replace potentially misleading "2K" term with 1440p, assuming the question has mobile devices in mind (current flagships are 1440p).
- Remove question about RTTI not being used in Godot, as RTTI is now always used to ensure type safety in projects exported in release mode.
- Remove reference to option to shrink images on load, which was removed in Godot 4.0 (devices with such low resolutions are very uncommon nowadays anyway).


mhilbrunner (approved):

skyace65 (approved):

mhilbrunner:
Thank you! 🙏

PR #7640, opened 2023-07-10
Update Best practices documentation for Godot 4.1

Calinou:
Note that the Data preferences page is still marked as outdated, as the internal implementation of Dictionary/HashMap has changed in Godot 4.0. This is complex to grasp; I may take a look at it later.

- See https://github.com/godotengine/godot-docs/issues/5121.

PR #8375, opened 2023-11-01
Fix incorrect C# example using `_GetConfigurationWarnings`

kleonc:
In #7640 only the method name was changed in that example but not the return type.

### DIFF
--- a/tutorials/best_practices/godot_interfaces.rst
+++ b/tutorials/best_practices/godot_interfaces.rst
@@ -108,11 +108,13 @@ access.
         };
 
         // Warn users if the value hasn't been set.
-        public String _GetConfigurationWarnings()
+        public string[] _GetConfigurationWarnings()
         {
             if (EnemyScn == null)
-                return "Must initialize property 'EnemyScn'.";
-            return "";
+            {
+                return new string[] { "Must initialize property 'EnemyScn'." };
+            }
+            return Array.Empty<string>();
         }
     }
 


tutorials/best_practices/godot_interfaces.rst
```diff
@@ -108,11 +108,13 @@ access.
         };
 
         // Warn users if the value hasn't been set.
-        public String _GetConfigurationWarnings()
+        public String[] _GetConfigurationWarnings()
```
raulsntos, line 111: Since you are changing this, could you also use the language keyword instead of the framework type name?

```suggestion
        public string[] _GetConfigurationWarnings()
```

tutorials/best_practices/godot_interfaces.rst
```diff
@@ -108,11 +108,13 @@ access.
...
         {
             if (EnemyScn == null)
-                return "Must initialize property 'EnemyScn'.";
-            return "";
+            {
+                return { "Must initialize property 'EnemyScn'." };
```
raulsntos, line 115: This syntax can only be used in a declaration, you have to use `new` here.

```suggestion
                return new string[] { "Must initialize property 'EnemyScn'." };
```

tutorials/best_practices/godot_interfaces.rst
```diff
@@ -108,11 +108,13 @@ access.
...
+            }
+            return {};
```
raulsntos, line 117: Same, but also it'd be better to use `Array.Empty` to reuse empty arrays instead of allocating a new one.

```suggestion
            return Array.Empty<string>();
```

raulsntos (commented):
Thank you for contributing to the .NET documentation!

raulsntos (approved):
LGTM.

mhilbrunner (approved):

mhilbrunner:
Thank you :) Merged.

mhilbrunner:
Cherry-picked to 4.1.

PR #8385, opened 2023-11-03
mention molten-vk when compiling for macos

suddjian:
I had to do some research to figure out what I was missing, based on an error when running `scons platform=macos arch=arm64`: "MoltenVK SDK installation directory not found, use 'vulkan_sdk_path' SCons parameter to specify SDK path". Thought I'd open a PR to document it.

### DIFF
--- a/contributing/development/compiling/compiling_for_macos.rst
+++ b/contributing/development/compiling/compiling_for_macos.rst
@@ -55,6 +55,9 @@ To compile for Apple Silicon (ARM64) powered Macs, use::
 
     scons platform=macos arch=arm64
 
+You may need to install MoltenVK first::
+    brew install molten-vk
+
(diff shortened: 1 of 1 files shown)


contributing/development/compiling/compiling_for_macos.rst
```diff
@@ -55,6 +55,9 @@ To compile for Apple Silicon (ARM64) powered Macs, use::
 
     scons platform=macos arch=arm64
 
+You may need to install MoltenVK first::
+    brew install molten-vk
```
AThousandShips, line 59:
```suggestion

    brew install molten-vk
```

bruvzg: It is already mentioned above (VulkanSDK). And in general you should not use homebrew for it, it includes both dynamic and static library, and linker will prefer dynamic library. It's OK for test builds, but with dynamic library your build will not be portable. Also, you won't be able to build for another architecture.

PR #8388, opened 2023-11-03
Update references to community GDExtensions to point to up-to-date resources

Calinou:
godot-python doesn't work with Godot 4.1 as described in https://github.com/touilleMan/godot-python/issues/324, but there are additional language bindings we can link to now.

The example for extending Godot was replaced by an actual library binding as opposed to a language binding (which is quite a different approach in itself).


mhilbrunner (approved):

mhilbrunner:
Thanks! Merged.

PR #8405, opened 2023-11-07
Update exporting_for_android.rst for linux

vivienprouvot:
I think we can simplify the steps when installing on Linux the android export: ->For the sdk, just follow one command line found here : https://stackoverflow.com/questions/34556884/how-to-install-android-sdk-on-ubuntu , ->And with no changes for the debug.keystore ,
It works perfectly for me (debian neon plasma kde 64x)

### DIFF
--- a/tutorials/export/exporting_for_android.rst
+++ b/tutorials/export/exporting_for_android.rst
@@ -50,8 +50,18 @@ Download and install the Android SDK.
 
 .. note::
 
-    If you are using Linux,
-    **do not use an Android SDK provided by your distribution's repositories as it will often be outdated**.
+If you are using Linux,
+    **do not use an Android SDK provided by your distribution's repositories as it will often be outdated**
+    Instead of the previous steps (both Install OpenJDK and download Android SDK), just apply the command " sudo apt update && sudo apt install android-sdk "
+    The location of Android SDK on Linux can be any of the following:
+
+    /home/AccountName/Android/Sdk 
+
+    /usr/lib/android-sdk 
+
+    /Library/Android/sdk/ 
+
+    /Users/[USER]/Library/Android/sdk 
 
 
 Create a debug.keystore


AThousandShips:

If you need more information about the update process please see: [PR Workflow - Updating your branch](https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html#updating-your-branch). If wanted I can fix the branch for you.

m4gr3d:
> I think we can simplify the steps when installing on Linux the android export: ->For the sdk, just follow one command line found here : https://stackoverflow.com/questions/34556884/how-to-install-android-sdk-on-ubuntu , ->And with no changes for the debug.keystore , It works perfectly for me (debian neon plasma kde 64x)

Doesn't this go against the `note` right above that states not to use the Android SDK provided by the linux distribution?

skyace65:
I don't think this change should be merged. as m4gr3d says it goes right against the note above it, which is still a relevant note.

And changing the wording to say something about only using this on a rolling release distro doesn't work because apt isn't used on every distro. Suse uses zypper and Fedora and it's derivatives use RPM. And having commands listed for all 3 package managers seems excessive.

Just pointing everyone to get it from google is a simple instruction that works, and anyone on a rolling release should know how to get it from their package manager if they want to, they don't need instructions here.

mhilbrunner:
Thanks for contributing :) I agree with the above though and will close the PR.

vivienprouvot:
The note says not to use a SDK **provided by your distribution** but dosn't mention where to find it instead.

m4gr3d:
> Thank you all for your answers. I understand this problem may occurs only on particular distributions. 
> I would just want to specify that the note says not to use a SDK **provided by your distribution** but dosn't mention where to find it instead. 
> 

The documentation mentions you can get the Android SDK from Android Studio, or by downloading the Android SDK command line tools, both provided with links for access.

vivienprouvot:
Yes I have seen and tried that. It didn't work for me.
Furthermore, the solution I found is much simplier because it does not require to download the full android studio or the command line tools.

PR #8409, opened 2023-11-07
Improve GDExtension documentation

Calinou:
- Mention additional language bindings that support 4.x (already listed on the FAQ and List of features pages).
- Document compatibility issues with double-precision builds.
- Remove `-j4` flag from example as SCons uses most CPU cores automatically by default in godot-cpp now (like core Godot).


mhilbrunner (approved):

mhilbrunner:
Thank you! Wow, you were busy. Great work. :)

PR #8411, opened 2023-11-07
Update `list-unused-images.sh` script for recent changes

Calinou:
- Exclude all images in the top-level `img/` folder.
- Take WebP images into account.
- Fix shellcheck warnings.
- Print message with destination path of the list of unused images.
- Mark the script as executable so it can be run more easily.

### DIFF
--- a/_tools/list-unused-images.sh
+++ b/_tools/list-unused-images.sh
@@ -1,4 +1,7 @@
-#!/bin/bash
+#!/usr/bin/env bash
+
+set -uo pipefail
+IFS=$'\n\t'
 
 check_git_history=false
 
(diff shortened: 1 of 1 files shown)


_tools/list-unused-images.sh
```diff
@@ -9,19 +12,17 @@ rm -f tmp-unused-images-history
 # Exceptions are ignored, and for .svg files we also look for potential .png
 # files with the same base name, as they might be sources.
 
-exceptions="docs_logo.svg tween_cheatsheet.png"
-
-files=$(find -name "_build" -prune -o \( -name "*.png" -o -name "*.jpg" -o -name "*.svg" -o -name "*.gif" \) -print | sort)
+files=$(find -name "_build" -prune -o \( -name "*.png" -o -name "*.webp" -o -name "*.jpg" -o -name "*.svg" -o -name "*.gif" \) -print | sort)
```
mhilbrunner, line 15: I guess in theory, this should/could use `-iname` to be case insensitive (to match things like `FILE.WEBP` or `file.WebP`), or use `*.[Ww][Ee][Bb][Pp]` if we don't want to depend on GNU `find`, as `-iname` seems GNU-specific.

mhilbrunner (approved):

mhilbrunner:
@Calinou Not sure if you think it worthwhile to switch to `-iname` as discussed above, besides that, this looks good to go for me.

mhilbrunner (approved):

mhilbrunner:
Thanks!

PR #8424, opened 2023-11-08
Update list of modules that can be disabled in Optimizing for size

Calinou:
This list was confirmed to be working for the 2D platformer demo (with missing background music if the Ogg/Vorbis modules are left disabled).

This adds instructions for disabling the advanced text server and using the fallback text server instead. The list has been reorganized to list options with the most space savings at the top.

This also removes the platform options from each example as SCons uses autodetection by default now.


mhilbrunner (approved):

mhilbrunner:
LGTM, thanks.

PR #8431, opened 2023-11-09
Clarify general usage of CharacterBody2D

Offlein:
As someone learning Godot, I had a great deal of confusion about the intended usage of CharacterBody2D. 

Specifying in the first sentence that it is used for "user-controlled" gives an (I think?) improper specification for the use of these nodes.

I saw some tutorials using CharacterBody2D/3D for NPCs and I went to Discord to ask if that was improper usage given the documentation, and was advised that the word "user-" (in "user-controlled") was intended to refer to the developer, and that the real purpose is for script-controlled physics body. But I do not think this is fully true, since the 2nd usage of "user-controlled characters" seems to more-clearly refer to player-controlled ones.

This PR hopefully clarifies the usage of CharacterBody2D/3D as "it's for scripted bodies", makes it clear that this is perhaps most commonly used for player-controlled characters, and specifies "player" instead of "user" in the other verbiage so as not to equivocate on what a "user" is (versus a "player").

My first Godot PR so apologies if I've done something wrong.

### DIFF
--- a/classes/class_characterbody2d.rst
+++ b/classes/class_characterbody2d.rst
@@ -19,7 +19,7 @@ A 2D physics body specialized for characters moved by script.
 Description
 -----------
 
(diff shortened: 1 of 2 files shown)


Calinou:

Offlein:
@Calinou My apologies! Thank you so much.

PR #8433, opened 2023-11-09
docs: add missing camera translation Y 15

rolbr:
necessary to match following `image8`

### DIFF
--- a/getting_started/first_3d_game/03.player_movement_code.rst
+++ b/getting_started/first_3d_game/03.player_movement_code.rst
@@ -404,6 +404,9 @@ one).
 
 |image7|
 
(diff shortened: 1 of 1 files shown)


skyace65:
I think you may have followed the tutorial incorrectly. While the images aren't a perfect match to what I'm seeing in the project adding 15 to the
 Y axis of the camera or camera pivot definitely does not get you that image. I took some screenshots of what that does:
CameraAdjustPivot
CameraAdjustCamera

Could you provide us a zip file of what your copy of that project so we can see what's going on?

skyace65:
Closing due to the lack of a response

PR #8436, opened 2023-11-09
Document glTF base path in Runtime file loading and saving

Calinou:
- See https://github.com/godotengine/godot-demo-projects/pull/993.

PR #8441, opened 2023-11-10
Document generating mipmaps in Runtime file loading and saving

Calinou:
- Follow-up to https://github.com/godotengine/godot-docs/pull/8436.

This is often needed when displaying loaded images on a 3D surface.

In a future PR, we should look into adding a code sample that loops over textures from a 3D scene (GLTFState has a method for this) and generate mipmaps on all textures that need it automatically.


mhilbrunner (approved):

mhilbrunner:
Thanks, merged :)

PR #8454, opened 2023-11-11
Add `--headless` parameter to "compiling with dotnet" examples

adamscott:
We already tell users to use `--headless` to generate the glue. This PR make it so that we use `--headless` in the examples.

https://github.com/godotengine/godot-docs/blob/9212f4f0a2205186e839e532d45a23a55dbdf917/contributing/development/compiling/compiling_with_dotnet.rst?plain=1#L28-L38


AThousandShips (approved):
Makes sense, and matches what we do in CI so works well

mhilbrunner (approved):

mhilbrunner:
Thanks! Merged. :)

PR #8457, opened 2023-11-12
Change comment to use new executable name: godot.linuxbsd.editor.dev.x86_64.llvm

rakkarage:
says:

`godot.linuxbsd.tools.64.llvm`

but, in my experience the file is actually named:

`godot.linuxbsd.editor.dev.x86_64.llvm`

and it will then match the subsequent lines.

### DIFF
--- a/contributing/development/configuring_an_ide/visual_studio_code.rst
+++ b/contributing/development/configuring_an_ide/visual_studio_code.rst
@@ -80,7 +80,7 @@ To run and debug the project you need to create a new configuration in the ``lau
       "name": "Launch Project",
       "type": "lldb",
       "request": "launch",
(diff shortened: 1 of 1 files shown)


Calinou:
The `.dev` prefix is only present for builds with `dev_build=yes` builds, which isn't the default. I suggest naming it just `godot.linuxbsd.editor.x86_64.llvm`.

rakkarage:
Thanks but both real commands that the comment suggest replacing also have dev in name?

skyace65 (approved):

skyace65:
rakkarage is correct, the actual code the comment is for uses "dev" so I'll merge this. If we don't want people to be doing it this way that can be changed in a separate PR.

PR #8458, opened 2023-11-12
Update recording_with_microphone.html to reflect GDscript code

brno32:
as can be seen in https://github.com/godotengine/godot-demo-projects/blob/master/audio/mic_record/MicRecord.gd, the methods that are connected to signals have different naming in the GDScript code than is reflected in the docs on the demo project


AThousandShips:

If you need more information about the update process please see: [PR Workflow - Updating your branch](https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html#updating-your-branch). If wanted I can fix the branch for you.

AThousandShips:
This matches the code in the demo project, but it does *not* match what you will get with the editor if you follow the tutorial, the naming matches the default settings

PR #8459, opened 2023-11-12
Add visual shader node interface explanation

ShlomiRex:
Added visual shader port types, which greatly increases clarity for newbies.

image


AThousandShips:

ShlomiRex:
@AThousandShips done

skyace65 (approved):

mhilbrunner (approved):
Looks good to me, but I wonder how maintainable the example screenshots are, as we struggle to keep screenshots updated with Godot updates.

So we should consider how many of these we would expect to be adding for this, how often we expect them to change, and how much value they add. Happy to hear y'alls thoughts :)

ShlomiRex:
@mhilbrunner I agree that the screenshot represent UI that can change over time. However, its just an example and not a direct fact.
The screenshots of the UI help users visualize different types of ports by their color (which is the main reason for the screenshots, since I can't post the colors of the ports).
I dont expect we would add more to this list, because the point was to demonstrate the basic port types: scalar, vector, matrix, and so on. It's not subject to change.

skyace65:
If it's just these screenshots I would consider this maintainable

skyace65 (changes requested):
Missed this in the original review, apologies. But the outline for the vs_node image needs to be a different color. We use a specific shade of yellow (hex color is fffb44, fffb44ff if you're using something with transparency)  so color blind people don't have issues reading the documentation.

ShlomiRex:
> Missed this in the original review, apologies. But the outline for the vs_node image needs to be a different color. We use a specific shade of yellow (hex color is fffb44, fffb44ff if you're using something with transparency) so color blind people don't have issues reading the documentation.

Thank you. I read the instructions in: https://docs.godotengine.org/en/stable/contributing/documentation/docs_image_guidelines.html

I should have read that before submitting. My mistake.
Fixed:
image

paddy-exe:
I would perhaps hold off this PR from this for a while longer as there's a redesign coming in terms of both UX and Color Design: https://github.com/godotengine/godot/pull/85017

skyace65:
> I would perhaps hold off this PR from this for a while longer as there's a redesign coming in terms of both UX and Color Design: [godotengine/godot#85017](https://github.com/godotengine/godot/pull/85017)

I get it, but the work that's done for this PR is already done. I don't see holding off on this PR as giving us less work in the future.

skyace65 (approved):

skyace65:
Thanks!

PR #8461, opened 2023-11-12
Increase clarity of scripting_player_input.rst

theDrake:
Added two commas before lesson titles, changed "properties like" to "properties, such as" (for greater clarity in the context of the full sentence), and changed "classes'" to "class's" (for consistency with the previous sentence).

### DIFF
--- a/getting_started/step_by_step/scripting_player_input.rst
+++ b/getting_started/step_by_step/scripting_player_input.rst
@@ -7,7 +7,7 @@
 Listening to player input
 =========================
 
-Building upon the previous lesson :ref:`doc_scripting_first_script`, let's look
+Building upon the previous lesson, :ref:`doc_scripting_first_script`, let's look
 at another important feature of any game: giving control to the player.
 To add this, we need to modify our ``sprite_2d.gd`` code.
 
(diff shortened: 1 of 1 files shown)


getting_started/step_by_step/scripting_player_input.rst
```diff
@@ -181,12 +181,12 @@ Summary
...
 In summary, every script in Godot represents a class and extends one of the
 engine's built-in classes. The node types your classes inherit from give you
-access to properties like ``rotation`` and ``position`` in our sprite's case.
+access to properties, such as ``rotation`` and ``position`` in our sprite's case.
 You also inherit many functions, which we didn't get to use in this example.
 
 In GDScript, the variables you put at the top of the file are your class's
 properties, also called member variables. Besides variables, you can define
-functions, which, for the most part, will be your classes' methods.
+functions, which, for the most part, will be your class's methods.
```
AThousandShips, line 189: This is supposed to be plural, the above line uses the singular, but here we're talking about classes in general

theDrake: Sure, fair enough. I think it works either way and can imagine arguments for both, but I'll go ahead and get this pull request changed (or submit a new one). Thanks for the feedback!

PR #8471, opened 2023-11-14
Document using EditorScript in Running code in the editor

Calinou:
EditorScript is useful for one-off operations such as:

- Use as a playground for GDScript or C# scripting without having to run a project.
- Scale all light nodes in the currently edited scene.
- Replace nodes that were copy-pasted with scene instances.


tutorials/plugins/running_code_in_the_editor.rst
```diff
@@ -286,6 +286,89 @@ By default, the warning only updates when closing and reopening the scene.
...
+
+    # This function is recursive: it calls itself to get lower levels of child nodes as needed.
+    # `children_acc` is the accumulator parameter that allows this function to work.
+    # It should be left to its default value when you call this function directly.
+    func get_all_children(in_node, children_acc = []):
+        children_acc.push_back(in_node)
+        for child in in_node.get_children():
+            children_acc = get_all_children(child, children_acc)
+
+        return children_acc
```
Calinou, line 362: I'd like this script not to affect children of instanced subscenes, but I'm not sure how to detect those (regardless of whether Editable Children is enabled or not on them).

YuriSizov:
I think the most obvious use case for it is as a GDScript playground. I use it all the time to check what some code would do or output without having to run a project or use external resources.

mhilbrunner (approved):

mhilbrunner:
Thank you!

PR #8486, opened 2023-11-16
Update exporting_pcks.rst

Heremeus:
Updating C# assembly loading instructions according to https://github.com/godotengine/godot/issues/75352#issuecomment-1807234258. Simply loading the assembly is no longer enough with godot 4 and @granitrocky identified how to load the dlls correctly.

Wasn't sure how to properly format the code inside the note block, so please suggest improvements, if necessary.

### DIFF
--- a/tutorials/export/exporting_pcks.rst
+++ b/tutorials/export/exporting_pcks.rst
@@ -132,7 +132,7 @@ The PCK file contains a “mod_scene.tscn” test scene in its root.
 .. note::
     For a C# project, you need to build the DLL and place it in the project directory first.
     Then, before loading the resource pack, you need to load its DLL as follows:
-    ``Assembly.LoadFile("mod.dll")``
+    ``var context = AssemblyLoadContext.GetLoadContext(typeof(Godot.Bridge.ScriptManagerBridge).Assembly); var assembly = context.LoadFromAssemblyPath("mod.dll"); Godot.Bridge.ScriptManagerBridge.LookupScriptsInAssembly(assembly);``
 
 Summary
 -------


tutorials/export/exporting_pcks.rst
```diff
@@ -132,7 +132,7 @@ The PCK file contains a “mod_scene.tscn” test scene in its root.
 .. note::
     For a C# project, you need to build the DLL and place it in the project directory first.
     Then, before loading the resource pack, you need to load its DLL as follows:
-    ``Assembly.LoadFile("mod.dll")``
+    ``var context = AssemblyLoadContext.GetLoadContext(typeof(Godot.Bridge.ScriptManagerBridge).Assembly); var assembly = context.LoadFromAssemblyPath("mod.dll"); Godot.Bridge.ScriptManagerBridge.LookupScriptsInAssembly(assembly);``
```
raulsntos, line 135:
```suggestion

    .. code-block:: csharp

        var context = Assembly.LoadContext.GetLoadContext(typeof(Godot.Bridge.ScriptManagerBridge).Assembly);
        var assembly = context.LoadFromAssemblyPath("mod.dll");
        Godot.Bridge.ScriptManagerBridge.LookupScriptsInAssembly(assembly);
```

raulsntos (commented):
The `LookupScriptsInAssembly` method (as most everything else in the `Godot.Bridge` namespace) is not meant to be used by users, if this API is exposed it's only because it's used by our source generators but otherwise it'd be internal API.

So I don't feel comfortable telling users to use this API. At least not without the proper explanation of what's happening.

Also, the plan was to make the type `ScriptManagerBridge` internal and replace the `LookupScriptsInAssembly` method with source generators (`ScriptRegistrarGenerator`), so it'd be better if users didn't start relying on this API since we'll likely break compatibility here.

granitrocky:
@raulsntos That was a genuine concern of mine since there is a comment in the `ScriptManagerBridge` saying that the `LookupScriptsInAssembly` method is planned to go private. An alternative could be setting up an `AssemblyLoad` callback that calls it internally

Heremeus:
> The `LookupScriptsInAssembly` method (as most everything else in the `Godot.Bridge` namespace) is not meant to be used by users, if this API is exposed it's only because it's used by our source generators but otherwise it'd be internal API.
> 
> So I don't feel comfortable telling users to use this API. At least not without the proper explanation of what's happening.
> 
> Also, the plan was to make the type `ScriptManagerBridge` internal and replace the `LookupScriptsInAssembly` method with source generators (`ScriptRegistrarGenerator`), so it'd be better if users didn't start relying on this API since we'll likely break compatibility here.

That's totally fair, thanks for the insight! Feel free to close this issue and notify other users on the [original issue](https://github.com/godotengine/godot/issues/75352#issuecomment-1807234258). If I understood it correctly, the assembly workflow will be reworked in a future version which should fix the issue?

Carsillas:
> The `LookupScriptsInAssembly` method (as most everything else in the `Godot.Bridge` namespace) is not meant to be used by users, if this API is exposed it's only because it's used by our source generators but otherwise it'd be internal API.
> 
> So I don't feel comfortable telling users to use this API. At least not without the proper explanation of what's happening.
> 
> Also, the plan was to make the type `ScriptManagerBridge` internal and replace the `LookupScriptsInAssembly` method with source generators (`ScriptRegistrarGenerator`), so it'd be better if users didn't start relying on this API since we'll likely break compatibility here.

I hope it is prioritized that the original issue is resolved prior to making this breaking change, I intend on using this in the meantime and would like to not be completely blocked from keeping my engine up to date.

haltingstate:
What is the method now, to load scripts that are in runtime loadable assemblies, or to load scripts that are outside of the Godot project folder?

vonweller:
How do I update mods in C#? Is there a reasonable suggestion from the authorities?

vonweller:
> The `LookupScriptsInAssembly` method (as most everything else in the `Godot.Bridge` namespace) is not meant to be used by users, if this API is exposed it's only because it's used by our source generators but otherwise it'd be internal API.`LookupScriptsInAssembly` 方法（就像 `Godot.Bridge` 命名空间中的大多数其他内容一样）不是为了供用户使用，如果这个 API 被公开，那只是因为它被我们的源代码生成器使用，否则它将是内部 API。
> 
> So I don't feel comfortable telling users to use this API. At least not without the proper explanation of what's happening.所以我不愿意告诉用户使用这个 API。至少在没有对正在发生的事情进行适当解释的情况下不会。
> 
> Also, the plan was to make the type `ScriptManagerBridge` internal and replace the `LookupScriptsInAssembly` method with source generators (`ScriptRegistrarGenerator`), so it'd be better if users didn't start relying on this API since we'll likely break compatibility here.此外，计划是将 `ScriptManagerBridge` 类型设为内部，并将 `LookupScriptsInAssembly` 方法替换为源生成器 （`ScriptRegistrarGenerator`），因此如果用户不要开始依赖此 API 会更好，因为我们可能会在这里破坏兼容性。

How do I update mods in C#? Is there a reasonable suggestion from the authorities?

vonweller:
@raulsntos How do I update mods in C#? Is there a reasonable suggestion from the authorities?

skyace65:
Closing this PR due to the lack of updates to it addressing raulsntos' review.

PR #8489, opened 2023-11-17
Fixing style guide example on 2 indent levels for continuation lines

Ero98:
The original example seems to be using 3 tabs for indentation, but the description said 2 should be used.


skyace65:

PR #7370, opened 2023-05-18
Snake-case .tscn, .gd and _on_* callbacks

mhilbrunner:
Salvaging the previous work in #6598.
Closes https://github.com/godotengine/godot-docs/issues/6245.

PR #8494, opened 2023-11-17
Sprite2D.gd not sprite_2d.gd in scripting_first_script.rst

rochamarcelo:
While following the doc I noticed the file name of the script created don't match the one informed. It should be Sprite2D.gd not sprite_2d.gd

https://docs.godotengine.org/en/stable/getting_started/step_by_step/scripting_first_script.html


AThousandShips:
This was already changed to the current format is the change should not be reverted:
* https://github.com/godotengine/godot-docs/pull/7370

PR #8497, opened 2023-11-18
"as long" -> "as long as" in 03.player_movement_code.rst

theDrake:
Changed "as long it is not on or collides" to "as long as it is not on or colliding".


mhilbrunner (approved):

mhilbrunner:
Thanks! :) Merged.

PR #8504, opened 2023-11-19
Clarify InputEventScreenDrag only applies to touch events

avniavdulla:
I was trying to use this event to check when I was clicking and dragging my mouse across the screen. It was not working, so I asked in Discord.  I was told this is only used for touchscreen. So I added this clarification so others are not confused.


AThousandShips:

Issue #7958, opened 2023-09-19
Documentation in tutorial clamp args inaccurate


soliforte:
**Your Godot version:**
4.1

**Issue description:**
In the tutorial, this line of code appears to be inaccurate:
`position = position.clamp(Vector2.ZERO, screen_size)`

Throws the error: Invalid type in function 'clamp' in base 'Vector2'. Cannot convert argument 2 from Rect2 to Vector2.

Issue was rectified by using screen_size.size.

**URL to the documentation page:**

https://docs.godotengine.org/en/stable/getting_started/first_2d_game/03.coding_the_player.html

If you know how to fix the issue you are reporting please
consider opening a pull request. We provide a tutorial on
using git here: https://docs.godotengine.org/en/stable/community/contributing/pr_workflow.html,
writing documentation at https://docs.godotengine.org/en/stable/community/contributing/docs_writing_guidelines.html
and contributing to the class reference here: https://docs.godotengine.org/en/stable/community/contributing/updating_the_class_reference.html

AThousandShips:
Please make sure you have copied the instructions correctly, you seem to have copied this part wrong and missed the `size` part:
```gdscript
func _ready():
    screen_size = get_viewport_rect().size
```
Because otherwise the value of `screen_size` is in fact a `Vector2`.

AThousandShips:
Closing as not a bug, if you have any other questions please reply here

PR #8514, opened 2023-11-21
Fix syntax error

jacola:
As mentioned in https://github.com/godotengine/godot-docs/issues/7958, the tutorial documentation is incorrect and will not run as is.

As is the code throws `Invalid type in function 'clamp' in base 'Vector2'. Cannot convert argument 2 from Rect2 to Vector2.` I corrected the code by changing `screen_size` to `screen_size.size`, however updating `_ready()` to use `get_viewport_rect().size` would also be a valid solution.

### DIFF
--- a/getting_started/first_2d_game/03.coding_the_player.rst
+++ b/getting_started/first_2d_game/03.coding_the_player.rst
@@ -232,7 +232,7 @@ the ``_process`` function (make sure it's not indented under the `else`):
  .. code-tab:: gdscript GDScript
 
         position += velocity * delta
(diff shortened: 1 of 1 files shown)


AThousandShips (changes requested):
This is not a valid fix, you have almost certainly copied this line incorrectly:
```gdscript
screen_size = get_viewport_rect().size
```

AThousandShips:
Closing as per the above and that the actual solution was already explained and the issue closed

jacola:
Thanks, I should have caught that😅

PR #8526, opened 2023-11-24
Fix release builds for new GDExtension developers by fixing release identifiers in `.gdextension`

MoralCode:
In working on [my own GDExtension](https://github.com/Computational-Geometry-G1/godot-wiimote) as a first-time godot developer, I have discovered (while debugging a performance related issue) that the example of a `.gdextension` In the tutorial (the C++ example) is incorrect.

When using scons to build my extension in release mode (`scons target=template_release`) my demo project would fail to open in godot without loading errors. it was only after I saw [this comment](https://github.com/godotengine/godot/issues/71090#issuecomment-1380386350) and tried changing the release identifier to `linux.x86_64` (from `linux.release.x86_64`) that godot successfully loaded with the release version of my extension.

godot 4.1

### DIFF
--- a/tutorials/scripting/gdextension/gdextension_cpp_example.rst
+++ b/tutorials/scripting/gdextension/gdextension_cpp_example.rst
@@ -368,21 +368,21 @@ loaded for each platform and the entry function for the module. It is called ``g
     [libraries]
 
     macos.debug = "res://bin/libgdexample.macos.template_debug.framework"
-    macos.release = "res://bin/libgdexample.macos.template_release.framework"
+    macos = "res://bin/libgdexample.macos.template_release.framework"
     windows.debug.x86_32 = "res://bin/libgdexample.windows.template_debug.x86_32.dll"
-    windows.release.x86_32 = "res://bin/libgdexample.windows.template_release.x86_32.dll"
+    windows.x86_32 = "res://bin/libgdexample.windows.template_release.x86_32.dll"
     windows.debug.x86_64 = "res://bin/libgdexample.windows.template_debug.x86_64.dll"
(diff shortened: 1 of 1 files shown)


AThousandShips:
If this is incorrect then the code in `godot-cpp` is also incorrect

bruvzg:
This is not an intended way to do it, target tag probably should be `template_release`, `template_debug` and `editor`, but definitely not empty. If it's not working, it's a bug in the engine, not a documentation issue.

AThousandShips:
It also seems to work for godot Jolt, and the standards, it uses the *tag* not the suffix, it should be `debug` and `release` respectively AFAIK

AThousandShips:
Assuming this is verified by the gdextension team by the above, I think this can be closed, based on my analysis of the code

MoralCode:
so if scons is building using the `template_release` configuration, should the values in `.gdextension` be `linux.template_release.arm64 = "res://bin/libgdexample.linux.template_release.arm64.so"`?

dsnopek:
> so if scons is building using the `template_release` configuration, should the values in `.gdextension` be `linux.template_release.arm64 = "res://bin/libgdexample.linux.template_release.arm64.so"`?

That should work, but so should:

`linux.release.arm64 = "res://bin/libgdexample.linux.template_release.arm64.so"`

The parts of the `linux.release.arm64` are _feature tags_, so `linux`, `release` and `arm64`. All feature tags must be present in the Godot build for a particular GDExtension to be selected. And both `release` and `template_release` are valid feature tags that are present in `template_release` builds of Godot.

See here for more information about feature tags:

https://docs.godotengine.org/en/stable/tutorials/export/feature_tags.html

PR #8553, opened 2023-12-01
Update 'Your first 3D shader' for better 4.2 compatibility

interpol-kun:
New mesh shaders are being generated with light() function, which affects the tutorial flow. When adding OmniLight3D, nothing will happen to your mesh, unless you delete light() function from a shader. I've just encountered that problem and fixed it only by chance (diffing my shader code and shader code from the end of tutorial)

### DIFF
--- a/tutorials/shaders/your_first_shader/your_first_3d_shader.rst
+++ b/tutorials/shaders/your_first_shader/your_first_3d_shader.rst
@@ -367,7 +367,8 @@ We can even drag the light around and the lighting will update automatically.
 .. note:: The default :ref:`FastNoiseLite <class_fastnoiselite>` implementation in Godot changed in Godot 4. If you want
 your mesh to look smoother, try lowering the number of Octaves in :ref:`FastNoiseLite <class_fastnoiselite>` of your normal map.
 
-Here is the full code for this tutorial. You can see it is not very long as Godot handles most of the difficult stuff for you.
+Here is the full code for this tutorial. You can see it is not very long as
+Godot handles most of the difficult stuff for you.
 
 .. code-block:: glsl
 
--- a/tutorials/shaders/your_first_shader/your_first_3d_shader.rst
+++ b/tutorials/shaders/your_first_shader/your_first_3d_shader.rst
@@ -292,9 +292,6 @@ the light is affecting the terrain as if it were a flat plane. This is because
 the light shader uses the normals from the :ref:`Mesh <class_mesh>` to calculate
 light.
 
-.. note:: If you don't see :ref:`OmniLight3D <class_OmniLight3D>` affecting your mesh,
-          delete the ``light()`` function from the shader.
-
 The normals are stored in the Mesh, but we are changing the shape of the Mesh in
 the shader, so the normals are no longer correct. To fix this, we can
 recalculate the normals in the shader or use a normal texture that corresponds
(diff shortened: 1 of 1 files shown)


tutorials/shaders/your_first_shader/your_first_3d_shader.rst
```diff
@@ -292,6 +292,9 @@ the light is affecting the terrain as if it were a flat plane. This is because
 the light shader uses the normals from the :ref:`Mesh <class_mesh>` to calculate
 light.
 
+.. note:: If you don't see OmniLight3D<class_OmniLight3D> affecting your mesh,
```
AThousandShips, line 295:
```suggestion
.. note:: If you don't see :ref:`OmniLight3D <class_OmniLight3D>` affecting your mesh,
```

mhilbrunner (approved):

clayjohn:
This comment will become partially out of date with 4.x/4.2.1 https://github.com/godotengine/godot/pull/84594 since the light() function won't be defined by default anymore

interpol-kun:
> This comment will become partially out of date with 4.x/4.2.1 [godotengine/godot#84594](https://github.com/godotengine/godot/pull/84594) since the light() function won't be defined by default anymore

That's good. Maybe I can update the other parts of this tutorial, can you navigate me a little bit? For example, this is how the mesh looks like in tutorial with fixed normals and OmniLight3D on the scene:
image

Here how it looks in the 4.2 editor with the same steps reproduced:
image

Notice, that albedo is being affected by noise texture, and I have completely different results. I've played with default environment and direction light, but couldn't get the same result as in tutorial.

The-oldone:
It's strange this pull request hasn't been approved fully but has been approved partially after so long. Putting that sentence would help people a lot. If something like having discussed this privately but planning to implement this with a bunch of other changes despite it being potentially misleading for new users has occurred, please let me know and I will stop doing this.

But for now, I believe @clayjohn needs a bump to review this

clayjohn:
> It's strange this pull request hasn't been approved fully but has been approved partially after so long. Putting that sentence would help people a lot. If something like having discussed this privately but planning to implement this with a bunch of other changes despite it being potentially misleading for new users has occurred, please let me know and I will stop doing this.
> 
> But for now, I believe @clayjohn needs a bump to review this

I'm not sure about this change. Its a note about recreating how the 3.x version of this tutorial looked. But I think that is confusing for new users and is only helpful for people who are already familiar with the 3.x version of the tutorial. Since our target audience is people who are new to writing shaders, I don't think it makes sense to add this.

The-oldone:
> Its a note about recreating how the 3.x version of this tutorial looked. But I think that is confusing for new users and is only helpful for people who are already familiar with the 3.x version of the tutorial.

OK wait, do you mean you will adjust the images to make it look how the tutorial will end up looking as in godot 4 by following the steps of the tutorial instead of modifying the code to make it look the same way as it did in godot 3?

clayjohn:
> OK wait, do you mean you will adjust the images to make it look how the tutorial will end up looking as in godot 4 by following the steps of the tutorial instead of modifying the code to make it look the same way as it did in godot 3?

If the images are out of date then they should be updated. The tutorial should not reference a different version of the same tutorial.

skyace65:
Closing this PR as I've made a PR to update the images here: https://github.com/godotengine/godot-docs/pull/10554

PR #8563, opened 2023-12-04
Update changing_application_icon_for_windows.rst

wlwl2:
The icons appear blurry if the order of the sizes in the commands are reversed. This happened in v3.5.stable, but I did a quick check (correct me if I'm wrong) and I think there has not been much change since then.

Related to https://github.com/godotengine/godot/issues/64073.

I'll add some tips to refresh the icon cache and thumbnails in another pull request.

### DIFF
--- a/tutorials/export/changing_application_icon_for_windows.rst
+++ b/tutorials/export/changing_application_icon_for_windows.rst
@@ -23,13 +23,13 @@ using this `ImageMagick <https://www.imagemagick.org/>`_ command:
 
 .. code-block:: none
 
-    magick convert icon.png -define icon:auto-resize=256,128,64,48,32,16 icon.ico
+    magick convert icon.png -define icon:auto-resize=16,32,48,64,128,256 icon.ico
 
 Depending on which version of ImageMagick you installed, you might need to leave out the ``magick`` and run this command instead:
 
 .. code-block:: none
 
-    convert icon.png -define icon:auto-resize=256,128,64,48,32,16 icon.ico
+    convert icon.png -define icon:auto-resize=16,32,48,64,128,256 icon.ico
 
 .. warning::
 


skyace65:
I made a test project in 4.2.1 and exported it with an ICO file that was created with the original size order, and with the new order. I'm not seeing any visual difference between the two on Windows 11 in the taskbar, or in file explorer with the extra large icons option. I will not be able to test win 10 until I'm back from the holidays. did you use godoticon to change the icon instead of rcedit?

Here's how the taskbar looks for me:
Taskbar

And file explorer:
File Explorer

wlwl2:
@skyace65 You don't have to do any of that. This is additionally done to keep the order consistent after generation. The inverse order can be seen using other tools.

skyace65:
I don't understand why you want to change the size order in the docs when there are merged PR's for the 3.x and 4.x branch that fixed this issue.

wlwl2:
@skyace65 Which PR's are you talking about exactly?

wlwl2:
@skyace65 The .ico file generated by GIMP has a different order than the magick command. Is there a reason for keeping the order inconsistent?

wlwl2:
> I don't understand why you want to change the size order in the docs when there are merged PR's for the 3.x and 4.x branch that fixed this issue.

I just realized something about my first post. I was talking about the command and order of the images changing not the blurriness being fixed (you can see this when you open those .ico files in GIMP).

bruvzg:
Source icon size order should not matter, it is rearranged to have correct order on export anyway, see https://github.com/godotengine/godot/pull/68828 and https://github.com/godotengine/godot/pull/68858.

bruvzg:
I think it's rcedit bug, it's not replacing all icon related resources so it should have the same order as the default icon.

wlwl2:
Alright, @bruvzg that's enough for now. Really appreciate that PR. I need to do a check on rcedit myself.

mhilbrunner:
Closing as per the above discussion. Thanks for taking the time to contribute!

wlwl2:
Just in case anyone was wondering (I do regret not closing this pull request and entire matter sooner), the latest state seems to be that rcedit is probably going to be completely dropped. https://github.com/godotengine/godot/pull/75950#issue-1663024371

PR #8568, opened 2023-12-04
Fix `/` keyboard shortcut to allow typing `/` within the search field

Calinou:
- Select all existing text in the field when pressing the shortcut.
- Remove broken Algolia integration (unused since 4.0) and move the shortcut code to the main `custom.js` file.


mhilbrunner (approved):

mhilbrunner:
Cherry-picked to 4.2.

PR #8569, opened 2023-12-04
Improve color constants cheat sheet

theowiik:
Improve the color constants cheat sheet from [PR](https://github.com/godotengine/godot-docs/pull/3881) which is linked in [this page](https://docs.godotengine.org/en/stable/classes/class_color.html) 

- Improve sorting. Sorted with "Step Sorting" as described [here](https://www.alanzucconi.com/2015/09/30/colour-sorting/) instead of hue
- Properly display transparency
- Increase resolution
- Use the default Godot font

_(Also ported the project which this was made with to Godot 4 [godolors](https://github.com/theowiik/godolors) 🙂)_

image


AThousandShips:
While you're at it, could you please convert the image to the `.webp` format.

timothyqiu:
Maybe also change the identifiers to match the Godot 4 version?

e.g. `whitesmoke` is now `WHITE_SMOKE`.

theowiik:
> Maybe also change the identifiers to match the Godot 4 version?
> 
> e.g. `whitesmoke` is now `WHITE_SMOKE`.

Updated the cheat sheet to use the new names
colors

mhilbrunner:
Great work! I'm unsure whether +440kb is a bit heavy though. Did you try using one of the various PNG optimizers, or maybe decreasing resolution a tiny bit?

theowiik:
> Great work! I'm unsure whether +440kb is a bit heavy though. Did you try using one of the various PNG optimizers, or maybe decreasing resolution a tiny bit?

Compressed it so it is now only +38.9 KB (120%) larger (while still being a much higher resolution)

theowiik:
@mhilbrunner anything else to fix? :slightly_smiling_face:

mhilbrunner:
@theowiik Thanks for the ping!

In theory, this should also now be WebP (we're in the process of moving the docs repository over to WebP, see [image creation guidelines](https://docs.godotengine.org/en/stable/contributing/documentation/docs_image_guidelines.html)).

However, this would also require a PR for the class reference to update the URL, and adding a redirect so the old link keeps working. If you want to do that, go ahead, but otherwise I'm not against just merging this considering how long it has been open. @skyace65 thoughts?

Calinou (approved):
Thanks!

I force-pushed a commit with `oxipng --strip --zopfli` to further reduce file size by 5.49%.

PR #8578, opened 2023-12-06
Fix typos the scene tree key_concepts_overview.rst

khoanguyen8496:
Hi team,

While going through the introduction, I don't know whether this definition of **The scene tree** is supposed to be tree of scenes instead of tree of nodes, which is repeated twice. If this is intention please ignore this pull request.

Thank you.


skyace65:
theowiik is correct, there's nothing wrong with "scenes" being used in that sentence.

PR #8593, opened 2023-12-09
Document Fullscreen vs Exclusive Fullscreen window mode for integer scaling

Calinou:
See https://forum.godotengine.org/t/integer-scale-mode-breaks-in-fullscreen/35119.

**Note:** Don't cherry-pick to 4.1, as integer scaling is only present in 4.2 and and later.

### DIFF
--- a/tutorials/rendering/multiple_resolutions.rst
+++ b/tutorials/rendering/multiple_resolutions.rst
@@ -328,6 +328,22 @@ that have a pixel art UI, so that the visible area in the 3D viewport doesn't
 reduce in size (which occurs when using ``canvas_items`` or ``viewport`` stretch
 mode with the ``integer`` scale mode).
 
+.. tip::
+
+    Games should use the **Exclusive Fullscreen** window mode, as opposed to
+    **Fullscreen** which is designed to prevent Windows from automatically
+    treating the window as if it was exclusive fullscreen.
+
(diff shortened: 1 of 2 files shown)


tutorials/rendering/multiple_resolutions.rst
```diff
@@ -328,6 +328,22 @@ that have a pixel art UI, so that the visible area in the 3D viewport doesn't
 reduce in size (which occurs when using ``canvas_items`` or ``viewport`` stretch
 mode with the ``integer`` scale mode).
 
+.. tip::
```
skyace65, line 331: The information here is good, but it almost feels like it should be it's own section on this page, and have the bit at the end about integer scaling remain a tip in this section. I get that fullscreen vs exclusive fullscreen isn't fully about multiple resolutions but there's no other place to put it really belongs in the docks.

Regardless I'd be ok with merging as is.

skyace65 (approved):

mhilbrunner:
Merging, as it is an improvement. Agree with skyace65 that this maybe warrants another pass to see if things should be moved to new sections; IMO its somewhat on the border of being too much for a tip, not sure. Thanks anyway, merged as is for now :)

mhilbrunner:
Cherry-picked to 4.2.

PR #8607, opened 2023-12-12
Areas have the input pickable property set to true as default, so I removed the part that said they didn't.

### LINKED ISSUE #6843: `_input_event` does not travel in reverse depth-first order as documented
**Your Godot version:**
4.0 rc6 and 3.5.1

**Issue description:**
On both versions of Godot, I made several Area2Ds with a script that would print its name when it receives a click event. They printed in a seemingly random order and not in tree order, and the order wasn't the same when trying again. All Area2Ds were the children of the same node.

Also:

> If this ray hits an object, it will call the [CollisionObject._input_event()](https://docs.godotengine.org/en/stable/classes/class_collisionobject.html#class-collisionobject-method-input-event) function in the relevant physics object (bodies receive this callback by default, but areas do not. This can be configured through [Area](https://docs.godotengine.org/en/stable/classes/class_area.html#class-area) properties).

When I instanced the Area2D the input pickable property was set to true by default. I think this might be wrong.

Also, on the latest docs:
> When sending events to its child and descencand nodes

Does that mean "descended"?

https://docs.godotengine.org/en/latest/tutorials/inputs/inputevent.html
https://docs.godotengine.org/en/stable/tutorials/inputs/inputevent.html

ShatReal:
Closes #6843 (the other stuff in it was fixed by other pull requests it seems).

### DIFF
--- a/tutorials/inputs/inputevent.rst
+++ b/tutorials/inputs/inputevent.rst
@@ -113,8 +113,7 @@ received input, in order:
    In the case of a 3D scene if a :ref:`Camera3D <class_Camera3D>` is assigned to the Viewport, a ray
    to the physics world (in the ray direction from the click) will be cast. If this ray hits an object,
    it will call the :ref:`CollisionObject3D._input_event() <class_CollisionObject3D_private_method__input_event>`
-   function in the relevant physics object (bodies receive this callback by default, but areas do
-   not. This can be configured through :ref:`Area3D <class_Area3D>` properties).
+   function in the relevant physics object.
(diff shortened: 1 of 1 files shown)


AThousandShips (approved):
Haven't confirmed but this seems to be a confusion from the 2D case, where it is in reverse, in 3D areas and bodies all have `input_ray_pickable` true by default

mhilbrunner (approved):

mhilbrunner:

mhilbrunner:
Cherry-picked to 4.2.

PR #8623, opened 2023-12-14
Clarify that EditorScript can only run using the internal script editor

AdriaandeJongh:
Unless there's some hidden functionality somewhere, the docs state incorrectly that you can run an C# EditorScript even though there is no File > Run for C# scripts.

Edit: found that both C# and GDScript scripts can run, but only using the internal editor. This PR adds a note about that to the docs.

### DIFF
--- a/tutorials/plugins/running_code_in_the_editor.rst
+++ b/tutorials/plugins/running_code_in_the_editor.rst
@@ -505,8 +505,8 @@ Scripts that extend EditorScript must be ``@tool`` scripts to function.
 
 .. note::
 
-    If you are using an external editor, open the script inside the Godot
-    script editor so you can execute the function.
+    EditorScripts can only be run from the Godot script editor. If you are using
+    an external editor, open the script inside the Godot script editor to run it.
 
 .. danger::
 
--- a/tutorials/plugins/running_code_in_the_editor.rst
+++ b/tutorials/plugins/running_code_in_the_editor.rst
@@ -503,6 +503,11 @@ currently focused on the script editor.
 
 Scripts that extend EditorScript must be ``@tool`` scripts to function.
 
+.. note::
+
+    If you are using an external editor, open the script inside the Godot
+    script editor so you can execute the function.
+
 .. danger::
 
     EditorScripts have no undo/redo functionality, so **make sure to save your


YuriSizov:
It totally works. 

image

What do you mean exactly by "there is no File > Run for C# scripts"? This describes the Godot editor UI. If you use an external editor for coding in C# (which is to be expected, of course), then no, there is probably no way to trigger this from there.

AdriaandeJongh:
I honestly didn't even know you could open C# scripts into the internal script editor. I tried but now I realize I had the editor setting 'Use External Editor' checked so I essentially never saw the internal editor, and there was no entry point for me to run the EditorScript.

I guess the larger issue here is that there is no entry point to run an EditorScript for people using an eternal script editor. And regardless of that UX overlook (good for an issue on the engine itself I guess) this fact is not in the documentation whatsoever. Will think about how to put that in the docs now.

YuriSizov:
This is a problem for GDScript-based EditorScripts as well. I want us to have a way to trigger any of the scripts from anywhere in the editor. This would also automatically resolve your issue, since you wouldn't need to open the script to run it. Although with C# you need to make sure to build the solution first, of course.

tetrapod00:
> This is still a bit of an awkward PR, because the underlying feature is awkward. Should I close this?

The awkwardness of the underlying feature may be improved by https://github.com/godotengine/godot/pull/99318.

In the meantime, I don't mind this PR, though it does now need a rebase.

AdriaandeJongh:
@tetrapod00 rebase done!

tutorials/plugins/running_code_in_the_editor.rst
```diff
@@ -503,6 +503,11 @@ currently focused on the script editor.
 
 Scripts that extend EditorScript must be ``@tool`` scripts to function.
 
+.. note::
+
+    If you are using an external editor, open the script inside the Godot
+    script editor so you can execute the function.
```
tetrapod00, line 509:
```suggestion
    EditorScripts can only be run from the Godot script editor. If you are using
    an external editor, open the script inside the Godot script editor to run it.
```

tetrapod00 (approved):
Seems good to document this limitation, IMO. This will need a second opinion from another maintainer, since as you noted this is a bit of an awkward PR.

I left an optional suggestion to (subjectively) clarify the note a bit.

skyace65 (approved):

skyace65:
Applied tetrapod's suggestion

skyace65:
Thanks!

PR #8654, opened 2023-12-19
Added section on common unsafe operations and their safe counterparts

AdriaandeJongh:
In trying to learn typed GDScript and enabling all the advanced warnings, I found that there was no guidance anywhere in the docs on how to actually fix those warnings. This is a first attempt at that.

### DIFF
--- a/tutorials/scripting/gdscript/static_typing.rst
+++ b/tutorials/scripting/gdscript/static_typing.rst
@@ -396,6 +396,79 @@ if you prefer a more readable and reliable, but more verbose syntax.
 ``UNSAFE_*`` warnings make unsafe operations more noticeable, than unsafe lines.
 Currently, ``UNSAFE_*`` warnings do not cover all cases that unsafe lines cover.
 
+Common unsafe operations and their safe counterparts
+----------------------------------------------------
+
+``UNSAFE_PROPERTY_ACCESS`` and ``UNSAFE_METHOD_ACCESS`` warnings
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In this example, we aim to set a property and call a method on an object
+that has a script attached with ``class_name MyScript`` and that ``extends
+Node2D``. If we have a reference to the object as a ``Node2D`` (for instance,
+as it was passed to us by the physics system), we can first check if the
+property and method exist and then set and call them if they do::
+    
+    if "some_property" in node_2d:
+        node_2d.some_property = 20  # Produces UNSAFE_PROPERTY_ACCESS warning.
+    
+    if node_2d.has_method("some_function"):
+        node_2d.some_function()  # Produces UNSAFE_METHOD_ACCESS warning.
+
+However, this code will produce ``UNSAFE_PROPERTY_ACCESS`` and
+``UNSAFE_METHOD_ACCESS`` warnings as the property and method are not present
+in the referenced type - in this case a ``Node2D``. To make these operations
+safe, you can first check if the object is of type ``MyScript`` using the
+``is`` keyword and then declare a variable with the type ``MyScript`` on
+which you can set its properties and call its methods::
+    
+    if node_2d is MyScript:
+        var my_script: MyScript = node_2d
+        my_script.some_property = 20
+        my_script.some_function()
+
+Alternatively, you can declare a variable and use the ``as`` operator to try
+to cast the object. You'll then want to check whether the cast was successful
+by confirming that the variable was assigned::
+
+    var my_script := node_2d as MyScript
+    if my_script != null:
+        my_script.some_property = 20
+        my_script.some_function()
+
+``UNSAFE_CAST`` warning
+~~~~~~~~~~~~~~~~~~~~~~~
+
+In this example, we would like the label connected to an object entering our
+collision area to show the area's name. Once the object enters the collision
+area, the physics system sends a signal with a ``Node2D`` object, and the most
+straightforward (but not statically typed) solution to do what we want could
+be achieved like this::
+    
+    func _on_body_entered(body: Node2D) -> void:
+        body.label.text = name  # Produces UNSAFE_PROPERTY_ACCESS warning.
+
+This piece of code produces an ``UNSAFE_PROPERTY_ACCESS`` warning because
+``label`` is not defined in ``Node2D``. To solve this, we could first check if the
+``label`` property exist and cast it to type ``Label`` before settings its text
+property like so::
+
+    func _on_body_entered(body: Node2D) -> void:
+        if "label" in body:
+            (body.label as Label).text = name  # Produces UNSAFE_CAST warning.
+
+However, this produces an ``UNSAFE_CAST`` warning because ``body.label`` is of a
+``Variant`` type. To safely get the property in the type you want, you can use the
+``Object.get()`` method which returns the object as a ``Variant`` value or returns
+``null`` if the property doesn't exist. You can then determine whether the
+property contains an object of the right type using the ``is`` keyword, and
+finally declare a statically typed variable with the object::
+
+    func _on_body_entered(body: Node2D) -> void:
+        var label_variant: Variant = body.get("label")
+        if label_variant is Label:
+            var label: Label = label_variant
+            label.text = name
+
 Cases where you can't specify types
 -----------------------------------
 


tutorials/scripting/gdscript/static_typing.rst
```diff
@@ -396,6 +396,38 @@ if you prefer a more readable and reliable, but more verbose syntax.
 ``UNSAFE_*`` warnings make unsafe operations more noticeable, than unsafe lines.
 Currently, ``UNSAFE_*`` warnings do not cover all cases that unsafe lines cover.
 
+Common unsafe operations and their safe counterparts
+----------------------------------------------------
+
+The example below gives an ``UNSAFE_CAST`` warning because ``body.health`` is of
+an unknown type:
```
AThousandShips, line 403:
```suggestion
an unknown type::
```
This is required to make it a block, see below and above

tutorials/scripting/gdscript/static_typing.rst
```diff
@@ -396,6 +396,38 @@ if you prefer a more readable and reliable, but more verbose syntax.
...
+
+    func _on_body_entered(body: Node2D) -> void:
+        if "health" in body:
+            (body.health as Health).take_damage(damage)
+
+This can be made safe using the following syntax, which works because 
+``Object.get()`` returns ``null`` if the property doesn't exist or is of another
+type:
```
AThousandShips, line 411:
```suggestion
type::
```

tutorials/scripting/gdscript/static_typing.rst
```diff
@@ -396,6 +396,38 @@ if you prefer a more readable and reliable, but more verbose syntax.
...
+
+    func _on_body_entered(body: Node2D) -> void:
+	    var health: Health = body.get("health")
+	    if health != null:
+	    	health.take_damage(damage)
+
+In another example below, we call a method that we know is on the object or
+script, but we will still get an ``UNSAFE_METHOD_ACCESS`` warning as the method
+is not present in the inferred type (in this case being a Node2D):
```
AThousandShips, line 420:
```suggestion
is not present in the inferred type (in this case being a Node2D)::
```

tutorials/scripting/gdscript/static_typing.rst
```diff
@@ -396,6 +396,38 @@ if you prefer a more readable and reliable, but more verbose syntax.
...
+
+    if node2d.has_method("some_function"):
+        node2d.some_function()
+
+To make this operation safe, you can use the following syntax:
```
AThousandShips, line 425:
```suggestion
To make this operation safe, you can use the following syntax::
```

tutorials/scripting/gdscript/static_typing.rst
```diff
@@ -396,6 +396,38 @@ if you prefer a more readable and reliable, but more verbose syntax.
...
+
+    var myScript: MyScript = node2d
+    if myScript != null:
+        myScript.some_function()
```
AThousandShips, line 429:
```suggestion
    var my_script: MyScript = node_2d
    if my_script != null:
        my_script.some_function()
```
Use `snake_case` for GDScript

tutorials/scripting/gdscript/static_typing.rst
```diff
@@ -396,6 +396,38 @@ if you prefer a more readable and reliable, but more verbose syntax.
...
+        var health: Health = body.get("health")
+        if health != null:
+            health.take_damage(damage)
+
+In another example below, we call a method that we know is on the object or
+script, but we will still get an ``UNSAFE_METHOD_ACCESS`` warning as the method
+is not present in the inferred type (in this case being a Node2D):
+
+    if node2d.has_method("some_function"):
+        node2d.some_function()
```
AThousandShips, line 423:
```suggestion
    if node_2d.has_method("some_function"):
        node_2d.some_function()
```

tutorials/scripting/gdscript/static_typing.rst
AThousandShips: It's easy with the PascalCase named types 🙂

adamscott:
Nice PR! I would add though some subtitles for each error, making it easy to understand which part applies to which error.

AdriaandeJongh:
> Nice PR! I would add though some subtitles for each error, making it easy to understand which part applies to which error.

Thanks!

I'm new to Sphinx / reST so I don't know if 'subtitles' hints at special reST markup or whether you're using subtitles figuratively ;) Perhaps you could give an example and I'll elaborate on the errors in the PR.

AThousandShips:
Subsections, which are done with `~` I think, see above with the type hints

AdriaandeJongh:
Ah, those kinds of subtitles! Added them.

AdriaandeJongh:
I just pushed a slightly more elaborate description of the examples, and changed the 'fix' of the second example. The old fix would work to get rid of the warning, but would lead to an error at runtime if the object would not be of the type it was casted into.

tutorials/scripting/gdscript/static_typing.rst
```diff
@@ -396,6 +396,53 @@ if you prefer a more readable and reliable, but more verbose syntax.
...
+``take_damage()``. However, this gives an ``UNSAFE_CAST`` warning because
+``body.health`` is of an unknown type::
+
+    func _on_body_entered(body: Node2D) -> void:
+        if "health" in body:
+            (body.health as Health).take_damage(damage)
+
+This operation can be made type safe using the ``Object.get()`` method.
+``body.get("health")`` will return ``null`` if the property doesn't exist or if
+it is of another type::
```
dalexeev, line 418:
```suggestion
``body.get("health")`` will silently return ``null`` if the property doesn't exist::
```

When calling `get()`, you cannot specify what type you are expecting. This is not a special syntax, but a regular method ([`Object.get()`](https://docs.godotengine.org/en/stable/classes/class_object.html#class-object-method-get)).

tutorials/scripting/gdscript/static_typing.rst
```diff
@@ -396,6 +396,53 @@ if you prefer a more readable and reliable, but more verbose syntax.
...
+
+    func _on_body_entered(body: Node2D) -> void:
+        var health: Health = body.get("health")
+        if health != null:
+            health.take_damage(damage)
```
dalexeev, line 423: It is not safe. The safe option is

```gdscript
        var health := body.get("health") as Health
        if health != null:
            health.take_damage(damage)
```

or

```gdscript
        if body.get("health") is Health:
            var health: Health = body.get("health")
            health.take_damage(damage)
```

Note that `body.get("health")` is executed twice, so a more correct option is

```gdscript
        var health_variant: Variant = body.get("health")
        if health_variant is Health:
            var health: Health = health_variant
            health.take_damage(damage)
```

tutorials/scripting/gdscript/static_typing.rst
```diff
@@ -396,6 +396,53 @@ if you prefer a more readable and reliable, but more verbose syntax.
...
+
+Warning: ``UNSAFE_CAST``
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+In the example below, we would like an object entering our collision area to take
+damage if it has a health variable that contains a ``Health`` object. The physics
+system passes us a ``Node2D`` object, and after we check whether that object has
+the health variable, we cast the variable to a ``Health`` object and call
+``take_damage()``. However, this gives an ``UNSAFE_CAST`` warning because
+``body.health`` is of an unknown type::
```
dalexeev, line 410:
```suggestion
``body.health`` is of ``Variant`` type::
```

tutorials/scripting/gdscript/static_typing.rst
```diff
@@ -396,6 +396,53 @@ if you prefer a more readable and reliable, but more verbose syntax.
...
 Currently, ``UNSAFE_*`` warnings do not cover all cases that unsafe lines cover.
 
+Common unsafe operations and their safe counterparts
+----------------------------------------------------
+
+Warning: ``UNSAFE_CAST``
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+In the example below, we would like an object entering our collision area to take
+damage if it has a health variable that contains a ``Health`` object. The physics
```
dalexeev, line 406: I think most often the `health` variable is of type `int` or `float`. Perhaps this example is not very successful in a practical sense. I often cite the `_input()` method and the `InputEvent` and `InputEventMouseMotion` classes as an example of appropriate use of the `as` operator.

tutorials/scripting/gdscript/static_typing.rst
```diff
@@ -396,6 +396,53 @@ if you prefer a more readable and reliable, but more verbose syntax.
...
+reference to the object as a ``Node2D`` (for instance, as it was passed to us by
+the physics system), we can first check if the method exists and then call the
+method if it does. However, we will get an ``UNSAFE_METHOD_ACCESS`` warning 
+as the method is not present in the referenced type (in this case a ``Node2D``)::
+
+    if node_2d.has_method("some_function"):
+        node_2d.some_function()
+
+To make this operation safe, you can first check if the object is of type
+``MyScript``and then declare a variable with the type ``MyScript`` on which you
```
dalexeev, line 439:
```suggestion
``MyScript`` and then declare a variable with the type ``MyScript`` on which you
```

tutorials/scripting/gdscript/static_typing.rst
AdriaandeJongh: Good point; came up with another scenario.

AdriaandeJongh:
Thanks for the review @dalexeev – made all the changes you requested, as well as reordered the two examples, and added ``UNSAFE_PROPERTY_ACCESS`` to the ``UNSAFE_METHOD_ACCESS`` example (because separating them would lead to nearly identical examples, which I don't think would be better? Please advice.)

tutorials/scripting/gdscript/static_typing.rst
```diff
@@ -396,6 +396,71 @@ if you prefer a more readable and reliable, but more verbose syntax.
...
+that has a script attached with ``class_name MyScript`` and that ``extends
+Node2D``. If we have a reference to the object as a ``Node2D`` (for instance,
+as it was passed to us by the physics system), we can first check if the
+property and method exist and then set and call them if they do::
+    
+    if "some_property" in node_2d:
+        node_2d.some_property = 20  # gives UNSAFE_PROPERTY_ACCESS warning
+    
+    if node_2d.has_method("some_function"):
+        node_2d.some_function()  # gives UNSAFE_METHOD_ACCESS warning
```
dalexeev, line 415:
```suggestion
    if "some_property" in node_2d:
        node_2d.some_property = 20 # Produces UNSAFE_PROPERTY_ACCESS warning.
    
    if node_2d.has_method("some_function"):
        node_2d.some_function() # Produces UNSAFE_METHOD_ACCESS warning.
```

tutorials/scripting/gdscript/static_typing.rst
```diff
@@ -396,6 +396,71 @@ if you prefer a more readable and reliable, but more verbose syntax.
...
+
+However, this code will return ``UNSAFE_PROPERTY_ACCESS`` and
```
dalexeev, line 417:
```suggestion
However, this code will produce ``UNSAFE_PROPERTY_ACCESS`` and
```

I'm not sure whether "gives" or "produces" is better, but "returns" looks incorrect to me.

tutorials/scripting/gdscript/static_typing.rst
```diff
@@ -396,6 +396,71 @@ if you prefer a more readable and reliable, but more verbose syntax.
...
+in the referenced type - in this case a ``Node2D``. To make these operations
+safe, you can first check if the object is of type ``MyScript`` using the
+``is`` keyword and then declare a variable with the type ``MyScript`` on
+which you can set its properties and call its methods::
+    
+    if node_2d is MyScript:
+        var my_script: MyScript = node_2d
+        my_script.some_property = 20
+        my_script.some_function()
+
```
dalexeev, line 428:
```suggestion

Alternatively, you can use the ``as`` operator::

    var my_script := node_2d as MyScript
    if my_script:
        my_script.some_property = 20
        my_script.some_function()
```

tutorials/scripting/gdscript/static_typing.rst
```diff
@@ -396,6 +396,71 @@ if you prefer a more readable and reliable, but more verbose syntax.
 ``UNSAFE_*`` warnings make unsafe operations more noticeable, than unsafe lines.
 Currently, ``UNSAFE_*`` warnings do not cover all cases that unsafe lines cover.
 
+Common unsafe operations and their safe counterparts
+----------------------------------------------------
+
+Warnings: ``UNSAFE_PROPERTY_ACCESS`` and ``UNSAFE_METHOD_ACCESS``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
dalexeev, line 403:
```suggestion
``UNSAFE_PROPERTY_ACCESS`` and ``UNSAFE_METHOD_ACCESS`` warnings
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Just a suggestion. I'm not sure what's best, but the colon in the header looks a bit weird to me.

tutorials/scripting/gdscript/static_typing.rst
```diff
@@ -396,6 +396,71 @@ if you prefer a more readable and reliable, but more verbose syntax.
...
+which you can set its properties and call its methods::
+    
+    if node_2d is MyScript:
+        var my_script: MyScript = node_2d
+        my_script.some_property = 20
+        my_script.some_function()
+
+
+Warning: ``UNSAFE_CAST``
+~~~~~~~~~~~~~~~~~~~~~~~~
```
dalexeev, line 431:
```suggestion
``UNSAFE_CAST`` warning
~~~~~~~~~~~~~~~~~~~~~~~
```

tutorials/scripting/gdscript/static_typing.rst
```diff
@@ -396,6 +396,71 @@ if you prefer a more readable and reliable, but more verbose syntax.
...
+
+In this example, we would like the label connected to an object entering our
+collision area to show the area's name. Once the object enters the collision
+area, the physics system sends a signal with a ``Node2D`` object, and the most
+straightforward (but not statically typed) solution to do what we want could
+be achieved like this::
+    
+    func _on_body_entered(body: Node2D) -> void:
+        body.label.text = name  # gives UNSAFE_PROPERTY_ACCESS warning
```
dalexeev, line 440:
```suggestion
        body.label.text = name # Produces UNSAFE_PROPERTY_ACCESS warning.
```

tutorials/scripting/gdscript/static_typing.rst
```diff
@@ -396,6 +396,71 @@ if you prefer a more readable and reliable, but more verbose syntax.
...
+
+This piece of code gives a ``UNSAFE_PROPERTY_ACCESS`` warning because ``label``
+is not defined in ``Node2D``. To solve this, we could first check if the
+``label`` property exist and cast it to type ``Label`` before settings its text
+property like so::
+
+    func _on_body_entered(body: Node2D) -> void:
+        if "label" in body:
+            (body.label as Label).text = name  # gives UNSAFE_CAST warning
```
dalexeev, line 449:
```suggestion
            (body.label as Label).text = name # Produces UNSAFE_CAST warning.
```

tutorials/scripting/gdscript/static_typing.rst
```diff
@@ -396,6 +396,71 @@ if you prefer a more readable and reliable, but more verbose syntax.
...
+
+However, this gives an ``UNSAFE_CAST`` warning because ``body.label`` is of a
+``Variant`` type. This operation can be made type safe using the
+``Object.get()`` method that returns the property as a ``Variant`` (or
+``null`` if the property doesn't exist), using the ``is`` keyword to determine
+whether the property contains an object of the right type, and using a
+variable declaration with a static type::
```
dalexeev, line 456:
```suggestion
``Variant`` type. To safely get a property you can use the ``Object.get()``
method, which silently returns ``null`` if the property doesn't exist.
Then, since the statically method returns a ``Variant`` value, you can check
to determine whether the property contains an object of the right type, and
using a variable declaration with a static type::
```

The original wording contrasts `Variant` and `null`, although `null` is a `Variant` value too. It also seems like `Object.get()` can choose which static type it returns, but the static type is predefined and does not depend on whether the property exists (which is checked at runtime, not at compile time). I hope my suggestion removes potential confusion.

dalexeev (commented):
@AdriaandeJongh Sorry for the delay and thanks for your patience! I found no factual errors, my comments contain only additions and style suggestions. @AThousandShips Could you please check the grammar and wording?

tutorials/scripting/gdscript/static_typing.rst
AdriaandeJongh: I'll included this alternative in a new commit, but I do want to note that this produces a INFERRED_DECLARATION warning. It doesn't feel entirely right to include example solutions which produce warnings ;)

tutorials/scripting/gdscript/static_typing.rst
AdriaandeJongh: Took your suggestion and slightly iterated on it; mostly added `using the ``is`` keyword` as it made the list of steps less vague.

AdriaandeJongh:
> @AdriaandeJongh Sorry for the delay and thanks for your patience! I found no factual errors, my comments contain only additions and style suggestions. @AThousandShips Could you please check the grammar and wording?

No worries about the delay – I presume you're drowned in coding yourself and reviewing PRs and whatnot! I appreciate the thorough review. Latest commit is ready for final grammar / style check by @AThousandShips.

AThousandShips (approved):
Wording and grammar is good

tutorials/scripting/gdscript/static_typing.rst
dalexeev: Fair point, but note that `INFERRED_DECLARATION` is added for those who prefer verbose code. This is a strict mode to the second power, we assume that few people enable the warning.

There is also currently no way to specify a metatype: for example if you preload a script or enum into a constant (use `const` as an analogue of `typedef`/`using` in C++). In this case, you are forced to trigger the `INFERRED_DECLARATION` warning and suppress it with `@warning_ignore`.

tutorials/scripting/gdscript/static_typing.rst
```diff
@@ -396,6 +396,77 @@ if you prefer a more readable and reliable, but more verbose syntax.
...
+    
+    if node_2d is MyScript:
+        var my_script: MyScript = node_2d
+        my_script.some_property = 20
+        my_script.some_function()
+
+Alternatively, you can use the ``as`` operator::
+
+    var my_script := node_2d as MyScript
+    if my_script:
```
AdriaandeJongh, line 432: @dalexeev sorry, last question! In my opinion, `if my_script:` reads as if my_script is a boolean. I know that it actually means `if my_script != null:` so I was thinking to write that out like that in this example. Agreed?

dalexeev: It's a bit complicated since in addition to `null` there is `Object#null` and `Freed Object` (see also godotengine/godot#73896), but I think that comparison with `null` will work here. There is also `is_instance_valid()`. On the other hand, implicit casting to `bool` has the advantage of fewer opcodes.

AdriaandeJongh: Hmm, complicated indeed. Perhaps the most logical way of thinking about this particular case is that `my_script` is explicitly set to `null` if `node_2d` is not of type `MyScript`, and the explicitly verbose `null` check would help illustrate what happens when the cast fails. So even though I agree the fewer opcodes makes sense in most situations, perhaps in this example the slightly more verbose comparison to `null` makes sense in this example? I've pushed a new commit.

dalexeev (approved):
Looks good to me.

mhilbrunner (approved):

mhilbrunner:
Awesome contribution, thanks for taking the time to add these docs. Also thanks everyone for the reviews! Merged. :)

mhilbrunner:
Cherry-picked to 4.2.

PR #8657, opened 2023-12-19
Update export template docs to use Godot 4.x target names

aaronfranke:
The compile template target docs for web, macOS, and iOS were still using the Godot 3.x names (opt/debug/tools) instead of the Godot 4.x names (target_release/target_debug/editor). This PR updates them. I have tested compiling macOS and iOS and tested exporting from macOS to verify those but compiling for web fails for me.


akien-mga (approved):

akien-mga:
Thanks!

mhilbrunner:
Cherry-picked to 4.2.

PR #8660, opened 2023-12-21
Update c_sharp_basics.rst use external editor instruction

fanjin-z:
Update use external editor instruction to reflect actual UI (see screenshot). 

Screenshot 2023-12-20 at 5 43 51 PM


raulsntos:
Thanks for contributing to the .NET documentation. I'm afraid this change is wrong, .NET uses a separate setting for the external editor and the current documentation is correct.

AThousandShips:
Indeed this is not correct

PR #8683, opened 2023-12-27
Update 04.creating_the_enemy.rst

cmacfadyen:
Change "it's" to possessive "its"


skyace65 (changes requested):
You need to change this PR so it's targeting the master branch instead of 4.2. Changes are made to the master branch then cherry picked to ones like 4.2 if applicable.

If you don't know how to change the branch this is targeting we have a guide [here](https://docs.godotengine.org/en/latest/contributing/workflow/pr_workflow.html#rebasing-onto-another-branch).

cmacfadyen:
My apologies. And I see now in master that there is already a commit #59336ef for the same issue so I'll close this.

PR #8684, opened 2023-12-29
class_name should be changed to actual GDScript keyword "class" in the example

the-loan-wolf:


AThousandShips (changes requested):
This is incorrect, this is the global class syntax, not a local class, you are thinking about:
```gdscript
class MyClass:
    ...
```

This is not the same

PR #8685, opened 2023-12-29
changed "class_name" with actual GDScript keyword "class" in the example

the-loan-wolf:
right now example contain "class_name" which create confusion in readers, it should be "class" keyword instead.

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_basics.rst
+++ b/tutorials/scripting/gdscript/gdscript_basics.rst
@@ -37,7 +37,7 @@ here's an example of how GDScript looks.
     @icon("res://path/to/optional/icon.svg")
 
     # (optional) class definition:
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):

mhilbrunner:
Actually, see comments in https://github.com/godotengine/godot-docs/pull/8684.

the-loan-wolf:
Ok, after reading more i think in example "class_name" keyword was used to name that class and add that class to the global scope, that's make sense! and it is also commented that it is optional. example is alright i thought it is wrong because it is using class_name keyword to create a local class. @mhilbrunner you can close this issue now.

Piralein:
The same code example shows the definition of an inner class on line 107.
Closing per above request. Thank you for the contribution nonetheless!

Issue #6372, opened 2022-11-08
Godot's "Your first 2d game" tutorial seems to currently spawn enemies at the wrong locations (not on the real screen borders)


WraithGlade:
**Your Godot version:** 3.5.1 stable win64

**Issue description:** The tutorial for "Your first 2D game" spawns enemies in the wrong locations unless `global_position` is used instead of `position` for the mob spawning locations along the mob path.

I was careful to ensure that my scene was the same as what following along with the tutorial word-for-word would result in. I only followed the tutorial document and didn't use any code from GitHub for it. 

By the end of [the tutorial page](https://docs.godotengine.org/en/stable/getting_started/first_2d_game/05.the_main_game_scene.html) the project spawns all enemies at the upper left corner of the screen as if the border for the `MobPath` were centered on the *origin* instead of the real screen. I checked that my mob path wasn't around the origin. I also attached a sprite of the Godot icon to the `MobSpawnLocation` and saw that it in contrast accounts does appear at the real screen edges, suggesting the problem was in code and not the node data. So, it seems that the tutorial's code actually grabs from the mob path's *local* scene coordinates and not the global coordinates that correspond to the real screen. So, I fixed the problem by simply changing this:

```GDScript
# Set the mob's position to a random location.
mob.position = mob_spawn_location.position
```

... to this:

```GDScript
# Set the mob's position to a random location.
mob.position = mob_spawn_location.global_position
```

I would submit a change normally, but since I'm not experienced with Godot I think it better to submit an issue.

The GitHub page for the tutorial page in question can be found [here](https://github.com/godotengine/godot-docs/blob/master/getting_started/first_2d_game/05.the_main_game_scene.rst).

**URL to the documentation page:** https://docs.godotengine.org/en/stable/getting_started/first_2d_game/05.the_main_game_scene.html

nickcernis:
I ran into this too and submitted a PR at https://github.com/godotengine/godot-docs/pull/10698.

PR #8700, opened 2024-01-01
Update 05.the_main_game_scene.rst for global_position

libroraptor:
Replace 
with

Please check csharp because I don't know it.

### DIFF
--- a/getting_started/first_2d_game/05.the_main_game_scene.rst
+++ b/getting_started/first_2d_game/05.the_main_game_scene.rst
@@ -205,7 +205,7 @@ Note that a new instance must be added to the scene using ``add_child()``.
         var direction = mob_spawn_location.rotation + PI / 2
 
         # Set the mob's position to a random location.
-        mob.position = mob_spawn_location.position
+        mob.position = mob_spawn_location.global_position
 
         # Add some randomness to the direction.
         direction += randf_range(-PI / 4, PI / 4)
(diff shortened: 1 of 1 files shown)


AThousandShips:
Why is this requried? The position here should be in local space as the `mob_spawn_location` is *also* in the same space, under the main scene, further this doesn't match the demo project which works correctly, so I do not think this is valid

Please explain why this should be done

libroraptor:
This change (or something else) is required because it doesn’t work in the tutorial – the effective path in local space functions as if it is only half as big as the screen, resulting in mobs spawning along the edges of the upper left quadrant, not the edges of the whole screen.

Maybe there a change needed in how the path is defined? Or to ensure that it points to the correct parent?

I am a novice; I just came across this problem while working through the tutorial so don’t know what is really wrong. Just posted this modification (which works) in order to start conversation about where the actual problem is.

libroraptor:
I found this:
https://github.com/godotengine/godot-docs/issues/6372 <https://github.com/godotengine/godot-docs/issues/6372>

and I wonder whether the actual problem is that the path’s datum point needs to be set as a tutorial step.

AThousandShips:
I tested the demo project, which is identical to this, and the code works perfectly as it is without this change, this change is not needed

This is not a valid change, you must have missed some detail in the tutorial as the code is fully valid

Closing as this change isn't required and isn't valid

I'd suggest asking in the other [community channels](https://godotengine.org/community) for help with your problem, chances are you missed some detail when applying this tutorial, check out the [demo code](https://github.com/godotengine/godot-demo-projects/tree/master/2d/dodge_the_creeps) and compare your code

libroraptor:
I have figured out the problem. You are right: it isn’t the code that needs to change. 

The problem is that the MobPath.Position was not (0,0), but at approximately mid-screen. 

Thus the MobPath looks to be in the right place in the construction interface but gets displaced during build to be centred on the upper left corner.

So I suggest considering an extra line to the tutorial for drawing the MobPath: 

Ensure that the path’s position is (0, 0). If not, set it to (0, 0) and move the MobPath nodes so that they are close to the corners of the display boundary.

PR #8721, opened 2024-01-05
Update "The Main Game Scene" for clarity

FishingSimulator:
Many people have reported having issues connecting HUD's start_game and the main node's new_game() in the next step of this tutorial. This occurred for me because the documentation states that "new_game()" should be removed from the _ready() function, but does not include instructions to add "pass" back (and pass is not left in the example code when "new_game()" is added to the function, so the initial one will not remain). Having the function declaration alone without "pass" breaks all the following functions, preventing new_game() (or any Main functions) from being connected.


mhilbrunner (approved):

mhilbrunner:
Yeah, this is a good call, this indeed seems to be easy confusion for beginners. Thanks for pointing it out and taking the time to fix it!

mhilbrunner:
Cherry-picked to 4.2.

Issue #8729, opened 2024-01-06
Lambda documentation does not mention "return" is required


Marenz:
Version: All

The documentation for lambda functions does not mention that to return a value you need to explicitly write `return` before the expression.
All of the examples use `print()` which also does not demonstrate that need.

https://github.com/godotengine/godot-docs/blob/master/tutorials/scripting/gdscript/gdscript_basics.rst#lambda-functions
https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_basics.html#lambda-functions

AThousandShips:
This is only a confusing thing if you expect it to work differently than normal methods

Marenz:
I think we argued enough in the PR, so I only want to add my user experience here so one can see where I come from:

First of, I have +1y experience with python and +10y with D/C++ and similar.

In my project, I had the line

```
winners.sort_custom(func (a, b): times[a]  < times[b]) # missing return
```

It did trigger a warning (`Standalone expression (the line has no effect).`) yes.
But I thought it was referencing the `sort_custom` function, though the documentation of that said it's `void` so I was confused
and assumed it's maybe a bug in the metadata of the function that the interpreter internally uses, things _seemed_ to work more or less as they should after all.

I read the documentation of the lambdas, nowhere was a `return` demonstrated or mentioned, so I was no wiser than before.

Only after I asked in the chat someone pointed the right thing out to me.
We could have saved both, the helpful chat persons and myself a lot of time, if this information was easily found.

More importantly, it would be a much improved user experience and in the end, isn't that what counts?
I believe that both the scenario where this happened (a lambda embedded in a call) as well as the target crowd (a person knowing python but being new to gdscript) are not rare and thus this would make an improved user experience for a non-trivial amount of people.

AThousandShips:
I was making my opinion clear here, that's what this is for 🙂 I didn't say it shouldn't be done

PR #8730, opened 2024-01-06
Lambdas need an explicit `return`

Marenz:
Extend the documentation so that it is clear that lambdas require an explicit return . 

Also add an according example.

fixes: https://github.com/godotengine/godot-docs/issues/8729

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_basics.rst
+++ b/tutorials/scripting/gdscript/gdscript_basics.rst
@@ -1347,6 +1347,12 @@ Lambda functions can be named for debugging purposes::
     var lambda = func my_lambda(x):
         print(x)
 
+Note that if you want to return a value from a lambda, an explicit ``return``
+is required (you can't omit ``return``)::
+
+    var lambda = func(x): return x ** 2
+    print(lambda.call(2)) # Prints `4`.
+
 Lambda functions capture the local environment. Local variables are passed by value, so they won't be updated in the lambda if changed in the local function::
 
     var x = 42


tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1342,6 +1342,13 @@ This can be useful to create Callables to pass around without polluting the clas
     var lambda = func(x): print(x)
     lambda.call(42) # Prints "42"
 
+Unlike python, Lambda functions require an explicit `return`.
```
AThousandShips, line 1345: Shouldn't reference python, GDScript isn't related to python and we shouldn't expect any knowledge like that

This should be expected as nowhere else do we have implicit return, so I don't think this is something to document

Marenz: I searched the document before deciding to mention python and found another part doing the same, so I thought it would be okay. 
It adds a valuable reference especially for people coming from python.
I would have never found out that they work this way if I hadn't ask in the chat about it.

At the very least an example with `return` could be provided.

AThousandShips: Then it's okay, didn't know there were, would need to be discussed still, as it's a very unusual assumption based on knowledge from another language

AThousandShips:
```suggestion
Unlike Python, lambda functions require an explicit `return`.
```
Name, capitalize, which is done elsewhere, also lambda isn't a name

Marenz: To be honest, I think it is very natural to use knowledge from the closest language to gdscript that is out there.
No matter how often it is written that it's independent, people with according experience will at first assume it is like a python with less features (which it is to a large degree).

If, even after explicitly reading the lambda documentation, one is not made aware of this little difference, then the documentation is lacking imo.

AThousandShips: IMO that's not how to approach new situations, but I'm fine with adding this, but we can't necessarily point out every single way GDScript differs from Python, because it isn't Python, it only similar because it uses indentation, and the documentation clearly states it has no relation to it

Marenz: Mind you, even the [style guide](https://github.com/godotengine/godot-docs/blob/3107a52ce7a4a07e5dac5a8e74f1930f02b98a5c/tutorials/scripting/gdscript/gdscript_styleguide.rst#gdscript-style-guide) references python in the very first paragraph (saying it's close), as well as this document itself (if only to mention the indentation). 

It is put in peoples heads, whether you like it or not..

AThousandShips: As I said:
> GDScript is entirely independent from Python and is not based on it

But I'm fine with adding this, but I don't agree with the fundamental logic, but it's clearly stated to not be at all connected in that sense 🙂

Marenz: > but please fix the spelling issues

Yah, I was struggling with the desktop github client (don't often use it), it seems to be lacking the "amend" option, or its well hidden.

> As I said:
> GDScript is entirely independent from Python and is not based on it

I am not refuting this, but you have to also see that there clearly is a relation at it would be naive to not expect people to behave accordingly :) 
I am trying a pragmatic approach really. In my mind, it is rather likely that a new comer to godot (but not to programming) has previous python experience and will draw from it for this.
Maybe I am too biased from my own experience, but the smallest common denominator here would then be to have at least the example.

Let's see what others say or decide from here.

AThousandShips: Note though that this would require people to ignore warnings, so I'm not sure this is a critical thing to point out, it tells you you're doing something wrong if you do like that code

Marenz: as you mentioned warnings..

> Though I'm against it also because it causes errors and warnings, so it's clear you're doing something wrong

I did have a warning, true, but it was hard to understand because I thought it referenced the outer function

> winners.sort_custom(func (a, b): times[a]  < times[b])

so I thought it was an issue with `sort_custom` and left me more confused than helpful. 
My first assumption was that `sort_custom` was somehow wrongly tagged as "a function that returns something when in fact it doesn't" than the lambda being wrong..

And given that lambdas are usually embedded in other calls, it's probably not a rare scenario that way...

tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1342,6 +1342,13 @@ This can be useful to create Callables to pass around without polluting the clas
     var lambda = func(x): print(x)
     lambda.call(42) # Prints "42"
 
+Unlike Python, lambda functions require an explicit `return`.
+
+::
+
+    var lambda = func(x): return x*x
+    print(lambda.call(2)) # prints "4"
```
dalexeev, line 1350:
```suggestion
Note that if you want to return a value from a lambda, an explicit ``return``
is required (you can't omit ``return``)::

    var lambda = func(x): return x ** 2
    print(lambda.call(2)) # Prints `4`.
```

1. Note that lambdas are not required to return a value. I agree with @AThousandShips that the comparison with Python is unnecessary here, since Python uses a special syntax for lambdas, completely different from GDScript.
2. I would also move this paragraph below, after the note about named lambdas.
3. See also godotengine/godot-proposals#5925.

Marenz:
Used your suggestion and moved it down

dalexeev (approved):

mhilbrunner (approved):

mhilbrunner:

mhilbrunner:
Cherry-picked to 4.2.

PR #8711, opened 2024-01-03
Fix export annotation being formatted as Godot 3.

IntangibleMatter:
While browsing the docs, I noticed that there was a bit of code obviously misformatted for Godot 4. Fixed it, hopefully.

PR #8732, opened 2024-01-07
Add c# script name note to scripting_first_script.rst

a-johnston:
Adds a short note regarding C# script naming. This is a somewhat common point of confusion for newer users. Please let me know if I should update the wording here to be clearer/more precise.

### DIFF
--- a/getting_started/step_by_step/scripting_first_script.rst
+++ b/getting_started/step_by_step/scripting_first_script.rst
@@ -85,6 +85,11 @@ other options set to their default values and click the Create button to create 
 
 .. image:: img/scripting_first_script_attach_node_script.webp
 
+.. note::
+
+    C# script names need to match their class name. In this case, you should name the
+    file ``MySprite2D.cs``.
+
 The Script workspace should appear with your new ``sprite_2d.gd`` file open and
 the following line of code:
 


raulsntos:
There's a note about this in the [_C# basics_](https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_basics.html#example) documentation page:

image

This page is already linked at the beginning of the _Creating your first script_ documentation page. Users that intend to use C# with Godot should read it.

https://github.com/godotengine/godot-docs/blob/d4bff01d23a09748862c847d8174a2ac9677c94d/getting_started/step_by_step/scripting_first_script.rst?plain=1#L31

Maybe the note should be reworded so it's a bit more _aggresive_?

---

Either way I think adding a note in the step where you add your first script in the _Getting Started_ tutorial makes sense, because I think it's meant to be very hand-holdy.

a-johnston:
I agree with your comments and honestly I think the current error is pretty clear, but yeah still a common issue I've seen. I think in some cases users get to this page after being linked by [Using signals](https://docs.godotengine.org/en/stable/getting_started/step_by_step/signals.html#scene-setup) and don't think to follow other links, or don't think they have anything to gain by going over the basics.

skyace65 (approved):

skyace65:
Thanks!

a-johnston:
Sorry if this isn't the best place to comment (is there an official place to discuss the docs repo?) but I noticed that many docs PRs marked as cherrypicked for a certain version still only show up under `latest` on the docs pages (for example https://github.com/godotengine/godot-docs/pull/8711). This seems like it potentially could be an issue in how the build actions are configured? I'm guessing most users don't bother changing to `latest` and as a result might be missing out on a lot of corrections/changes.

Piralein:
@a-johnston PR's with the cherry-pick label are still to be cherry-picked into the branch the label mentions. This is done by a maintainer, not an deploy/github action. The label will be removed after it was cherry-picked. The linked PR will show up in the `stable` and `4.2` version of the docs, after it was picked.

mhilbrunner:
Cherry-picked to 4.2.

PR #8735, opened 2024-01-07
Fix method bind instructions and style

AThousandShips:
This file needs some more general cleaning up but leaving that for later

* Closes: https://github.com/godotengine/godot-cpp/issues/1352

### DIFF
--- a/contributing/development/core_and_modules/object_class.rst
+++ b/contributing/development/core_and_modules/object_class.rst
@@ -80,7 +80,7 @@ Default values for arguments can be passed as parameters at the end:
     ClassDB::bind_method(D_METHOD("methodname", "arg1name", "arg2name", "arg3name"), &MyCustomType::method, DEFVAL(-1), DEFVAL(-2)); // Default values for arg2name (-1) and arg3name (-2).
 
 Default values must be provided in the same order as they are declared,
-skipping requried arguments and then providing default values for the optional ones.
+skipping required arguments and then providing default values for the optional ones.
 This matches the syntax for declaring methods in C++.
 
 ``D_METHOD`` is a macro that converts "methodname" to a StringName for more
--- a/contributing/development/core_and_modules/object_class.rst
+++ b/contributing/development/core_and_modules/object_class.rst
@@ -71,13 +71,17 @@ Registering functions is one:
 
 .. code-block:: cpp
 
-    ClassDB::bind_method(D_METHOD("methodname", "arg1name", "arg2name"), &MyCustomMethod);
+    ClassDB::bind_method(D_METHOD("methodname", "arg1name", "arg2name", "arg3name"), &MyCustomType::method);
 
-Default values for arguments can be passed in reverse order:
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):

contributing/development/core_and_modules/object_class.rst
```diff
@@ -71,13 +71,13 @@ Registering functions is one:
 
 .. code-block:: cpp
 
-    ClassDB::bind_method(D_METHOD("methodname", "arg1name", "arg2name"), &MyCustomMethod);
+    ClassDB::bind_method(D_METHOD("methodname", "arg1name", "arg2name", "arg3name"), &MyCustomType::method);
 
-Default values for arguments can be passed in reverse order:
+Default values for arguments can be passed:
```
akien-mga, line 76: What "reverse order" tried to convey is that it applies to the X last arguments matching the X default values provided. One could expect otherwise that it would set default values for `arg1name` and `arg2name` below, and that `arg3name` would be `null`, or that it would fail compiling.

I think we might still want to try to make this clear here. But I'm not sure how to phrase it (reverse order was the wrong term for sure).

YuriSizov: Also ending the sentence like that is abrupt and doesn't work without "in reverse order" :)

We can adjust this sentence to say something like "Default values for arguments can also be passed as parameters at the end:". And after the sample we could add an explanation about the order more verbosely. Something like...

"Default values must be supplied in the same order as function's declared arguments are. You must skip all required arguments, and then provide default values for the optional ones."

AThousandShips: Updated, see what you think, the last part comparing to c++ is optional but thought it would be helpful

contributing/development/core_and_modules/object_class.rst
```diff
@@ -71,13 +71,17 @@ Registering functions is one:
...
-Default values for arguments can be passed in reverse order:
+Default values for arguments can be passed as parameters at the end:
 
 .. code-block:: cpp
 
-    ClassDB::bind_method(D_METHOD("methodname", "arg1name", "arg2name"), &MyCustomType::method, DEFVAL(-1)); // default value for arg2name
+    ClassDB::bind_method(D_METHOD("methodname", "arg1name", "arg2name", "arg3name"), &MyCustomType::method, DEFVAL(-1), DEFVAL(-2)); // Default values for arg2name (-1) and arg3name (-2).
+
+Default values must be provided in the same order as they are declared,
+skipping requried arguments and then providing default values for the optional ones.
```
akien-mga, line 83:
```suggestion
skipping required arguments and then providing default values for the optional ones.
```

akien-mga (approved):

mhilbrunner:
Thanks! Merged. :)

AThousandShips:
Thank you!

mhilbrunner:
Cherry-picked to 4.2.

PR #8736, opened 2024-01-07
Update list of advantages and disadvantages to more accurately refle…

Emynator:
Updated list of advantages and disadvantages of dynamic typed languages to be more accurate to real world software developing. Saying "Less code written means less errors & mistakes to fix." is an extremely misleading phrasing considering the amount of developer time that went to fix typing related errors and bugs just in the javascript ecosystem.

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_advanced.rst
+++ b/tutorials/scripting/gdscript/gdscript_advanced.rst
@@ -32,7 +32,6 @@ are that:
 While the main disadvantages are:
 
 -  Less performance than statically typed languages.
--  Prone to type errors which can cause bugs that are hard to trace and fix.
 -  More difficult to refactor (symbols can't be traced).
 -  Some errors that would typically be detected at compile time in
    statically typed languages only appear while running the code
--- a/tutorials/scripting/gdscript/gdscript_advanced.rst
+++ b/tutorials/scripting/gdscript/gdscript_advanced.rst
@@ -24,7 +24,6 @@ are that:
 
 -  The language is easy to get started with.
 -  Most code can be written and changed quickly and without hassle.
--  Less code written means less errors & mistakes to fix.
 -  The code is easy to read (little clutter).
 -  No compilation is required to test.
 -  Runtime is tiny.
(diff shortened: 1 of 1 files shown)


tutorials/scripting/gdscript/gdscript_advanced.rst
```diff
@@ -33,6 +32,8 @@ are that:
 While the main disadvantages are:
 
 -  Less performance than statically typed languages.
+-  Prone to type errors which can cause bugs that are hard
+   to trace and fix.
```
AThousandShips, line 36:
```suggestion
-  Prone to type errors which can cause bugs that are hard to trace and fix.
```
Keep to one line, easier to read and no wider than other lines here

MarcoN96: done

skyace65 (approved):

mhilbrunner (changes requested):
I think the newly added disadvantage is already covered by the point two lines below about errors that are only detected at runtime instead of compile time. Even more, now with optional typing, even that argument/downside has become a bit weaker.

I'm fine with the removed point, I think that claim is indeed a bit weak.

I think this whole page is in need of an update, it has become a bit outdated. I also don't think we should spend our resources documenting "general knowledge on dynamically typed languages" - there are other people and places better suited for that, so this page should be refocused on GDScript specifically, IMO.

MarcoN96:
@mhilbrunner I agree with your point that the godot documentation should be focused on godot and GDScript instead of "general knowledge on dynamically typed languages".

I disagree that this disadvantage is covered by the point two lines below it though. while technically the point is included in this statement imo it's an important enough point to be listed on it's own. especially because typing is optional in GDScript having that point mentioned extra highlights the importance of it.

I would also say that "that claim is indeed a bit weak" is an understatement.

tutorials/scripting/gdscript/gdscript_advanced.rst
```diff
@@ -33,6 +32,7 @@ are that:
 While the main disadvantages are:
 
 -  Less performance than statically typed languages.
+-  Prone to type errors which can cause bugs that are hard to trace and fix.
 -  More difficult to refactor (symbols can't be traced).
```
skyace65, line 36:
```suggestion
-  Less performance than statically typed languages.
-  More difficult to refactor (symbols can't be traced).
```

skyace65 (changes requested):

skyace65:
Given the lack of others weighing in with their opinion in favor or against, and how long this PR has been open, I've removed the item added to the list of disadvantages so this can be merged.

Emynator thank you for contributing, and congrats on your first merged PR!

Emynator:
@skyace65 thank you very much! I honestly forgot about this PR after all that time

PR #8741, opened 2024-01-08
Update viewports.rst

MichaelMacha:
Changed the order in which the Cameras section is introduced, and added necessary information to the provided example. The behavior only occurs as specified when SubViewport has its own World3D, or a World3D assigned to it; it does not exclude rendering of MeshB from the root Viewport if it does not.

### DIFF
--- a/tutorials/rendering/viewports.rst
+++ b/tutorials/rendering/viewports.rst
@@ -56,35 +56,6 @@ audible, the :ref:`Viewport <class_Viewport>` needs to be enabled as a listener 
 If you are using a :ref:`SubViewport <class_SubViewport>` to display your :ref:`World3D <class_World3D>` or
 :ref:`World2D <class_World2D>`, don't forget to enable this!
 
-Cameras (2D & 3D)
------------------
-
-When using a :ref:`Camera3D <class_Camera3D>` or
-:ref:`Camera2D <class_Camera2D>`, it will always display on the
-closest parent :ref:`Viewport <class_Viewport>` (going towards the root). For example, in the
-following hierarchy:
-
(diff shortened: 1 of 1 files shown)


tutorials/rendering/viewports.rst
```diff
@@ -130,6 +101,45 @@ is possible to do so by setting :ref:`world_2d<class_Viewport_property_world_2d>
...
+
+``CameraA`` will display on the Root :ref:`Viewport <class_Viewport>` and it will draw ``MeshA``. ``CameraB``
+will be captured by the :ref:`SubViewport <class_SubViewport>` along with ``MeshB``. Even though ``MeshB`` is in the scene
+hierarchy, it will still not be drawn to the Root Viewport. Similarly, ``MeshA`` will not
+be visible from the SubViewport because SubViewports and their :ref:`World3D <class_World3D>` only
+capture nodes which are their children in the graph.
+
+.. note::
+
+    This behavior only occurs when the SubViewport has its own :ref:`World3D <class_World3D>` to guide its rendering. Otherwise, ``MeshB`` will continue up the graph, and be rendered on the first viewport which does have an assigned :ref:`World3D <class_World3D>`; in this case, the Root :ref:`Viewport <class_Viewport>`.
```
AThousandShips, line 126:
```suggestion
    This behavior only occurs when the SubViewport has its own :ref:`World3D <class_World3D>` to guide its rendering. Otherwise, ``MeshB`` will continue up the graph,
    and be rendered on the first viewport which does have an assigned :ref:`World3D <class_World3D>`; in this case, the Root :ref:`Viewport <class_Viewport>`.
```
Keep the lines around the same length

PR #8743, opened 2024-01-09
Rename SCons command line options for CCFLAGS, etc. to lowercase

akien-mga:
Matching the changes in https://github.com/godotengine/godot/pull/86964.

Shouldn't be cherry-picked, depends on the above PR.

---

Since it's the only mention in the docs, and it's for a very niche use case, I chose not to add a prominent warnings about the change of case in 4.3.


skyace65 (approved):

skyace65:
Thanks!

PR #8749, opened 2024-01-11
Update thread_safe_apis.rst to mention don't call certain rendering functions on other threads

jsjtxietian:
For https://github.com/godotengine/godot/issues/86796

But I bet my writing is not good, feel free to correct me if any.

### DIFF
--- a/tutorials/performance/thread_safe_apis.rst
+++ b/tutorials/performance/thread_safe_apis.rst
@@ -61,7 +61,7 @@ Note that the Multi-Threaded thread model has several known bugs, so it may not 
 in all scenarios.
 
 You should avoid calling functions involving direct interaction with the GPU on other threads, such as creating new textures
-or modifying and retrieving image data, these can lead to performance stalls because these operations require synchronization 
+or modifying and retrieving image data, these operations can lead to performance stalls because they require synchronization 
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):

tutorials/performance/thread_safe_apis.rst
```diff
@@ -60,6 +60,10 @@ To make rendering thread-safe, set the **Rendering > Driver > Thread Model** pro
 Note that the Multi-Threaded thread model has several known bugs, so it may not be usable
 in all scenarios.
 
+You should avoid calling functions involving direct interaction with the GPU on other threads, such as creating new textures
+or modifying and retrieving image data, these can lead to performance stalls because these operations require synchronization
```
AThousandShips, line 64:
```suggestion
or modifying and retrieving image data, these operations can lead to performance stalls because they require synchronization 
```

AThousandShips (approved):
Looks good otherwise, just think this makes the sentence flow better

mhilbrunner:

DarioSamo:
We may want to keep an eye out for this doc change because the work I'm doing for transfer queues means this will no longer hold true, as texture and buffer creation will be thread-safe and not imply synchronization.

mhilbrunner:
Cherry-picked to 4.2.

PR #8756, opened 2024-01-12
Fix properties' names stretched on mobile

Cronos87:
This is my first contribution to Godot, I hope I did things well 😊

Fix issue https://github.com/godotengine/godot-docs/issues/7504 where properties' names are stretched on mobile.

Before:

CleanShot 2024-01-12 at 12 18 53@2x

After:

CleanShot 2024-01-12 at 12 19 09@2x

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -766,6 +766,11 @@ html.writer-html5 .rst-content table.docutils th {
 .wy-table-responsive table.wrap-normal th {
     white-space: normal;
 }
+/* Properties' name must not be streched */
+.classref-reftable-group#properties td:nth-child(2),
+.classref-reftable-group#properties th:nth-child(2) {
+    white-space: nowrap;
+}
 
 /* Make sure line blocks don't stretch tables */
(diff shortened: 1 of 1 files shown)


AThousandShips:
See also:
* https://github.com/godotengine/godot-docs/pull/7542

Which already works on this issue 🙂

Cronos87:
Yes indeed, I didn't know about this issue. It could be deployed as a temporary fix until this one fixes it better 🤔

AThousandShips:
It was linked to the relevant issue, good to know in the future that there's a list of linked PRs on issues that says they are there

Temporary fixes are usually not supported, but we'll see which solution ends up being the correct one 🙂

YuriSizov:
> It could be deployed as a temporary fix until this one fixes it better 🤔

I think your PR doesn't help with usability, so fixing it even temporarily this way would just exchange one inconvenience for another. We really need to customize the layout itself on mobile, which https://github.com/godotengine/godot-docs/pull/7542 attempts to do.

Any help with testing and finishing it would be appreciated!

PR #8777, opened 2024-01-15
Remove Giscus from search page

adamscott:
This moves the Giscus insertion logic inside custom.js.

Technically, it could have worked if I would have done this instead:
```
{% if (not meta or meta.get('allow_comments') != 'False') and godot_show_article_comments and pagename != 'search' %}
```
... but it's easier to maintain my way, as we can add easily new pages to ban.

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -1763,3 +1763,8 @@ p + .classref-constant {
         padding: 3px 5px;
     }
 }
+
+/* Giscus */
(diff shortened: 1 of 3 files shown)


mhilbrunner (approved):

mhilbrunner:
cc @Calinou

Calinou (approved):
Tested locally, it works as expected.

PR #8795, opened 2024-01-18
Document `use_debanding` render mode in Spatial shader and Sky shader

Calinou:


Calinou:
Closing, as https://github.com/godotengine/godot/pull/87350 was reworked not to use a `use_debanding` render mode anymore (it's either enabled on all materials, or none at all).

PR #8805, opened 2024-01-21
Update compiling_with_dotnet.rst: clarify NuGet source creation

31:
Someone building on Windows and unfamiliar with Linuxy paths, in particular `~`, will try to run `dotnet nuget add source ~/MyLocalNugetSource --name MyLocalNugetSource` verbatim, get an error having to do with `~/MyLocalNugetSource`, and not know where to go from there. (This has brought a few people to the C# Discord help channel.)

I think this page deserves to be friendly even though it's focused on engine devs: people come here without being engine devs to e.g. compile the engine with double precision.

I added a little bit of info on how/where to set up the dir depending on OS, and used `<my_local_source>` to refer to the dir. (Similar to using `<godot_binary>` earlier in the doc.)

The paragraph starting with `The Godot NuGet packages must be added to that local source.` seemed to me like it could be confused for something the doc is instructing the dev to do right then and there, and it seems redundant vs. the explanation that's already there a few paragraphs later, so I tried removing it. Maybe the doc isn't as good at convincing the dev to use the `--push-nupkgs-local` arg, but I'm not sure it's necessary to convince them. I like the flow after the removal.

### DIFF
--- a/contributing/development/compiling/compiling_with_dotnet.rst
+++ b/contributing/development/compiling/compiling_with_dotnet.rst
@@ -100,23 +100,29 @@ distributed as NuGet packages. This is all transparent to the user, but it can
 make things complicated during development.
 
 In order to use Godot with a development version of those packages, a local
-NuGet source must be created where MSBuild can find them. This can be done with
-the .NET CLI:
+NuGet source must be created where MSBuild can find them.
+
+First, pick a location for the local NuGet source. If you don't have a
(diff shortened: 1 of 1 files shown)


raulsntos (approved):
Looks good, thank you!

skyace65 (approved):

skyace65:
Thanks!

mhilbrunner:
Cherry-picked to 4.2.

PR #8812, opened 2024-01-22
Elaborate on C# Connect/Disconnect

31:
The C# signal page mentions the `Connect` and `Disconnect` APIs (defined on `GodotObject`), but not when you might want to use them, or how. When looking at the class doc, there's technically an example in there, but it seems limited by only translating the code examples rather than having a C#-specific description. It's never mentioned why you might still want/need to use these. I added some examples and some details about this.

This also clarifies the note about https://github.com/godotengine/godot/issues/70414. Right now it doesn't link to the active issue, so I added a link, and it also doesn't mention that using `.Connect` is another way around it. Depending on how much a particular dev hates manual cleanup, this might be significantly more appealing.

I imagine the note might have been kept brief rather than mentioning more about it because this was imagined at the time to perhaps be a short-lived problem. But there doesn't seem to me to be a clear, consensus path forward at https://github.com/godotengine/godot/pull/73730, so I think it's worth the detail.

/cc @paulloz (https://github.com/godotengine/godot-docs/pull/8538)

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_signals.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_signals.rst
@@ -6,8 +6,14 @@ C# signals
 For a detailed explanation of signals in general, see the :ref:`doc_signals` section in the step
 by step tutorial.
 
-While it is still possible to use signals through the ``Connect``/``Disconnect`` API, C# gives us
-a more idiomatic way to implement the :ref:`observer pattern<doc_key_concepts_signals>`.
+Signals are implemented using C# events, the idiomatic way to represent
+:ref:`the observer pattern<doc_key_concepts_signals>` in C#. This is the
+recommended way to use signals in C# and the focus of this page.
+
+In some cases it's necessary to use the older
+:ref:`Connect()<class_object_method_connect>` and
+:ref:`Disconnect()<class_object_method_disconnect>` APIs.
+See :ref:`using_connect_and_disconnect` for more details.
 
 Signals as C# events
 --------------------
@@ -30,8 +36,13 @@ In addition, you can always access signal names associated with a node type thro
 .. warning::
 
     While all engine signals connected as events are automatically disconnected when nodes are freed, custom
-    signals aren't. Meaning that: you will need to manually disconnect (using ``-=``) all the custom signals you
-    connected as C# events (using ``+=``).
+    signals connected using ``+=`` aren't. Meaning that: you will need to manually disconnect (using ``-=``)
+    all the custom signals you connected as C# events (using ``+=``).
+
+    An alternative to manually disconnecting using ``-=`` is to
+    :ref:`use Connect <using_connect_and_disconnect>` rather than ``+=``.
+
+    See `Godot issue #70414 <https://github.com/godotengine/godot/issues/70414>`_.
 
 Custom signals as C# events
 ---------------------------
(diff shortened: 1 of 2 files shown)


tutorials/scripting/c_sharp/c_sharp_signals.rst
```diff
@@ -30,8 +36,13 @@ In addition, you can always access signal names associated with a node type thro
 .. warning::
 
     While all engine signals connected as events are automatically disconnected when nodes are freed, custom
-    signals aren't. Meaning that: you will need to manually disconnect (using ``-=``) all the custom signals you
-    connected as C# events (using ``+=``).
+    signals connected using ``+=`` aren't. Meaning that: you will need to manually disconnect (using ``-=``)
```
paulloz, line 39: The "as events" on the line above already implies the "connected using `+=`" for me, that's why it was worded like this.

AThousandShips, line 39:
```suggestion
    signals connected using ``+=`` aren't. Meaning that you will need to manually disconnect (using ``-=``)
```

tutorials/scripting/c_sharp/c_sharp_signals.rst
```diff
@@ -30,8 +36,13 @@ In addition, you can always access signal names associated with a node type thro
...
+    all the custom signals you connected as C# events (using ``+=``).
+
+    An alternative to manually disconnecting using ``-=`` is to
+    :ref:`use Connect <using_connect_and_disconnect>` rather than ``+=``.
+
+    See `Godot issue #70414 <https://github.com/godotengine/godot/issues/70414>`_.
```
paulloz, line 45: When I did the other PR, we elected to not mention/link bug issues in the docs.

tutorials/scripting/cross_language_scripting.rst
```diff
@@ -37,6 +42,8 @@ The following two scripts will be used as references throughout this page.
     {
         public string myField = "bar";
 
+        [Signal] public delegate void MySignal();
```
paulloz, line 45: Signal delegates need to end with `EventHandler`.
```suggestion
        [Signal] public delegate void MySignalEventHandler();
```

tutorials/scripting/c_sharp/c_sharp_signals.rst
31: I think this is ambiguous:

> While all engine signals connected as events are automatically disconnected when nodes are freed, custom signals aren't.

While all (engine signals connected as events) are automatically disconnected [...], (custom signals) aren't.
vs.
While all (engine signals) (connected as events) are automatically disconnected [...], (custom signals) (connected as events [implied]) aren't.

I don't know if that's enough to convey what I mean, but I didn't even consider the second interpretation--I thought this line was entirely mistaken when I first read it. 😄

paulloz: Yeah, the whole sentence is too complicated for me either way, so I don't really have a strong opinion. I hate we don't have a simple way to convey `+=`.

31: To me, without the `:`, it's a conversational sentence fragment, but it doesn't read all that well with the `:`. Maybe `Meaning:` would sound ok to me. I think `This means` is safe/natural/boring (and pushed that for now), but I'm open to whatever.

31: I'll take another look later with fresh eyes... yeah, pulling apart the complexity a bit is probably the right move.

And yeah, this is all already a nightmare in casual convos, getting it right in a doc seems like it may be tough. 😅

raulsntos (approved):
As you say I did imagine it wouldn't take too long to fix and didn't want to discourage users from the event syntax. Unfortunately, there's problems with the proposed fix (https://github.com/godotengine/godot/pull/73730) so it may still take a while. Therefore, I'm more than willing to correct the documentation now (as it should've been done long ago).

Thank you for contributing this PR, it looks good to me so I'll approve but let's wait for @paulloz approval too.

paulloz:
It’s fine by me. And (re)documenting the calls to `Connect` and `Disconnect` is a good thing.

>

skyace65 (approved):

skyace65:
Thanks!

tutorials/scripting/c_sharp/c_sharp_signals.rst
31: After playing with the behavior a bit, I found a whole new chapter to this about capturing variables, so ended up writing a decent amount at https://github.com/godotengine/godot-docs/pull/8815.

mhilbrunner:

PR #8813, opened 2024-01-22
Mention `@GlobalScope` can be omitted in Class reference

Mickeon:
Dipping my toes into modifying this document. I wanted to say more but for now this is a pretty noteworthy thing to mention.

### DIFF
--- a/contributing/documentation/class_reference_primer.rst
+++ b/contributing/documentation/class_reference_primer.rst
@@ -132,7 +132,7 @@ Linking
 """""""
 
 Whenever you link to a member of another class, you need to specify the class name.
-For links to the same class, the class name is optional and can be omitted.
+For links to the same class or to :ref:`class_@GlobalScope`, the class name is optional and can be omitted.
 
 +-------------------------------+-----------------------------------------+----------------------------------------------------------------------+
 | Tag and Description           | Example                                 | Result                                                               |


YuriSizov:
I wouldn't do that to avoid potential conflicts.

AThousandShips:
Was this intended to be specific to 4.2? Otherwise it's the wrong branch

Mickeon:
Oops. Nope, it should be on master.

Mickeon:
Either way, I thought about it and I believe it is still worth mentioning. I tested having `@GlobalScope` excluded from the text, and it does look very nice but there are too many occasions where it's ambiguous (all of the Variant's `is_equal_approx()` methods.).

So, having control over when it appears is nice, I wouldn't discourage mentioning it because it makes a lot of descriptions (such as ones that talk about error constants) much more readable.

YuriSizov:
> I wouldn't discourage mentioning it

So we agree on that point and this PR should be archived?

Mickeon:
Filtering both negatives in the sentence I meant "I would encourage mentioning it".
It should be archived regardless because it's targeting the wrong branch. I'll try again another time, sorry for the inconvenience.

YuriSizov:
> Filtering both negatives in the sentence I meant "I would encourage mentioning it".

So, again, we agree then? 🙃 Current version of the PR suggests that not mentioning it is okay, which I disagree with. But if you want to change it to actively encourage mentioning the global scope, then I support that.

PR #8815, opened 2024-01-23
Add C# signal automatic disconnection info

31:
After https://github.com/godotengine/godot-docs/pull/8812, I looked a bit further at this `Delegate.Target` quirk that I remember noticing a while ago but hadn't looked into. This is just based on experimentation rather than tracking it down in Godot/.NET source, so I'm not 100% sure my reasoning is correct.

I don't know if all this specific stuff is already tracked by https://github.com/godotengine/godot/issues/70414 / https://github.com/godotengine/godot/pull/73730, or if there's more to pull out into a separate functional issue. This has been at the back of my mind for a while since it comes up in the C# Discord chat fairly often, so I figured now that I've looked into it a bit more, I'd just write it as a doc either way. 😅

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_signals.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_signals.rst
@@ -15,6 +15,10 @@ In some cases it's necessary to use the older
 :ref:`Disconnect()<class_object_method_disconnect>` APIs.
 See :ref:`using_connect_and_disconnect` for more details.
 
+If you encounter a ``System.ObjectDisposedException`` while handling a signal,
+you might be missing a signal disconnection. See
+:ref:`disconnecting_automatically_when_the_receiver_is_freed` for more details.
+
 Signals as C# events
 --------------------
 
@@ -32,15 +36,6 @@ In addition, you can always access signal names associated with a node type thro
 .. code-block:: csharp
 
     await ToSignal(GetTree(), SceneTree.SignalName.ProcessFrame);
-
-.. warning::
-
-    While all engine signals connected as events are automatically disconnected when nodes are freed, custom
-    signals connected using ``+=`` aren't. This means you will need to manually disconnect (using ``-=``)
-    all the custom signals you connected as C# events (using ``+=``).
-
-    An alternative to manually disconnecting using ``-=`` is to
-    :ref:`use Connect <using_connect_and_disconnect>` rather than ``+=``.
 
 Custom signals as C# events
 ---------------------------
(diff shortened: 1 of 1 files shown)


paulloz:
This seems fine to me. The ordering within the page is just a bit odd now:
- Signals as C# events
- Disconnecting automatically when the receiver is freed
- Custom signals as C# events
- Signal emission
- Bound values
- Signal creation at runtime
- Using Connect and Disconnect

Maybe the new section should instead be moved at the very end, or before the last one?

31:
Good point, I didn't really consider the order all that carefully. Adding it at the bottom makes sense to me.

That makes me think about how I expect people to actually get here... I think calling it out in the summary with the type of exception someone is probably seeing if they've hit this quirk is a good way to go.

Pushed the changes.

tutorials/scripting/c_sharp/c_sharp_signals.rst
```diff
@@ -184,3 +179,164 @@ does nothing.
...
+On tick 4, the lambda expression tries to access the ``Name`` property of the
+node, but the node has already been freed. This causes the exception.
+
+To disconnect, keep a reference to the delegate created by the lambda expression
+and pass that to ``-=``. For example, this node connects and disconnects using
+the ``_EnterTree`` and ``_ExitTree`` lifecycle methods:
+
+.. code-block:: csharp
+
+    [Export] public Timer myTimer;
```
raulsntos, line 264: I think we usually add a new line after attributes in our documentation examples so we should follow that style to be consistent. Also, public fields should be properties.

```suggestion
    [Export]
    public Timer MyTimer { get; set; }
```

raulsntos (commented):
All of this has made me realize how complicated C# signals are at the moment. I feel like users shouldn't need to know all this just to use signals in C#, but since that's the current state of things it's better to document it than to leave users confused.

tutorials/scripting/c_sharp/c_sharp_signals.rst
```diff
@@ -184,3 +179,164 @@ does nothing.
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Connecting to a custom signal using ``+=`` doesn't disconnect automatically when
+the receiving node is freed.
+
+To disconnect, use ``-=`` at an appropriate time. For example:
+
+.. code-block:: csharp
+
+    [Export] public MyClass myClass;
```
31, line 320: To change this to a property in my own code I would use `public MyClass MyClass { get; set; }` (typical C# convention), but in docs I don't like to, to make it totally clear for someone who isn't used to it.

31:
Yeah, initially I thought using `.Connect` was solving everything and the situation was ok, but this is quite unfortunate.

If we end up stuck here, I think it might be simpler if the docs included manual disconnection everywhere for C#, and this page mentioned that automatic disconnection happens in a few specific cases. That's a pretty big change though, and if https://github.com/godotengine/godot/pull/73730 ends up making things simpler (or ends up with us manually tying references to our delegates to the lifetime of a Godot object, or... something), not sure it's worth too much churn simplifying where we're at now.

mhilbrunner (approved):

mhilbrunner:
Thank you!

PR #8852, opened 2024-01-27
singletons_autoload.rst: add C# example similar to "Global Variable"

31:
The singleton page leaves C# devs with the clunky `GetNode<PlayerVariables>("/root/PlayerVariables")`. In the C# discord help channel, adding a `public static PlayerVariables Instance { get; private set; }` is probably one of the most repeated bits of code. This PR tries adding it to the doc.

### DIFF
--- a/tutorials/scripting/singletons_autoload.rst
+++ b/tutorials/scripting/singletons_autoload.rst
@@ -94,6 +94,33 @@ be accessed directly in GDScript, without requiring ``get_node()``:
 
    PlayerVariables.health -= 10
 
+The **Enable** column has no effect in C# code. However, if the singleton is a
+C# script, a similar effect can be achieved by including a static property
+called ``Instance`` and assigning it in ``_Ready()``:
+
+.. tabs::
+ .. code-tab:: csharp
+
+    public partial class PlayerVariables : Node
+    {
+        public static PlayerVariables Instance { get; private set; }
+
+        public int Health { get; set; }
+
+        public override void _Ready()
+        {
+            Instance = this;
+        }
+    }
+
+This allows the singleton to be accessed from C# code without ``GetNode()`` and
+without a typecast:
+
+.. tabs::
+ .. code-tab:: csharp
+
+    PlayerVariables.Instance.Health -= 10;
+
 Note that autoload objects (scripts and/or scenes) are accessed just like any
 other node in the scene tree. In fact, if you look at the running scene tree,
 you'll see the autoloaded nodes appear:


tutorials/scripting/singletons_autoload.rst
```diff
@@ -94,6 +94,27 @@ be accessed directly in GDScript, without requiring ``get_node()``:
 
    PlayerVariables.health -= 10
 
+The **Enable** column has no effect in C# code. However, if the singleton is a
+C# script, a similar effect can be achieved by adding a static property:
```
31, line 98: I think it's awkward enough having a GDScript-only code tab switcher, so I don't think it's great to also add a C#-only one. If there's some way to put it all under C#/GDScript tabs, it seems like that would be nicer? I don't imagine it would be ok to change all the explanation into a code comment, though, and that's the only way I know how to do it.

Another idea is to put this example in the C# docs somewhere (in the "C# language features" page?) and add a note here that links to it. I don't really know how important it is to document C# stuff up front vs. siloing it.

I don't have any strong preferences here, just want to get this in the docs somewhere. 😄

raulsntos: Code can't be translated so I don't think we should move the explanation to code comments.

As for tab switchers with only one tab, we can add code snippets without tabs. See: https://github.com/godotengine/godot-docs/blob/67b17958ca317cec3230d09ed01f45e799ecf949/tutorials/scripting/c_sharp/c_sharp_basics.rst?plain=1#L176

But this is from the C# basics page so one could assume that all the code snippets are in C#, therefore there's no need to indicate the language. In the singletons page we have snippets in GDScript too, so maybe the tab is useful as a language indicator idk.

tutorials/scripting/singletons_autoload.rst
```diff
@@ -94,6 +94,27 @@ be accessed directly in GDScript, without requiring ``get_node()``:
...
+
+.. tabs::
+ .. code-tab:: csharp
+
+    public static PlayerVariables Instance { get; private set; }
+
+    public override void _Ready()
+    {
+        Instance = this;
+    }
```
raulsntos, line 108: Without the class definition I'm worried that users may not understand where this code snippet is supposed to go.

raulsntos (commented):
I think it's good to try and provide good examples of practical C# like this throughout the documentation. So I'm in favor of adding more C#-specific examples rather than 1-to-1 translations where we can (related: https://github.com/godotengine/godot-docs/issues/6622).

tutorials/scripting/singletons_autoload.rst
31: Yeah, language indicator is how I interpreted the single tag, and as long as we can't make whole sections swap, IMO it does its job fine. Interesting note about translation, didn't realize that but it makes sense.

tutorials/scripting/singletons_autoload.rst
31: Hmm, agreed. I'll fill it out. Maybe it could be declared beforehand (for GDScript too) but writing out the whole C# class here at least makes it totally clear.

raulsntos (approved):
Looks good to me, thanks!

mhilbrunner (approved):

mhilbrunner:
Thank you!

mhilbrunner:

PR #8853, opened 2024-01-28
c_sharp_exports.rst: fix private naming style; minor grammar

31:
I took a peek at this doc for reference for https://github.com/godotengine/godot-docs/pull/8815#discussion_r1468713871 and saw that `private Foo _foo` style wasn't being used consistently here. Also a couple minor grammar changes.

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_exports.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_exports.rst
@@ -15,7 +15,7 @@ Exporting is done by using the ``[Export]`` attribute.
     public partial class ExportExample : Node3D
     {
         [Export]
-        private int Number = 5;
+        public int Number { get; set; } = 5;
     }
 
 In that example the value ``5`` will be saved, and after building the current project
@@ -36,7 +36,7 @@ Exporting can only be done with :ref:`Variant-compatible <doc_c_sharp_variant>` 
 Basic use
 ---------
 
-Exporting can work with fields and properties.
+Exporting works with fields and properties. They can have any access modifier.
 
 .. code-block:: csharp
 
@@ -54,10 +54,10 @@ because ``string`` is a reference type.
 .. code-block:: csharp
 
     [Export]
-    private int _number;
+    public int Number { get; set; }
 
     [Export]
-    private string _text;
+    public string Text { get; set; }
 
 Default values can be specified for fields and properties.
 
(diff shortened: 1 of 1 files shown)


paulloz (approved):

AThousandShips:
I think these should be public instead as they're exported

CC @raulsntos

paulloz:
We use private exported fields in a lot of places, and we explicitly discourage using public fields in the style guide.

AThousandShips:
Where? The [styleguide](https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_style_guide.html) makes no mention of exports, and the [exports page](https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_exports.html) uses public almost all the time

The official demo projects also do not use private fields exclusively

paulloz:
> The [styleguide](https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_style_guide.html) makes no mention of exports

Didn’t say the styleguide mentioned exports, it mentions preferring properties to non public fields. And yes, the docs also show a lot of exported public properties. Which is why I’m saying if this is intended to be shown as public, it should be a property, else it should stay private.

AThousandShips:
Gotcha, did mean public and fields, now necessarily public plain variables 🙂 sorry for being unclear

31:
> I think these should be public instead as they're exported

Personally, I agree. `[Export]` to me is simply annotating a piece of my public API saying that Godot should show it in the inspector.

However, many people I've talked with (e.g. on the C# help discord channel) are accustomed to C# serialization libraries that readily get and set `private` state. From this perspective, `[Export] private int _x;` is a form of encapsulation: it allows Godot to change these members while protecting them from inadvertent changes by other C# scripts that interact with them.

I think it's important for this doc to point out that fields and properties, public and private, all work, so nobody thinks Godot is "forcing" them to do something they feel is unnatural--regardless of Godot or broader industry style.

That said, making all but the first few examples default to `public Foo Foo { get; set; }` doesn't seem bad to me as long as it's still obvious that private fields/properties work.

raulsntos (approved):
Making them public makes sense to me. As long as, like you said, it's clear that it's also possible to export fields, and that the access modifier doesn't matter.

I think, with the current changes, it's clear enough.

mhilbrunner (approved):

mhilbrunner:
Thank you! Merged. 🎉

PR #8856, opened 2024-01-28
Clarify what Navlink does and does not do

### LINKED ISSUE #7598: Using NavigationLinks page is misleading
**Your Godot version:**
v4.1.rc2.official.46424488e

**Issue description:**

All this info at the bottom of the page seems incorrect.
image

What happens is the `NavigationLink*` nodes seem to provide a valid navigation path that the AI can traverse just fine.

https://docs.godotengine.org/en/stable/tutorials/navigation/navigation_using_navigationlinks.html

I don't know if I checked enough edge cases, but from my tests `NavigationLink*` nodes enable the AI to traverse using its own path strategy without the need to manually move the agents with animation or teleportation or any other way.

Here's a small demo showing how the navigation agent moves from island to island using `NavigationLink2D`.

https://github.com/godotengine/godot-docs/assets/1177508/9883f562-dd49-4b6d-98c4-9c00b5e67772

*Note* that in the latter half I'm trying to get the agent to move to the edge of the island by giving it unreachable targets and it doesn't do what I expect it to do. Some times it moves to the edge and others it does't. Seems to be dependent on the distance to the navigation region, but on the convex one to the right it goes to incorrect placements. I think this is related to https://github.com/godotengine/godot/issues/78943

[godot-4-dbg.zip](https://github.com/godotengine/godot-docs/files/11929479/godot-4-dbg.zip)

skyace65:
Clarifies how navlink works, and what happens if you don't setup special movement. Closes #7598. That's the second issue that's been opened about this specific section so I figure it needs more clarification.

### DIFF
--- a/tutorials/navigation/navigation_using_navigationlinks.rst
+++ b/tutorials/navigation/navigation_using_navigationlinks.rst
@@ -44,13 +44,12 @@ If no valid polygon is found within the search radius the navigation link gets d
 The link debug visuals can be changed in the Editor :ref:`ProjectSettings<class_ProjectSettings>` under ``debug/shapes/navigation``.
 The visibility of the debug can also be controlled in the Editor 3D Viewport gizmo menu.
 
-.. note::
-
-    NavigationLinks do not move agents between the two link positions by themselves.
-
-A navigation link does not provide any automated movement through the link. Instead, when
-an agent reaches the position of a link, game code needs to react (e.g. through area triggers) and provide means for the agent
-to move through the link to end up at the links other position (e.g. through teleport or animation) to continue along the path.
+A navigation link does not provide any specialized movement through the link. Instead, when
+an agent reaches the position of a link, game code needs to react (e.g. through area triggers)
+and provide means for the agent to move through the link to end up at the links other position
+(e.g. through teleport or animation). Without that an agent will attempt to move itself along
+the path of the link. You could end up with an agent walking over a bottomless pit instead of
+waiting for a moving platform, or walking through a teleporter and proceeding through a wall.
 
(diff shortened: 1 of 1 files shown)


mhilbrunner:
cc @smix8 :)

mhilbrunner (approved):

mhilbrunner:
Thanks!

kthecoder:
I don't know if it's just me, but I still fail to understand how to, "and provide means for the agent to move through the link to end up at the links other position".

In, [Godot Forum Post](https://forum.godotengine.org/t/how-to-use-the-navigation-links/42351/6); it is suggested to use the NavigationAgent3D's signal called “link_reached”.

In GDscript I have created a Navigation Link 3D for an Agent to jump off the platform. (Platform is dynamically spawned during gameplay) But the "link_reached" signal does not come with an Owner. I assume this is because I created it in code and so there is no Node for my link? So I tried storing the dynamically created Navigation Link's into a Dictionary that can be checked and when the Agent hits a link in the Dictionary it then knows it should jump. So the Agent jumps but then gets stuck.

_**What is the proper way of managing the traversal of the navigation link path in code then?**_

Here is a GIF of what happens when I create a link on top of my platform 10m above the ground navmesh. The agent is spawned on top of a platform and tries to jump off but then for an unknown reason gets stuck hanging in mid air.

https://imgur.com/shRgIBt

I already tried messing with the 'Path Desired Distance' and 'Target Desired Distance' but they cause the agent to move along the path weirdly and the Agent still doesn't update its NavigationMeshRID map to the ground nav mesh.

mhilbrunner:

PR #8763, opened 2024-01-13
Pin sphinx-contrib libraries to pre-5.0 version

mhilbrunner:
Manual squash and cherrypick of https://github.com/godotengine/godot-docs/pull/8760 and the followup PRs https://github.com/godotengine/godot-docs/pull/8761 and https://github.com/godotengine/godot-docs/pull/8762.

After updates on the side of Sphinx/Sphinx requirements, we need to pin these versions of these requirements that are included by Sphinx as the just published ones depend on Sphinx 5.0+, while we are still on 4.x.

This needs to be cherrypicked to all active branches that are built on ReadTheDocs to ensure builds keep working there.

PR #8857, opened 2024-01-29
Pin sphinx extensions to fix offline builds for 3.6 branch

### LINKED ISSUE #8849: Build for offline tasks failing
The 3.6 doc builder is failing, which is causing the 4.x builders to not execute. There are no available offline versions as the last successful build was too long ago.

skyace65:
Offline builds for 3.6 aren't working for this branch and it's affecting others, see [here](https://github.com/godotengine/godot-docs/actions/runs/7689407863). Closes #8849. I think this is what we need, this is the first time I'm doing something like this, I'm just guessing based on how this file is in the master branch.


mhilbrunner (approved):

mhilbrunner:
See also: https://github.com/godotengine/godot-docs/pull/8763 (which has not been cherrypicked everywhere yet). Thanks!

PR #8859, opened 2024-01-29
c_sharp_exports.rst: add complex getter/setter note

31:
I think that with an example saying you can use a backing field, it's worth pointing out this quirk. It doesn't come up that often but people do run into it. I think that explaining that the C# code is analyzed also helps make some other behavior more reasonable without needing to point it out: e.g. that constructors aren't involved.


tutorials/scripting/c_sharp/c_sharp_exports.rst
```diff
@@ -62,15 +62,50 @@ Exported members can specify a default value; otherwise, the `default value of t
...
+
+    However, this won't work: the analyzer doesn't understand this code, and the
+    default value will show up in the property editor as ``0``. However, when
+    running the scene, ``_number`` will be ``2``, and ``NumberWithBackingField``
+    will be ``5``.
+
+    Godot will execute the getter if the script is in
+    :ref:`tool mode <doc_running_code_in_the_editor>`. However, the behavior
+    becomes inconsistent. With ``[Tool]``, the default value may act as ``5``
+    or ``0`` depending on the situation.
```
raulsntos, line 105: I'm not sure what you mean by this. The default value should be the same as when the script wasn't a `[Tool]`.

31: Going into this, my understanding was that `[Tool]` actually intentionally brought back to the 3.x workflow where Godot constructs the object and gets the post-construction value and uses that as the default. But I found that isn't true (anymore?), it just makes the UI appear buggy in 4.2.1 and 4.3dev2. The clearest issue is that the inspector shows the reset button in a fresh scene.

It sounds like this isn't widely known, happy to file a bug. I don't know what the intentional behavior here is (or should be).

31: In case we're talking past each other, by "default value" I just mean how it's represented in the editor. I don't have a deeper understanding of how the default value is represented internally or what the "real" default value is, vs. a UI bug that misrepresents it.

raulsntos: Yes, I think we were talking past each other. By "default value" I meant the value that the editor _understands_ as the default, which is specified by the C# source generators. This value is used to determine whether the reset button needs to be displayed and what value to set the property to when the reset button is pressed.

> The clearest issue is that the inspector shows the reset button in a fresh scene.

I think calling this an issue is a matter of perspective. From the editor's side this property's "default value" is `default` (or `0` if you prefer) just like when the script is not `[Tool]`. The only difference is that `[Tool]` scripts are run in the editor and, as a side-effect, it changes the value of the property from the "default value". And when you save the scene it will also serialize this value since, for the editor, this is not the "default value".

But I understand how this can be confusing, specially when users don't realize that the "default value" may not correspond to what they've written in code (like in your complex property example).

> Going into this, my understanding was that `[Tool]` actually intentionally brought back to the 3.x workflow where Godot constructs the object and gets the post-construction value and uses that as the default.

I don't think so. We intentionally changed this behavior because users were complaining about side-effects on their project when constructing their types just to get the default values. So now your types are "never" constructed. Unless your script is a `[Tool]`, in this case we have to construct them, not to retrieve the default values, but because we need an instance of your script to make it run in the editor (this applies to the entire engine lifetime callbacks as you'd expect).

The difference between a `[Tool]` and a non-`[Tool]` script for the inspector is that the `[Tool]` script is an instance of the real script and the non-`[Tool]` uses a placeholder so we don't have to construct the real thing. Since the `[Tool]` script is a real instance, when the inspector tries to retrieve the property values to display them it will naturally call your property getters and as a result the inspector can show a different value from the "default" when opening the scene.

31: > And when you save the scene it will also serialize this value since, for the editor, this is not the "default value".

Ah... the scene's tab doesn't mark it as modified in the tab when this happens (just `Scene`, not `Scene (*)`) but it *does* serialize `5` into the tscn if I save it. That's interesting, and it probably explains why it just seemed like nonsense behavior: because I wasn't aware that I was saving any changes.

Thanks for the explanation. I've tried simplifying the note to make it useful without getting into the weeds.

raulsntos (approved):
LGTM. Thanks!

mhilbrunner (approved):

mhilbrunner:
Thank you!

PR #8862, opened 2024-01-29
Update ray-casting.rst to make 3D example simpler

Nodragem:
Hello, I hope I did the right steps for contributing!

The Raycast 3D API is a hot topic, and I noticed a couple of times where new users find it confusing or overcomplicated. 

I believe that one reason is that the current example is a bit too long, and seems complicated at first look. 

I suggest to uncluter the example by:
- use `PhysicsRayQueryParameters3D.new()` to highlight that it's just a two-step process: we first fill in the parameters, then we call intersect_ray
- unclutter the example by removing unnecessary variable assignments
- remove `cam.project_ray_origin`, which basically just returns the camera's origin, so that the reader does not get distrcated/confused
- replace `origin + cam.project_ray_normal(mousepos) * RAY_LENGTH` with `project_position(mousepos, RAY_LENGTH)` (needs to be tested, but I think it is equivalent)
- I renamed `query` to `params` to highlight again that we are just filling a form. It also better match the last noun in `PhysicsRayQueryParameters3D`. But if we go for `params`, we need to change it for the other examples as well.

### DIFF
--- a/tutorials/physics/ray-casting.rst
+++ b/tutorials/physics/ray-casting.rst
@@ -157,14 +157,14 @@ The data is similar in 3D space, using Vector3 coordinates. Here an example assu
  .. code-tab:: gdscript GDScript
 
         const RAY_LENGTH = 1000
+        var parameters = PhysicsRayQueryParameters3D.new()
 
-        func _physics_process(delta):
-            var params = PhysicsRayQueryParameters3D.new()
-            params.from = $Camera3D.origin
-            params.to = $Camera3D.project_position(
+        func _physics_process(delta):            
+            parameters.from = $Camera3D.position
+            parameters.to = $Camera3D.project_position(
                     get_viewport().get_mouse_position(),
                     RAY_LENGTH)
-            var result = get_world_3d().direct_space_state.intersect_ray(params)
+            var result = get_world_3d().direct_space_state.intersect_ray(parameters)
 
 Note that to enable collisions with Area3D, the boolean parameter ``params.collide_with_areas`` must be set to ``true``.
 


tutorials/physics/ray-casting.rst
```diff
@@ -151,25 +151,22 @@ data:
        metadata: Variant() # metadata of collider
     }
 
-The data is similar in 3D space, using Vector3 coordinates. Note that to enable collisions
-with Area3D, the boolean parameter ``collide_with_areas`` must be set to ``true``.
+The data is similar in 3D space, using Vector3 coordinates. Here an example assuming that `$Camera3D` is the main camera node:
```
AThousandShips, line 154:
```suggestion
The data is similar in 3D space, using Vector3 coordinates. Here an example assuming that ``$Camera3D`` is the main camera node:
```

tutorials/physics/ray-casting.rst
```diff
@@ -151,25 +151,22 @@ data:
...
-            var query = PhysicsRayQueryParameters3D.create(origin, end)
-            query.collide_with_areas = true
-
-            var result = space_state.intersect_ray(query)
+        	var params = PhysicsRayQueryParameters3D.new()
+        	params.from = $Camera3D.origin
+        	params.to = $Camera3D.project_position(
+        							get_viewport().get_mouse_position(), 
+        							RAY_LENGTH )
+        	var result = get_world_3d().direct_space_state.intersect_ray(params)
```
AThousandShips, line 167:
```suggestion
            var params = PhysicsRayQueryParameters3D.new()
            params.from = $Camera3D.origin
            params.to = $Camera3D.project_position(
                    get_viewport().get_mouse_position(),
                    RAY_LENGTH)
            var result = get_world_3d().direct_space_state.intersect_ray(params)
```
Two indents for continuations, and use four spaces for indentation, alternatively:
```suggestion
            var params = PhysicsRayQueryParameters3D.new()
            params.from = $Camera3D.origin
            params.to = $Camera3D.project_position(get_viewport().get_mouse_position(), RAY_LENGTH)
            var result = get_world_3d().direct_space_state.intersect_ray(params)
```

Nodragem: if possible I like lines not to be too long.

Nodragem:
let me know your thoughts concerning replacing the variable name `query` with `params`.

smix8:
I think the physics examples have this general problem that they teach users to create new throwaway parameter objects inside physics loops all the time which is not good. I would move and create the parameters object outside the physics loop so it can be reused by just updating the parameters. Else the physics loop might spam up to 8 new parameter objects each main loop iteration for nothing.

```gdscript
var parameters: PhysicsRayQueryParameters3D = PhysicsRayQueryParameters3D.new()
```

Personally I also try to avoid using abbreviations for variable names in documentation examples. There is always "that one" user that does not understand the example due to an abbreviation and it makes things harder to read for many others. So I would name it just `parameters` instead of `params` even if it is for this example a neglectable difference.

Nodragem:
yes, I will check if it works when moving the `PhysicsRayQueryParameters3D.new()` outside the loop. Not sure what is the actual performance impact of doing so. I agree that we should use full names when possible in the doc.

Nodragem:
I push the corrections:
- change params to parameters;
-  origin to position; 
- and move .new() out of _physics_process().

And I tested the code in Godot 4.2.2.

Let me know if I should update the 2D examples as well.

PR #8865, opened 2024-01-29
Update recommended MSVC version in Compiling for Windows

Calinou:
- Mention that official Godot binaries are compiled using MinGW (and describe why).


mhilbrunner:
Thanks!

PR #8878, opened 2024-02-02
c_sharp_variant.rst: add more examples, fix explicit vs. implicit

31:
This page deserves examples and a little correction about the implicit vs. explicit conversions.

The example about comparing two `Variant`s using `.Obj` and `==` might seem oddly specific, but it's come up at least twice now in the C# discord help channel.

I also assumed a while back that `.As<T>` would behave more like a `a as B`, and the explicit conversion `(B)a` would throw an exception, but it's not the case (and the `Packed{Type}Array` empty array case is interesting), so worth mentioning IMO, but not sure in how much detail.

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_variant.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_variant.rst
@@ -10,7 +10,7 @@ We recommend avoiding ``Godot.Variant`` unless it is necessary to interact with 
 Take advantage of C#'s type safety when possible.
 
 Converting from a Variant-compatible C# type to ``Godot.Variant`` can be done using implicit
-conversions. Also available are ``CreateFrom`` method overloads and the generic ``Variant.From<T>``
+conversions. There are also ``CreateFrom`` method overloads and the generic ``Variant.From<T>``
 methods. Only the syntax is different: the behavior is the same.
 
 .. code-block:: csharp
@@ -31,9 +31,9 @@ specifying the final color of the tween is a ``Godot.Variant``.
     Tween tween = CreateTween();
     tween.TweenProperty(GetNode("Sprite"), "modulate", Colors.Red, 1.0f);
 
-Converting from ``Godot.Variant`` to a C# type can be done using explicit conversions. Also
-available are ``Variant.As{TYPE}`` methods or the generic ``Variant.As<T>`` method. All of these
-behave the same.
+Converting from ``Godot.Variant`` to a C# type can be done using explicit conversions. There are
+also ``Variant.As{TYPE}`` methods and the generic ``Variant.As<T>`` method. All of these behave the
+same.
 
 .. code-block:: csharp
 
@@ -70,7 +70,7 @@ acceptable, consider using a ``Variant.As<GodotObject>() is MyNode n`` type patt
 .. note::
 
     Since the Variant type in C# is a struct, it can't be null. To create a "null"
-    Variant, use the ``default`` keyword or ``Godot.Variant`` parameterless constructor.
+    Variant, use the ``default`` keyword or the ``Godot.Variant`` parameterless constructor.
 
 Variant-compatible types
 ------------------------
--- a/tutorials/scripting/c_sharp/c_sharp_variant.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_variant.rst
@@ -9,62 +9,63 @@ For a detailed explanation of Variant in general, see the :ref:`Variant <class_V
 We recommend avoiding ``Godot.Variant`` unless it is necessary to interact with untyped engine APIs.
 Take advantage of C#'s type safety when possible.
 
-Converting from a supported type to ``Godot.Variant`` can be done using implicit conversions. Also
-available are ``CreateFrom`` method overloads and the generic ``Variant.From<T>`` methods.
+Converting from a Variant-compatible C# type to ``Godot.Variant`` can be done using implicit
+conversions. Also available are ``CreateFrom`` method overloads and the generic ``Variant.From<T>``
+methods. Only the syntax is different: the behavior is the same.
 
 .. code-block:: csharp
 
     int x = 42;
-    Godot.Variant number = x;
-    Godot.Variant hello = "Hello, World!";
+    Variant numberVariant = x;
+    Variant helloVariant = "Hello, World!";
 
-    Godot.Variant number2 = Godot.Variant.CreateFrom(x);
-    Godot.Variant number3 = Godot.Variant.From(x);
+    Variant numberVariant2 = Variant.CreateFrom(x);
+    Variant numberVariant3 = Variant.From(x);
 
 Implicit conversions to ``Godot.Variant`` make passing variants as method arguments very convenient.
-For example, ``final_val`` of :ref:`tween_property<class_Tween_method_tween_property>`.
+For example, the third argument of :ref:`tween_property<class_Tween_method_tween_property>`
+specifying the final color of the tween is a ``Godot.Variant``.
 
 .. code-block:: csharp
 
     Tween tween = CreateTween();
     tween.TweenProperty(GetNode("Sprite"), "modulate", Colors.Red, 1.0f);
 
-Converting from ``Godot.Variant`` to a supported type can be done using explicit conversions. Also
-available are ``Variant.As{TYPE}`` methods or the generic ``Variant.As<T>`` method.
+Converting from ``Godot.Variant`` to a C# type can be done using explicit conversions. Also
+available are ``Variant.As{TYPE}`` methods or the generic ``Variant.As<T>`` method. All of these
+behave the same.
 
 .. code-block:: csharp
 
-    int cSharpNumber = (int)number;
-    string cSharpHello = (string)hello;
+    int number = (int)numberVariant;
+    string hello = (string)helloVariant;
 
-    int cSharpNumber2 = number.As<int>();
-    string cSharpHello2 = hello.AsString();
+    int number2 = numberVariant.As<int>();
+    int number3 = numberVariant.AsInt32();
 
 .. note::
 
-    All methods of converting from ``Godot.Variant`` to a C# type behave the same way. If conversion
-    is not possible, the default value of the target type or an empty array is returned. An
-    exception is not thrown.
+    The ``Variant.As{TYPE}`` methods are typically named after C# types (``Int32``), not C# keywords
+    (``int``).
 
-A matching type is not necessary for a conversion to succeed. For example, all variants are
-convertible to ``string``.
+If the Variant type doesn't match the conversion target type, the consequences vary depending on the
+source and target values.
 
-.. code-block:: csharp
+- The conversion may examine the value and return a similar but potentially unexpected value of the
+  target type. For example, the string ``"42a"`` may be converted to the integer ``42``.
+- The default value of the target type may be returned.
+- An empty array may be returned.
+- An exception may be thrown.
 
-    string s = Variant.From(new Vector3(1, 3, 9)).As<string>();
-    GD.Print(s); // (1, 3, 9)
+Converting to the correct type avoids complicated behavior and should be preferred.
 
-Some C# types are not directly represented by ``Variant.Type``. The ``Variant.As{TYPE}`` method uses
-a more specific name to represent them.
+The ``Variant.Obj`` property returns a C# ``object`` with the correct value for any variant. This
+may be useful when the type of Variant is completely unknown. However, when possible, prefer more
+specific conversions. ``Variant.Obj`` evaluates a ``switch`` on ``Variant.VariantType`` and it may
+not be necessary. Also, if the result is a value type, it may be boxed when it normally wouldn't be.
 
-.. code-block:: csharp
-
-    int cSharpNumber3 = number.AsInt32();
-
-To convert ``Godot.Variant`` to a C# ``object`` without specifying a more specific target type, use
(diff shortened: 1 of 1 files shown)


tutorials/scripting/c_sharp/c_sharp_variant.rst
```diff
@@ -9,17 +9,67 @@ For a detailed explanation of Variant in general, see the :ref:`Variant <class_V
...
+    string cSharpHello2 = hello.AsString();
+
+.. note::
+
+    All methods of converting from ``Godot.Variant`` to a C# type behave the same way. If conversion
+    is not possible, the default value of the target type or an empty array is returned. An
+    exception is not thrown.
+
+A matching type is not necessary for a conversion to succeed. For example, all variants are
+convertible to ``string``.
```
raulsntos, line 50: In general, converting between Variant types will always "succeed" but I don't think we should be encouraging this.

I guess it depends on what you consider success, the Variant conversion will try to get the "closest" value but that can be confusing in some cases. Here are some interesting examples:

```csharp
Variant v;
int n;

v = "42"; // This is a string Variant.
n = v.As<int>(); // Converts the string to the int '42'.

v = "42.5";
n = v.As<int>(); // Converts the string to the int '42' discarding the decimals.

v = "42a";
n = v.As<int>(); // Converts the string to the int '42' discarding the suffix.
                 // Honestly, I didn't think this would work (I expected '0').

v = "Hello world";
n = v.As<int>(); // Converts the string to the int '0' because it can't convert the text,
                 // but doesn't print any errors.

v = new Vector3(1, 3, 9); // This is a Vector3 Variant.
n = v.As<int>(); // Converts the Vector3 to the int '0' because it can't convert the Vector3,
                 // but doesn't print any errors.

v = 42; // This is a int Variant.
v.As<Vector3>(); // Converts the int to the Vector3 '(0, 0, 0)' because it can't convert the int,
                 // but doesn't print any errors.
```

tutorials/scripting/c_sharp/c_sharp_variant.rst
```diff
@@ -9,17 +9,67 @@ For a detailed explanation of Variant in general, see the :ref:`Variant <class_V
...
+
+.. code-block:: csharp
+
+    string s = Variant.From(new Vector3(1, 3, 9)).As<string>();
+    GD.Print(s); // (1, 3, 9)
+
+Some C# types are not directly represented by ``Variant.Type``. The ``Variant.As{TYPE}`` method uses
+a more specific name to represent them.
```
raulsntos, line 58: If this is meant to help users trying to figure out the naming for the `As{TYPE}` methods, I think we can just say that the name will match the name of the C# type (not the keyword):

```
bool -> Bool // This really should have been Boolean.
int -> Int32
long -> Int64
float -> Single
double -> Double
```

tutorials/scripting/c_sharp/c_sharp_variant.rst
```diff
@@ -9,17 +9,67 @@ For a detailed explanation of Variant in general, see the :ref:`Variant <class_V
...
+
+.. code-block:: csharp
+
+    int cSharpNumber3 = number.AsInt32();
+
+To convert ``Godot.Variant`` to a C# ``object`` without specifying a more specific target type, use
+the ``Variant.Obj`` property. This can be used to compare two variants that both have unknown types
+using the equality operator ``==``. When possible, prefer more specific conversions. ``Variant.Obj``
+may result in unnecessary boxing of value types.
```
raulsntos, line 67: I don't know if it's a good idea to use `Variant.Obj` to compare Variant values, and in general I'd prefer to not encourage the usage of that property.

Using `==` with boxed values will fail:

```csharp
object a = 42;
object b = 42;
Console.WriteLine(a == b); // This prints False.
Console.WriteLine(a.Equals(b)); // This prints True.
```

It may be better to check the `VariantType` first, and then use the `As{TYPE}` or `As<T>` methods:

```csharp
public static bool IsEqualsTo(this Variant left, Variant right)
{
    if (left.VariantType != right.VariantType)
    {
        return false;
    }

    return switch left.VariantType
    {
        Variant.Type.Nil => true,
        Variant.Type.Bool => left.AsBool() == right.AsBool(),
        Variant.Type.Int => left.AsInt64() == right.AsInt64(),
        Variant.Type.Float => left.AsDouble() == right.AsDouble(),
        // ...
    }
}
```

raulsntos (commented):
I appreciate the effort to improve the `Variant` documentation. I want to be careful about how we explain the type to avoid encouraging its usage more than necessary, I hope I wasn't too nit-picky.

tutorials/scripting/c_sharp/c_sharp_variant.rst
31: I'm fine cutting it down to just a heads up. I could see this paragraph and example being misconstrued as something to use rather than `ToString`.

> ```cs
> v = "42a";
> n = v.As<int>(); // Converts the string to the int '42' discarding the suffix.
>                  // Honestly, I didn't think this would work (I expected '0').
> ```

It looks like this was documented in 3.0, and it's more than just suffixes: https://docs.godotengine.org/en/3.0/classes/class_int.html, not sure where this text went (if anywhere):

> This method will ignore all non-number characters, so calling int('1e3') will return 13.

It seems like ideally the Godot Engine conversion logic would be documented somewhere, and this page could just say "conversions have the same loose logic as that ->".

tutorials/scripting/c_sharp/c_sharp_variant.rst
31: I was thinking that it might be enough to hint to people that the `AsInt32` in their intellisense is correct, but yeah, can get more specific. (With I guess a caveat "in most cases". 😄)

tutorials/scripting/c_sharp/c_sharp_variant.rst
31: Hmm. I'll just remove this paragraph for now.

Adding it to the Variant API could raise questions like whether Int 1 == Float 1 (should it attempt to convert), or less likely Bool true, etc. Not sure what would be expected based on how GDScript/Godot behaves.

tutorials/scripting/c_sharp/c_sharp_variant.rst
```diff
@@ -9,17 +9,67 @@ For a detailed explanation of Variant in general, see the :ref:`Variant <class_V
...
+Converting from ``Godot.Variant`` to a supported type can be done using explicit conversions. Also
+available are ``Variant.As{TYPE}`` methods or the generic ``Variant.As<T>`` method.
+
+.. code-block:: csharp
+
+    int cSharpNumber = (int)number;
+    string cSharpHello = (string)hello;
+
+    int cSharpNumber2 = number.As<int>();
+    string cSharpHello2 = hello.AsString();
```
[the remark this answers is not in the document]
raulsntos: I don't mind having examples for `As<T>` and `As{TYPE}`, but if you feel it's not needed feel free to remove them.

tutorials/scripting/c_sharp/c_sharp_variant.rst
raulsntos: > It looks like this was documented in 3.0 [...] not sure where this text went

Oh, you're right. It looks like the documentation for the `int(String)` constructor now points to the `String.to_int` method that has this in the description:

> This method removes any non-number character and stops at the first decimal point (`.`)

I agree that ideally all this `Variant` conversion is an engine thing rather than a C#-specific thing so it should probably be documented in the general `Variant` documentation (maybe [here](https://docs.godotengine.org/en/stable/classes/class_variant.html) or [here](https://docs.godotengine.org/en/stable/contributing/development/core_and_modules/variant_class.html)?) and just linked from the C# documentation.

tutorials/scripting/c_sharp/c_sharp_variant.rst
raulsntos: Yeah I don't know either, I think it's a complicated topic where different users may have different needs (e.g.: JavaScript's `==` vs `===`). So I feel like it's better to leave it to a user-defined method.

tutorials/scripting/c_sharp/c_sharp_variant.rst
raulsntos: I think it's enough, I just don't know if users will make the connection that the name comes from the C# type name and is not just some random name we chose.

tutorials/scripting/c_sharp/c_sharp_variant.rst
31: I'll leave an exhaustive explanation for later--for now, simplifying this to a scary list.

tutorials/scripting/c_sharp/c_sharp_variant.rst
```diff
@@ -9,17 +9,68 @@ For a detailed explanation of Variant in general, see the :ref:`Variant <class_V
 We recommend avoiding ``Godot.Variant`` unless it is necessary to interact with untyped engine APIs.
 Take advantage of C#'s type safety when possible.
 
-Any of ``Variant.As{TYPE}`` methods or the generic ``Variant.As<T>`` method can be used to convert
-a ``Godot.Variant`` to a C# type. Since the ``Godot.Variant`` type contains implicit conversions
-defined for all the supported types, calling these methods directly is usually not necessary.
+Converting from a Variant-compatible C# type to ``Godot.Variant`` can be done using implicit
+conversions. Also available are ``CreateFrom`` method overloads and the generic ``Variant.From<T>``
```
AThousandShips, line 13:
```suggestion
conversions. You also have ``CreateFrom`` method overloads and the generic ``Variant.From<T>``
```
Or
```suggestion
conversions. There are also ``CreateFrom`` method overloads and the generic ``Variant.From<T>``
```
"Also available are" sounds clunky in English IMO

tutorials/scripting/c_sharp/c_sharp_variant.rst
```diff
@@ -9,17 +9,68 @@ For a detailed explanation of Variant in general, see the :ref:`Variant <class_V
...
+For example, the third argument of :ref:`tween_property<class_Tween_method_tween_property>`
+specifying the final color of the tween is a ``Godot.Variant``.
+
+.. code-block:: csharp
+
+    Tween tween = CreateTween();
+    tween.TweenProperty(GetNode("Sprite"), "modulate", Colors.Red, 1.0f);
+
+Converting from ``Godot.Variant`` to a C# type can be done using explicit conversions. Also
+available are ``Variant.As{TYPE}`` methods or the generic ``Variant.As<T>`` method. All of these
```
AThousandShips, line 35: Same here

tutorials/scripting/c_sharp/c_sharp_variant.rst
```diff
@@ -9,17 +9,68 @@ For a detailed explanation of Variant in general, see the :ref:`Variant <class_V
...
+not be necessary. Also, if the result is a value type, it may be boxed when it normally wouldn't be.
+
+For example, if the potential for ``Variant.As<MyNode>()`` to throw a invalid cast exception isn't
+acceptable, consider using a ``Variant.As<GodotObject>() is MyNode n`` type pattern instead.
 
 .. note::
 
     Since the Variant type in C# is a struct, it can't be null. To create a "null"
-    Variant use the ``default`` keyword or the parameterless constructor.
+    Variant, use the ``default`` keyword or ``Godot.Variant`` parameterless constructor.
```
AThousandShips, line 73:
```suggestion
    Variant, use the ``default`` keyword or the ``Godot.Variant`` parameterless constructor.
```

tutorials/scripting/c_sharp/c_sharp_variant.rst
31: FWIW to me (en-us) it's a common phrase that implies the first option is the default. `also available` shows up elsewhere in the docs too, just not followed with `are` in particular. I do think `There are also` works fine though, happy to change it.

tutorials/scripting/c_sharp/c_sharp_variant.rst
```diff
@@ -9,17 +9,68 @@ For a detailed explanation of Variant in general, see the :ref:`Variant <class_V
...
+- The default value of the target type may be returned.
+- An empty array may be returned.
+- An exception may be thrown.
+
+Converting to the correct type avoids complicated behavior and should be preferred.
+
+The ``Variant.Obj`` property returns a C# ``object`` with the correct value for any variant. This
+may be useful when the type of Variant is completely unknown. However, when possible, prefer more
+specific conversions. ``Variant.Obj`` evaluates a ``switch`` on ``Variant.VariantType`` and it may
+not be necessary. Also, if the result is a value type, it may be boxed when it normally wouldn't be.
```
raulsntos, line 65: It will _always_ be boxed. That's the only way to return a value type as a `System.Object`.

raulsntos (approved):
I think it looks great already, thank you for improving the C# Variant documentation. I only left a minor nit-pick.

tutorials/scripting/c_sharp/c_sharp_variant.rst
31: For some reason what I had in mind here was `object o = v.As<Vector3>();`, where `.Obj` wouldn't make a difference to how much boxing ends up happening total. But at best, this sentence is ambiguous--sticking to the direct fact here is definitely better, good spot. 😄

mhilbrunner (approved):

mhilbrunner:
Whoops, should have squashed. Anyway, merged! Thanks, and thanks for the reviews everyone.

mhilbrunner:

PR #8894, opened 2024-02-04
Link to C# Variant-compatible section directly

31:
* After https://github.com/godotengine/godot-docs/pull/8878, the C# variant page's first section is longer. Now the variant compatibility rules aren't necessarily visible when you get there, which could be confusing/annoying.

Added a reference to the variant compatibility rules and linked to it from the top of the page, to start with. Then, I looked for references to `doc_c_sharp_variant` that seem like they should probably directly point at the variant compatibility section. And I added a brief description to the section to establish some context when you click a direct reference to that section.

In most cases, `Variant` isn't actually being used--the compatibility rules just have consequences all over the place.

I was tempted to change the section name from `Variant-compatible types` to `Variant-compatible type` to avoid a lot of manual reference naming, but resisted it to avoid unnecessary churn and external link breakage.

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_collections.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_collections.rst
@@ -46,8 +46,8 @@ To choose which collection type to use for each situation, consider the followin
 * Does your collection need to interact with the Godot engine?
   (e.g.: the type of an exported property, calling a Godot method).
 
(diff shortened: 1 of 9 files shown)


raulsntos (approved):
Makes sense to me. Thank you!

mhilbrunner (approved):

mhilbrunner:
Thanks! Merged.

mhilbrunner:

PR #8900, opened 2024-02-04
Remove duplicate information in autoload page

### LINKED ISSUE #8312: Confusing wording on Autoload / Singleton article
Godot 4.1.2 under Windows, internal Godot editor, although unrelated.

The Singleton / Autoload page in the docs https://docs.godotengine.org/en/stable/tutorials/scripting/singletons_autoload.html , at the "This means that any node can access a singleton named "PlayerVariables" with:" passage and following text encourages users to put their singletons in a variable to access them, as if its a best practice, which ends up blocking Autocomplete and other Godot internal features.

The article DOES explain right after it what you are actually supposed to do (leave the Singleton enabled).

My suggestion here is certainly to reverse the order of the explanations, maybe mentioning the Singleton starts enabled by default and can be accessed like a global object, then explaining to place it in a variable if its not

skyace65:
Removes duplicate info which made the page confusing to read. Closes #8312.

### DIFF
--- a/tutorials/scripting/singletons_autoload.rst
+++ b/tutorials/scripting/singletons_autoload.rst
@@ -73,21 +73,8 @@ in top-to-bottom order.
 
 .. image:: img/autoload_example.webp
 
-This means that any node can access a singleton named "PlayerVariables" with:
-
-.. tabs::
- .. code-tab:: gdscript GDScript
-
-   var player_vars = get_node("/root/PlayerVariables")
-   player_vars.health -= 10
-
- .. code-tab:: csharp
-
-    var playerVariables = GetNode<PlayerVariables>("/root/PlayerVariables");
-    playerVariables.Health -= 10; // Instance field.
-
 If the **Enable** column is checked (which is the default), then the singleton can
-be accessed directly in GDScript, without requiring ``get_node()``:
+be accessed directly in GDScript:
 
 .. tabs::
  .. code-tab:: gdscript GDScript


tutorials/scripting/singletons_autoload.rst
```diff
@@ -86,14 +87,6 @@ This means that any node can access a singleton named "PlayerVariables" with:
...
     playerVariables.Health -= 10; // Instance field.
 
-If the **Enable** column is checked (which is the default), then the singleton can
-be accessed directly in GDScript, without requiring ``get_node()``:
-
-.. tabs::
- .. code-tab:: gdscript GDScript
-
-   PlayerVariables.health -= 10
-
```
Piralein, line 96: Isn't this the recommended example, which should remain and the other removed?

tutorials/scripting/singletons_autoload.rst
```diff
@@ -73,7 +73,8 @@ in top-to-bottom order.
 
 .. image:: img/autoload_example.webp
 
-This means that any node can access a singleton named "PlayerVariables" with:
+If the **Enable** column is checked (which is the default), then the singleton can
+be accessed directly in GDScript, without requiring ``get_node()``:
```
Piralein, line 77: If we remove the `get_node()` example, this could be rewritten to not mention the function at all.
Otherwise we should probably reverse it and add it as: "If you deactivate the `global variable` checkbox, the node can no longer be accessed by its name. Use the `get_node()` function to access it. + example". As the issue suggests.

skyace65:
Yeah I probably shouldn't have made this PR while tired. Updated it. Switched which code example is deleted and removed the reference to ``get_node()``.

Piralein (approved):
The C# sections still mentions the `GetNode()` with typecast. Maybe the example should just be moved below the:

> Note that autoload objects (scripts and/or scenes) are accessed just like any other node in the scene tree

as it contains both the GDscript and C# example. Either way solves the linked issue.

mhilbrunner (approved):

mhilbrunner:
Thanks!

mhilbrunner:
Cherrypicked to 4.2.

PR #8903, opened 2024-02-05
Fixes #4790: Loading a page with some google context breaks the CSS

### LINKED ISSUE #4790: Loading a page with some google context breaks the CSS
**Your Godot version:**
https://docs.godotengine.org/en/stable

**Issue description:**

When I load [this URL from google](https://docs.godotengine.org/en/stable/tutorials/misc/gles2_gles3_differences.html#:~:text=%22GLES2%22%20and%20%22GLES3%22,WebGL%202.0%20on%20the%20web.), which includes search parameters, the pages CSS is broken.

Screenshot 2021-03-29 at 14 52 59

When I load [the url with the context parameters stripped](, the page renders correctly

Screenshot 2021-03-29 at 14 53 38

[URL](https://docs.godotengine.org/en/stable/tutorials/misc/gles2_gles3_differences.html#:~:text=%22GLES2%22%20and%20%22GLES3%22,WebGL%202.0%20on%20the%20web.)

panreyes:
Fixes https://github.com/godotengine/godot-docs/issues/4790

This fixes the scroll horizontal position when accessing the documentation from a highlited URL, like the ones that Google search produces. 

Let me know if you think of any cases that the scroll.x resetting might be troublesome.

### DIFF
--- a/_static/js/custom.js
+++ b/_static/js/custom.js
@@ -487,6 +487,11 @@ $(document).ready(() => {
 
   // Giscus
   registerGiscus();
+
+  // Fixes "Loading a page with some google context breaks the CSS"
+  if (window.scrollX != 0) {
+    window.scrollTo(0, window.scrollY);
+  }
 });
 
 // Override the default implementation from doctools.js to avoid this behavior.


coppolaemilio:
Without really knowing too much about why this is happening, resetting the scroll like this looks a bit like a hack. I wasn't able to reproduce the code, but the hack doesn't seem too invasive. 

I would add a comment linking to the original issue because maybe the origin of the error is somewhere in our code, and that should be the thing to fix first instead of doing this hack, but as a temporary solution it could work.

mhilbrunner:
Thanks for contributing! For now, we went with https://github.com/godotengine/godot-docs/pull/9015, which hopefully will fix this. If not, we can reconsider this hack.

PR #8905, opened 2024-02-05
Update the Emscripten minimum requirements to 3.1.39

### LINKED ISSUE #8589: Wrong recommended Emscripten version for compiling web
**Your Godot version:**
Godot 4.2
**Issue description:**
The recommended version in the document is Emscripten 1.39.9+,which will cause compilation errors. Too newer version will cause a LTO error, but an older version (Within the recommended version range) will cause a GL_WORKAROUND_SAFARI_GETCONTEXT_BUG not found error.
image

https://docs.godotengine.org/en/latest/contributing/development/compiling/compiling_for_web.html

After my testing, Emscripten version 3.1.26 can perform compilation tasks correctly.

_Production edit: corrected LTO typo_

adamscott:
Fix #8589

Recommending 3.1.39 as this is our current battle-tested version.

### DIFF
--- a/contributing/development/compiling/compiling_for_web.rst
+++ b/contributing/development/compiling/compiling_for_web.rst
@@ -15,7 +15,7 @@ Requirements
 
 To compile export templates for the Web, the following is required:
 
-- `Emscripten 1.39.9+ <https://emscripten.org>`__.
+- `Emscripten 3.1.39+ <https://emscripten.org>`__.
 - `Python 3.6+ <https://www.python.org/>`__.
 - `SCons 3.0+ <https://scons.org/pages/download.html>`__ build system.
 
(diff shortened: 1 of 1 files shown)


akien-mga (approved):
Technically earlier versions can still work, but I don't know how far back.

So it's fine as is, but we might want to clarify this in case someone really needs to use e.g. 3.1.20 for some reason, which would still work.

I think the min requirement for GDExtension support is 3.1.14, but for users who don't care about dynamic loading it may work with older versions too.

skyace65 (approved):

skyace65:
Thanks!

mhilbrunner:
Cherrypicked to 4.2.

PR #8923, opened 2024-02-08
Clarify C# GlobalClass exported node behavior

31:
When following this link on the C# exports page:

> Custom node classes can also be used, see [C# global classes](https://docs.godotengine.org/en/latest/tutorials/scripting/c_sharp/c_sharp_global_classes.html#doc-c-sharp-global-classes).

This sentence in the intro paragraph seems relatively easy to misinterpret:

> ... [exported properties](https://docs.godotengine.org/en/latest/tutorials/scripting/c_sharp/c_sharp_exports.html#doc-c-sharp-exports) are restricted to instances of the global class or derived classes.

Someone took this to mean that exported custom node properties must be global classes (or derived from a global class). I hadn't read the intro to this doc yet when they brought it up, and reading it for the first time with this interpretation in mind (even though I was fairly sure it wasn't right), I also interpreted it that way.

After digging into it, I understand what it's trying to say, and I think it's worth clarifying. I added a bit more about how the restriction and filtering is helpful.

I retook the `img/globalclasses_addnode.webp` screenshot to get rid of some dead space that was making it harder to scroll through the page. (Seemed more noticeable once I added more content.)


tutorials/scripting/c_sharp/c_sharp_global_classes.rst
```diff
@@ -21,6 +26,35 @@ The ``MyNode`` type will be registered as a global class with the same name as t
...
+The inheritance hierarchy is used if a custom type isn't registered.
+
+.. code-block:: csharp
+
+    public partial class MySpecializedNode : MyNode
+    {
+    }
+
+Inspector assignments to an export of type ``MySpecializedNode`` are restricted
+to ``MyNode`` and derivative types. This is because ``MySpecializedNode`` isn't
```
paulloz, line 52: For consistency, we seem to generally use "derived type" and not "derivative types".
```suggestion
to ``MyNode`` and derived types. This is because ``MySpecializedNode`` isn't
```

tutorials/scripting/c_sharp/c_sharp_global_classes.rst
```diff
@@ -21,6 +26,35 @@ The ``MyNode`` type will be registered as a global class with the same name as t
...
+    public partial class Main : Node
+    {
+        [Export]
+        public MyNode MyNode { get; set; }
+    }
+
+
+.. image:: img/globalclasses_exportednode.webp
+
+The inheritance hierarchy is used if a custom type isn't registered.
```
paulloz, line 43: This sentence is a bit unclear IMHO.

paulloz (commented):
Thanks! It is way more understandable. It took me a couple of times re-reading the current page to understand what it meant.

Also, the new screenshot is a bit tiny, which is aesthetically unpleasing for me, but probably not a deal-breaker.

tutorials/scripting/c_sharp/c_sharp_exports.rst
```diff
@@ -293,7 +293,18 @@ Since Godot 4.0, nodes can be directly exported without having to use NodePaths.
...
 
-Custom node classes can also be used, see :ref:`doc_c_sharp_global_classes`.
+A specific type of node can also be directly exported. The list of nodes shown
+after pressing "Assign" in the inspector is filtered to the specified type, and
+only a correct node can be assigned.
+
+.. code-block:: csharp
+
+    [Export]
+    public Sprite Sprite { get; set; }
```
raulsntos, line 303: `Sprite` is not a real type, did you mean `Sprite2D`?

tutorials/scripting/c_sharp/c_sharp_global_classes.rst
31: Hmm, agreed. I'm thinking merging this with the `MySpecializedNode` paragraph is the right first step to making it clearer, I'll take a look.

31:
> the new screenshot is a bit tiny

On my screen (100% scaling, non-4k, FWIW) the font size seems the same as the existing images, which I was thinking is what matters. Just to make sure there's no rendering difference, here's what old (left) and new (right) look like to me:

image

And the unchanged screenshots:

image

I would agree the editor font used on this page overall is too small though.

Is there a contrib guide page that has something like "recommended screenshot settings"? I think having that on https://docs.godotengine.org/en/stable/contributing/documentation/docs_image_guidelines.html would be nice. (Maybe even with a note that it's ok to have mismatched sizes if a page's existing shots don't fit the guidelines.)

tutorials/scripting/c_sharp/c_sharp_global_classes.rst
```diff
@@ -21,6 +26,35 @@ The ``MyNode`` type will be registered as a global class with the same name as t
...
+
+.. code-block:: csharp
+
+    public partial class MySpecializedNode : MyNode
+    {
+    }
+
+Inspector assignments to an export of type ``MySpecializedNode`` are restricted
+to ``MyNode`` and derivative types. This is because ``MySpecializedNode`` isn't
+registered, but its parent, ``MyNode``, is.
```
31, line 53: I tried this again, and it doesn't behave like I wrote it here. It only restricts to `Node`. 😕 I'm not sure if the behavior actually changed, if I just read my results wrong on my first attempt, or if something's subtly different in this attempt.

It seems the most specific built-in node type is used in this case, too.

31:
Updated with fixes from review. I was able to simplify the talk about the inheritance hierarchy because I seem to have been wrong about how that works anyway. Also, updated all the screenshots to use the default font size rather than a smaller one. (Wasn't able to find a StatsIcon.svg so made a new super simple one.)

paulloz (approved):
It's OK for me. Thanks ❤️

raulsntos (approved):
Looks good to me too. And thanks for the new icon 🙏.

skyace65 (approved):

skyace65:
Thanks!

PR #8928, opened 2024-02-08
Add a paragraph about self

elenakrittik:
Adds a paragraph about the ``self`` keyword and documents that it can be used to refer to variables
defined in subclasses of current class.

Closes godotengine/godot#87944


tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1454,6 +1454,32 @@ Here are some examples of expressions::
 Identifiers, attributes, and subscripts are valid assignment targets. Other expressions cannot be on the left side of
 an assignment.
 
+self
+""""
+
+``self`` can be used to refer to the current instance and is often equivalent to directly referring to symbols available in
+current script, however, while may seem surprising, it also allows you to access properties, methods and other names in
```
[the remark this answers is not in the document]
elenakrittik: Thanks for fixes! Not sure what do you mean by "incomplete sentence"? It continues (and ends) on the line following the last line you quoted.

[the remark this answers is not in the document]
AThousandShips: "while may" is an incomplete sentence, you need an object there or it's not a valid sentence :)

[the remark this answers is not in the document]
AThousandShips: Meant that the sentence *without* `it` was incomplete, that's what I was correcting :)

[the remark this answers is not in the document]
elenakrittik: I guess i'm a "little" worse at English than you are.. I really don't see anything wrong with this sentence, but to avoid this conversation turning into a debate over English grammar i just removed the problematic wording. Hopefully it's correct now 😅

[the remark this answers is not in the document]
AThousandShips: It was good with my suggestion :) but no worries it looks better this way

dalexeev:
Thanks for the PR! This is greatly appreciated as it is one of the oldest undocumented features of GDScript. However, in my opinion, in the current edition the emphasis is shifted a little in a wrong direction. This feature is not intended solely for inheritance, as the reader might think. I think `member` implies a static context, and `self.member` implies a dynamic context. For example, you can declare **in the same class** dynamic properties using `_set()` and `_get()`, and to access them you can use `set("member", value)` and `get("member")` or `self.member`, but you can't use just `member`.

In GDScript there is no `_call()`/`_invoke()` method for dynamically defining methods, so the only example is defining a method in a descendant class. Note that often considered bad practice, the base class should not know about of its descendants. It is generally recommended to define an empty ("abstract" or "virtual") method in the base class. But for the purposes of illustrating the `self` syntax features, we could use an example like:

```gdscript
if has_method("test"):
    test() # Error.
    self.test() # OK.
    call("test") # OK.
```

tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1454,6 +1454,31 @@ Here are some examples of expressions::
 Identifiers, attributes, and subscripts are valid assignment targets. Other expressions cannot be on the left side of
 an assignment.
 
+self
+^^^^
+
+``self`` can be used to refer to the current instance and is often equivalent to directly referring to symbols available in
+the current script, however, it also allows you to access properties, methods, and other names in subclasses of the current script.
```
dalexeev, line 1461: You probably meant "subtypes". "Subclasses" can be confused with inner classes. It is better to use the term "inherited class", "child class" or "descendant class".

tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1454,6 +1454,31 @@ Here are some examples of expressions::
...
+
+::
+
+    class_name Weapon extends Node
+
+    func _ready() -> void:
+        print(self.damage)
+
+``damage`` is not defined in ``Weapon``, but we (through, e.g., :ref:`documenting our code <doc_gdscript_documentation_comments>`)
+can assume that it will exist in subclasses of ``Weapon``. For example:
```
dalexeev, line 1471: I think mentioning doc comments is unnecessary here. You cannot document non-existent members, except by mentioning them in the class description.

tutorials/scripting/gdscript/gdscript_basics.rst
elenakrittik: You're right, sorry! It's probably my Python bias again :p

elenakrittik:
@dalexeev I (hopefully) addressed most of your concerns in dead3eb9a, but i am still not sure how to approach the examples. I have something like this in mind:

```gdscript
class_name Item extends Node

# Returns the expected size an item will take up in
# player's inventory (or -1 if it cannot be collected).
func get_size() -> int:
	var size = get(&"size")

	return size if size else -1

# When player touches an item, collect it if it is
# collectible (i.e. has a `collect` method).
func on_player_touch() -> void:
	if has_method(&"collect"):
		# collect() # Would be an error!
		# self.collect() # @dalexeev This is also appears to be an error
						 # on latest 4.2, perhaps you misremembered this part?
		call(&"collect")

# .. another file ..

class_name Potion extends Item

var size := 2

func collect() -> void:
	print("Collected a potion!")
```

But it looks pretty verbose and (in my view) not immediately understandable without additional comments. Do you have any better ideas?

tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1454,6 +1454,32 @@ Here are some examples of expressions::
 Identifiers, attributes, and subscripts are valid assignment targets. Other expressions cannot be on the left side of
 an assignment.
 
+self
+^^^^
+
+``self`` can be used to refer to the current instance and is often equivalent to directly referring to symbols available in
+the current script, however, while this is generally considered a bad practice, ``self`` also allows you to access properties,
+methods, and other names that are defined dynamically (i.e. are expected to exist in subtypes of the current script, or are
```
dalexeev, line 1462: Accessing dynamically defined properties is not a bad practice, it's a different approach, a different paradigm. But accessing members of child classes in the base class can be considered bad practice.

Also "in the current script" can be confusing, since there are inner classes, I would prefer "in the current class".

tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1454,6 +1454,32 @@ Here are some examples of expressions::
...
+    class_name Weapon extends Node
+
+    func _ready() -> void:
+        print(self.damage)
+
+``damage`` is not defined in ``Weapon``, but using ``self`` we can assume that it will exist at runtime. For example:
+
+::
+
+    class_name Pistol extends Weapon
```
dalexeev, line 1476: In my opinion, it would be better to separate the example with dynamic properties and inheritance, this section may be difficult for a beginner to understand.

tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1454,6 +1454,32 @@ Here are some examples of expressions::
...
+::
+
+    class_name Weapon extends Node
+
+    func _ready() -> void:
+        print(self.damage)
+
+``damage`` is not defined in ``Weapon``, but using ``self`` we can assume that it will exist at runtime. For example:
+
+::
```
dalexeev, line 1474:
```suggestion
``damage`` is not defined in ``Weapon``, but using ``self`` we can assume that it will exist at runtime. For example::
```

tutorials/scripting/gdscript/gdscript_basics.rst
elenakrittik: There are examples in `Object._get()`, `Object._set()` and related methods' docs already. I don't think it would be beneficial to duplicate them here?

tutorials/scripting/gdscript/gdscript_basics.rst
elenakrittik: Done in 22b1ea7db!

elenakrittik:
@dalexeev As i haven't heard anything negative towards the proposed example, i went ahead and changed the weapon/pistol example with this new one. I also added a `.. warning::`` section, and would like to hear your opinion on it, especially the second paragraph (as i am not an expert on GDScript and what is considered the best solution for each problem).

tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1454,6 +1454,55 @@ Here are some examples of expressions::
...
+    func get_size() -> int:
+        var size = get(&"size")
+
+        return size if size else -1
+
+    # When player touches an item, collect it if it is collectible (i.e. has a `collect` method).
+    func on_player_touch() -> void:
+        if has_method(&"collect"):
+            # collect() # Would be an error!
+            # self.collect() # An error, too! `self` only bypasses property checks.
```
dalexeev, line 1478: This should not be an error unless you have set `UNSAFE_METHOD_ACCESS` to Error.

tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1454,6 +1454,55 @@ Here are some examples of expressions::
 Identifiers, attributes, and subscripts are valid assignment targets. Other expressions cannot be on the left side of
 an assignment.
 
+self
+^^^^
+
+``self`` can be used to refer to the current instance and is often equivalent to directly referring to symbols available in
+the current script, however, ``self`` also allows you to access properties, methods, and other names that are defined dynamically
```
RedMser, line 1461:
```suggestion
the current script. However, ``self`` also allows you to access properties, methods, and other names that are defined dynamically
```

Calinou:
@elenakrittik Could you look into replying to the above suggestions (or applying them directly) so this PR can be merged? Thanks in advance 🙂

elenakrittik:
Much sorry, i'll apply the automatic change right away but i unfortunately will not be able to address dalexeev's feedback until after a week or so. Once i'm back i'll make sure to drive this to the finish line!

elenakrittik:
Sorry for the long wait! I believe everything should be good now.

dalexeev:
* See also https://github.com/godotengine/godot/pull/100237.

```gdscript
set("not_exist", 123) # No error or warning.
self["not_exist"] = 123 # Runtime error.
self.not_exist = 123 # Runtime error.
not_exist = 123 # Compile time error.

print(get("not_exist")) # No error or warning.
print(self["not_exist"]) # Runtime error.
print(self.not_exist) # Runtime error.
print(not_exist) # Compile time error.
```

tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -180,7 +180,7 @@ in case you want to take a look under the hood.
 +------------+---------------------------------------------------------------------------------------------------------------------------------------------------+
 | as         | Cast the value to a given type if possible.                                                                                                       |
 +------------+---------------------------------------------------------------------------------------------------------------------------------------------------+
-| self       | Refers to current class instance.                                                                                                                 |
+| self       | See `self`_.                                                                                                                                      |
```
dalexeev, line 183:
```suggestion
| self       | Refers to current class instance. See `self`_.                                                                                                    |
```

tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1530,6 +1530,36 @@ Here are some examples of expressions::
...
+        my_func()
+        # Checked at runtime, thus may work for descendant classes.
+        self.my_func()
+
+.. warning::
+
+    Beware that accessing members of child classes in the base class is often considered a bad practice because this
+    makes the relationships between parts of your game's code harder to reason about. It is easy to forget to define
+    a variable that the base class expected or forget about an invariant that the base class assumed to be true when
+    implementing a function needed by the base class.
```
dalexeev, line 1561: The first sentence looks good, but I doubt the second one is a good reasoning/illustration. I would suggest removing it or replacing it to take the emphasis off "easy to forget".

I don't think it's considered bad practice because it's easy to forget to add a required member when inheriting. In my opinion, the main problem is that the class is accessing non-existent properties/methods that are outside its area of ​​responsibility. Subclasses are in a sense undefined until the base class is fully defined, so it shouldn't know about their existence.

If you want to use abstract methods (which currently don't exist in GDScript), it's better to define a "prototype" with an empty body or an assertion.

elenakrittik: Makes perfect sense, i included your reasoning into the `:: warning` section!

dalexeev (approved):
Looks good to me.

tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1530,6 +1530,36 @@ Here are some examples of expressions::
 Identifiers, attributes, and subscripts are valid assignment targets. Other expressions cannot be on the left side of
 an assignment.
 
+self
+^^^^
```
tetrapod00, line 1534:
```suggestion
~~~~
```
Maybe promote the header by one level? I don't see how `self` is semantically a subsection of `Expressions`. 

I'm also not sure if this is exactly the right section of the page to put this - is `Statements and control flow` really more fitting than either `Classes` or `Properties (setters and getters)`? 

However, since I'm not sure *exactly* where the best place to put it is, it is okay to merge this PR with the new section where it is, and consider moving it in a followup PR.

tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1530,6 +1530,36 @@ Here are some examples of expressions::
...
+
+``self`` can be used to refer to the current instance and is often equivalent to directly referring to symbols available in
+the current script. However, ``self`` also allows you to access properties, methods, and other names that are defined dynamically
+(i.e. are expected to exist in subtypes of the current class, or are provided using :ref:`_set() <class_Object_private_method__set>`
+and/or :ref:`_get() <class_Object_private_method__get>`).
```
tetrapod00, line 1539:
```suggestion
``self`` can be used to refer to the current instance and is often equivalent to
directly referring to symbols available in the current script. However, ``self``
also allows you to access properties, methods, and other names that are defined
dynamically (i.e. are expected to exist in subtypes of the current class, or are
provided using :ref:`_set() <class_Object_private_method__set>` and/or
:ref:`_get() <class_Object_private_method__get>`).
```
Wrap to 80 characters. (See [here](https://docs.godotengine.org/en/latest/contributing/documentation/docs_writing_guidelines.html#manually-wrapping-lines) for details)

tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1530,6 +1530,36 @@ Here are some examples of expressions::
...
+        my_func()
+        # Checked at runtime, thus may work for descendant classes.
+        self.my_func()
+
+.. warning::
+
+    Beware that accessing members of child classes in the base class is often considered a bad practice,
+    because this blurs the area of responsibility of any given piece of code, making the overall relationship
+    between parts of your game harder to reason about. Besides that, one can simply forget that the parent
+    class had some expectations about it's descendants.
```
tetrapod00, line 1561:
```suggestion
    Beware that accessing members of child classes in the base class is often
    considered a bad practice, because this blurs the area of responsibility of
    any given piece of code, making the overall relationship between parts of
    your game harder to reason about. Besides that, one can simply forget that
    the parent class had some expectations about it's descendants.
```
Wrap to 80 characters

tetrapod00 (commented):
If dalexeev approves it's good.

I suggested a couple changes to better follow the style guide.

tutorials/scripting/gdscript/gdscript_basics.rst
elenakrittik: If we ignore the specifics of how the engine treats `self` (most notably, for the engine it's a keyword and not a "name"/identifier), `self` is just another variable that refers to the current class instance, so, personally, it makes sense to me to display `self` as a special-case of an identifier expression.

If we assume that `self` does not indeed belong to `Expressions` and we should move it somewhere else, i also can't find a fitting place for it. Moving it one level up is definitely not a solution (`self` is not a statement (||unless you just write consisting of a standalone `self`, but that's a technicality||)). The best i can find is the `Classes` section, however, given it's current content, i struggle to see `self` there without rewriting parts or all of the section.

tutorials/scripting/gdscript/gdscript_basics.rst
elenakrittik: Right, sorry!

tutorials/scripting/gdscript/gdscript_basics.rst
tetrapod00: Okay, that reasoning makes sense. If there's no other obvious place to put it, it's good here.

tetrapod00 (approved):

elenakrittik:
Is there someone "final" i should request a review from, or does the "review queue" move automatically?

tetrapod00:
It's pretty informal for the docs repo, we don't have as much of a process as the main repo does. In most cases PRs are good to merge once it gets an approval from a subject matter expert (dalexeev in this case) and a general docs reviewer (me in this case). Sometimes we merge things with less approvals than that, even, if it's a very obvious change.

This is good to merge I just didn't get around to it yet.

tetrapod00:
Merged! Thank you.

PR #8934, opened 2024-02-09
Clarify which track is being talked about

### LINKED ISSUE #8141: Animation tracks missing Track Settings panel (4.1)
4.1.stable.mono

The documentation on the tutorial [Introduction to the animation features](https://docs.godotengine.org/en/stable/tutorials/animation/introduction.html) mentions a track Settings panel, which does not exist in the Animation tab of the Godot 4.1 editor.

As an example, I have a basic Elevator scene containing a `CollisionShape3D` and two meshes (`Platform` and `Shroud`), each of which has properties animated in the `Ascend` and `Descend` animations (which are separate to allow configurable delays with corresponding looping animations to play in between). There are no children of the `AnimationPlayer` node in the Scene Tree; the meshes and collision shape are sibling nodes. The `Ascend` animation is shown here, with the tracks occupying the full width of the panel:

image

I cannot find where such settings have been moved to, even when toggling into the bezier curve editor. No such settings are visible in the Inspector panel, even when selecting Animation -> Open in Inspector.

skyace65:
Clarifies that the documentation is referring to property tracks specifically. While this section is part of a tutorial I can understand how people could get the misconception that this is talking about all track types. Closes #8141.

### DIFF
--- a/tutorials/animation/introduction.rst
+++ b/tutorials/animation/introduction.rst
@@ -240,7 +240,7 @@ in the next chapter.
 Track settings
 ~~~~~~~~~~~~~~
 
-Each track has a settings panel at the end, where you can set its update
+Each property track has a settings panel at the end, where you can set its update
 mode, track interpolation, and loop mode.
 
 .. figure:: img/animation_track_settings.webp


AThousandShips (approved):

AThousandShips:
Thank you!

mhilbrunner:

PR #8949, opened 2024-02-12
Add missing class definition whitespace example in style guide

sydist:
It is stated that both class and function definitions should be surrounded by two blank lines, but only functions are shown. This leaves some ambiguity and could be missed by skimming users.


AThousandShips:

If you need more information about the update process please see: [here](https://docs.godotengine.org/en/latest/contributing/workflow/pr_workflow.html#rebasing-onto-another-branch). If wanted I can fix the branch for you.

Issue #8950, opened 2024-02-12
Clarify suffix usage when importing resources with ResourceImporterScene


emre0altan:
**Your Godot version: 4.2.1-stable**

**Issue description: Suffix usage when importing resources with ResourceImporterScene is not clear**

**URL to the documentation page: https://docs.godotengine.org/en/stable/tutorials/assets_pipeline/importing_3d_scenes/node_type_customization.html#create-navigation-navmesh**

Explained suffix on the page is not clear that it works with "$" (case sensitive), "-" (case insensitive) and "_" (case insensitive). This issue came up here: https://github.com/godotengine/godot/issues/87377. I'll create a PR for this.

skyace65:
Fixed by #8951.

PR #8951, opened 2024-02-12
Node type customization warning updated to explain suffix options #8950

emre0altan:
Suffix options for resource importing in 3d scenes can actually be used with multiple symbols with different case sensitivity. Pointed in this issue: https://github.com/godotengine/godot-docs/issues/8950

### DIFF
--- a/tutorials/assets_pipeline/importing_3d_scenes/node_type_customization.rst
+++ b/tutorials/assets_pipeline/importing_3d_scenes/node_type_customization.rst
@@ -17,7 +17,8 @@ will detect suffixes in object names and will perform actions automatically.
 
 .. warning::
 
-    All the suffixes described below are **case-sensitive**.
+    All the suffixes described below can be used with ``-``, ``$``, and ``_`` and are
+	**case-insensitive**.
 
 Remove nodes (-noimp)
 ^^^^^^^^^^^^^^^^^^^^^


skyace65:
Do you have a test 3D model you could share for this? I tested out the behavior you described and everything was imported correctly with all 3 suffixes regardless of case. Here's a picture of a test glb file I imported to Godot 4.2.1: 
ImportTest

emre0altan:
Thanks for testing. It's my mistake not testing "$" symbol. Case conversion for it happens differently. Other two symbols are lower cased and then checked, but "$" is used with `findn()` which I realised is a case-insensitive version of `find()`.

https://github.com/godotengine/godot/blob/b4e2a24c1f62088b3f7ce0197afc90832fc25009/editor/import/3d/resource_importer_scene.cpp#L321-L331

So, I assume components are being added and should I just write down all three symbols are case-insensitive?

skyace65:
Yes, if you're seeing that everything is case insensitive for you as well update the PR. When you do please keep your PR to 1 commit, if you don't know how to do that we have a guide [here](https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html#modifying-a-pull-request).

emre0altan:
Thanks, updated now.

tutorials/assets_pipeline/importing_3d_scenes/node_type_customization.rst
```diff
@@ -17,7 +17,8 @@ will detect suffixes in object names and will perform actions automatically.
 
 .. warning::
 
-    All the suffixes described below are **case-sensitive**.
+    All the suffixes described below can be used with ``$``, ``-`` and ``_`` as
```
skyace65, line 20:
```suggestion
    All the suffixes described below can be used with ``$``, ``-`` and ``_`` and are
```

One minor critique.

skyace65 (changes requested):

tutorials/assets_pipeline/importing_3d_scenes/node_type_customization.rst
```diff
@@ -17,7 +17,8 @@ will detect suffixes in object names and will perform actions automatically.
 
 .. warning::
 
-    All the suffixes described below are **case-sensitive**.
+    All the suffixes described below can be used with ``$``, ``-`` and ``_`` and are
```
AThousandShips, line 20:
```suggestion
    All the suffixes described below can be used with ``$`` and ``_`` instead of ``-`` and are 
```
Or
```suggestion
    All the suffixes described below can be used with ``$``, ``-``, and ``_`` and are 
```
If original form is kept

AThousandShips, line 20:
```suggestion
    All the suffixes described below can be used with ``$``, ``-``, and ``_`` and are 
```
Please use [Oxford comma](https://docs.godotengine.org/en/latest/contributing/documentation/docs_writing_guidelines.html#use-the-oxford-comma-to-enumerate-anything).

skyace65 (approved):

skyace65:
Thanks! Congrats on your first merged PR! Also just so you know github has a feature where if you write "closes #issue number" in your PR description it will automatically close that issue when the PR is merged.

emre0altan:
Good to know, thank you!

PR #8953, opened 2024-02-13
Update method name in object_class.rst

thiagola92:
`register_virtual_class()` changed to `register_abstract_class()` in [2022](https://github.com/godotengine/godot/commit/6f51eca1e3045571ccc68414a922e8b0229111f0#diff-f94cd5be526d3d95f0e3e7dd0f2c7a6c6205ad2f94aa96ef4165c2bb0051938cL172-R174)  

Documentation page: https://docs.godotengine.org/en/stable/contributing/development/core_and_modules/object_class.html#registering-an-object

### DIFF
--- a/contributing/development/core_and_modules/object_class.rst
+++ b/contributing/development/core_and_modules/object_class.rst
@@ -57,7 +57,7 @@ Registering as virtual is the same but it can't be instanced.
 
 .. code-block:: cpp
 
-    ClassDB::register_virtual_class<MyCustomClass>()
+    ClassDB::register_abstract_class<MyCustomClass>()
 
 Object-derived classes can override the static function
 ``static void _bind_methods()``. When one class is registered, this


contributing/development/core_and_modules/object_class.rst
```diff
@@ -57,7 +57,7 @@ Registering as virtual is the same but it can't be instanced.
 
 .. code-block:: cpp
 
-    ClassDB::register_virtual_class<MyCustomClass>()
+    ClassDB::register_abstract_class<MyCustomClass>()
```
AThousandShips, line 60:
```suggestion
    ClassDB::register_class<MyCustomClass>(true)
```
This is how you register a virtual class, abstract is different, and if we should mention it we should have it talk about abstract classes instead IMO

thiagola92: I shouldn't have assume that it only changed the name, my bad.

AThousandShips: These two should probably be replaced with the better macros:
```cpp
GDREGISTER_CLASS(MyCustomClass);
GDREGISTER_VIRTUAL_CLASS(MyCustomClass);
```

PR #8961, opened 2024-02-14
Add note on enabling driver input in recording_with_microphone.rst to allow mic recording

Skyvastern:
I've written a note that tells user to enable the option `Audio -> Driver -> Enable Input` in `Project Settings`. This is disabled by **default** (in the example project provided in doc its enabled). So without this enabled in new projects, microphone recording won't work.


AThousandShips:
Please open it on the `master` branch as instructed

PR #8836, opened 2024-01-26
multiple_resolutions.rst: Remove confusing stretch.png graphic

jpaugh:
Remove confusing stretch.png graphic, which does not correspond to the available 4.x stretch modes, nor is it referred to by any of the surrounding text. In fact, the actual stretch modes behave quite differently than what this graphic implies.

PR #8970, opened 2024-02-15
Update stretch image label

### LINKED ISSUE #8964: Image in stretch mode docs may be out of date
**Your Godot version:** 4.2

**Issue description:** The first image in [the Stretch Mode section](https://docs.godotengine.org/en/stable/tutorials/rendering/multiple_resolutions.html#stretch-mode) of the Multiple Resolutions page lists the three options as "Disabled," "2D," and "Viewport." It's my first day using Godot, but it appears that the "2D" setting was either renamed to "Canvas Items" or replaced with this option.

If this was just a rename, I'd be happy to open a PR that updates the image to have the second option with the correct label. However, since I'm so new to Godot I'm not actually sure if "Canvas Items" behaves the same as "2D" did; if they're different, then updating the visual representation of the option may be more appropriate than simply changing the label in the image.

I'm trying to learn Godot and get into contributing more to the docs, so if anyone could point me in the direction of the best fix for this, I'd love to open a PR. Thanks so much!

**URL to the documentation page:** https://docs.godotengine.org/en/stable/tutorials/rendering/multiple_resolutions.html#stretch-mode

jamesplease:
Resolves #8964 

----

This updates the SVG source to reflect the new name of this stretch mode option. I noticed that there's also a PNG next to this image. I wasn't sure if that png gets updated as a part of the build step, or if there's another system in place to generate that PNG from the SVG.

Lmk if there's a specific process you'd like me to do to create that PNG. Otherwise I can just export it from Figma really fast and update this PR.

Note: I also made a slight tweak to the y positioning of the 3 labels so that they're vertically aligned with one another.


AThousandShips:
It should preferably be exported as `webp`, you can read more about it in the [Doc image guidelines](https://docs.godotengine.org/en/stable/contributing/documentation/docs_image_guidelines.html#format-conversion).

AThousandShips:
See also:
* https://github.com/godotengine/godot-docs/pull/8836

PR #8988, opened 2024-02-17
Fix function description for Font.get_char_size()

DaltonSW:
Removed the reference to past functionality involving "passing the next character to account for kerning"


AThousandShips:

PR #8993, opened 2024-02-19
Update 02.player_input.rst

MarsAstro:
Set :article_outdated to False.

I have followed the tutorial and found no discrepancies between the text or pictures and what I saw when I followed the tutorial using Godot 4.2.

I was told to commit to master last time, but I'm hoping it's right to commit to 4.2 given that this is specifically about the documentation for 4.2? I see that the master version has no :artice_outdated property, so I assume I don't need to target master as well as 4.2, as suggested in the description comments.

### DIFF
--- a/getting_started/first_3d_game/02.player_input.rst
+++ b/getting_started/first_3d_game/02.player_input.rst
@@ -1,4 +1,4 @@
-:article_outdated: True
+:article_outdated: False
 
 .. _doc_first_3d_game_player_scene_and_input:
 


MarsAstro:
> Unless this is _only_ for 4.2 this should be done on `master`, you should target only `master` if it applies to both

As far as I can tell, this only applies to 4.2. The property I changed (whether article is outdated or not) doesn't seem to exist in the file on the master branch, and I figured that was intentional since an article being outdated or not is entirely dependent on which engine version you've selected for the documentation.

AThousandShips:
If it's been updated in `master` then it'll be cherry picked for 4.2 at some point

Thank you for your contribution nonetheless :)

Issue #7348, opened 2023-05-14
Saving tutorial only mentions JSON


nlupugla:
**Your Godot version:**
4.0.1

**Issue description:**
The current tutorial on saving games  only mentions JSON for serializing save data. I understand that generally, we want to discourage users from using JSON as opposed to Godot's built in resource serialization as JSON is less efficient and harder to work with in the engine.

It would be great to have an article discussing the pros and cons of different serialization methods, as well as a few common "gotchas" that people might run in to or perhaps an FAQ section. For instance, why should or shouldn't I save my whole game as one giant packed scene? What security issues should I be aware of when loading game data from custom resources? When I load my game and something breaks, what strategies can I use to debug the problem?

I'm willing to take the lead on writing these docs, but I will need help as I haven't actually made a fully fledged save system in Godot yet and I'm still a new contributor. If you'd like to help out, feel free to @nlupugla in Rocket chat.

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/stable/tutorials/io/saving_games.html

Calinou:
It may also be worth mentioning that if you're saving replays, you should store transforms on a regular basis (every frame, or less often to reduce filesize) and not inputs as Godot's physics engine is not deterministic.

and-rad:
@nlupugla Here is the example project. I kept it minimalistic, but the savegame system has all the necessary features to be used in a simple game right now. You can ignore the `character` folder, it just contains a character controller to be able to move around the level. The important stuff happens in the `savegame` and `ui` folders, and in the game's main level.

I documented everything extensively, but you can always hit me up if things are unclear.

[SavegameExamples.zip](https://github.com/godotengine/godot-docs/files/11506706/SavegameExamples.zip)

eswartz:
FWIW, I came across this same docs issue and think it's a little more serious than it seems. 

Think about this from the point of view of a new Godot user. If you look at the Manual section for "File and data I/O", I think most people will come away with the conclusion that Godot provides support only for reading/writing 3rd party data ("Runtime file loading and saving") while users need to use JSON for custom purposes ("Saving games").

I understand the appeal of a completely standalone save game format which doesn't leak any details of one's scene tree or the like. But as a high-level entry in the manual, it does a disservice to those learning Godot not to mention up front that `Resource` and `PackedScene` can be extended for data serialization, since this allows you to store any other Resource-based object by extension. 

[...]

PR #8995, opened 2024-02-19
Explain how PackedScene can be used as a JSON alternative for savegames

eswartz:
This attempts to provide a (relatively) brief explanation of using `.tscn`/`.scn` and its caveats as mentioned in
https://github.com/godotengine/godot-docs/issues/7348. 

It does not cover Calinou's comment
(https://github.com/godotengine/godot-docs/issues/7348#issuecomment-1546957474) which is a more general issue not related to the
storage format.

### DIFF
--- a/tutorials/io/saving_games.rst
+++ b/tutorials/io/saving_games.rst
@@ -336,14 +336,14 @@ call when child objects are loaded. You will also need a way to link
 children to parents as the :ref:`NodePath
 <class_nodepath>` will likely be invalid.
 
-JSON vs binary serialization
-----------------------------
+JSON serialization alternatives
+-------------------------------
 
 For simple game state, JSON may work and it generates human-readable files that are easy to debug.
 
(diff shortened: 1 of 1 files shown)


dalexeev:
I think this is a bad alternative, both in terms of security (see gdquest-demos/godot-demos-2022#6) and compatibility with future versions of the game (you are tightly coupling _data_ to _implementation_ in the form of the scene tree structure).

eswartz:
As for the security issues, I won't argue about the dangers. I see (as hinted in your discussion thread) and verified that someone could change the few `ext_resource` references to the `.gd` files to point to different files entirely and it will run that node's `_init` code on load. Oof. Point taken!

As for tying the game state to save game data, the text acknowledges this and tries to make clear that it is a data container which you must explicitly convert back and forth at load and save time. Beyond that, compatibility issues arise in any serialization format. (Unfortunately, the original text makes no mention of this either.)

I'll leave the hot potato for someone else :)

PR #8998, opened 2024-02-20
Update class_sprite2d.rst

dustdfg:
Clarify that `is_pixel_opaque` function gets position in local coordinates

**DISCLAIMER:** I am sure for 99% it is true because I tried it and see that it use local coords but maybe I am just a fool...


AThousandShips:

Issue #8077, opened 2023-09-27
The "Your first/second 3D shader" tutorials are outdated and provide incorrect results


SaadTheGlad:
**Your Godot version:** 
v4.1.1

**Issue description:**
The tutorials on 3D shaders do not work properly and give unexpected results, especially when calculating normals

What I expected and what I recieved: 
![image](https://github.com/godotengine/godot-docs/assets/39242060/9bd2d89c-af4b-45fb-85c1-6e08adf24d07)

In general the entire tutorial is outdated and provides unexpected results in-engine and should be massively revised.

**URL to the documentation page:**
https://docs.godotengine.org/en/stable/tutorials/shaders/your_first_shader/your_second_3d_shader.html

If you know how to fix the issue you are reporting please
consider opening a pull request. We provide a tutorial on
using git here: https://docs.godotengine.org/en/stable/community/contributing/pr_workflow.html,
writing documentation at https://docs.godotengine.org/en/stable/community/contributing/docs_writing_guidelines.html
and contributing to the class reference here: https://docs.godotengine.org/en/stable/community/contributing/updating_the_class_reference.html

taterboom:
😥 I'm a beginner learning shaders, and it's truly frustrating that despite following the official tutorial, the results I achieved are different from what the tutorial demonstrates.

nathancarter:
Note:  The seams in the image I showed above come from assigning a noise texture without checking the "seamless" box.

nathancarter:
Note that a YouTuber has made a sophisticated water shader tutorial for Godot 4 and also shared a written version with project files here.  This may be a good place to start to figure out what needs updating.

https://stayathomedev.com/tutorials/single-plane-water-shader/

skyace65:
The first 3D shader tutorial was fixed with this PR #10554. The second shader still needs to be updated.

PR #9013, opened 2024-02-24
Marking tutorials out of date as per issue #8077

nathancarter:
Did this wrong the first time; I think this fixes it.


nathancarter:
I don't know why the checks have failed.  Is that something to do with the tiny tweaks I made?

AThousandShips:
The pages are incorrectly named, they're supposed to be "3d shaders", but the doc line above naming them is incorrect, as are crosslinks, your changes doesn't affect this but it causes the files to be re-processed

PR #8932, opened 2024-02-09
Replace "alt" with "opt" for macOS shortcuts

skyace65:
In the past Mac keyboards would have "alt" written on the option key in addition to "option", these days it's just "option". I'm using "Opt" instead of "Option" for these based on discussions with bruvzg in rocket chat. Closes #7959.

PR #9019, opened 2024-02-26
Fix macOS default mappings

edmundito:
I noticed that for the Mac editor, it was referring to the `Alt` key. However, Mac keyboards do not have an Alt key, and when I looked at the mappings in my Gotot editor (4.2.0), they were set to other combinations. I have not made any mapping changes in my editor, so I reviewed the mappings and made updates.

Additionally, I updated the order compared to Windows/Linux because in the mac keyboard keys as in the following order (the Godot Editor mapping follows this arrangement: `shift`, `fn`, `control`, `option`, `command`. Although Cmd tends to be analogous to Ctrl in other systems, when it comes to modifiers, they tend to come first. Instead of `Cmd + Ctrl`, on Mac it's `Ctrl + Cmd`.


skyace65:
There's already a PR open for this which covers more. #8932.

edmundito:
See the comment above. I will close this in favor of the earlier PR.

Issue #9009, opened 2024-02-24
CharacterBody2D move_and_slide example is using delta, despite warning not to use it


tograf:
**Your Godot version:**
4.2

**Issue description:**
The **move_and_sclide()** example for **CharacterBody2D** is using delta to calculate the y-Velocity. However the warning above the example is mentioning not to apply delta to the velocity vector, as it is done internaly. Is this an error or done on purpose?

**URL to the documentation page:**
https://docs.godotengine.org/en/stable/tutorials/physics/physics_introduction.html#doc-physics-introduction

skyace65:
If you remove the ``* delta`` or that whole line from the script it breaks movement. But the note was added in the same PR that added the ``* delta`` (#1176). So it's probably accurate somehow just doing a bad job of explaining what that is.

AThousandShips:
Gravity is an *acceleration* not a velocity, so using delta here is correct, we're updating the velocity based on the acceleration, if we were using some other method we'd actually apply delta twice to the gravity

This isn't a bug, but might need to clarify it

PR #9024, opened 2024-02-28
Removed delta in CharacterBody2D move_and_slide example

PyScrap789:
Issue #9009 is solved.


skyace65:
This does not fix that issue. Removing the delta multiplication breaks jumping. And if you remove the entire line gravity doesn't work. I appreciate that you want to contribute to the docs, but please test changes you're making to example code before opening a PR.

PR #9025, opened 2024-02-28
Update 03.player_movement_code.rst

The-oldone:
added a note to elevate the camera and added code to update the position of the player

### DIFF
--- a/getting_started/first_3d_game/03.player_movement_code.rst
+++ b/getting_started/first_3d_game/03.player_movement_code.rst
@@ -161,7 +161,7 @@ zero, which means the player is pressing a direction key.
 We compute the direction the ``$Pivot`` is looking by creating a :ref:`Basis <class_Basis>`
 that looks in the ``direction`` direction.
 
-Then, we update the velocity. We have to calculate the ground velocity and the
+Then, we update the velocity and position. We have to calculate the ground velocity and the
 fall speed separately. Be sure to go back one tab so the lines are inside the
 ``_physics_process()`` function but outside the condition we just wrote above.
 
@@ -183,6 +183,7 @@ fall speed separately. Be sure to go back one tab so the lines are inside the
 
         # Moving the Character
         velocity = target_velocity
+        position += delta * velocity
         move_and_slide()
 
  .. code-tab:: csharp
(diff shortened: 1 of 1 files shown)


getting_started/first_3d_game/03.player_movement_code.rst
```diff
@@ -183,6 +183,7 @@ fall speed separately. Be sure to go back one tab so the lines are inside the
 
         # Moving the Character
         velocity = target_velocity
+        position += delta * velocity
```
AThousandShips, line 186: This is not correct, this is what the `move_and_slide` method does, you shouldn't do this manually

getting_started/first_3d_game/03.player_movement_code.rst
```diff
@@ -392,7 +393,7 @@ Preview by clicking the checkbox.
 |image6|
 
 In the top view, make sure your *Camera3D* is selected and move the camera about
-``19`` units on the Z axis (drag the blue arrow).
+``19`` units on the Z axis (drag the blue arrow) then ``9`` units on the Y axis (drag the green arrow).
```
AThousandShips, line 396: This is also not correct, see right below, you *rotate* the camera so it lifts up like this

AThousandShips (changes requested):

getting_started/first_3d_game/03.player_movement_code.rst
The-oldone: i believe there is some confusion
i have downloaded the most recent version of godot(4.2) and move_and_slide doesn't do that
i have attached a file for clarity, could you let me know if i am accidentally using an earlier version?

https://github.com/godotengine/godot-docs/assets/139103754/f94632a0-f00a-499e-9231-72c091ef3d83

AThousandShips: I can't tell from just this video, however I can tell that your code isn't exact and you haven't copied the code exactly from the demo, I'd suggest testing the official demo and seeing if it works, and if you've made any errors copying the code:
https://github.com/godotengine/godot-demo-projects/tree/master/3d/squash_the_creeps

getting_started/first_3d_game/03.player_movement_code.rst
AThousandShips: Yes, then you are not following the instructions correctly, so please make sure you've followed *every* step exactly

AThousandShips:
I'd suggest that you ask in the [forums](https://forum.godotengine.org/) and find out what might be wrong with your project, it looks like you've failed to follow some instruction, if there is some issue with the tutorial you can come back, but the parts you've changed are not wrong and your code instead doesn't work correctly

PR #9035, opened 2024-02-29
Update export option

timkrief:
The Godot 3 --export option was changed to more explicit --export-release / --export-debug / --export-pack options.


AThousandShips:

You will need to make a new PR targeting `master`

However this page generally needs a rework, so this small change would be best done as part of that work instead of a small single change

PR #9052, opened 2024-03-05
Remove confusing sentence for changing scenes

### LINKED ISSUE #8846: 'Change scenes manually' page unclear
**Your Godot version:** 4.2.1.stable.mono

**Issue description:** Unclear reference in documentation

On the 'Change scenes manually' page, the first paragraph reads:

> Sometimes it helps to have more control over how one swaps scenes around. As mentioned above, a [:ref:`Viewport <class_Viewport>`](https://github.com/godotengine/godot-docs/blob/master/tutorials/scripting/change_scenes_manually.rst#id1)'s child nodes will render to the image it generates. This holds true even for nodes outside of the "current" scene. Autoloads fall into this category, but so do scenes which one instances and adds to the tree at runtime:

I found this page via google when searching for 'godot changing scenes'. I'm not clear what `As mentioned above` references to (The section above is 'Evaluating expressions' which, after a quick scan, doesn't seem to reference how to change scenes. Up a level is the Core features contents page. I'm guessing the documentation was restructured as some point and this soft reference was lost.

`Sometimes it helps to have more control over how one swaps scenes around` seems to suggest that the manual changing of scenes is an alternative to a different (maybe more basic but simpler?) approach. As a new user I'd like to know what that approach is, so I can evaluate the best choice for my situation.

**URL to the documentation page (if already existing):** https://github.com/godotengine/godot-docs/blob/master/tutorials/scripting/change_scenes_manually.rst // https://docs.godotengine.org/en/stable/tutorials/scripting/change_scenes_manually.html

skyace65:
Removes a sentence that references a page that doesn't exist, there's no mention of viewport on the previous page. I even checked the doc branch from when this page and sentence was added (3.1), and there was nothing about viewport in the previous page back then. Closes #8846.

### DIFF
--- a/tutorials/scripting/change_scenes_manually.rst
+++ b/tutorials/scripting/change_scenes_manually.rst
@@ -4,8 +4,8 @@ Change scenes manually
 ======================
 
 Sometimes it helps to have more control over how one swaps scenes around.
-As mentioned above, a :ref:`Viewport <class_Viewport>`'s child nodes
-will render to the image it generates. This holds true even for nodes outside
+A :ref:`Viewport <class_Viewport>`'s child nodes will render to the image
+it generates, this holds true even for nodes outside
 of the "current" scene. Autoloads fall into this category, but so do
 scenes which one instances and adds to the tree at runtime:
 


mhilbrunner (approved):

mhilbrunner:
Thanks! Merged.

mhilbrunner:
Cherrypicked to 4.2.

PR #9054, opened 2024-03-05
[Navigation] Document 3D support is required for navigation

AThousandShips:
Should *only* be cherry picked for 4.2 as this does not apply for 4.1

* Closes: https://github.com/godotengine/godot/issues/89185

### DIFF
--- a/contributing/development/compiling/optimizing_for_size.rst
+++ b/contributing/development/compiling/optimizing_for_size.rst
@@ -144,6 +144,11 @@ Tools must be disabled in order to use this flag, as the editor is not designed
 to operate without 3D support. Without it, the binary size can be reduced
 by about 15%.
 
+.. note::
+
+    Disabling 3D support also disables all navigation. This includes 2D navigation,
+    as it uses the 3D navigation system internally.
+
 Disabling advanced GUI objects
 ------------------------------
 


mhilbrunner (approved):

mhilbrunner:
It might be worth it to add a single sentence of *why* this is? The reasoning makes sense to me, but if you are new, this may come as a surprise and seem weird at first glance.

AThousandShips:
There really isn't a reason as such, see:
* https://github.com/godotengine/godot/pull/88681

It was done as the code depends on the 3D side and at the time it failed to compile because of missing guards, so not sure what it should say other than that it depends on the 3D side which isn't enabled

mhilbrunner:
Well, "because there is no 2D navigation, and everything uses 3D navigation under the hood" IMO makes (some amount of) sense once you know that, and experienced people will probably infer that from that note, but I could see newer people being confused. Maybe not, I guess :)

contributing/development/compiling/optimizing_for_size.rst
```diff
@@ -144,6 +144,10 @@ Tools must be disabled in order to use this flag, as the editor is not designed
 to operate without 3D support. Without it, the binary size can be reduced
 by about 15%.
 
+.. note::
+
+    Disabling 3D support also disables all navigation, including 2D navigation.
```
AThousandShips, line 149:
```suggestion
    Disabling 3D support also disables all navigation, including 2D navigation
    as it uses 3D navigation internally, there's no dedicated 2D system.
```
Maybe?

growingbrain: Probably better to replace 'it' with '2D navigation' to be complete clear.
    Disabling 3D support also disables all navigation, including 2D navigation
    since 2D navigation uses 3D navigation internally, there's no dedicated 2D system.

mhilbrunner (approved):

mhilbrunner:
Thanks, and thanks for making the change!

AThousandShips:
Thank *you*!

mhilbrunner:

PR #9074, opened 2024-03-10
Update signals.rst

LunarKitsune:
I am going through the tutorials to learn godot.

Upon hitting this tutorial I had to find where my issue was specifically on halting the processing of the 2d sprite.

Upon finding where c# was using functions for _on_button_pressed() with godot, the example used "OnButtonPressed()". Fixing it to the exact name property helped as I am finding that godot is case sensitive and wants the exact spelling just like the default godot programming. 

From that instance learning that I am commiting fixes to the timer function for C# as well.

It should be noted for the user if they need the exact name and casing for C#, (or other languages for that matter they can go into the node properties of an object, right click the connection, click edit, and then copy and paste the function name to avoid syntax errors that can break the program.

### DIFF
--- a/getting_started/step_by_step/signals.rst
+++ b/getting_started/step_by_step/signals.rst
@@ -156,7 +156,7 @@ the ``not`` keyword to invert the value.
 
  .. code-tab:: csharp C#
 
-    private void OnButtonPressed()
+    private void _on_button_pressed()
     {
         SetProcess(!IsProcessing());
     }
@@ -221,7 +221,7 @@ Your complete ``sprite_2d.gd`` code should look like the following.
             Position += velocity * (float)delta;
         }
 
-        private void OnButtonPressed()
+        private void _on_button_pressed()
         {
             SetProcess(!IsProcessing());
         }
@@ -326,7 +326,7 @@ bottom of our script and use it to toggle our sprite's visibility.
 
  .. code-tab:: csharp C#
 
-    private void OnTimerTimeout()
+    private void _on_timer_timeout()
     {
         Visible = !Visible;
     }
@@ -393,12 +393,12 @@ Here is the complete ``sprite_2d.gd`` file for reference.
             Position += velocity * (float)delta;
         }
 
-        private void OnButtonPressed()
+        private void _on_button_pressed()
         {
             SetProcess(!IsProcessing());
         }
 
-        private void OnTimerTimeout()
+        private void _on_timer_timeout()
         {
             Visible = !Visible;
         }


AThousandShips:
This is not correct, the naming standard for C# is correct in these files

CC @raulsntos

LunarKitsune:
> This is not correct, the naming standard for C# is correct in these files
> 
> As you can see if you read this note:
> 
> > By convention, we name these callback methods in GDScript as "_on_node_name_signal_name" and in C# as "OnNodeNameSignalName". Here, it'll be "_on_timer_timeout" for GDScript and OnTimerTimeout() for C#.
> 
> CC @raulsntos

Normally yes the naming convention would be correct, but when I use it instead of the Godot naming convention with C# the toggle button actually breaks, meaning it doesn't toggle the processing of the sprite on and off. 

I can reproduce the issue too. I am using godot 4.1 if that is what is making the diffference.

AThousandShips:
That sounds like either a bug or user error, not a fault in the documentation

getting_started/step_by_step/signals.rst
```diff
@@ -393,12 +393,12 @@ Here is the complete ``sprite_2d.gd`` file for reference.
...
         }
 
-        private void OnButtonPressed()
+        private void _on_button_pressed()
         {
             SetProcess(!IsProcessing());
         }
 
-        private void OnTimerTimeout()
+        private void _on_timer_timeout()
```
AThousandShips, line 401: This absolutely won't work as this is connected via script above, did you actually test this code?

getting_started/step_by_step/signals.rst
```diff
@@ -326,7 +326,7 @@ bottom of our script and use it to toggle our sprite's visibility.
 
  .. code-tab:: csharp C#
 
-    private void OnTimerTimeout()
+    private void _on_timer_timeout()
```
AThousandShips, line 329: Same for this, see on the example above

AThousandShips:
Id say that instead of making the changes for the button case (the timer case is not correct) there should be a note about naming for signals, or entering the name manually, there's settings for this to use

getting_started/step_by_step/signals.rst
LunarKitsune: changing the delegate and inserting method into the delegate did work for timer either way so I stand corrected on the timer. I do apologize I on that bit so the original did work on this.

The OnButtonPressed does seem to be bugged. I will look into whether I have a version issue, or some issue in the way my project is handled and comment little later.

getting_started/step_by_step/signals.rst
LunarKitsune: Okay I was very dumb. Please close this out. Following the tutorial I didn't look at the obvious of editing the reciever method.

PR #9076, opened 2024-03-10
Update CSG note with info on reliability

### LINKED ISSUE #8767: Document the CSG system's expected level of reliability
**Your Godot version:**
4.2.1

**Issue description:**
I think https://docs.godotengine.org/en/stable/tutorials/3d/csg_tools.html should mention that Godot's CSG can't reliably produce meshes without holes, and based on the conversation in this issue, my understanding is the Godot team thinks this isn't a totally fixable problem:

* https://github.com/godotengine/godot/issues/58637

> Accurate CSG algorithms (that handle coplanar faces correctly) are much slower, so I don't know if it's a viable tradeoff given many users are using those nodes at run-time.

I think the note at the top of the page is good ("mainly intended for prototyping"), but the second sentence seems to me to imply that the only reason that's the case is because of the missing features, not its reliability:

> The CSG nodes in Godot are mainly intended for prototyping. There is no built-in support for UV mapping or editing 3D polygons (though extruded 2D polygons can be used with the CSGPolygon3D node).

I think reliability should be mentioned directly.

Having to keep an eye out for holes and jiggling the nodes around to get it to work might work for some situations, but not others, and I think it would be good for this doc to inform people about this so they can more carefully evaluate whether it fits their needs. Otherwise, they could depend on it without thinking they need to evaluate how reliable it is, and end up in trouble later in development.

I filed a issue with some other CSG meshing issues (before finding the older issue) that made me think about suggesting this doc change: https://github.com/godotengine/godot/issues/86567

https://docs.godotengine.org/en/stable/tutorials/3d/csg_tools.html

skyace65:
CSG can't reliably create meshes made up of multiple nodes without holes in them. See discussion in https://github.com/godotengine/godot/issues/58637. Closes #8767.

### DIFF
--- a/tutorials/3d/csg_tools.rst
+++ b/tutorials/3d/csg_tools.rst
@@ -15,7 +15,9 @@ Interior environments can be created by using inverted primitives.
 
 .. note:: The CSG nodes in Godot are mainly intended for prototyping. There is
           no built-in support for UV mapping or editing 3D polygons (though
-          extruded 2D polygons can be used with the CSGPolygon3D node).
+          extruded 2D polygons can be used with the CSGPolygon3D node). In
+          addition CSG can't reliably create meshes made up of multiple nodes
+          without holes.
 
           If you're looking for an easy to use level design tool for a project,
           you may want to use `Qodot <https://github.com/QodotPlugin/Qodot>`__


mhilbrunner (approved):

mhilbrunner:
Thanks, merged.

mhilbrunner:

Issue #7728, opened 2023-07-26
Add support for embedding videos


Calinou:
We should use videos instead of animated GIFs, animated PNGs or animated WebPs for the following reasons:

- Better compression, which allows using higher resolutions and framerates.
- Support for playback controls.
- Support for audio if needed.

I have a WIP branch for this: https://github.com/Calinou/godot-docs/tree/add-video-support

However, this is blocked on https://github.com/sphinx-contrib/video/issues/33, as we'd have to include all videos in a top-level `_static` folder otherwise.

**Edit:** The above issue was resolved by sphinx-contrib-video 0.2.1.rc0.

PR #9092, opened 2024-03-13
Add support for embedding videos in documentation pages

Calinou:
This is more efficient[^1] than GIFs, animated PNGs or WebPs, while also providing controls and audio playback if needed.

Ideally, we'd ship AV1 versions of the videos for more optimized file sizes, but [macOS/iOS Safari only support these on very recent devices](https://caniuse.com/av1) (there's no CPU decoding fallback). Therefore, we'll settle on VP9 for now, which is supported on all platforms (including iOS) since iOS 17.4 was recently released.

[^1]: Not necessarily smaller in file size, but there's a potential to achieve much higher quality and smoother framerates. There's also more control over framerates compared to GIF – proper 30 FPS and 60 FPS are feasible with videos, not GIFs as they can only specify frame duration by 10 ms intervals.

A video has been committed to replace the introductory GIF in the *Prototyping levels with CSG* page.

- This closes https://github.com/godotengine/godot-docs/issues/7728.


skyace65:
Are there any codec issues with this on Linux? I know some patented ones got turned off by default last year in several distros.

EDIT: also how consistent is compression size for MP4? Because I know there is variation with images like webp (squoosh will get you a smaller lossless webp than krita).

Calinou:
> Are there any codec issues with this on Linux? I know some patented ones got turned off by default last year in several distros.

This is still an issue on Fedora out of the box, but pretty much everyone enables RPMFusion anyway as other places on the web will often use H.264 without any fallback. Other distributions don't have this issue.

Calinou:
It turns out VP9 is supported in iOS 17.4 and later, so we could actually use VP9 everywhere without a fallback: https://caniuse.com/webm

This means we can also support Linux distributions without RPMFusion enabled out of the box.

skyace65:
I'd say go for VP9 then.

Calinou:
> I'd say go for VP9 then.

Done. I've also updated the documentation image guidelines to add recommendations for creating videos for the documentation.

mhilbrunner (approved):

mhilbrunner:
The only thing I'm wary of here is adding another dependency, but it seems small and simple enough, hopefully. Let's merge and see if we run into any issues in practice on `master`/`4.3`.

Issue #9080, opened 2024-03-11
.


daveTheOldCoder:
.

aaronfranke:
> This method returns a time interval. It doesn't make sense to state "returns the time in UTC". A time interval is independent of any time zone.

No, it doesn't return a time interval, it returns the timestamp in UTC. A timestamp could technically be described as a time interval but it is the special case of the start being t=0 (1970-01-01 at 00:00:00) and the end being now. This method in particular always uses UTC for the timestamp.

You may be thinking of `get_ticks_msec` and `get_ticks_usec` which return the amount of time since the engine started.

> This sentence in the class description should also be removed or changed: "However, the get_unix_time_from_system method always returns the time in UTC."

No, that sentence is correct.

daveTheOldCoder:
>No, it doesn't return a time interval

Of course it returns a time interval. It's the interval between the Unix epoch and the current time.

> it returns the timestamp in UTC.

That makes no sense. Think about it. The number of seconds from the Unix epoch to the current time is the same regardless of the time zone.

Please re-read my original post.

aaronfranke:
> The number of seconds from the Unix epoch to the current time is the same regardless of the time zone.

No, because different timezones have the date/time 1970-01-01 00:00:00 occurring at different physical times.

This method specifically treats t=0 as 1970-01-01 00:00:00 in UTC. The other methods in the Time class are designed to work with any timezone, with t=0 as 1970-01-01 00:00:00 in your timezone of choice.

daveTheOldCoder:
>No, because different timezones have the date/time 1970-01-01 00:00:00 occurring at different physical times.

The Unix epoch is 1970-01-01 00:00:00 **UTC**. It's a fixed reference point in time, independent of time zones.

aaronfranke:
We agree that this method returns the timestamp in UTC, but not all timestamps are based on the Unix epoch. This method in particular is different from the rest of the methods because it works with UTC, while the others are designed to work with any timezone. That is why the description clarifies that this method works with UTC. It is not incorrect and not "doesn't make sense" to state that this method uses UTC, stating this is at worst redundant, at best extremely helpful.

daveTheOldCoder:
>It is not common or useful to refer to a full date and time as a time interval.

It's useful when it **is** a time interval.

I give up.

I know what the function does, in spite of the incorrect documentation. If you want to keep it that way, that's up to you.

PR #9093, opened 2024-03-13
Change description for get_unix_time_from_system for clarity regarding Unix Epoch and UTC

iltenahmet:
This fixes the issue #9080


AThousandShips:

PR #9095, opened 2024-03-13
Add commands for converting videos to animated WebP in Creating movies

Calinou:
This is useful when you need lossless-compressed output that can be viewed on the web, or when you need to upload an animation where only images are allowed (`.webp` is considered as an image by hosts, even if it's animated).

APNG and GIF can be created with similar methods, but these formats are more situational nowadays and the optimal way of creating a GIF is more involved in FFmpeg. (Ideally, [gifski](https://gif.ski/) should be used but it's an external program.)

### DIFF
--- a/tutorials/animation/creating_movies.rst
+++ b/tutorials/animation/creating_movies.rst
@@ -351,8 +351,10 @@ Some common post-processing steps are listed below.
 
 .. _doc_creating_movies_converting_avi:
 
-Converting AVI video to MP4
-^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Converting AVI video to another format
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+**MP4:**
 
 While some platforms such as YouTube support uploading the AVI file directly, many
 others will require a conversion step beforehand. `HandBrake <https://handbrake.fr/>`__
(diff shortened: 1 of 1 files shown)


skyace65:
Are there any codec issues with this on Linux? I know some patented ones got turned off by default last year in several distros.

mhilbrunner:
While useful, I'm not sure whether this is striving a bit outside the scope of *Godot* documentation. We're not `ffmpeg`, and video converting is really at the fringe of being Godot related. Though I guess this command should be expected to stay stable, so its unlikely we need to keep this page updated with `ffmpeg` changes?

skyace65:
I have to agree with mhilbrunner, this whole section is starting to feel like a bit much. Especially since there's a wide variety of ways to convert video, not just FFMPEG and handbrake. This is opening a can of worms where we wouldn't have a reason to turn down docs for other video editors relevant to this, which would just make the docs bigger and harder to maintain.

And it's not like external code editors where we can justify it because there's stuff specific to Godot that needs to be documented. An AVI file from Godot is the same as an AVI file from any other program and would be edited the same way.

chrisl8:
I'm in favor of it.

This kind of "here is exactly what you need to do this thing" explanation is what makes Godot beginner friendly.

There are hundreds of ways to encode video, with many specific arguments to the tools like `ffmpeg`, and one can spend hours fighting to get it right with little feedback from the engine about why it is failing.

I think that this will reduce support requests by giving people a simple "this will work" answer.

mhilbrunner:
> There are hundreds of ways to encode video, with many specific arguments to the tools like ffmpeg, and one can spend hours fighting to get it right with little feedback from the engine about why it is failing.

If I am not missing anything, this isn't about doing *anything* in the engine (and getting little feedback from what you're doing wrong in Godot).

At this point, you already have your finished `.avi` movie file. This tutorial is doing about something else with that, converting it to a different format. That is indeed useful, but not sure it is really content for *Godot*'s documentation. :)

mhilbrunner:
I still like the content, but I also still don't want to have to maintain for eternity on the docs, and don't want to give reason for this page to keep growing and growing with tutorials on external tooling.

@Calinou Would it make sense to salvage it into a small blog post about how to make movies? Or make sure it gets published somewhere else, I guess.

For the docs for now, I'll close this due to the lack of consensus, but if we can put this elsewhere, that'd be nice :)

PR #9098, opened 2024-03-15
Add extension support info for web export

### LINKED ISSUE #6970: Exporting for web page is possibly outdated
**Your Godot version:**

v4.0.stable.official [92bee43ad]

**Issue description:**

The exporting for web page currently says:

> You can choose the Export Type to select which features will be available:
> 
> Regular: is the most compatible across browsers, will not support threads, nor GDExtension.
> 
> Threads: will require the browser to support [SharedArrayBuffer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer). See [Can I use SharedArrayBuffer](https://caniuse.com/sharedarraybuffer) for details.

I don't see these options in Godot 4.0.  Instead, I see:

image

I might be missing something, but based on what I'm observing in Godot 4.0 I think this section needs two changes:

1. To reference "Variant/Extensions Support" instead of "Export Type"
2. To explain that exporting for web always requires SharedArrayBuffer

If I'm wrong about the behavior above, please consider this a request for improved documentation.  I can't find "Export Type" anywhere, and I haven't been able to find a way to disable SharedArrayBuffer.

https://docs.godotengine.org/en/stable/tutorials/export/exporting_for_web.html

skyace65:
States that "extension support" needs to be enabled if a project uses GDExtension. I'm not sure if there's more to add to this or not, that's basically the info people on Discord gave me. If anything else should be added I'll gladly do it. Closes #6970.

### DIFF
--- a/tutorials/export/exporting_for_web.rst
+++ b/tutorials/export/exporting_for_web.rst
@@ -57,6 +57,8 @@ If a runnable web export template is available, a button appears between the
 *Stop scene* and *Play edited Scene* buttons in the editor to quickly open the
 game in the default browser for testing.
 
+If your project uses GDExtension **Extension Support** needs to be enabled.
+
 If you plan to use :ref:`VRAM compression <doc_importing_images>` make sure that
 **Vram Texture Compression** is enabled for the targeted platforms (enabling
 both **For Desktop** and **For Mobile** will result in a bigger, but more


mhilbrunner (approved):

mhilbrunner:
Thanks! Merged.

mhilbrunner:

PR #9101, opened 2024-03-16
Add more tool script info and add a section

### LINKED ISSUE #8692: "Running Code in the Editor" missing info on Resources
**Your Godot version:**
Godot 4.2

**Issue description:**
It seems like Nodes that use `[Tool]` can only use Resources that also have `[Tool]`. Without it, Godot will throw the following error:

> Unable to cast object of type 'Godot.Resource' to type 'ExampleResourceType'.

This is not exactly a clear error and [has caused confusion for several people](https://github.com/godotengine/godot/issues/85459#issuecomment-1870709144). It also does not appear that the docs explain this behavior.

I think it would make sense to include a few sentences explaining that Resources are affected by `[Tool]` as well. The wording is not important to me, so long as it conveys what the developer needs to do.

https://docs.godotengine.org/en/stable/tutorials/plugins/running_code_in_the_editor.html

skyace65:
Adds that any script a tool script interacts with must also be a tool script. Instead of creating another note in this section I've combined that info with the two existing notes. Having three back to back would be excessive. Closes #8692

### DIFF
--- a/tutorials/plugins/running_code_in_the_editor.rst
+++ b/tutorials/plugins/running_code_in_the_editor.rst
@@ -105,18 +105,20 @@ Here is how a ``_process()`` function might look for you:
         // Code to execute both in editor and in game.
     }
 
-.. note::
+Important information
+---------------------
 
-    Modifications in the editor are permanent. For example, in the following
-    case, when we remove the script, the node will keep its rotation. Be careful
-    to avoid making unwanted modifications.
+Any other GDScript that your tool script uses must *also* be a tool. Any
+GDScript without ``@tool`` used by the editor will act like an empty file!
 
-.. note::
+Extending a ``@tool`` script does not automatically make the extending script
+a ``@tool``. Omitting ``@tool`` from the extending script will disable tool
+behavior from the super class. Therefore the extending script should also
+specify the ``@tool`` annotation.
 
-    Extending a ``@tool`` script does not automatically make the extending script
-    a ``@tool``. Omitting ``@tool`` from the extending script will disable tool
-    behavior from the super class. Therefore the extending script should also
-    specify the ``@tool`` annotation.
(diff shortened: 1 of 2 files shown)


mhilbrunner (approved):

tutorials/plugins/running_code_in_the_editor.rst
```diff
@@ -105,18 +105,20 @@ Here is how a ``_process()`` function might look for you:
...
 
-.. note::
+Important information
+---------------------
 
-    Modifications in the editor are permanent. For example, in the following
-    case, when we remove the script, the node will keep its rotation. Be careful
-    to avoid making unwanted modifications.
+Any other GDScript that your tool script uses must *also* be a tool. Any
+GDScript without ``@tool`` imported into the editor will act like an empty file!
```
AThousandShips, line 112: Do we "import" scripts? Is this terminology used elsewhere? I don't think it makes sense to use it, imports are files that are modified and converted no?

skyace65: That terminology comes from this note on the making plugins page, which I modified a little for this PR

"In addition to the EditorPlugin script, any other GDScript that your plugin uses must also be a tool. Any GDScript without @tool imported into the editor will act like an empty file!"

AThousandShips: That should probably be modified separately later, but I don't think using "imported" here as it's not correct for scripts

skyace65:
@AThousandShips I've updated the phrasing to say "used by" instead of "imported into", does that work?

AThousandShips (approved):

mhilbrunner (approved):

mhilbrunner:
Thanks, merged.

mhilbrunner:

PR #9106, opened 2024-03-17
Update compiling_for_macos.rst with double precision app bundling example.

micheal-w-wells:
Add full app bundling cmd for mac silicon + double precision binary, tested on M1 air Sonoma 14.2.1.  The other bundle cmd did not work for me with `universal` swapped for `arm64`.

### DIFF
--- a/contributing/development/compiling/compiling_for_macos.rst
+++ b/contributing/development/compiling/compiling_for_macos.rst
@@ -152,6 +152,14 @@ with the following commands (assuming a universal build, otherwise replace the
     cp bin/godot.macos.template_debug.universal macos_template.app/Contents/MacOS/godot_macos_debug.universal
     chmod +x macos_template.app/Contents/MacOS/godot_macos*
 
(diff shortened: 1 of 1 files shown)


AThousandShips:

Please open a new PR on the correct branch, if needed (check the latest version first).

contributing/development/compiling/compiling_for_macos.rst
```diff
@@ -152,6 +152,14 @@ with the following commands (assuming a universal build, otherwise replace the
     cp bin/godot.macos.template_debug.universal macos_template.app/Contents/MacOS/godot_macos_debug.universal
     chmod +x macos_template.app/Contents/MacOS/godot_macos*
 
+For a binary compiled with double precision, (for example `scons platform=macos arch=arm64 precision=double`) the following will work:
```
AThousandShips, line 155: For when you reopen a new version, you need to have two colons here, like so:
```suggestion
For a binary compiled with double precision, (for example `scons platform=macos arch=arm64 precision=double`) the following will work::
```
See the one above

PR #9111, opened 2024-03-17
Update getting_started/step_by_step/scripting_player_input.rst

DonutsInBelly:
Hi!
Was following the Godot docs to learn a bit and found this bit that didn't make sense. Opened the code and realized that something wasn't rendering, so here's a small fix.

Page in question: https://docs.godotengine.org/en/stable/getting_started/step_by_step/scripting_player_input.html

* Removed a space that cause code block to not render

Screenshot 2024-03-17 at 1 42 43 PM

Cheers!


DonutsInBelly:
hmm I'm gonna close this because I don't think I totally understood which branches got released when I made this PR. Code in this repo seems to get merged into versioned branches from master branch and then deployed on the version branch. When I made this, I thought everything was just deployed from master and the branches were just archived versions.

PR #9113, opened 2024-03-18
Add VRS/Foveated rendering to XR start script

BastiaanOlij:
This PR adds code to our startup script that:
- ensures VRS is enabled if the mobile/forward+ renderer is used
- warns the user if foveated rendering is not turned on in the OpenXR project settings if the compatibility renderer is used.

Companion to: https://github.com/godotengine/godot-demo-projects/pull/1022

### DIFF
--- a/tutorials/xr/a_better_xr_start_script.rst
+++ b/tutorials/xr/a_better_xr_start_script.rst
@@ -91,7 +91,16 @@ We introduce a few new variables to our script as well:
 Our updated ready function
 --------------------------
 
-The ready function mostly remains the same but we hook up a number of signals that will be emitted by the :ref:`XRInterface <class_xrinterface>`.
+We add a few things to the ready function.
+
+If we're using the mobile or forward+ renderer we set the viewports ``vrs_mode`` to ``VRS_XR``.
+On platforms that support this, this will enable foveated rendering.
+
+If we're using the compatibility renderer, we check if the OpenXR foveated rendering settings
+are configured and if not, we output a warning.
+See :ref:`OpenXR Settings <doc_openxr_settings>` for further details.
+
+We hook up a number of signals that will be emitted by the :ref:`XRInterface <class_xrinterface>`.
 We'll provide more detail about these signals as we implement them.
 
 We also quit our application if we couldn't successfully initialise OpenXR.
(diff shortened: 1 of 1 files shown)


BastiaanOlij:
@Malcolmnixon would you mind adding this to https://github.com/Malcolmnixon/godot_openxr_demo/tree/csharp so we can also include the C# version of this?

Malcolmnixon:
> @Malcolmnixon would you mind adding this to https://github.com/Malcolmnixon/godot_openxr_demo/tree/csharp so we can also include the C# version of this?

It's updated.

BastiaanOlij:
Ok, CSharp code has been added and a little bit more of an introduction to the ready section.

BastiaanOlij:
Note that https://github.com/godotengine/godot-demo-projects/pull/1022 applies these changes to our demo projects.

tutorials/xr/a_better_xr_start_script.rst
```diff
@@ -91,7 +91,16 @@ We introduce a few new variables to our script as well:
 Our updated ready function
 --------------------------
 
-The ready function mostly remains the same but we hook up a number of signals that will be emitted by the :ref:`XRInterface <class_xrinterface>`.
+We add a few things to the ready function.
+
+If we're using the mobile or forward+ renderer we set the viewports ``vrs_mode`` to ``VRS_XR``.
+On platforms that support this, this will enable foveated rendering.
```
skyace65, line 97: Would a platform in this case be referring to the headset or the OpenXR runtime(SteamVR, Mixed Reality Portal)?

tutorials/xr/a_better_xr_start_script.rst
```diff
@@ -91,7 +91,16 @@ We introduce a few new variables to our script as well:
...
+
+If we're using the compatibility renderer, we check if the OpenXR foveated rendering settings
+are configured and if not, we output a warning.
```
skyace65, line 100: This part is a bit confusing to me. The previous paragraph says to only change settings for Mobile and Forward+, but we put in a warning for when we intentionally don't do anything for the compatibility renderer?

Also while I understand the point of having the warning in the code for the XRTools addon, I'm not sure it makes sense to include it here where users would be the ones typing out the code. Especially since users would have just enabled the setting.

tutorials/xr/a_better_xr_start_script.rst
BastiaanOlij: In this case it's unrelated to the XR side, VRS is implemented in Vulkan so it's up to the capabilities of the GPU driver.

tutorials/xr/a_better_xr_start_script.rst
BastiaanOlij: The idea here is that we include this script in every XR entry on godot-demo-projects and I'm trying to keep the script the same over all projects. This help page just further explains each part of that script.

So you are right, someone who uses this help page to create their own version of this script, would skip over this and just configure things the way they want depending on their choice of renderer.

But those who grab the script as is from the demo projects to include in their own projects, for them it's a little extra to nudge them to enable foveated rendering as they may not realise there are separate settings for compatibility.

skyace65 (approved):

skyace65:
Thanks!

Issue #9117, opened 2024-03-19
"Why does Godot use Vulkan or OpenGL instead of Direct3D" in faq is outdated


jsjtxietian:
**Your Godot version:**

Current master

**Issue description:**

In faq it says that "In the long term, we may develop a Direct3D 12 renderer for Godot (mainly for Xbox), but Vulkan and OpenGL will remain the default rendering backends on all platforms, including Windows." 

While this driver is still experimental and only available in Godot 4.3 and later, it is present now.

**URL to the documentation page:**

https://docs.godotengine.org/en/latest/about/faq.html#why-does-godot-use-vulkan-or-opengl-instead-of-direct3d

PR #9118, opened 2024-03-19
Update faq's directx12 question

jsjtxietian:
Fixes https://github.com/godotengine/godot-docs/issues/9117

But the title can be changed too, any good suggestion?


skyace65 (commented):
You could go with something like "Why does Godot prioritize Vulkan and OpenGL over Direct3D?"

Also I think the second paragraph in this section can just be completely removed.

skyace65 (approved):

mhilbrunner (approved):

mhilbrunner:
Thank you! Merged.

PR #9121, opened 2024-03-20
OpenXR: Add a note about the vendor plugin

BastiaanOlij:
For a few versions the vendor plugin needed to be enabled but now that it has been rewritten to GDExtensions this is no longer the case.

I though of removing the section all together but seeing this info is features in a fair amount of places, I thought a note would be more communicative so people know this is no longer needed.


mhilbrunner (approved):

skyace65 (approved):
I think it would make more sense to have that note before the instructions. But I'm fine with it as is.

skyace65:
Thanks!

mhilbrunner:

PR #9126, opened 2024-03-21
Android export: mention special chars in passwords caveat

jefemagril:
See for more info about this: https://stackoverflow.com/questions/10338543/what-causes-keytool-error-failed-to-decrypt-safe-contents-entry

### DIFF
--- a/tutorials/export/exporting_for_android.rst
+++ b/tutorials/export/exporting_for_android.rst
@@ -144,6 +144,7 @@ keystore file; such file can be generated like this:
     keytool -v -genkey -keystore mygame.keystore -alias mygame -keyalg RSA -validity 10000
 
 This keystore and key are used to verify your developer identity, remember the password and keep it in a safe place!
+It is suggested to use only upper and lowercase letters and numbers. Special characters may cause errors.
 Use Google's Android Developer guides to learn more about `APK signing <https://developer.android.com/studio/publish/app-signing>`__.
 
 Now fill in the following forms in your Android Export Presets:


mhilbrunner (approved):

mhilbrunner:
Makes sense, although I'm not sure whether it is good to recommend reducing the possible keyspace like this officially, or whether users should just be cautioned to avoid UTF-8 special chars or to enclose passphrases containing special chars in quotes as usual on CLI. Is there something official on the Android side recommending this?

AThousandShips:
This is targeting 4.2, not master, is that desired?

mhilbrunner:
Ah, good catch. Yeah, this should definitely target the master branch.

jefemagril:
Sorry re: targeting, I'm a noob to this. No, I'm not aware of any official info from Android about this; there's no mention of password requirements in their docs. I made this because it resolved what I was stuck on for days. I'm not sure quotes would resolve this, because the password is entered as a prompted and hidden input on the CLI, not as an argument.

mhilbrunner:
Ah, I always forget how basic the Github base branch switch is, the PR needs to be rebased to master manually.

> I made this because it resolved what I was stuck on for days. I'm not sure quotes would resolve this, because the password is entered as a prompted and hidden input on the CLI, not as an argument.

Gotcha. In that case I think the recommendation makes sense.

mhilbrunner (approved):

mhilbrunner:
Alright, fixed the branch and merged. Thanks and congrats on your first merged contribution to Godot's docs! 🎉

mhilbrunner:

PR #9128, opened 2024-03-22
Update first 2D game input mapping event configuration image

idosavion:
The previous photo contained a typo which is no longer present in Godot 4.2 ("Poistion" -> "Position").

Adding current and new photo for review convenience:

Current:
current

New:
image


mhilbrunner:
Thanks for taking the time to update this image!

The new image contains noticeable compression artifacts and looks more blurry to me (while increasing the file size). It also includes OS-specific window chrome, and lacks the selection highlight (albeit for the latter, I'm not sure how important it was, if at all).

Would be good if we could get it closer to the original in these respects.

skyace65:
In addition to what mhilbrunner said, if you're not that familiar with editing images we do have a guide on editing them for the documentation [here](https://docs.godotengine.org/en/stable/contributing/documentation/docs_image_guidelines.html).

idosavion:
Thanks @mhilbrunner @skyace65 !
1. I took another photo and converted using `Squoosh` as suggested (updated the PR message to contain the new photo (the photo I attached is converted to jpeg due to github technical constraints).
2. I changed the photo so it won't contain the macOS buttons at the top in order to make it more consistent with the rest of the pictures in the tutorial (although, there is a minor difference in the "Additional Options" section).
3. The selection highlight is toggled on my machine, the color seems to be different.

Hope that works as I used all the resources I found relevant in the suggested documentation.

mhilbrunner (approved):

mhilbrunner:
Thanks for making those changes, that indeed looks way better! :)

Congrats on your first merged PR here, and thanks for taking your time to contribute to Godot!

mhilbrunner:
Cherrypicked to 4.2.

PR #9135, opened 2024-03-24
Fixed Typo in astargrid2d method get_point_path

brandtware:
The docs for AStar2D say that they would return a PackedVector**3**Array while inside a thread but the typing for that function is PackedVector**2**Array.

### DIFF
--- a/classes/class_astargrid2d.rst
+++ b/classes/class_astargrid2d.rst
@@ -553,7 +553,7 @@ Returns an array with the IDs of the points that form the path found by AStar2D 
 
 Returns an array with the points that are in the path found by **AStarGrid2D** between the given points. The array is ordered from the starting point to the ending point of the path.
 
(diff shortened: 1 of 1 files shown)


AThousandShips:

brandtware:
Ok, now I am confused, in the other repo you mentioned it was already fixed 3 years ago, yet it is still online in the docs (see below). This shouldn't happen... right?
https://docs.godotengine.org/en/stable/classes/class_astargrid2d.html#class-astargrid2d-method-get-point-path

AThousandShips:
It's not fixed there:
https://github.com/godotengine/godot/blob/99ff024f78f65ba0bc54fb409cfeca43ba2008fe/doc/classes/AStarGrid2D.xml#L88

brandtware:
Dang, I checked AStar2D (non-**Grid**) sorry... I'll try again, first time contributing

PR #9158, opened 2024-03-29
Document how to handle compatibility breakages

AThousandShips:
Basic instructions and examples for using this feature

### DIFF
--- a/contributing/development/handling_compatibility_breakages.rst
+++ b/contributing/development/handling_compatibility_breakages.rst
@@ -1,0 +1,141 @@
+.. _doc_handling_compatibility_breakages:
+
+Handling compatibility breakages
+================================
+
+.. TODO: Elaborate on types of compatibility and procedure.
+
+So you've added a new parameter to a method, changed the return type,
+changed the type of a parameter, or changed its default value,
+and now the automated testing is complaining about compatibility breakages?
+
+Breaking compatibility should be avoided, but when necessary there are systems in place
+to handle this in a way that makes the transition as smooth as possible.
+
+A practical example
+-------------------
+
+.. TODO: Add example that showcases more details like original default arguments etc.
+
+These changes are taken from `pull request #88047 <https://github.com/godotengine/godot/pull/88047>`_, which added
+new pathing options to ``AStarGrid2D`` and other AStar classes.
+Among other changes, these methods were modified in ``core/math/a_star_grid_2d.h``:
+
+.. code-block:: cpp
+
+    Vector<Vector2> get_point_path(const Vector2i &p_from, const Vector2i &p_to);
+    TypedArray<Vector2i> get_id_path(const Vector2i &p_from, const Vector2i &p_to);
+
+To:
+
+.. code-block:: cpp
+
+    Vector<Vector2> get_point_path(const Vector2i &p_from, const Vector2i &p_to, bool p_allow_partial_path = false);
+    TypedArray<Vector2i> get_id_path(const Vector2i &p_from, const Vector2i &p_to, bool p_allow_partial_path = false);
+
+This meant adding new compatibility method bindings to the file, which should be in the ``protected`` section of
+the code, usually placed next to ``_bind_methods()``:
+
+.. code-block:: cpp
+
+    #ifndef DISABLE_DEPRECATED
+        TypedArray<Vector2i> _get_id_path_bind_compat_88047(const Vector2i &p_from, const Vector2i &p_to);
+        Vector<Vector2> _get_point_path_bind_compat_88047(const Vector2i &p_from, const Vector2i &p_to);
+        static void _bind_compatibility_methods();
+    #endif
+
+They should start with a ``_`` to indicate that they are internal, and end with ``_bind_compat_`` followed by the PR number
+that introduced the change (``88047`` in this example). These compatibility methods need to be implemented in a dedicated file,
+like ``core/math/a_star_grid_2d.compat.inc`` in this case:
+
+.. code-block:: cpp
+
+    /**************************************************************************/
+    /*  a_star_grid_2d.compat.inc                                             */
+    /**************************************************************************/
+    /*                         This file is part of:                          */
+    /*                             GODOT ENGINE                               */
+    /*                        https://godotengine.org                         */
+    /**************************************************************************/
+    /* Copyright (c) 2014-present Godot Engine contributors (see AUTHORS.md). */
+    /* Copyright (c) 2007-2014 Juan Linietsky, Ariel Manzur.                  */
+    /*                                                                        */
+    /* Permission is hereby granted, free of charge, to any person obtaining  */
+    /* a copy of this software and associated documentation files (the        */
+    /* "Software"), to deal in the Software without restriction, including    */
+    /* without limitation the rights to use, copy, modify, merge, publish,    */
+    /* distribute, sublicense, and/or sell copies of the Software, and to     */
+    /* permit persons to whom the Software is furnished to do so, subject to  */
+    /* the following conditions:                                              */
+    /*                                                                        */
+    /* The above copyright notice and this permission notice shall be         */
+    /* included in all copies or substantial portions of the Software.        */
+    /*                                                                        */
+    /* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,        */
+    /* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF     */
+    /* MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. */
+    /* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+    /* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+    /* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+    /* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+    /**************************************************************************/
+
+    #ifndef DISABLE_DEPRECATED
+
+    #include "core/variant/typed_array.h"
+
+    TypedArray<Vector2i> AStarGrid2D::_get_id_path_bind_compat_88047(const Vector2i &p_from_id, const Vector2i &p_to_id) {
+        return get_id_path(p_from_id, p_to_id, false);
+    }
+
+    Vector<Vector2> AStarGrid2D::_get_point_path_bind_compat_88047(const Vector2i &p_from_id, const Vector2i &p_to_id) {
+        return get_point_path(p_from_id, p_to_id, false);
+    }
+
+    void AStarGrid2D::_bind_compatibility_methods() {
+        ClassDB::bind_compatibility_method(D_METHOD("get_id_path", "from_id", "to_id"), &AStarGrid2D::_get_id_path_bind_compat_88047);
+        ClassDB::bind_compatibility_method(D_METHOD("get_point_path", "from_id", "to_id"), &AStarGrid2D::_get_point_path_bind_compat_88047);
+    }
+
+    #endif // DISABLE_DEPRECATED
+
+Unless the change in compatibility is complex, the compatibility method should simply call the modified method directly,
+instead of duplicating that method. Make sure to match the default arguments for that method (in the example above this would be ``false``).
+
+This file should always be placed next to the original file, and have ``.compat.inc`` at the end instead of ``.cpp`` or ``.h``.
+Next, this should be included in the ``.cpp`` file we're adding compatibility methods to, so ``core/math/a_star_grid_2d.cpp``:
+
+.. code-block:: cpp
+
+    #include "a_star_grid_2d.h"
+    #include "a_star_grid_2d.compat.inc"
+
+    #include "core/variant/typed_array.h"
+
+And finally, the changes reported by the API validation step should be added to the relevant validation file. Because this was
+done during the development of 4.3, this would be ``misc/extension_api_validation/4.2-stable.expected`` (including changes not shown in
+this example):
+
+.. code-block:: text
+
+    GH-88047
+    --------
+    Validate extension JSON: Error: Field 'classes/AStar2D/methods/get_id_path/arguments': size changed value in new API, from 2 to 3.
+    Validate extension JSON: Error: Field 'classes/AStar2D/methods/get_point_path/arguments': size changed value in new API, from 2 to 3.
+    Validate extension JSON: Error: Field 'classes/AStar3D/methods/get_id_path/arguments': size changed value in new API, from 2 to 3.
+    Validate extension JSON: Error: Field 'classes/AStar3D/methods/get_point_path/arguments': size changed value in new API, from 2 to 3.
+    Validate extension JSON: Error: Field 'classes/AStarGrid2D/methods/get_id_path/arguments': size changed value in new API, from 2 to 3.
+    Validate extension JSON: Error: Field 'classes/AStarGrid2D/methods/get_point_path/arguments': size changed value in new API, from 2 to 3.
+
+    Added optional "allow_partial_path" argument to get_id_path and get_point_path methods in AStar classes.
+    Compatibility methods registered.
+
+The instructions for how to add to that file are at the top of the file itself.
+
+If you get a "Hash changed" error for a method, it means that the compatibility binding is missing or incorrect.
+Such lines shouldn't be added to the ``.expected`` file, but fixed by binding the proper compatibility method.
+
+And that's it! You might run into a bit more complicated cases, like rearranging arguments,
+changing return types, etc., but this covers the basic on how to use this system.
+
+For more information, see `pull request #76446 <https://github.com/godotengine/godot/pull/76446>`_.
--- a/about/release_policy.rst
+++ b/about/release_policy.rst
@@ -275,5 +275,4 @@ compatibility-breaking changes of this kind.
       optional parameter), a GDExtension compatibility method must be created.
       This ensures that existing GDExtensions continue to work across patch and
       minor releases, so that users don't have to recompile them.
-      See `pull request #76446 <https://github.com/godotengine/godot/pull/76446>`_
-      for more information.
+      See :ref:`doc_handling_compatibility_breakages` for more information.
--- a/contributing/development/index.rst
+++ b/contributing/development/index.rst
@@ -19,6 +19,7 @@ especially if you're new to using Git or GitHub.
    best_practices_for_engine_contributors
    code_style_guidelines
    cpp_usage_guidelines
+   handling_compatibility_breakages
 
 Buildsystem and work environment
 --------------------------------


contributing/development/handling_compatibility_breakages.rst
```diff
@@ -0,0 +1,110 @@
+.. _doc_handling_compatibility_breakages:
+
+Handling compatibility breakages
+=================================
+
+So you've added a new argument to a method, changed the return type,
+changed the type of an argument, or changed the default value of an argument,
+and now the automated testing is complaining about compatibility breakages?
```
Bromeon, line 8: Technically those are parameters (declared in the signature) rather than arguments (values passed in). But I've seen Godot often use the latter... so not sure?

contributing/development/handling_compatibility_breakages.rst
```diff
@@ -0,0 +1,110 @@
...
+
+Not to worry, there's a system to handle just that!
+
+Compatibility is handled in a few different ways, in GDScript this is simple,
+new arguments are usually given default values when added so existing scripts
+won't break, and types are flexible in GDScript which makes this work well.
```
Bromeon, line 14: Quite long sentence, some punctuation might help the flow:
```suggestion
Compatibility is handled in a few different ways. In GDScript this is simple:
new arguments are usually given default values when added, so existing scripts
won't break. Types are flexible in GDScript which makes this work well.
```

I'm not exactly sure what you mean by "Types are flexible in GDScript which makes this work well." -- what does "flexible" refer to here?

contributing/development/handling_compatibility_breakages.rst
```diff
@@ -0,0 +1,110 @@
...
+
+However in other cases, like GDExtension and C#, the exact way the method is
+declared matters and must be matched exactly. This is handled by creating
+compatibility methods that then gets translated to the new versions of these
+methods. All of this happens *almost* automatically, there are just a few
+things you need to do.
```
Bromeon, line 20:
```suggestion
However in other cases, like GDExtension and C#, the exact way the method is
declared matters and must be matched exactly. This is handled by creating
compatibility methods that then get translated to the new versions of these
methods. All of this happens *almost* automatically, there are just a few
things you need to do.
```
And maybe "mapped" is a bit clearer than "translated" here?

contributing/development/handling_compatibility_breakages.rst
```diff
@@ -0,0 +1,110 @@
...
+
+Binding compatibility methods
+-----------------------------
+
+Just like how you use ``ClassDB`` to bind methods there's a way to bind
+compatibility methods, and conveniently these are named very similarly to their
+normal counterparts.
```
Bromeon, line 27:
```suggestion
Just like how you use ``ClassDB`` to bind methods, there's a way to bind
compatibility methods, and conveniently these are named very similarly to their
normal counterparts.
```

Bromeon (commented):
Thanks a lot! Sounds good so far, my feedback is very nitpicky 😅 

Regarding an example, maybe we should use a real one that occurred? Then we don't need to make up things. It probably illustrates the principle better than an abstract one with "arg1", "arg2" etc.

dsnopek:
Thanks for starting on this! I look forward to linking to it in the future :-)

contributing/development/handling_compatibility_breakages.rst
AThousandShips: It was related to how weak typing and similar works in GDScript but it's not necessarily helpful here, so removing it, it was how you can change a signature more safely than a compiled language

contributing/development/handling_compatibility_breakages.rst
```diff
@@ -0,0 +1,158 @@
...
+-----------------------------
+
+Just like how you use ``ClassDB`` to bind methods, there's a way to bind
+compatibility methods, and conveniently these are named very similarly to their
+ordinary counterparts.
+
+.. code-block:: cpp
+
+    ClassDB::bind_method(D_METHOD("my_method", "arg1", "arg2"), &MyClass::my_method);
+    ClassDB::bind_compatibility_method(D_METHOD("my_method", "arg1", "arg2", "arg3"), &MyClass::_my_method_bind_compat_1234, DEFVAL(0));
```
dsnopek, line 32: While this is a totally valid example of an argument being removed, I feel like the reverse case is more common: adding a new argument to method. So, I'd personally reverse these:

```suggestion
    ClassDB::bind_method(D_METHOD("my_method", "arg1", "arg2", "arg3"), &MyClass::my_method, DEFVAL(0));
    ClassDB::bind_compatibility_method(D_METHOD("my_method", "arg1", "arg2"), &MyClass::_my_method_bind_compat_1234);
```

contributing/development/handling_compatibility_breakages.rst
```diff
@@ -0,0 +1,158 @@
...
+.. code-block:: cpp
+
+    #ifndef DISABLE_DEPRECATED
+        TypedArray<Vector2i> _get_id_path_bind_compat_88047(const Vector2i &p_from, const Vector2i &p_to);
+        Vector<Vector2> _get_point_path_bind_compat_88047(const Vector2i &p_from, const Vector2i &p_to);
+        static void _bind_compatibility_methods();
+    #endif
+
+Note the ``88047`` at the end of the method names should match the PR number, they should also start with a ``_`` to
+indicate that they are internal, and end with ``_bind_compat`` followed by the PR number. These compatibility methods
```
dsnopek, line 71: This redundantly mentions the "PR number" twice in the same sentence. Maybe:

```suggestion
They should start with a ``_`` to indicate that they are internal, and end with ``_bind_compat`` followed by the PR number that introduced the change (``88047`` in this example). These compatibility methods
```

contributing/development/handling_compatibility_breakages.rst
```diff
@@ -0,0 +1,158 @@
...
+need to be defined as well, and these were added in a dedicated file, in this case ``core/math/a_star_grid_2d.compat.inc``:
```
dsnopek, line 72:
```suggestion
need to be implemented in a dedicated file, like ``core/math/a_star_grid_2d.compat.inc`` in this case:
```

contributing/development/handling_compatibility_breakages.rst
```diff
@@ -0,0 +1,158 @@
...
+
+.. code-block:: cpp
+
+    #include "a_star_grid_2d.h"
+    #include "a_star_grid_2d.compat.inc"
+
+    #include "core/variant/typed_array.h"
+
+And finally the changes reported by the API validation step should be added to the relevant validation file, which because this was
+done during the development of 4.3 would be ``misc/extension_api_validation/4.2-stable.expected`` (including changes not shown in
```
dsnopek, line 136:
```suggestion
And finally, the changes reported by the API validation step should be added to the relevant validation file. Because this was
done during the development of 4.3, this would be ``misc/extension_api_validation/4.2-stable.expected`` (including changes not shown in
```

contributing/development/handling_compatibility_breakages.rst
```diff
@@ -0,0 +1,158 @@
...
+    Validate extension JSON: Error: Field 'classes/AStar2D/methods/get_point_path/arguments': size changed value in new API, from 2 to 3.
+    Validate extension JSON: Error: Field 'classes/AStar3D/methods/get_id_path/arguments': size changed value in new API, from 2 to 3.
+    Validate extension JSON: Error: Field 'classes/AStar3D/methods/get_point_path/arguments': size changed value in new API, from 2 to 3.
+    Validate extension JSON: Error: Field 'classes/AStarGrid2D/methods/get_id_path/arguments': size changed value in new API, from 2 to 3.
+    Validate extension JSON: Error: Field 'classes/AStarGrid2D/methods/get_point_path/arguments': size changed value in new API, from 2 to 3.
+
+    Added optional "allow_partial_path" argument to get_id_path and get_point_path methods in AStar classes.
+    Compatibility methods registered.
+
+The instructions for how to add to that file are at the top of the file itself.
```
dsnopek, line 153: Either here, or in the forth-coming instructions above, we should mention that in 99.99% of cases, we shouldn't be ignore the "hash changed" messages. If those don't go away on their own, then the developer will know they did something wrong with the compatibility methods.

dsnopek (commented):
Thanks! This is looking really great so far.

contributing/development/handling_compatibility_breakages.rst
AThousandShips: Think this belongs in the expected files to ensure they're found

contributing/development/handling_compatibility_breakages.rst
```diff
@@ -0,0 +1,168 @@
...
+This checks the API generated by the engine against that of previous versions, as well as files containing the expected
+results.
+
+To tell the system that the changes made in a PR were not by mistake but intentional compatibility breakages you need
+to add the results of this check to the relevant validation file. These files are stored under ``misc/extension_api_validation/``
+and are named in two ways. Either a short name with just one version in it, like ``4.2-stable.expected``, or two versions,
+one after another, like ``4.1-stable_4.2-stable.expected``. Most of the time you only need to care about the first type, which will
+have the version number of the most recently released version.
+
+The specific instructions for what to do when this check fails are at the top of each of these files.
```
AThousandShips, line 49: Might be a bit rambling, any suggestions would be welcome

contributing/development/handling_compatibility_breakages.rst
akien-mga: I agree this is worth calling out here.
```suggestion
The instructions for how to add to that file are at the top of the file itself.

If you get a "Hash changed" error for a method, it means that the compatibility binding is missing or incorrect. Such lines shouldn't be added to the ``.expected`` file, but fixed by binding the proper compatibility method.
```

contributing/development/handling_compatibility_breakages.rst
```diff
@@ -0,0 +1,168 @@
+.. _doc_handling_compatibility_breakages:
+
+Handling compatibility breakages
+=================================
```
akien-mga, line 4:
```suggestion
Handling compatibility breakages
================================
```

contributing/development/handling_compatibility_breakages.rst
```diff
@@ -0,0 +1,168 @@
...
+
+Compatibility is handled in a few different ways. In GDScript this is simple,
+new arguments are usually given default values when added, so existing scripts
+won't break in such cases.
+
+However in other cases, like GDExtension and C#, the exact way the method is
+declared matters and must be matched exactly. This is handled by creating
+compatibility methods that then get mapped to the new versions of these
+methods. All of this happens *almost* automatically, there are just a few
+things you need to do.
```
akien-mga, line 20: I feel we need to call out the different types of "compatibility breakages" more explicitly:
- For GDScript, adding a new parameter at the end of a method prototype with a default value does not break compatibility, but it does for GDExtension and C#, as described here. This is the only "compat breakage" that is ok to do in general.
- All other compatibility breakages break GDScript even if compat methods are added, so they can only be done exceptionally with approval for the release management team.

Not sure how to clarify this, with the introduction already being a bit long. I think my main issue is that the light hearted "no worry, there's a system that lets you break compat" may give the wrong impression about what kind of changes are allowed by default.

contributing/development/handling_compatibility_breakages.rst
akien-mga: I feel those two generic sections tend to overlap with explanations given below in the practical example, especially the one about binding compat methods which I would expect to explain how to name the compat method, where to place it, etc.

Maybe we could introduce the example from the start and not try to have standalone generic explanations, but use the example throughout to make the generic bits clearer?

contributing/development/handling_compatibility_breakages.rst
```diff
@@ -0,0 +1,168 @@
...
+    TypedArray<Vector2i> get_id_path(const Vector2i &p_from, const Vector2i &p_to);
+
+To:
+
+.. code-block:: cpp
+
+    Vector<Vector2> get_point_path(const Vector2i &p_from, const Vector2i &p_to, bool p_allow_partial_path = false);
+    TypedArray<Vector2i> get_id_path(const Vector2i &p_from, const Vector2i &p_to, bool p_allow_partial_path = false);
+
+This meant adding new compatibility bind methods to the file, which should be in the ``protected`` section of
```
akien-mga, line 69:
```suggestion
This meant adding new compatibility method bindings to the file, which should be in the ``protected`` section of
```

contributing/development/handling_compatibility_breakages.rst
```diff
@@ -0,0 +1,168 @@
...
+one after another, like ``4.1-stable_4.2-stable.expected``. Most of the time you only need to care about the first type, which will
+have the version number of the most recently released version.
+
+The specific instructions for what to do when this check fails are at the top of each of these files.
+
+A practical example
+-------------------
+
+These changes are taken from `pull request #88047 <https://github.com/godotengine/godot/pull/88047>`_, which added
+new pathing options to ``AStarGrid2D``. Which among other changes changed these methods in ``core/math/a_star_grid_2d.h``:
```
akien-mga, line 55:
```suggestion
new pathing options to ``AStarGrid2D`` and other AStar classes. Among other changes, these methods were modified in ``core/math/a_star_grid_2d.h``:
```

contributing/development/handling_compatibility_breakages.rst
```diff
@@ -0,0 +1,168 @@
...
+
+To:
+
+.. code-block:: cpp
+
+    Vector<Vector2> get_point_path(const Vector2i &p_from, const Vector2i &p_to, bool p_allow_partial_path = false);
+    TypedArray<Vector2i> get_id_path(const Vector2i &p_from, const Vector2i &p_to, bool p_allow_partial_path = false);
+
+This meant adding new compatibility bind methods to the file, which should be in the ``protected`` section of
+the code, usually placed next to ``bind_methods()``:
```
akien-mga, line 70:
```suggestion
the code, usually placed next to ``_bind_methods()``:
```

contributing/development/handling_compatibility_breakages.rst
```diff
@@ -0,0 +1,168 @@
...
+
+.. code-block:: cpp
+
+    #ifndef DISABLE_DEPRECATED
+        TypedArray<Vector2i> _get_id_path_bind_compat_88047(const Vector2i &p_from, const Vector2i &p_to);
+        Vector<Vector2> _get_point_path_bind_compat_88047(const Vector2i &p_from, const Vector2i &p_to);
+        static void _bind_compatibility_methods();
+    #endif
+
+They should start with a ``_`` to indicate that they are internal, and end with ``_bind_compat`` followed by the PR number
```
akien-mga, line 80:
```suggestion
They should start with a ``_`` to indicate that they are internal, and end with ``_bind_compat_`` followed by the PR number
```

contributing/development/handling_compatibility_breakages.rst
```diff
@@ -0,0 +1,168 @@
...
+        return get_id_path(p_from_id, p_to_id, false);
+    }
+
+    Vector<Vector2> AStarGrid2D::_get_point_path_bind_compat_88047(const Vector2i &p_from_id, const Vector2i &p_to_id) {
+        return get_point_path(p_from_id, p_to_id, false);
+    }
+
+    void AStarGrid2D::_bind_compatibility_methods() {
+        ClassDB::bind_compatibility_method(D_METHOD("get_id_path", "from_id", "to_id"), &AStarGrid2D::_get_id_path_bind_compat_88047);
+        ClassDB::bind_compatibility_method(D_METHOD("get_point_path", "from_id", "to_id"), &AStarGrid2D::_get_point_path_bind_compat_88047);
```
akien-mga, line 130: Would be good to have an example that uses `DEFVAL()`, or at least this needs to be called out that if the previous binding of a method had default values assigned, the compatibility method should have them too.

akien-mga (approved):
Left some suggestions, but aside from a couple typo/style nitpicks, I think this could be merged as is, and improved later. It's already a much needed and sufficient addition to our contributor docs.

AThousandShips:
Thank you for the review! Will try to update later today

contributing/development/handling_compatibility_breakages.rst
AThousandShips: Agreed will see what I can come up with, maybe we can cut this down and make a separate page discussing the general topic of compatibility focused on the abstract instead of the specific

contributing/development/handling_compatibility_breakages.rst
AThousandShips: Will see if I can find another example that has that to replace this with to make the example more complete

contributing/development/handling_compatibility_breakages.rst
AThousandShips: Dropping the generics, didn't feel any of it was useful to integrate into the practical example

contributing/development/handling_compatibility_breakages.rst
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    #endif // DISABLE_DEPRECATED
+
+Unless the change in compatibility is complex the compatibility method should simply call the normal method directly, instead of duplicating
+the method, making sure to match the default arguments of the bind.
```
AThousandShips, line 108: Added this to emphasize this part, might be worded better

AThousandShips:
Will look at the intro section and a more complex example case soon

AThousandShips:
There, reduced the intro section and made it clearer that breakgages should be avoided, and added a few `TODO`s, I'm comfortable merging it in this state and improving the details later, if the current wording is good

dsnopek (approved):
I just re-read this after the latest changes. Looks fantastic! Great work :-)

contributing/development/handling_compatibility_breakages.rst
```diff
@@ -0,0 +1,141 @@
...
+
+To:
+
+.. code-block:: cpp
+
+    Vector<Vector2> get_point_path(const Vector2i &p_from, const Vector2i &p_to, bool p_allow_partial_path = false);
+    TypedArray<Vector2i> get_id_path(const Vector2i &p_from, const Vector2i &p_to, bool p_allow_partial_path = false);
+
+This meant adding new compatibility method bindings to the file, which should be in the ``protected`` section of
+the code, usually placed next to ``_bind_methods()``:
```
Bromeon, line 37: You could briefly explain the concept of compatibility methods and explain why they're needed in the first place, something along:

> For the GDExtension API, we provide backwards compatibility with earlier versions: for example, a C++ extension compiled with Godot 4.1 still runs under Godot 4.3. For this to be possible, old member functions must remain available in their original signature, in addition to new version of those functions. We call the old function "compatibility method".

contributing/development/handling_compatibility_breakages.rst
```diff
@@ -0,0 +1,141 @@
...
+
+    void AStarGrid2D::_bind_compatibility_methods() {
+        ClassDB::bind_compatibility_method(D_METHOD("get_id_path", "from_id", "to_id"), &AStarGrid2D::_get_id_path_bind_compat_88047);
+        ClassDB::bind_compatibility_method(D_METHOD("get_point_path", "from_id", "to_id"), &AStarGrid2D::_get_point_path_bind_compat_88047);
+    }
+
+    #endif // DISABLE_DEPRECATED
+
+Unless the change in compatibility is complex the compatibility method should simply call the normal method directly, instead of duplicating
+the method, making sure to match the default arguments of the bind.
```
Bromeon, line 103: Slightly shorter sentences:
```suggestion
Unless the change in compatibility is complex, the compatibility method should simply call the new method directly, instead of duplicating
it. This makes sure that default arguments of the bind are matched.
```

Not quite sure about the last part -- for a user, what does "match the bind" mean? Maybe this can be written in easier-to-understand terms?

contributing/development/handling_compatibility_breakages.rst
```diff
@@ -0,0 +1,141 @@
...
+
+This file should always be placed next to the original file, and have ``.compat.inc`` at the end instead of ``.cpp`` or ``.h``.
+Next this should be included in the ``.cpp`` file we're adding compatibility methods to, so ``core/math/a_star_grid_2d.cpp``:
```
Bromeon, line 106:
```suggestion
Next, this should be included in the ``.cpp`` file we're adding compatibility methods to, so ``core/math/a_star_grid_2d.cpp``:
```

contributing/development/handling_compatibility_breakages.rst
```diff
@@ -0,0 +1,141 @@
...
+
+The instructions for how to add to that file are at the top of the file itself.
+
+If you get a "Hash changed" error for a method, it means that the compatibility binding is missing or incorrect.
+Such lines shouldn't be added to the ``.expected`` file, but fixed by binding the proper compatibility method.
+
+And that's it! You might run into a bit more complicated cases, like rearranging arguments,
+changing return types, etc., but this covers the basic on how to use this system.
+
+For more information see `pull request #76446 <https://github.com/godotengine/godot/pull/76446>`_
```
Bromeon, line 141:
```suggestion
For more information, see `pull request #76446 <https://github.com/godotengine/godot/pull/76446>`._
```

contributing/development/handling_compatibility_breakages.rst
```diff
@@ -0,0 +1,141 @@
...
+================================
+
+.. TODO: Elaborate on types of compatibility and procedure.
+
+So you've added a new parameter to a method, changed the return type,
+changed the type of a parameter, or changed its default value,
+and now the automated testing is complaining about compatibility breakages?
+
+Breaking compatibility should be avoided, but when necessary there are systems in place
+to handle this in a way that helps to make it as smooth as possible.
```
Bromeon, line 13:
```suggestion
Breaking compatibility should be avoided, but when necessary there are systems in place
to handle this in a way, which makes the transition as smooth as possible.
```

Bromeon (commented):
Thanks a lot for the updates, I agree it sounds really good!

Some minor comments, admittedly a bit nitpicky 😀

contributing/development/handling_compatibility_breakages.rst
AThousandShips: This is not what it is meant to say, and there's no "new" method, the method has just changed signature, but can clarify the description of the bind part, but to clarify it means that it should match the method bind `ClassDB::bind_method`

contributing/development/handling_compatibility_breakages.rst
AThousandShips: Did so but it's hard not to get verbose, leaving this for a follow up PR

contributing/development/handling_compatibility_breakages.rst
Bromeon: "New" as in "with the new signature", maybe this could be clarified.
I'm just not sure if "normal" is that obvious in this context, but you can also leave it if you think it is 🙂

AThousandShips: I think the example code should make it obvious what the text means, but can look at clarifying some details

contributing/development/handling_compatibility_breakages.rst
AThousandShips: I'm not sure about breaking the sentence there it feels quite abrupt to me

contributing/development/handling_compatibility_breakages.rst
AThousandShips: I don't think it's necessarily too verbose either, but I think some of the details like that can be added later and just get the functional examples needed to help with development out there

contributing/development/handling_compatibility_breakages.rst
Bromeon: I'm pretty sure it's correct though. "Next" acts as an adverb like "Furthermore, ..." or "Additionally, ...".

I didn't find a very good source on this, here are still some: [1](https://www.pristineword.com/comma-next/) [2](https://linguaholic.com/linguablog/comma-after-next/) [3](https://www.chicagomanualofstyle.org/qanda/data/faq/topics/Punctuation/faq0001.html) [4](https://www.scribbr.com/category/commas/)

contributing/development/handling_compatibility_breakages.rst
```diff
@@ -0,0 +1,141 @@
...
+================================
+
+.. TODO: Elaborate on types of compatibility and procedure.
+
+So you've added a new parameter to a method, changed the return type,
+changed the type of a parameter, or changed its default value,
+and now the automated testing is complaining about compatibility breakages?
+
+Breaking compatibility should be avoided, but when necessary there are systems in place
+to handle this in a way that makes the transition as smoothly as possible.
```
Bromeon, line 13:
```suggestion
to handle this in a way that makes the transition as smooth as possible.
```
This was better before, "smooth" is an adjective referring to "transition".

"smoothly" as an adverb would refer to a verb, e.g. "it transitions smoothly" :wink:

AThousandShips: Yep, realized after, only read part of the sentence when fixing

Bromeon (approved):
Thanks a lot for writing this down and the continuous updates!

AThousandShips:
I've been planning to do it for a long time and it'll save myself work too as I've largely been the one to instruct how to do this on PRs 😅

akien-mga (approved):
Looks great!

akien-mga:
Thanks, that's a great piece of documentation for contributors :tada:

AThousandShips:
Thank you! I'll experiment with some improvements to some of the wording later and we can discuss how to phrase the policy as well here

mhilbrunner:

Issue #9108, opened 2024-03-17
Complete the example for "skip_vertex_transform"


huwpascoe:
**Issue description:**

The example for skip_vertex_transform stops with a comment:
> same as above for binormal and tangent, if normal mapping is used

Yes, the reader could google or gpt the correct algorithms eventually and transcribe them into godot shading language, but please let's have the full example as "how godot does it"

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/4.2/tutorials/shaders/shader_reference/spatial_shader.html

```
shader_type spatial;
render_mode skip_vertex_transform;

void vertex() {
    VERTEX = (MODELVIEW_MATRIX * vec4(VERTEX, 1.0)).xyz;
    NORMAL = normalize((MODELVIEW_MATRIX * vec4(NORMAL, 0.0)).xyz);
    // same as above for binormal and tangent, if normal mapping is used
}
```

clayjohn:
I guess it wouldn't hurt to add two lines, overall it's only adding one line and it would be much clearer to spell out the code even though it literally is the exact same as for NORMAL

```
BINORMAL = normalize((MODELVIEW_MATRIX * vec4(BINORMAL, 0.0)).xyz);
TANGENT = normalize((MODELVIEW_MATRIX * vec4(TANGENT, 0.0)).xyz);
```

PR #9181, opened 2024-04-03
Complete the example for "skip_vertex_transform"

dawei-wang:
Fix godotengine/godot-docs#9108


clayjohn (approved):

AThousandShips (changes requested):

PR #9185, opened 2024-04-05
Specified that get_floor_normal() does not return the surface normal

Gamepro5:
Specified that get_floor_normal() does not return the surface normal, but rather the collision normal.

Also see https://github.com/godotengine/godot-proposals/issues/8324


AThousandShips:

PR #9213, opened 2024-04-11
Add information on how to enable debugging in Visual Studio for C#

fyndor:
Added instructions in the C# section on how to setup debugging in Visual Studio.  When I first tried Godot a few weeks ago, I gave up as soon as I thought there was no debugging support yet.  I think if you support C#, you really need to make it easy for people to be able to figure out how to debug their code.   I believe the instructions for VS Code would be fairly similar.  I have not contributed before.  I assume note is the appropriate choice here, but not 100%.  If there is a better way to format the fields I reference or anything else, I would be happy to tweak it.


raulsntos:
Thanks for contributing to the documentation! I think adding a note about debugging in the getting started section makes sense but I'd rather not duplicate documentation, and the instructions for using external editors (including debugging with them) should already be mentioned in the [_C# basics_](https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_basics.html#configuring-an-external-editor) documentation page.

It does look like our _C# basics_ documentation page currently lacks any information about debugging with Visual Studio, so how about adding this information there and linking to this page from the getting started section? This page already includes instructions to setup debugging with VSCode so you can use that as reference.

Also, it looks like you need to rebase in order to get rid of unrelated commits. Take a look at the [_Pull request workflow_](https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html#the-interactive-rebase) documentation page for more information about rebasing, and don't hesitate to reach out in the [development chat](https://chat.godotengine.org) if you need any help.

fyndor:
Sounds good.  I am not very experienced with contributing to open source, but I could tell that commit somehow got mucked up.  I haven't read it yet, but the plan is to start over fresh with a new commit/pull request following your workflow, so this PR can be killed if that works for you.  

Also glancing at it, the VS Code instructions points to a GitHub issue as an example configuration.  While this works, this doesn't seem a like a great long term solution.  It is an external reference to the docs themselves, and it links to a comment of a person.  I'm sure she is a great person and would never do anything nefarious, but just as a general rule I wouldn't want my project unnecessarily using someone's comment, that I don't have control over, as documentation content.  Would you guys be ok with moving that VS Code configuration example into the docs proper, so the docs repo has full control over that content rather than a 3rd party?

raulsntos:
> I haven't read it yet, but the plan is to start over fresh with a new commit/pull request following your workflow, so this PR can be killed if that works for you.

Yes, that works. Just add a comment here linking to the new PR so we can keep track of it.

> Would you guys be ok with moving that VS Code configuration example into the docs proper,

Yes, the current instructions were only meant to be a temporary solution until we updated the VSCode extension. Unfortunately this is taking longer than expected, so it makes sense to me to move the instructions to the docs and we can update them later when we have a working 4.x extension.

fyndor:
I created a new pull request https://github.com/godotengine/godot-docs/pull/9222

PR #9215, opened 2024-04-11
Simplify the MSVC install instructions

chrisl8:
This line can be simplified from:
`If installing Visual Studio 2017 or 2019, make sure to enable **C++** in the list of workflows to install.`
To:
`Make sure to enable **C++** in the list of workflows to install.`
As we **only** support 2019+ anyway.

This line is not needed as we don't support 2015 anymore: `If installing Visual Studio 2015, make sure to run a **Custom** installation instead of **Typical** and select **C++** as a language there.`

This can be simplified from:
`If you've already made the mistake of installing Visual Studio without C++ support, run the installer again; it should present you a **Modify** button.` To:
`If you've already installed Visual Studio without C++ support, run the installer again; it should present you a **Modify** button.`

From my checking on my system, this line is no longer true, so we can remove it: `Running the installer from *Add/Remove Programs* will only give you a **Repair** option, which won't let you install C++ tools.`

Now we can move the simplified text up so that the reader does not have to dig through the entire document to find that important bit of information, while not making the text at the top much longer due removing outdated information.

### DIFF
--- a/contributing/development/compiling/compiling_for_windows.rst
+++ b/contributing/development/compiling/compiling_for_windows.rst
@@ -17,8 +17,9 @@ For compiling under Windows, the following is required:
 
 - `Visual Studio Community <https://www.visualstudio.com/vs/community/>`_,
   version 2019 or later. Visual Studio 2022 is recommended.
-  **Make sure to read "Installing Visual Studio caveats" below or you
-  will have to run/download the installer again.**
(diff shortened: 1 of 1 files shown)


aaronfranke (approved):

skyace65 (approved):

skyace65:
Thanks!

PR #9225, opened 2024-04-13
Add links to the stable documentation offline builds

pizzafaceopossum:
Stable documentation is built but not linked on the main page, so I added links to the stable build next to the links for the latest build.


mhilbrunner:
I would actually switch those around. Stable should be the default link, with master being the second, optional one. Most users probably don't want docs for an in-development version that may be wrong, untested and still in flux.

Would be also great if you could squash your commits and rebase on recent master, as we fixed an error that causes the CI failure in this PR. See [here](https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html#the-interactive-rebase) for some info on how to do that, or ask for assistance here or in Rocketchat if needed :)

Thanks for contributing!

PR #9242, opened 2024-04-17
Add new `--import` command-line option

mihe:
This adds documentation for the new `--import` command-line option that was introduced in godotengine/godot#90431.


Calinou (approved):
Thanks!

PR #9246, opened 2024-04-18
Release policy: Bump 4.3 ETA to June

akien-mga:
We're close to beta, but not started yet, so it's definitely not releasing in April. A release in May is plausible, but I don't want to make too optimistic promises, so let's communicate June as a more realistic estimate.


mhilbrunner (approved):

mhilbrunner:
Makes sense. Merged.

akien-mga:
Cherry-picked for 4.2.

PR #9297, opened 2024-04-29
Remove unused references that carried over from Godot 3.

RadiantUwU:


AThousandShips:
I suspect they're there to make it clear to people how to access this, but might be moved instead, but I think just removing them will make it harder for people to migrate code easily, they're not there just because they were copied over, hence the note

clayjohn:
Yes, these are documented intentionally to point users to use the new `hint_screen_texture` and `hint_depth_texture` instead. We should not remove them. 

Thank you anyway! We appreciate the effort.

PR #9303, opened 2024-04-29
document 0, 90, 180, and 270 degree TileMap cell rotations

Goldenlion5648:
Documents how to simulate 0, 90, 180, and 270 degree rotations for tiles


AThousandShips:

PR #9308, opened 2024-05-02
Added external_resources.rst

computer-science-student:
I would add a page that provides links to step-by-step tutorials as a marketing device that generates awareness of Godot, gets people to use Godot and gets people to make recommendations on what Godot can do to be more useful to them.

The intended audience is potential game developers. These are game developers who are considering using Godot and need something to help them make the decision. I propose to provide links to video tutorials, so that these potential new game developers can explore the ins-and-outs of Godot so that they can make an informed decision. A secondary audience are people who want to provide how-to step-by-step tutorials. This will provide a place for them to put links to their tutorials.

When I was first getting started with Godot, I had trouble finding step-by-step tutorials. This is not the point of Godot’s documentation. For this reason, I think adding another page to Godot’s documentation that shows a list of external links for learning Godot would help. Godot’s documentation wasn’t intended to include step-by-step tutorials or how-to guides. I find those very useful, so I would add a page to the documentation that links external resources for those how-to-guides and step-by-step tutorials. New game developers should find this list of resources in the documentation useful for learning how to get started.

### DIFF
--- a/getting_started/introduction/external_resources.rst
+++ b/getting_started/introduction/external_resources.rst
@@ -1,0 +1,118 @@
+.. _doc_introduction_to_godot:
+
+Introduction to Godot
+=====================
+
+This article is here to help you figure out whether Godot might be a good fit
+for you. We will introduce some broad features of the engine to give you a feel
+for what you can achieve with it and answer questions such as "what do I need to
+know to get started?".
+
+This is by no means an exhaustive overview. We will introduce many more features
+in this getting started series.
+
+What is Godot?
+--------------
+
+Godot is a general-purpose 2D and 3D game engine designed to support all sorts
+of projects. You can use it to create games or applications you can then release
+on desktop or mobile, as well as on the web.
+
+You can also create console games with it, although you either need strong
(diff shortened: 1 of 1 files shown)


skyace65:
We already have a page that does this in the community section of the documentation. Page can be found [here](https://docs.godotengine.org/en/stable/community/tutorials.html).

We do not currently recommend any paid resources like this in the official docs, nor do I think we should. That's opening a large can of worms. We do direct people towards companies for console ports, however right now the foundation reviews who gets listed and makes decisions on a case by case basis. Putting all that aside. while I can appreciate that you want to give people a humble bundle coupon. The docs do not get updated frequently enough that people not on the master branch would see something like that.

Two more things. One, we ask that you keep all PRs for the documentation and engine to 1 commit, if you don't know how to do that we have a guide [here](https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html#modifying-a-pull-request). And lastly if you want to make a PR but don't want it reviewed yet you can mark it as a draft pull request to make that clear. there should be a dropdown button for the "mark PR" button or whatever it's called.

PR #9309, opened 2024-05-02
Add info on script template folders

### LINKED ISSUE #6815: Script Templates should state that each template must be in their own subdirectory
**Your Godot version:**
v4.0.rc4.official [e0de3573f]

**Issue description:**
Script Templates don't work unless they're in their own subdirectories such as `/script_templates/node/`, the documentation doesn't make any mention of this.

I also still have no idea at all how to create a Shader Template.

https://docs.godotengine.org/en/stable/tutorials/scripting/creating_script_templates.html

skyace65:
State that casing doesn't matter, but adding a "_" to a node name will not work. Also explicitly state that if a script isn't in the proper folder it will not be detected. Closes #6815.

### DIFF
--- a/tutorials/scripting/creating_script_templates.rst
+++ b/tutorials/scripting/creating_script_templates.rst
@@ -63,7 +63,9 @@ where:
 
 * ``template_path`` is one of the 2 locations discussed in the previous two sections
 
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):

mhilbrunner:
Thanks!

mhilbrunner:

PR #9315, opened 2024-05-02
Add a note to the vendor plugin info that it no longer needs to be ma…

haplionheart:
…nually enabled

Backporting @BastiaanOlij note from the master branch (commit #dd4c243) as it also applies to Godot 4.2


AThousandShips:
Thank you for your contribution, however this will be done as part of a cherry pick, not manually

Closing, but thank you for your contribution nonetheless

PR #9316, opened 2024-05-02
Add a `super` keyword to the table of keywords

Chaosus:


tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -187,6 +187,8 @@ in case you want to take a look under the hood.
 +------------+---------------------------------------------------------------------------------------------------------------------------------------------------+
 | self       | Refers to current class instance.                                                                                                                 |
 +------------+---------------------------------------------------------------------------------------------------------------------------------------------------+
+| super      | Refers to base class instance. See `Inheritance`_.                                                                                                |
```
dalexeev, line 190: I think that `super` does not refer to anything, but _resolves_ the scope of the parent method. You can pass `self` as a first class value, but you cannot use `super` on its own.

Chaosus: "Resolves the scope of the base class" - sound better?

Chaosus: Done, check again.

dalexeev (approved):

PR #9329, opened 2024-05-05
Change "Editor > Convert Icons With Editor Theme" to "Editor > Convert Colors With Editor Theme"

StefanH-AT:
That's what the option is actually called https://docs.godotengine.org/en/stable/tutorials/assets_pipeline/importing_images.html#doc-importing-images-editor-import-options

Also are you sure png icons should be 16x16? I created a 16x16 icon and it was only half as big as it had to be. However, I'm using a 4K screen with 200% application scaling, so this might just be because of that.


skyace65 (approved):

skyace65:
Thanks! Unfortunately I don't have an answer for you on png icons. Also, congrats on your first merged PR!

PR #9333, opened 2024-05-06
Add "Using System;" line to default C# script

brgillis:
This line appears in the default created script, but was missing the section showing what we should expect to see.

### DIFF
--- a/getting_started/step_by_step/scripting_first_script.rst
+++ b/getting_started/step_by_step/scripting_first_script.rst
@@ -101,6 +101,7 @@ the following line of code:
  .. code-tab:: csharp C#
 
     using Godot;
+    using System;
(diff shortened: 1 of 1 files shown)


raulsntos (commented):

We usually don't add unused `using`s to code snippets, but in this case since this is the first script tutorial it makes sense to match the exact contents of the template to avoid confusing beginners.

We should also change the _Complete script_ code snippet in line 330:

https://github.com/godotengine/godot-docs/blob/97eb0806e618fc85c185ccd2e3dcf5f65321d637/getting_started/step_by_step/scripting_first_script.rst?plain=1#L330-L331

skyace65:
Superseded by #9603

PR #9335, opened 2024-05-06
Update Range step default value and overrides

VojtaStruhar:
According to following snippet, `Range`'s default `step` value is 1, not 0.01. I updated the docs accordingly

https://github.com/godotengine/godot/blob/d8aa2c65a9f857e86d0c1fc1cc6b95b8ccf23099/scene/gui/range.h#L39-L52

- Update `Range`'s default `step` value to 1
- `EditorSpinSlider`, `Slider`, `SpinBox` and `TextureProgressBar` don't override the step value anymore, so there is no need to mention it in their properties
- `ProgressBar` is now the only one that truly overrides the step value, so I added this to its properties


AThousandShips:

VojtaStruhar:
Oops, sorry about that. I proposed the same changes at https://github.com/godotengine/godot/pull/91629

PR #9337, opened 2024-05-07
Updated step 3 in the scene setup section

fanelenan7:
I'm very new to Godot, and as a good first step I have decided to follow the official docs for guidance. When I got to the Signals portion, I have followed everything as closely as possible but still ended up getting errors. Turns out step 3 was incorrect - perhaps it was correct at some point but it wasn't updated. Instead of dragging the `sprite_2d.tscn` on top of Node2D you needed to put it under. That cost me an hour of pulling my hair out. In the end, good folks in the Discord channel helped me figure it out. But I figured I might as well fix this so someone else who is just starting with Godot doesn't spend an hour thinking they are bad at programming.


Calinou:
If your scene currently only has a single node, drag-and-dropping another scene or node on top is identical to drag-and-dropping it under.

Also, the screenshot was made with a non-default theme and scale. Please use the default editor theme and 100% editor scale for screenshots (or downsize the screenshot in an image editor if taken at an editor scale above 100%).

PR #9358, opened 2024-05-10
Document behavior of modulo operator

qwertystop:
Different languages and contexts have different default behaviors for modulo where one or both operands are negative. "The remainder of an integer division" is underspecified. As such, clarification is in order.

### DIFF
--- a/tutorials/scripting/evaluating_expressions.rst
+++ b/tutorials/scripting/evaluating_expressions.rst
@@ -44,7 +44,7 @@ The following operators are available:
 | Division (``/``)       | Performs and integer division if both operands are integers.                        |
 |                        | If at least one of them is a floating-point number, returns a floating-point value. |
 +------------------------+-------------------------------------------------------------------------------------+
-| Remainder (``%``)      | Returns the remainder of an integer division.                                       |
+| Remainder (``%``)      | Returns the remainder of an integer division (modulo).                              |
 |                        | The result will always have the sign of the dividend.                               |
 +------------------------+-------------------------------------------------------------------------------------+
 
--- a/tutorials/scripting/evaluating_expressions.rst
+++ b/tutorials/scripting/evaluating_expressions.rst
@@ -45,6 +45,7 @@ The following operators are available:
 |                        | If at least one of them is a floating-point number, returns a floating-point value. |
 +------------------------+-------------------------------------------------------------------------------------+
 | Modulo (``%``)         | Returns the remainder of an integer division.                                       |
+|                        | The result will always have the sign of the dividend.                               |
 +------------------------+-------------------------------------------------------------------------------------+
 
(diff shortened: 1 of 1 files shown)


dalexeev:
It would be great to use a description consistent with the GDScript modulo operator. `Expression` supports utility functions too.

https://github.com/godotengine/godot-docs/blob/183d0e3147de1c9de6bcc58d33a6db97d920673b/tutorials/scripting/gdscript/gdscript_basics.rst?plain=1#L330-L331

tutorials/scripting/evaluating_expressions.rst
```diff
@@ -45,6 +45,7 @@ The following operators are available:
 |                        | If at least one of them is a floating-point number, returns a floating-point value. |
 +------------------------+-------------------------------------------------------------------------------------+
 | Modulo (``%``)         | Returns the remainder of an integer division.                                       |
+|                        | The result will always have the sign of the dividend.                               |
```
aaronfranke, line 48: Could be improved further:

```suggestion
| Remainder (``%``)      | Returns the remainder of an integer division.                                       |
|                        | The result will always have the sign of the dividend.                               |
```

The behavior in GDScript is Remainder, not Modulo, which is what Python's `%` operator does.

qwertystop: That this page uses "modulo" and the other (linked by @dalexeev ) uses "remainder" led to me only finding this and not the other when I searched – terminology around this is often a bit fuzzy. Even if one is more technically accurate than the other, it might be useful to have both in docs to make sure it is relatively searchable.

qwertystop:
That this page uses "modulo" and the other (linked by @dalexeev ) uses "remainder" led to me only finding this and not the other when I searched – terminology around this is often a bit fuzzy. Even if one is more technically accurate than the other, it might be useful to have both in docs to make sure it is relatively searchable.

tutorials/scripting/evaluating_expressions.rst
```diff
@@ -44,7 +44,8 @@ The following operators are available:
 | Division (``/``)       | Performs and integer division if both operands are integers.                        |
 |                        | If at least one of them is a floating-point number, returns a floating-point value. |
 +------------------------+-------------------------------------------------------------------------------------+
-| Modulo (``%``)         | Returns the remainder of an integer division.                                       |
+| Remainder (``%``)      | Returns the remainder of an integer division.                                       |
```
Calinou, line 47: For SEO goodness 🙂 

```suggestion
| Remainder (``%``)      | Returns the remainder of an integer division (modulo).                              |
```

Calinou (approved):

aaronfranke (approved):

aaronfranke:
Thanks!

PR #9366, opened 2024-05-12
[4.2] Update dependencies in Compiling for Linux/*BSD

Calinou:
- `4.2` version of https://github.com/godotengine/godot-docs/pull/9357 with Wayland references removed.

___

- Use `sudo` on distributions that are set up to use it by default.
- Use unattended install-friendly commands (e.g. `-y` switch).
- Mention required commands to run before installing, such as `apt-get update`.
- Fix install commands on some distributions:
  - Alpine no longer packages libexecinfo since 3.17, so it was removed from the list.
  - Solus requires `-c` for `system.devel`.
- Reorder distributions so they're in alphabetical order.

### DIFF
--- a/contributing/development/compiling/compiling_for_linuxbsd.rst
+++ b/contributing/development/compiling/compiling_for_linuxbsd.rst
@@ -33,7 +33,7 @@ required:
 - Development libraries:
 
   - X11, Xcursor, Xinerama, Xi and XRandR.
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):

mhilbrunner:
Thanks for the backport!

PR #9369, opened 2024-05-13
Fix hashing context example

davidpeck-littlemonkey:
The example function produces invalid hashes for files that aren't a multiple of `CHUNK_SIZE` bytes in size

Fixes two issues:

+ replaces use of `eof_reached()` who's docs say to use `file.get_position() < file.get_len()` instead when looping
+ uses remaining number of bytes instead of `CHUNK_SIZE` if `CHUNK_SIZE` would be too large


AThousandShips:

PR #9373, opened 2024-05-14
OpenXR: Add documentation page about the new composition layers

BastiaanOlij:
This new help page documents the new composition layer support @dsnopek added to Godot.

I've detailed a simple example project for this. I'll see if I can upload that somewhere so we can reference the source.


tutorials/xr/openxr_composition_layers.rst
```diff
@@ -0,0 +1,294 @@
+.. _doc_openxr_composition_layers:
+
+OpenXR composition layers
+=========================
+
+Introduction
+------------
+
+In XR games you generally want to create user interactions that happen in 3D space 
+and involve users touching objects asif they are touching them in real life.
```
mhilbrunner, line 10:
```suggestion
and involve users touching objects as if they are touching them in real life.
```

tutorials/xr/openxr_composition_layers.rst
```diff
@@ -0,0 +1,294 @@
...
+
+Sometimes however it is unavoidable to create more complex UIs
+using a more traditional 2D interface.
```
mhilbrunner, line 13:
```suggestion
Sometimes however it is unavoidable to create a more traditional 2D interface.
```

tutorials/xr/openxr_composition_layers.rst
```diff
@@ -0,0 +1,294 @@
...
+In XR however you can't just add 2D components to your scene.
+Godot needs depth information to properly position these elements so they appear at
+a comfortable place for the user.
+Even with depth information there are headsets with slanted displays that make it impossible 
+for the standard 2D pipeline to correctly render the 2D elements.
+
+The solution then is to render the UI to a :ref:`SubViewport <class_subviewport>`
+and display the result of this using a :ref:`ViewportTexture <class_viewporttexture>` on a 3D mesh.
+The build in :ref:`QuadMesh <class_quadmesh>` is suitable option for this.
```
mhilbrunner, line 22:
```suggestion
The :ref:`QuadMesh <class_quadmesh>` is a suitable option for this.
```

tutorials/xr/openxr_composition_layers.rst
```diff
@@ -0,0 +1,294 @@
...
+
+.. note::
+    See the `Gui in 3D <https://github.com/godotengine/godot-demo-projects/tree/master/viewport/gui_in_3d>`_
```
mhilbrunner, line 25:
```suggestion
    See the `GUI in 3D <https://github.com/godotengine/godot-demo-projects/tree/master/viewport/gui_in_3d>`_
```

tutorials/xr/openxr_composition_layers.rst
```diff
@@ -0,0 +1,294 @@
...
+    example project for an example of this approach.
+
+The problem with displaying the viewport in this way is that the rendered result
+is sampled for lens distortion by the XR runtime and this can lead to a lot of quality loss 
+making text in the UI hard to read.
```
mhilbrunner, line 30:
```suggestion
is sampled for lens distortion by the XR runtime and the resulting quality loss 
can make UI text hard to read.
```

tutorials/xr/openxr_composition_layers.rst
```diff
@@ -0,0 +1,294 @@
...
+
+OpenXR offers a solution to this problem through its composition layers implementation.
```
mhilbrunner, line 32:
```suggestion
OpenXR offers a solution to this problem through composition layers.
```

mhilbrunner (commented):
Only reviewed the first half so far.

tutorials/xr/openxr_composition_layers.rst
```diff
@@ -0,0 +1,294 @@
...
+
+
+Hole punching
+-------------
+
+This is a feature that is specifically targeted at AR use cases as it requires our
+main viewport to be configured with a transparent background.
+
+As the composition layer is composited on top of the render result,
+it can be rendered infront of objects that are actually forward of the viewport.
```
devloglogan, line 285:
```suggestion
it can be rendered in front of objects that are actually forward of the viewport.
```

tutorials/xr/openxr_composition_layers.rst
```diff
@@ -0,0 +1,294 @@
...
+
+So far we're only displaying our UI, to make it work we need to add some code.
+For this example we're going to keep things simple and
+make one of the controllers work as a pointer.
+We'll then simulate mouse actions with this pointer.
+
+This code also requires a ``MeshInstance3D`` node called ``Pointer`` to be added
+as a child to our ``OpenXRCompositionLayerQuad`` node.
+We configure a ``SphereMesh`` with a radius ``0.01`` meters.
+We'll be using this as a helper to visualise where the user is pointing.
```
devloglogan, line 103:
```suggestion
We'll be using this as a helper to visualize where the user is pointing.
```

tutorials/xr/openxr_composition_layers.rst
```diff
@@ -0,0 +1,294 @@
...
+main viewport to be configured with a transparent background.
+
+As the composition layer is composited on top of the render result,
+it can be rendered infront of objects that are actually forward of the viewport.
+
+By enabling hole punch you instruct Godot to render a transparent object
+where our viewport is displayed.
+It does this in a way that fills the depth buffer and clears the current rendering result.
+Anything behind our viewport will now be cleared,
+while anything infront of our viewport will be rendered as usual.
```
devloglogan, line 291:
```suggestion
while anything in front of our viewport will be rendered as usual.
```

devloglogan (commented):
A few more typos

tutorials/xr/openxr_composition_layers.rst
```diff
@@ -0,0 +1,293 @@
...
+
+            else:
+                was_pressed = false
+                was_intersect = NO_INTERSECTION
+
+
+Hole punching
+-------------
+
+This is a feature that is specifically targeted at AR use cases as it requires our
```
m4gr3d, line 280: I'd add `MR` in addition to `AR` in that sentence to avoid giving the impression we're only referring to AR glasses:

> This is a feature that is specifically targeted at AR and MR use cases...

tutorials/xr/openxr_composition_layers.rst
```diff
@@ -0,0 +1,293 @@
...
+.. code:: gdscript
+
+    ...
+
+            else:
+                was_pressed = false
+                was_intersect = NO_INTERSECTION
+
+
+Hole punching
```
m4gr3d, line 277: It may be worth giving a use-case to illustrate how this is being used in actual project (e.g: adding a virtual window to the environment with passthrough in the 'background'.

devloglogan: I said I was on board with this use case last night at the XR meeting but I need to run that back: hole punched composition layers aren't the correct method for a "window into a virtual world". Since a subviewport can't be rendered in stereo to make it look actually 3D. If you wanted to achieve that effect you'd be better off having the "virtual world" be the main scene, and then have everything that isn't the window be covered up by (hole punched) geometry passthrough.

The main point of hole punching in composition layers is to just let things be displayed in front of the composition layer, like hand models, etc. Adding that use case would make sense.

BastiaanOlij: @devloglogan actually for those type of use cases you're better off rendering geometry directly with the shadow to opacity option and skip composition layers all together.

Hole punching in relation to composition layers is purely meant to allow the composition layer to be rendered behind the main content so that you can have virtual foreground objects occlusion part of the content of the composition layer.

Right now composition layers are always visible on top even if they clip through geometry.

tutorials/xr/openxr_composition_layers.rst
BastiaanOlij: As mentioned during the XR meeting, should have written this down before :) In my humble opinion I want to avoid the MR terminology here. AR and VR reference two distinct approaches, VR referencing fully immersive virtual worlds, while AR references any approach where the real world is augmented with virtual content.

Just for completeness sake in for this discussion, XR is a grouping term that was adopted by Khronos and as such by a part of the industry (including by Godot).

MR is a confusing term that means something different depending on who you talk to.
- Originally it referred to mixing real world capture with virtual world capture to create a composite such as solutions as MixCast, Liv and Oculus own Mixed Reality iPad app implement.
- Then Microsoft adopted MR to mean Mixed Reality as the name for their WMR platform, in this case defining it as a platform supported both AR (through Hololens) and VR (through other WMR devices, none of which do passthrough I believe). I also guess they hoped people would see it as an abbreviation of Microsoft Reality, but that's pure speculation on my part :)
- Then in most recent years MR meaning Mixed Reality has become an alternative to XR as an encompassing term for people who don't like the XR moniker.
- And finally, Meta seems to have rebranded it recently to specifically mean AR through the means of passthrough though I am not sure if that has created more clarification or more confusion in the market space. All though Metas marketing here IMHO is inconsistent as its also often referring to the above meaning as a catch all for AR and VR.

So MR in its meaning in the context of this document as I understand it, e.g. AR through passthrough, is a meaning that is fairly contested.

As Godot is a platform agnostic development tool and we try not to care **how** AR is achieved but want our applications to be portable between various AR devices, it is my personal opinion that AR in this context is sufficient capturing all use cases of this feature, while MR is confusing.

There are features that are only possible on devices that use a camera feed to achieve AR and where passthrough should be named specifically, but this is no such feature as it is as applicable on a passthrough device such as the Quest, as it is on an optical AR device such as the Magic Leap 2.

**However**, I stress that this is my personal standpoint on the matter, so this is absolutely up for debate.

BastiaanOlij:
One more bit of feedback from the XR meeting was the suggestion to add some image captures to the documentation.

Especially seeing we're naming this by the technical solution, not by what you can achieve with this solution, a picture tells a thousand words to let people know what this functionality is used for.

tutorials/xr/openxr_composition_layers.rst
devloglogan: @BastiaanOlij I totally agree, and was attempting to say as much in my previous comment. :) Would not use composition layers when going for that use case.

BastiaanOlij: Ah we're on the same page then :)

devloglogan:
Perhaps an image like this would suffice to illustrate hole punching?

CompositionLayerHolePunch

BastiaanOlij:
Owh thats pretty clear!

m4gr3d:
> Owh thats pretty clear!
> 
> Interesting btw that this shows its also viable for VR, so possibly I should rewrite that part of the text to not point out AR at all.

@BastiaanOlij  +1! A VR specific example would be to use a quad composition layer for text display to improve the text rendering quality.

> hole punched composition layers aren't the correct method for a "window into a virtual world". 

@devloglogan During the meeting, I was suggesting the opposite, showing a window into the real world, which should be doable using the [`OpenXRFbPassthroughGeometry`](https://github.com/GodotVR/godot_openxr_vendors/pull/139) node.

> If you wanted to achieve that effect you'd be better off having the "virtual world" be the main scene, and then have everything that isn't the window be covered up by (hole punched) geometry passthrough.

This would make for an interesting sample, especially if the geometry is destructable so the user can  'break out' from the real world into the virtual world (e.g: similar to what First Encounters does).

BastiaanOlij:
@m4gr3d asked this on rocketchat as well, but I don't see whats special about FB passthrough geometry, we can already do this with the `shadow_to_opacity` shader and it will work cross platform on any AR capable device? 
I'm probably missing something?

tutorials/xr/openxr_composition_layers.rst
```diff
@@ -0,0 +1,292 @@
...
+    When the composition layer is supported,
+    it is the XR runtime that presents the subviewport.
+    This means the UI is only visible in the headset,
+    it will not be accessible by Godot and will thus
+    not be shown when you have a spectator view on the desktop.
+
+There are currently 3 nodes that expose this functionality:
+
+- :ref:`OpenXRCompositionLayerCylinder <class_OpenXRCompositionLayerCylinder>` shows the contents of the SubViewport on the inside of a cylinder.
+- :ref:`OpenXRCompositionLayerEquirect <class_OpenXRCompositionLayerEquirect>` shows the contents of the SubViewport on a warped rectangle.
```
dsnopek, line 51: I think describing an equirect as the interior of sphere makes it easiest to understand:

```suggestion
- :ref:`OpenXRCompositionLayerEquirect <class_OpenXRCompositionLayerEquirect>` shows the contents of the SubViewport on the interior of a sphere (or "slice" of a sphere).
```

tutorials/xr/openxr_composition_layers.rst
```diff
@@ -0,0 +1,292 @@
...
+.. warning::
+    When the composition layer is supported,
+    it is the XR runtime that presents the subviewport.
+    This means the UI is only visible in the headset,
+    it will not be accessible by Godot and will thus
+    not be shown when you have a spectator view on the desktop.
+
+There are currently 3 nodes that expose this functionality:
+
+- :ref:`OpenXRCompositionLayerCylinder <class_OpenXRCompositionLayerCylinder>` shows the contents of the SubViewport on the inside of a cylinder.
```
dsnopek, line 50:
```suggestion
- :ref:`OpenXRCompositionLayerCylinder <class_OpenXRCompositionLayerCylinder>` shows the contents of the SubViewport on the inside of a cylinder (or "slice" of a cylinder).
```

tutorials/xr/openxr_composition_layers.rst
```diff
@@ -0,0 +1,292 @@
...
+        $Pointer.visible = false
+
+        if controller and layer_viewport:
+            var controller_t : Transform3D = controller.global_transform
+            var intersect : Vector2 = intersects_ray(controller_t.origin, -controller_t.basis.z)
+
+    ...
+
+Next we check if we're intersecting with our viewport.
+If so we check if our button is pressed and place our pointer at our intersection point.
```
dsnopek, line 181:
```suggestion
If so, we check if our button is pressed and place our pointer at our intersection point.
```

tutorials/xr/openxr_composition_layers.rst
```diff
@@ -0,0 +1,292 @@
...
+            else:
+                was_pressed = false
+                was_intersect = NO_INTERSECTION
+
+
+Hole punching
+-------------
+
+As the composition layer is composited on top of the render result,
+it can be rendered in front of objects that are actually forward of the viewport.
```
dsnopek, line 281: Is it worth specifically saying something about "virtual hands appearing behind the composition layer"? I feel like that's the first thing developers are going to notice about composition layers in VR.

dsnopek (commented):
This is looking really great! I have just a couple minor notes.

Also, I wonder if it's worth pointing out that an equirect layer can also be used for rendering a sky box? While Godot can of course render the sky, using a composition layer has the same advantages with regard to clarity and performance that using a composition layer for 2D UI has.

This is mentioned specifically in the OpenXR spec:

Selection_163

tutorials/xr/openxr_composition_layers.rst
BastiaanOlij: I just copied these descriptions from the class definition, so I think if we change it here, we should also change it on the class?

dsnopek: Oops, probably my fault then. :-) Yeah, we should update it there as well.

tutorials/xr/openxr_composition_layers.rst
```diff
@@ -0,0 +1,292 @@
...
+
+            else:
+                was_pressed = false
+                was_intersect = NO_INTERSECTION
+
+
+Hole punching
+-------------
+
+As the composition layer is composited on top of the render result,
```
m4gr3d, line 280: This is dependent upon the sort order, so it may be worth clarifying that.

m4gr3d (approved):
Minor comments, but looks good overall!

tutorials/xr/openxr_composition_layers.rst
```diff
@@ -0,0 +1,292 @@
...
+
+If we've just released our button we also prepare
+a :ref:`InputEventMouseButton <class_InputEventMouseButton>` object
+to simulate a button release and send that to our viewport for further processing.
+
+.. code:: gdscript
+
+    ...
+
+                if !is_pressed and was_pressed:
```
AThousandShips, line 226:
```suggestion
                if not is_pressed and was_pressed:
```

tutorials/xr/openxr_composition_layers.rst
```diff
@@ -0,0 +1,292 @@
...
+
+Or if we've just pressed our button we prepare
+a :ref:`InputEventMouseButton <class_InputEventMouseButton>` object
+to simulate a button press and send that to our viewport for further processing.
+
+.. code:: gdscript
+
+    ...
+
+                elif is_pressed and !was_pressed:
```
AThousandShips, line 244:
```suggestion
                elif is_pressed and not was_pressed:
```

tutorials/xr/openxr_composition_layers.rst
BastiaanOlij: I'll adjust it here and then we can update the documentation later.

tutorials/xr/openxr_composition_layers.rst
BastiaanOlij: It does say that later on:
```
You also need to set ``Sort Order`` to a negative value,
the XR compositor will now draw the viewport first, and then overlay our rendering result.
```

tutorials/xr/openxr_composition_layers.rst
BastiaanOlij: Well, we do have the screenshot logan provided, it kind of shows that use case.

dsnopek (approved):

skyace65 (approved):

tutorials/xr/openxr_composition_layers.rst
```diff
@@ -0,0 +1,296 @@
...
+OpenXR composition layers
+=========================
+
+Introduction
+------------
+
+In XR games you generally want to create user interactions that happen in 3D space 
+and involve users touching objects as if they are touching them in real life.
+
+Sometimes however it is unavoidable to create a more traditional 2D interface.
```
skyace65, line 12:
```suggestion
Sometimes however creating a more traditional 2D interface is unavoidable.
```

skyace65 (changes requested):
Only one minor phrasing issue, but after that is fixed this should be good to merge.

skyace65 (approved):

skyace65:
Thanks!

PR #9378, opened 2024-05-15
Update using_tilesets.rst

Josaphat:
Because of https://github.com/godotengine/godot/issues/88440 the guide cannot currently be followed as written. It seems that the behavior change was intentional, but this fixes the guide regardless of whether the old behavior is revived.

I'm not sure exactly what the protocol is for this, but I would imagine this ought to be "backported" since the issue is present since 4.2.0.


skyace65 (changes requested):
Sorry for the long wait for a review. The change itself looks good, I just need you to delete those two image files. using_tilesets_load_tilesheet.webp and using_tilesets_create_new_atlas.webp

skyace65 (approved):

skyace65:
Thanks! And congrats on your first merged PR! In the future please try and keep PRs to 1 commit, if you don't know how to do that we have a guide [here](https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html#modifying-a-pull-request).

PR #9379, opened 2024-05-15
Update compiling_with_script_encryption_key.rst

Morokiane:
Adding in the compiling a PCK with encryption. It may be necessary to note, that when adding the env var to Windows with command prompt, terminal, or Powershell, it might require being ran as admin. Not doing this can lead the the engine unable to unencrypt the PCK at runtime.

### DIFF
--- a/contributing/development/compiling/compiling_with_script_encryption_key.rst
+++ b/contributing/development/compiling/compiling_with_script_encryption_key.rst
@@ -52,7 +52,7 @@ Step by step
    that way you can minimize the risk of exposing the key.
 
 2. Set this key as environment variable in the console that you will use to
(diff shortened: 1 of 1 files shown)


Calinou:
I don't see why you'd need to set the variable as administrator. These commands are only valid for the current terminal session either way (unlike `setx` which is permanent, and user-wide when not run as administrator).

Morokiane:
I would suggest then that the documentation says that it is only for the current terminal session.

skyace65:
Closing per Calinou's comment, if you need to run something as an administrator Godot's been put in a location it probably shouldn't be in. I've confirmed this on my own Windows 10 PC

Issue #9355, opened 2024-05-10
Please remove this code block, or add some explanation to it


anmolp476:
**Your Godot version:** 4.2

**Issue description:** I am confused what this code block is there for. I've read older posts from the r/godot subreddit, and one comment is suggesting that it may be due to the collision between the floor and the player being handled elsewhere. Although I can't see that anymore. Overall, im just super confused what this code block is for. 

```
# If the collision is with ground
		if collision.get_collider() == null:
			continue
```

**URL to the documentation page (if already existing):** https://docs.godotengine.org/en/stable/getting_started/first_3d_game/06.jump_and_squash.html

The referenced code block is in the **Squashing monsters** section

fmnjose:
Just stumbled into this on the documentation as well. The comment is misleading.
This block of code is not to identify if the collision is with the ground, but is indeed to prevent a null pointer that occurs when there are multiple collisions with an enemy in one single frame.

Without this block of code, if there is more than one collision with a mob detected in a single frame, the following will happen:

- The first collision calls the squash function and deletes the mob object
- The second collision will try to get the collider, which will then be null because the object was just deleted
- Calling the collision.getCollider().is_in_group() will fail with a null pointer exception.

PR #9382, opened 2024-05-15
Change misleading comment in 'Your first 3D game'

fmnjose:
On the collision processing code (Chapter 6 - Jumping and squashing monsters) there is a misleading comment saying that the collision.get_collider() == null has the purpose of checking for a collision with the ground, while the purpose is to prevent a null pointer that happens when there are multiple collisions with a mob in a single frame.

This addresses the following open issue: https://github.com/godotengine/godot-docs/issues/9355

### DIFF
--- a/getting_started/first_3d_game/06.jump_and_squash.rst
+++ b/getting_started/first_3d_game/06.jump_and_squash.rst
@@ -241,7 +241,11 @@ With this code, if no collisions occurred on a given frame, the loop won't run.
             # We get one of the collisions with the player
             var collision = get_slide_collision(index)
 
-            # If the collision is with ground
+            # If there are duplicate collisions with a mob in a single frame
+            # the mob will be deleted after the first collision, and a second call to 
+            # get_collider will return null, leading to a null pointer when calling 
+            # collision.get_collider().is_in_group("mob").
+            # This block of code prevents processing duplicate collisions.
             if collision.get_collider() == null:
                 continue
 


Bamarin:
Not against this change, but rather curious: how can the collider be null?

You mentioned this happens after we squash the mob and delete the object (for subsequent collisions with the same mob, the collider will be null).
But we are breaking out of the look right after the squash, so that null check shouldn't be necessary.

Indeed, it's missing in the C# counterpart.
Is it because the `break` statement acts differently in GDscript?

raulsntos:
> Indeed, it's missing in the C# counterpart.
> Is it because the `break` statement acts differently in GDscript?

No, the check also exists in C#:

```csharp
if (collision.GetCollider() is Mob mob)
```

This checks that the collider instance is of type `Mob` **and** that it's not null.

fmnjose:
> Not against this change, but rather curious: how can the collider be null?
> 
> You mentioned this happens after we squash the mob and delete the object (for subsequent collisions with the same mob, the collider will be null). But we are breaking out of the look right after the squash, so that null check shouldn't be necessary.
> 
> Indeed, it's missing in the C# counterpart. Is it because the `break` statement acts differently in GDscript?

I've tried removing the code block, and I started getting null pointers when hitting the mobs from certain angles. 
I printed the collision counters and noticed that in those cases there were multiple collisions with the mob, and the first collision processed would delete the object, inducing a null pointer for the next ones being processed.

Either way the block of code does not serve the purpose of checking for a collision with the ground, as it will never have a null collider.

mhilbrunner (approved):

mhilbrunner:
Merged. Thanks and congrats on your first merged contribution! Sorry this took a while :)

PR #9399, opened 2024-05-19
Fix C# main screen plugin example and document C# `EditorInterface` singleton

easterNday:
Correct the syntax errors in the plugin development section, specifically by modifying `EditorInterface` to `EditorInterface.Singleton` in the C# portion. It is imperative to utilise the singleton pattern here to ensure the plugin successfully compiles.

* Docteam edit, closes: https://github.com/godotengine/godot-docs/issues/9626*

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_differences.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_differences.rst
@@ -363,6 +363,17 @@ Example:
 
     Input.Singleton.JoyConnectionChanged += Input_JoyConnectionChanged;
 
+If you are developing main screen plugins using C#, it is essential to note the 
+distinctions between ``GDScript`` and ``C#`` during setup. Unlike in ``GDScript``, 
+this portion is not a static class in ``C#``. Consequently, you must employ the singleton 
+pattern to obtain an instance of the ``EditorInterface``:
+
+====================  ==============================================================
+GDScript              C#
+====================  ==============================================================
+``EditorInterface``        ``EditorInterface.Singleton``
+====================  ==============================================================
+
 String
 ------
 
--- a/tutorials/plugins/editor/making_main_screen_plugins.rst
+++ b/tutorials/plugins/editor/making_main_screen_plugins.rst
@@ -92,7 +92,7 @@ Add five extra methods such that the script looks like this:
 
         public override Texture2D _GetPluginIcon()
         {
-            return EditorInterface.GetEditorTheme().GetIcon("Node", "EditorIcons");
+            return EditorInterface.Singleton.GetEditorTheme().GetIcon("Node", "EditorIcons");
         }
     }
     #endif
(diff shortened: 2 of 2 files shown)


AThousandShips:
Are you sure this is needed? See [here](https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_differences.html#singletons)

easterNday:
The ·EditorInterface· isn't a static class, and in the course of developing plugins for Godot's main screen, it's essential to utilise the singleton instance of EditorInterface; I fancy this to be a prerequisite. Indeed, attempts to proceed without incorporating this singleton would meet with compilation obstacles, rendering the process unfeasible.
image

paulloz:
> Are you sure this is needed?

Yes, it is totally needed 🙂

paulloz (approved):

AThousandShips:
Then this singleton is an exception, should probably be added to the documentation

easterNday:
I'm not entirely sure if this fully meets your requirements, but I have added a specific exception explanation to that [section](https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_differences.html#singletons).

paulloz:
I think if we want to make a table listing everywhere `.Singleton` is necessary, we should do it exhaustively (could be done in its own PR). IMHO, an incomplete list is as bad as the current status quo.

easterNday:
> I think if we want to make a table listing everywhere `.Singleton` is necessary, we should do it exhaustively (could be done in its own PR). IMHO, an incomplete list is as bad as the current status quo.

You're right, I will follow your advice and complete this list accordingly.

raulsntos:
`EditorInterface` is the only one:

https://github.com/godotengine/godot/blob/daa81bbb7d1c6d75d1711595604178ee62a5801d/modules/mono/editor/bindings_generator.cpp#L4801-L4803

This is because historically this class wasn't registered as a singleton, so when it became one in https://github.com/godotengine/godot/pull/75694 we couldn't make it static without breaking compat.

All the other singletons are static classes, although they also have a `Singleton` property. In this case using the static class or the instance makes no difference.

tutorials/scripting/c_sharp/c_sharp_differences.rst
```diff
@@ -363,6 +363,17 @@ Example:
 
     Input.Singleton.JoyConnectionChanged += Input_JoyConnectionChanged;
 
+If you are developing main screen plugins using C#, it is essential to note the 
+distinctions between ``GDScript`` and ``C#`` during setup. Unlike in ``GDScript``, 
+this portion is not a static class in ``C#``. Consequently, you must employ the singleton 
+pattern to obtain an instance of the ``EditorInterface``:
```
tetrapod00, line 369:
```suggestion
If you are developing main screen plugins, it is essential to note that ``EditorInterface``
is not a static class in C#, unlike in GDScript. Consequently, you must use the
singleton pattern to obtain an instance of the ``EditorInterface``:
```

tetrapod00 (approved):
The current state was also verbally approved by raulsntos in the RC: https://chat.godotengine.org/channel/dotnet?msg=mJnHFdxJM75PhnRWN

Since the author turned off edits by maintainers, I am merging the current acceptable state and will make a followup PR for the wording adjustments.

tetrapod00:
Thank you, and congrats on your first merged contribution to the Godot documentation!

PR #9401, opened 2024-05-20
Document HardLimiter audio bus effect

skyace65:
Documents the new HardLimiter audio bus effect introduced [here](https://github.com/godotengine/godot/pull/89088). The new description for the old limiter effect was taken from that PR. The new webp image has gone through squoosh compression, this is as small as it will get without being lossy.


mhilbrunner (approved):

mhilbrunner:
Thanks!

PR #9403, opened 2024-05-20
Point user to further reading at end of GDExtension example

TvdSmagt:
Based on a discussion I had [here](https://github.com/godotengine/godot-cpp/issues/1467#issuecomment-2118832190), I thought it would be best to point the reader of the GDExtension C++ example to the location of other existing C++ documentation. 

For someone that is new to the documentation, it might not make sense to look for explanation of the core types in the "Contributing" section. This is something that I myself ran into.

### DIFF
--- a/tutorials/scripting/gdextension/gdextension_cpp_example.rst
+++ b/tutorials/scripting/gdextension/gdextension_cpp_example.rst
@@ -698,3 +698,7 @@ Next steps
 We hope the above example showed you the basics. You can
 build upon this example to create full-fledged scripts to control nodes in Godot
 using C++.
+
+To learn more about the types and objects you can use, see the following:
+- :ref:`doc_core_types`
+- :ref:`doc_object_class`


paddy-exe:
I don't think it's a good idea to have this at the end of the this page. To be honest I think even the icons section is unnecessary and should be in a separate page. There's way more pages that could be added here and just having a list of all of them at the end looks more like this is going to be the only docs page for GDExtension. I am already working on more pages that go into details of the system and I would rather have a dedicated "additional resources/docs" section where these pages are linked.

TvdSmagt:
Fair point. I just wanted to make sure that the reader is also pointed to that section in the documentation, as it being listed under "Contributing" is not an intuitive place to look IMO.

If I can help you in any way with these pages you are creating, or if you want me to put this in a separate page with "additional resources/docs", let me know.

paddy-exe:
> Fair point. I just wanted to make sure that the reader is also pointed to that section in the documentation, as it being listed under "Contributing" is not an intuitive place to look IMO.
> 
> If I can help you in any way with these pages you are creating, or if you want me to put this in a separate page with "additional resources/docs", let me know.

Yeah absolutely understandable and I generally fully agree that this information should be linked in the GDExtension docs. Thank you first and foremost for creating this PR in the first place. I think having a separate page called "Additional resources" would be imho great but this should be probably be talked about in the [engine developer chat](https://chat.godotengine.org/channel/gdextension). Would be great to have more people working on these pages so thank you already in advance!

tetrapod00:
If this isn't going to be added *here*, can this PR be closed? Or should we leave this PR open just to track the idea of adding these links?

skyace65:
> If this isn't going to be added _here_, can this PR be closed? Or should we leave this PR open just to track the idea of adding these links?

I'm gonna say close it. We shouldn't have PRs open that we know for a fact will not be merged. Tracking this in an issue would be fine.

PR #9407, opened 2024-05-21
Remove Note from Constants

ceejbro:
From PR [#71051
](https://github.com/godotengine/godot/pull/71051)
Documentation not updated to reflect deep read-only change to const Array and const Dictionary

Remove Note Section describing that Arrays and Dictionaries could be modified after declared constant when that is no longer the case.


dalexeev (approved):
Looks good to me. Constant packed arrays can still be modified, but this is a bug that will probably be fixed at some point. Constant objects (`preload`ed resources) can still be modified, but I don't think we need to emphasize that.

skyace65:

mhilbrunner:

PR #9433, opened 2024-05-29
Add guidelines on inline comments in GDScript style guide

Calinou:
The "two spaces before inline comment" rule follows Python's code style guidelines.
Most languages also tend to favor comments written on their own lines as opposed to inline comments (although using inline comments is not *forbidden*).

For reference, this is a rule I've followed in various demo projects.

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_styleguide.rst
+++ b/tutorials/scripting/gdscript/gdscript_styleguide.rst
@@ -483,8 +483,7 @@ comments from disabled code.
 
 Prefer writing comments on their own line as opposed to inline comments
 (comments written on the same line as code). Inline comments are best used for
-short comments, typically a few words at most. When using inline comments, use
-two spaces before the inline comment to better separate it from code:
+short comments, typically a few words at most:
 
 **Good**:
 
--- a/tutorials/scripting/gdscript/gdscript_styleguide.rst
+++ b/tutorials/scripting/gdscript/gdscript_styleguide.rst
@@ -492,7 +492,7 @@ short comments, typically a few words at most:
 ::
 
     # This is a long comment that would make the line below too long if written inline.
-    print("Example")  # Short comment.
+    print("Example") # Short comment.
 
 **Bad**:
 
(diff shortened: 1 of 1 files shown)


dalexeev:
Looks reasonable to me. The only concern is the double space before inline comments. Yes, this matches the Python style guides, but we have a lot of single space code in the docs and demo projects. Should we now correct this everywhere?

Personally, I don't like double spaces. We do not use double space before `//` inline comments in C++, JavaScript and do not find it less readable. Why do we suddenly need another space before `#` comments? Comments are highlighted in a different color in most editors, GitHub web interface, etc.

Mickeon:
It may look a little prettier, but there's no other guideline about double-spaces anywhere else. This feels like an inconsistency that is easy to forget even if you religiously follow these guidelines.

skyace65:
For whatever my opinion is worth as a doc maintainer I agree that double spacing is unnecessary.

tutorials/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -481,6 +481,28 @@ comments from disabled code.
     :kbd:`Ctrl + K`. This feature adds/removes a single ``#`` sign before any
     code on the selected lines.
 
+Prefer writing comments on their own line as opposed to inline comments
+(comments written on the same line as code). Inline comments are best used for
+short comments, typically a few words at most. When using inline comments, use
+two spaces before the inline comment to better separate it from code:
```
tetrapod00, line 487:
```suggestion
short comments, typically a few words at most:
```

tutorials/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -481,6 +481,28 @@ comments from disabled code.
...
+
+**Good**:
+
+.. rst-class:: code-example-good
+
+::
+
+    # This is a long comment that would make the line below too long if written inline.
+    print("Example")  # Short comment.
```
tetrapod00, line 496:
```suggestion
    print("Example") # Short comment.
```

tetrapod00 (commented):
If it's contentious, I'd recommend not specifying the number of spaces in the text specification (and implicitly using a single space in the examples, as seems to be the general consensus so far)

skyace65 (approved):
Given the current consensus against specifying two lines before inline comments I've committed tetrapod's suggestions.

PR #9434, opened 2024-05-29
Write `class_name` and `extends` on the same line in GDScript style guide

Calinou:
- Follow-up to https://github.com/godotengine/godot-docs/pull/9433.

This saves one line per script with similar readability as before. The syntax also closely follows class declarations in C# or Java. This syntax may be more subject to debate than https://github.com/godotengine/godot-docs/pull/9433, but I've started using it in demos and it grew on me :slightly_smiling_face:

This also renames "docstring" to "class documentation comment" as the term "docstring" is Python jargon that may not be well-understood by everyone.

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_styleguide.rst
+++ b/tutorials/scripting/gdscript/gdscript_styleguide.rst
@@ -28,8 +28,7 @@ Here is a complete class example based on these guidelines:
 
 ::
 
-    class_name StateMachine
-    extends Node
+    class_name StateMachine extends Node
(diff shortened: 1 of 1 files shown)


AThousandShips:
This makes all the script templates no longer follow the style, and it significantly reduces readability to me at least

dalexeev:
> This also renames "docstring" to "class documentation comment" as the term "docstring" is Python jargon that may not be well-understood by everyone.

This is a good change. The first occurrence was proposed for correction in #8920, but we did not notice the second.

PR #9435, opened 2024-05-29
Tweak guidelines on dictionary and enum formatting in GDScript style guide

Calinou:
- Follow-up to https://github.com/godotengine/godot-docs/pull/9434.

___

- Recommend writing each enum item on its own line, similar to C++.
- Recommend using spaces around inline dictionary declarations to make them easier to distinguish from arrays (due to similar-looking `[]` and `{}` characters in most fonts). This is commonly found in Python or JavaScript style guides.


tutorials/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -35,7 +35,6 @@ Here is a complete class example based on these guidelines:
     ## Initializes states and delegates engine callbacks ([method Node._physics_process],
     ## [method Node._unhandled_input]) to the state.
 
-
```
AThousandShips, line 38: This seems unrelated and breaks the general style which has two lines between entries

tutorials/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -46,7 +45,6 @@ Here is a complete class example based on these guidelines:
         set = set_state
     @onready var _state_name = _state.name
 
-
```
AThousandShips, line 49: Same here

tutorials/scripting/gdscript/gdscript_styleguide.rst
Calinou: For this particular case, there isn't a class or function just below (but a variable with a setter), so I don't think two blank lines are needed here.

tutorials/scripting/gdscript/gdscript_styleguide.rst
Calinou: I double-checked and `black` *does* apply two blank lines before the first method, so this is the correct approach.

mhilbrunner (approved):

mhilbrunner:
Thank you!

mhilbrunner:

PR #9438, opened 2024-05-30
change shrink to enlarge in "Your first 3D game" tutorial

aGuyWhoMadeGames:
This fixes an issue mentioned in #6744. In Godot 4, it seems the default size of a spherical collision shape changed from `1` to `0.5`, therefore, the size needs to be increased rather than reduced to get to the final size of `0.8`.


skyace65:

PR #9440, opened 2024-05-31
Fix false positives in codespell CI and fix reported typos

Calinou:
- Don't install codespell when building offline docs as it's not needed in that workflow.


AThousandShips (approved):

mhilbrunner:
Thanks!

PR #9460, opened 2024-06-06
Add 'Signals, concise example' documentation page

manujarvinen:
Hi,
First time documentation updater here.

I'm aware of the Content Guidelines.

This concise Signal example is as short as it possibly can. Unlike the other Signal examples. I for myself tried for almost a month to get Signals to work by reading the manual but simply couldn't understand the Signal mechanics no matter how precisely I read the manual.

So I made this tutorial first and foremost for me to wrap my head around Signals. 
I finally undestand them!

Image-version of this documentation page of mine:
**https://images.manujarvinen.com/signals_concise_example.jpg**

This example of mine tells all the essentials and nothing more. Really hoping it would get included.

**I'll add the C# code portions if you think this is a good idea.**


manujarvinen:
Marked this 'ready for review'.

Hoping for comments before I learn some C# to update the code examples for that language also.

tutorials/scripting/signals_concise_example.rst
```diff
@@ -0,0 +1,52 @@
...
+========================
+
+.. image:: img/signals_concise_example.webp
+
+Goal: Rotate the **yellow gate** by sending a signal from the **red button**.
+
+``red_button.gd``
+
+.. tabs::
+  .. code-tab:: gdscript GDScript
```
AThousandShips, line 13:
```suggestion
 .. code-tab:: gdscript GDScript
```
For all, or the format won't work correctly

manujarvinen: Thank you.
(Still learning the correct procedures)

Initially I used this page to copy the formatting code from, it had the whitespace there - but it seems to work:
https://raw.githubusercontent.com/godotengine/godot-docs/master/tutorials/scripting/resources.rst
https://docs.godotengine.org/en/stable/tutorials/scripting/resources.html

Image snippet of that code from the resources.rst:

Screenshot 2024-06-09 at 18 56 00

AThousandShips: Yes that has one space before `.. code-tab` you have two, you didn't copy it directly

manujarvinen: Oh my, I tried to be very prudent but it seemed not to be enough.

You have some sharp eyes, nice catch.

I'll fix the errors soon and build it locally to see if it works nicely.

Calinou:
If you want a concise example, I think it would be better to improve the [Signal class reference page](https://docs.godotengine.org/en/latest/classes/class_signal.html) and link it at the top of the existing documentation.

manujarvinen:
I added the shak2's suggestions to my proposal, I hope they don't mind. 
Also fixed the ReST formatting typos.
Take a look:
https://images.manujarvinen.com/signals_concise_example_v002.png

But I desperately need help with the CSharp portion of this proposal. I tried to add something but I don't understand it.

Also, in the image there's the **red arrow** which is me asking if this page is put to correct spot?

Other than these, in my opinion this is ready to be merged. What do you think?

Here you can download the Godot Scene in order to implement and test out your CSharp code (this is only GDScript now):
https://files.gearnoodle.com/godot_4.2_signals_concise_example.zip

tutorials/scripting/signals_concise_example.rst
```diff
@@ -0,0 +1,110 @@
+.. _doc_signals_concise_example:
+
+Signals, concise example
+========================
+
+Signals are easy.
+
+Make Signal (in sender):
+
+``Signal SIGNALNAME``
```
AThousandShips, line 10:
```suggestion
``signal SIGNALNAME``
```

AThousandShips (changes requested):

tutorials/scripting/signals_concise_example.rst
```diff
@@ -0,0 +1,110 @@
+.. _doc_signals_concise_example:
+
+Signals, concise example
+========================
```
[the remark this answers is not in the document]
manujarvinen: I'm open to suggestions, but please let's not bloat this up. In my opinion too many words is bad design and certainly won't help anyone, either. In my mind PHP language's manual pages are superior to Godot's because of their concise examples: https://www.php.net/manual/en/function.array - That could serve as inspiration (but not as strict guide, of course)

tutorials/scripting/signals_concise_example.rst
```diff
@@ -0,0 +1,110 @@
...
+
+``Signal SIGNALNAME``
+
+
+Emit Signal (in sender):
+
+``SIGNALNAME.emit(arg1, arg2, ...)``
+
+
+Connect Signal (in receiver):
```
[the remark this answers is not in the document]
manujarvinen: I'm still new to Godot and Signals. That's why I needed this concise clarification. Yes, let's work together to open a little bit more how the developer can utilize signals, but again, being very careful of not adding too long sentences and wording. Even too many examples can be detrimental for understanding.

manujarvinen:
I fixed some code and added a new description, but I'm starting to hate how this is already getting too bloated. Not much more should be added to this concise example.

I would love to get this merged soon to get over it.

Please comment a lot, criticize, point out flaws.

Also, still, the C# code portion is needed :/

Here's how this looks like now:
signals_concise_example_v003

tutorials/scripting/signals_concise_example.rst
```diff
@@ -0,0 +1,113 @@
...
+
+
+Connect Signal (in receivers):
+
+``sender.SIGNALNAME.connect(FUNCTIONNAME)``
+
+
+Make functions (in receivers):
+
+``func FUNCTIONNAME(arg1, arg2, ...):``
```
raulsntos, line 25: All of this is GDScript. You should use `code-tab` to provide the C# equivalent.

I'm a bit worried about using all caps names since users may misunderstand this as the naming convention. I'm also hesitant to provide examples with invalid syntax (such as ellipsis). So I refrained from doing so in the C# examples.

```suggestion
Make Signal (in sender):

.. tabs::
 .. code-tab:: gdscript GDScript

    signal SIGNALNAME

 .. code-tab:: csharp

     [Signal]
     public delegate void SignalNameEventHandler();

Emit Signal (in sender):

.. tabs::
 .. code-tab:: gdscript GDScript

    SIGNALNAME.emit(arg1, arg2, ...)

 .. code-tab:: csharp

     EmitSignal(SignalName.SignalName, arg1, arg2);

Connect Signal (in receivers):

.. tabs::
 .. code-tab:: gdscript GDScript

    sender.SIGNALNAME.connect(FUNCTIONNAME)

 .. code-tab:: csharp

    sender.SignalName += FunctionName;

Make functions (in receivers):

.. tabs::
 .. code-tab:: gdscript GDScript

    func FUNCTIONNAME(arg1, arg2, ...):

 .. code-tab:: csharp

    private void FunctionName(int arg1, int arg2)
```

raulsntos (commented):
I'm only reviewing the C# code.

Overall I'm not sure how useful this page is, considering we already have multiple documentation pages for signals.

Specifically the [_C# signals_](https://docs.godotengine.org/en/latest/tutorials/scripting/c_sharp/c_sharp_signals.html) documentation page covers the same thing and goes into a bit more detail about C#-specific behavior (such as the difference between using `Connect` and the C# event to connect to signals).

The _GDScript basics_ documentation page also has a [_signals_](https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_basics.html#signals) section. But it looks like GDScript doesn't have a dedicated page for signals.

And then there's the [_Using signals_](https://docs.godotengine.org/en/stable/getting_started/step_by_step/signals.html) tutorial which goes in depth on how to use signals, as well as the [`Signal`](https://docs.godotengine.org/en/stable/classes/class_signal.html#class-signal) class reference which has a simple example about declaring signals.

tutorials/scripting/signals_concise_example.rst
```diff
@@ -0,0 +1,110 @@
...
+        {
+            var redButtonArea = GetNode<Area3D>("RedButtonArea3D");
+            redButtonArea.Connect("RedTrigger", this, nameof(OnRedTriggerPress));
+        }
+    
+        private void OnRedTriggerPress(int degreesReceived, Color playerColorReceived)
+        {
+            Rotation = new Vector3(-Mathf.DegToRad(degreesReceived), Rotation.y, Rotation.z);
+            var playerMesh = GetNode<MeshInstance3D>("Player/Mesh");
+            if (playerMesh.Mesh.SurfaceGetMaterial(0) is SpatialMaterial material)
```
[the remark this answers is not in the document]
manujarvinen: Fixed this and tested the CSharp code in a CSharp Godot project

manujarvinen:
Okay, things are quite polished now.
I'll give a shot to squashing commits now (via a tutorial)

manujarvinen:
Is it really this hard?
https://docs.github.com/en/desktop/managing-commits/squashing-commits-in-github-desktop

Should I manually go through every single commit?

Can I just remove this Pull Request and make a new one with a fresh simple small files that are changed?

manujarvinen:
Okay, pheww. Finally squashed the whole thing into one commit. What a mess at first but I think I got hang of it. Not easy gitting, I must say.

So, what's the next step?

manujarvinen:
Now would be a great time to merge this PR since the 4.3 got released, what do you think? This is ready to go. Imo, worthy addition - I've grown to love Signals nowadays, after understanding them from a simple example like this.

manujarvinen:
@AThousandShips @Calinou I'd be pleased to know what to do about this now? Should this go through some tough review still? Thank you for your time :)

tutorials/scripting/signals_concise_example.rst
```diff
@@ -0,0 +1,146 @@
+.. _doc_signals_concise_example:
+
+Signals, concise example
+========================
+
+Signals are easy.
```
adamscott, line 6: I know what you mean, but this is more an opinion than anything else. I'm more in the mindset of showing how easy it is instead of telling how easy it is.

tutorials/scripting/signals_concise_example.rst
```diff
@@ -0,0 +1,146 @@
...
+
+.. tabs::
+ .. code-tab:: gdscript GDScript
+
+    signal SIGNALNAME
+
+ .. code-tab:: csharp
+
+     [Signal]
+     public delegate void SignalNameEventHandler();
```
adamscott, line 18: This should be technically called "Declare signal", but I don't really like the format here. Like @AThousandShips said:

> While concise this section is far too bare, this won't help anyone unless they know some basics, it needs some slightly less minimal descriptions, and it just repeats the same as below in simplified form, which I'd say is just redundant

tutorials/scripting/signals_concise_example.rst
```diff
@@ -0,0 +1,146 @@
...
+
+    func FUNCTIONNAME(arg1, arg2, ...):
+
+ .. code-tab:: csharp
+
+    private void FunctionName(int arg1, int arg2)
+
+Note that there can be multiple receivers receiving the same signal and each of them can have a different function for the same signal.
+
+.. image:: img/signals_concise_example1.webp
```
adamscott, line 55: To be honest, I would be lost trying to decipher the image if I was a pure beginner.

tutorials/scripting/signals_concise_example.rst
```diff
@@ -0,0 +1,146 @@
...
+
+---------
+
+Real-world simple example:
+
+.. image:: img/signals_concise_example2.webp
```
adamscott, line 61: The "real life example" is way to complicated. It assumes a lot, like people knowing about `Area3D` and how they behave. It may seem simple to you, but I don't find this simple.

And to be honest also, the 3D image is way too busy.

tutorials/scripting/signals_concise_example.rst
```diff
@@ -0,0 +1,146 @@
...
+    # yellow_gate.gd
+ 
+    extends StaticBody3D
+
+    func _ready():
+        %RedButtonArea3D.red_trigger.connect(_on_red_trigger_press)
+
+    func _on_red_trigger_press(degrees_received, player_color_received):
+        rotation.x = -deg_to_rad(degrees_received)
+        %"Player/Mesh".mesh.material.albedo_color = player_color_received
```
adamscott, line 118: This doesn't seem simple from my point of view.

adamscott (commented):
Hi there, @manujarvinen ! First and foremost, thanks for your PR.

For me, I don't see your example as a "concise example". The way you present the signals seem to clash with the rest of the manual, which could be fine if it was super clear. But I don't think it's the case.

I know what you're trying to do, and I find this venerable. But I don't think you reached the level you claim. For this, I would argue to not merge this, unfortunately.

tetrapod00:
Thanks for contributing, and for sticking with the review process. However, I am inclined to agree with other reviewers that as-is, this doesn't seem like a great addition to the documentation. As such, we won't be merging this.

In the manual, we typically don't include this kind of short "cheat sheet" example which doesn't include a lot of explanation. A concise example is better suited for the class reference, as Calinou and AThousandShips mentioned. Additionally, there are several other places where signals are already documented, as raulsntos pointed out.

This page *could* be expanded to be more in line with the format of the manual, but the changes would be so drastic that it definitely would not be a "concise example" any more. And since you mentioned that you feel like the page as-is is already getting too bloated, I don't think we'll be able to come to a consensus on what the page should look like.

Not every bit of information is a good fit for the official documentation. This seems like it would be a good community resource, maybe as an infographic or a short blogpost. You could also post it as a comment on any of the existing signals pages, if you want to make it visible to readers of the official documentation (note that you can write and edit such comments on [GitHub directly](https://github.com/godotengine/godot-docs-user-notes/discussions) if you want more advanced formatting options).

Thank you again for contributing, and apologies that this isn't getting merged.

PR #9470, opened 2024-06-07
Update C++ example to use version 4.2

nathantalewis:
I was going through the C++ GDExtension tutorial and noticed that in some places it references 4.0 and in others it references 4.1.

This PR updates it to reference 4.2 everywhere.

It looks like something similar was done on the `4.2` docs branch, but never made it into master. https://github.com/godotengine/godot-docs/tree/1ef005b81b6baa190522bae37fd0fe96ff04b654

### DIFF
--- a/tutorials/scripting/gdextension/gdextension_cpp_example.rst
+++ b/tutorials/scripting/gdextension/gdextension_cpp_example.rst
@@ -363,7 +363,7 @@ loaded for each platform and the entry function for the module. It is called ``g
     [configuration]
 
     entry_symbol = "example_library_init"
-    compatibility_minimum = "4.2"
+    compatibility_minimum = "4.1"
 
     [libraries]
 
--- a/tutorials/scripting/gdextension/gdextension_cpp_example.rst
+++ b/tutorials/scripting/gdextension/gdextension_cpp_example.rst
@@ -35,7 +35,7 @@ of Godot. GDExtensions will not work in older versions of Godot (only Godot 4 an
 .. note::
     To use `GDExtension <https://godotengine.org/article/introducing-gd-extensions>`__
     you need to use the godot-cpp branch that matches the version of Godot that you are
-    targeting. For example, if you're targeting Godot 4.1, use the ``4.1`` branch,
+    targeting. For example, if you're targeting Godot 4.2, use the ``4.2`` branch,
     which is what is shown through out this tutorial.
 
     The ``master`` branch is the development branch which is updated regularly
@@ -59,7 +59,7 @@ a Git submodule:
     mkdir gdextension_cpp_example
     cd gdextension_cpp_example
     git init
-    git submodule add -b 4.1 https://github.com/godotengine/godot-cpp
+    git submodule add -b 4.2 https://github.com/godotengine/godot-cpp
     cd godot-cpp
     git submodule update --init
 
@@ -69,7 +69,7 @@ Alternatively, you can also clone it to the project folder:
 
     mkdir gdextension_cpp_example
     cd gdextension_cpp_example
-    git clone -b 4.1 https://github.com/godotengine/godot-cpp
+    git clone -b 4.2 https://github.com/godotengine/godot-cpp
 
 .. note::
 
@@ -332,7 +332,7 @@ build files in a subsequent tutorial.
 
     This ``SConstruct`` file was written to be used with the latest ``godot-cpp``
     master, you may need to make small changes using it with older versions or
-    refer to the ``SConstruct`` file in the Godot 4.0 documentation.
+    refer to the ``SConstruct`` file in the Godot 4.2 documentation.
 
 Once you've downloaded the ``SConstruct`` file, place it in your GDExtension folder
 structure alongside ``godot-cpp``, ``src`` and ``demo``, then run:
@@ -363,7 +363,7 @@ loaded for each platform and the entry function for the module. It is called ``g
     [configuration]
 
     entry_symbol = "example_library_init"
-    compatibility_minimum = "4.1"
+    compatibility_minimum = "4.2"
 
     [libraries]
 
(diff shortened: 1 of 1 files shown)


skyace65 (commented):
The text changes look good. Does the SConstruct file this page talks about need to be updated?

tutorials/scripting/gdextension/gdextension_cpp_example.rst
```diff
@@ -363,7 +363,7 @@ loaded for each platform and the entry function for the module. It is called ``g
     [configuration]
 
     entry_symbol = "example_library_init"
-    compatibility_minimum = "4.1"
+    compatibility_minimum = "4.2"
```
paddy-exe, line 366: This change is not necessary to do as it's the minimum compatibility (which was introduced in 4.1) for the extension to work and doesn't need an update.

paddy-exe (changes requested):
Looks good except for the change to the `.gdextension file` as there's no need to update that. There is already a WIP docs PR to explain the purpose of these flags.

paddy-exe:
> The text changes look good. Does the SConstruct file this page talks about need to be updated?

Not that I am aware of. It should be very simple and there haven't been any open issues regarding the example file AFAIK

tutorials/scripting/gdextension/gdextension_cpp_example.rst
nathantalewis: Reverted. Thanks for pointing it out!

paddy-exe (approved):

dsnopek:
For `master`, does it makes sense to go all the way to `4.3` (rather than changing it to `4.2)?

paddy-exe:
Hmm now thinking about that it may make more sense to abstract the used version so that we can use the same for every minor version because updating these very minor details in the docs will take time with every release.

nathantalewis:
What type of abstraction would be best? In the tutorial there is already abstraction used in some places, like "platform".

> To generate and compile the bindings, use this command (replacing \<platform\> with windows, linux or macos depending on your OS):

If we want to continue with that type of abstraction, then I can change anywhere that currently has `4.2` to instead use `<godot_version>` with the first instance getting `(replacing <godot_version> with the version number of godot that you are targeting)` or something similar nearby.

dsnopek:
I think that could work, although, it is helpful for folks to see a valid example file so they know the "shape" of the data.

I could also see an argument for just keeping this value as `"4.1"` in all of our examples? This always needs to be specified to avoid compatibility issues with Godot 4.0. But if you compile against the Godot 4.2 version of godot-cpp, it'll do its own checks to ensure you are on 4.2, and other bindings probably do similar checks. So, it isn't necessary to update this to the version of Godot you are targeting in most cases.

paddy-exe:
@dsnopek My main point for having it abstracted a bit is to ease the burden of maintenance. We could still write `we will use version 4.2 as an example here`. This project works the same on all different 4.x versions.

nathantalewis:
So maybe as a compromise between having a valid example file that shows the "shape" of the data and using abstraction to make docs maintenance easier, we can use `4.x` instead of `4.2` or `4.3` with a caveat that they should replace `x` with the minor version they are targeting? This would be instead of the idea of `<godot_version>` for abstraction that was proposed earlier.

So instead of what is currently in the PR:
```
    This ``SConstruct`` file was written to be used with the latest ``godot-cpp``
    master, you may need to make small changes using it with older versions or
    refer to the ``SConstruct`` file in the Godot 4.2 documentation.
```
We would have:
```
    This ``SConstruct`` file was written to be used with the latest ``godot-cpp``
    master, you may need to make small changes using it with older versions or
    refer to the ``SConstruct`` file in the Godot 4.x documentation.
```

It will be a bit less clear in places that people might expect to just copy and paste:
```
git submodule add -b 4.x https://github.com/godotengine/godot-cpp
```

For commands like that, `<godot_version>` makes it more obvious that something needs to be replaced to work:
```
git submodule add -b <godot_version> https://github.com/godotengine/godot-cpp
```

Opinions? @paddy-exe in particular, would you mind offering an example of the type of abstraction you are thinking of? The tutorial already included:
> For example, if you're targeting Godot 4.1, use the ``4.1`` branch, which is what is shown through out this tutorial.

paddy-exe:
@nathantalewis Your suggestion is great and exactly what I imagined in addition to the last comment you referenced. I think this should make it clear enough IMO. What do you think @dsnopek?

nathantalewis:
Alright, PR updated to abstract the minor version where it doesn't lose clarity.

tutorials/scripting/gdextension/gdextension_cpp_example.rst
```diff
@@ -35,16 +35,17 @@ of Godot. GDExtensions will not work in older versions of Godot (only Godot 4 an
 .. note::
     To use `GDExtension <https://godotengine.org/article/introducing-gd-extensions>`__
     you need to use the godot-cpp branch that matches the version of Godot that you are
-    targeting. For example, if you're targeting Godot 4.1, use the ``4.1`` branch,
-    which is what is shown through out this tutorial.
+    targeting. For example, if you're targeting Godot 4.1, use the ``4.1`` branch. Throughout
+    this tutorial we use ``4.x`,` which will need to be replaced with the version of Godot you
```
AThousandShips, line 39:
```suggestion
    this tutorial we use ``4.x``, which will need to be replaced with the version of Godot you
```

nathantalewis: Thanks for catching the typo!

skyace65 (approved):

paddy-exe (approved):
Awesome! Thanks for your patience and ongoing work on this 👍🏻

dsnopek (approved):
Thanks! This seems good to me :-)

skyace65:

nathantalewis:
Thank you!

PR #9480, opened 2024-06-11

mrwnphs:
Updating the Main scene page in the Your First 2D Game tutorial.  Just wanted to clarify the procedure for connecting the signals from the timers to the Main scene.

### DIFF
--- a/getting_started/first_2d_game/05.the_main_game_scene.rst
+++ b/getting_started/first_2d_game/05.the_main_game_scene.rst
@@ -152,8 +152,8 @@ everything up for a new game:
         GetNode<Timer>("StartTimer").Start();
     }
 
-Now connect the ``timeout()`` signal of each of the Timer nodes (``StartTimer``,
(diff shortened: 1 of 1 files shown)


skyace65:
I'm not sure if this is a good change. This feels like an excessive amount of detail considering signals are already explained earlier in the getting started section. Plus the workflow of creating the function and code before connecting is a bit... weird I guess. It works but I don't see it as any better than just copying the code into the function.

mrwnphs:
@skyace65 I cannot speak to your concern right now.  I wrote this a few weeks ago when I was trying to go through the tutorial as a total newbie.  Since then I have gone through many tutorials and I am much more familiar with signals and timers.   When I submit feedback like this, it's usually because something was unclear to me and I spent a lot of time trying to figure out what I was supposed to do.  My suggestions are based on what I thought would have been helpful to me at the time.

skyace65:
Closing due to a lack of support for these changes, and also not updating based on AThousandShips review.

PR #9489, opened 2024-06-14
Add smoothstep graphs

Malcolmnixon:
This PR adds two graphs showing the behavior of the smoothstep function with positive and negative ranges.


BastiaanOlij:
Just for clarity as it's not specified clearly in the OP, Malcolm is adding just images, no documentation changes, here because they are included in the documentation of the smoothstep function inside the main godot repo.

I don't know if there is precedence for this. I don't know if this is the correct way to go about adding images in the class reference, especially seeing the way they are referenced? The images do provide a lot of context to the class reference.

Malcolmnixon:
I decided to merge the graphs into a single comparison image, which also includes showing the zero-range behavior. Additionally the graphs now show the area under the curve to better highlight the function response.

BastiaanOlij (approved):
Looks good, leaving it up to the documentation team to verify this is the way forward.

mhilbrunner:
IIRC this is fine and already done in other parts of the class reference.

Please convert the PNG file to WebP (we're in the process of moving the docs repository over to WebP, see [image creation guidelines](https://docs.godotengine.org/en/stable/contributing/documentation/docs_image_guidelines.html)).

That the linked PR using this image (https://github.com/godotengine/godot/pull/93149) does other stuff than just embedding the image makes it a bit hard to merge, as the main repo PR may get rejected. Please ping me if it gets another approval.

Malcolmnixon:
> Please convert the PNG file to WebP (we're in the process of moving the docs repository over to WebP, see [image creation guidelines](https://docs.godotengine.org/en/stable/contributing/documentation/docs_image_guidelines.html)).

The image has been converted to WebP format.

Calinou (approved):
Image looks good to me. Note that the main repository PR needs to be merged first before this PR can be merged.

Malcolmnixon:
> Image looks good to me. Note that the main repository PR needs to be merged first before this PR can be merged.

It might be better for this to go in first. If the main repository PR goes in then users trying to access the smoothstep documentation will get missing images. Pulling this in first shouldn't break anything - it'll just be available for use when the main repository merge goes through.

clayjohn:
That makes sense. Merging now to be safe

PR #9497, opened 2024-06-16
fix typo in animationmixer

nklbdev:
Small typo fix in code block


skyace65 (approved):

skyace65:
While the fix itself is correct the class reference needs to be edited in the main engine repository. The class reference files in this repository are generated from that. The file you want to edit is here: https://github.com/godotengine/godot/blob/master/doc/classes/AnimationMixer.xml

nklbdev:
Thank you, @skyace65!

PR #9507, opened 2024-06-18
Fix typo in "Optimizing 3D performance"

aGuyWhoMadeGames:
From context, it seems like origin shifting is being introduced as an alternative to large world coordinates, but this is not reflected in the actual wording.


mhilbrunner (approved):

skyace65 (approved):

skyace65:
Thanks!

PR #9512, opened 2024-06-19
Ubuntu APT oneliner to install system builtin alternatives

gio3k:
Might work on Debian, not sure - tested on Ubuntu 24.04 LTS


Calinou:
For reference, I've tested this on a fresh Ubuntu 24.04 and Debian stable Docker containers and can confirm it works.

PR #9514, opened 2024-06-19
Fix wrong lines of code in the "Using NavigationMaps" page

OverloadedOrama:
NavigationServer2D and 3D's `map_set_active()` method [takes two arguments](https://docs.godotengine.org/en/stable/classes/class_navigationserver3d.html#class-navigationserver3d-method-map-set-active), the map RID and the boolean that makes the map active/inactive. The docs only contained the boolean, resulting in wrong code.


smix8 (approved):

mhilbrunner (approved):

mhilbrunner:
Thank you! Merged.

PR #9539, opened 2024-06-24
Fix for loop error.

### LINKED ISSUE #9535: Fix "for loop" error in GDScript reference
**Your Godot version:**
Docs 4.2

**Issue description:**
The section about for loops says:
```
for i in 2.2:
	statement # Similar to range(ceil(2.2)).
```
but this is incorrect as this will output "0, 1" in the Output console.
It should be:
```
for i in 2.2:
	statement # Similar to range(floor(2.2)).
```

https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_basics.html

aGuyWhoMadeGames:
fixes #9535

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_basics.rst
+++ b/tutorials/scripting/gdscript/gdscript_basics.rst
@@ -1565,7 +1565,7 @@ in the loop variable.
         statement # Similar to range(3).
 
     for i in 2.2:
-        statement # Similar to range(ceil(2.2)).
+        statement # Similar to range(round(2.2)).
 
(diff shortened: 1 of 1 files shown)


dalexeev (changes requested):
The current documentation is correct, although probably not clear enough:

```gdscript
func test(num: float) -> void:
    print("===")
    for x in num:
        print(var_to_str(x))

func _run() -> void:
    test(2.0)
    test(2.2)
    test(2.8)
```

```
===
0.0
1.0
===
0.0
1.0
2.0
===
0.0
1.0
2.0
```

PR #9548, opened 2024-06-27
Add warnings for classes named "Editor" in GDscript and C# pages, class_name keyword/global class attribute

DDoop:
Currently classes with the "Editor" prefix in their names are hidden in the Godot editor. This documents that behavior.

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_global_classes.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_global_classes.rst
@@ -80,3 +80,11 @@ will let you create and load instances of this type easily.
 .. image:: img/globalclasses_exportedproperty1.webp
 
 .. image:: img/globalclasses_exportedproperty2.webp
+
+.. warning::
+
+    The Godot editor will hide these custom classes with names that beging with the prefix
+    "Editor" in the 'Create New Node' or 'Create New Scene' dialog windows. The classes 
+    are available for instantiation at runtime via their class names, but are 
+    automatically hidden by the editor windows along with the built-in editor nodes used 
+    by the Godot editor.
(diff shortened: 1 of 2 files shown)


mhilbrunner (approved):
Good to know, this may be something that should be more clearly pointed out in-editor as well (if it isn't already). Documenting it is a good idea.

Mickeon:
> Documenting it is a good idea.

A theoretical warning should be generated automatically, not documented manually for every editor class.

DDoop:
> Good to know, this may be something that should be more clearly pointed out in-editor as well (if it isn't already). Documenting it is a good idea.
> 
> Would be nice if you could squash the commits ([see here](https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html#pull-request-workflow)). Otherwise, I'll just squash while merging.

Thanks for pointing me in this direction, sorry I didn't really follow up. I'll give it a good read or 3. This is my first real contribution to a oss project so thanks for your patience.

tutorials/scripting/c_sharp/c_sharp_global_classes.rst
```diff
@@ -80,3 +80,11 @@ will let you create and load instances of this type easily.
...
 
 .. image:: img/globalclasses_exportedproperty2.webp
+
+.. warning::
+
+    The Godot editor will hide these custom classes with names that beging with the prefix
+    "Editor" in the 'Create New Node' or 'Create New Scene' dialog windows. The classes 
+    are available for instantiation at runtime via their class names, but are 
+    automatically hidden by the editor windows along with the built-in editor nodes used 
+    by the Godot editor.
```
mhilbrunner, line 90:
```suggestion
    by the Godot editor.
    
```

mhilbrunner (approved):

mhilbrunner:
Merged. Thank you!

PR #9551, opened 2024-06-28
Modification of the section title of a_better_xr_start_script.rst

15x3:
## The original section title is somewhat misleading; 
**On session begun** should refer to the **_on_openxr_session_begun** signal in the script code. The same applies to **On visible state**, **On focussed state**, **On stopping state**, and **On pose recentered** (**_on_openxr_visible_state**, **_on_openxr_focused_state**, **_on_openxr_stopping**, **_on_openxr_pose_recentered**).

### DIFF
--- a/tutorials/xr/a_better_xr_start_script.rst
+++ b/tutorials/xr/a_better_xr_start_script.rst
@@ -192,7 +192,7 @@ it is nicer to exit on failure than to hang the system.
     ...
 
 
-On session begun
+_on_openxr_session_begun
 ----------------
 
 This signal is emitted by OpenXR when our session is setup.
@@ -294,7 +294,7 @@ Not matching the physics update rate will cause stuttering as frames are rendere
 
     ...
 
-On visible state
+_on_openxr_visible_state
 ----------------
 
 This signal is emitted by OpenXR when our game becomes visible but is not focussed.
@@ -363,7 +363,7 @@ If you haven't, you can connect a method to the signal that performs additional 
 
     ...
 
-On focussed state
+_on_openxr_focused_state
 -----------------
 
 This signal is emitted by OpenXR when our game gets focus.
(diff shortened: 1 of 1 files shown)


AThousandShips (changes requested):
This is not correct, the way to add these references is like is done at the top of the page, not by replacing the title text

tutorials/xr/a_better_xr_start_script.rst
```diff
@@ -192,8 +192,8 @@ it is nicer to exit on failure than to hang the system.
     ...
 
 
-On session begun
-----------------
+_on_openxr_session_begun
```
AThousandShips, line 195:
```suggestion
.. _on_openxr_session_begun:

On session begun
```
For all these, but only add the ones that are actually referenced

AThousandShips: Unless you want it to link to somewhere else, this wouldn't work here I think, but that would be:
```
 :ref:`On session begun <TARGET>`
```
But there's nothing to link to in the code like that, and it's just there so I don't think that would make sense

Tbh I don't know what you're trying to achieve

15x3: These sections all begin with "This signal" without mentioning the name of the specific signal before or after. Specific  name can only be recognized by looking at the code more closely.

I'm not sure if first time XR developers with Godot will have the same problem when they see this paragraph? So it may no need to linking to other places, the modification is just to replace section title to mention the signals.

AThousandShips: But the name of the signal is the title, it's `session_begun`, not `_on_openxr_session_begun`

I don't think this change makes sense, the only think I'd say would make sense is to link to the signal, which would be:
`class_OpenXRInterface_signal_session_begun`

15x3: I agree that we shouldn't get too hung up on details like that, and I'm not particularly deep into the source code, apologise for taking up some of your time. 🙏

PR #9552, opened 2024-06-28
Remove unnecessary opinion about languages.

Iakobs:
Removing this unnecessary opinion about languages in the `Internationalizing games` section. Since it can be a controversial opinion (it can trigger anti-colonialism folks, or language lovers for that matter), I propose to just remove it.

I understand it is there just as a funny comment, but to be honest, it always shocks me when I see it, and doesn't add any value to the documentation. I also believe that the docs should be as neutral as possible, and this is the main reason why I opened the PR.

### DIFF
--- a/tutorials/i18n/internationalizing_games.rst
+++ b/tutorials/i18n/internationalizing_games.rst
@@ -6,9 +6,7 @@ Internationalizing games
 Introduction
 ------------
 
-Sería excelente que el mundo hablara solo un idioma (It would be great if the
-world spoke only one language). Unfortunately for
-us developers, that is not the case. While indie or niche games usually
+While indie or niche games usually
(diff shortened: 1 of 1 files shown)


AThousandShips:
I think keeping some form of translation example would be very useful, to highlight the need, and especially shift the all too common complacency with English especially not having to understand or engage with other languages. Though I'm not sure what other sentence might fit better

I'm all in favor of dropping the current wording though, and potentially add something else back later.

AThousandShips (approved):

mhilbrunner (approved):

mhilbrunner:
Thank you! Merged. Makes sense.

mhilbrunner:

PR #9558, opened 2024-06-29
Update class_animationmixer.rst

scotmcp:
Improve description of how to set root_node_track property.

The current wording in the documentation isn't clear, I added an example of what this "parent node" might be...In the case of the animation player or the animation tree, this is usually the root node.

### DIFF
--- a/classes/class_animationmixer.rst
+++ b/classes/class_animationmixer.rst
@@ -381,7 +381,7 @@ This makes it more convenient to preview and edit animations in the editor, as c
 - void **set_root_motion_track** **(** :ref:`NodePath<class_NodePath>` value **)**
 - :ref:`NodePath<class_NodePath>` **get_root_motion_track** **(** **)**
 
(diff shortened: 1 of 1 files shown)


AThousandShips:

scotmcp:
Can you please clarify? I have make commits and PRs to and from the repo before, and the 1st link you sent me to sends me to the engine repo. The 2nd link sends me to 4.3 docs, whilst I was working with 4.2.2.

AThousandShips:
Yes, you need to update the documentation in the main repo, see the top of the file you edited, and you should fix the 4.3 file first, then that can be cherry picked for 4.2:
> .. DO NOT EDIT THIS FILE!!!
> .. Generated automatically from Godot engine sources.

scotmcp:
I see the difference now, previously I was contributed the Manual, not the class reference....Got it, thanks.

PR #9560, opened 2024-06-29
Update class_editorplugin.rst

### LINKED ISSUE #9555: "icon" parameter in EditorPlugin.add_custom_type is not optional, but documentation says it is
**Your Godot version**

4.2.1.stable.mono

**Issue description**

The up-to-date-for-4.2 documentation says that the `icon` parameter in `EditorPlugin.add_custom_type()` is optional.

> Adds a custom type, which will appear in the list of nodes or resources. An icon can be optionally passed.

However, this is not an optional parameter. It's clear that it doesn't have a default value. An editor error is generated when this function is called without the `icon` parameter.

**URL to the documentation page**
[EditorPlugin > add_custom_type](https://docs.godotengine.org/en/4.2/classes/class_editorplugin.html#class-editorplugin-method-add-custom-type)

ashish0kumar:
fixes #9555

fixed an error: "icon" parameter in EditorPlugin.add_custom_type is not optional, but documentation says it is.
Removed the line "An icon can be optionally passed." and added this "An editor error is generated when this function is called without the icon parameter."

### DIFF
--- a/classes/class_editorplugin.rst
+++ b/classes/class_editorplugin.rst
@@ -185,7 +185,7 @@ Methods
 .. rst-class:: classref-descriptions-group
 
 Signals
--------
+-------     
 
 .. _class_EditorPlugin_signal_main_screen_changed:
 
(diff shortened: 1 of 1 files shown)


AThousandShips:

PR #9568, opened 2024-07-01
Update Cross-compiling for iOS on Linux

Calinou:
- Update link to Xcode download. Mention that it requires being logged into an Apple ID for the page to be visible.
- Remove DMG method as Xcode has been distributed in XIP archives since 2016.
- Update page formatting to make copying commands easier and use variables when possible.

For Windows users, it should be possible to follow these steps on WSL but I haven't tested. It might be worth adding a note on this if someone can get it to work there.

### DIFF
--- a/contributing/development/compiling/cross-compiling_for_ios_on_linux.rst
+++ b/contributing/development/compiling/cross-compiling_for_ios_on_linux.rst
@@ -25,94 +25,60 @@ Requirements
 ------------
 
 - `XCode with the iOS SDK <https://developer.apple.com/download/all/?q=Xcode>`__
-  (a dmg image, for newer versions a **xip** file is going to be downloaded.
-  You must be logged into an Apple ID to download Xcode.)
+  (you must be logged into an Apple ID to download Xcode).
 - `Clang >= 3.5 <https://clang.llvm.org>`__ for your development
   machine installed and in the ``PATH``. It has to be version >= 3.5
   to target ``arm64`` architecture.
-- `Fuse <https://github.com/libfuse/libfuse>`__ for mounting and unmounting
-  the dmg image.
-- `darling-dmg <https://github.com/darlinghq/darling-dmg>`__, which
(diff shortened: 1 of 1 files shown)


contributing/development/compiling/cross-compiling_for_ios_on_linux.rst
```diff
@@ -150,11 +114,11 @@ environment variable defined to anything.
...
-    $ export OSXCROSS_IOS=anything
+    export OSXCROSS_IOS="anything"
 
 Now you can compile for iPhone using SCons like the standard Godot
 way, with some additional arguments to provide the correct paths:
 
 ::
 
-    $ scons -j 4 platform=ios arch=arm64 target=template_release IOS_SDK_PATH="/path/to/iPhoneSDK" IOS_TOOLCHAIN_PATH="/path/to/iostoolchain" ios_triple="arm-apple-darwin11-"
+    scons platform=ios arch=arm64 target=template_release IOS_SDK_PATH="/path/to/iPhoneSDK" IOS_TOOLCHAIN_PATH="/path/to/iostoolchain" ios_triple="arm-apple-darwin11-"
```
Calinou, line 124: Is the iOS triple still up-to-date? Also, now that iOS is 64-bit only, could SCons default to a sensible value automatically so the user doesn't have to define it?

bruvzg: It's the same value as used for official builds:

https://github.com/godotengine/godot-build-scripts/blob/d36a865ff1ced6e856628c3e2c82e01993bc56dc/build-ios/build.sh#L28-L31

It's only for osxcross, for building on macOS it should be empty, I guess it can be auto set based on arch when building with osxcross.

mhilbrunner (approved):

mhilbrunner:
Merged. Thanks!

mhilbrunner:

Issue #9462, opened 2024-06-06
Add a sample of GDExtention's `dependencies` syntax 


greenfox1505:
**Your Godot version: **
4.2.2

**Issue description:**
`dependencies` is mentioned in the docs here, but there is no example here of what that syntax looks like.

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/4.0/tutorials/scripting/gdextension/gdextension_cpp_example.html#using-the-gdextension-module

dsnopek:
PR https://github.com/godotengine/godot-docs/pull/9383 is attempting to add more documentation about .gdextension files, but it doesn't cover `dependencies` either

PR #9383, opened 2024-05-15
Add documentation about .gdextension file

paddy-exe:
This is a PR in the effort to improve the GDExtension docs for users

Adds details how to configure the `.gdextension` file and aids as a quick reference guide to add support for platforms

PR #9569, opened 2024-07-01
Add dependency section to gdextension file docs

paddy-exe:
Continuation of https://github.com/godotengine/godot-docs/pull/9383
Fixes https://github.com/godotengine/godot-docs/issues/9462

also removes a double word

### DIFF
--- a/tutorials/scripting/gdextension/gdextension_file.rst
+++ b/tutorials/scripting/gdextension/gdextension_file.rst
@@ -129,7 +129,7 @@ Icons section
 -------------
 
 By default, Godot uses the Node icon in the scene dock for GDExtension nodes. A custom icon can be
-set set by reference to its name and resource path of an SVG file.
+set by reference to its name and resource path of an SVG file.
 
 For example:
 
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):

mhilbrunner:
Merged! Feedback by @dsnopek still appreciated, saw the review request too late, oops.

paddy-exe:
@mhilbrunner No worries. I can patch it up very quickly if there is any feedback 👍🏻

dsnopek:
This looks good to me!

The only note that I would add, is that the most common way to use `[dependencies]` is with a path of `""`, which means to copy the dependency into the same folder as the Godot executable, and so it'd be good to have an example of that, ie:

```
windows.debug = {
    "res://bin/libdependency.windows.template_debug.x86_64.dll" : "",
    "res://bin/libdependency.windows.template_debug.x86_32.dll" : ""
}
```

On MacOS, I think using the path `"Contents/Resources"` or `"Content/Frameworks/"` is common, but I'm not 100% sure about those.

In any case, I think we should try to include some paths that are actually used in the wild, in order to help folks figure out how to use this.

YuriSizov:
Hey, @paddy-exe and @dsnopek! Am I missing something about these dependencies? When I looked into Godot codebase, I only see `OS_Android` using this data. `GDExtension::open_library` takes the info and extracts paths from it, puts it into `OS::GDExtensionData::library_dependencies`, and then `library_dependencies` is only accessed in `OS_Android::open_dynamic_library`.

If this is only implemented on Android, then this must be reflected in the documentation in two ways: an explicit note and all examples for other platforms must be removed. Or am I missing some indirect use?

PS. Well, export plugins system also extracts this information for its own use via `GDExtension::find_extension_dependencies`, but that doesn't seem to be related to loading GDExtension libraries.

paddy-exe:
@YuriSizov I haven't looked at the implementation for specific platforms but I assure you I have tested it both on Linux and even more extensively on Mac to make sure it works. I am not aware if there are more specific requirements for loading on mobile platforms though.

mhilbrunner:

PR #9572, opened 2024-07-02
Add more precise information about library paths

paddy-exe:
Continuation of https://github.com/godotengine/godot-docs/pull/9569

I looked at various application on my Mac and found that all are loaded into some folder named with ``Frameworks`` inside the ``Contens`` directory so the structure would be ``Game.app/Contents/Frameworks/libexample.dylib``

Here XCode as example:
image

I added the explanation that when no path is supplied, the libraries will be moved to the same directory as the game executable. 
For Windows at least the libraries are usually located in the same directory but not sure about Linux.

### DIFF
--- a/tutorials/scripting/gdextension/gdextension_file.rst
+++ b/tutorials/scripting/gdextension/gdextension_file.rst
@@ -148,34 +148,38 @@ Dependencies section
 In this section you set the paths of the GDExtension dependencies. This is used internally to export the dependencies
 when exporting your game executable. You are able to set which dependency is loaded depending on the feature flags
 of the exported executable. In addition, you are able to set an optional subdirectory to move your dependencies into.
-In the following example the libraries are moved into a subdirectory called ``libraries``.
+If no path is supplied Godot will move the libraries into the same directory as your game executable.
+
+.. warning::
+    In MacOS it is necessary to have shared libraries inside a folder called ``Frameworks`` with a directory structure
+    like this: ``Game.app/Contents/Frameworks``.
 
 .. code-block:: none
 
     [dependencies]
 
     macos.debug = {
-        "res://bin/libdependency.macos.template_debug.framework" : "libraries"
+        "res://bin/libdependency.macos.template_debug.framework" : "Contents/Frameworks"
     }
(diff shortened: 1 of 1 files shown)


tutorials/scripting/gdextension/gdextension_file.rst
```diff
@@ -148,34 +148,39 @@ Dependencies section
 In this section you set the paths of the GDExtension dependencies. This is used internally to export the dependencies
 when exporting your game executable. You are able to set which dependency is loaded depending on the feature flags
 of the exported executable. In addition, you are able to set an optional subdirectory to move your dependencies into.
-In the following example the libraries are moved into a subdirectory called ``libraries``.
+If no path is supplied Godot will move the libraries into the same directory as your game executable.
+
+.. note::
+    In MacOS it is common to have shared libraries inside a folder called ``Frameworks`` with a directory structure
+    like this: ``Game.app/Contents/Frameworks``. As long as the libraries are located inside the ``Contents`` directory
+    Godot will be able to load them into your game executable.
```
dsnopek, line 156: Are you sure that the libraries can be anywhere under `Contents`? Or, do they need to be in exactly `Contents/Frameworks`? I'm not familiar enough with MacOS.

dsnopek (approved):
Thanks!

I'm not entirely sure about the MacOS information, but otherwise, this looks great to me :-)

tutorials/scripting/gdextension/gdextension_file.rst
paddy-exe: They can be in any folder inside the `Contents` folder. They can even be just directly inside and are picked up correctly. In XCode for instance you have differently named folders not just `Frameworks` but also `Library` where you can also find `.framework` folders and their respective libraries.

paddy-exe:
Do not merge yet

paddy-exe:
Updated with correct information that the libraries **must** be inside the ``Frameworks`` folder. Thanks @dsnopek for the feedback!

mhilbrunner (approved):

mhilbrunner:
Thank you!

PR #9587, opened 2024-07-09
Add 2D Parallax documentation page

markdibarry:
This adds a page to clarify using `Parallax2D`, provides visual examples, and explains some common pitfalls.


tutorials/2d/2d_parallax.rst
```diff
@@ -0,0 +1,227 @@
...
+The infinite repeat effect is easiest to work with when you have an image designed to repeat seamlessly and is the same
+size or larger than your viewport **before** setting the :ref:`repeat_size<class_parallax2d_property_repeat_size>`. If
+you aren't able to obtain assets that are designed for this task, there are some other things you can do to better
+prepare your image in regards to size.
+
+Here is an example of a texture that is too small for its viewport:
+
+.. image:: img/2d_parallax_size_bad.webp
+
+We can see that the viewport size is ``500 x 300`` but the texture is ``288 x 208``. If we set the
```
mhilbrunner, line 68:
```suggestion
We can see that the viewport size is 500×300 but the texture is 288×208. If we set the
```
Either no formatting, or alternatively just for the numbers. I think this is fine? Also unsure whether we should add "pixel", but this is probably clear enough as is.

tutorials/2d/2d_parallax.rst
```diff
@@ -0,0 +1,227 @@
...
+half of the image is cut off before it jumps forward like in the image below:
+
+.. image:: img/2d_parallax_repeat_bad.webp
+
+Scroll offset
+-------------
+
+If your parallax textures are already working correctly, but you prefer it to start at a different point,
+:ref:`Parallax2D<class_parallax2d>` comes with a :ref:`scroll_offset<class_parallax2d_property_scroll_offset>` property
+used to offset where the infinite repeat canvas starts. As an example, if your image is ``288 x 208``, setting
```
mhilbrunner, line 160:
```suggestion
used to offset where the infinite repeat canvas starts. As an example, if your image is 288×208, setting
```

tutorials/2d/2d_parallax.rst
```diff
@@ -0,0 +1,227 @@
...
+the :ref:`scroll_offset<class_parallax2d_property_scroll_offset>` to ``(-144,0)`` or ``(144,0)`` allows it to begin
+halfway across the image.
+
+Repeat times
+------------
+
+Ideally, following this guide, your parallax textures are large enough to cover the screen even when zoomed out.
+Until now, we have had a perfectly fitting ``288 x 208`` texture inside of a ``288 x 208`` viewport. However, problems
```
mhilbrunner, line 168:
```suggestion
Until now, we have had a perfectly fitting 288×208 texture inside of a 288×208 viewport. However, problems 
```

tutorials/2d/2d_parallax.rst
```diff
@@ -0,0 +1,227 @@
+.. doc_2d_parallax:
+
+2D Parallax
+===========
+
+Introduction
+------------
+
+Parallax is an effect used to simulate depth by having textures move at different speeds relative to the camera. It's
+also something that even experienced users struggle to wrap their heads around. Godot provides the
```
mhilbrunner, line 10:
```suggestion
Parallax is an effect used to simulate depth by having textures move at different speeds relative to the camera. Godot provides the
```
IMO that does not add much :)

mhilbrunner (commented):
Great work, thank you! A few thoughts below, I gave it a read and found no issues otherwise.

markdibarry:
Thanks! I'm not as familiar with rst file syntax, so if you have any suggestions for formatting, feel free!

There's a lot of users I've talked to who get stuck on parallax after following tutorials that also misunderstand how the effect works. I've been compiling some FAQs regarding parallax from the community and wanted to at least get some thoughts down and a basis to build off of. This should at least serve as a reference for any new tutorials that come out after 4.3. There's a lot to improve on this, so if you see anything tone-related, grammar, or something to expand/condense, please let me know!

tutorials/2d/2d_parallax.rst
```diff
@@ -0,0 +1,226 @@
...
+
+2D Parallax
+===========
+
+Introduction
+------------
+
+Parallax is an effect used to simulate depth by having textures move at different speeds relative to the camera. Godot
+provides the :ref:`Parallax2D<class_parallax2d>` node to make this easier for users. It can still be easy to get tripped
+up though, so this page provides in-depth descriptions of some properties and how to fix some common mistakes.
```
skyace65, line 11:
```suggestion
provides the :ref:`Parallax2D<class_parallax2d>` node to achieve this effect. This node is still experimental, so the
implementation might change in future versions of Godot. However it is still recommended to use this over the old method, using the
:ref:`ParallaxLayer<class_parallaxlayer>` and :ref:`ParallaxBackground<class_parallaxbackground>` nodes. As such this page
will only cover how to use :ref:`Parallax2D<class_parallax2d>`.
```

skyace65 (changes requested):
Content is good, only one thing really needs to be changed. We need to be clear that the node is experimental, but that we still recommend using it over the old method.

markdibarry:
Great catch! I added a note.

skyace65 (approved):

mhilbrunner (approved):
Built this locally and had a quick look, spotted nothing else. Great work!

PR #9600, opened 2024-07-13
Adding note about alpha channels to Image class method blit_rect

davidrautert:
Adding a note about the alpha channel behavior when using Image.blit_rect() and guiding the reader to blend_rect() for alpha-blending.


AThousandShips:

PR #9604, opened 2024-07-13
Docs - Fixed input action C# issue

SrdanJokic:
The `SetAction` and `SetPressed` are internal methods and can't be invoked directly outside of that assembly. 

InputEventAction.cs methods

Instead, the `Action` and `Pressed` property setters can be used to invoke them instead.

InputEventAction.cs setters

### DIFF
--- a/tutorials/inputs/inputevent.rst
+++ b/tutorials/inputs/inputevent.rst
@@ -234,8 +234,8 @@ The Input singleton has a method for this:
 
     var ev = new InputEventAction();
     // Set as ui_left, pressed.
-    ev.SetAction("ui_left");
-    ev.SetPressed(true);
+    ev.Action = "ui_left";
+    ev.Pressed = true;
     // Feedback.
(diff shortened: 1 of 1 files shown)


raulsntos:
This change is correct and as far as I can see it's still not fixed in `master`. Not sure why you closed the PR, but if you need help feel free to reach out in the [development chat](https://chat.godotengine.org).

SrdanJokic:
> This change is correct and as far as I can see it's still not fixed in `master`. Not sure why you closed the PR, but if you need help feel free to reach out in the [development chat](https://chat.godotengine.org).

Hi, sorry for closing this PR without commenting as to why. The issue was that the PR branch was started from 4.2 instead of master, causing the only valid base branch to be 4.2 instead of master; which is not aligned with the contribution guidelines.

I've since created the same PR pointing to the master branch here: https://github.com/godotengine/godot-docs/pull/9605

PR #9615, opened 2024-07-16
Fix wrong information on NavigationLayers page

smix8:
Removes the parts that state that the navigation layers have anything to do with how a navigation map merges the navigation regions or navigation mesh polygons.

Also corrects wrong statement about changing the navigation layers of a region as that change requires a server sync and is not applied immediately.

This can be backported to all Godot 4.0+ version that are still updated as it applies to all versions.

This wrong documentation was done years ago build on a lot of hearsay from developers how they thought the system worked at that time. After learning how it actually functions at least I forgot to correct the page 3 times by now - so lucky number. :sweat: :unamused:

### DIFF
--- a/tutorials/navigation/navigation_using_navigationlayers.rst
+++ b/tutorials/navigation/navigation_using_navigationlayers.rst
@@ -3,14 +3,12 @@
 Using NavigationLayers
 ======================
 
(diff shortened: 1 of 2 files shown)


skyace65 (approved):

mhilbrunner (approved):

mhilbrunner:

mhilbrunner:

PR #9685, opened 2024-07-31
Upgrade to 4.3: Add note about Multiplayer protocol compat breakage

akien-mga:
Closes https://github.com/godotengine/godot/issues/94769.

### DIFF
--- a/tutorials/migrating/upgrading_to_godot_4.3.rst
+++ b/tutorials/migrating/upgrading_to_godot_4.3.rst
@@ -237,7 +237,7 @@ Core
 .. note::
 
     ``PackedByteArray`` is now able to use a more compact base64 encoding for storage. But the trade-off is that it breaks
-    compatibility, meaning that older versions of Godot may not be able to open resources saved by 4.3. (`GH-89186`_).
+    compatibility, meaning that older versions of Godot may not be able to open resources saved by 4.3 (`GH-89186`_).
 
     To maximize compatibility, this new storage format will only be enabled for resources and scenes that contain large
     PackedByteArrays for now. Support for this new format will also be added in patch updates for older versions of Godot.
@@ -281,7 +281,7 @@ GUI nodes
 
 .. note::
 
-    The default font outline color was changed from white to black. (`GH-54641`_).
+    The default font outline color was changed from white to black (`GH-54641`_).
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


skyace65 (approved):

tutorials/migrating/upgrading_to_godot_4.3.rst
```diff
@@ -290,6 +290,16 @@ GUI nodes
...
 
+Multiplayer
+^^^^^^^^^^^
+
+.. note::
+
+    The ``SceneMultiplayer`` caching protocol was changed to send the received ID instead of the Node path when sending a node removal confirmation packet (`GH-90027`_).
+
+    This is a breaking change for the high-level multiplayer protocol making it incompatible with previous Godot versions.
+    Upgrade both your server and client versions to Godot 4.3 to handle this change gracefully.
```
mhilbrunner, line 301:
```suggestion
    Upgrade both your server and client versions to Godot 4.3 to handle this change gracefully.
    
    Note that high-level multiplayer facilities are only ever meant to be compatible with server and client using the same Godot version. It is recommended to implement some kind of version checking.
```

Wording can be changed, and having this explicit note on this page is good, but we should make sure this does not imply this being a special case. We're just nice enough to give a headsup here. We *always* only guaruantee it working between the same Godot versions, anything else working is accidental and should not be relied on (and is *bad* to rely on, because some breakage or corruption may not be immediately obvious, but still happening).

Calinou: There might be a case for having this version checking built-in, at least across minor Godot releases. (Patch releases are likely compatible between each other, even if it's not 100% guaranteed.)

mhilbrunner: It should both check Godot version (yes, even minor version) and the game/project version.

Realistically, any complex multiplayer game using the high-level networking facilities will likely update more often than Godot, and even those updates are unlikely to be compatible. So outside small demos, the game version is really what you want to check for.

Maybe worth opening a proposal / discussing. Currently, AFAIK the networking only provides the hooks for the initial handshake and auth, assuming that needs are going to be too different to implement a one-size-fits-all solution, so its left as an excercise for the user. We could discuss if there is some reasonable defaults we can offer while more complicated games still are expected to provide their own.

Calinou: I've opened a proposal for this: https://github.com/godotengine/godot-proposals/issues/10348

mhilbrunner (approved):

mhilbrunner:
Thanks ❤️

PR #9686, opened 2024-07-31
Document Windows Defender causing slow startup in large projects

Jordyfel:
https://github.com/godotengine/godot/issues/47053#issuecomment-2143579317


mhilbrunner (approved):

mhilbrunner:
We already have a similar note as a warning on the [Compiling for Windows](https://docs.godotengine.org/en/latest/contributing/development/compiling/compiling_for_windows.html) page. Repeating it here makes sense.

PR #9695, opened 2024-08-01
Update 3d_gizmos.rst

The-Cyber-Captain:
Addresses "Unnamed Gizmo" error when following example code; now overriding _get_gizmo_name function, not get_name. Possible from a typo or legacy-era docs?


Calinou (approved):
Thanks!

PR #9709, opened 2024-08-04
Mention that the NavigationMesh border_size only works along xz-axis

smix8:
Adds a warning that the NavigationMesh `border_size` property only works along xz-axis as a limitation of the Recast third-party baking library.

This is because the border_size offset was developed for tile baking with the Recast tile mesh. This tile mesh never supported  verticality, you never can or could use it to bake level "voxel" chunks.


mhilbrunner (approved):

mhilbrunner:
Thanks! Merged.

PR #9710, opened 2024-08-05
Add note about BackBufferCopy and Control nodes to Screen-reading shaders page

perroboc:
Added note on Control derived nodes, found in the class documentation.

### DIFF
--- a/tutorials/shaders/screen-reading_shaders.rst
+++ b/tutorials/shaders/screen-reading_shaders.rst
@@ -115,6 +115,14 @@ With correct back-buffer copying, the two spheres blend correctly:
     If you plan to instance a scene that uses a material with ``hint_screen_texture``,
     you will need to use a BackBufferCopy node.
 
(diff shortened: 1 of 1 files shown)


tetrapod00 (approved):
Mild approval. The information added is copied verbatim from the the [BackBufferCopy class description](https://docs.godotengine.org/en/stable/classes/class_backbuffercopy.html#description), so is correct as far as it goes.

I was around in the help chat which prompted the author to make this PR, so adding this information to this page does solve a tangible problem. But I'm also slightly wary of stuffing more information into this already long and slightly disorganized page.

skyace65:
Given tetrapod's reservations I'm going to close this PR. In addition, it's not hidden that this node inherits from Node2D and not Control. Not behaving like a control node should be the expected behavior.

Perroboc thank you for your work on this regardless, and please consider contributing again in the future if you find something you want changed.

PR #9720, opened 2024-08-08
 Mention SceneMultiplayer in high_level_multiplayer

### LINKED ISSUE #9717: The default MultiplayerAPI is actually a SceneMultiplayer
**Your Godot version:**
v4.2.2.stable.official [15073afe3]

**Issue description:**
In the "High-level multiplayer" overview (Arguably the first link someone stumbles upon when googling "Godot multiplayer") it is said that by default, the multiplayer node already has a `MultiplayerAPI` configured. However, it's actually a [`SceneMultiplayer`](https://docs.godotengine.org/en/stable/classes/class_scenemultiplayer.html), that *inherits* MultiplayerAPI.
This is actually really important however, because you need to set `allow_object_decoding` in the `SceneMultiplayer` to true to send objects over the network, something which probably every Godot multiplayer game does. Each time I made a new multiplayer project I always stumbled over [this issue](https://github.com/godotengine/godot/issues/82718) and only then remembered that.
Considering how essential that option is, maybe it should also be mentioned in the "high-level multiplayer" overview.

https://docs.godotengine.org/en/stable/tutorials/networking/high_level_multiplayer.html#initializing-the-network

IfgiU:
This fixes #9717 by mentioning in the "High Level Multiplayer" Overview that the default implementation is a SceneMultiplayer. It also mentions that for object serializing allow_object_decoding is required.

### DIFF
--- a/tutorials/networking/high_level_multiplayer.rst
+++ b/tutorials/networking/high_level_multiplayer.rst
@@ -100,7 +100,7 @@ Initializing the network
 High level networking in Godot is managed by the :ref:`SceneTree <class_SceneTree>`.
 
 Each node has a ``multiplayer`` property, which is a reference to the ``MultiplayerAPI`` instance configured for it
-by the scene tree. Initially, every node is configured with the same default ``MultiplayerAPI`` object.
+by the scene tree. Initially, every node is configured with the same default ``SceneMultiplayer`` object, which inherits ``MultiplayerAPI``.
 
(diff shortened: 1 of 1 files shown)


Jordyfel:
The change is technically incorrect, it could be a `SceneMultiplayer` (as it is by default), or it could be something else, that's why its a base class. The fact that it is the default implementation is mentioned clearly enough in `MultiplayerAPI` and `SceneMultiplayer` class reference.

That said, I agree that it would be good to have it mentioned here as well, but it should be worded better.

PR #9722, opened 2024-08-09
Fix return type of `EditorImportPlugin._Import`

esainane:
The following is the currently generated `EditorImportPlugin.cs`:

```csharp
    public virtual Error _Import(string sourceFile, string savePath, Dictionary options, Array<string> platformVariants, Array<string> genFiles)
    {
        return Error.Ok;
    }
```

This fixes the type signature in the documentation's example to match the actual type signature.


AThousandShips:

esainane:
My apologies, and thank you for the pointer!

PR #9731, opened 2024-08-10
Add 2D versions of NavigationAgent script templates

smix8:
Adds 2D versions of the NavigationAgent script templates so they are easier to just copy & paste for 2D users, and condenses all the templates in a tab menu for easier filtering.

Note this uses the newer `get_iteration_id()` function to avoid empty paths or error msgs from querying unsynced navigation maps. This function is only available in Godot 4.3. If those lines are removed the scripts work in 4.2 as well.

agent_scripts


mhilbrunner (approved):

skyace65 (approved):

skyace65:
Thanks!

PR #9734, opened 2024-08-11
Add description of TIME to shaders, add label to global uniform header

tetrapod00:
Added/changed description of TIME in CanvasItem, Fog, Particle, Sky, and Spatial shader pages. Description now links to other relevant docs pages. Added a label to the global uniform header of shading_languge.rst, so it can be cross referenced from elsewhere.

I made some changes in addition to simply copying the existing description from CanvasItem shaders to the other pages.

I wrote for a user is new-ish to shaders, notices that ``TIME`` does not exactly match CPU time behavior, and finds any of the shader pages. Ideally they can come away the exact behavior of the built-in ``TIME`` uniform, as well as a way to make their own if need be. I based this on my interactions with users in various discords who asked about ``TIME``.

New description which appears on each shader page:
> Global time since the engine has started, in seconds. It repeats after every 3,600 seconds (which can  be changed with the ``rollover`` setting). It's not affected by ``time_scale`` or pausing. If you need  a ``TIME`` variable that can be scaled or paused, add your own ``global shader uniform`` and update it each frame.

Old description, from the CanvasItem shader page: 
> Global time since the engine has started, in seconds (always positive). It's subject to the rollover setting (which is 3,600 seconds by default). It's not affected by ``time_scale`` or pausing, but you can define a global shader uniform to add a "scaled" ``TIME`` variable if desired.

Changes:
- Now links to the [rollover setting](https://docs.godotengine.org/en/latest/classes/class_projectsettings.html#class-projectsettings-property-rendering-limits-time-time-rollover-secs) in project settings.
- Now links to the [global shader uniform](https://docs.godotengine.org/en/latest/tutorials/shaders/shader_reference/shading_language.html#global-uniforms) documentation. A label is added to the global uniform header in that docs page. I may have gotten the preferred syntax wrong for this.
- Uses more active language for the sentence about adding your own ``TIME`` uniform.
- Clarifies that the value *repeats* rather than merely "being subject to" the rollover setting.
- Emphasizes the default value of 3,600 seconds first, because a user finding this page is unlikely to have changed that value.
- No longer mentions that the value is "always positive". If this is necessary, I would suggest "always in the range of 0 to 3,600" or "always in the range of 0 to ``rollover``".

If any of these changes break docs style, or are otherwise too much, I can revert to more minimal changes.

### DIFF
--- a/tutorials/shaders/shader_reference/canvas_item_shader.rst
+++ b/tutorials/shaders/shader_reference/canvas_item_shader.rst
@@ -50,24 +50,26 @@ Global built-ins
 
 Global built-ins are available everywhere, including custom functions.
 
-+-------------------+----------------------------------------------------------------------------------------+
-| Built-in          | Description                                                                            |
-+===================+========================================================================================+
-| in float **TIME** | Global time since the engine has started, in seconds (always positive).                |
-|                   | It's subject to the rollover setting (which is 3,600 seconds by default).              |
-|                   | It's not affected by :ref:`time_scale<class_Engine_property_time_scale>`               |
-|                   | or pausing, but you can define a global shader uniform to add a "scaled"               |
-|                   | ``TIME`` variable if desired.                                                          |
-+-------------------+----------------------------------------------------------------------------------------+
-| in float **PI**   | A ``PI`` constant (``3.141592``).                                                      |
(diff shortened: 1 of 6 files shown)


clayjohn (approved):
Looks great to me! Good work

skyace65 (approved):

PR #9736, opened 2024-08-11
Add tip and warning about TileMap build-in navigation

smix8:
- Adds a tip to TileMap navigation section that informs users to prefer baking the TileMap navigation.
- Adds a warning that 2D navigation meshes can not overlap.

I don't think this big warning block is too much considering that basically every single TileMap user gets this wrong and the TileMap and its editor does a very poor job informing users about all the TileMap specific navigation related constrains and issues.

New users spend a lot of time trying to make the ill-designed TileMapLayer feature work until they are deeply frustated with it (you can't "layer" navigation meshes). Only then they start to ask or read the navigation documentation. I hope by placing it more prominently inside the TileMap doc that this information reaches more users before that frustation happens.

tilemap_nav_warning

### DIFF
--- a/tutorials/2d/using_tilemaps.rst
+++ b/tutorials/2d/using_tilemaps.rst
@@ -81,6 +81,15 @@ Navigation
 - **Navigation Visible** Whether or not the TileMapLayer's navigation meshes are
   visible. If set to default then it depends on the show navigation debug settings.
 
+.. tip::
+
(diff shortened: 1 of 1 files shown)


skyace65 (approved):

mhilbrunner (approved):

mhilbrunner:
Merged. Thanks for caring so much about the documentation for your work!

PR #9740, opened 2024-08-12
bone_map property not available to post import scripts

scotmcp:
You cannot load a bone_map with import scripts. :ref:`class_bonemap` is not accessible outside of the UI. In order to map the bones with an import script, you must load the bone map file into an array and assign the bones one at a time.

### DIFF
--- a/tutorials/assets_pipeline/retargeting_3d_skeletons.rst
+++ b/tutorials/assets_pipeline/retargeting_3d_skeletons.rst
@@ -47,6 +47,9 @@ With the Skeleton node selected, first set up a new :ref:`class_bonemap` and :re
 Godot has a preset called :ref:`class_skeletonprofilehumanoid` for humanoid models.
 This tutorial proceeds with the assumption that you are using :ref:`class_skeletonprofilehumanoid`.
 
+..note:: You cannot load a bone_map with import scripts. :ref:`class_bonemap` is not accessible outside of the UI. In order to
+         map the bones with an import script, you must load the bone map file into an array and assign the bones one at a time.
+
 .. note:: If you need a profile that is different from :ref:`class_skeletonprofilehumanoid`, you can export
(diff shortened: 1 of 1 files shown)


scotmcp:
The reason for adding this note is a LOT of time was wasted trying to find a way to do this, and it cannot be done. This should be noted at the ONE AND ONLY reference to the bone_map property in the documentation. I added the means to get the job accomplished, albiet it is an arduous means to and end.

tutorials/assets_pipeline/retargeting_3d_skeletons.rst
```diff
@@ -47,6 +47,9 @@ With the Skeleton node selected, first set up a new :ref:`class_bonemap` and :re
 Godot has a preset called :ref:`class_skeletonprofilehumanoid` for humanoid models.
 This tutorial proceeds with the assumption that you are using :ref:`class_skeletonprofilehumanoid`.
 
+..note:: You cannot load a bone_map with import scripts. :ref:`class_bonemap` is not accessible outside of the UI. In order to
+         map the bones with an import script, you must load the bone map file into an array and assign the bones one at a time.
```
AThousandShips, line 51:
```suggestion
.. note:: You cannot load a bone_map with import scripts. :ref:`class_bonemap` is not accessible outside of the UI. In order to
          map the bones with an import script, you must load the bone map file into an array and assign the bones one at a time.
```

lyuma:
Sorry but I simply don't understand.

is there a MRP or issue that shows what doesn't work, or can you share an example of a script that should have worked in your mind, but doesn't 

what does "outside of the UI" mean?

also, what does class_bonemap mean? The BoneMap class should be accessible to scripts.

scotmcp:
@tool # Needed so it runs in editor.
extends EditorScenePostImport
func _post_import(scene):
	var bone_map = load("res://explosive_bone_map.tres")
	skeleton = scene.get_node("Armature/Skeleton3D")
	var bonemap : BoneMap = skeleton.bone_map

PR #9741, opened 2024-08-12
Add pre-cond command to install Godot using Scoop on Windows

HotariTobu:
Added `scoop bucket add extras` in the editor command line tutorial because scoop cannot find Godot without the command before.


skyace65:

PR #9777, opened 2024-08-17
Use the newly introduced ``is not`` operator

0stam:
Use the newly introduced ``is not`` operator instead of ``not (...) is``.

Running the following regular expression: ``(^|\`\`)\s*(if|while|.*=)(.)* not (|.* )is `` I have found only one occurrence of the outdated code in the entire documentation.

### DIFF
--- a/tutorials/scripting/gdscript/static_typing.rst
+++ b/tutorials/scripting/gdscript/static_typing.rst
@@ -269,7 +269,12 @@ get full autocompletion on the player variable thanks to that cast.
 
         player.damage()
 
-    or ``assert()`` statement::
+    You can also simplify the code by using the ``is not`` operator::
+
+        if body is not PlayerController:
+            push_error("Bug: body is not PlayerController")
+
+    Alternatively, you can use the ``assert()`` statement::
 
         assert(body is PlayerController, "Bug: body is not PlayerController.")
 
@@ -278,6 +283,7 @@ get full autocompletion on the player variable thanks to that cast.
             return
 
         player.damage()
+
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


Mickeon:
I wonder if there's any `not (...) is` in the class reference, too...

0stam:
> I wonder if there's any `not (...) is` in the class reference, too...

After all the searching, I have read the documentation guidelines. It turns out that, to make everything more concise, you should avoid using typed code. So it makes sense that the operator is not used outside of the typing tutorial, and I don't expect to find it in the class reference.

At first I wanted to replace the operator in the expression, but then realised that this page also introduces the `is` operator itself, and introducing both at the same time could be misleading to someone.

tutorials/scripting/gdscript/static_typing.rst
```diff
@@ -279,6 +279,11 @@ get full autocompletion on the player variable thanks to that cast.
 
         player.damage()
 
+You can also simplify the expression using the ``is not`` operator::
+
+    if body is not PlayerController:
+        push_error("Bug: body is not PlayerController")
```
tetrapod00, line 285:
```suggestion
    You can also simplify the expression using the ``is not`` operator::
    
        if body is not PlayerController:
            push_error("Bug: body is not PlayerController")
```
This needs to be indented one level deeper, so it appears in the note block, since it is conceptually related to the earlier example.

You should also relocate this new `is not` example to after the initial `is` example and before the `assert()` example. Note that you'll have to adjust the existing `or ``assert()`` statement::` line to make sense.

tetrapod00 (commented):
Another PR, https://github.com/godotengine/godot-docs/pull/10085, also tries to document `is not` in this section (though that PR only replaces the existing usage). One of these two PRs should be accepted.

0stam:
Thanks for the review, I have applied all the feedback. I have also replaced an incorrectly used "expression".

We don't have a separate documentation for the `is` keyword, so I would personally stick with explaining the basic version before introducing `is not`, but I'm no expert on the topic.

The other PR also suggest removing parentheses from `if not (body is PlayerController):`. It's technically correct, because `is` has a higher precedence than `not`, but it's not a very common knowledge (at least I had to check it) and I believe it's much easier to understand in the current version, especially if you see it for the first time.

tetrapod00 (approved):
Looks good now. This PR is a better way to document `is not` than the alternate PR https://github.com/godotengine/godot-docs/pull/10085.

The reasoning to keep the parentheses in the existing `if not (body is PlayerController):` line makes sense to me.

mhilbrunner (approved):

mhilbrunner:
Merged. Thanks and congrats on your first merged contribution! Sorry this one took a bit :)

0stam:
Thank you too. I'm glad I could help :)

PR #9781, opened 2024-08-18
Add section on working with collections in C#

fractaal:
Clarified that for passing collections, use Godot's collections (or raw arrays) instead of C#'s provided collections as these currently don't work with the marshaller

### DIFF
--- a/tutorials/scripting/cross_language_scripting.rst
+++ b/tutorials/scripting/cross_language_scripting.rst
@@ -173,25 +173,22 @@ If that's impossible, you'll see the following error: ``Invalid call. Nonexisten
 
 Receiving C# collections from GDScript
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-When passing or returning collections from C# methods, types from the `System.Collections` and `System.Collections.Generic` namespaces are not compatible with Godot's marshalling system. For example:
+When passing or returning collections from C# methods, types from the `System.Collections` and `System.Collections.Generic` namespaces are not compatible with Godot's marshalling system. For example, this method will not be visible to GDScript, and Godot will throw the "Invalid call" error:
 
 .. code-block:: csharp
 
-    // These methods will not be visible to GDScript, and Godot will throw the "Invalid call" error
     public System.Collections.Generic.List<string> GetStrings() { ... }
 
-To ensure your collections can be used in GDScript, use Godot's collection types from the `Godot.Collections` namespace. The types in your collection also must be :ref:`Variant-compatible <c_sharp_variant_compatible_types>`:
+To ensure your collections can be used in GDScript, use Godot's collection types from the `Godot.Collections` namespace. The types in your collection also must be :ref:`Variant-compatible <c_sharp_variant_compatible_types>`. This method *will* be visible to GDScript:
 
 .. code-block:: csharp
 
-    // This method can be called from GDScript
     public Godot.Collections.Array<string> GetStrings() { ... }
 
-Alternatively, you can use raw arrays, which are also compatible with the marshalling system.
+Alternatively, you can use raw arrays, which are also compatible with the marshalling system. This method will also be visible to GDScript:
 
 .. code-block:: csharp
 
-    // This method can be called from GDScript
     public string[] GetStrings() { ... }
 
 
--- a/tutorials/scripting/cross_language_scripting.rst
+++ b/tutorials/scripting/cross_language_scripting.rst
@@ -178,7 +178,7 @@ When passing or returning collections from C# methods, types from the `System.Co
 .. code-block:: csharp
 
     // These methods will not be visible to GDScript, and Godot will throw the "Invalid call" error
-    public List<string> GetStrings() { ... }
+    public System.Collections.Generic.List<string> GetStrings() { ... }
 
 To ensure your collections can be used in GDScript, use Godot's collection types from the `Godot.Collections` namespace:
 
--- a/tutorials/scripting/cross_language_scripting.rst
+++ b/tutorials/scripting/cross_language_scripting.rst
@@ -195,6 +195,8 @@ Alternatively, you can use raw arrays, which are also compatible with the marsha
     public string[] GetStrings() { ... }
 
 
+For more details on how to choose the right C# collection for your use case, see :ref:`C# collections<doc_c_sharp_collections>`.
+
 Calling GDScript methods from C#
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/tutorials/scripting/cross_language_scripting.rst
+++ b/tutorials/scripting/cross_language_scripting.rst
@@ -171,6 +171,30 @@ If that's impossible, you'll see the following error: ``Invalid call. Nonexisten
     my_csharp_node.PrintArray(["a", "b", "c"]) # a, b, c
     my_csharp_node.PrintArray([1, 2, 3]) # 1, 2, 3
 
+Receiving C# collections from GDScript
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+When passing or returning collections from C# methods, types from the `System.Collections` and `System.Collections.Generic` namespaces are not compatible with Godot's marshalling system. For example:
+
+.. code-block:: csharp
+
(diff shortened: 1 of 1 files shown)


tutorials/scripting/cross_language_scripting.rst
```diff
@@ -171,6 +171,30 @@ If that's impossible, you'll see the following error: ``Invalid call. Nonexisten
...
+    public Godot.Collections.Array<string> GetStrings() { ... }
+
+Alternatively, you can use raw arrays, which are also compatible with the marshalling system.
+
+.. code-block:: csharp
+
+    // This method can be called from GDScript
+    public string[] GetStrings() { ... }
+
+
```
tetrapod00, line 197:
```suggestion

For more details on how to choose the right C# collection for your use case, see :ref:`C# collections<doc_c_sharp_collections>`.

```

Cross-link the page that goes into more detail.

tutorials/scripting/cross_language_scripting.rst
```diff
@@ -171,6 +171,30 @@ If that's impossible, you'll see the following error: ``Invalid call. Nonexisten
...
     my_csharp_node.PrintArray([1, 2, 3]) # 1, 2, 3
 
+Receiving C# collections from GDScript
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+When passing or returning collections from C# methods, types from the `System.Collections` and `System.Collections.Generic` namespaces are not compatible with Godot's marshalling system. For example:
+
+.. code-block:: csharp
+
+    // These methods will not be visible to GDScript, and Godot will throw the "Invalid call" error
+    public List<string> GetStrings() { ... }
```
tetrapod00, line 181:
```suggestion
    public System.Collections.Generic.List<string> GetStrings() { ... }
```

I'd be explicit here, given that this article is for both GDScript and C# users.

tutorials/scripting/cross_language_scripting.rst
```diff
@@ -171,6 +171,30 @@ If that's impossible, you'll see the following error: ``Invalid call. Nonexisten
     my_csharp_node.PrintArray(["a", "b", "c"]) # a, b, c
     my_csharp_node.PrintArray([1, 2, 3]) # 1, 2, 3
 
+Receiving C# collections from GDScript
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+When passing or returning collections from C# methods, types from the `System.Collections` and `System.Collections.Generic` namespaces are not compatible with Godot's marshalling system. For example:
+
+.. code-block:: csharp
+
+    // These methods will not be visible to GDScript, and Godot will throw the "Invalid call" error
```
tetrapod00, line 180: I believe that comments should be brought outside into body text if possible, as comments don't currently get translated? See discussion [here](https://github.com/godotengine/godot-docs/issues/4188).

tutorials/scripting/cross_language_scripting.rst
```diff
@@ -171,6 +171,30 @@ If that's impossible, you'll see the following error: ``Invalid call. Nonexisten
...
+    public List<string> GetStrings() { ... }
+
+To ensure your collections can be used in GDScript, use Godot's collection types from the `Godot.Collections` namespace:
```
tetrapod00, line 183:
```suggestion
To ensure your collections can be used in GDScript, use Godot's collection types from the `Godot.Collections` namespace. The types in your collection also must be :ref:`Variant-compatible <c_sharp_variant_compatible_types>`:
```
I'd also mention the variant compatibility requirement. Feel free to reword or move it to somewhere else in the section.

tetrapod00 (changes requested):

tutorials/scripting/cross_language_scripting.rst
fractaal: Thanks for the suggestions! Made all of the changes. 

I thought this apt to put into the docs as I think i've been burned by this fact for way too long and didn't realize what exactly was happening until now.

tutorials/scripting/cross_language_scripting.rst
```diff
@@ -173,25 +173,22 @@ If that's impossible, you'll see the following error: ``Invalid call. Nonexisten
 
 Receiving C# collections from GDScript
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-When passing or returning collections from C# methods, types from the `System.Collections` and `System.Collections.Generic` namespaces are not compatible with Godot's marshalling system. For example:
+When passing or returning collections from C# methods, types from the `System.Collections` and `System.Collections.Generic` namespaces are not compatible with Godot's marshalling system. For example, this method will not be visible to GDScript, and Godot will throw the "Invalid call" error:
```
tetrapod00, line 176:
```suggestion
When passing or returning collections from C# methods, types from the `System.Collections` and `System.Collections.Generic` namespaces are not compatible with Godot's marshalling system. For example, this method will not be visible to GDScript, and Godot will throw an ``Invalid call`` error:
```

Style nitpick to match the usage of `code` for errors in the rest of the page; and use "an" to not assume the reader has read the whole page. Other than that, LGTM!

tetrapod00:
Good addition! This section definitely would have helped me when I started using Godot.

You'll probably have to squash your commits as described [here](https://docs.godotengine.org/en/latest/contributing/workflow/pr_workflow.html#the-interactive-rebase), since you used the web editor and added some commits for review. You could either do so now or wait until maintainers comment on the PR.

raulsntos (commented):
I feel like this new section can be merged with the previous one (_Calling C# methods from GDScript_). The previous section already mentions that GDScript won't find methods that contain parameter types that can't be marshalled but perhaps the explanation could be improved.

For example, we could be more explicit about what makes Godot unable to find a method with a matching signature. And link to `c_sharp_variant_compatible_types` documentation like you do in the section you added.

Then, mentioning collections specifically seem unnecessary because I feel like it's already covered in the `c_sharp_variant_compatible_types` documentation which explains what makes a type _Variant-compatible_ and even includes a table with all the compatible types (including collections).

fractaal:
Making the explanation of what types Godot can recognize more comprehensive works for me, then we can omit this new section entirely. 

I still think there is some value in explicitly clarifying in this article that .NET collections won't work with the marshaller, then link to the more detailed articles regarding variant compatible types and C# collections, because this is probably going to be one of the top most visited pages in the doc if people are going to be working with C# in Godot. Leaving a small aside that lets you know about this gotcha would save a world of hurt. I know it would've helped me! 

Not to mention: most of the wording in `cross_language_scripting.rst` is about how accessing fields and calling methods in C# from GDScript is "straightforward," so I think it's important to highlight a situation where it might not be as straightforward.

For instance, I've found at least just [one other dev](https://forum.godotengine.org/t/nonexistent-function-in-c-sharp-script/54275) experiencing the same problem (using List<int>, which, as discussed, won't be seen by GDScript):

raulsntos:
> I still think there is some value in explicitly clarifying in this article that .NET collections won't work with the marshaller, then link to the more detailed articles regarding variant compatible types and C# collections

Sure, I think it's good to add those links. But I don't think it's necessary to single out collections, since they work the same way as any other type.

> most of the wording in `cross_language_scripting.rst` is about how accessing fields and calling methods in C# from GDScript is "straightforward,"

I agree and in most of the C# documentation we try to clarify that only Variant-compatible types work. This page could be updated to be more clear about that, but I don't think we should duplicate information that is already available in other pages since they can go out of sync.

paulloz:
FWIW that page dates back to 3.x where the marshaller was more lenient with conversions, hence how it is worded. It has only been vaguely updated since.

fractaal:
Thanks for your guys' input! The collections thing might just be me, they just live in a "different headspace" so to speak - so a .NET collection with Variant compatible types could be assumed to be handled even though it isn't. Something with the type in the collection as opposed to the collection itself being Variant compatible.

We all agree on clarifying that only Variant compatible types work on this page, which *can* include the quick aside that this *includes stuff like Collections* and then lead to the more comprehensive pages talking about these topics. I don't think this is duplicating information, just makes the docs more practical to read as it addresses a common use case, and is a good pointer to the other pages.

I'm adamant to have that in as otherwise it might not be immediately apparent, and other devs might go down the same rabbit hole I did. But if it ultimately doesn't align with the writing style of the Godot Docs, I'll defer to your input. Thanks again!

skyace65:
Closing this PR per raulsntos' review

PR #9796, opened 2024-08-21
Fix comment directive to prevent webpage display

esainane:
`..` immediately following paragraph text at the same indentation is interpreted as a continuation of that text, as seen on the website:

https://docs.godotengine.org/en/stable/contributing/how_to_contribute.html#technical-contributions

> Upload your plugins to the Godot Asset Library to make them available to others. .. update to talk about Asset Store later

I feel the intent was still clear, but it does look a bit weird at the moment.

### DIFF
--- a/contributing/how_to_contribute.rst
+++ b/contributing/how_to_contribute.rst
@@ -74,8 +74,9 @@ Technical contributions
 - **Write Plugins (GDScript, C#, & more)**
   Community addons are not directly included in the core engine download or repository, yet they provide essential quality of life upgrades for your fellow game developers.
   Upload your plugins to the `Godot Asset Library <https://godotengine.org/asset-library/asset>`_ to make them available to others.
-  .. update to talk about Asset Store later
 
+  ..
+    update to talk about Asset Store later
 - **Demo projects (GDScript, C#, and making Assets)**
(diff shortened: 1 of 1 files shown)


esainane:
There must be a blank line between the text block and the comment for the comment to be properly interpreted as a comment.
 
There must not be two blank lines between the two list items, once the comment is ignored. Otherwise, the generated output produces two separate lists. This a perceptibly different amount of padding compared to when they were structurally part of the same list in HTML output, since Godot adds extra padding around `<ul>` tags.

That last point is not respected elsewhere in practice. For example, earlier on the same page, `Publish Godot Games` has a noticeably wider gap between it and the preceding `Donation Drives` item.

This is due to the following code:

```rst
  [...]
  If you do live streams, perhaps think about organizing a stream with donation incentives.

.. - **Buy Official Merch**

- **Publish Godot Games.**
```

This is probably going to drive me absolutely crazy, now that I'm conscious of it.

But, fixing this everywhere is something that will take much more work, and should be less important than comments emitted as user-visible text anyway. I'll stick to that for this PR.

mhilbrunner (approved):

mhilbrunner:
Thanks, merged.

mhilbrunner:

PR #9807, opened 2024-08-23
Updated the reference to AudioStreamPlayer2D in Tutorial

ninlar:
The tutorial and C# code samples reference the node type AudioStreamPlayer which is now AudioStreamPlayer2D.


ninlar:
The documentation is accurate, I completely missed that there is still an AudioStreamPlayer that is non-positional

PR #9809, opened 2024-08-23
Correct note in bbcode tutorial about faux fonts

aXu-AP:
https://docs.godotengine.org/en/4.3/tutorials/ui/bbcode_in_richtextlabel.html#reference (Note box) Godot doesn't generate fonts via RichTextLabel as the text suggests. Corrected text to reflect what actually is happening: use default theme italics and bolds (which are fake variations, but that info isn't relevant on this page).

Of note: if the `default_font` of the current theme is set, no font variations occur until user overrides the fonts. Should the tutorial also mention about this?


bruvzg:
> Of note: if the default_font of the current theme is set, no font variations occur until user overrides the fonts. Should the tutorial also mention about this?

If you set custom font for the default theme (in the project settings) it should generate variations, if it's fully a custom theme it won't.

skyace65:
@Calinou, you added that information in #6991 to update the page for Godot 4. Was that functionality removed? Or is this feature not working an engine bug?

Calinou:
> @Calinou, you added that information in #6991 to update the page for Godot 4. Was that functionality removed? Or is this feature not working an engine bug?

The behavior I described is the one that should be occurring, as per @bruvzg's comment. It should work even with a normal font override, unless you use a fully custom theme.

skyace65:
Closing this per Calinou's comment

PR #9821, opened 2024-08-23
Add GDExtension C++ code snippets to GettingStarted

dementive:
I have been writing a project entirely in GDExtension C++ and I noticed the Godot 3.6 documentation had C++ snippets all over but there are still not any C++ snippets for anything in the 4.0 docs.

So I'd like to start an effort to get all of these filled out because GDExtension is extremely usable at this point in my opinion. Unfortunately it is not very accessible to most due to the current lack of documentation (unless you are sadistic and actually enjoy reading C++ source code to find out how to do things like me).

I figured it would be best to start at the beginning of the docs with the first place code blocks found in ` scripting_first_script.rst`. I want to get some feedback on this and if the way I am doing it is acceptable I can go through and do a lot more of them, transcribing Gdscript to C++ is a pretty trivial and mindless task for me at this point so I can get through a lot of them pretty quickly I feel, especially with some help from other C++ fellas.

To ensure the documentation I am writing is both consistently formatted and actually correct. I have a specific setup to make sure I'm not writing nonsense code that doesn't work.

1. Setting up a basic GDExtension project to write the documentation in so I can use `clangd` to ensure all the code I'm writing and adding to the docs will actually compile without benign easy to find errors.

2. Compiling my code, testing it in the editor (if possible) and making sure it is actually doing what the documentation describes.

3. Before I add any of my code to the documentation blocks I run `clang-format -i ./File.cpp` from inside the Godot repository to ensure the formatting uses the exact same rules Godot source code would. There are some formatting things clang-format won't enforce, so I'm also comparing my style to the code style of godot-cpp's header files and trying to make sure they are always the exact same.

4. Building the docs and testing that my changes actually did what they were supposed to.

If I am doing anything wrong here or you think my process could be improved let me know! If everything here looks good let me know and I'll keep working through these to add C++ code blocks everywhere.

### DIFF
--- a/getting_started/step_by_step/scripting_first_script.rst
+++ b/getting_started/step_by_step/scripting_first_script.rst
@@ -107,6 +107,28 @@ the following line of code:
     {
     }
 
+  .. code-tab:: GDExtension C++
+
+    #ifndef MY_SPRITE_2D_H
+    #define MY_SPRITE_2D_H
+
+    #include <godot_cpp/classes/sprite2d.hpp>
+
+    // MySprite2D would need to be bound in your GDExtensions register_types.cpp file.
+    // More info on this in the GDExtension example setup tutorial.
+    class MySprite2D : public godot::Sprite2D {
+        GDCLASS(godot::MySprite2D, godot::Object);
+
+    protected:
+        static void _bind_methods() {}  // _bind_methods is required for all GDCLASS's or compilation will fail!
+
+    public:
+        // Empty default constructor is required or compilation will fail!
+        // GDExtension constructors also are not allowed to have additional parameters or compilation will fail!
+        MySprite2D() {}
+    };
+    #endif // MY_SPRITE_2D_H
+
 Every GDScript file is implicitly a class. The ``extends`` keyword defines the
 class this script inherits or extends. In this case, it's ``Sprite2D``, meaning
 our script will get access to all the properties and functions of the Sprite2D
@@ -150,6 +172,12 @@ Add the following code to your script:
         GD.Print("Hello, world!");
     }
 
+ .. code-tab:: GDExtension C++
+
+    MySprite2D() {
+        godot::UtilityFunctions::print("Hello, world!");
+    }
+
 
 Let's break it down. The ``func`` keyword defines a new function named
 ``_init``. This is a special name for our class's constructor. The engine calls
@@ -187,6 +215,11 @@ angular speed in radians per second.  Add the following after the ``extends Spri
 
     private int _speed = 400;
     private float _angularSpeed = Mathf.Pi;
+
+ .. code-tab:: GDExtension C++
+
+    int speed = 400;
+    float angular_speed = Math_PI;
 
 Member variables sit near the top of the script, after any "extends" lines,
 but before functions. Every node
@@ -229,6 +262,14 @@ At the bottom of the script, define the function:
     public override void _Process(double delta)
     {
         Rotation += _angularSpeed * (float)delta;
+    }
+
+ .. code-tab:: GDExtension C++
+
+    void _process(const double p_delta) override {
+        // Note that ``rotation`` is a private member variable of  Node2D so the ``set_rotation`` function must be used.
+        // You'll find that a lot of member variables that are public in gdscript/C# are instead private in GDExtension C++.
+        set_rotation(get_rotation() + (angular_speed * p_delta));
     }
 
 The ``func`` keyword defines a new function. After it, we have to write the
@@ -277,6 +318,14 @@ them.
     var velocity = Vector2.Up.Rotated(Rotation) * _speed;
 
     Position += velocity * (float)delta;
+
+ .. code-tab:: GDExtension C++
+
+    // Note that the directional Vector2 constants do not exist in GDExtension right now. So Vector2(0, -1) must be used.
+    Vector2 velocity = Vector2(0, -1).rotated(get_rotation()) * speed;
+
+    set_position(get_position() + (velocity * p_delta));
+
 
 As we already saw, the ``var`` keyword defines a new variable. If you put it at
 the top of the script, it defines a property of the class. Inside a function, it
@@ -343,3 +392,35 @@ Here is the complete ``sprite_2d.gd`` file for reference.
             Position += velocity * (float)delta;
         }
     }
+ 
+  .. code-tab:: GDExtension C++
+
+    #ifndef MY_SPRITE_2D_H
+    #define MY_SPRITE_2D_H
+
+    #include "godot_cpp/variant/utility_functions.hpp"
+    #include <godot_cpp/classes/sprite2d.hpp>
+    #include <godot_cpp/core/math.hpp>
+
+    class MySprite2D : public Sprite2D {
+        GDCLASS(MySprite2D, godot::Object);
+
+        int speed = 400;
+        float angular_speed = Math_PI;
+
+    protected:
+        static void _bind_methods() {}
+
+    public:
+        MySprite2D() {
+            godot::UtilityFunctions::print("Hello, world!");
+        }
+
+        void _process(const double p_delta) override {
+            set_rotation(get_rotation() + (angular_speed * delta));
+
+            Vector2 velocity = Vector2(0, -1).rotated(get_rotation()) * speed;
+            set_position(get_position() + (velocity * delta));
+        }
+    };
+    #endif // MY_SPRITE_2D_H
--- a/getting_started/step_by_step/scripting_first_script.rst
+++ b/getting_started/step_by_step/scripting_first_script.rst
@@ -30,6 +30,8 @@ The equivalent C# code has been included in another tab for convenience.
 
 .. seealso:: To learn more about C#, head to the :ref:`C# basics <doc_c_sharp>` page.
 
+.. seealso:: To learn more about GDExtension C++, head to the :ref:`What is GDExtension? <doc_what_is_gdextension>` page.
+
 Project setup
 -------------
 
@@ -107,26 +109,31 @@ the following line of code:
     {
     }
 
-  .. code-tab:: GDExtension C++
+ .. code-tab:: cpp C++
 
     #ifndef MY_SPRITE_2D_H
     #define MY_SPRITE_2D_H
 
     #include <godot_cpp/classes/sprite2d.hpp>
 
-    // MySprite2D would need to be bound in your GDExtensions register_types.cpp file.
+    namespace CustomNamespace {
+
+    // MySprite2D also needs to be bound in your GDExtension's register_types.cpp file.
     // More info on this in the GDExtension example setup tutorial.
     class MySprite2D : public godot::Sprite2D {
-        GDCLASS(godot::MySprite2D, godot::Object);
+        GDCLASS(godot::MySprite2D, godot::Object)
 
     protected:
-        static void _bind_methods() {}  // _bind_methods is required for all GDCLASS's or compilation will fail!
+        // _bind_methods is required for all GDCLASS's or compilation will fail!
+        static void _bind_methods() {}
 
     public:
-        // Empty default constructor is required or compilation will fail!
-        // GDExtension constructors also are not allowed to have additional parameters or compilation will fail!
+        // Default constructor with no parameters is required or compilation will fail!
         MySprite2D() {}
     };
+
+    } // CustomNamespace
+
     #endif // MY_SPRITE_2D_H
 
 Every GDScript file is implicitly a class. The ``extends`` keyword defines the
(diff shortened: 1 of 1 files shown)


getting_started/step_by_step/scripting_first_script.rst
```diff
@@ -343,3 +399,40 @@ Here is the complete ``sprite_2d.gd`` file for reference.
...
+    protected:
+        static void _bind_methods() {}
+
+    public:
+        MySprite2D() {
+            godot::UtilityFunctions::print("Hello, world!");
+        }
+
+        void _process(const double p_delta) override {
+            set_rotation(get_rotation() + (angular_speed * delta));
```
AThousandShips, line 429:
```suggestion
            set_rotation(get_rotation() + angular_speed * p_delta);
```
Unnecessary and hurts readability IMO when adding parentheses without being needed for precedence 

Also correct variable name

getting_started/step_by_step/scripting_first_script.rst
```diff
@@ -343,3 +399,40 @@ Here is the complete ``sprite_2d.gd`` file for reference.
...
+
+            Vector2 velocity = Vector2(0, -1).rotated(get_rotation()) * speed;
+            set_position(get_position() + (velocity * delta));
```
AThousandShips, line 432:
```suggestion
            set_position(get_position() + velocity * p_delta);
```

getting_started/step_by_step/scripting_first_script.rst
```diff
@@ -343,3 +399,40 @@ Here is the complete ``sprite_2d.gd`` file for reference.
...
+
+    protected:
+        static void _bind_methods() {}
+
+    public:
+        MySprite2D() {
+            godot::UtilityFunctions::print("Hello, world!");
+        }
+
+        void _process(const double p_delta) override {
```
AThousandShips, line 428:
```suggestion
        void _process(double p_delta) override {
```
No `const` for simple types

getting_started/step_by_step/scripting_first_script.rst
```diff
@@ -231,6 +271,14 @@ At the bottom of the script, define the function:
         Rotation += _angularSpeed * (float)delta;
     }
 
+ .. code-tab:: cpp C++
+
+    void _process(const double p_delta) override {
```
AThousandShips, line 276:
```suggestion
    void _process(double p_delta) override {
```

getting_started/step_by_step/scripting_first_script.rst
```diff
@@ -231,6 +271,14 @@ At the bottom of the script, define the function:
...
+        // Note that rotation is a private member variable of Node2D, so the set_rotation function must be used.
+        // You'll find that a lot of member variables that are public in gdscript/C# are instead private in GDExtension C++.
+        set_rotation(get_rotation() + (angular_speed * p_delta));
```
AThousandShips, line 279:
```suggestion
        set_rotation(get_rotation() + angular_speed * p_delta);
```

getting_started/step_by_step/scripting_first_script.rst
```diff
@@ -278,6 +326,14 @@ them.
 
     Position += velocity * (float)delta;
 
+ .. code-tab:: cpp C++
+
+    // Note that the directional Vector2 constants do not exist in GDExtension right now. So Vector2(0, -1) must be used.
+    Vector2 velocity = Vector2(0, -1).rotated(get_rotation()) * speed;
+
+    set_position(get_position() + (velocity * p_delta));
```
AThousandShips, line 334:
```suggestion
    set_position(get_position() + velocity * p_delta);
```

getting_started/step_by_step/scripting_first_script.rst
```diff
@@ -107,6 +109,33 @@ the following line of code:
...
+    #define MY_SPRITE_2D_H
+
+    #include <godot_cpp/classes/sprite2d.hpp>
+
+    namespace CustomNamespace {
+
+    // MySprite2D also needs to be bound in your GDExtension's register_types.cpp file.
+    // More info on this in the GDExtension example setup tutorial.
+    class MySprite2D : public godot::Sprite2D {
+        GDCLASS(godot::MySprite2D, godot::Object)
```
AThousandShips, line 124:
```suggestion
        GDCLASS(MySprite2D, godot::Sprite2D)
```
Correct type, also this class isn't in the `godot` namespace

getting_started/step_by_step/scripting_first_script.rst
```diff
@@ -278,6 +326,14 @@ them.
 
     Position += velocity * (float)delta;
 
+ .. code-tab:: cpp C++
+
+    // Note that the directional Vector2 constants do not exist in GDExtension right now. So Vector2(0, -1) must be used.
+    Vector2 velocity = Vector2(0, -1).rotated(get_rotation()) * speed;
```
AThousandShips, line 332: Unless you use `using namespace Godot;` you need to prefix the namespace
```suggestion
    godot::Vector2 velocity = godot::Vector2(0, -1).rotated(get_rotation()) * speed;
```

getting_started/step_by_step/scripting_first_script.rst
```diff
@@ -343,3 +399,40 @@ Here is the complete ``sprite_2d.gd`` file for reference.
...
+    #ifndef MY_SPRITE_2D_H
+    #define MY_SPRITE_2D_H
+
+    #include <godot_cpp/variant/utility_functions.hpp>
+    #include <godot_cpp/classes/sprite2d.hpp>
+    #include <godot_cpp/core/math.hpp>
+
+    namespace CustomNamespace {
+
+    class MySprite2D : public Sprite2D {
```
AThousandShips, line 414:
```suggestion
    class MySprite2D : public godot::Sprite2D {
```
Or use `using namespace godot;` above

getting_started/step_by_step/scripting_first_script.rst
```diff
@@ -343,3 +399,40 @@ Here is the complete ``sprite_2d.gd`` file for reference.
...
+        GDCLASS(MySprite2D, godot::Object)
```
AThousandShips, line 415:
```suggestion
        GDCLASS(MySprite2D, godot::Sprite2D)
```

getting_started/step_by_step/scripting_first_script.rst
```diff
@@ -343,3 +399,40 @@ Here is the complete ``sprite_2d.gd`` file for reference.
...
+
+    public:
+        MySprite2D() {
+            godot::UtilityFunctions::print("Hello, world!");
+        }
+
+        void _process(const double p_delta) override {
+            set_rotation(get_rotation() + (angular_speed * delta));
+
+            Vector2 velocity = Vector2(0, -1).rotated(get_rotation()) * speed;
```
AThousandShips, line 431:
```suggestion
            godot::Vector2 velocity = godot::Vector2(0, -1).rotated(get_rotation()) * speed;
```
Same note

AThousandShips (changes requested):

getting_started/step_by_step/scripting_first_script.rst
dementive: Will remove the parentheses agreed those do nothing. p_delta seems like the idomatic way to do it in C++ in both godot source code and GDExtension source code so I think it's best to not just copy the other scripting languages for that. Don't mind changing it though if you disagree.

AThousandShips: You are using `delta` not `p_delta`, you've got the suggestion backwards

getting_started/step_by_step/scripting_first_script.rst
dementive: Does adding/removing const here make any difference? I just copied the function signature from the source code so I actually don't know.

AThousandShips: It's per the style, will check if the extension code generates `const` in godot-cpp and if so it should be fixed for simple types

getting_started/step_by_step/scripting_first_script.rst
dementive: Do you think it would be best for the C++ documentation to use `using namespace godot;`? Or would throwing `godot::` on everything be better? The additional characters from using `godot::` might increase the mental load on reading the docs for begginers for no actual gain but then again `using namespace` is generally discouraged so it's hard what to say is best, up to you.

AThousandShips: The godot-cpp example uses `using` so I'd use that here too

getting_started/step_by_step/scripting_first_script.rst
dementive: Oops I see now it seems like I forgot my brain this morning

getting_started/step_by_step/scripting_first_script.rst
dementive: It looks like GDExtension does not use const in their _process function, so I will go with that

getting_started/step_by_step/scripting_first_script.rst
dementive: Nice that makes sense, i'll go with `using namespace godot;` then. One last question then i'll commit these tweaks, do you think it would be best to wrap the custom namespace like this:

```cpp
namespace CustomNamespace {

}
```

Or just do it with `using namespace CustomNamespace;` (this is what I do in my extension code, it makes the debug symbols easier to parse through having a custom namespace but idk what would be idomatic)

Or would it be best in the docs to not even worry about the custom namespace at all? Again totally up to you.

AThousandShips: You shouldn't use the `using` here as you're in the header (if you put the code in a `.cpp` file you should)

dementive: Gotcha, I'll leave it as is then and commit what I got. Makes sense why you wouldn't want that in the header 👍

dementive:
Should be all good now. I worked through everything you mentioned, thanks for the detailed feedback!

I am going to try to finish up the "Listening to player input" section over the weekend. Also will add a info box to the top of the "Using Signals" article with a link to the bottom of the "GDExtension Example" documentation that discussed signals. This seems like it will be sufficient since the articles are both basically identical, I don't see a point in filling out the "Using Signals" article with C++ snippets when the same info is in another place.

Looking forward I don't really see a point in adding C++ snippets to either of the "Your first X game" sections because if it is a users first attempt at using Godot they really shouldn't use C++ (in my opinion) since they will need to know a good amount of gdscript to effectively use GDExtension anyway.

So I'm thinking this will be good to merge once I finish up "Listening to player input" and "Using Signals".

getting_started/step_by_step/scripting_player_input.rst
```diff
@@ -95,6 +107,12 @@ Comment out the lines ``var velocity = Vector2.UP.rotated(rotation) * speed`` an
 
     //Position += velocity * (float)delta;
 
+ .. code-tab:: cpp C++
+
+    // Vector2 velocity = Vector2(0, -1).rotated(get_rotation()) * speed;
```
AThousandShips, line 112:
```suggestion
    //Vector2 velocity = Vector2(0, -1).rotated(get_rotation()) * speed;
```
Commented out code has no leading space

Maarrk:
This is absolutely lovely, I wish knew about this and #9837 when learning GDExtension!

I wonder how to link this best from the GDExtension tutorial? I only found out by browsing PR list, after I already went through the path of reading source code and error traces.

getting_started/step_by_step/scripting_first_script.rst
```diff
@@ -231,6 +278,14 @@ At the bottom of the script, define the function:
         Rotation += _angularSpeed * (float)delta;
     }
 
+ .. code-tab:: cpp C++
+
+    void _process(double p_delta) override {
+        // Note that rotation is a private member variable of Node2D, so the set_rotation function must be used.
+        // You'll find that a lot of member variables that are public in gdscript/C# are instead private in GDExtension C++.
```
dsnopek, line 285: It isn't so much a "private member" as properties in godot-cpp must be accessed via setters and getters.

```suggestion
        // Note that properties (like rotation) are accessed via setters and getters in godot-cpp.
GDExtension C++.
```

dsnopek (commented):
Thanks!

For the most part this looks really good, and would be nice to have, just to show people it's possible and what it looks like.

dementive:
Merged in your changes and squashed, should be good to go now.

kburchfiel:
Thanks so much @dementive ! Regarding the Your First Game tutorials: When I was learning to use C++ with GDNative in Godot 3, I found the C++ snippets to be extremely helpful in translating between GDScript and C++. So while I understand your reasoning behind skipping those, I would still love to help fill them in.

Fortunately, @j-dax [has already written C++ code ](https://github.com/j-dax/gd-cpp) for version 4.3's Your First 2D Game tutorial. My plan is to try to convert that code into snippets (and update them for 4.6 where needed). Once they're ready, or if I need some debugging help, I'll post an update here!

PR #9822, opened 2024-08-23
Update Random number generation for Godot 4.2+

Calinou:
- Document `Array.pick_random()` as a shorthand.
- Use global scope random methods as they are all available in global scope now.
- Remove calls to `randomize()` that are not needed for code samples to be functional, as the engine now calls `randomize()` on startup once automatically.

___

- See https://github.com/godotengine/godot-docs-user-notes/discussions/74#discussioncomment-10433139.

### DIFF
--- a/tutorials/math/random_number_generation.rst
+++ b/tutorials/math/random_number_generation.rst
@@ -33,9 +33,6 @@ Global scope methods are easier to set up, but they don't offer as much control.
 RandomNumberGenerator requires more code to use, but allows creating
 multiple instances, each with their own seed and state.
 
-This tutorial uses global scope methods, except when the method only exists in
-the RandomNumberGenerator class.
-
 The randomize() method
 ----------------------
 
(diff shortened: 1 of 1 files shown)


tutorials/math/random_number_generation.rst
```diff
@@ -186,20 +186,24 @@ and ``to``, and returns a random integer between ``from`` and ``to``:
...
-        randomize()
-
         for i in range(100):
             # Pick 100 fruits randomly.
             print(get_fruit())
 
+        for i in range(100):
+            # Pick 100 fruits randomly, this time using the `Array.pick_random()`
+            # helper method. This has the same behavior as `get_fruit()`.
+            print(_fruits.pick_random())
```
raulsntos, line 206: `pick_random` is also available in C# (as `PickRandom` in `Godot.Collections.Array`) so you could add this to the C# code, but you'll have to change the the type of the `_fruits` field like so:

```csharp
private Godot.Collections.Array<string> _fruits = new Godot.Collections.Array<string> { "apple", "orange", "pear", "banana" };
```

We usually prefer the BCL collections which is probably why this example was using `string[]` but I guess it makes sense to use Godot collections in this case, in order to document this API.

KleidonBD:
I also noticed that the RandomNumberGenerator.randi_range() method is still listed here, but there is a global scope randi_range() method. The page states that the tutorial only uses the RandomNumberGenerator class for methods that only exist within that class, so I think this should be updated to use the global scope method instead (which I assume didn't exist when the page was written).
https://docs.godotengine.org/en/4.3/classes/class_%40globalscope.html#class-globalscope-method-randi-range

mhilbrunner (approved):

Mickeon:
[rand_weighted](https://docs.godotengine.org/en/stable/classes/class_randomnumbergenerator.html#class-randomnumbergenerator-method-rand-weighted) exists too now, so that could be documented in another PR as well (as that can't be cherrypicked for 4.2).

mhilbrunner:

PR #9830, opened 2024-08-25
added note on calling behavior

Sheepy3:
added a note on calling behavior to the documentation in lieu of my own trouble with the signals related to the synchronizer and this issue: https://github.com/godotengine/godot/issues/88587


AThousandShips:

PR #9842, opened 2024-08-26
Corrected error in sort_custom section of class_array.rst

MajorGonzo:
Corrected the array indices used in the lambda function of the sort descending example to match the indices used in the function  sort_ascending.  

As written:
```
func sort_ascending(a, b):
	if a[1] < b[1]:
		return true
	return false
```

in ready function:

```
	my_items.sort_custom(func(a, b): return a[0] > b[0])
	print(my_items) # Prints [["Apple", 9], ["Tomato", 5], ["Rice", 4]]
```

As written, the output is incorrect, as it would print Tomato, Rice, Apple order


AThousandShips:

This is already covered by:
* https://github.com/godotengine/godot/pull/96118

PR #9846, opened 2024-08-27
Fix shader language matrix row/col convention

### LINKED ISSUE #5805: Confusing matrix row/col convention in shader language documentation
**Your Godot version:**
3.5 and 4

**Issue description:**
Documentation is confusing about the row / column convention for matrices creation and access in shaders.

### In the data type section
`mat{x}` types are described as :
> {x}x{x} matrix, in column major order.

The term “column major” normally refers to the internal storage layout of the matrix in memory (in this case, one column after another, left to right).
No matter if it’s correct or not, this is an implementation detail the user should not have to care about.

**👉 I suggest to remove this information from the chart.**

Instead, what matters for the user is :
1. the meaning of the indexing convention the matrix’ components (scalar or sub-vector) are accessed with
2. the convention used when creating a matrix from a literal

That’s what im focusing on below

### 1. Access (“Members” section)
The doc says :
> For matrices, use the m[column][row] indexing syntax to access each scalar, or m[idx] to access a vector by row index. For example, for accessing the y position of an object in a mat4 you use m[3][1].

2 suggestions here :
- if godot follows glsl conventions, then `m[column][row]` is correct, but `m[idx] to access a vector by row index` is incorrect. **👉 I don’t know the right answer, but whatever it is it would be nice to explicit mention commonalities / differences with the glsl standard.**
- the example given right after sounds completely opaque to me (so I assume the same for most users !) **👉 I suggest to replace by anything more meaningful, or to simply remove it**

### 2. Constructing section
The documentation doesn’t specify whether `vec`s count for lines or columns in the following statement :
>mat2 m2 = mat2(vec2(1.0, 0.0), vec2(0.0, 1.0));

**👉 Here again I don’t know the right answer but the doc should be explicit on that**

https://docs.godotengine.org/en/3.5/tutorials/shaders/shader_reference/shading_language.html

### LINKED ISSUE #9616: Shader matrix column/row major ordering very unclear
**Your Godot version:**

Godot v4.2.2

**Issue description:**

The Godot shader language documentation for matrix column- vs. row-major ordering is very unclear:

1. There is no specification as to whether `matx( vecx, vecx... )` [constructor form](https://github.com/godotengine/godot-docs/blob/4.2/tutorials/shaders/shader_reference/shading_language.rst?plain=1#L145) is column- or row- major.
2. The documentation states that [matrix index access is row-major](https://github.com/godotengine/godot-docs/blob/4.2/tutorials/shaders/shader_reference/shading_language.rst?plain=1#L120) (i.e. `matrix[2]` is a row not a column) which completely diverts from GLSL, and also seems like a terrible decision since two-dimensional array access is column-major (`matrix[2][1]` is the 2nd column, then the 1st row). Is this true? Or is it a documentation discrepancy?

Live documentation:
https://docs.godotengine.org/en/stable/tutorials/shaders/shader_reference/shading_language.html#members

GitHub file:
https://github.com/godotengine/godot-docs/blob/4.2/tutorials/shaders/shader_reference/shading_language.rst?plain=1#L112

Flarkk:
Fixes #5805
Fixes #9616

### DIFF
--- a/tutorials/shaders/shader_reference/shading_language.rst
+++ b/tutorials/shaders/shader_reference/shading_language.rst
@@ -155,8 +155,8 @@ Individual scalar members of vector types are accessed via the "x", "y", "z" and
 equivalent. Use whatever fits best for your needs.
 
 For matrices, use the ``m[column][row]`` indexing syntax to access each scalar,
-or ``m[idx]`` to access a vector by row index. For example, for accessing the y
-position of an object in a mat4 you use ``m[3][1]``.
+or ``m[column]`` to access a vector by column index. For example, for accessing the
+y-component of the translation from a mat4 transform matrix (4th column, 2nd line) you use ``m[3][1]`` or ``m[3].y``.
 
 Constructing
 ~~~~~~~~~~~~
@@ -174,7 +174,7 @@ Construction of vector types must always pass:
     vec4 a = vec4(0.0);
 
 Construction of matrix types requires vectors of the same dimension as the
-matrix. You can also build a diagonal matrix using ``matx(float)`` syntax.
+matrix, interpreted as columns. You can also build a diagonal matrix using ``matx(float)`` syntax.
 Accordingly, ``mat4(1.0)`` is an identity matrix.
 
 .. code-block:: glsl


Mickeon (commented):
This seems to be factually correct and what I have come to know myself, but an additional look by one more shader-savvy person may be necessary.

clayjohn (approved):

mhilbrunner (approved):

mhilbrunner:
Merged. Thanks and congrats on your first merged contribution! Sorry this took a bit :)

mhilbrunner:

PR #9853, opened 2024-08-27
Note that MingW64 DLL location needs to be in $PATH on Windows

danila-zol:
Based on this issue: https://github.com/godotengine/godot/issues/96190

### DIFF
--- a/contributing/development/compiling/compiling_for_windows.rst
+++ b/contributing/development/compiling/compiling_for_windows.rst
@@ -113,7 +113,10 @@ If you do not have Visual Studio installed, it will attempt to use
 MinGW instead. If you already have Visual Studio installed and want to
 use MinGW-w64, pass ``use_mingw=yes`` to the SCons command line. Note that MSVC
 builds cannot be performed from the MSYS2 or MinGW shells. Use either
-``cmd.exe`` or PowerShell instead. If you are using MinGW-LLVM, pass both
+``cmd.exe`` or PowerShell instead. You also need to add the directory
+that MingW64 is installed to (for exmaple `C:\msys64\mingw64\bin\`) to your `$PATH`,
+otherwise it will fail to locate DLL files required by the toolchain. 
+If you are using MinGW-LLVM, pass both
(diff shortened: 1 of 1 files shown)


contributing/development/compiling/compiling_for_windows.rst
```diff
@@ -113,7 +113,10 @@ If you do not have Visual Studio installed, it will attempt to use
 MinGW instead. If you already have Visual Studio installed and want to
 use MinGW-w64, pass ``use_mingw=yes`` to the SCons command line. Note that MSVC
 builds cannot be performed from the MSYS2 or MinGW shells. Use either
-``cmd.exe`` or PowerShell instead. If you are using MinGW-LLVM, pass both
+``cmd.exe`` or PowerShell instead. You also need to add the directory
```
bruvzg, line 116: If you are using MSYS2 you should use MSYS2 [own shell](https://www.msys2.org/docs/terminals/) instead of `cmd` or `PowerShell`. Adding parts of `$PATH` might work in some cases, but it's not how it is supposed to be used. And MSYS2 shell should have environment already pre-configured.

danila-zol: That is correct in general, but the same page explicitly states: 
> Note that MSVC builds cannot be performed from the MSYS2 or MinGW shells. Use either cmd.exe or PowerShell instead. [[link](https://docs.godotengine.org/en/latest/contributing/development/compiling/compiling_for_windows.html#selecting-a-compiler)]   

So, you have to rely on this crutch to build Godot and I think this should be documented.

bruvzg: > That is correct in general, but the same page explicitly states:

How is it related to MSVC builds? 

- For MSVC build, use MSVC shell (for target arch) or generic cmd/PowerShell.
- For MSYS2/MinGW build, use MSYS2 shell (for target arch).
- For other MinGW distributions build, set `MINGW_PREFIX` and use cmd/PowerShell.

Setting `PATH` usually should not be necessary.

PR #9858, opened 2024-08-29
Added openKylin system installation dependencies

DSOE1024:
Although openKylin also uses apt as the package manager, compiling the Godot dependent installation on the openKylin system is slightly different, requiring python3-pip to be installed before scons is installed with pip, which makes it easier for Godot developers on the openKylin system to copy commands from the document


mhilbrunner:
Thank you! Merged this in https://github.com/godotengine/godot-docs/pull/10277, moving this above `openSUSE`, giving you credit.

PR #9870, opened 2024-08-31
Update TileMap node with Layer postfix in using_tilemaps.rst

Lukas-Fletcher:
There are a mix of references between TileMap and TileMapLayer nodes but if I've understood correctly it should just reference the latter. "TileMap editor" references seem correct however.

PR #9871, opened 2024-08-31
Add commas to sentence in using_tilemaps.rst

Lukas-Fletcher:
Will conflict with #9870. Could be changed to parentheses instead of commas.
I got confused when reading the sentence because "the layers you aren't currently editing from" all flow together but then what's left of sentence doesn't make sense "the same TileMap node will appear grayed out while in the TileMap editor". Hope the extra commas will help with the flow.
Another solution would be to rephrase it as "In the 2D editor, the layers from the same TileMap node that you aren't currently editing will appear grayed out while in the TileMap editor." or something along those lines.

### DIFF
--- a/tutorials/2d/using_tilemaps.rst
+++ b/tutorials/2d/using_tilemaps.rst
@@ -126,9 +126,9 @@ layer you wish to paint on:
 
 .. tip::
 
-    In the 2D editor, the layers you aren't currently editing from the same
-    TileMap node will appear grayed out while in the TileMap editor. You can
-    disable this behavior by clicking the icon next to the layer selection menu
+    In the 2D editor, the layers you aren't currently editing,
+    from the same TileMap node, will appear grayed out while in the TileMap editor.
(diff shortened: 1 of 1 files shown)


Geometror (commented):
What about "In the 2D editor, the layers of a TileMap node you aren't currently editing will appear grayed out." ?
I've checked the behavior and the last part ("... in the TileMap editor") isn't actually the case (or I'm misunderstanding) so I think it can be removed, which makes the sentence even clearer.

Lukas-Fletcher:
@Geometror After some testing my understanding is also that it isn't actually the case. I agree that part probably should be removed. I don't know at what stage the transitioning from TileMap to TileMapLayer in the documentation is in but "In the 2D editor, the TileMapLayer nodes you aren't currently editing will appear grayed out." might be more fitting.

skyace65:
This needs to be updated since #9870 was merged

skyace65:
Closing due to a lack of updates, also not 100% sure if this is grammatically necessary or not

PR #9873, opened 2024-08-31
Document sub plugins

### LINKED ISSUE #9571: Adding autoloads from custom plugin: example in the doc is bad practice
**Your Godot version:** 4.2.2 and 4.3.beta-2

**Issue description:** see the doc at link below. It shows an example of adding and removing autoloads in `_enter_tree()`, `_exit_tree()`. I think this is not correct way to to do this because:
- This messes up the order of autoloads that the user may have defined
- This does not preserve the order of autoloads, if the user reordered them
- This sometimes causes problems, e.g. an autoload is added by such plugin much later than compiler checks for the presence of this autoload when it is already referenced from other code / autoloads
- This is just ineffective. No need to add & remove autoload each time the editor starts up and shuts down

The problem is much worse than it may look at the first glance, because some plugins out there already follow this bad practice.

Correct me if I'm wrong, but the proper way to add autoloads from a plugin is:
```gdscript
@tool
extends EditorPlugin

func _enable_plugin() -> void:
	add_autoload_singleton("MyAutoload", "res://my_autoload.gd")

func _disable_plugin() -> void:
	remove_autoload_singleton("MyAutoload")
```

**URL to the documentation page (if already existing):** https://docs.godotengine.org/en/stable/tutorials/plugins/editor/making_plugins.html#registering-autoloads-singletons-in-plugins

HolonProduction:
https://github.com/godotengine/godot-proposals/issues/3782 and https://github.com/godotengine/godot-proposals/issues/10556 ask for a way to add sub plugins, which as it turns out is already possible. With this PR it is also a documented feature.

I also made some other small changes:
- fixed #9571
- moved the "Going beyond" section to the very end, adding autoloads is a very basic thing to do in a plugin and shouldn't be placed after it. One could argue that sub-plugins should be placed behind it, but the section feels more like a closing to the whole page. (The complete plugin section might need a little restructuring at some point though, we have installing plugins then multiple pages named as "making ... plugins" and then some named just as "... plugins" which also talk about making plugins, so they should either be renamed, or "making plugins" should become a collection of pages with everything related to making plugins below it.)
- Upgraded "A custom panel" to a real heading, since it isn't a subsection of "A custom node"


tutorials/plugins/editor/making_plugins.rst
```diff
@@ -469,15 +457,62 @@ Use the following code to register a singleton from an editor plugin:
...
             RemoveAutoloadSingleton(AutoloadName);
         }
     }
     #endif
+
+Using sub-plugins
+~~~~~~~~~~~~~~~~~
+
+Often a plugin adds multiple things, for example a custom node and a panel.
+In those cases it might be easier to have an own plugin script for each of those features.
```
AThousandShips, line 477:
```suggestion
In those cases it might be easier to have a separate plugin script for each of those features.
```

tutorials/plugins/editor/making_plugins.rst
```diff
@@ -469,15 +457,62 @@ Use the following code to register a singleton from an editor plugin:
...
+Sub-plugin can be used for this.
```
AThousandShips, line 478:
```suggestion
Sub-pluginz can be used for this.
```

tutorials/plugins/editor/making_plugins.rst
```diff
@@ -469,15 +457,62 @@ Use the following code to register a singleton from an editor plugin:
...
+
+First create all plugins and sub plugins as normal plugins:
+
+.. image:: img/sub_plugin_creation.webp
+
+Then move the sub plugins into the main plugin folder:
+
+.. image:: img/sub_plugin_moved.webp
+
+Godot will hide sub-plugins from the plugin list, so to a user can't enable or disable them.
```
AThousandShips, line 488:
```suggestion
Godot will hide sub-plugins from the plugin list, so that a user can't enable or disable them.
```

tutorials/plugins/editor/making_plugins.rst
HolonProduction: Do you mean `Sub-plugins`?

AThousandShips: Yes my bad

HolonProduction: Well with English I'm never sure about anything 😅

AThousandShips: English is massively confusing and just three or four languages in a trenchcoat 🙃

mhilbrunner (approved):

mhilbrunner:
Thank you! Merged. Oof, this took a bit, sorry :)

PR #9878, opened 2024-09-01
Reduce imprecision in Multiple Resolutions doc

michaelgundlach:
I tried to make this doc page clearer and remove bugs, because I find it hard to understand as is.

- The resizing discussion mixed talking about resizing to fit a screen and resizing to fit a window.  Because the animated gifs do a great job of demonstrating the Stretch Mode differences using resizing of windows, I changed most "screen" references to "window" references.

- Replaced the word "display" with "screen" in most cases, as it's an unneeded synonym that made me wonder if there was a difference.

- Corrected several instances of referring to the viewport width and height settings as window width and height, or base window width and base window height.

- Standardized referring to "base size" in most places instead of sometimes referring to "viewport size" or to "base window size".

- Fleshed out the description of Canvas Item vs Viewport stretch mode.  Maybe I made it worse, but I had to go onto reddit to find a clearer description than what existed.  What clicked for me was a reddit summary saying "Canvas Items first scales the images, and then renders them. Viewport renders the images, and then scales them." I was then able to (hopefully!) clarify the mode descriptions.

- Replaced Keep Height description with basically "like Keep Width but for height", rather than making the user check the copy-and-modified description vs Keep Width's for differences.

- Asserted that Stretch Mode=Viewport completely avoids scaling artifacts when Stretch Scale Mode is set to Integer.  If I'm wrong, please let me know.  It's still not clear to me why the Stretch Mode=Viewport example animated gif never looks fuzzy, assuming the author recorded it using Stretch Scale Mode=Fractional, since the Stretch Scale Mode section seems to imply that random resizing of small resolution sprites should cause artifacts unless Stretch Scale Mode=Integer.

I may have introduced bugs due to not perfectly understanding the subject matter, and am happy to revise further if you explain what I got wrong.

### DIFF
--- a/tutorials/rendering/multiple_resolutions.rst
+++ b/tutorials/rendering/multiple_resolutions.rst
@@ -56,8 +56,8 @@ Godot provides several useful tools to do this easily.
 Base size
 ---------
 
-A base size for the window can be specified in the Project Settings under
-**Display → Window**.
+A base size for the game window can be specified in the Project Settings
+under **Display → Window**.
 
 .. image:: img/screenres.webp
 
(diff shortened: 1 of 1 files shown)


AThousandShips:

Please open a new PR and a new branch based on the correct branch, if needed (check the latest version first).

PR #9880, opened 2024-09-01

ffslight:
I followed the guide to create the first game with Godot, and i found it confusing that Mobs would not spawn following the MobSpawnLocation.  After some debugging i figured it was because they spawn at local position and not global position. To make it more clear to newcomers i suggest either using GlobalPosition OR adding a note that explains the difference between local and global position. I opted for the first solution and created this pull request with the change.

### DIFF
--- a/getting_started/first_2d_game/05.the_main_game_scene.rst
+++ b/getting_started/first_2d_game/05.the_main_game_scene.rst
@@ -206,7 +206,7 @@ Note that a new instance must be added to the scene using ``add_child()``.
         var direction = mob_spawn_location.rotation + PI / 2
 
         # Set the mob's position to a random location.
-        mob.position = mob_spawn_location.position
+        mob.position = mob_spawn_location.global_position
 
         # Add some randomness to the direction.
         direction += randf_range(-PI / 4, PI / 4)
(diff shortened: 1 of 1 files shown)


AThousandShips:
Thank you for your contribution, however this is not a necessary change and the code will work correctly when applied correctly, this is a duplicate of:
* https://github.com/godotengine/godot-docs/pull/8700

Which was closed because it was caused by incorrectly following the instructions and putting the nodes in the wrong places in the hierarchy, causing the positions to be wrong

Please use [this code](https://github.com/godotengine/godot-demo-projects/blob/master/2d/dodge_the_creeps/main.gd) to confirm that your code is actually correct, and [this scene](https://github.com/godotengine/godot-demo-projects/blob/master/2d/dodge_the_creeps/main.tscn) to make sure you set the scene up correctly

Closing as this is not a necessary or valid change, but thank you for your contribution nonetheless

ffslight:
I agree the code works. However the confusing part is the same as in the duplicate. We draw a rectangle and would expect mobs to spawn along that rectangle but they do not, which is hard for newcomers to understand. And as such the tutorial makes a poor example of the followpath object. So I still suggest that either explain the difference between global and local position or set it to global would be clear. The point of the suggested change is solely to make the tutorial easier to understand.Den 2. sep. 2024 kl. 14.49 skrev A Thousand Ships ***@***.***>:﻿
Thank you for your contribution, however this is not a necessary change and the code will work correctly when applied correctly, this is a duplicate of:

#8700

Which was closed because it was caused by incorrectly following the instructions and putting the nodes in the wrong places in the hierarchy, causing the positions to be wrong
Please use this code to confirm that your code is actually correct, and this scene to make sure you set the scene up correctly
Closing as this is not a necessary or valid change, but thank you for your contribution nonetheless

AThousandShips:
As the changes here are incorrect please open an issue to discuss adding some form of explanation

The drawn rectangle *is* where the mobs will spawn, so there's no confusion there and that confusion isn't related at all to the global position, and IMO adding unrelated information just makes the tutorial harder by introducing unnecessary things

So please open an issue to discuss if this should be clarified and if it is agreed on you can open a new PR with the relevant change

PR #9881, opened 2024-09-01
Improve clarity in Multiple Resolutions doc

michaelgundlach:
I tried to make this doc page clearer and remove bugs, because I find it hard to understand as is.

- The resizing discussion mixed talking about resizing to fit a screen and resizing to fit a window.  Because the animated gifs do a great job of demonstrating the Stretch Mode differences using resizing of windows, I changed most "screen" references to "window" references.

- Replaced the word "display" with "screen" in most cases, as it's an unneeded synonym that made me wonder if there was a difference.

- Corrected several instances of referring to the viewport width and height settings as window width and height, or base window width and base window height.

- Standardized referring to "base size" in most places instead of sometimes referring to "viewport size" or to "base window size".

- Fleshed out the description of Canvas Item vs Viewport stretch mode. Maybe I made it worse, but I had to go onto reddit to find a clearer description than what existed.  What clicked for me was a reddit summary saying "Canvas Items first scales the images, and then renders them. Viewport renders the images, and then scales them." I was then able to (hopefully!) clarify the mode descriptions.

- Replaced Keep Height description with basically "like Keep Width but for height", rather than making the user check the copy-and-modified description vs Keep Width's for differences.

- Asserted that Stretch Mode=Viewport completely avoids scaling artifacts when Stretch Scale Mode is set to Integer.  If I'm wrong, please let me know.  It's still not clear to me why the Stretch Mode=Viewport example animated gif never looks fuzzy, assuming the author recorded it using Stretch Scale Mode=Fractional, since the Stretch Scale Mode section seems to imply that random resizing of small resolution sprites should cause artifacts unless Stretch Scale Mode=Integer.

- Applied style guide to some long lines.

I may have introduced bugs due to not perfectly understanding the subject matter, and am happy to revise further if you explain what I got wrong.

### DIFF
--- a/tutorials/rendering/multiple_resolutions.rst
+++ b/tutorials/rendering/multiple_resolutions.rst
@@ -56,8 +56,8 @@ Godot provides several useful tools to do this easily.
 Base size
 ---------
 
-A base size for the window can be specified in the Project Settings under
-**Display → Window**.
+A base size for the game window can be specified in the Project Settings
+under **Display → Window**.
 
 .. image:: img/screenres.webp
 
@@ -70,17 +70,9 @@ blue rectangle in the 2D editor.
 There is often a need to support devices with screen and window sizes
 that are different from this base size. Godot offers many ways to
 control how the viewport will be resized and stretched to different
-screen sizes.
+screen and window sizes.
 
-.. note::
-
-   On this page, *window* refers to the screen area allotted to your game
-   by the system, while *viewport* refers to the root object (accessible
-   from ``get_tree().root``) which the game controls to fill this screen area.
-   This viewport is a :ref:`Window <class_Window>` instance. Recall from the
-   :ref:`introduction <doc_viewports>` that *all* Window objects are viewports.
-
-To configure the stretch base size at runtime from a script, use the
+To configure the base size stretch at runtime from a script, use the
 ``get_tree().root.content_scale_size`` property (see
 :ref:`Window.content_scale_size <class_Window_property_content_scale_size>`).
 Changing this value can indirectly change the size of 2D elements. However, to
@@ -122,6 +114,21 @@ most flexible way to deal with the problem, it can be a lot of work,
 code and guessing, so Godot provides a set of parameters in the
 project settings to handle multiple resolutions.
 
+.. note::
+
+   While games are usually run in fullscreen mode, so that the game window's
+   resolution is equal to the screen resolution, the same challenges arise
+   when the game is run in a window resized by the user to an arbitrary
+   width and height. For the rest of this page, we will discuss resizing
+   your game to fit a *window*, but the same logic applies when the game is 
+   running in fullscreen mode.
+
+   On this page, *window* refers to the screen area allotted to your game
+   by the system, while *viewport* refers to the root object (accessible
+   from ``get_tree().root``) which the game controls to fill this screen area.
+   This viewport is a :ref:`Window <class_Window>` instance. Recall from the
+   :ref:`introduction <doc_viewports>` that *all* Window objects are viewports.
+
 Stretch settings
 ----------------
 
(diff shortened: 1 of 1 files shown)


michaelgundlach:
If this is too large a change please let me know and I can break it up.  Or, if it made it too hard to review because I fixed line spacing, I can undo that.

tetrapod00:
Hey, sorry for the long wait on any feedback. I skimmed this and it generally looks good, and I've been meaning to review it more deeply.

> If this is too large a change please let me know and I can break it up. 

This is more for future reference, but:

Splitting things can definitely make things easier to review, if you can split out the easily verifiable changes from the changes that require more knowledge to verify (and therefore require a reviewer with niche knowledge who may not be available). In this case, the page seems to contain a mixture of both intermixed, so I'm not sure splitting things would be easy or help.

Splitting out *new* content from changes to existing content can also be useful.

Splitting merely into *sections of the page* might not help, if each section still requires a specific area reviewer.

tutorials/rendering/multiple_resolutions.rst
```diff
@@ -132,39 +139,48 @@ Stretch settings are located in the project settings and provide several options
...
-   setting into account). This means that everything is rendered
-   directly at the target resolution. 3D is unaffected,
-   while in 2D, there is no longer a 1:1 correspondence between sprite
-   pixels and screen pixels, which may result in scaling artifacts.
+-  **Stretch Mode = Canvas Items**: The root :ref:`Viewport <class_Viewport>`
+   is resized from the base size to the window's resolution (taking the
+   **Stretch Aspect** setting into account). Each CanvasItem is then
+   scaled by the same amount as the base size was scaled.  For example, 
+   if the base size is 16x9 and the window is 160x180, each CanvasItem
+   will be stretched by a factor of 10 horizontally and 20 vertically.
```
tetrapod00, line 167: >  For example, if the base size is 16x9 and the window is 160x180, each CanvasItem will be stretched by a factor of 10 horizontally and 20 vertically.

I don't think this is true in every case, and it seems to directly contradict the gif just below it? It would be true if `Stretch Aspect` was `ignore`.

tutorials/rendering/multiple_resolutions.rst
```diff
@@ -132,39 +139,48 @@ Stretch settings are located in the project settings and provide several options
...
+   This means that everything is rendered at the window's resolution.
+   3D is unaffected, while in 2D, there is no longer a 1:1 correspondence
+   between sprite pixels and window pixels, which may result in scaling
+   artifacts. For example, in the animation below, the "16:9" sprite is
+   being scaled by arbitrary amounts, which rarely aligns the edges of 
+   the sprite image's pixels with the window pixels.
```
tetrapod00, line 173: The rest of this paragraph needs a review from someone very familiar with viewport scaling.

tutorials/rendering/multiple_resolutions.rst
```diff
@@ -265,17 +281,17 @@ what the **Stretch** options above already provide. The default value of ``1.0``
...
-**Disabled**, each unit in your scene will correspond to 2×2 pixels on the
-screen. This is a good way to provide scaling options for non-game applications.
+**Disabled**, each unit in your scene will correspond to 2×2 pixels in the
+window. This is a good way to provide scaling options for non-game applications.
 
-If **Stretch Mode** is set to **canvas_items**, 2D elements will be scaled
-relative to the base window size, then multiplied by the **Scale** setting. This
-can be exposed to players to allow them to adjust the automatically determined
-scale to their liking, for better accessibility.
+If **Stretch Mode** is set to **Canvas Items**, 2D elements will be scaled
```
tetrapod00, line 287:
```suggestion
If **Stretch Mode** is set to ``canvas_items``, 2D elements will be scaled
```
This seemed right before, assuming we're talking about how it appears in the project settings. Though `code` style is preferred to **bold** here.

tutorials/rendering/multiple_resolutions.rst
```diff
@@ -265,17 +281,17 @@ what the **Stretch** options above already provide. The default value of ``1.0``
...
+relative to the viewport's base size, then multiplied by the **Scale** setting.
+This can be exposed to players to allow them to adjust the automatically
+determined scale to their liking, for better accessibility.
 
-If **Stretch Mode** is set to **viewport**, the viewport's resolution is divided
+If **Stretch Mode** is set to **Viewport**, the viewport's base size is divided
```
tetrapod00, line 292:
```suggestion
If **Stretch Mode** is set to ``viewport``, the viewport's base size is divided
```
See above.

tutorials/rendering/multiple_resolutions.rst
```diff
@@ -265,17 +281,17 @@ what the **Stretch** options above already provide. The default value of ``1.0``
 means that no additional scaling occurs.
 
 For example, if you set **Scale** to ``2.0`` and leave **Stretch Mode** on
-**Disabled**, each unit in your scene will correspond to 2×2 pixels on the
-screen. This is a good way to provide scaling options for non-game applications.
+**Disabled**, each unit in your scene will correspond to 2×2 pixels in the
```
tetrapod00, line 284:
```suggestion
``disabled``, each unit in your scene will correspond to 2×2 pixels in the
```

tetrapod00 (changes requested):
Overall I think this is an improvement, once changes are applied. The revised page reads about as well to me as the original, and I didn't catch any obvious inaccuracies besides the paragraph I pointed out. I like reducing synonyms within this page.

This really needs a review from someone more familiar with the content.

skyace65:
Closing due to the lack of a response to tetrapod's review.

PR #9882, opened 2024-09-01
Add hyphens to adjective phrases

JohnVeness:
Hi. I have done some searching and replacing within the docs, adding hyphens where grammatically correct, that is, when something is used as an adjective ("open-source license") versus a noun ("Godot is open source"). Hope this is welcome!

### DIFF
--- a/about/complying_with_licenses.rst
+++ b/about/complying_with_licenses.rst
@@ -88,14 +88,14 @@ Credits screen
 
 Include the above license text somewhere in the credits screen. It can be at the
 bottom after showing the rest of the credits. Most large studios use this
-approach with open source licenses.
+approach with open-source licenses.
 
 Licenses screen
 ^^^^^^^^^^^^^^^
 
 Some games have a special menu (often in the settings) to display licenses.
 This menu is typically accessed with a button called **Third-party Licenses**
-or **Open Source Licenses**.
+or **Open-source Licenses**.
 
 Output log
 ^^^^^^^^^^
@@ -147,7 +147,7 @@ Godot itself contains software written by
 `third parties <https://github.com/godotengine/godot/blob/master/thirdparty/README.md>`_,
 which is compatible with, but not covered by Godot's MIT license.
 
-Many of these dependencies are distributed under permissive open source licenses
+Many of these dependencies are distributed under permissive open-source licenses
 which require attribution by explicitly citing their copyright statement and
 license text in the final product's documentation.
 
--- a/about/faq.rst
+++ b/about/faq.rst
@@ -11,7 +11,7 @@ Frequently asked questions
 What can I do with Godot? How much does it cost? What are the license terms?
 ----------------------------------------------------------------------------
 
-Godot is `Free and open source Software <https://en.wikipedia.org/wiki/Free_and_open_source_software>`_
+Godot is `Free and open-source Software <https://en.wikipedia.org/wiki/Free_and_open-source_software>`_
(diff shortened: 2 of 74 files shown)


AThousandShips:
This being a minor language quirk and nitpick makes it important to consider the impact on translations as this will invalidate existing translated content

skyace65 (changes requested):
In addition to what AThousandShips said "Open-Source" doesn't seem to actually be grammatically correct. it's written "open source" without hyphen on wikipedia as well as the Red Hat website.

JohnVeness:
"Open-source" when it's an adjective, not a noun. E.g. https://en.wikipedia.org/wiki/Free_and_open-source_software

tetrapod00:
I'm generally in favor of these kinds of hyphenation standardizations for the docs, so I'm not per se opposed to this. However, it's better to start from *existing usages* and standardize, rather than choosing the form that is most correct in the abstract.

In the current manual (excluding the class ref pages):
- "open-source" is used 3 times, "open source" 42 times. This one I don't think makes sense to change even if it's technically wrong.
- "cross-platform" 4 times,  "cross platform" 3 times. Sure, change this.
- "command-line" 19 times, "command line" 107 times. I don't know about changing this one.
- "platform-specific" 11 times, "platform specific" 3 times. This one makes sense to change

I'm not sure if this is an exhaustive list of the phrases you checked, just took a cursory look. Seems like adjective phrases that *only* exist as adjective phrases are good candidates to use hyphens for, but adjective phrases that also exist as nouns perhaps not.

I actually *did* change some of the "built in" instances you found to "built-in" in https://github.com/godotengine/godot-docs/pull/10235. Note that Godot uses "built-in" as a noun quite often in place of the full form "built-in variable".

I'd suggest maybe making another PR for the less controversial set of changes here (everything that isn't "open source" or "command line".

mhilbrunner:
I agree with what was discussed here. Closing as per the above. Also, if we are going to make (a subset of) these changes, we should keep them to `master` to release in the next version, although even then breaking translations for this isn't the best.

Thanks for taking the time to contribute and add all these hyphens, even if we're not going ahead with this :)

JohnVeness:
Thanks for all the notes above. If I do decide to suggest similar changes again, I'll likely break it down into single words, to make it easier to accept/reject a PR.

I like the look of @tetrapod00's work in #10218 and related PRs. Apologies if I've overlooked one, but it would be great if there was a Godot documentation style guide, such as the Raspberry Pi one at https://github.com/raspberrypi/style-guide/blob/master/style-guide.md

tetrapod00:
Hey, @JohnVeness, I went ahead and extracted out the improvements which I think have a good chance of being merged into two new PRs (https://github.com/godotengine/godot-docs/pull/10284 and https://github.com/godotengine/godot-docs/pull/10285). You're listed as the co-author on the commits.

JohnVeness:
Thanks for picking this up and submitting the subset PRs. And thanks for pointing me to the existing style guide, and for your work improving it.

I agree that it can be hard to enforce a style guide, and I think "enforce" is too strong a word anyway! I'm very happy for non-technical writers to write documentation within their abilities and for other people to proofread/edit it later if necessary. It's just that without such a style guide, different people might have different ideas about (or might just plain forget!) what typical changes should be made.

tetrapod00:
Yeah, I think my ideal outcome with a detailed style guide is that we have a short version that we expect *most* casual contributors to read and apply as best they can, and also a long detailed one which we expect maintainers to read and remember at least some of. Casual contributors can read it if they wish. During review, suggested changes can optionally come with links to the relevant sections of the style guide.

PR #9885, opened 2024-09-02
Fix ambiguity in angle_difference function

LunaTheFoxgirl:
The current wording does not make it clear that the unit received by the function is in radians (or -PI to PI).
To make it more clear to the end user that the function should get a signed radian instead of angles, this has been explicitly added.


AThousandShips:

PR #9889, opened 2024-09-02
Add clarity around uncommenting code

LTangaF:
The commenting out of the position isn't obviously restored by the code snippets, so drawing a little extra attention to it in the descriptive text seems advisable as there's a reasonable chance that it won't be noticed right away and the "student" will be initially confused that they're sprite doesn't move forward.

### DIFF
--- a/getting_started/step_by_step/scripting_player_input.rst
+++ b/getting_started/step_by_step/scripting_player_input.rst
@@ -104,7 +104,7 @@ Moving when pressing "up"
 -------------------------
 
 To only move when pressing a key, we need to modify the code that calculates the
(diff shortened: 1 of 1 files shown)


skyace65:
I don't think this is necessary, there's no other comments in the script that would confuse users. You can only uncomment what was just commented out.

LTangaF:
I see your point, but it happened to me, which is why I suggested this edit. When I'm learning from a tutorial, I'm not always a careful reader and I ran into this behavior and couldn't quite understand why right away because the code blocks themselves didn't show the code being uncommented. I know it's shown in total down below, but I didn't go there right away.

If you still want to reject it, I'll understand.

PR #9893, opened 2024-09-03
Add info to turn off preview sunlight before working with 3D shader lighting.

FarzadBT:
By default in Godot 4.3, preview sunlight is enabled and only disabled by a DirectionalLight3D being added to the scene. If you follow the guide and add an OmniLight3D, the preview sunlight stays enabled and causes a drastically different look compared to the images in the guide.

With preview sunlight (default).
image

Without preview sunlight.
image

Optionally also add an image to point out the toggle button;
image


AThousandShips:

You should target `master` not `4.3`, changes can then be cherry picked, see [here](https://docs.godotengine.org/en/latest/contributing/documentation/contributing_to_the_documentation.html#contributing-changes).

Please open a new PR and a new branch based on the correct branch, if needed (check the latest version first).

PR #9903, opened 2024-09-04
Explain that `@export` loads `Resource`s when script instances are loaded

AdriaandeJongh:
I somehow couldn't find this anywhere in the docs even though this is very important and relevant info! 

~~I also removed a very vague and illogical sentence that would have been just above or below my addition.~~

All feedback is welcome.

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_exports.rst
+++ b/tutorials/scripting/gdscript/gdscript_exports.rst
@@ -302,6 +302,17 @@ Therefore, if you specify an extension of Resource such as:
 The drop-down menu will be limited to AnimationNode and all
 its derived classes.
 
+.. note::
+
+    Using ``@export`` variables for :ref:`Resource <class_Resource>` objects
+    makes them a dependency of the instance, meaning that all the resources
+    referenced by ``@export`` variables are loaded when the scene
+    containing the script is loaded. If you want to reference a
+    :ref:`Resource <class_Resource>` object but load it manually when you need
+    it (which, for example, is often the case for
+    :ref:`PackedScenes <class_PackedScene>` containing a whole level), use
+    ``@export_file`` or ``@export_file_path`` instead.
+
 .. _doc_gdscript_exports_exporting_bit_flags:
 
 Exporting bit flags


tutorials/scripting/gdscript/gdscript_exports.rst
```diff
@@ -244,9 +244,15 @@ Therefore, if you specify an extension of Resource such as:
...
-in conjunction with a :ref:`script in "tool" mode <doc_gdscript_tool_mode>`.
+One consideration in using ``@export`` variables for classes that 
+derive from :ref:`Resource <class_Resource>` is that it loads the
+resource when the instance of the script is loaded. This is often
+desirable as you'll likely want to use that resource directly in
+your script. However, if the resource in question is a large
+resource commonly loaded during a game's loading screen (like a
+:ref:`PackedScene <class_PackedScene>` containing a whole level),
+it may be desirable to use ``@export_file`` so you have the file
+path to the resource file and can load it manually when you need it.
```
HolonProduction, line 255: I think it is too easy to understand this as "if your resource is big, use export_file" which I don't really agree with since in most cases the big resource is embedded into a big scene which gets loaded async anyways.

I think this should be a note to make it less conclusive and also an example of a case where it would make sense and were it wouldn't might help to convey that this requires a case by case decision. Examples that come to mind:
- use export_file : a level manager scene which holds the player and instantiates and frees levels as the player transitions through them
- don't use export_file : a big level scene which instantiates some subscenes using a script (for whatever reason :sweat_smile:)

HolonProduction (commented):
From my understanding this makes the resource a dependency of the scene in which it is assigned to an export property, which means it will get loaded together with that scene.

The formulation from this PR seems a bit abstract to me, if this is what we want to convey. Or am I missing a second case which has nothing to do with scenes?

tutorials/scripting/gdscript/gdscript_exports.rst
```diff
@@ -244,9 +244,15 @@ Therefore, if you specify an extension of Resource such as:
 The drop-down menu will be limited to AnimationNode and all
 its derived classes.
 
-It must be noted that even if the script is not being run while in the
-editor, the exported properties are still editable. This can be used
-in conjunction with a :ref:`script in "tool" mode <doc_gdscript_tool_mode>`.
```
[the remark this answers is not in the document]
AdriaandeJongh: What you suggest it says perfectly contradicts the first sentence of what I removed, so I am honestly confused. The first line seems to suggest that property setters are executed when NOT in tool mode.

I'll leave this line as-is, but it is honestly so vague and confusing that it should be removed or replaced by something else.

tutorials/scripting/gdscript/gdscript_exports.rst
```diff
@@ -244,9 +244,15 @@ Therefore, if you specify an extension of Resource such as:
...
+One consideration in using ``@export`` variables for classes that 
+derive from :ref:`Resource <class_Resource>` is that it loads the
+resource when the instance of the script is loaded. This is often
```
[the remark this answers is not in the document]
AdriaandeJongh: Right! Will do.

AdriaandeJongh:
@HolonProduction iterated on the wording based on your comments and made the whole thing a note. Also reverted removing the extremely vague line so that you or someone else can decide what to do with it. My opinion remains that it should be removed.

tutorials/scripting/gdscript/gdscript_exports.rst
```diff
@@ -298,7 +248,16 @@ It must be noted that even if the script is not being run while in the
 editor, the exported properties are still editable. This can be used
 in conjunction with a :ref:`script in "tool" mode <doc_gdscript_tool_mode>`.
 
-.. _doc_gdscript_exports_exporting_bit_flags:
+.. note::
+
+    Using ``@export`` variables for :ref:`Resource <class_Resource>` objects
+    makes them a dependency of the script, meaning that all the resources
+    referenced by ``@export`` variables are loaded when the scene
+    containing the script is instantiated. If you want to reference a
```
HolonProduction:
```suggestion
    containing the script is loaded. If you want to reference a
```

Based on my more recent testing on cyclic dependencies.

HolonProduction (changes requested):
So yeah this, slipped though my mailbox. Sorry!
Overall it is nice to document this. Just needs an adjustment regarding this specific piece of behaviour:

tutorials/scripting/gdscript/gdscript_exports.rst
```diff
@@ -304,6 +304,17 @@ its derived classes.
...
+.. note::
+
+    Using ``@export`` variables for :ref:`Resource <class_Resource>` objects
+    makes them a dependency of the script, meaning that all the resources
+    referenced by ``@export`` variables are loaded when the scene
+    containing the script is loaded. If you want to reference a
+    :ref:`Resource <class_Resource>` object but load it manually when you need
+    it (which, for example, is often the case for
+    :ref:`PackedScenes <class_PackedScene>` containing a whole level), use
+    ``@export_file`` instead.
```
dalexeev:
```suggestion
    ``@export_file`` or ``@export_file_path`` instead.
```

Note that `@export_file` and `@export_file_path` only allow you to specify a filename mask, while `@export` filters resources by type.

tutorials/scripting/gdscript/gdscript_exports.rst
```diff
@@ -304,6 +304,17 @@ its derived classes.
 
 .. _doc_gdscript_exports_exporting_bit_flags:
 
+.. note::
+
+    Using ``@export`` variables for :ref:`Resource <class_Resource>` objects
+    makes them a dependency of the script, meaning that all the resources
```
HolonProduction:
```suggestion
    makes them a dependency of the instance, meaning that all the resources
```
Just a nitpick I guess. But the script does not depend on them, only the specific instance of the script does.

HolonProduction (approved):
LGTM (based on content)

tutorials/scripting/gdscript/gdscript_exports.rst
```diff
@@ -304,6 +304,17 @@ its derived classes.
 
 .. _doc_gdscript_exports_exporting_bit_flags:
```
HolonProduction: This will need to go below your note. It's a label for the next heading. I think that's whats failing in CI.

AdriaandeJongh: Ah whoops, yeah, silly of me.

skyace65 (approved):
Grammar and spelling wise it looks good to me.

skyace65:
Thanks!

mhilbrunner:

PR #9908, opened 2024-09-05
Document file path String methods in File paths in Godot projects

Calinou:
- See https://github.com/godotengine/godot-docs-user-notes/discussions/106#discussioncomment-10525790.

### DIFF
--- a/tutorials/io/data_paths.rst
+++ b/tutorials/io/data_paths.rst
@@ -25,6 +25,23 @@ This makes it possible to work with paths returned by other Windows
 applications. We still recommend using only forward slashes in your own code to
 guarantee that everything will work as intended.
 
+.. tip::
+
+    The String class offers over a dozen methods to work with strings that represent file paths:
+
+    - :ref:`String.filecasecmp_to() <class_String_method_filecasecmp_to>`
+    - :ref:`String.filenocasecmp_to() <class_String_method_filenocasecmp_to>`
+    - :ref:`String.get_base_dir() <class_String_method_get_base_dir>`
+    - :ref:`String.get_basename() <class_String_method_get_basename>`
+    - :ref:`String.get_extension() <class_String_method_get_extension>`
+    - :ref:`String.get_file() <class_String_method_get_file>`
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):
Mild approval, because I wonder if just that sentence (and maybe one example) is enough, with a link to the class ref.

I'm not a big fan of reproducing the class ref like this, because it tends to produce lots of additional effort with the list getting outdated, people creating issues and PRs, us reviewing + merging + cherrypicking etc. If we can, we should have information once and link to it, instead of duplicating. That said, it can be worth it if its especially important, I'm just not sure thats the case here?

skyace65 (commented):
I agree with mhilbrunner that this isn't ideal. I think it would be better to just say something like "The string class offers several methods to work with strings that represent file paths, you can find a list of it's class reference page here" and link to the class reference page.

Calinou:
> I think it would be better to just say something like "The string class offers several methods to work with strings that represent file paths, you can find a list of it's class reference page here" and link to the class reference page.

The String class has a lot of methods unrelated to path manipulation, which is why I opted to explicitly list all relevant methods.

mhilbrunner:
@Calinou That makes sense. Do we maybe want to maintain the list in the class reference for `String` then, and link to it here? That at least increases the chance that we remember to update it when changing `String`.

tetrapod00:
As a *user* of the engine, a list in the File Paths page is clearly the right place for this information. Whether the maintenance burden is worth it is up to you.

Putting the list in the class reference page (presumably in the intro section?) and linking also seems fine, but is there any real precedent for listing a subset of methods like that in the class reference? It means that there is a redundant list of the same methods twice (once as a subset, and once in the main methods list), and it clutters the class reference for String if you're looking at it for other reasons.

skyace65 (approved):
Given the discussion I'm in favor of merging this as is. I'd also recommend that instead of copying this list in the class reference we should link to this page in the "tutorials" section of the string class.

mhilbrunner (approved):

mhilbrunner:
Merging as per the above. Thank you!

mhilbrunner:

PR #9929, opened 2024-09-09
Update release support timelines: 4.4 ETA, 3.6 is released

akien-mga:
For the 3.5, 4.0 and 4.1 docs branches, we should likely nuke the contents of that page and replace it with a short paragraph that redirects users to the `latest` branch.

And later on we should remove this from the docs and move it to the website, so we don't keep maintaining multiple versions of our release policy and timelines each time we branch off.


clayjohn (approved):

mhilbrunner (approved):

mhilbrunner:
Thanks!

mhilbrunner:

PR #9940, opened 2024-09-10
Add 3->4 upgrading information about NDC, light(), and reverse-z

tetrapod00:
Resolves https://github.com/godotengine/godot/issues/96704 by changing the docs.

- Adds information about normalized device coordinates changing from 3.x/Compatibility to Forward+/Mobile.
- Adds a note that shaders with custom lighting will need changes.
- Links to reverse-z changes. As more and more 4.x versions are released, reverse-z will become the default, and most users finding this page will need to apply reverse-z changes if their shader needs it. We should not expect users to visit each minor version conversion page in sequence in order to find this information.

---

Summary of the original issue: a simplified version of the shader in https://github.com/paddy-exe/Godot-RealTimeCaustics was converted automatically from 3 to 4, then all the manual fixes mentioned [here](https://docs.godotengine.org/en/stable/tutorials/migrating/upgrading_to_godot_4.html#updating-shaders) were applied. But the shader still has no visible output.
Two additional changes had to be made so the shader works in 4.3 Forward+:
- Convert the NDC reconstruction from ` vec3 ndc = vec3(SCREEN_UV, depth) * 2.0 - 1.0;` to ` vec3 ndc = vec3(SCREEN_UV* 2.0 - 1.0, depth);`. 
- Change ALBEDO = vec3(0.0); to ALBEDO = vec3(1.0);. The shader is doing additive lighting using the `light()` function, and requires a change to ALBEDO to look similar to 3.x.

These kinds of changes cannot be supported by an automatic conversion tool, and they can't even be exhaustively listed. We can list common cases, and clarify the warning that advanced shaders will need more manual work.

### DIFF
--- a/tutorials/migrating/upgrading_to_godot_4.rst
+++ b/tutorials/migrating/upgrading_to_godot_4.rst
@@ -509,13 +509,14 @@ environment effect and its visual knobs remain within the Environment resource.
 Updating shaders
 ^^^^^^^^^^^^^^^^
 
-There have been some changes to shaders that aren't covered by the upgrade tool.
+There have been some changes to shaders that aren't covered by the upgrade tool. 
+You will need to make some manual changes, especially if your shader is advanced.
 
 The ``.shader`` file extension is no longer supported, which means you must
 rename ``.shader`` files to ``.gdshader`` and update references accordingly in
 scene/resource files using an external text editor.
 
-Some notable renames you will need to perform in shaders are:
+Some notable changes you will need to perform in shaders are:
 
 - Texture filter and repeat modes are now set on individual uniforms, rather
   than the texture files themselves.
@@ -524,6 +525,15 @@ Some notable renames you will need to perform in shaders are:
 - :ref:`Built in matrix variables were renamed. <doc_spatial_shader>`
 - Particles shaders no longer use the ``vertex()`` processor function. Instead
   they use ``start()`` and ``process()``.
+- In Forward+ and Mobile, normalized device coordinates now have a Z-range of [0,1]
+  instead of [-1,1]. When reconstructing NDC from ``SCREEN_UV`` and depth, use 
+  ``vec3 ndc = vec3(SCREEN_UV * 2.0 - 1.0, depth);`` instead of 
+  ``vec3 ndc = vec3(SCREEN_UV, depth) * 2.0 - 1.0;``. Compatibility is unchanged.
+- The lighting model changed. If your shader has a custom ``light()`` function,
+  you may need to make changes to get the same visual result.
+- In 4.3 and up, the reverse Z depth buffer technique is now implemented, which 
+  may break advanced shaders. See 
+  `Introducing Reverse Z (AKA I'm sorry for breaking your shader) <https://godotengine.org/article/introducing-reverse-z/>`__.
 
 See :ref:`doc_shading_language` for more information.
 


tutorials/migrating/upgrading_to_godot_4.rst
```diff
@@ -509,13 +509,14 @@ environment effect and its visual knobs remain within the Environment resource.
 Updating shaders
 ^^^^^^^^^^^^^^^^
 
-There have been some changes to shaders that aren't covered by the upgrade tool.
+There have been some changes to shaders that aren't covered by the upgrade tool. 
+You will need to make some manual changes, especially if your shader is advanced.
```
Flarkk, line 513: ... especially if your shader _performs advanced operations_ ?

tutorials/migrating/upgrading_to_godot_4.rst
```diff
@@ -524,6 +525,15 @@ Some notable renames you will need to perform in shaders are:
 - :ref:`Built in matrix variables were renamed. <doc_spatial_shader>`
 - Particles shaders no longer use the ``vertex()`` processor function. Instead
   they use ``start()`` and ``process()``.
+- In Forward+ and Mobile, normalized device coordinates now have a Z-range of [0,1]
+  instead of [-1,1]. When reconstructing NDC from ``SCREEN_UV`` and depth, use 
+  ``vec3 ndc = vec3(SCREEN_UV * 2.0 - 1.0, depth);`` instead of 
+  ``vec3 ndc = vec3(SCREEN_UV, depth) * 2.0 - 1.0;``. Compatibility is unchanged.
```
Flarkk, line 531: Compatibility _renderer_ is unchanged ?

tutorials/migrating/upgrading_to_godot_4.rst
```diff
@@ -524,6 +525,15 @@ Some notable renames you will need to perform in shaders are:
 - :ref:`Built in matrix variables were renamed. <doc_spatial_shader>`
 - Particles shaders no longer use the ``vertex()`` processor function. Instead
   they use ``start()`` and ``process()``.
+- In Forward+ and Mobile, normalized device coordinates now have a Z-range of [0,1]
```
Flarkk, line 528: In Forward+ and Mobile _renderers_ ... ?

tutorials/migrating/upgrading_to_godot_4.rst
tetrapod00: "Advanced operations" isn't *quite* right, either. What I want here is a disclaimer that you still need to make manual changes, and that the list here can't catch literally every change.

Things that should be covered by "advanced" or its replacement, without explicitly listed them out:
- complex `light()` functions that make assumptions about defaults or the standard shader lighting function
- assumptions about the specifics of coordinate spaces, like the NDC change or the internals of the transform matrixes. 
- using clip space directly (all the cases in the reverse-z blogpost that don't have example snippets)

Just using a transform matrix to convert coordinates is not an "advanced operation", but taking the matrix apart and manually reassembling it for some other purpose is.

I should probably just add an explicit note that this list is *not* exhaustive, and your shader may not work even if all the changes are made.

clayjohn: I think you can simplify and say something like _especially if your shader uses a custom `light()` function or includes coordinate space transformations._

Advanced users will be flagged about the risks, while beginner users will know that those are advanced things that they are not using.

Flarkk: Yeah I get it. I guess there is no sweet spot as it's hard to qualify what may require some rewrite and what may not.
It's just that "advanced shader" doesn't speak much to me (the shader itself is not advanced, but more the actions it performs, or the use you make of the shading functionalities). But I'm not a native speaker so others may feel differently.

tetrapod00: I ended up using the specific cases that clayjohn mentioned, rather than trying to generalize too much.

I also added another paragraph at the end which solves what I saw as the actual problem.

tutorials/migrating/upgrading_to_godot_4.rst
```diff
@@ -524,9 +526,23 @@ Some notable renames you will need to perform in shaders are:
...
+- In 4.3 and up, the reverse Z depth buffer technique is now implemented, which 
+  may break advanced shaders. See 
+  `Introducing Reverse Z (AKA I'm sorry for breaking your shader) <https://godotengine.org/article/introducing-reverse-z/>`__.
 
 See :ref:`doc_shading_language` for more information.
 
+This list is not exhaustive. If you made all the changes mentioned here and your 
+shader still doesn't work, try asking for help in one of the community channels,
+or open an issue on GitHub.
+
```
tetrapod00, line 545: This might be asking for more support issues on github, but it's also how we found out about this issue...

clayjohn: I wouldn't tell people to open a bug report. We don't offer support through Github and asking for help porting from 3.x to 4.x is definitely something that should not be on Github.

tutorials/migrating/upgrading_to_godot_4.rst
```diff
@@ -524,9 +526,23 @@ Some notable renames you will need to perform in shaders are:
...
+  you may need to make changes to get the same visual result.
+- In 4.3 and up, the reverse Z depth buffer technique is now implemented, which 
+  may break advanced shaders. See 
+  `Introducing Reverse Z (AKA I'm sorry for breaking your shader) <https://godotengine.org/article/introducing-reverse-z/>`__.
 
 See :ref:`doc_shading_language` for more information.
 
+This list is not exhaustive. If you made all the changes mentioned here and your 
+shader still doesn't work, try asking for help in one of the community channels,
+or open an issue on GitHub.
```
clayjohn, line 544:
```suggestion
This list is not exhaustive. If you made all the changes mentioned here and your 
shader still doesn't work, try asking for help in one of the `community channels <https://godotengine.org/community/>`__.
```

tetrapod00:
I think that's everything. Thank you for the reviews!

clayjohn (approved):
Thank you!

mhilbrunner (approved):

mhilbrunner:
@tetrapod00 Thank you! Merged. This should apply to all Godot version going back to 4.0, right?

PR #9972, opened 2024-09-17
Modified PCKPacker Description for Clarity

Rynzier:
Changed pck_start() description so functionality was more clear. Replaced "with the name" with "at the file path," since the pck_name argument actually represents the path that the .pck file will be written to (including the name), and not just the name of the file.


AThousandShips:

PR #9977, opened 2024-09-19
Add hint that navmesh geometry nodes can be parsed by groupname

smix8:
Adds hint that navmesh geometry nodes can be parsed by groupname.

This is just missed way too often by new users that struggle around how they can get their nodes below the NavigationRegion node while it is as simple as just switching the parse settings on the NavigationPolygon/NavigationMesh resource.


mhilbrunner (approved):

mhilbrunner:
Thanks!

mhilbrunner:

PR #9980, opened 2024-09-20
Clearify behaviour of `@export` in combination with setters

HolonProduction:
This piece of text came up in #9903. When it was added it wasn't placed under export resources, so it makes sense to move it away from there. Also I changed the wording to something easier to understand (IMO), behavior was tested on 4.3.stable.


tutorials/scripting/gdscript/gdscript_exports.rst
```diff
@@ -52,6 +52,10 @@ Resources and nodes can be exported.
     @export var resource: Resource
     @export var node: Node
 
+Even if a script is not executed in the editor, exported properties
+can still be edited. However getters and setters will only be used
```
AdriaandeJongh, line 56: A comma after `However`? -> `However, getters and setters will only be used`

skyace65 (approved):

skyace65:
Thanks!

PR #9984, opened 2024-09-21
Rename shading language to GDShader

tetrapod00:
Closes https://github.com/godotengine/godot-proposals/issues/4350.
Closes https://github.com/godotengine/godot-proposals/issues/10784.

Companion PR to https://github.com/godotengine/godot/pull/97298.

The places that definitely need to be changed are [Shading Language/Introduction](https://docs.godotengine.org/en/stable/tutorials/shaders/shader_reference/shading_language.html#introduction), the [Shader class reference](https://docs.godotengine.org/en/stable/classes/class_shader.html), and the [GLSL Migration Guide](https://docs.godotengine.org/en/stable/tutorials/shaders/converting_glsl_to_godot_shaders.html#doc-converting-glsl-to-godot-shaders). In places like compute shaders where godot *does* use GLSL, we should make the distinction that "GLSL is used, not GDShader"

Unlike the class ref, for the manual we don't need to replace every reference to the shading language with "GDShader" immediately, to avoid breaking translations.

_Bugsquad edit: Waiting on main repo PR https://github.com/godotengine/godot/pull/97298_

### DIFF
--- a/contributing/development/core_and_modules/internal_rendering_architecture.rst
+++ b/contributing/development/core_and_modules/internal_rendering_architecture.rst
@@ -307,9 +307,8 @@ recompile the editor or export template binary.
 
 Some material features such as height mapping, refraction and proximity fade are
 not part of core shaders, and are performed in the default BaseMaterial3D using
-the Godot shader language instead (not GLSL). This is done by procedurally
-generating the required shader code depending on the features enabled in the
-material.
+GDShader instead (not GLSL). This is done by procedurally generating the required
+shader code depending on the features enabled in the material.
 
 By convention, shader files with ``_inc`` in their name are included in other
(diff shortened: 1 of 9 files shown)


tetrapod00:
This is ready for review. I looked this over again, and couldn't find any more references to "Godot Shader Language" or similar that need to be changed.

mhilbrunner:
Let's wait for the main repo PR https://github.com/godotengine/godot/pull/97298 to get approved/closed, as this decision should be made there IMO. :) Also, this will need a rebase, but I'd wait for a decision first.

tetrapod00:
This specific set of terminology was turned down in https://github.com/godotengine/godot/pull/97298#issuecomment-2466837691. I'm going to close this PR, and open a separate one once some set of terminology is finalized in the main repo issue.

Issue #9133, opened 2024-03-24
Improve discoverability of Engine core and modules pages


allenwp:
**Your Godot version:**
Godot 4

**Issue description:**
The [Engine core and modules](https://docs.godotengine.org/en/latest/contributing/development/core_and_modules/index.html#getting-started-with-godot-s-source-code) section of the docs provides information to experienced programmers who are new to Godot.

Personally, I have had a hard time finding and referring back to these pages because they are nested so deep that the pages do not appear on the sidebar. My instinct when looking for these pages is to look for items on the sidebar titled “Core types”, “Variant class”, or “Object class” under the engine development section. These pages do exist, but are nested deep enough that they do not appear on the sidebar.

My proposal is to remove the current intermediate [Engine core and modules](https://docs.godotengine.org/en/latest/contributing/development/core_and_modules/index.html#getting-started-with-godot-s-source-code) landing page and replace it with its two sub-pages. The new structure would look something like this:

- Engine development
    - Engine architecture
        - Engine core (and modules?)
            - Godot's architecture diagram
            - Common engine methods and
            - Core types
            - Variant class
            - Object class
            - Inheritance class tree
            - Internal rendering architecture
            - 2D coordinate systems and 2D 
        - Extending Godot by modifying its source code
            - Custom modules in C++
            - Binding to external libraries
            - Custom Godot servers
            - Custom resource format loaders
            - Custom AudioStreams
            - Custom platform ports
            - Unit testing

I expect that no content would be lost with this new structure and promoting these leaf-level pages to the sidebar would make them more discoverable and faster to refer back to.

You can see the history of these menu items being buried across docs version 3.6, 4.0, and 4.1:
![image](https://github.com/godotengine/godot-docs/assets/17506573/5b19d4a7-a24d-47d5-9e48-443ebef9d7bc)

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/latest/contributing/development/core_and_modules/index.html

[Edit: Originally I referenced #5877 as an example of users having a hard time finding this documentation, but this issue is from before these menu items became hidden.]

Issue #8567, opened 2023-12-04
Deeply nested pages have unexpected breadcrumbs


31:
## Your Godot version:
4.2 / stable

## Issue description:

https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_signals.html shows this incorrect breadcrumb at the top, making it hard to get back to the list of deeply nested articles:

![image](https://github.com/godotengine/godot-docs/assets/331300/7ef4d3f4-0a40-47c4-83d2-1ce84d60b8c8)

I would expect it to say `Scripting` / `C#/.NET` / `Godot API for C#` / `C# signals`, and link me to the `Godot API for C#` index.

---

The reason I expect that is that `Godot API for C#` is highlighted on the sidebar:

![image](https://github.com/godotengine/godot-docs/assets/331300/6949cfb5-99a8-4850-9913-8090ce467e7a)

For context, I recently saw and started trying to use the newly reorganized pages. So far, it feels unusual and inconsistent. For example: the first few sections under `C#/.NET` are pages, but the last two are header anchor links within the "parent" section. This seems very strange to mix together in that order, especially without any visual distinction.

So, my expectation about the breadcrumb may have more to do with somehow clearing up the large-scale confusion rather than fixing a specific small thing. 😄

31:
I might have misinterpreted how the sidebar is working, it's very confusing. 😕 

The first few sections under `C#/.NET` are pages, but the last two are header anchor links. They way they're mixed together without any kind of marking makes it very hard to follow.

Issue #8566, opened 2023-12-04
New deeply nested sections cause sidebar to not start at the right place


31:
## Your Godot version:
4.2 / stable

Chrome Version 119.0.6045.200 (Official Build) (64-bit)

## Issue description:

When visiting https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/index.html, the sidebar scrolls to the open section.

![image](https://github.com/godotengine/godot-docs/assets/331300/ed0b4518-6562-47a4-a70d-d58ba2422b06)

When visiting https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_differences.html, which lacks an entry in the sidebar, the sidebar starts at the top.

![image](https://github.com/godotengine/godot-docs/assets/331300/2bf4e21d-5d84-44a3-a2fb-e830ce25b25c)

I'd expect the `Godot API for C#` element to be visible, because that's what highlighted.

Issue #8792, opened 2024-01-18
Unreachable pages via sidebar


Cerno-b:
**Your Godot version:** 4.2

**Issue description:**

Currently there are docs pages that are not reachable via the sidebar due to deep nesting.
The only way to access them is via link from another page or by reading the docs sequentially using the "next page" button.

I think the manual should be structured so that each rst file should be accessible via the sidebar.
This might require some reordering of pages in order to reduce nesting depth.

The maximum nesting level for sidebar visibility is 5.

For example, this page with nesting level 6 cannot be directly accessed via the sidebar:
https://docs.godotengine.org/en/stable/contributing/development/compiling/compiling_for_android.html

![image](https://github.com/godotengine/godot-docs/assets/16470072/2f2393d2-cf9c-456f-9dea-715f28e12412)

**URL to the documentation page:**

Here are the inaccessible level 6 pages:

https://docs.godotengine.org/en/stable/contributing/development/compiling/compiling_for_android.html
https://docs.godotengine.org/en/stable/contributing/development/compiling/compiling_for_ios.html
https://docs.godotengine.org/en/stable/contributing/development/compiling/compiling_for_linuxbsd.html
https://docs.godotengine.org/en/stable/contributing/development/compiling/compiling_for_macos.html
https://docs.godotengine.org/en/stable/contributing/development/compiling/compiling_for_web.html
https://docs.godotengine.org/en/stable/contributing/development/compiling/compiling_for_windows.html
https://docs.godotengine.org/en/stable/contributing/development/compiling/compiling_with_dotnet.html
https://docs.godotengine.org/en/stable/contributing/development/compiling/compiling_with_script_encryption_key.html
https://docs.godotengine.org/en/stable/contributing/development/compiling/cross-compiling_for_ios_on_linux.html
https://docs.godotengine.org/en/stable/contributing/development/compiling/getting_source.html
https://docs.godotengine.org/en/stable/contributing/development/compiling/introduction_to_the_buildsystem.html
https://docs.godotengine.org/en/stable/contributing/development/compiling/optimizing_for_size.html

https://docs.godotengine.org/en/stable/contributing/development/core_and_modules/2d_coordinate_systems.html
https://docs.godotengine.org/en/stable/contributing/development/core_and_modules/binding_to_external_libraries.html
https://docs.godotengine.org/en/stable/contributing/development/core_and_modules/common_engine_methods_and_macros.html
[...]

PR #9989, opened 2024-09-23
Fix unreachable pages in sidebar by using fake headers in index pages

tetrapod00:
Fixes some cases of https://github.com/godotengine/godot-docs/issues/8792
May mitigate or fix https://github.com/godotengine/godot-docs/issues/8567
May mitigate or fix https://github.com/godotengine/godot-docs/issues/8566
Fixes https://github.com/godotengine/godot-docs/issues/9133

Ensures that most pages are reachable through the sidebar, by replacing real semantic headers with fake headers on index pages. As a side effect, this also means that some pages now have navigable section headers again.

---

There are many pages which do not show up in the sidebar (listed in https://github.com/godotengine/godot-docs/issues/8792). It's especially bad on the scripting pages, which are used often by end users. The contributing pages are a bit less of a problem, because they are referenced less often and only by contributors.

Any header will show up in the navigation sidebar. Currently, there is a maximum depth of 5 in the navigation sidebar. Currently, there are several index pages which serve as empty "table of contents" pages, but which use headers.

The solution I arrived at is to replace the semantic headers in these pages with a fake header, using the `.. rubric::` syntax. I used [this hack](https://stackoverflow.com/a/23544142). I am not a sphinx, RTD, or reStructuredText expert. I hacked this together because using the docs has become actively annoying. If there's a less hacky way to solve this, I'd love to hear it.

If using `rubric` is the right solution, I think the current rubric styling looks *okay*, but it's also possible to [make custom styles to match the header styles.](https://stackoverflow.com/a/23549957)

**Visual summary of changes:**
| Page | Master | This PR |
|-|-|-|
| GDScript exported properties | gdscript exported master | firefox_XVs3eaDzdk |
| C# Signals | c# signals master | firefox_k2JFaZhM4B |
| Building from Source | firefox_MM6ydYbh10 | building source pr |
| Engine core and modules | firefox_ROdbWjTOT5 | engine core pr |

I don't know the technical reason for the current nesting limit of 5. *Even if* that limit was increased, we would not be free to use as many headers as we wanted - this problem of deep nesting would show up again sooner or later. Note that even with this PR, some pages get to have section headers in the sidebar, and some only have a title.

### DIFF
--- a/contributing/development/compiling/index.rst
+++ b/contributing/development/compiling/index.rst
@@ -18,8 +18,7 @@ stripped of extra modules, or an executable targeting engine development.
 The articles below should help you navigate configuration options available, as well as
 prerequisites required to compile Godot exactly the way you need.
 
-Basics of building Godot
-------------------------
+.. rubric:: Basics of building Godot
 
 Let's start with basics, and learn how to get Godot's source code, and then which options
 to use to compile it regardless of your target platform.
@@ -31,8 +30,7 @@ to use to compile it regardless of your target platform.
    getting_source
    introduction_to_the_buildsystem
 
-Building for target platforms
------------------------------
+.. rubric:: Building for target platforms
 
 Below you can find instructions for compiling the engine for your specific target platform.
 Note that Godot supports cross-compilation, which means you can compile it for a target platform
@@ -51,8 +49,7 @@ will try their best to cover all possible situations.
    cross-compiling_for_ios_on_linux
    compiling_for_web
 
-Other compilation targets and options
--------------------------------------
+.. rubric:: Other compilation targets and options
 
 Some additional universal compilation options require further setup. Namely, while Godot
 does have C#/.NET support as a part of its main codebase, it does not get compiled by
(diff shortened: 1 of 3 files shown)


skyace65 (approved):
Built the docs with this and everything works!

tetrapod00:
1. Should the custom styles for `rubric` be added to match existing header styles?
2. There is one more large set of hidden pages, the children of https://docs.godotengine.org/en/stable/tutorials/3d/particles/index.html. I didn't fix those yet because the index page also has *significant* actual info. Should I try to fix that in this PR too?

mhilbrunner (approved):

mhilbrunner:
Better to keep PRs smaller and easy to digest :) Those sound like good followup PRs. Thank you, merged! And thanks Sky for testing.

tetrapod00:
One side effect to note is that fake headers such as `Programming Languages` no longer have a copyable permalink:
firefox_wCcl8ykuHw

But old links like this one still redirect to the page itself: https://docs.godotengine.org/en/latest/tutorials/scripting/index.html#programming-languages

Even if the permalinks can't be added back in, I think this is a fine tradeoff for improved usability.

mhilbrunner:

PR #9995, opened 2024-09-24
Document bitmask C# API difference

LemmaEOF:
I spent a good couple hours today beating my head against an inscrutable C# compile error tonight that ended up being because of some methods that use bitmasks as an argument wanting a `uint` in C# instead of an `int` in the type signature, something that doesn't appear on GDScript-focused documentation as a difference between signed and unsigned ints doesn't exist in GDScript. Due to C# type coercion in most cases this is niche enough that not many people will run into it, but I want to prevent other people running into this wall as well, so I've put a quick note of bitmasks using `uint` for method overrides in the C# API differences doc.

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_differences.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_differences.rst
@@ -805,7 +805,7 @@ GDScript              C#
 
 Bitmasks
 -------
-Some methods, such as :ref:`ResourceFormatSaver _Save <class-resourceformatsaver-private-method-save>`
+Some methods, such as :ref:`ResourceFormatSaver._Save <class-resourceformatsaver-private-method-save>`,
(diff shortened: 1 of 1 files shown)


paulloz:
Hello, and thank you for your contribution.
This is not something specific to bitmasks, it is a more general thing about the GDScript types not always aligning with C# types (e.g. `int` in the documentation might mean `long` in C#, or `float` might mean `double`). We are currently working on a dedicated C# documentation. That will IMHO solve this issue, since it'll always refer to the actual C# API types. In the meantime, it's probably fine to add a note somewhere to explain this, but I'm not sure that we need to focus on bitmasks. I'm also not a big fan of wording errors as being "confusing" in our own documentation.

LemmaEOF:
Makes sense to me! The C# docs will definitely solve the issue. I can look into adding a note directly on things that use `uint` possibly, but that's a matter for another PR.

PR #10007, opened 2024-09-26
Add tutorial for reducing pipeline compilation stutters.

DarioSamo:
Tutorial for the new functionality added by https://github.com/godotengine/godot/pull/90400.

Doesn't need to be merged until the PR is merged.

The tutorial is aimed at users familiar with the engine but it should not require them to be rendering experts. Suggestions on what content to include and adjusting the grammar are welcome.


tutorials/performance/pipeline_compilations.rst
```diff
@@ -0,0 +1,77 @@
...
+   Pipeline compilations of one of the demo projects.
+
+.. note:: We can see the pipelines compiled during gameplay and
+          verify which steps could possibly cause stuttters. Note
+          that these values will only increase and never go down,
+          as deleted pipelines are not tracked by these monitors
+          and pipelines may be erased and recreated during gameplay.
+
+- **Canvas**: Compiled when drawing a 2D node. The engine does not currently feature pre-compilation for 2D elements and the stutters will show up when the node is drawn for the first time.
+- **Mesh**: Compiled as part of loading a 3D mesh and identifying what pipelines can be pre-compiled from its properties. These can lead to stutters if a mesh is loaded from the game loop, but they can be mitigated if the mesh is loaded by using a background thread. **Modifiers that are part of nodes such as material overrides can't be compiled on this step**.
```
clayjohn, line 35:
```suggestion
- **Mesh**: Compiled as part of loading a 3D mesh and identifying what pipelines can be pre-compiled from its properties. These can lead to stutters if a mesh is loaded during gameplay, but they can be mitigated if the mesh is loaded by using a background thread. **Modifiers that are part of nodes such as material overrides can't be compiled on this step**. 
```

tutorials/performance/pipeline_compilations.rst
```diff
@@ -0,0 +1,77 @@
...
+          that these values will only increase and never go down,
+          as deleted pipelines are not tracked by these monitors
+          and pipelines may be erased and recreated during gameplay.
+
+- **Canvas**: Compiled when drawing a 2D node. The engine does not currently feature pre-compilation for 2D elements and the stutters will show up when the node is drawn for the first time.
+- **Mesh**: Compiled as part of loading a 3D mesh and identifying what pipelines can be pre-compiled from its properties. These can lead to stutters if a mesh is loaded from the game loop, but they can be mitigated if the mesh is loaded by using a background thread. **Modifiers that are part of nodes such as material overrides can't be compiled on this step**. 
+- **Surface**: Compiled when a frame is about to be drawn and 3D objects were instanced on the scene tree for the first time. This can also include compilation for nodes that aren't even visible on the scene tree. The stutter will occur only on the first frame the node is added to the scene, which won't result in an obvious stutter if it happens right after a loading screen.
```
clayjohn, line 36:
```suggestion
- **Surface**: Compiled when a frame is about to be drawn and 3D objects were instanced on the scene tree for the first time. This can also include compilation for nodes that aren't even visible on the scene tree. The stutter will occur only on the first frame the node is added to the scene, which won't result in an obvious stutter if it happens right after a loading screen. Commonly, compilations happen here as a result of using material overrides.
```

tutorials/performance/pipeline_compilations.rst
```diff
@@ -0,0 +1,77 @@
...
+- **Draw**: Compiled on demand when a 3D object needs to be drawn and an ubershader was not pre-compiled ahead of time. The engine is unable to pre-compile this pipeline due to triggering a case that hasn't been covered yet or a modification that was done to the engine's code. Leads to stutters during gameplay. This is identical to Godot versions before **4.4**.
```
clayjohn, line 37: If I read this correctly. This should always be 0 for users right?

tutorials/performance/pipeline_compilations.rst
```diff
@@ -0,0 +1,77 @@
...
+- **Specialization**: Compiled in the background during gameplay to optimize the framerate. Unable to cause stutters.
```
clayjohn, line 38:
```suggestion
- **Specialization**: Compiled in the background during gameplay to optimize the framerate. Unable to cause stutters, but may result in reduced framerates if there are many happening per frame.
```

tutorials/performance/pipeline_compilations.rst
```diff
@@ -0,0 +1,77 @@
...
+
+- **MSAA Level**: Enabled when the level of 3D MSAA is changed on the project settings. Unfortunately, different MSAA levels being used on different viewports will lead to stutters as the engine only keeps track of one level at a time to perform pre-compilation.
+- **Reflection Probes**: Enabled when a ReflectionProbe node is placed on the scene.
+- **Separate Specular**: Enabled when using effects like sub-surface scattering or a compositor effect that relies on sampling the specularity directly off the screen.
+- **Motion Vectors**: Enabled when using effects such as TAA, FSR2 or a compositor effect that requires motion vectors (such as motion blur).
+- **Normal and Roughness**: Enabled when using an effect or shader that relies on sampling the normals directly off the screen.
```
clayjohn, line 51:
```suggestion
- **Normal and Roughness**: Enabled when using SDFGI, VoxelGI, Screen Space Reflections, SSAO, SSIL, or using the normal_roughness_buffer in a custom shader or CompositorEffect. 
```

tutorials/performance/pipeline_compilations.rst
```diff
@@ -0,0 +1,77 @@
...
+- **Lightmaps**: Enabled when a LightmapGI node is placed on the scene and a node uses a baked lightmap.
+- **VoxelGI**: Enabled when a VoxelGI node is placed on the scene.
+- **SDFGI**: Enabled when the WorldEnvironment enables SDFGI.
+- **Multiview**: Enabled for XR projects.
+- **16/32-bit Shadows**: Enabled when the configuration of the depth of shadowmaps is changed on the project settings.
+- **Omni Shadow Dual Paraboloid**: Enabled when an omni light casts shadows and uses the dual paraboloid mode.
+- **Omni Shadow Cubemap**: Enabled when an omni light casts shadows and uses the cubemap mode.
+
+If you witness stutters during gameplay and the monitors report a sudden increase in compilations during the **Surface** step, it is very likely a feature was not enabled ahead of time. Making use of this effect in the initial scene that your game loads will likely mitigate the issue.
```
clayjohn, line 60:
```suggestion
If you witness stutters during gameplay and the monitors report a sudden increase in compilations during the **Surface** step, it is very likely a feature was not enabled ahead of time. Ensuring that this effect is enabled while loading your game will likely mitigate the issue.
```

tutorials/performance/pipeline_compilations.rst
```diff
@@ -0,0 +1,77 @@
...
+
+Pipeline Pre-Compilation Instancing
+-----------------------------------
+
+One common source of stutters in games is the fact that some effects are only instanced on the scene because of interactions that only happen during gameplay. This is often the case with particle effects, which are only added to the scene through scripts when the player does an action. Even if the scene is preloaded, the engine might be unable to pre-compile the pipelines until the effect is added to the scene at least once. Luckily, it's possible for **Godot 4.4** to pre-compile these pipelines as long as the scene is instantiated at least once on the scene, even if it's completely invisible or outside of the camera's view.
```
clayjohn, line 65:
```suggestion
One common source of stutters in games is the fact that some effects are only instanced on the scene because of interactions that only happen during gameplay. For example, if you have a particle effect that is only added to the scene through a script when a player does an action. Even if the scene is preloaded, the engine might be unable to pre-compile the pipelines until the effect is added to the scene at least once. Luckily, it's possible for **Godot 4.4** to pre-compile these pipelines as long as the scene is instantiated at least once on the scene, even if it's completely invisible or outside of the camera's view.
```

clayjohn (commented):
Great work! I left a few suggestions to add more detail and improve grammar. This is going to be very helpful for users

tutorials/performance/pipeline_compilations.rst
```diff
@@ -0,0 +1,77 @@
+Reducing Stutter from Shader (Pipeline) Compilations
+====================================================
+
+Pipeline compilation, also commonly known as shader compilation, is an expensive operation required by the engine to be able to draw any kind of content with the GPU. 
+
+In more precise terms, *Shader Compilation* involves the translation of the GLSL code that Godot generates into an intermediate format that can be shared across systems (such as SPIR-V when using Vulkan). However, this format can't be used by the GPU directly. *Pipeline Compilation* is the step the GPU driver performs where it'll convert the intermediate shader format (the result from shader compilation) to something the GPU can actually use for rendering. Drivers usually keep a cache of pipelines stored somewhere in the system to avoid repeating the process every time a game is run. This cache is usually deleted when the driver is updated.
```
tetrapod00, line 6: This section might be made more clear by explicitly listing out the shader compilation stages, ideally with some graphic:
`VisualShader`/`StandardMaterial` (this generates GDShader in some cases, right?) -> `GDShader` (Godot Shading Language) -> `GLSL` -> intermediate format (`SPIR-V`) -> pipeline. 

As is, this assumes some knowledge of the shader compilation process (the fact that Godot compiles to GLSL as an intermediate step isn't commonly referenced throughout the rest of the docs, IIRC).

tetrapod00: Something like this
ShaderCompilationSteps

(I'm not sure if all this is accurate, and labelling just the single step as "shader compilation" may be misleading")

tetrapod00:
We might want to link to this page from https://docs.godotengine.org/en/stable/tutorials/rendering/jitter_stutter.html.

tutorials/performance/pipeline_compilations.rst
```diff
@@ -0,0 +1,77 @@
...
+
+Pipelines contain more information than just the shader code, which makes it difficult for an engine to compile them ahead of time due to the amount of possible combinations. On top of that, this step can only be performed on the user's system and it is very tough to share the result between users unless they have the exact same hardware and driver version.
+
+Before **Godot 4.4**, there was no solution to pipeline compilation other than generating them when an object shows up inside the camera's view, leading to the infamous *shader stutter* or hitches that only occur during the first playthrough. With **Godot 4.4**, new mechanisms have been introduced to mitigate stutters from pipeline compilation.
+
+- **Ubershaders**: Godot makes use of specialization constants, a feature that allows the driver to optimize a pipeline's code around a set of parameters such as lighting, shadow quality, etc. Ubershaders are able to read these constants while rendering, which means Godot can pre-compile a pipeline ahead of time and compile the optimized version on the background during gameplay. This reduces the amount of pipelines that need to be created significantly.
```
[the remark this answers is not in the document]
DarioSamo: > Ubershaders are special version of the shader are able to change these constants while rendering, which means Godot can pre-compile just one pipeline ahead of time and compile the more optimized versions on the background during gameplay.

This one seems to introduce a grammar error, could you provide the corrected version?

tutorials/performance/pipeline_compilations.rst
DarioSamo: I feel this one may be a bit misleading as currently Surface compilations are a pretty big portion due to reasons such as just not being covered by the use ArrayMesh. We probably don't want to lead users down the path of taking out all overrides and not seeing that reflected in the monitor just yet.

tutorials/performance/pipeline_compilations.rst
DarioSamo: Correct, anything landing here is usually something we should address if possible. But since pre-compilation is not an automatic matter at the moment from a code maintenance point of view, it's a good metric to keep an eye on during development or for people who fork the engine and make their modifications to the renderer.

tutorials/performance/pipeline_compilations.rst
DarioSamo: I like it! Can I include this graph as is? Or do we need to handle it in some way to take transparency or background into account?

tutorials/performance/pipeline_compilations.rst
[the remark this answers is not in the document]
clayjohn: I updated the suggestion directly!

tutorials/performance/pipeline_compilations.rst
clayjohn:
```suggestion
- **Draw**: Compiled on demand when a 3D object needs to be drawn and an ubershader was not pre-compiled ahead of time. The engine is unable to pre-compile this pipeline due to triggering a case that hasn't been covered yet or a modification that was done to the engine's code. Leads to stutters during gameplay. This is identical to Godot versions before **4.4**. If you see compilations here, please let the devs know as this should never happen with the Ubershader system.
```

tutorials/performance/pipeline_compilations.rst
DarioSamo: Delaying for later.

DarioSamo:
The PR has been merged, marking ready for review.

KeyboardDanni:
It's nice to finally get some documentation on what can cause compilation stutters, but we'll also need to be sure to keep it up to date as the renderer changes. Is there some process we could come up with to make sure this happens?

DarioSamo:
> It's nice to finally get some documentation on what can cause compilation stutters, but we'll also need to be sure to keep it up to date as the renderer changes. Is there some process we could come up with to make sure this happens?

Probably not worth it. My end goal in the long term would be for this article to not even need to exist in the first place.

It's just that to get there it'll require the generalization of a lot of systems to go through something that dictates the rendering at a higher level and allows the engine to know in advance everything that will be required.

I think manual updates will suffice for the time being.

AThousandShips:

DarioSamo:
Thanks, done!

mhilbrunner (approved):

mhilbrunner:
Thank you for taking the time to document this awesome work of yours :) Merged.

PR #10013, opened 2024-09-28
Remove "degrees" redundancy

KleidonBD:
The phrase, "90° degrees", is a redundancy as the "°" symbol already means "degrees". It should just be ,"90°".

### DIFF
--- a/tutorials/math/vectors_advanced.rst
+++ b/tutorials/math/vectors_advanced.rst
@@ -172,7 +172,7 @@ For two points in space, there are actually two planes that pass through
 them, sharing the same space but with normal pointing to the opposite
 directions. To compute the normal from the two points, the direction
 vector must be obtained first, and then it needs to be rotated 90°
-degrees to either side:
+to either side:
 
 .. tabs::
  .. code-tab:: gdscript GDScript


tutorials/math/vectors_advanced.rst
```diff
@@ -172,7 +172,7 @@ For two points in space, there are actually two planes that pass through
 them, sharing the same space but with normal pointing to the opposite
 directions. To compute the normal from the two points, the direction
 vector must be obtained first, and then it needs to be rotated 90°
```
tetrapod00, line 174:
```suggestion
vector must be obtained first, and then it needs to be rotated 90
```
Existing usage across the docs definitely favors spelling out "degrees":
4 usages of the symbol: https://github.com/search?q=repo%3Agodotengine%2Fgodot-docs%20%C2%B0&type=code
Many usages of the full word: https://github.com/search?q=repo%3Agodotengine%2Fgodot-docs+degrees&type=code

tetrapod00: (If amending your PR as described [here](https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html#modifying-a-pull-request) is intimidating, feel free to close this PR and open another one.)

KleidonBD: You're right. I'll make a new PR. I was looking only at the previous page on vector maths where the symbol was used exclusively rather than the full word. Perhaps the other uses of the symbol should also be changed for consistency.

PR #10016, opened 2024-09-28
remove inexistent find_custom and rfind_custom functions from Array section

aziomo:
changelog:
- removed `find_custom` method from the Array section
- removed `rfind_custom` method from the Array section

for cleanup reasons, Array class does not have those methods as of 4.4.dev2

why is this in the docs? aside from community requests, there is not even a hint of this feature


AThousandShips:
This is incorrect, these methods exist in the latest version, they were added in:
* https://github.com/godotengine/godot/pull/95449

Also these are generated from the source and shouldn't be edited here (see at the top)

PR #10028, opened 2024-10-02
Add notes on bit depth for ints and floats

BrianBHuynh:
Added a small note about the bit depth of integers and floats in Godot's shading language as it is not explicitly stated anywhere.

The bit depth of integer and floats in GDscript and Godot's shading language are different, which can cause problems with lost precision in calculations when integers are set from GDscript as floats/ints in GDscript are 64 bits instead of 32 bits (the standard in GLSL ES 3.0).

While Added a small note about the bit depth of integers and floats in Godot's shading language as it is not explicitly stated anywhere.

The bit depth of integer and floats in GDscript and Godot's shading language are different, which can cause problems with lost precision in calculations when integers are set from GDscript as floats/ints in GDscript are 64 bits instead of 32 bits (the standard in GLSL ES 3.0).

While most are unlikely to run into problems due to this difference in bit depth, it can cause mathematical errors in edge cases. As stated by previous contributors, no error will be thrown if types do not match while setting a shader uniform. This includes GDscript floats being set as Godot shader floats (which may not be intuitive).

Examples of problems this may cause:
When setting two floats with time.get_unix_time_from_system() in GDscript (which returns a 64 bit float) a few seconds apart, when compared in the shader they will be equal to each other, and when subtracted from one another they will equal 0.0 due to the 32 bit depth of floats in the shader language.

This is not intuitive to debug without documentation as when using the get function in GDscript, they will still subtract correctly so that the second float will be greater than the first float in GDscript, even if they won’t subtract correctly within the shader.most are unlikely to run into problems due to this difference in bit depth, it can cause mathematical errors in edge cases. As stated by previous contributors, no error will be thrown if types do not match while setting a shader uniform. This includes GDscript floats being set as Godot shader floats (which may not be intuitive).

Additionally some functions mention 32 bit floating point numbers (look at packHalf2x16 for example) however I could not find anywhere that states that the default bit depth of int / floats was 32 bit and not 64 bit like in GDscript.


mhilbrunner (approved):

mhilbrunner:

mhilbrunner:

BrianBHuynh:
LESSGOOOO FIRST COMMIT

PR #10045, opened 2024-10-04
Document expression evaluator in Debugger panel

Calinou:
It might be worth finding more concrete use cases to document, but as someone who doesn't deal with REPLs often, I'm not sure what to put in exactly.

This also renames the Stack Trace tab to match its name since Godot 4.2.

Don't cherry-pick to `4.3`, as this feature is only in 4.4.

- See https://github.com/godotengine/godot/pull/97647.


mhilbrunner (approved):

mhilbrunner:
Thank you!

PR #10052, opened 2024-10-06
Use dedicated `Note` section instead of plain text

dustdfg:

### DIFF
--- a/tutorials/io/runtime_file_loading_and_saving.rst
+++ b/tutorials/io/runtime_file_loading_and_saving.rst
@@ -168,10 +168,10 @@ Example of loading an image and displaying it in a :ref:`class_TextureRect` node
 Audio/video files
 -----------------
 
-Godot supports loading Ogg Vorbis audio at runtime. Note that not *all* files
-with an ``.ogg`` extension may be Ogg Vorbis files. Some may be Ogg Theora
-videos, or contain Opus audio within an Ogg container. These files will **not**
-load correctly as audio files in Godot.
+Godot supports loading Ogg Vorbis audio at runtime.
+
+.. note::
+    Note that not *all* files with an ``.ogg`` extension may be Ogg Vorbis files.         Some may be Ogg Theora videos, or contain Opus audio within an Ogg container.         These files will **not** load correctly as audio files in Godot.
 
 Example of loading an Ogg Vorbis audio file in an :ref:`class_AudioStreamPlayer` node:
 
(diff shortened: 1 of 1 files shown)


tutorials/io/runtime_file_loading_and_saving.rst
```diff
@@ -168,10 +168,10 @@ Example of loading an image and displaying it in a :ref:`class_TextureRect` node
...
 -----------------
 
-Godot supports loading Ogg Vorbis audio at runtime. Note that not *all* files
-with an ``.ogg`` extension may be Ogg Vorbis files. Some may be Ogg Theora
-videos, or contain Opus audio within an Ogg container. These files will **not**
-load correctly as audio files in Godot.
+Godot supports loading Ogg Vorbis audio at runtime.
+
+.. note::
+    Note that not *all* files with an ``.ogg`` extension may be Ogg Vorbis files.         Some may be Ogg Theora videos, or contain Opus audio within an Ogg container.         These files will **not** load correctly as audio files in Godot.
```
tetrapod00, line 174:
```suggestion
    Not *all* files with an ``.ogg`` extension may be Ogg Vorbis files. Some may be
    Ogg Theora videos, or contain Opus audio within an Ogg container. These files
    will **not** load correctly as audio files in Godot.
```
The formatting got a bit mangled, this should fix it.

I think if this note block is moved to after the examples and before the note about WAV and MP3, it interrupts the flow a bit less.

tetrapod00 (approved):
I'm neutral on this change. I don't think it majorly improves readability to make every "Note that" a note block. But I'm not against it either.

tutorials/io/runtime_file_loading_and_saving.rst
dustdfg: About note about WAV and MP3. After merging https://github.com/godotengine/godot/pull/97855 and https://github.com/godotengine/godot/pull/97849 and similar PR for wav (which isn't ready because I don't know if @MaxIsJoe is going to implement for themselves) this notee will become outdated and wrong so I decided to convert it to draft and ship all the changes to this section in batch...

tetrapod00: It's okay to make smaller self-contained docs PRs. If you think this PR is worth it on its own, I would leave it up, and make a second PR for the WAV and MP3 changes when they are approved. But one big PR also works.

skyace65:
Closing this PR as I don't see any benefit to making it a note, and a lack of other people supporting this change

PR #10055, opened 2024-10-06
Update idle_and_physics_processing.rst: Specify that processing happens in *tree order*.

0x53A:
Specify that processing happens in *tree order*.

It was quite hard to find out in which order processing is called. Someone linked me to https://docs.godotengine.org/en/4.3/tutorials/scripting/scene_tree.html#tree-order which explains it, but it would be great to also have this information directly here.

For comparison, for ``ready`` it's clearly articulated that it's called in reverse, that is, first the children then the parent.

I'd like to also add this information to the API docs of [``Node``](https://docs.godotengine.org/en/stable/classes/class_node.html#class-node-private-method-process). Here again, it's spelled out for ``ready`` but omitted for ``process``.

I assume that would need to be changed in the source code repository?

### DIFF
--- a/tutorials/scripting/idle_and_physics_processing.rst
+++ b/tutorials/scripting/idle_and_physics_processing.rst
@@ -9,6 +9,8 @@ class to do so: :ref:`Node._process() <class_Node_private_method__process>` and
 :ref:`Node._physics_process() <class_Node_private_method__physics_process>`. If you
 define either or both in a script, the engine will call them automatically.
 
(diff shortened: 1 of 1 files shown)


0x53A:
It's slightly more complex, with ``process_priority`` and ``process_physics_priority ``. I already added this info to the docs for the two functions themselves (https://github.com/godotengine/godot/pull/97905) and I think that's good enough. Also adding it here might just be bloat.

PR #10060, opened 2024-10-08
Fixed typo in Control._can_drop_data C# example

mh114:
Fixed a typo (`dict` -> `data`) in `Control._drop_data()` **C#**-example. The typo is in the `_CanDropData(...)` method of the `_DropData(...)` example.


AThousandShips:

mh114:
> 

Ah, my bad! 🤦‍♂️ Sorry!

AThousandShips:
No worries the website makes it hard to do things right unfortunately, we're trying to make it easier but we haven't found a good solution yet

Issue #8196, opened 2023-10-08
Remove "Blender ESCN exporter" page from docs


tetrapod00:
**Your Godot version:**
4.1
**Issue description:**
The addon [Godot Blender Exporter](https://github.com/godotengine/godot-blender-exporter) appears to be out of date and no longer officially supported. Exporting from Blender to Godot has been superseded by other workflows. I suggest removing the [docs page](https://docs.godotengine.org/en/stable/tutorials/assets_pipeline/escn_exporter/index.html) for this addon entirely.

It's out of date:
- The last commit was nearly 2 years ago
- There is an [issue](https://github.com/godotengine/godot-blender-exporter/issues/418) that involves Blender 3.4 and Godot 4.X support. The apparent [answer ](https://github.com/godotengine/godot-blender-exporter/issues/418#issuecomment-1382176328 )is that the preferred way is now to use GLTF import.
- It is unclear if it even _works_ in 4.x - the same issue reports it doesn't work in a 4.0 alpha. To me it is a bit beside the point if it compiles and functions in 4.x. It is a _non-supported_ addon, with functionality that is officially supported elsewhere. There's no reason to be directing users to it via a docs page.

It's superseded by other functionality - the recommended way to export from Blender to Godot is either:
- Export GLTF from Blender, import GLTF to Godot.
- Directly import a .blend file (which implicitly does the above, with less manual control.

Anecdotally, I have seen one new Godot user try using this addon with 4.1, and they cited the "Up to date" marker to me in conversation (as an implicit reason for why they were using this addon). 
![image](https://github.com/godotengine/godot-docs/assets/145553014/20b79097-2b1b-4024-9de5-3094e806bfc2)
The fact that this pages appears appears in the outline at the same level as "Importing 3D scenes" and has "Blender" in the title means that users are directed to this page rather than the more up to date supported methods.

**Proposal:**
Remove the page.

If after discussion the page should stay, I propose adding a paragraph that notes the lack of support and links to the current recommended methods.

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/stable/tutorials/assets_pipeline/escn_exporter/index.html
**URL to the addon:**
https://github.com/godotengine/godot-blender-exporter

Uradamus:
[...]

PR #10064, opened 2024-10-09
Remove Blender ESCN exporter pages

tetrapod00:
Closes https://github.com/godotengine/godot-docs/issues/8196.
Alternate or complement to https://github.com/godotengine/godot-docs/pull/10067.

The addon [Godot Blender Exporter](https://github.com/godotengine/godot-blender-exporter) has not been officially supported since Godot 4.x. While it partially works in some cases (see https://github.com/godotengine/godot-docs/issues/8196#issuecomment-2400982178), since Blender version 4.0 it does not work at all. We should not be linking to an unsupported and unmaintained addon, much less using the official documentation as the documentation for an unmaintained addon!

I left a stub page which with a link to currently supported Blender exporting methods, as well as some information about the addon.

PR #10067, opened 2024-10-09
Mark Blender ESCN exporter as deprecated

tetrapod00:
Closes https://github.com/godotengine/godot-docs/issues/8196.
Alternate or complement to https://github.com/godotengine/godot-docs/pull/10064.

This PR adds a warning to each page under [Blender ESCN exporter](https://docs.godotengine.org/en/latest/tutorials/assets_pipeline/escn_exporter/index.html) that the blender plugin is depreacted in Godot 4.x. Personally, I believe even calling the plugin "deprecated" is promising too much, since it has not been maintained at all since Godot 4.0, and is producing shader errors for all tested Godot 4 versions (see https://github.com/godotengine/godot-docs/issues/8196#issuecomment-2400982178). "Deprecated" implies that the plugin functions in the current Godot version, but is not receiving updates.

Some combination of this PR and https://github.com/godotengine/godot-docs/pull/10064 (which removes these pages entirely, except for a stub redirect page) should be applied to all Godot 4.x docs versions. I see three options:
1. Apply this PR to 4.4 and cherrypick it to all 4.x docs versions (or as far back as we cherrypick).
2. Apply https://github.com/godotengine/godot-docs/pull/10064 to 4.4, and apply this PR to all previous 4.x docs versions (or as far back as we cherrypick).
3. Apply https://github.com/godotengine/godot-docs/pull/10064 to 4.4, and cherrypick it back to all previous 4.x docs versions (or as far back as we cherrypick).

Option 1 seems like the wrong choice to me. I don't think carrying these docs pages forward to future Godot 4.x versions makes sense. The plugin is unmaintained, unsupported, and does not even export at all since Blender version 4.0. The plugin does not work at all in Blender 4.2, which is the current Blender LTS version as of Godot 4.3. 

Option 2 makes sense to me, since it corrects the problem for future 4.x docs versions and keeps the existing docs pages for anyone still relying on them.

Option 3 also makes sense, and is my preferred solution. The Blender addon has in practice been unmaintained and unsupported since Godot 4.0. It should have been marked deprecated since Godot version 4.x, and perhaps not included in the Godot 4.x docs at all.

### DIFF
--- a/tutorials/assets_pipeline/escn_exporter/animation.rst
+++ b/tutorials/assets_pipeline/escn_exporter/animation.rst
@@ -1,5 +1,12 @@
 Animation
 =========
+
+.. warning::
+   This chapter covers the Blender plugin
+   `Godot Blender Exporter <https://github.com/godotengine/godot-blender-exporter>`__,
+   which is deprecated in Godot 4.x. To export from Blender to Godot 4.x, use
+   one of the :ref:`available 3D formats <doc_importing_3d_scenes_available_formats>`.
+
 Animation supported:
  - transform animation of all types of objects
  - transform animation of pose bones
--- a/tutorials/assets_pipeline/escn_exporter/index.rst
+++ b/tutorials/assets_pipeline/escn_exporter/index.rst
@@ -3,8 +3,11 @@
 Blender ESCN exporter
 =====================
 
-.. note:: This chapter relates to the Blender plugin called "Godot Blender Exporter",
-          which can be downloaded here: https://github.com/godotengine/godot-blender-exporter
+.. warning::
+   This chapter covers the Blender plugin
+   `Godot Blender Exporter <https://github.com/godotengine/godot-blender-exporter>`__,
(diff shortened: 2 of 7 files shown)


fire:
We have a similar problem with the collada importer.

tetrapod00:
>We have a similar problem with the collada importer.

There's just a single reference to that addon. Made a PR: https://github.com/godotengine/godot-docs/pull/10074

fire:
I think using the same strategy as the collada plugin would work here. Not sure which one that is on your list

tetrapod00:
There is no exact analogue to the the Collada plugin strategy, since this plugin also has several actual pages of docs. 

https://github.com/godotengine/godot-docs/pull/10064 is approved by two docs maintainers, so I'd suggest you review that and see if it's acceptable?

The current state of that PR removes the pages except for a stub, calls the plugin "unmaintained and unsupported", but does not call the plugin "deprecated", since it was not officially deprecated. It also manually links to an older version of the full docs, in case someone still needs them.

skyace65:
Closing this as #10064 was merged

PR #10073, opened 2024-10-11
Fixing some small errors in examples - Update class_nodepath.rst

SimonCle:
Fixing some small errors in the example of NodePath.get_name
I know it's small, but stuff like this can be confusing.


tetrapod00:
Thanks for the contribution! Unfortunately, this is the wrong repo.

Changes to the class reference should be made in the [main repository](https://github.com/godotengine/godot), not here. See https://docs.godotengine.org/en/stable/contributing/documentation/updating_the_class_reference.html for how to contribute to the class ref.

PR #10078, opened 2024-10-12
Document filename requirements in C# global classes

### LINKED ISSUE #9083: Should be made clear that classes marked with GlobalClass must be named the same as their file
**Your Godot version:**
4.2 Stable
**Issue description:**
The C# attribute [GlobalClass] which is used to mark classes which should be made available within the editor's Node selection dialogues does not work correctly unless the class name matches the file it's in. I'm not sure whether this is considered a bug or expected behaviour, but either way I think the documentation should indicate this limitation. It's difficult to determine what the issue is in this case since the project will still build without issue and the editor gives no indication as to why the class or derived classes are not available.
https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_global_classes.html

Calinou:
- Flesh out the introduction.

___

- This closes #9083.
- See https://github.com/godotengine/godot-docs-user-notes/discussions/194#discussioncomment-10919431.

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_global_classes.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_global_classes.rst
@@ -3,8 +3,11 @@
 C# global classes
 =================
 
-Global classes (also known as named scripts) are types registered in Godot's editor so they can be used
-more conveniently.
+Global classes (also known as named scripts) are types registered in Godot's
+editor so they can be used more conveniently.
+:ref:`In GDScript <doc_gdscript_basics_class_name>`, this is achieved
+using the ``class_name`` keyword at the top of a script. This page describes how
+to achieve the same effect in C#.
 
 - Global classes show up in the *Add Node* and *Create Resource* dialogs.
 - If an :ref:`exported property <doc_c_sharp_exports>` is a global class, the
@@ -21,6 +24,12 @@ Global classes are registered with the ``[GlobalClass]`` attribute.
     public partial class MyNode : Node
     {
     }
+
+.. warning::
+
+    The file name must match the class name in **case-sensitive** fashion.
+    For example, a global class named "MyNode" must have a file name of
+    ``MyNode.cs``, not ``myNode.cs``.
 
 The ``MyNode`` type will be registered as a global class with the same name as the type's name.
 
@@ -84,8 +93,7 @@ will let you create and load instances of this type easily.
 .. warning::
 
     The Godot editor will hide these custom classes with names that begin with the prefix
-    "Editor" in the 'Create New Node' or 'Create New Scene' dialog windows. The classes 
-    are available for instantiation at runtime via their class names, but are 
-    automatically hidden by the editor windows along with the built-in editor nodes used 
+    "Editor" in the "Create New Node" or "Create New Scene" dialog windows. The classes
+    are available for instantiation at runtime via their class names, but are
+    automatically hidden by the editor windows along with the built-in editor nodes used
     by the Godot editor.
-    


tutorials/scripting/c_sharp/c_sharp_global_classes.rst
```diff
@@ -22,6 +24,11 @@ Global classes are registered with the ``[GlobalClass]`` attribute.
     {
     }
 
+.. warning::
+
+    The file name (without the ``.cs`` extension) must match the class name
+    in **case-sensitive** fashion, as per C# language requirements.
```
raulsntos, line 30: I feel like this sound as if it's a restriction imposed by the C# language. The C# guidelines **do** recommend to do this, but it's not enforced. It's only enforced by Godot on C# classes that are used as scripts.

To be honest, I can't come up with something better. Maybe we can just remove the "as per C# language requirements" part?

raulsntos (commented):
Thanks, I think it makes sense to repeat this here because of how often users run into it.

This is already mentioned in the [_C# basics_](https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_basics.html) page (in the [_Current gotchas and known issues_](https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_basics.html#current-gotchas-and-known-issues) section) among other things to look out for. Maybe we should also link to that page?

tutorials/scripting/c_sharp/c_sharp_global_classes.rst
tetrapod00, line 30:
```suggestion
    The file name must match the class name in **case-sensitive** fashion,
    as per C# language requirements. For example, a global class named 
    "MyNode" must have a file name of ``MyNode.cs``, not ``myNode.cs``.

```

The phrase "without the ``.cs`` extension" read as slightly ambiguous to me.

tetrapod00 (approved):
Makes sense. The requirement that C# class names match their filenames is already documented in the note [here](https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_basics.html#example) and in the limitations [here](https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_basics.html#current-gotchas-and-known-issues), but adding another note on this page is good.

tutorials/scripting/c_sharp/c_sharp_global_classes.rst
```diff
@@ -3,8 +3,10 @@
 C# global classes
 =================
 
-Global classes (also known as named scripts) are types registered in Godot's editor so they can be used
-more conveniently.
+Global classes (also known as named scripts) are types registered in Godot's
+editor so they can be used more conveniently. In GDScript, this is achieved
+using the ``class_name`` keyword at the top of a script. This page describes how
+to achieve the same effect in C#.
```
tetrapod00, line 9:
```suggestion
Global classes (also known as named scripts) are types registered in Godot's
editor so they can be used more conveniently.
:ref:`In GDScript <doc_gdscript_basics_class_name>`, this is achieved
using the ``class_name`` keyword at the top of a script. This page describes how
to achieve the same effect in C#.
```
If we mention the GDScript implementation, we might as well link to it.

tutorials/scripting/c_sharp/c_sharp_global_classes.rst
```diff
@@ -84,8 +91,7 @@ will let you create and load instances of this type easily.
 .. warning::
 
     The Godot editor will hide these custom classes with names that begin with the prefix
-    "Editor" in the 'Create New Node' or 'Create New Scene' dialog windows. The classes 
-    are available for instantiation at runtime via their class names, but are 
-    automatically hidden by the editor windows along with the built-in editor nodes used 
+    "Editor" in the 'Create New Node' or 'Create New Scene' dialog windows. The classes
+    are available for instantiation at runtime via their class names, but are
+    automatically hidden by the editor windows along with the built-in editor nodes used
```
tetrapod00, line 96:
```suggestion
    "Editor" in the Create New Node or Create New Scene dialog windows. The classes
    are available for instantiation at runtime via their class names, but are
    automatically hidden by the editor windows along with the built-in editor nodes used
```
Use of single quotes `'` for the window names struck me as out of place. There's no real standard here, but no quotes seems most common.
Code_IwuKlHLJl0

tetrapod00 (commented):
Forgot to add these two comments.

tetrapod00:
This closes https://github.com/godotengine/godot-docs/issues/9083.

raulsntos (approved):

mhilbrunner (approved):

mhilbrunner:
Thank you!

mhilbrunner:

Issue #5701, opened 2022-03-21
Very long pages with no TOC


setanarut:
**Your Godot version:**
Godot 3.4.3 or Godot 4 alpha 4

**Issue description:**
There are very long reference pages with no TOC. I have to use the search bar on the left first and then the browser's search bar to find what I'm looking for. Then It's easy to get lost in the huge pages.

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_basics.html#classes

<img width="802" alt="Ekran Resmi 2022-03-21 22 35 43" src="https://user-images.githubusercontent.com/36481442/159355019-2673d72a-0ef9-4964-b67e-3addbbc85c78.png">

Github's auto-generated table of contents in the RST file. [godotengine.org](https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_basics.html#) doesn't have this TOC.

https://user-images.githubusercontent.com/36481442/159352131-4b275e79-dccf-4740-a02c-b8e9512e3da9.mov

# Enhancement request

1. As an alternative to the slow and cumbersome docs.godotengine.org site. Documents should be adapted to GitHub Flavored Markdown format and put on Github. So anyone who wants can fork and find what they're looking for.
2. or add TOC to the docs.godotengine.org. (My second reluctant choice.)

Calinou:
> As an alternative to the slow and cumbersome docs.godotengine.org site. Documents should be adapted to GitHub Flavored Markdown format and put on Github. So anyone who wants can fork and find what they're looking for.

Godot used to use a GitHub wiki (with Markdown formatting), but it ended up being too limited, especially since note/warning blocks can't be created with just Markdown. There are nonstandard Markdown extensions that make this possible, but GitHub doesn't support rendering those.

hazarek:
> It ended up being too limited, especially since note/warning blocks can't be created with just Markdown.

Note and warning boxes can be easily converted into quote blocks with a script. I even tried it and converted it. only links need to be rendered to relative links.

> **NOTE**: _Lorem ipsum dolor sit amet._

> **WARNING**: _Be realistic, demand the impossible!_

---

> so I guess we need to add a manual table of contents at the top of long pages such as GDScript reference.

It's useless if it's not one of the fixed menus in the corner like this;

https://user-images.githubusercontent.com/36481442/159363665-0e06931b-65b8-445d-b491-9e81bb2bd22b.mov


[...]

PR #10080, opened 2024-10-13
Improve GDScript Basics sidebar navigation by changing header levels

tetrapod00:
Resolves https://github.com/godotengine/godot-docs/issues/5701. IMO fixing the sidebar navigation will sufficiently fix the problem mentioned in that issue, although the OP is asking for a slightly different solution.

Promotes many headers up one level, so they appear in the sidebar. Removes the redundant Language header, since the whole page is about the language.

Each separate topic is promoted to a top-level subheader on this page, so it appears in the sidebar. The sidebar is long, but still fits on a single screen on a 1080 monitor. Ideally I would have liked to have two levels of navigable headers, but this is the best we can do with the current Sphinx theme. Subheaders under each promoted header are also promoted one level.

The Language and Data headers were removed. In the current organization, it was a parent header for Identifiers, Keywords, Operators, and a few others. However, this whole page is technically a description of the language, and because all the other headers were promoted to top-level, keeping the Language header seemed redundant to me. The Data header was similarly meant to have Variables, Constants, etc as subheaders, but if those subheaders are promoted it is now just an empty sibling header, and is redundant. I don't strongly care about this change; these two headers can be added back if anyone feels strongly.

| 4.4 Master | This PR |
| - | - |
| firefox_Nj6CIJujyC | firefox_mkA8FxDoGE |

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_basics.rst
+++ b/tutorials/scripting/gdscript/gdscript_basics.rst
@@ -118,15 +118,8 @@ If you have previous experience with statically typed languages such as
 C, C++, or C# but never used a dynamically typed one before, it is advised you
 read this tutorial: :ref:`doc_gdscript_more_efficiently`.
 
-Language
---------
-
-In the following, an overview is given to GDScript. Details, such as which
-methods are available to arrays or other objects, should be looked up in
-the linked class descriptions.
-
 Identifiers
-~~~~~~~~~~~
+-----------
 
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):
IMO an improvement.

mhilbrunner:
Thank you!

mhilbrunner:

PR #10081, opened 2024-10-14
Adding Warnings against possible Remote Code Execution when loading untrusted resources

betalars:
This issue is already mentioned in the wiki when you look up the comments under the [saving games](https://docs.godotengine.org/en/stable/tutorials/io/saving_games.html) tutorial, however I think it is nice to add these safeguards until a proper solution has been implemented.

As this security vulnerability is already publicly discussed and assumed to be known by the maintainers and the educated public, I think it is fair to be at least honest about this limitation.


AThousandShips:

betalars:
Thanks, created https://github.com/godotengine/godot/pull/98168 intead.

PR #10084, opened 2024-10-14
Fix shader function overloading paragraph

Chaosus:
Doc fix after https://github.com/godotengine/godot/pull/92441 merge.

### DIFF
--- a/tutorials/shaders/shader_reference/shading_language.rst
+++ b/tutorials/shaders/shader_reference/shading_language.rst
@@ -637,13 +637,23 @@ Example below:
         result = a + b;
     }
 
-.. note::
+Function overloading is supported. You can define multiple functions with the same
+name, but different arguments. Note that `implicit casting <Casting_>`_ in overloaded
+function calls is not allowed, such as from ``int`` to ``float`` (``1`` to ``1.0``).
 
-    Unlike GLSL, Godot's shader language does **not** support function
-    overloading. This means that a function cannot be defined several times with
-    different argument types or numbers of arguments. As a workaround, use
-    different names for functions that accept a different number of arguments or
-    arguments of different types.
+.. code-block:: glsl
+
+    vec3 get_color(int t) {
+        return vec3(1, 0, 0); // Red color.
+    }
+    vec3 get_color(float t) {
+        return vec3(0, 1, 0); // Green color.
+    }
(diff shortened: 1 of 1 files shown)


tutorials/shaders/shader_reference/shading_language.rst
```diff
@@ -637,13 +637,25 @@ Example below:
         result = a + b;
     }
 
-.. note::
+Starting off Godot 4.4 you can now define multiple functions with the same name but different
+arguments (**overloading**). An implicit constant conversion (e.g. `1` -> `1.0`) will be disabled
+on function call for such cases to prevent ambiguity.
```
tetrapod00, line 642:
```suggestion
Function overloading is supported. You can define multiple functions with the same
name, but different arguments. Note that `implicit casting <Casting_>`_ in function
calls is not allowed, such as from ``int`` to ``float`` (``1`` to ``1.0``).
```

In most cases the docs do not need to specify the first version in which a feature is supported, since it's only relevant for the first few versions after a change (and only if you re-read the docs page on each release).

In Godot 4.3, implicit casting is already forbidden for GDShader, including for function calls:
Godot_v4 3-stable_win64_QCagaFSYj7

Also revised the paragraph for grammar.

tutorials/shaders/shader_reference/shading_language.rst
```diff
@@ -637,13 +637,25 @@ Example below:
...
+    }
+
+    vec3 foo(float t) {
+        return vec3(0, 1, 0); // Green color.
+    }
+
+    void foo() {
+        vec3 red = foo(1);
+        vec3 green = foo(1.0);
+    }
```
tetrapod00, line 657:
```suggestion
    vec3 get_color(int t) {
        return vec3(1, 0, 0); // Red color.
    }

    vec3 get_color(float t) {
        return vec3(0, 1, 0); // Green color.
    }

    void fragment() {
        vec3 red = get_color(1);
        vec3 green = get_color(1.0);
    }
```
Nitpick, but avoid `foo()` and `bar()` style examples if possible. And in this case, use `fragment()` instead of `foo()` as the calling function to avoid confusion.

mhilbrunner (approved):

mhilbrunner:
Thank you!

PR #10085, opened 2024-10-15
Update code example to use "is not" in Static Typing

KleidonBD:
Godot 4.3 added the ability to use "is not" instead of "not a is b" for improved readability. This code example should be updated to use the new feature to be more up to date with the latest version.

Also, the original line used parentheses in its conditional statement, which is explicitly against the [style guide](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html#avoid-unnecessary-parentheses).

### DIFF
--- a/tutorials/scripting/gdscript/static_typing.rst
+++ b/tutorials/scripting/gdscript/static_typing.rst
@@ -260,7 +260,7 @@ get full autocompletion on the player variable thanks to that cast.
     in some cases, it can also lead to bugs. Use the ``as`` keyword only if this
     behavior is intended. A safer alternative is to use the ``is`` keyword::
 
-        if not (body is PlayerController):
+        if body is not PlayerController:
             push_error("Bug: body is not PlayerController.")
 
         var player: PlayerController = body


tetrapod00:
Duplicate of https://github.com/godotengine/godot-docs/pull/9777.

tetrapod00 (approved):
This seems like a fine change. 

Using the `is not` keyword without first documenting `is` is a bit strange. But this note block is meant to document type casting, not the `is` keyword itself, and there are two other examples on this page that use the `is` keyword:
https://github.com/godotengine/godot-docs/blob/9a44730c9297b058b0186dd99a9f3579c1a1e132/tutorials/scripting/gdscript/static_typing.rst?plain=1#L424

https://github.com/godotengine/godot-docs/blob/9a44730c9297b058b0186dd99a9f3579c1a1e132/tutorials/scripting/gdscript/static_typing.rst?plain=1#L466

Between this PR and https://github.com/godotengine/godot-docs/pull/9777, I slightly prefer the other PR. But this one is good and does not need additional changes.

KleidonBD:
Completely understandable. Like you said, the keyword itself isn't the focus of the code block which is why I didn't feel the need to call attention to it. The main thing I sought to change was removing the parentheses to align with the GDScript style guide, and figured this would kill two birds with one stone.

mhilbrunner:
Finally got around to this - I just merged https://github.com/godotengine/godot-docs/pull/9777, so I'm going to close this one. Thanks for contributing :)

Issue #10091, opened 2024-10-15
On "Making main screen plugins" page, EditorInterface methods called as if static on C#, when they are not


dougVanny:
**Your Godot version:** v4.3.stable.mono.official [77dcf97d8]

**Issue description:** References to `EditorInterface` methods in the C# examples are wrong. Methods such as `GetEditorTheme()` is are not static method, so instead it should be `EditorInterface.Singleton`

**URL to the documentation page:** https://docs.godotengine.org/en/stable/tutorials/plugins/editor/making_main_screen_plugins.html

![image](https://github.com/user-attachments/assets/6e0e1b6b-92d0-4960-8a60-ca2043908ee1)
![image](https://github.com/user-attachments/assets/9639ac2e-4872-49e6-baeb-b74f54f288a9)
![image](https://github.com/user-attachments/assets/6dcf4067-15f1-43f1-a304-844a50c46593)

AThousandShips:
See:
* https://github.com/godotengine/godot-docs/pull/9399#issuecomment-2119334224
* https://github.com/godotengine/godot-docs/issues/9626

This is not wrong use as such

tetrapod00:
Fixed by https://github.com/godotengine/godot-docs/pull/9399.

PR #10092, opened 2024-10-15
On "Making main screen plugins" page, EditorInterface methods called as if static on C#, when they are not

dougVanny:
Fixes https://github.com/godotengine/godot-docs/issues/10091


AThousandShips:
Thank you for your contribution, this is however already handled in:
* https://github.com/godotengine/godot-docs/pull/9399

Which also has some relevant feedback to how this isn't incorrect use

Closing as there's already an open PR for this specific code

PR #10105, opened 2024-10-17
Add warning about uninitialized variables in shaders

tetrapod00:
Closes https://github.com/godotengine/godot/issues/98223.

Adds a warning note to shading_language.rst. GDShader does not initialize local variables to 0, and an uninitialized local variable can contain an arbitrary value.

There is no section specifically about variables in this page, so I added the note to the [Data Types](https://docs.godotengine.org/en/stable/tutorials/shaders/shader_reference/shading_language.html#data-types) section.

I briefly tested both varyings and uniforms, and both appear to be consistently zero-initialized if not set manually.


mhilbrunner (approved):

mhilbrunner:

PR #10114, opened 2024-10-19
Add new Renderers page

tetrapod00:
This PR adds a new page called "Renderers". It provides a short description of each renderer, as well as a section on how to choose a renderer. It also contains a table listing all the feature differences between the three renderers. This page reimplements something like the 3.x page [Differences between GLES2 and GLES3](https://docs.godotengine.org/en/3.6/tutorials/rendering/gles2_gles3_differences.html).

Initially I only intended to write a feature comparison page, but I realized that there is not actually a good user-facing description of the renderers at this time. They are briefly described in [List of Features](https://docs.godotengine.org/en/latest/about/list_of_features.html#rendering), and described in a very technical way in [Internal rendering architecture](https://docs.godotengine.org/en/stable/contributing/development/core_and_modules/internal_rendering_architecture.html#internal-rendering-architecture), which is more for developers and power users. This page adds something more in-between, and is a starting point place for future documentation of the renderers.

This page is called "Renderers" to be slightly future-proof. I foresee first adding some longer sections about each renderer, then as those get longer splitting this page into multiple pages. For now it's a nice clean page though.

To fill out the tables, I looked for every instance where the docs describe a feature as being limited to one renderer or another. I trusted the docs and did not verify all this in-engine.

Here's a screenshot of how the tables currently look:

firefox_faWi45Tnb7


tutorials/rendering/renderers.rst
```diff
@@ -0,0 +1,300 @@
...
++-------------------------+--------------------------+--------------------------+--------------------------+
+| Unlit color buffer      | ❌ Not supported.*       | ❌ Not supported.*       | ❌ Not supported.*       |
++-------------------------+--------------------------+--------------------------+--------------------------+
+| Lighting buffer         | ❌ Not supported.*       | ❌ Not supported.*       | ❌ Not supported.*       |
++-------------------------+--------------------------+--------------------------+--------------------------+
+| Stencil buffer          | ❌ Not supported.        | ⚠️ Planned.              | ⚠️ Planned.              |
++-------------------------+--------------------------+--------------------------+--------------------------+
+| Custom buffer           | ❌ Not supported.        | ❌ Not supported.        | ❌ Not supported.        |
++-------------------------+--------------------------+--------------------------+--------------------------+
+| Compute shaders         | ❌ Not supported.        | ⚠️ Limited.              | ✔️ Supported.            |
```
tetrapod00, line 275: How do we feel about using unicode emojis here? Alternately we can use the green, yellow, and red dot images that are used in [Godot release policy](https://docs.godotengine.org/en/stable/about/release_policy.html#release-support-timeline) and a few other places.

tutorials/rendering/renderers.rst
```diff
@@ -0,0 +1,300 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+See :ref:`doc_standard_material_3d` for more information.
+
++-------------------------+--------------------------+--------------------------+--------------------------+
+| Feature                 | Compatibility            | Mobile                   | Forward+                 |
++=========================+==========================+==========================+==========================+
+| Sub-surface scattering  | ❌ Not supported.        | ❌ Not supported.        | ✔️ Supported.            |
+|                         |                          |                          |                          |
++-------------------------+--------------------------+--------------------------+--------------------------+
```
tetrapod00, line 247: I focused on only adding features that *varied* between the renderers, but that means that some tables such as this one are very short. I could add in some more features, like `triplanar` or `refraction` that are supported by all renderers, too.

tutorials/rendering/renderers.rst
```diff
@@ -0,0 +1,300 @@
...
+| Normal/Roughness buffer | ❌ Not supported.        | ✔️ Supported.            | ✔️ Supported.            |
++-------------------------+--------------------------+--------------------------+--------------------------+
+| Unlit color buffer      | ❌ Not supported.*       | ❌ Not supported.*       | ❌ Not supported.*       |
++-------------------------+--------------------------+--------------------------+--------------------------+
+| Lighting buffer         | ❌ Not supported.*       | ❌ Not supported.*       | ❌ Not supported.*       |
++-------------------------+--------------------------+--------------------------+--------------------------+
+| Stencil buffer          | ❌ Not supported.        | ⚠️ Planned.              | ⚠️ Planned.              |
++-------------------------+--------------------------+--------------------------+--------------------------+
+| Custom buffer           | ❌ Not supported.        | ❌ Not supported.        | ❌ Not supported.        |
```
tetrapod00, line 273: Documenting the features that are *not* supported at all could be out of scope for this page.

mhilbrunner: Per-instance shader uniforms are supported in Compat as of a few days ago in 4.4.

Features that are available in all or in none of the backends IMO should be out of scope here, because otherwise the page just ends up a giant table of all hypothetical features a renderer could have.

tutorials/rendering/renderers.rst
mhilbrunner: Fine either way, IMO.

tutorials/rendering/renderers.rst
mhilbrunner: Left a comment for this on the other one.

mhilbrunner:
I wonder if there is somewhere in the class reference this could fit (do we have documented classes for the renderers? Or maybe project settings?) - having these directly available as a comparison in-engine when staring at the renderer selection would be useful.

I guess if nothing else we could at least fit some links to this new page in there somewhere. Having it in the class ref takes it closer to the source however, which always makes it more likely to be kept updated as the renderers change.

tutorials/rendering/renderers.rst
tetrapod00: I would make an exception for *sets* of features where one one of the set is supported by all renderers. For example, I want to make clear that the normals/rough texture is not supported on Compatibility, but the depth texture is. I also want to list any antialiasing methods that are supported on all renderers, because most are not.

Either way the completely unsupported ones in this table are clearly out of scope.

tetrapod00:
> I wonder if there is somewhere in the class reference this could fit

There is the project setting [rendering/renderer/rendering_method](https://docs.godotengine.org/en/latest/classes/class_projectsettings.html#class-projectsettings-property-rendering-renderer-rendering-method) (only place the renderer is exposed, we don't even have a separate way to get it at runtime yet). Can the in-engine class ref even do tables like this well?

To me this really does feel more like a manual page rather than a class ref page.

tetrapod00:
Pushed a change to the overall comparison table. Now it uses less text and more icons, so should be easier to read at a glance.

tutorials/rendering/renderers.rst
clayjohn: Ya, your examples make sense. But the ones highlighted by this comment should be removed. 

Also, as a side note, stencil should be supported in all three renderers eventually. And custom buffer will likely be supported even before stencil! But I wouldn't mention them here until they are implemented.

mhilbrunner:
> To me this really does feel more like a manual page rather than a class ref page.

Lets do this here then and link in the class ref where possible/useful.

mhilbrunner (approved):
> To me this really does feel more like a manual page rather than a class ref page.

Lets do this here then and link in the class ref where possible/useful.

Content looks good to me. I haven't build this locally yet to verify all the tables render fine, some rows look like they may possibly be malformed (off by a single space, and also not sure whether all fields need to be enclosed by spaces, some directly end in the `|`).

tetrapod00:
The tables likely appear malformed due to the use of unicode emojis. In VS Code, emojis are not monospaced, so each emoji introduces a half-character offset. It's quite possible the same is true in the GitHub preview. I've been building locally and the tables render fine. (Of course, this is one more point against using emojis instead of the existing color dots.)

> some directly end in the `|`

This also renders fine for me locally, and it's already used in other tables in the docs. If this is poor style I don't mind changing it though.

mhilbrunner:
No, if they render fine thats alright to me. I'm fine with merging as is then, I'll let it stay open a bit more in case the others still want to chime in and/or review.

tetrapod00:
There is no need to rush merging, and this needs a deeper review from at least one person very familiar with the renderers. It's likely that I made at least one mistake when transcribing data from various places in the docs.

Also, I'd like to make sure that the way I characterized the renderers in the intro and overview is correct - for example I described Forward+ as "the focus of new development", which seems true from the outside looking in, but isn't necessarily something that should be stated in the docs.

tutorials/rendering/renderers.rst
Calinou: The emojis make more sense to use nowadays, especially since they scale better to hiDPI displays and show up directly in most text editors.

tutorials/rendering/renderers.rst
```diff
@@ -0,0 +1,315 @@
...
+
+See :ref:`doc_introduction_to_global_illumination` for more information.
+
++-------------------------+--------------------------+--------------------------+--------------------------+
+| Feature                 | Compatibility            | Mobile                   | Forward+                 |
++=========================+==========================+==========================+==========================+
+| ReflectionProbe         | ⚠️ Planned.              | ✔️ Supported.            | ✔️ Supported.            |
+|                         |                          |                          |                          |
++-------------------------+--------------------------+--------------------------+--------------------------+
+| LightmapGI              | ⚠️ Planned.              | ✔️ Supported.            | ✔️ Supported.            |
```
Calinou, line 190: LightmapGI baking and rendering are both supported in Compatibility since 4.3.

Baking requires a GPU with RenderingDevice support (I'd mention that in a footnote).

tutorials/rendering/renderers.rst
```diff
@@ -0,0 +1,315 @@
...
++---------------------+--------------------------+--------------------------+--------------------------+
+| Feature             | Compatibility            | Mobile                   | Forward+                 |
++=====================+==========================+==========================+==========================+
+| **Required**        | Older or low-end.        | Newer or high-end.       | Newer or high-end.       |
+| **hardware**        |                          | Requires Vulkan support. | Requires Vulkan support. |
++---------------------+--------------------------+--------------------------+--------------------------+
+| Runs on new hardware| ✔️ Yes.                  | ✔️ Yes.                  | ✔️ Yes.                  |
++---------------------+--------------------------+--------------------------+--------------------------+
+| Runs on old and     | ✔️ Yes.                  | ❌ No.                   | ❌ No.                   |
+| low-end hardware    |                          |                          |                          |
```
Calinou, line 98: Forward+/Mobile run on low-end hardware, but they won't run as well as Compatibility.

Mobile runs better on low-end hardware compared to Forward+.

tutorials/rendering/renderers.rst
```diff
@@ -0,0 +1,315 @@
...
+| **Feature set**     | 2D and core 3D features. | Most rendering features. | All rendering features.  |
++---------------------+--------------------------+--------------------------+--------------------------+
+| 2D rendering        | ✔️ Yes.                  | ✔️ Yes.                  | ✔️ Yes.                  |
+| features            |                          |                          |                          |
++---------------------+--------------------------+--------------------------+--------------------------+
+| Core 3D rendering   | ✔️ Yes.                  | ✔️ Yes.                  | ✔️ Yes.                  |
+| features            |                          |                          |                          |
++---------------------+--------------------------+--------------------------+--------------------------+
+| Advanced            | ❌ No.                   | ✔️ Yes, limited by       | ✔️ Yes. All rendering    |
+| rendering features  |                          | mobile hardware.         | features are supported.  |
```
Calinou, line 130: Use warning triangle icon for Mobile, since features like SDFGI and volumetric fog are missing:

```suggestion
| Advanced            | ❌ No.                   | ⚠️ Yes, limited by       | ✔️ Yes. All rendering    |
| rendering features  |                          | mobile hardware.         | features are supported.  |
```

tutorials/rendering/renderers.rst
```diff
@@ -0,0 +1,315 @@
...
++---------------------+--------------------------+--------------------------+--------------------------+
+| New features        | ⚠️ Some new rendering    | ✔️ Most new rendering    | ✔️ All new features are  |
+|                     | features are added to    | features are added to    | added to Forward+. As the|
+|                     | Compatibility. Features  | Mobile. Mobile usually   | focus of new development,|
+|                     | are added after Mobile   | gets new features as     | Forward+ gets features   |
+|                     | and Forward+.            | Forward+ does.           | first.                   |
++---------------------+--------------------------+--------------------------+--------------------------+
+| Rendering cost      | Low base cost, but       | Medium base cost, and    | Highest base cost, and   |
+|                     | high scaling cost.       | low scaling cost.        | low scaling cost.        |
```
Calinou, line 139: Mobile does not use clustered lighting, so it doesn't scale as well to complex scenes as Forward+:

```suggestion
| Rendering cost      | Low base cost, but       | Medium base cost, and    | Highest base cost, and   |
|                     | high scaling cost.       | medium scaling cost.        | low scaling cost.        |
```

tutorials/rendering/renderers.rst
```diff
@@ -0,0 +1,315 @@
...
+
+Global Illumination
+~~~~~~~~~~~~~~~~~~~
+
+See :ref:`doc_introduction_to_global_illumination` for more information.
+
++-------------------------+--------------------------+--------------------------+--------------------------+
+| Feature                 | Compatibility            | Mobile                   | Forward+                 |
++=========================+==========================+==========================+==========================+
+| ReflectionProbe         | ⚠️ Planned.              | ✔️ Supported.            | ✔️ Supported.            |
```
Calinou, line 187: Limits vary depending on renderer. Also, Compatibility supports reflection probes since 4.3:

```suggestion
| ReflectionProbe         | ⚠️ Supported, 2 per mesh.              | ✔️ Supported, 8 per mesh.            | ✔️ Supported, unlimited.            |
```

tutorials/rendering/renderers.rst
```diff
@@ -0,0 +1,315 @@
...
++-------------------------+--------------------------+--------------------------+--------------------------+
+| Screen-Space            | ❌ Not supported.        | ❌ Not supported.        | ✔️ Supported.            |
+| Indirect Lighting (SSIL)|                          |                          |                          |
++-------------------------+--------------------------+--------------------------+--------------------------+
+| Signed Distance Field   | ❌ Not supported.        | ❌ Not supported.        | ✔️ Supported.            |
+| Global Illumination     |                          |                          |                          |
+| (SDFGI)                 |                          |                          |                          |
++-------------------------+--------------------------+--------------------------+--------------------------+
+| Reflection probes       | ❌ Not supported.        | ✔️ Supported.            | ✔️ Supported.            |
++-------------------------+--------------------------+--------------------------+--------------------------+
```
Calinou, line 204: This is already listed above, so it should be removed.

tutorials/rendering/renderers.rst
```diff
@@ -0,0 +1,315 @@
...
+| MSAA              | ✔️ Supported.            | ✔️ Supported.            | ✔️ Supported.            |
++-------------------+--------------------------+--------------------------+--------------------------+
+| TAA               | ❌ Not supported.        | ❌ Not supported.        | ✔️ Supported.            |
++-------------------+--------------------------+--------------------------+--------------------------+
+| FSR2              | ❌ Not supported.        | ❌ Not supported.        | ✔️ Supported.            |
++-------------------+--------------------------+--------------------------+--------------------------+
+| FXAA              | ❌ Not supported.        | ✔️ Supported.            | ✔️ Supported.            |
++-------------------+--------------------------+--------------------------+--------------------------+
+| SSAA              | ❌ Not supported.        | ✔️ Supported.            | ✔️ Supported.            |
++-------------------+--------------------------+--------------------------+--------------------------+
```
Calinou, line 260: 3D scaling in Compatibility is supported since 4.3, so SSAA is supported now (it just involves setting the render scale above `1.0`).

tutorials/rendering/renderers.rst
```diff
@@ -0,0 +1,315 @@
...
+
+Antialiasing
+~~~~~~~~~~~~
+
+See :ref:`doc_3d_antialiasing` for more information.
+
++-------------------+--------------------------+--------------------------+--------------------------+
+| Feature           | Compatibility            | Mobile                   | Forward+                 |
++===================+==========================+==========================+==========================+
+| MSAA              | ✔️ Supported.            | ✔️ Supported.            | ✔️ Supported.            |
```
Calinou, line 251:
```suggestion
| MSAA              | ⚠️ Supported (3D only).            | ✔️ Supported (2D + 3D).            | ✔️ Supported (2D + 3D).            |
```

tutorials/rendering/renderers.rst
```diff
@@ -0,0 +1,315 @@
...
+Choosing a renderer
+-------------------
+
+Choosing a renderer is a complex question, and depends on your hardware and the
+which platforms you are developing for. As a starting point:
+
+Choose **Forward+** if:
+
+    - You are developing for desktop.
+    - You have relatively new hardware which supports Vulkan.
```
Calinou, line 40:
```suggestion
    - You have relatively new hardware which supports Vulkan, Direct3D 12 or Metal.
```

tutorials/rendering/renderers.rst
```diff
@@ -0,0 +1,315 @@
...
+    - You are developing a 3D game.
+    - You want to use the most advanced rendering features.
+
+Choose **Mobile** if:
+
+    - You are developing for newer mobile devices, XR, or desktop.
+    - You have relatively new hardware which supports Vulkan.
```
Calinou, line 47:
```suggestion
    - You have relatively new hardware which supports Vulkan, Direct3D 12 or Metal.
```

tutorials/rendering/renderers.rst
```diff
@@ -0,0 +1,315 @@
...
+
+In the editor, you can always switch between renderers by clicking on the renderer
+name in the upper-right corner of the editor.
+
+Switching between renderers may require some manual tweaks to your scene, lighting,
+and environment, since each renderer is different. In general, switching between
+the Forward+ or Mobile renderers and the Compatibility renderer will require more
+adjustments than switching between the Mobile and Forward+ renderers.
+
+Switching between different renderers at runtime is not yet supported.
```
Calinou, line 77:
```suggestion
Since Godot 4.4, when using Forward+ or Mobile, if Vulkan is not supported, the engine will fall back to Direct3D 12 and vice versa. If the attempted fallback driver is not supported either, the engine will then fall back to Compatibility when RenderingDevice backend is not supported. This behavior can be disabled in the project settings by unchecking **Rendering > Rendering Device > Fallback to OpenGL 3**.

Switching between different renderers at runtime is not yet supported.
```

tutorials/rendering/renderers.rst
tetrapod00: Sure. In that case, I think there should be another line in the table specifically for requiring Vulkan/D3D12/Metal support. The editor not running at all is a hard limit that deserves a solid X.

tutorials/rendering/renderers.rst
AThousandShips:
```suggestion
    - You have relatively new hardware which supports Vulkan, Direct3D 12, or Metal.
```

tutorials/rendering/renderers.rst
AThousandShips:
```suggestion
    - You have relatively new hardware which supports Vulkan, Direct3D 12, or Metal.
```

tutorials/rendering/renderers.rst
tetrapod00: Split this into two rows for 2D and 3D MSAA instead.

tutorials/rendering/renderers.rst
tetrapod00: This page also isn't the first place to use them. Emojis are used on the [minor version upgrading pages,](https://docs.godotengine.org/en/latest/tutorials/migrating/upgrading_to_godot_4.3.html#gdextension) too.

tutorials/rendering/renderers.rst
tetrapod00: RenderingDevice is only mentioned briefly as an "other feature". It might be worth introducing the concept of RenderingDevice as an abstraction layer in the introduction. That means we can use "RenderingDevice support" in some places instead of the verbose "Vulkan, Direct3D 12, or Metal".

tetrapod00:
Changes made. Other pages were also updated in cases where they were incorrect.

This page can no longer be cherrypicked to 4.3, since it references the new 4.4 setting `Rendering > Rendering Device > Fallback to OpenGL 3`.

I also added a section which describes the distinction between a renderer, a rendering driver, and the RenderingDevice backend. Included in this new section is a diagram, made with draw.io, using colors picked from the docs theme. 
Source files: [rendering_layers.drawio.xml.zip](https://github.com/user-attachments/files/17532097/rendering_layers.drawio.xml.zip)
Preview:

renderers_rendering_layers

tutorials/rendering/renderers.rst
```diff
@@ -0,0 +1,357 @@
...
+Choosing a renderer
+-------------------
+
+Choosing a renderer is a complex question, and depends on your hardware and the
+which platforms you are developing for. As a starting point:
+
+Choose **Forward+** if:
+
+    - You are developing for desktop.
+    - You have relatively new hardware which supports Vulkan, Direct3D 12 or Metal.
```
AThousandShips, line 63:
```suggestion
    - You have relatively new hardware which supports Vulkan, Direct3D 12, or Metal.
```

tutorials/rendering/renderers.rst
```diff
@@ -0,0 +1,357 @@
...
+    - You are developing a 3D game.
+    - You want to use the most advanced rendering features.
+
+Choose **Mobile** if:
+
+    - You are developing for newer mobile devices, XR, or desktop.
+    - You have relatively new hardware which supports Vulkan, Direct3D 12 or Metal.
```
AThousandShips, line 70:
```suggestion
    - You have relatively new hardware which supports Vulkan, Direct3D 12, or Metal.
```

tutorials/rendering/renderers.rst
```diff
@@ -0,0 +1,357 @@
...
+Introduction
+------------
+
+Godot 4 includes three renderers:
+
+- **Forward+**, sometimes called **Clustered Forward**. The most advanced renderer,
+  suited for desktop platforms only. Used by default on desktop platforms. This
+  renderer uses **Vulkan**, **Direct3D 12**, or **Metal** as the rendering driver,
+  and it uses the **RenderingDevice** backend.
+- **Mobile**, sometimes called **Forward Mobile**. Less features, but renders
```
AThousandShips, line 21:
```suggestion
- **Mobile**, sometimes called **Forward Mobile**. Fewer features, but renders
```

tutorials/rendering/renderers.rst
```diff
@@ -0,0 +1,357 @@
...
+
+Feature comparison
+------------------
+
+This is not a complete list of the features of each renderer. If a feature is
+not listed here, it is available in all renderers, though it may be much faster
+on some renderers. For a list of *all* features in Godot, see :ref:`doc_list_of_features`. 
+
+Hardware with RenderingDevice support is hardware which can run Vulkan, Direct3D
+12 or Metal.
```
AThousandShips, line 116:
```suggestion
12, or Metal.
```

tutorials/3d/standard_material_3d.rst
```diff
@@ -468,6 +468,8 @@ mapping.
 Subsurface Scattering
 ---------------------
 
+*This is only available in the Forward+ backend, not the Forward Mobile
+or Compatibility backends.*
```
AThousandShips, line 472:
```suggestion
or Compatibility renderer.*
```
Matching the others

tutorials/3d/global_illumination/introduction_to_global_illumination.rst
```diff
@@ -121,7 +121,7 @@ In order of performance from fastest to slowest:
   - ReflectionProbes with their update mode set to **Always** are much more
     expensive than probes with their update mode set to **Once** (the default).
     Suited for integrated graphics when using the **Once** update mode.
-    *Available when using the Forward Mobile backend. Will be available in the Compatibility backend in later releases.*
+    *Available when using the Mobile and Compatibility renderers.*
```
AThousandShips, line 124: Are reflection probes unavailable in `Forward+`? Otherwise I'd remove this as it is available in all renderers

tetrapod00: That was a copy-paste error. It's available in all three renderers, but you're right it makes sense to leave out the note in that case.

tetrapod00: Nevermind. The conventions in this section are weird. All global illumination methods are available in Forward+, so the other notes here only describe whether the global illumation method is available in Mobile or Compatibility, for example:
`**Not available** *when using the Forward Mobile or Compatibility backends.*`
`*Available when using the Forward Mobile backend. Will be available in the Compatibility backend in later releases.*`
`**Not available** *when using the Forward Mobile or Compatibility backends.*`

With that context, the note I wrote fits local conventions.

I'm working on this section anyway as part of changes for https://github.com/godotengine/godot-docs/issues/10139, changing "Forward Mobile" to "Mobile". I'll change all these notes to be more sensible as part of that change, in another PR.

tutorials/rendering/renderers.rst
```diff
@@ -0,0 +1,357 @@
...
+
+See :ref:`doc_lights_and_shadows` for more information.
+
++-------------------------+--------------------------+--------------------------+--------------------------+
+| Feature                 | Compatibility            | Mobile                   | Forward+                 |
++=========================+==========================+==========================+==========================+
+| Lighting approach       | Forward                  | Forward                  | Clustered Forward        |
+|                         |                          |                          |                          |
++-------------------------+--------------------------+--------------------------+--------------------------+
+| Maximum                 | 8 per mesh. Can be       | 8 per mesh, 256 per view.| Unlimited.               |
```
clayjohn, line 194: I don't think it is true that the Forward+ renderer supports Unlimited lights. 

The clustered element limit is per-clustered item i.e. ( the limit is 512 omni lights, 512 spot lights, 512 reflectionprobes, 512 decals, per cluster)

I would remove the clustered elements line, and instead record it here:
"512 per cluster. Can be increased."

tutorials/rendering/renderers.rst
clayjohn: This comment got resolved and the note was removed in the 3D antialising page, but it is still incorrect here.

tutorials/rendering/renderers.rst
```diff
@@ -0,0 +1,357 @@
...
+
++-------------------------+--------------------------+--------------------------+--------------------------+
+| Feature                 | Compatibility            | Mobile                   | Forward+                 |
++=========================+==========================+==========================+==========================+
+| Screen texture          | ✔️ Supported.            | ✔️ Supported.            | ✔️ Supported.            |
++-------------------------+--------------------------+--------------------------+--------------------------+
+| Depth texture           | ✔️ Supported.            | ✔️ Supported.            | ✔️ Supported.            |
+|                         |                          |                          |                          |
++-------------------------+--------------------------+--------------------------+--------------------------+
+| Normal/Roughness texture| ❌ Not supported.        | ✔️ Supported.            | ✔️ Supported.            |
```
clayjohn, line 330: This isn't supported in the mobile renderer

tutorials/rendering/renderers.rst
```diff
@@ -0,0 +1,357 @@
...
++-------------------------+--------------------------+--------------------------+--------------------------+
+| Compute shaders         | ❌ Not supported.        | ⚠️ Supported, but most   | ✔️ Supported.            |
```
clayjohn, line 332: The note for mobile is inaccurate. I would instead say:

```
Supported, but comes with a performance penalty on older devices.
```

clayjohn (commented):
Great work! I left a few suggestions, but this is already really good

tutorials/3d/global_illumination/introduction_to_global_illumination.rst
tetrapod00: In the case of global illumination or antialiasing, I think it does make sense to note when a method is available in all renderers, because most methods are not. That *is* a change from what these two pages currently do.

I'll make this change and change the others for consistency.

tutorials/rendering/renderers.rst
```diff
@@ -0,0 +1,357 @@
...
+not listed here, it is available in all renderers, though it may be much faster
+on some renderers. For a list of *all* features in Godot, see :ref:`doc_list_of_features`. 
+
+Hardware with RenderingDevice support is hardware which can run Vulkan, Direct3D
+12, or Metal.
+
+Overall comparison
+~~~~~~~~~~~~~~~~~~
+
++---------------------+--------------------------+--------------------------+--------------------------+
```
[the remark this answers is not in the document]
tetrapod00: That's due to the use of emojis, which annoyingly don't show up as monospaced. The table renders fine when compiled.

tetrapod00:
Suggestions applied. On the `Introduction to global illumination` and `3D antialiasing` pages, now each GI/AA method explicitly lists which renderers it is available on, since most methods are not available in all renderers, and it's worth pointing out when they are rather than leaving it implicit.

mhilbrunner (approved):

mhilbrunner:
Thank you, amazing work! And thank you all for reviewing.

PR #10116, opened 2024-10-19
[Draft] Add manual about usage of Poedit.

Vovkiv:
More projects starts to using gettext system for internationalization support for their games.
And for dealing with gettext from translators perspective - Poedit exists - nice free and open source program for redacting po files. And probably the only a good one program for offline editing around.

The problem is: gettext is more complex then your typical csv or ini or json file and developers in general don't spend much time writing documentation for tools need to translate their games/programs (if they document this process at all). And because of that, most of the time, translator is alone, trying to figure out how to use all of this tools and workflows.

To help with it, I'm proposing to add this manual page that aimed at casuals/newbies in translation world with Poedit program. Developers of games/programs with Godot that uses gettext system could link this manual for their users or use it as base for their own manual. It will cover basics and workflow with Poedit, explain features of gettext (such as comments, updating from pot file, notice for translators, usage of plural editor, etc, etc) in somewhat easy to understand way.

As of now, this manual is... not exactly ready, there still proofreading needed, fixing typos, wording, better screenshots, etc, etc. But I want to know opinions of maintainers/developers if Godot docs is right place for this manual. If they are okay with it, I will continue to work on it!


tetrapod00:
Get a second opinion from an actual maintainer, but I don't think this is in scope for the official docs. We do currently document the usage of an external program in a few places:
- [Configuring an IDE](https://docs.godotengine.org/en/stable/contributing/development/configuring_an_ide/index.html)
- [Configuring an external editor (C#)](https://docs.godotengine.org/en/latest/tutorials/scripting/c_sharp/c_sharp_basics.html#configuring-an-external-editor)
- [Using the Blender ESCN exporter](https://docs.godotengine.org/en/latest/tutorials/assets_pipeline/escn_exporter/index.html) (removed in 4.4)

But those pages are focused on setting up an external program to work *with* Godot, not just using the external program on its own.

Adding a short note about Poedit and linking to a well-maintained external tutorial (or official Poedit docs) is likely the right way to document usage of Poedit.

Vovkiv:
That's the thing that I found: there not much good manuals for Poedit around, especially in text form and especially in different languages and especially aimed and total newbies.

Official documentation is... https://poedit.net/trac/wiki/Doc empty, mostly.

There some sites that provide usage examples, but they either don't focus on totally new users in translations (like my manual tries to) or they hosted on sites with aggressive ads (which I'm not big fan of).

There just seems to be no great choice that would fill this niche. At least, I'm not aware of such manual around.

There of course option, that I could host such documentation myself and possible it could be linked in official docs, but then there might be not much people, for example, try to translate it as other Godot manuals, since Godot has working Weblate and I don't have.

skyace65 (commented):
I'm inclined to agree with Tetrapod. While I understand there might not be a ton of resources elsewhere, this is just too much documentation to add to Godot for a completely separate program that isn't interacting directly with Godot.

PR #10119, opened 2024-10-20
Clarify use of source_color in Shading Language

tetrapod00:
- Rewrite section on using the source_color uniform hint. Content is mostly the same, but is more clear.
- Rename one instance of "shader hint" to "uniform hint". Both terms are only used in one place, and neither term is used in the engine itself. So this solidifies the name as "uniform hints".
- Adds a section header for Uniform hints.

The rest of this section (Uniforms) can use some more work. It needs reorganization, more subheaders, and some rewrites for clarity. That will be a separate PR though.

### DIFF
--- a/tutorials/shaders/shader_reference/shading_language.rst
+++ b/tutorials/shaders/shader_reference/shading_language.rst
@@ -780,7 +780,7 @@ Uniforms
 Passing values to shaders is possible. These are global to the whole shader and
 are called *uniforms*. When a shader is later assigned to a material, the
 uniforms will appear as editable parameters in it. Uniforms can't be written
-from within the shader.
+from within the shader. Any GLSL type except for ``void`` can be a uniform.
 
 .. code-block:: glsl
 
@@ -803,9 +803,12 @@ GDScript:
           in the shader. It must match *exactly* to the name of the uniform in
           the shader or else it will not be recognized.
 
-Any GLSL type except for *void* can be a uniform. Additionally, Godot provides
-optional shader hints to make the compiler understand for what the uniform is
-used, and how the editor should allow users to modify it.
+
+Uniform hints
+~~~~~~~~~~~~~
+
+Godot provides optional uniform hints to make the compiler understand what the
+uniform is used for, and how the editor should allow users to modify it.
 
 .. code-block:: glsl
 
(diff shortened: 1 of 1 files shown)


tutorials/shaders/shader_reference/shading_language.rst
```diff
@@ -816,20 +819,26 @@ used, and how the editor should allow users to modify it.
     uniform vec4 other_color : source_color = vec4(1.0); // Default values go after the hint.
     uniform sampler2D image : source_color;
 
-It's important to understand that textures *that are supplied as color* require
-hints for proper sRGB -> linear conversion (i.e. ``source_color``), as Godot's
-3D engine renders in linear color space. If this is not done, the texture will
-appear washed out.
+.. admonition:: Source Color
```
tetrapod00, line 822: This is the first use of an admonition note with a custom title in the whole docs repo. I think it makes this readable, but if this is against some style guide I can change this to use a subheader or standard note block instead.

tutorials/shaders/shader_reference/shading_language.rst
```diff
@@ -803,9 +803,12 @@ GDScript:
...
 
-Any GLSL type except for *void* can be a uniform. Additionally, Godot provides
-optional shader hints to make the compiler understand for what the uniform is
-used, and how the editor should allow users to modify it.
+
+Uniform hints
+~~~~~~~~~~~~~
+
+Additionally, Godot provides optional uniform hints to make the compiler understand
+what the uniform is used for, and how the editor should allow users to modify it.
```
clayjohn, line 811:
```suggestion
Godot provides optional uniform hints to make the compiler understand
what the uniform is used for, and how the editor should allow users to modify it.
```
Since this is the start of a new section and paragraph

tutorials/shaders/shader_reference/shading_language.rst
clayjohn: Looks good to me

tutorials/shaders/shader_reference/shading_language.rst
```diff
@@ -816,20 +819,26 @@ used, and how the editor should allow users to modify it.
...
-appear washed out.
+.. admonition:: Source Color
 
-.. note::
+    Any texture which contains *sRGB color data* requires a ``source_color`` hint
+    in order to be correctly sampled. This is because Godot renders in linear
+    color space, but some textures contain sRGB color data. If this hint is not
+    used, the texture will appear washed out.
+
+    Albedo and color textures should have a ``source_color`` hint. Normal,
```
clayjohn, line 829:
```suggestion
    Albedo and color textures should typically have a ``source_color`` hint. Normal,
```
I suggest softening this as it isn't always true, it depends on the source image itself. It's surprising, but once in awhile someone uses a texture with linear space colors in it

clayjohn (commented):
Overall this is a great change, I've left a couple small suggestions for your consideration

tetrapod00:
Changes applied, plus one minor change to wrap the paragraphs to ~80 characters.

tutorials/shaders/shader_reference/shading_language.rst
tetrapod00: I'll be trying this out whenever a note block could use a title, then. I don't mind reverting such changes back if it proves to be a problem. I think in most cases, a note with a title should be a subsection, though?

clayjohn: I think so. I suspect that most notes will be fine without a special title

clayjohn (approved):

mhilbrunner (approved):

mhilbrunner:
Thank you!

PR #10122, opened 2024-10-20
Remove references to 3.x minor versions

tetrapod00:
Removes notes about features being introduced in 3.x minor versions. In 4.x, these details are no longer relevant. If the feature has been supported since before 4.0, it does not matter which minor version it was introduced in.

Note that there are some changes to pages which have an `:article_outdated: True` tag. I'm not sure if it's poor form to make minor changes to outdated articles without actually reviewing them?

A few references to 3.x minor version differences/feature introductions are still in the docs, in cases where I wasn't sure removing the reference was correct.


skyace65 (approved):

mhilbrunner:
Thanks!

mhilbrunner:

PR #10143, opened 2024-10-25
Note limitation of shader world coordinates with large world coordinates

tetrapod00:
This came from helping a user use world space coordinates in a `fragment()` shader while using a double precision build. The easy solution of `vec3 world = (INV_VIEW_MATRIX * vec4(VERTEX, 1.0)).xyz;` didn't work, but passing a varying did. It's a niche problem and a niche solution, but I think that's okay on this page.

After confirming that user's problem was fixed by passing a varying, I also tested it myself on a build of https://github.com/roalyr/godot-for-3d-open-worlds, which is a double-precision build and also has some other changes which I don't think affect this. I couldn't find any official double precision builds.

### DIFF
--- a/tutorials/physics/large_world_coordinates.rst
+++ b/tutorials/physics/large_world_coordinates.rst
@@ -229,6 +229,23 @@ some limitations when it comes to 3D rendering precision:
 - :ref:`Triplanar mapping <doc_standard_material_3d_triplanar_mapping>` doesn't
   benefit from increased precision. Materials using triplanar mapping will exhibit
   visible jittering when far away from the world origin.
+- In double-precision builds, world space coordinates in a shader ``fragment()``
+  function can't be reconstructed from view space, for example:
+
+  .. code-block:: glsl
+
+    vec3 world = (INV_VIEW_MATRIX * vec4(VERTEX, 1.0)).xyz;
+    
+  Instead, calculate the world space coordinates in the ``vertex()`` function and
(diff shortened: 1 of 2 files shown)


tutorials/physics/large_world_coordinates.rst
```diff
@@ -229,6 +229,23 @@ some limitations when it comes to 3D rendering precision:
 - :ref:`Triplanar mapping <doc_standard_material_3d_triplanar_mapping>` doesn't
   benefit from increased precision. Materials using triplanar mapping will exhibit
   visible jittering when far away from the world origin.
+- World space coordinates in a shader ``fragment()`` function can't be
+  reconstructed from view space, for example:
```
clayjohn, line 233: This should have a qualifier that it only applies to double precision builds

tetrapod00: Isn't that implied by being on the Large world coordinates page? The whole page is about double precision builds, which are also called "large world coordinates".

clayjohn: Yes, please ignore me. I didn't look at the title of the page

clayjohn (approved):

tutorials/physics/large_world_coordinates.rst
tetrapod00: I specified it anyway. It doesn't hurt and there's a chance that the section is referenced out of context.

skyace65 (approved):

skyace65:
Thanks!

PR #10144, opened 2024-10-25
Add diagram of shader compilation steps to Pipeline Compilations page

tetrapod00:
Followup to https://github.com/godotengine/godot-docs/pull/10007#discussion_r1779208789.

Diagram made with draw.io. Colors were picked from existing docs theme. Font is Noto Sans.

Export process:
- Exported as SVG from draw.io.
- Imported to inkscape, then exported as webp.
- Imported to krita, and exported once again as webp to optimize for filesize.

There might be a better process, but the default webp and png exports from draw.io are bad.

PNG preview of diagram (exported directly from draw.io, so the font rendering is worse than the actual webp):
shader_compilation_png

PNG alternate, with colors picked from the editor instead of the docs. It's not bad, but it's dull.
shader_compilation_editor_colors

Source for diagram: 
[shader_compilation.drawio.xml.zip](https://github.com/user-attachments/files/17528220/shader_compilation.drawio.xml.zip)

Aesthetic feedback on these diagrams is also appreciated.


tetrapod00:
Oh, since the diagram uses "GDShader" this is waiting on an outcome for https://github.com/godotengine/godot/pull/97298, too.
I can make a variant on the diagram that uses "Godot Shading Language" instead.

clayjohn (approved):

mhilbrunner (approved):

tetrapod00:
Rebased.

PR #10158, opened 2024-10-28
Update Sphinx and requirements

Piralein:
supersede #7443

- Update CI OS version for sphinx 8.x support (I guess phyton version?)
- disable social cards, because of warnings
- remove `html_theme_path` and `display_version` as this are outdated options
- remove search css, as this shouldn't be required anymore
- remove cache busters, as this is now integrated
- remove pinned dependencies
- update dependencies

- `class_scripteditor.rst: document is referenced in multiple toctrees: ['classes/index', 'classes/index'], selecting: classes/index <- classes/class_scripteditor` is an information and can't be suppressed, like a warning. (as far as I know)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ concurrency:
 
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - name: Checkout
         uses: actions/checkout@v4
--- a/_static/css/algolia.css
+++ b/_static/css/algolia.css
@@ -1,6 +1,0 @@
-.wy-nav-side { overflow: visible; }
-.wy-side-scroll { overflow-x: inherit; }
-
-.algolia-autocomplete {
-    display: block !important;
-}
--- a/conf.py
+++ b/conf.py
@@ -9,7 +9,7 @@ import os
 
 # -- General configuration ------------------------------------------------
 
-needs_sphinx = "1.3"
+needs_sphinx = "8.1"
 
 # Sphinx extension module names and templates location
 sys.path.append(os.path.abspath("_extensions"))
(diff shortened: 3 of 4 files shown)


mhilbrunner:
I've been testing this these past few days and had no issues, so I think we can go ahead with this, considering the current trouble with the outdated build we currently use. Both the web/HTML build and the EPUB one seem to work.

> class_scripteditor.rst: document is referenced in multiple toctrees: ['classes/index', 'classes/index'], selecting: classes/index <- classes/class_scripteditor is an information and can't be suppressed, like a warning. (as far as I know)

The [discussion](https://github.com/orgs/sphinx-doc/discussions/12999) on the Sphinx 8.1 release at first made me think this was just an error on the Sphinx side (see this issue: https://github.com/sphinx-doc/sphinx/issues/13012). And to quote:
>  for now, the lines are safe to ignore -- they're informational only. We will see if there is something we can do to improve the experience

However, it felt a bit odd that this only happens to a very limited set of files in the class reference: Editor-only classes, and Node/Resource/Object.

It seems they are the classes that are listed more than once in the class reference index (`classes/index.rst`, autogenerated by [make_rst.py](https://github.com/godotengine/godot/tree/master/doc/tools/make_rst.py) in the main engine repo).

For example, `class_editorcommandpalette` is listed once in the general `Nodes` toctree in that file, and a second time in the `Editor-only` toctree. For now, this only seems to result in a a warning; both generated links in the sidebar appear to work. However, we may need to think about fixing this behaviour for the future.

Piralein:
> I guess for social cards, you mean these warnings?
> 
> > WARNING: [Social card] image cannot be an SVG image, skipping...

Yes, I haven't checked if this is still an issue. I added it at the time and never removed it as we probably don't want them anyway.

mhilbrunner (approved):
It definitely is still an issue. Not sure if wanted or not, I think social cards were at least nice once upon a time for Twitter, but with how that whole thing is going, well. Alright, not that much of a priority anyway, let's go ahead and merge this and see how it pans out for everyone on `master`/`latest` at least. Thanks for your work :)

mhilbrunner:

Issue #1650, opened 2018-08-05
Adding tooltips to the manual


BrahRah:
I highly recommend tooltips! It's much more work but definitely worth it! Anything that is hard to grasp can get a small description, this saves the user a lot of time on searching outside the document and makes the reading process much smoother. The godot docs are full with terms that beginners might not be familiar with, for example: API, VMs, threads, bottlenecks etc.

FeralBytes:
How should they look in Restructered text?

BrahRah:
There are a ton, I'd say every word that is a technical term and not part of the normal spoken language should have such a description. If you mean the Python programming "ReStructuredText" I'm sorry I'm not familiar with that.

Calinou:
This can be done using the `:abbr:` role provided on Sphinx (it's not part of vanilla reStructuredText):

```rst
:abbr:`PHP (PHP Hypertext Preprocessor)`
```

cbscribe:
Closing as there's been no activity for some time. Feel free to open a new issue if you feel the need is still there.

Cerno-b:
@BrahRah @Calinou @FeralBytes In case this ever gets revisited, there is also the option to do this via [hoverxref](https://sphinx-hoverxref.readthedocs.io/en/latest/). That way we could have some kind of glossary in the docs and refer to the text blocks defined in it from the places where terms need to be explained.

I also think this is a good idea, and it could potentially enforce naming discipline. I have been doing quite a lot of translation work in the manual and I encountered multiple places where I wasn't sure whether a term was a new technical term or just a re-wording of a different term.

Generally, the whole 3D-section is full of terminology that could make use of a quick explanation or even just a reminder. Sure, a lot of terms are actually explained in other sections of the manual, and there are links to those places sometimes, but having a glossary tooltip the first time a technical term occurs in a page can be great help for someone who just landed on a single page and who does not want to jump around the manual chasing terminology. 

On the other hand it could be argued that it's not helpful to jump into 3D without getting some overview over terminology, so having tooltips might not really help a lot, which is why I am not creating a new issue. I still think tooltips are useful, but I am not feeling the burn so much as to create a new ticket ;)

PR #10159, opened 2024-10-29
[Draft] Add glossary of common terms

tetrapod00:
Adds a glossary of common terms to the online manual.

Uses the sphinx [glossary](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#glossary) directive.

On its own, this is perhaps a solution in search of a problem. But it's also relatively low-risk and low-cost to add, since it links to other pages, but no other pages depend on it.

With the addition of links *to* the glossary, this could be a much more useful improvement. We can link to these terms with the [term role](https://www.sphinx-doc.org/en/master/usage/referencing.html#role-term). Alternately, we can also integrate [hoverxref](https://sphinx-hoverxref.readthedocs.io/en/latest/) to add tooltips for common terms like acronyms, which link into the glossary.

See also previous discussion in https://github.com/godotengine/godot-docs/issues/1650#issuecomment-1870252031.

**Kinds of terms that will be included in the glossary:**
**Very common nodes or godot concepts:** control, node, viewport, signal, resource
**Godot-specific terms:** renderer, rendering driver, server (in the Godot sense)
**Godot proper nouns:** GDScript, GDExtension, Forward+ renderer, Mobile renderer, Compatibility renderer, RenderingDevice
**Other proper nouns:** Vulkan, Metal, Direct3D 12
**Acronyms, including common gamedev/graphics terms:** CPU, GPU, DOF, FXAA, MSAA, SSAA, SSAO, SSIL, SDFGI, HDR, SDR, sRGB

The current definitions are roughly representative of the final page in terms of length. Each term should be briefly defined, with 1-3 sentences, and link to a relevant portion of the documentation. However, the current *list* of terms is not yet complete.

I'm leaving this as a draft to discuss whether this is something worth including at all, and if so, whether there would need to be major changes to implementation.

The page renders like this with the current theme:
msedge_R4lc0e487K

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -101,6 +101,7 @@ the ``GodotEngine.epub`` file in an e-book reader application.
    tutorials/troubleshooting
    tutorials/editor/index
    tutorials/migrating/index
+   tutorials/glossary
 
    tutorials/2d/index
    tutorials/3d/index
--- a/tutorials/3d/3d_antialiasing.rst
+++ b/tutorials/3d/3d_antialiasing.rst
@@ -38,6 +38,8 @@ detailed below.
     You can compare antialiasing algorithms in action using the
     `3D Antialiasing demo project <https://github.com/godotengine/godot-demo-projects/tree/master/3d/antialiasing>`__.
 
+.. _doc_3d_antialiasing_msaa:
+
 Multisample antialiasing (MSAA)
 -------------------------------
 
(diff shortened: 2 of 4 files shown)


tetrapod00:
I don't think this makes sense in its current state. It would require adding a dependency on hoverxref to actually be useful.

PR #10162, opened 2024-10-29
Mention callback argument count in The JavaScript bridge singleton

Calinou:
- See https://github.com/godotengine/godot-docs-user-notes/discussions/195#discussioncomment-10920548.

### DIFF
--- a/tutorials/platform/web/javascript_bridge.rst
+++ b/tutorials/platform/web/javascript_bridge.rst
@@ -1,6 +1,6 @@
 .. _doc_web_javascript_bridge:
 
-The JavaScriptBridge Singleton
+The JavaScriptBridge singleton
 ==============================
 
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):

mhilbrunner:
Thank you!

mhilbrunner:

Faless:
I missed this.

This is wrong.

Callbacks **always get one argument**, which is the JS [`arguments`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments) array. See the [implementation for reference](https://github.com/godotengine/godot/blob/master/platform/web/js/libs/library_godot_javascript_singleton.js#L230).

That should probably also be mentioned in the `JavaScriptBridge.create_callback` doc.

Issue #10154, opened 2024-10-28
Remove redundant section 3D rendering limitations/MSAA, add any missing information to 3D antialiasing/MSAA


tetrapod00:
**Your Godot version:**
4.4
**Issue description:**
There is a whole section on the 3D rendering limitations page for MSAA, though much of it is spent describing *other* antialiasing techniques. All this information should be folded into the main 3D antialiasing page, either into the MSAA section, or the section comparing different antialiasing methods.
**URL to the documentation page (if already existing):**
Redundant page: https://docs.godotengine.org/en/stable/tutorials/3d/3d_rendering_limitations.html#multi-sample-antialiasing
Main 3D antialiasing page: https://docs.godotengine.org/en/stable/tutorials/3d/3d_antialiasing.html#multisample-antialiasing-msaa

PR #10167, opened 2024-10-29
Add Antialiasing comparison table, remove redundant MSAA limitations section

tetrapod00:
Closes https://github.com/godotengine/godot-docs/issues/10154.

Removes a redundant section on MSAA (which, at the time of writing, apparently functioned as general antialiasing documentation). Adds a new comparison table to the `3D antialiasing` page.

Information in the table is taken from the rest of the page. There's a chance that some of the information in the rest of the page will need to be updated, too.

I don't believe that a redirect or stub section on `3D rendering limitations` is needed in this case.

In the table, ❌✔️ are used for renderer support, and 🔴🟡🟢 are used for comparison of features (because the ⚠️ emoji is not a good match for "medium performance", for example). Table renders like this:
msedge_VuN7vNQxb2


tutorials/3d/3d_antialiasing.rst
```diff
@@ -279,3 +284,36 @@ usually unnecessary, but it can provide better visuals on high-end GPUs or for
...
+| Mobile                   | ✔️ Yes                   | ❌ No                    | ❌ No                    | ✔️ Yes                   | ✔️ Yes                   | ✔️ Yes                   |
++--------------------------+--------------------------+--------------------------+--------------------------+--------------------------+--------------------------+--------------------------+
+| Compatibility            | ✔️ Yes                   | ❌ No                    | ❌ No                    | ❌ No                    | ✔️ Yes                   | ❌ No                    |
++--------------------------+--------------------------+--------------------------+--------------------------+--------------------------+--------------------------+--------------------------+
+
+
+.. [#] MSAA does not work well with materials with Alpha Scissor (1-bit transparency). This can be mitigated by enabling ``alpha antialiasing`` on the material.
+.. [#] TAA transparency antialiasing is most effective when using Alpha Scissor.
+.. [#] FSR2 transparency antialiasing is most effective when using Alpha Scissor.
```
clayjohn, line 318: Is there any way to combine these? I think it would be a bit cleaner if they could share a note. It would also help communicate that the limitation is the same in both, so they are equivalent.

tetrapod00: Yeah, I think the other footnote syntax can do that

clayjohn: I have a slight preference for this alternative

clayjohn (approved):
Looks good to me!

mhilbrunner (approved):

mhilbrunner:
Looks good to me too, and if Clay approves, may as well merge :) Thanks!

PR #10171, opened 2024-10-30
Fix inaccurate/misleading note about shader global arrays

tetrapod00:
This note was originally written before uniform arrays were merged.


mhilbrunner:
Thank you!

PR #10175, opened 2024-10-30
Note that `render_mode wireframe` doesn't work in Compatibility

tetrapod00:
See https://github.com/godotengine/godot/issues/98532, https://github.com/godotengine/godot/issues/98532#issuecomment-2448127668.


clayjohn:
I don't think this is an appropriate note in the documentation. We avoid documenting known bugs because inevitably the documentation will stay outdated much longer than the bug itself impacts users.

tetrapod00:
That makes sense! I thought it was more of a missing feature than a bug, in this case.

PR #10190, opened 2024-11-02
Update navigation_introduction_2d.rst

leiofrivia:
There is no actor_setup.call_deferred() method, so instead it's ok just using actor_setup() function as specified below.


mhilbrunner:
Can you elaborate? `actor_setup()` exists in the code below, and `call_deferred()` is an engine feature that should always be available. It **should** exist.

Also `actor_setup()` uses `await` and as the comment points out, should be called deferred to the next frame when called from `ready()`.

mhilbrunner:
Closing as per the above, ping me if you believe this to be a mistake / have further feedback to the above.

PR #10197, opened 2024-11-03
Improve interpolation smoothing example

tetrapod00:
Supersedes https://github.com/godotengine/godot-docs/pull/6134.

Improves explanation of interpolation smoothing ("lerp smoothing"). 
Adds a note about framerate-independent version, with off-site links to full explanations.

I believe keeping the current, framerate-dependent version is good. It is simpler, and is required context appreciate the framerate-independent version. (I would also *like* to include the truly naive version with no `delta`: `$Sprite2D.position = $Sprite2D.position.lerp(mouse_pos, FOLLOW_SPEED)`. But the explanation would be too long.)

The explanation of the framerate-independent version belongs in a note. The simple version does actually work in most cases, because physics delta is consistent in normal gameplay, more so than process delta. 

Unfortunately, while it's easy enough to explain why just multiplying by `delta` doesn't work, it's hard to concisely explain why the framerate-independent version works. So I also included links to the a short-ish article deriving the better formula, and to a good but very long video on the problem and how to derive the better formula.

Both GDScript and C# examples have been tested and confirmed to work. I did not deeply test the formulas at varying framerates, though.

Note that the form used in this PR:
```
value = value.lerp(target, 1 - exp(-rate* delta))
```
seems different than the form used in [the article](https://www.gamedeveloper.com/programming/improved-lerp-smoothing-) (after changing the names and `lerp()` function signature): 
```
value = target.lerp(value, exp(-rate*delta))
```
It's equivalent if you flip the `target` and `value`, and compensate with a `1 - x` though:
```
value = value.lerp(target, 1 - exp(-rate*delta))
```

Note also that we use `exp()` in place of `exp2()`. This is still valid; only the `rate` value needs to change to compensate for the change in exponential base.

### DIFF
--- a/tutorials/math/interpolation.rst
+++ b/tutorials/math/interpolation.rst
@@ -3,7 +3,10 @@
 Interpolation
 =============
 
-Interpolation is a very basic operation in graphics programming. It's good to become familiar with it in order to expand your horizons as a graphics developer.
+Interpolation is a common operation in graphics programming, which is used to
+blend or transition between two values. Interpolation can also be used to smooth
+movement, rotation, etc. It's good to become familiar with it in order to expand
+your horizons as a game developer.
 
 The basic idea is that you want to transition from A to B. A value ``t``, represents the states in-between.
 
@@ -109,7 +112,11 @@ And again, it will produce the following motion:
 Smoothing motion
 ----------------
 
-Interpolation can be used to smooth movement, rotation, etc. Here is an example of a circle following the mouse using smoothed motion:
+Interpolation can be used to smoothly follow a moving target value, such as a
+position or a rotation. Each frame, ``lerp()`` moves the current value towards
+the target value by a fixed percentage of the remaining difference between the values.
+The current value will smoothly move towards the target, slowing down as it gets
+closer. Here is an example of a circle following the mouse using interpolation smoothing:
 
 .. tabs::
  .. code-tab:: gdscript GDScript
@@ -138,4 +145,42 @@ Here is how it looks:
 
 .. image:: img/interpolation_follow.gif
 
-This is useful for smoothing camera movement, allies following you (ensuring they stay within a certain range), and many other common game patterns.
+This is useful for smoothing camera movement, for allies following the player
+(ensuring they stay within a certain range), and for many other common game patterns.
+
+.. note::
+    Despite using ``delta``, the formula used above is framerate-dependent, because
+    the ``weight`` parameter of ``lerp()`` represents a *percentage* of the remaining
+    difference in values, not an *absolute amount to change*. In ``_physics_process()``,
+    this is usually fine because physics is expected to maintain a constant framerate,
+    and therefore ``delta`` is expected to remain constant.
+
+    For a framerate-independent version of interpolation smoothing that can also
+    be used in ``process()``, use the following formula instead:
+
+    .. tabs::
+        .. code-tab:: gdscript GDScript
+
+            const FOLLOW_SPEED = 4.0
+
+            func _process(delta):
+                var mouse_pos = get_local_mouse_position()
+                var weight = 1 - exp(-FOLLOW_SPEED * delta)
+                $Sprite2D.position = $Sprite2D.position.lerp(mouse_pos, weight)
+
+        .. code-tab:: csharp
+
+            private const float FollowSpeed = 4.0f;
+
+            public override void _Process(double delta)
+            {
+                Vector2 mousePos = GetLocalMousePosition();
+
+                Sprite2D sprite = GetNode<Sprite2D>("Sprite2D");
+                float weight = 1f - Mathf.Exp(-FollowSpeed * (float)delta);
+                sprite.Position = sprite.Position.Lerp(mousePos, weight);
+            }
+    
+    Deriving this formula is beyond the scope of this page. For an explanation, 
+    see `Improved Lerp Smoothing <https://www.gamedeveloper.com/programming/improved-lerp-smoothing->`__
+    or watch `Lerp smoothing is broken <https://www.youtube.com/watch?v=LSNQuFEDOyQ>`__.


tutorials/math/interpolation.rst
```diff
@@ -109,7 +109,11 @@ And again, it will produce the following motion:
 Smoothing motion
 ----------------
 
-Interpolation can be used to smooth movement, rotation, etc. Here is an example of a circle following the mouse using smoothed motion:
+Interpolation can be used to smoothly follow a moving target value, such as a
+position or a rotation. Each frame, ``lerp()`` moves the current value towards
+the target value by a fixed percentage of the remaining difference between the values.
```
tetrapod00, line 114: "Percentage" is technically incorrect here, since the value is in the range of `[0,1]` rather than `[0,100]`. "Portion" or maybe "fraction" is more technically correct, but have the wrong connotations - in English, "portion" can also mean a fixed amount rather, rather than a value that depends on the total value.

tutorials/math/interpolation.rst
```diff
@@ -138,4 +142,48 @@ Here is how it looks:
...
+(ensuring they stay within a certain range), and for many other common game patterns.
+
+.. note::
+ Despite using ``delta``, the formula used above is framerate-dependent, because
+ the ``weight`` parameter of ``lerp()`` represents a *percentage* of the remaining
+ difference in values, not an *absolute amount to change*. In ``_physics_process()``,
+ this is usually fine because physics is expected to maintain a constant framerate,
+ and therefore ``delta`` is expected to remain constant. For example, with an FPS
+ of ``60`` and a ``FOLLOW_SPEED`` of ``4.0``, ``delta`` will be ``1/60`` or ``0.0166``,
+ and ``FOLLOW_SPEED * delta`` will be a constant value of ``0.0666``.
```
tetrapod00, line 155: This last sentence feels perhaps unecessary, but I wanted to illustrate that even with `delta` involved, the value `FOLLOW_SPEED * delta` tends to be constant.

tutorials/math/interpolation.rst
```diff
@@ -138,4 +142,48 @@ Here is how it looks:
...
 .. image:: img/interpolation_follow.gif
 
-This is useful for smoothing camera movement, allies following you (ensuring they stay within a certain range), and many other common game patterns.
+This is useful for smoothing camera movement, for allies following the player
+(ensuring they stay within a certain range), and for many other common game patterns.
+
+.. note::
+ Despite using ``delta``, the formula used above is framerate-dependent, because
+ the ``weight`` parameter of ``lerp()`` represents a *percentage* of the remaining
+ difference in values, not an *absolute amount to change*. In ``_physics_process()``,
```
tetrapod00, line 151: If it's not clear from a single sentence, explaining why `weight` being a percentage rather than an absolute amount matters feels beyond the scope of the docs. There are two good links to get a deeper explanation, if the reader wants them.

tutorials/math/interpolation.rst
mhilbrunner: I think this is fine as is, I agree with the reasoning.

tutorials/math/interpolation.rst
mhilbrunner: I think this is fine.

tutorials/math/interpolation.rst
mhilbrunner: I'd lean towards removing the last sentence here. Examples help when they provide a simple illustration of a complex concept, here, `In _physics_process() this is usually fine because physics is expected to maintain a constant framerate` is easier for me to parse and conceptualize than the example with all those numbers.

tutorials/math/interpolation.rst
tetrapod00: I'd be happy to do a pass on the rest of the page too, but I think that's slightly out of scope of this PR which just aims to improve the section on using lerp() to follow a moving target (or "lerp smoothing").

mhilbrunner: In that case I'm fine with this as is, too. That example is better for the beginning rather halfway down the page, anyway.

aXu-AP:
LGTM, keeps it simple while explaining the topic with enough depth.

tutorials/math/interpolation.rst
```diff
@@ -138,4 +142,48 @@ Here is how it looks:
 
 .. image:: img/interpolation_follow.gif
 
-This is useful for smoothing camera movement, allies following you (ensuring they stay within a certain range), and many other common game patterns.
+This is useful for smoothing camera movement, for allies following the player
+(ensuring they stay within a certain range), and for many other common game patterns.
+
+.. note::
```
AThousandShips, line 148: I'd say this whole section in the note needs to be further indented or it's nearly impossible to tell what is in the note or not, I'd say to make it at least 4 spaces

tetrapod00: Agree, it was bothering me as I wrote it. The whole page might need to change, I'm not sure if RST/Sphinx handles pages with mixed spaces per indentation well.

tetrapod00: Ah, it looks like I was doing something wrong last time I tried using mixed spaces per indentation in a single page. It compiles fine for me now.

Just to be clear, your recommendation is 4 spaces per indentation like this?
```
.. note::
    Paragraph of text.

    .. tabs::    

            code

        .. code-tab:: csharp

            code {}

    Paragraph of text.
```

Not 2 spaces per indentation, like this?
```
.. note::
  Paragraph of text.

  .. tabs::    

        code

    .. code-tab:: csharp

        code {}

  Paragraph of text.

```

AThousandShips: I'd say either goes, 3 might be good too to line up with the "note" text, but it should be more than 1

tetrapod00: It's definitely spaces, not tabs, in the manual.

3 seems wrong. Most pages use either 2 or 4. This page is an exception in using 1.

As a rule I prefer 4 spaces per indentation because it makes the page more legible, but I believe 2 may be chosen in some cases because it minimizes total line length?

AThousandShips: I'd say 4 then, to make this unusually long note less confusing

tutorials/math/interpolation.rst
tetrapod00: I made a change to the very first paragraph:
- bringing the removed sentence to the beginning of the page
- removing one of the banned words "basic"
- changing "graphics developer" to "game developer"

To be clear, the rest of this page still needs more work, but it's beyond the scope of this PR.

tutorials/math/interpolation.rst
tetrapod00: Removed.

mhilbrunner (approved):

tutorials/math/interpolation.rst
```diff
@@ -138,4 +145,42 @@ Here is how it looks:
...
+            {
+                Vector2 mousePos = GetLocalMousePosition();
+
+                Sprite2D sprite = GetNode<Sprite2D>("Sprite2D");
+                float weight = 1f - Mathf.Exp(-FollowSpeed * (float)delta);
+                sprite.Position = sprite.Position.Lerp(mousePos, weight);
+            }
+    
+    Deriving this formula is beyond the scope of this page. For an explanation, 
+    read `Improved Lerp Smoothing <https://www.gamedeveloper.com/programming/improved-lerp-smoothing->`__
```
AThousandShips, line 185:
```suggestion
    see `Improved Lerp Smoothing <https://www.gamedeveloper.com/programming/improved-lerp-smoothing->`__
```
It seems that the imperative "read" is seen as more "aggressive" than "see"

AThousandShips (approved):

tutorials/math/interpolation.rst
tetrapod00: I wanted to emphasize that one is text and the other is video. Is there another good way to do that? Maybe "see the article Improved Lerp Smoothing"?

AThousandShips: I think it should be obvious, but maybe:
```suggestion
    see the `Improved Lerp Smoothing article <https://www.gamedeveloper.com/programming/improved-lerp-smoothing->`__
```

tetrapod00: After looking again, your original suggestion was fine, actually. The alternatives are all more awkward

mhilbrunner:
Thank you!

PR #10207, opened 2024-11-05
Offline docs: actions/checkout@v4 migration

mhilbrunner:
As of the `v4` of the `actions/checkout` Github action, artifacts can't be overwritten by default anymore, so we need to enable that manually.

`if-no-files-found: error` makes sure we error if the file we want (and expect) to upload isn't found. By default, only a warning is generated.

See: https://github.com/actions/upload-artifact?tab=readme-ov-file#usage

This should be cherry-picked to all branches https://github.com/godotengine/godot-docs/pull/8627 has been cherry-picked to (all branches where the Github action was updated to `v4`).


mhilbrunner:
This needs further investigation on my side.

PR #10208, opened 2024-11-05
CI: Skip class reference sync and offline docs on forks

mhilbrunner:
Currently, we automatically run scheduled Github actions to sync the class reference for `master` and to build offline documentation for `master`,  `stable` and `3.6`.

However, this also happens on all forks by default if they have actions enabled, which generates unnecessary notification spam and wastes computing resources needlessly; the large majority of forks are docs contributors unlikely to have any interest in either of these.

With this, scheduled runs are skipped by default on forks. They can be re-enabled by setting repository variables. Manual runs can be triggered as normal and remain unaffected.

### DIFF
--- a/.github/workflows/build_offline_docs.yml
+++ b/.github/workflows/build_offline_docs.yml
@@ -8,6 +8,9 @@ on:
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


mhilbrunner:

PR #10209, opened 2024-11-05
I18N: Remove canonical/hreflang tags, let RTD handle SEO/locale negotiation

mhilbrunner:
This was originally added as a stopgap (by me) and turned out to be less than satisfactory. In the mean time, RTD now generates sitemaps to handle this (compare https://docs.godotengine.org/sitemap.xml), and Google actually considers our generated links here broken due to no proper backlinks if you check via their Webmaster tooling.

Removing this should hopefully mean Google uses the Sitemap and the `<html lang="">` tag to find links between alternate language versions, and prefer to show users the pages meant for their configured preferences for language/locale.

If that turns out not enough, we need to revisit this, but fix the implementation to provide proper backlinks for all languages.

### DIFF
--- a/_templates/layout.html
+++ b/_templates/layout.html
@@ -13,16 +13,6 @@
 
 {% block linktags -%}
   <meta name="theme-color" content="#3d8fcc" />
-  {% if godot_inject_language_links -%}
-  {% for alternate_lang in godot_docs_supported_languages -%}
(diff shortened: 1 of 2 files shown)


mhilbrunner:
Let's see how this works in practice.

mhilbrunner:

PR #10219, opened 2024-11-07
Improve Bug triage guidelines

tetrapod00:
A couple small improvements.
- Link to the up-to-date list of labels on github itself.
- Update some of the labels to match *observed behavior* of the usage of these labels. I could be wrong, though!
    - In the main repo, the `documentation` tag is not really used for documentation improvement requests (from users). Instead it is used for issues that are reported as bugs, but are actually limitations that can be (further) documented.
    - Similarly, the `feature request` label is no longer used much for feature request issues (since those are closed and turned into proposals), but it is used to tag PRs that do not have a corresponding proposal.
    - The `needs work` label is used sometimes for very incomplete bug reports, as opposed to the `needs testing` label for more complete reports that just need to be tested.

### DIFF
--- a/contributing/workflow/bug_triage_guidelines.rst
+++ b/contributing/workflow/bug_triage_guidelines.rst
@@ -34,7 +34,8 @@ to both issues and pull requests.
 Labels
 ~~~~~~
 
-The following labels are currently defined in the Godot repository:
+The following `labels <https://github.com/godotengine/godot/labels>`__ are
+currently defined in the Godot repository:
 
 **Categories:**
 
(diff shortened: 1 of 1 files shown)


contributing/workflow/bug_triage_guidelines.rst
```diff
@@ -58,16 +59,19 @@ The following labels are currently defined in the Godot repository:
 -  *Discussion*: the issue is not consensual and needs further
    discussion to define what exactly should be done to address the
    topic.
--  *Documentation*: issue related to the documentation. Mainly to request
-   enhancements in the API documentation. Issues related to the ReadTheDocs
-   documentation should be filed on the
+-  *Documentation*: related to the documentation. PRs with this label improve the
+   class reference. Issues with this label are not bugs, but limitations to be
+   further documented. Often paired with *Discussion*. Issues related to the
```
mhilbrunner, line 64: We actually consider wrong documentation bugs too, and those issues do get labeled both with "bug" and "documentation", so "issues with this label are not bugs" isn't entirely technically correct.

mhilbrunner (approved):

mhilbrunner:
Thanks!

mhilbrunner:

PR #10220, opened 2024-11-07
Fix codespell not running on some files

tetrapod00:
Codespell was not running on some files.

The current command should check multiple directory levels, but it does not:
```
codespell -I _tools/codespell-ignore.txt -x _tools/codespell-ignore-lines.txt -S tutorials/i18n/locales.rst {about,community,contributing,getting_started,tutorials}/**/*.rst
```
 It was only checking `.rst` files that were one directory below the top-level directories (`about`, `community`, `contributing`, etc). It was not checking direct children of the top-level directories, nor was it checking children that were more deeply nested within subdirectories. This means that the entire `about` section was not being checked at all, since all the pages there are direct children of top-level folders.

This updated command is a crude way to check multiple levels. There should be a better way to do this, but for some reason that way is *not* just using a single `/**/`.
```
codespell -I _tools/codespell-ignore.txt -x _tools/codespell-ignore-lines.txt -S tutorials/i18n/locales.rst {about,community,contributing,getting_started,tutorials}/{*.rst,**/*.rst,**/**/*.rst,**/**/**/*.rst}
```

One uncaught typo was found with the updated command.

If you want to test out this updated codespell command, or a better alternate, try putting a typo like "1nd" in one file in each folder nesting level, then run the command (either locally or in CI).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Linter checks
         run: |
           bash _tools/format.sh
-          codespell -I _tools/codespell-ignore.txt -x _tools/codespell-ignore-lines.txt -S tutorials/i18n/locales.rst {about,community,contributing,getting_started,tutorials}/**/*.rst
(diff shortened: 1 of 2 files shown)


mhilbrunner (approved):

mhilbrunner:
Good catch, thank you.

mhilbrunner:

Issue #10218, opened 2024-11-06
Standardize usage of compound words, and document standards in docs style guide


tetrapod00:
**Your Godot version:**
4.4
**Issue description:**
There are some compound words, like "antialiasing", that appear in the docs in both hyphenated form "anti-aliasing" and in single-word form "antialiasing". For consistent style, we should choose one form and standardize.

Some examples:
- antialiasing / anti-aliasing 
- runtime / run-time
- screen space / screen-space

This is a fairly low priority issue. We should not break translations just to change this. However, we should choose which form to use for each compound word, document that choice in the style guide, and change existing usages whenever other changes are being made.

**URL to the documentation page (if already existing):**
The whole manual and class reference.

Mickeon:
Bit of a side note, we've broken translations [for much less](https://github.com/godotengine/godot/pull/89036). I think going into a big pass for this is somewhat justified. I find consistency to be very important even for subtle details like these.

tetrapod00:
I've been doing the manual changes one word at a time in multiple individual PRs. When that's done, maybe we can collect them and do the class reference changes in a single big PR?

Mickeon:
Sounds fine by me.

betalars:
Controversial take, but I think I like anti-aliasing a lot more than antialiasing.

Both forms are grammatically fine and the form with the hyphen is more accessible.

Where both forms are viable, I strongly suggest using the more accessible option.

tetrapod00:
> Controversial take, but I think I like anti-aliasing a lot more than antialiasing.

This one has already been resolved (at least in the manual), so if we decided to change it again we should probably get some really formal guidelines written down (or pick an existing style guide) so we can justify the changes against hard rules, rather than just trying to pick whichever is more common.

> Where both forms are viable, I strongly suggest using the more accessible option.

When resolving these, I'm been standardizing on whichever form is more common already. Usually one form is overwhelmingly more common. When things are close, I've been erring on the side of *more compact*, so "antialiasing" is preferred over "anti-aliasing", and "anti-aliasing" is preferred over "anti aliasing". As far as I know, the general historical trend for an individual compound word is to become more compact over time.

Instead, if existing usage is mixed, we could decide in favor of the more accessible form. Presumably this means preferring hyphenation over single words. Between hyphenation and spaced multiple words, which is more accessible? I couldn't find any best practices on that with a quick search.

betalars:
I think going with the more widely-used form is a sensible idea tho.

[...]

PR #10221, opened 2024-11-07
Enforce style guidelines with custom codespell dictionary

tetrapod00:
Adds a custom dictionary to codespell. Both the default codespell dictionary and our custom dictionary are checked, using multiple copies of the `-D` argument.

Currently, this dictionary only contains "anti-aliasing -> antialiasing". Later, we can add other style guidelines that we want to enforce, like perhaps "run-time -> runtime". https://github.com/godotengine/godot-docs/issues/10218 tracks potential words to add.

Note that codespell is limited here. We can't check for phrases with spaces or for specific capitalizations. We also can't check for phrases which we don't want to enforce a style for 99%+ of the time, since each exception needs to be listed in an ignore file. So the number of entries in the custom dictionary will likely remain low.

---
Addresses https://github.com/godotengine/godot-docs/issues/10218, but does not close.
This PR is a pure superset of https://github.com/godotengine/godot-docs/pull/10220. It uses the same fix for codespell not checking some files. It also includes the one typo found by the other PR, so that the CI passes. However, I made two separate PRs because we probably want to merge the simple bug fix before this feature change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,8 @@ jobs:
       - name: Linter checks
         run: |
           bash _tools/format.sh
-          codespell -I _tools/codespell-ignore.txt -x _tools/codespell-ignore-lines.txt -S tutorials/i18n/locales.rst {about,community,contributing,getting_started,tutorials}/{*.rst,**/*.rst,**/**/*.rst,**/**/**/*.rst}
+
+          codespell -D- -D _tools/codespell-dict.txt -I _tools/codespell-ignore.txt -x _tools/codespell-ignore-lines.txt -S tutorials/i18n/locales.rst {about,community,contributing,getting_started,tutorials}/{*.rst,**/*.rst,**/**/*.rst,**/**/**/*.rst}
 
       # Use dummy builder to improve performance as we don't need the generated HTML in this workflow.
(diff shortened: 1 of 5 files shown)


mhilbrunner:
@tetrapod00 Reviewed and merged the other one first, so this now needs a rebase :) Looks good to me though.

tetrapod00:
Are you sure this one is a cherrypick all the way back? Unless I'm misunderstanding, won't the CI start complaining about some usages of "anti-aliasing" in old docs versions?

mhilbrunner:
Yeah, this would definitely also need to fix those occurences in older branches. Applying the labels is more of a suggestion to myself that this in theory does apply to old branches as well, and there is something to be said to have the "meta" stuff like Github Actions, CI, build system and ReadTheDocs stuff stay mostly in sync so we don't end up with 10 different setups eventually.

When actually doing cherry-picks, there is still some consideration that goes into what is actually cherry-picked. I'm not 100% decided on this one, and hey, now we got this comment trail to remind me to revisit that discussion :)

Thanks for the rebase.

mhilbrunner (approved):

mhilbrunner:
In this goes 🎉 Thank you!

mhilbrunner:

PR #10222, opened 2024-11-07
Change all usages of "run-time" to "runtime"

tetrapod00:
Addresses https://github.com/godotengine/godot-docs/issues/10218.

Changes all instances (in the manual) of "run-time" to "runtime". Prior to this change, "run-time" had ~50 usages in the manual, and "runtime" had ~250. The class ref has a similar preference for "runtime".

One exception: a link to the demo project `Run-time File Saving and Loading` stays as it is, because it's the name of a project and should be searchable in the asset library.

Not sure if changing all instances at once is the right way to go about this. There is definitely precedent for it.

### DIFF
--- a/about/faq.rst
+++ b/about/faq.rst
@@ -567,7 +567,7 @@ resulted in better usability while still being fast enough for most use cases.
 
 That said, nothing prevents you from making use of composition in your project
 by creating child Nodes with individual scripts. These nodes can then be added and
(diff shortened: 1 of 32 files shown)


mhilbrunner (approved):

mhilbrunner:
Yup, and makes sense. Thank you!

PR #10223, opened 2024-11-07
Tweak SSRL performance cost in 3D antialiasing comparison table

Calinou:
- Follow-up to https://github.com/godotengine/godot-docs/pull/10167.

The screen-space roughness limiter's performance cost is very low compared to other antialiasing methods (typically in the same ballpark as FXAA).


clayjohn (approved):
That's right. I would say even lower than FXAA

tetrapod00:
Thanks. The current SSRL body text also calls the cost "small", so the table now matches the current description.

(The cost was called "moderate" in the removed MSAA limitations section, so I probably used that as the source for "medium cost" instead of the more up-to-date info in the AA page itself)

mhilbrunner (approved):

mhilbrunner:
Thanks!

PR #10227, opened 2024-11-07
Improve GDScript format strings tutorial

tetrapod00:
- Remove usages of "integral". The adjective "integral" is pretty much only used on this page. While "integral number" is technically more correct, "integer" is a much more commonly understood word. The rest of the docs already uses "integer number" or "integer value" for similar cases.
- Add section on string concatenation. Since this page already compares two ways of modifying strings, this one should be mentioned too.
- Improve organization. Information about the ``format()`` function was interleaved in a confusing way. I extracted it out to its own section. The introduction to format strings was also simplified.
- Fix incorrect examples. A couple examples in the ``format()`` example table did not match their listed output.

Arguably this page should also be renamed, but I think that will break things?

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_format_string.rst
+++ b/tutorials/scripting/gdscript/gdscript_format_string.rst
@@ -3,21 +3,24 @@
 GDScript format strings
 =======================
 
-GDScript offers a feature called *format strings*, which allows reusing text
-templates to succinctly create different but similar strings.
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):

mhilbrunner:
Thanks! Merged.

mhilbrunner:

PR #10240, opened 2024-11-10
Rename Renderers page to List of renderers

Calinou:
- Follow-up to https://github.com/godotengine/godot-docs/pull/10114.

___

- Put it first in the list of pages in the Rendering section, as it makes more sense to be read first.

### DIFF
--- a/tutorials/rendering/index.rst
+++ b/tutorials/rendering/index.rst
@@ -12,8 +12,8 @@ Rendering
    :maxdepth: 1
    :name: toc-learn-features-rendering
 
+   renderers
    viewports
    multiple_resolutions
    jitter_stutter
    compositor
-   renderers
--- a/tutorials/rendering/renderers.rst
+++ b/tutorials/rendering/renderers.rst
@@ -1,7 +1,7 @@
 .. _doc_renderers:
 
-Renderers
-=========
+List of renderers
+=================
 
 .. seealso::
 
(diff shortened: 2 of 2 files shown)


tetrapod00 (approved):
This seems good, but I want to clarify why the page is called `renderers.rst`. I initially wanted to write a page with just the comparison table, but saw that there was a lack of information about the renderers (besides the technical `Internal Rendering Architecture` page). Over time I believe that the `Renderers` page will grow like this:

1. Add the page (we are here):
- Renderers **page**
  - Intro **section**
  - Comparison table **section**

2. Add info about each renderer to a *section*, as we add or transfer information about renderers:
- Renderers **page**
  - Intro
  - Forward+ **section**
  - Mobile **section**
  - Compatibility **section**
  - Comparison

3. The section on each renderer becomes a page (this is as far as I'm pretty sure we'll get):
- Renderers **index page** (this one still has the `doc_renderers` anchor despite all the changes)
  - Intro + Comparison **page**
  - Forward+ **page**
  - Mobile **page**
  - Compatibility **page**

4. Potentially add some other pages which are about *renderers* but don't fit in any existing page, and split out the table to a full page:
- Renderers **index page**
  - Intro
  - Forward+
  - Mobile
  - Compatibility
  - Supporting multiple renderers **page**
  - Comparison of renderers **page**

So I named the page `rendererers.rst` and "Renderers", rather than "Renderer comparison" or "List of renderers", in anticipation of this. I think "List of renderers" or "Renderer comparison" is the better name for the page as it currently exists, so this change seems good, as long as the anchor and filename are left as they are and the page title can be renamed again later if needed.

mhilbrunner (approved):

skyace65:
What is the issue with the current name? The page isn't just a list of renderers, it gives information on which is best for your needs and a breakdown of what features are supported on which. I just don't see much of a point with this change.

mhilbrunner:
Yeah, I don't disagree. I'm fine with the change (and tetrapod seems too), but I'm not sure "Renderers" isn't fine as is.

mhilbrunner:
I think with all three of us not really against it, but also not seeing much of a point, I'll close this for now. If you disagree though, feel free to ping for a reopen Calinou, and we can discuss :)

Issue #10204, opened 2024-11-04
Document writing shaders that work in multiple renderers


tetrapod00:
**Your Godot version:**
4.4
**Issue description:**
Since https://github.com/godotengine/godot/pull/97142 allowed falling back from Vulkan/D3D12/Metal to OpenGL, it is now more expected that rendering, including shaders, works on multiple renderers.

We should document writing shaders that work in multiple renderers, using `OUTPUT_SRGB`, `CLIP_SPACE_FAR`, and/or `CURRENT_RENDERER` (once https://github.com/godotengine/godot/pull/98549 is merged).

This is currently half-documented in two places:
- The NDC change from always using OpenGL NDC in 3.x to sometimes using Vulkan NDC in 4.x is documented in https://docs.godotengine.org/en/latest/tutorials/migrating/upgrading_to_godot_4.html#updating-shaders.
- The existence of Vulkan NDC is mentioned in https://docs.godotengine.org/en/latest/tutorials/shaders/advanced_postprocessing.html, but that page does not describe how to write a shader that works with Compatibility which uses OpenGL NDC.

Changes to make:
- Advanced postprocessing should include a snippet for using `CLIP_SPACE_FAR` to create correct NDCs for all renderers.
- Possibly, some shader page should include a general section on writing shaders that work in all renderers. Possibly, the Renderers page should link to that shader section.

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/latest/tutorials/migrating/upgrading_to_godot_4.html#updating-shaders
https://docs.godotengine.org/en/latest/tutorials/shaders/advanced_postprocessing.html

PR #10244, opened 2024-11-11
Add multiple renderer support to Advanced Postprocessing

tetrapod00:
Addresses https://github.com/godotengine/godot-docs/issues/10204.

Adds code for the Compatibility renderer to the Advanced Postprocessing page.

With https://github.com/godotengine/godot/pull/98549 merged, there is now a way to detect the renderer in shaders that has low no runtime cost. Unlike using `OUTPUT_IS_SRGB`, it also is obvious what it does. Alternately, `CLIP_SPACE_FAR` could be used in the example here, since NDC depends directly on that value. But the math to do so is non-obvious, and using a renderer define here is more obvious and instructive.

### DIFF
--- a/tutorials/shaders/advanced_postprocessing.rst
+++ b/tutorials/shaders/advanced_postprocessing.rst
@@ -108,10 +108,6 @@ from ``0.0`` to ``1.0`` in the ``z`` direction when using the Vulkan backend.
 Reconstruct the NDC using ``SCREEN_UV`` for the ``x`` and ``y`` axis, and
 the depth value for ``z``.
 
-.. note::
-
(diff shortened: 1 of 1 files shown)


Calinou (approved):
Looks good to me.

**Note:** Not cherry-pickable to `4.3` as this feature is only in 4.4.

tetrapod00:
I can make a cherrypickable version that either 
- does not mention supporting multiple renderers, only adds the Compatibility NDC snippet
- uses the jankier `OUTPUT_IS_SRGB` and runtime if() to detect the renderer

mhilbrunner (approved):

mhilbrunner:
Thank you!

Issue #9694, opened 2024-08-01
LightMapGI node only bakes light data for child nodes of its parent, but that is not mentioned in the documentation.


Arnklit:
**Your Godot version:**
4.2.2 and 4.3

**Issue description:**
LightMapGI node only bakes light data for child nodes of its parent, but that is not mentioned in the documentation.
Another note should probably be added:

```Note: The LightmapGI node only bakes light data for child nodes of its parent. Nodes further up the hierarchy of the scene will not be baked.```

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/latest/classes/class_lightmapgi.html

skyace65:
This also should be mentioned in the manual page here: https://docs.godotengine.org/en/stable/tutorials/3d/global_illumination/using_lightmap_gi.html

skyace65:
Fixed by #10247 and https://github.com/godotengine/godot/pull/99079

PR #10247, opened 2024-11-11
Add note about LightmapGI only baking nodes under its parent

FlooferLand:
`LightmapGI`'s documentation currently doesn't mention the fact it only bakes nodes under its parent. 
This can make the user think there is something wrong with their scene setup or 3D models, as it refuses to bake when the user's models / world isn't under the same parent as the `LightmapGI`.

I personally encountered this issue because I tried adding a `LightmapGI` underneath a blank node called "Environment" under which a `WorldEnvironment` was also present, and the `LightmapGI` complained about having nothing to bake.

Made for the #9694 issue
Linked class reference PR: [godotengine/godot#99079](https://github.com/godotengine/godot/pull/99079)

### DIFF
--- a/tutorials/3d/global_illumination/using_lightmap_gi.rst
+++ b/tutorials/3d/global_illumination/using_lightmap_gi.rst
@@ -82,8 +82,10 @@ Setting up
 
 .. note::
 
-    The LightmapGI node only bakes nodes under its parent.
-    This is because some people may require several separate lightmaps in the same scene.
+    The LightmapGI node only bakes nodes that are on the same level as the
+    LightmapGI node (siblings), or nodes that are children of the
+    LightmapGI node. This allows you to use several LightmapGI nodes to bake
+    different parts of the scene, independently from each other.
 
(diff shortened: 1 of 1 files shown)


tutorials/3d/global_illumination/using_lightmap_gi.rst
```diff
@@ -80,6 +80,11 @@ Setting up
     graphics API limitations on those devices. On Android and web platforms,
     only *rendering* lightmaps that were baked on a desktop PC is supported.
 
+.. note::
+
+    The LightmapGI node only bakes nodes under its parent.
+    This is because some people may require several separate lightmaps in the same scene.
```
Calinou, line 86:
```suggestion
    The LightmapGI node only bakes nodes that are on the same level (siblings),
    or nodes that are children of the LightmapGI node. This allows you to use
    several LightmapGI nodes to bake different parts of the scene,
    independently from each other.
```

mhilbrunner (approved):

mhilbrunner:

PR #10252, opened 2024-11-12
Change warning boxes to danger boxes if data loss can occur

tetrapod00:
Promotes warning blocks to danger blocks if any loss of data can occur.

Prior to this PR, the documentation seems to use:
- "Note" for optional information, used often, 
- "Caution" only twice, 
- "Attention" only 8 times,
- "Danger" once, in the tool script page,
- and "Warning" for the vast majority of everything else. "Warning" covers a very wide range of severity, from side notes that *should* be read, potential misconceptions, conventions that are different than usual, or performance problems. It's used often enough that you might start ignoring the warning boxes.

There are a few instances where data loss or actual harm can occur, and these instances should use "Danger" instead of "Warning" to stand out. I checked all the existing "Warning" boxes, and also checked for "backup" and "version control" mentions to find anything else.

Later, I'd like to investigate maybe using the lesser-used Attention and Caution notes more often, for notes that are less severe. Both of these are still the same color as Warning, but the words have different severity. I think Attention could be used in many cases where Warning currently is.

### DIFF
--- a/contributing/development/core_and_modules/custom_platform_ports.rst
+++ b/contributing/development/core_and_modules/custom_platform_ports.rst
@@ -169,7 +169,7 @@ you have two options:
 Distributing a custom platform port
 -----------------------------------
 
-.. warning::
+.. danger::
 
     Before distributing a custom platform port, make sure you're allowed to
     distribute all the code that is being linked against. Console SDKs are
--- a/contributing/workflow/bisecting_regressions.rst
+++ b/contributing/workflow/bisecting_regressions.rst
@@ -42,7 +42,7 @@ whether the issue is a regression in 4.0 or not.
 - If the issue is **not present** in 3.x, then you can try older 4.0 alphas and
   betas to determine when the regression started.
 
-.. warning::
+.. danger::
 
     Project files may be incompatible between Godot versions.
     **Make a backup of your project** before starting the bisection process.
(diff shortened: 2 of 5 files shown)


contributing/development/core_and_modules/custom_platform_ports.rst
```diff
@@ -169,7 +169,7 @@ you have two options:
 Distributing a custom platform port
 -----------------------------------
 
-.. warning::
+.. danger::
```
tetrapod00, line 172: Based on the same standard of potential legal harm used to change this warning, the warning in Complying with licenses could also be a danger:
https://github.com/godotengine/godot-docs/blob/905b1f4f0805894d1a09927e384eb234ae243efc/about/complying_with_licenses.rst?plain=1#L8-L12
However, that one occurs early in the docs in the About section, so a danger might scare people off. And I think the potential harm of getting the MIT license a little wrong is less than the potential harm of distributing an NDA console port 🤷

mhilbrunner:
> Later, I'd like to investigate maybe using the lesser-used Attention and Caution notes more often, for notes that are less severe. Both of these are still the same color as Warning, but the words have different severity. I think Attention could be used in many cases where Warning currently is.

I agree. Also, and more long-term, I'd like to review every place where we use these notes and reconsider whether they are necessary, need to be in those boxes, or whether there are better alternatives. Whenver I stumble upon a section that has multiple of them plastered all over it, it looks... busy? Confusing? Dangerous, like there's thousands of little gotchas to keep in mind?

I'd like if we could avoid stacking these boxes. No easy task, however :)

mhilbrunner (approved):

mhilbrunner:
Thanks!

PR #10258, opened 2024-11-13
"Base tile size equal or bigger than" in TileSetAtlasSource

berarma:


AThousandShips:

These changes needs to be done over [here](https://github.com/godotengine/godot), changing the generated files is not allowed, see [updating the class reference](https://docs.godotengine.org/en/latest/contributing/documentation/updating_the_class_reference.html)

Closing, but thank you nonetheless

Issue #10250, opened 2024-11-11
`VERTEX` shader built-in description confusing


tetrapod00:
**Your Godot version:**
4.4
**Issue description:**
This issue came from a help chat discussion, making this issue to track for myself.

`VERTEX` in `fragment()` assumes a bit of shader knowledge. The current description "Vertex position that comes from the vertex() function (default, in view space)" assumes that you already know that the *pixel/fragment* position is the interpolated *vertex* position. I'm not even sure if the docs currently document how values get interpolated like this. The table can't explain how everything in shaders works, of course, but this seems like it can be made more clear with only a few words.

Possible improvements:
- Improve the `VERTEX` description itself, and likely the vertex() `VERTEX` and maybe vertex() `POSITION` ones too, to avoid confusion when searching for "position" in the page.
- Check that interpolation between vertex() and fragment() is covered *somewhere*, maybe mention it briefly in this page.

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/latest/tutorials/shaders/shader_reference/spatial_shader.html

Issue #7927, opened 2023-09-16
Clarify what "local coordinates" means in the spatial shader docs


mieldepoche:
**Your Godot version:** 4.1.1

**Issue description:**

![image](https://github.com/godotengine/godot-docs/assets/71937306/9449c742-900f-4235-b6e3-fad44c59ba8e)

it is unclear what "local coordinates" refer to. I think it's view space.
It should use the same terminology as other entries that all refer to a kind of space (model, view, world).

**URL to the documentation page:** https://docs.godotengine.org/en/stable/tutorials/shaders/shader_reference/spatial_shader.html

AThousandShips:
Look above:
> Vertex data (VERTEX, NORMAL, TANGENT, BITANGENT) are presented in local model space. If not written to, these values will not be modified and be passed through as they came.
>
> They can optionally be presented in world space by using the world_vertex_coords render mode.
>
> Users can disable the built-in modelview transform (projection will still happen later) and do it manually with the following code:

mieldepoche:
Ah, I see.
Well I think there should be a nuance then, considering the variety of scenarios.
The fragment NORMAL has a hint to that:
![image](https://github.com/godotengine/godot-docs/assets/71937306/311aa0ea-abcf-4f22-a25b-01cbe7e76491)
but it's not enough I think.
Ideally, I'd say it should describe all of the scenarios:
- model space by default
- world space when using the `world_vertex_coords` render mode
- ??? (idk) when using the `skip_vertex_transform` render mode

The table is the first thing you look at when looking for that information, so I think it deserves to be there.

AThousandShips:
The table would be very busy if we were to add information that's right there, repeating things over and over isn't a good idea IMO

mieldepoche:
maybe saying it can be in various spaces and adding a link that scrolls up to the explanation is a better solution then

Issue #9780, opened 2024-08-18
Add documentation about shader coordinate spaces


tetrapod00:
**Your Godot version:**
4.3
**Issue description:**
Information about coordinate spaces is lacking in the shader pages. They are referred to by multiple names ("model space" = "local space"). They use names which may be unfamiliar to users coming from other renderers. The pages assume that you are already familiar with coordinate space from prior shader experience. Knowledge of how to transform between spaces is assumed.

I propose two things to solve this:

### Brief documentation in Spatial, CanvasItem, etc pages

I am already working on this half in https://github.com/godotengine/godot-docs/pull/9757.

Add a section about coordinate spaces to each shader documentation page. It will have a table of the various coordinate spaces, each with a paragraph long description of the coordinate space. The description will use as simple language as possible without being misleading. It will avoid technical terms if it does aid understanding. The range of each coordinate space will be documented, along with alternate names you may see ("local" or "object" for "model"; "camera" or "eye" for "view").

Below the table will be a section containing snippets on how to use coordinate spaces and convert between them, such as the common `vec3 position_world = (MODEL_MATRIX * vec4(VERTEX, 1.0)).xyz;`.

The coordinate space section should be in the individual shader pages, not the shading reference page, because the spaces are different between CanvasItem and Spatial.

Each page needs a different level of attention:
-   Spatial shaders will have a section with a table and some conversion snippets.
-   CanvasItem shaders will have a section with a table and some conversion snippets. The language on the CanvasItem page may be aimed even more towards beginners, as 2D shaders are a good introduction. It's okay to assume a little more prior knowledge for 3D. CanvasItem can also link to [this page](https://docs.godotengine.org/en/stable/contributing/development/core_and_modules/2d_coordinate_systems.html).
-   Particle shaders will link to the Spatial and CanvasItem sections, since Particle shaders can use both.
-   Fog and Sky could either link to Spatial, copy the entire section, or only list the world spaces they use.

### In-depth documentation in a new page
Once that section is complete, we may want to add a new page that describes the entire user-accessible shading pipeline, from `vertex()` to `light()`, and including the clip, NDC, and screen space operations that happen after your shader code runs. This page would be aimed at those looking to supplement their practical knowledge with precise technical knowledge of exactly how Godot renders a frame.

[...]

PR #10259, opened 2024-11-13
Clarify VERTEX and other shader built-ins

tetrapod00:
Closes https://github.com/godotengine/godot-docs/issues/10250 and closes https://github.com/godotengine/godot-docs/issues/7927. Note that this PR resolves half of https://github.com/godotengine/godot-docs/issues/7927, with the other half, an explanation of local coordinates, now tracked in https://github.com/godotengine/godot-docs/issues/9780.

The description for `VERTEX` in `fragment()` is longer than the corresponding descriptions for `NORMAL`, `BINORMAL`, and `TANGENT`, briefly describing how values are interpolated. While this is basic shader knowledge, it actually seems to be under-explained in the Godot docs in other places I would expect to see it. Since using `VERTEX` is so common, and since the name is somewhat confusing, I think it's okay to include such an explanatory note here.

I also added notes like "If ``skip_vertex_transform`` is enabled, it may not be in view space." to several other built-ins. As we add more and more built-ins, the table gets longer, and I don't think it's expected that users read the *whole page* to understand each description. I think we should start treating the built-in descriptions as closer to class member descriptions - they are expected to be read individually, with the *optional* context of the rest of the page. In this case, including a mention of `skip_vertex_transform` will prompt a reader to read the rest of the page, if applicable.

In the future we should consider changing these pages to something more like the class ref, with slightly longer descriptions, and copyable links to each individual shader built-in variable.

### DIFF
--- a/tutorials/shaders/shader_reference/spatial_shader.rst
+++ b/tutorials/shaders/shader_reference/spatial_shader.rst
@@ -59,9 +59,11 @@ For visual examples of these render modes, see :ref:`Standard Material 3D and OR
 +-------------------------------+------------------------------------------------------------------------------------------------------+
 | **specular_disabled**         | Disable specular.                                                                                    |
 +-------------------------------+------------------------------------------------------------------------------------------------------+
-| **skip_vertex_transform**     | ``VERTEX``/``NORMAL``/etc. need to be transformed manually in the ``vertex()`` function.             |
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):

mhilbrunner:
Thanks!

PR #10262, opened 2024-11-14
Add comments about non-monospaced tables

tetrapod00:
This came up a couple times in reviews of the original PRs for these tables, and in a followup PR. Hopefully this saves some frustration in future maintenance.

The upgrading to 4.1, 4.2, 4.3 pages also use emojis in a table, but those are subject to much less future maintenance so I think its fine to omit this note in those pages.


skyace65 (approved):

mhilbrunner (approved):

mhilbrunner:
Thanks!

tetrapod00:
Since this PR only exists for future maintenance reasons, there is no need to cherrypick it.

PR #10263, opened 2024-11-14
Update link in Upgrading from Godot 3 to Godot 4

tetrapod00:
Changes a link to "new features present in 4.0" to point to a [Godot release blogpost](https://godotengine.org/article/godot-4-0-sets-sail/). There might be another better link than this.
The old link went to a [github changelog that would not load](https://github.com/godotengine/godot/blob/master/CHANGELOG.md). It was not exactly broken, but it also wasn't useful.


mhilbrunner (approved):

mhilbrunner:
Thanks!

PR #10281, opened 2024-11-18
Add back information about per-vertex shading

tetrapod00:
Documentation for https://github.com/godotengine/godot/pull/83360.

Rather than use the existing screenshots from prior docs versions, I took new ones at higher resolution.

I also re-took the brick cube screenshot for this section, using a CC0 bricks texture from [AmbientCG](https://ambientcg.com/view?id=Bricks051). But we might want to revert that since now this section does not match the rest of this page.

This whole page kind of needs a rewrite with all new screenshots, so I'm not particularly worried about introducing a little more visual inconsistency.

I did my best to search the rest of the docs for "vertex shading" and variations, and this is all I could find.


Calinou (approved):
Looks good to me.

mhilbrunner (approved):

PR #10294, opened 2024-11-20
Make syncref work on stable

adikhoff:


tetrapod00:
I think you have the wrong repo.

PR #10299, opened 2024-11-21
Add `samplerExternalOES` type to variable types of shading language

Chaosus:
To reflect change from https://github.com/godotengine/godot/pull/96982


tutorials/shaders/shader_reference/shading_language.rst
```diff
@@ -943,88 +943,91 @@ GDScript uses different variable types than GLSL does, so when passing variables
...
++------------------------+-------------------------+------------------------------------------------------------+
+| **usampler3D**         | **Texture3D**           |                                                            |
++------------------------+-------------------------+------------------------------------------------------------+
+| **samplerCube**        | **Cubemap**             | See :ref:`doc_importing_images_changing_import_type` for   |
+|                        |                         | instructions on importing cubemaps for use in Godot.       |
++------------------------+-------------------------+------------------------------------------------------------+
+| **samplerCubeArray**   | **CubemapArray**        | Only supported in Forward+ and Mobile, not Compatibility.  |
++------------------------+-------------------------+------------------------------------------------------------+
+| **samplerExternalOES** | **ExternalTexture**     | External texture.                                          |
+|                        |                         | Only supported in Compatibility/Android platform.          |
```
dsnopek, line 1029: We don't need to say it here now (because we don't support Vulkan yet), but just a note about when we do eventually add support for Vulkan: it can use `sampler2D` for external textures which is what `samplerExternalOES` is translated to on Vulkan, so if you use `samplerExternalOES`, you can write a shader that'll work both on OpenGL and Vulkan.

dsnopek (approved):
Looks good to me!

tetrapod00:
This should be added to the table of data types at the top of the Shading Language page, too. Looks like you only added it to the table comparing GDScript types

Chaosus:
@tetrapod00 Check again, I think I've fixed it.

tetrapod00:
Thanks!

PR #10301, opened 2024-11-21
How to bind to external libraries with GDExtension

GreenCrowDev:
Since https://docs.godotengine.org/en/stable/contributing/development/core_and_modules/binding_to_external_libraries.html exists for C++ modules, it could be a good idea to have a manual about how to setup external libraries with GDExtension.

I think it's important to explain the most common issues a user can encounter when trying to bind external libraries, such as compiler errors, and how to fix them.

Godot Engine is probably one of the best platforms to develop on top for beginners and intermediates that want to grow as open source contributors and developers.

I open this draft so that I can gather suggestions while I try to write this section during my current trial and error.

The worst stumbling blocks right have been linking and https://github.com/godotengine/godot-cpp/pull/1469 https://github.com/godotengine/godot-cpp/pull/1319 https://github.com/godotengine/godot-cpp/pull/957

### DIFF
--- a/tutorials/scripting/gdextension/gdextension_binding_to_external_libraries.rst
+++ b/tutorials/scripting/gdextension/gdextension_binding_to_external_libraries.rst
@@ -1,0 +1,4 @@
+.. _gdextension_binding_to_external_libraries:
+
+Bind to external libraries with GDExtension
+===========================================


tutorials/scripting/gdextension/gdextension_binding_to_external_libraries.rst
```diff
@@ -0,0 +1,4 @@
+.. _gdextension_binding_to_external_libraries:
```
tetrapod00, line 1:
```suggestion
.. _doc_gdextension_binding_to_external_libraries:
```
I know this is nowhere near review but FYI we prefix the doc anchors like this

skyace65:
@GreenCrowDev is this PR something you're still working on or plan to work on?

tetrapod00:
It looks like @paddy-exe already has plans to write a similar page as mentioned in https://github.com/godotengine/godot-docs/pull/10570#pullrequestreview-2581363851. Since this is in such an early state, I think I'd recommend closing to avoid duplicating work.

GreenCrowDev:
I'm too busy right now, sorry.
We can close this as you mentioned someone is doing a similar contribution!

tetrapod00:
Closed per the above comments. Thanks nonetheless, it is a good idea to document this.

In the future, if you only have an overall outline for a new page or addition, it might be better to make an issue first so that the work can be picked up by other contributors later

Issue #10300, opened 2024-11-21
Contradictory performance information about `Packed*Array` types


mechalynx:
**Your Godot version:**

Both stable (4.3) documentation and latest ([0ceba58](https://github.com/godotengine/godot-docs/commit/0ceba58667e0fd08118638057a9a983e0a94e6a5))

**Issue description:**

There is contradictory information within the class reference and the gdscript reference in regards to the performance of `Packed*Array` types versus regular `Array` and typed array classes.

Specifically the class references for:
  - [PackedVector2Array](https://docs.godotengine.org/en/latest/classes/class_packedvector2array.html)
  - [PackedVector3Array](https://docs.godotengine.org/en/latest/classes/class_packedvector3array.html)
  - [PackedStringArray](https://docs.godotengine.org/en/latest/classes/class_packedstringarray.html)
  - [PackedInt64Array](https://docs.godotengine.org/en/latest/classes/class_packedint64array.html)
  - [PackedFloat64Array](https://docs.godotengine.org/en/latest/classes/class_packedfloat64array.html)
  - [PackedColorArray](https://docs.godotengine.org/en/latest/classes/class_packedcolorarray.html)

(but not the class references for `PackedInt32Array`, `PackedFloat32Array`, `PackedVector4Array` and `PackedByteArray`)

all have this paragraph:
> Differences between packed arrays, typed arrays, and untyped arrays: <ins>**Packed arrays are generally faster to iterate on and modify compared to a typed array of the same type**</ins> (e.g. [PackedInt32Array](https://docs.godotengine.org/en/latest/classes/class_packedint32array.html#class-packedint32array) versus Array[int]). Also, packed arrays consume less memory. As a downside, packed arrays are less flexible as they don't offer as many convenience methods such as [Array.map](https://docs.godotengine.org/en/latest/classes/class_array.html#class-array-method-map). Typed arrays are in turn faster to iterate on and modify than untyped arrays.

Reading this, I understand that for performance, `Packed` arrays are optimal, followed by typed `Arrays`, followed by untyped `Arrays` and that `Packed` arrays simply don't have some convenience methods. So if I just want to store an unknown amount of data in an array and I know I only care to iterate over it and add to it, it makes sense to use a `Packed` array type.

Yet in the [gdscript reference](https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_basics.html#packed-arrays) it says:

[...]

PR #10304, opened 2024-11-22
Update PackedArray explanation to match Godot 4.0 behavior

mechalynx:
Closes godotengine/godot-docs#10300

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_basics.rst
+++ b/tutorials/scripting/gdscript/gdscript_basics.rst
@@ -893,12 +893,26 @@ native or user class, or enum. Nested array types (like ``Array[Array[int]]``) a
 Packed arrays
 ^^^^^^^^^^^^^
 
-GDScript arrays are allocated linearly in memory for speed.
-Large arrays (more than tens of thousands of elements) may however cause
-memory fragmentation. If this is a concern, special types of
-arrays are available. These only accept a single data type. They avoid memory
-fragmentation and use less memory, but are atomic and tend to run slower than generic
-arrays. They are therefore only recommended to use for large data sets:
+PackedArrays are generally faster to iterate on and modify compared to a typed
+Array of the same type (e.g. PackedInt32Array versus Array[int]) and consume
+less memory. In the worst case, they are expected to be as fast as an untyped
+Array. Conversely, non-Packed Arrays (typed or not) have extra convenience
+methods such as :ref:`Array.map <class_Array_method_map>` that PackedArrays
+lack. Consult the :ref:`class reference <class_PackedFloat32Array>` for details
+on the methods available. Typed Arrays are generally faster to iterate on and
+modify than untyped Arrays.
+
+While all Arrays can cause memory fragmentation when they become large enough,
+if memory usage and performance (iteration and modification speed) is a concern
+and the type of data you're storing is compatible with one of the `Packed`
+Array types, then using those may yield improvements. However, if you do not
+have such concerns (e.g. the size of your array does not reach the tens of
+thousands of elements) it is likely more helpful to use regular or typed
+Arrays, as they provide convenience methods that can make your code easier to
+write and maintain (and potentially faster if your data requires such
+operations a lot). If the data you will store is of a known type (including
+your own defined classes), prefer to use a typed Array as it may yield better
+performance in iteration and modification compared to an untyped Array.
 
 - :ref:`PackedByteArray <class_PackedByteArray>`: An array of bytes (integers from 0 to 255).
 - :ref:`PackedInt32Array <class_PackedInt32Array>`: An array of 32-bit integers.
--- a/tutorials/scripting/gdscript/gdscript_basics.rst
+++ b/tutorials/scripting/gdscript/gdscript_basics.rst
@@ -894,7 +894,7 @@ Packed arrays
 ^^^^^^^^^^^^^
 
 PackedArrays are generally faster to iterate on and modify compared to a typed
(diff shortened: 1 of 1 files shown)


clayjohn (approved):
Looks great!

tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -893,12 +893,26 @@ native or user class, or enum. Nested array types (like ``Array[Array[int]]``) a
...
 ^^^^^^^^^^^^^
 
-GDScript arrays are allocated linearly in memory for speed.
-Large arrays (more than tens of thousands of elements) may however cause
-memory fragmentation. If this is a concern, special types of
-arrays are available. These only accept a single data type. They avoid memory
-fragmentation and use less memory, but are atomic and tend to run slower than generic
-arrays. They are therefore only recommended to use for large data sets:
+PackedArrays are generally faster to iterate on and modify compared to a typed
+Array of the same type (e.g. PackedInt32Array versus Array[int]) and consume
```
tetrapod00, line 897:
```suggestion
Array of the same type (e.g. PackedInt64Array versus Array[int]) and consume
```
Isn't `int` is [64-bit](https://docs.godotengine.org/en/latest/classes/class_int.html#class-int) in GDScript?

tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -893,12 +893,26 @@ native or user class, or enum. Nested array types (like ``Array[Array[int]]``) a
...
+less memory. In the worst case, they are expected to be as fast as an untyped
+Array. Conversely, non-Packed Arrays (typed or not) have extra convenience
+methods such as :ref:`Array.map <class_Array_method_map>` that PackedArrays
+lack. Consult the :ref:`class reference <class_PackedFloat32Array>` for details
+on the methods available. Typed Arrays are generally faster to iterate on and
+modify than untyped Arrays.
+
+While all Arrays can cause memory fragmentation when they become large enough,
+if memory usage and performance (iteration and modification speed) is a concern
+and the type of data you're storing is compatible with one of the `Packed`
```
tetrapod00, line 907:
```suggestion
and the type of data you're storing is compatible with one of the ``Packed``
```
Use two backticks if you wanted `code` formatting here

tetrapod00 (commented):
Reviewing for docs style, I don't know enough to review the content deeply

tutorials/scripting/gdscript/gdscript_basics.rst
mechalynx: That part is taken verbatim from the class reference for the PackedArray classes (see: https://docs.godotengine.org/en/latest/classes/class_packedint64array.html which is where I got it from). I think it's just a general comparison. Default `int` might be 64-bit but we have PackedArrays for Int32 and Int64. Unless I've misunderstood what the correction is.

tetrapod00: Ah, it's probably fine then 👍

tetrapod00 (approved):
Looks good to me style-wise

tutorials/scripting/gdscript/gdscript_basics.rst
tetrapod00: I just saw your edits now. I think `Color` would be a less controversial class to use in the comparison.

PackedInt32Array/Array[int] isn't *wrong* it's just potentially confusing. Since this PackedInt32Array/Array[int] comparison already exists in the class reference, I'm personally okay with leaving it and merging this PR as-is. But I'll leave this conversation open so others can chime in if they want to.

mechalynx: Github wasn't showing your comments after mine... I accepted the change. I'll just make the same adjustment to the class reference when I make a PR for that. After thinking about it, I felt that `Color` might be less "raw" of a comparison, since that's a more complex data type than an `int`. I felt keeping it to ints keeps it simpler, so changing it to PackedInt64Array feels cleaner to me.

tetrapod00:
Thank you! Congratulations on your first merged documentation contribution 🎉

tetrapod00:
@mechalynx I'm not actually sure, I asked in the contributors chat.

akien-mga:
I amended the commit to use the anonymized GitHub email: https://github.com/godotengine/godot-docs/commit/0f8c24b071bba7f7955f1c8cb8c0fa30003f0d81.patch

This is a destructive operation that requires a force push so normally not something we'd do, but here it was still the latest commit and no new PR had been opened in the meantime, so it shouldn't cause too much trouble.

mechalynx:
Thankyou. I did not expect you to go to the lengths to even replace it with the anyonymized version which is more than kind on your part. I understand the nature of the operation and that it was presumptuous to ask for this in the first place, especially for a repository of this size and scope. It being the last commit was what made me decide to ask but I still feel deeply embarassed for having to do so. I understand this was likely a stressful decision and I am very sorry for the trouble. If necessary, I can delete my comments related to this, if there is concern that it might give the wrong idea about project policy.

I looked into this and it seems github at some point added a dropdown to select email used for merge commits in 2019 but since at least 2022 it seems the option is gone again. The commits on my fork branch had the email I chose and I assumed it would use that one but it instead used my primary. I have since changed email to private and use the anonymized one in my local git installation.

Thanks again and sorry.

mhilbrunner:
No worries, shit happens, especially if something as complex as Git is involved :) Thanks for contributing!

mhilbrunner:
Cherrypicked to 4.3 in https://github.com/godotengine/godot-docs/pull/10347.

PR #10315, opened 2024-11-23
Explain difference between global shader uniforms and global scope in shaders

tetrapod00:
The term "global" is used for two things in the Shading Langauge page:
- Global scope, that is everything not inside a process function
- Global shader uniforms, which are shader uniforms that can be accessed from every shader in the project.

This PR adds a note clarifying the difference.


mhilbrunner (approved):

mhilbrunner:
Not sure how this one slipped through. Merged!

PR #10319, opened 2024-11-25
Update translation priorities in Localization contributing page

tetrapod00:
Updated the word counts to roughly match the current "All strings" count on Weblate.

Also changed the recommended priority of translations to editor interface, then class reference, then manual. As I don't have first-hand experience with translation, I'm not 100% sure about that order. I asked in the translation RC and got a couple agreements, but I'm not sure how unanimous that is. It does match the concept of treating the class reference as more of a source of truth, since it's in the main repo and included in the engine itself.

### DIFF
--- a/contributing/documentation/editor_and_docs_localization.rst
+++ b/contributing/documentation/editor_and_docs_localization.rst
@@ -12,12 +12,11 @@ translation effort of the community.
 
 These resources include:
 
-1. The `Godot editor's interface <https://hosted.weblate.org/projects/godot-engine/godot/>`__
-   (ca. 15,000 words).
-2. The `online documentation <https://hosted.weblate.org/projects/godot-engine/godot-docs/>`__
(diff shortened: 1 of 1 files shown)


Mickeon (approved):

mhilbrunner (approved):
I'm not sure whether the word count is useful or should be dropped. If it is, maybe we should add one of those `UPDATE` comments here so try to remember?

Mickeon:
Maybe it's best to drop it, yeah. If people want to see the count is right there on Weblate. The actual numbers are only going to grow and cannot be maintained accurate every time.

tetrapod00:
It seems reasonable to drop the word count entirely.

As an alternative, we could list the total word count which right now is ~1,000,000, but not the subcounts

mhilbrunner:
This seems fine to me :) Thank you, merged.

PR #10326, opened 2024-11-27
Document Light3D shadow caster mask property

tetrapod00:
Documentation for https://github.com/godotengine/godot/pull/85338.

Note that this page does duplicate information that is in the class reference, so I opted to write a longer description than the very short description from the class reference.


mhilbrunner (approved):

mhilbrunner:
Thanks!

Issue #10146, opened 2024-10-26
Docs use old and removed `PoolVector`


dustdfg:
**Your Godot version:** 4.4.dev/master

**Issue description:** Docs use old and removed `PoolVector`

**URL to the documentation page:**  https://docs.godotengine.org/en/latest/contributing/development/core_and_modules/core_types.html#allocating-memory

tetrapod00:
See also https://github.com/godotengine/godot-docs/issues/3192, maybe https://github.com/godotengine/godot-docs/issues/6259.

Issue #6259, opened 2022-10-05
Pool Vector Dead Link on 4.0


TheYellowArchitect:
PoolVectorArrays were a thing in 3.x it seems
https://docs.godotengine.org/en/stable/classes/class_poolvector2array.html?highlight=poolvector2array

They don't exist in 4.x https://docs.godotengine.org/en/latest/classes/class_poolvector2array.html?highlight=poolvector2array
[PackedVectorXArrays](https://docs.godotengine.org/en/latest/classes/class_packedvector2array.html?highlight=packedvector2array) replaced them.

Yet, in the docs, there is still a location they are referenced:
https://docs.godotengine.org/en/latest/development/cpp/core_types.html?highlight=poolvector#allocating-memory
pool_vector.h especially is a dead link and has been long deleted. memory.h seems to have completely remade it to use MemoryStaticPool (can't confirm as I can't read the source code)

Related:
- https://github.com/godotengine/godot/pull/36311
- https://github.com/godotengine/godot-docs/issues/3192

My suggestions to close this and the above issue:
- [ ] Make https://docs.godotengine.org/en/latest/classes/class_poolvector2array.html?highlight=poolvector2array redirect to https://docs.godotengine.org/en/latest/classes/class_packedvector2array.html?highlight=packedvector2array
Same applies for Vector3 and whatever converted from pool -> packed

- [ ] Update the https://docs.godotengine.org/en/latest/development/cpp/core_types.html?highlight=poolvector#allocating-memory with whatever is the new memory return/format

Calinou:
Pool*Vector was replaced by Packed*Vector in 4.0. The documentation referencing them should be rewritten (not just renamed) as how packed vectors work internally is different from pooled vectors. You can find some information about it in [Core refactoring progress report #1](https://godotengine.org/article/core-refactoring-progress-report-1).

Issue #3192, opened 2020-02-19
Every tutorials that refers to PoolVector (C++) or Pool*Array (scripting) should be reviewed for 4.0


pyacier:
After this [Convert all references and instances of PoolVector to Vector #36311](https://github.com/godotengine/godot/pull/36311) PoolVector do no longer exist, all tutorials referring to it need review.

tetrapod00:
Closing, since after https://github.com/godotengine/godot-docs/pull/10329 the remaining references to Pool*Array, including a single tutorial, have been removed. I have to trust that in the four years since this was reported, the rest of the tutorials were updated correctly.

PR #10329, opened 2024-11-28
Remove references to PoolVector

esainane:
PoolVector has not existed since 2020. It was removed in godotengine/godot@3205a92a.

_Bugsquad edit: closes https://github.com/godotengine/godot-docs/issues/10146, possibly also https://github.com/godotengine/godot-docs/issues/3192 and https://github.com/godotengine/godot-docs/issues/6259_

### DIFF
--- a/contributing/development/core_and_modules/core_types.rst
+++ b/contributing/development/core_and_modules/core_types.rst
@@ -103,36 +103,12 @@ which are equivalent to new, delete, new[] and delete[].
 memnew/memdelete also use a little C++ magic and notify Objects right
 after they are created, and right before they are deleted.
 
-For dynamic memory, the PoolVector<> template is provided. PoolVector is a
-standard vector class, and is very similar to vector in the C++ standard library.
-To create a PoolVector buffer, use this:
-
-.. code-block:: cpp
-
-    PoolVector<int> data;
-
-PoolVector can be accessed using the [] operator and a few helpers exist for this:
-
-.. code-block:: cpp
-
-    PoolVector<int>::Read r = data.read()
-    int someint = r[4]
-
-.. code-block:: cpp
-
-    PoolVector<int>::Write w = data.write()
-    w[4] = 22;
-
-These operations allow fast read/write from PoolVectors and keep it
-locked until they go out of scope. However, PoolVectors should be used
-for small, dynamic memory operations, as read() and write() are too slow for a
-large amount of accesses.
+For dynamic memory, use Vector<>.
 
 References:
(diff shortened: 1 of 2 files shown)


contributing/development/core_and_modules/core_types.rst
```diff
@@ -103,36 +103,12 @@ which are equivalent to new, delete, new[] and delete[].
...
-.. code-block:: cpp
-
-    PoolVector<int>::Write w = data.write()
-    w[4] = 22;
-
-These operations allow fast read/write from PoolVectors and keep it
-locked until they go out of scope. However, PoolVectors should be used
-for small, dynamic memory operations, as read() and write() are too slow for a
-large amount of accesses.
+For dynamic memory, use Vector<>.
```
bruvzg, line 106: This page should also mention [`LocalVector`](https://github.com/godotengine/godot/blob/master/core/templates/local_vector.h), and [`Packed*Vector`](https://github.com/godotengine/godot/blob/0eadbdb5d0709e4e557e52377fa075d3e2f0ad1f/core/variant/variant.h#L73-L83).

- `Vector<>` - copy-on-write and can be safely passed via public API.
- `Packed*Vector` - aliases for specific `Vector<*>` types accessible via GDScript.
- `LocalVector<>` - non COW version, with less overhead, for the internal use.

bruvzg: Yes, `For dynamic memory, use Vector<>` is correct, and in general a safe option to use, so it's OK as is.

tetrapod00: If it's OK as a bugfix, then let's merge this one as-is.

@esainane, you're welcome to document the other options mentioned here if you want to, as another PR, after this one is merged. If you don't feel knowledgeable enough, that is also OK.

esainane: Sure, I should be able to do something proper this Sunday.

esainane: I think I have a suitable follow up ready to go, in case people were waiting?

tetrapod00: It's fine to bundle it into this PR, since it's not actually merged yet.

(I thought we might have merged this already, but without an explicit second approval from an experienced engine dev I ended up not merging it yet)

tetrapod00:
Thank you!

I decided to merge this as a bugfix as-is. If you have a followup PR ready to document the other options, it would be very welcome 🙂

PR #10330, opened 2024-11-28
Mark Haxe GDExtension bindings as archived

esainane:
This project was archived October 2024.

### DIFF
--- a/about/list_of_features.rst
+++ b/about/list_of_features.rst
@@ -492,8 +492,6 @@ Scripting
   `Swift <https://github.com/migueldeicaza/SwiftGodot>`__, and `Rust <https://github.com/godot-rust/gdextension>`__
   bindings provided by the community. (Some of these bindings may be experimental and not production-ready).
 
-- Archived partial GDExtension bindings for `Haxe <https://github.com/HxGodot/hxgodot>`__.
-
 Audio
 -----
 
--- a/about/list_of_features.rst
+++ b/about/list_of_features.rst
@@ -489,8 +489,10 @@ Scripting
    - Use any build system and language features you wish.
 
 - Actively developed GDExtension bindings for `D <https://github.com/godot-dlang/godot-dlang>`__,
-  `Haxe <https://hxgodot.github.io/>`__, `Swift <https://github.com/migueldeicaza/SwiftGodot>`__, and `Rust <https://github.com/godot-rust/gdextension>`__
+  `Swift <https://github.com/migueldeicaza/SwiftGodot>`__, and `Rust <https://github.com/godot-rust/gdextension>`__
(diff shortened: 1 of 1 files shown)


about/list_of_features.rst
```diff
@@ -489,9 +489,11 @@ Scripting
    - Use any build system and language features you wish.
 
 - Actively developed GDExtension bindings for `D <https://github.com/godot-dlang/godot-dlang>`__,
-  `Haxe <https://hxgodot.github.io/>`__, `Swift <https://github.com/migueldeicaza/SwiftGodot>`__, and `Rust <https://github.com/godot-rust/gdextension>`__
+  `Swift <https://github.com/migueldeicaza/SwiftGodot>`__, and `Rust <https://github.com/godot-rust/gdextension>`__
   bindings provided by the community. (Some of these bindings may be experimental and not production-ready).
 
+- Archived partial GDExtension bindings for `Haxe <https://github.com/HxGodot/hxgodot>`__.
+
```
tetrapod00, line 496:
```suggestion
```
From the linked repo:
> This project is no longer maintained! This was an interesting experiment and journey which was in the end not able to live up to the vision I had in mind. In the end it was hard to justify the time and complexity required in relation to the factual outcome of a semi-working solution. Thanks!

Since it is both *archived* and *incomplete* I don't think this should be linked from the official docs.

If it was either *complete and archived* or *incomplete but actively developed* I think it may still have a place.

tetrapod00: Also, these language bindings are listed in the page [About / List of Features](https://docs.godotengine.org/en/stable/about/list_of_features.html). Archived partial bindings are *not* a feature of Godot, and should not be listed in a page in the About section (which is meant to be read by prospective users of Godot).

If these bindings were listed somewhere in the GDExtension pages, or in the Contributing section, there would be a bit more leeway here.

esainane: That seems reasonable. So this should be converted to just removing the link, then?

tetrapod00: Yeah. I tried to suggest the exact change (remove line 495 and 496 entirely) but for some reason it's not showing up quite right in my GitHub UI.

skyace65: I agree with tetrapod, I think it's best to just remove the link.

mhilbrunner: I likewise agree with just removing the reference.

esainane: Done!

tetrapod00 (approved):

skyace65 (approved):

skyace65:
Thanks!

mhilbrunner:
Cherry-picked to 4.3 in https://github.com/godotengine/godot-docs/pull/10363.

PR #10333, opened 2024-11-29
Add a page on the Output bottom panel

Calinou:
While `print_rich()` works independently of RichTextLabel, it still makes use of BBCode. (I'm not sure if there is another location where this paragraph would be more suitable.)


mhilbrunner (approved):

tutorials/ui/bbcode_in_richtextlabel.rst
```diff
@@ -1147,3 +1147,26 @@ This will add a few new BBCode commands, which can be used like so:
...
+different depending on the user's theme, while colors in the editor will use the
+same colors as they would in the project.
+
+.. note::
+
+    To print error and warning messages that can be filtered by type in the
+    editor Output panel, use
+    :ref:`@GlobalScope.push_error<class_@GlobalScope_method_push_error>` and
+    :ref:`@GlobalScope.push_warning<class_@GlobalScope_method_push_warning>`
+    instead of ``print_rich()``.
```
tetrapod00, line 1172: I don't think this is quite technically correct in all cases, either - `push_error` and `push_warning` do print to the Output panel when the debugger is *not* running, such as in the editor. But when the debugger is running, they only appear in the debugger. You mentioned this in https://github.com/godotengine/godot-proposals/issues/10568#issuecomment-2322758852

I think this paragraph needs slight wording tweaks to be technically correct, and/or we should change the behavior of push_warning and push_error as mentioned in the linked proposal.

---

As a side note, the description in the class ref of [print_rich()](https://docs.godotengine.org/en/latest/classes/class_@globalscope.html#class-globalscope-method-print-rich) doesn't say that `push_warning` and `push_error` are always present in the Output panel, but it does slightly imply it:
> Note: Consider using [push_error](https://docs.godotengine.org/en/latest/classes/class_@globalscope.html#class-globalscope-method-push-error) and [push_warning](https://docs.godotengine.org/en/latest/classes/class_@globalscope.html#class-globalscope-method-push-warning) to print error and warning messages instead of [print](https://docs.godotengine.org/en/latest/classes/class_@globalscope.html#class-globalscope-method-print) or [print_rich](https://docs.godotengine.org/en/latest/classes/class_@globalscope.html#class-globalscope-method-print-rich). This distinguishes them from print messages used for debugging purposes, while also displaying a stack trace when an error or warning is printed.

I think the current class ref description is *fine* because it's not *directly* wrong.

Calinou:
This PR now adds a new page on the Output bottom panel, which is linked to on the BBCode in RichTextLabel page.

tutorials/ui/bbcode_in_richtextlabel.rst
Calinou: I've added a mention of this behavior in the newly added page.

tetrapod00 (approved):
Looks good to me now, thanks for making extensive changes like this!

tutorials/scripting/debug/output_panel.rst
```diff
@@ -0,0 +1,93 @@
...
+  When an error is printed in a running project, it's displayed in the **Debugger > Errors**
+  tab instead.
+- :ref:`push_warning() <class_@GlobalScope_method_push_warning>`: Prints a warning message.
+  When a warning is printed in a running project, it's displayed in the **Debugger > Errors**
+  tab instead.
+
+To get more advanced formatting capabilities, consider using
+:ref:`doc_gdscript_printf` along with the above printing functions.
+
+.. _doc_bottom_panel_printing_rich_text:
```
tetrapod00, line 69:
```suggestion
.. _doc_output_panel_printing_rich_text:
```
CI is erroring because there is a mismatch between this and the reference to it.

tetrapod00 (approved):

mhilbrunner (approved):

mhilbrunner:
Thank you!

PR #5593, opened 2022-02-10
Document using Pyston to speed up builds on Linux

Calinou:
See https://github.com/godotengine/godot/issues/41271#issuecomment-909193308.

PR #10335, opened 2024-11-29
Remove instructions on setting up Pyston for faster development

Calinou:
- Follow-up to https://github.com/godotengine/godot-docs/pull/5593.

This is done for several reasons:

- Recent CPython versions such as 3.13 have mostly caught up with Pyston in terms of performance. If the trend continues, CPython may supersede Pyston's performance in the long term. This is especially the case if the experimental JIT is eventually able to run Godot's SCons setup.
- Pyston is not maintained anymore (its last release was in 2022), and the last Python version it can run is 3.8. Python 3.8 is now end-of-life. Godot's SCons setup (as well as SCons itself) will eventually require Python versions newer than 3.8.

### DIFF
--- a/contributing/development/compiling/compiling_for_linuxbsd.rst
+++ b/contributing/development/compiling/compiling_for_linuxbsd.rst
@@ -272,7 +272,7 @@ Start a terminal, go to the root dir of the engine source code and type:
 
 .. tip::
     If you are compiling Godot to make changes or contribute to the engine,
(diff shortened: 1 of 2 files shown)


mhilbrunner (approved):

mhilbrunner:
Thanks!

mhilbrunner:
Cherrypicked to 4.3 in https://github.com/godotengine/godot-docs/pull/10347.

PR #10126, opened 2024-10-21
Remove unnecessary note from doc page

Jedrzej1940518:
Removed unnecessary note, as this page got rid of "$" signs in below …

https://github.com/godotengine/godot-docs/pull/9956

before:
image

after:
image

PR #9956, opened 2024-09-15
Harmonize Bash command syntax for easier line selection and copy-pasting

Calinou:
- Remove prompt or `$` sign which makes triple-click based selection more difficult and time-consuming.
- Use `platform` instead of `p` alias in all SCons examples.

___

- See https://github.com/godotengine/godot-docs-user-notes/discussions/133#discussioncomment-10616085.

PR #10337, opened 2024-11-29
Removed irrelevant Note

OS-of-S:
Remove forgoted "Note", that is not actual anymore.

image
image

### DIFF
--- a/contributing/workflow/pr_workflow.rst
+++ b/contributing/workflow/pr_workflow.rst
@@ -88,9 +88,6 @@ repository (in Git speak, the *origin remote*). If you haven't already,
 download Git from `its website <https://git-scm.com>`_ if you're using Windows or
 macOS, or install it through your package manager if you're using Linux.
 
-.. note:: If you are on Windows, open Git Bash to type commands. macOS and Linux users
-          can use their respective terminals.
-
 To clone your fork from GitHub, use the following command:
 
 ::


tetrapod00:
Can you elaborate more on the reasoning for the removal of this note?

skyace65:
In addition to what tetrapod said, the note you have a screenshot of here is not the one that you removed.

tetrapod00:
I think I can follow the logic - since the command line examples were changed in https://github.com/godotengine/godot-docs/pull/9956 to no longer use a `$` (as demonstrated by the screenshotted note and the comments on a docs page), now we no longer need to recommend specifically Git Bash on Windows (which also uses a `$` for the command line prompt).

But the reasoning for this change needs to be clearly stated by the PR author in the description itself, not inferred from a chain of logic.

OS-of-S:
> In addition to what tetrapod said, the note you have a screenshot of here is not the one that you removed.

Oh, yes... It was mistake. I planned to remove only one note. Should I return the other one by myself to fix it?

tetrapod00:
The note in your screenshot was removed in https://github.com/godotengine/godot-docs/pull/10126 already. I'm not sure if you meant this, but please do not add the removed note back to the codebase.

Your PR currently removes another note. Can you please explain why you want to remove it?

OS-of-S:
> The note in your screenshot was removed in #10126 already.

Ah... I see... Probably the documentation had not been updated at that time, and I deleted the wrong note becouse due to my mistake. I beg your pardon, request should be closed.

OS-of-S:
1) A mistake was made
2) The need has disappeared

tetrapod00:
Yeah, mistakes happen, and our docs versioning can be a bit confusing at times. Thanks for contributing nonetheless.

PR #10353, opened 2024-11-30
add flex layout for left nav container

### LINKED ISSUE #10352: Elements are not scrollable in the left nav
**Your Godot version:** Not applicable.

**Issue description:** This issue only happens when viewing the "stable" or versioned docs. It doesn't happen when viewing the "latest" docs. If pages/subpages are expanded in the left nav, the container element becomes scrollable as expected, but it becomes impossible to scroll elements listed at the very top into view. This is due to the fact that container's top margin is set to a fixed number and it's not large enough to accommodate for the "version" and "language" dropdowns that appear when viewing the docs in "stable" or versioned mode. The expected behavior is that the scroll container should dynamically fill the available space in the left nav.

**URL to the documentation page:** Any page. For example: [docs.godotengine.org/en/stable](https://docs.godotengine.org/en/stable/)

hakanguclu:
fixes #10352 

this PR makes the left nav a flex container. by doing that, it also removes the manual calculation that happens in custom.js using fixed margin numbers. that manual calculation makes parts of the left nav non-scrollable in e.g. "stable" mode of the docs. with this PR, any changes to the top portion of the left nav's height will not matter as the bottom part will flex on the available height automatically.

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -1368,13 +1368,12 @@ p + .classref-constant {
 }
 
 .wy-side-nav-search {
+    flex-grow: 0;
+    flex-shrink: 0;
+    flex-basis: max-content;
     background-color: var(--navbar-background-color);
     color: var(--navbar-level-1-color);
     margin-right: 8px;
-}
-
-.wy-side-nav-search.fixed {
-    position: fixed;
 }
 
 @media only screen and (min-width: 769px) {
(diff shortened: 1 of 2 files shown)


hakanguclu:
Closing this PR as [commit 7a79288](https://github.com/godotengine/godot-docs/commit/7a792883194982759aa4286103f172d1579c5fc2) closes the issue this PR was attached to.

PR #10355, opened 2024-11-30
Update "exporting_pcks" page to change our stance about mods and pcks

adamscott:
As `.pck` files can have GDScript, and that there's no current way to limit the scope of the API of a `.pck` file, we should just change our stance in the "exporting_pcks" article to mention that we don't support this method for modding purposes anymore.

### DIFF
--- a/tutorials/export/exporting_pcks.rst
+++ b/tutorials/export/exporting_pcks.rst
@@ -1,7 +1,14 @@
 .. _doc_exporting_pcks:
 
-Exporting packs, patches, and mods
-==================================
+Exporting packs and patches
+===========================
+
+.. warning::
+    We changed our stance about exporting mods using PCK files for security reasons.
+
+    **We don't recommend anymore using PCK files to add modding functionality to your game.**
+
+    Please read :ref:`doc_exporting_pcks_about_mods_using_pck_files` for more information.
 
 Use cases
 ---------
@@ -15,8 +22,10 @@ Examples of this include...
 - Patches: the ability to fix a bug that is present in a shipped product.
 - Mods: grant other people the ability to create content for one's game.
 
-These tools help developers to extend their development beyond the initial
+Godot offers some tools to help developers to extend their development beyond the initial
 release.
+
+.. _doc_exporting_pcks_overview_of_pck_files:
 
 Overview of PCK files
 ---------------------
@@ -28,7 +37,7 @@ with extension ``.pck``).
 
 - incremental updates/patches
 - offer DLCs
-- offer mod support
+- offer mod support (see :ref:`doc_exporting_pcks_about_mods_using_pck_files` for more info)
 - no source code disclosure needed for mods
 - more modular project structure
 - users don't have to replace the entire game
@@ -51,6 +60,23 @@ PCK files usually contain, but are not limited to:
 The PCK files can even be an entirely different Godot project, which the
 original game loads in at runtime.
 
+.. _doc_exporting_pcks_about_mods_using_pck_files:
+
+About mods using PCK files
+--------------------------
+
+While it's easy enough to create mods for a Godot game using PCK files, **we don't recommend doing it**.
+
+The reason is simple: :ref:`PCK files are very powerful <doc_exporting_pcks_overview_of_pck_files>`. They can override and replace content in a Godot game.
+While this seems ideal for modding purposes (and yes, it is), it comes with the downside of giving the keys to the castle to any bad actor.
+
+While replacing shaders and models is quite benign, replacing scripts is problematic, as the GDScript scripting engine doesn't have actually
+any concept of a `sandbox <https://en.wikipedia.org/wiki/Sandbox_(computer_security)>`__. Every mod containing GDScript files have access
+to the entire Godot Engine API, including making network calls and accessing the filesystem.
+
+We do currently encourage game developers to create their own modding system for the time being. To be really secure, any modding system
+should limit the scope of scripting calls being made by a mod, as it may be compromised by an untrustworthy actor.
+
 Generating PCK files
 --------------------
 
@@ -66,7 +92,8 @@ process will build that type of file for the chosen platform.
 
 .. note::
 
-    If one wishes to support mods for their game, they will need their users to
+    If one wishes to support mods for their game using PCK files, even after reading
+    :ref:`doc_exporting_pcks_about_mods_using_pck_files`, they will need their users to
     create similarly exported files. Assuming the original game expects a
     certain structure for the PCK's resources and/or a certain interface for
     its scripts, then either...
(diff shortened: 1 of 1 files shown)


tetrapod00:
As a general comment on the structure of this PR: Usually we should change the docs to be correct in the present rather than leaving this kind of messy recommendation with caveats in a page, which mentions something we *used* to recommend and then says why we don't anymore.

I suppose in this case it does make sense to leave in a remnant of the old recommendation, since there are security concerns.

adamscott:
> As a general comment on the structure of this PR: Usually we should change the docs to be correct in the present rather than leaving this kind of messy recommendation with caveats in a page, which mentions something we _used_ to recommend and then says why we don't anymore.
> 
> I suppose in this case it does make sense to leave in a remnant of the old recommendation, since there are security concerns. There are slightly cleaner ways to go about that, I'll try to get around to a specific review for that, but others might want to discuss the PR overall first

I can remove the remnants, but then, we need a blog post to clearly mention our change of stance. Maybe a blog post is needed either way.

tetrapod00:
I think if the intent is to strongly signal that we no longer officially recommend using PCKs as mods (presumably at least partly in response to recent news?), a blog post is needed.

> I reread the article and I think the mentions are needed. 

That's okay with me, every rule has its exceptions and the reasons here make sense. In that case I do have suggestions for how to reword it to retain the information while being more idiomatic to the docs.

adamscott:
> In that case I do have suggestions for how to reword it to retain the information while being more idiomatic to the docs.

Go ahead, @tetrapod00 ! I'm not a native English speaker, so I welcome suggestions.

tutorials/export/exporting_pcks.rst
```diff
@@ -1,7 +1,14 @@
 .. _doc_exporting_pcks:
 
-Exporting packs, patches, and mods
-==================================
+Exporting packs and patches
+===========================
+
+.. warning::
```
tetrapod00, line 6:
```suggestion
.. danger::
```
Since the docs use `warning` for so many things, we should use `danger` for anything that is a legal, security, or data loss risk. See https://github.com/godotengine/godot-docs/pull/10252.

tutorials/export/exporting_pcks.rst
```diff
@@ -1,7 +1,14 @@
...
+    We changed our stance about exporting mods using PCK files for security reasons.
+
+    **We don't recommend anymore using PCK files to add modding functionality to your game.**
+
+    Please read :ref:`doc_exporting_pcks_about_mods_using_pck_files` for more information.
```
tetrapod00, line 11:
```suggestion
    See :ref:`doc_exporting_pcks_about_mods_using_pck_files` for more information.
```
Standard phrasing, avoids imperative "read".

tutorials/export/exporting_pcks.rst
```diff
@@ -1,7 +1,14 @@
...
 
-Exporting packs, patches, and mods
-==================================
+Exporting packs and patches
+===========================
+
+.. warning::
+    We changed our stance about exporting mods using PCK files for security reasons.
+
+    **We don't recommend anymore using PCK files to add modding functionality to your game.**
```
tetrapod00, line 9:
```suggestion
    **We no longer recommend using PCK files to add modding functionality to your game.**
```

mhilbrunner:
I disagree with this change. Security is always a necessary consideration, especially with implementing mods.

However, "only downloads mods you trust" is already a thing outside of Godot, same as for any other executable. The vast majority of mods for games I'm familiar with are similarly able to execute arbitrary, unsandboxed code.

"Just spend hundreds or thousands of hours developing your own sandbox" is also not really a useful suggestion. The people who have the technical chops or resources to already know and would do this. For everyone else, suggesting they absolutely *should* spend their scarce resources specifically developing their own sandbox to implement modding when, say, Minecraft or most Unity game's mods get away without sandboxing seems not really dev friendly and maybe a bit overzealous.

Even with sandboxing, "only download mods you trust" is important. We had asset importing/format vulnerabilities before, same for unpacking (as 7-Zip just had). So even if we had full sandboxing, one should not download and use any random file they find if they care about that machine and its local network.

tutorials/export/exporting_pcks.rst
```diff
@@ -51,6 +60,23 @@ PCK files usually contain, but are not limited to:
...
+About mods using PCK files
+--------------------------
+
+While it's easy enough to create mods for a Godot game using PCK files, **we don't recommend doing it**.
+
+The reason is simple: :ref:`PCK files are very powerful <doc_exporting_pcks_overview_of_pck_files>`. They can override and replace content in a Godot game.
+While this seems ideal for modding purposes (and yes, it is), it comes with the downside of giving the keys to the castle to any bad actor.
+
+While replacing shaders and models is quite benign, replacing scripts is problematic, as the GDScript scripting engine doesn't have actually
+any concept of a `sandbox <https://en.wikipedia.org/wiki/Sandbox_(computer_security)>`__. Every mod containing GDScript files have access
```
tetrapod00, line 74:
```suggestion
any concept of a `sandbox <https://en.wikipedia.org/wiki/Sandbox_(computer_security)>`__. Every mod containing GDScript files has access
```

tutorials/export/exporting_pcks.rst
```diff
@@ -51,6 +60,23 @@ PCK files usually contain, but are not limited to:
...
+to the entire Godot Engine API, including making network calls and accessing the filesystem.
+
+We do currently encourage game developers to create their own modding system for the time being. To be really secure, any modding system
```
tetrapod00, line 77:
```suggestion
We do currently encourage game developers to create their own modding system. To be really secure, any modding system
```
No need to specify both "currently" and "for the time being".

We *could* also mention the criteria for no longer recommending this (presumably some implementation of sandboxing?), but I think that doing so is optional.

tutorials/export/exporting_pcks.rst
```diff
@@ -51,6 +60,23 @@ PCK files usually contain, but are not limited to:
 The PCK files can even be an entirely different Godot project, which the
 original game loads in at runtime.
 
+.. _doc_exporting_pcks_about_mods_using_pck_files:
+
+About mods using PCK files
```
tetrapod00, line 65: Usually headers are are either in the form "Using X for Y" or "Doing X with Y" or sometimes just "X", where X is the name of the feature. One of these suggestions is more idiomatic:
```suggestion
Using PCK files for modding
```
or
```suggestion
Using PCK files for mods
```
or
```suggestion
Modding games with PCK files
```
or
```suggestion
Supporting mods with PCK files
```
We should wait for other docs maintainers to comment before changing this. Note that the anchor and links to this section need to change, too.

tutorials/export/exporting_pcks.rst
```diff
@@ -51,6 +60,23 @@ PCK files usually contain, but are not limited to:
 The PCK files can even be an entirely different Godot project, which the
 original game loads in at runtime.
 
+.. _doc_exporting_pcks_about_mods_using_pck_files:
+
```
tetrapod00, line 64:
```suggestion

.. UPDATE: Recommendation may change. If we ever implement sandboxing or another solution
.. change this section.

```
Recently we started using these kind of comments to keep track of information in the docs that is inherently likely to change.

tetrapod00:
I don't have a strong opinion about this change overall (I really don't disagree with a lot of Max's points). I think that decision should probably be made with more than just the docs team as decisionmakers, though.

If we do go through with these changes, they look alright to me after my suggestions.

skyace65:
Closing this PR. We've discussed that this could work if heavily re-worded. Given that plus the lack of updates I'd rather close this than leave it open indefinitely. Adam, feel free to make a new PR for this subject with updated wording whenever you have the time.

PR #10366, opened 2024-12-04
Add Nim GDExtension

xix-xeaon:
I've been testing the extension for a few different things which have worked well and it has an active developer so I think it can be added to the list.

### DIFF
--- a/tutorials/scripting/gdextension/what_is_gdextension.rst
+++ b/tutorials/scripting/gdextension/what_is_gdextension.rst
@@ -97,6 +97,7 @@ The bindings below are developed and maintained by the community:
 
 - `D <https://github.com/godot-dlang/godot-dlang>`__
 - `Go <https://github.com/grow-graphics/gd>`__
+- `Nim <https://github.com/godot-nim/gdext-nim>`__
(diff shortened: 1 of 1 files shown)


tetrapod00 (approved):
Approval in theory. Someone from GDExtension needs to confirm that is considered well-developed enough to be added.

dsnopek:
> Someone from GDExtension needs to confirm that is considered well-developed enough to be added.

I don't know that anyone on the GDExtension team has directly used the Nim binding, and the developer hasn't (to my knowledge) been in contact with the team either. Given that I don't personally know Nim, and the binding itself is written in Nim (as opposed to being written in C or C++, which sometimes is the case), it's hard to review it. :-)

Poking around on GitHub, it appears the current [Nim binding](https://github.com/godot-nim/gdext-nim) is a continuation of [this project](https://github.com/panno8M/godot-nim), which has been around since May 2023. The two projects have 58 and 73 stars, respectively. That's smaller than the [Nim GDNative bindings for Godot 3](https://github.com/pragmagic/godot-nim) which goes back many more years and has ~500 stars, but that still seems fairly promising to me?

I'd probably err on the side of including it.

mhilbrunner:

PR #10368, opened 2024-12-04
Update functions names in example code to match 4.x nomenclature

Takapapatapaka:
Changed the function to override theme font color from "add_color_override" to "add_theme_color_override", as the previous version is not recognized by my 4.3 Godot Editor. 

Judging by the documentation archives, it seems that "add_color_override" was the version used in Godot 3.

From Control's 4.3 documentation, it seems C# functions have also been updated to include Theme in their name. It would make sense, so I updated it too, but I'm not 100% sure it actually works this way.


tetrapod00 (approved):
This was indeed changed in 4.x.
The C# function names are correct.

mhilbrunner (approved):

mhilbrunner:

PR #10369, opened 2024-12-04
Standardize header syntax within About section

tetrapod00:
This is a proof of concept, before making changes to the whole online documentation.

This standardizes our RST header syntax as follows:
```
Page title
==========
Renders as h1. Every page has this.

Header
------
Renders as h2. Usually appears in sidebar. Many pages only need one level of nested headers.

Sub-header
~~~~~~~~~~
Renders as h3. Appears in sidebar in some pages, depending on how deeply nested the *page* is.

Sub-sub-header
^^^^^^^^^^^^^^
Renders as h4. Usually won't appear in the sidebar.

Sub-sub-sub-header
""""""""""""""""""
Renders as h5. If you got this far something went wrong. I'm not actually sure if we have any 5-layer header nesting in the docs.
```

Feel free to bikeshed the syntax order here, if you care. `=, -, ~` is fairly common, but so is `=, -, ^`. `"""` is rarely used, but it does follow `^^^` when it is used. `+++` is occasionally used.

Edit: after going through a lot more pages, `=, -, ^` is common on pages with only 3 heading levels, but on pages with four heading levels, `=, -, ~, ^` is standard, and `=, -, ^, ~` or any variant is almost never used. So I think my original guideline is still correct.

Note that RST assigns headers based on the order they appear, *within a particular page*. So this PR will not change the header level of any existing pages. It only standardizes our syntax. This also means that this PR is not a one-to-one replacement of particular characters with other particular characters. It depends on the existing page!

Approval of this PR indicates the header syntax that I will standardize the rest of the docs with, in https://github.com/godotengine/godot-docs/pull/10370.

### DIFF
--- a/about/complying_with_licenses.rst
+++ b/about/complying_with_licenses.rst
@@ -84,39 +84,39 @@ how the text has to be included, but here are the most common approaches (you
 only need to implement one of them, not all).
 
 Credits screen
-^^^^^^^^^^^^^^
+~~~~~~~~~~~~~~
 
 Include the above license text somewhere in the credits screen. It can be at the
 bottom after showing the rest of the credits. Most large studios use this
 approach with open source licenses.
 
 Licenses screen
-^^^^^^^^^^^^^^^
+~~~~~~~~~~~~~~~
 
 Some games have a special menu (often in the settings) to display licenses.
 This menu is typically accessed with a button called **Third-party Licenses**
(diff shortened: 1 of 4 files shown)


skyace65 (approved):

mhilbrunner (approved):

skyace65:
Thanks!

PR #10371, opened 2024-12-05
Clarify sentence in Overview of Godot's key concepts

AturAseman:
… to understand


tetrapod00 (approved):
Mild approval. This was discussed in RC, and a couple people thought that the existing version was a bit confusing. I agreed and suggested that OP make a PR for it.

It's worth nothing that this is a small, subjective change, and often wouldn't be worth breaking translations for. But OP is also translating the docs which is how they noticed this at all, so I think it's fine in this case.

mhilbrunner (approved):

mhilbrunner:

PR #10374, opened 2024-12-05
Fix instruction for RISC-V (the previous instruction didn't work)

OS-of-S:
I checked this for Godot 4.3 sources on a Virtual Machine with Ubuntu 22.04.3(LTE). This pull request is the only instruction that worked for me. In all other cases I get an error.
I came to the conclusion that the user does not need to install clang at all to compile for RISC-V, since clang comes with the tool specified in the documentation. Moreover, only this toolset can be used for compilation. Check out details of my experiment:

As you can see below, I have clang installed (version 18.1.8, much better then 16-minimum mentioned in documentation). When I try to use command from actual version of "Compiling for Linux, *BSD" page, I get an error at the end of Godot compilation. ( "scons platform=linuxbsd use_llvm=yes linker=lld" works well therefore it's not problem with my clang )
image

Now I gonna uninstall official clang and check the same scons command after adding $RISCV_TOOLCHAIN_PATH/bin into PATH. Let's uninstall clang and llvm via apt-get purge command. To check that clang were trully uninstalled I trying to check it's version again.
image
image
image
(here I open a new terminal)
image
image
As you can see, now my Linux environment use built-in clang from riscv-gnu-toolchain path and it compiles Godot without errors. But, of course, adding $RISCV_TOOLCHAIN_PATH/bin into PATH would automatically "replaces" the installed Clang with another one if you have it on your system. The same is true vice versa — if the user installs a separate Сlang, then when trying to compile for RISC-V, an inappropriate version can be selected. That's why I recommend not to make adding to PATH permanent and do it every time before RISC-V-compiling.
image

In actual page version said: "RISC-V GCC has bugs with its atomic operations which prevent it from compiling Godot correctly". I **can not** prove that Godot compiled in my case have no such bugs, since I don't have any RISC-V devices. This requires testing (I attached the compiled file at the end of this comment). But, as I understang, GCC and Clang are different compilers, and I don't use GCC in suggested instruction.

[godot.linuxbsd.editor.rv64.llvm](https://drive.google.com/file/d/1KqfWt9hYatyGaub9bML87hPmTj7B9KuK/view?usp=sharing)

### DIFF
--- a/contributing/development/compiling/compiling_for_linuxbsd.rst
+++ b/contributing/development/compiling/compiling_for_linuxbsd.rst
@@ -423,7 +423,7 @@ You also need to add ``$RISCV_TOOLCHAIN_PATH/bin`` to your user's PATH environme
 
 ::
 
-	export PATH="$RISCV_TOOLCHAIN_PATH/bin:$PATH"
+	PATH="$RISCV_TOOLCHAIN_PATH/bin:$PATH"
 
 .. warning:: Since riscv-gnu-toolchain uses its own Clang located
 			 in the ``bin`` folder, this command may block you from
@@ -442,6 +442,13 @@ Go to the root of the source code, and execute the following build command:
     scons arch=rv64 use_llvm=yes linker=mold lto=none target=editor \
         ccflags="--sysroot=$RISCV_TOOLCHAIN_PATH/sysroot --gcc-toolchain=$RISCV_TOOLCHAIN_PATH -target riscv64-unknown-linux-gnu" \
         linkflags="--sysroot=$RISCV_TOOLCHAIN_PATH/sysroot --gcc-toolchain=$RISCV_TOOLCHAIN_PATH -target riscv64-unknown-linux-gnu"
+
+.. note::
+
+    RISC-V GCC has `bugs with its atomic operations <https://github.com/riscv-collab/riscv-gcc/issues/15>`__
+    which prevent it from compiling Godot correctly. That's why Clang is used instead. Make sure that
+    it *can* compile to RISC-V. You can verify by executing this command ``clang -print-targets``,
+    make sure you see ``riscv64`` on the list of targets.
 
 The command is similar in nature, but with some key changes. ``ccflags`` and
 ``linkflags`` append additional flags to the build. ``--sysroot`` points to
--- a/contributing/development/compiling/compiling_for_linuxbsd.rst
+++ b/contributing/development/compiling/compiling_for_linuxbsd.rst
@@ -419,26 +419,13 @@ variable like this:
 This way, we won't have to manually set the directory location
 each time we want to reference it.
 
-You also need to add ``$RISCV_TOOLCHAIN_PATH/bin`` to your user's PATH environment variable.
-
-::
-
-	PATH="$RISCV_TOOLCHAIN_PATH/bin:$PATH"
-
-.. warning:: Since riscv-gnu-toolchain uses its own Clang located
-			 in the ``bin`` folder, this command may block you from
-			 accessing another version of Clang if one is installed.
-			 So it's recommended to run this command every time
-			 before compiling Godot for RISC-V devices instead of adding it to the
-			 ``$HOME/.bash_profile`` file.
-
-
 With all the above setup, we are now ready to build Godot.
 
 Go to the root of the source code, and execute the following build command:
 
 ::
 
+    PATH="$RISCV_TOOLCHAIN_PATH/bin:$PATH"
     scons arch=rv64 use_llvm=yes linker=mold lto=none target=editor \
         ccflags="--sysroot=$RISCV_TOOLCHAIN_PATH/sysroot --gcc-toolchain=$RISCV_TOOLCHAIN_PATH -target riscv64-unknown-linux-gnu" \
         linkflags="--sysroot=$RISCV_TOOLCHAIN_PATH/sysroot --gcc-toolchain=$RISCV_TOOLCHAIN_PATH -target riscv64-unknown-linux-gnu"
@@ -449,6 +436,15 @@ Go to the root of the source code, and execute the following build command:
     which prevent it from compiling Godot correctly. That's why Clang is used instead. Make sure that
     it *can* compile to RISC-V. You can verify by executing this command ``clang -print-targets``,
     make sure you see ``riscv64`` on the list of targets.
+
+.. warning:: The code above includes adding ``$RISCV_TOOLCHAIN_PATH/bin`` to the PATH,
+             but only for the following ``scons`` command. Since riscv-gnu-toolchain uses
+             its own Clang located in the ``bin`` folder, adding ``$RISCV_TOOLCHAIN_PATH/bin``
+             to your user's PATH environment variable may block you from accessing another
+             version of Clang if one is installed. For this reason not recommended to make
+             it permanent. You can also omit the ``PATH="$RISCV_TOOLCHAIN_PATH/bin:$PATH"`` line
+             if you want to use scons with self-installed version of Clang, but it may have
+             compatibility issues with riscv-gnu-toolchain.
 
 The command is similar in nature, but with some key changes. ``ccflags`` and
 ``linkflags`` append additional flags to the build. ``--sysroot`` points to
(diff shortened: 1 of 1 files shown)


skyace65:
@aaronfranke Please review this whenever you get the chance.

contributing/development/compiling/compiling_for_linuxbsd.rst
```diff
@@ -425,6 +419,20 @@ variable like this:
 This way, we won't have to manually set the directory location
 each time we want to reference it.
 
+You also need to add ``$RISCV_TOOLCHAIN_PATH/bin`` to your user's PATH environment variable.
+
+::
+
+	export PATH="$RISCV_TOOLCHAIN_PATH/bin:$PATH"
```
aaronfranke, line 426: As you've mentioned in the warning below, exporting this into your PATH may break things. We shouldn't encourage users to write this command, even just in the terminal. Instead, we should override the variable just for the command, instead of exporting it into the PATH, like this:

```
PATH="$RISCV_TOOLCHAIN_PATH/bin:$PATH" scons arch=rv64 use_llvm=yes linker=mold ...
```

Adjusting the code block below to include this at the start would be welcome.

aaronfranke (commented):
I don't like either of the changes in this PR as-is.

contributing/development/compiling/compiling_for_linuxbsd.rst
```diff
@@ -403,12 +403,6 @@ To cross-compile Godot for RISC-V devices, we need to setup the following items:
   If in doubt, `use this version <https://github.com/riscv-collab/riscv-gnu-toolchain/releases/tag/2021.12.22>`__,
   and download ``riscv64-glibc-ubuntu-18.04-nightly-2021.12.22-nightly.tar.gz``. Extract
   it somewhere and remember its path.
-- Clang. RISC-V GCC has
-  `bugs with its atomic operations <https://github.com/riscv-collab/riscv-gcc/issues/15>`__
-  which prevent it from compiling Godot correctly. Any version of Clang from 16.0.0 upwards
-  will suffice. Download it from the package manager of your distro, and make sure that
-  it *can* compile to RISC-V. You can verify by executing this command ``clang -print-targets``,
-  make sure you see ``riscv64`` on the list of targets.
```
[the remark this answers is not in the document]
OS-of-S: Well, the Clang dosn't work for me! It may be matter on the version you're using. What versions of Clang and riscv-gnu-toolchain do you use?

My versions: Clang 18.1.8 and riscv-gnu-toolchain 2023.06.09

Also a small note about riscv-gnu-toolchain: the official Godot documentation says: "the older the toolchain, the more compatible our final binaries will be", and recommends installing version 2021.12.22-ubuntu-18.04. But this script installs version 2023.06.09-ubuntu-20.04, since this is the oldest version with which I was able to compile Godot. For newer ones I get the error
 	mold: fatal: huf_decompress_amd64.linuxbsd.editor.rv64.llvm.o: incompatible file type: riscv64 is expected but got x86_64
Most likely, a more correct way to install the necessary tools exists (I suspect, related to configuring Clang). It should also be noted that I did not check the functionality of the resulting Godot builds, and given the content of the forms on this topic, there is a high chance that they won't even start.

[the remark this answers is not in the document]
OS-of-S: The rest of the comments make sense and I fully agree with them. I will make corrections regarding the gcc and correct the command for the terminal.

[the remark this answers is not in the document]
aaronfranke: For RISC-V, this platform is so new that the advice about compiling for older Ubuntu versions for better compatibility is superseded by the need to use newer systems for including critical bugfixes. I have upgraded my Ubuntu systems to 24.04 since I last tried this, and I think that is fine as a baseline, given that RISC-V devices are still to this day very niche and will be for several years.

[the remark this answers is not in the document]
OS-of-S: > I have upgraded my Ubuntu systems to 24.04 since I last tried this, and I think that is fine as a baseline, given that RISC-V devices are still to this day very niche and will be for several years.

Ok, so... No changes on this page needed?

I still think that if the toolchain have it's own Clang with which it works fine insted of newer one version of Clang we shouldn't insist on installing different Clang. I don't know how it works, but I have an issue with it. Otherwise we need to add a comment that the user should rely on newer versions of Linux (which is already contrary to the documentation and makes it more complicated for user).

I made changes on all other claims in my PR. Maybe we could add such note in page:

"You can use a self-installed version of Clang, but it may have compatibility issues with riscv-gnu-toolchain."

Will this be enough?

contributing/development/compiling/compiling_for_linuxbsd.rst
```diff
@@ -425,6 +419,20 @@ variable like this:
 This way, we won't have to manually set the directory location
 each time we want to reference it.
 
+You also need to add ``$RISCV_TOOLCHAIN_PATH/bin`` to your user's PATH environment variable.
+
+::
+
+	PATH="$RISCV_TOOLCHAIN_PATH/bin:$PATH"
```
aaronfranke, line 426: Setting this will only work if it's a part of the command as a prefix to it. Otherwise, it will only be changed in the shell itself. https://stackoverflow.com/questions/1158091/defining-a-variable-with-or-without-export

The command below should be edited to include this, it doesn't need its own separate code block.

OS-of-S: Fixed.

contributing/development/compiling/compiling_for_linuxbsd.rst
```diff
@@ -431,10 +425,27 @@ Go to the root of the source code, and execute the following build command:
 
 ::
 
+    PATH="$RISCV_TOOLCHAIN_PATH/bin:$PATH"
     scons arch=rv64 use_llvm=yes linker=mold lto=none target=editor \
```
aaronfranke, line 429: Won't work as-is, you either need to put it on one line or add a `\` to the end.

```suggestion
    PATH="$RISCV_TOOLCHAIN_PATH/bin:$PATH" scons arch=rv64 use_llvm=yes linker=mold lto=none target=editor \
```

aaronfranke (commented):
Looks good to me except for one minor issue.

contributing/development/compiling/compiling_for_linuxbsd.rst
OS-of-S: Ok, I thought about it, but wasn't shure. Done!

aaronfranke (approved):

contributing/development/compiling/compiling_for_linuxbsd.rst
```diff
@@ -431,10 +425,27 @@ Go to the root of the source code, and execute the following build command:
...
+    which prevent it from compiling Godot correctly. That's why Clang is used instead. Make sure that
+    it *can* compile to RISC-V. You can verify by executing this command ``clang -print-targets``,
+    make sure you see ``riscv64`` on the list of targets.
+
+.. warning:: The code above includes adding ``$RISCV_TOOLCHAIN_PATH/bin`` to the PATH,
+             but only for the following ``scons`` command. Since riscv-gnu-toolchain uses
+             its own Clang located in the ``bin`` folder, adding ``$RISCV_TOOLCHAIN_PATH/bin``
+             to your user's PATH environment variable may block you from accessing another
+             version of Clang if one is installed. For this reason not recommended to make
+             it permanent. You can also omit the ``PATH="$RISCV_TOOLCHAIN_PATH/bin:$PATH"`` line
```
skyace65, line 445:
```suggestion
             version of Clang if one is installed. For this reason it's not recommended to make
             adding the bin folder permanent. You can also omit the ``PATH="$RISCV_TOOLCHAIN_PATH/bin:$PATH"`` line
```

skyace65 (changes requested):
Only two minor changes need to be made. This should be good to merge after that.

contributing/development/compiling/compiling_for_linuxbsd.rst
OS-of-S: Done.

skyace65 (approved):
Grammar and spelling looks good to me.

skyace65:

PR #10375, opened 2024-12-06
Add note on Quest3, how to pick renderer, how to contribute to docs

haydenflinner:


about/faq.rst
```diff
@@ -206,6 +206,12 @@ You can find detailed information on supported formats, how to export them from
 your 3D modeling software, and how to import them for Godot in the
 :ref:`doc_importing_3d_scenes` documentation.
 
+What renderer should I use?
+---------------------------
+
+To choose Renderer when making a new project, see: :ref:`doc_internal_rendering_architecture`
```
tetrapod00, line 212: You should check the [latest version of this page](https://docs.godotengine.org/en/latest/about/faq.html#should-i-use-the-forward-mobile-or-compatibility-renderer) 🙂

about/faq.rst
```diff
@@ -622,7 +628,8 @@ for the rest of the game.
 How can I support Godot development or contribute?
 --------------------------------------------------
 
-See :ref:`doc_ways_to_contribute`.
+See :ref:`doc_ways_to_contribute`, or click "Edit on GitHub" on the top-right to add
```
tetrapod00, line 631: I'm not against this kind of call to action to edit the docs directly, but it's probably better suited to be on the page itself, not in the FAQ linking to the page.

about/faq.rst
```diff
@@ -44,15 +44,15 @@ Which platforms are supported by Godot?
 * Windows
 * macOS
 * Linux, \*BSD
-* Android (experimental)
+* Android (experimental, including Meta Quest 3)
```
tetrapod00, line 47: This still seems valid in latest.

tetrapod00 (commented):
Hi! It looks like you're making changes against `4.3` instead of `master`, probably because you used the "Edit on GitHub" button? PRs should be made against `master`, see [here](https://docs.godotengine.org/en/latest/contributing/documentation/contributing_to_the_documentation.html#contributing-changes). (Unfortunately our "Edit on GitHub" links still link to `4.3`).

I'd recommend remaking this PR targeting `master` instead, since one of your changes was actually addressed with a new page in the master/latest version.

about/faq.rst
```diff
@@ -44,15 +44,15 @@ Which platforms are supported by Godot?
...
 * `Web <https://editor.godotengine.org/>`__ (experimental)
 
 **For exporting your games:**
 
 * Windows
 * macOS
 * Linux, \*BSD
-* Android
+* Android (including Meta Quest 3)
```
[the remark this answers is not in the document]
tetrapod00: Yeah, I think it's somewhat reasonable to single out Meta Quest devices for the *XR editor*, since that seems to be the [scope of current support](https://docs.godotengine.org/en/latest/tutorials/editor/using_the_xr_editor.html#xr-devices-support). But for exported projects we shouldn't be this specific.

skyace65:
Closing this due to a lack of a response to reviews

PR #10376, opened 2024-12-06
[4.3] Change Edit on GitHub link to point to master

tetrapod00:
This has been discussed a few times in RC. The choice here is between having the link point to 4.3, and receiving 4.3 PRs which as a general rule we do not accept, or having the link point to master, and a slight chance that a page is renamed/moved and 404s.

From my cursory knowledge of RTD, I think this changing this field won't have any side effects, it only affects the construction of the URL. ~~But I'm not 100% sure and I couldn't find out for sure in the RTD source.~~ It was in the RTD Sphinx theme, not the RTD main repo. This seems to be the only usage of this config value, so it seems like there will be no side effects:
https://github.com/readthedocs/sphinx_rtd_theme/blob/8d4d394dad2d55cf9a4db880effac6aa5c7b12e6/sphinx_rtd_theme/breadcrumbs.html#L39

(And, when we branch the docs to a stable minor version in the future, we should leave this field as `master` rather than the new minor version.)

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -184,7 +184,7 @@ html_context = {
     "display_github": not is_i18n,  # Integrate GitHub
     "github_user": "godotengine",  # Username
     "github_repo": "godot-docs",  # Repo name
-    "github_version": "4.3",  # Version
+    "github_version": "master",  # Version
     "conf_py_path": "/",  # Path in the checkout to the docs root
     "godot_docs_title": supported_languages[language],
     "godot_docs_basepath": "https://docs.godotengine.org/",


mhilbrunner (approved):
Agree with this in principle. I'll leave a chance for others to chime in :)

AThousandShips (approved):
I think this is appropriate, those who are familiar with contributing are more likely to contribute directly by editing the files on GitHub or forking the repo, but the link pointing to a branch other than `master` is a trap for new contributors

The only risk I can see is people who know what they're doing and *do* want to edit 4.3 use this, but that requires them to not pay attention when editing so I think it's safe to make this change, at least worth it to hopefully eliminate PRs on the wrong branch by beginners

mhilbrunner:
Thank you!

tetrapod00:
Edit links on stable still point to 4.3, FYI. Are we just waiting until the next set of cherrypicks to update stable from 4.3, since this isn't the highest priority thing to change/fix (and we may not be keeping stable around, anyway)?

akien-mga:
I mistakenly redid the same change (`master` -> `4.4`) when branching off `4.4`.
We should clarify this in `conf.py` with a comment.

PR #10377, opened 2024-12-06
Update broken link to RTD docs in comment

tetrapod00:
Since RTD doesn't seem to keep old versions of their docs around, I had to guess what the correct link was.

Broken link: https://docs.readthedocs.io/en/latest/vcs.html#github

Best guess at new link: https://docs.readthedocs.io/en/latest/guides/edit-source-links-sphinx.html


mhilbrunner:
Thanks!

PR #10380, opened 2024-12-07
Use more idiomatic English in How to Contribute

tetrapod00:
Pretty marginal change, take it or leave it.
Plural "cooperations" is not really idiomatic English (at least where I am from!). Plural "collaborations" can be used in a similar way.
"Opens the door" is typically more common than "opens the doors", but I don't think the latter form is *wrong*. We're already changing the string, though, so might as well change this too.


clayjohn (approved):

mhilbrunner (approved):

mhilbrunner:
Thanks!

PR #10382, opened 2024-12-07
Briefly document Vector<> variations

esainane:
This continues #10329.

`Packed*Array` aliases seem universally preferred where available, so a link to the list of types seems appropriate.

`LocalVector` is used sparingly, so mentioning the intent and rough tradeoff involved seems right for an overview.

_Bugsquad edit: closes https://github.com/godotengine/godot-docs/issues/6259._

### DIFF
--- a/contributing/development/core_and_modules/core_types.rst
+++ b/contributing/development/core_and_modules/core_types.rst
@@ -104,7 +104,7 @@ memnew/memdelete also use a little C++ magic and notify Objects right
 after they are created, and right before they are deleted.
 
 For dynamic memory, use one of Godot's sequence types such as ``Vector<>``
-or ``LocalVector<>``. ``Vector<>`` behaves much like an STL ``Vector<>``,
+or ``LocalVector<>``. ``Vector<>`` behaves much like an STL ``std::vector<>``,
 but is simpler and uses Copy-On-Write (CoW) semantics. CoW copies of
 ``Vector<>`` can safely access the same data from different threads, but
 several threads cannot access the same ``Vector<>`` instance safely.
--- a/contributing/development/core_and_modules/core_types.rst
+++ b/contributing/development/core_and_modules/core_types.rst
@@ -103,18 +103,25 @@ which are equivalent to new, delete, new[] and delete[].
 memnew/memdelete also use a little C++ magic and notify Objects right
 after they are created, and right before they are deleted.
 
-For dynamic memory, use Godot's ``Vector<>`` or one of its variations.
-Godot's ``Vector<>`` behaves much like an STL ``Vector<>``, but is simpler,
-thread safe, and uses Copy-On-Write semantics.
-It can be safely passed via public API.
+For dynamic memory, use one of Godot's sequence types such as ``Vector<>``
+or ``LocalVector<>``. ``Vector<>`` behaves much like an STL ``Vector<>``,
+but is simpler and uses Copy-On-Write (CoW) semantics. CoW copies of
+``Vector<>`` can safely access the same data from different threads, but
+several threads cannot access the same ``Vector<>`` instance safely.
+It can be safely passed via public API if it has a ``Packed`` alias.
 
-The ``Packed*Array`` :ref:`types <doc_gdscript_packed_arrays>` are aliases for
-specific ``Vector<*>`` types (e.g., ``PackedByteArray``, ``PackedInt32Array``)
-that are accessible via GDScript. Prefer using the ``Packed*Array`` aliases
-when available.
+The ``Packed*Array`` :ref:`types <doc_gdscript_packed_arrays>` are aliases
+for specific ``Vector<*>`` types (e.g., ``PackedByteArray``,
+``PackedInt32Array``) that are accessible via GDScript. Outside of core,
+prefer using the ``Packed*Array`` aliases for functions exposed to scripts,
+and ``Vector<>`` for other occasions.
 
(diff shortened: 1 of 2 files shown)


esainane:
Please do check my understanding is correct - this is based off my reading of the source, and commentary from #10329, but there may be nuances which I'm missing.

tetrapod00 (approved):
Approving *only* for style, and for not seeing anything obviously wrong from my casual understanding of the Godot codebase.

It's good to go as long as it is correct, IMO.

contributing/development/core_and_modules/core_types.rst
```diff
@@ -103,7 +103,18 @@ which are equivalent to new, delete, new[] and delete[].
 memnew/memdelete also use a little C++ magic and notify Objects right
 after they are created, and right before they are deleted.
 
-For dynamic memory, use Vector<>.
+For dynamic memory, use Godot's ``Vector<>`` or one of its variations.
+Godot's ``Vector<>`` behaves much like an STL ``std::vector<>``, but is simpler,
```
Ivorforce, line 107: `Vector` is not fully thread safe. CoW copies of `Vector` can access the same data safely through different threads, but several threads cannot access the same instance of one `Vector` safely.

contributing/development/core_and_modules/core_types.rst
```diff
@@ -103,7 +103,18 @@ which are equivalent to new, delete, new[] and delete[].
...
+thread safe, and uses Copy-On-Write semantics.
+It can be safely passed via public API.
+
+The ``Packed*Array`` :ref:`types <doc_gdscript_packed_arrays>` are aliases for
+specific ``Vector<*>`` types (e.g., ``PackedByteArray``, ``PackedInt32Array``)
+that are accessible via GDScript. Prefer using the ``Packed*Array`` aliases
```
Ivorforce, line 113: I haven't seen this be done very much, is using `Packed` aliases actually recommended?

contributing/development/core_and_modules/core_types.rst
```diff
@@ -103,7 +103,18 @@ which are equivalent to new, delete, new[] and delete[].
...
+when available.
+
+``LocalVector<>`` is a non-COW version, with less overhead. It is intended for
```
Ivorforce, line 116: `LocalVector` is much more like `std::vector` than `Vector`. Maybe it should be covered first, with the intro sentence of comparing to STL `std::vector`?

contributing/development/core_and_modules/core_types.rst
```diff
@@ -103,7 +103,18 @@ which are equivalent to new, delete, new[] and delete[].
 memnew/memdelete also use a little C++ magic and notify Objects right
 after they are created, and right before they are deleted.
 
-For dynamic memory, use Vector<>.
+For dynamic memory, use Godot's ``Vector<>`` or one of its variations.
```
Ivorforce, line 106: `LocalVector` is not a variation of `Vector`. They're two unrelated types. I'd rather phrase it like "Use one of the sequence types" or something like that. `List` has a very different backing technique than both `Vector` and `LocalVector`, but should probably be mentioned here too.

contributing/development/core_and_modules/core_types.rst
```diff
@@ -103,7 +103,18 @@ which are equivalent to new, delete, new[] and delete[].
...
+Godot's ``Vector<>`` behaves much like an STL ``std::vector<>``, but is simpler,
+thread safe, and uses Copy-On-Write semantics.
+It can be safely passed via public API.
```
Ivorforce, line 109: Only if it has a `Packed` alias, I think :)

contributing/development/core_and_modules/core_types.rst
Repiteo: `core` might be the exception, as the aliases are defined in `Variant`. Virtually everything outside of `core` will include `Variant` by default, so in those instances I've generally seen the aliases used more.

timothyqiu: My personal interpretation is to use `Packed*Array` for functions exposed to scripts, and `Vector<>` for other occasions.

esainane:
Cheers for all the reviews, and sorry about the delay! I should be able to take another look at this tonight.

skyace65:
@esainane no rush, but are you planning to continue working on this?

contributing/development/core_and_modules/core_types.rst
```diff
@@ -103,7 +103,25 @@ which are equivalent to new, delete, new[] and delete[].
 memnew/memdelete also use a little C++ magic and notify Objects right
 after they are created, and right before they are deleted.
 
-For dynamic memory, use Vector<>.
+For dynamic memory, use one of Godot's sequence types such as ``Vector<>``
+or ``LocalVector<>``. ``Vector<>`` behaves much like an STL ``Vector<>``,
```
AThousandShips, line 107:
```suggestion
or ``LocalVector<>``. ``Vector<>`` behaves much like an STL ``std::vector<>``,
```

esainane: ...and that definitely counts as something silly which slipped through. Thank you for the catch!

esainane:
I suspect this would read better if moved to the "Containers" section which immediately follows "Allocating memory", but wanted to provide an update for existing feedback first.

Ivorforce (approved):
Looks correct information-wise now.

skyace65 (approved):
Grammar and spelling wise everything looks good

skyace65:
Thanks!

PR #10386, opened 2024-12-08
TIME is affected by time_scale

Ryan-000:
Update docs to say that TIME is affected by time_scale.

### DIFF
--- a/tutorials/shaders/shader_reference/spatial_shader.rst
+++ b/tutorials/shaders/shader_reference/spatial_shader.rst
@@ -98,32 +98,32 @@ Global built-ins
 
 Global built-ins are available everywhere, including custom functions.
 
-+-----------------------------+------------------------------------------------------------------------------------------+
-| Built-in                    | Description                                                                              |
-+=============================+==========================================================================================+
-| in float **TIME**           | Global time since the engine has started, in seconds. It repeats after every ``3,600``   |
-|                             | seconds (which can  be changed with the                                                  |
-|                             | :ref:`rollover<class_ProjectSettings_property_rendering/limits/time/time_rollover_secs>` |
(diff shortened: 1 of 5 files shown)


tutorials/shaders/shader_reference/spatial_shader.rst
```diff
@@ -98,32 +98,32 @@ Global built-ins
...
++-----------------------------+-----------------------------------------------------------------------------------------------------+
+| in bool **OUTPUT_IS_SRGB**  | ``true`` when output is in sRGB color space (this is ``true`` in the Compatibility                  |
+|                             | renderer, ``false`` in Forward+ and Mobile).                                                        |
++-----------------------------+-----------------------------------------------------------------------------------------------------+
+| in float **CLIP_SPACE_FAR** | Clip space far ``z`` value.                                                                         |
+|                             | In the Forward+ or Mobile renderers, it's ``0.0``.                                                  |
+|                             | In the Compatibility renderer, it's ``-1.0``.                                                       |
++-----------------------------+-----------------------------------------------------------------------------------------------------+
```
tetrapod00, line 126: Since the description still fits here without making the table wider, we might instead want to rewrap only the lines which are part of the changed description, and not make the whole table wider.

These tables are already so inconsistent across files it doesn't really matter though.

tetrapod00 (approved):
Good catch, I guess this changed since the original description was written in 3.x.

I experimentally checked this with a simple test script, and it does indeed seem to be affected by time_scale but not by pausing.

tutorials/shaders/shader_reference/spatial_shader.rst
tetrapod00: Yeah, it's okay as-is.

mhilbrunner (approved):

mhilbrunner:

PR #10392, opened 2024-12-09
Improve Physics Interpolation pages

tetrapod00:
Quick editing pass on the Physics interpolation pages to make various minor improvements and bugfixes:
- Reordered the index so that troubleshooting is still the last page.
- Removed commenting on the physics interpolation index page.
- Removed some "Spatial" node references.
- Link to project settings and properties in a few places.
- Fix RST headers. RST headers like `---` and `^^^` do not have an inherent meaning, but are determined by first usage of each in a page. So a couple pages here weren't quite rendering right because `^^^` was used before `---`, despite the apparent intention being that `---` was `h2` and `^^^` was `h3`.

### DIFF
--- a/tutorials/physics/index.rst
+++ b/tutorials/physics/index.rst
@@ -20,5 +20,5 @@ Physics
    collision_shapes_2d
    collision_shapes_3d
    large_world_coordinates
-   troubleshooting_physics_issues
    interpolation/index
+   troubleshooting_physics_issues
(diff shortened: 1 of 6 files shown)


mhilbrunner (approved):

mhilbrunner:
Great work and good catches, quite a haul :) Thanks for the attention to detail, merged.

PR #10394, opened 2024-12-10
Improve C# code sample for saving nodes

tombasche:
I noticed this code sample in the tutorial wasn't very idiomatic C# code, and there is perhaps a more type-safe way to handle saving of nodes within a project using interfaces.

### DIFF
--- a/tutorials/io/saving_games.rst
+++ b/tutorials/io/saving_games.rst
@@ -167,7 +167,7 @@ way to pull the data out of the file as well.
 
     // Note: This can be called from anywhere inside the tree. This function is
     // path independent.
-    // Go through everything in the persist category and ask them to return a
+    // Go through everything in the persist category that implements the ISaveable interface and ask them to return a
     // dict of relevant variables.
     public void SaveGame()
     {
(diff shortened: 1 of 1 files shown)


tetrapod00:
It's worth noting that it was previously considered to use interfaces for this exact page in an earlier issue https://github.com/godotengine/godot-docs/issues/6622#issuecomment-1384679731:
> There are plenty of examples of this in the documentation, see the [Saving Games](https://docs.godotengine.org/en/latest/tutorials/io/saving_games.html) page, which uses duck typing and Godot's JSON serializer while it may better to use interfaces and one of .NET's more powerful serializers.

tombasche:
Interestingly, from the linked issue, this was also one of my concerns when writing this PR 😅 I'm glad it isn't just me.
>  though I'm worried that users may be confused as to where IHittable is coming from, since it's not a type provided by Godot or the BCL, we'll likely want to add some text explaining what it is and then that may be confusing to GDScript users.

tetrapod00:
You might be able to use the second codeblock (under https://docs.godotengine.org/en/latest/tutorials/io/saving_games.html#serializing) to also include the definition of `ISaveable`. Maybe use `//` comments to distinguish which file each line is in, which is often bad practice (separate codeblocks are better) but in this case since the text and codeblocks in these tutorials are designed for GDScript then adapted for C#, there's only so much you can do.

tutorials/io/saving_games.rst
```diff
@@ -118,6 +118,14 @@ The save function will look like this:
         };
     }
 
+    // ISaveable.cs
+    // Define an interface to be used by the overarching Save() method
+    // which the Node you wish to persist will need to implement.
+    public interface ISaveable
+    {
+        public Godot.Collections.Dictionary<string, Variant> Save();
+    }
```
raulsntos, line 127: I don't know if this is clear enough, since we never explicitly show the Node implementing the interface. Also, comments in code can't be translated so that may make the C# code less accessible.

raulsntos (commented):
In general I'm in favor of making C# examples more idiomatic but, as mentioned in the linked issue (https://github.com/godotengine/godot-docs/issues/6622#issuecomment-1384679731), since the text around the code examples is common to all scripting languages we have to be careful that it still makes sense. I'm also a bit concerned about adding important information directly in the code since it can't be translated.

We usually just translate the GDScript examples directly into C# which is easier to maintain. For example, if the GDScript example gets updated, the diff when comparing with the C# example will be smaller. If we want to take a different approach, we may need to be a bit more active maintaining the C# docs so I'd like to hear what other contributors think (feel free to continue this conversation on https://github.com/godotengine/godot-docs/issues/6622).

tutorials/io/saving_games.rst
```diff
@@ -183,21 +191,24 @@ way to pull the data out of the file as well.
                 continue;
             }
 
-            // Check the node has a save function.
-            if (!saveNode.HasMethod("Save"))
+            // Check the node implements the expected interface
+            if (saveNode is ISaveable saveableNode)
```
raulsntos, line 195: You could make the diff smaller by inverting the condition with `is not` and keeping the early continue as before.

```suggestion
            if (saveNode is not ISaveable saveableNode)
```

skyace65:
Closing this due to the lack of a response to raulsntos' review.

PR #10101, opened 2024-10-17
Tweak animation file names to be less confusing in Creating the player scene

Calinou:
- Follow-up to https://github.com/godotengine/godot-docs/pull/9958.

It's more verbose, but it's easier to understand and doesn't rely on the reader knowing how to interpret the `[1/2]` part.

- See https://github.com/godotengine/godot-docs-user-notes/discussions/105#discussioncomment-10967761.

PR #10397, opened 2024-12-10
Correct names for player sprites

JosephHill:


tetrapod00:
Thanks for the contribution! However, there was already an open PR which addressed this, https://github.com/godotengine/godot-docs/pull/10101. Since it is a bit more comprehensive, we merged that one instead. Thanks for reminding us of the problem.

As such, I'm closing this as a duplicate.

(FYI, this PR also targets `4.3` instead of `master`, so it would have needed to be rebased to target `master`).

PR #10409, opened 2024-12-15
Simply remove some words

tetrapod00:
Do we still believe in banning [these words](https://docs.godotengine.org/en/stable/contributing/documentation/docs_writing_guidelines.html#ban-these-8-words)? If so, these are the usages of "simply" that I could find that are unambiguously in the spirit of the rule. They are cases where "simply" is an empty adverb, or where it trivializes the difficulty of learning something new. This includes most cases where the reader is instructed to "simply do something" with an imperative verb.

I left unchanged many examples of "simply" where it is a valid comparison against a more complex approach, or where it really *is* simple to do something in Godot, and this is worth remarking upon. I also left unchanged instances in outdated tutorials where I'm pretty sure the whole page needs to be rewritten.

To justify the changes, I made additional improvements to changed lines where I could find them. (But translations break at the paragraph level, so maybe I should do a second pass on the changed *paragraphs* instead?)

### DIFF
--- a/contributing/development/handling_compatibility_breakages.rst
+++ b/contributing/development/handling_compatibility_breakages.rst
@@ -100,7 +100,7 @@ like ``core/math/a_star_grid_2d.compat.inc`` in this case:
 
     #endif // DISABLE_DEPRECATED
 
-Unless the change in compatibility is complex, the compatibility method should simply call the modified method directly,
+Unless the change in compatibility is complex, the compatibility method should call the modified method directly,
 instead of duplicating that method. Make sure to match the default arguments for that method (in the example above this would be ``false``).
 
 This file should always be placed next to the original file, and have ``.compat.inc`` at the end instead of ``.cpp`` or ``.h``.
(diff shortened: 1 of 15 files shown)


contributing/development/handling_compatibility_breakages.rst
```diff
@@ -100,7 +100,7 @@ like ``core/math/a_star_grid_2d.compat.inc`` in this case:
 
     #endif // DISABLE_DEPRECATED
 
-Unless the change in compatibility is complex, the compatibility method should simply call the modified method directly,
+Unless the change in compatibility is complex, the compatibility method should call the modified method directly,
```
mhilbrunner, line 103: I'd argue this one is not in the spirit of the rule, as the meaning here does not seem to be "this is obvious and/or simple to you, reader", but rather "the compat method should do a simple (as in direct) call in this case, instead of something more involved".

mhilbrunner (approved):

mhilbrunner:
> Do we still believe in banning [these words](https://docs.godotengine.org/en/stable/contributing/documentation/docs_writing_guidelines.html#ban-these-8-words)?

While I'm not sure I'm in favor of it being a hard rule, this improving all these sentences IMO shows that it *is* a good rule, if maybe not one that should be applied 100% in all cases. :)

contributing/development/handling_compatibility_breakages.rst
tetrapod00: The rule also covers empty adverbs like this!

I'd argue that if the reader should *directly* call the method, you could use the world "directly" instead of "simply" here.

PR #10410, opened 2024-12-15
Fix article outdated status for spring arm page

tetrapod00:
Marking this as false is redundant and not idiomatic.


mhilbrunner (approved):

mhilbrunner:
Thank you!

PR #10416, opened 2024-12-17
Clarify creating a global in Singletons (Autoload)

The-oldone:
The code written below in the page to be added to scene 1 and 2 didn't say how to define the global variable, and neither did this, so I fixed it

### DIFF
--- a/tutorials/scripting/singletons_autoload.rst
+++ b/tutorials/scripting/singletons_autoload.rst
@@ -153,8 +153,8 @@ Starting from the menu, open
 **Project > Project Settings > Globals > Autoload** and
 select the script by clicking the browse button or typing its path:
 ``res://global.gd``. Press **Add** to add it to the autoload list
-and name it "Global"(this is required for scripts to access it
-by the name "Global"):
+and name it "Global", which is required for scripts to access it
+by the name "Global":
 
 .. image:: img/autoload_tutorial1.webp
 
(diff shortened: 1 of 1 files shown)


tutorials/scripting/singletons_autoload.rst
```diff
@@ -152,7 +152,9 @@ The next step is to add this script to the autoLoad list.
 Starting from the menu, open
 **Project > Project Settings > Globals > Autoload** and
 select the script by clicking the browse button or typing its path:
-``res://global.gd``. Press **Add** to add it to the autoload list:
+``res://global.gd``. Press **Add** to add it to the autoload list
+and name it "Global"(this is required for scripts to access it
```
AThousandShips, line 156:
```suggestion
and name it "Global" (this is required for scripts to access it
```

tutorials/scripting/singletons_autoload.rst
```diff
@@ -152,7 +152,9 @@ The next step is to add this script to the autoLoad list.
...
+by the name "Global"):
```
tetrapod00, line 157:
```suggestion
and name it "Global", which is required for scripts to access it
by the name "Global":
```
I'd phrase it like this to avoid a parenthetical aside

tetrapod00 (approved):
It's good once either of my suggestion or ATS's is applied.

Worth noting that "Global" does get autofilled when adding the script as a global. So the current text of the page, while technically incomplete, is a valid set of instructions, and this change is not *strictly* needed.

The-oldone:
@tetrapod00 added them, you could merge it now

and yes, it does come in the autocomplete, but as a beginner would copy paste the code of scene 1 and 2 in the example, they might not understand that they have to do that

tetrapod00:
Thanks, and congratulations on your first merged contribution!

PR #10422, opened 2024-12-18
Add Instructions to disable swipe-to-dismiss on WearOS

syntaxerror247:
Closes https://github.com/godotengine/godot-proposals/issues/11357

Implementing the requested feature is not feasible for the Legacy export and can only be achieved for Gradle build exports. As a result, the most practical approach is to handle this manually. This PR adds documentation with clear instructions on how to achieve this manually for Gradle builds.

> [!NOTE]
> The **Gradle build export** section seemed like the most appropriate place to incorporate this instruction. However, if you believe there’s a better location for it, please let me know!

### DIFF
--- a/tutorials/export/android_gradle_build.rst
+++ b/tutorials/export/android_gradle_build.rst
@@ -57,3 +57,20 @@ configuration is needed.
 
     For example, ``_example/image.png`` will **not** be included as an asset,
     but ``_image.png`` will.
+
+Disable Swipe-to-Dismiss on Wear OS
+-----------------------------------
+
+To disable swipe-to-dismiss functionality on Wear OS devices, follow these steps:
+
+1. Open the file ``res://android/build/res/values/themes.xml`` in your file editor.
+2. Add or update the ``android:windowSwipeToDismiss`` attribute to ``false`` in the `GodotAppMainTheme` style.
+
+Example:
+
+.. code-block:: xml
+
+    <style name="GodotAppMainTheme" parent="@android:style/Theme.DeviceDefault.NoActionBar">
+        <item name="android:windowDrawsSystemBarBackgrounds">false</item>
+        <item name="android:windowSwipeToDismiss">false</item>
+    </style>


tutorials/export/android_gradle_build.rst
```diff
@@ -57,3 +57,20 @@ configuration is needed.
 
     For example, ``_example/image.png`` will **not** be included as an asset,
     but ``_image.png`` will.
+
+Disable Swipe-to-Dismiss on Wear OS
+-----------------------------------
+
+To disable swipe-to-dismiss functionality on Wear OS devices, follow these steps:
+
+1. Open the file ``res://android/build/res/values/themes.xml`` in your file editor.
```
tetrapod00, line 66:
```suggestion
1. Open the file ``res://android/build/res/values/themes.xml`` in a text editor.
```
Possibly this?

tutorials/export/android_gradle_build.rst
```diff
@@ -57,3 +57,20 @@ configuration is needed.
...
+2. Add or update the ``android:windowSwipeToDismiss`` attribute to ``false`` in the `GodotAppMainTheme` style.
```
tetrapod00, line 67:
```suggestion
2. Add or update the ``android:windowSwipeToDismiss`` attribute to ``false`` in the ``GodotAppMainTheme`` style.
```
Like this if you want `code`

```suggestion
2. Add or update the ``android:windowSwipeToDismiss`` attribute to ``false`` in the *GodotAppMainTheme* style.
```
Like this if you actually want *italics*

In RST single backticks does "default style" which is a link/reference if you also put _ at the end, and italics otherwise

Also, it's currently slightly ungrammatical, since "Add the attribute to false" isn't grammatical. But the grammatically correct options both seem a bit more awkward, so IDK, it's maybe fine as-is:
```suggestion
2. Add the ``android:windowSwipeToDismiss`` attribute in ``GodotAppMainTheme`` style if you don't already have it. Then set it to ``false``.
```
```suggestion
2. Update the ``android:windowSwipeToDismiss`` attribute to ``false`` in the `GodotAppMainTheme` style. If you don't already have it, add it first.
```

tutorials/export/android_gradle_build.rst
```diff
@@ -57,3 +57,20 @@ configuration is needed.
 
     For example, ``_example/image.png`` will **not** be included as an asset,
     but ``_image.png`` will.
+
+Disable Swipe-to-Dismiss on Wear OS
+-----------------------------------
```
tetrapod00, line 62:
```suggestion
Disabling Swipe-to-Dismiss on Wear OS
-------------------------------------
```
Slightly more idiomatic format for titles

Also, *possibly* also this. The current style guide for the manual is to avoid Actual Title Case for header titles, and only capitalize proper nouns and the first word. Seems like people refer to "swipe to dismiss" several ways, but here is [one instance](https://developer.android.com/training/wearables/views/exit#swipe-to-dismiss) of the official android docs using "swipe-to-dismiss"
```suggestion
Disabling swipe-to-dismiss on Wear OS
-------------------------------------
```

tetrapod00 (approved):
Approving in theory. This seems okay to put in the official docs, but I'd like a second opinion on it, since it's a bit specific.

Putting a specific mini tutorial in this page is slightly awkward. But at a glance I don't really see any better place to put it, so let's put it here and potentially move it if the Android docs grow larger and a better place appears.

tutorials/export/android_gradle_build.rst
```diff
@@ -57,3 +57,20 @@ configuration is needed.
...
+1. Open the file ``res://android/build/res/values/themes.xml`` in your file editor.
+2. Add or update the ``android:windowSwipeToDismiss`` attribute to ``false`` in the `GodotAppMainTheme` style.
+
+Example:
+
+.. code-block:: xml
+
+    <style name="GodotAppMainTheme" parent="@android:style/Theme.DeviceDefault.NoActionBar">
+        <item name="android:windowDrawsSystemBarBackgrounds">false</item>
+        <item name="android:windowSwipeToDismiss">false</item>
```
m4gr3d, line 75: @syntaxerror247 We could add an advanced export option to automatically update this value for gradle builds.

m4gr3d:
> Implementing the requested feature is not feasible for the Legacy export and can only be achieved for Gradle build exports.

It's okay to have functionality that only works for gradle builds - for example, all of the XR functionality requires gradle builds - so we can add the requested feature, and include a warning when it's enabled for the legacy build.

tutorials/export/android_gradle_build.rst
syntaxerror247: @m4gr3d I initially considered this approach but decided to document instructions instead, as it seemed feasible only for Gradle export. However, after seeing your other comment, if it's acceptable to have Gradle-specific build options, I can close this PR and open a new one in the Godot repository to implement it.

Do you think adding this to the project settings in `pointing` would be a better solution?

syntaxerror247:
@m4gr3d Thank you for the clarification! Since it's okay to have functionality that works only for Gradle builds, I will close this PR and work on implementing the feature.

tetrapod00:
Thanks for the quick review here. Adding the feature does seem like a better solution than documenting something niche

PR #10429, opened 2024-12-20
Fix Introduction to 2D scale hot key

lulange:
Change description of the scale tool to correctly suggest the 'S' hot key instead of the 'R' hot key. This issue is likely a carry over of the 'R' hot key being used for scale in the 3D workspace.


mhilbrunner:

PR #10430, opened 2024-12-20
OpenXR: add binding modifiers documentation to action map

BastiaanOlij:
Adding documentation about the new binding modifier logic in the action map.

*Contributed by Khronos Group through the Godot Integration Project*


tutorials/xr/xr_action_map.rst
```diff
@@ -385,3 +385,93 @@ This is why many XR runtimes only use it as a last resort and will attempt to us
...
+.. image:: img/openxr_dpad_modifier.webp
+
+These settings are used as follows:
+
+  * ``Action Set`` defines the action set to which these settings are applied.
+  * ``Input Path`` defines the original input to which these settings are applied.
+  * ``Threshold`` specifies the threshold value that will enable a dpad action, e.g. a value of ``0.6`` means that if the distance from center goes above ``0.6`` the dpad action is pressed.
+  * ``Threshold Released`` specifies the threshold value that will disable a dpad action, e.g. a value of ``0.4`` means that if the distance from center goes below ``0.4`` the dpad action is released.
+  * ``Center Region`` specifies the distance from center that enabled the center action, this is only supported for trackpads.
+  * ``Wedge Angle`` specifies the angle of each wedge.
```
BastiaanOlij, line 443: Requires upstream fix, noticed that it's in radians, will correct that.

BastiaanOlij: Fixed here: https://github.com/godotengine/godot/pull/100633

tutorials/xr/xr_action_map.rst
```diff
@@ -385,3 +385,124 @@ This is why many XR runtimes only use it as a last resort and will attempt to us
   This is our advice as well: limit your action map to the interaction profiles for devices you have actually tested your game with.
   The Oculus Touch controller is widely used as a fallback controller by many runtimes.
   If you are able to test your game using a Meta Rift or Quest and add this profile there is a high probability your game will work with other headsets.
+
+Binding Modifiers
+-----------------
+
+While one of the main goals of the action map is to remove the need for the application to know the hardware used, sometimes the hardware has physical differences that require inputs to be altered in ways other than how they are bound to actions. This need ranges from setting thresholds, to altering the inputs available on a controller.
```
mhilbrunner, line 392: > While one of the main goals of the action map is to remove the need for the application to know the hardware used, sometimes the hardware has physical differences that require inputs to be altered in ways other than how they are bound to actions.

This is a complex sentence that took multiple attempts to parse for me as a non-native speaker. I suggest breaking it up a bit or simplifying if possible :)

tutorials/xr/xr_action_map.rst
```diff
@@ -385,3 +385,124 @@ This is why many XR runtimes only use it as a last resort and will attempt to us
...
+
+Currently OpenXR supports two binding modifiers that have been implemented in Godot but more are likely to follow in the future.
```
mhilbrunner, line 394: > Currently OpenXR supports two binding modifiers that have been implemented in Godot **but more are likely to follow in the future.**

I'd suggest removing that part; its generally not super helpful to users *now*, and also looks bad if we don't end up following it up in the (near) future. We should IMO try to document the current state, not possible futures?

tutorials/xr/xr_action_map.rst
```diff
@@ -385,3 +385,124 @@ This is why many XR runtimes only use it as a last resort and will attempt to us
...
+
+These binding modifiers are not enabled by default and require enabling in the OpenXR project settings. Also there is no guarantee that these modifiers are supported by every runtime. You will need to consult the support for the runtimes you are targeting and decide whether to rely on the modifiers or implement some form of fallback mechanism.
+
+If you are targeting multiple runtimes that have support for the same controllers, you may need to create separate action maps for each runtime. You can control which action map Godot uses by using different export templates for each runtime and using a custom :ref:`feature tag <doc_feature_tags>` to set the action map.
+
+In Godot, Binding modifiers are divided into two groups. Modifiers that work on the interaction profile level, and modifiers that work on individual bindings.
```
mhilbrunner, line 400:
```suggestion
In Godot, binding modifiers are divided into two groups: modifiers that work on the interaction profile level, and modifiers that work on individual bindings.
```

tutorials/xr/xr_action_map.rst
```diff
@@ -385,3 +385,124 @@ This is why many XR runtimes only use it as a last resort and will attempt to us
...
+Binding modifiers that are applied to the whole interaction profile can be accessed through the modifier button on the right side of the interaction profile editor.
+
+.. image:: img/openxr_ip_binding_modifier.webp
+
+You can add a new modifier by pressing the ``Add binding modifier`` button.
+
+.. warning::
+  As Godot doesn't know which controllers and runtimes support a modifier,
+  there is no restriction to adding modifiers.
+  Unsupported modifiers will be filtered out at runtime.
```
mhilbrunner, line 414: Is "will be ignored" more fitting here instead of "filtered out"? Not important tbh.

tutorials/xr/xr_action_map.rst
tetrapod00:
```suggestion
One of the main goals of the action map is to remove the need for the application to know the hardware used. However, sometimes the hardware has physical differences that require inputs to be altered in ways other than how they are bound to actions. This need ranges from setting thresholds, to altering the inputs available on a controller.
```
Quickest fix is to use "however". There's probably an even better fix though

tutorials/xr/xr_action_map.rst
tetrapod00: Yes. In general the docs should:
- Document current state, written in present tense, in a way that will still read well in the future
- Avoid using the word "currently", though! It's usually rendundant
- Avoid mentioning future plans like this

Source: I looked through the docs recently and found a lot of instances where we didn't follow these guidelines and things needed to be corrected

tutorials/xr/xr_action_map.rst
```diff
@@ -385,3 +385,124 @@ This is why many XR runtimes only use it as a last resort and will attempt to us
...
+
+Dpad Binding modifier
```
tetrapod00, line 416: I don't have time to look thoroughly right now but "dpad" is a word we should check the standardization of. From a quick search I see "DPad", "dpad", and "D-pad" even with XR docs specifically

This can probably be solved in a future PR just making a note

tutorials/xr/xr_action_map.rst
```diff
@@ -385,3 +385,124 @@ This is why many XR runtimes only use it as a last resort and will attempt to us
...
+^^^^^^^^^^^^^^^^^^^^
+
+Modifiers can support automatic haptic output that is triggered when thresholds are reached.
+
+.. note::
+  Currently both available modifiers support this feature however there is no rule future modifiers also have this capability.
+  Only one type of haptic feedback is supported but in the future other options may become available.
+
+Haptic vibration
+================
```
tetrapod00, line 498:
```suggestion
""""""""""""""
```
`===` is already used for the title of this page. I think you mean to use `"""` here, judging by existing usages

tetrapod00 (commented):
Not a full review

tutorials/xr/xr_action_map.rst
```diff
@@ -385,3 +385,124 @@ This is why many XR runtimes only use it as a last resort and will attempt to us
...
   If you are able to test your game using a Meta Rift or Quest and add this profile there is a high probability your game will work with other headsets.
+
+Binding Modifiers
+-----------------
+
+While one of the main goals of the action map is to remove the need for the application to know the hardware used, sometimes the hardware has physical differences that require inputs to be altered in ways other than how they are bound to actions. This need ranges from setting thresholds, to altering the inputs available on a controller.
+
+Currently OpenXR supports two binding modifiers that have been implemented in Godot but more are likely to follow in the future.
+
+These binding modifiers are not enabled by default and require enabling in the OpenXR project settings. Also there is no guarantee that these modifiers are supported by every runtime. You will need to consult the support for the runtimes you are targeting and decide whether to rely on the modifiers or implement some form of fallback mechanism.
```
tetrapod00, line 396:
```suggestion
These binding modifiers are not enabled by default and require enabling in the
OpenXR project settings. Also there is no guarantee that these modifiers are
supported by every runtime. You will need to consult the support for the
runtimes you are targeting and decide whether to rely on the modifiers or
implement some form of fallback mechanism.
```
One other thing, which applies to this whole PR.

In the manual the rule is usually to wrap long lines by hand after 80-100 characters. This was an informal rule for a long time and I recently [wrote it down](https://docs.godotengine.org/en/latest/contributing/documentation/docs_writing_guidelines.html#manually-wrapping-lines). (I'd say wrapping at 80 is ideal and wrapping after 100 is mandatory. Allowing some room between 80 and 100 reduces some diff noise since later minor changes don't *force* a re-wrap.)

As it stands, the manual isn't exactly standardized. But I think we should avoid introducing *new* very long lines like this one.

tetrapod00: (I also think that it would be very nice if we could avoid this rule somehow, and instead use one line per paragraph and let the tools automatically wrap, like the class reference XML. But I don't know if the tooling support is quite there yet.

tutorials/xr/xr_action_map.rst
BastiaanOlij: Merged upstream.

tutorials/xr/xr_action_map.rst
BastiaanOlij: As a non-native speaker, I'm struggling to put this in the right words for sure :)

The main thing is that just binding inputs to actions isn't enough to cater for the physical differences between devices, further modifications to the inputs are required per device. Hence the binding modifier system being introduced.

tutorials/xr/xr_action_map.rst
BastiaanOlij: Agreed, removing this.

tutorials/xr/xr_action_map.rst
BastiaanOlij: Yeah I think that makes sense. We're filtering them out and never supplying them to OpenXR (OpenXR will fail action maps that include things that are not supported on the runtime) but that is a technical implementation detail.
Towards a user saying they are ignored is probably better.

tutorials/xr/xr_action_map.rst
BastiaanOlij: I'll change it to "dpad" for the OpenXR docs. Thats the name they are using in the binding modifier spec: https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#XR_EXT_dpad_binding

tetrapod00: Makes sense, seems fine to me.

At some point I might take a look at what we use in general when referring to the dpad and standardize some more. It may be the case that we end up using "DPad" or "D-pad" generally and only using "dpad" when referring to the spec.

tutorials/xr/xr_action_map.rst
BastiaanOlij: Yeah its annoying to follow through on this, especially when you're still editing things and suddenly the linebreaks need to be in different places. Not to mention the trailing spaces that you easily introduce.

But I'll try and make sure the document is properly hard-wrapped.

tutorials/xr/xr_action_map.rst
BastiaanOlij: Yeah sounds good, it makes sense but it sounds like a lot of work :)

BastiaanOlij:
Ok I think I've got all feedback covered.

I hard-wrapped the whole document so including the existing part of the documentation..

tetrapod00 (approved):

tetrapod00:
Seems okay to merge but ideally someone else can review for XR first

mhilbrunner (approved):

mhilbrunner:
Gonna trust Bastiaan on XR stuff. :) Looks good to me, thank you and happy holidays!

PR #10023, opened 2024-10-01
Remove individual youtube channel links from tutorials page

skyace65:
Does what the title says.

Currently on the community tutorial page we link to individual youtube channels that provide Godot tutorials. I believe we should stop doing things this way for several reasons.

One, the foundation has to review any PR that adds a channel to this page. I'm not against that but the foundation doesn't seem to be actually reviewing anything. The oldest open video PR is from April this year which is almost 5 months old. I'm not saying this to criticize the foundation for not doing this, merely pointing out that the process has created a backlog. I also think their time is better spent elsewhere for this next reason.

Two, going to our tutorials page is less efficient than just going to YouTube and searching for videos. On the tutorials page we only break down creators into broad categories, "2D" covers a wide array of topics. If someone was trying to find a creator going over, for example, tilemaps, it's not easy to find out who is doing that without going through every channel. And in comparison if someone goes straight to YouTube they just have to search for "Godot TileMapLayer" tutorial to immediately find several relevant results.

There's currently only one video tutorial on the page that links somewhere other than YouTube that I've also removed. It just feels odd to keep that linked when there's no other sites we also link to. Though I'm not against re-adding it to this PR.

I welcome all discussion on if these channel links should be removed or not.

Lastly, I've also edited the sentence before the video section to sound better.

PR #10431, opened 2024-12-20
Fixed grammar in tutorials.rst, and added my Godot YouTube channel to the devlogs section

QuebleGameDev:

### DIFF
--- a/community/tutorials.rst
+++ b/community/tutorials.rst
@@ -35,6 +35,11 @@ Text tutorials
 - `Game Dev Artisan website <https://gamedevartisan.com/>`__
 - `Night Quest Games Blog <https://www.nightquestgames.com/blog-articles/>`__
 
+Video tutorials
+--------------
+
+- `Queble (2D, GDScript) <https://www.youtube.com/@queblegamedevelopment4143/videos>`_
+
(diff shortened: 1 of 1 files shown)


tetrapod00:
Hi! Thanks for your interest in contributing. However, I don't think this is in scope to link to from the official documentation.

We recently made the decision to remove existing links to individual youtube tutorials, and no longer curate such a list in the official documentation. See https://github.com/godotengine/godot-docs/pull/10023. I'm actually not sure why we left the two links to "devlogs" when removing links to video tutorials. That might have been an oversight.

After looking at your channel, the videos definitely look like tutorials, so I don't think this is in scope.

(The grammar fix is still very welcome. You can either make a change to this PR or submit another one with just the grammar fix)

QuebleGameDev:
Ah I see, I guess I wasn't sure why I had been removed initially, but that makes more sense 👍

I'll change this asap to remove the devlogs section, and just keep the grammar change.

tetrapod00:
I opened a new PR just for the grammar fix. Going to close this, since the tutorial link won't be added with our current policy. Thanks for the contribution nonetheless!

PR #10432, opened 2024-12-20
Remove links to YouTube devlogs

tetrapod00:
I'm not sure about this, and it needs full consensus from the docs team.

In https://github.com/godotengine/godot-docs/pull/10023, we removed a curated list of YouTube tutorials. However, we left two links to YouTube devlogs:
- https://www.youtube.com/@bitbraindev/videos
- https://www.youtube.com/@devduck/videos

I suggest removing those, since it seems a bit strange to leave an exception on the page specifically for devlogs. Was there a specific reason to do so? With devlog links we still have the same problem of needing a foundation review for new additions.

Additionally, one of the *two* currently linked devlogs includes a mix of short tutorials and devlog content. If we leave devlogs on the page, it may function as a loophole to get a specific channel linked from the official docs. (To be clear I don't mean to single out a specific creator here, but it's hard when the page only links two of them.)

### DIFF
--- a/community/tutorials.rst
+++ b/community/tutorials.rst
@@ -35,12 +35,6 @@ Text tutorials
 - `Game Dev Artisan website <https://gamedevartisan.com/>`__
 - `Night Quest Games Blog <https://www.nightquestgames.com/blog-articles/>`__
 
-Devlogs
--------
-
-- `bitbrain <https://www.youtube.com/@bitbraindev>`_
-- `DevDuck (2D) <https://www.youtube.com/@devduck/videos>`_
-
 Resources
 ---------
 


mhilbrunner (approved):
Approving, although I guess we are starting to reach the point where we should be re-considering the whole page. :') Starting to be pretty bare bones.

It may make sense to change it to focus mostly on the resources?

tetrapod00:
There was a bit of prior discussion of reconsidering the whole page in the RC a few months back (https://chat.godotengine.org/channel/documentation?msg=D5bBD6pJW7p4dqDaT)

I'm okay with holding off on this particular change in favor of a dedicated discussion about the future purpose of this page. We should figure out what our *actual* policy is for linking to stuff rather than changing it ad-hoc in bits and pieces, I think.

Note that if we focus more on just linking to resources, one of the major resources linked to is https://github.com/godotengine/awesome-godot. But the [section of that readme file about tutorials](https://github.com/godotengine/awesome-godot?tab=readme-ov-file#tutorials) just links back to the docs page. So that's something we would want to untangle.

skyace65 (approved):
Honestly I just didn't see the devlogs when I made my PR that removed the youtube links. I'm fine with removing these as well.

fire (approved):

skyace65:
Thanks!

PR #10449, opened 2024-12-30
Update the compilation instruction for linux

xsellier:
Since the master branch crashes using GCC 12, I think updating the requirement to GCC 13 might be a good idea (see issue: https://github.com/godotengine/godot/issues/100768 )

Also, since there's a publicly available toolchian that's easy to install, and doesn't mess up with the user's machine, I'd rather add this as an option.

### DIFF
--- a/contributing/development/compiling/compiling_for_linuxbsd.rst
+++ b/contributing/development/compiling/compiling_for_linuxbsd.rst
@@ -16,7 +16,7 @@ Requirements
 For compiling under Linux or other Unix variants, the following is
 required:
 
-- GCC 9+ or Clang 6+.
+- GCC 13+ or Clang 6+.
 - `Python 3.8+ <https://www.python.org/downloads/>`_.
 - `SCons 4.0+ <https://scons.org/pages/download.html>`_ build system.
 - pkg-config (used to detect the development libraries listed below).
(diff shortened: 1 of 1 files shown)


contributing/development/compiling/compiling_for_linuxbsd.rst
```diff
@@ -16,7 +16,7 @@ Requirements
 For compiling under Linux or other Unix variants, the following is
 required:
 
-- GCC 9+ or Clang 6+.
+- GCC 13+ or Clang 6+.
```
akien-mga, line 19: I don't think we should change the minimum supported version due to a bug we caused. We should just fix that bug.

contributing/development/compiling/compiling_for_linuxbsd.rst
```diff
@@ -42,6 +42,19 @@ Distro-specific one-liners
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 .. tabs::
+    .. tab:: All distros x84_64
```
akien-mga, line 45: This is worth documenting, but this is a bit too power user with only a script and no explanations about what this is for and why it's a good option.

It's also not a distro one-liner.

I'd make a separate section for it, and explain the purpose of this Linux SDK and why we recommend it for production builds.

contributing/development/compiling/compiling_for_linuxbsd.rst
xsellier: Sounds good to me

contributing/development/compiling/compiling_for_linuxbsd.rst
xsellier: Made a new section at the very end of the documentation

contributing/development/compiling/compiling_for_linuxbsd.rst
```diff
@@ -612,3 +612,24 @@ running ``scons -h``, then looking for options starting with ``builtin_``.
     across Linux distributions anymore. Do not use this approach for creating
     binaries you intend to distribute to others, unless you're creating a
     package for a Linux distribution.
+
+Using Godot official toolchain for production build
+---------------------------------------------------
+
+Contains the Godot buildroot to generate toolchains for building the Godot engine in
+a portable way for Linux. Using these toolchains is the best way to distribute Linux
+builds of your custom-compiled Godot game.
```
skyace65, line 621: Shouldn't this be custom-compiled Godot engine, not game?

contributing/development/compiling/compiling_for_linuxbsd.rst
```diff
@@ -612,3 +612,24 @@ running ``scons -h``, then looking for options starting with ``builtin_``.
     across Linux distributions anymore. Do not use this approach for creating
     binaries you intend to distribute to others, unless you're creating a
     package for a Linux distribution.
+
+Using Godot official toolchain for production build
+---------------------------------------------------
+
+Contains the Godot buildroot to generate toolchains for building the Godot engine in
```
skyace65, line 619:
```suggestion
This creates the Godot buildroot to generate toolchains for building the Godot engine in
```

skyace65 (changes requested):

skyace65:
Closing this due to the lack of a response. Feel free to ping me if you want to continue working on this.

PR #10451, opened 2024-12-30
Clarify FlowContainer get_line_count explanation

idbrii:
Make it more obvious that it's not giving you the number of items in  a line and avoid restating the name of the function as  documentation.


tetrapod00:
These changes have to be made in the main repo, in this case [here](https://github.com/godotengine/godot/blob/master/doc/classes/FlowContainer.xml). The files in this repo are just a copy, not the source of truth. See [this page](https://docs.godotengine.org/en/latest/contributing/documentation/updating_the_class_reference.html) for more info

PR #10452, opened 2024-12-30

koubaa:
Make the 2d tutorial slightly more straightforward

Handle assigning the position before setting the direction


tetrapod00:
In theory I agree, since the position is set between two lines of code about setting the direction.
Note that the demo project has a completed version, too, and if we change the tutorial, the completed version will no longer match. The relevant line in the completed project is: https://github.com/godotengine/godot-demo-projects/blob/master/2d/dodge_the_creeps/main.gd#L32-L36

koubaa:
https://github.com/godotengine/godot-demo-projects/pull/1151

PR #10463, opened 2025-01-03
Document using browser search shortcuts to open issues faster

aaronfranke:
@akien-mga mentioned that this was possible in RocketChat, and I figured it was good to put in the docs too.

### DIFF
--- a/contributing/workflow/index.rst
+++ b/contributing/workflow/index.rst
@@ -20,3 +20,4 @@ approach the project.
    pr_workflow
    pr_review_guidelines
    testing_pull_requests
+   tips_and_tricks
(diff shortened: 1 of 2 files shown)


akien-mga:
I think it's a cool trick but a bit too much of a power user advice to be featured so prominently in an intro page like this IMO.

mhilbrunner:
I agree with Akien. Cool, but this is more a "how do I use my browser / other software and Github well" than really Godot specific. Considering we're already trying to limit which *Godot-specific* content we add and maintain to keep it manageable, this doesn't quite meet the bar for me, IMO. Happy to hear others however :)

aaronfranke:
¯\\\_(ツ)_/¯

PR #10474, opened 2025-01-06
CI updates (master branch)

mhilbrunner:
Happy 2025! Time for some ~~spring~~ cleaning in the new year:
- Update from Ubuntu `22.04` to `24.04` (and replace all use of `ubuntu-latest` with `24.04` for consistency)
    - Update workflows to use Python `venv`s, as newer Ubuntu (and Debian) versions don't want us trashing global packages with Python packages (see [here](https://www.omgubuntu.co.uk/2023/04/pip-install-error-externally-managed-environment-fix))
      - Using virtualenvs also means we now need to exclude directories starting with `.` from Sphinx builds, to support `.venv` (this would have been a good idea anyway, for `.git`, `.vscode` and so on)
- Fix permission issues (by adding write permissions to repo contents and/or PRs where those are now needed, see [here](https://graphite.dev/guides/github-actions-permissions), originally added around 2021, now mandatory), this should hopefully fix the issues with https://github.com/godotengine/godot-docs/pull/10361)
-  (Hopefully) improve CI build times by utilizing up to 4 CPUs (which according to [docs](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners) seems now save to assume for Github action runners)
- Change all references to `github.ref` to `github.ref_name`, which should be equivalent to the old behaviour (see [here](https://github.blog/changelog/2023-09-13-github-actions-updates-to-github_ref-and-github-ref/))
- Set timeouts (by default, Github waits 6 hours before killing a hung job, whereas all our stuff should be killed with fire way earlier, most take a few minutes a most, with building the offline docs being the longest at 1-2 hours, not 6)
- Update some comments that have grown old by now

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,5 +9,7 @@ updates:
     schedule:
       interval: "daily"
     ignore:
-      # ReadTheDocs is staying on v1.
+      # We need to decide on when we upgrade Sphinx manually,
+      # as historically, this has been proven to often imply larger changes
+      # (RTD compat, upgrading extensions, other dependencies, our content, etc.).
       - dependency-name: "sphinx"
--- a/.github/workflows/sync_class_ref.yml
+++ b/.github/workflows/sync_class_ref.yml
@@ -3,9 +3,10 @@ name: Sync Class Reference
 on:
   workflow_dispatch:
   # Scheduled updates only run on the default/master branch.
+  # Other branches need to be run manually (usually after a new release for that branch).
   schedule:
-    # Run it at night (European time) every Saturday.
-    # The offset is there to try and avoid the high load times.
+    # Run it at (European) night time every Saturday.
+    # The offset is there to try and avoid high load times.
     - cron: '15 3 * * 6'
 
 # Make sure jobs cannot overlap.
--- a/.github/workflows/sync_class_ref.yml
+++ b/.github/workflows/sync_class_ref.yml
@@ -22,6 +22,9 @@ jobs:
     runs-on: ubuntu-24.04
     env:
       engine_rev: 'master'
+    permissions:
+      contents: write
+      pull-requests: write
 
     steps:
       - name: Checkout the documentation repository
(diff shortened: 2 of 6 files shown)


.github/workflows/sync_class_ref.yml
```diff
@@ -22,6 +23,9 @@ jobs:
     runs-on: ubuntu-latest
```
tetrapod00, line 23: Should this and the cherrypick workflow be pinned to `ubuntu-24.04` rather than `ubuntu-latest` too? I'm pretty new to GitHub Actions so I'm not sure if there's a relevant difference here.

When I wrote the auto-cherrypicker I think I copied the existing `ubuntu-latest`, FWIW.

.github/dependabot.yml
```diff
@@ -9,5 +9,5 @@ updates:
     schedule:
       interval: "daily"
     ignore:
-      # ReadTheDocs is staying on v1.
+      # We need to decide on this one manually.
```
tetrapod00, line 12: This comment perhaps is unnecessary? There's no mention of *why* we decide this one manually. And as-is it just describes what the `ignore:` section does.

.github/workflows/sync_class_ref.yml
mhilbrunner: Eh, tradeoffs. Pinning it means new contributors have less chance of *their* specific PR suddenly failing for reasons outside their control because the underlying dependencies changed at just the right (or wrong) moment, but also means we don't get small breakages as they happen and can quickly fix them, but rather we get a whole mess of them once we upgrade.

I think *may* be worth favoring new contributor experience over our ease of upgrading? Anyway, in this case we should just do it so everything is the same for consistency.

.github/dependabot.yml
mhilbrunner: Updated to add some reasoning. Good point. :) Also considered just deleting, but we may as well write down some tribal knowledge.

.github/workflows/sync_class_ref.yml
mhilbrunner: Done, replaced all remaining `ubuntu-latest` with `24.04`. CI passed, so seems to work fine. 🤞

mhilbrunner:
Let's see if anything breaks, all tests so far seem good 👀 Merging, as I want to do further work on GH actions and this needs to go in first... thanks for having a look, tetra!

PR #10489, opened 2025-01-09
Update screenshot of Editor Settings location for Exporting for Android

AKKamath:
The Exporting for Android webpage currently shows the "Editor Settings" being under the Editor tab in the screenshot:
https://docs.godotengine.org/en/stable/tutorials/export/exporting_for_android.html

In Godot v4.3, this is under the Godot tab, which is updated in this screenshot.


Calinou:
> In Godot v4.3, this is under the Godot tab, which is updated in this screenshot.

This is only the case on macOS, and it's been like this since 4.0 at least. On other platforms (or if you disable **Expand to Title** in the Editor Settings), it's still in the Editor tab.

AKKamath:
I'm sorry, I was not aware of this difference. I've updated the PR and removed the screenshot change, and instead directly edited the text as you've mentioned.

skyace65 (approved):

skyace65:
Thanks!

PR #10492, opened 2025-01-10
Update random_number_generation.rst

PierreDessare:
Fixed a typo in the C# version of randi_range().
It was noted RandiRange instead of RandRange.
Correction based on :
https://docs.godotengine.org/en/stable/classes/class_%40globalscope.html#class-globalscope-method-randi-range

### DIFF
--- a/tutorials/math/random_number_generation.rst
+++ b/tutorials/math/random_number_generation.rst
@@ -179,7 +179,7 @@ and ``to``, and returns a random integer between ``from`` and ``to``:
  .. code-tab:: csharp
 
     // Prints a random integer between -10 and 10.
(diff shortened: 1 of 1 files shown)


raulsntos (approved):
Nice catch, thanks! The C# bindings use `GD.RandRange` for both "int" and "float" by providing 2 overloads. It'd be nice if you could also fix the `GD.RandfRange` example just above.

Adding cherry-pick labels since this applies since Godot 4.0.

raulsntos:
My bad, it looks like this has already been fixed by https://github.com/godotengine/godot-docs/pull/8072. Thanks for the contribution nonetheless.

PR #10500, opened 2025-01-13
Add refresher tips for how to do tasks, explanations, minor style changes

y-almannaee:
This pull request changes the following things on the "Creating the enemy" page under the "Your first 2D game" tutorial:

- Explains how to change scenes, reminds the user to save their new Mob scene
- Adds a tip to remind users how to group a node with its children
- Explains what the Layer and Mask selections do in the CollisionObject2D
- Adds a tip to remind users how to add animation sprites
- Adds a comment for other contributors to update a code section with changes to match the example project
- Adds a tip to remind users how to connect signals to nodes
- Adds a note on how users can learn what an unknown function does in the Script Editor
- Minor style updates, manually wrap lines

Please review some of the additions for verbosity. Thank you!

### DIFF
--- a/getting_started/first_2d_game/04.creating_the_enemy.rst
+++ b/getting_started/first_2d_game/04.creating_the_enemy.rst
@@ -10,19 +10,38 @@ choose a random direction, and move in a straight line.
 We'll create a ``Mob`` scene, which we can then *instance* to create any number
 of independent mobs in the game.
 
+To create the ``Mob`` scene, click Scene > New Scene from the main menus at the top.
+
+You can switch Scenes using the tabs above the Toolbar. There should be a Player
+scene (from the previous section), and a new scene called "[empty]", which is what
+we've just created. Ensure you are editing the new "[empty]" scene by clicking
+on it. Save this scene as "Mob".
+
 Node setup
 ----------
 
-Click Scene -> New Scene from the top menu and add the following nodes:
+Add the following nodes to the mob's scene:
 
-- :ref:`RigidBody2D <class_RigidBody2D>` (named ``Mob``)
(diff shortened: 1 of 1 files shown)


skyace65 (commented):
In my opinion none of the tips, notes and reminders in this PR should be added. The best thing for a text tutorial is to make it as short and concise as possible. Adding notes and reminders for things people were already taught how to do in this tutorial, and the introduction and step by step sections of the getting started section (which we explicitly tell people to read first), goes against that. Anyone who needs these reminders should be watching a video tutorial instead so they can always see what someone is clicking on.

Additionally, you should not be adding a note asking other people to update code. For starters if you find something out of date and can't update it yourself you should open an issue. The amount of comments we would have if every out of date part of the docs had one would be huge considering we have over 800 open issues. And two, you already have the code you need to update this, if you think the demo repository code is more correct/works better then you should change it yourself in this PR, we have no restrictions on who can update code examples

tetrapod00:
I mostly agree with skyace here. In isolation, one or two of these tips might be appropriate additions. In general I tend to favor including *small* additions and clarifications to these pages, since each addition represents a reader's confusion that they considered worth making a PR for, and is in most cases shared by other readers. But taken all together these changes are too much, for the reasons skyace mentioned.

I think one alternative to consider is to add some more header sections within each Getting Started tutorial page. Then, when a later tutorial needs to reference the content of an earlier tutorial, we could link directly to the header section rather than either vaguely referencing the contents (like we currently do with `Don't forget to set the children so that they cannot be selected, like you did with the Player scene`), or explicitly listing the steps again as this PR does. That would be a much larger project, though, and is definitely outside the scope of just this PR.

skyace65:
Closing due to the lack of a response to reviews after 2 weeks

PR #10501, opened 2025-01-13
Clarify why the planemesh looks odd

The-oldone:
It confused me for a decent amount of time why it said the lighting looked flat, I believe this will clarify it as the mesh isn't a flat plane as currently written. It isn't mentioned anywhere either that this randomness is to make the material seem more grainy.

### DIFF
--- a/tutorials/shaders/your_first_shader/your_first_3d_shader.rst
+++ b/tutorials/shaders/your_first_shader/your_first_3d_shader.rst
@@ -289,9 +289,9 @@ First, we will add an :ref:`OmniLight3D<class_OmniLight3D>` to the scene.
 .. image:: img/light.png
 
 You can see the light affecting the terrain, but it looks odd. The problem is
-the light is affecting the terrain as if it were a flat plane. This is because
-the light shader uses the normals from the :ref:`Mesh <class_mesh>` to calculate
-light.
+the light is affecting the terrain as if it were made of a perfectly flat
+material without irregularities. This is because the light shader uses the
+normals from the :ref:`Mesh <class_mesh>` to calculate light.
 
 The normals are stored in the Mesh, but we are changing the shape of the Mesh in
 the shader, so the normals are no longer correct. To fix this, we can


tetrapod00:
I think the unchanged version is more correct. The problem really is that the `Mesh` resource being used is a flat plane, and the vertex normal values are those of a flat plane. So "as if it were a flat plane" is correct. When the vertices are offset in the `vertex()` function, that doesn't change the vertices of the Mesh resource, and it doesn't change the normals of those vertices. The next paragraph explains this already.

Changing it to "perfectly flat material without irregularities" is less correct, and potentially misleading. "Flat" in the context of *materials* or *shaders* usually refers to "flat shading" or "flat interpolation" of vertex values.

There may still be some wording or clarity improvements that can be made here. But as is, I don't think this change is good to merge.

skyace65 (commented):
I agree with Tetrapod. I don't see these changes as an improvement, and I don't the existing phrasing as incorrect.

The-oldone:
> The problem really is that the `Mesh` resource being used is a flat plane, and the vertex normal values are those of a flat plane.

Does this mean that the lighting will land on the mesh according to the modified vertices, but will get reflected as if the normals were of a flat plane? Sorry I'm new to this

And if that is the case, why does adding a normal map only change the texture of the material? As far as I can see, the texture of the material obtains some bumps all across the material in the documentation. Am I wrong and the lighting actually changes from being that of a flat plane to that of a bumpy plane? And final question, if we have to change the lighting to reflect the light from the changed mesh properly, why are we using two different randomly generated textures? Unless they are of the same seed, they will have a different set of bumps, and so will not produce the same set of normals as that of the first texture.


P.S. I'm on mobile, so I can't select the comment to reply to

tetrapod00:
> Does this mean that the lighting will land on the mesh according to the modified vertices, but will get reflected as if the normals were of a flat plane?

It will have the position of the modified (moved) vertices, but the normals of the flat plane. I think that's what you mean?

> if we have to change the lighting to reflect the light from the changed mesh properly, why are we using two different randomly generated textures?

I believe the intention is that the two textures *should* be using the same settings and seed, with the only difference one is a normal map. If that is the case, then we can use the normal map version of the noise texture as the normals of the mesh, since the two match. If that isn't the case, something is wrong with the tutorial. It might be addressed in https://github.com/godotengine/godot-docs/pull/8553? I remember these tutorials do need an update, but it's been a little while since I looked and I don't remember if this is one of the current problems with them.

The-oldone:
> It will have the position of the modified (moved) vertices, but the normals of the flat plane. I think that's what you mean?
Indeed, that's what I wanted to confirm

> I believe the intention is that the two textures should be using the same settings and seed, with the only difference one is a normal map. If that is the case, then we can use the normal map version of the noise texture as the normals of the mesh, since the two match. If that isn't the case, something is wrong with the tutorial. 
Very well, if using FastNoiseLite without changing anything produces the exact same result i.e. operates with the exact same seed, the tutorial is good.

I'm not entirely sure the normal map is used to adjust the lighting normals, since using a normal map to adjust the vertices in the shader gives such a drastically different result, but anyways, if you guys feel this is redundant, feel free to close this.

PR #10502, opened 2025-01-13
Fix `Switch` node boolean conversion description in Visual shaders

The-oldone:
I believe a "true boolean vector" is a vector which evaluates to true. This occurs if any element of the vector is non-zero

### DIFF
--- a/tutorials/shaders/visual_shaders.rst
+++ b/tutorials/shaders/visual_shaders.rst
@@ -224,7 +224,7 @@ Switch node
 ~~~~~~~~~~~
 
 The ``Switch`` node returns a vector if the boolean condition is ``true`` or
-``false``. ``Boolean`` was introduced above. If you convert a vector to a true
-boolean, all components of the vector should be above zero.
+``false``. ``Boolean`` was introduced above. If you want to convert a vector
+to a true boolean, all components of the vector should be non-zero.
 
 .. image:: img/vs_switch.webp


tutorials/shaders/visual_shaders.rst
```diff
@@ -225,6 +225,6 @@ Switch node
 
 The ``Switch`` node returns a vector if the boolean condition is ``true`` or
 ``false``. ``Boolean`` was introduced above. If you convert a vector to a true
-boolean, all components of the vector should be above zero.
+boolean, atleast one component of the vector should be non-zero.
```
tetrapod00, line 228:
```suggestion
boolean, all components of the vector should be non-zero.
```
I don't think the behavior difference you have suggested is quite correct.
Cursory testing in the visual shader editor suggests that it is all components, and that it is non-zero values. Cyan is true, magenta is false:
Godot_v4 3-stable_mono_win64_V6cBtseiaQ

And it compiles to Godot's text shading language as something like the following, with an explicit `all()` in there:
```
// Switch:2
	n_out2p0 = mix(vec3(n_out11p0.xyz), vec3(n_out10p0.xyz), float(all(bvec2(n_out3p0))));
```

The-oldone: Is it?
I shall double check later today and close this if I'm wrong

tetrapod00 (approved):
Looks good to me. The description now matches the observed behavior.

For other reviewers, this is what I used to verify the behavior, cyan is true magenta is false:
Godot_v4 3-stable_mono_win64_V6cBtseiaQ

paddy-exe:
> @tetrapod00 what does pending review from godotengine/shaders mean? It doesn't seem like an actual user

It means a member of the shaders maintainer team has to review it. You can take a look [here for the teams](https://godotengine.org/teams/)

paddy-exe (approved):
LGTM

The-oldone:
@paddy-exe thanks, it took me a while but I got it working. You can merge it now

mhilbrunner (approved):

mhilbrunner:
Thank you! Merged. And thanks everyone for the reviews :)

PR #10505, opened 2025-01-13
Update language on gui_theme_type_variations.rst

lackhand:
I was very confused about how to use themes, and felt the documentation didn't resolve my confusion.

1. Tighten up the language 28-32 referring to extending types (if you're not familiar, its use of the specialized word "type" ("theme type") could refer to a "type system", ie, class extension for a new kind of Node rather than type variation). By ensuring "type" is not used raw but "type variation" or "theme type", we discourage that use.
2. Add an explanatory paragraph 42-43 to creating a type variation that explains every step is necessary and summarizes what they are. Without this it is unclear how critical the para 54+ is  -- count the number of "cans" and "shoulds", which makes it sound very optional!

Thank you!

### DIFF
--- a/tutorials/ui/gui_theme_type_variations.rst
+++ b/tutorials/ui/gui_theme_type_variations.rst
@@ -22,8 +22,8 @@ of being self-sufficient and standalone they extend another, base type.
 
 Following the previous example, your theme can have some styles, colors, and fonts
 defined for the ``Button`` type, customizing the looks of every button element in your UI.
-To then have a gray, red, or blue button you would create a new type, e.g. ``GrayButton``, and
-mark it as a variation of the base ``Button`` type.
+To then have a gray, red, or blue button you would create a new theme type, 
+e.g. ``GrayButton``, and mark it as a variation of the ``Button`` base type.
 
 Type variations can replace some aspects of the base type, but keep others.
 They can also define properties that the base style hasn't defined. For example,
@@ -38,6 +38,9 @@ a combination of both types giving priority to the type variation.
 
 Creating a type variation
 -------------------------
+
+Type variations are created from the theme editor and require both a name and
+a base type from which to extend.
 
 To create a type variation open the theme editor, then click the plus icon
 next to the **Type** dropdown on the right side of the editor. Type in what
@@ -67,7 +70,7 @@ you can find the **Theme Type Variation** property. It is empty by default,
 which means that only the base type has an effect on this node.
 
 You can either select a type variation from a dropdown list, or input its name
-manually. Variations appear on the list only if the type variation belongs to
+as free text. Variations appear on the list only if the type variation belongs to
 the project-wide theme, which you can configure in the project settings. For
 any other case you have to input the name of the variation manually. Click on
 the pencil icon to the right. Then type in the name of the type variation and click the


tutorials/ui/gui_theme_type_variations.rst
```diff
@@ -39,6 +39,9 @@ a combination of both types giving priority to the type variation.
 Creating a type variation
 -------------------------
 
+Type variations are created from the theme editor and require both a name and
+a base type from which to extend.
```
skyace65, line 43: Adding this seems redundant to me since the rest of the section walks you through using the theme editor to do this, and choosing a base type.

tutorials/ui/gui_theme_type_variations.rst
```diff
@@ -67,7 +70,7 @@ you can find the **Theme Type Variation** property. It is empty by default,
 which means that only the base type has an effect on this node.
 
 You can either select a type variation from a dropdown list, or input its name
-manually. Variations appear on the list only if the type variation belongs to
+as free text. Variations appear on the list only if the type variation belongs to
```
skyace65, line 73: I'm not sure what confusion the existing phrasing was causing? This might be because I'm more artist than programmer but I don't see how "manually" could be misinterpreted, and I've never heard the term "free text" used before.

skyace65 (changes requested):

tutorials/ui/gui_theme_type_variations.rst
lackhand: Agreed that this is all just quibbling over wording, so if we all don't think it's better, that's fine, I can drop the PR or drop this section.
[Free text or freeform text seem pretty common ways to discuss this kind of field IMO](https://www.google.com/search?q=form+free+text+field+site:stackoverflow.com).
My concern is that selecting an option from a list is also literally manual input in that a human is doing it, versus the option being selected by the system automatically.

And the system _does_ select options automatically for control nodes!
While debugging wtf was going wrong, I tried to define a gdscript that extended a control node* (`class_name MyLabel extends Label`), hoping that what this was referring to was the mapping of the first default nameless option to the defined variation (versus manual selection of its full name). OFC that also failed, but this section is why I even tried it 😁 

---
 * I now understand why this isn't how it works: they're not _really_ MyLabel nodes, they're Label nodes with an attached MyLabel script. The theme lookup system doesn't pay attention to that. But IMO that's pretty subtle so worth documentation about correct usage!

tutorials/ui/gui_theme_type_variations.rst
lackhand: I'm very open to a different fix, but I believe some kind of fix is warranted.
My PR description explains my objection to the rest of the section: it presents itself as optional and skippable if you are not already aware of exactly how the feature works.

Much much much more description below.

The rest of the paragraph includes the following snippets, emphasis mine:

1. "You _can_ select the base type there"
2. "_typically_ the name of a control node class"
3. "you _should_ now be able to see its properties ..."
4. "you _can_ edit them as normal"

Note that one might come away from this thinking that they don't need to specify a base type (the use of "can" in points 1, where it means required, and 4, where it means optional).

I suppose another fix would be to tighten up the language:
1. "You _must_ select a base type there"
2. "typically the name of a _built-in_ control node class" (_edit: Is this accurate language? I don't know the right terms. Merely extending a node in gdscript isn't enough afaict to work with this feature._)
3. I'm not sure how to fix this one! I notice that if you chain custom types -- `Foo extends Label` and `Bar extends Foo` -- `Bar` doesn't show any properties. It might be cosmetic, I'm not sure. Together with point #2, how the user should proceed is pretty unclear IMO. So: "You should now be able to see properties _if you inherited from a control node, which is recommended for beginners_"? Still messy, maybe both sentences get a rewrite.
4. "you can edit them as normal" (_no notes, this is fine 😁!_)

Given all of that, I _didn't_ try to fix the language, and instead just added a guiding "goal" statement; in jokier language "you're here to use the UI to add a theme and set its base type, without which it might not work. There are other reasons it might not work too, which are buried in the details. Good luck."

If this is unpersuasive, perhaps there's a better fix here? But the feature is complex enough that being clear about the absolute requirements seems good to me.

lackhand:
Thank you very much for your time and attention! 
FWIW I respond very well to something along the lines of "skyace has bigger fish to fry than this so lackhand should wait until it's worth fixing this" -- I'm fine leaving this branch open for _decades_ until the glorious future where this is our biggest problem 😁

tutorials/ui/gui_theme_type_variations.rst
tetrapod00: I don't have specific feelings about the other change, but I think "manually" is already pretty clear, and IMO more clear than "as free text".

skyace65:
Closing this PR due to a lack of support from other maintainers.

PR #10514, opened 2025-01-17
Document adding scene-unique nodes by prefixing with '%'

Synzorasize:
This feature was added in godotengine/godot#101163. Therefore it should probably be documented here so that users don't miss it. Feel free to request changes.

### DIFF
--- a/tutorials/scripting/scene_unique_nodes.rst
+++ b/tutorials/scripting/scene_unique_nodes.rst
@@ -18,6 +18,8 @@ the node's path is changed.
 Creation and usage
 ------------------
 
+There are two ways to create a scene unique node.
+
 In the Scene tree dock, right-click on a node and select
 **Access as Unique Name** in the context menu.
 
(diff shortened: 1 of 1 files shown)


tutorials/scripting/scene_unique_nodes.rst
```diff
@@ -18,6 +18,8 @@ the node's path is changed.
 Creation and usage
 ------------------
 
+There are two ways to do this.
```
tetrapod00, line 21:
```suggestion
There are two ways to create a scene unique node.
```

tutorials/scripting/scene_unique_nodes.rst
```diff
@@ -28,6 +30,9 @@ to its name in the scene tree:
 
 .. image:: img/percent.webp
 
+You can also do this while renaming the node and adding % to the beginning. Once you
+confirm, the percent symbol will appear next to its name.
```
tetrapod00, line 34:
```suggestion
You can also do this while renaming the node by adding "%" to the beginning of the name.
Once you confirm, the percent symbol will appear next to its name.
```

tetrapod00 (approved):
Okay to merge as-is, but better with the changes suggested here.

As a side note, in a later PR we might want to double check all the places we refer to "scene unique nodes" and make sure they are aligned in terminology. As best I can tell, in the Godot editor UI itself, the phrase "scene unique node" is not used, but "Access as Unique Name" is. The tooltip also doesn't mention "scene unique node" anywhere:
Godot_v4 4-beta1_win64_zynAMYiYEW

So it's an interesting situation where the docs call it one thing but the editor doesn't seem to have a name for it.

tetrapod00 (approved):

mhilbrunner (approved):

mhilbrunner:

PR #10520, opened 2025-01-18
Update animation tracks page for Godot 4.4

skyace65:
Technically for 4.4 since two of the images needed to be updated for the new FPS option. Other ones needed to be updated for reasons that have probably been in a few prior releases. All images have gone through squoosh compression.

- Updated two images for the snap to FPS option in the animation editor
- Replaced one because the right click menu got the cut and copy option
- Replaced one image for the new audio stream player icon from Godot 4.3
- Also added clarification to the right click menu instructions. When I was going through the page I thought I could just right click anywhere or right click with a handle selected, but the key specifically has to be selected


tutorials/animation/animation_track_types.rst
```diff
@@ -157,7 +157,8 @@ diamonds connected to them by a line control curve's shape.
 
 .. image:: img/bezier_curves.webp
 
-In the right click panel of the editor, you can select the handle mode:
+While a keyframe is selected (Not the handle), in the right click panel of the
```
mhilbrunner, line 160:
```suggestion
While a keyframe is selected (not the handle), in the right click panel of the
```

mhilbrunner (approved):

skyace65:
Fixed

mhilbrunner:
Thanks!

PR #10523, opened 2025-01-19
Update first look at Godot's interface for Godot 4.4

skyace65:
Between the new "game" workspace at the top, toolbar changes for the 2D and 3D workspaces, and a few minor tweaks, almost every image on this page needed to be updated. I also added a small section for the new "Game" workspace, and updated a sentence to reference the movie maker mode toggle which is from an earlier release.


mhilbrunner (approved):

mhilbrunner:
Thank you!

PR #10526, opened 2025-01-20
Document Jolt built-in module in 4.4

skyace65:
This whole PR is largely the information from the PR that added Jolt (https://github.com/godotengine/godot/pull/99895) copy and pasted, and then heavily edited for proper formatting, adding links, rephrasing and cutting out content that only makes sense for the PR ("maybe we should do this in the future" stuff).

There are not links for the project settings listed at the bottom for name changes as that didn't seem necessary, but I'll add them if people think they should be there.

I've looked over the Jolt PRs merged since this initial PR was merged and I believe the information here is accurate.

@mihe let me know if there's anything at all you wanted added, changed or removed

### DIFF
--- a/tutorials/physics/using_jolt_physics.rst
+++ b/tutorials/physics/using_jolt_physics.rst
@@ -1,0 +1,278 @@
+.. _doc_using_jolt_physics:
+
+Using Jolt Physics
+==================
+
+Introduction
+------------
+
+The Jolt physics engine was added as an alternative to the existing Godot Physics
+physics engine in 4.4. Jolt is developed by Jorrit Rouwe with a focus on games and
+VR applications. Previously it was available as a extension but is now built into
+Godot.
+
+It is important to note that the built-in Jolt Physics module is considered
+**not finished**, **experimental**, and **lacks feature parity** with both
+Godot Physics and the Godot Jolt extension. Behavior may change as it is developed
+further. Please keep that in mind when choosing what to use for your project.
+
+The existing extension is now considered in maintenance mode. That means bug fixes
+will be merged, and it will be kept compatible with new versions of Godot until
+the built-in module has feature parity with the extension. The extension can be
+found `here on GitHub <https://github.com/godot-jolt/godot-jolt>`_ and in Godot's asset
+library.
+
+To change the 3D physics engine to be Jolt Physics, set
+:ref:`Project Settings > Physics > 3D > Physics Engine<class_ProjectSettings_property_physics/3D/Physics_Engine>`
+to ``Jolt Physics``. Once you've done that, click the **Save & Restart** button.
+When the editor opens again, 3D scenes should now be using Jolt for physics.
+
+Notable differences to Godot Physics
+------------------------------------
+
+There are many differences between the existing Godot Physics engine and Jolt.
+
+Area3D and static bodies
+~~~~~~~~~~~~~~~~~~~~~~~~
+When using Jolt, :ref:`class_Area3D` will not detect overlaps with :ref:`class_StaticBody3D`
+(nor a :ref:`class_RigidBody3D` frozen with ``FREEZE_MODE_STATIC``) by default, for
+performance reasons. If you have many/large :ref:`class_Area3D` overlapping with
+complex static geometry, such as :ref:`class_ConcavePolygonShape3D` or
+:ref:`class_HeightMapShape3D`, you can end up wasting a significant amount of CPU
+performance and memory without realizing it.
+
+For this reason this behavior is opt-in through the project setting
+:ref:`Physics > Jolt Physics 3D > Simulation > Areas Detect Static Bodies<class_ProjectSettings_property_physics/jolt_physics_3d/simulation/areas_detect_static_bodies>`,
+with the recommendation that you set up your collision layers and masks in such a
+way that only a few small :ref:`class_Area3D` are able to detect collisions with
+static bodies.
+
+Joint properties
+~~~~~~~~~~~~~~~~
+
+The current interfaces for the 3D joint nodes don't quite line up with the interface
+of Jolt's own joints. As such, there are a number of joint properties that are not
+supported, mainly ones related to configuring the joint's soft limits.
+
+The unsupported properties are:
+
+- PinJoint3D: ``bias``, ``damping``, ``impulse_clamp``
+- HingeJoint3D: ``bias``, ``softness``, ``relaxation``
+- SliderJoint3D: ``angular_\*``, ``\*_limit/softness``, ``\*_limit/restitution``, ``\*_limit/damping``
+- ConeTwistJoint3D: ``bias``, ``relaxation``, ``softness``
+- Generic6DOFJoint3D: ``*_limit_*/softness``, ``*_limit_*/restitution``, ``*_limit_*/damping``, ``*_limit_*/erp``
+
+Currently a warning is emitted if you set these properties to anything but their
+default values.
+
+Single-body joints
+~~~~~~~~~~~~~~~~~~
+
+You can, in Godot, omit one of the joint bodies for a two-body joint and effectively
+have "the world" be the other body. However, the node path that you assign your body
+to (:ref:`node_a<class_Joint3D_property_node_a>` vs :ref:`node_b<class_Joint3D_property_node_b>`)
+is ignored. Godot Physics will always behave as if you
+assigned it to ``node_a``, and since ``node_a`` is also what defines the frame of reference
+for the joint limits, you end up with inverted limits and a potentially strange
+limit shape, especially if your limits allow both linear and angular degrees of
+freedom.
+
+Jolt will behave as if you assigned the body to ``node_b`` instead, with ``node_a``
+representing "the world". There is a project setting called :ref:`Physics > Jolt Physics 3D > Joints > World Node<class_ProjectSettings_property_physics/jolt_physics_3d/joints/world_node>`
+that lets you toggle this behavior, if you need compatibility for an existing project.
+
+Collision margins
+~~~~~~~~~~~~~~~~~
+
+Jolt (and other similar physics engines) uses something that Jolt refers to as
+"convex radius" to help improve the performance and behavior of the types of
+collision detection that Jolt relies on for convex shapes. Other physics engines
+(Godot included) might refer to these as "collision margins" instead. Godot exposes
+these as the ``margin`` property on every Shape3D-derived class, but Godot Physics
+itself does not use them for anything.
+
+What these collision margins sometimes do in other engines (as described in Godot's
+documentation) is effectively add a "shell" around the shape, slightly increasing
+its size while also rounding off any edges/corners. In Jolt however, these margins
+are first used to shrink the shape, and then the "shell" is applied, resulting in
+edges/corners being similarly rounded off, but without increasing the size of the
+shape.
+
+To prevent having to tweak this margin property manually, since its default value
+can be problematic for smaller shapes, the Jolt module exposes a project setting
+called :ref:`Physics > Jolt Physics 3D > Collisions > Collision Margin Fraction<class_ProjectSettings_property_physics/jolt_physics_3d/collisions/collision_margin_fraction>`
+which is multiplied with the smallest axis of the shape's AABB to calculate the
+actual margin. The margin property of the shape is then instead used as an upper
+bound.
+
+These margins should, for most use-cases, be more or less transparent, but can
+sometimes result in odd collision normals when performing shape queries. You can
+lower the above mentioned project setting to mitigate some of this, including
+setting it to ``0.0``, but too small of a margin can also cause odd collision results,
+so is generally not recommended.
+
+Baumgarte stabilization
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Baumgarte stabilization is a method to resolve penetrating bodies and push them to a
+state where they are just touching. In Godot Physics this works like a spring. This
+means that bodies can accelerate and may cause the bodies to overshoot and separate
+completely. With Jolt, the stabilization is only applied to the position and not to
+the velocity of the body. This means it cannot overshoot but it may take longer to
+resolve the penetration.
+
+The strength of this stabilization can be tweaked using the project setting
+:ref:`Physics > Jolt Physics 3D > Simulation > Baumgarte Stabilization Factor<class_ProjectSettings_property_physics/jolt_physics_3d/simulation/baumgarte_stabilization_factor>`.
+Setting this project setting to ``0.0`` will turn Baumgarte stabilization off.
+Setting it to ``1.0`` will resolve penetration in 1 simulation step. This is fast
+but often also unstable.
+
+Ghost collisions
+~~~~~~~~~~~~~~~~
+
+Jolt employs two techniques to mitigate ghost collisions, meaning collisions with
+internal edges of shapes/bodies that result in collision normals that oppose the
+direction of movement.
+
+The first technique, called "active edge detection", marks edges of triangles in
+:ref:`class_ConcavePolygonShape3D` or :ref:`class_HeightMapShape3D` as either "active" or "inactive", based on
+the angle to the neighboring triangle. When a collision happens with an inactive
+edge the collision normal will be replaced with the triangle's normal instead, to
+lessen the effect of ghost collisions.
+
+The angle threshold for this active edge detection is configurable through the
+project setting :ref:`Physics >Jolt Physics 3D > Collisions > Active Edge Threshold<class_ProjectSettings_property_physics/jolt_physics_3d/collisions/active_edge_threshold>`.
+
+The second technique, called "enhanced internal edge removal", instead adds runtime
+checks to detect whether an edge is active or inactive, based on the contact points
+of the two bodies. This has the benefit of applying not only to collisions with
+:ref:`class_ConcavePolygonShape3D` and :ref:`class_HeightMapShape3D`, but also edges between any shapes within
+the same body.
+
+Enhanced internal edge removal can be toggled on and off for the various contexts to
+which it's applied, using the :ref:`Physics >Jolt Physics 3D > Simulation > Use Enhanced Internal Edge Removal<class_ProjectSettings_property_physics/jolt_physics_3d/simulation/use_enhanced_internal_edge_removal>`,
+project setting, and the similar settings for :ref:`queries<class_ProjectSettings_property_physics/jolt_physics_3d/queries/use_enhanced_internal_edge_removal>`
+and :ref:`motion queries<class_ProjectSettings_property_physics/jolt_physics_3d/motion_queries/use_enhanced_internal_edge_removal>`.
+
+Note that neither the active edge detection nor enhanced internal edge removal apply
+when dealing with ghost collisions between two different bodies.
+
+Memory usage
+~~~~~~~~~~~~
+
+Jolt uses a stack allocator for temporary allocations within its simulation step.
+This stack allocator requires allocating a set amount of memory up front, which can
+be configured using the :ref:`Physics > Jolt Physics 3D > Limits > Temporary Memory Buffer Size<class_ProjectSettings_property_physics/jolt_physics_3d/limits/temporary_memory_buffer_size>`
+project setting.
+
+Ray-cast face index
+~~~~~~~~~~~~~~~~~~~
+
+The ``face_index`` property returned in the results of :ref:`intersect_ray()<class_PhysicsDirectSpaceState3D_method_intersect_ray>`
+and RayCast3D will by default always be ``-1`` with Jolt. The project setting :ref:`Physics > Jolt Physics 3D > Queries > Enable Ray Cast Face Index<class_ProjectSettings_property_physics/jolt_physics_3d/queries/enable_ray_cast_face_index>`
+will enable them.
+
+Note that enabling this setting will increase the memory requirement of :ref:`class_ConcavePolygonShape3D`
+with about 25%.
+
+Kinematic RigidBody3D contacts
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When using Jolt, a :ref:`class_RigidBody3D` frozen with :ref:`FREEZE_MODE_KINEMATIC<class_RigidBody3D_constant_FREEZE_MODE_KINEMATIC>`
+will by default not report contacts from collisions with other static/kinematic
+bodies, for performance reasons, even when setting a non-zero :ref:`max_contacts_reported<class_RigidBody3D_property_max_contacts_reported>`.
+If you have many/large kinematic bodies overlapping with complex static geometry,
+such as :ref:`class_ConcavePolygonShape3D` or :ref:`class_HeightMapShape3D`, you can
+end up wasting a significant amount of CPU performance and memory without realizing
+it.
+
+For this reason this behavior is opt-in through the project setting
+:ref:`Physics > Jolt Physics 3D > Simulation > Generate All Kinematic Contacts<class_ProjectSettings_property_physics/jolt_physics_3d/simulation/generate_all_kinematic_contacts>`.
+
+Contact impulses
+~~~~~~~~~~~~~~~~
+
+Due to limitations internal to Jolt, the contact impulses provided by :ref:`PhysicsDirectBodyState3D.get_contact_impulse()<class_physicsdirectbodystate3d_method_get_contact_impulse>`
+are estimated ahead of time based on things like the contact manifold and velocities
+of the colliding bodies. This means that the reported impulses will only be accurate
+in cases where the two bodies in question are not colliding with any other bodies.
+
+Area3D and SoftBody3D
+~~~~~~~~~~~~~~~~~~~~~
+
+Jolt does not currently support any interactions between :ref:`class_SoftBody3D`
+and :ref:`class_Area3D`, such as overlap events, or the wind properties found on
+:ref:`class_Area3D`.
+
+WorldBoundaryShape3D
+~~~~~~~~~~~~~~~~~~~~
+
+:ref:`class_WorldBoundaryShape3D`, which is meant to represent an infinite plane, is
+implemented a bit differently in Jolt compared to Godot Physics. Both engines have
+an upper limit for how big the effective size of this plane can be, but this size is
+much smaller when using Jolt, in order to avoid precision issues.
+
+You can configure this size using the :ref:`Physics > Jolt Physics 3D > Limits > World Boundary Shape Size<class_ProjectSettings_Property_physics/jolt_physics_3d/limits/world_boundary_shape_size>`
+project setting.
+
+Notable differences to the Godot Jolt extension
+-----------------------------------------------
+
+While the built-in Jolt module is largely a straight port of the Godot Jolt
+extension, there are a few things that are different.
+
+Project settings
+~~~~~~~~~~~~~~~~
+
+All project settings have been moved from the ``physics/jolt_3d`` category to
+``physics/jolt_physics_3d``.
+
+On top of that, there's been some renaming and refactoring of the individual project
+settings as well. These include:
+
+- ``sleep/enabled`` is now ``simulation/allow_sleep.``
+- ``sleep/velocity_threshold`` is now ``simulation/sleep_velocity_threshold.``
+- ``sleep/time_threshold`` is now ``simulation/sleep_time_threshold.``
+- ``collisions/use_shape_margins`` is now ``collisions/collision_margin_fraction``,
+  where a value of 0 is equivalent to disabling it.
+- ``collisions/use_enhanced_internal_edge_removal`` is now ``simulation/use_enhanced_internal_edge_removal.``
+- ``collisions/areas_detect_static_bodies`` is now ``simulation/areas_detect_static_bodies.``
+- ``collisions/report_all_kinematic_contacts`` is now ``simulation/generate_all_kinematic_contacts.``
+- ``collisions/soft_body_point_margin`` is now ``simulation/soft_body_point_radius.``
+- ``collisions/body_pair_cache_enabled is now simulation/body_pair_contact_cache_enabled.``
+- ``collisions/body_pair_cache_distance_threshold`` is ``now simulation/body_pair_contact_cache_distance_threshold.``
+- ``collisions/body_pair_cache_angle_threshold is now simulation/body_pair_contact_cache_angle_threshold.``
+- ``continuous_cd/movement_threshold`` is now ``simulation/continuous_cd_movement_threshold``,
+  but expressed as a fraction instead of a percentage.
+- ``continuous_cd/max_penetration`` is now ``simulation/continuous_cd_max_penetration``,
+  but expressed as a fraction instead of a percentage.
+- ``kinematics/use_enhanced_internal_edge_removal`` is now ``motion_queries/use_enhanced_internal_edge_removal.``
+- ``kinematics/recovery_iterations`` is now ``motion_queries/recovery_iterations``,
+  but expressed as a fraction instead of a percentage.
+- ``kinematics/recovery_amount`` is now ``motion_queries/recovery_amount.``
+- ``queries/use_legacy_ray_casting`` has been removed.
+- ``solver/position_iterations`` is now ``simulation/position_steps.``
+- ``solver/velocity_iterations`` is now ``simulation/velocity_steps.``
+- ``solver/position_correction`` is now ``simulation/baumgarte_stabilization_factor``,
+  but expressed as a fraction instead of a percentage.
+- ``solver/active_edge_threshold`` is now ``collisions/active_edge_threshold.``
+- ``solver/bounce_velocity_threshold`` is now ``simulation/bounce_velocity_threshold.``
+- ``solver/contact_speculative_distance`` is now ``simulation/speculative_contact_distance.``
+- ``solver/contact_allowed_penetration`` is now ``simulation/penetration_slop.``
+- ``limits/max_angular_velocity`` is now stored as radians instead.
+- ``limits/max_temporary_memory`` is now ``limits/temporary_memory_buffer_size.``
+
+Joint nodes
+~~~~~~~~~~~
+
+The joint nodes that are exposed in the Godot Jolt extension (JoltPinJoint3D,
+JoltHingeJoint3D, JoltSliderJoint3D, JoltConeTwistJoint3D, and JoltGeneric6DOFJoint)
+have not been included in the Jolt module.
+
+Thread safety
+~~~~~~~~~~~~~
+
+Unlike the Godot Jolt extension, the Jolt module does have thread-safety,
+including support for the :ref:`Physics > 3D > Run On Separate Thread<class_ProjectSettings_Property_physics/3d/run_on_separate_thread>`
+project setting. However this has not been tested very thoroughly, so it should be
+considered experimental.
--- a/tutorials/physics/index.rst
+++ b/tutorials/physics/index.rst
@@ -10,6 +10,7 @@ Physics
    :name: toc-learn-features-physics
 
    physics_introduction
+   using_jolt_physics
    rigid_body
    using_area_2d
    using_character_body_2d


tetrapod00:
For my part, this seems better than nothing. If we can guarantee that new user-facing documentation will be written between now and 4.5stable releasing (which is presumably the earliest that Jolt will be marked not experimental), it's fine to wait for that documentation rather than merging this PR. But if it's unclear that @mihe would have time for that, either, I'm mildly in favor of including some form of this current page with the knowledge that it can be streamlined or removed later.

skyace65:
I do not want 4.4 releasing without some kind of Jolt documentation because I can see that leading to a lot of confusion. If you think too much of this shouldn't be user facing @mihe then just tell me what chunks of it to remove and I'll do that.

Alternatively, If you could give me bullet point lists of what's missing in terms of feature parity compared to the extension and GodotPhysics3D I'd be fine with adding those and removing everything else except for the overview section I wrote, the renamed project settings subsection, and the subsection on thread safety.

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,504 @@
...
+The Godot Jolt extension has the ability to generate what Jolt refers to as
+"snapshots", where it serializes the state of the physics simulation to a file.
+These can then be loaded in Jolt's own Samples application, to debug issues more
+closely there, without needing to deal with Godot itself, which has proved to be
+quite useful when reporting issues upstream to Jolt.
+
+The code for this is technically included with this module, and can be found as
+``JoltPhysicsServer3D::dump_debug_snapshots``, but it is currently unexposed for now.
+
+Roadmap
```
mhilbrunner, line 475: I would strike this whole section, it feels out of place with other docs - the docs are generally not a place to talk about future plans, and we don't use it for roadmaps for other teams AFAIK.

The whole future being experimental is fair, but IMO some other place (a blog post? the original PR?) are a better time capsule to capture the current state of what is in, what is going to be in, and what won't be in than the docs.

skyace65:
I've removed the roadmap section

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,473 @@
...
+that wraps StaticCompoundShape, but which will always defer its building/committing
+only until absolutely necessary, meaning either right before a simulation step or
+when performing queries against the body.
+
+Scaling shapes/bodies/queries
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Godot Physics supports scaling the transform of collision shapes, shape queries, as
+well as static and kinematic bodies, meaning :ref:`class_StaticBody3D`,
+:ref:`class_CharacterBody3D`, :ref:`class_AnimatableBody3D` and frozen :ref:`class_RigidBody3D`.
```
AThousandShips, line 148:
```suggestion
:ref:`class_CharacterBody3D`, :ref:`class_AnimatableBody3D`, and frozen :ref:`class_RigidBody3D`.
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,473 @@
...
+consequently this module) the number of steps is dynamically calculated  based on
+the cast distance, aiming for roughly millimeter precision, and then clamped between
+4 and 16 steps.
+
+This does however mean that :ref:`cast_motion<class_PhysicsDirectSpaceState3D_method_cast_motion>`
+will technically, when using Jolt, cost more CPU performance the farther you cast
+the shape.
+
+Note that this also applies to :ref:`body_test_motion<class_PhysicsServer3D_method_body_test_motion>`, and consequently :ref:`test_move<class_PhysicsBody3D_method_test_move>`,
+:ref:`move_and_collide<class_PhysicsBody3D_method_move_and_collide>` and :ref:`move_and_slide<class_CharacterBody3D_method_move_and_slide>`.
```
AThousandShips, line 187:
```suggestion
:ref:`move_and_collide<class_PhysicsBody3D_method_move_and_collide>`, and :ref:`move_and_slide<class_CharacterBody3D_method_move_and_slide>`.
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,473 @@
...
+Setting this project setting to ``1.0`` will effectively disable the technique,
+resulting in a simulation that more closely resembles Godot Physics, but which is
+also more unstable.
+
+Motion queries (move_and_slide, etc.)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Physics servers in Godot are meant to implement the :ref:`PhysicsServer3D.body_test_motion()<class_PhysicsServer3D_method_body_test_motion>`
+method, which in turn powers methods like :ref:`PhysicsBody3D.test_move()<class_PhysicsBody3D_method_test_move>`,
+:ref:`PhysicsBody3D.move_and_collide()<class_PhysicsBody3D_method_move_and_collide>`
```
AThousandShips, line 208:
```suggestion
:ref:`PhysicsBody3D.move_and_collide()<class_PhysicsBody3D_method_move_and_collide>`,
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,473 @@
...
+
+Jolt uses a stack allocator for temporary allocations within its simulation step.
+This stack allocator requires allocating a set amount of memory up front, which can
+be configured using the :ref:`physics/jolt_physics_3d/limits/temporary_memory_buffer_size<class_ProjectSettings_property_physics/jolt_physics_3d/limits/temporary_memory_buffer_size>`
+project setting.
+
+Jolt also makes use of aligned allocations for some of its memory usage, which it
+acquires through function pointers that the implementer assigns. Aligned allocations
+were recently added to Godot in the form of ``Memory::alloc_aligned_static``,
+``Memory::realloc_aligned_static`` and ``Memory::free_aligned_static``, which have all been
```
AThousandShips, line 283:
```suggestion
``Memory::realloc_aligned_static``, and ``Memory::free_aligned_static``, which have all been
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,473 @@
...
+
+The Godot Jolt extension utilizes a custom container (typically referred to as an
+"inline vector" or "small vector") in order to avoid heap allocations for physics
+queries that return more than a single hit, meaning :ref:`intersect_point<class_PhysicsDirectSpaceState3D_method_intersect_point>`,
+:ref:`intersect_shape<class_PhysicsDirectSpaceState3D_method_intersect_shape>`,
+:ref:`collide_shape<class_PhysicsDirectSpaceState3D_method_collide_shape>`,
+:ref:`get_rest_info<class_PhysicsDirectSpaceState3D_method_get_rest_info>`,
+:ref:`cast_motion<class_PhysicsDirectSpaceState3D_method_cast_motion>`,
+:ref:`body_test_motion<class_PhysicsServer3D_method_body_test_motion>`,
+:ref:`test_move<class_PhysicsBody3D_method_test_move>`, :ref:`move_and_collide<class_PhysicsBody3D_method_move_and_collide>`
```
AThousandShips, line 445:
```suggestion
:ref:`test_move<class_PhysicsBody3D_method_test_move>`, :ref:`move_and_collide<class_PhysicsBody3D_method_move_and_collide>`,
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,473 @@
...
+- ``limits/max_temporary_memory`` is now ``limits/temporary_memory_buffer_size.``
+
+There might be some discussion to be had with regards to migrating the settings
+values for projects who have previously been relying on the extension.
+
+Joint Nodes
+~~~~~~~~~~~
+
+The joint nodes that are exposed in the Godot Jolt extension (JoltPinJoint3D,
+JoltHingeJoint3D, JoltSliderJoint3D, JoltConeTwistJoint3D and JoltGeneric6DOFJoint)
```
AThousandShips, line 418:
```suggestion
JoltHingeJoint3D, JoltSliderJoint3D, JoltConeTwistJoint3D, and JoltGeneric6DOFJoint)
```

mihe:
No, sorry, not yet. It's been in the back of my mind, but time has been scarce. I'll try and make some time for it this weekend.

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,473 @@
...
+and :ref:`CharacterBody3D.move_and_slide()<class_CharacterBody3D_method_move_and_slide>`,
+which are largely meant to be used for moving player characters around.
+
+:ref:`PhysicsServer3D.body_test_motion<class_PhysicsServer3D_method_body_test_motion>`
+is split into three parts, the first being depenetration (called "recovery" in
+Godot), the second being a shape-cast from the "recovered" position, and the third
+being the actual collision check, at the "unsafe" fraction of the shape-cast.
+
+The "recovery" step in Godot Physics is hardcoded to always do 4 iterations with 40%
+depenetration per iteration. I figured these constants might be useful to expose, so
```
jrouwe, line 218: 'I figured' is probably not what you want in the documentation.

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,473 @@
...
+
+While these discrepancies do seem to result in less ghost collisions when using
+Jolt, they also introduce new problems:
+
+- :ref:`move_and_slide<class_CharacterBody3D_method_move_and_slide>` is generally
+  slower than in Godot Physics, due to :ref:`class_CharacterBody3D`::_snap_on_floor(PR NOTE: Apply_floor_snap?)
+  triggering on every call.
+- :ref:`class_CharacterBody3D` will not report wall collisions unless you're moving towards them.
+
+It's not clear to me how to fix these issues, but I suspect they will require
```
jrouwe, line 244: Do we want to use 'I' in documentation?

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,473 @@
...
+lessen the effect of ghost collisions.
+
+The angle threshold for this active edge detection is configurable through the
+project setting :ref:`physics/jolt_physics_3d/collisions/active_edge_threshold<class_ProjectSettings_property_physics/jolt_physics_3d/collisions/active_edge_threshold>`.
+
+The second technique, called "enhanced internal edge removal", instead adds runtime
+checks to detect whether an edge is active or inactive, based on the contact points
+of the two bodies. This has the benefit of applying not only to collisions with
+:ref:`class_ConcavePolygonShape3D` and :ref:`class_HeightMapShape3D`, but also edges between any shapes within
+the same body.
```
jrouwe, line 266: This is an important point and good to stress: If you want to avoid ghost collisions, make sure to group multiple shapes within the same body. It does not work if an object is sliding over the boundary between two static bodies.

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,473 @@
...
+including support for the :ref:`physics/3d/run_on_separate_thread<class_ProjectSettings_Property_physics/3d/run_on_separate_thread>`
+project setting. This is achieved by utilizing the same wrapper server that's used
+by Godot Physics, called ``PhysicsServer3DWrapMT``, as well as introducing a mutex
+around the rebuilding of shapes, since concurrent shape-casts could otherwise
+trigger a race condition.
+
+This has however not been tested very thoroughly, so should be considered
+experimental.
+
+Query Performance
```
jrouwe, line 434: I think this has been fixed by https://github.com/godotengine/godot/pull/101071 so this section can probably be removed.

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,473 @@
...
+~~~~~~~~~~~~~~~
+
+The Godot Jolt extension has the ability to generate what Jolt refers to as
+"snapshots", where it serializes the state of the physics simulation to a file.
+These can then be loaded in Jolt's own Samples application, to debug issues more
+closely there, without needing to deal with Godot itself, which has proved to be
+quite useful when reporting issues upstream to Jolt.
+
+The code for this is technically included with this module, and can be found as
+``JoltPhysicsServer3D::dump_debug_snapshots``, but it is currently unexposed for now.
```
jrouwe, line 473: remove 'for now' as it says the same thing twice.

mihe:
Thank you for the feedback, @jrouwe.

I was in the middle of typing up some of my own thoughts, but they're a bit more abstract, as I've struggled a bit with distilling this down to something more minimal, like I said I would.

I figured I would at least share my thoughts on the broader topics and see if maybe we could work our way down from there, but it got messier than I'd hoped.

Here they are anyway:

#### Things that probably should be included

- Interactions between `Area3D` and static bodies
  - This seems to have turned into one of the bigger stumbling blocks for people migrating to Jolt, so should be mentioned.
- Interactions between `Area3D` and `SoftBody3D`
  - This is somewhat useful functionality that's missing compared to Godot Physics, so should be mentioned.
- The missing joint properties
  - I struggle to see anyone relying on these in Godot Physics currently, given their current state, but it's technically functionality that's missing, so should be mentioned I guess.
- Collision margins
  - The majority of people will likely never have to care about these, but it's proven to add some friction for certain advanced use-cases, so should probably be mentioned in some form.
- Single-body joints
  - This should rightfully be included in some form, especially considering the project setting, but I find explaining it to be messy to an unfortunate degree.
- Ghost collisions
  - I'm torn on this one. I feel like it gets overly technical, but it's an important topic that a lot of people will likely run into, and knowing how the mitigation techniques work (and their limitations) might be useful.
- Baumgarte stabilization
  - Given @jrouwe's feedback above I guess this needs some rewording. In general I feel like this might be better presented through some kind of FAQ that speaks to the user scenario of bodies depenetrating slowly instead.
- Ray cast face index
  - It's not obvious that this is opt-in with Jolt, so should be mentioned I guess.
- Contact impulses
  - Probably worth mentioning.
- Thread-safety
  - Probably worth mentioning, although I must admit it's unclear when my self-imposed experimental status for this would be lifted exactly.

#### Things I'm not sure should be included

- Scaling shapes/bodies/queries
  - I would probably say this shouldn't be included. You'll still get node configuration warnings when trying to scale things in ways that Godot Physics doesn't support, so putting this out there as a supported feature doesn't seem right.
  - Any invalid scaling will emit somewhat clear errors anyway, both in the editor and at runtime.
- Memory usage
  - It might be worth mentioning that memory used by Jolt isn't currently tracked by Godot, but even that isn't unique to third-party libraries in Godot (as far as I know) so seems excessive.
- ConvexPolygonShape3D
  - I'm not sure this is worth mentioning, given that you likely wouldn't have 1-to-1 behavior between the two engines anyway.
- Axis-locking
  - I'm not sure this is worth mentioning. An error is emitted if you try to lock all axes anyway.
- Joint nodes (missing from the extension)
  - Might be better to incorporate into the general topic of the joint properties/interfaces.

#### Things I don't think should be included

- Compound shapes
  - This was mostly just an implementation detail. I don't think it needs to be in any user-facing documentation.
- Shape-casting
  - More implementation details that don't matter much for the user.
- Motion queries
  - More implementation details that don't matter much for the user.
- Query performance (compared to the extension)
  - As mentioned by @jrouwe already, this has since been resolved, so can be removed.
- Debug renderer (missing from the extension)
  - This refers to a development feature, with no real importance for the user.
- Debug snapshots (missing from the extension)
  - This refers to a development feature, with no real importance for the user.

---

I can turn this into review comments if you'd all prefer. My hope was still that we wouldn't necessarily base the documentation on my PR description to begin with, but I guess that's where this is headed, and I can't really justify a rephrasing for a lot of these topics anyway.

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,473 @@
...
+:ref:`class_CharacterBody3D`, :ref:`class_AnimatableBody3D` and frozen :ref:`class_RigidBody3D`.
+It does not however support scaling simulated/dynamic bodies, such as a non-frozen :ref:`class_RigidBody3D`,
+and will effectively discard any such scaling, instead treating it as (1, 1, 1).
+
+Jolt does however support scaling everywhere, which means that :ref:`class_RigidBody3D`
+will support scaling when using Jolt, despite the warning shown on the node
+currently.
+
+Jolt also supports non-uniform scaling, so long as the inherent primitive shape is
+preserved. For example, you can scale a cylinder along its height axis but not along
```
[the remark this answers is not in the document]
mihe: Yeah, I was referring to scaling of one axis at a time, but that was poorly worded of me.

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,473 @@
...
+its other axes. You will however currently see warnings on the shape node when doing
+this, due to Godot Physics not supporting this.
+
+Since invalid scaling can cause a number of weird artifacts, and sometimes outright
```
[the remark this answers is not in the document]
mihe: Right, but `MakeScaleValid` is part of the "sanitizing" mentioned in that paragraph, although I'll admit I had overlooked that `MakeScaleValid` covered the zero scaling case as well. Without it you could find yourself in trouble, if I remember correctly, although "crash the simulation" might've been me jumping the gun a bit.

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,473 @@
...
+will technically, when using Jolt, cost more CPU performance the farther you cast
+the shape.
+
+Note that this also applies to :ref:`body_test_motion<class_PhysicsServer3D_method_body_test_motion>`, and consequently :ref:`test_move<class_PhysicsBody3D_method_test_move>`,
+:ref:`move_and_collide<class_PhysicsBody3D_method_move_and_collide>` and :ref:`move_and_slide<class_CharacterBody3D_method_move_and_slide>`.
+
+Baumgarte stabilization
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Jolt employs a technique in its solver called Baumgarte stabilization, which is
```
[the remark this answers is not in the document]
mihe: > As far as I can see Godot Physics has it as well

That is news to me! I always wondered what that custom bias stuff was.

I guess all the more reason to not lead with the technique but instead present this in some other form, like addressing the discrepancy of slower convergence, through some FAQ-like thing or something.

tutorials/physics/using_jolt_physics.rst
mihe: See [my comment](https://github.com/godotengine/godot-docs/pull/10526#issuecomment-2664160981) below. This entire section should be removed in my opinion.

tutorials/physics/using_jolt_physics.rst
mihe: Same as the previous comment. This entire section should be removed in my opinion.

tutorials/physics/using_jolt_physics.rst
mihe: See [my comment](https://github.com/godotengine/godot-docs/pull/10526#issuecomment-2664160981) below. This entire section should be removed in my opinion.

skyace65:
I've gone through and made changes from all 3 reviews. All of AThousandShips changes have been applied, I've made some changes based on jrouwe's feedback.

@mihe I've removed every section you definitely think should be removed. I've also removed Axis-locking, ConvexPolygonShape3D,Scaling shapes/bodies/queries. I've kept the memory section just because the first paragraph does explain what a project setting is for, I think that would be useful to know or at least have written down here, the second paragraph I could take or leave. If you've got more specific changes you'd like me to make please do it as review comments.

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,286 @@
...
+  but expressed as a fraction instead of a percentage.
+- ``solver/active_edge_threshold`` is now ``collisions/active_edge_threshold.``
+- ``solver/bounce_velocity_threshold`` is now ``simulation/bounce_velocity_threshold.``
+- ``solver/contact_speculative_distance`` is now ``simulation/speculative_contact_distance.``
+- ``solver/contact_allowed_penetration`` is now ``simulation/penetration_slop.``
+- ``limits/max_angular_velocity`` is now stored as radians instead.
+- ``limits/max_temporary_memory`` is now ``limits/temporary_memory_buffer_size.``
+
+There might be some discussion to be had with regards to migrating the settings
+values for projects who have previously been relying on the extension.
```
tetrapod00, line 266:
```suggestion
```
Not relevant for docs.

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,286 @@
...
+Overview
+--------
+
+The Jolt physics engine was added as an alternative to the existing GodotPhysics3D
+physics engine in 4.4. Jolt is developed by Jorrit Rouwe with a focus on games and
+VR applications. Previously it was available as a extension but is now built into
+Godot.
+
+It is important to note that the built in Jolt Physics module is considered
+**not finished**, **experimental**, and **lacks feature partiy** with both
```
tetrapod00, line 15:
```suggestion
**not finished**, **experimental**, and **lacks feature parity** with both
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,286 @@
...
+GodotPhysics3D, and the Jolt Extension. Behavior may change as it is developed
+further, please keep that in mind when choosing what to use for your project. A
+roadmap of this module can be found at the bottom of this page.
```
tetrapod00, line 18:
```suggestion
GodotPhysics3D and the Jolt Extension. Behavior may change as it is developed
further. Please keep that in mind when choosing what to use for your project.
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,286 @@
...
+will be merged, and it will be kept compatible with new versions of Godot until
+the built-in module has feature parity with the extension. The extension can be
+found `here on GitHub <https://github.com/godot-jolt/godot-jolt>`_ and in Godot's asset
+library.
+
+To change the 3D physics engine to the Jolt module go to
+:ref:`**Project Settings > Physics > 3D > Physics Engine**<class_ProjectSettings_property_physics/3D/Physics_Engine>`
+and select ``Jolt Physics`` from the dropdown menu. Once you've done that click the
+``Save & Restart`` button, when the editor opens again 3D scenes should now be using
+Jolt for physics.
```
tetrapod00, line 30:
```suggestion
To change the 3D physics engine to the Jolt module, set
:ref:`Project Settings > Physics > 3D > Physics Engine<class_ProjectSettings_property_physics/3D/Physics_Engine>`
to ``Jolt Physics``. Once you've done that, click the **Save & Restart** button.
When the editor opens again, 3D scenes should now be using Jolt for physics.
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,286 @@
+.. _doc_using_jolt_physics:
+
+Using Jolt Physics
+==================
+
+Overview
+--------
```
tetrapod00, line 7:
```suggestion
Introduction
------------
```
"Overview" isn't wrong but "Introduction" is more commonly used 🤷‍♀️

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,286 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~
+When using Jolt, :ref:`class_Area3D` will not detect overlaps with :ref:`class_StaticBody3D`
+(nor a :ref:`class_RigidBody3D` frozen with ``FREEZE_MODE_STATIC``) by default, for
+performance reasons. If you have many/large :ref:`class_Area3D` overlapping with
+complex static geometry, such as :ref:`class_ConcavePolygonShape3D` or
+:ref:`class_HeightMapShape3D`, you can end up wasting a significant amount of CPU
+performance without realizing it.
+
+For this reason this behavior is opt-in through the project setting
+:ref:`physics/jolt_physics_3d/simulation/areas_detect_static_bodies<class_ProjectSettings_property_physics/jolt_physics_3d/simulation/areas_detect_static_bodies>`,
```
tetrapod00, line 47:
```suggestion
:ref:`Physics > Jolt Physics 3D > Simulation > Areas Detect Static Bodies<class_ProjectSettings_property_physics/jolt_physics_3d/simulation/areas_detect_static_bodies>`,
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,286 @@
...
+with the recommendation that you set up your collision layers/masks in such a way that only
```
tetrapod00, line 48:
```suggestion
with the recommendation that you set up your collision layers and masks in such a way that only
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,286 @@
...
+
+Currently an error is emitted if you set these properties to anything but their
+default values.
+
+Single-body joints
+~~~~~~~~~~~~~~~~~~
+
+You can, in Godot, omit one of the joint bodies for a two-body joint and effectively
+have "the world" be the other body. However, the node path that you assign your body
+to (node_a vs node_b) is ignored. Godot Physics will always behave as if you
```
tetrapod00, line 74:
```suggestion
to (:ref:`node_a<class_Joint3D_property_node_a>` vs
:ref:`node_b<class_Joint3D_property_node_b>`) is ignored.
Godot Physics will always behave as if you
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,286 @@
...
+assigned it to node_a, and since node_a is also what defines the frame of reference
+for the joint limits, you end up with inverted limits and a potentially strange
+limit shape, especially if your limits allow both linear and angular degrees of
+freedom.
+
+Jolt will behave as if you assigned the body to node_b instead, with node_a
+representing "the world". There is a project setting called :ref:`physics/jolt_physics_3d/joints/world_node<class_ProjectSettings_property_physics/jolt_physics_3d/joints/world_node>`
```
tetrapod00, line 81:
```suggestion
representing "the world". There is a project setting called :ref:`Physics > Jolt Physics 3D > Joints > World Node<class_ProjectSettings_property_physics/jolt_physics_3d/joints/world_node>`
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,286 @@
...
+Collision margins
+~~~~~~~~~~~~~~~~~
+
+Jolt (and other similar physics engines) uses something that Jolt refers to as
+"convex radius" to help improve the performance and behavior of the types of
+collision detection that Jolt relies on for convex shapes. Other physics engines
+(Godot included) might refer to these as "collision margins" instead. Godot exposes
+these as the margin property on every Shape3D-derived class, as a leftover from the
+Bullet integration in Godot 3, but Godot Physics itself does not use them for
+anything.
```
tetrapod00, line 93:
```suggestion
(Godot included) might refer to these as "collision margins" instead. Godot exposes
these as the ``margin`` property on every Shape3D-derived class, but Godot Physics
itself does not use them for anything.
```
Unsure if the historical trivia is relevant for a docs page (rather than a PR)

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,286 @@
...
+What these collision margins sometimes do in other engines (as described in Godot's
+documentation) is effectively add a "shell" around the shape, slightly increasing
+its size while also rounding off any edges/corners. In Jolt however, these margins
+are first used to shrink the shape, and then the "shell" is applied, resulting in
+edges/corners being similarly rounded off, but without increasing the size of the
+shape.
+
+To prevent having to tweak this margin property manually, since its default value
+can be problematic for smaller shapes, this module exposes a project setting called
+:ref:`physics/jolt_physics_3d/collisions/collision_margin_fraction<class_ProjectSettings_property_physics/jolt_physics_3d/collisions/collision_margin_fraction>`
```
tetrapod00, line 104:
```suggestion
:ref:`Physics > Jolt Physics 3D > Collisions > Collision Margin Fraction<class_ProjectSettings_property_physics/jolt_physics_3d/collisions/collision_margin_fraction>`
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,286 @@
...
+which is multiplied with the smallest axis of the shape's AABB to calculate the
+actual margin. The margin property of the shape is then instead used as an upper
+bound.
+
+These margins should, for most use-cases, be more or less transparent, but can
+sometimes result in odd collision normals when performing shape queries. You can
+lower the above mentioned project setting to mitigate some of this, including
+setting it to ``0``, but too small of a margin can also cause odd collision results,
```
tetrapod00, line 112:
```suggestion
setting it to ``0.0``, but too small of a margin can also cause odd collision results,
```
For floats

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,286 @@
...
+so is generally not recommended.
+
+Baumgarte stabilization
+~~~~~~~~~~~~~~~~~~~~~~~
+
+the baumgarte stabilization jolt uses can result in some artifacts, like piles of
```
tetrapod00, line 118:
```suggestion
The Baumgarte stabilization Jolt uses can result in some artifacts, like piles of
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,286 @@
...
+bodies not separating as quickly as one might expect.
+
+The strength of this stabilization can be tweaked using the project setting
+:ref:`physics/jolt_physics_3d/simulation/baumgarte_stabilization_factor<class_ProjectSettings_property_physics/jolt_physics_3d/simulation/baumgarte_stabilization_factor>`.
```
tetrapod00, line 122:
```suggestion
:ref:`Physics > Jolt Physics 3D > Simulation > Baumgarte Stabilization Factor<class_ProjectSettings_property_physics/jolt_physics_3d/simulation/baumgarte_stabilization_factor>`.
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,286 @@
...
+internal edges of shapes/bodies.
+
+The first technique, called "active edge detection", marks edges of triangles in
+:ref:`class_ConcavePolygonShape3D` or :ref:`class_HeightMapShape3D` as either "active" or "inactive", based on
+the angle to the neighboring triangle. When a collision happens with an inactive
+edge the collision normal will be replaced with the triangle's normal instead, to
+lessen the effect of ghost collisions.
+
+The angle threshold for this active edge detection is configurable through the
+project setting :ref:`physics/jolt_physics_3d/collisions/active_edge_threshold<class_ProjectSettings_property_physics/jolt_physics_3d/collisions/active_edge_threshold>`.
```
tetrapod00, line 140:
```suggestion
project setting :ref:`Physics >Jolt Physics 3D > Collisions > Active Edge Threshold<class_ProjectSettings_property_physics/jolt_physics_3d/collisions/active_edge_threshold>`.
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,286 @@
...
+
+The second technique, called "enhanced internal edge removal", instead adds runtime
+checks to detect whether an edge is active or inactive, based on the contact points
+of the two bodies. This has the benefit of applying not only to collisions with
+:ref:`class_ConcavePolygonShape3D` and :ref:`class_HeightMapShape3D`, but also edges between any shapes within
+the same body.
+
+Enhanced internal edge removal can be toggled on and off for the various contexts to
+which it's applied, using the ``physics/jolt_physics_3d/*/use_enhanced_internal_edge_removal``
+project settings.
```
tetrapod00, line 150:
```suggestion
which it's applied, using the
:ref:`Physics >Jolt Physics 3D > Simulation > Use Enhanced Internal Edge Removal<class_ProjectSettings_property_physics/jolt_physics_3d/simulation/use_enhanced_internal_edge_removal>`,
project setting, and the similar settings for :ref:`queries<class_ProjectSettings_property_physics/jolt_physics_3d/queries/use_enhanced_internal_edge_removal>` and :ref:`motion queries<class_ProjectSettings_property_physics/jolt_physics_3d/motion_queries/use_enhanced_internal_edge_removal>`.
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,286 @@
...
+
+Memory usage
+~~~~~~~~~~~~
+
+Jolt uses a stack allocator for temporary allocations within its simulation step.
+This stack allocator requires allocating a set amount of memory up front, which can
+be configured using the :ref:`physics/jolt_physics_3d/limits/temporary_memory_buffer_size<class_ProjectSettings_property_physics/jolt_physics_3d/limits/temporary_memory_buffer_size>`
```
tetrapod00, line 157:
```suggestion
be configured using the :ref:`Physics > Jolt Physics 3D > Limits > Temporary Memory Buffer Size<class_ProjectSettings_property_physics/jolt_physics_3d/limits/temporary_memory_buffer_size>`
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,286 @@
...
+project setting.
+
+Jolt also makes use of aligned allocations for some of its memory usage, which it
+acquires through function pointers that the implementer assigns. Aligned allocations
+were recently added to Godot in the form of ``Memory::alloc_aligned_static``,
+``Memory::realloc_aligned_static``, and ``Memory::free_aligned_static``, which have all been
+hooked up to Jolt. However, these aligned allocation functions don't currently touch
+``Memory::mem_usage``, which means that some of Jolt's memory won't be tracked, and thus
+the performance monitors in Godot won't be accurate.
```
tetrapod00, line 166:
```suggestion
```
This seems mostly irrelevant to the end user, and touches on Godot internals in a way that we don't do outside of the contributing section

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,286 @@
...
+
+Ray-Cast face index
+~~~~~~~~~~~~~~~~~~~
+
+The face_index property returned in the results of intersect_ray and RayCast3D will
+by default always be ``-1`` with Jolt, the project setting :ref:`physics/jolt_physics_3d/queries/enable_ray_cast_face_index<class_ProjectSettings_property_physics/jolt_physics_3d/queries/enable_ray_cast_face_index>`
+will enable them. The reason for this being that Jolt does not store these indices,
```
tetrapod00, line 173:
```suggestion
The face_index property returned in the results of
:ref:`intersect_ray()<class_PhysicsDirectSpaceState3D_method_intersect_ray>` and RayCast3D will
by default always be ``-1`` with Jolt. The project setting :ref:`Physics > Jolt Physics 3D > Queries > Enable Ray Cast Face Index<class_ProjectSettings_property_physics/jolt_physics_3d/queries/enable_ray_cast_face_index>`
will enable them. The reason for this being that Jolt does not store these indices,
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,286 @@
...
+and for them to be stored per-triangle userdata must be enabled, which adds about
+25% extra memory usage to the underlying Jolt implementation of :ref:`class_ConcavePolygonShape3D`.
+
+Kinematic RigidBody3D contacts
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When using Jolt, a :ref:`class_RigidBody3D` frozen with ``FREEZE_MODE_KINEMATIC`` will by default not
```
tetrapod00, line 180:
```suggestion
When using Jolt, a :ref:`class_RigidBody3D` frozen with
:ref:`FREEZE_MODE_KINEMATIC<class_RigidBody3D_constant_FREEZE_MODE_KINEMATIC>`
will by default not
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,286 @@
...
+report contacts from collisions with other static/kinematic bodies, for performance
+reasons, even when setting a non-zero max_contacts_reported. If you have many/large
+kinematic bodies overlapping with complex static geometry, such as :ref:`class_ConcavePolygonShape3D`
+or :ref:`class_HeightMapShape3D`, you can end up wasting a significant amount of CPU
+performance without realizing it.
+
+For this reason this behavior is opt-in through the project setting
+:ref:`physics/jolt_physics_3d/simulation/generate_all_kinematic_contacts<class_ProjectSettings_property_physics/jolt_physics_3d/simulation/generate_all_kinematic_contacts>`.
```
tetrapod00, line 188:
```suggestion
:ref:`Physics > Jolt Physics 3D > Simulation > Generate All Kinematic Contacts<class_ProjectSettings_property_physics/jolt_physics_3d/simulation/generate_all_kinematic_contacts>`.
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,286 @@
...
+
+WorldBoundaryShape3D
+~~~~~~~~~~~~~~~~~~~~
+
+:ref:`class_WorldBoundaryShape3D`, which is meant to represent an infinite plane, is
+implemented a bit differently in Jolt compared to Godot Physics. Both engines have
+an upper limit for how big the effective size of this plane can be, but this size is
+much smaller when using Jolt, in order to avoid precision issues.
+
+You can configure this size using the :ref:`physics/jolt_physics_3d/limits/world_boundary_shape_size<Class_ProjectSettings_Property_physics/jolt_physics_3d/limits/world_boundary_shape_size>`
```
tetrapod00, line 215:
```suggestion
You can configure this size using the :ref:`Physics > Jolt Physics 3D > Limits > World Boundary Shape Size<class_ProjectSettings_Property_physics/jolt_physics_3d/limits/world_boundary_shape_size>`
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,286 @@
...
+project setting.
+
+Notable differences to Godot Jolt
+---------------------------------
```
tetrapod00, line 219:
```suggestion
Notable differences to the Godot Jolt extension
-----------------------------------------------
```
To avoid confusion

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,286 @@
...
+
+The joint nodes that are exposed in the Godot Jolt extension (JoltPinJoint3D,
+JoltHingeJoint3D, JoltSliderJoint3D, JoltConeTwistJoint3D, and JoltGeneric6DOFJoint)
+have not been included with this module.
+
+Thread Safety
+~~~~~~~~~~~~~
+
+Unlike the Godot Jolt extension, this module does have experimental thread-safety,
+including support for the :ref:`physics/3d/run_on_separate_thread<class_ProjectSettings_Property_physics/3d/run_on_separate_thread>`
```
tetrapod00, line 279:
```suggestion
including support for the :ref:`Physics > 3D > Run On Separate Thread<class_ProjectSettings_Property_physics/3d/run_on_separate_thread>`
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,286 @@
...
+project setting. This is achieved by utilizing the same wrapper server that's used
+by Godot Physics, called ``PhysicsServer3DWrapMT``, as well as introducing a mutex
+around the rebuilding of shapes, since concurrent shape-casts could otherwise
+trigger a race condition.
+
+This has however not been tested very thoroughly, so should be considered
+experimental.
```
tetrapod00, line 286:
```suggestion
project setting. This should be considered experimental.
```
Implementation details are not relevant for users. Don't put "hasn't been tested thoroughly" in the docs :)

tetrapod00 (commented):
Reviewed for docs style stuff, and pointed out sections that I don't think are relevant for end users.

Can't review for physics.

I assume that this page will get significant changes during the duration of 4.4, to better convey information that is relevant to users (rather than PR reviewers). I think that the current page is a lot better than nothing even if it isn't the thing that would have been written from scratch for end users.

skyace65:
Applied changes from tetrapod's review

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,272 @@
...
+These margins should, for most use-cases, be more or less transparent, but can
+sometimes result in odd collision normals when performing shape queries. You can
+lower the above mentioned project setting to mitigate some of this, including
+setting it to ``0.0``, but too small of a margin can also cause odd collision results,
+so is generally not recommended.
+
+Baumgarte stabilization
+~~~~~~~~~~~~~~~~~~~~~~~
+
+the Baumgarte stabilization Jolt uses can result in some artifacts, like piles of
```
tetrapod00, line 117:
```suggestion
The Baumgarte stabilization Jolt uses can result in some artifacts, like piles of
```

tetrapod00 (approved):

mihe (changes requested):
I did a rough pass on what's there currently. Apologies for not formatting the hard wrapping correctly in my suggestions, as well as not formatting the references correctly. I wasn't sure what the exact format was.

Outside of the review comments, I have some other general nitpicks:

1. There seems to be a mix of referring to Godot Physics as either "Godot Physics" or "GodotPhysics3D". I would probably suggest we pick one or the other. I don't believe it has any truly official canonical name though. "GodotPhysics3D" is what's shown in the list when picking the physics engine, but there are several places where the engine is referred to as "Godot Physics", including the documentation.
2. There's a couple of places referring to "this module". Talking about engine modules in general seems a bit like "contributor speak" to me. Maybe it's worth rephrasing it to be "the Jolt Physics integration" instead or something?
3. There's a mix of different casings for the headings, where some have the first word capitalized (e.g. "Joint properties") but some are all capitalized (e.g. "Thread Safety").

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,272 @@
+.. _doc_using_jolt_physics:
+
+Using Jolt Physics
+==================
+
+Introduction
+------------
+
+The Jolt physics engine was added as an alternative to the existing GodotPhysics3D
```
tetrapod00, line 9:
```suggestion
The Jolt physics engine was added as an alternative to the existing Godot Physics
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,272 @@
...
+
+To change the 3D physics engine to the Jolt module, set
+:ref:`Project Settings > Physics > 3D > Physics Engine<class_ProjectSettings_property_physics/3D/Physics_Engine>`
+to ``Jolt Physics``. Once you've done that, click the **Save & Restart** button.
+When the editor opens again, 3D scenes should now be using Jolt for physics.
+
+Notable differences to Godot Physics
+------------------------------------
+
+There are many differences between the existing GodotPhysics3D engine and Jolt.
```
tetrapod00, line 33:
```suggestion
There are many differences between the Godot Physics engine and Jolt.
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,272 @@
...
+------------
+
+The Jolt physics engine was added as an alternative to the existing GodotPhysics3D
+physics engine in 4.4. Jolt is developed by Jorrit Rouwe with a focus on games and
+VR applications. Previously it was available as a extension but is now built into
+Godot.
+
+It is important to note that the built in Jolt Physics module is considered
+**not finished**, **experimental**, and **lacks feature parity** with both
+GodotPhysics3D and the Jolt Extension. Behavior may change as it is developed
```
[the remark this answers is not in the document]
tetrapod00:
```suggestion
Godot Physics and the Jolt Extension. Behavior may change as it is developed
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,272 @@
...
+of Jolt's own joints. As such, there are a number of joint properties that are not
+supported, mainly ones related to configuring the joint's soft limits.
+
+The unsupported properties are:
+
+- PinJoint3D: ``bias``, ``damping``, ``impulse_clamp``
+- HingeJoint3D: ``bias``, ``softness``, ``relaxation``
+- SliderJoint3D: ``angular_\*``, ``\*_limit/softness``, ``\*_limit/restitution``, ``\*_limit/damping``
+- ConeTwistJoint3D: ``bias``, ``relaxation``, ``softness``
+- Generic6DOFJoint3D: ``*_limit_*/softness``, ``*_limit_*/restitution``, ``*_limit_*/damping``, ``*_limit_*/erp``
```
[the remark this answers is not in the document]
tetrapod00: Usually I'm in favor of explicitly listing, but since there are so many (24?), in this case what you already have is fine.

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,272 @@
...
+to (:ref:`node_a<class_Joint3D_property_node_a>` vs :ref:`node_b<class_Joint3D_property_node_b>`)
+is ignored. Godot Physics will always behave as if you
+assigned it to node_a, and since node_a is also what defines the frame of reference
+for the joint limits, you end up with inverted limits and a potentially strange
+limit shape, especially if your limits allow both linear and angular degrees of
+freedom.
+
+Jolt will behave as if you assigned the body to node_b instead, with node_a
+representing "the world". There is a project setting called :ref:`Physics > Jolt Physics 3D > Joints > World Node<class_ProjectSettings_property_physics/jolt_physics_3d/joints/world_node>`
+that lets you toggle this behavior, if you need compatibility for an existing project.
```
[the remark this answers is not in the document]
tetrapod00: Formatted as code is ideal. They don't need links after the first mention. I was a bit lazy with my style suggestions overall, so I missed some cases like these.

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,272 @@
...
+actual margin. The margin property of the shape is then instead used as an upper
+bound.
+
+These margins should, for most use-cases, be more or less transparent, but can
+sometimes result in odd collision normals when performing shape queries. You can
+lower the above mentioned project setting to mitigate some of this, including
+setting it to ``0.0``, but too small of a margin can also cause odd collision results,
+so is generally not recommended.
+
+Baumgarte stabilization
```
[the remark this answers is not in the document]
tetrapod00: I'm okay with removing and adding a new section during the 4.5 cycle (and cherrypicking). I'd want a second maintainer opinion on that.

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,272 @@
...
+will enable them. The reason for this being that Jolt does not store these indices,
+and for them to be stored per-triangle userdata must be enabled, which adds about
+25% extra memory usage to the underlying Jolt implementation of :ref:`class_ConcavePolygonShape3D`.
+
+Kinematic RigidBody3D contacts
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When using Jolt, a :ref:`class_RigidBody3D` frozen with :ref:`FREEZE_MODE_KINEMATIC<class_RigidBody3D_constant_FREEZE_MODE_KINEMATIC>`
+will by default not report contacts from collisions with other static/kinematic
+bodies, for performance reasons, even when setting a non-zero max_contacts_reported.
```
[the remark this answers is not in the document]
tetrapod00:
```suggestion
bodies, for performance reasons, even when setting a non-zero :ref:`max_contacts_reported<class_RigidBody3D_property_max_contacts_reported>`.
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,272 @@
...
+- ``solver/position_correction`` is now ``simulation/baumgarte_stabilization_factor``,
+  but expressed as a fraction instead of a percentage.
+- ``solver/active_edge_threshold`` is now ``collisions/active_edge_threshold.``
+- ``solver/bounce_velocity_threshold`` is now ``simulation/bounce_velocity_threshold.``
+- ``solver/contact_speculative_distance`` is now ``simulation/speculative_contact_distance.``
+- ``solver/contact_allowed_penetration`` is now ``simulation/penetration_slop.``
+- ``limits/max_angular_velocity`` is now stored as radians instead.
+- ``limits/max_temporary_memory`` is now ``limits/temporary_memory_buffer_size.``
+
+Joint Nodes
```
tetrapod00, line 257:
```suggestion
Joint nodes
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,272 @@
...
+~~~~~~~~~~~
+
+The joint nodes that are exposed in the Godot Jolt extension (JoltPinJoint3D,
+JoltHingeJoint3D, JoltSliderJoint3D, JoltConeTwistJoint3D, and JoltGeneric6DOFJoint)
+have not been included with this module.
+
+Thread Safety
```
tetrapod00, line 264:
```suggestion
Thread safety
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,272 @@
...
+You can configure this size using the :ref:`Physics > Jolt Physics 3D > Limits > World Boundary Shape Size<class_ProjectSettings_Property_physics/jolt_physics_3d/limits/world_boundary_shape_size>`
+project setting.
+
+Notable differences to the Godot Jolt extension
+-----------------------------------------------
+
+While this module is largely a straight port of the Godot Jolt extension, with a lot
+of cosmetic changes, there are a few things that are different.
+
+Project Settings
```
tetrapod00, line 216:
```suggestion
Project settings
```
This one could go either way - "project settings" is used throughout the docs to refer to individual settings. So if capitalized with our rules (only the first word and proper nouns or node class names capitalized), "Project settings" is correct here.

"Project Settings" is also the name of a window in the editor so I guess it's fine as is too.

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,272 @@
...
+sometimes result in odd collision normals when performing shape queries. You can
+lower the above mentioned project setting to mitigate some of this, including
+setting it to ``0.0``, but too small of a margin can also cause odd collision results,
+so is generally not recommended.
+
+Baumgarte stabilization
+~~~~~~~~~~~~~~~~~~~~~~~
+
+the Baumgarte stabilization Jolt uses can result in some artifacts, like piles of
+bodies not separating as quickly as one might expect.
```
jrouwe, line 118:
```suggestion
Baumgarte stabilization is a method to resolve penetrating bodies and push them in a state where they are just touching. In Godot Physics this works like a spring. This means that bodies can accellerate and may cause the bodies to overshoot and separate completely. With Jolt, the stabilization is only applied to the position and not to the velocity of the body. This means it cannot overshoot but it may take longer to resolve the penetration.
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,272 @@
...
+
+The strength of this stabilization can be tweaked using the project setting
+:ref:`Physics > Jolt Physics 3D > Simulation > Baumgarte Stabilization Factor<class_ProjectSettings_property_physics/jolt_physics_3d/simulation/baumgarte_stabilization_factor>`.
+Setting this project setting to ``1.0`` will resolve penetration in 1 simulation
+step, resulting in a simulation that more closely resembles Godot Physics, but which
+is also more unstable.
```
jrouwe, line 124:
```suggestion
Setting this project setting to ``0.0`` will turn Baumgarte stabilization off. Setting it to ``1.0`` will resolve penetration in 1 simulation step. This is fast but often also unstable.
```

jrouwe (commented):
@mihe How about this?

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,272 @@
...
+The joint nodes that are exposed in the Godot Jolt extension (JoltPinJoint3D,
+JoltHingeJoint3D, JoltSliderJoint3D, JoltConeTwistJoint3D, and JoltGeneric6DOFJoint)
+have not been included with this module.
+
+Thread Safety
+~~~~~~~~~~~~~
+
+Unlike the Godot Jolt extension, this module does have experimental thread-safety,
+including support for the :ref:`Physics > 3D > Run On Separate Thread<class_ProjectSettings_Property_physics/3d/run_on_separate_thread>`
+project setting. This should be considered experimental.
```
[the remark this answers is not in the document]
skyace65: Going to rephrase this section a bit more since it being experimental is mentioned on the first line as well

skyace65:
Applied changes from @mihe and tetrapod's reviews. Waiting for Mihe to review jrouwe's comments on bugmarite before changing anything.

GodotPhysics3D is now Godot Physics everywhere, and section name casing has been standardized. In regards to the use of "this module" Maybe it would be better to just say Jolt?

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,269 @@
...
+much smaller when using Jolt, in order to avoid precision issues.
+
+You can configure this size using the :ref:`Physics > Jolt Physics 3D > Limits > World Boundary Shape Size<class_ProjectSettings_Property_physics/jolt_physics_3d/limits/world_boundary_shape_size>`
+project setting.
+
+Notable differences to the Godot Jolt extension
+-----------------------------------------------
+
+While this module is largely a straight port of the Godot Jolt extension, there are
+a few things that are different.
```
tetrapod00, line 213:
```suggestion
While the built-in Jolt module is largely a straight port of the Godot Jolt extension,
there are a few things that are different.
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,269 @@
...
+- ``solver/contact_allowed_penetration`` is now ``simulation/penetration_slop.``
+- ``limits/max_angular_velocity`` is now stored as radians instead.
+- ``limits/max_temporary_memory`` is now ``limits/temporary_memory_buffer_size.``
+
+Joint nodes
+~~~~~~~~~~~
+
+The joint nodes that are exposed in the Godot Jolt extension (JoltPinJoint3D,
+JoltHingeJoint3D, JoltSliderJoint3D, JoltConeTwistJoint3D, and JoltGeneric6DOFJoint)
+have not been included with this module.
```
tetrapod00, line 261:
```suggestion
JoltHingeJoint3D, JoltSliderJoint3D, JoltConeTwistJoint3D, and JoltGeneric6DOFJoint)
have not been included in the Jolt module.
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,269 @@
...
+
+What these collision margins sometimes do in other engines (as described in Godot's
+documentation) is effectively add a "shell" around the shape, slightly increasing
+its size while also rounding off any edges/corners. In Jolt however, these margins
+are first used to shrink the shape, and then the "shell" is applied, resulting in
+edges/corners being similarly rounded off, but without increasing the size of the
+shape.
+
+To prevent having to tweak this margin property manually, since its default value
+can be problematic for smaller shapes, this module exposes a project setting called
```
tetrapod00, line 102:
```suggestion
can be problematic for smaller shapes, Jolt exposes a project setting called
```
This is technically not quite correct, since Jolt itself != Jolt module in Godot != Jolt extension. However, I think that for most users of Godot, there isn't really a difference between Jolt and Jolt as implemented within Godot. So except for the cases where we do need to distinguish between the module and the extension in this page, we could just say "Jolt"?

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,269 @@
...
+
+Due to limitations internal to Jolt, the contact impulses provided by :ref:`PhysicsDirectBodyState3D.get_contact_impulse()<class_physicsdirectbodystate3d_method_get_contact_impulse>`
+are estimated ahead of time based on things like the contact manifold and velocities
+of the colliding bodies. This means that the reported impulses will only be accurate
+in cases where the two bodies in question are not colliding with any other bodies.
+
+Area3D and SoftBody3D
+~~~~~~~~~~~~~~~~~~~~~
+
+This module does not currently support any interactions between :ref:`class_SoftBody3D`
```
tetrapod00, line 194:
```suggestion
Jolt does not currently support any interactions between :ref:`class_SoftBody3D`
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,269 @@
...
+~~~~~~~~~~~
+
+The joint nodes that are exposed in the Godot Jolt extension (JoltPinJoint3D,
+JoltHingeJoint3D, JoltSliderJoint3D, JoltConeTwistJoint3D, and JoltGeneric6DOFJoint)
+have not been included with this module.
+
+Thread safety
+~~~~~~~~~~~~~
+
+Unlike the Godot Jolt extension, this module does have thread-safety,
```
tetrapod00, line 266:
```suggestion
Unlike the Godot Jolt extension, the Jolt module does have thread-safety,
```

tutorials/physics/using_jolt_physics.rst
skyace65: I was actually thinking of going with "Jolt has a project setting" or would that still be technically incorrect?

mihe: I would personally prefer that we differentiate between Jolt itself and Jolt as integrated in Godot. This setting in particular is not something that Jolt itself offers, and is purely something introduced by the integration.

But yes, I realize users talk about Jolt as integrated in Godot as just "Jolt", which I find somewhat unfortunate myself, as people then start to attribute quirks or shortcomings of the integration to be shortcomings of Jolt itself, so I would rather see that we don't exacerbate that.

I don't see it being that much more verbose to just say "the Jolt module" instead.

```suggestion
can be problematic for smaller shapes, the Jolt module exposes a project setting called
```

I guess the question is whether referring to it as "the Jolt module" makes that distinction clear enough, or whether it would be better served by calling it "the Jolt integration" (or "the Jolt Physics integration", to use its canonical name).

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,269 @@
...
+
+The unsupported properties are:
+
+- PinJoint3D: ``bias``, ``damping``, ``impulse_clamp``
+- HingeJoint3D: ``bias``, ``softness``, ``relaxation``
+- SliderJoint3D: ``angular_\*``, ``\*_limit/softness``, ``\*_limit/restitution``, ``\*_limit/damping``
+- ConeTwistJoint3D: ``bias``, ``relaxation``, ``softness``
+- Generic6DOFJoint3D: ``*_limit_*/softness``, ``*_limit_*/restitution``, ``*_limit_*/damping``, ``*_limit_*/erp``
+
+Currently an warning is emitted if you set these properties to anything but their
```
mihe, line 65:
```suggestion
Currently a warning is emitted if you set these properties to anything but their
```

tutorials/physics/using_jolt_physics.rst
mihe: I can't speak to the technical details, but it seems alright to me.

I still think we need to rethink this section in general, as people who find the slower convergence to be a problem aren't likely to look for Baumgarte stabilization in the documentation, but that can be done later.

Minor spelling/grammatical notes:
- "push them in a state" -> "push them to a state"?
- "accellerate" -> "accelerate"

tutorials/physics/using_jolt_physics.rst
mihe: Is it all that useful to mention the `0.0` case? As far as I can tell it just prevents the bodies from moving at all.

Clearly the statement of "more closely resembles Godot Physics" was incorrect on my part, but I must admit I'm a bit confused by this notion that `0.0` would be what disables it, as opposed to `1.0`. I would have thought that a 100% error correction would be the default case in any solver, and that anything less is what constitutes the effect of sacrificing correctness for stability, but perhaps I've got that all wrong.

mihe (changes requested):
I haven't done a full pass on it, but there are a number of comments from the previous review that haven't been addressed (maybe because GitHub collapsed them as hidden?).

I went through and resolved the comments that were addressed (or stale).

jrouwe:
> Is it all that useful to mention the 0.0 case? As far as I can tell it just prevents the bodies from moving at all.

The Baumgarte constant is multiplied by the position error. The position error of a contact constraint is basically how much the two bodies penetrate. So multiplying by 0 means that the system is off, multiplying by 1 means the error is fully corrected. The bodies should still move with a value of 0, they will just not be able to be pushed apart anymore once they penetrate.

I mentioned 0 because the value can be between 0 and 1 and I thought it would be useful to specify what the extreme values do.

tutorials/physics/using_jolt_physics.rst
tetrapod00: The reasoning to distinguish Jolt and Jolt in Godot makes sense, agreed.

"the Jolt module" seems okay. But usually we don't refer to things as modules outside of contributing/compiling documentation, since it doesn't matter for users. It makes sense in this case because this page is partly for users migrating away from the extension.

"the Jolt integration" or "the Jolt Physics integration" seem better to me. It will be more future-proof after the module reaches feature parity and stability, and the section on differences to the extension is (likely) removed as no longer relevant in several minor versions.

One question though - is it okay to use more colloquial and non-specific "when using Jolt", "enable Jolt", etc, in *other* documentation pages that aren't specifically about the Jolt integration? Or should we try to enforce a "Jolt integration" or "Jolt Physics integration" phrasing where possible?

mihe:
> The Baumgarte constant is multiplied by the position error. The position error of a contact constraint is basically how much the two bodies penetrate. So multiplying by 0 means that the system is off, multiplying by 1 means the error is fully corrected.

Right. I see what you mean. Perhaps I'm reading too much into the name/description of the constant.

> The bodies should still move with a value of 0, they will just not be able to be pushed apart anymore once they penetrate.

Yes, I guess what I meant to say was "prevent the bodies from depenetrating at all".

> I mentioned 0 because the value can be between 0 and 1 and I thought it would be useful to specify what the extreme values do.

Fair enough!

tutorials/physics/using_jolt_physics.rst
mihe: > One question though - is it okay to use more colloquial and non-specific "when using Jolt", "enable Jolt", etc, in _other_ documentation pages that aren't specifically about the Jolt integration? Or should we try to enforce a "Jolt integration" or "Jolt Physics integration" phrasing where possible?

I guess it might depend on the context, but in general I don't personally see any problem with calling it just "Jolt" or "Jolt Physics" in those cases.

@jrouwe I'm curious, I've tried my best to refer to the library as "Jolt Physics" as opposed to just "Jolt" ever since the integration, since that is its name after all, and is what's prominently displayed in the repository, but you yourself often refer to it as just "Jolt", so how strongly do you feel about one over the other in more "official mediums", like this kind of documentation?

jrouwe: I agree that it's probably best to use Jolt Physics everywhere, but I also often just type Jolt because it's shorter and I'm lazy. In any case I don't really mind if people call it Jolt.

tetrapod00: Then probably best to try to use "Jolt Physics" when possible as a rule/guideline in the docs. But it's okay if just "Jolt" slips in. Our documentation is often a bit informal.

skyace65:
Just went through and applied changes from all the reviews. Apologies about the ones I missed before.

jrouwe (approved):
Looks good to me!

tetrapod00 (approved):

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,278 @@
...
+You can, in Godot, omit one of the joint bodies for a two-body joint and effectively
+have "the world" be the other body. However, the node path that you assign your body
+to (:ref:`node_a<class_Joint3D_property_node_a>` vs :ref:`node_b<class_Joint3D_property_node_b>`)
+is ignored. Godot Physics will always behave as if you
+assigned it to ``node_a``, and since ``node_a`` is also what defines the frame of reference
+for the joint limits, you end up with inverted limits and a potentially strange
+limit shape, especially if your limits allow both linear and angular degrees of
+freedom.
+
+Jolt will behave as if you assigned the body to node_b instead, with node_a
```
mihe, line 80:
```suggestion
Jolt will behave as if you assigned the body to ``node_b`` instead, with ``node_a``
```

tutorials/physics/using_jolt_physics.rst
```diff
@@ -0,0 +1,278 @@
...
+
+What these collision margins sometimes do in other engines (as described in Godot's
+documentation) is effectively add a "shell" around the shape, slightly increasing
+its size while also rounding off any edges/corners. In Jolt however, these margins
+are first used to shrink the shape, and then the "shell" is applied, resulting in
+edges/corners being similarly rounded off, but without increasing the size of the
+shape.
+
+To prevent having to tweak this margin property manually, since its default value
+can be problematic for smaller shapes, the Jolt Module exposes a project setting
```
mihe, line 102:
```suggestion
can be problematic for smaller shapes, the Jolt module exposes a project setting
```

mihe (approved):
Some minor nitpicks left, but what's there looks fine otherwise.

skyace65:
I've fixed the nitpicks

skyace65:
@tetrapod00 if you think we're good on reviews now please merge this, there's nothing else I want to change

tetrapod00:
Thank you, and thank you everyone for the thorough reviews!

PR #10532, opened 2025-01-21
Document recovery mode

skyace65:
Documents the recovery mode added in https://github.com/godotengine/godot/pull/92563. The information is basically what shows up in the popup asking if you want to open a project in recovery mode, which appears after opening a project that crashed.

I added this to the project manager page since that's where you'd go to then open a project in recovery mode. It felt a bit weird to have all the information there even though it does fit, so I added a small section to the troubleshooting page pointing people towards it. If someone thinks this would fit better somewhere else I'm open to suggestions.


mhilbrunner (approved):

mhilbrunner:
Thanks!

Issue #10417, opened 2024-12-17
Provide a Blender to Godot Engine manifold tutorial


fire:
**Your Godot version:**

Godot Engine master. https://github.com/godotengine/godot/commit/4364ed6ccd001cbfe7cb781d074100695c878d90

**Issue description:**

@tetrapod00 I have a draft of the process for making manifold shapes using Blender's fixing tool. ~~https://docs.google.com/document/d/11KPDwMuZl3c4NSwanhatekAIvUxktURmXJVPkP_a17o/edit?tab=t.0#heading=h.r41iui30hwv7~~

Let me know if I should work on it more.

[Techniques for making non-manifold meshes manifold via mathematics.pdf](https://github.com/user-attachments/files/18445825/Techniques.for.making.non-manifold.meshes.manifold.via.mathematics.pdf)

fire:
I'll use this extremely technical and precise language for manifold.

> Every edge of every triangle must contain the same two vertices (by index) as exactly one other triangle edge, and the start and end vertices must switch places between these two edges. The triangle vertices must appear in clockwise order when viewed from the outside of the Godot Engine manifold mesh.

Which is modified from https://github.com/elalish/manifold/wiki/Manifold-Library#manifoldness-definition

@elalish Is this modification correct?

tetrapod00:
I think it would make sense to split this into two PRs, actually:
1. Update the existing CSG mesh requirements to mention "manifold". IMO we should include both technical and non-technical explanations of manifoldness. So maybe keep the existing one, and put the technical one after. Linking to the definition used by the library is good to do, too.
2. Add a tutorial for using blender to fix manifold issues, which may need more reviews.

fire:
I would appreciate assistance on those two pull requests when there is consensus.

tetrapod00:
As in, help making the PR at all? If so, sure, I don't mind doing that for either or both.

I think the change to manifold descriptions is pretty uncontroversial so that can be made now and we can discuss changes in the PR.
For the tutorial we should wait for another maintainer to comment maybe

elalish:
LGTM

fire:
We didn't get much pushback does it mean we can do it?

fire:
I'll close this for now; feel free to salvage the pull request.

tetrapod00:
[...]

PR #10540, opened 2025-01-21
Add small tutorial on making existing mesh manifold with Blender

tetrapod00:
Closes https://github.com/godotengine/godot-docs/issues/10417.

Personally I do think this is a marginal inclusion, since it's documentation on using a specific addon for another program. But it's also a pretty small inclusion and addresses a potential pain point for users updating to 4.4.

I wrote the version of the tutorial that I think is the right scope for the official docs, and we can either approve it or not.

CC @fire


skyace65 (approved):
While an argument could be made that this is out of scope I think this is small enough and justified enough to be included in the docs, I'm also ok with linking to a non-godot created extension since it's hosted on blender.org. We've done similar things to this in the past I still agree with like showing users how to make ico files because it's an obscure process.

mhilbrunner (approved):

mhilbrunner:
I agree with Sky. Merged! @fire Feel free to ping here or on RC #docs if there are any concerns with this :) Noticed the request for your review too late.

fire:
Additional tips for people who want to make manifold in Blender coming from a search engine.

1. You can merging "near" vertices.
2. Another trick when there is a few vertices left is go into edit mode and select with traits of non-manifold and delete them.
3. The make manifold tutorial in this pull request can help here too.
4. Mix and match the listed techniques.

PR #10541, opened 2025-01-21
Add links from GDScript keywords table to relevant sections

tetrapod00:
Matches the style of the existing entries with the `See Thing.` format, which also avoids the need for a new anchor.

Made in response to user feedback about the enum entry in discord; and expanded to all the cases I could find with an obvious link target.


skyace65 (approved):

skyace65:
Thanks!

PR #10543, opened 2025-01-22
Update `rainbow` bbcode options - add `speed`

havi05:
https://github.com/godotengine/godot/pull/100494 added a new option to the bbcode effect `rainbow` called `speed` to fix the animation direction. Updated the documentation to include this.

### DIFF
--- a/tutorials/ui/bbcode_in_richtextlabel.rst
+++ b/tutorials/ui/bbcode_in_richtextlabel.rst
@@ -1015,12 +1015,13 @@ Rainbow
 .. image:: img/bbcode_in_richtextlabel_effect_rainbow.webp
 
 Rainbow gives the text a rainbow color that changes over time. Its tag format is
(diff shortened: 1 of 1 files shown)


skyace65 (approved):

skyace65:
Thanks a ton! We really appreciate it when engine contributors document their PRs

havi05:
Since the mentioned pr is considered for cherry-picking, do I also have to create pr's for these branches (4.3, 4.2) or will this be done by the maintainers if needed?

Thanks for the kind words!

tetrapod00:
@havi05 Yeah, you don't need to take any more actions.

To clarify a little more, while the engine PR is marked for cherrypicking, so far it has *not* been cherrypicked, and therefore the documentation PR isn't marked for cherrypicking. We wouldn't want to document features that don't exist yet. Right now we usually only cherrypick back a single minor version, so there is a chance that the documentation for this feature will only ever exist in the 4.4 and forward version of the documentation, depending on the timing of everything.

PR #10554, opened 2025-01-28
Fully update first 3D Shader page

skyace65:
Looks like we went through the page at some point and updated the editor screenshots, but not the screenshots of the shader in the editor.

- All the shader screenshots as well as 2 editor ones that had slight changes have been updated (All have gone through squoosh compression)
- Updated `"[empty]"` in the text since it's now `"<empty>"`
- Added a line of text about using default shader settings
- Changed instructions for getting the shader editor open to reflect current behavior
- Updated the text on what the shader is generated with since ``vertex()`` is now generated by default
- Simplifying the text related to the ``fragment()`` function.


tutorials/shaders/your_first_shader/your_first_3d_shader.rst
```diff
@@ -63,7 +63,7 @@ Setting up
 
 Add a new :ref:`MeshInstance3D <class_MeshInstance3D>` node to your scene.
 
-In the inspector tab beside "Mesh" click "[empty]" and select "New PlaneMesh".
+In the inspector tab beside "Mesh" click "<empty>" and select "New PlaneMesh".
```
tetrapod00, line 66:
```suggestion
In the inspector tab, set the **Mesh** property to a new PlaneMesh.
```
I'm not a fan of how this page spells out how to create new resources by `clicking on "<empty>"` each time. I feel like a shader tutorial should expect a little more familiarity with the inspector UI than that?

tutorials/shaders/your_first_shader/your_first_3d_shader.rst
```diff
@@ -315,19 +304,10 @@ Set this second uniform texture to another :ref:`NoiseTexture2D <class_noisetext
...
-
-  void fragment() {
-  }
-
 When we have normals that correspond to a specific vertex we set ``NORMAL``, but
 if you have a normalmap that comes from a texture, set the normal using
-``NORMAL_MAP``. This way Godot will handle the wrapping of texture around the
-mesh automatically.
+``NORMAL_MAP`` in the ``fragment()`` function. This way Godot will handle the
+wrapping of texture around the mesh automatically.
```
tetrapod00, line 310: The section on setting the normals *is* a little broken, as indicated in https://github.com/godotengine/godot-docs/pull/10501. I think that PR suggests the wrong fix, though.

I believe the intention with the tutorial is that the plane mesh normals are to be corrected by reading from a normal map. But since the normal map texture has high resolution, and the height is changed in `vertex()` but the normals are changed in `fragment()`, it not only corrects for the height, but also adds a bunch of extra normal detail. I *think* the right solution here might be to set the NORMAL value in `vertex()` rather than setting a normal map here. I'm not 100% sure, though.

tetrapod00 (approved):
I followed along with the tutorial. The images in the page now match the results of following along in 4.4beta1.

I approve the image changes. However, this page still needs a more significant refresh to fix a few things:
- Nitpicks around the way we stylize things like setting properties in the inspector
- Some confusion around the purpose and method of setting the normal values
- A few other inaccuracies and nitpicks I noticed

I would recommend merging this as-is, and I can try a followup PR to make the more extensive changes.

tutorials/shaders/your_first_shader/your_first_3d_shader.rst
skyace65: This is the only page I can think of that actually describes clicking on "<empty>"

tetrapod00: Search the page for "empty" and it comes up at least four times

tetrapod00: Code_WCtdPb9psl
It comes up a couple other times. I think it's fine and justified in the Getting Started tutorials, since it makes it extremely clear what to do. My guess is that it's a stylistic convention that was being followed at the time these pages were written.

I think we should move towards describing cases like this as "setting the Whatever property to a new Thing". If necessary we could also add in a general tutorial somewhere in the docs (either in Getting Started or maybe in the Inspector page) about which buttons to click to set properties, so this is explained in one place rather than piecemeal wherever it comes up. Something for another PR.

mhilbrunner (approved):
Approving the changes here, I'll leave it to you two whether you're merging this as is and do a followup or do the further fixes in this PR. :)

Good work!

tetrapod00:
@skyace65 If you want to do any of the rewording away from `click on "<empty>"` in this PR feel free.

Other than that I would recommend merging this one as-is and leaving the rest for a followup PR, since the changes are a combination of nitpicky style stuff, also may require a proper review from shaders/rendering if we end up changing the instructions around setting the normals. This is already a great improvement as is.

tutorials/shaders/your_first_shader/your_first_3d_shader.rst
AThousandShips, line 310:
```suggestion
``NORMAL_MAP`` in the ``fragment()`` function. This way Godot will handle
wrapping the texture around the mesh automatically.
```
Or

```suggestion
``NORMAL_MAP`` in the ``fragment()`` function. This way Godot will handle the
wrapping of the texture around the mesh automatically.
```

AThousandShips (approved):

tetrapod00:
Thank you, merged! I'll get started on the more extensive changes when I get a chance.

PR #10562, opened 2025-01-29
Update compiling_for_windows.rst

f2l2pe:
mingw already includes make and gcc

the scons version that scoop installs is also outdated

this was causing a build error on windows since it was using an older version of GCC and Scons

Scons need to be installed from pip


skyace65 (changes requested):
Tested the command and everything worked. Just need to add that scons still needs to be installed via pip

skyace65 (approved):

skyace65:

Issue #10458, opened 2025-01-01
Add clarification about what CharacterBody is used for


JamesC01:
**Your Godot version:**
4.3

**Issue description:**
I've used Godot casually for a few years, and I'm still a bit uncertain what the recommended usage of CharacterBody is. The name indicates that you should use it for characters, like the player, or enemies that move, but after reading the docs, I'm a bit confused. It seems like it's absolutely usable for stuff other than characters, considering all of the extra collision detection information it gives you, and it doesn't seem to mention that it's only for characters, but I'm still unsure of if it's recommended outside of them, since the name makes it seem like that's all it's for.

If CharacterBody _is_ meant to be used outside of just characters, and intended for anything where you want to have more control over the physics of an object, then I think, outside of a name change, it would be helpful to clarify in the documentation, something like "Despite the name, CharacterBody can be used outside of characters, whenever you want more control over physics objects." Maybe inside one of the little blue notes, or in the main description in the reference pages.

I just feel like it would really help if it was fully clarified. Of course, I only think it should say this if it was actually intended to be used for things other than characters. If there's a better way to implement a custom physics body, then it doesn't need to mention what I said above.

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/stable/tutorials/physics/physics_introduction.html#characterbody2d
https://docs.godotengine.org/en/stable/tutorials/physics/using_character_body_2d.html
https://docs.godotengine.org/en/stable/classes/class_characterbody2d.html
https://docs.godotengine.org/en/stable/classes/class_characterbody3d.html

tetrapod00:
In 3.x, this was called a [KinematicBody](https://docs.godotengine.org/en/3.6/classes/class_kinematicbody.html) (see also [here](https://docs.godotengine.org/en/3.6/tutorials/physics/kinematic_character_2d.html)), which was a more general name. I'm actually not sure why it was renamed to CharacterBody; happened before my time. A rename won't be considered at least for 4.x, and probably not even after that, since it was already changed once.

[...]

PR #10564, opened 2025-01-29
Clarify CharacterBody's purpose

Vroodot:
issue: https://github.com/godotengine/godot-docs/issues/10458 Added:
Despite the name, this class is meant to be used for more than just "characters." In this case, a character body is any physics body you want to control manually.


AThousandShips:

You should target `master` not `4.3`, changes can then be cherry picked, see [here](https://docs.godotengine.org/en/latest/contributing/documentation/contributing_to_the_documentation.html#contributing-changes).

PR #10565, opened 2025-01-29
Add Getting Started for Parallax2D

### LINKED ISSUE #10512: Parallax2d and "2d Parallax" tutorial pages do not explain how to set up the nodes
**Your Godot version:** 4.3

**Issue description:** 

I cant figure out how to set up the nodes for the parallax2d node. What nodes do I add under it? How do i set the scroll speed? Is every node supposed to have it's own Parallax2d or do they all go under one? 

There's no setup guide.

https://docs.godotengine.org/en/stable/classes/class_parallax2d.html

https://docs.godotengine.org/en/stable/tutorials/2d/2d_parallax.html

markdibarry:
Adds a short "Getting Started" section to clarify where to place textures. Also, updates image to show the example's node tree.

Closes #10512


tutorials/2d/2d_parallax.rst
```diff
@@ -16,17 +16,22 @@ up though, so this page provides in-depth descriptions of some properties and ho
     recommended to use over the :ref:`ParallaxLayer<class_parallaxlayer>` and
     :ref:`ParallaxBackground<class_parallaxbackground>` nodes.
 
-Scroll scale
-------------
+Getting started
+---------------
+
+The parallax node supports adding textures as child nodes, so you can use one or many textures to make up each layer.
+To begin, place each image or images you want to have scroll independently as a child of their own parallax node.
```
tetrapod00, line 23: "Texture" and "image" aren't node types. I'm still uncertain which kinds of nodes are able to be used here. Presumably Sprite2D is okay? Can I use this with a `TextureRect`, which is the closest thing that comes up when searching "texture", but is a control node which probably shouldn't work with Node2Ds very well?

Can Parallax2D be used with any kind of node that includes the `texture` property (this is how the `texture` property is read by shaders, for example)?

I'd recommend listing at least the most common specific node types which can be used. Ideally you can link directly to the class reference like:
```
:ref:`Sprite2D<class_sprite2d>`
```

tetrapod00 (changes requested):
Thanks for taking the time to add this, however I think this could still be a little more specific about exactly which nodes can be used

~~Also, for posterity's sake, in https://github.com/godotengine/godot-docs/pull/10462 we updated our dependency on [sphinxcontrib-video](https://github.com/sphinx-contrib/video), which added the ability to specify alignment for videos in https://github.com/sphinx-contrib/video/pull/40. However that changed the existing unspecified alignment behavior, so in this PR the alignment of the video is specified as `center` to match the prior behavior.~~ The video alignment is handled by https://github.com/godotengine/godot-docs/pull/10566 instead now

tutorials/2d/2d_parallax.rst
```diff
@@ -43,6 +48,7 @@ The video below displays how these values affect scrolling while in-game:
    :autoplay:
    :loop:
    :muted:
+   :align: center
```
tetrapod00, line 51: This bit is superseded by https://github.com/godotengine/godot-docs/pull/10566 now. I don't think `:align: center` is the right solution, but if it is we'll make it in the other PR as a batch.
```suggestion
```

tutorials/2d/2d_parallax.rst
```diff
@@ -16,17 +16,22 @@ up though, so this page provides in-depth descriptions of some properties and ho
...
 
-Scroll scale
-------------
+Getting started
+---------------
+
+The parallax node supports adding textures as child nodes, so you can use one or many textures to make up each layer.
+To begin, place each image or images you want to have scroll independently as a child of their own parallax node.
+Make sure that the top left of your images are at the ``(0,0)`` crossing, like the image below. See the section on
+:ref:`positioning<positioning>` for why this is important.
```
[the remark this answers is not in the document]
markdibarry: Ah didn't realize that would work. I wondered why the only other one referencing a section of the same page was https://docs.godotengine.org/en/stable/tutorials/shaders/converting_glsl_to_godot_shaders.html#glsl-coordinates

tutorials/2d/2d_parallax.rst
```diff
@@ -16,17 +16,25 @@ up though, so this page provides in-depth descriptions of some properties and ho
...
+
+The parallax node supports adding nodes that render things as children, so you can use one or many nodes to make up each
+layer. To begin, place each node or nodes you want to have scroll independently as a child of their own parallax node.
+Make sure that the top left of the textures used are at the ``(0, 0)`` crossing, like in the image below. See the section
+on :ref:`positioning <doc_2d_parallax_positioning>` for why this is important.
+
+.. image:: img/2d_parallax_size_viewport.webp
+
+The scene above uses one prepared texture for the higher clouds in a :ref:`Sprite2D <class_sprite2d>`, but you could
+just as easily use multiple nodes spaced out to compose the layer.
```
tetrapod00, line 30:
```suggestion
just as easily use multiple nodes spaced out to compose the layer.

Scroll scale
------------
```
Did you intend to remove the existing `Scroll scale` section header and replace it with the `Getting started` one?

It reads okay to me with or without it, so your call

tetrapod00 (approved):
Looks good to me, besides a possibly-forgotten section header

tutorials/2d/2d_parallax.rst
markdibarry: I removed it since it seemed too short to be its own section now, but if you think they should be separated I'm fine to change it.

markdibarry: Done!

skyace65 (approved):

skyace65:
Thanks!

PR #10569, opened 2025-01-29
Point mapping contributions to Godot in Controllers, gamepads and joysticks

Calinou:
[GameControllerDB's upstream](https://github.com/mdqinc/SDL_GameControllerDB) does not accept contributions from Godot since it does not use SDL. Users can still create custom mappings, but they should be submitted to Godot's controller database instead of the SDL one.

This also adds a reference to `godotcontrollerdb.txt` which was missing in the first place. This file is used as a complement to `gamecontrollerdb.txt`.

### DIFF
--- a/tutorials/inputs/controllers_gamepads_joysticks.rst
+++ b/tutorials/inputs/controllers_gamepads_joysticks.rst
@@ -3,9 +3,7 @@
 Controllers, gamepads, and joysticks
 ====================================
 
-Godot supports hundreds of controller models thanks to the community-sourced
-`SDL game controller database <https://github.com/gabomdq/SDL_GameControllerDB>`__.
-
+Godot supports hundreds of controller models out of the box.
 Controllers are supported on Windows, macOS, Linux, Android, iOS, and HTML5.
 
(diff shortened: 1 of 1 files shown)


AThousandShips (approved):
I think this looks good, I think not explicitly stating to not report it upstream or contribute there is safest, to not create any bad feelings

akien-mga:
I would remove any explicit reference to https://github.com/gabomdq/SDL_GameControllerDB/ as we shouldn't be sending users there, since they made it clear our use case is out of scope for them.

That's their prerogative and it's totally fair, we seem to have misinterpreted how much overlap there is between our needs for SDL-compatible mappings, and SDL's own need for a mapping DB for those not supported out of the box by its driver.

If we do find bugs in mappings needed for SDL itself, we can contribute them back, but this should be done by Godot maintainers after properly evaluating that they're in scope for SDL, and not by Godot users.

FYI @offalynne

tetrapod00 (approved):

akien-mga (approved):

tetrapod00:
Thanks, and thanks for the reviews!

PR #10571, opened 2025-01-29
Add a command snippet to build only modified documentation

Maarrk:
I imagine the main reason to use `FILELIST` is to only rebuild the files that are being worked on.

I don't usually work in ~~PowerShell~~ Windows but if you require a ~~PS~~ Windows version to accept the contribution, I can try to make it work.

There are some more nice ways to use it like `git diff --staged` or `git diff master` but I thought it would be too much here. But let me know if you'd actually like it included, I imagine it could help reviewers.

### DIFF
--- a/contributing/documentation/building_the_manual.rst
+++ b/contributing/documentation/building_the_manual.rst
@@ -176,7 +176,7 @@ You can specify a list of files to build, which can greatly speed up compilation
     make html FILELIST='classes/class_node.rst classes/class_resource.rst'
 
 The list of files can also be provided by the ``git`` command.
-This way you can automatically get names of all files that were changed since
+This way you can automatically get the names of all files that have changed since
 the last commit (``sed`` is used to put them on the same line).
 
 .. code:: sh
--- a/contributing/documentation/building_the_manual.rst
+++ b/contributing/documentation/building_the_manual.rst
@@ -162,8 +162,33 @@ RAM for Sphinx alone.
 Specifying a list of files
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+.. warning::
+
+    This section will not work on Windows, since the repository is using
+    a simplified ``make.bat`` script instead of the real GNU Make program.
+    If you would like to get a Linux terminal on your system, consider using
+    `Windows Subsystem for Linux (WSL) <https://learn.microsoft.com/en-us/windows/wsl/>`__.
+
 You can specify a list of files to build, which can greatly speed up compilation:
 
 .. code:: sh
 
     make html FILELIST='classes/class_node.rst classes/class_resource.rst'
+
+The list of files can also be provided by the ``git`` command.
+This way you can automatically get names of all files that were changed since
+the last commit (``sed`` is used to put them on the same line).
+
+.. code:: sh
+
+    make html FILELIST="$(git diff HEAD --name-only | sed -z 's/\n/ /g')"
+
+You can replace ``HEAD`` with ``master`` to return all files changed from the
+``master`` branch:
+
+.. code:: sh
+
+    make html FILELIST="$(git diff master --name-only | sed -z 's/\n/ /g')"
+
+If any images were modified, the output will contain some warnings about them,
+but the build will proceed correctly.


contributing/documentation/building_the_manual.rst
```diff
@@ -167,3 +167,10 @@ You can specify a list of files to build, which can greatly speed up compilation
 .. code:: sh
 
     make html FILELIST='classes/class_node.rst classes/class_resource.rst'
+
+You can also use command substitution, so ``git`` passes to ``make`` names of
+all files that were changed since the last commit (``sed`` adjusts formatting):
+
+.. code:: sh
+
+    FILELIST="\"$(git diff HEAD --name-only | sed -z 's/\n/\" \"/')\"" make html
```
skyace65, line 176: the ``make html`` part needs to be at the front, I'm also getting this error on windows in command prompt ``'sed' is not recognized as an internal or external command,`` It also gives me an error with ``sed`` in PowerShell

Also, not a issue specific to this PR but I'm not even sure if the ``FILELIST`` command actually works, I tried the above example and it just builds the entire docs

skyace65 (changes requested):

contributing/documentation/building_the_manual.rst
tetrapod00: `FILELIST` definitely works in general, I tested just now on Ubuntu via WSL.

Maarrk: I'm sorry but on Windows I'm unable to successfully specify the FILELIST in any way. Even not doing the git trickery, it outputs classes whatever I try in PowerShell and Command Prompt, using an environment variable, before html, after html etc.

It's already a wonder that something resembling `make` works at all in Windows. If more people reproduce @skyace65's results, maybe we should mark the whole `FILELIST` section as non-Windows? (I don't count myself because I know next to nothing about Windows commandline usage)

tetrapod00: AFAIK `set SPHINXOPTS=-j2 && make html` doesn't work on Windows either, but I can't track down the issue that confirms that. (It's why I switched to WSL for docs builds.)

We may want to suggest using WSL as a general purpose solution for users who frequently build the docs on windows.

skyace65: From what I'm seeing ``sed`` is only on Linux. I'm fine with adding this to the docs but we need to make sure it's clear that command will only work on Linux.

Maarrk: I know about `sed` only being a Linux thing. There is a counterpart on Windows, that's not the issue.

The question now is whether you're able to get the FILELIST to work on Windows even passing the filenames manually

Maarrk:
> Also, not a issue specific to this PR but I'm not even sure if the FILELIST command actually works, I tried the example above what you added in this PR and it just builds the entire docs

> I'm fine with adding this to the docs but we need to make sure it's clear that command will only work on Linux.

> We may want to suggest using WSL as a general purpose solution for users who frequently build the docs on windows.

I had a look at the `make.bat` script, and it doesn't handle `FILELIST` in any way. So I clarified that with a `.. warning` with WSL suggestion.

https://github.com/godotengine/godot-docs/blob/9abcbd707e35d4137cb70321b1e795794e7c9d0f/make.bat

I also rewrote the explanation and modified the snippet to hopefully make it a bit easier to follow

contributing/documentation/building_the_manual.rst
tetrapod00: I can confirm the inability to set FILELIST on windows; tested in both cmd and powershell and before and after the `make html` or `.\make html` (in powershell). I trust Maarrk's testing here. I also tried `set FILELIST='about.rst'` since that is recommended for setting SPHINXOPTS and that doesn't work either.

If only the basic `make html` command works on windows, I think it's good to add a warning about that and also suggest WSL if you're building often.

tetrapod00:
Oh, and to be clear I tested the commands on Ubuntu via WSL and they work well. Personally the `master` version is going to be the most useful for me, to test PRs, so I recommend adding that as a copyable snippet. 

Worth noting that it *can* pull in invalid files like images, but that gets filtered out (and writes warnings to the console). Also if the PR hasn't been recently rebased against master it can pull in some irrelevant files. But overall it should still be a lot faster than building the whole docs, and easier than adding in the relevant pages manually in some cases

contributing/documentation/building_the_manual.rst
```diff
@@ -162,8 +162,29 @@ RAM for Sphinx alone.
...
+
+.. code:: sh
+
+    make html FILELIST="$(git diff HEAD --name-only | sed -z 's/\n/ /g')"
+
+You can replace ``HEAD`` with ``master`` to return all files changed from the
+``master`` branch.
+
+.. code:: sh
+    make html FILELIST="$(git diff master --name-only | sed -z 's/\n/ /g')"
```
tetrapod00, line 190:
```suggestion
.. code:: sh

    make html FILELIST="$(git diff master --name-only | sed -z 's/\n/ /g')"
```
My bad, I gave you a bad RST suggestion

Maarrk: fixed

Maarrk:
> Worth noting that it can pull in invalid files like images, but that gets filtered out (and writes warnings to the console). 

@tetrapod00 can you have a look while you're here?

tetrapod00 (approved):
I think this looks good now! Seems like a useful addition.

contributing/documentation/building_the_manual.rst
```diff
@@ -162,8 +162,26 @@ RAM for Sphinx alone.
...
+    `Windows Subsystem for Linux (WSL) <https://learn.microsoft.com/en-us/windows/wsl/>`__.
+
 You can specify a list of files to build, which can greatly speed up compilation:
 
 .. code:: sh
 
     make html FILELIST='classes/class_node.rst classes/class_resource.rst'
+
+The list of files can also be read from ``FILELIST`` environment variable.
+This way you can automatically get names of all files that were changed since
```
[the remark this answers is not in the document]
tetrapod00:
```suggestion
This way you can automatically get the names of all files that have changed since
```

skyace65 (approved):

skyace65:
Thanks!

PR #10572, opened 2025-01-29
Add more Python-related files to gitignore

Maarrk:
I tend to name all my environments `venv`, but there's also a bunch of tooling assuming that. Makes the workflow more friendly because you don't see an explosion of files in git.

Based on https://www.toptal.com/developers/gitignore/api/python which is under CC0 license, took a moment to organize the file similarly

> Yo dawg, I heard you like `venv`, so we put more `venv` in your `.gitignore` so you can ignore `venv` while you `git`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,14 +3,28 @@
 .env
 
 _build/
-env/
-__pycache__
-*.pyc
 *~
 .directory
 .vs/
 .vscode/
 *.mo
+
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# Common environment files
+.env
+.venv
+env/
+venv/
+ENV/
+env.bak/
+venv.bak/
+
+# User created Python virtual environment as described in the docs
+godot-docs-venv/
 
 # Vim temp files
 *.swo
@@ -44,9 +58,6 @@ logo.h
 # Output of list-unused-images.sh tool
 tmp-unused-images
 tmp-unused-images-history
-
-# User created Python virtual environement as described in the docs
-godot-docs-venv/
 
 # Jetbrains IDE files
 /.idea/


tetrapod00 (approved):
The reasoning makes sense, and it makes sense to cover a wide range of common environment names since the docs are edited by a lot of contributors that we can't enforce a common workflow on.

I don't know enough python to review the details, though. Would appreciate other reviews.

For future reference the CC0 license is specified [here](https://github.com/toptal/gitignore.io?tab=readme-ov-file#license-of-the-generated-files) ([permalink](https://github.com/toptal/gitignore.io/blob/fd3331d935ac5029c59d4f4455c4bd1b29dcb3b1/README.md?plain=1#L29)); it's not specified in the linked file in the PR itself. But a gitignore likely falls under fair use anyway?

Maarrk:
> For future reference the CC0 license is specified [here](https://github.com/toptal/gitignore.io?tab=readme-ov-file#license-of-the-generated-files) ([permalink](https://github.com/toptal/gitignore.io/blob/fd3331d935ac5029c59d4f4455c4bd1b29dcb3b1/README.md?plain=1#L29)); it's not specified in the linked file in the PR itself.

Note that it's also in the commit message, but on this website you need to click on the three dots on the side to show more lines.

> But a gitignore likely falls under fair use anyway?

I am not a lawyer, but these few lines are too trivial to have any discussion of copyright over them.

I only added that remark to show that they aren't my idea but come from a widely used source, to make it easier to trust for maintainers. Here is the same content under the same license from GitHub itself:

https://github.com/github/gitignore/blob/297239c101dcfdfae7e75757ed17ed993df0b4eb/Python.gitignore#L130-L137

mhilbrunner (approved):

mhilbrunner:
As a Python dev, these make sense to me. Not sure I ever encountered the `.bak` ones, but whatever. :)

PR #10573, opened 2025-01-29
scripting_player_input.rst: fix small inaccuracy about conditionals

joepie91:
In its original version, the text implies that the colon is part of the conditional, but that is not correct. This fixes that.

### DIFF
--- a/getting_started/step_by_step/scripting_player_input.rst
+++ b/getting_started/step_by_step/scripting_player_input.rst
@@ -63,7 +63,8 @@ right, and ``-1`` means they want to turn left.
 
 To produce these values, we introduce conditions and the use of ``Input``. A
 condition starts with the ``if`` keyword in GDScript and ends with a colon. The
-condition is the expression between the keyword and the end of the line.
+condition is the expression between the keyword and the colon at the end of the
+line.
 
 To check if a key was pressed this frame, we call ``Input.is_action_pressed()``.
 The method takes a text string representing an input action and returns ``true``


getting_started/step_by_step/scripting_player_input.rst
```diff
@@ -63,7 +63,8 @@ right, and ``-1`` means they want to turn left.
 
 To produce these values, we introduce conditions and the use of ``Input``. A
 condition starts with the ``if`` keyword in GDScript and ends with a colon. The
-condition is the expression between the keyword and the end of the line.
+condition is the expression between the keyword and the colon at the end of the
+line.
```
tetrapod00, line 67: I suggest something like:
```
To produce these values, we introduce conditional statements and the use of
``Input``. A conditional statement starts with the ``if`` keyword in GDScript
and ends with a colon. The condition is specifically the expression between
the `if` keyword and the colon at the end of the line.
```

If we're going to clarify things this precisely, let's make sure it's right, and not use two seemingly distinct meanings for "condition" in two lines right next to each other.

This is a "conditional statement" or "conditional":
```
if Input.is_action_pressed("ui_left"):
```

This is a "conditional expression" or "condition":
```
Input.is_action_pressed("ui_left")
```

The [rest of the Godot docs](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_basics.html#if-else-elif) don't draw a super clear distinction between these two at all times, but we [do use the term "conditional statement"](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html#avoid-unnecessary-parentheses).

This does seem to be the general terminology (from https://en.wikipedia.org/wiki/Conditional_(computer_programming)):
> In the example code above, the part represented by (Boolean condition) constitutes a conditional expression, having intrinsic value (e.g., it may be substituted by either of the values True or False) but having no intrinsic meaning. In contrast, the combination of this expression, the If and Then surrounding it, and the consequent that follows afterward constitute a conditional statement, having intrinsic meaning (e.g., expressing a coherent logical rule) but no intrinsic value.

skyace65 (approved):
Applied Tetrapod's suggestions

skyace65:

PR #10580, opened 2025-01-30
Add notes about automatic `.app` bundle creation to compiling for macOS page.

bruvzg:
Adds build instructions and notes about automatic editor bundle generation (the same is already added for export templates).

### DIFF
--- a/contributing/development/compiling/compiling_for_macos.rst
+++ b/contributing/development/compiling/compiling_for_macos.rst
@@ -57,10 +57,6 @@ To compile for Apple Silicon (ARM64) powered Macs, use::
 
     scons platform=macos arch=arm64
 
-To support both architectures in a single "Universal 2" binary, run the above two commands and then use ``lipo`` to bundle them together::
-
-    lipo -create bin/godot.macos.editor.x86_64 bin/godot.macos.editor.arm64 -output bin/godot.macos.editor.universal
-
 .. tip::
     If you are compiling Godot to make changes or contribute to the engine,
     you may want to use the SCons options ``dev_build=yes`` or ``dev_mode=yes``.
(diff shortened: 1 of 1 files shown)


contributing/development/compiling/compiling_for_macos.rst
```diff
@@ -72,20 +68,39 @@ If all goes well, the resulting binary executable will be placed in the
 runs without any dependencies. Executing it will bring up the Project
 Manager.
 
+.. note:: Using a standalone editor executable is not recommended, it should be always packaged into
```
skyace65, line 71:
```suggestion
.. note:: Using a standalone editor executable is not recommended, it should be always packaged into a
```

skyace65 (changes requested):
One minor nitpick. Can't speak to technical accuracy but grammar and spelling for everything else looks good.

contributing/development/compiling/compiling_for_macos.rst
bruvzg: Done.

contributing/development/compiling/compiling_for_macos.rst
```diff
@@ -72,20 +68,39 @@ If all goes well, the resulting binary executable will be placed in the
 runs without any dependencies. Executing it will bring up the Project
 Manager.
 
+.. note:: Using a standalone editor executable is not recommended, it should be always packaged into a
```
AThousandShips, line 71:
```suggestion
.. note:: Using a standalone editor executable is not recommended, it should be always packaged into an
```
To match the rest (it can be read as either `dot-app` or `app`, the other cases assume the latter)

skyace65 (approved):
While I do think it should be read as "dot-app" and use "a" instead of "an" I'd rather merge this as is instead of holding this up for my nitpicking.

skyace65:
Thanks!

Issue #10604, opened 2025-02-02
INSTANCE_CUSTOM information regarding accesibility only in vertex shader.


DJLink:
**Your Godot version:4.3**

**[Issue description:]**
Today I spent some time trying to figure out why INSTANCE_CUSTOM was throwing errors until I bumped into this post explaining that INSTANCE_CUSTOM is only accessible in vertex shader and not fragment. I was searching docs but can't find it, so if it's not there I think there should be more information regarding this feature in shader docs or mesh instances.

https://github.com/godotengine/godot/issues/57584#issuecomment-2568180588

tetrapod00:
INSTANCE_CUSTOM is listed in the spatial shader [vertex() built-ins table](https://docs.godotengine.org/en/4.3/tutorials/shaders/shader_reference/spatial_shader.html#vertex-built-ins), and it's not listed in the spatial shader [fragment() built-ins table](https://docs.godotengine.org/en/4.3/tutorials/shaders/shader_reference/spatial_shader.html#fragment-built-ins) in the same page. It's the same situation in the [CanvasItem shader docs](https://docs.godotengine.org/en/4.3/tutorials/shaders/shader_reference/canvas_item_shader.html). So from that information I would not expect it to be available in `fragment()`.

IMO this is already at least *sufficiently* documented, but the fact that you're not the first person to run into problems suggests it could be improved. Do you have any suggestions for where else you would be looking for the information? I can think of a few possibilities:
- Is there a place in the docs where INSTANCE_CUSTOM is *used* in another tutorial that you think could use clarification?
- Perhaps the CanvasItem and Spatial shader reference pages should include a note about not all built-ins being available everywhere, and tell you about using varyings.
- On the engine side, it would be nice if there was an specific error message for using a builtin that is recognized but not available in one function or another. But that's a whole other issue that's beyond the scope of docs :)

DJLink:
@tetrapod00 Hi, thanks for checking this. I think my problem comes from trying to understand how MultiMeshInstance2D works, saw the set_instance_custom_data and moved from there "how can I get it on the shader side".

Because you are right, now that I see it, only vertex has it, but I was approaching my case from another situation and bumped into this.


-  Is there a place in the docs where INSTANCE_CUSTOM is used in another tutorial that you think could use clarification?

Considering set_instance_custom_data() I would probably put some quick mention there, followed by "please see shader docs for detailed information" because it's there. 

[...]

PR #10605, opened 2025-02-02
Add note to use varyings to access builtins in shader pages

tetrapod00:
Addresses https://github.com/godotengine/godot-docs/issues/10604.

In theory, the fact that shader builtins are different depending on the function is obvious and already noted in [Shading language](https://docs.godotengine.org/en/latest/tutorials/shaders/shader_reference/shading_language.html#built-in-variables). In practice, it's easy to miss that stub section and go straight to the Spatial or CanvasItem pages instead. So I think duplicating this note in two pages is appropriate.


mhilbrunner (approved):

mhilbrunner:
Agreed. :) Thanks, merged.

Issue #10503, opened 2025-01-13
Incorrect information on editing creating scenes and instances


Pezz-x:
**Your Godot version:**
4.3

**Issue description:**
Incorrect information on editing creating scenes and instances 
The guide shows the ball as a RigidBody2D when in the [instancing_starter.zip](https://github.com/godotengine/godot-docs-project-starters/releases/download/latest-4.x/instancing_starter.zip). the ball is a CollisionShape2D.

**URL to the documentation page:**
https://docs.godotengine.org/en/stable/getting_started/step_by_step/instancing.html#editing-scenes-and-instances

If you know how to fix the issue you are reporting please
consider opening a pull request. We provide a tutorial on
using git here: https://docs.godotengine.org/en/stable/community/contributing/pr_workflow.html,
writing documentation at https://docs.godotengine.org/en/stable/community/contributing/docs_writing_guidelines.html
and contributing to the class reference here: https://docs.godotengine.org/en/stable/community/contributing/updating_the_class_reference.html

AThousandShips:
The ball scene *is* a `RigidBody2D` though, please explain what is not correct here

AThousandShips:
That's not the node, you need to select the root of the scene called `Ball`

Pezz-x:
OOOHHHHHH I get it now!
Sorry, clearly a lapse in my memory from the nodes and scenes section.

I do think that maybe some rewording might be needed for the line:
_"Select the Ball node. In the Inspector on the right, click on the PhysicsMaterial property to expand it."_ 
Perhaps to:
_"Select the Ball node within the scene tab on the Left. Now in the Inspector on the right, click on the PhysicsMaterial property to expand it."_

No doubt I'm not the first idiot to just select the ball in the scene viewfinder.

Many thanks

PR #10608, opened 2025-02-03
Add clarifying text for instance docs

chellygel:
Added some clarifying text to help make the transition from double clicking Ball.tscn to editing the specific Ball node itself. Now it is clear to first click within the left nav before moving over to the inspector tab.

### DIFF
--- a/getting_started/step_by_step/instancing.rst
+++ b/getting_started/step_by_step/instancing.rst
@@ -121,7 +121,7 @@ Let's try this. Double-click ``ball.tscn`` in the FileSystem to open it.
 
 .. image:: img/instancing_ball_scene_open.webp
 
-Select the Ball node. In the Inspector on the right, click on the PhysicsMaterial
+In the Scene dock on the left, select the Ball node. Then, in the Inspector on the right, click on the PhysicsMaterial
 property to expand it.
 
 .. image:: img/instancing_physics_material_expand.webp


getting_started/step_by_step/instancing.rst
```diff
@@ -121,7 +121,7 @@ Let's try this. Double-click ``ball.tscn`` in the FileSystem to open it.
 
 .. image:: img/instancing_ball_scene_open.webp
 
-Select the Ball node. In the Inspector on the right, click on the PhysicsMaterial
+In the left hand nav, under the Scene tab, select  the Ball node. Then, in the Inspector on the right, click on the PhysicsMaterial
```
tetrapod00, line 124:
```suggestion
In the Scene dock, select the Ball node. Then, in the Inspector
on the right, click on the PhysicsMaterial
```
or
```suggestion
In the Scene dock on the left, select the Ball node. Then, in the
Inspector on the right, click on the PhysicsMaterial
```
Since the concept of the scene dock was introduced earlier in the page, I think we can shorten this a bit. Also, "left hand nav" also isn't an official term that we use anywhere. I would also split up the line so it doesn't exceed 80 characters.

tetrapod00:
Implements the suggestion in https://github.com/godotengine/godot-docs/issues/10503, so closes that issue.

getting_started/step_by_step/instancing.rst
chellygel: Yes I agree with your second suggestion, since before we use "on the right" it would bring the eye back to the correct left side. Will address it tomorrow if not taken care of before then. Great suggestion.

tetrapod00 (approved):

skyace65 (approved):

chellygel:
@tetrapod00 thanks again for the hand holding. New computer and never bothered to set. All sorted on my end, I believe. Sorry for the extra trouble. Hoping for easier PRs moving forward and less work for you 😅

tetrapod00:
I'm not sure what you did to change the email, but it looks like the commit email is still a different email than the one used by GitHub: `Chelsea Winfree authored and chellygel committed`, you can also check the [commit patch directly](https://github.com/godotengine/godot-docs/commit/fc1f052d3263dd904ef6fc95fcc481355672bbba.patch). 

You might need to change some settings on GitHub itself too to fix this? At this point it's really up to you if it's worth the effort to change things further. This looks fine to merge on our end.

chellygel:
@tetrapod00 go ahead and merge, i'll fix it on my own!

skyace65:

PR #10611, opened 2025-02-03
Fix video alignment after dependency update without CSS

tetrapod00:
Retry of https://github.com/godotengine/godot-docs/pull/10566, see that PR for details of the problem we're solving.

Tested on Firefox and Edge, checked videos and tables.

This time, use `:align: default` for each video, but don't add any custom CSS. This will align videos on the left without floating:
Screenshot 2025-02-03 112449
image
Since no CSS is changed, tables still render correctly:
Screenshot 2025-02-03 112312
Screenshot 2025-02-03 112408

Left alignment matches the behavior of 4.3 videos. While I think center alignment looks a bit better, I don't see a clear way to do it without introducing additional tags at every video usage, and left alignment looks fine and matches existing behavior.

**Why we can't use `:align: left`, as the plugin intends:**
It left-aligns with `float` due to our Sphinx theme, see the original PR for details.

**Why we can't use `:align: center` to center, as the plugin intends:**
It left-aligns due to our Sphinx theme.

**Why we can't use `:align: default` with custom CSS just for videos, to center:**
The video plugin places the video inside a div which only has the `align-default` class, so there's no way to specify `align-default` just for the videos. We could add a class with `:class:`, see [here](https://sphinxcontrib-video.readthedocs.io/en/latest/quickstart.html#options). But after reverting the first fix, I'm not sure trying to maintain custom CSS just to center videos is worth the maintenance, and this would also require the use of an extra tag at every video instance.

### DIFF
--- a/contributing/documentation/docs_image_guidelines.rst
+++ b/contributing/documentation/docs_image_guidelines.rst
@@ -242,6 +242,7 @@ videos should be included with the following code snippet::
        :autoplay:
        :loop:
        :muted:
+       :align: default
 
 Where ``documentation_video.webp`` would be changed to the name of the video you
 created. Name your videos in a way that makes their meaning clear, possibly with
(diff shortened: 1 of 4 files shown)


mhilbrunner (approved):

mhilbrunner:
Thanks!

PR #10618, opened 2025-02-05
Split GDExtension category

tetrapod00:
Alternate to https://github.com/godotengine/godot-docs/pull/10617. Reference implemtation of reorganization discussed in https://github.com/godotengine/godot-docs/pull/10617#issuecomment-2635333025. Try to keep general discussion of the reorganization in that PR, not this one

The text would need to be reworded to reflect the fact that it's in a single index page and not multiple pages.

Looks like this, just like the other implementation:
image

But unlike that implementation, it avoids creating extra directories, keeping everything in the base `gdextension` folder, which should avoid the need for extra redirects.

### DIFF
--- a/tutorials/scripting/gdextension/index.rst
+++ b/tutorials/scripting/gdextension/index.rst
@@ -8,7 +8,30 @@ GDExtension
    :name: toc-tutorials-gdnative
 
    what_is_gdextension
-   gdextension_cpp_example
-   gdextension_c_example
    gdextension_file
+
+C++ (godot-cpp)
+---------------
+
+This section describes how to get started with
+`godot-cpp <https://github.com/godotengine/godot-cpp>`__,
+the official C++ GDExtension bindings maintained as part of the Godot project.
+
+.. toctree::
+   :maxdepth: 1
+   :name: toc-tutorials-godot-cpp
+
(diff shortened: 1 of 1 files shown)


tutorials/scripting/gdextension/index.rst
```diff
@@ -8,7 +8,34 @@ GDExtension
...
+
+GDExtension Internals
+---------------------
+
+This section is aimed at people who want to make a new GDExtension from scratch.
+This is mainly people creating new language bindings.
+
+If you're looking to make a Godot extension with existing language bindings,
+check out the :ref:`C++ documentation (godot-cpp) <doc_gdextension_docs_system>`,
+or documentations for :ref:`other languages <doc_what_is_gdextension>` instead.
```
AThousandShips, line 35:
```suggestion
or documentation for :ref:`other languages <doc_what_is_gdextension>` instead.
```

tetrapod00: I scratched that whole paragraph since it doesn't really make sense if this is all on one index page.

tetrapod00:
Closing this one since we seem to be going ahead with https://github.com/godotengine/godot-docs/pull/10631

PR #9271, opened 2024-04-23
Fixed out of date binding syntax for `Signal.connect`

TheJP:
fixes godotengine/godot-docs#9236

PR #10625, opened 2025-02-05
Update binding Signal connection args docs

### LINKED ISSUE #8755: Signal documentation not updated for Godot 4 regarding passing additional variables with signal (binds array)
**Your Godot version:**
4.2.1

**Issue description:**
When adding a signal connection you can chose to add additional parameters to the signal call, beyond those emitted by the original signal. in Godot 3 you could apparently do this by adding the additional variables in a ["binds array" as an additional parameter in the connect()](https://forum.godotengine.org/t/get-the-node-that-trigger-a-signal/25576/2) call.  And this is what the current Godot 4 also describes.

Screenshot 2024-01-12 at 10 57 45

However, it seems that in Godot 4 this behavior has changed, and using the code above will give you an error as the second parameter of the connect call is now a set of connection flags as described here: https://docs.godotengine.org/en/4.2/classes/class_signal.html#class-signal-method-connect

Instead, the Godot 4 way of doing this is to use bind() to add the additional parameters you want to pass to the callable in the connect() call.

So the incorrect current example:
`character_node.health_changed.connect(battle_log_node._on_Character_health_changed, [character_node.name])`

Should be 
`character_node.health_changed.connect(battle_log_node._on_Character_health_changed.bind(character_node.name))`

Along with some changes in the text to describe this new way of doing things.

https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_basics.html#signals

Vocco:
Updated GDScript Reference `Signals` section to better reflect Godot 4's approach to binding additional arguments to connections. Intended to close #8755.

This was partially fixed with #9271, but the surronding text still referenced the binds array approach from earlier versions. This is my attempt to update that text.

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_basics.rst
+++ b/tutorials/scripting/gdscript/gdscript_basics.rst
@@ -2439,20 +2439,19 @@ arguments when you emit signals; it's up to you to emit the correct values.
 
 .. image:: img/gdscript_basics_signals_node_tab_1.png
 
-GDScript can bind an array of values to connections between a signal
-and a method. When the signal is emitted, the callback method receives
-the bound values. These bound arguments are unique to each connection,
-and the values will stay the same.
+You can also create copies of GDScript Callable objects which accept additional
+arguments using :ref:`Callable.bind() <class_Callable_method_bind>`. This
+allows you to add extra information to the connection if the emitted signal
+itself doesn't give you access to all the data that you need.
 
-You can use this array of values to add extra constant information to the
-connection if the emitted signal itself doesn't give you access to all the data
-that you need.
+When the signal is emitted, the callback method receives the bound values, in
+addition to those provided by the signal.
 
 Building on the example above, let's say we want to display a log of the damage
 taken by each character on the screen, like ``Player1 took 22 damage.``. The
 ``health_changed`` signal doesn't give us the name of the character that took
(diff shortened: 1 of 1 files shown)


tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -2439,20 +2439,20 @@ arguments when you emit signals; it's up to you to emit the correct values.
...
+arguments using :ref:`Callable.bind() <class_Callable_method_bind>`. This
+allows you to add extra information to the connection if the emitted signal
+itself doesn't give you access to all the data that you need.
 
-You can use this array of values to add extra constant information to the
-connection if the emitted signal itself doesn't give you access to all the data
-that you need.
+When the signal is emitted, the callback method receives the bound values, in
+addition to those provided by the signal. These bound arguments are unique to
+each connection, and the values will stay the same.
```
AThousandShips, line 2449: This might be confusing for data that is shared, like resources, which are shared even in binds, depending on the context, not sure if `Array`s are shared in this case or made unique

Vocco: Verified the behavior with resources and `Array`s, both are shared, not made unique and affected by side effects in the callbacks. Removed the sentence.

skyace65 (approved):

skyace65:
Thanks!

PR #10628, opened 2025-02-05
Remove shader materials page

skyace65:
I was going through the shader docs looking for anything that I'd be able to easily update and noticed that most of this page, at least screenshot wise is out of date. After going through the page itself I don't think it currently adds any value to the docs and only increases the maintenance burden.

My reasoning is that almost everything covered on this page is covered by the first shader and visual shader page. There's not really any unique information you need to know about shader materials themselves, they just exist to hold a shader. The only piece of information not on those pages is that you can convert other materials to shader materials. That's not done in this PR right now, I wanted to get second opinions on if it would be worth adding that to the first shader and visual shader pages, and will do so if requested.

@tetrapod00 @clayjohn what are your thoughts?


clayjohn (approved):
Ya, the text is partially out of date too. I think its fine to remove

tetrapod00 (approved):
Agree with the removal.

For the `Converting to ShaderMaterial` section: I would *prefer* if that ended up a reference page rather than a tutorial page, since the first shader pages are in theory optional if you already know shaders.

However I don't really see a better spot for it, so placing in one of the first shader pages is fine with me as long as it keeps a similar section title so it can be cleanly linked to from elsewhere.

skyace65:
@tetrapod00 what if I made a new page in the shading reference that's just for this? It would be a small page though.

tetrapod00:
Eh, that doesn't seem right either. I'd just put it in the first shader page and if we ever find a better place we can move or duplicate it. It's easier to extract a section out of a page than vice versa given the current redirect situation.

mhilbrunner (approved):

skyace65:
@tetrapod00 I've added a note to the first 3D shader and 2D shader page, let me know what you think.

tutorials/shaders/your_first_shader/your_first_2d_shader.rst
```diff
@@ -53,6 +53,11 @@ you are writing a CanvasItem Shader or a Spatial Shader and it previews the
 output of spatial shaders. So what you are seeing is the output of the default
 Spatial Shader.
 
+.. note::
+  Materials that inherit from the :ref:`class_Material` resource, such as :ref:`class_StandardMaterial3D`
+  and :ref:`class_ParticleProcessMaterial`, can be converted to a :ref:`class_ShaderMaterial`
+  and their existing properties will be converted to an accompanying text shader.
```
tetrapod00, line 59:
```suggestion
  and their existing properties will be converted to an accompanying text shader.
  To do so, right-click on the material in the FileSystem dock and choose
  **Convert to ShaderMaterial**. You can also do so by right-clicking on any
  property holding a reference to the material in the inspector.
```
Note that the filesystem access for this is new in 4.4.
I'm not sure if "any property holding a reference to the material in the inspector" is technically correct with Godot terminology but it's colloquially correct.

skyace65:
Added that info to both notes. Also added this PR to the 4.4 milestone since that's 4.4 functionality.

tetrapod00:
Thanks!

PR #10633, opened 2025-02-06
Update random_number_generation.rst to reference correct C# random int range method.

maxwell32:
On Godot 4.3 GD.RandiRange() is not an available method.

master branch has the correct documentation.

### DIFF
--- a/tutorials/math/random_number_generation.rst
+++ b/tutorials/math/random_number_generation.rst
@@ -179,7 +179,7 @@ and ``to``, and returns a random integer between ``from`` and ``to``:
  .. code-tab:: csharp
 
     // Prints a random integer between -10 and 10.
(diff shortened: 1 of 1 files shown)


tetrapod00:
This was changed in master/latest in https://github.com/godotengine/godot-docs/pull/8072, which is currently marked for cherrypicking. The `RandfRange` example immediately before this `RandiRange` example is also still incorrect on 4.3.
We should probably cherrypick the PR as a whole rather than fixing cases piecemeal.

skyace65:
While we appreciate the PR, I'm going to close this as the PR tetrapod linked to that will be cherrypicked is a more extensive update for the page, and merging this would cause a conflict with cherrypicking that one. I hope you consider contributing in the future if you find something else you want to improve.

Issue #4408, opened 2020-11-26
Document truthiness of various GDScript data types in one place


pycbouh:
**Your Godot version:**
Any.

**Issue description:**
As mentioned by @Calinou on Discord, we should probably provide a table of truthy/falsy characteristics for various data types of GDScript in one place. Currently this information is sparsely mentioned across the class reference and other docs, but an aggregated table can be pretty useful as a quick reference.

Question is where to put it though. GDScript basics, somewhere else, or maybe its own article?

**URL to the documentation page (if already existing):**

Animus-Surge:
Perhaps there could be another subsection under scripting/gdscript that contains the data values and their truth tables

NathanLovato:
If it's GDScript-specific, it should probably be in scripting/gdscript. If it's shared, it can be in the above directory, `scripting/`.

If there are only several types that are gdscript-specific but others are shared (dunno how vector or transform types work with C# for instance), I'd put it in `scripting/` with a heading dedicated to the specifics of GDScript.

copygirl:
Getting close to 5 years and I still can't find such a table anywhere.

The page for `bool` might not be the worst place to put such information. It's where I tried to look, first.

PR #10635, opened 2025-02-07
Add GDScript warning pages

### LINKED ISSUE #10627: Add documentation for GDScript warnings
**Your Godot version:** 4.4

**Issue description:**

Related to #368 .

There is currently no documentation for GDScript warnings in the Godot manual. I believe it may be helpful to create pages describing each warning in more detail, along with examples of code that causes them and an explanation of how to fix them.

Microsoft's official C# documentation has pages describing exceptions that can be thrown, with similar sections: https://learn.microsoft.com/en-us/dotnet/api/system.nullreferenceexception?view=net-9.0

For reference, here is my shot at writing a page for the `GET_NODE_DEFAULT_WITHOUT_ONREADY` warning:
> # `GET_NODE_DEFAULT_WITHOUT_ONREADY`
> The warning that appears when a node's default value is set to a location in the scene tree without using the `@onready` annotation.
> The message will say something like
> ```
> The default value is using "$" which won't return nodes in the scene tree before "_ready()" is called. Use the "@onready" annotation to solve this.
> ```
> ## When this warning occurs
> In GDScript, instance variables can be set by declaring them outside of a method. Additionally, they can be given a default value using `=`:
> ```gdscript
> extends Area2D
> var my_num = 3
> ```
> This way, the variable `my_num` will always start out with a value of `3` in each instance of this class.
> GDScript also has methods to retrieve specific nodes from the scene tree: namely, the `get_node` method, and its shorthand versions `$` (and `%` for unique nodes). Thus, if you want to have an instance variable default to a child of the node with a script, it may be tempting to write something like the following:
> ```gdscript
> extends Area2D
> var my_collision_shape = $CollisionShape2D
> ```
> However, class instance variables' default values are evaluated and assigned before the scene tree is set up. This means that at the time of assigning the default value, the node may not be in the scene tree, and the variable will be set to `null` instead.
> 
> ## How to fix this warning
> The most straightforward solution is to add the `@onready` annotation before your variable declaration:
> ```gdscript
> extends Area2D
> @onready var my_collision_shape = $CollisionShape2D
> ```
> Now, the default value of the variable will not be assigned until the scene tree has been initialized, and the target node will be present.
> 
> Alternatively, you can set the value of the variable at the beginning of your `_ready` method:
> ```gdscript
> extends Area2D
> var my_collision_shape
> 
> func _ready():
>     my_collision_shape = $CollisionShape2D
> ```

While I tried to keep everything accurate to the best of my knowledge, there might be some inaccuracies in this write-up. Of course, if this proposal were to move forward, we would want the content of the pages to be reviewed by multiple people before going live.
Hopefully, though, this example conveys the spirit of what I'm proposing.

The editor could potentially even provide a link to the corresponding page when a user encounters a particular warning, so that the user can read more about it.

Meorge:
Closes #10627.

This begins the implementation of pages describing each warning that can be given by GDScript. A page with all of the warnings can be found at `/tutorials/scripting/gdscript/warnings/index.html` or **Manual > Scripting > GDScript > GDScript warnings**. 

This initial PR adds pages for six warnings commonly encountered by beginners:
- GET_NODE_DEFAULT_WITHOUT_ONREADY
- INTEGER_DIVISION
- RETURN_VALUE_DISCARDED
- SHADOWED_VARIABLE
- UNASSIGNED_VARIABLE
- UNUSED_VARIABLE

If this PR is successful, small batches of additional warning pages can be added in subsequent PRs, so that they can be reviewed and verified before being published.

Each page has the following sections:
- An example of the warning message that appears in Godot, so that it can be easily found if someone pastes the message into a web search
- An explanation of what causes the warning to appear, including sample code
- Instructions on how to fix or suppress the warning

The target audience for the pages is relative newcomers to GDScript and/or programming in general; people who may be using code from tutorials without fully understanding it, and thus could be confused by the warning messages Godot provides. As such, the explanations aim to be extra beginner-friendly.

### DIFF
--- a/tutorials/scripting/gdscript/warnings/get_node_default_without_onready.rst
+++ b/tutorials/scripting/gdscript/warnings/get_node_default_without_onready.rst
@@ -1,0 +1,54 @@
+GET_NODE_DEFAULT_WITHOUT_ONREADY
+================================
+
+This warning appears when a node's property is initialized to a node in the scene tree
+without using the ``@onready`` annotation.
+
+Depending on how you attempt to access the scene tree, the warning message will
+be one of the following:
+
+.. code-block:: none
+
+    The default value uses "get_node()" which won't return nodes in the scene tree before "_ready()" is called. Use the "@onready" annotation to solve this.
+
+    The default value uses "$" which won't return nodes in the scene tree before "_ready()" is called. Use the "@onready" annotation to solve this.
+
+    The default value uses "%" which won't return nodes in the scene tree before "_ready()" is called. Use the "@onready" annotation to solve this.
+
+The default warning level for this warning is **Error**.
+To modify it, set :ref:`Project Settings > Debug > GDScript > Warnings > Get Node Default Without Onready<class_ProjectSettings_property_debug/gdscript/warnings/get_node_default_without_onready>`.
+
+When this warning occurs
+------------------------
+
+This warning may appear when attempting to assign the default value of a property
+to a node in the scene tree, like so:
+
+.. code-block:: gdscript
+
+    extends Area2D
+
+    # Will give warning GET_NODE_DEFAULT_WITHOUT_ONREADY.
+    var my_collision_shape = $CollisionShape2D
+
+If you want a property to refer to another node in the scene tree, you may be inclined
+to reference it using :ref:`get_node() <class_Node_method_get_node>`
+(or the ``$`` and ``%`` shorthand versions) when setting its default value.
+
+However, properties' default values are evaluated and assigned *before* the scene tree is set up.
+This means that when Godot tries to assign that value, the node you're referring to may not be in
+the scene tree yet, and the property will be set to ``null`` instead.
+
+How to fix this warning
+-----------------------
+
+Add the ``@onready`` annotation before your property declaration:
+
+.. code-block:: gdscript
+
+    extends Area2D
+
+    @onready var my_collision_shape = $CollisionShape2D
+
+Now, the default value of the property will not be assigned until the scene tree has been initialized,
+at which time the node will be present.
--- a/tutorials/scripting/gdscript/warnings/get_node_default_without_onready.rst
+++ b/tutorials/scripting/gdscript/warnings/get_node_default_without_onready.rst
@@ -1,0 +1,55 @@
+GET_NODE_DEFAULT_WITHOUT_ONREADY
+================================
+
+This warning appears when a node's property is initialized to a node in the scene tree
+without using the ``@onready`` annotation.
+
+Depending on how you attempt to access the scene tree, the warning message will
+be one of the following:
+
+.. code-block:: none
+
+    The default value uses "get_node()" which won't return nodes in the scene tree before "_ready()" is called. Use the "@onready" annotation to solve this.
+
+    The default value uses "$" which won't return nodes in the scene tree before "_ready()" is called. Use the "@onready" annotation to solve this.
+
+    The default value uses "%" which won't return nodes in the scene tree before "_ready()" is called. Use the "@onready" annotation to solve this.
+
+The default warning level for this warning is **Error**.
+To modify it, set :ref:`Project Settings > Debug > GDScript > Warnings > Get Node Default Without Onready<class_ProjectSettings_property_debug/gdscript/warnings/get_node_default_without_onready>`.
+
+When this warning occurs
+------------------------
+
+This warning may appear when attempting to assign the default value of a property
+to a node in the scene tree, like so:
+
+.. code-block:: gdscript
+
+    extends Area2D
+
+    # Will give warning GET_NODE_DEFAULT_WITHOUT_ONREADY.
+    var my_collision_shape = $CollisionShape2D
+
+If you want a property to refer to another node in the scene tree, you may be inclined
+to reference it using :ref:`get_node() <class_Node_method_get_node>`
+(or the ``$`` and ``%`` shorthand versions) when setting its default value.
+
+However, properties' default values are evaluated and assigned *before* nodes are added to the scene tree.
+This means that when the script tries to assign that value, it won't be able to find the node it's looking for,
+because it isn't a part of the scene tree containing that node yet.
+As a result, the property will be set to ``null`` instead.
+
+How to fix this warning
+-----------------------
+
+Add the ``@onready`` annotation before your property declaration:
+
+.. code-block:: gdscript
+
+    extends Area2D
+
+    @onready var my_collision_shape = $CollisionShape2D
+
+Now, the default value of the property will not be assigned until the scene tree has been initialized,
+at which time the node will be present.
--- a/tutorials/scripting/gdscript/warnings/index.rst
+++ b/tutorials/scripting/gdscript/warnings/index.rst
@@ -1,0 +1,27 @@
+:allow_comments: False
+
+.. _doc_gdscript_warnings:
+
+GDScript warnings
+=================
+
+This is a collection of warnings that GDScript can emit, and information on how to fix the code that causes them.
+
+.. note::
+
+   The list of warnings is currently incomplete. We hope to add pages for more warnings over time,
+   so that each one can be thoroughly reviewed for correctness and clarity.
+
+.. rubric:: Warnings
+   :heading-level: 2
+
+.. toctree::
+   :maxdepth: 1
+   :name: toc-gdscript-warnings
+
+   get_node_default_without_onready
+   integer_division
+   return_value_discarded
+   shadowed_variable
+   unassigned_variable
+   unused_variable
--- a/tutorials/scripting/gdscript/warnings/integer_division.rst
+++ b/tutorials/scripting/gdscript/warnings/integer_division.rst
@@ -1,0 +1,43 @@
+INTEGER_DIVISION
+================
+
+The warning message is:
+
+.. code-block:: none
+
+    Integer division. Decimal part will be discarded.
+
+The default warning level for this warning is **Warn**.
+To modify it, set :ref:`Project Settings > Debug > GDScript > Warnings > Integer Division<class_ProjectSettings_property_debug/gdscript/warnings/integer_division>`.
+
+When this warning occurs
+------------------------
+
+This warning may appear when attempting to divide two integers:
+
+.. code-block::
+
+    var result = 5 / 3  # Will give warning INTEGER_DIVISION.
+
+Because both operands are integers, the result will be an integer as well.
+Integers can't store fractional parts of numbers, so the result must be a whole number.
+Godot discards anything after the decimal point in the mathematical result to obtain the integer result.
+**Note that the number is not rounded to the nearest whole number.**
+
+
+How to fix this warning
+-----------------------
+
+Use a floating-point number (``float``) for at least one operand of the division operation:
+
+.. code-block::
+
+    var result = 5.0 / 3
+
+If the integers being divided are variables, cast them to ``float``:
+
+.. code-block::
+
+    var a: int = 5
+    var b: int = 3
+    var result = float(a) / float(b)
--- a/tutorials/scripting/gdscript/warnings/return_value_discarded.rst
+++ b/tutorials/scripting/gdscript/warnings/return_value_discarded.rst
@@ -1,0 +1,42 @@
+RETURN_VALUE_DISCARDED
+======================
+
+The warning message is:
+
+.. code-block:: none
+
+    The function "get_number()" returns a value that will be discarded if not used.
+
+The default warning level for this warning is **Ignore**.
+To modify it, set :ref:`Project Settings > Debug > GDScript > Warnings > Return Value Discarded<class_ProjectSettings_property_debug/gdscript/warnings/return_value_discarded>`.
+
+When this warning occurs
+------------------------
+
+This warning may appear if a method returns a value, but that value is not used
+in an expression or assigned to a variable:
+
+.. code-block::
+
+    func _ready():
+        print("About to get a number...")
+        get_number()  # Will give warning RETURN_VALUE_DISCARDED.
+        print("Got a number!")
+
+    func get_number() -> int:
+        return 5
+
+How to fix this warning
+-----------------------
+
+Assign the returned value to a variable for use later.
+
+.. code-block::
+    
+    func _ready():
+        print("About to get a number...")
+        var num = get_number()
+        print("Got a number! It's %s" % num)
+
+However, some methods in Godot's APIs return values that are not necessary to store.
+As such, depending on the situation, it may make more sense to ignore this warning.
--- a/tutorials/scripting/gdscript/warnings/return_value_discarded.rst
+++ b/tutorials/scripting/gdscript/warnings/return_value_discarded.rst
@@ -1,0 +1,42 @@
+RETURN_VALUE_DISCARDED
+======================
+
+The warning message is:
+
+.. code-block:: none
+
+    The function "get_number()" returns a value that will be discarded if not used.
+
+The default warning level for this warning is **Ignore**.
+To modify it, set :ref:`Project Settings > Debug > GDScript > Warnings > Return Value Discarded<class_ProjectSettings_property_debug/gdscript/warnings/return_value_discarded>`.
+
+When this warning occurs
+------------------------
+
+This warning may appear if a method returns a value, but that value is not used
+in an expression or assigned to a variable:
+
+.. code-block::
+
+    func _ready():
+        print("About to get a number...")
+        get_number()  # Will give warning RETURN_VALUE_DISCARDED.
+        print("Got a number!")
+
+    func get_number() -> int:
+        return 5
+
+How to fix this warning
+-----------------------
+
+Assign the returned value to a variable for use later.
+
+.. code-block::
+
+    func _ready():
+        print("About to get a number...")
+        var num = get_number()
+        print("Got a number! It's %s" % num)
+
+However, some methods in Godot's APIs return values that are not necessary to store.
+As such, depending on the situation, it may make more sense to ignore this warning.
--- a/tutorials/scripting/gdscript/warnings/shadowed_variable.rst
+++ b/tutorials/scripting/gdscript/warnings/shadowed_variable.rst
@@ -1,0 +1,40 @@
+SHADOWED_VARIABLE
+=================
+
+The warning message is:
+
+.. code-block:: none
+
+    The local variable "level" is shadowing an already-declared variable at line 3 in the current class.
+
+The default warning level for this warning is **Warn**.
+To modify it, set :ref:`Project Settings > Debug > GDScript > Warnings > Shadowed Variable<class_ProjectSettings_property_debug/gdscript/warnings/shadowed_variable>`.
+
+When this warning occurs
+------------------------
+
+This warning may appear when giving something the same name as a variable previously
+defined in the class.
+
+.. code-block::
+
+    extends Node
+
+    var level = 3
+
+    func _ready():
+        # Will give warning SHADOWED_VARIABLE.
+        var level = 1
+        print("Time for level %s" % level)
+
+In this example, the script class has a property ``level`` which can be accessed from its functions.
+However, at the first line of ``_ready()``, a new ``level`` variable is declared for that function specifically.
+After this declaration, any references to ``level`` will go to that version and not the shared variable for the class.
+This is called *shadowing*.
+
+
+How to fix this warning
+-----------------------
+
+Change the name to something that isn't being used by the class.
+For example, if receiving a warning about using the identifier ``level``, consider using something more descriptive like ``new_level``.
--- a/tutorials/scripting/gdscript/warnings/unassigned_variable.rst
+++ b/tutorials/scripting/gdscript/warnings/unassigned_variable.rst
@@ -1,0 +1,35 @@
+UNASSIGNED_VARIABLE
+===================
+
+The warning message is:
+
+.. code-block:: none
+
+    The variable "my_var" is used before being assigned a value.
+
+The default warning level for this warning is **Warn**.
+To modify it, set :ref:`Project Settings > Debug > GDScript > Warnings > Unassigned Variable<class_ProjectSettings_property_debug/gdscript/warnings/unassigned_variable>`.
+
+When this warning occurs
+------------------------
+
+This warning may appear when attempting to use a variable that hasn't had a value
+assigned to it yet.
+
+.. code-block::
+
+    var my_var
+    print(my_var)
+
+By default, the variable will be ``null``. However, Godot considers this a warning
+because the user did not explicitly assign the value, and as such might be unaware of it.
+
+How to fix this warning
+-----------------------
+
+Assign a value to the variable before including it in an expression or function call:
+
+.. code-block::
+
+    var my_var = 5
+    print(my_var)
--- a/tutorials/scripting/gdscript/warnings/unused_variable.rst
+++ b/tutorials/scripting/gdscript/warnings/unused_variable.rst
@@ -1,0 +1,52 @@
+UNUSED_VARIABLE
+===============
+
+The warning message is:
+
+.. code-block:: none
+
+    The local variable "counter" is declared but never used in the block. If this is intended, prefix it with an underscore: "_counter".
+
+The default warning level for this warning is **Warn**.
+To modify it, set :ref:`Project Settings > Debug > GDScript > Warnings > Unused Variable<class_ProjectSettings_property_debug/gdscript/warnings/unused_variable>`.
+
+When this warning occurs
+------------------------
+
+This warning may appear when a variable is declared, but never used before its
+function or scope ends:
+
+.. code-block::
+
+    extends CharacterBody2D
+
+    func _process(delta):
+        var player_speed = 5.0  # Will give warning UNUSED_VARIABLE.
+        velocity = Input.get_vector("ui_left", "ui_right", "ui_up", "ui_down")
+        move_and_slide()
+
+In this example, the variable ``player_speed`` is presumably meant to control how
+fast the player moves. However, ``velocity`` never actually takes into account
+this speed variable; it only uses the result of ``Input.get_vector()``.
+It's likely that the programmer meant to include ``player_speed`` here somehow but forgot to write it.
+
+How to fix this warning
+-----------------------
+
+If a variable is being marked as unused, you probably meant to use it somewhere
+but forgot to include it in the relevant statement. Following the example above,
+the likely solution would be to incorporate it into the calculation of ``velocity``:
+
+.. code-block::
+
+    velocity = player_speed * Input.get_vector("ui_left", "ui_right", "ui_up", "ui_down")
+
+If you're certain you don't need the variable for anything right now, but might
+need its value later on, prefix its name with an underscore (``_``) as the warning text suggests.
+Once you start referencing it elsewhere in the code, you can remove the underscore.
+
+.. code-block::
+
+    var _player_speed = 5.0  # Will not give warning UNUSED_VARIABLE.
+
+If you know you won't ever need the variable, then simply delete it.
--- a/tutorials/scripting/gdscript/gdscript_basics.rst
+++ b/tutorials/scripting/gdscript/gdscript_basics.rst
@@ -1592,6 +1592,9 @@ Lambda functions capture the local environment:
         lambda.call()
         print(a) # Prints `[1]`.
 
+
+.. _doc_gdscript_basics_static_functions:
+
 Static functions
 ~~~~~~~~~~~~~~~~
 
--- a/tutorials/scripting/gdscript/gdscript_basics.rst
+++ b/tutorials/scripting/gdscript/gdscript_basics.rst
@@ -1595,6 +1595,9 @@ Lambda functions capture the local environment:
         lambda.call()
         print(a) # Prints `[1]`.
 
+
+.. _doc_gdscript_basics_static_functions:
+
 Static functions
 ~~~~~~~~~~~~~~~~
 
--- a/tutorials/scripting/gdscript/index.rst
+++ b/tutorials/scripting/gdscript/index.rst
@@ -16,4 +16,5 @@ GDScript
    gdscript_styleguide
    static_typing
    warning_system
+   warnings/index
    gdscript_format_string


tutorials/scripting/gdscript/warnings/GET_NODE_DEFAULT_WITHOUT_ONREADY.rst
```diff
@@ -0,0 +1,41 @@
+``GET_NODE_DEFAULT_WITHOUT_ONREADY``
+====================================
+
```
tetrapod00, line 3:
```suggestion

Desccription
------------

```
Perhaps?

tutorials/scripting/gdscript/warnings/GET_NODE_DEFAULT_WITHOUT_ONREADY.rst
```diff
@@ -0,0 +1,41 @@
...
+The warning that appears when a node's default value is set to a location in the scene tree without using the ``@onready`` annotation.
+The message will say something like:
+
+.. code-block:: none
```
tetrapod00, line 7: It's unfortunate that the error here is wider than the screen and requires scrolling with a code block. If there was a way to use word-wrapped code blocks we should do that. Otherwise we could consider using indented and maybe italicized or codestyle text, which is used elsewhere in the manual for quotes.

This may also be solvable with some custom CSS but that should be avoided if possible.

I'll look around and see if there's any better way here.

tutorials/scripting/gdscript/warnings/GET_NODE_DEFAULT_WITHOUT_ONREADY.rst
```diff
@@ -0,0 +1,41 @@
...
+
+    The default value is using "$" which won't return nodes in the scene tree before "_ready()" is called. Use the "@onready" annotation to solve this.
+
+When this warning occurs
+------------------------
+In GDScript, instance variables can be set by declaring them outside of a method. Additionally, they can be given a default value using ``=``::
+
+    extends Area2D
```
tetrapod00, line 15:
```suggestion
```
I'm not sure we *need* the `extends Area2D` here for the example to work, but maybe it increases clarity to keep in?

tutorials/scripting/gdscript/warnings/GET_NODE_DEFAULT_WITHOUT_ONREADY.rst
```diff
@@ -0,0 +1,41 @@
+``GET_NODE_DEFAULT_WITHOUT_ONREADY``
+====================================
+
+The warning that appears when a node's default value is set to a location in the scene tree without using the ``@onready`` annotation.
```
tetrapod00, line 4:
```suggestion
This warning appears when a node's default value is set to a location in the scene tree without using the ``@onready`` annotation.
```
Personal preference, the existing form is fine too.

tutorials/scripting/gdscript/warnings/GET_NODE_DEFAULT_WITHOUT_ONREADY.rst
```diff
@@ -0,0 +1,41 @@
...
+This way, the variable ``my_num`` will always start out with a value of ``3`` in each instance of this class.
+GDScript also has methods to retrieve specific nodes from the scene tree: namely, the ``get_node`` method, and its shorthand versions ``$`` (and ``%`` for unique nodes). Thus, if you want to have an instance variable default to a child of the node with a script, it may be tempting to write something like the following::
+
+    extends Area2D
+    var my_collision_shape = $CollisionShape2D
+
+However, class instance variables' default values are evaluated and assigned before the scene tree is set up. This means that at the time of assigning the default value, the node may not be in the scene tree, and the variable will be set to ``null`` instead.
+
+How to fix this warning
+-----------------------
```
tetrapod00, line 27:
```suggestion
-----------------------

```
Usually we put an empty line after headers

tutorials/scripting/gdscript/warnings/index.rst
```diff
@@ -0,0 +1,63 @@
...
+This is a collection of the warnings that GDScript can emit, and information on how to fix the code that causes them.
+
+.. rubric:: Warnings
+   :heading-level: 2
+
+.. toctree::
+   :maxdepth: 1
+   :name: toc-gdscript-warnings
+
+   UNASSIGNED_VARIABLE
```
tetrapod00, line 17: If there's a natural way to break up this very long toctree into multiple smaller lists that would be good. Does the engine categorize these warnings in any way already?

To be clear you would want to make a few new toctrees each like this:

```
.. rubric:: Category of warnings
   :heading-level: 2

.. toctree::
   :maxdepth: 1
   :name: toc-gdscript-warnings-category

   UNASSIGNED_VARIABLE_OP_ASSIGN
   UNUSED_VARIABLE
   UNUSED_LOCAL_CONSTANT

```

They should still appear in one single long list in the sidebar of the site.

tutorials/scripting/gdscript/warnings/GET_NODE_DEFAULT_WITHOUT_ONREADY.rst
```diff
@@ -0,0 +1,41 @@
+``GET_NODE_DEFAULT_WITHOUT_ONREADY``
```
tetrapod00, line 1:
```suggestion
GET_NODE_DEFAULT_WITHOUT_ONREADY
```
We may not want to use codestyle within the title, since it makes it a bit unclear that these are clickable links:
Screenshot 2025-02-06 195145

tetrapod00 (commented):
I have no complaints with the overall structure of the warning page. The example you've written already seems quite high quality, and you've matched the tone and style of the manual quite well already. I at least would be happy merging this more or less as-is.

The content reads as overall correct but I'm not a deep expert in GDScript so that will need another review.

I left several suggestions for style and minor structure changes, but I'm not certain about some of them. Feel free to disagree and push back on any of them.

For the manual, unlike the class reference, we currently wrap lines to between 80 and 100 characters, as described [here](https://docs.godotengine.org/en/latest/contributing/documentation/docs_writing_guidelines.html#manually-wrapping-lines). There are plugins that can do this automatically in your IDE too.

tetrapod00:
These warnings are also documented (rather briefly) in the project settings, for example https://docs.godotengine.org/en/latest/classes/class_projectsettings.html#class-projectsettings-property-debug-gdscript-warnings-get-node-default-without-onready. We may want to link from each new page to the projectsettings class reference too.

tutorials/scripting/gdscript/warnings/GET_NODE_DEFAULT_WITHOUT_ONREADY.rst
Meorge: Overall, I agree - it's not used in this example script, so it *could* be removed if need be. The main reason I kept it is so that it implies we're looking at an *entire* script, rather than just an isolated snippet.

In the later code blocks, in addition to serving that purpose, it also provides a practical reason for attempting to assign a node type variable (saving the CollisionShape2D of an Area2D seems like a likely thing to want to do).

tutorials/scripting/gdscript/warnings/GET_NODE_DEFAULT_WITHOUT_ONREADY.rst
Meorge: What about manually inserting a line break?

tutorials/scripting/gdscript/warnings/GET_NODE_DEFAULT_WITHOUT_ONREADY.rst
Meorge: I agree; I was rather bummed to see how not-link-like they were once the code style was applied. Since they are code constant/enum names, I thought it'd make sense to give them the code style, but considering this issue, and the fact that in many other places code style isn't used for code names (like function and property names), I'll remove it.

tutorials/scripting/gdscript/warnings/GET_NODE_DEFAULT_WITHOUT_ONREADY.rst
tetrapod00: That might work, if there is a natural sentence break. We definitely should not manually wrap these to 80 characters or similiar.

In the engine these strings are on a single line, though, and you have to scroll horizontally there too. So maybe it's fine as-is.

tutorials/scripting/gdscript/warnings/GET_NODE_DEFAULT_WITHOUT_ONREADY.rst
```diff
@@ -0,0 +1,41 @@
...
+
+When this warning occurs
+------------------------
+In GDScript, instance variables can be set by declaring them outside of a method. Additionally, they can be given a default value using ``=``::
+
+    extends Area2D
+    var my_num = 3
+
+This way, the variable ``my_num`` will always start out with a value of ``3`` in each instance of this class.
+GDScript also has methods to retrieve specific nodes from the scene tree: namely, the ``get_node`` method, and its shorthand versions ``$`` (and ``%`` for unique nodes). Thus, if you want to have an instance variable default to a child of the node with a script, it may be tempting to write something like the following::
```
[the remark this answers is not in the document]
Meorge: Personally, I do like having the parentheses to distinguish methods/functions from properties, so I'll add this in. 

Perhaps we should open another issue for it (if it doesn't exist already) but the use of parentheses before method names seems to be inconsistent - specifically, I noticed that in the [Tween documentation](https://docs.godotengine.org/en/stable/classes/class_tween.html) there are lots of references to methods that don't include the parentheses.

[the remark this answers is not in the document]
tetrapod00: Yeah, let's split this out into another issue(s).

In the engine, the class reference includes `()` in links to methods. In the online class reference, `()` is not included. In the manual (that is, the rest of the docs, all the stuff with the source of truth in this repo), it's currently mixed whether we include `()` or not in links and mentions of functions.

Including `()` is definitely correct, and it's *overall* currently favored, so there are two issues here:
- We should change the RST generation of the class reference to include `()` for the online class reference.
- We should write down the standard that we include `()` in the manual - I didn't specify anything when I wrote [the guideline](https://docs.godotengine.org/en/latest/contributing/documentation/docs_writing_guidelines.html#classes-properties-and-methods).

I'll write the issues for these.

tutorials/scripting/gdscript/warnings/index.rst
Meorge: As far as I know, there isn't any existing categorization for the warnings. This is how they appear in the Project Settings:

CleanShot 2025-02-07 at 12 49 05@2x

That being said, I do like the idea of breaking them up into categories, and it seems the order has been set up such that it wouldn't be too difficult.

tetrapod00: If there is no existing categorization, I'd suggest either leaving them in the exact order they appear in the project settings, or alphabetizing them.

We can probably defer splitting into categories for a future update. It's possible that we would want the categorization to be driven by the engine side of things rather than the docs side.

Meorge: Sounds good!

Meorge:
I like the idea of providing links to the class reference. Additionally, we could use the descriptions already written in the class reference as the short one-line descriptions for the new pages, before further sections go into more detail and provide examples.

tetrapod00:
Using the existing short descriptions as placeholders makes sense

Meorge:
I haven't added the short descriptions as placeholders, but I did paste in the warning message strings for each warning 😅 This way, if someone searches the warning message they receive, they'd hopefully find the associated page in the docs.

A big issue with this is that messages that have details filled in at runtime are tricky to include. For now, I've just kept them verbatim from the source code with `%s` where text is filled in.

dalexeev:
It would also be nice to mention the default level (Ignore, Warn, Error), in which cases it is recommended or not recommended to enable/disable/suppress the warning. For the warning message, we could use human-readable placeholders instead of `%s`.

Can we use snake_case or kebab-case for file names, while UPPER_CASE for warning names and TOC items?

tetrapod00:
> Can we use snake_case or kebab-case for file names, while UPPER_CASE for warning names and TOC items?
For RST filenames, snake case is currently idiomatic

Meorge:
I can certainly change the filenames to use snake case, and replace the instances of `%s` with something more natural.
Adding the default warning level also makes sense to me - I'll try to add these things to my prototype pages soon.

tutorials/scripting/gdscript/warnings/ASSERT_ALWAYS_FALSE.rst
```diff
@@ -10,27 +10,39 @@ The warning message is:
...
+        # Despite false being passed, this won't raise ASSERT_ALWAYS_FALSE.
+        assert(false, "False is false")
+
+        # This evaluates to a boolean which is false, so it also won't raise
+        # the warning.
+        assert(3 == 4, "3 isn't equal to 4")
+
+    This is because ``assert(false)`` calls are often used in development to forcibly halt program execution and avoid strange errors later on.
+
+    See `issue #58087 <https://github.com/godotengine/godot/issues/58087>`_ for more information.
```
tetrapod00, line 40:
```suggestion
    See `GH-58087 <https://github.com/godotengine/godot/issues/58087>`_ for more information.
```
Usually it's styled like this.

Linking to the original issue/PR from the docs seems like an antipattern though, usually we would rewrite the relevant content of the issue/PR into a form suitable for the manual. I'm not wholly against linking though

tutorials/scripting/gdscript/warnings/ASSERT_ALWAYS_FALSE.rst
```diff
@@ -10,27 +10,39 @@ The warning message is:
 When this warning occurs
 ------------------------
 
-The :ref:`assert() <class_@GDScript_method_assert>` keyword can be used to ensure that a given condition is met before allowing code execution to continue. If the first argument passed to it evaluates to ``true``, the rest of the function will run as expected; if it is ``false``, then the project will stop.
+The :ref:`assert() <class_@GDScript_method_assert>` keyword can be used to ensure that a given condition is met before allowing code execution to continue. If the first argument passed to it is truthy, the rest of the function will run as expected; if it is falsy, then the project will stop.
```
tetrapod00, line 13: It would be nice if we defined truthy/falsy somewhere else in the GDScript docs that we could link to, see also https://github.com/godotengine/godot-docs/issues/4408

Meorge: I agree. In general I'm nervous about even using the terms "truthy" and "falsy" right now, because neither of them currently appear in the docs at all from what I can see (searching either of them on the online docs turns up no results).

tutorials/scripting/gdscript/warnings/ASSERT_ALWAYS_FALSE.rst
Meorge: I like the idea of linking the issue/PR, similar to how one might paraphrase a primary source in an essay and then provide a citation to the primary source itself, so the reader can look at it themselves if they so wish. But the reasoning provided in the issue might be general enough that it doesn't "need" a citation, and it could also make it unclear when it's appropriate to link to GitHub, resulting in the docs becoming more cluttered.

tutorials/scripting/gdscript/warnings/GET_NODE_DEFAULT_WITHOUT_ONREADY.rst
Mickeon: I feel like this is not enough. You could make more obvious that it's an instance variable (basically a property) by declaring a method below that uses it.

tutorials/scripting/gdscript/warnings/ASSERT_ALWAYS_FALSE.rst
Mickeon: I wanted to do this in https://github.com/godotengine/godot/pull/95487 but I haven't had the time to retouch it considerably.

Meorge: Thanks, I may use that wording instead to remain consistent!

Meorge:
I'm opening this PR for review now. My goal is to have at least a first version complete for all pages before merging, so that users don't go to look up a warning and just find "TODO". There's definitely a lot of work in writing the pages out, so perhaps that is too much to hope for. Regardless, it'll be very nice to have people reading and reviewing what's been written so far.

If there's a way to allow other people to contribute pages, that would be great too. I suppose they'd submit a PR on my fork?

tutorials/scripting/gdscript/warnings/standalone_expression.rst
```diff
@@ -0,0 +1,51 @@
...
+
+    Standalone expression (the line may have no effect).
+
+The default warning level for this warning is **Warn**.
+To modify it, see :ref:`ProjectSettings.debug/gdscript/warnings/standalone_expression<class_ProjectSettings_property_debug/gdscript/warnings/standalone_expression>`.
+
+When this warning occurs
+------------------------
+
+This warning may appear when writing an expression (a combination of values) that aren't assigned to anything or passed to a function:
```
Mickeon, line 16: Not necessarily true. `+++++1` is also an expression.
```suggestion
This warning may appear when writing an expression that is not assigned to anything or passed to a function:
```

tutorials/scripting/gdscript/warnings/standalone_expression.rst
```diff
@@ -0,0 +1,51 @@
...
+
+    func _ready():
+        # Will give warning STANDALONE_EXPRESSION.
+        (5 * 3) + 2 * add_to_val(3)
+
+    func add_to_val(val) -> int:
+        counter += 1
+        return val + 2
+
+Here, the line in ``_ready()`` receives the ``STANDALONE_EXPRESSION`` warning, even though the call to ``add_to_val()`` changes the ``counter`` property. (The other math calculations still won't have any effect, though.)
```
Mickeon, line 41: "still" and "though" would mean the same thing here. *Though*, we could do without them, I think
```suggestion
Here, the line in ``_ready()`` receives the ``STANDALONE_EXPRESSION`` warning, even though the call to ``add_to_val()`` changes the ``counter`` property. The other math calculations won't have any effect.
```

tutorials/scripting/gdscript/warnings/standalone_expression.rst
```diff
@@ -0,0 +1,51 @@
...
+
+How to fix this warning
+-----------------------
+
+If you've written an expression in your code, it's likely because you intended to use it for something. Make sure you're assigning it to a variable or passing it to a function if so.
```
Mickeon, line 46: I first replaced that "if so" with "if that's the case", but then I realised it would basically repeat the same prior sentence.
```suggestion
If you've written an expression in your code, it's likely because you intended to use it for something. Make sure you're assigning it to a variable or passing it to a function.
```

tutorials/scripting/gdscript/warnings/static_called_on_instance.rst
```diff
@@ -0,0 +1,45 @@
...
+        static func subtract_two(val):
+            return val - 2
+
+    func _ready():
+        var my_math_funcs = MathFuncs.new()
+
+        # Will give warning STATIC_CALLED_ON_INSTANCE.
+        var result = my_math_funcs.subtract_two(5)
+
+When a function is *static*, it belongs to the class as a whole, not any one specific instance of the class. This can be quite handy for writing functions that you want to access from anywhere, without needing a class instance.
```
Mickeon, line 33: It's true, it is handy, but this doesn't explain why this warning exists. Most warnings exist because the code may not be working as intended, but this one is mainly to avoid confusion when reading.

tutorials/scripting/gdscript/warnings/unassigned_variable.rst
```diff
@@ -0,0 +1,36 @@
...
+
+    The variable "my_var" is used before being assigned a value.
+
+The default warning level for this warning is **Warn**.
+To modify it, see :ref:`ProjectSettings.debug/gdscript/warnings/unassigned_variable<class_ProjectSettings_property_debug/gdscript/warnings/unassigned_variable>`.
+
+When this warning occurs
+------------------------
+
+This warning may appear when attempting to use a variable that hasn't had a value assigned to it yet.
```
Mickeon, line 16: May be worth mentioning that in Godot an unassigned variable is essentially `null`

tutorials/scripting/gdscript/warnings/unreachable_code.rst
```diff
@@ -0,0 +1,62 @@
...
+
+    Unreachable code (statement after return) in function "_ready()".
+
+The default warning level for this warning is **Warn**.
+To modify it, see :ref:`ProjectSettings.debug/gdscript/warnings/unreachable_code<class_ProjectSettings_property_debug/gdscript/warnings/unreachable_code>`.
+
+When this warning occurs
+------------------------
+
+This warning may appear if there is code in a function that can never be reached because all possible code paths before it would hit a ``return`` statement and leave the function:
```
Mickeon, line 16: A bit ambiguous because both the next "_that_" and "_it_" may refer to "_in a function_" as the subject.
```suggestion
This warning may appear if there is code that can never be reached because all possible code paths before it would hit a ``return`` statement and leave a function:
```
or something else to mitigate the ambiguity, I am not sure

tutorials/scripting/gdscript/warnings/unsafe_call_argument.rst
```diff
@@ -0,0 +1,58 @@
...
+
+    var original_health = 10
+
+    func _ready():
+        set_health(original_health)  # Will give warning UNSAFE_CALL_ARGUMENT.
+
+    func set_health(new_health: int):
+        current_health = new_health
+
+Because the type of ``original_health`` was not explicitly stated, it is inferred to be a ``Variant``. The ``Variant`` type is broader than the ``int`` type expected by the ``set_health()`` function, so it is possible that the variable ``original_health`` could hold a kind of value incompatible with it.
```
Mickeon, line 28:
```suggestion
Because the type of ``original_health`` was not explicitly stated, it is inferred to be a ``Variant``. The ``Variant`` type is broader than the ``int`` type expected by the ``set_health()`` function's parameter, so it is possible that the variable ``original_health`` could be an incompatible type.
```
or similar. I kind of understand why "kind of value" was said, so I am not sure if mine is the right, either.

Mickeon (commented):
If this is accepted each page should also be linked in the corresponding setting in the class reference. 

This is quite a lot to review at once, and yet, as of this swoop, it's quite nice. But oh boy

tutorials/scripting/gdscript/warnings/get_node_default_without_onready.rst
```diff
@@ -0,0 +1,63 @@
...
+This way, the variable ``my_num`` will always start out with a value of ``3`` in each instance of this class.
+GDScript also has methods to retrieve specific nodes from the scene tree: namely, the :ref:`get_node() <class_Node_method_get_node>` method, and its shorthand versions ``$`` (and ``%`` for unique nodes). Thus, if you want to have an instance variable default to a child of the node with a script, it may be tempting to write something like the following::
+
+.. code-block::
+
+    extends Area2D
+
+    var my_collision_shape = $CollisionShape2D
+
+However, class instance variables' default values are evaluated and assigned before the scene tree is set up. This means that at the time of assigning the default value, the node may not be in the scene tree, and the variable will be set to ``null`` instead.
```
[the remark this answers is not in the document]
Meorge: I agree; since GDScript is so object-oriented, the fact that the variable belongs to a class is implied. I could see a few ways to tighten up the phrasing:

- "However, variables' default values are evaluated..."
- "However, a variable's default values are evaluated..."
- "However, the default values of a variable are evaluated..."

I'm not sure if removing "class instance" or "instance" could cause people to be confused about the behavior of local variables, though. Would it be proper to use the term "property" here instead of "variable"?

tutorials/scripting/gdscript/warnings/int_as_enum_without_match.rst
```diff
@@ -0,0 +1,55 @@
...
+Remember that while Godot will assign integer values to enum members by default, you can also explicitly define their corresponding values:
+
+.. code-block::
+
+    enum MyEnum { ZERO, ONE, TWO, TEN_THOUSAND = 10000 }
+    
+    func _ready():
+        var my_var = 10000 as MyEnum  # Will correspond to MyEnum.TEN_THOUSAND.
+
+Alternatively, you may just need to change the integer number you're attempting to cast as an enum value to be within the range of valid values.
```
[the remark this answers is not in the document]
Meorge: I agree that sentence is quite dense 😅. Will see what I can come up with to make it clearer and improve the accuracy of the valid values part, thanks!

tutorials/scripting/gdscript/warnings/return_value_discarded.rst
```diff
@@ -0,0 +1,41 @@
...
+Assign the returned value to a variable for use later.
+
+.. code-block::
+    
+    func _ready():
+        print("About to get a number...")
+        var num = get_number()
+        print("Got a number! It's %s" % num)
+
+However, some methods in Godot's APIs return values that are not necessary to store. As such, depending on the situation it may make more sense to ignore this warning.
```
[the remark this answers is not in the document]
Meorge: This is a good suggestion for sentence flow, but I'm wondering if it may make sense to just remove "depending on the situation" overall - it feels like it may bloat the sentence more than necessary

tutorials/scripting/gdscript/warnings/get_node_default_without_onready.rst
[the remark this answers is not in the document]
Mickeon: I personally use "property" all the time when referring to member variables (so not local). It is much less ambiguous. The term is also fairly consistent in the class reference at least. So yes

tutorials/scripting/gdscript/warnings/standalone_expression.rst
Meorge: Good point. Perhaps we don't need to try to explicitly define an expression here. Maybe there's somewhere else in the documentation that describes expressions, and we can simply link a reference to that, in case someone's confused about what the warning means by an "expression"?

Mickeon: I think expressions themselves are something you intrinsically learn about when programming. I don't think it should be explained here, but I guess it wouldn't hurt to link https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_basics.html#expressions

tutorials/scripting/gdscript/warnings/confusable_identifier.rst
```diff
@@ -0,0 +1,31 @@
...
+------------------------
+
+Some alphabets such as Cyrillic have characters that look like Latin (i.e., English, Spanish, etc.) characters, but are actually different.
+
+.. code-block::
+
+    var engine_nаme = "Godot"
+    print(engine_name)
+
+In this code snippet, the ``print`` statement would fail, because ``engine_name`` is actually not defined. The identifier in the ``print`` statement uses the Latin character "a" (U+0061), while the identifier in the variable declaration above uses the Cyrillic character "а" (U+0430).
```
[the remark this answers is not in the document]
Meorge: I can double-check later, but I recall going into the code to try and find the "look-alike" table that Godot may have been using to detect these, and was unable to find it (perhaps it was being processed by a third-party library?). The letters being indistinguishable might be helpful here, though. They look identical to each other, yet aren't the same, which is the point of the warning.

tutorials/scripting/gdscript/warnings/assert_always_false.rst
```diff
@@ -0,0 +1,51 @@
...
+
+    Assert statement will raise an error because the expression is always false.
+
+The default warning level for this warning is **Warn**.
+To modify it, see :ref:`ProjectSettings.debug/gdscript/warnings/assert_always_false<class_ProjectSettings_property_debug/gdscript/warnings/assert_always_false>`.
+
+When this warning occurs
+------------------------
+
+The :ref:`assert() <class_@GDScript_method_assert>` keyword can be used to ensure that a given condition is met before allowing code execution to continue. If the first argument passed to it is truthy, the rest of the function will run as expected; if it is falsy, then the project will stop.
```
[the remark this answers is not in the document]
Meorge: As I recall, someone else made the same (or a similar) suggestion a while back, so this is definitely worth revising.

tutorials/scripting/gdscript/warnings/assert_always_false.rst
```diff
@@ -0,0 +1,51 @@
+ASSERT_ALWAYS_FALSE
+=======================
+
+The warning message is:
+
+.. code-block:: none
+
+    Assert statement will raise an error because the expression is always false.
+
+The default warning level for this warning is **Warn**.
```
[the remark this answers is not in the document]
Meorge: Looks like I can define a series of substitutions like
```rst
.. |default_is_ignore| replace:: The default warning level for this warning is **Ignore**.
.. |default_is_warn| replace:: The default warning level for this warning is **Warn**.
.. |default_is_error| replace:: The default warning level for this warning is **Error**.
```
and then use them (like `|default_is_warn|`) to paste that message in. Automating the ProjectSettings links might take a bit more work, since it involves changing text within the reference link. I would really love to figure out a way to automate this well, since as you pointed out so much of the text is the same.

tutorials/scripting/gdscript/warnings/confusable_identifier.rst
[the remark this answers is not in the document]
Meorge: I found this page that visualizes the Unicode confusable characters: https://util.unicode.org/UnicodeJsps/confusables.jsp

For this example, I could replace the Cyrillic character with ɑ (lowercase alpha), like so:
```gdscript
var engine_nɑme = "Godot"
print(engine_name)
```
There are a few other characters (α, ⍺, 𝐚, 𝑎, etc) that could also be used instead.

tutorials/scripting/gdscript/warnings/assert_always_false.rst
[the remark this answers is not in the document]
Mickeon: Unless I'm mistaken I don't think this would work across multiple pages so there may still be a need to merge everything into one.

tutorials/scripting/gdscript/warnings/assert_always_false.rst
```diff
@@ -0,0 +1,47 @@
...
+To modify it, see :ref:`ProjectSettings.debug/gdscript/warnings/assert_always_false<class_ProjectSettings_property_debug/gdscript/warnings/assert_always_false>`.
```
tetrapod00, line 11: Project settings links are styled like this, with setting names in title case as they appear in the editor, separated by `>`. Also "see" should be "set" or "change". "See" is mostly used for pure information reference, but in this case we're telling the reader to change a project setting.

```suggestion
To modify it, set :ref:`Project Settings > Debug > GDScript > Warnings > Assert Always False<class_ProjectSettings_property_debug/gdscript/warnings/assert_always_false>`.
```

(And apply a similar change to all links)

tutorials/scripting/gdscript/warnings/assert_always_false.rst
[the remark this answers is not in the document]
tetrapod00: - I wouldn't let this narrow issue of substitutions determine if we make one page or multiple pages for these warnings. (FWIW a page for each warning seems natural to me, and there is prededent with the C# diagnostics, but this is a larger decision and it seems like other reviewers are leaning towards a single page)

- It's likely not worth the effort to automate the class reference links, since that would involve dynamic replacement and IIRC that isn't straightforward with RST substitutions. 

- You can make a substitution accessible from all pages `rst_epilog`, we already use it at least once:
https://github.com/godotengine/godot-docs/blob/6ae63ddd5f294f195ec5be06e4e92294e49cf2c4/conf.py#L289-L298 

However, we *don't* use that for cases like the upgrading pages or the class reference, where *each* page redefines the same substitions:
https://github.com/godotengine/godot-docs/blob/6ae63ddd5f294f195ec5be06e4e92294e49cf2c4/tutorials/migrating/upgrading_to_godot_4.1.rst?plain=1#L211-L213
https://github.com/godotengine/godot-docs/blob/6ae63ddd5f294f195ec5be06e4e92294e49cf2c4/classes/class_aabb.rst?plain=1#L770-L777
There might be a compile time increase when using rst_epilog, which is why it's not used more. I haven't looked closely at this, though. 

There's also `.. include` ([docs](https://docutils.sourceforge.io/docs/ref/rst/directives.html#including-an-external-document-fragment)) which could be used to only include the substitutions on the pages that use it (and therefore reduce compile times if using rst_epilog is too heavy). We don't currently use `.. include`, but it's usage looks relatively straightforward.

(I still have to do a deeper review of the rest of this later)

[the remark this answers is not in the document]
Meorge: Using an RST file and `.. include` is how I implemented the test on my end.

`warning_page_substitutions.rst`:
```rst
.. |default_is_ignore| replace:: The default warning level for this warning is **Ignore**.
.. |default_is_warn| replace:: The default warning level for this warning is **Warn**.
.. |default_is_error| replace:: The default warning level for this warning is **Error**.
```

`assert_always_false.rst`:
```rst
ASSERT_ALWAYS_FALSE
=======================

The warning message is:

.. code-block:: none

    Assert statement will raise an error because the expression is always false.

|default_is_warn|
To modify it, see :ref:`ProjectSettings.debug/gdscript/warnings/assert_always_false<class_ProjectSettings_property_debug/gdscript/warnings/assert_always_false>`.

.. include:: warning_page_substitutions.rst
```
With this, we'd only have to edit or translate the sentences once in `warning_page_substitutions.rst` and they'd apply everywhere, which is nice. Dynamic substitution would be very helpful for this, but as you said it sounds like it would be quite involved.

tutorials/scripting/gdscript/warnings/assert_always_false.rst
Meorge: Thanks for the catch, will get that fixed!

Meorge:
During one of the previous GDScript meetings, we discussed only including a small subset of the GDScript warnings in the initial PR. This way, they'd be easier for others to thoroughly review and approve before merging.

I think it'd make sense, then, to try and choose the most common warnings first, so that they're the most likely to help people. The more obscure warnings can get pages later.

Based on what I've seen from people in online help communities, I think these pages could make for a good initial batch:
- GET_NODE_DEFAULT_WITHOUT_ONREADY
- INTEGER_DIVISION
- RETURN_VALUE_DISCARDED
- SHADOWED_VARIABLE
- UNASSIGNED_VARIABLE
- UNUSED_VARIABLE

Are there any other warnings that would be good to include? On the flip-side, are any of the ones I've listed here unnecessary to start with? Once I have a solid response/consensus on that, I can work on editing the PR's contents to just include those warning pages.

tutorials/scripting/gdscript/warnings/unassigned_variable_op_assign.rst
```diff
@@ -0,0 +1,43 @@
...
+------------------------
+
+This warning may appear when attempting to use one of the compound-assignment operators, like ``+=``, ``-=``, ``*=``, or ``/=``, on a variable that hasn't had a value assigned to it yet.
+
+.. code-block::
+
+    var my_counter: int
+    my_counter += 1
+
+Compound-assignment operators are shorthand for performing a binary math operation and an assignment operation. The above example is equivalent to:
```
AThousandShips:
```suggestion
Compound-assignment operators are a shorthand for performing a binary math operation and an assignment operation. The above example is equivalent to:
```

tutorials/scripting/gdscript/warnings/static_called_on_instance.rst
```diff
@@ -0,0 +1,47 @@
...
+
+    func _ready():
+        var my_math_funcs = MathFuncs.new()
+
+        # Will give warning STATIC_CALLED_ON_INSTANCE.
+        var result = my_math_funcs.subtract_two(5)
+
+When a function is *static*, it belongs to the class as a whole, not any one specific instance of the class. See :ref:`Static variables <doc_gdscript_basics_static_variables>` and :ref:`Static functions <doc_gdscript_basics_static_functions>` for more information.
+
+There is often no need to call a static function on an instance of a class. In the example above, the function ``subtract_two()`` will always perform the same operation; one instance of the ``MathFuncs`` class would never return a different value than another instance. As such, when the user tries to call ``subtract_two()`` from an instance of ``MathFuncs``, it suggests the user may not understand what the function does.
```
AThousandShips: > one instance of the ``MathFuncs`` class would never return a different value than another instance

This is not necessarily true, only assuming all other conditions are the same, static does not mean stateless, so it might be good to qualify here

tutorials/scripting/gdscript/warnings/int_as_enum_without_match.rst
```diff
@@ -0,0 +1,56 @@
...
+This warning may appear when attempting to cast an integer value to an enum type, but the enum type doesn't have a member with a corresponding value.
+
+.. code-block::
+
+    enum MyEnum { ZERO, ONE, TWO }
+    
+    func _ready():
+        var my_var = 3 as MyEnum  # Will give warning INT_AS_ENUM_WITHOUT_MATCH.
+
+The purpose of the enum is to keep track of a pre-determined number of possible values. In most cases, there is no reason to assign a value outside of those pre-determined ones, so it is considered to be likely a mistake.
```
AThousandShips:
```suggestion
The purpose of the enum is to keep track of a pre-determined number of possible values. In most cases, there is no reason to assign a value outside of those pre-determined ones, so it is considered likely to be a mistake.
```

tutorials/scripting/gdscript/warnings/confusable_identifier.rst
```diff
@@ -0,0 +1,31 @@
...
+
+How to fix this warning
+-----------------------
+
+Avoid using Cyrillic or other alphabets' characters that are visually similar to Latin ones. A good rule of thumb is to prefer the Latin alphabet for program identifiers.
```
AThousandShips:
```suggestion
Avoid using Cyrillic or other scripts' characters that are visually similar to Latin ones. A good rule of thumb is to prefer the Latin alphabet for program identifiers.
```
Not all writing systems are alphabets (granted most would not be confusable with alphabets, but you can spoof with both Hebrew and Thai)

tutorials/scripting/gdscript/warnings/confusable_identifier.rst
```diff
@@ -0,0 +1,31 @@
...
+
+    The identifier "my_vаr" has misleading characters and might be confused with something else.
+
+The default warning level for this warning is **Warn**.
+To modify it, set :ref:`Project Settings > Debug > GDScript > Warnings > Confusable Identifier<class_ProjectSettings_property_debug/gdscript/warnings/confusable_identifier>`.
+
+When this warning occurs
+------------------------
+
+Some alphabets such as Cyrillic have characters that look like Latin (such as English, Spanish, etc.) characters, but are actually different.
```
AThousandShips:
```suggestion
Some scripts such as Cyrillic have characters that look like Latin (such as English, Spanish, etc.) characters, but are actually different.
```

tutorials/scripting/gdscript/warnings/assert_always_false.rst
```diff
@@ -0,0 +1,47 @@
...
+
+    Assert statement will raise an error because the expression is always false.
+
+The default warning level for this warning is **Warn**.
+To modify it, set :ref:`Project Settings > Debug > GDScript > Warnings > Assert Always False<class_ProjectSettings_property_debug/gdscript/warnings/assert_always_false>`.
+
+When this warning occurs
+------------------------
+
+The :ref:`assert() <class_@GDScript_method_assert>` keyword can be used to ensure that a given condition is met before allowing code execution to continue. If the first argument passed to it is true in a boolean context, the rest of the function will run as expected; if it is false in a boolean context, then the project will stop.
```
AThousandShips, line 16:
```suggestion
The :ref:`assert() <class_@GDScript_method_assert>` method can be used to ensure that a given condition is met before allowing code execution to continue. If the first argument passed to it is true in a boolean context, the rest of the function will run as expected; if it is false in a boolean context, then the project will stop.
```

tutorials/scripting/gdscript/warnings/assert_always_true.rst
```diff
@@ -0,0 +1,36 @@
...
+
+    Assert statement is redundant because the expression is always true.
+
+The default warning level for this warning is **Warn**.
+To modify it, set :ref:`Project Settings > Debug > GDScript > Warnings > Assert Always True<class_ProjectSettings_property_debug/gdscript/warnings/assert_always_true>`.
+
+When this warning occurs
+------------------------
+
+The :ref:`assert() <class_@GDScript_method_assert>` keyword can be used to ensure that a given condition is met before allowing code execution to continue. If the first argument passed to it evaluates to ``true``, the rest of the function will run as expected; if it is ``false``, then the project will stop.
```
AThousandShips, line 16:
```suggestion
The :ref:`assert() <class_@GDScript_method_assert>` method can be used to ensure that a given condition is met before allowing code execution to continue. If the first argument passed to it evaluates to ``true``, the rest of the function will run as expected; if it is ``false``, then the project will stop.
```

tutorials/scripting/gdscript/warnings/confusable_identifier.rst
Mickeon: The term "script" is consistent to how the rest of the engine refers to the concept, but I believe here "alphabet" was used because otherwise it'd be very much confusable with GDScript itself.

AThousandShips: I don't think anyone will confuse this with GDScript, why would anyone think "Cyrillic" is a GDScript?

Meorge: Admittedly I think here I just used the term "alphabet" because I was thinking of individual characters; I don't remember taking the confusion with GDScript scripts into consideration 😅 It is a really good point, though. I feel like for now we can use "scripts" here since it is more correct, but revisit it if we get the feeling it's confusing people.

tutorials/scripting/gdscript/warnings/static_called_on_instance.rst
Meorge: In this case, I think I was specifically referring to the `subtract_two()` function I defined here. Since it only uses its input argument and a constant to calculate its result, all `MathFuncs` instances would give the same output. However this could probably be made clearer, and/or the example as a whole reworked.

tutorials/scripting/gdscript/warnings/confusable_identifier.rst
Meorge: Weird, I didn't get a notification about your comment until a while after I'd posted mine...

> I don't think anyone will confuse this with GDScript, why would anyone think "Cyrillic" is a GDScript?

That's fair, although I could also see someone parsing the sentence as "Cyrillic characters, or characters from other GDScript scripts", possibly? As I said in my previous comment, though, it's probably okay to just use "scripts" for now and then revisit if people do find it confusing.

Meorge:
This PR has been updated to only contain warning pages for the six warnings listed above:

- GET_NODE_DEFAULT_WITHOUT_ONREADY
- INTEGER_DIVISION
- RETURN_VALUE_DISCARDED
- SHADOWED_VARIABLE
- UNASSIGNED_VARIABLE
- UNUSED_VARIABLE

We can start adding more pages soon, but hopefully this will make for a good initial batch.

tutorials/scripting/gdscript/warnings/get_node_default_without_onready.rst
```diff
@@ -0,0 +1,63 @@
+GET_NODE_DEFAULT_WITHOUT_ONREADY
+====================================
+
+This warning appears when the default value of a node's property is set to a location in the scene tree without using the ``@onready`` annotation.
```
Shadows-of-Fire: This reads as... well, really overly technical to me. It took me a couple re-reads to figure out what it was trying to tell me.

Maybe:

```suggestion
This warning appears when a node's property is initialized to a location in the scene tree without using the ``@onready`` annotation.
```

Shadows-of-Fire: Also, we should replace `node's property` here with `member variable`, and also use `member variable` more consistently over `property` for the remainder of these documents. The GDScript reference declares them as member variables [here](https://docs.godotengine.org/en/4.4/tutorials/scripting/gdscript/gdscript_basics.html#example-of-gdscript)

tutorials/scripting/gdscript/warnings/get_node_default_without_onready.rst
```diff
@@ -0,0 +1,63 @@
...
+Instance properties can be set by declaring them outside of a method. Additionally, they can be given a default value using ``=``:
+
+.. code-block:: gdscript
+
+    extends Area2D
+
+    var my_num = 3
+
+The property ``my_num`` will always start out with a value of ``3`` in each instance of this class.
+GDScript also has methods to retrieve specific nodes from the scene tree: namely the :ref:`get_node() <class_Node_method_get_node>` method, and its shorthand versions ``$`` (and ``%`` for unique nodes). Thus, if you want to have a property's default value be a particular child node, it may be tempting to write something like the following:
```
Shadows-of-Fire: I'm not sure we need to explain the concept of member variable initialization within this document. I think we can assume that the reader is familiar with it by the time they reach here.

tutorials/scripting/gdscript/warnings/get_node_default_without_onready.rst
Mickeon: A grand majority of the class reference refers to them as properties. We _generally_ refer to them as properties outside the docs, as well. So I don't necessarily agree with this.

A proper change should be done in bulk based on consensus.

Meorge: I like the suggestion for the sentence and will add that in, thanks!

As for the "property" vs "member variable" situation, I think I was mostly inspired by the class reference, where they are referred to as "properties". Granted, those are for built-in classes that are more of a black box for users, whereas the things we're talking about here users have direct access to modify and inspect. But for now, "property" is more succinct and used frequently in documentation, so I think it makes sense to keep it that way.

tutorials/scripting/gdscript/warnings/get_node_default_without_onready.rst
Meorge: That's a good point. Originally, my thought was that the target audience for these docs is new programmers who could be confused by the warning messages themselves. They may be following tutorials and pasting code without fully understanding it, so having the full thought process leading to the code that causes the warning explained could help them.

That being said, explaining the basics of member variable initialization like this might be a bit *too* much background. This could be simplified to just the `get_node()` (and shorthand) methods, as well as how member variable initialization happens before building the scene tree. If there's a guide elsewhere in the docs that explains member variables for newcomers, we could link it here.

Meorge:
I've tried to refine the GET_NODE_DEFAULT_WITHOUT_ONREADY page, bringing it more in line with the other, later-written pages as well as removing unnecessary background information.

tutorials/scripting/gdscript/warnings/get_node_default_without_onready.rst
```diff
@@ -0,0 +1,46 @@
+GET_NODE_DEFAULT_WITHOUT_ONREADY
+====================================
```
Calinou:
```suggestion
GET_NODE_DEFAULT_WITHOUT_ONREADY
================================
```

tutorials/scripting/gdscript/warnings/integer_division.rst
```diff
@@ -0,0 +1,41 @@
+INTEGER_DIVISION
+====================
```
Calinou:
```suggestion
INTEGER_DIVISION
================
```

tutorials/scripting/gdscript/warnings/return_value_discarded.rst
```diff
@@ -0,0 +1,41 @@
+RETURN_VALUE_DISCARDED
+==========================
```
Calinou:
```suggestion
RETURN_VALUE_DISCARDED
======================
```

tutorials/scripting/gdscript/warnings/shadowed_variable.rst
```diff
@@ -0,0 +1,38 @@
+SHADOWED_VARIABLE
+=====================
```
Calinou:
```suggestion
SHADOWED_VARIABLE
=================
```

tutorials/scripting/gdscript/warnings/unassigned_variable.rst
```diff
@@ -0,0 +1,36 @@
+UNASSIGNED_VARIABLE
+=======================
```
Calinou:
```suggestion
UNASSIGNED_VARIABLE
===================
```

tutorials/scripting/gdscript/warnings/unused_variable.rst
```diff
@@ -0,0 +1,45 @@
+UNUSED_VARIABLE
+===================
```
Calinou:
```suggestion
UNUSED_VARIABLE
===============
```

Calinou (commented):
Adjust underline lengths to avoid Sphinx build warnings:

Calinou (changes requested):
Looks good to me.

However, several of the pages added here contain extraneous blank lines at the end. There should only be one newline at the end of each file (as added by most text editors). On GitHub, it should look like there are no newlines at the end of each file (but no red warning sign should appear either).

Meorge:
Thanks for the review! The files should hopefully now be fixed to have the correct endings.

Calinou (approved):
Looks good to me.

tutorials/scripting/gdscript/warnings/get_node_default_without_onready.rst
```diff
@@ -0,0 +1,46 @@
+GET_NODE_DEFAULT_WITHOUT_ONREADY
+================================
+
+This warning appears when a node's property is initialized to a location in the scene tree without using the ``@onready`` annotation.
```
AThousandShips:
```suggestion
This warning appears when a node's property is initialized to a node in the scene tree without using the ``@onready`` annotation.
```
I'd say

tutorials/scripting/gdscript/warnings/get_node_default_without_onready.rst
```diff
@@ -0,0 +1,46 @@
...
+This warning may appear when attempting to assign the default value of a property to a node in the scene tree, like so:
+
+.. code-block:: gdscript
+
+    extends Area2D
+
+    # Will give warning GET_NODE_DEFAULT_WITHOUT_ONREADY.
+    var my_collision_shape = $CollisionShape2D
+
+If you want a property to refer to another node in the scene tree, you may be inclined to reference it using :ref:`get_node() <class_Node_method_get_node>` (or the ``$`` and ``%`` shorthand versions) when setting its default value.
```
AThousandShips:
```suggestion
If you want a property to refer to another node in the scene tree, you may be inclined to reference it using
:ref:`get_node() <class_Node_method_get_node>` (or the ``$`` and ``%`` shorthand versions) when setting its default value.
```
For all, keep to a reasonable length

tutorials/scripting/gdscript/warnings/index.rst
```diff
@@ -0,0 +1,25 @@
...
+
+.. _doc_gdscript_warnings:
+
+GDScript warnings
+=================
+
+This is a collection of warnings that GDScript can emit, and information on how to fix the code that causes them.
+
+.. note::
+   The list of warnings is currently incomplete. We hope to add pages for more warnings over time, so that each one can be thoroughly reviewed for correctness and clarity.
```
AThousandShips:
```suggestion

   The list of warnings is currently incomplete. We hope to add pages for more warnings over time,
   so that each one can be thoroughly reviewed for correctness and clarity.
```

tutorials/scripting/gdscript/warnings/get_node_default_without_onready.rst
```diff
@@ -0,0 +1,46 @@
...
+When this warning occurs
+------------------------
+
+This warning may appear when attempting to assign the default value of a property to a node in the scene tree, like so:
+
+.. code-block:: gdscript
+
+    extends Area2D
+
+    # Will give warning GET_NODE_DEFAULT_WITHOUT_ONREADY.
```
AThousandShips, line 31: I don't think "give warning" is an established terminology? Where has this been used elsewhere?

Meorge: From what I can see in the Godot Docs, there's not generally a lot of discussion of GDScript warnings. The verb "give" is used in at least one spot, [Static typing in GDScript](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/static_typing.html#warning-system):

> Godot gives you warnings about your code as you write it.

[Later on the same page](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/static_typing.html#unsafe-property-access-and-unsafe-method-access-warnings), "produce" is used:

> ```gdscript
> if "some_property" in node_2d:
> 	node_2d.some_property = 20  # Produces UNSAFE_PROPERTY_ACCESS warning.
> 
> if node_2d.has_method("some_function"):
> 	node_2d.some_function()  # Produces UNSAFE_METHOD_ACCESS warning.
> ```
> However, this code will produce `UNSAFE_PROPERTY_ACCESS` and `UNSAFE_METHOD_ACCESS` warnings...

In the dedicated page for the [GDScript warning system](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/warning_system.html#gdscript-warning-system), "trigger" is used:

> Godot will add an annotation above the corresponding line and the code won't trigger the corresponding warning anymore

---

So, overall it appears that "give", "produce", and "trigger" all have precedent in the docs currently.

AThousandShips: Produce seems to be the most common, but I think we can leave it as is

AThousandShips (approved):

tutorials/scripting/gdscript/warnings/get_node_default_without_onready.rst
Meorge: Alright, I will leave it for now then 🙂

tutorials/scripting/gdscript/warnings/return_value_discarded.rst
```diff
@@ -0,0 +1,42 @@
...
+
+.. code-block::
+    
+    func _ready():
+        print("About to get a number...")
+        var num = get_number()
+        print("Got a number! It's %s" % num)
+
+However, some methods in Godot's APIs return values that are not necessary to store.
+As such, depending on the situation, it may make more sense to ignore this warning.
```
Cykyrios, line 42: It may be worth mentioning that `Variant` return types are optional (e.g. `type_convert()` returns `Variant` and will not throw a warning if you don't assign it to a variable). This would make it clearer than just "some methods" (but I'm not sure if there are other cases or exceptions).

HolonProduction: I wouldn't go into too much detail on this. To me it's not exactly clear if that's intended in the first place, and even if it is, it's probably just to reduce false positives.

Keeping this in mind:

> The target audience for the pages is relative newcomers to GDScript and/or programming in general; people who may be using code from tutorials without fully understanding it, and thus could be confused by the warning messages Godot provides. As such, the explanations aim to be extra beginner-friendly.

I see no point in going into details on potential false negatives.

tutorials/scripting/gdscript/warnings/get_node_default_without_onready.rst
```diff
@@ -0,0 +1,54 @@
...
+    var my_collision_shape = $CollisionShape2D
+
+If you want a property to refer to another node in the scene tree, you may be inclined
+to reference it using :ref:`get_node() <class_Node_method_get_node>`
+(or the ``$`` and ``%`` shorthand versions) when setting its default value.
+
+However, properties' default values are evaluated and assigned *before* the scene tree is set up.
```
HolonProduction: > *before* the scene tree is set up

That's not really the case though. If you instantiate a scene while the game is running, the scene tree is set up. It might be more accurate to say "before a node gets added to the scene tree"

tutorials/scripting/gdscript/warnings/get_node_default_without_onready.rst
```diff
@@ -0,0 +1,54 @@
...
+This means that when Godot tries to assign that value, the node you're referring to may not be in
+the scene tree yet, and the property will be set to ``null`` instead.
```
HolonProduction: The issue isn't the node you are referring to, it's the node with the script attached that is not in the scene tree.

tutorials/scripting/gdscript/warnings/get_node_default_without_onready.rst
```diff
@@ -0,0 +1,54 @@
...
+.. code-block:: none
+
+    The default value uses "get_node()" which won't return nodes in the scene tree before "_ready()" is called. Use the "@onready" annotation to solve this.
+
+    The default value uses "$" which won't return nodes in the scene tree before "_ready()" is called. Use the "@onready" annotation to solve this.
+
+    The default value uses "%" which won't return nodes in the scene tree before "_ready()" is called. Use the "@onready" annotation to solve this.
+
+The default warning level for this warning is **Error**.
+To modify it, set :ref:`Project Settings > Debug > GDScript > Warnings > Get Node Default Without Onready<class_ProjectSettings_property_debug/gdscript/warnings/get_node_default_without_onready>`.
```
HolonProduction, line 19: I don't quite understand who this section is meant for. I thought this page was aimed at beginners to give them a short explanation on the common warnings.

I don't see how the default value is relevant to them.

HolonProduction (commented):
I'm not exactly on top of the discussion, so that might have been discussed already. But I'm wondering whether it wouldn't be more helpful to beginners to expand the documentation on the settings with code samples. That way we could probably link to it directly from within the editor.

tutorials/scripting/gdscript/warnings/return_value_discarded.rst
Meorge: I'm not personally familiar with the `Variant` return types such as for `type_convert()` you brought up, but admittedly it not throwing a warning sounds more to me like it might be a bug on that side 😅

I have thought about having expandable sections that could go into more detail when necessary, but I don't recall if Godot's documentation supports such text sections, and we'd also want to have those details verified before asserting them in the documentation.

tutorials/scripting/gdscript/warnings/get_node_default_without_onready.rst
Meorge: Eventually, the hope is to have every warning documented with examples, explanations, and fixes. This initial PR just adds pages for a few warnings that might be the most common, so that the review burden isn't as great.

Can you clarify what you mean by "I don't see how the default value is relevant to [beginners for a short explanation on the common warnings]"? I'm not sure I currently follow; the `GET_NODE_DEFAULT_WITHOUT_ONREADY` warning involves setting a property's default value with `get_node()` (or a shorthand equivalent) which is why default values are discussed here.

tutorials/scripting/gdscript/warnings/return_value_discarded.rst
Cykyrios: I certainly hope `Variant` returns not throwing the warning is not a bug, because unless I'm mistaken, functions can "return void" as a valid `Variant`. E.g. with a function that returns a `Variant`, you can do `return some_callable.callv(args)`, where `some_callable` can return `-> void`, so having the warning would be a problem in that scenario, and having to `@warning_ignore` the call would also be problematic.

But yeah, this is way out of scope for a document aimed at beginners.

tutorials/scripting/gdscript/warnings/get_node_default_without_onready.rst
HolonProduction: > Can you clarify what you mean by "I don't see how the default value is relevant to [beginners for a short explanation on the common warnings]"?

That was referring to this part:
>> The default warning level for this warning is **Error**.

The default value is already in the class reference, and there it's actually always up-to-date (although displaying it as int is a bit terrible I guess).

I don't think including this information here makes that much sense.

> Eventually, the hope is to have every warning documented with examples, explanations, and fixes.

Yeah I get that. I was talking about the general structure that you used here. I feel like we might just want to strike this initial section, and start with "When this warning occurs" directly.

>> This warning appears when a node's property is initialized to a node in the scene tree
without using the ``@onready`` annotation.

It seems like this is basically saying the same as the explanation you provide below.

>> Depending on how you attempt to access the scene tree, the warning message will
be one of the following:

I see no point in spelling out all variations. That's bound to get outdated, and users already know this message if they looked up the error.
(If you did that for SEO, that might be a point I guess. But we have well defined warning names, so that seems unnecessary to me 🤔.)

Meorge:
Thanks for the reviews!

> I'm wondering whether it wouldn't be more helpful to beginners to expand the documentation on the settings with code samples. That way we could probably link to it directly from within the editor.

I recall this may have been discussed a while ago (or something similar – perhaps one page for all warning descriptions), but I felt that having all of the warnings' content on one page would lead to a massive page and/or force each warning's description to be too short. For me, only having one warning's content on a page gives the content more room to breathe and be on the longer side if that's necessary to help people understand the warning.

Linking to it directly from within the editor would be great, but I'm not sure that putting it in the class reference under the settings is a strict requirement for that. The current built-in class reference for Godot already contains links to manual pages which open in the user's web browser when clicked:

CleanShot 2026-07-09 at 08 45 36@2x

If there was a clean way to include these manual pages in the offline download, that'd be even better. Perhaps we could look into a way to add a section to the class reference for GDScript warnings...? Otherwise though, I think manual pages linked to from the class reference would be sufficient.

HolonProduction:
> Otherwise though, I think manual pages linked to from the class reference would be sufficient.

The audience this is targeting is not going to go to the class ref. If we want to link it, it needs to be from the warning panel.

But thinking about it, I think the shader editor also links directly to online documentation. So I guess we are fine with linking to the online docs from editor UI.

tutorials/scripting/gdscript/warnings/get_node_default_without_onready.rst
Meorge: > That was referring to this part

Ah, I see, thanks for the clarification! Thinking about it now, I agree that it might make more sense to just link to the class reference from these pages, and only specify the default value there. That way, updates to the default value won't make these pages outdated or inaccurate.

I recall there being discussion of adding links from these pages to the settings class reference; perhaps the inclusion of the default values was to make them flow more naturally? I'll see what I can find in the previous conversations, and if I can't find anything or what I find isn't especially compelling, I'll remove the default values and only link to the class reference.

---

> I feel like we might just want to strike this initial section, and start with "When this warning occurs" directly.
> ...
> It seems like this is basically saying the same as the explanation you provide below.

I definitely see where you're coming from here; the first sentence of the page is largely retreading ("pre-treading"?) the description given in the "When this warning occurs" section. The intent here is to provide a short (one- or two-sentence) description of the warning, along with the warning messages that might be displayed to the user (and thus that they might paste into a search engine to figure it out). 

The "short description" being largely repeated below has precedent in the Godot documentation. For example, in the [Node3D class reference](https://docs.godotengine.org/en/stable/classes/class_node.html), the initial short description says
> Base object in 3D space, inherited by all 3D nodes.

and just below, the long description starts with
> The **Node3D** node is the base representation of a node in 3D space. All other 3D nodes inherit from this class.

Bringing it back to the warning page, I suppose we could remove the short description of the warning and put the examples of warning messages in a section named something like "What this warning looks like". But given that other areas of the manual/class reference often start with an overview/intro section that is expanded upon in the following sections, I think it might be best to keep as-is.

---

> If you did that for SEO, that might be a point I guess

Yep, that was pretty much the reason. I would expect users to copy and paste the warning messages they receive into a search engine to look for help with it, so I want to make sure the messages themselves (not just the codes/codenames, like `GET_NODE_DEFAULT_WITHOUT_ONREADY`) cause these pages to show up in search results.

Meorge:
Yep, I think linking to it from the warning panel would be great. When the code actions PR is merged and this docs PR is merged as well, I've also been thinking of doing a follow-up PR that lets users access the warning page from the code actions menu itself. But that's still a ways away 😅

Issue #10599, opened 2025-02-02
Shouldn't Area2D Make Reference to TileMapLayer Instead of TileMap?


CouchSloth:
**Your Godot version:** 4.3 Stable

Since '`TileMap` is deprecated, wouldn't it be best to update `Area2D` documents to reflect that? Or changing it to reference both `TileMapLayer` and `TileMap`?
Unless there has been a change on how `Area2D` and `TileMapLayers` interact compared to `TileMap`. 

![Image](https://github.com/user-attachments/assets/69bce11f-39f0-49fc-9709-f7043327d8f1)

![Image](https://github.com/user-attachments/assets/927889e4-3a6c-4d1b-87f4-f651507c0545)

With `TileMapLayer` being where a lot of effort and utility has been put into, it would be nice to see it referenced more in documentation. Especially since `TileMapLayer` does not seem to inherit from `TileMap`.

![Image](https://github.com/user-attachments/assets/45229eab-2b13-45ad-87c5-655780d2e6e0)

**URL to the documentation page (if already existing):** https://docs.godotengine.org/en/stable/classes/class_area2d.html

tetrapod00:
I'm not a 2D expert, but assuming that it works as a drop-in replacement, it seems like it's time to update all these links to TileMap in the class reference. There are more cases than just this Area2D class.
One instance was already changed in https://github.com/godotengine/godot/pull/98782.

CouchSloth:
I imagine there are a lot, such as `TileSet` still references `TileMap` and not `TileMapLayer` in the documentation [here](https://docs.godotengine.org/en/stable/classes/class_tileset.html#class-tileset) but the "[Using TileSets](https://docs.godotengine.org/en/stable/tutorials/2d/using_tilesets.html#doc-using-tilesets)" it is properly updated. However, I just noticed this one since I'm working through some StackOverflow questions. 
It seems to be an issue where the tutorials are updated but the documents have fallen behind on updates and usefulness.

tetrapod00:
The TileSet class reference is actually updated in the [latest branch](https://docs.godotengine.org/en/latest/classes/class_tileset.html). Since the class reference source is in the main godot repo, not here, and that repo is a lot more conservative with cherrypicking changes back a version, I'm not actually certain that 4.3 will end up getting that change. If it does, it might come whenever the 4.3.1 patch update happens.

CouchSloth:
[...]

PR #10638, opened 2025-02-07
Update multiple files to reference TileMapLayer instead of TileMap

Dowsley:
Should I create an issue for this first?


AThousandShips:

Dowsley:
will do!

tetrapod00:
> Should I create an issue for this first?

Arguably https://github.com/godotengine/godot-docs/issues/10599 already represents this problem.

PR #10648, opened 2025-02-08
Fix "<>" symbols in sphinx.

NickHagopianZ:
In Sphinx the symbols <> cause text inside them to disappear. For example here's an include that should have "&lt;godot_cpp/classes/sprite2d.hpp&gt;" following it:

image

Unfortunately, with this change, "& gt;" and "& lt;" (without spaces) appear in the code preview in GitHub. I couldn't find a great solution for fixing this in both GitHub and Sphinx, but assume Sphinx is the priority.

### DIFF
--- a/tutorials/scripting/gdextension/gdextension_cpp_example.rst
+++ b/tutorials/scripting/gdextension/gdextension_cpp_example.rst
@@ -104,10 +104,11 @@ call the Godot executable:
     godot --dump-extension-api
 
 The resulting ``extension_api.json`` file will be created in the executable's
-directory. Copy it to the project folder and add ``custom_api_file=<PATH_TO_FILE>``
+directory. Copy it to the project folder and add ``custom_api_file=&lt;PATH_TO_FILE&gt;``
 to the scons command below.
 
-To generate and compile the bindings, use this command (replacing ``<platform>``
+
+To generate and compile the bindings, use this command (replacing ``&lt;platform&gt;``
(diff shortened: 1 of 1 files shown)


tetrapod00 (changes requested):
I don't think this PR is needed. I tested the page (https://docs.godotengine.org/en/latest/tutorials/scripting/gdextension/gdextension_cpp_example.html) on both `stable` and `latest` on Windows 11 in Edge and Firefox, and every `<` renders just fine for me.

The specific example in your post renders fine, too:
image

How did you test to come to the conclusion that Sphinx can't handle `<` and `>` symbols?

skyace65:
I'm also not seeing it on Windows 10 and Firefox

NickHagopianZ:
For a minute I thought it was just my Brave browser, but I also see this issue in Edge and Chrome. When running the docs from source, I don't see this issue in Brave. When testing here https://snippets.documatt.com/, I do see the issue.

Feel free to close this if no one else can replicate. I don't plan on investing any more time on this issue if it's just a me problem.

skyace65:
@NickHagopianZ given the issues this would cause with code preview I'm going to close this PR. I'm assuming since you mention Edge that you're running on Windows(If not, what OS are you on?). Do you have any plugins installed in all of your browsers? if so I'd recommend turning them off and on to see if that's a cause.

Also, I just tested Edge on my machine and I'm not seeing the issue there either.

PR #10651, opened 2025-02-09
Add note on pressing F6 for current scene vs. F5 for default scene

raziel444:
Add note within the "Using Signals" tutorial to clarify usage of F5 (run default scene) vs. F6 (run current scene) #10650
Adds a note within the "Using Signals" tutorial to clarify usage of F5 (run default scene) vs. F6 (run current scene).

When following the tutorial instructions explicitly, the user will have the node_2d scene selected and be instructed to run the scene. If the user chooses to do so by pressing F5 or the Play button, the default scene will run, which does not contain the button and creates confusion.

This change will add a note immediately following those instructions to explain the difference in behavior between F5 and F6. This should ease people over a possible stumbling block without having to seek additional resources beyond the documentation.

This issue has been encountered by others as well, as seen here:
https://www.reddit.com/r/godot/comments/16regpg/godot_411_button_not_showing_up/

### DIFF
--- a/getting_started/step_by_step/signals.rst
+++ b/getting_started/step_by_step/signals.rst
@@ -252,7 +252,7 @@ Your complete ``sprite_2d.gd`` code should look like the following.
         }
     }
 
-Run the currrent scene by pressing :kbd:`F6` (:kbd:`Cmd + R` on macOS),
+Run the current scene by pressing :kbd:`F6` (:kbd:`Cmd + R` on macOS),
 and click the button to see the sprite start and stop.
 
 Connecting a signal via code
--- a/getting_started/step_by_step/signals.rst
+++ b/getting_started/step_by_step/signals.rst
@@ -252,12 +252,8 @@ Your complete ``sprite_2d.gd`` code should look like the following.
         }
     }
 
-Run the scene now and click the button to see the sprite start and stop.
-
-.. note:: If you try to run your scene and you don't see the button you created, you may have
-          run the default scene by pressing :kbd:`F5` (:kbd:`Cmd + B` on macOS) rather than the 
-          current scene by pressing :kbd:`F6` (:kbd:`Cmd + R` on macOS).  Make sure you have the
-          ``node_2d`` scene selected and then press :kbd:`F6` (:kbd:`Cmd + R` on macOS).
+Run the currrent scene by pressing :kbd:`F6` (:kbd:`Cmd + R` on macOS),
+and click the button to see the sprite start and stop.
 
 Connecting a signal via code
 ----------------------------
--- a/getting_started/step_by_step/signals.rst
+++ b/getting_started/step_by_step/signals.rst
@@ -254,6 +254,11 @@ Your complete ``sprite_2d.gd`` code should look like the following.
 
 Run the scene now and click the button to see the sprite start and stop.
 
+.. note:: If you try to run your scene and you don't see the button you created, you may have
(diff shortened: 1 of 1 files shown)


getting_started/step_by_step/signals.rst
```diff
@@ -254,6 +254,11 @@ Your complete ``sprite_2d.gd`` code should look like the following.
 
 Run the scene now and click the button to see the sprite start and stop.
 
+.. note:: If you try to run your scene and you don't see the button you created, you may have
+          run the default scene by pressing :kbd:`F5` (:kbd:`Cmd + B` on macOS) rather than the 
+          current scene by pressing :kbd:`F6` (:kbd:`Cmd + R` on macOS).  Make sure you have the
+          ``node_2d`` scene selected and then press :kbd:`F6` (:kbd:`Cmd + R` on macOS).
+
```
tetrapod00, line 261:
```suggestion
```
I'd delete the new note and just change the sentence above to the following. (Note that the line break is correct to stay under 80 characters wide)
```
Run the currrent scene by pressing :kbd:`F6` (:kbd:`Cmd + R` on macOS),
and click the button to see the sprite start and stop.
```

We use similar phrasing already in a few places in these Getting Started tutorials, including in this page itself:
https://github.com/godotengine/godot-docs/blob/1d9c4a4ae1d81117548ca2c7b25faa4a10f62fe4/getting_started/step_by_step/signals.rst?plain=1#L104-L107

I don't think we need to add a whole note clarifying the difference between `Run Project` and `Run Current Scene` every time this comes up. If such a note is needed and doesn't exist (I'd need to check), it should come a *lot* earlier in the intro sequence than here.

tetrapod00 (changes requested):

tetrapod00:
I think if we want to add a note really explaining the difference between pressing `F5` and `F6`, we should do it earlier in the tutorial sequence. I'm sympathetic to the idea, since it seems to be a common confusion as you noted, so maybe we should split that out into another issue to add a note somewhere else.

getting_started/step_by_step/signals.rst
raziel444: Agreed, this was just the first time I encountered difficulty due to the contrast between the two options.  I think your suggestion is more appropriate than mine.

I checked the docs and the first reference to the "run button" is immediately following creation of the "Hello World" label, here:
https://docs.godotengine.org/en/stable/getting_started/step_by_step/nodes_and_scenes.html#running-the-scene

I will create a PR to add a note/context at that earlier point of the "Getting Started" guide.

skyace65 (approved):

getting_started/step_by_step/signals.rst
```diff
@@ -252,7 +252,8 @@ Your complete ``sprite_2d.gd`` code should look like the following.
         }
     }
 
-Run the scene now and click the button to see the sprite start and stop.
+Run the currrent scene by pressing :kbd:`F6` (:kbd:`Cmd + R` on macOS),
```
skyace65, line 255:
```suggestion
Run the current scene by pressing :kbd:`F6` (:kbd:`Cmd + R` on macOS),
```

skyace65 (changes requested):

skyace65 (approved):

skyace65:

getting_started/step_by_step/signals.rst
raziel444: Created #10656 for this other topic.

PR #10655, opened 2025-02-09
Wrong function name in documentation

fsantos98:
In the [Removing old creeps](https://docs.godotengine.org/en/stable/getting_started/first_2d_game/06.heads_up_display.html#removing-old-creeps) section, the documentation states that the line to remove creeps when the game is over should be added to the `new_game()` function. However, this is incorrect, the creeps should be cleared in the game_over() function instead. This PR corrects that mistake.

### DIFF
--- a/getting_started/first_2d_game/06.heads_up_display.rst
+++ b/getting_started/first_2d_game/06.heads_up_display.rst
@@ -298,7 +298,7 @@ Now all mobs will be in the "mobs" group.
 
 .. image:: img/scene_group_mobs.webp
 
-We can then add the following line to the ``new_game()`` function in ``Main``:
+We can then add the following line to the ``game_over()`` function in ``Main``:
 
 .. tabs::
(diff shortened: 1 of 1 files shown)


skyace65:
This is not correct. The code is intended to go into the ``new_game()`` function, you can see it in the demo repository here: https://github.com/godotengine/godot-demo-projects/blob/master/2d/dodge_the_creeps/main.gd

That said, if you want to change the function it goes in you must open a PR in the demo repository changing the code there as well. And please provide a reasoning for why the code should go in one function vs the other in a PR for that.

fsantos98:
I opened [PR#1164](https://github.com/godotengine/godot-demo-projects/pull/1164) with my reasoning. If that does not make sense I'll close this PR.

aaronfranke:
Thank you for your contribution! Unfortunately, we have decided to reject this change, and the corresponding demos PR.

PR #10659, opened 2025-02-10
Update links to C# and GDScript sections

tetrapod00:
- Change all links to `_doc` anchors instead of `toc`s. Referencing `toc`s directly really isn't idiomatic anymore, since we have intro pages with additional intro paragraphs. It usually doesn't come up since the index pages are short, but these links technically point to the middle of the page not the top.
- Move `doc_c_sharp` and `doc_gdscript` to the index pages. In most cases, when linking to the GDScript or C# language in general, it should go to the general index page, not the introductory page within a section. Making these two anchors be on the index page is also more resilient to future doc reorganization.
- Add `doc_gdscript_reference` anchor. In some cases we do want to point to the reference page specifically, though.
- Update all existing links to point to the right place, updating grammar if needed. In most cases I pointed links at the section, not the GDScript reference page. In some cases I specified a specific header in the *long* GDScript reference page which was previously unspecified.

### DIFF
--- a/getting_started/step_by_step/scripting_first_script.rst
+++ b/getting_started/step_by_step/scripting_first_script.rst
@@ -26,9 +26,8 @@ The equivalent C# code has been included in another tab for convenience.
 .. image:: img/scripting_first_script_rotating_godot.gif
 
 .. seealso:: To learn more about GDScript, its keywords, and its syntax, head to
-             the :ref:`GDScript reference<doc_gdscript>`.
-
-.. seealso:: To learn more about C#, head to the :ref:`C# basics <doc_c_sharp>` page.
+             the :ref:`doc_gdscript` section. To learn more about C#,
+             head to the :ref:`doc_c_sharp` section.
 
 Project setup
 -------------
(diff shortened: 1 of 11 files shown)


getting_started/step_by_step/scripting_first_script.rst
```diff
@@ -26,9 +26,8 @@ The equivalent C# code has been included in another tab for convenience.
 .. image:: img/scripting_first_script_rotating_godot.gif
 
 .. seealso:: To learn more about GDScript, its keywords, and its syntax, head to
-             the :ref:`GDScript reference<doc_gdscript>`.
-
-.. seealso:: To learn more about C#, head to the :ref:`C# basics <doc_c_sharp>` page.
+             the :ref:`doc_gdscript` section. To learn more about C#,
+             head to the :ref:`doc_c_sharp` section.
```
tetrapod00, line 30:
```suggestion
             :ref:`doc_gdscript`. To learn more about C#, head to
             :ref:`doc_c_sharp`.
```
Could be like this to be more future-proof.

skyace65 (approved):

skyace65:
Finally got around to reviewing this, looks good to me! I think still using "section" is fine, I don't anticipate that being changing even in the long term, if it does we can make a change at that point.

skyace65:
Thanks!

PR #8718, opened 2024-01-04
Updated mob initialization in "Your first 3D game" tutorial to prevent unwanted rotation

omzz15:

PR #10666, opened 2025-02-10
Fix `look_at_from_position()` usage in Your first 2D game tutorial

Calinou:
The mob's orientation was previously shifted according to the player's height, which could lead to collision and movement issues that were difficult to diagnose.

- See https://github.com/godotengine/godot-docs-user-notes/discussions/118#discussioncomment-12113928.


aaronfranke (approved):

getting_started/first_3d_game/04.mob_scene.rst
```diff
@@ -310,7 +322,11 @@ Here is the complete ``mob.gd`` script for reference.
         {
             // We position the mob by placing it at startPosition
             // and rotate it towards playerPosition, so it looks at the player.
-            LookAtFromPosition(startPosition, playerPosition, Vector3.Up);
+            //
+            // Ignore the player's height, so that the mob's orientation is not slightly
+            // shifted if the mob spawns while the player is jumping.
+            Vector3 target = new Vector3(player_position.x, start_position.y, player_position.z);
```
tetrapod00, line 328:
```suggestion
        Vector3 target = new Vector3(playerPosition.X, startPosition.Y, playerPosition.Z);
```

tetrapod00 (changes requested):
Needs the C# fix but the motivation makes sense and the change looks right.

skyace65:
This would supersede #8718

aaronfranke:
PR #8718 is simpler and would accomplish the same thing, but I'm not sure if we want to encourage the practice of modifying function parameters.

skyace65:
Found another one this would supersede #7123

PR #10671, opened 2025-02-11
Remove outdated disadvantages to upgrading to Godot 4

tetrapod00:
- Godot 4 is no longer in "the early stages". However, we may want to add back in a bullet point about Godot 3.x/3.6's focus on stability, and that it is still a good choice *if you already have a 3.x project* that is shipped or near to shipping. I'm not sure if such projects would need us to tell them that, though 🙂 
- There are now plenty of Godot 4 tutorials.

This page and https://docs.godotengine.org/en/latest/about/release_policy.html#which-version-should-i-use-for-a-new-project do need a more extensive update, tweaking wording and recommendations. I split out these two bullet points because they seemed like clear removals.


Calinou (approved):

dsnopek (approved):

skyace65 (approved):

skyace65:
Thanks!

Issue #10674, opened 2025-02-12
Bisecting Regressions page out of date, including bad link


tetrapod00:
**Your Godot version:**
4.4
**Issue description:**
The link to the download mirror (https://downloads.tuxfamily.org/godotengine/) leads to a connection timeout error. I'm not sure if this problem is temporary or just on my end.

**URL to the documentation page:**
https://docs.godotengine.org/en/stable/contributing/workflow/bisecting_regressions.html#using-official-builds-to-speed-up-bisecting

PR #10676, opened 2025-02-12
Update Bisecting regressions page

tetrapod00:
Fixes https://github.com/godotengine/godot-docs/issues/10674.

### DIFF
--- a/contributing/workflow/bisecting_regressions.rst
+++ b/contributing/workflow/bisecting_regressions.rst
@@ -30,17 +30,7 @@ Before using Git's ``bisect`` command, we strongly recommend trying to reproduce
 the bug with an older (or newer) official release. This greatly reduces the
 range of commits that potentially need to be built from source and tested.
 You can find binaries of official releases, as well as alphas, betas,
-and release candidates `here <https://downloads.tuxfamily.org/godotengine/>`__.
-
-If you have experience with Godot 3.x and can reproduce an issue with Godot 4.0,
-we recommend trying to reproduce the issue in the latest Godot 3.x version (if
-the feature exhibiting the bug is present in 3.x). This can be used to check
-whether the issue is a regression in 4.0 or not.
-
-- If the issue **is present** in 3.x, then you'll need to check whether the issue
-  occurs in older 3.x versions as well.
-- If the issue is **not present** in 3.x, then you can try older 4.0 alphas and
-  betas to determine when the regression started.
+and release candidates `here <https://godotengine.org/download/archive/>`__.
 
 .. danger::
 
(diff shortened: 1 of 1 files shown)


contributing/workflow/bisecting_regressions.rst
```diff
@@ -116,23 +113,6 @@ following commit hashes depending on the version:
...
-            | grep 'from commit' \
-            | sed 's/^Built from commit \(.*\)\.$/\1/'
-    }
-
-Example usage:
-
-.. code-block:: shell
-
-    gd_snapshot_commit 4.0 beta4
-
```
tetrapod00, line 135: I'm not sure how to update this section, considering the tuxfamily page is down. I opted to remove rather than leaving a nonfunctional command in the docs. If we can update it, that's ideal.

contributing/workflow/bisecting_regressions.rst
```diff
@@ -116,23 +113,6 @@ following commit hashes depending on the version:
     3.1-stable
     3.0-stable
```
tetrapod00, line 114: This table should probably be updated or removed?

skyace65: I say remove it. It's really unnecessary.

skyace65 (approved):

contributing/workflow/bisecting_regressions.rst
```diff
@@ -81,57 +71,22 @@ Determine the commit hashes
...
+look something like ``v4.4.beta3.official [06acfccf8]``, and the commit hash is
+within the brackets, in this case ``06acfccf8``. You can click on the version
+information to copy it, including the commit hash.
+
+Alternately, you can browse the `Godot Archive
+<https://godotengine.org/download/archive/>`__, and find the commit hash within
+the release page linked from the **News** button.
+
+If you're using a stable release as the "good" or "bad" build, you can use the
+`release tag <https://github.com/godotengine/godot/tags>`__ as the commit hash,
```
tetrapod00, line 88: I'm not sure what the technically correct name for these is - they're called "tags" on github?

skyace65, line 88: From what I'm seeing on google just "tag" would be correct. So I'd recommend phrasing like this
```suggestion
If you're using a stable release as the "good" or "bad" build, you can find the
commit hash for that release in its `tag <https://github.com/godotengine/godot/tags>`__.
```

tetrapod00: That's true too, but the point is that you can use the tags directly in git bisect, like `git bisect bad 4.1.1-stable`.

I think the phrasing of the page before any changes might be technically wrong too, since the list of branch names/tags might not actually be "commit hashes"? You can use either `77dcf97` or `4.3-stable` and point to the same thing.

mhilbrunner: Tag is correct.

mhilbrunner: Tags are basically commit aliases. Human-friendly names assigned to a specific commit. And yeah, using tags here is fine.

tetrapod00: I did another pass on this whole section, adding a paragraph that lists the three options: commit hash directly, branch, or release tag. I think now it's both explicitly clear and correct?

contributing/workflow/bisecting_regressions.rst
```diff
@@ -81,57 +71,27 @@ Determine the commit hashes
...
+within the brackets, in this case ``06acfccf8``. You can click on the version
+information to copy it, including the commit hash.
+
+Alternately, you can browse the `Godot Archive
+<https://godotengine.org/download/archive/>`__, and find the commit hash within
+the release page linked from the **News** button.
+
+If you're using a stable release as the "good" or "bad" build, you can use the
+tag of that release directly, such as ``4.2-stable`` or ``4.2.1-stable``. A full
+list of release tags is listed `on GitHub
```
skyace65, line 92:
```suggestion
list of release tags is available `on GitHub
```

Feels weird to say that a list is listed somewhere. Other than that this change looks good!

skyace65 (changes requested):

skyace65 (approved):

contributing/workflow/bisecting_regressions.rst
```diff
@@ -81,57 +71,27 @@ Determine the commit hashes
...
+If you're using a pre-release build as the "good" or "bad" build, you can find
+the commit hash in the Project Manager in the lower-right corner, or in in the
+**Help > About Godot** dialog in the Godot editor. The version information will
+look something like ``v4.4.beta3.official [06acfccf8]``, and the commit hash is
+within the brackets, in this case ``06acfccf8``. You can click on the version
+information to copy it, including the commit hash.
+
+Alternately, you can browse the `Godot Archive
+<https://godotengine.org/download/archive/>`__, and find the commit hash within
+the release page linked from the **News** button.
```
bruvzg, line 88: Might be also worth mentioning https://godotengine.github.io/godot-interactive-changelog/, it's much more convenient for getting commit hashes for dev/beta releases (as well as PR list with individual commit hashes for each) than looking at each release page.

tetrapod00: Added some information on how to use that.

bruvzg (approved):

skyace65:
Thanks!

PR #10679, opened 2025-02-12
Clarify the global registration of named_classes 

thibault:
Hi,

This is my first constribution to the great Godot's documentation, I hope it will be useful.

While reading the doc about gdscript classes, I found the usage of the keyword `class_name` confusing. Notably, the fact that a named class becomes available globally does not seem intuitive to me, as it's not clearly stated.

Seing that gdscript draws some inspiration from Python, as a newcomer it seems reasonable to think that any class must be imported before being used, and that was my first assumption.

Here is a clarification suggestion for the documentation. Feel free to tell me if you think it is a useful addition.

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_basics.rst
+++ b/tutorials/scripting/gdscript/gdscript_basics.rst
@@ -1951,6 +1951,20 @@ If you want to use ``extends`` too, you can keep both on the same line::
 
     class_name MyNode extends Node
 
+Named classes are globally registered, hence they become available in other scripts without the need to `load` or `preload` them. For example::
+
+    # In a file 'game.gd'.
+
+    # If you uncomment the next line, you will trigger a warning:
+    # The variable "Character" has the same name as a global class defined in "character.gd".
+    # var Character = load("character.gd")
+
+    var player: Character
+
+    func new_game() -> void:
+        player = Character.new()
+
+
 .. note::
 
     Godot initializes non-static variables every time you create an instance,
--- a/tutorials/scripting/gdscript/gdscript_basics.rst
+++ b/tutorials/scripting/gdscript/gdscript_basics.rst
@@ -1951,19 +1951,24 @@ If you want to use ``extends`` too, you can keep both on the same line::
 
     class_name MyNode extends Node
 
-Named classes are globally registered, hence they become available in other scripts without the need to `load` or `preload` them. For example::
+Named classes are globally registered, which means they become available to use
+in other scripts without the need to ``load`` or ``preload`` them:
 
-    # In a file 'game.gd'.
+.. code-block:: gdscript
 
-    # If you uncomment the next line, you will trigger a warning:
-    # The variable "Character" has the same name as a global class defined in "character.gd".
-    # var Character = load("character.gd")
+    var player
 
-    var player: Character
-
-    func new_game() -> void:
+    func _ready():
         player = Character.new()
 
+Using the name of a named class as a variable name will cause a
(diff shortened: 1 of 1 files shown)


tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1951,6 +1951,20 @@ If you want to use ``extends`` too, you can keep both on the same line::
...
+    # If you uncomment the next line, you will trigger a warning:
+    # The variable "Character" has the same name as a global class defined in "character.gd".
+    # var Character = load("character.gd")
+
+    var player: Character
+
+    func new_game() -> void:
+        player = Character.new()
+
+
```
tetrapod00, line 1967:
```suggestion
Named classes are globally registered, which means they become available to use
in other scripts without the need to ``load`` or ``preload`` them:

.. code-block:: gdscript

    var player

    func _ready():
        player = Character.new()

Using the name of a named class as a variable name will cause a
:ref:`SHADOWED_GLOBAL_IDENTIFIER <class_ProjectSettings_property_debug/gdscript/warnings/shadowed_global_identifier>`
warning. For example, using ``Character`` as a variable name will shadow the
``Character`` class you just defined, and cause a warning:

.. code-block:: gdscript

    var Character

```
I rewrote it pretty extensively. From what I see, this PR is trying to communicate two things:
- Named classes are globally registered.
- Reusing the name of a global class as a variable name will cause a warning.

I split those out into two shorter and more direct examples. For the shadowing example, I used a generic `var Character` rather than confusing the example by making `Character` also be loading a script.

Additionally, I changed some documentation style things:
- Link to the class reference where this warning is documented (this will get better after https://github.com/godotengine/godot-docs/pull/10635).
- Use a different syntax for the code examples that uses GDScript highlighting.
- Use dynamic typing for the examples, since we're still early in the process of converting examples to use static typing, and the GDScript reference page should favor dynamic typing possibly even after that is done.
- In Sphinx/RST we use two backticks for code style not one

If I missed the intent of your PR or you think I missed anything in the rewrite, feel free to suggest additional changes

tetrapod00 (changes requested):

tutorials/scripting/gdscript/gdscript_basics.rst
thibault: @tetrapod00 This is much clearer that way. Thank you for your time.

tetrapod00 (approved):

tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1951,6 +1951,25 @@ If you want to use ``extends`` too, you can keep both on the same line::
...
+Using the name of a named class as a variable name will cause a
+:ref:`SHADOWED_GLOBAL_IDENTIFIER <class_ProjectSettings_property_debug/gdscript/warnings/shadowed_global_identifier>`
+warning. For example, using ``Character`` as a variable name will shadow the
+``Character`` class you just defined, and cause a warning:
+
+.. code-block:: gdscript
+
+    var Character
+
```
dalexeev, line 1972: Looks unrelated to me, we should write a separate section about name lookup and shadowing. As for the global registration clarification, that change looks ok to me.

thibault: I don't have any problem with that. I don't have enough knowledge about the subject to contribute anything meaningul about lookups and shadowing, though. Feel free to update the current PR as you see fit.

tetrapod00, line 1972:
```suggestion
```

tetrapod00: Removed the other section.
Made an issue to track adding the shadowing section.

skyace65 (approved):

PR #10680, opened 2025-02-12
GitHub Actions: Check URLs using lychee

### LINKED ISSUE #10674: Bisecting Regressions page out of date, including bad link

Calinou:
This makes it possible to be aware of dead links and replace them with other resources (such as archived versions) as soon as possible.

Note that class reference URLs not checked within this repository, as I plan to also integrate lychee into [godotengine/godot](https://github.com/godotengine/godot)'s CI (which has the source of the class reference files).

> [!NOTE]
>
> CI will fail until we resolve https://github.com/godotengine/godot-docs/issues/10674, so that we remove the `download.tuxfamily.org` links from the docs.

### DIFF
--- a/.lycheeignore
+++ b/.lycheeignore
@@ -1,0 +1,30 @@
+# Don't read Godot's own URL schemes as web URLs.
+user:\/\/.*
+res:\/\/.*
+uid:\/\/.*
+
+# Don't read WebSockets or TCP URLs as web URLs.
+ws(s?):\/\/.*
+tcp:\/\/.*
+
+# Security checks prevent checking the URLs of these websites automatically,
+# typically returning 403 errors.
+.*asecuritysite\.com.*
+.*intel\.com.*
+.*reddit\.com.*
+
+# Don't check URLs from these websites due to frequent rate limits (error 429) or timeouts.
+.*adobe\.com.*
+.*gamedevartisan\.com.*
+.*github\.com.*
+.*gnu\.org.*
+.*loopit\.dk.*
+.*meta\.com.*
+.*sourceforge\.io.*
+
+# Not a valid URL with the GET method, which lychee always sends.
+# Mentioned in the class reference.
+.*httpbin\.org\/post
+
+# Class reference mentions `example.com/index.php` in an example, which is 404.
+.*example\.com.*
--- a/.github/workflows/check_urls.yml
+++ b/.github/workflows/check_urls.yml
@@ -1,0 +1,38 @@
+name: 🌐 Check URLs
+on:
+  push:
+  pull_request:
+  schedule:
+    # Every day at 18:00 UTC.
+    # URLs can decay over time. Setting up a schedule makes it possible to be warned
+    # about dead links as soon as possible.
+    - cron: "0 18 * * *"
+
+jobs:
+  check-urls:
+    runs-on: ubuntu-24.04
+    steps:
+
+      - uses: actions/checkout@v4
+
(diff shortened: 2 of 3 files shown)


adamscott (approved):

.lycheeignore
```diff
@@ -0,0 +1,29 @@
...
+.*gnu\.org.*
+.*loopit\.dk.*
+.*meta\.com.*
+.*sourceforge\.io.*
+
+# Not a valid URL with the GET method, which lychee always sends.
+# Mentioned in the class reference.
+.*httpbin\.org\/post
+
+# Class reference metnions `example.com/index.php` in an example, which is 404.
```
tetrapod00, line 28:
```suggestion
# Class reference mentions `example.com/index.php` in an example, which is 404.
```

tetrapod00 (approved):
Approve the idea. Shader function replacement links are correct.

Can't review the lychee action implementation

skyace65:
I rebased this PR since tetrapod's PR was merged and now the URL check is failing on a different link. I'm not sure why though because the link works for me.

Calinou:
> I rebased this PR since tetrapod's PR was merged and now the URL check is failing on a different link. I'm not sure why though because the link works for me.

I've added `gamedevartisan.com` to exclusions. Some websites tend to block link checkers and will return various error codes (typically 429, but sometimes 403 or even 503 in this case).

skyace65 (approved):

skyace65:
@mhilbrunner I'd like your review before merging this since it's adding another check

mhilbrunner (approved):
I think something like this is a good idea, and I didn't spot anything in the workflow. Looks good to me!

Some things can be expanded, but I think we want to see this in action first & iterate. So, lets go ahead.

mhilbrunner:
Thank you!

brotheramai:
@Calinou 
This action has been failing since 2025. Perhaps github updated in a way that broke compatibility? Here's a sample of the failure output:

Run lycheeverse/lychee-action@v2
Run echo "$HOME/.local/bin" >> "$GITHUB_PATH"
Run # Remove any existing lychee binary to prevent conflicts
Run # Create a temporary directory for downloads and extraction
Downloading from: https://github.com/lycheeverse/lychee/releases/download/lychee-v0.23.0/lychee-x86_64-unknown-linux-gnu.tar.gz
Extracting lychee-x86_64-unknown-linux-gnu.tar.gz
lychee
docs/
docs/README.md
docs/TROUBLESHOOTING.md
docs/lychee.1
docs/PRE_COMMIT.md
Run # Install lychee from the temporary directory
Run /home/runner/work/_actions/lycheeverse/lychee-action/v2/entrypoint.sh
error: invalid value '.' for '--base <BASE>': Error with base dir `.` : Base must either be a URL (with scheme) or an absolute local path. See `--help` for more information. If you want to resolve root-relative links in local files, also see `--root-dir`.

For more information, try '--help'.

Calinou:
@brotheramai Thanks for the reminder! I've opened a PR to fix this:

- https://github.com/godotengine/godot-docs/pull/11993

There is still one outstanding error I didn't manage to fix though.

PR #10685, opened 2025-02-14
Remove unneeded dependencies from Fedora installation in Compiling for Linux/*BSD

Calinou:
Only a few packages are actually needed, as most libraries used by Godot are either embedded or `dlopen()`'d at runtime.
The new list matches what's present in https://github.com/godotengine/godot/pull/43377.

- See https://github.com/godotengine/godot/pull/43377#issuecomment-2126354623.


mhilbrunner (approved):

skyace65 (approved):
I made a Fedora 41 VM to test this. The command works perfectly, I was able to compile the engine and it runs.

PR #10687, opened 2025-02-15
Update first 3D game setup page images for 4.4

skyace65:
When I started working on this I thought I'd go through the whole tutorial at once like the first 2D game, that's not happening due to the sheer number of images that need to be replaced. Not even for being outdated but for how bad the compression and size is on them. Some of these really should never have been added.

In addition to images I've updated one line of text due to a UI change. I've also added a line stating that you may need to open a project from the project manager after importing it. That's happening for me on 4.4 when I build from the latest commit.

Due to that I'm only doing that game setup page for now, no idea if I'll do the rest before 4.4 because of how much work it is and there's other stuff I'd like to spend time on instead.


mhilbrunner (approved):

getting_started/first_3d_game/img/01.game_setup/07.collision_shape_warning.webp
tetrapod00, line 1: Maybe this is a larger discussion beyond the scope of this PR, but I feel like popup/dialog screenshots benefit from including the title bar.

Notably if you set the editor to single window mode, in 4.4beta the screenshot still looks like the existing screenshot:
Screenshot 2025-02-15 124351

If we choose to not include the titles in screenshots of dialogs (and right now it is really inconsistent across the docs whether we do or don't), IMO we should use a figure with a caption with the name of the popup, or mention the popup window title in the text of the page.

tetrapod00 (approved):
I checked the images and they all seem to be appropriate replacements, but I didn't build the page and check them in context.

getting_started/first_3d_game/img/01.game_setup/07.collision_shape_warning.webp
skyace65: Gonna be honest, I didn't even know the single window mode option was a thing. I've been taking out the title bar to avoid having something OS specific in the screenshots. That said I'm not sure how much use we get out of including it. And while I know this is still minor all things considered, it does save on screenshot size, both in terms of pixels and file size.

tetrapod00: > That said I'm not sure how much use we get out of including it.

Personally, I think it's pretty important context. When you are *using* Godot, the title of these popups contains important information. When you are reading the documentation, seeing the title of the popup is one thing that confirms you're following along correctly.

It's for similar reasons that we include the tab/title of docks like the Inspector or Scene docks in screenshots. While you *could* follow along well enough with a more cropped screenshot and accompanying text that is correct, seeing the title of the dock helps confirm that you're in the right place.

I believe that in some cases we don't include the title because certain screenshot tools crop to the content of the window and not the title (sharex included, which I use sometimes). When I took the screenshots for the [Project Settings](https://docs.godotengine.org/en/latest/tutorials/editor/project_settings.html) page update, I followed the observed convention that we often don't include the title, but I compensated by adding the window title in a caption. I think leaving out the window title entirely isn't ideal to make the docs most usable.

So, if we decide on a convention moving forward, IMO it should either be to include the title in the screenshot, or to exclude the title but include that information in the caption or surrounding text.

tetrapod00:
Thanks!

Issue #10692, opened 2025-02-15
Add `else if` to flow control statements in shading reference


GlitchCode24:
**Your Godot version:**
4.4.beta3
**Issue description:**
The `else if` statement is missing from flow control statements in shading reference.
**URL to the documentation page:**
[Shading Language documentation page ](https://docs.godotengine.org/en/stable/tutorials/shaders/shader_reference/shading_language.html#flow-control)

GlitchCode24:
I made a PR to fix the issue

tetrapod00:
Fixed by the linked PR

PR #10693, opened 2025-02-15
Add `else if` to shading reference 

GlitchCode24:
Fixes issue #10692

### DIFF
--- a/tutorials/shaders/shader_reference/shading_language.rst
+++ b/tutorials/shaders/shader_reference/shading_language.rst
@@ -554,7 +554,7 @@ Godot Shading language supports the most common types of flow control:
     // `if`, `else if` and `else`.
     if (cond) {
 
-    } else if (cond) {
+    } else if (other_cond) {
 
     } else {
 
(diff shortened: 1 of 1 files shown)


tutorials/shaders/shader_reference/shading_language.rst
```diff
@@ -551,9 +551,11 @@ Godot Shading language supports the most common types of flow control:
 
 .. code-block:: glsl
 
-    // `if` and `else`.
+    // `if`, `else if` and `else`.
     if (cond) {
 
+    } else if (cond) {
```
tetrapod00, line 557:
```suggestion
    } else if (other_cond) {
```

tetrapod00 (commented):
Makes sense. I think we should make sure that each example of flow control makes sense in isolation, so we shouldn't use `cond` for both cases.

tutorials/shaders/shader_reference/shading_language.rst
GlitchCode24: That makes sense, it's not the same condition

skyace65 (approved):

skyace65:

PR #10694, opened 2025-02-16
Switch terms *error* and *warning* in output panel debug docs.

harrykeightley:
In the output panel docs, *Error* and *Warning* were incorrectly given each-other's descriptions. This change switches the sub-titles for each section.

PR #10695, opened 2025-02-16
Reorder Output panel message categories to match order in editor

tetrapod00:
Followup to https://github.com/godotengine/godot-docs/pull/10694. After the descriptions were corrected, the listed order no longer matches the order in the editor. This PR returns the original order so that it matches the editor.


skyace65 (approved):

skyace65:
Thanks!

PR #10698, opened 2025-02-17
2D game: Use global position for mob spawn location

### LINKED ISSUE #6372: Godot's "Your first 2d game" tutorial seems to currently spawn enemies at the wrong locations (not on the real screen borders)

nickcernis:
Fixes #6372 by correcting the spawn location of creeps, accounting for the MobPath's position in the main scene.

Without this change, creeps spawn from an offset position around the origin 0,0, resulting in creeps appearing in the middle of the screen when they spawn from the right edge, killing the player without a chance for them to dodge.

Using global_position ensures creeps spawn on actual screen edges using the world position of the MobSpawnLocation instead of its relative position to its parent MobPath.

Docs page:
https://docs.godotengine.org/en/stable/getting_started/first_2d_game/05.the_main_game_scene.html

| Before (.position) | After (.global_position)
| - | - |
| Using position | Using global_position |

CleanShot 2025-02-17 at 19 21 00@2x

CleanShot 2025-02-17 at 19 21 18@2x

### DIFF
--- a/getting_started/first_2d_game/05.the_main_game_scene.rst
+++ b/getting_started/first_2d_game/05.the_main_game_scene.rst
@@ -218,7 +218,7 @@ Note that a new instance must be added to the scene using ``add_child()``.
         mob_spawn_location.progress_ratio = randf()
 
         # Set the mob's position to the random location.
-        mob.position = mob_spawn_location.position
+        mob.position = mob_spawn_location.global_position
 
         # Set the mob's direction perpendicular to the path direction.
         var direction = mob_spawn_location.rotation + PI / 2


AThousandShips:
This is not an appropriate change, see:
* https://github.com/godotengine/godot-docs/pull/8700

skyace65:
Please provide us with a zip file of your game project where you're seeing the issue with position.

nickcernis:
@skyace65 @AThousandShips Ah! That explains it — it's due to the x,y position of MobPath (visible in my screenshot above), which should be 0,0 as in the demo code. Thanks for checking in on this and for the link to #8700.

I think what's happening here (for me, the people in #6372 and #8700, and others like [this report in Reddit](https://www.reddit.com/r/godot/comments/80z47c/path_placing_creeps_in_center_of_scene/) from a user who says they stopped using Godot partly because they couldn't figure this issue out) is this:

1. The last instruction on the [main game scene page](https://docs.godotengine.org/en/stable/getting_started/first_2d_game/05.the_main_game_scene.html) before creating the Path2D is to make a Marker2D named StartPosition with position 240, 450.
2. We're then asked to, "add a [Path2D](https://docs.godotengine.org/en/stable/classes/class_path2d.html#class-path2d) node named MobPath as a child of Main". But if you accidentally add a node without selecting Main first, you end up with MobPath as a child of StartPosition, which already has a transform.
3. If you then try to correct this after drawing your path (when seeing the expected scene tree at the end of the page, for example) by reparenting to Main in the Scene Tree with drag-and-drop, you end up with a Path2D with position 240, 450 instead of 0,0 and the position logic breaks due to the offset.

So — totally user error and sorry to have missed that — but it's not obvious to a beginner that reparenting has a side-effect of changing the offset based on the new parent's transform relative to the previous parent.

Should I make a separate PR to change the steps to be more explicit since this has caught a few people out?

```diff
-Add a Path2D node named MobPath as a child of Main.
+Click the Main node in the Scene Tree, then add a child Path2D node named MobPath.
```

Or feel free to close this out, perhaps along with https://github.com/godotengine/godot-docs/issues/6372 since it sounds like that's invalid now.

tetrapod00:
A new PR with the clarification of what to select makes sense to me. You can mention this one in the description so the context is preserved

As a small tweak I suggest one of the following, for style consistency reasons (the dock/tab you click on in the editor is just called "Scene", not "Scene Tree", so it should either be called "Scene dock" or "scene tree" in lowercase):
```
Click the Main node in the Scene dock, then add a child Path2D node named MobPath.
```
or
```
Click the Main node in the scene tree, then add a child Path2D node named MobPath.
```

nickcernis:
@tetrapod00 Thanks for your guidance! PR at https://github.com/godotengine/godot-docs/pull/10699, set up to close this PR and the issue at https://github.com/godotengine/godot-docs/issues/6372.

Issue #10678, opened 2025-02-12
Specify that `get_contact_local_normal` is in global space


Eole211:
**Your Godot version:** 4.3

**Issue description:**
In the documentation of the `PhysicsDirectBodyState3D`, could it be possible to specify that the [get_contact_local_normal](https://docs.godotengine.org/en/stable/classes/class_physicsdirectbodystate3d.html#class-physicsdirectbodystate3d-method-get-contact-local-normal) method returns the normal is global space ? I wasted some time thinking "Oh there is local in the name, then it should returns the normal in local space", until I got some problem with the results, and also read the [get_contact_local_position](https://docs.godotengine.org/en/stable/classes/class_physicsdirectbodystate3d.html#class-physicsdirectbodystate3d-method-get-contact-local-position) doc which specify it is in global coordinate

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/stable/classes/class_physicsdirectbodystate3d.html

Damnae:
I'm having the same issue with `get_contact_collider_position` and `get_contact_local_position`:

get_contact_collider_position:
`Returns the position of the contact point on the collider in the global coordinate system.`
I know shapes and bodies, what is a collider? Specially with the next one;

get_contact_local_position:
`Returns the position of the contact point on the body in the global coordinate system.`
This one is about body, that's the only difference as per the description, but like `get_contact_local_normal` the method name and its description are conflicting and confusing.

In the end both methods appear to return the same value in global coordinates.

PR #10702, opened 2025-02-21
Clarify local normal is returned in global coordinate space for physicsdirectbodystate2d/3d

AidanCooper2003:
Fixes [#10678](https://github.com/godotengine/godot-docs/issues/10678).

Originally did not specify the coordinate space that the normal was returned in, which could lead to confusion as the function returns the local normal.


AThousandShips:

PR #10721, opened 2025-02-27
Add information about video seeks

berarma:
This depends on godotengine/godot#102360 being merged.

I've included a note about seek times that might seem too verbose. I wanted to make sure it's understood correctly.


skyace65:
Thanks!

PR #10727, opened 2025-02-28
Create xr full screen effects doc page

devloglogan:
Adds a doc page informing on how users can take asymmetric FOV into account in full screen quad ~~post-processing~~ shaders.

@BastiaanOlij mentioned that attempting to do this while reading from the screen texture is a big no-no so I also included a section at the bottom attempting to warn users about that.


BastiaanOlij:
Personally i feel we shouldn't be talking about post processing here. Post processing effects suggest that you modify the rendered output and that is where the big no-no's come in.

Vignettes' etc are not post processes. They are screen space effects that use a quad to render their output (actually a single triangle would be better but we don't have an easy primitive for that). So I think we need to come up with better naming.

In the same line, I don't think that referencing the original post processing document is a good idea. Again, we just set people on the wrong footing, they will read that, figure out they can do that in XR as well, and then when they are too far along find out that doing so is what is causing their performance to never get up to spec. By then it's too late.

devloglogan:
@BastiaanOlij understood, and now that I understand the term a bit better I fully agree! @dsnopek had mentioned something similar to me elsewhere. I'll rework this soon to swap that term out for something like "full screen effects" and also remove the reference to the other doc.

BastiaanOlij:
@devloglogan it does feel like nitpicking but I feel we can prevent ourselves a bit of hardship if we get that bit right :)

devloglogan:
Alright, I've opted for "full screen effects" as the term instead and removed the reference to the other doc entirely (which adds a touch more on the basic full screen quad setup).

tutorials/xr/xr_full_screen_effects.rst
```diff
@@ -0,0 +1,81 @@
+.. _doc_xr_full_screen_effects:
+
+XR full screen effects
+======================
+
+When adding custom full screen effects to your XR application, one valid approach is
+applying a full screen quad to the XR camera and applying effects to that quad's shader.
```
dsnopek, line 7:
```suggestion
When adding custom full screen effects to your XR application, one approach is
using a full screen quad and applying effects to that quad's shader.
```

tutorials/xr/xr_full_screen_effects.rst
```diff
@@ -0,0 +1,81 @@
...
+  render_mode depth_test_disabled, skip_vertex_transform, unshaded, cull_disabled;
+
+  void vertex() {
+	  POSITION = vec4(VERTEX.xy, 1.0, 1.0);
+  }
+
+  void fragment() {
+	  ALBEDO = vec3(0.0);
+	  ALPHA = dot(UV * 2.0 - 1.0, UV * 2.0 - 1.0) * 2.0;
+  }
```
dsnopek, line 26: I'm not sure about including the incorrect shader code in full, because folks may not read very closely and just copy-paste this.

I think it may be better to only show full shader code for the correct one, and maybe put comments calling out the part that adjusts for the asymmetric projection matrix? Saying like, "if you remove this, the effect won't be correctly centered")

tutorials/xr/xr_full_screen_effects.rst
```diff
@@ -0,0 +1,81 @@
...
+XR full screen effects
+======================
+
+When adding custom full screen effects to your XR application, one valid approach is
+applying a full screen quad to the XR camera and applying effects to that quad's shader.
+
+.. image:: img/xr_full_screen_effects_starting_quad.webp
+
+Here is a simple vignette shader that you might apply to the quad, darkening the edges
+of a user's screen:
```
dsnopek, line 12: If we're not linking to https://docs.godotengine.org/en/latest/tutorials/shaders/advanced_postprocessing.html, then I think we need to include some more information than this, explaining:

1. The `MeshInstance3D` node they need to add, the size of the quad mesh, and to put it as a child of the camera. Some of this is shown in the images, but I think we should have it in words too.
2. That we're making it full screen by setting the `POSITION` in the shader. (This ends up being implied in the "Applying the projection matrix" section, but I feel like it'd be best to state it up front)

tutorials/xr/xr_full_screen_effects.rst
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  void fragment() {
+	  ALBEDO = vec3(0.0);
+	  ALPHA = dot(UV * 2.0 - 1.0, UV * 2.0 - 1.0) * 2.0;
+  }
+
+
+However, when creating an effect that is centered straight ahead in the user's view
+(such as the above vignette effect), the end result may look incorrect in XR.
```
dsnopek, line 30: This is shown wonderfully in the picture, but I think we should also say in words, why it doesn't look correct in XR (that the center of the rendered output we are producing isn't necessarily the center of the user's view in XR, because our output will be deformed to match the lenses of the headset)

Also, I think it would also be cool to show a screenshot from inside the headset showing how the correct and incorrect versions look, in addition to the screenshots of the render target texture. Maybe it could be shown in a 2x2 grid, with the headset image above the render target image? Otherwise, it might not be entirely clear that the off-center effect on the render target is actually the correct one that will look centered in the headset

tutorials/xr/xr_full_screen_effects.rst
```diff
@@ -0,0 +1,81 @@
...
+.. note:: For more info on asymmetric FOV and its purpose, see this
+          `Meta Asymmetric Field of View FAQ <https://developers.meta.com/horizon/documentation/unity/unity-asymmetric-fov-faq/>`_.
+
+Limitations
+-----------
+
+Currently, full screen effects that require reading from the screen texture effectively disable all
+rendering performance optimizations in XR. This is because, when reading from the screen texture,
+Godot makes a full copy of the render buffer. Since this may create performance issues, it is recommended
+that custom effects be limited to per-pixel ones such as the above vignette shader.
```
dsnopek, line 81: I think it could be made a little clearer that this limitation _doesn't_ apply to the technique shown here. Having a "Limitations" section makes it sound like this is a limitation of what was shown above, rather than a limitation of a totally different way to approach doing this, which we didn't show :-)

devloglogan:
@dsnopek Thanks for the review! I believe I've addressed all of your suggestions, please take another look at the doc when you get a chance. :)

dsnopek (approved):
Thanks! This is looking good to me now :-)

skyace65 (changes requested):
Grammer and spelling should be good after these two changes. @BastiaanOlij did you want to look this over again?

BastiaanOlij (approved):
LGTM, nice writeup Logan! Indeed just need the small grammar fixes and its ready to merge IMHO.

tutorials/xr/xr_full_screen_effects.rst
```diff
@@ -0,0 +1,73 @@
...
+
+You can then add a shader to your quad to make it cover the screen. This is done by setting the
+vertex shader's ``POSITION`` built-in to ``vec4(VERTEX.xy, 1.0, 1.0)``.
+However, when creating an effect that is centered straight ahead in the user's view
+(such as a vignette effect), the end result may look incorrect in XR.
+
+Below shows captures of the right-eye view with a vignette shader, both from the headset and the render target itself.
+The left captures are an unmodified shader; the right captures adjust the full screen quad using the projection matrix.
+While the capture on the left is centered in the render target, it is off-center in the headset view.
+But, after applying the projecton matrix, we see that the effect is centered in the headset itelf.
```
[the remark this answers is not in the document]
devloglogan: Applied!

tutorials/xr/xr_full_screen_effects.rst
```diff
@@ -0,0 +1,73 @@
+.. _doc_xr_full_screen_effects:
+
+XR full screen effects
+======================
+
+When adding custom full screen effects to your XR application, one approach is
+using a full screen quad and applying effects to that quad's shader.
+As a child to your :ref:`XRCamera3D <class_XRCamera3D>`,
+add a :ref:`MeshInstance3D <class_MeshInstance3D>` node
```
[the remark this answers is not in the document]
devloglogan: Applied!

skyace65 (approved):

skyace65:
Thanks!

PR #10730, opened 2025-03-02
class_scriptextension.rst Github URL showed

XxDavid123xX:
Github_url showed

I'd seed that the ScriptExtension page Is Very undone but i do not have any idea to what put on  where, so i just show the githud URL to make easier to edit.


tetrapod00:
The link to edit this page is not shown because the class reference pages are edited in the main Godot repository, in this case https://github.com/godotengine/godot/blob/master/doc/classes/ScriptExtension.xml

PR #10732, opened 2025-03-03
Remove mention of Handbrake from playing_videos.rst

JormaRysky:
Handbrake removed support for OGG in 2009(!) so it is not valid recommendation of tool for the job. Downloaded Handbrake and confirmed it can't export Ogg Theora

### DIFF
--- a/tutorials/animation/playing_videos.rst
+++ b/tutorials/animation/playing_videos.rst
@@ -182,7 +182,7 @@ you should use a lossless or uncompressed format as an intermediate format to
 maximize the quality of the output Ogg Theora video, but this can require a lot
 of disk space.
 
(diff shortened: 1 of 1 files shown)


skyace65 (changes requested):
I downloaded handbrake and confirmed that converting to ``.ogg`` is not an option.

skyace65 (approved):

skyace65:

berarma:
Could we have this backported to 4.4 and 3.6 at least? Users keep trying Handbrake and they think the problem is in their end.

Also, we could add a mention for [Shutter Encoder](https://www.shutterencoder.com/), an opensource and cross-platform GUI that can write OGV files.

PR #10733, opened 2025-03-03
Update release policy for 4.4 release

akien-mga:
Should be cherry-picked for `4.3` at least.

For older branches, they all have even more outdated states. I would suggest we drop any dynamic content from any branch that's not `latest`/`stable`, so we don't need to constantly cherry-pick this.


akien-mga:
Cherry-picked for 4.3.

PR #10739, opened 2025-03-04
Add a reminder to free RID's in the Compute Shader tutorial

### LINKED ISSUE #8447: Add More Reminders to Free RID's
**Your Godot version:** 4.1

**Issue description:** Whenever a resource is created and identified with an RID, like when using RenderingServer or TextServer, it needs to be manually freed. While there's reminders to free them in the descriptions of functions that return RID's, there's a few more places I think should have reminders as well.

First, I think there should be a brief note at the end of the description of the RID class reference, reminding that the RID should be freed using whatever object created it. Second, in the compute shader tutorial, there's currently no information about freeing the shader's resources after it's completed running. I think there should at least be a note at the end about what resources need to be manually freed.

RID Page: https://docs.godotengine.org/en/stable/classes/class_rid.html#class-rid
Compute Shader Tutorial: https://docs.godotengine.org/en/stable/tutorials/shaders/compute_shaders.html

DevPoodle:
Closes #8447.

I've added a little section at the end of the compute shader tutorial to remind the reader to free RID's when they're finished using them. I also went through and removed some static type hints from the code blocks on this page to make it more consistent with the rest of the manual.

### DIFF
--- a/tutorials/shaders/compute_shaders.rst
+++ b/tutorials/shaders/compute_shaders.rst
@@ -163,7 +163,7 @@ which can be created using the :ref:`class_RenderingServer`:
  .. code-tab:: gdscript GDScript
 
     # Create a local rendering device.
-    var rd := RenderingServer.create_local_rendering_device()
+    var rd = RenderingServer.create_local_rendering_device()
 
  .. code-tab:: csharp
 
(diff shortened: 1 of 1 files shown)


DevPoodle:
I really messed up when trying to open this PR, and couldn't figure out how to fix it using Git. So, I closed it and opened it under a different PR: #10740, sorry about that.

PR #10740, opened 2025-03-04
Add reminder to free RID's in compute shader tutorial

DevPoodle:
Closes https://github.com/godotengine/godot-docs/issues/8447.

I've added a little section at the end of the compute shader tutorial to remind the reader to free RID's when they're finished using them. I also went through and removed some static type hints from the code blocks on this page to make it more consistent with the rest of the manual.

### DIFF
--- a/tutorials/shaders/compute_shaders.rst
+++ b/tutorials/shaders/compute_shaders.rst
@@ -366,6 +366,16 @@ the data and print the results to our console.
     GD.Print("Input: ", string.Join(", ", input));
     GD.Print("Output: ", string.Join(", ", output));
 
+Freeing memory
+------------------
+
+The ``buffer``, ``pipeline``, and ``uniform_set`` variables we've been using are
+each an :ref:`class_RID`. Because RenderingDevice is meant to be a lower-level
+API, RIDs aren't freed automatically. This means that once you're done using
+``buffer`` or any other RID, you are responsible for freeing its memory
+manually using the RenderingDevice's
+:ref:`free_rid()<class_RenderingDevice_method_free_rid>` method.
+
 With that, you have everything you need to get started working with compute
 shaders.
 


tutorials/shaders/compute_shaders.rst
```diff
@@ -366,6 +366,11 @@ the data and print the results to our console.
     GD.Print("Input: ", string.Join(", ", input));
     GD.Print("Output: ", string.Join(", ", output));
 
+Freeing memory
+------------------
+
+The ``buffer``, ``pipeline``, and ``uniform_set`` variables we've been using are each an :ref:`class_RID`. Because RenderingDevice is meant to be a lower-level API, RID's aren't freed automatically. This means that once you're done using ``buffer`` or any other RID object, you are responsible for freeing its memory manually using the :ref:`class_RenderingDevice` ``free_rid`` method.
```
tetrapod00, line 372:
```suggestion
The ``buffer``, ``pipeline``, and ``uniform_set`` variables we've been using are
each an :ref:`class_RID`. Because RenderingDevice is meant to be a lower-level
API, RIDs aren't freed automatically. This means that once you're done using
``buffer`` or any other RID object, you are responsible for freeing its memory
manually using the RenderingDevice's
:ref:`free_rid()<class_RenderingDevice_method_free_rid>` method.
```
- Break up long lines
- Link to relevant function
- RIDs not RID's

Optionally, we might not want to use the term "object" for a rid, since "object" can refer to specifically anything derived from Object:
```suggestion
The ``buffer``, ``pipeline``, and ``uniform_set`` variables we've been using are
each an :ref:`class_RID`. Because RenderingDevice is meant to be a lower-level
API, RIDs aren't freed automatically. This means that once you're done using
``buffer`` or any other RID, you are responsible for freeing its memory
manually using the RenderingDevice's
:ref:`free_rid()<class_RenderingDevice_method_free_rid>` method.
```

tutorials/shaders/compute_shaders.rst
```diff
@@ -163,7 +163,7 @@ which can be created using the :ref:`class_RenderingServer`:
  .. code-tab:: gdscript GDScript
 
     # Create a local rendering device.
-    var rd := RenderingServer.create_local_rendering_device()
+    var rd = RenderingServer.create_local_rendering_device()
```
tetrapod00, line 166: I would revert the changes to the code examples.

While we do currently recommend untyped GDScript in the documentation style guide, there is a lot of support among documentation maintainers to switch to typed GDScript when we can. So I don't think it makes sense to change some typed examples back to untyped if we may be changing them again.

If we do make the call to remove all typed GDScript for the time being, we should do so all at once like https://github.com/godotengine/godot/pull/99925 did.

tetrapod00 (approved):

thibault:
Hi. I stumbled upon this PR while searching about this topic. While this addition is very useful, I feel it would be interesting to add more details about the reason it is necessary to clear rids. Like, you have to clear rids, but *why*? Does the entire resource stays in memory forever? Or is there a limited amount of available rids?

thibault:
Another comment : On the `water_plane` demo on `godot-demo-projects`, [you can find this comment](https://github.com/godotengine/godot-demo-projects/blob/master/compute/texture/water_plane/water_plane.gd#L239C1-L239C92) related to rid freeing code:

> Note that our sets and pipeline are cleaned up automatically as they are dependencies :P

I find that rather confusing, so it seems there's a little more to it than just clearing every single created RID? (sorry if it outside the current PR initial scope).

DevPoodle:
@thibault Clearing RID's is important because by default their memory isn't cleared when they go out of scope, meaning you can very easily cause a memory leak when using them. As for the types of RID's that are cleared automatically, if you free the shader first, then I believe it also frees the pipeline and uniform sets. I think that would be better documented under the create_shader and free_rid methods than here though.

skyace65 (approved):

skyace65:

PR #10756, opened 2025-03-10
Style: Integrate new `#pragma once` syntax

Repiteo:
- See godotengine/godot#102298

With `#pragma once` being the new stylistic syntax utilized in the main repo, this aims to update the docs to reflect that

### DIFF
--- a/contributing/development/cpp_usage_guidelines.rst
+++ b/contributing/development/cpp_usage_guidelines.rst
@@ -94,11 +94,12 @@ Lambdas should be used conservatively when they make code effectively faster or
 simpler, and do not impede readability. Please ask before using lambdas in a
 pull request.
 
-``#pragma once`` directive
-~~~~~~~~~~~~~~~~~~~~~~~~~~
+``#ifdef``-based include guards
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-To follow the existing style, please use standard ``#ifdef``-based include
-guards instead of ``#pragma once`` in new files.
+Starting with 4.5, all files now use the ``#pragma once`` directive, as they
+improve readability and declutter macros. Use of ``#ifdef``-based include
+guards are now actively discouraged.
 
 ``try``-``catch`` blocks
 ~~~~~~~~~~~~~~~~~~~~~~~~
--- a/contributing/development/code_style_guidelines.rst
+++ b/contributing/development/code_style_guidelines.rst
@@ -203,18 +203,13 @@ Example:
     /* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
     /**************************************************************************/
 
-    #ifndef MY_NEW_FILE_H
-    #define MY_NEW_FILE_H
+    #pragma once
 
     #include "core/hash_map.h"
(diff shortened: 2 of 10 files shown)


contributing/development/cpp_usage_guidelines.rst
```diff
@@ -94,11 +94,12 @@ Lambdas should be used conservatively when they make code effectively faster or
 simpler, and do not impede readability. Please ask before using lambdas in a
 pull request.
 
-``#pragma once`` directive
-~~~~~~~~~~~~~~~~~~~~~~~~~~
+``#ifdef``-based include guards
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
tetrapod00, line 98:
```suggestion
``#ifndef``-based include guards
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

contributing/development/cpp_usage_guidelines.rst
```diff
@@ -94,11 +94,12 @@ Lambdas should be used conservatively when they make code effectively faster or
...
 
-To follow the existing style, please use standard ``#ifdef``-based include
-guards instead of ``#pragma once`` in new files.
+Starting with 4.5, all files now use the ``#pragma once`` directive, as they
+improve readability and declutter macros. Use of ``#ifdef``-based include
+guards are now actively discouraged.
```
tetrapod00, line 102:
```suggestion
Starting with Godot 4.5, all files now use the ``#pragma once`` directive, as they
improve readability and declutter macros. Use of ``#ifndef``-based include
guards are now actively discouraged.
```
Both suggestions optional.

These are technically `ifndef`-based not `ifdef`-based, right? But maybe calling them colloquially "ifdef-based" still makes sense?

skyace65:
I know technically this is about changes in 4.5. But is there any reason not to have it in the 4.4 docs? I don't do coding for the engine so maybe I'm off here, but I wouldn't expect people to automatically assume they need to use the master branch for engine development.

mhilbrunner (approved):

mhilbrunner:
> But is there any reason not to have it in the 4.4 docs?

There could be an argument to keep guidelines for a version to the state that applied to that versions, i.e., to have Godot 3 docs reflect Godot 3 guidelines. I think that argument is stronger for major versions however, I don't think the complexity is worth it for minor versions.

And kinda moot anyway considering we mostly decided to eventually move to one (current) set of contributor documentation and to move it out of the (user-facing) docs. I see no harm in cherrypicking these.

contributing/development/cpp_usage_guidelines.rst
mhilbrunner: I think `ifdef` is fine, IMO it refers more to the mechanism than the exact (inverted) keyword. You're technically very much correct though :)

contributing/development/cpp_usage_guidelines.rst
bruvzg: "ifdef" seems fine here, since it's referring to the language feature in general, and is not a specific code example.

bruvzg (approved):

skyace65:
Thanks!

PR #10072, opened 2024-10-11
AnimationTree documentation simplification and expansion

saabdo42:
Simplified AnimationTree language, added documentation for Advance Expression, made images webp and colorblind friendly with more descriptive names.

Fixes: https://github.com/godotengine/godot-docs/issues/9699
Fixes: https://github.com/godotengine/godot-docs/issues/7753
Fixes: https://github.com/godotengine/godot-docs/issues/7658

PR #10759, opened 2025-03-12
Update documentation on animation tree for godot 4

Jesusemora:
trying to improve the seriously outdated tutorial for animation tree. it is missing documentation on expressions and the images and information are from godot 3.

it still needs some work but it's a start.


Jesusemora:
sorry, found https://github.com/godotengine/godot-docs/pull/10072 that does the same thing.

PR #10769, opened 2025-03-15
Update C# HelloWorld code block

bk0704:
The original code block was in a "public 2DScript()" method with no return type(or void) what so ever. I edited the C# version to be in a initializer/constructor, just like how the GDScript version is like

### DIFF
--- a/getting_started/step_by_step/scripting_first_script.rst
+++ b/getting_started/step_by_step/scripting_first_script.rst
@@ -148,10 +148,10 @@ Add the following code to your script:
 
  .. code-tab:: csharp C#
 
-	public override void _Ready()
-	{
-		GD.Print("Hello, world!");
-	}
+    public override void _Ready()
+    {
+        GD.Print("Hello, world!");
+    }
 
 
 Let's break it down. The ``func`` keyword defines a new function named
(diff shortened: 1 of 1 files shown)


getting_started/step_by_step/scripting_first_script.rst
```diff
@@ -148,10 +148,10 @@ Add the following code to your script:
...
  .. code-tab:: csharp C#
 
-    public MySprite2D()
-    {
-        GD.Print("Hello, world!");
-    }
+	public override void _Ready()
+	{
+		GD.Print("Hello, world!");
+	}
```
AThousandShips, line 154: The indentation should be restored to use spaces

AThousandShips (dismissed):

getting_started/step_by_step/scripting_first_script.rst
bk0704: So like the indent should be 4 spaces instead of tab?

AThousandShips: Yes, as it was originally (and the GDScript example)

bk0704: I just replaced the tab indent with a 4 spaces idendent

raulsntos:
The `public MySprite2D()` method is the [constructor](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/constructors) of the class, that's why it doesn't have a return type. The C# class constructor is the equivalent to GDScript's `_init()`, changing it for `_Ready()` is not correct.

skyace65:
Closing per raulsntos' comment

PR #5254, opened 2021-09-14
Optimize PNG and SVG images using oxipng and svgcleaner

aaronfranke:
This PR optimizes the images in this repo. This decreases the image sizes, at the cost of increasing the Git history size.

The `oxipng` command used is `~/.cargo/bin/oxipng -o6 --strip all --zopfli --fix`.

I was unable to optimize the below SVG images. `svgcleaner` runs into an end of stream error when processing them.

```
tutorials/math/img/tutovec10.svg
tutorials/math/img/tutovec11.svg
tutorials/math/img/tutovec12.svg
```

PR #10770, opened 2025-03-16
Replace axes colors in image

### LINKED ISSUE #10765: Swapped colors of y,z axes in docs figure
**Your Godot version:**
4.4 - latest docs

**Issue description:**
Hi,  I have noticed in the image of the axes (euler angles) in Using 3D transform page,
the color of y and z are swapped in regards to their colors in the editor.

In editor, y is green and z is blue, which makes great sense, because it matches their aliases... but in the image y is blue and z is green.

should the image in the docs be fixed?😅

Thank you

https://docs.godotengine.org/en/latest/tutorials/3d/using_transforms.html

BLooperZ:
This fixes #10765
The colors in the modified axes image now matches the colors in editor

Modification process:
I have swapped the green and blue color channels in the image
then optimized it using the same command from #5254

Thank you


skyace65 (changes requested):
The change itself is good but please convert this image to a webp file. You can find information on doing that in our documentation here: https://docs.godotengine.org/en/stable/contributing/documentation/docs_image_guidelines.html

kleonc (commented):
Not only the axes colors don't match Godot, but also the orientation of the Z axis. The current image shows a left-handed coordinate system but Godot uses right-handed coordinate system. See: https://docs.godotengine.org/en/4.4/tutorials/3d/introduction_to_3d.html#coordinate-system.

Corrected:<br>yBTpgncaO|pmQJnqGdX2
-|-

And BTW the rotation direction indicators are easy to misinterpret (the solid/intersection part goes behind or in front of the given axis? :thinking: (answer: assuming X/Y rotation directions are shown correctly, the solid/intersection parts are behind the axes)). So would be nice if they could be improved. But that would be a bonus / shouldn't be necessary.

skyace65 (approved):

skyace65:

PR #10773, opened 2025-03-16
Specified return value of resize() relating to [this issue](https://g…

Monochrome-debug:
Made it so that return value of resize() is specified


AThousandShips:

Closing as this is not a valid change

PR #10780, opened 2025-03-18
Update Enum names, improve enum member names, and add related suggestion

Lexyth:
- Changed the grammatical number of all enum names to be singular.
- Removed the enum names from their members, (e.g.: Tile.TILE_BRICK -> Tile.BRICK).
- Added a suggestion for the reader to use singular case for enum names.

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_styleguide.rst
+++ b/tutorials/scripting/gdscript/gdscript_styleguide.rst
@@ -186,11 +186,11 @@ indentation level to distinguish continuation lines:
         "Job": "Mechanic",
     }
 
-    enum Tiles {
-        TILE_BRICK,
-        TILE_FLOOR,
-        TILE_SPIKE,
-        TILE_TELEPORT,
+    enum Tile {
+        BRICK,
+        FLOOR,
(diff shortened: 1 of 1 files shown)


Lexyth:
I'm a bit lost. 
I don't think this error was caused by anything in my PR specifically, so I've just been waiting for someone to approve/reject it. But now it's been a while and my other PRs from the same day have been merged already, so it feels like that error is blocking the process here. 
If someone could tell me whether I have to do something, it'd be appreciated.

dalexeev (approved):
Looks good to me. Please try to rebase the branch to fix the failed test.

skyace65 (approved):

skyace65:
Thanks! Don't worry about that build issue, it was the URL checker, you didn't cause the problem and we don't run it on every PR now.

Lexyth:
Oh, good.
I was pretty sure I didn't cause it, since I didn't touch anything URL related, nor change anything escape-codey or similar, but you never know. 
I think I tried rerunning the action and got the same result, but maybe I just misremember that. Will try to rebase if it happens again on another PR.

PR #10781, opened 2025-03-18
Clarify type-safety of as

Lexyth:
Changed the statement about the as keyword's type-safety to clarify that it is less type-safe than type hints.

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_styleguide.rst
+++ b/tutorials/scripting/gdscript/gdscript_styleguide.rst
@@ -1053,7 +1053,7 @@ that type will be used to infer the type of the var.
 
 .. note::
 
-    This option is considered less :ref:`type-safe<doc_gdscript_static_typing_safe_lines>` than type hints, as it silently casts the variable to null in case of a type mismatch at runtime, without an error/warning
+    This option is considered more :ref:`type-safe<doc_gdscript_static_typing_safe_lines>` than type hints, but also less null-safe as it silently casts the variable to null in case of a type mismatch at runtime, without an error/warning.
 
 **Bad**:
 
--- a/tutorials/scripting/gdscript/gdscript_styleguide.rst
+++ b/tutorials/scripting/gdscript/gdscript_styleguide.rst
@@ -1050,7 +1050,10 @@ that type will be used to infer the type of the var.
     @onready var health_bar := get_node("UI/LifeBar") as ProgressBar
     # health_bar will be typed as ProgressBar
 
-This option is also considered more :ref:`type-safe<doc_gdscript_static_typing_safe_lines>` than the first.
+
+.. note::
+
+    This option is considered less :ref:`type-safe<doc_gdscript_static_typing_safe_lines>` than type hints, as it silently casts the variable to null in case of a type mismatch at runtime, without an error/warning
 
 **Bad**:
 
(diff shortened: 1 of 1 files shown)


dalexeev:
I find your correction useful in a practical sense, but imprecise in terminology.

_Type safety_ means a guarantee that you won't get a type error at runtime. The `as` operator is more type safe, but also more error-prone.

The problem is that objects are nullable, _null safety_ is not guaranteed by the current type system for objects.

Lexyth:
> I find your correction useful in a practical sense, but imprecise in terminology.
> 
> _Type safety_ means a guarantee that you won't get a type error at runtime. The `as` operator is more type safe, but also more error-prone.
> 
> The problem is that objects are nullable, _null safety_ is not guaranteed by the current type system for objects.

I agree. I have edited it to clarify that.
https://github.com/Lexyth/godot-docs/blob/58c9fdb6b5e5dd3ded7963a7be2681e72e642181/tutorials/scripting/gdscript/gdscript_styleguide.rst?plain=1#L1056

tutorials/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -1050,7 +1050,10 @@ that type will be used to infer the type of the var.
     @onready var health_bar := get_node("UI/LifeBar") as ProgressBar
     # health_bar will be typed as ProgressBar
 
-This option is also considered more :ref:`type-safe<doc_gdscript_static_typing_safe_lines>` than the first.
+
+.. note::
+
+    This option is considered more :ref:`type-safe<doc_gdscript_static_typing_safe_lines>` than type hints, but also less null-safe as it silently casts the variable to null in case of a type mismatch at runtime, without an error/warning.
```
dalexeev, line 1056:
```suggestion
    This option is considered more :ref:`type-safe<doc_gdscript_static_typing_safe_lines>` than type hints, but also less null-safe as it silently casts the variable to ``null`` in case of a type mismatch at runtime, without an error/warning.
```

dalexeev (approved):

tutorials/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -1050,7 +1050,10 @@ that type will be used to infer the type of the var.
     @onready var health_bar := get_node("UI/LifeBar") as ProgressBar
     # health_bar will be typed as ProgressBar
 
-This option is also considered more :ref:`type-safe<doc_gdscript_static_typing_safe_lines>` than the first.
+
+.. note::
+
+    This option is considered more :ref:`type-safe<doc_gdscript_static_typing_safe_lines>` than type hints, but also less null-safe as it silently casts the variable to ``null`` in case of a type mismatch at runtime, without an error/warning.
```
AThousandShips, line 1056:
```suggestion
    This option is considered more :ref:`type-safe<doc_gdscript_static_typing_safe_lines>` than type hints,
    but also less null-safe as it silently casts the variable to ``null`` in case of a type mismatch at runtime,
    without an error/warning.
```
Break up long lines

Lexyth: Thanks! Didn't think of that, because I only had the end text in mind, which auto-wraps.

skyace65 (approved):
Checked for grammar

skyace65:
Thanks!

PR #10782, opened 2025-03-18
Fix misplaced bad example

Lexyth:
Moved the bad example for the inferred type for functions that return super types directly after its related statement.

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_styleguide.rst
+++ b/tutorials/scripting/gdscript/gdscript_styleguide.rst
@@ -1032,6 +1032,14 @@ the function's return type. For example, ``get_node()`` cannot infer a type
 unless the scene or file of the node is loaded in memory. In this case, you
 should set the type explicitly.
 
+**Good**:
+
+.. rst-class:: code-example-good
+
+::
+
+    @onready var health_bar: ProgressBar = get_node("UI/LifeBar")
+
 **Bad**:
 
 .. rst-class:: code-example-bad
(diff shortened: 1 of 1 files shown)


skyace65 (commented):
What about this example makes you think it's misplaced? The entire page has the good example shown first and the bad example shown second.

Lexyth:
> What about this example makes you think it's misplaced? The entire page has the good example shown first and the bad example shown second.

The `as` keyword represents an alternative to state the type explicitly other than type hints, but is placed before a bad example that ONLY uses a type hint.

It was like this:

Statement about explicitly stating type, useful for get_node().
Good example of using a type hint.
Statement about the `as` keyword.
Good example of using the `as` keyword.
Bad example if using a type hint.

The good and bad examples both regard explicit type stating, but the bad only features type hints, so is not related to the `as` keyword statement, which therefore can as well come afterwards, putting the related examples closer together while still staying within the section it belongs to (i.e.: explicit type stating). 

I'd say, it's confusing to add an alternative in the middle of a list of examples. It's not done anywhere else in the page, either. Explanations – of what was good/bad –, sure, they're related to the example, but an alternative doesn't fit there, as it's related to the topic the examples are about, not the examples themselves. 
Alternatives could fit directly inside an example, maybe, but that, also, is only a place for smaller syntax alternatives, not whole keywords (e.g.: an example using `var x = 5 # or const x = 5` to express that both can be used regarding topic).

skyace65 (approved):

skyace65:

PR #10797, opened 2025-03-25
Update setting_up_xr.rst to reflect Main node name

Daynvheur:
The screen shows the default node as `Main`, the C# code should reflect that.

### DIFF
--- a/tutorials/xr/setting_up_xr.rst
+++ b/tutorials/xr/setting_up_xr.rst
@@ -108,7 +108,7 @@ Next we need to add a script to our root node. Add the following code into this 
 
     using Godot;
 
-    public partial class MyNode3D : Node3D
+    public partial class Main : Node3D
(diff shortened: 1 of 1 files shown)


AThousandShips:
This makes this script not match the [later step](https://docs.godotengine.org/en/latest/tutorials/xr/a_better_xr_start_script.html) in the tutorial, which will be confusing

tetrapod00 (commented):
This change is unnecessary. For C# code examples, we default to the `MyNode3D` naming convention whenever the GDScript example doesn't name the class. In this case the GDScript version starts with `extends Node3D` and has no `class_name`. While `Main` is probably a good name for the class in your project, for the purposes of a script example it's better to follow our current docs convention for naming.

skyace65:
Closing per tetrapod's comment

PR #10798, opened 2025-03-25
Document known issues with HDR 2D alpha blending in Environment and post-processing

Calinou:
- This closes https://github.com/godotengine/godot/issues/80868
and https://github.com/godotengine/godot/issues/80912.

### DIFF
--- a/tutorials/3d/environment_and_post_processing.rst
+++ b/tutorials/3d/environment_and_post_processing.rst
@@ -705,6 +705,12 @@ There are 2 ways to use glow in 2D:
     ``canvas_item`` shaders, so it's recommend to use it when relevant either
     way.
 
+    There is a `known issue <https://github.com/godotengine/godot/issues/80868>`__
+    where enabling 2D HDR will change how alpha blending works. Sprites with low
+    opacity values generally become more visible, and font rendering will look
+    bolder due to the low-opacity pixels from the font antialiasing becoming
+    more visible. This also affects the editor's own rendering.
+
 .. _doc_environment_and_post_processing_using_glow_to_blur_the_screen:
 
 Using glow to blur the screen


skyace65 (approved):
I know we generally don't like documenting bugs but I'm fine with it in this case, especially since it links to the issue so we'll know when it's been fixed. I'd like to get at least one other approval before a merge though.

tetrapod00 (approved):
Soft **approval**. It's better than the status quo, and broadly I think it would be okay to link to known issues more than we currently do. We definitely **should not close** the linked issue.

In the documentation, IMO calling something a "known issue" implies that it is a bug and we would like to fix it. But the current discussion in https://github.com/godotengine/godot/issues/80868 from maintainers indicates that this is a known *limitation* that should be documented, not a bug to be fixed. The issue is labeled `discussion` and `documentation`. The correct way to resolve that would be to rewrite some of the comments in that issue into a version that fits the manual. Instead, this PR links to the issue, essentially using https://github.com/godotengine/godot/issues/80868#issuecomment-1989326624 (and other comments) as documentation, rather than rewriting it into the manual format. Again, okay with me as better than status quo, but not ideal.

I'd also be okay with not merging, because this:
- links to an explanation rather than rewriting it into manual format, which is unusual.
- the linked issue is not a bug, so is unlikely to be resolved soon, and therefore isn't a great fit for a "known issue" link from the docs.

skyace65 (changes requested):
After thinking this over some more I agree with tetrapod, though I'm giving this a hard no as is now. It should be explained as a limitation, and we shouldn't be linking to an engine issue, all info should be in the docs.

mhilbrunner:
Closing as per the above comment, I just merged https://github.com/godotengine/godot-docs/pull/10891. Thank you!

PR #10808, opened 2025-03-27
Fix C# example about interpolating with quaternions

raulsntos:
The `GetQuaternion` method was unexposed in 4.0. The example now uses the `Quaternion(Basis)` constructor which matches the GDScript example.

- See https://github.com/godotengine/godot-docs-user-notes/discussions/243#discussioncomment-12622914


skyace65 (approved):

skyace65:
Thanks!

PR #10811, opened 2025-03-28
CI: Check URLs on schedule only to reduce false positives

Calinou:
- Follow-up to https://github.com/godotengine/godot-docs/pull/10680.

Links occasionally fail due to network issues or other transient problems. This should not cause pull requests to fail checks.

The check is now performed weekly at a random-looking time on Fridays (when server load isn't too high).


skyace65 (approved):

skyace65:
Thanks!

PR #10820, opened 2025-03-30
Autoload nodes can be accessed in the editor if they're using tool scripts

ryevdokimov:
Autoload nodes are not an exception. If a tool script is attached to them, they also can be accessed. They become a child of the root node and can be used in the editor in the same way that they would be used during runtime. 

image


skyace65 (approved):
Tested and confirmed this change is accurate. You just have to reload a project after an autoload has ``@tool`` added to it for it to be accessible.

skyace65:

PR #10822, opened 2025-03-31
Link to class reference for each platform that can be exported to

Calinou:
Each page provides a list of all export options for the given platform, which can be searched through using <kbd>Ctrl + F</kbd>. While this information is also available in the editor by hovering export option names, it can't be quickly searched through.

- See https://github.com/godotengine/godot-docs-user-notes/discussions/15#discussioncomment-12664437.


skyace65 (approved):

skyace65:
Thanks!

PR #10631, opened 2025-02-06
Split C++ (godot-cpp) and GDExtension system info into separate categories, children of Scripting.

Ivorforce:
Supersedes #10617.

This PR specifically rearranges articles such that `Scripting` now has these children:

SCR-20250207-omjc

`C++ (godot-cpp)` and `The GDExtension System` are split off from one another because they cater to different target groups (godot-cpp for C++ users, GDExtension to language bindings implementers).
Other Languages is taken from the previous GDExtension article, and is targeted at people that want to code in other languages.

I kept the changes as minimal as possible, but rearranged some articles and added intros to make it flow better with the new structure.

PR #10827, opened 2025-04-01
Stop referring to GDExtension as experimental

dsnopek:
Per the discussion at the previous GDExtension meeting, we think it's time to stop saying that GDExtension is experimental :-)

I'm going to mark this as a DRAFT so that it doesn't cause conflicts with https://github.com/godotengine/godot-docs/pull/10631 - it'll be much easier to rebase this PR than that one

But we can still discuss these specific changes to the text


skyace65:
Are we considering GDExtension non experimental as of 4.4, or as of 4.5 when that releases?

dsnopek:
We discussed this at the GDExtension meeting on 2025-03-18, and I believe the consensus was that we were considering it non-experimental since 4.1 retroactively, because the only technical change we can point to as far as when it became non-experimental was the compatibility break in 4.1. Whereas if we were to say it became non-experimental in 4.4, it would be kind of artificial because nothing in particular changed in 4.4 in order to cross that threshold.

However, I'm curious what other folks think!

Relatedly, we're also planning to do a "GDExtension retrospective" blog post in order to "announce" that it's no longer experimental.

tutorials/scripting/gdextension/gdextension_cpp_example.rst
```diff
@@ -44,14 +44,12 @@ of Godot. GDExtensions will not work in older versions of Godot (only Godot 4 an
...
-    in later minor versions, but not vice-versa. For example, a GDExtension targeting Godot 4.1
-    should work just fine in Godot 4.2, but one targeting Godot 4.2 won't work in Godot 4.1.
-
-    However, GDExtension is currently *experimental*, which means that we may break compatibility
-    in order to fix major bugs or include critical features. For example, GDExtensions created
-    for Godot 4.0 aren't compatible with Godot 4.1
-    (see :ref:`updating_your_gdextension_for_godot_4_1`).
+    Usually, GDExtensions targeting an earlier version of Godot will work in later
+    minor versions, but not vice-versa. For example, a GDExtension targeting Godot 4.2
+    should work just fine in Godot 4.3, but one targeting Godot 4.3 won't work in Godot 4.2.
```
paddy-exe, line 49:
```suggestion
    GDExtensions targeting an earlier version of Godot should work in later
    minor versions, but not vice-versa. For example, a GDExtension targeting Godot 4.2
    should work just fine in Godot 4.3, but one targeting Godot 4.3 won't work in Godot 4.2.
```

I am not sure if having "Usually" in the front of the sentence is a good phrasing. "Usually" implies there are other factors but these aren't named here. If we state it that they "should" work in later versions it implies that if they don't -> it's a bug of either the extension or godot-cpp

paddy-exe (changes requested):
Just some minor language nitpick IMHO.

I also think 4.1 should be recognized as stable as David explained. We also cherry-picked a LOT of PRs back to it if that was possible.

tutorials/scripting/gdextension/gdextension_cpp_example.rst
dsnopek: Thanks! Updated in my latest push

paddy-exe (approved):
Looks good to me now 👍🏻

dsnopek:
Now that https://github.com/godotengine/godot-docs/pull/10631 is merged, I've rebased and taken this out of DRAFT

tutorials/scripting/cpp/about_godot_cpp.rst
```diff
@@ -71,18 +65,16 @@ godot-cpp (or another GDExtension system) isn't enough:
...
 
 For this reason, when creating GDExtensions, you may want to target the lowest version of
 Godot that has the features you need, *not* the most recent version of Godot. This can
 save you from needing to create multiple builds for different versions of Godot.
 
-However, GDExtension is currently *experimental*, which means that we may
-break compatibility in order to fix major bugs or include critical features.
-For example, GDExtensions created for Godot 4.0 aren't compatible with Godot
-4.1 (see :ref:`updating_your_gdextension_for_godot_4_1`).
+There is one exception to this: extensions targeting Godot 4.0 will _not_ work with
```
skyace65, line 76:
```suggestion
There is one exception to this: extensions targeting Godot 4.0 will **not** work with
```

tutorials/scripting/cpp/gdextension_cpp_example.rst
```diff
@@ -53,14 +53,12 @@ of Godot. GDExtensions will not work in older versions of Godot (only Godot 4 an
...
-
-    However, GDExtension is currently *experimental*, which means that we may break compatibility
-    in order to fix major bugs or include critical features. For example, GDExtensions created
-    for Godot 4.0 aren't compatible with Godot 4.1
-    (see :ref:`updating_your_gdextension_for_godot_4_1`).
+    GDExtensions targeting an earlier version of Godot should work in later
+    minor versions, but not vice-versa. For example, a GDExtension targeting Godot 4.2
+    should work just fine in Godot 4.3, but one targeting Godot 4.3 won't work in Godot 4.2.
+
+    There is one exception to this: extensions targeting Godot 4.0 will _not_ work with
```
skyace65, line 60:
```suggestion
    There is one exception to this: extensions targeting Godot 4.0 will **not** work with
```

skyace65 (changes requested):

tutorials/scripting/cpp/about_godot_cpp.rst
dsnopek: Thanks! Fixed in my latest push

skyace65 (approved):

skyace65:
Thanks!

PR #10832, opened 2025-04-03
Fix Export Enum Example in c_sharp_exports.rst ( Line 458)

DeerUwU:
changed line 458:
    public MyEnum MyEnum { get; set; }
to:
    public MyEnum MyEnumCurrent { get; set; }

because it throws an error if you reuse the already declared name


tetrapod00 (approved):
It does fix an error. "MyEnumCurrent" seems like a fine name, unless we have a more specific naming convention for cases like this.

(We could also avoid the naming problem entirely by using a more concrete example than "MyEnum" and "Thing1", but that seems like a larger change to the whole page we could make later.)

skyace65 (approved):

skyace65:

PR #10834, opened 2025-04-04
Fix description of `shadows_disabled` shader render mode

pixcai:


tetrapod00 (approved):
It's true. Casting shadows is controlled by `GeometryInstance3D.cast_shadow`. `shadows_disabled` and the BaseMaterial3D equivalent [disable_receive_shadows](https://docs.godotengine.org/en/stable/classes/class_basematerial3d.html#class-basematerial3d-property-disable-receive-shadows) only control the received shadows. The base material property is named better than the shader render mode.

You can verify this pretty easily in the editor with a scene like this, with the top two meshes in each stack having shadows disabled, but still casting shadows:
image

skyace65 (approved):

skyace65:

PR #10846, opened 2025-04-08
C#: Remove outdated bits defines

raulsntos:
These defines were removed in 4.0 but we missed them when we previously updated this documentation page.


skyace65 (approved):

skyace65:
Thanks!

PR #10853, opened 2025-04-10
Fix @export_custom hint format in gdscript_exports.rst

starsJuly:
Corrects the `@export_custom` example to use `"suffix:m"` instead of `"altitude:m"`.
The engine seems to define a suffix if the comma delimited hint string has a slice which starts with `suffix:`.

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_exports.rst
+++ b/tutorials/scripting/gdscript/gdscript_exports.rst
@@ -475,7 +475,7 @@ For example, this exposes the ``altitude`` property with no range limits but a
 
 ::
 
-    @export_custom(PROPERTY_HINT_NONE, "suffix:m") var altitude: Vector3
+    @export_custom(PROPERTY_HINT_NONE, "suffix:m") var altitude: float
 
 The above is normally not feasible with the standard ``@export_range`` syntax,
 since it requires defining a range.
--- a/tutorials/scripting/gdscript/gdscript_exports.rst
+++ b/tutorials/scripting/gdscript/gdscript_exports.rst
@@ -475,7 +475,7 @@ For example, this exposes the ``altitude`` property with no range limits but a
 
 ::
 
-    @export_custom(PROPERTY_HINT_NONE, "altitude:m") var altitude: Vector3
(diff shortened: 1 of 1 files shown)


tetrapod00 (approved):
Thanks! The original just looks like a typo.

tutorials/scripting/gdscript/gdscript_exports.rst
```diff
@@ -475,7 +475,7 @@ For example, this exposes the ``altitude`` property with no range limits but a
 
 ::
 
-    @export_custom(PROPERTY_HINT_NONE, "altitude:m") var altitude: Vector3
+    @export_custom(PROPERTY_HINT_NONE, "suffix:m") var altitude: Vector3
```
skyace65, line 478:
```suggestion
    @export_custom(PROPERTY_HINT_NONE, "suffix:m") var suffix: Vector3
```

So, admittedly this is something I don't understand well, but I checked the class reference and the var is named "suffix" in the example there. Is that necessary or no?

tetrapod00: The variable should be named altitude in this page.

There is no need to name the variable "suffix" to have the suffix export hint work. The current example in the class reference does compile, but arguably it's confusing to name the variable "suffix":
```
@export_custom(PROPERTY_HINT_NONE, "suffix:m") var suffix: Vector3
```

I'd suggest changing the class reference example to something like:
```
@export_custom(PROPERTY_HINT_NONE, "suffix:m") var altitude: Vector3
```
```
@export_custom(PROPERTY_HINT_NONE, "suffix:m") var length: Vector3
```
or this, if we *must* include "suffix" in the name:
```
@export_custom(PROPERTY_HINT_NONE, "suffix:m") var suffix_example: Vector3
```

tetrapod00: Wait, actually it's weird that in this page the variable is called "altitude" but it's a Vector3, that likely also comes from copy-pasting the other example.
While we're here let's change this page to:
```suggestion
    @export_custom(PROPERTY_HINT_NONE, "suffix:m") var altitude: float
```

skyace65 (approved):

skyace65:

PR #10854, opened 2025-04-10
Add C# integer casts where needed in Your first 2D game

Calinou:
- See https://github.com/godotengine/godot-docs-user-notes/discussions/118#discussioncomment-12793095.

### DIFF
--- a/getting_started/first_3d_game/04.mob_scene.rst
+++ b/getting_started/first_3d_game/04.mob_scene.rst
@@ -202,7 +202,7 @@ We got a random position, now we need a ``random_speed``. ``randi_range()`` will
         // ...
 
         // We calculate a random speed (integer).
-        int randomSpeed = GD.RandRange(MinSpeed, MaxSpeed);
+        int randomSpeed = (int)GD.RandRange(MinSpeed, MaxSpeed);
         // We calculate a forward velocity that represents the speed.
         Velocity = Vector3.Forward * randomSpeed;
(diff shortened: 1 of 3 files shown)


raulsntos:
These shouldn't be needed. The `MinSpeed` and `MaxSpeed` properties are defined as `int` in the tutorial from what I can see, so the `int GD.RandRange(int, int)` overload would be used and that already returns `int`.

As for `var multiplier = GD.RandRange(.8, 1.2);` I don't think it needs a cast either, since the `multiplier` variable is only used to add it to an untyped Godot Array.

tetrapod00:
Agree, these don't look needed. In fact if you add the cast, your editor might suggest removing them (I get an [IDE0004](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0004) suggestion in VSCode).

The original docs comment says:
> GD.RandRange() requires an explicit cast so it would be great to add [these casts]

But I think they may be extrapolating from the earlier cast from C# double to float, which *is* required because `RotateY()` takes a float argument:
https://github.com/godotengine/godot-docs/blob/e87d427253dd4d0ab71b0ada8fd17060515b4621/getting_started/first_3d_game/04.mob_scene.rst?plain=1#L316

MetalWhine:
Cheers mate, seems like I used the wrong override by making my `MinSpeed` & `MaxSpeed` properties as floats and extrapolated from the mentioned previous section of tutorial.

Issue #8929, opened 2024-02-08
Add short section on git lfs


jtbourke:
**4.2.1:**

**Version Control chapter covers gitignore but doesn't mention git lfs setup, which is also important:**

[**URL to the documentation page (if already existing):**](https://docs.godotengine.org/en/stable/tutorials/best_practices/version_control_systems.html#files-to-exclude-from-vcs)

I just asked a question on the forum and then realized other people might have the same question and it might be nice to add this to the docs: https://forum.godotengine.org/t/godot-binary-file-types-for-git-lfs-tracking/46681

The information on file types is found in the documentation, but spread out.  Everyone who uses git lfs will need the same list of file types so it would be pretty simple to just provide a .gitattributes file that will work for everyone, and to provide a link to that in the documentation.

Maybe this is already provided somewhere and I missed it.  I'm new to Godot.

I did find this list of binary files which is helpful for source art, but not game specific, and it is missing some of the files that Godot imports, such as dds:
https://confluence.atlassian.com/bbkb/add-commonly-used-binaries-to-git-lfs-1188409676.html

Jim-Bourke:
Based on the comment by Calinou in the forum thread, these Godot-specific files are binary and therefore they should be lfs tracked:

- scn (not tscn)
- res (not tres)
- material
- anim
- mesh

This demonstrates the need for the documentation.  If I hadn't asked the question and if Calinou hadn't been willing to assist, I don't know how I would discover those file types without discovering them during use.

kostabekre:
A command to add Godot specific binaries to git LFS:
```bash
 git lfs track {"*.scn","*.res","*.material","*.anim","*.mesh"}
```

ash-dvlpr:
I too agree that this information should be listed somewhere inside the documentation.

PR #10857, opened 2025-04-12
Add Git LFS section

jaydensipe:
Closes https://github.com/godotengine/godot-docs/issues/8929 

I've recently added Git LFS to my project and thought I could write this as the setup is still fresh in my mind. I'm still new to Git LFS, so this might not contain all the _nuances_, but could still be useful to have.

### DIFF
--- a/tutorials/best_practices/version_control_systems.rst
+++ b/tutorials/best_practices/version_control_systems.rst
@@ -82,3 +82,95 @@ It is better to set this option as:
 Creating version control metadata using the project manager or editor will
 automatically enforce LF line endings using the ``.gitattributes`` file.
 In this case, you don't need to change your Git configuration.
+
+Git LFS
+-------
+
+Git LFS (Large File Storage) is a Git extension that allows you to manage large
+files in your repository. It replaces large files with text pointers inside Git,
+while storing the file contents on a remote server. This is useful for
+managing large assets, such as textures, audio files, and 3D models, without
+bloating your Git repository.  
+
+.. note::
+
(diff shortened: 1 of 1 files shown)


skyace65 (commented):
Grammar and spelling wise this looks good. However I'd like to get consensus from other maintainers before a merge. Unlike the existing section of this page none of this is Godot specific except for the example gitattributes files. And even that isn't something people strictly need to adhere to like the excluded files and folders at the top of the page. Some may choose to use LFS for image files like pngs, and others may choose not to, there's no technical reason you have to do it one way.

tetrapod00:
I'm okay with including this, even though it is only marginally in scope for the *Godot* docs. Including an example `.gitattributes` file also seems okay as a starting point (the text already clarifies it is a *starting point*). Maybe add another sentence explaining why these types are chosen?

jaydensipe:
> Maybe add another sentence explaining why these types are chosen?

Done!

jaydensipe:
Should we include a warning about potential costs with Git LFS due to file storage, or would the official docs suffice for something like this?

mhilbrunner (approved):
I'm somewhat hesitant to add more docs for external, non-Godot-specific tooling, even though I don't particularly like slippery slope arguments, we need to try to avoid scope creep.

However, seeing as (a tiny bit) of this configuration is actually Godot specific and saves Godot users some time, and that searching for `git lfs godot` did turn up a *few*, but not a ton of specific tutorials, I think I am fine with merging this for now, unless we have some better authority to link to.

Thanks for writing it!

skyace65 (approved):

skyace65:

PR #10863, opened 2025-04-14
Remove outdated #pragma once note, clarify auto usage.

bruvzg:
- Remove outdated `#pragma once` note, it is now preferred style.
- Changes `auto` notes to reflect actual state of the codebase (it is used a lot in two mentioned scenarios).


contributing/development/cpp_usage_guidelines.rst
```diff
@@ -84,8 +84,26 @@ Keep in mind hover documentation often isn't readily available for pull request
 reviewers. Most of the time, reviewers will use GitHub's online viewer to review
 pull requests.
 
-We chose to forbid ``auto`` instead of allowing it on a case-by-case basis to
-avoid having to decide on difficult edge cases. Thank you for your understanding.
+``auto`` keyword is can be used in some special cases, like C++ lambda or Objective-C block
```
skyace65, line 87:
```suggestion
The ``auto`` keyword can be used in some special cases, like C++ lambda or Objective-C block
```

skyace65 (changes requested):
There's an existing PR open for pragma once that covers more #10756. Any review on that would be appreciated,

Other than the one issue I found the auto keyword stuff looks good.

contributing/development/cpp_usage_guidelines.rst
```diff
@@ -84,8 +84,26 @@ Keep in mind hover documentation often isn't readily available for pull request
...
+definitions and C++ templates. Please ask before using ``auto`` in a pull request.
```
AThousandShips, line 88: Unsure what is meant by asking to use it here, below we say we forbid all other uses, but that we allow it for lambdads, Objective-C block definitions, and templates

Does we mean "please ask before using `auto` for these cases", or "please ask before using `auto` in other cases"?

IMO we should suggest asking for the template case, and having the general lambda/block cases be assumed okay (with reviewers correcting specific cases, for example where it should be `std::function<>` or similar)

---

I'd say to clarify here, as I'd say we don't want people asking about using `auto` in the cases we already forbid, and that we also don't want noise over normal cases, for example with lambdas where the use is fully normalized

i.e. suggest asking when using it for template cases exclusively

bruvzg: Changed it `Please ask before using templates with auto in a pull request`, we have `Please ask before using lambdas in a pull request.` in the next section, so it's redundant for lambdas.

> I'd say to clarify here, as I'd say we don't want people asking about using auto in the cases we already forbid, and that we also don't want noise over normal cases, for example with lambdas where the use is fully normalized

Yes, I agree. Lambdas are good case to use it, templates are special cases and should be reviewed individually, and the rest should be forbidden.

bruvzg:
> There's an existing PR open for pragma once that covers more https://github.com/godotengine/godot-docs/pull/10756.

Reverted this part.

AThousandShips (approved):

mhilbrunner (approved):

Calinou (approved):

skyace65 (approved):

skyace65:
Thanks!

PR #10864, opened 2025-04-14
Several minor XR page improvements

skyace65:
Makes several changes on a few XR pages

- Update setting up XR Tools page to reflect that it is available on the asset library for Godot 4.x
- Change see the "chapter" on hand tracking to page.
- Add a link to the binding modifiers section of the action map page on the XR Settings page
- Add a new image for enabling OpenXR settings. The image was updated in a PR that I made for the settings page, I was unaware at the time that it was also used in two other places and it doesn't fit now.


mhilbrunner:
Thanks!

PR #10867, opened 2025-04-15
Document sub-pixel morphological antialiasing (SMAA)

MJacred:
* For Godot 4.5, depends on merging https://github.com/godotengine/godot/pull/102330


tetrapod00:
The demo for generating the images is https://github.com/Calinou/godot-antialiasing-comparison, see this comment on the page:
https://github.com/godotengine/godot-docs/blob/f8ce2c9a0a35e778f3029213c589049d5dfd9386/tutorials/3d/3d_antialiasing.rst?plain=1#L6-L8

tutorials/3d/3d_antialiasing.rst
```diff
@@ -295,33 +316,34 @@ Antialiasing comparison
...
+| Compatibility            | ✔️ Yes                   | ❌ No                    | ❌ No                    | ❌ No                    |  ❌ No                   | ✔️ Yes                   | ❌ No                    |
++--------------------------+--------------------------+--------------------------+--------------------------+--------------------------+--------------------------+--------------------------+--------------------------+
 
 
 .. [1] MSAA does not work well with materials with Alpha Scissor (1-bit transparency).
        This can be mitigated by enabling ``alpha antialiasing`` on the material.
 .. [2] TAA/FSR2 transparency antialiasing is most effective when using Alpha Scissor.
 .. [3] SSAA has some blur from bilinear downscaling. This can be mitigated by
        using an integer scaling factor of ``2.0``.
+.. [4] SMAA 1x introduces less blurriness than FXAA at a slightly higher cost than FXAA.
```
tetrapod00, line 349: I'm not sure if this should be a footnote in the table. It's the kind of relatively small difference that IMO should be made in the section about SMAA earlier in the page. SMAA and FXAA both have a "low" performance cost by the standards of this table.

I do understand the desire to differentiate between the two postprocessing techniques, but I think if we start adding footnotes for every comparison the table loses it's value as a succinct summary. I'd like to hear what other maintainers think though.

MJacred: I fully understand your concern.

Perhaps it is agreeable to introduce new adjectives; proposals:
* Added blur: Light / Less / Lowest / Minimal / Minimum / Reduced
* Performance cost: Low-Medium / Mild

Or use "Low" for "Performance cost" and change cost value for `FXAA` to "Very Low"

tetrapod00: I approve of "very low" and "low" for the performance cost FXAA and SMAA. I'd prefer adding more distinctions at the "very low" and "very high" levels rather than the "low-medium" or "medium-high" levels for clarity's sake.

However, can you (at least briefly/casually) test the performance difference between FXAA and SMAA, and verify that the performance difference is actually big enough to justify using both "very low" and "low"? If the difference is very minimal, I think it would be more appropriate to keep using "low" for both.

For added blur, I approve of adding at least one more descriptor. I would avoid "lowest" IMO because the lowest amount is "none" 🙂. For blur, I think I would personally go with one of these (preference on the second one):
- "none", "less", "some", "more"
- **"none", "low", "some", "high"**

The "more" or "high" category might not be needed.

This would require re-evaluating the blurriness of the existing options to categorize them in the new system (most should remain as "some"). There should be enough information in the page to do so, I mostly used the info already on the page when I added the table.

MJacred: > I would avoid "lowest" IMO because the lowest amount is "none" 🙂

that… makes sense

---

For blur, your preferred counter-suggestion makes sense to me.

---

> However, can you (at least briefly/casually) test the performance difference between FXAA and SMAA […]

I used Calinou's project for testing the performance, but FXAA does not appear in the visual profile, SMAA appeared (I turned off the viewport 1 which shows the same w/o antialiasing), and SMAA had most of the time 0.09 ms GPU, and randomly spikes to ~1.25 ms. FXAA cannot be tracked.

Afterwards, I tested the FPS demo, and when using FXAA, the max FPS on my computer is 49/50 FPS. For SMAA, I get 47/48 FPS.  
Funny (sarcasm) fact: I reset 3D scaling mode from FSR 2 to default Bilinear, and it still appeared as "FSR2" in the the visual profiler consuming GPU and CPU. Though I didn't see it in the other project.

Here some screenshots of the stats from Calinou's project: first FXAA, then SMAA
fxaa
smaa

tetrapod00: > Afterwards, I tested the FPS demo (pressed "Play" so it was the actual game), and when using FXAA, the max FPS on my computer is 49/50 FPS. For SMAA, I get 47/48 FPS.

I know FPS comparisons are not that robust, but IMO this seems close enough that we could still call both SMAA and FXAA "Low" performance cost for the purposes of the table. But, calling FXAA "very low" also seems reasonable, so I'm happy with the current state of the PR.

I'm going to leave this conversation unresolved in case anyone else wants to see the context when this gets further review.

Calinou (approved):
Should be good to merge after adding a note about mixed SMAA + TAA/FSR2 usage, as mentioned above.

MJacred:
Tip was added.

I just remembered reading in some Godot issue or PR that when combined with FSR2 it may yield worse results at higher cost and it's better to just use FSR2.  
Sadly I can't find the source anymore… Is that true?

skyace65:
Thanks!

PR #10868, opened 2025-04-15
Remove outdated reference to fog transmission in Environment and post-processing

Calinou:
This was a 3.x feature that wasn't reimplemented as-is in 4.x. Instead, sun scatter and aerial perspective supersede it.

- See https://github.com/godotengine/godot-docs-user-notes/discussions/417#discussioncomment-12821268.


macduy:
Thanks for updating the docs! Is there an equivalent option for light transmittance? I have a omni light placed far away that I'd like to be more visible through the fog.

skyace65:
Thanks!

PR #10873, opened 2025-04-18
Add a table listing STL datatypes and their Godot equivalents in C++ usage guidelines

Calinou:
This is more comprehensive than the previous advice.

@Ivorforce Could you take a look at this? There are probably some inaccuracies in here, as I'm not fully up-to-date on which scenarios are most optimal for each datatype. I also don't know whether OAHashMap, AHashMap and HashMap all preserve insertion order or only some of them (the class comments don't make an explicit notice of this).

### DIFF
--- a/contributing/development/cpp_usage_guidelines.rst
+++ b/contributing/development/cpp_usage_guidelines.rst
@@ -53,18 +53,72 @@ as Godot provides its own data types (among other things).
 See :ref:`doc_faq_why_not_stl` for more information.
 
 This means that pull requests should **not** use ``std::string``,
-``std::vector`` and the like. Instead, use Godot's datatypes as described below:
+``std::vector`` and the like. Instead, use Godot's datatypes as described below.
+A 📜 icon denotes the type is part of :ref:`Variant <doc_variant_class>`. This
+means it can be used as a parameter or return value of a method exposed to the
+scripting API.
 
-- Use ``String`` instead of ``std::string``.
-- Use ``Vector`` instead of ``std::vector``. In some cases, ``LocalVector``
-  can be used as an alternative (ask core developers first).
-- Use ``Array`` instead of ``std::array``.
-
-.. note::
-
-    Godot also has a List datatype (which is a linked list). While List is already used
-    in the codebase, it typically performs worse than other datatypes like Vector
-    and Array. Therefore, List should be avoided in new code unless necessary.
++------------------------+--------------------------+---------------------------------------------------------------------------------------+
+| Godot datatype         | Closest C++ STL datatype | Comment                                                                               |
++========================+==========================+=======================================================================================+
+| ``String`` 📜          | ``std::string``          | **Use this as the "default" string type.** ``String`` uses UTF-32 encoding            |
+|                        |                          | to improve performance thanks to its fixed character size.                            |
++------------------------+--------------------------+---------------------------------------------------------------------------------------+
+| ``StringName`` 📜      | ``std::string``          | Uses string interning for fast comparisons. Use this for static strings that are      |
+|                        |                          | referenced frequently and used in multiple locations in the engine.                   |
++------------------------+--------------------------+---------------------------------------------------------------------------------------+
+| ``Vector``             | ``std::vector``          | **Use this as the "default" vector type.** Uses copy-on-write (COW) semantics.        |
+|                        |                          | This means it's generally slower but can be copied around almost for free.            |
++------------------------+--------------------------+---------------------------------------------------------------------------------------+
+| ``LocalVector``        | ``std::vector``          | Closer to ``std::vector`` in semantics. In most situations, ``Vector`` should be      |
+|                        |                          | preferred.                                                                            |
++------------------------+--------------------------+---------------------------------------------------------------------------------------+
+| ``Array`` 📜           | ``std::vector``          | Values can be of any Variant type. No static typing is imposed.                       |
+|                        |                          | Uses shared reference counting, similar to ``std::shared_ptr``.                       |
++------------------------+--------------------------+---------------------------------------------------------------------------------------+
+| ``TypedArray`` 📜      | ``std::vector``          | Subclass of ``Array`` but with static typing for its elements.                        |
+|                        |                          | Not to be confused with ``Packed*Array``, which is internally a ``Vector``.           |
++------------------------+--------------------------+---------------------------------------------------------------------------------------+
+| ``Packed*Array`` 📜    | ``std::vector``          | Alias of ``Vector``, e.g. ``PackedColorArray = Vector<Color>``.                       |
+|                        |                          | Only a limited list of packed array types are available                               |
+|                        |                          | (use ``TypedArray`` otherwise).                                                       |
++------------------------+--------------------------+---------------------------------------------------------------------------------------+
+| ``List``               | ``std::list``            | Linked list type. Generally slower than other array/vector types. Prefer using        |
+|                        |                          | other types in new code, unless using ``List`` avoids the need for type conversions.  |
++------------------------+--------------------------+---------------------------------------------------------------------------------------+
+| ``Span``               | ``std::span``            | Represents read-only access to a contiguous array without needing to copy any data.   |
+|                        |                          | See `pull request description <https://github.com/godotengine/godot/pull/100293>`__   |
+|                        |                          | for details.                                                                          |
++------------------------+--------------------------+---------------------------------------------------------------------------------------+
+| ``HashSet``            | ``std::unordered_set``   | **Use this as the "default" set type.**                                               |
++------------------------+--------------------------+---------------------------------------------------------------------------------------+
+| ``RBSet``              | ``std::set``             | Uses a `red-black tree <https://en.wikipedia.org/wiki/Red-black_tree>`__              |
+|                        |                          | for faster access.                                                                    |
++------------------------+--------------------------+---------------------------------------------------------------------------------------+
+| ``VSet``               | ``std::flat_set``        | Uses copy-on-write (COW) semantics.                                                   |
+|                        |                          | This means it's generally slower but can be copied around almost for free.            |
+|                        |                          | The performance benefits of ``VSet`` aren't established, so prefer using other types. |
++------------------------+--------------------------+---------------------------------------------------------------------------------------+
+| ``HashMap``            | ``std::unordered_map``   | **Use this as the "default" map type.** Does not preserve insertion order.            |
++------------------------+--------------------------+---------------------------------------------------------------------------------------+
+| ``AHashMap``           | ``std::unordered_map``   | Array-based implementation of a hash map. Does not preserve insertion order.          |
++------------------------+--------------------------+---------------------------------------------------------------------------------------+
+| ``OAHashMap``          | ``std::unordered_map``   | Does not preserve insertion order.                                                    |
++------------------------+--------------------------+---------------------------------------------------------------------------------------+
+| ``RBMap``              | ``std::map``             | Uses a `red-black tree <https://en.wikipedia.org/wiki/Red-black_tree>`__              |
+|                        |                          | for faster access.                                                                    |
++------------------------+--------------------------+---------------------------------------------------------------------------------------+
+| ``VMap``               | ``std::flat_map``        | Uses copy-on-write (COW) semantics.                                                   |
+|                        |                          | This means it's generally slower but can be copied around almost for free.            |
+|                        |                          | The performance benefits of ``VMap`` aren't established, so prefer using other types. |
++------------------------+--------------------------+---------------------------------------------------------------------------------------+
+| ``Dictionary`` 📜      | ``std::unordered_map``   | Keys and values can be of any Variant type. No static typing is imposed.              |
+|                        |                          | Uses shared reference counting, similar to ``std::shared_ptr``.                       |
++------------------------+--------------------------+---------------------------------------------------------------------------------------+
+| ``TypedDictionary`` 📜 | ``std::unordered_map``   | Subclass of ``Dictionary`` but with static typing for its keys and values.            |
++------------------------+--------------------------+---------------------------------------------------------------------------------------+
+| ``Pair``               | ``std::pair``            | Stores a single key-value pair.                                                       |
++------------------------+--------------------------+---------------------------------------------------------------------------------------+
 
 ``auto`` keyword
 ~~~~~~~~~~~~~~~~


Ivorforce (commented):
I like this a lot! I think a table like this could be really helpful to get people started with the codebase.
Perhaps I'd rather structure it like this:

```
Godot Datatype  |  Closest STL datatype  |  Comments
Vector   |   std::vector   |   Uses copy-on-write semantics. That means it's generally slower but can be copied around almost for free.
```
etc.

What do you think?

contributing/development/cpp_usage_guidelines.rst
```diff
@@ -55,16 +55,39 @@ See :ref:`doc_faq_why_not_stl` for more information.
...
 ``std::vector`` and the like. Instead, use Godot's datatypes as described below:
 
-- Use ``String`` instead of ``std::string``.
-- Use ``Vector`` instead of ``std::vector``. In some cases, ``LocalVector``
-  can be used as an alternative (ask core developers first).
-- Use ``Array`` instead of ``std::array``.
++------------------------+----------------------------+-------------------------------------------+
+| C++ STL datatype       | Recommended Godot datatype | Alternative Godot datatype*               |
++========================+============================+===========================================+
+| ``std::string``        | ``String``                 | ``StringName``                            |
```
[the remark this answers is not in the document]
Calinou: They are pretty specific, so I won't list them here.

contributing/development/cpp_usage_guidelines.rst
```diff
@@ -55,16 +55,39 @@ See :ref:`doc_faq_why_not_stl` for more information.
...
++------------------------+----------------------------+-------------------------------------------+
+| ``std::set``           | ``HashSet``                | ``RBSet``                                 |
++------------------------+----------------------------+-------------------------------------------+
+| ``std::map``           | ``OAHashMap``              | ``AHashMap``, ``HashMap``, ``Dictionary`` |
++------------------------+----------------------------+-------------------------------------------+
+| ``std::unordered_map`` | ``OAHashMap``              | ``AHashMap``, ``HashMap``, ``Dictionary`` |
++------------------------+----------------------------+-------------------------------------------+
+| ``std::pair``          | ``Pair``                   |                                           |
++------------------------+----------------------------+-------------------------------------------+
+| ``std::variant``       | ``Variant``                |                                           |
```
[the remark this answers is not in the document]
Calinou: OK, I'll remove it then.

Calinou:
> As I was reading this, I was thinking that perhaps it might be better to structure it by Godot datatype than STL datatype? That would allow us to add some more nuance.

Done :slightly_smiling_face:

contributing/development/cpp_usage_guidelines.rst
```diff
@@ -53,18 +53,69 @@ as Godot provides its own data types (among other things).
...
+|                        |                          | Only a limited list of packed array types are available                             |
+|                        |                          | (use ``TypedArray`` otherwise).                                                     |
++------------------------+--------------------------+-------------------------------------------------------------------------------------+
+| ``Span``               | ``std::span``            | Represents read-only access to a contiguous array without needing to copy any data. |
+|                        |                          | See `pull request description <https://github.com/godotengine/godot/pull/100293>`__ |
+|                        |                          | for details.                                                                        |
++------------------------+--------------------------+-------------------------------------------------------------------------------------+
+| ``HashSet``            | ``std::set``             | **Use this as the "default" set type.**                                             |
```
AThousandShips, line 88: As both are hash containers, especially since `HashSet` doesn't offer any insertion ordering
```suggestion
| ``HashSet``            | ``std::unordered_set``   | **Use this as the "default" set type.**                                             |
```

contributing/development/cpp_usage_guidelines.rst
```diff
@@ -53,18 +53,69 @@ as Godot provides its own data types (among other things).
...
++------------------------+--------------------------+-------------------------------------------------------------------------------------+
+| ``RBSet``              | ``std::set``             | Uses a `red-black tree <https://en.wikipedia.org/wiki/Red-black_tree>`__            |
+|                        |                          | for faster access.                                                                  |
++------------------------+--------------------------+-------------------------------------------------------------------------------------+
+| ``HashMap``            | ``std::map``             | **Use this as the "default" map type.** Iteration is different from ``std::map``    |
```
AThousandShips, line 93: I'd argue that the relevant difference isn't the ordering, but the keying, `std::map` is a key based system, which is mostly relevant for things like time complexity, so I'd say this should be `std::unordered_map`

contributing/development/cpp_usage_guidelines.rst
```diff
@@ -53,18 +53,69 @@ as Godot provides its own data types (among other things).
...
+| ``AHashMap``           | ``std::map``             | Array-based implementation of a hash map.                                           |
++------------------------+--------------------------+-------------------------------------------------------------------------------------+
+| ``OAHashMap``          | ``std::unordered_map``   | Does not preserve insertion order. Generally faster than ``HashMap`` and            |
+|                        |                          | ``AHashMap``.                                                                       |
++------------------------+--------------------------+-------------------------------------------------------------------------------------+
+| ``Dictionary`` 📜      | ``std::map``             | Keys and values can be of any Variant type. No static typing is imposed.            |
++------------------------+--------------------------+-------------------------------------------------------------------------------------+
+| ``TypedDictionary`` 📜 | ``std::map``             | Equivalent to ``Dictionary`` but with static typing for its keys and values.        |
```
AThousandShips, line 103: Same for `Dictionary`, as they are just `HashMap`, and again aren't ordered on key (unless sorted)

mhilbrunner (approved):
Approving the addition, the other feedback still applies :) Thanks for working on this!

contributing/development/cpp_usage_guidelines.rst
```diff
@@ -53,18 +53,69 @@ as Godot provides its own data types (among other things).
...
-- Use ``Array`` instead of ``std::array``.
+``std::vector`` and the like. Instead, use Godot's datatypes as described below.
+A 📜 icon denotes the type is part of :ref:`Variant <doc_variant_class>`. This
+means it can be used as a parameter or return value of a method exposed to the
+scripting API.
+
++------------------------+--------------------------+-------------------------------------------------------------------------------------+
+| Godot datatype         | Closest C++ STL datatype | Comment                                                                             |
++========================+==========================+=====================================================================================+
+| ``String`` 📜          | ``std::string``          | **Use this as the "default" string type.**                                          |
```
Ivorforce, line 64:
```suggestion
| ``String`` 📜          | ``std::string``          | **Use this as the "default" string type.** ``String`` uses UTF-32 encoding.                                         |
```

contributing/development/cpp_usage_guidelines.rst
```diff
@@ -53,18 +53,69 @@ as Godot provides its own data types (among other things).
...
++------------------------+--------------------------+-------------------------------------------------------------------------------------+
+| ``StringName`` 📜      | ``std::string``          | Uses string interning for fast lookups. Use this for static strings that are        |
```
Ivorforce, line 66:
```suggestion
| ``StringName`` 📜      | ``std::string``          | Uses string interning for fast comparisons. Use this for static strings that are        |
```

contributing/development/cpp_usage_guidelines.rst
```diff
@@ -53,18 +53,69 @@ as Godot provides its own data types (among other things).
...
+|                        |                          | This means it's generally slower but can be copied around almost for free.          |
++------------------------+--------------------------+-------------------------------------------------------------------------------------+
+| ``LocalVector``        | ``std::vector``          | Closer to ``std::vector`` in semantics. In most situations, ``Vector`` should be    |
+|                        |                          | preferred.                                                                          |
++------------------------+--------------------------+-------------------------------------------------------------------------------------+
+| ``Array`` 📜           | ``std::vector``          | Values can be of any Variant type. No static typing is imposed.                     |
++------------------------+--------------------------+-------------------------------------------------------------------------------------+
+| ``TypedArray`` 📜      | ``std::vector``          | Equivalent to ``Array`` but with static typing for its elements.                    |
```
Ivorforce, line 77:
```suggestion
| ``TypedArray`` 📜      | ``std::vector``          | Subclass of ``Array`` but with static typing for its elements.                    |
```

contributing/development/cpp_usage_guidelines.rst
```diff
@@ -53,18 +53,69 @@ as Godot provides its own data types (among other things).
...
+|                        |                          | Not to be confused with ``Packed*Array``, which is internally a ``Vector``.         |
++------------------------+--------------------------+-------------------------------------------------------------------------------------+
+| ``Packed*Array`` 📜    | ``std::vector``          | Equivalent to ``Vector``, e.g. ``PackedColorArray = Vector<Color>``.                |
```
Ivorforce, line 80:
```suggestion
| ``Packed*Array`` 📜    | ``std::vector``          | Alias of ``Vector``, e.g. ``PackedColorArray = Vector<Color>``.                |
```

contributing/development/cpp_usage_guidelines.rst
```diff
@@ -53,18 +53,69 @@ as Godot provides its own data types (among other things).
...
++------------------------+--------------------------+-------------------------------------------------------------------------------------+
+| ``HashMap``            | ``std::map``             | **Use this as the "default" map type.** Iteration is different from ``std::map``    |
+|                        |                          | (``HashMap`` is insertion-ordered, while ``std::map`` is key-ordered).              |
++------------------------+--------------------------+-------------------------------------------------------------------------------------+
+| ``AHashMap``           | ``std::map``             | Array-based implementation of a hash map.                                           |
++------------------------+--------------------------+-------------------------------------------------------------------------------------+
+| ``OAHashMap``          | ``std::unordered_map``   | Does not preserve insertion order. Generally faster than ``HashMap`` and            |
+|                        |                          | ``AHashMap``.                                                                       |
```
Ivorforce, line 99: I'm not sure it's faster than `AHashMap`. I don't quite remember how they compare in terms of speed.

contributing/development/cpp_usage_guidelines.rst
Ivorforce, line 103:
```suggestion
| ``TypedDictionary`` 📜 | ``std::map``             | Alias of ``Dictionary`` but with static typing for its keys and values.        |
```

contributing/development/cpp_usage_guidelines.rst
```diff
@@ -53,18 +53,69 @@ as Godot provides its own data types (among other things).
...
++------------------------+--------------------------+-------------------------------------------------------------------------------------+
+| ``Dictionary`` 📜      | ``std::map``             | Keys and values can be of any Variant type. No static typing is imposed.            |
++------------------------+--------------------------+-------------------------------------------------------------------------------------+
+| ``TypedDictionary`` 📜 | ``std::map``             | Equivalent to ``Dictionary`` but with static typing for its keys and values.        |
++------------------------+--------------------------+-------------------------------------------------------------------------------------+
+| ``Pair``               | ``std::pair``            | Stores a single key-value pair.                                                     |
++------------------------+--------------------------+-------------------------------------------------------------------------------------+
+
+\*: For certain use cases, this alternative datatype is more performant or has
```
Ivorforce, line 108: I don't think this disclaimer is relevant anymore?

contributing/development/cpp_usage_guidelines.rst
```diff
@@ -53,18 +53,69 @@ as Godot provides its own data types (among other things).
...
+characteristics that are more suitable for the task at hand. However, it is not
+always a direct replacement for the recommended datatype. Ask for advice if you
+are unsure.
 
 .. note::
 
-    Godot also has a List datatype (which is a linked list). While List is already used
-    in the codebase, it typically performs worse than other datatypes like Vector
-    and Array. Therefore, List should be avoided in new code unless necessary.
+    Godot also has a ``List`` datatype (which is a linked list). While ``List``
```
Ivorforce, line 115: It's probably fine to move `List` into the table, and just put the usage disclaimer in the comment box. Slow or not, it _is_ a regular collection type.

contributing/development/cpp_usage_guidelines.rst
```diff
@@ -53,18 +53,69 @@ as Godot provides its own data types (among other things).
...
++------------------------+--------------------------+-------------------------------------------------------------------------------------+
+| ``Vector``             | ``std::vector``          | **Use this as the "default" vector type.** Uses copy-on-write semantics.            |
+|                        |                          | This means it's generally slower but can be copied around almost for free.          |
++------------------------+--------------------------+-------------------------------------------------------------------------------------+
+| ``LocalVector``        | ``std::vector``          | Closer to ``std::vector`` in semantics. In most situations, ``Vector`` should be    |
+|                        |                          | preferred.                                                                          |
++------------------------+--------------------------+-------------------------------------------------------------------------------------+
+| ``Array`` 📜           | ``std::vector``          | Values can be of any Variant type. No static typing is imposed.                     |
```
Ivorforce, line 75:
```suggestion
| ``Array`` 📜           | ``std::vector``          | Values can be of any Variant type. No static typing is imposed. Uses shared reference counting, similar to ``std::shared_ptr``.                    |
```

contributing/development/cpp_usage_guidelines.rst
```diff
@@ -53,18 +53,69 @@ as Godot provides its own data types (among other things).
...
+|                        |                          | (``HashMap`` is insertion-ordered, while ``std::map`` is key-ordered).              |
++------------------------+--------------------------+-------------------------------------------------------------------------------------+
+| ``AHashMap``           | ``std::map``             | Array-based implementation of a hash map.                                           |
++------------------------+--------------------------+-------------------------------------------------------------------------------------+
+| ``OAHashMap``          | ``std::unordered_map``   | Does not preserve insertion order. Generally faster than ``HashMap`` and            |
+|                        |                          | ``AHashMap``.                                                                       |
++------------------------+--------------------------+-------------------------------------------------------------------------------------+
+| ``Dictionary`` 📜      | ``std::map``             | Keys and values can be of any Variant type. No static typing is imposed.            |
```
Ivorforce, line 101:
```suggestion
| ``Dictionary`` 📜      | ``std::map``             | Keys and values can be of any Variant type. No static typing is imposed. Uses shared reference counting, similar to ``std::shared_ptr``.           |
```

contributing/development/cpp_usage_guidelines.rst
Calinou: I remember reduz mentioning this a few years ago (along with many pull requests that replaced uses of `HashMap` and `AHashMap` with `OAHashMap` for performance reasons. It's possible the performance situation has changed since then, so I've removed the note.

Ivorforce (approved):
This is looking great! I'm happy we finally have a reference point for this info ☺️

contributing/development/cpp_usage_guidelines.rst
```diff
@@ -53,18 +53,61 @@ as Godot provides its own data types (among other things).
...
++------------------------+--------------------------+--------------------------------------------------------------------------------------+
+| ``Span``               | ``std::span``            | Represents read-only access to a contiguous array without needing to copy any data.  |
+|                        |                          | See `pull request description <https://github.com/godotengine/godot/pull/100293>`__  |
+|                        |                          | for details.                                                                         |
++------------------------+--------------------------+--------------------------------------------------------------------------------------+
+| ``HashSet``            | ``std::unordered_set``   | **Use this as the "default" set type.**                                              |
++------------------------+--------------------------+--------------------------------------------------------------------------------------+
+| ``RBSet``              | ``std::unordered_set``   | Uses a `red-black tree <https://en.wikipedia.org/wiki/Red-black_tree>`__             |
```
AThousandShips, line 95:
```suggestion
| ``RBSet``              | ``std::set``             | Uses a `red-black tree <https://en.wikipedia.org/wiki/Red-black_tree>`__             |
```
This was correctly in that sense, also I think a description of `RBMap` is needed as well, might also want to describe `VSet` and `VMap`, which are equivalent to `std::flat_set/map` (which are in C++23, or in Boost)

Ivorforce: `VSet` and `VMap` also notably use copy-on-write semantics. But they aren't used much in the engine at all, and the benefit is unclear, so it has been proposed to remove them again. 
I guess there's no harm in covering them since they do exist, but I'd probably attach a note that the benefits aren't clearly established.

Ivorforce (approved):
Re-approving after changes

mhilbrunner:
Thank you! Merged.

PR #10902, opened 2025-04-28
Hide user notes on 404 page

Calinou:
This switches to `404.rst` instead of an embedded 404 page in `conf.py`, so we can specify page metadata.

I've tested this change locally and it stops the user notes block from generating on the generated `404.html`.


mhilbrunner (approved):

mhilbrunner:
Thanks!

PR #10907, opened 2025-04-29
Document CollisionObject3D requires object visibility

lackhand:
this is pretty unwieldy; maybe it's _actually_ a bug in how mouse_entered and _exit work. But per https://github.com/godotengine/godot/blob/ce94b26de74a0a0f2463c5dc61b9b683a1e3676b/scene/3d/physics/collision_object_3d.cpp#L338 this is directly how it's implemented, so we might as well be honest about it.

(in general, there's some confusion about how to "turn off" physics objects -- via processing? (there's a setting for that). Via setting a parent object to invisible? (it's not ray pickable anymore, but there's no configuration/notes around it. Probably fine, therefore: doc fix.))


AThousandShips:

AThousandShips:
No worries!

PR #10909, opened 2025-04-30
Clarify that `Tree.create_item()` may fail and when

Kv-Lam:
This change addresses [`_OnTreeItemSelected()` signal receiver method using C# doesn't allow adding of items #103101](https://github.com/godotengine/godot/issues/103101). 

This update clarifies the `Tree.create_item()` documentation to indicate that the method returns null if called during a mouse selection event.

Note: This is my first time contributing, so I apologize for any mistakes!


AThousandShips:

PR #10912, opened 2025-05-01
Add new `FixedVector` type to the types table.

Ivorforce:
- FixedVector was _just_ added in https://github.com/godotengine/godot/pull/104055.

### DIFF
--- a/contributing/development/cpp_usage_guidelines.rst
+++ b/contributing/development/cpp_usage_guidelines.rst
@@ -86,6 +86,10 @@ scripting API.
 | ``List``               | ``std::list``            | Linked list type. Generally slower than other array/vector types. Prefer using        |
 |                        |                          | other types in new code, unless using ``List`` avoids the need for type conversions.  |
 +------------------------+--------------------------+---------------------------------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


contributing/development/cpp_usage_guidelines.rst
```diff
@@ -86,6 +86,8 @@ scripting API.
 | ``List``               | ``std::list``            | Linked list type. Generally slower than other array/vector types. Prefer using        |
 |                        |                          | other types in new code, unless using ``List`` avoids the need for type conversions.  |
 +------------------------+--------------------------+---------------------------------------------------------------------------------------+
+| ``FixedVector``        | ``std::array``           | Vector with a fixed capacity (more similar to ``boost::container::static_vector``).   |
```
clayjohn, line 89: I think we should mention that the data is allocated on the stack so there are no heap allocations

Ivorforce: I didn't find a good formatulation for it, but "makes no heap allocations" is great! I'll add that.

mhilbrunner (approved):

mhilbrunner:
Thanks!

PR #10914, opened 2025-05-01
Add notes to `Dictionary`, `HashMap` and `Array`.

Ivorforce:
- Split from #10912 to be cherry-pickable

I noticed these after https://github.com/godotengine/godot-docs/pull/10873 was merged:

- `HashMap` and `Dictionary` _do_ preserve insertion order.
- `Dictionary` and `Array` use `HashMap` and `Vector` internally, respectively.
- I think we should link to the types in the FAQ entry where people may look for them.

### DIFF
--- a/contributing/development/cpp_usage_guidelines.rst
+++ b/contributing/development/cpp_usage_guidelines.rst
@@ -45,6 +45,8 @@ variables and ``nullptr`` is encouraged when possible. Still, try to keep your
 use of modern C++ features conservative. Their use needs to serve a real
 purpose, such as improving code readability or performance.
 
+.. _doc_cpp_godot_types:
+
 Standard Template Library
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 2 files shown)


contributing/development/cpp_usage_guidelines.rst
```diff
@@ -99,7 +102,7 @@ scripting API.
 |                        |                          | This means it's generally slower but can be copied around almost for free.            |
 |                        |                          | The performance benefits of ``VSet`` aren't established, so prefer using other types. |
 +------------------------+--------------------------+---------------------------------------------------------------------------------------+
-| ``HashMap``            | ``std::unordered_map``   | **Use this as the "default" map type.** Does not preserve insertion order.            |
+| ``HashMap``            | ``std::unordered_map``   | **Use this as the "default" map type.** Preserves insertion order.                    |
```
clayjohn, line 105: Is `std::unordered_map` still the closest analogue? Or does `std` have a map type that preserves insertion order?

On a related note. I have very vague memories that the fact that HashMap preserves insertion order was an implementation detail and we didn't want that fact to be part of the contract with users. 

I'll do a quick search on RC and see if I can find some discussion about it to refresh my memory

clayjohn: I couldn't find anything conclusive, but it seems that we use HashMap in a lot of places that rely on preserving insertion order.

Ivorforce: It's also [documented behavior](https://docs.godotengine.org/en/stable/classes/class_dictionary.html): 

> Dictionaries will preserve the insertion order when adding new entries.

I think the window to change this has closed 😅

mhilbrunner: Only for exposed APIs, but yeah. I think considering that both JS and Python do also preserve insertion order on their dicts (with some caveats) and the low barrier we aim for with GDScript, this seems fine.

mhilbrunner (approved):

mhilbrunner:
Thank you! Merged.

PR #10920, opened 2025-05-02
Clarify UTF encoding between C# strings and Godot Strings

### LINKED ISSUE #7682: Document that C# `string`s are UTF16 while Godot `String`s are UTF32
See the discussion in https://github.com/godotengine/godot-docs/pull/7612

It is a very subtle difference that doesn't matter for most cases, but can cause some issues when porting code or calling some text server functions.

Athenr:
Fix #7682 

Updated c_sharp_differences.rst to include the difference between the UTF encoding for C# strings and Godot Strings.

### DIFF
--- a/tutorials/scripting/c_sharp/c_sharp_differences.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_differences.rst
@@ -381,7 +381,7 @@ Use ``System.String`` (``string``). Most of Godot's String methods have an
 equivalent in ``System.String`` or are provided by the ``StringExtensions``
 class as extension methods.
 
-Note that C# strings uses UTF-16 encoding, while Godot Strings uses UTF-32 encoding.
+Note that C# strings use UTF-16 encoding, while Godot Strings use UTF-32 encoding.
 
 Example:
 


tutorials/scripting/c_sharp/c_sharp_differences.rst
```diff
@@ -381,6 +381,8 @@ Use ``System.String`` (``string``). Most of Godot's String methods have an
 equivalent in ``System.String`` or are provided by the ``StringExtensions``
 class as extension methods.
 
+Note that C# strings uses UTF-16 encoding, while Godot Strings uses UTF-32 encoding.
```
AThousandShips, line 384:
```suggestion
Note that C# strings use UTF-16 encoding, while Godot Strings use UTF-32 encoding.
```

raulsntos (commented):
Thanks for contributing to the C# documentation. I think we had in mind something more extensive that clarifies why this can be a problem.

Normally, the encoding is not a problem since we convert between C# strings and Godot strings automatically, so ideally users wouldn't even need to think about it. So mentioning the encoding difference would not be important if that's all there is to say.

We wanted to add a note about this in the documentation because of the problems that it may cause in some APIs. The example given in the discussion from https://github.com/godotengine/godot-docs/pull/7612 was `TextServer::string_get_word_breaks`. This API breaks the text into words and returns an array of character indices, but these indices will be wrong for C# strings **in some cases**.

For example, for the string `"ℌ𝔢𝔩𝔩𝔬 𝔚𝔬𝔯𝔩𝔡"` the returned array would be `[0, 5, 6, 11]`. But those indices don't correspond in the C# string because the characters may take more than a single UTF-16 character. In C# the indices should be `[0, 9, 10, 20]` or you should use `System.Rune` instead.

RubenGarcia (approved):

skyace65:

PR #10935, opened 2025-05-09
Remove outdated 2D batching note

Jordyfel:
This note about 2D batching being used only in OpenGL rendering has been outdated for a while now.

Maybe it would be good to leave the mention that 2D batching is more important when using OpenGL?


clayjohn (approved):
This looks great. I think removing the section on why 2D batching wasn't implemented in the RD backend is a good call

mhilbrunner (approved):
Any idea how far back we should cherrypick this?

clayjohn:
> Any idea how far back we should cherrypick this?

Just 4.4

mhilbrunner:
Thank you! Merged.

PR #10937, opened 2025-05-09
Recommend GPU Screen Recorder instead of SimpleScreenRecorder for recording on Linux

v1993:
Switch to recommending GPU Screen Recorder instead of SimpleScreenRecorder for recording for Linux. It's a newer project that offers the following features over SSR:

1. Full Wayland support, including Wayland hotkeys. X11 is supported as well.
2. Very efficient utilization of GPU encoding across all GPU vendors, even better than OBS Studio
3. Not quite relevant for Godot at the moment, but HDR recording support
4. Importantly for end users, it's [available](https://flathub.org/apps/com.dec05eba.gpu_screen_recorder) as a Flatpak, meaning it's easy to install on about every distro out there

Other than that, GPU Screen Recorder is just generally the go-to solution for screen recording on Linux these days due to being both efficient and easy to use, so it makes sense to recommend it.

### DIFF
--- a/classes/class_moviewriter.rst
+++ b/classes/class_moviewriter.rst
@@ -31,7 +31,7 @@ If you need to encode to a different format or pipe a stream through third-party
 
 \ **Editor usage:** A default movie file path can be specified in :ref:`ProjectSettings.editor/movie_writer/movie_file<class_ProjectSettings_property_editor/movie_writer/movie_file>`. Alternatively, for running single scenes, a ``movie_file`` metadata can be added to the root node, specifying the path to a movie file that will be used when recording that scene. Once a path is set, click the video reel icon in the top-right corner of the editor to enable Movie Maker mode, then run any scene as usual. The engine will start recording as soon as the splash screen is finished, and it will only stop recording when the engine quits. Click the video reel icon again to disable Movie Maker mode. Note that toggling Movie Maker mode does not affect project instances that are already running.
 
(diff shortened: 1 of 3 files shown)


classes/class_moviewriter.rst
```diff
@@ -31,7 +31,7 @@ If you need to encode to a different format or pipe a stream through third-party
...
 
-\ **Note:** MovieWriter is available for use in both the editor and exported projects, but it is *not* designed for use by end users to record videos while playing. Players wishing to record gameplay videos should install tools such as `OBS Studio <https://obsproject.com/>`__ or `SimpleScreenRecorder <https://www.maartenbaert.be/simplescreenrecorder/>`__ instead.
+\ **Note:** MovieWriter is available for use in both the editor and exported projects, but it is *not* designed for use by end users to record videos while playing. Players wishing to record gameplay videos should install tools such as `OBS Studio <https://obsproject.com/>`__ or `GPU Screen Recorder <https://git.dec05eba.com/gpu-screen-recorder/about/>`__ instead.
```
Calinou, line 34: Files in `classes/` should not be edited manually, as they are automatically generated. Instead, you should open a pull request on the main Godot repository: https://github.com/godotengine/godot/blob/master/doc/classes/MovieWriter.xml

skyace65 (commented):
I'm in agreement with Calinou. Adding GPU Screen Recorder is fine but I don't see a reason to remove SimpleScreenRecorder

PR #10938, opened 2025-05-09
Add `call_deferred` to UI Navigation code example

### LINKED ISSUE #10136: UI GUI Navigation documentation is incorrect
**Your Godot version:**
4.3

**Issue description:**
Section entitled "Necessary code" is incorrect or godot has a bug.  A VBoxContainer was created in a scene with a root Control node.  the VBoxContainer contains 4 buttons with a custom theme set.  The Control has a script:

```
func _ready():
	$StartButton.grab_focus()
```

The scene was instantiated and added to an active scene that has a Node2D root node. 

This does not appear to result in the StartButton taking focus.  Instead, it appears it is necessary to set focus somewhere in _process() or with a timer signal. 

Necessary code[¶](https://docs.godotengine.org/en/4.3/tutorials/ui/gui_navigation.html#necessary-code)

ProgrammerOnCoffee:
Adds `call_deferred` to the code example in the UI Navigation tutorial so that the code works more reliably. Also fixes a couple of small typos.
Closes #10136

### DIFF
--- a/tutorials/ui/gui_navigation.rst
+++ b/tutorials/ui/gui_navigation.rst
@@ -58,22 +58,25 @@ have well-defined vertical or horizontal navigation flow.
 Necessary code
 --------------
 
-For keyboard and controller navigation to work correctly, any node must be focused on
+For keyboard and controller navigation to work correctly, any node must be focused by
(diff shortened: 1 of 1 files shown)


skyace65:
Thanks!

PR #10943, opened 2025-05-12
Use bitshift and binary `OR` operator in to set layers in Physics introduction

Calinou:
This approach is more idiomatic, more reliable and also more efficient than using `pow()`.

- See https://github.com/godotengine/godot-docs-user-notes/discussions/2#discussioncomment-13108109.


mhilbrunner:
Thanks!

PR #10959, opened 2025-05-23
Remove suggestion to do a partial upgrade

tgpholly:
The Compiling for Windows page currently suggests doing a partial upgrade on arch, this is potentially dangerous and can lead to a broken system.

See: https://wiki.archlinux.org/title/System_maintenance#Partial_upgrades_are_unsupported


Calinou (approved):
`pacman -S` is the standard command to install a package, so this is valid.

skyace65 (approved):

skyace65:

Issue #9277, opened 2024-04-24
Add clear instructions on how an end user can use a pre-compiled GDExtension in their project


muno777:
**Your Godot version:** 4.2

**Issue description:** The current documentation for GDExtension seems to focus on the developer experience, without much to offer for users who wish to use an extension in their own project by downloading a `.gdextension` file. This threw me for a loop initially until I realised that it's as simple as adding the file to the project folder (unlike addons, which need to be enabled in the project settings). To help others learn this more easily, it would be nice if the documentation included a brief section addressed to end users.

**URL to the documentation page (if already existing):** [Link](https://docs.godotengine.org/en/stable/tutorials/scripting/gdextension/what_is_gdextension.html)

PR #10960, opened 2025-05-23
Add instructions on how to use a pre-compiled GDExtension, + update a link in the GDExtension class reference

muno777:
Currently, a Godot end-user who downloaded a pre-made GDExtension (such as [SG Physics 2D](https://gitlab.com/snopek-games/sg-physics-2d)) would struggle to find out how to use it in their own project. These instructions are not listed anywhere, and the current GDExtension docs only cater to people *writing* a GDExtension.

This PR adds a short section to the main GDExtension page describing this, and also changes the link in the GDExtension class doc to lead to the main GDExtension page instead of a sub-page.

Addresses #9277.

### DIFF
--- a/classes/class_gdextension.rst
+++ b/classes/class_gdextension.rst
@@ -28,7 +28,7 @@ The **GDExtension** resource type represents a `shared library <https://en.wikip
 Tutorials
 ---------
 
-- :doc:`GDExtension overview <../tutorials/scripting/gdextension/what_is_gdextension>`
+- :doc:`GDExtension overview <../tutorials/scripting/gdextension/index>`
 
 - :doc:`GDExtension example in C++ <../tutorials/scripting/cpp/gdextension_cpp_example>`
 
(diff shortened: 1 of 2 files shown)


classes/class_gdextension.rst
AThousandShips, line 1: This change needs to be done over [here](https://github.com/godotengine/godot), changing the generated files is not allowed, see [here](https://docs.godotengine.org/en/latest/contributing/documentation/updating_the_class_reference.html)

tutorials/scripting/gdextension/index.rst
```diff
@@ -9,6 +9,9 @@ The GDExtension system
 native `shared libraries <https://en.wikipedia.org/wiki/Library_(computing)#Shared_libraries>`__
 at runtime. You can use it to run native code without compiling it with the engine.
 
+In order to use a pre-compiled GDExtension in your Godot project, place the `.gdextension` file somewhere inside your
+project's directory. Godot will then load the extension and you should be able to use it in your project.
```
dsnopek, line 13: I think some folks may interpret this to mean that you only need to copy in the `.gdextension` file and nothing else.

Maybe something like:

```suggestion
In order to use a pre-compiled GDExtension in your Godot project, place all the files (including the `.gdextension` file) somewhere inside your
project's directory. Godot will then load the extension and you should be able to use it in your project.
```

skyace65:
Closing due to the lack of a response to reviews

PR #10969, opened 2025-05-27
Replace uses of shorthand codeblock syntax to avoid issues with localization

Calinou:
Previously, many paragraphs in the documentation used `::` at the end of a line as a replacement for a single colon (and `::` on the following line). Due to this, the `::` was part of the extracted string for localization. Weblate also showed a warning about the trailing `::`, as it's not normally expected to be there.

This change requires translators to revalidate all modified strings. However, considering many of those have broken code blocks due to translators replacing `::` with `:`, it can be a good idea.

This also adds documentation writing guidelines for code blocks and admonitions. This commit does not modify existing admonitions in any way (as this would lead to a much larger diff, and require more localizable text to be revalidated by translators).


contributing/development/compiling/compiling_for_windows.rst
```diff
@@ -205,7 +211,9 @@ optional PIX and Agility SDK components).
              which is needed to generate some files.
           2. Clone the `godot-nir-static <https://github.com/godotengine/godot-nir-static>`_
              directory and navigate to it.
-          3. Run the following::
+          3. Run the following:
+
+::
```
Mickeon, line 216:
```suggestion
          ::
```

contributing/documentation/docs_writing_guidelines.rst
```diff
@@ -643,6 +643,58 @@ rendered, prefer splitting the setting name and the section when the link is lon
...
+
+::
+
+    This is a code block:
+
+    ::
+
+        The code block's contents.
+
+As for admonitions (note/warning/etc. blocks), you should write the admonition tag on its own line,
```
Mickeon, line 674:
```suggestion
As for admonitions (note, warning, etc. blocks), you should write the admonition tag on its own line,
```

tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -1810,7 +1870,9 @@ The following pattern types are available:
 
 - Binding pattern
     A binding pattern introduces a new variable. Like the wildcard pattern, it matches everything - and also gives that value a name.
-    It's especially useful in array and dictionary patterns::
+    It's especially useful in array and dictionary patterns:
+
+::
```
Mickeon, line 1875:
```suggestion
    ::
```

Mickeon (commented):
I may be misunderstanding how this syntax works but from what I do know...

Issue #10942, opened 2025-05-12
Better document the use of HandBrake external tool or directly remove it


MarianoGnu:
**Your Godot version:**
4.4.1-stable

**Issue description:**
I am dealing with video conversions to be used in an application we are building. This application has some embeded video tutorials.
The section [Recommended Theora encoding settings](https://docs.godotengine.org/en/stable/tutorials/animation/playing_videos.html#recommended-theora-encoding-settings) recommends to use HandBrake to convert videos, but HandBrake does not seem to export to ogv files, yu can configure mkv with libtheora decode, but that will not work. Maybe ogv was removed and the documentation is outdated

**URL to the documentation page:**
https://docs.godotengine.org/en/stable/tutorials/animation/playing_videos.html#recommended-theora-encoding-settings

Calinou:
I wouldn't be surprised if HandBrake removed Theora export, given the format isn't used often anymore.

We can list an alternative open source GUI tool if we find one, but I can't think of any right now.

berarma:
> I wouldn't be surprised if HandBrake removed Theora export, given the format isn't used often anymore.

It looks like it can encode to Theora and Vorbis but only inside the MKV container that isn't supported by Godot.

> > We can list an alternative open source GUI tool if we find one, but I can't think of any right now.
> 
> I have used [Theora converter .NET](https://sourceforge.net/projects/theoraconverter/) on Windows. However, i have had [some issues](https://github.com/godotengine/godot/issues/106311) with the video it produced, but i am still not sure of it is the converter's fault or Godot's fault, because video looks okay on Windows media player

I'd recommend using something newer than that. The best option is a recent FFmpeg nightly build. You can try FFmpegGUI, but if you're not afraid, learning to use the command line for simple conversions is affordable to anyone.

I only use the command line on Linux so if anyone can recommend a GUI we could put it in the documentation.

berarma:
[Shutter Encoder](https://www.shutterencoder.com/) is another GUI that looks very good, powerful, cross-platform, opensource and seems to support Theora/ogg.

berarma:
#10732 has been tagged to be cherrypicked for the current stable versions. And I'll propose to add Shutter Encoder in another PR.

skyace65:
Closing this as #10732 was merged.

berarma:
For the record, Shutter Encoder hasn't been added to the docs because it uses pretty inefficient encoding parameters. Pushing godotengine/godot#108600 instead.

PR #10614, opened 2025-02-04
Update playing videos

berarma:
Documents changes from godotengine/godot#101958.

- Add information about using `-g:v` to encode videos with a higher GOP size.
- Add notes and warnings about problems when encoding Theora with FFmpeg.
- ~~Remove note about not being able to seek Theora files.~~

Please, feel free to correct my writing.

PR #10972, opened 2025-05-28
Update Playing Videos page for Godot 4.5

berarma:
Removed another limitation: change playback speed.

~~Added Shutter Encoder as a FFmpeg GUI. Related to #10942.~~

I've decided to remove the date from the FFmpeg warning since they have already released a point version but without the fix. And I don't know how many more they will release yet. I hoped that the next major version was closer to being released.

Changed the GOP description a bit to hopefully better fit the rest of the documentation.

### DIFF
--- a/tutorials/animation/playing_videos.rst
+++ b/tutorials/animation/playing_videos.rst
@@ -150,10 +150,8 @@ To ensure your videos decode smoothly on varied hardware:
 Playback limitations
 --------------------
 
-There are several limitations with the current implementation of video playback in Godot:
+There are some limitations with the current implementation of video playback in Godot:
 
-- Changing playback speed is not supported. VideoStreamPlayer also won't follow
-  :ref:`Engine.time_scale<class_Engine_property_time_scale>`.
 - Streaming a video from a URL is not supported.
 - Only mono and stereo audio output is supported. Videos with 4, 5.1 and 7.1
   audio channels are supported but down-mixed to stereo.
@@ -183,7 +181,7 @@ maximize the quality of the output Ogg Theora video, but this can require a lot
 of disk space.
 
 `FFmpeg <https://ffmpeg.org/>`__ (CLI) is a popular open source tool
-for this purpose. FFmpeg has a steep learning curve, but it's powerful tool.
+for this purpose. FFmpeg has a steep learning curve, but it's a powerful tool.
 
 Here are example FFmpeg commands to convert an MP4 video to Ogg Theora. Since
 FFmpeg supports a lot of input formats, you should be able to use the commands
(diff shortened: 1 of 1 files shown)


berarma:
> What specifically in this PR only applies to 4.5?

I've removed some limitations. I've taken the opportunity to rework some of the text I did previously, and reference Shutter Encoder.

skyace65:
Got it. This needs to be rebased since I just merged #10721. Also is there a reason this is marked as a draft?

berarma:
Oh, sorry, I had forgot about that PR. There's another limitation to remove. I'll do it in this PR.

It's marked as a draft because one of the new features wasn't yet merged. Now that it is I'll update this PR.

The reason I've reworked some of my previous additions to this page is that I feel a bit insecure writing documentation. When I read some parts of it, it feels too verbose or more intricate than necessary. Your judgement on this would help me. Thanks.

berarma:
Description updated.

berarma:
After testing Shutter Encoder myself I've decided against including it in the docs because [it uses the worst possible encoding parameters and offers no way to change them](https://github.com/paulpacifico/shutter-encoder/issues/298).

tutorials/animation/playing_videos.rst
```diff
@@ -202,10 +200,9 @@ below with almost any input video format (AVI, MOV, WebM, …).
 
 .. warning::
 
-   All FFmpeg releases before Feb 20th, 2025 could produce bad video streams
-   due to a couple of bugs. It's highly recommended to use one of the latest
-   static daily builds, or build FFmpeg from their master branch where they're
-   already fixed.
+   Current official FFmpeg releases have some bugs in their Ogg/Theora
+   multiplexer. It's highly recommended to use one of the latest static daily
+   builds, or build from their master branch where they are already fixed.
```
skyace65: Could you please specify an exact version here? Something like "at the time of writing daily builds are fixed, expect that anything newer than 7.1.1 should be fine." But it doesn't have to be worded like that exactly. Just something so that in the future people aren't wondering what exact version fixed the issue.

skyace65 (changes requested):
Looks good! Only one minor change needs to be made.

tutorials/animation/playing_videos.rst
berarma: I don't know. It could be fixed in a hypotetical 7.1.2 (less likely), in 7.2.0 (more likely?), or in 8.0 (most probably). I don't know how the release process works. I avoided mentioning versions because of this.

The FFmpeg project page recommends that end users use the git source anyway because it will have the latest fixes. The releases are meant for repackagers and integrators. Maybe we should do the same.

berarma: This is from ffmpeg.org:

> Approximately every 6 months the FFmpeg project makes a new major release. Between major releases point releases will appear that add important bug fixes but no new features. Note that these releases are intended for distributors and system integrators. Users that wish to compile from source themselves are strongly encouraged to consider using the development branch (see above), this is the only version on which FFmpeg developers actively work. The release branches only cherry pick selected changes from the development branch, which therefore receives much more and much faster bug fixes such as additional features and security patches. 

Maybe saying something along the lines of "it's recommended to build from the sources using the development branch or use the daily builds to get the latest features and fixes" as a permanent notice. What do you think?

skyace65: How about this, do we have a link to the (presumably closed) theora issue in their repository? Or whatever PR fixed it? If we do I'd like it to be linked in a doc comment. That way in 6 months or whenever we can check to see "is this in a release yet?" I just don't like idea of an indefinite "build from source" warning with no easy way to see if the issue fix has been included in a release yet.

berarma: I had initially linked to the closed issues in their [Trac page](https://trac.ffmpeg.org), but [I was suggested that it was too much](https://github.com/godotengine/godot-docs/pull/10614#discussion_r1972490661), and I agreed. Also, it doesn't seem to provide information on which release fixes them.

Note that although FFmpeg suggests building from source, the [download page](https://ffmpeg.org/download.html) has links to daily builds under "More downloading options". So it isn't mandatory to build from source.

These are the issues:

https://trac.ffmpeg.org/ticket/11451
https://trac.ffmpeg.org/ticket/11454

They don't work with PRs, they send patches to a mailing list instead.

These are the commits fixing those issues:

FFmpeg/FFmpeg@22aa71d4da37a4ad2b0d28deeace64b57aa2ef50
FFmpeg/FFmpeg@84d85e7ad4ace228265af0c8c5caccc0730042fd
FFmpeg/FFmpeg@6e26f57f672b05e7b8b052007a83aef99dc81ccb

I don't think there's an easy way to know that they're included in a release.

skyace65: I think it's too much to have them as regular text. But as a doc comment so maintainers can see it should be fine. Since FFmpeg doesn't have any specific way for us to tell based on the issue I'd say make a doc comment that links to the commits. In the future we'd be able to download the source code for whatever the newest release is and check if the file changes from the commits are there. It's not easy, but not exactly hard either.

berarma:
I've moved the doc comment below the warning that it refers to so that it's more clear where it belongs. I've added the commit refs to the comment and a minor rewording of the warning.

skyace65:
Thanks!

PR #10982, opened 2025-06-01
doc: Fix phrasing

matthieucan:
I believe the intent here was to specify "debugging", as both options are forms of "exporting".


mhilbrunner:
Hey, thanks for taking the time to contribute!

I don't think this is the intent. The editor can actually export projects, the export templates are just able to run exported projects, not export them.

Both can have debugging on or off, as stated in the line above that.

matthieucan:
Hi @mhilbrunner, thanks for the quick feedback, and apologies for the confusion!

PR #10983, opened 2025-06-01
Use SpeedScale in Player.cs in part 9 of 3d tutorial

JRahmatiNL:
The C# alternative for the Godot code, was using a non existing `PlaybackSpeed` property instead of the same property used by the GDScript alternative: speed_scale (SpeedScale in C#).

As a result the C# alternative would not run, while the GDScript alternative worked just fine.


skyace65 (approved):

skyace65:

PR #10990, opened 2025-06-03
Remove warning on build caused by non-existent pygment

### LINKED ISSUE #10989: Building godot docs produces WARNING: Pygments lexer name 'gitignore' is not known
**Your Godot version:**

This is HEAD of docs at the moment (6b8faf39fcce30a9)

**Issue description:**

When rebuilding Godot's documentation, I found a WARNING that didn't use to happen beforehand for file tutorials/best_practices/version_control_systems.html

When investigating, it was introduced at pull request  #10857 when a snippet of:

`.. code-block:: gitignore`

Was added. I double checked that I had the latest version of pygments and there is not such a code for syntax-highligting of git. This is not a me-only issue, as official docs can be seen with this code block not being correctly highlighted.

https://docs.godotengine.org/en/latest/tutorials/best_practices/version_control_systems.html

**Suggested fix:**

There is, however, a package called `pygments-git` that -if installed-, could provide the potentially desired highlighting, but with the slightly different name:

`.. code-block:: git-ignore`

Resulting in this:

Image

**Let me know if you prefer the added package to requirements for correct highlighting, or falling back to a more  generic "unixconfig" or "ini" supported code block**. I would love to get rid of warnings on build (happy to send either pull request). 💙

jynus:
At #10857 a gitignore codeblock was added. Addition seems correct, but a pygment formatting was defined as:

.. code-block:: gitignore

No such a pygment exists, and although there is an aditional package that could parse git-related files, we chose to not add an extra package dependency for building the docs.

Fallback to unixconfig instead.

Closes #10989

### DIFF
--- a/tutorials/best_practices/version_control_systems.rst
+++ b/tutorials/best_practices/version_control_systems.rst
@@ -130,7 +130,7 @@ order to access the LFS files.
 Below is an example ``.gitattributes`` file that you can use as a starting point for Git LFS. 
 These file types were chosen because they are commonly used, but you can modify the list to include any binary types you may have in your project.
 
-.. code-block:: gitignore
+.. code-block:: unixconfig
 
     # Normalize EOL for all files that Git considers text files.
     * text=auto eol=lf


jynus:
Here is the list of alternatives: https://pygments.org/languages/

Calinou (approved):
Looks good to me. `unixconfig` is a close enough match (it'll gray out comment lines starting with `#`).

skyace65 (approved):

skyace65:
Thanks!

PR #10994, opened 2025-06-05
Fix check of effect_callback_type

Radivarig:
Fix for the hardcoded check of `effect_callback_type` and makes the example code work after switching it from the UI.

### DIFF
--- a/tutorials/rendering/compositor.rst
+++ b/tutorials/rendering/compositor.rst
@@ -247,7 +247,7 @@ this at the right stage of rendering.
 
     # Called by the rendering thread every frame.
     func _render_callback(p_effect_callback_type, p_render_data):
-        if rd and p_effect_callback_type == EFFECT_CALLBACK_TYPE_POST_TRANSPARENT and _check_shader():
+        if rd and p_effect_callback_type == effect_callback_type and _check_shader():
             # Get our render scene buffers object, this gives us access to our render buffers.
             # Note that implementation differs per renderer hence the need for the cast.
             var render_scene_buffers: RenderSceneBuffersRD = p_render_data.get_render_scene_buffers()


Radivarig:
@BastiaanOlij can this be merged?

BastiaanOlij:
Hmmm, it was hardcoded in the example on purpose.

Often when you write a compositor effect, you do so for a specific callback type, or possibly a limited number of callback types. Making it run for any callback type even those its not designed for, leads to issues. Hence its double checked in the callback.

It is an unfortunate side effect that the Godot property system allows the user to override the callback type whether desired or not.

Radivarig:
@BastiaanOlij I see, then we can hide those properties and make type subsets as custom variables.

Would this be acceptable?

image

```gdscript
func _init() -> void:
	needs_normal_roughness = true
	effect_callback_type_ = EFFECT_CALLBACK_TYPE_PRE_TRANSPARENT
	notify_property_list_changed() # initial _validate_property

const HIDDEN_PROPERTIES := [
	'effect_callback_type',
	'needs_motion_vectors',
	'needs_normal_roughness',
	'access_resolved_color',
	'access_resolved_depth',
]

func _validate_property(property: Dictionary) -> void:
	if property.name in HIDDEN_PROPERTIES:
		property.usage = PROPERTY_USAGE_NO_EDITOR

	if property.name == 'effect_callback_type_':
		property.hint = PROPERTY_HINT_ENUM
		property.hint_string = ','.join([
			'Pre Transparent:' + str(EFFECT_CALLBACK_TYPE_PRE_TRANSPARENT),
			'Post Transparent:' + str(EFFECT_CALLBACK_TYPE_POST_TRANSPARENT),
		])

@export var effect_callback_type_ := EFFECT_CALLBACK_TYPE_PRE_TRANSPARENT:
	set(value):
		effect_callback_type_ = value
		effect_callback_type = value

```

PR #10996, opened 2025-06-05
Add specular occlusion documentation to StandardMaterial3D tutorial

LunaCapra:
This adds documentation for both the "Disable Specular Occlusion" toggle as well as bent normal maps.

- *Production edit: See https://github.com/godotengine/godot/pull/89988
and https://github.com/godotengine/godot/pull/106145.*


tutorials/3d/img/spatial_material_bentnormals.webp
Calinou, line 1: This is a bit large for the documentation. Here's a version compressed to 264 KB: 
[spatial_material_bentnormals.zip](https://github.com/user-attachments/files/20618106/spatial_material_bentnormals.zip)

It's very slightly lossy (WebP at quality 100 and preprocess dithering enabled to reduce banding), but is more than twice as small.

tutorials/3d/standard_material_3d.rst
```diff
@@ -414,6 +419,41 @@ and wider compatibility.
...
+
+When creating a bent normal map, there are three things required for it to
+work correctly in Godot:
+
+* A **cosine distribution** of rays has to be used when baking.
+* The texture must be created in **tangent space**.
+* The bent normal map needs to use the X+, Y+ and Z+ coordinates, this is
+  known as OpenGL style. If you've imported a material made to be used with
+  another engine it may be DirectX style, in which case the bent normal map
+  needs to be converted so its Y axis is flipped.
```
Calinou:
```suggestion
* The bent normal map needs to use the X+, Y+ and Z+ coordinates, which is
  known as OpenGL-style. If you've imported a material made to be used with
  another engine, it may be DirectX-style. In this case, the bent normal map
  needs to be converted so its Y axis is flipped. This can be achieved by
  checking **Normal Map Invert Y** in the Import dock.
```

LunaCapra: Normal Map Invert Y in the import settings seems to be marked as deprecated in v4.5 dev5 and it says that the same result can be achieved by setting the green channel mapping to inverted green instead. Should I document that instead?

tutorials/3d/img/spatial_material_bentnormals.webp
hsvfan-jan: I think this could be zoomed or cropped in way more, so that it's easier to focus on the differences. Especially on smaller screens this helps

tutorials/3d/standard_material_3d.rst
Calinou: Yes, I'd point to the new setting.

tutorials/3d/standard_material_3d.rst
```diff
@@ -414,6 +419,43 @@ and wider compatibility.
...
+
+Godot only uses the red and green channel of a bent normal map for better
+compression and wider compatibility.
+
+When creating a bent normal map, there are three things required for it to
+work correctly in Godot:
+
+* A **cosine distribution** of rays has to be used when baking.
+* The texture must be created in **tangent space**.
+* The bent normal map needs to use the X+, Y+ and Z+ coordinates, this is
```
AThousandShips:
```suggestion
* The bent normal map needs to use the X+, Y+, and Z+ coordinates, this is
```
Please use [Oxford comma](https://docs.godotengine.org/en/latest/contributing/documentation/docs_writing_guidelines.html#use-the-oxford-comma-to-enumerate-anything).

tutorials/3d/standard_material_3d.rst
```diff
@@ -414,6 +419,43 @@ and wider compatibility.
...
+
+* Indirect diffuse lighting is made to match global illumination more closely.
+* If specular occlusion is enabled, it is calculated using the bent normals and
+  ambient occlusion instead of just from ambient light.
+  This includes screen-space ambient occlusion (SSAO) and other sources of
+  ambient occlusion.
+
+.. image:: img/spatial_material_bentnormals.webp
+
+Godot only uses the red and green channel of a bent normal map for better
```
AThousandShips:
```suggestion
Godot only uses the red and green channels of a bent normal map for better
```

tutorials/3d/standard_material_3d.rst
LunaCapra: This was copied from the normal map section so I wasn't sure whether to keep it as it is or not

AThousandShips: It should be adjusted to follow the style rules, the other case can be fixed separately

Calinou (approved):
Looks good to me.

We should probably include instructions on baking bent normal maps from various programs in a future PR, as the process for doing so is *much* less standardized (and less well-documented) than standard normal maps.

v1993:
> We should probably include instructions on baking bent normal maps from various programs in a future PR, as the process for doing so is _much_ less standardized (and less well-documented) than standard normal maps.

To elaborate on this, from software suggested in the original PR:

* xNormal doesn't work on Linux even under Wine/Proton
* ArmorPaint documentation points to bent normal maps requiring experimental builds and an RT-capable GPU, plus the program itself seems to not be particularly robust (Vulkan validation layer errors, crash when just trying to open a menu)
* Blender support requires quite some manual setup and is apparently documented/implemented in a sole Stack Exchange [answer](https://blender.stackexchange.com/a/158804) + Github Gist from 2019; it's also unclear if its results match the format Godot expects.
* Commercial solutions are... not nearly as readily available to the average user and also lack native Linux versions; preliminary research suggests that at least Adobe products might run okay under Proton. Knald's documentation website also seems to be down and I was unable to find anything about anyone trying to get it working under Wine.

LunaCapra:
>  it's also unclear if its results match the format Godot expects.

It does match the format, I've checked the code. The only really "obscure" requirement is cosine weighted hemisphere sampling for the ambient occlusion/bent normal baking process, which to my knowledge is usually an optimization for nicer results in most places due to how the rendering equation includes the cosine of the light angle already. It is also the default in the programs I've used to bake AO and bent normals.

> and also lack native Linux versions

There's actually an exception to this and it's on steam. It offers both [Substance Painter](https://store.steampowered.com/app/3366290/Substance_3D_Painter_2025/) and [Substance Designer](https://store.steampowered.com/app/3366300/Substance_3D_Designer_2025/) - albeit at the usual fairly big price - as a perpetual license with native linux builds, which I have been using to generate my bent normal maps.

I have also been thinking about how if anyone could write a more readily available tool to bake ambient occlusion maps, the bent normal maps would practically come for free as well since the bent normals are derived from the same process as baking ambient occlusion, it just stores the average direction of the incoming ambient light instead of just its brightness. But I'm not sure how feasible that would be overall, though I did come across https://github.com/Fewes/BakerBoy which implements this kind of thing in unity.

skyace65 (approved):

skyace65:
Thanks!

PR #10998, opened 2025-06-06
Update Android tools version

dragonhunt02:
Bump up Android tools versions in documentation

Reference:
- https://github.com/godotengine/godot/blob/master/platform/android/java/app/config.gradle


skyace65:

Note to other maintainers. The PRs might be cherry picked but haven't yet, so this shouldn't be labelled for cherry picking, but should also stay off the 4.5 milestone.

PR #11008, opened 2025-06-09
Fix type of 'what' parameter to _Notifications In C# usage example

ptlis:
The type is wrong in the example ¯\_(ツ)_/¯

### DIFF
--- a/tutorials/best_practices/godot_notifications.rst
+++ b/tutorials/best_practices/godot_notifications.rst
@@ -362,7 +362,7 @@ nodes that one might create at runtime.
 
         public override void _Notification(int what)
         {
-            switch (what)
+            switch ((long)what)
             {
                 case NotificationParented:
                     _parentCache = GetParent();


tutorials/best_practices/godot_notifications.rst
```diff
@@ -360,7 +360,7 @@ nodes that one might create at runtime.
             return _parentCache.HasUserSignal("InteractedWith");
         }
 
-        public void _Notification(int what)
+        public void _Notification(long what)
```
raulsntos: This is not correct, the `_Notification` method takes an `int` parameter. But it is missing the `override` keyword.

```suggestion
        public override void _Notification(int what)
```

tutorials/best_practices/godot_notifications.rst
```diff
@@ -360,7 +360,7 @@ nodes that one might create at runtime.
...
         {
             switch (what)
```
raulsntos: However, it's true that the `what` parameter needs to be converted to `long`.

```suggestion
            switch ((long)what)
```

raulsntos (changes requested):

I've reviewed the change, and while the current example certainly needs a fix (so thanks for spotting that!), there are a couple of adjustments needed.

ptlis:
Hi, thanks for that - as you probably guessed I ended up not needing to use this, but was tripped up nevertheless in my exploratory coding.

I've updated the PR to include both requested changes.

raulsntos (approved):
Thanks! Looks good to me.

skyace65 (approved):

skyace65:
Thanks! And  congrats on your first merged PR!

PR #11012, opened 2025-06-10
Clarify FileAccess.get_path(): return value, virtual path format, and usage example

### LINKED ISSUE #10966: Clarify return behavior and format for `FileAccess.get_path()` in class reference
**Your Godot version:**
4.2.2.stable.official

**Issue description:**
The documentation for FileAccess.get_path() currently states only:

Returns the path as a string.

This is vague compared to other FileAccess methods like get_open_error() or get_length(), which include detailed notes on return values and behavior under edge cases. It is not clear:

Whether the path returned is absolute, relative, or virtual (e.g., res://, user://)

What is returned if no file has been opened yet

Whether it can return an empty string, null, or throw an error

This ambiguity can lead to confusion and unnecessary guesswork when developers interact with the file system API.

https://docs.godotengine.org/en/stable/classes/class_fileaccess.html#class-fileaccess-method-get-path

Suggested improvement:
Clarify the following in the method description:

Return value when no file is open

Expected path format (absolute, virtual, etc.)

Example usage if applicable

This would align with the Godot Documentation Style Guide and improve consistency across the FileAccess API reference.

I’ve checked the Godot docs issues and didn’t see a similar request already posted. Happy to submit a pull request for this once approved or discussed.

Xylia-Rippy:
This improves the documentation for `FileAccess.get_path()` by:
- Clarifying that the return value uses Godot's virtual file system (e.g., res://, user://)
- Specifying that it returns an empty string if no file is open
- Confirming that it does not throw an error in this case
- Providing an example of typical usage

This brings `get_path()` in line with the documentation style of other FileAccess methods like `get_length()` and `get_open_error()`, and removes ambiguity for new users working with the file system API.

Closes #10966

### DIFF
--- a/classes/class_fileaccess.rst
+++ b/classes/class_fileaccess.rst
@@ -830,7 +830,22 @@ Text is interpreted as being UTF-8 encoded.
 
 :ref:`String<class_String>` **get_path**\ (\ ) |const| :ref:`🔗<class_FileAccess_method_get_path>`
 
-Returns the path as a :ref:`String<class_String>` for the current open file.
+
+Returns the path that was used to open the current file.
+
+The returned path uses Godot's virtual file system format, such as ``res://`` or ``user://``, depending on how the file was opened using :ref:`FileAccess.open() <class_FileAccess_method_open>`.
+
+If no file is currently open, this method returns an empty string (``""``). It does not throw an error in this case.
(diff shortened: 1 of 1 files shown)


AThousandShips:

PR #11013, opened 2025-06-11
Document LightmapGI rendering limits in Mobile and Compatibility renderers

Calinou:
The limit in Compatibility was determined through testing on 4.4.1.stable. The testing project with 144 LightmapGI instances is able to render all of them at once when using Compatibility, but only 8 in Forward+ and 2 in Mobile.

- See https://github.com/godotengine/godot-proposals/issues/12577#issuecomment-2960734329.

**Testing project:** [test-lightmap-count.zip](https://github.com/user-attachments/files/20681506/test-lightmap-count.zip)

### DIFF
--- a/tutorials/3d/global_illumination/using_lightmap_gi.rst
+++ b/tutorials/3d/global_illumination/using_lightmap_gi.rst
@@ -645,6 +645,14 @@ Reducing LightmapGI artifacts
 
 If you notice LightmapGI nodes popping in and out of existence as the camera
 moves, this is most likely because the engine is rendering too many LightmapGI
-instances at once. Godot is limited to rendering 8 LightmapGI nodes at once,
-which means up to 8 instances can be in the camera view before some of them will
-start flickering.
+instances at once.
+
+Here are the limits for each renderer:
+
+- **Forward+**: Up to 8 LightmapGI nodes can be rendered at once.
+- **Forward Mobile**: Up to 2 LightmapGI nodes can be rendered at once.
+- **Compatibility**: An unlimited number of LightmapGI nodes can be rendered at once.
+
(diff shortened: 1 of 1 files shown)


clayjohn:
See https://github.com/godotengine/godot/pull/106907

The compatibility renderer definitely doesn't support an unlimited number of LightmapGI nodes. I wouldn't promise that we support more than 8 as likely the fact that it doesn't break when more are used is an implementation detail that may change at any time.

tutorials/3d/global_illumination/using_lightmap_gi.rst
```diff
@@ -645,6 +645,14 @@ Reducing LightmapGI artifacts
...
-instances at once. Godot is limited to rendering 8 LightmapGI nodes at once,
-which means up to 8 instances can be in the camera view before some of them will
-start flickering.
+instances at once.
+
+Here are the limits for each renderer:
+
+- **Forward+**: Up to 8 LightmapGI nodes can be rendered at once.
+- **Forward Mobile**: Up to 2 LightmapGI nodes can be rendered at once.
+- **Compatibility**: An unlimited number of LightmapGI nodes can be rendered at once.
```
skyace65, line 654:
```suggestion
- **Compatibility**: No hard limit currently however that may change. Design your scene as if 8 is the maximum
```

@clayjohn would something like this be ok?

clayjohn: No, there is a hard limit, someone just needs to go look at the code to figure out what it is. 

Not knowing the limit is not the same as having an infinite limit.

skyace65:
Closing this due to the lack of an update for over 6 months

PR #11024, opened 2025-06-14
Remove outdated note referencing a file which no longer exists

Empika1:
`core/method_bind_ext.gen.inc` is no longer a file, and the parameter limit for `ClassDB::bind_method` is now greater than 5 (although I'm not sure exactly what it is). This note doesn't make any sense anymore.


skyace65:

PR #11026, opened 2025-06-14

jchengjr77:
I went through the tutorial w VS Code and the function auto-creation works.

Updating the doc language to indicate that the bug may or may not occur (don't want to confuse people).


skyace65 (commented):
One, you left in the docs comment on what engine bug this is referencing. Two, that engine bug is still open. I do not want the docs warning removed until that engine issue is closed. Doing a quick glance at it, it seems like this might only be a C# issue. Please comment on it with what you're seeing. If this does indeed only affect C# I'd be fine with updating the warning to reflect that.

skyace65:
Closing due to the lack of a response.

PR #11030, opened 2025-06-16
Document `--lsp-port` command line argument

HolonProduction:
Descriptions follow the adjusted spelling from https://github.com/godotengine/godot/pull/107607

Also there are definitely other arguments missing which are available in 4.4.1. This page either needs an overall sync or some sort of disclaimer, that the page only documents the most important parameters.


skyace65 (approved):

skyace65:
Agreed on a page cleanup/sync, that definitely needs to be done at some point

PR #11035, opened 2025-06-18
default_key_mapping: Fix "Soft Reload Script" shortcut

mvuets:
In Text Editor/Script Editor "Ctrl + Shift + R" is "Replace in files". The correct shortcut for "Soft Reload Script" is "Ctrl + Alt + R".

Reference: <https://github.com/godotengine/godot/blob/4.4.1-stable/editor/plugins/script_editor_plugin.cpp#L4303>:

    file_menu->get_popup()->add_shortcut(ED_SHORTCUT("script_editor/reload_script_soft", TTRC("Soft Reload Tool Script"), KeyModifierMask::CMD_OR_CTRL | KeyModifierMask::ALT | Key::R), FILE_TOOL_RELOAD_SOFT);


tutorials/editor/default_key_mapping.rst
```diff
@@ -359,7 +359,7 @@ Script editor
 +----------------------+---------------------------------+---------------------------------+----------------------------------------+
 | Save All             | :kbd:`Ctrl + Shift + Alt + S`   | :kbd:`Cmd + Shift + Opt + S`    | ``script_editor/save_all``             |
 +----------------------+---------------------------------+---------------------------------+----------------------------------------+
-| Soft Reload Script   | :kbd:`Ctrl + Shift + R`         | :kbd:`Cmd + Shift + R`          | ``script_editor/reload_script_soft``   |
+| Soft Reload Script   | :kbd:`Ctrl + Alt + R`           | :kbd:`Opt + Cmd + R`            | ``script_editor/reload_script_soft``   |
```
skyace65, line 362:
```suggestion
| Soft Reload Tool Script   | :kbd:`Ctrl + Alt + R`           | :kbd:`Opt + Cmd + R`            | ``script_editor/reload_script_soft``   |
```
So unfortunately, the name of the shortcut has changed and would need to be updated. That also means every other box on this section would need to be updated as well. I don't want to merge this without the rename, but this might be better left for a PR that goes through everything to clean it up. I'm sure more of these are out of date, and also not in the same order in the editor anymore.

skyace65 (changes requested):

tutorials/editor/default_key_mapping.rst
mvuets: Ah, I see. Shall I file a regular issue instead then?

skyace65: Yes, you can open a regular issue if you'd like to.

PR #11038, opened 2025-06-20
fix: corrected propetyinfo field

rosshadden:
Updated Object docs to correct `GDExtensionPropertyInfo#name` which is actually a `StringName`.


AThousandShips:

This is also wrong, the type is `String`, this isn't `GDExtensionPropertyInfo`, it's the method of `Object`, and references this type:
https://github.com/godotengine/godot/blob/5abed52fd9d99c4aab7d8c62d505f9ad618e4533/core/object/object.h#L165

PR #11050, opened 2025-06-23
Various grammar and spelling fixes

AThousandShips:
For context see:
* https://github.com/godotengine/godot/pull/107895

But further fixing some uses of "its"/"it's", also treating individual characters such as `_` as their spoken forms, in this case "underscore", and some fixes to plural/possessive based on context, and some UK vs US English fixes

Also treating all annotations as not being pronounced with `@`, as well as `#` for precompile directives and `_` for methods (i.e. `_ready` is just treated as `ready` for pronounciation, which is the general style, otherwise they'd be "an `_ready`" etc. as it'd be "underscore `ready`")


skyace65 (approved):

skyace65:
Thanks a ton for this!

PR #11052, opened 2025-06-23
Update command line arguments to match master branch

skyace65:
Follow up to #11030. I've gone through https://github.com/godotengine/godot/blob/master/main/main.cpp (options start at line 536) and updated what's in the docs to match what's in there. I haven't gone through all of the descriptions though so there may be some discrepancies. But every option that exists in master exists here.

Some options are marked as deprecated, I haven't changed those since there's no current formatting for that. 

The options ``--accurate-breadcrumbs`` and ``--extra-gpu-memory-tracking`` seem to require debug and dev to be enabled. I'm using the existing ``|debug|`` for them, but let me know if I should do something different.

### DIFF
--- a/tutorials/editor/command_line_tutorial.rst
+++ b/tutorials/editor/command_line_tutorial.rst
@@ -43,17 +43,19 @@ given build type.
 
 **General options**
 
-+----------------------------+-----------------------------------------------------------------------------+
-| Command                    | Description                                                                 |
-+----------------------------+-----------------------------------------------------------------------------+
-| ``-h``, ``--help``         | |release| Display the list of command line options.                         |
-+----------------------------+-----------------------------------------------------------------------------+
-| ``--version``              | |release| Display the version string.                                       |
-+----------------------------+-----------------------------------------------------------------------------+
-| ``-v``, ``--verbose``      | |release| Use verbose stdout mode.                                          |
-+----------------------------+-----------------------------------------------------------------------------+
-| ``-q``, ``--quiet``        | |release| Quiet mode, silences stdout messages. Errors are still displayed. |
-+----------------------------+-----------------------------------------------------------------------------+
++----------------------------+-------------------------------------------------------------------------------+
+| Command                    | Description                                                                   |
++----------------------------+-------------------------------------------------------------------------------+
+| ``-h``, ``--help``         | |release| Display the list of command line options.                           |
++----------------------------+-------------------------------------------------------------------------------+
+| ``--version``              | |release| Display the version string.                                         |
++----------------------------+-------------------------------------------------------------------------------+
+| ``-v``, ``--verbose``      | |release| Use verbose stdout mode.                                            |
++----------------------------+-------------------------------------------------------------------------------+
+| ``-q``, ``--quiet``        | |release| Quiet mode, silences stdout messages. Errors are still displayed.   |
++----------------------------+-------------------------------------------------------------------------------+
+| ``--no-header``            | |release| Do not print engine version and rendering method header on startup. |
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):

tutorials/editor/command_line_tutorial.rst
```diff
@@ -232,6 +279,9 @@ given build type.
 +------------------------------------------------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------+
 | ``--benchmark-file <path>``                                      | |editor| Benchmark the run time and save it to a given file in JSON format. The path should be absolute.                                                |
 +------------------------------------------------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------+
+|``"--test [--help]``                                              | |editor| Run unit tests. Use --test --help for more information.                                                                                        |
```
mhilbrunner: Extra `"` here?

skyace65: Fixed

Calinou:
> The options `--accurate-breadcrumbs` and `--extra-gpu-memory-tracking` seem to require debug and dev to be enabled. I'm using the existing |debug| for them, but let me know if I should do something different.

We could add a new `dev` visibility level for options, but this should be done at an engine level first in 4.6.

**Edit:** The exact `#if` is `defined(DEBUG_ENABLED) || defined(DEV_ENABLED)`, so it's available in all debug builds anyway. We can probably leave things as-is.

tutorials/editor/command_line_tutorial.rst
```diff
@@ -108,6 +118,9 @@ given build type.
 +------------------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------------+
 | ``--headless``                           | |release| Enable headless mode (``--display-driver headless --audio-driver Dummy``). Useful for servers and with ``--script``.                               |
 +------------------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| ``--log-file``                           | |release| "Write output/error log to the specified path instead of the default location defined by the project. <file> path should be absolute or relative to|
```
mhilbrunner: Another stray `"` here, sorry I missed that on the first go.

tutorials/editor/command_line_tutorial.rst
```diff
@@ -195,19 +234,27 @@ given build type.
...
+| ``--export-patch <preset> <path>``                               | |editor| Export pack with changed files only. See --export-pack description for other considerations.                                                   |
++------------------------------------------------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------+
+| ``--patches <paths>``                                            | |editor| List of patches to use with --export-patch. The list is comma-separated.                                                                       |
++------------------------------------------------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------+
+| ``--install-android-build-template``                             | |editor| Install the Android build template. Used in conjunction with --export-release or --export-debug.                                               |
```
AThousandShips, line 256:
```suggestion
| ``--install-android-build-template``                             | |editor| Install an Android build template. Used in conjunction with --export-release or --export-debug.                                                |
```
Matching the phrasing elsewhere

skyace65: That's an exact copy of the text from main.cpp. If it needs to be changed then it needs to be changed there first.

tutorials/editor/command_line_tutorial.rst
skyace65: Fixed

mhilbrunner:
Thank you! Merged.

PR #11054, opened 2025-06-25
Update section about using system libraries in Compiling for Linux/*BSD

Calinou:
- Follow-up to https://github.com/godotengine/godot/pull/106218.

___

- Mention how to use system SDL library now that SDL 3 is used for gamepad support.

Not cherry-pickable to `4.4`, as SDL is only used for gamepad support in 4.5 onwards.


skyace65:
Thanks!

PR #11059, opened 2025-06-27
Update resources.rst to make inner class note more clear

Zshandi:
The term "subclass" is often used as a synonym for "child class", so using it here to refer to an inner class is just confusing, even with the "i.e. an inner class" part. The GDScript Reference page simply uses the term "inner class" and never refers to them as "subclass", so this is also consistent with that.


skyace65:
Thanks!

PR #11060, opened 2025-06-28
Add some notes about importing meshes for use with SoftBody3D

simpkins:
This mentions that users may want to change the default LOD generation settings when importing meshes for use with SoftBody3D.

See godotengine/godot#108025


tutorials/physics/soft_body.rst
```diff
@@ -67,3 +67,12 @@ Play the scene and the cloak should simulate correctly.
 .. image:: img/softbody_cloak_finish.png
 
 This covers the basic settings of softbody, experiment with the parameters to achieve the effect you are aiming for when making your game.
+
+Using Imported Meshes
+~~~~~~~~~~~~~~~~~~~~~
+
+When importing a 3D scene, you can use the **Save to File** option in the mesh import settings to save the mesh as a standalone resource file that you can then attach to ``SoftBody3D`` nodes.
+
+When importing a mesh for use with ``SoftBody3D``, you may also want to change the default LOD generation options.  The default import settings will produce an LOD that merges adjacent faces that are nearly flat with respect to each other, even at very close render distances.  This works well for static meshes, but is often undesirable for use with ``SoftBody3D``, if you want these faces to be able to bend and move with respect to each other, instead of being rendered as a single plane.
```
skyace65:
```suggestion
When importing a mesh for use with ``SoftBody3D``, you may also want to change the
default LOD generation options. The default import settings will produce an LOD
that merges adjacent faces that are nearly flat with respect to each other, even at
very close render distances. This works well for static meshes, but is often
undesirable for use with ``SoftBody3D`` if you want these faces to be able to bend
and move with respect to each other, instead of being rendered as a single plane.
```

simpkins: Sure, your suggestion sounds good to me.  I had assumed that non-wrapping was the convention based on the other lines in this file and the class reference, but wrapping definitely seems preferable.  I'll also tweak this slightly to try and better comply with the writing guidelines from the contributing guide.

skyace65: Not all of the files are as good with this as they should be. And the class reference is an exception to this

skyace65:
Thanks!

PR #11061, opened 2025-06-28
Remove reference to "Normal Split Angle" from the mesh import dialog

simpkins:
This setting was removed in godotengine/godot#98620


simpkins:
Note that I just manually edited the image of the import dialog to remove this setting, since I didn't have the source mesh originally used in it.  Let me know if there is a more preferred way of regenerating this image.

skyace65 (approved):

skyace65:

PR #11066, opened 2025-06-28
Update static_typing.rst

NeonMika:
The documentations said it shows the same callback two times, one time with dynamic typing and one time with type hints. In reality, two different callbacks were shown. This has been fixed.


skyace65:

PR #11067, opened 2025-06-28
Use `SoftBody3D` consistently in the soft body tutorial

simpkins:
This file had only been partially updated after `SoftBody` was renamed to `SoftBody3D` in Godot 4.  This commit changes the remaining uses of to `SoftBody` to `SoftBody3D`.


skyace65 (approved):

skyace65:
Thanks!

PR #11069, opened 2025-06-29
Fix return type of _GetOptionVisibility example in c#

dogboydog:


dogboydog:
Oops, misread the docs here, will raise the PR for this the right way

PR #11070, opened 2025-06-29
GDScript: Update annotation description and development guidelines

dalexeev:
See:

* #9188
* https://github.com/godotengine/godot/pull/107717

### DIFF
--- a/engine_details/architecture/scripting_development.rst
+++ b/engine_details/architecture/scripting_development.rst
@@ -9,30 +9,57 @@ GDScript
 Annotation guidelines
 ~~~~~~~~~~~~~~~~~~~~~
 
-..
-    This description intentionally avoids mention of implementation and
-    compilation details because these are often inconsistent between annotations
+In Godot 3 we had a lot of keywords (including ``tool``, ``export``, ``onready``).
+This had the following disadvantages:
 
-Create annotations for modifiers that act on the script or its code.
-Additionally, create annotations for behavior that is specific to the Godot
-engine and editor; if the primary purpose is to affect the way that the engine
-or editor treats or interacts with the script, implement the token as an
-annotation.
+- Overcomplication of the language grammar.
+- Possible conflicts with user identifiers.
+- Need to touch the parser in a major way to add a new keyword.
 
-Do not create annotations for general programming language features.
+So in Godot 4 we introduced the concept of annotations. An annotation is a modifier
+(an attribute, a marker) that can be applied to an entire script, a declaration,
+a statement, or a location in the source code. Annotations can optionally take
+additional arguments, listed in parentheses. Annotation arguments must be constant
+expressions or string literals (in a few special cases where the argument value
+must be resolved in the parser rather than in the analyzer).
 
-::
+Currently, an annotation target can be a combination of the following
+``GDScriptParser::AnnotationInfo::TargetKind`` flags:
 
-    # Affects how the editor treats this script.
-    @icon("res://path/to/class/icon.svg")
-    
-    # Affects how the engine interacts with this script.
-    @onready var character_name = $Label
++-----------------+----------------------------------------------+-----------------------------+
+| **Target Kind** | **Description**                              | **Example**                 |
++=================+==============================================+=============================+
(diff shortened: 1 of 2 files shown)


tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -449,9 +449,9 @@ GDScript also supports :ref:`format strings <doc_gdscript_printf>`.
 Annotations
 -----------
 
-Annotations are special tokens in GDScript that act as modifiers to a script or
-its code and may affect how the script is treated by the Godot engine or
-editor.
+Annotations are special tokens in GDScript that act as modifiers to an entire script,
```
Meorge, line 452: I recognize it's rather nitpicky, but I feel like it might be better to move "an entire script" to the last item in this list, since it encompasses the most. I can't think of a formal rule in English that specifies this, but my intuition just says that including the largest option first causes the later ones to be a bit more confusing.

dalexeev: I understand what you meant, but I went from up to down. If we invert the direction, `a location in the source code` will appear first, which is also not good in my opinion, since it is a more specific case.

contributing/development/core_and_modules/scripting_development.rst
```diff
@@ -9,30 +9,57 @@ GDScript
...
-For historical reasons, some existing annotations and keywords do not strictly
-follow these guidelines. Choosing between implementing a feature as an
-annotation or as a language keyword is a nuanced decision that should be made
-through discussion with other GDScript developers.
+- **General rule:** Do not use keywords unless they are required by the language grammar
+  and the parser. Do not use annotations for independent syntactic units.
+- Use keywords for class member declarators (``var``, ``const``, ``func``).
+- Use annotations for class member modifiers (``@export``, ``@onready``, ``@abstract``).
+- Use keywords for operators within expressions (``and``), for delimiters in statements
+  (``in``, ``when``) and declarations (``extends``).
```
AThousandShips:
```suggestion
  (``in``, ``when``), and declarations (``extends``).
```

Ivorforce:
This needs a rebase after https://github.com/godotengine/godot-docs/pull/11188.

engine_details/architecture/scripting_development.rst
```diff
@@ -9,30 +9,57 @@ GDScript
 Annotation guidelines
 ~~~~~~~~~~~~~~~~~~~~~
 
-..
-    This description intentionally avoids mention of implementation and
-    compilation details because these are often inconsistent between annotations
+In Godot 3 we had a lot of keywords (including ``tool``, ``export``, ``onready``).
```
Ivorforce, line 12: In a document concerned with the present, i would avoid leading by talking about the past.

I would instead reframe the introduction by explaining what annotations are and what their design goal is. The Godot 3 history background can be in an info box or separate section, if we really need to explain it (which I'm not sure we do).
For example, we could remove everything up until "An annotation is a modifier..."

dalexeev: I partially agree with you. In user documentation, historical context is often redundant (with the exception of migration guides). However, we still sometimes have exceptions for important cases that reference Godot 3 (in infoboxes). For example, we mention that the special behavior of `continue` in `match` was removed in 4.0.

As for this documentation, its difference is that it is intended for contributors, not regular users. The historical context here is necessary to explain the motivation behind the introduction of annotations. What problems existed previously, are now solved by annotations, and whose recurrence we aim to avoid in the future. What problems are annotations intended to solve in the present and future, and what is their scope?

I wouldn't mind clearly separating historical context from the current state of affairs, "rules" for when to use keywords and when to use annotations. Perhaps even putting the context in the infobox, as you suggested. But I think the historical context is essential to understanding the language design we're aiming for.

Ivorforce: Hmhm, yea that makes sense to me. Including the historical context feels more appropriate when I imagine the page as part of the contributing docs, as opposed to when it's in the docs.

mhilbrunner:
Thank you!

PR #11072, opened 2025-07-01
Remove PCK encryption note for Android.

bruvzg:
Since https://github.com/godotengine/godot/pull/105984 was merged, PCK encryption is supported for APK / AAB exports as well.

### DIFF
--- a/contributing/development/compiling/compiling_with_script_encryption_key.rst
+++ b/contributing/development/compiling/compiling_with_script_encryption_key.rst
@@ -22,12 +22,6 @@ with that same key.
     It is absolutely **required** to compile your own export templates to use
     PCK encryption.
 
-.. warning::
-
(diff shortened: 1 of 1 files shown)


Calinou (commented):
Looks good to me.

However, I'm not sure if we should remove the note entirely right now, given how "important" it is. Maybe we should change it to say "Prior to Godot 4.5, ..." instead, so that users on 4.4 who happen to be reading `stable` once 4.5 is released will still be aware of this limitation.

We can remove the note entirely once 4.6 is out instead.

bruvzg:
Not sure if it's important, encryption has extremely limited use, and this note most likely was not and will not be noticed by most users.

mhilbrunner (approved):

mhilbrunner:
Same, I think this is fine.

skyace65:
Thanks!

PR #11074, opened 2025-07-02
Update Pineapple Works entry in consoles.rst

wkubiak:
- UWP got deprecated some time ago by Microsoft so having a GDK native port on Xbox is the only way of doing things now so we remove the GDK highlight next to Xbox platforms
- added Switch 2 and PlayStation 5 to our list of supported platforms


skyace65:
Thanks!

wkubiak:
Thanks for the fast approval @skyace65 :)

PR #11078, opened 2025-07-03
Fix NavigationAgent.height note about being limited to 3D

MJacred:
Is used in 2D additionally to 3D, see code. https://github.com/godotengine/godot/blob/master/modules/navigation_3d/nav_agent_3d.cpp

I reckon this is outdated since this PR:
* https://github.com/godotengine/godot/pull/69988


MJacred:
Ah, I believe what was meant is "works only in 3D games", not "works only in 3D avoidance"

PR #11081, opened 2025-07-03
Fix incorrect notice about multithreading in Introduction to the buildsystem

Calinou:
- Use `-j12` in the example line to reflect the most popular CPU core topology in Steam hardware survey as of writing (6 cores/12 threads).

___

- See https://github.com/godotengine/godot-docs-user-notes/discussions/215#discussioncomment-13656190.


mhilbrunner:
Thanks!

PR #11082, opened 2025-07-04
Document unsupported custom feature tags in EditorExportPlugin in Feature tags

Calinou:
- See https://github.com/godotengine/godot-docs-user-notes/discussions/307#discussioncomment-13663221.

### DIFF
--- a/tutorials/export/feature_tags.rst
+++ b/tutorials/export/feature_tags.rst
@@ -153,6 +153,10 @@ field in the *export preset* used to generate it:
     running the project from the editor, even if the export preset marked as
     **Runnable** for your current platform has custom feature tags defined.
 
+    Custom feature tags are also not used in :ref:`class_EditorExportPlugin`
+    scripts. Instead, feature tags in :ref:`class_EditorExportPlugin`
+    will reflect the device the editor is currently running on.
+
 Overriding project settings
 ---------------------------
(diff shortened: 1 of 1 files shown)


AndreasOM:
Thanks.
This would have saved me hours of work.

Still a bit weird that feature overrides are not applied on export.
I do have a workaround, but it feels very counter intuitive -- leaving quite a few marks from my keyboard on my forehead.

Calinou:
> I do have a workaround, but it feels very counter intuitive -- leaving quite a few marks from my keyboard on my forehead.

What workaround did you use? Maybe we can add it to the documentation this way :slightly_smiling_face:

AndreasOM:
I ended up creating a tiny EditorExportPlugin, that scans for overrides based on the used features, and swaps it into the base settings:

https://github.com/AndreasOM/fiiish-v3/blob/main/godot/addons/omg-export_overrides/omg-export_overrides_editor_export_plugin.gd

Might turn this into a standalone plugin at some point, but I'm just getting started with Godot.

Too be fully honest:
I think the current export behaviour is a bug, not a feature.

mhilbrunner (approved):

skyace65 (approved):

skyace65:
Thanks!

PR #11084, opened 2025-07-04

Feipian:
change $Player.start($StartPosition.position) to
$Player.Start($StartPosition.position)
ensure the project working


Calinou:
This is not the correct fix for GDScript code, which uses the `snake_case` naming convention for method names. Notice how the line just below uses `start()`.

If it doesn't work locally, double-check that you defined the player's start function as `start()` and not `Start()`.

Feipian:
Thank you for the reply

skyace65:
Closing this per Calinou's comment.

PR #11098, opened 2025-07-08
Update custom template images on compiling pages

skyace65:
I've updated the custom template images for the Android, Linux and Windows compiling pages. Instructions have also been updated since you have to enable advanced options to see those settings now. I've removed the architecture note from the windows one. It was added when we added Windows ARM support, but it feels excessive to me. I would expect people to be aware of what they're exporting for, but I'm fine with changing it back if people think I should, it would only take 10 minutes at most to update these images.


Calinou (approved):
Looks good to me.

mhilbrunner (approved):

mhilbrunner:
Looks good to me too. Thank you :)

PR #11099, opened 2025-07-09
Update Standard Material 3D page for new transform options

skyace65:
Documents settings added in https://github.com/godotengine/godot/pull/93142 as well as the particle trails setting, not sure when that was added. The descriptions are straight from the class reference.

I've also removed the transmission section, when I updated the page for 4.0 I moved it upwards since it was renamed to blacklight but forgot to delete the original section. The image from that section has been moved to the backlight section.

### DIFF
--- a/tutorials/3d/standard_material_3d.rst
+++ b/tutorials/3d/standard_material_3d.rst
@@ -767,6 +767,7 @@ Use FOV Override
 Overrides the ``Camera3D``'s field of view angle (in degrees).
 
 .. note::
+
   This behaves as if the field of view is set on a ``Camera3D`` with
   ``Camera3D.keep_aspect`` set to ``Camera3D.KEEP_HEIGHT``. Additionally, it may not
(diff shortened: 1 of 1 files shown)


tutorials/3d/standard_material_3d.rst
```diff
@@ -741,14 +743,34 @@ Point Size
...
+may break with lower scales. Therefore, try to keep this setting as close to 1.0 as
+possible.
+
+Use FOV Override
+~~~~~~~~~~~~~~~~
+
+Overrides the ``Camera3D``'s field of view angle (in degrees).
+
+.. note::
+  This behaves as if the field of view is set on a ``Camera3D`` with
```
AThousandShips, line 771:
```suggestion

  This behaves as if the field of view is set on a ``Camera3D`` with
```

Calinou (approved):
Looks good to me.

mhilbrunner:
Looks good to me. :P

PR #11101, opened 2025-07-09
Update SpaceCast collision_result

### LINKED ISSUE #9737: ShapeCast3D documentation issue
v4.3.rc1.mono.official [e343dbbcc]

ShapeCast3D collision_result documentation it contains the same data is in PhysicsDirectSpaceState3D.get_rest_info, but it contains also collider which is not in the PhysicsDirectSpaceState3D.get_rest_info data (or at least in its documentation). So this collider data needs to be added to one page or the other.

[https://docs.godotengine.org/en/stable/classes/class_shapecast3d.html#class-shapecast3d-property-collision-result](url)

charjr:
Resolves #9737

[Shapecast3D.collision_result](https://docs.godotengine.org/en/3.6/classes/class_shapecast.html#class-shapecast-property-collision-result) claims it returns data the same as `PhysicsDirectSpaceState3D.get_rest_info()`

Checking Godot's Core I can see it returns an **additional** collider field: [shape_cast_3d.cpp](https://github.com/godotengine/godot/blob/d7cc121e6469d0d998bf214e3a87e37961e6f1bb/scene/3d/physics/shape_cast_3d.cpp#L477)

```cpp
Array ShapeCast3D::get_collision_result() const {
	Array ret;

	for (int i = 0; i < result.size(); ++i) {
		const PhysicsDirectSpaceState3D::ShapeRestInfo &sri = result[i];

		Dictionary col;
		col["point"] = sri.point;
		col["normal"] = sri.normal;
		col["rid"] = sri.rid;
		col["collider"] = ObjectDB::get_instance(sri.collider_id);
		col["collider_id"] = sri.collider_id;
		col["shape"] = sri.shape;
		col["linear_velocity"] = sri.linear_velocity;

		ret.push_back(col);
	}
	return ret;
}
```

This is the same for Shapecast2d: [shape_cast_2d.cpp](https://github.com/godotengine/godot/blob/d7cc121e6469d0d998bf214e3a87e37961e6f1bb/scene/2d/physics/shape_cast_2d.cpp#L384)

```cpp
Array ShapeCast2D::get_collision_result() const {
	Array ret;

	for (int i = 0; i < result.size(); ++i) {
		const PhysicsDirectSpaceState2D::ShapeRestInfo &sri = result[i];

		Dictionary col;
		col["point"] = sri.point;
		col["normal"] = sri.normal;
		col["rid"] = sri.rid;
		col["collider"] = ObjectDB::get_instance(sri.collider_id);
		col["collider_id"] = sri.collider_id;
		col["shape"] = sri.shape;
		col["linear_velocity"] = sri.linear_velocity;

		ret.push_back(col);
	}
	return ret;
}
```

### DIFF
--- a/classes/class_shapecast2d.rst
+++ b/classes/class_shapecast2d.rst
@@ -169,7 +169,7 @@ The shape's collision mask. Only objects in at least one collision layer enabled
 
 - :ref:`Array<class_Array>` **get_collision_result**\ (\ )
 
-Returns the complete collision information from the collision sweep. The data returned is the same as in the :ref:`PhysicsDirectSpaceState2D.get_rest_info()<class_PhysicsDirectSpaceState2D_method_get_rest_info>` method.
+Returns the complete collision information from the collision sweep. Returns a :ref:`Dictionary<class_Dictionary>` containing all fields from the :ref:`PhysicsDirectSpaceState2D.get_rest_info()<class_PhysicsDirectSpaceState2D_method_get_rest_info>` method, with an additional `collider` field for the collided :ref:`Object<class_Object>`.
 
 .. rst-class:: classref-item-separator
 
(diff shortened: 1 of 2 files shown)


AThousandShips:

charjr:
> 

Ah, thank you. I'll sort it out once I'm off the clock

PR #11120, opened 2025-07-18
Remove references to reset_physics_interpolation

### LINKED ISSUE #11119: reset_physics_interpolation needs to be marked as deprecated / removed?
**Your Godot version:**
Godot v4.5.beta3

**Issue description:**
When calling `reset_physics_interpolation` the following warning is printed to the console:
```
instance_reset_physics_interpolation() is deprecated.
  <C++ Source>  servers/rendering_server.compat.inc:58 @ _instance_reset_physics_interpolation_bind_compat_104269()
```

https://github.com/godotengine/godot/blob/f92f1ce9c07d35b0e3843415bf99a509d3a51add/servers/rendering_server.compat.inc#L53-L59

Okay, but I'm not sure what "deprecated" means here. Is this no longer necessary to call? Is there some other function to call instead? Looking at the PR where this was added, it seems like it does not need to be called at all anymore.

https://docs.godotengine.org/en/latest/tutorials/physics/interpolation/using_physics_interpolation.html
https://docs.godotengine.org/en/latest/classes/class_node.html#class-node-method-reset-physics-interpolation

TranquilMarmot:
Closes #11119

### DIFF
--- a/tutorials/physics/interpolation/using_physics_interpolation.rst
+++ b/tutorials/physics/interpolation/using_physics_interpolation.rst
@@ -97,7 +97,8 @@ Testing and debugging tips
 
 Even if you intend to run physics at 60 ticks per second, in order to thoroughly test your
 interpolation and get the smoothest gameplay, it is highly recommended to
-temporarily set the physics tick rate to a low value such as 10 ticks per second.
+temporarily set the physics tick rate to a low value such as 10 ticks per second. This can be
+done in the project settings under Physics -> Common -> Physics Ticks per Second.
 
 The gameplay may not work perfectly, but it should enable you to more easily see
 cases where you should be using your own custom interpolation on e.g. a


tutorials/physics/interpolation/using_physics_interpolation.rst
```diff
@@ -92,61 +92,15 @@ As a rough guide:
 .. note:: You can always change the tick rate as you develop, it is as simple as
           changing the project setting.
 
-Call ``reset_physics_interpolation()`` when teleporting objects
```
TranquilMarmot, line 95: @lawnjelly is it correct that this does not need to be done anymore?

lawnjelly: No, it still needs to be done.

There are some auto-resets in some situations, but even in those you will want to manually reset in some situations (standing starts etc).

As such this PR is largely incorrect.

Calinou:
See my comment here:

https://github.com/godotengine/godot-docs/issues/11119#issuecomment-3089803686

Outside of the `classes/` folder (which should not be modified by godot-docs PRs), the current godot-docs repository does not contain any references to `instance_reset_physics_interpolation()`. Therefore, we don't need to make any changes to the documentation.

Moreover, the method was removed entirely from the scripting API in 4.5:

https://github.com/godotengine/godot/blob/f92f1ce9c07d35b0e3843415bf99a509d3a51add/misc/extension_api_validation/4.4-stable.expected#L87-L92

The MultiMesh-related methods remain exposed to scripting as they're still useful.

PR #11123, opened 2025-07-19
Update descriptions of `AHashMap` and `HashMap`. Recommend `AHashMap` over `HashMap` by default.

Ivorforce:
**Note:** Will need to go through a core meeting before this can be merged. (Edit: Approved in core meeting)

`AHashMap` is _way_ faster than `HashMap` in nearly all circumstances. `HashMap` on the other hand has some 'defensive' affordances built in. However, these affordances are not expected for C++ map types: No `std` map type grants these, probably because they are what makes `HashMap` so slow. In addition, most users of map types do not need these affordances. 
In the interest of performance, this changes the recommended map type to `AHashMap`.


dsnopek (approved):
This looks good to me!

mhilbrunner (approved):
Needs a rebase to fix CI :)

Calinou (approved):
Makes sense to me.

skyace65:
Thanks!

PR #11130, opened 2025-07-21
Add missing feature tags `threads` and `nothreads` in documentation table

leandro-benedet-garcia:
The tag `nothread` was added by https://github.com/godotengine/godot/pull/93563, however [this table](https://docs.godotengine.org/en/stable/tutorials/export/feature_tags.html#default-features) does not have both threads tags, this PR adds `threads` and `nothreads` to it.

This PR also adds a note to the gdextension tutorial explaining the order of operation and how they should work.

Discussion about the subject in rocket chat in this link: https://chat.godotengine.org/channel/gdextension/thread/kkLYt23hP93n2xvtP

### DIFF
--- a/tutorials/export/feature_tags.rst
+++ b/tutorials/export/feature_tags.rst
@@ -108,6 +108,10 @@ Here is a list of most feature tags in Godot. Keep in mind they are **case-sensi
 +--------------------+----------------------------------------------------------+
 | **web**            | Host OS is a Web browser                                 |
 +--------------------+----------------------------------------------------------+
+| **nothreads**      | Running without threading support                        |
++--------------------+----------------------------------------------------------+
+| **threads**        | Running with threading support                           |
++--------------------+----------------------------------------------------------+
 | **web_android**    | Host OS is a Web browser running on Android              |
 +--------------------+----------------------------------------------------------+
 | **web_ios**        | Host OS is a Web browser running on iOS                  |
--- a/tutorials/scripting/gdextension/gdextension_file.rst
+++ b/tutorials/scripting/gdextension/gdextension_file.rst
@@ -66,6 +66,16 @@ Here is an example of what that can look like:
     linux.release.rv64 = "res://bin/libgdexample.linux.template_release.rv64.so"
 
 
+Entries are matched in order, so if two sets of feature tags could match
+the same system, be sure to put the more specific ones first:
+
+.. code-block:: none
+
+    [libraries]
+
+    linux.release.editor.x86_64 = "res://bin/libgdexample.linux.template_release.x86_64.so"
+    linux.release.x86_64 = "res://bin/libgdexample.linux.noeditor.template_release.x86_64.so"
+
 Here are lists of some of the available built-in options (for more look at the :ref:`feature tags <doc_feature_tags>`):
 
 Running system


tutorials/scripting/gdextension/gdextension_file.rst
```diff
@@ -68,6 +68,16 @@ Here is an example of what that can look like:
 
 Here are lists of some of the available built-in options (for more look at the :ref:`feature tags <doc_feature_tags>`):
 
+Tags that have a `no` suffix (like `nothreads`) should be placed after their
+counterparts since they are evaluated in order. Like so:
```
dsnopek: If the `nothreads` feature tag is supported (in Godot 4.3+), you don't need to worry about the order - any order should be fine.

It's if you're trying to support Godot 4.1 and 4.2 (which don't have `nothreads`) then you need to make sure to put the single-threaded one second (and not use the `nothreads` feature tag, because it doesn't exist):

```
    web.release.threads.wasm32 = "res://bin/libgdexample.web.template_release.wasm32.wasm"
    web.release.wasm32 = "res://bin/libgdexample.web.template_release.wasm32.nothreads.wasm"
```

So, using a tag that doesn't have an opposite might be a better example of the way these are evaluated in order

leandro-benedet-garcia: Ooh, I see.

I am using the `editor` tag now as an example

tutorials/scripting/gdextension/gdextension_file.rst
```diff
@@ -68,6 +68,18 @@ Here is an example of what that can look like:
 
 Here are lists of some of the available built-in options (for more look at the :ref:`feature tags <doc_feature_tags>`):
 
+To use a build that is not compatible with a tag and doesn't have a `no`
+counterpart (like `nothreads` and `threads`), they should be placed after the
+build that is compatible with said feature, for example, if you wish to use a
+build without `editor` support, you may do like so:
```
dsnopek: The explanation here is kind of complex. Here's an attempt at simplifying:

```suggestion
Libraries are processed in order, so if two sets of feature tags could match the same system, be sure to put the more specific ones first:
```

Also, this whole section should probably go above the previous paragraph:

> Here are lists of some of the available built-in options (for more look at the :ref:`feature tags <doc_feature_tags>`):

That's meant to be above the headings that come after (hence it ending on a colon)

Calinou (approved):
Looks good to me.

tutorials/scripting/gdextension/gdextension_file.rst
```diff
@@ -66,6 +66,16 @@ Here is an example of what that can look like:
     linux.release.rv64 = "res://bin/libgdexample.linux.template_release.rv64.so"
 
 
+Libraries are processed in order, so if two sets of feature tags could match
```
Ivorforce:
```suggestion
Entries are matched in order, so if two sets of feature tags could match
```

I'm not sure why, but this would be easier to understand for me.

tutorials/export/feature_tags.rst
```diff
@@ -108,6 +108,10 @@ Here is a list of most feature tags in Godot. Keep in mind they are **case-sensi
 +--------------------+----------------------------------------------------------+
 | **web**            | Host OS is a Web browser                                 |
 +--------------------+----------------------------------------------------------+
+| **nothreads**      | Running without threads support                          |
```
Ivorforce:
```suggestion
| **nothreads**      | Running without threading support                        |
```

tutorials/export/feature_tags.rst
```diff
@@ -108,6 +108,10 @@ Here is a list of most feature tags in Godot. Keep in mind they are **case-sensi
...
++--------------------+----------------------------------------------------------+
+| **threads**        | Running with threads support                             |
```
Ivorforce: This is how it's put in the command line for this feature, at least.

```suggestion
| **threads**        | Running with threading support                           |
```

Ivorforce (approved):
I only got debatable nitpicks, looks good to me :)

skyace65 (approved):

skyace65:

PR #11131, opened 2025-07-21
Manual's csharp event handling instructions fixed. The previous versi...

atherman2:
…on didn't work with csharp.

Changes made to csharp section at scene organization at best practices at manual. It concerns csharp event handling instructions. The previous instructions didn't work with csharp.

### DIFF
--- a/tutorials/best_practices/scene_organization.rst
+++ b/tutorials/best_practices/scene_organization.rst
@@ -63,11 +63,13 @@ initialize it:
 
      .. code-tab:: csharp
 
-       // Parent
-       GetNode("Child").Connect("SignalName", Callable.From(ObjectWithMethod.MethodOnTheObject));
-
        // Child
-       EmitSignal("SignalName"); // Triggers parent-defined behavior.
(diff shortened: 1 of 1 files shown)


raulsntos:
The previous version of the instructions should work on C# and they match the GDScript instructions, although they may not be the best practice to follow. And keep in mind the names for the signal (`SignalName`) and method (`ObjectWithMethod.MethodOnTheObject`) are just placeholders.

It may be a good idea to show how to declare the signal in the child, as you do in your changes. However, this would also apply to the GDScript example. To be honest, I don't think the goal of the example is to show you how to use signals, there's a dedicated [C# Signals](https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_signals.html) page that goes more in depth about that.

paulloz:
In addition, it can be confusing to show the delegate definition and the `EmitSignal()` call in the same scope. We've had similar issues on other pages with readers learning C# not understanding this wouldn't be actual valid code.

skyace65:
Closing this per raulsntos and paulloz's comments and the lack of a response

PR #11135, opened 2025-07-23
Link to `Why not STL?` section in FAQ from core engine containers.

Ivorforce:
We've just had someone find this section, and wonder why we don't use STL.
It seems sensible to link to the "Why not STL?" section from here specifically.


Calinou (approved):
Looks good to me.

skyace65:
Thanks!

PR #11140, opened 2025-07-24
Document that VisionOS is supported

skyace65:
Pretty much does what the title says. I've added VisionOS to the list of features page in the XR section, I've also linked to that page from the Intro to XR page. It seemed less clunky than adding a note or warning box to the intro page. I've also updated a screenshot on the exporting projects page.

This PR does NOT cover exporting to VisionOS itself right now. I assume the process is near identical to what's on the iPhone export page but I'm not comfortable setting up a page based on a guess. There's also some odd stuff related to exporting that should be documented. Since it's built on top of what's already in place for iPhone's there's options in the VisionOS export that only make sense in that context such as "Performance Gaming Tier" and "Performance A 12".


about/list_of_features.rst
```diff
@@ -663,6 +665,11 @@ XR support (AR and VR)
 
   - Including support for popular stand alone headsets like the Meta Quest 1/2/3 and Pro, Pico 4, Magic Leap 2, and Lynx R1.
 
+- Out of the box limited support for VisionOS Apple Headsets.
+
+  - Currently only exporting a application for use on a flat plane within the
+    headset is supported. Immersive experiences are not supported.
```
Calinou: visionOS is has a lowercase "v" as per the [official site](https://www.apple.com/os/visionos/):

```suggestion
- Out of the box limited support for visionOS Apple headsets.

  - Currently only exporting a application for use on a flat plane within the
    headset is supported. Immersive experiences are not supported.
```

skyace65:
Fixed

about/list_of_features.rst
```diff
@@ -663,6 +665,11 @@ XR support (AR and VR)
 
   - Including support for popular stand alone headsets like the Meta Quest 1/2/3 and Pro, Pico 4, Magic Leap 2, and Lynx R1.
 
+- Out of the box limited support for visionOS Apple headsets.
+
+  - Currently only exporting a application for use on a flat plane within the
```
hsvfan-jan: I think this should be "an application"

skyace65:
Fixed

mhilbrunner:
poke @BastiaanOlij

tutorials/xr/setting_up_xr.rst
```diff
@@ -9,7 +9,12 @@ Introduction to the XR system in Godot
 Godot provides a modular XR system that abstracts many of the different XR platform specifics away from the user.
 At the core sits the :ref:`XRServer <class_xrserver>` which acts as a central interface to the XR system that allows users to discover interfaces and interact with the components of the XR system.
 
-Each supported XR platform is implemented as an :ref:`XRInterface <class_xrinterface>`. Supported interfaces register themselves with the :ref:`XRServer <class_xrserver>` and can be queried with the ``find_interface`` method on the :ref:`XRServer <class_xrserver>`. When the desired interface is found it can be initialized by calling ``initialize`` on the interface.
+Each supported XR platform is implemented as an :ref:`XRInterface <class_xrinterface>`.
+A list of supported platforms can be found on the list of features page :ref:`here <doc_xr_support>`.
+Supported interfaces register themselves with the :ref:`XRServer <class_xrserver>`
+and can be queried with the ``find_interface`` method on the :ref:`XRServer <class_xrserver>`.
+When the desired interface is found it can be initialized by calling ``initialize``
```
BastiaanOlij, line 16: Owh dang, this is a little outdated now. It's still correct and how it was designed to work, but for OpenXR we've had to early initialise in the core. While we talk about that elsewhere, wonder if we should add a note about it here.

Totally offtopic for VisionOs though so something to do separately.

BastiaanOlij:
My main concern here is that in 4.5, we don't really have any XR specific functionality for VisionOS.
We're just supporting it as a platform to export to so you can run normal Godot games and apps in the shared environment.

In 4.6 when we actually should get specific XR functionality added to this, the doc changes here will already become outdated. Might still be worth adding as this is where people will look but IDK..

about/list_of_features.rst
```diff
@@ -663,6 +665,11 @@ XR support (AR and VR)
 
   - Including support for popular stand alone headsets like the Meta Quest 1/2/3 and Pro, Pico 4, Magic Leap 2, and Lynx R1.
 
+- Out of the box limited support for visionOS Apple headsets.
+
+  - Currently only exporting an application for use on a flat plane within the
+    headset is supported. Immersive experiences are not supported.
+
```
stuartcarnie, line 672: Apple describes iPhone and iPad apps compiled for visionOS as _Compatible Apps_:

* [Create for visionOS](https://developer.apple.com/visionos/get-started/#create)

  > Note: People can also run your existing iPadOS or iOS app as a **compatible app** in visionOS. Your app appears as a single, scalable window in the person’s surroundings.
* [Making you app compatible with visionOS](https://developer.apple.com/documentation/visionos/making-your-app-compatible-with-visionos)

We could include these links to give developers more information.

skyace65:
> My main concern here is that in 4.5, we don't really have any XR specific functionality for VisionOS.
> We're just supporting it as a platform to export to so you can run normal Godot games and apps in the shared environment.

Yes but since it's in the engine now, we should be telling people somewhere that it's not full support. The stuff in this PR also isn't hard to update when full support comes. The XR page change is just pointing at the list of features for supported platforms, which I think we should have been doing anyway, and we'd just have to update a small amount of text in the list of features when full XR functionality is available.

BastiaanOlij:
@skyace65  I agree we should have mention of VisionOS support in the manual, the question is whether this belongs in the XR sections of the manual as we do not have specific XR functionality yet.

I was under the impression we already had manual changes addressing VisionOS support around the platform that either Recardo or Stuart submitted. Or maybe those are still outstanding PRs?

skyace65:
@BastiaanOlij there have been no opened or merged documentation PRs for the manual regarding visionOS other than this one.

I think it makes sense to mention it in the XR system requirements section. If we leave it out I think people will just assume that it wasn't updated, rather than assume it doesn't actually have full functionality

BastiaanOlij:
> @BastiaanOlij there have been no opened or merged documentation PRs for the manual regarding visionOS other than this one.

I stand corrected, I thought that had happened but guess not. 
 
> I think it makes sense to mention it in the XR system requirements section. If we leave it out I think people will just assume that it wasn't updated, rather than assume it doesn't actually have full functionality

Ok, we just need to make sure to update it again once the features are there :)

mhilbrunner:
@BastiaanOlij So we can go ahead and merge this?

BastiaanOlij:
> @BastiaanOlij So we can go ahead and merge this?

Yes I think thats ok, maybe just check with @stuartcarnie , he made a suggestion for a change that got no reaction.

stuartcarnie:
@mhilbrunner @BastiaanOlij is referring to the following:

- https://github.com/godotengine/godot-docs/pull/11140#discussion_r2240950153

My suggestion was to consider including those links, but it is not required.

skyace65:
> @mhilbrunner @BastiaanOlij is referring to the following:
> 
>     * [Document that VisionOS is supported #11140 (comment)](https://github.com/godotengine/godot-docs/pull/11140#discussion_r2240950153)
> 
> 
> My suggestion was to consider including those links, but it is not required.

I'm not sure what exactly the benefit would be with including that link, as anyone with an existing iOS app in Godot would be able to just use the VisionOS export option instead of iOS.

stuartcarnie:
> I'm not sure what exactly the benefit would be with including that link, as anyone with an existing iOS app in Godot would be able to just use the VisionOS export option instead of iOS.

Totally agree, which is why I didn't feel too strongly about it 👍🏻

stuartcarnie (approved):

BastiaanOlij (approved):

mhilbrunner:
Well, in this goes :P Thanks, and thanks for all your reviews!

PR #11142, opened 2025-07-25
Improve descriptions of `LocalVector`, recommend it when appropriate

beicause:
The current documentation doesn't highlight the advantages of `LocalVector` and may lead to misunderstandings and discourage its use.
In fact I rarely noticed the differences between `LocalVector` and `Vector` before, until seeing many PRs in 4.5 replacing `Vector` with `LocalVector`. I think we should provide a clearer explanation of `LocalVector` in the documentation.

### DIFF
--- a/engine_details/architecture/core_types.rst
+++ b/engine_details/architecture/core_types.rst
@@ -146,6 +146,7 @@ scripting API.
 +-----------------------+--------------------------+---------------------------------------------------------------------------------------+
 | |vector|              | ``std::vector``          | **Use this as the "default" vector type.** Uses copy-on-write (COW) semantics.        |
 |                       |                          | This means it's generally slower but can be copied around almost for free.            |
+|                       |                          | Use ``LocalVector`` instead where COW isn't needed and performance matters.           |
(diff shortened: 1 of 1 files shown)


smix8:
The main benefit of LocalVector, apart from no copy-on-write, is that a LocalVector keeps its allocated memory. So when you frequently edit and clear stuff and it is used internally only you want to use a LocalVector to not waste performance on all the frequent memory alloc.

beicause:
>The main benefit of LocalVector, apart from no copy-on-write, is that a LocalVector keeps its allocated memory. So when you frequently edit and clear stuff and it is used internally only you want to use a LocalVector to not waste performance on all the frequent memory alloc.

Will this be changed? Related: https://github.com/godotengine/godot/pull/105928#pullrequestreview-2940389572

beicause:
> The main benefit of LocalVector, apart from no copy-on-write, is that a LocalVector keeps its allocated memory. So when you frequently edit and clear stuff and it is used internally only you want to use a LocalVector to not waste performance on all the frequent memory alloc.

I updated the description to address this.

mhilbrunner:
Merged. Thank you, and thanks for the reviews!

PR #11156, opened 2025-07-30
Fix code tab title in `SurfaceTool` tutorial

Cykyrios:
There was a missing empty line after the `.. code-tab::` line from #6405, causing all following lines to be part of the tab name.
I took this opportunity to have a look through the docs, and this was the only instance I found; there are some places where spaces are left in the empty line, but they don't cause issues, so I left them.


skyace65:
Thanks!

PR #11160, opened 2025-07-31
Fix incorrect Intel IGP minimum requirement in System requirements

Calinou:
Intel Broadwell IGPs can run Forward+/Mobile on Linux thanks to Mesa drivers, but the Windows driver has never supported Vulkan nor Direct3D 12 with a sufficient feature level (and shader model).

This increases the listed requirement to use Forward+/Mobile to Skylake (released in 2015).

We could opt into listing Broadwell specifically for Linux, but it would make the table a lot more complex for a rare configuration nowadays (and one that's probably better served by Compatibility performance-wise anyway).

- See https://github.com/godotengine/godot/issues/109071.


skyace65 (approved):

skyace65:
Going to leave this open for a bit for discussion. But I agree it would be excessive to edit the chart to list it as supported only on Linux.

mhilbrunner:
Thank you!

PR #11164, opened 2025-08-02
Update c_sharp_exports.rst to remove note about typed dictionary support

AtelierFerrofell:
The note,

> Typed dictionaries are currently unsupported in the Godot editor, so the Inspector will not restrict the types that can be assigned, potentially resulting in runtime exceptions.

appears to be obsolete as of the current stable release. I tried various ways to get around inspector type restrictions, but could not.

image

The dropdown only accepts the specified type.

image

"Quick Load..." filters for the specified type.

image

"Load..." will not allow you to manually insert objects of the wrong type.

Therefore, I believe it's safe to remove this note.


mhilbrunner (approved):

mhilbrunner:
Thank you! Merged.

PR #11173, opened 2025-08-04
[Contributing] Update suggested Ubuntu version

AThousandShips:
As of 4.5 the minimum requirements are "Linux distribution released after 2018", this aligns with that (SCons versions on 16.04 also don't apppear to work)

* Closes: https://github.com/godotengine/godot/issues/109303

### DIFF
--- a/contributing/development/compiling/compiling_for_linuxbsd.rst
+++ b/contributing/development/compiling/compiling_for_linuxbsd.rst
@@ -332,7 +332,7 @@ Building export templates
 .. warning:: Linux binaries usually won't run on distributions that are
              older than the distribution they were built on. If you wish to
              distribute binaries that work on most distributions,
-             you should build them on an old distribution such as Ubuntu 16.04.
+             you should build them on an old distribution such as Ubuntu 20.04.
              You can use a virtual machine or a container to set up a suitable
              build environment.
 


Calinou:
I would go all the way and list Ubuntu 22.04, as both Ubuntu 18.04 and 20.04 are EOL since April 2025.

Ubuntu 22.04 should cover most distributions in use by now, especially since Debian `trixie` is releasing very soon (which means Debian oldstable, i.e. `bookworm` will still be supported).

AThousandShips:
The official minimum requirements remain 2018, so changing this to 22.04 risks causing issues for people, but I can change it if it's not going to cause issues with that regardless of the system requirements

bruvzg:
Maybe we should recommend using SDK instead of outdated Ubuntu versions - https://github.com/godotengine/buildroot 

If I remember correctly the only part required to support old distros is `glibc` version, official builds use 2.28, Ubuntu 18.04 use 2.27, and Ubuntu 22.04 use 2.35.

domdfcoding:
I would be miffed if the minimum glibc version went beyond 2.31, as I'm still actively using Ubuntu 20.04 with an ESM subscription. I don't mind if the official builds won't run, as I build from source anyway, so as long as I can continue to do that for the foreseeable future I'm happy.

Is a 2018 to 2020 jump in requirements perhaps more reasonable in short term? How long ago did the 2018 requirement come in (i.e. how many years behind have been supported in the past)?

Calinou:
> I would be miffed if the minimum glibc version went beyond 2.31, as I'm still actively using Ubuntu 20.04 with an ESM subscription. I don't mind if the official builds won't run, as I build from source anyway, so as long as I can continue to do that for the foreseeable future I'm happy.

I forgot Ubuntu 20.04 + ESM was still supported, so I guess it makes sense to keep listing 20.04 for now.

contributing/development/compiling/compiling_for_linuxbsd.rst
```diff
@@ -332,7 +332,7 @@ Building export templates
 .. warning:: Linux binaries usually won't run on distributions that are
              older than the distribution they were built on. If you wish to
              distribute binaries that work on most distributions,
-             you should build them on an old distribution such as Ubuntu 16.04.
+             you should build them on an old distribution such as Ubuntu 18.04.
```
Calinou:
```suggestion
             you should build them on an old distribution such as Ubuntu 20.04.
```

bruvzg: glibc is `2.31` in 20.04, so slightly less compatible than official builds.

AThousandShips:
I'll update to 20.04 and look at a separate PR suggesting using build root as a follow up for both 4.5 and 4.4!

mhilbrunner (approved):

skyace65 (approved):

skyace65:
Thanks!

AThousandShips:
Thank *you*!

PR #11175, opened 2025-08-04
Fix incorrect `[ul bullet=...]` description in BBCode in RichTextLabel

Calinou:
- Mention `[br]` behavior when used in a list item.

___

- See https://github.com/godotengine/godot-docs-user-notes/discussions/141#discussioncomment-13981107.

### DIFF
--- a/tutorials/ui/bbcode_in_richtextlabel.rst
+++ b/tutorials/ui/bbcode_in_richtextlabel.rst
@@ -278,7 +278,9 @@ Reference
       | ``[p {options}]{text}[/p]``
 
   * - | **br**
-      | Adds line break in a text.
+      | Adds line break in a text, without adding a new paragraph.
+        If used within a list, this won't create a new list item,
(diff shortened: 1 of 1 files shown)


tutorials/ui/bbcode_in_richtextlabel.rst
```diff
@@ -278,7 +278,9 @@ Reference
       | ``[p {options}]{text}[/p]``
 
   * - | **br**
-      | Adds line break in a text.
+      | Adds line break in a text. If used within a list, this won't
```
bruvzg:
```suggestion
      | Adds line break in a text, without adding a new paragraph. If used within a list, this won't
```

`[br]` will add a break, but it is similar to automatic line breaks added to fit control width, not a new paragraph which is processed independently and can have different properties (alignment, direction, language).

mhilbrunner:
Thank you!

PR #11178, opened 2025-08-05
Update running_code_in_the_editor.rst to fix error in example code

AtelierFerrofell:
The example code does not account for instances where new_resource/value becomes null, which happens when the resource is cleared or reset in the inspector or through code. This will cause an error: `Invalid access to property or key 'changed' on a base object of type 'Nil'.` whenever the resource becomes null.

The fix checks to make sure the resource has not become null before attempting to connect the new signal, just like it's already doing before attempting to disconnect the old signal.


skyace65 (approved):

skyace65:
Thanks!

PR #11183, opened 2025-08-06
Add exclamation point to example

Nadav-Ruskin:
"!" was never concat'ed

### DIFF
--- a/tutorials/best_practices/godot_notifications.rst
+++ b/tutorials/best_practices/godot_notifications.rst
@@ -292,7 +292,7 @@ values will set up according to the following sequence:
 
     public:
         String get_test() { return test; }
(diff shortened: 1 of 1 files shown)


tutorials/best_practices/godot_notifications.rst
```diff
@@ -292,7 +292,7 @@ values will set up according to the following sequence:
 
     public:
         String get_test() { return test; }
-        void set_test(String p_test) { return test = p_test; }
+        void set_test(String p_test) { return test = p_test + "!"; }
```
Calinou: The `return` here was also extraneous, as this method doesn't return a value:

```suggestion
        void set_test(String p_test) { test = p_test + "!"; }
```

skyace65:
Thanks!

PR #11184, opened 2025-08-06
Update RISC-V toolchain version

MBCX:
Should've done this way earlier to be honest.

I've got reports from users who have tested my [binaries](https://drive.google.com/drive/folders/1APeB0xDLsHNMjoTnY7Xv0IHywcpshmf4?usp=sharing) and reported to be slow, and that's almost certainly because of the toolchain (I'm compiling with LLVM 21.0.0git).

`2023.07.07` I think is the best middle ground. RISC-V devices are still niche, and while I can't confirm this in anyway, I don't believe it would be easy to find devices older than 2023. Besides, this version of the SDK contains many performance improvements for a much, much more usable experience.

### DIFF
--- a/contributing/development/compiling/compiling_for_linuxbsd.rst
+++ b/contributing/development/compiling/compiling_for_linuxbsd.rst
@@ -397,8 +397,8 @@ To cross-compile Godot for RISC-V devices, we need to setup the following items:
   While we are not going to use this directly, it provides us with a sysroot, as well
   as header and libraries files that we will need. There are many versions to choose
   from, however, the older the toolchain, the more compatible our final binaries will be.
-  If in doubt, `use this version <https://github.com/riscv-collab/riscv-gnu-toolchain/releases/tag/2021.12.22>`__,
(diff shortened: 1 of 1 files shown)


skyace65 (approved):
As someone whose been keeping a close eye on RISC-V ecosystem I think this makes sense. I don't think anyone considers the current hardware something you will use long term compared to say a Raspberry Pi 5, especially since nothing today works with RV23.

skyace65:
@aaronfranke any thoughts?

skyace65:
Thanks!

MBCX:
> > Should've done this way earlier to be honest.
> 
> For compatibility reasons, it's good to compile against an older version when possible. However, if it's causing issues, then yes we should update.

I thought about adding a note saying that you're not forced to only one SDK (I personally have custom builds of the 2022.08 and 2025.01 SDKs), and that you can compile a latest - 2025.01 - export templates and a "compatibility" - 2022.08 - export templates, which is what I'll personally do.

But that's probably overkill if you _just_ want to make your game available to RISC-V users.

PR #11196, opened 2025-08-10
Update class_vector2.rst

SasaSuppenknecht:
Fixed mix up in description of cross product for Vector2

### DIFF
--- a/classes/class_vector2.rst
+++ b/classes/class_vector2.rst
@@ -561,7 +561,7 @@ Returns a new vector with all components clamped between ``min`` and ``max``, by
 
 Returns the 2D analog of the cross product for this vector and ``with``.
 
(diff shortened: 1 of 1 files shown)


AThousandShips:

AThousandShips:
Also I'm not sure this is accurate, have you considered the fact that this is a coordinate system that has -Y up and not down

SasaSuppenknecht:
> 

Ohh, pardon me! I thought it was supposed to be done in godot-doc, since that is also be the place for raising issues with the documentation. I will do better next time!


> Also I'm not sure this is accurate, have you considered the fact that this is a coordinate system that has -Y up and not down

Ooohh, I see. I have been scratching my head for the past 20 minutes wondering where I missed that, but it's directly in the vector math documentation... thanks for letting me know!

PR #11198, opened 2025-08-13
Add note about bug when exporting with custom boot splash on Godot 4.3 and 4.4.

domdfcoding:
This addresses the bug in https://github.com/godotengine/godot/issues/109355 that had already been unknowingly fixed in 4.5, and doesn't affect 4.2 and earlier as it needs the `editor_hint` feature. I suggested in that issue only noting in the docs rather than backporting the fix as it is easy to work around.


AThousandShips:

AThousandShips:
Additionally we don't add notes to the class reference about past bugs, this should be documented in the online manual for 4.4 and older instead if anything, under export probably

PR #11201, opened 2025-08-13
Update shading_language.rst

stuckupfool:
I'm proposing a small change to the shading language reference docs to make it clear that per-instance uniform *arrays* are not supported. Currently the docs mention that per-instance uniforms do not support *textures* and proceeds to give an example involving uniform texture arrays. However when I read the docs, it wasn't clear to me that non-texture uniform arrays (e.g: `instance uniform vec3 my_arr[3]`) are also not supported.

### DIFF
--- a/tutorials/shaders/shader_reference/shading_language.rst
+++ b/tutorials/shaders/shader_reference/shading_language.rst
@@ -1104,7 +1104,7 @@ method on a node that inherits from :ref:`class_GeometryInstance3D`:
 
 When using per-instance uniforms, there are some restrictions you should be aware of:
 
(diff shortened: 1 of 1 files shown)


stuckupfool:
Thanks @Calinou.

For reference, the shader editor surfaces this error when using instance with a uniform array:
image

Calinou:
> For reference, the shader editor surfaces this error when using instance with a uniform array:

This makes sense, although we should probably change that error message to be more user-friendly (it kind of reads like an internal error right now).

mhilbrunner:
Well, in this goes. Thank you!

PR #11202, opened 2025-08-14
velocity.length should not be > but >=

Falling10fruit:
because non diagonal movement is only equal to one


aaronfranke:
No, that's wrong. Your change would make this `if` statement always true.

PR #11149, opened 2025-07-27
Overhaul "Getting Started - Your first 3D game" images

Piralein:
Update all images for consistent look, similar size, version 4.5 and add more yellow outlines as help.

PR #11203, opened 2025-08-14
Optimize PNG images using oxipng

aaronfranke:
This PR losslessly optimizes the PNG images in this repo with `oxipng`. This decreases the image sizes, at the cost of increasing the Git history size. Specifically, the command used is `oxipng -o max --zopfli --cache --fix` where `--cache` is a feature I added in my fork: https://github.com/oxipng/oxipng/pull/564

Follow-up to this PR: https://github.com/godotengine/godot-docs/pull/5254

Alternatively, we could switch to WebP images, but unless somebody does that, this PR is available.

Other PRs changing images like #11149 should be merged ahead of this one.


Calinou:
> Alternatively, we could switch to WebP images, but unless somebody does that, this PR is available.

I would prefer switching all PNGs to lossless WebP, which is in fact something I did in the past (...but I apparently never opened a PR for). Lossless WebP basically always wins compared to an highly optimized PNG (although lossy PNG with pngquant might still win, but it's lossy at this point).

This script should do it:

```bash
# Install `parallel`, `cwebp` and `sd` first.
shopt -s nullglob globstar
parallel cwebp -z 9 {} -o {.}.webp ::: {about,community,contributing,getting_started,tutorials}/**/*.png
sd ".png" ".webp" {about,community,contributing,getting_started,tutorials}/**/*.rst
# Review replacements and revert those that acted on non-img fields (it will replace text instances in tutorials too).

# The top-level `img/` folder is excluded as it's linked within the engine class reference.
```

skyace65:
I'd rather we not do this, and just continue to convert stuff to webp as we go. For some pngs that aren't going to change this makes sense, pictures of stuff that's outside of the editor. But for stuff like the translation tutorials the proper solution to the files being too large should be to replace them with up to date files that are webp.

Edit: Also worth pointing out, I very frequently find that the webp's we replace pngs with do end up being larger, which we knew would happen when we decided to move to that format. So again, if we do this, then replace those outdated images with webps eventually we're going to be right back where we started in terms of image size. If that's a big problem right now, we should be figuring out a permanent solution, not a one time fix like this that will have less benefits over time as the docs are updated.

PR #11214, opened 2025-08-17
Make explicit that string comparison is by default case sensitive

belzecue:


belzecue:
The problem here is, if you click inside the Godot editor to go to the doco for e.g. String.begins_with(), it does not state in the method description if the search is case sensitive or not.  You have to scroll up to the class description, and even then it's only inferred (as it is now) by talking about variant methods that are case insensitive.  In my case, I just wanted a quick answer: is begins_with() case sensitive or not?  It would be overkill to add this clarification to every string method description, so making it explicit in the class description is the best option, IMO.

"are reversed, and start from the end of the string, instead of the beginning."

This is misleading.  The string itself is not 'reversed'; the meaning of the 'r' variant method is 'reverse' as in 'reverse search'.  As is, the doco implies a string is first reversed and then searched from the end of the string, which would return the same as the non-r method!  In my edit I did not explicitly state that the 'r' method variation stands for 'reversed' because it's implicit in the description of what the method does: start from the end of the string instead of the beginning.  And there is no clarification righ now in the class description for that 'n' means (I assume it means 'not case sensitive').  But this could be made explicit in the edit if desired: i.e. that 'r' means 'reversed'.

AThousandShips:

PR #11215, opened 2025-08-18
Move Renderers page at the top of the Rendering section

Calinou:
- Rename to "Overview of renderers" to be more descriptive.

This page is usually meant to be read first to give an overview, while other pages are on more specific subjects.


mhilbrunner:
Thank you!

PR #11217, opened 2025-08-19
Improve the pygments lexer to cover GDScript 2 better

Qubus0:
we've [recently updated the gdscript lexer](https://github.com/GodotModding/pygments-gdscript) for our own documentation, this is a port of those changes to fit the godot docs

please check as many pages as you can for errors i might have missed. 

also let me know if the colors are fine like this. i think some of the existing colors may be from an older theme and not quite up to date, but i didn't touch them for now.

main changes are functions having proper blue color now, global functions being purple, all classes getting the correct greens (including builtins and types having the different green), and proper colors for $String, string names and node paths


Qubus0:
<details>
<summary>Side by side screenshot</summary>

image

LexerUpdate2

LexerUpdate3

</details>

_extensions/gdscript.py
```diff
@@ -9,45 +9,37 @@
...
 class GDScriptLexer(RegexLexer):
     """
-    For `GDScript source code <https://www.godotengine.org>`_.
+    For GDScript source code.
     """
 
     name = "GDScript"
-    aliases = ["gdscript", "gd"]
+    url = "https://www.godotengine.org"
+    aliases = ["gdscript", "gd", "gdscript2", "gd2"]
```
Calinou: I would remove the aliases we don't use:

```suggestion
    aliases = ["gdscript", "gd"]
```

This lexer should still do a good enough job with Godot 3.x GDScript code anyway, so the distinction isn't really relevant.

Qubus0: ah yep, we're running the full own lexer here. in our case we had a conflict with the builtin pygments lexer forcing us to add an alias. will remove

_extensions/gdscript.py
```diff
@@ -160,15 +93,14 @@ def innerstring_rules(ttype):
                         "super",
                         "yield", # Reserved for potential future use.
                         # Not really keywords, but used in property syntax.
-                        "set",
-                        "get",
+                        # also colored like functions, not keywords
+                        # "set",
```
AThousandShips:
```suggestion
                        #"set",
```
No leading space for commented out code

_extensions/gdscript.py
```diff
@@ -160,15 +93,14 @@ def innerstring_rules(ttype):
...
+                        # "get",
```
AThousandShips:
```suggestion
                        #"get",
```

_extensions/gdscript.py
```diff
@@ -193,7 +126,17 @@ def innerstring_rules(ttype):
                 Keyword.ControlFlow,
             ),
         ],
+
```
AThousandShips:
```suggestion
```
No space for the other entries

_extensions/gdscript.py
```diff
@@ -440,50 +1406,172 @@ def innerstring_rules(ttype):
...
+            (r"(\.)([a-zA-Z_]\w*)\b(?!\s*\()", bygroups(Operator, Name.Variable.Instance)),
+            include("operator"),
+            include("keywords"),
+            (r"(func)(\s+)", bygroups(Keyword, Whitespace), "funcname"),
+            include("functions"),
+            # NOTE:
+            #   This matches all PascalCase as a class. If this raises issues
+            #   please report it.
+            # see: https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html#naming-conventions
+            # (r"\s*([A-Z][a-zA-Z0-9_]*)", Name.Class),
```
AThousandShips:
```suggestion
            #(r"\s*([A-Z][a-zA-Z0-9_]*)", Name.Class),
```

_extensions/gdscript.py
```diff
@@ -440,50 +1406,172 @@ def innerstring_rules(ttype):
                 Name.Decorator,
             ),
         ],
-        "numbers": [
+        "operator": [
```
AThousandShips, line 1409: Any reason this is singular as opposed to other cases?

_extensions/gdscript.py
```diff
@@ -440,50 +1406,172 @@ def innerstring_rules(ttype):
...
             (
-                r"(-)?((\d|(?<=\d)_)+\.(\d|(?<=\d)_)*|(\d|(?<=\d)_)*\.(\d|(?<=\d)_)+)([eE][+-]?(\d|(?<=\d)_)+)?j?",
-                Number.Float,
+                r"!=|==|<<|>>|&&|\+=|-=|\*=|/=|%=|&=|\|=|\|\||:=|[-~+/*%=<>&^.!|$]",
```
AThousandShips: Note that the GDScript reference doesn't consider `:=` an operator strictly speaking, and it wasn't in the previous version for that reason IMO

_extensions/gdscript.py
```diff
@@ -440,50 +1406,172 @@ def innerstring_rules(ttype):
...
+                Operator,
             ),
-            (r"(-)?(\d|(?<=\d)_)+[eE][+-]?(\d|(?<=\d)_)+j?", Number.Float),
-            (r"(-)?0[xX]([a-fA-F0-9]|(?<=[a-fA-F0-9])_)+", Number.Hex),
+            (r"(in|and|or|not)\b", Operator.Word),
```
AThousandShips:
```suggestion
            (r"(in|is|and|as|or|not)\b", Operator.Word),
```
You dropped two operators here

_extensions/gdscript.py
Qubus0: i guess removing it won't make a big difference since `=` is covered as single operator and `:` is covered by punctuation
```suggestion
                r"!=|==|<<|>>|&&|\+=|-=|\*=|/=|%=|&=|\|=|\|\||[-~+/*%=<>&^.!|$]",
```

_extensions/gdscript.py
Qubus0: good question. somewhat arbitrary. `name`, `stringescape`, `comment` and a few others are singular too. for the most part plural would make sense, except `whitespace` seems awkward. perhaps singular everywhere makes the most sense?

Qubus0:
Name and Punctuation didn't have css classes yet. adding them makes the () and = bluish, pretty hard to notice, and fixes code in warning callouts being yellow

image

_extensions/gdscript.py
Calinou: I would use singular everywhere. Most highlighters seem to do this in their local definitions.

Calinou (commented):
Looks great now!

Dark | Light
-|-
Screenshot_20250829_225651 | Screenshot_20250829_225706

I can confirm the code color in admonitions is fixed now.

Qubus0:
singulars done, squashed everything, should be ready - let me know if anything else stands out to you

_extensions/gdscript.py
```diff
@@ -148,27 +81,27 @@ def innerstring_rules(ttype):
...
-                        "trait", # Reserved for potential future use.
+                        "trait",  # Reserved for potential future use.
                         "var",
                         # Other keywords.
                         "await",
                         "breakpoint",
                         "self",
                         "super",
-                        "yield", # Reserved for potential future use.
+                        "null",  # type, but colored like a keyword
```
Qubus0: added null. color-wise correct, does it make sense semantically too?
(yield was not removed, not sure why it shows red here but not in the full file view)

Qubus0: nvm, docs say it's a literal moved it to literals. which don't have a css color yet, getting that fixed too

Calinou: This matches the GDScript editor behavior, so I'd say it makes sense.

Qubus0:
okay, realized most of the light theme colors were outdated, so i brought everything closer to the editor theme. including the yellowed background which the editor doesn't have (improves clarity imo, might split some opinions though)

|before|after|
|---|---|
|image|image|

Qubus0:
Updated preview
<details>

| dark | light |
|---|---|
|image|image|
|note the function declaration colors above are fixed but | everything below still has the wrong ones but i didn't want to redo everything|
|image|image|
|image|image|
|image|image|
|image|image|

</details>

Qubus0:
fixed member vars color and `...args` 
|dark|light|
|---|---|
|image|image|

code not changed, just added for simpler screenshot

_extensions/gdscript.py
```diff
@@ -391,8 +333,1032 @@ def innerstring_rules(ttype):
...
-        "decorators": [
+            # copied from https://docs.godotengine.org/en/stable/classes/index.html
+            (
+                words(
+                    (
+                        # Nodes
+                        "Node",
+                        "AcceptDialog",
+                        "AnimatableBody2D",
+                        "AnimatableBody3D",
```
mhilbrunner, line 344: I have concerns about this. It feels like this will grow out of date quickly and would need constant updates. If there is any way at all to automate this part, I'd appreciate it.

mhilbrunner:
Thanks for working on this! Overall it looks quite good to me, besides the single concern noted above.

_extensions/gdscript.py
Qubus0: While I agree automation would be nice (and I'll have a look into it), realistically I don't think this going out of date is much of an issue because 

1. The most important classes have had the same name since 4.x, they aren't renamed particularly often. 
2. New classes are rarely added to the docs right away
3. Classes are rarely removed between major updates too
4. If there are any outdated classes, the hue will only be off slightly
5. And considering all of that together, the benefit of covering most of the important classes far outweighs  risking missing a few less important ones

mhilbrunner: Those are some fair points. :) Maybe we can make grabbing the list from that URL during the build a future todo then, and see how it goes.

mhilbrunner:
Thank you! Merged.

akien-mga:
Cherry-picked for 4.5.

PR #11220, opened 2025-08-21
COLOR is now in, use ALBEDO and ALPHA instead

Illiander42:
Not got everything, someone with more knowledge should fix.

### DIFF
--- a/tutorials/shaders/your_first_shader/your_first_2d_shader.rst
+++ b/tutorials/shaders/your_first_shader/your_first_2d_shader.rst
@@ -107,7 +107,8 @@ an input variable to the fragment function and the final output from it.
 .. code-block:: glsl
 
   void fragment(){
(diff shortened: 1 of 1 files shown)


AThousandShips (changes requested):
This is incorrect, there is no such fragment value in `CanvasItem` shaders, you're thinking of `Spatial` shaders, see the [shader reference](https://docs.godotengine.org/en/latest/tutorials/shaders/shader_reference/canvas_item_shader.html#fragment-built-ins)

But even if this was the correct change you still need to change the text as well, the text talks about `COLOR` so it will be very confusing

Illiander42:
I guess I got confused between "3d shaders" and volumetric shaders, and assumed that 2d shaders are the normal ones for meshes.  That's what I get for skimming.

AThousandShips:

PR #10546, opened 2025-01-24
Add "Upgrading from Godot 4.3 to Godot 4.4" page

raulsntos:
This documents all the breaking changes in 4.4, like we did for previous versions:

- https://github.com/godotengine/godot-docs/pull/9250
- https://github.com/godotengine/godot-docs/pull/8488
- https://github.com/godotengine/godot-docs/pull/7611

Area maintainers should review the changes listed for their area. Also feel free to suggest any addition that I might have missed or make follow-up PRs.

_Updated as of commit godotengine/godot@39c201ca58a8ddf1e99e62c8453294a2662a2fd8_

_Preview: https://github.com/raulsntos/godot-docs/blob/breaking_changes_4.4/tutorials/migrating/upgrading_to_godot_4.4.rst_

PR #8488, opened 2023-11-16
Add "Upgrading from Godot 4.1 to Godot 4.2" page

raulsntos:
Adds the migration guide for 4.2, just like we did for 4.1 (https://github.com/godotengine/godot-docs/pull/7611).

Last time we didn't include GDExtension compatibility because 4.1 completely broke compatibility, we added a warning at the top of the page instead. I'm not sure how we want to handle GDExtension compatibility this time, the @godotengine/gdextension team can suggest changes or make a follow-up PR.

Area maintainers should review the changes listed for their area. Also feel free to suggest any addition that I might have missed or make follow-up PRs.

PR #11223, opened 2025-08-21
Add "Upgrading from Godot 4.4 to Godot 4.5" page

raulsntos:
This documents all the breaking changes in 4.5, like we did for previous versions:

_Updated as of commit godotengine/godot@bfa330dd5dc6f4894efc956db1be20e4bc2d0064_

_Preview: https://github.com/raulsntos/godot-docs/blob/breaking_changes_4.5/tutorials/migrating/upgrading_to_godot_4.5.rst_

### DIFF
--- a/tutorials/migrating/upgrading_to_godot_4.5.rst
+++ b/tutorials/migrating/upgrading_to_godot_4.5.rst
@@ -1,0 +1,259 @@
+.. _doc_upgrading_to_godot_4.5:
+
+Upgrading from Godot 4.4 to Godot 4.5
+=====================================
+
+For most games and apps made with 4.4 it should be relatively safe to migrate to 4.5.
+This page intends to cover everything you need to pay attention to when migrating
+your project.
+
+Breaking changes
+----------------
+
+If you are migrating from 4.4 to 4.5, the breaking changes listed here might
+affect you. Changes are grouped by areas/systems.
+
+.. warning::
+
+    In order to support `new Google Play requirements`_ Android now requires
+    targeting .NET 9 when exporting C# projects to Android, other platforms
+    continue to use .NET 8 as the minimum required version but newer versions
+    are supported and encouraged.
+
+    If you are using C# in your project and want to export to Android, you will
+    need to upgrade your project to .NET 9 (see `Upgrading to a new .NET version`_
+    for instructions).
+
+This article indicates whether each breaking change affects GDScript and whether
+the C# breaking change is *binary compatible* or *source compatible*:
+
+- **Binary compatible** - Existing binaries will load and execute successfully without
+  recompilation, and the run-time behavior won't change.
+- **Source compatible** - Source code will compile successfully without changes when
+  upgrading Godot.
+
+Core
+~~~~
+
+========================================================================================================================  ===================  ====================  ====================  ============
+Change                                                                                                                    GDScript Compatible  C# Binary Compatible  C# Source Compatible  Introduced
+========================================================================================================================  ===================  ====================  ====================  ============
+**JSONRPC**
+Method ``set_scope`` replaced by ``set_method`` optional parameter                                                        |❌|                 |✔️ with compat|      |✔️ with compat|      `GH-104890`_
+**Node**
+Method ``get_rpc_config`` renamed to ``get_node_rpc_config``                                                              |❌|                 |✔️ with compat|      |✔️ with compat|      `GH-106848`_
+Method ``set_name`` changes ``name`` parameter type from ``String`` to ``StringName``                                     |✔️|                 |✔️ with compat|      |✔️ with compat|      `GH-76560`_
+========================================================================================================================  ===================  ====================  ====================  ============
+
+Rendering
+~~~~~~~~~
+
+========================================================================================================================  ===================  ====================  ====================  ============
+Change                                                                                                                    GDScript Compatible  C# Binary Compatible  C# Source Compatible  Introduced
+========================================================================================================================  ===================  ====================  ====================  ============
+**DisplayServer**
+Method ``file_dialog_show`` adds a new ``parent_window_id`` optional parameter                                            |✔️|                 |✔️ with compat|      |✔️ with compat|      `GH-98194`_
+Method ``file_dialog_with_options_show`` adds a new ``parent_window_id`` optional parameter                               |✔️|                 |✔️ with compat|      |✔️ with compat|      `GH-98194`_
+**RenderingDevice**
+Method ``texture_create_from_extension`` adds a new ``mipmaps`` optional parameter                                        |✔️|                 |✔️ with compat|      |✔️ with compat|      `GH-105570`_
+**RenderingServer**
+Method ``instance_reset_physics_interpolation`` removed                                                                   |❌|                 |✔️ with compat|      |✔️ with compat|      `GH-104269`_
+Method ``instance_set_interpolated`` removed                                                                              |❌|                 |✔️ with compat|      |✔️ with compat|      `GH-104269`_
+========================================================================================================================  ===================  ====================  ====================  ============
+
+.. note::
+
+    In C#, the enum ``RenderingDevice.Features`` breaks compatibility because of the way the bindings generator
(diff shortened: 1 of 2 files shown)


tutorials/migrating/upgrading_to_godot_4.5.rst
```diff
@@ -0,0 +1,258 @@
...
+Property ``sparse_indices_component_type`` changes type from ``int`` to ``GLTFAccessor::GLTFComponentType``               |✔️|                 |❌|                  |❌|                  `GH-106220`_
+Property ``sparse_values_byte_offset`` changes type metadata from ``int32`` to ``int64``                                  |✔️|                 |❌|                  |❌|                  `GH-106220`_
+**GLTFBufferView**
+Property ``byte_length`` changes type metadata from ``int32`` to ``int64``                                                |✔️|                 |❌|                  |❌|                  `GH-106220`_
+Property ``byte_offset`` changes type metadata from ``int32`` to ``int64``                                                |✔️|                 |❌|                  |❌|                  `GH-106220`_
+Property ``byte_stride`` changes type metadata from ``int32`` to ``int64``                                                |✔️|                 |❌|                  |❌|                  `GH-106220`_
```
raulsntos, line 87: This seem like a big breaking change, and these APIs seem to have existed for a while (since 4.0). I don't know if there are many C# users that were using them and would be affected, but it seems we should avoid breaking compatibility in these APIs.

cc @godotengine/import

tutorials/migrating/upgrading_to_godot_4.5.rst
```diff
@@ -0,0 +1,258 @@
...
+========================================================================================================================  ===================  ====================  ====================  ============
+**TextServerExtension**
+Method ``_font_draw_glyph`` adds a new ``oversampling`` optional parameter                                                |❌|                 |❌|                  |❌|                  `GH-104872`_
+Method ``_font_draw_glyph_outline`` adds a new ``oversampling`` optional parameter                                        |❌|                 |❌|                  |❌|                  `GH-104872`_
+Method ``_shaped_text_draw`` adds a new ``oversampling`` optional parameter                                               |❌|                 |❌|                  |❌|                  `GH-104872`_
+Method ``_shaped_text_draw_outline`` adds a new ``oversampling`` optional parameter                                       |❌|                 |❌|                  |❌|                  `GH-104872`_
```
raulsntos, line 142: It seems users could not have been implementing `TextServerExtension` in a scripting language (like GDScript or C#) because it has required virtual methods that use pointers (e.g. `shaped_text_get_ellipsis_glyphs`). As for GDExtensions, compat methods were registered. So it seems fine to break compatibility here.

cc @godotengine/gui-nodes

tutorials/migrating/upgrading_to_godot_4.5.rst
```diff
@@ -0,0 +1,258 @@
...
+========================================================================================================================  ===================  ====================  ====================  ============
+
+Editor plugins
+~~~~~~~~~~~~~~
+
+========================================================================================================================  ===================  ====================  ====================  ============
+Change                                                                                                                    GDScript Compatible  C# Binary Compatible  C# Source Compatible  Introduced
+========================================================================================================================  ===================  ====================  ====================  ============
+**EditorExportPlatformExtension**
+Method ``_get_option_icon`` changes return type from ``ImageTexture`` to ``Texture2D``                                    |✔️|                 |❌|                  |❌|                  `GH-108825`_
```
raulsntos, line 185: Unfortunately, we can't avoid breaking compat in C# if the method changed the return type. However, this class was added recently (in 4.4), and it's probably not implemented often by C# users (although we can't know for sure). So maybe breaking compatibility here is fine.

cc @godotengine/docks @godotengine/_platforms

tutorials/migrating/upgrading_to_godot_4.5.rst
bruvzg: Implementing text server with scripts doesn't make much sense. The only practical use case might be adding support for system font/text rendering API (e.g, macOS CoreText or Windows DirectWrite).

tutorials/migrating/upgrading_to_godot_4.5.rst
```diff
@@ -0,0 +1,258 @@
...
+
+In 4.5 some behavior changes have been introduced, which might require you to adjust your project.
+
+Core
+~~~~
+
+.. note::
+
+    In C#, ``StringExtensions.PathJoin`` now avoids adding an extra path separator when the original string is empty,
+    and when the appended path starts with a path separator (`GH-105281`_).
```
AThousandShips:
```suggestion
    or when the appended path starts with a path separator (`GH-105281`_).
```
If these are two different cases, which based on the PR seems to be the case, both are correct but this is less confusing

tutorials/migrating/upgrading_to_godot_4.5.rst
```diff
@@ -0,0 +1,258 @@
...
+Method ``texture_create_from_extension`` adds a new ``mipmaps`` optional parameter                                        |✔️|                 |✔️ with compat|      |✔️ with compat|      `GH-105570`_
+**RenderingServer**
+Method ``instance_reset_physics_interpolation`` removed                                                                   |❌|                 |✔️ with compat|      |✔️ with compat|      `GH-104269`_
+Method ``instance_set_interpolated`` removed                                                                              |❌|                 |✔️ with compat|      |✔️ with compat|      `GH-104269`_
+========================================================================================================================  ===================  ====================  ====================  ============
+
+.. note::
+
+    In C#, the enum ``RenderingDevice.Features`` breaks compatibility because of the way the bindings generator
+    detects the enum prefix. New members where added in `GH-103941`_ to the enum that caused the enum member
```
AThousandShips:
```suggestion
    detects the enum prefix. New members where added to the enum in `GH-103941`_ that caused the enum member
```

tutorials/migrating/upgrading_to_godot_4.5.rst
fire: @aaronfranke Has changed these. Would like your opinion.

fire: I think we silently upgraded our apis from 32bit to 64bit due to various errors.

aaronfranke: Compatibility methods are registered for these. I know that fixes GDExtension, but I'm not sure about C#? Anyway...

> I don't know if there are many C# users that were using them and would be affected

No, almost nobody was using these APIs. Before PR https://github.com/godotengine/godot/pull/108853 is merged for Godot 4.6, the functions to read/write glTF accessor and buffer data are internal to GLTFDocument, so in the current master of Godot, most of these exposed values aren't useful to user scripts unless they do a lot of manual work.

adamscott:
@raulsntos You should also mention in this document that Android now requires .NET 9

mihe:
It might be good to mention something about godotengine/godot#105746 under "Behavior changes".

People who are already using Jolt in 4.4 and are migrating to 4.5 will start to have overlaps reported between `Area3D` and static bodies that they didn't have before, assuming they didn't already enable the `physics/jolt_physics_3d/simulation/areas_detect_static_bodies` project setting. If the new behavior isn't desirable they'll need to adjust their collision layers/masks to emulate the old behavior.

Maybe something like:

```rst
Physics
~~~~~~~

.. note::

	When the 3D physics engine is set to Jolt Physics, you will now always have overlaps between ``Area3D`` and static
	bodies reported by default, as the ```physics/jolt_physics_3d/simulation/areas_detect_static_bodies``` project setting
	has been removed (`GH-105746`_). If you still want such overlaps to be ignored, you will need change the collision mask
	or layer of either the ``Area3D`` or the static body instead.
```

mhilbrunner (approved):

mhilbrunner:
As we're running out of time, I'm going to merge this now after giving it a read. If more crops up, we can add to it in further PRs and cherrypick those, as I'm going to branch off `4.5` in a sec. Thanks for working on it!

PR #11241, opened 2025-08-26
Add export_file warning note about uids

Lexyth:
The warning note was added to 4.5 in master, but not 4.4, even though the behavior is present there, too. Stable should be updated too, but that's done "manually" and I'm not sure what that entails.


Calinou:

For reference, this line of documentation was added to https://github.com/godotengine/godot/pull/105414, which can't be cherry-picked to `4.4` in its entirety as it's a new feature. So we still need a dedicated documentation PR on godotengine/godot targeting the `4.4` branch.

Lexyth:
Oh, alright. I just read that the documentation lives in a separate repository and it can't be modified in the PR in godot, along a link here, so I thought it would be modified here directly. 

Just thought I'd add the note quickly since it wasn't in the docs for 4.4, which lead me to dig deeper on the behavior of export_file and to the related issues in the first place.

PR #11246, opened 2025-08-30
Update instructions for resolution scaling in Multiple resolutions

Calinou:
The previously described approach was tailored for Godot 3.x, back when built-in 3D resolution scaling support didn't exist. Now that resolution scaling works in all rendering methods, it's better to use it instead.


mhilbrunner (approved):

skyace65:
Thanks!

PR #11269, opened 2025-09-09
Fix VSYNC_DISABLED description

hokuboros:
When vertical synchronization is disabled, assigning a value to Engine.max_fps does limit the framerate.


AThousandShips:

PR #11273, opened 2025-09-11
Document baking lightmaps on Android in Using Lightmap global illumination

Calinou:
This is supported since 4.5 thanks to https://github.com/godotengine/godot/pull/107635.


skyace65:
Thanks!

PR #11279, opened 2025-09-12
Link to GDExtension articles from the Godot introduction.

Ivorforce:
Ran across this. I think it's worth it linking to these technologies here, especially since it's intended for new users who are probably not familiar with all this.

I changes the wording somewhat too. C++ is an officially supported binding through godot-cpp, but C is not (we merely have a tutorial for it, for GDExtension binding authors, which is not intended for general use). Other languages are likely more interesting to many people, so I also link the article that contains links to those (e.g. Rust or Swift).


mhilbrunner:
Thanks!

PR #11289, opened 2025-09-13
read classes from file when lexing

Qubus0:
as requested by @mhilbrunner in https://github.com/godotengine/godot-docs/pull/11217#discussion_r2345757072 we can read the class reference index and grab all the file names, making the lexer exactly as up-to-date as the class reference.

- this does read pretty much all class files once because the class names in the index are not capitalised, which isn't quite optimal
- no web requests at least

the full repo is available during pipeline run, right? so the files should be readable


mhilbrunner:
Thank you for implementing that suggestion!

PR #11266, opened 2025-09-09
Add logging documentation

Calinou:
- Docs counterpart of https://github.com/godotengine/godot-demo-projects/pull/1242. This docs PR should be merged before, so I can link it in the demo README.

This also covers how to create custom loggers.

- See https://github.com/godotengine/godot/pull/91006.

PR #11299, opened 2025-09-17
Mention importance of thread-safety in `Logger` example

mihe:
Follow-up to #11266.

Given that not accommodating for the potential thread-safety concerns in these methods can crash the entire engine in the worst case, even when the user doesn't spin up any threads of their own, we should probably mention it in our example implementation.

### DIFF
--- a/tutorials/scripting/logging.rst
+++ b/tutorials/scripting/logging.rst
@@ -225,6 +225,8 @@ Here is a minimal working example of a custom logger, with the script added as a
         func _log_message(message: String, error: bool) -> void:
             # Do something with `message`.
             # `error` is `true` for messages printed to the standard error stream (stderr) with `print_error()`.
+            # Note that this method will be called from threads other than the main thread, possibly at the same
+            # time, so you will need to have some kind of thread-safety as part of it, like a Mutex.
             pass
 
         func _log_error(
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):

tutorials/scripting/logging.rst
```diff
@@ -225,6 +225,8 @@ Here is a minimal working example of a custom logger, with the script added as a
         func _log_message(message: String, error: bool) -> void:
             # Do something with `message`.
             # `error` is `true` for messages printed to the standard error stream (stderr) with `print_error()`.
+            # Note that this method will be called from threads other than the main thread, possibly at the same
+            # time, so you will need to have some kind of thread-safety as part of it, like a [Mutex].
```
Calinou: There's no syntax highlighting in standard comments, so we can't use a link syntax:

```suggestion
            # Note that this method will be called from threads other than the main thread, possibly at the same
            # time, so you will need to have some kind of thread-safety as part of it, like a Mutex.
```

tutorials/scripting/logging.rst
```diff
@@ -239,6 +241,8 @@ Here is a minimal working example of a custom logger, with the script added as a
         ) -> void:
             # Do something with the error. The error text is in `rationale`.
             # See the Logger class reference for details on other parameters.
+            # Note that this method will be called from threads other than the main thread, possibly at the same
+            # time, so you will need to have some kind of thread-safety as part of it, like a [Mutex].
```
Calinou:
```suggestion
            # Note that this method will be called from threads other than the main thread, possibly at the same
            # time, so you will need to have some kind of thread-safety as part of it, like a Mutex.
```

tutorials/scripting/logging.rst
mihe: Oh! That was not meant to be like that, my bad.

I'm seeing other places use backticks for class names in comments like this. Would it make sense to use those instead?

mhilbrunner:
Thank you!

mhilbrunner:

PR #11304, opened 2025-09-17
Add JavaScript-based redirects to supplement Read the Docs redirects

Calinou:
Read the Docs has an upper limit on the number of redirects that canapply. This limit is exceeded by the current `redirects.csv`,which means many redirects don't apply correctly.

While this JavaScript-based solution is not as seamless from the user's perspective, it has no limit on how many redirects can be defined.

To test this PR, build the website locally and run <code><a href="https://github.com/TheWaWaR/simple-http-server">simple-http-server</a> --try-file 404.html</code> in `_build/html`. It's important to instruct the local web server to use `404.html` as a custom 404 page, or you won't be able to test this.

Note that `redirects.csv` isn't modified by this PR yet, so this won't perform all redirects from the previous `contributing` section to `contributing.godotengine.org`.

We'll also probably want to change the redirects CSV URL to match the current branch in the long run. I'm not sure how we should do this in a way that keeps working with local web servers (which may not have `/latest`, `/stable` or `/4.5` in the URL). Maybe we should just modify `custom.js` for each branch?


Ivorforce:
> Note that `redirects.csv` isn't modified by this PR yet, so this won't perform all redirects from the previous `contributing` section to `contributing.godotengine.org`.

This is addressed in https://github.com/godotengine/godot-docs/pull/11315.

> We'll also probably want to change the redirects CSV URL to match the current branch in the long run. I'm not sure how we should do this in a way that keeps working with local web servers (which may not have `/latest`, `/stable` or `/4.5` in the URL). Maybe we should just modify `custom.js` for each branch?

I think the easiest solution might be to place `redirects.csv` in the public folder, and simply grab it by using the base url + `/redirects.csv`.

Calinou:
> I think the easiest solution might be to place redirects.csv in the public folder, and simply grab it by using the base url + `/redirects.csv`.

Wouldn't that break RTD redirects though? We still want them to work when possible to reduce the reliance on JavaScript, as it's less optimal.

**Edit:** Nevermind, this is handled by https://github.com/godotengine/godot-docs/blob/master/_tools/redirects/create_redirects.py rather than CSV file itself.

**Edit 2:** Done. The redirects CSV is now in `_static` and is loaded directly, so it also works using the local `redirects.csv` during local testing. I've also rebased the PR on top of https://github.com/godotengine/godot-docs/pull/11315.

Ivorforce:
Awesome! Does that leave the PR in a reviewable / potentially mergable state?

_static/js/custom.js
```diff
@@ -298,6 +297,37 @@ const registerGiscus = function () {
 };
 
 $(document).ready(() => {
+  const httpResponseStatus = window.performance.getEntries()[0].responseStatus;
+  if (httpResponseStatus === 404) {
+    // Check for redirects if on a currently invalid page.
+    // This is done in JavaScript, as Read the Docs' own redirect functionality
+    // is unreliable with large amounts of redirects defined.
+    fetch("https://raw.githubusercontent.com/godotengine/godot-docs/refs/heads/master/_tools/redirects/redirects.csv")
```
Ivorforce: Shouldn't this fetch from the base of the website now?

Calinou: Should be fixed now.

Note that during local development, the URL will be different (`/_static/redirects.csv` instead of `/en/latest/_static/redirects.csv`), but I don't know a good way to catch this automatically. I've just hardcoded it to `/en/latest/_static/redirects.csv` for now, so it works in production.

Calinou:
I've added support for redirecting to absolute URLs. Note that the existing `redirects.csv` has some typos which I didn't notice while reviewing https://github.com/godotengine/godot-docs/pull/11315 (search for instances of `//architecture`). Should I fix them in this PR, or should you open a new PR?

Ivorforce:
> I've added support for redirecting to absolute URLs. Note that the existing `redirects.csv` has some typos which I didn't notice while reviewing #11315 (search for instances of `//architecture`). Should I fix them in this PR, or should you open a new PR?

Ah, dang, sorry. I think it's fine to add them here.

Calinou:
> Ah, dang, sorry. I think it's fine to add them here.

Done :slightly_smiling_face:

_static/js/custom.js
```diff
@@ -298,6 +297,40 @@ const registerGiscus = function () {
 };
 
 $(document).ready(() => {
+  const httpResponseStatus = window.performance.getEntries()[0].responseStatus;
+  if (httpResponseStatus === 404) {
+    // Check for redirects if on a currently invalid page.
+    // This is done in JavaScript, as Read the Docs' own redirect functionality
+    // is unreliable with large amounts of redirects defined.
```
mhilbrunner:
```suggestion
    // This is done in JavaScript, as we exceed Read the Docs' limit for the amount of redirects configurable.
```

mhilbrunner (changes requested):
Looks good to me. Would be good to add a comment to `create_redirects.py` that it is now obsolete, pointing to `custom.js` instead.

Will test this a bit.

_static/js/custom.js
```diff
@@ -298,6 +297,40 @@ const registerGiscus = function () {
...
+    fetch("/_static/redirects.csv")
```
Ivorforce: Does this work with the full sub URL too, e.g. `/stable/en`? Normal fetches usually use the base URL, so this would become `docs.godotengine.org/_static/redirects.csv`, which I'm not sure if it would resolve.

_static/js/custom.js
```diff
@@ -298,6 +297,40 @@ const registerGiscus = function () {
...
+      .then(response => response.text())
+      .then(csvText => {
+        const lines = csvText.trim().split('\n');
+        const redirects = {};
```
Ivorforce: I'm not sure we need to make a hash table for this. In the loop, it should be possible to simply test that `window.location` ends with `from`, and if so, replace with `to`. It should technically be faster because less data is churned. Though I'm not sure if it would make a practical difference.

_static/js/custom.js
Calinou: Fixed (I had this in my local copy, but forgot to copy it out of `_build/html/`).

_static/js/custom.js
Calinou: Done (I've moved the logic within the `for` loop to remove the object).

mhilbrunner (approved):

mhilbrunner:
Alright, lets try this. Thanks both of you! We'll see how this goes.

mhilbrunner:

PR #11306, opened 2025-09-18
Add reference to the new "Logger" class

AdrienUfferte:
If you already use a global script or custom class named ``Logger``, it will conflict with the built-in one and produce errors such as "Static function not found". You must rename your global (e.g. from ``Logger`` to ``Log``) and update references.


AThousandShips:
We don't track new classes like this, also if people name their own classes like this those will throw an error because it conflicts with a built-in class

It's also misleading to declare this change a compatibility breakage, it does not break compatibility with any of the languages inherently

This would also require us to add any new class *or* method for any class you might extend, so this would be far too noisy

AdrienUfferte:
> We don't track new classes like this, also if people name their own classes like this those will throw an error because it conflicts with a built-in class
> 
> It's also misleading to declare this change a compatibility breakage, it does not break compatibility with any of the languages inherently
> 
> This would also require us to add any new class _or_ method for any class you might extend, so this would be far too noisy

Ok thank you for your reply.
So, shoul I close this PR ?

AThousandShips:
It's not up to me to completely reject this, so if you want a second opinion you can leave this open

AdrienUfferte:
> It's not up to me to completely reject this, so if you want a second opinion you can leave this open

Thank you, your answer made perfect sense, I'm closing it.

PR #11311, opened 2025-09-19
Add 4.5 upgrade Navigation notes

smix8:
Adds 4.5 upgrade Navigation notes.

### DIFF
--- a/tutorials/migrating/upgrading_to_godot_4.5.rst
+++ b/tutorials/migrating/upgrading_to_godot_4.5.rst
@@ -231,6 +231,25 @@ Core
     the shortest arc between the two input vectors. Previously, it would return incorrect values for certain inputs
     (`GH-107618`_).
 
+Navigation
+~~~~~~~~~~
+
+.. note::
+
+    By default, the regions in a NavigationServer map now update asynchronously using threads to improve performance.
+    This can cause additional delay in the update due to thread synchronisation.
+    The asynchronous region update can be toggled with the ``navigation/world/region_use_async_iterations`` project setting.
+
+.. note::
+    The merging of navmeshes in the NavigationServer has changed processing order. Regions now merge and cache
+    internal navmeshes first, then the remaining free edges are merged by the navigation map.
+    If a project had navigation map synchronisation errors before, it might now have shifted
+    affected edges, making already existing errors in a layout more noticeable in the pathfinding.
+    The ``navigation/2d_or_3d/merge_rasterizer_cell_scale`` project setting can be set to a lower value
+    to increase the detail of the rasterization grid (with `0.01` being the smallest cell size possible).
+    If edge merge errors still persist with the lowest possible rasterization scale value,
+    the error may be caused by overlap: two navmeshes are stacked on top of each other, causing geometry conflict.
+
 Physics
 ~~~~~~~
 


tutorials/migrating/upgrading_to_godot_4.5.rst
```diff
@@ -231,6 +231,25 @@ Core
...
     (`GH-107618`_).
 
+Navigation
+~~~~~~~~~~
+
+.. note::
+
+    NavigationServer regions now by default update async using threads to improve performance.
+    This can cause additional delay in the update due to thread synchronisation.
+    The async region update can be toggled with the ``navigation/world/region_use_async_iterations`` project setting.
```
Mickeon:
```suggestion
    By default, the regions in a NavigationServer regions now update asynchronously using threads to improve performance.
    This can cause additional delay in the update due to thread synchronisation.
    The asynchronous region update can be toggled with the ``navigation/world/region_use_async_iterations`` project setting.
```
Is it one, or more than one update, too?

tutorials/migrating/upgrading_to_godot_4.5.rst
```diff
@@ -231,6 +231,25 @@ Core
...
+
+.. note::
+    NavigationServer navmesh merging has changed processing order as regions now merge and
+    cache internal navmesh first before remaining free edges are merged by the navigation map.
+    If a project had navigation map synchronisation errors before it might now have shifted
+    affected edges making already existing errors in a layout more noticeable in the pathfinding.
```
Mickeon:
```suggestion
    The merging of navmeshes in the NavigationServer has changed processing order. Regions now merge and cache
    internal navmeshes first, then the remaining free edges are merged by the navigation map.
    If a project had navigation map synchronisation errors before, it might now have shifted
    affected edges, making already existing errors in a layout more noticeable in the pathfinding.
```

tutorials/migrating/upgrading_to_godot_4.5.rst
```diff
@@ -231,6 +231,25 @@ Core
...
+    The ``navigation/2d_or_3d/merge_rasterizer_cell_scale`` project setting can be used at a lower value
+    to increase the detail of the rasterization grid (with 0.01 being the smallest cell size possible).
+    If a navigation mesh edge merge error still persists with the lowest possible rasterization scale value
+    the error is caused by overlap, e.g. two navmeshes stacked on top of each other causing geometry conflict.
```
Mickeon:
```suggestion
    The ``navigation/2d_or_3d/merge_rasterizer_cell_scale`` project setting can be set to a lower value
    to increase the detail of the rasterization grid (with `0.01` being the smallest cell size possible).
    If edge merge errors still persist with the lowest possible rasterization scale value,
    the error may be caused by overlap: two navmeshes are stacked on top of each other, causing geometry conflict.
```
Or similar.

Mickeon (commented):
Thank you for your service

tutorials/migrating/upgrading_to_godot_4.5.rst
smix8: What do you mean "more than one"? What is "one"? The update? The sync? The iteration? The Wookie behind the tree laughing at both of us? I have so many possible "one"s here :D

There will be one sync point to join each thread task but how long that task takes, and how many possible "server syncs" and main loops happen on the main thread in the meantime, that is unknown.

It depends entirely on how heavy the task is and if a thread is even free to do the task immediately. It might be waiting in the WorkerThreadPool task queue for a while when no thread is free.

After the async region task is done there is also the async map task. So just because a region task has finished does not mean that the map is already updated.

The update state of any object can be polled by querying its current iteration id.
The iteration id is incremented by 1 everytime the object finished an update.

So after pushing an update to a navmesh, by first waiting for the region iteration to increase, followed by waiting for the next map iteration increase, you can make an educated guess when all the threaded updates went through. Although the moment users start to involve their own threaded queries that may delay map iteration builds the timing will be off again.

Nothing is simple in thread-land when you actually run things async-async instead of start-thread-and-immediately-wait_to_finish-yolo-block-the main-thread like a physics-server maniac.

mhilbrunner (approved):

mhilbrunner:
Thanks!

mhilbrunner:

PR #11314, opened 2025-09-21
Fix typo: "attenutation" --> "attenuation"

CypherPoet:


mhilbrunner:
Thanks for contributing! The class reference files (everything under `classes/`) are generated from the [main repository](https://github.com/godotengine/godot/tree/master/doc/classes) ([see here](https://github.com/godotengine/godot/tree/master/doc/classes), so the source files need to be edited there, and PRs need to be made against that repository, otherwise any changes are going to be lost during the next update.

PR #11317, opened 2025-09-22
Edit "Core types" descriptions

Ivorforce:
- `Pair`: Mention `KeyValue`. `KeyValue` could get its own entry, but there is nothing much to say, and `KeyValue` is unlikely to be used outside of containers.
- `RBMap`: Discourage its use. We might actually phase it out fully in the future.
- `Span`: Make it clear it's low level API. People should be careful when using this.
- `String`: Clarify UTF-32 motivation. UTF-32 isn't actually more performant in all respects (it's a RAM hog), but it _is_ simpler to work with.

### DIFF
--- a/engine_details/architecture/core_types.rst
+++ b/engine_details/architecture/core_types.rst
@@ -142,7 +142,7 @@ scripting API.
 | Godot datatype        | Closest C++ STL datatype | Comment                                                                               |
 +=======================+==========================+=======================================================================================+
 | |string| 📜           | ``std::string``          | **Use this as the "default" string type.** ``String`` uses UTF-32 encoding            |
-|                       |                          | to improve performance thanks to its fixed character size.                            |
+|                       |                          | to simplify processing thanks to its fixed character size.                            |
 +-----------------------+--------------------------+---------------------------------------------------------------------------------------+
 | |vector|              | ``std::vector``          | **Use this as the "default" vector type.** Uses copy-on-write (COW) semantics.        |
 |                       |                          | This means it's generally slower but can be copied around almost for free.            |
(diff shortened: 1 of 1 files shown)


engine_details/architecture/core_types.rst
```diff
@@ -180,8 +180,10 @@ scripting API.
 |                       |                          | no heap allocations.                                                                  |
 +-----------------------+--------------------------+---------------------------------------------------------------------------------------+
 | |span|                | ``std::span``            | Represents read-only access to a contiguous array without needing to copy any data.   |
-|                       |                          | See `pull request description <https://github.com/godotengine/godot/pull/100293>`__   |
-|                       |                          | for details.                                                                          |
+|                       |                          | Note that ``Span`` is designed to be a high performance API: It does not perform      |
```
Calinou:
```suggestion
|                       |                          | Note that ``Span`` is designed to be a high performance API: it does not perform      |
```

Calinou (approved):
Looks good to me.

engine_details/architecture/core_types.rst
```diff
@@ -180,8 +180,10 @@ scripting API.
 |                       |                          | no heap allocations.                                                                  |
 +-----------------------+--------------------------+---------------------------------------------------------------------------------------+
 | |span|                | ``std::span``            | Represents read-only access to a contiguous array without needing to copy any data.   |
-|                       |                          | See `pull request description <https://github.com/godotengine/godot/pull/100293>`__   |
-|                       |                          | for details.                                                                          |
+|                       |                          | Note that ``Span`` is designed to be a high performance API: it does not perform      |
+|                       |                          | parameter correctness checks in same way you might be used to with other Godot        |
```
AThousandShips:
```suggestion
|                       |                          | parameter correctness checks in the same way you might be used to with other Godot    |
```

engine_details/architecture/core_types.rst
AThousandShips: You reverted this change, was that intentional?

Ivorforce: When I looked at this, the discussion was already resolved, so I assumed it wasn't relevant after all. Doesn't the sentence after a semicolon begin uppercase if it's a full sentence?

mhilbrunner:
Thank you!

mhilbrunner:

PR #11333, opened 2025-10-01
Drop the experimental label for the Jolt Physics integration

mihe:
Related to godotengine/godot-proposals#12289.

See more detailed motivation in godotengine/godot#111115.

There might be some argument for keeping the "not finished" and "lacks feature parity" parts of this section, since they're technically still true, but as talked about in godotengine/godot#111115 I don't think these have proven to be significant enough of an issue to warrant this big and scary warning anymore, especially now that Jolt is the default 3D physics engine for newly created projects.

### DIFF
--- a/tutorials/physics/using_jolt_physics.rst
+++ b/tutorials/physics/using_jolt_physics.rst
@@ -11,11 +11,6 @@ physics engine in 4.4. Jolt is developed by Jorrit Rouwe with a focus on games a
 VR applications. Previously it was available as an extension but is now built into
 Godot.
 
-It is important to note that the built-in Jolt Physics module is considered
-**not finished**, **experimental**, and **lacks feature parity** with both
-Godot Physics and the Godot Jolt extension. Behavior may change as it is developed
-further. Please keep that in mind when choosing what to use for your project.
-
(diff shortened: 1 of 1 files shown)


Calinou (approved):
Looks good to me.

Note that full feature parity isn't achieved yet (e.g. Generic6DOFJoint3D is still missing), but the rest is basically there at this point.

mihe:
> Generic6DOFJoint3D is still missing

`Generic6DOFJoint3D` is implemented, with support for limits, motors and (unlike Godot Physics) springs, which is true for the other joints as well, but they're missing the properties for configuring soft limits, i.e. limit `softness`, `restitution` and `damping`. Those will likely never be implemented in their current form, since Jolt does that in a different way.

The exact properties are laid out in the documentation that this PR changes.

mhilbrunner:
Marking this a draft for now, as we currently try to keep `master` close to backwards compatible with `4.5`/`stable` a while longer, until we start to get more 4.6-only changes. That said, thanks for already being on top of that!

skyace65:
Merging this as we've started merging 4.6 only changes.

skyace65:
Thanks!

PR #11334, opened 2025-10-02
Fix a c# method name in class_control.rst

TomWerner:
As is, the example code fails to compile with `'Theme' does not contain a definition for 'SetStyleBox' and no accessible extension method 'SetStyleBox' accepting a first argument of type 'Theme' could be found`


mhilbrunner:
Thanks for contributing! The class reference files (everything under `classes/`) are generated from the [main repository](https://github.com/godotengine/godot/tree/master/doc/classes) ([see here](https://github.com/godotengine/godot/tree/master/doc/classes), so the source files need to be edited there, and PRs need to be made against that repository, otherwise any changes are going to be lost during the next update.

For more information, see the contributing docs for [Contributing to the class reference](https://contributing.godotengine.org/en/latest/documentation/class_reference.html).

PR #11342, opened 2025-10-04
Fix redirects in Safari

Ivorforce:
- Follow-up of #11304 

Redirects currently don't work in Safari. The reason is that `window.performance.getEntries()[0].responseStatus` is unsupported.

Instead of checking in js whether we're 404-ing, I moved the script into the 404 page. This has the additional benefit that it is not loaded / run on any other pages except those that 404.


mhilbrunner:
cc @Calinou

Calinou (approved):
Looks good to me.

Note that https://github.com/godotengine/godot-docs/pull/11349 should be rebased following this merge.

mhilbrunner:
Thank you! Merged.

mhilbrunner:

PR #11344, opened 2025-10-05
Move godot-cpp custom bindings information from "Getting Started" to build system information

Ivorforce:
On Discord, we regularly get people asking about or being directed to the "Getting Started" section of godot-cpp. This is great, because that means it's really useful!

However, we also regularly get people building with custom bindings, although they do not intend to target the latest `master` or a custom branch. I think they just do it because nobody reads the text carefully enough to realize that this step is optional.

Since most people do not need to build custom bindings, and will target a stable branch instead, I don't think this information should be part of the "Getting Started" doc. I therefore propose to move it to "Build system" in a dedicated subsection instead.

### DIFF
--- a/tutorials/scripting/cpp/build_system/scons.rst
+++ b/tutorials/scripting/cpp/build_system/scons.rst
@@ -73,3 +73,37 @@ There are two popular ways by which cross platform builds can be achieved:
 `godot-cpp-template <https://github.com/godotengine/godot-cpp-template>`__ contains an
 `example setup <https://github.com/godotengine/godot-cpp-template/tree/main/.github/workflows>`__
 for a GitHub based CI workflow.
+
+Using custom bindings
+---------------------
+
+Every branch of godot-cpp comes with bindings appropriate for the respective Godot version
+(e.g. the ``4.3`` branch comes with bindings compatible with Godot version ``4.3`` and
+later).
+
+However, you may want to use custom bindings, for example:
+
+* If you want to use the latest bindings from Godot ``master``.
+* If you want to use bindings exposed by custom modules.
+
+To use custom bindings, you first have to generate them from the appropriate Godot
+executable:
+
+.. code-block:: shell
+
+    godot --dump-extension-api
+
+The resulting ``extension_api.json`` file will be created in the executable's
+directory. To use these custom bindings, you can add ``custom_api_file`` to
+your build command:
+
+.. code-block:: shell
+
+    scons platform=<platform> custom_api_file=<PATH_TO_FILE>
+
+Alternatively, you can add them as the default bindings to your project by adding
+the following line to your SConstruct file:
+
+.. code-block:: python
+
+    localEnv["custom_api_file"] = "extension_api.json"
--- a/tutorials/scripting/cpp/gdextension_cpp_example.rst
+++ b/tutorials/scripting/cpp/gdextension_cpp_example.rst
@@ -97,44 +97,6 @@ following commands:
 
 This will initialize the repository in your project folder.
 
-Building the C++ bindings
--------------------------
-
-Now that we've downloaded our prerequisites, it is time to build the C++
-bindings.
-
-The repository contains a copy of the metadata for the current Godot release,
-but if you need to build these bindings for a newer version of Godot, call
-the Godot executable:
-
-.. code-block:: none
-
-    godot --dump-extension-api
-
-The resulting ``extension_api.json`` file will be created in the executable's
-directory. Copy it to the project folder and add ``custom_api_file=<PATH_TO_FILE>``
-to the scons command below.
-
-To generate and compile the bindings, use this command (replacing ``<platform>``
-with ``windows``, ``linux`` or ``macos`` depending on your OS):
-
-The build process automatically detects the number of CPU threads to use for
-parallel builds. To specify a number of CPU threads to use, add ``-jN`` at the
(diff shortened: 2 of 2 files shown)


tutorials/scripting/cpp/build_system/scons.rst
```diff
@@ -73,3 +73,37 @@ There are two popular ways by which cross platform builds can be achieved:
...
+---------------------
+
+Every branch of godot-cpp comes with bindings appropriate for the respective Godot version
+(e.g. the ``4.3`` branch comes with bindings compatible with Godot version ``4.3`` and
+later).
+
+However, you may want to use custom bindings, for example:
+
+* If you want to use the latest bindings from Godot ``master``.
+* If you want to use bindings exposed by custom modules.
```
dsnopek: We should also include something allow the lines of "* If you build Godot yourself with different options than the official builds" because you should do this if you built with options like `scons precision=double` or `scons disable_3d=yes` or `scons deprecated=no` as well

tutorials/scripting/cpp/build_system/scons.rst
```diff
@@ -73,3 +73,37 @@ There are two popular ways by which cross platform builds can be achieved:
 `godot-cpp-template <https://github.com/godotengine/godot-cpp-template>`__ contains an
 `example setup <https://github.com/godotengine/godot-cpp-template/tree/main/.github/workflows>`__
 for a GitHub based CI workflow.
+
+Using custom bindings
```
dsnopek: I'm not sure how I feel about calling this "custom bindings". To me, a "binding" is like godot-cpp or godot-rust -- the whole library that can interact with Godot. This is more like using a custom API file to build your bindings with

Maybe this should be like:

```suggestion
Using a custom API file
```

... or similar?

tutorials/scripting/cpp/build_system/scons.rst
```diff
@@ -73,3 +73,37 @@ There are two popular ways by which cross platform builds can be achieved:
...
+---------------------
+
+Every branch of godot-cpp comes with bindings appropriate for the respective Godot version
+(e.g. the ``4.3`` branch comes with bindings compatible with Godot version ``4.3`` and
```
dsnopek:
```suggestion
Every branch of godot-cpp comes with an API file (``extension_api.json``) appropriate for the respective Godot version
(e.g. the ``4.3`` branch comes with the API file compatible with Godot version ``4.3`` and
```

tutorials/scripting/cpp/build_system/scons.rst
dsnopek:
```suggestion
However, you may want to use a custom ``extension_api.json``, for example:

* If you want to use the latest APIs from Godot ``master``.
* If you want to use APIs exposed by custom modules.
```

tutorials/scripting/cpp/build_system/scons.rst
```diff
@@ -73,3 +73,37 @@ There are two popular ways by which cross platform builds can be achieved:
...
+later).
+
+However, you may want to use custom bindings, for example:
+
+* If you want to use the latest bindings from Godot ``master``.
+* If you want to use bindings exposed by custom modules.
+
+To use custom bindings, you first have to generate them from the appropriate Godot
```
dsnopek:
```suggestion
To use a custom API file, you first have to generate it from the appropriate Godot
```

tutorials/scripting/cpp/build_system/scons.rst
```diff
@@ -73,3 +73,37 @@ There are two popular ways by which cross platform builds can be achieved:
...
+executable:
+
+.. code-block:: shell
+
+    godot --dump-extension-api
+
+The resulting ``extension_api.json`` file will be created in the executable's
+directory. To use these custom bindings, you can add ``custom_api_file`` to
```
dsnopek:
```suggestion
directory. To use this, you can add ``custom_api_file`` to
```

tutorials/scripting/cpp/build_system/scons.rst
```diff
@@ -73,3 +73,37 @@ There are two popular ways by which cross platform builds can be achieved:
...
+your build command:
+
+.. code-block:: shell
+
+    scons platform=<platform> custom_api_file=<PATH_TO_FILE>
+
+Alternatively, you can add them as the default bindings to your project by adding
```
dsnopek:
```suggestion
Alternatively, you can add it as the default API file to your project by adding
```

tutorials/scripting/cpp/build_system/scons.rst
Ivorforce: Makes sense to me. I reworded the text section below accordingly as well.
Edit: Ah, those are what your remaining comments were about anyway :D

tutorials/scripting/cpp/build_system/scons.rst
Ivorforce: Good point! I included the `precision` and `disable_3d` in there as examples.

mhilbrunner (approved):

mhilbrunner:
LGTM. Should we cherrypick this for any branches?

Ivorforce:
> Should we cherrypick this for any branches?

It applies to all branches technically. Can't cherry-pick further than 4.5 though because of previous changes.

mhilbrunner:
Thanks! Merged.

Issue #11129, opened 2025-07-21
Switching languages shows page not found


berarma:
English uses different version tags in the URL, thus a page in english in `/4.4/` would be located in `/4.x/' in Spanish or German. Same with tags `latest` and `stable`, they only exist in English. This makes switching languages a bit hard and confusing, because instead of redirecting to the correct URL or the home page, a 404 not found error page is shown.

M-Banasiak:
Up. This is quite annoying issue. Documentation automatically redirects to local language version (usually incomplete and badly translated) and it is hard to switch to english.

Calinou:
This should be addressed by https://github.com/godotengine/godot-docs/pull/11349.

PR #11349, opened 2025-10-06
Fix RTD language switcher redirecting to nonexistent `/en/4.x/` branch

Calinou:
If we were redirected to `/en/4.x/` as a result of the RTD language selector, redirect to `/en/stable/` instead, as English does not have a `4.x` branch.

For maintenance reasons, non-English languages have a single `4.x` branch that is the counterpart of `4.3`, `4.4`, `4.5` and so on.).

image  

_Bugsquad edit:_ Followup to https://github.com/godotengine/godot-docs/pull/11304

- This closes https://github.com/godotengine/godot-docs/issues/11129.

### DIFF
--- a/404.rst
+++ b/404.rst
@@ -24,6 +24,15 @@ Page not found
     </p>
 
     <script>
+    // If we were redirected to `/en/4.x/` as a result of the RTD language selector,
+    // redirect to `/en/stable/` instead, as English does not have a `4.x` branch.
+    // (For maintenance reasons, non-English languages have a single `4.x` branch
+    // that is the counterpart of `4.3`, `4.4`, `4.5`, and so on.)
+    if (window.location.pathname.includes('/en/4.x/')) {
+      const newUrl = window.location.pathname.replace('/en/4.x/', '/en/stable/');
+      window.location.replace(newUrl);
+    }
+
     // Check for redirects if on a currently invalid page.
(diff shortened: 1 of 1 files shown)


rburing (approved):
Makes sense to me and the implementation looks fine.

It would be nice to match and replace only the start of the path, but realistically it doesn't seem like the current approach will cause any issue.

Mickeon (approved):
This bothered me for the longest time.

_static/js/custom.js
```diff
@@ -299,6 +299,15 @@ const registerGiscus = function () {
 $(document).ready(() => {
   const httpResponseStatus = window.performance.getEntries()[0].responseStatus;
   if (httpResponseStatus === 404) {
+    // If we were redirected to `/en/4.x/` as a result of the RTD language selector,
+    // redirect to `/en/stable/` instead, as English does not have a `4.x` branch.
+    // (For maintenance reasons, non-English languages have a single `4.x` branch
+    // that is the counterpart of `4.3`, `4.4`, `4.5` and so on.)
```
AThousandShips:
```suggestion
    // that is the counterpart of `4.3`, `4.4`, `4.5`, and so on.)
```

mhilbrunner:
@Calinou Needs a rebase now after https://github.com/godotengine/godot-docs/pull/11342

mhilbrunner:
Thanks!

mhilbrunner:

PR #11350, opened 2025-10-06
Edit per-instance texture array workaround

### LINKED ISSUE #11331: Per-instance uniforms texture workaround issue?
**Your Godot version:**
4.5

**Issue description:**
The given workaround does not seem to work? I may be doing something wrong, but I've been searching all around and I can't find anything else giving more details.

The workaround in question is:
> Per-instance uniforms do not support textures or arrays, only regular scalar and vector types. As a workaround, you can pass a texture array as a regular uniform, then pass the index of the texture to be drawn using a per-instance uniform.

The problem is when creating an ``uniform sampler2D textures[100]`` array, you can't actually select specific indexes via the per-instance uniform as you can only use constant expressions to select array indexes. Here the error from godot, when attepmting the workaround:

```
E 0:00:00:657   _display_error_with_code: CanvasShaderGLES3: Fragment shader compilation failed:
0:492(19): error: sampler arrays indexed with non-constant expressions are forbidden in GLSL 1.30 and later
0:496(19): error: sampler arrays indexed with non-constant expressions are forbidden in GLSL 1.30 and later
0:500(19): error: sampler arrays indexed with non-constant expressions are forbidden in GLSL 1.30 and later
0:504(19): error: sampler arrays indexed with non-constant expressions are forbidden in GLSL 1.30 and later

  <C++ Source>  drivers/gles3/shader_gles3.cpp:259 @ _display_error_with_code()
```

https://docs.godotengine.org/en/stable/tutorials/shaders/shader_reference/shading_language.html

JellyBoonz:
Fixes the workaround for per-instance uniforms with texture arrays.
   
The previous documentation suggested indexing could work directly, but GLSL requires sampler array indices to be compile-time constants. 

This PR corrects the documentation to explain the limitation and provides a working switch statement workaround with code example.
   
   Fixes #11331

### DIFF
--- a/tutorials/shaders/shader_reference/shading_language.rst
+++ b/tutorials/shaders/shader_reference/shading_language.rst
@@ -1104,10 +1104,28 @@ method on a node that inherits from :ref:`class_GeometryInstance3D`:
 
 When using per-instance uniforms, there are some restrictions you should be aware of:
 
-- **Per-instance uniforms do not support textures or arrays**, only regular scalar and
-  vector types. As a workaround, you can pass a texture array as a regular
-  uniform, then pass the index of the texture to be drawn using a per-instance
-  uniform.
+- **Per-instance uniforms do not support textures or arrays**, only regular scalar and vector types. 
+
+.. note::
+   Due to GLSL limitations, you cannot directly index a texture array using a per-instance uniform. Sampler arrays can only be indexed by compile-time constant expressions.
+   
+   As a workaround, pass a texture array as a regular uniform and the desired texture index as a per-instance uniform. Then use a switch statement to select the texture:
+   
+   .. code-block:: glsl
+   
+      uniform sampler2D texture_array[4];
+      instance uniform int texture_index;
+      
+      void fragment() {
+          vec4 color;
+          switch(texture_index) {
+              case 0: color = texture(texture_array[0], UV); break;
+              case 1: color = texture(texture_array[1], UV); break;
+              case 2: color = texture(texture_array[2], UV); break;
+              case 3: color = texture(texture_array[3], UV); break;
+          }
+          COLOR = color;
+      }
 - There is a practical maximum limit of 16 instance uniforms per shader.
 - If your mesh uses multiple materials, the parameters for the first mesh
   material found will "win" over the subsequent ones, unless they have the same


tutorials/shaders/shader_reference/shading_language.rst
```diff
@@ -1104,10 +1104,28 @@ method on a node that inherits from :ref:`class_GeometryInstance3D`:
...
+      void fragment() {
+          vec4 color;
+          switch(texture_index) {
+              case 0: color = texture(texture_array[0], UV); break;
+              case 1: color = texture(texture_array[1], UV); break;
+              case 2: color = texture(texture_array[2], UV); break;
+              case 3: color = texture(texture_array[3], UV); break;
+          }
+          COLOR = color;
+      }
```
Calinou, line 1142:
```suggestion
.. note::

    Due to GLSL limitations, you cannot directly index a texture array
    using a per-instance uniform. Sampler arrays can only be indexed by
    compile-time constant expressions.
   
    As a workaround, pass a texture array as a regular uniform and the
    desired texture index as a per-instance uniform. Then use a ``switch``
    statement to select the texture:
   
   .. code-block:: glsl
   
      uniform sampler2D texture_array[4];
      instance uniform int texture_index;
      
      void fragment() {
          vec4 color;
          switch (texture_index) {
              case 0:
                  color = texture(texture_array[0], UV);
                  break;
              case 1:
                  color = texture(texture_array[1], UV);
                  break;
              case 2:
                  color = texture(texture_array[2], UV);
                  break;
              case 3:
                  color = texture(texture_array[3], UV);
                  break;
          }
          
          COLOR = color;
      }
```

tutorials/shaders/shader_reference/shading_language.rst
```diff
@@ -1104,10 +1104,42 @@ method on a node that inherits from :ref:`class_GeometryInstance3D`:
...
+              case 2:
+                  color = texture(texture_array[2], UV);
+                  break;
+              case 3:
+                  color = texture(texture_array[3], UV);
+                  break;
+          }
+          
+          COLOR = color;
+      }
```
Calinou, line 1142:
```suggestion
.. note::

    Due to GLSL limitations, you cannot directly index a texture array
    using a per-instance uniform. Sampler arrays can only be indexed by
    compile-time constant expressions.
   
    As a workaround, pass a texture array as a regular uniform and the
    desired texture index as a per-instance uniform. Then use a ``switch``
    statement to select the texture:
   
   .. code-block:: glsl
   
      uniform sampler2D texture_array[4];
      instance uniform int texture_index;
      
      void fragment() {
          vec4 color;
          switch (texture_index) {
              case 0:
                  color = texture(texture_array[0], UV);
                  break;
              case 1:
                  color = texture(texture_array[1], UV);
                  break;
              case 2:
                  color = texture(texture_array[2], UV);
                  break;
              case 3:
                  color = texture(texture_array[3], UV);
                  break;
          }
          
          COLOR = color;
      }

```

Calinou (approved):

mhilbrunner (approved):

mhilbrunner:
Thank you! Merged.

akien-mga:
Cherry-picked for `4.5`.

FyiurAmron:
@JellyBoonz Hmm. Either I'm missing something, or
```
shader_type spatial;

uniform sampler2D texture_array[4];
instance uniform int texture_index;

void fragment() {
    ALBEDO = texture(texture_array[texture_index], UV).rgb;
}
```
(or any similar code) just works - at least it works for me? Supposedly, starting from GLSL 4.0 the above syntax is valid... and GLSL 4.0 is already 15 years old. See https://stackoverflow.com/a/45167941/17825391 etc. I don't know how to reproduce https://github.com/godotengine/godot-docs/issues/11331 , but currently I'm seeing docs that are incorrect by stating something that seems to not be universally true.

Also, Godot is using `#version 450` (4.5) internally I believe? As such, the whole SO thread doesn't really apply. I'm really stumped why the issue happened for @CabalCrow. I was able to compile the above both on Forward+ and Mobile.

Can @akien-mga and/or @mhilbrunner and/or @Calinou verify whether #11331 actually affects your shader compilation, i.e. whether my minimal shader above compiles for you?

Calinou:
That shader compiles without errors for me on 4.5.1 on Windows + NVIDIA, in all renderers. Which OS and graphics card model are you using?

> Also, Godot is using `#version 450` (4.5) internally I believe?

Yes on Forward+/Mobile, but not in Compatibility where GLSL ES 3.0 is used.

FyiurAmron:
@Calinou I've tested on two Windows 10 (build 19045) machines, in all the modes as well (from Direct3D 12 (Forward+) to OpenGL 3 (Compatibility)), both for spatials and for canvas_items, both on dedicated NVIDIA GeForce GTX 1660 Ti (NVIDIA; 32.0.15.7602) on my laptop and on Radeon RX 580 Series (Advanced Micro Devices, Inc.; 31.0.21923.1000) on my office box. Neither of those is new by any means. No need for switch shenanigans on any of those.

I've asked the original reporter for repro or more info. I have a hunch this problem is either actually nonexistent or happens only on some very strange combo of old hardware/system. If that would be the case, I'd propose amending the docs to either former wording or a more relaxed phrasing (i.e. adding "on some very old hardware" etc.) - IMO there's no need to teach people smelly workarounds that ain't needed in normal situations for decades :)

PR #11369, opened 2025-10-12
Remove mentions of mobile platforms from godot-cpp "Getting started" document

Ivorforce:
Since this article was written, we've figured out that `.framework` and `.xcframework` is actually created from GDExtensions automatically, and the GDExtension does not (necessarily) have to provide the frameworks itself. It is simpler to just build and link the binaries, as with other platforms. Hence, I have removed this information from the article.

Furthermore, I have removed some entries from the `.gdextension` file. The file is incomplete anyway, and does not reflect our latest naming schemes. It is better to provide a short file that will suffice for the purposes of the article, and refer people to the template for a full file template.


mhilbrunner:
Thanks!

PR #11372, opened 2025-10-14
Document function parameter defaults in GDScript reference

Calinou:
The feature is indirectly mentioned just below, but there was no explicit mention of how it works and its caveats.

This also fixes some reStructuredText syntax issues.


mhilbrunner (approved):

mhilbrunner:
Thanks!

PR #11381, opened 2025-10-16
Document relative paths in `.gdextension` files being supported

Calinou:
- Mention support for comments in `.gdextension` files.
- List recommended import options for GDExtension icons.

___

- See https://github.com/godotengine/godot-docs-user-notes/discussions/266#discussioncomment-14696014.

### DIFF
--- a/tutorials/scripting/gdextension/gdextension_file.rst
+++ b/tutorials/scripting/gdextension/gdextension_file.rst
@@ -50,21 +50,25 @@ Here is an example of what that can look like:
 
 .. code-block:: none
 
+    ; A comment line starts with a semicolon. This line is ignored by the engine.
     [libraries]
 
-    macos.debug = "res://bin/libgdexample.macos.template_debug.framework"
-    macos.release = "res://bin/libgdexample.macos.template_release.framework"
-    windows.debug.x86_32 = "res://bin/libgdexample.windows.template_debug.x86_32.dll"
-    windows.release.x86_32 = "res://bin/libgdexample.windows.template_release.x86_32.dll"
-    windows.debug.x86_64 = "res://bin/libgdexample.windows.template_debug.x86_64.dll"
-    windows.release.x86_64 = "res://bin/libgdexample.windows.template_release.x86_64.dll"
-    linux.debug.x86_64 = "res://bin/libgdexample.linux.template_debug.x86_64.so"
-    linux.release.x86_64 = "res://bin/libgdexample.linux.template_release.x86_64.so"
-    linux.debug.arm64 = "res://bin/libgdexample.linux.template_debug.arm64.so"
-    linux.release.arm64 = "res://bin/libgdexample.linux.template_release.arm64.so"
-    linux.debug.rv64 = "res://bin/libgdexample.linux.template_debug.rv64.so"
-    linux.release.rv64 = "res://bin/libgdexample.linux.template_release.rv64.so"
+    macos.debug = "./bin/libgdexample.macos.template_debug.dylib" ; Inline comments are also allowed.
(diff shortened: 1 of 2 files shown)


Ivorforce (commented):
I would prefer to switch the example to use relative names. 
We discussed this in GDExtension meetings before, and concluded (iirc) that relative paths are preferable since this supports moving around and renaming the gdextension without having to worry about the `.gdextension` file. This is also the reason we changed it in the template.
I don't think there are any advantages to using absolute paths, so I'm not sure I'd even mention them.

Calinou:
> I would prefer to switch the example to use relative names.

Done.

Ivorforce (approved):

tutorials/scripting/gdextension/gdextension_file.rst
```diff
@@ -50,21 +50,25 @@ Here is an example of what that can look like:
...
-    windows.release.x86_64 = "res://bin/libgdexample.windows.template_release.x86_64.dll"
-    linux.debug.x86_64 = "res://bin/libgdexample.linux.template_debug.x86_64.so"
-    linux.release.x86_64 = "res://bin/libgdexample.linux.template_release.x86_64.so"
-    linux.debug.arm64 = "res://bin/libgdexample.linux.template_debug.arm64.so"
-    linux.release.arm64 = "res://bin/libgdexample.linux.template_release.arm64.so"
-    linux.debug.rv64 = "res://bin/libgdexample.linux.template_debug.rv64.so"
-    linux.release.rv64 = "res://bin/libgdexample.linux.template_release.rv64.so"
-
+    macos.debug = "bin/libgdexample.macos.template_debug.framework" ; Inline comments are also allowed.
+    macos.release = "bin/libgdexample.macos.template_release.framework"
```
dsnopek: In the template, we did all the paths starting with `./`, which, if I remember correctly was to fix a MacOS issue? And we also use `.dylib` in the template and not `.framework`, which I also remember discussing being to workaround some issue.

Do you folks remember the details?

Anyway, if there was a reason for those changes, we should do them in the docs as well

Ivorforce: Good point! Yea, I believe the macOS load issue is still around, so changing to `./` format would be good.

And yea, the change from framework to dylib also definitely makes sense.

mhilbrunner (approved):

tutorials/scripting/gdextension/gdextension_file.rst
mhilbrunner: @Calinou ^ Gentle poke :)

Calinou: Done. I've also taken the opportunity to change some paths that were still absolute elsewhere in the documentation, as well as fix some notes on feature tags.

tutorials/scripting/gdextension/gdextension_file.rst
```diff
@@ -149,46 +153,54 @@ For example:
...
+    with a directory structure like this: ``Game.app/Contents/Frameworks``.
 
 .. code-block:: none
 
     [dependencies]
 
     macos.debug = {
-        "res://bin/libdependency.macos.template_debug.framework" : "Contents/Frameworks"
+        "./bin/libdependency.macos.template_debug.dylib" : "Contents/Frameworks"
     }
```
Calinou, line 185: I'm not sure if I should have changed this part; let me know if I should revert that.

Ivorforce: That change is included in https://github.com/godotengine/godot-docs/pull/11405, and it requires changes to the `SConstruct` too. So better revert it in this PR.

mhilbrunner:
Thanks! Merged.

PR #11383, opened 2025-10-17
Recommend building just `compile_commands.json` in the `CLion` documentation

Ivorforce:
Running the command without the explicit `compile_commands.json` compiles the entire codebase. This isn't necessary. By providing the target, just the file can be generated.
This version is documented elsewhere too.


mhilbrunner:
Thanks!

PR #11388, opened 2025-10-19
Mention Zed in external editor exec flags

DeeJayLSP:
Editor counterpart: https://github.com/godotengine/godot/pull/111821

Although Zed is still in development, its state is already good enough for use with GDScript.

Flags are the same as Sublime Text, so it may be placed at its side.


Calinou (approved):

mhilbrunner (approved):

mhilbrunner:
Thanks!

Issue #11387, opened 2025-10-19
GDExtension C Example: ODR violations in `api.h`


charjr:
Following the [GDExtension C example](https://docs.godotengine.org/en/stable/tutorials/scripting/gdextension/gdextension_c_example.html) leads to compilation errors with `scons` due to multiple definition warnings.

These errors are specific to the `api`, `constructors` and `destructors` variables defined in [api.h](https://docs.godotengine.org/en/stable/tutorials/scripting/gdextension/gdextension_c_example.html#a-wrapper-api) which break the [one definition rule](https://en.wikipedia.org/wiki/One_Definition_Rule) by defining variables inside a header file:

https://github.com/godotengine/godot-docs/blob/e4d463e746f0c15e83da79075d42c23e09bc9ca2/tutorials/scripting/gdextension/gdextension_c_example.rst?plain=1#L291-L324

My suspicion is that these variables should be `static`, however I cannot confirm that because I have not managed to compile a working extension.

dsnopek:
Hm. I think the One Definition Rule is a C++ thing (not C), but perhaps these were meant to be `extern`?

I don't think we should be marking variables as `static` in a header - I think that would lead to a unique version of the variable existing in each translation unit that included the header.

dsnopek:
Oh, it looks like @vnen merged a PR to his complete version of that tutorial which marked them as `extern`:

https://github.com/vnen/gdextension-c-example/pull/1

Looks like it didn't make it's way back to the docs for some reason, though

dsnopek:
I just posted PR https://github.com/godotengine/godot-docs/pull/11389 to add `extern` at the variable declaration, and a definition in the `.c` file. I haven't tested it, but it seems right to me

@charjr Can you test if making the same change fixes the compilation errors you are seeing?

charjr:
@dsnopek I confirm that it solved my issues **with some additional changes**: https://github.com/godotengine/godot-docs/pull/11389#issuecomment-3419668109.

I do still have one _warning_: 
```
src/api.c: In function 'load_api':
src/api.c:23:52: warning: assignment to 'GDExtensionInterfaceStringNameNewWithLatin1Chars' {aka 'void (*)(void *, const char *, unsigned char)'} from incompatible pointer type 'GDExtensionInterfaceFunctionPtr' {aka 'void (*)()'} [-Wincompatible-pointer-types]
[...]

PR #11385, opened 2025-10-18
Remove erroneous include guard

charjr:
The tail end of an include guard is provided
within the defs.h code blocks,
however #pragma once is used at the top instead.

https://github.com/godotengine/godot-docs/blob/e4d463e746f0c15e83da79075d42c23e09bc9ca2/tutorials/scripting/gdextension/gdextension_c_example.rst?plain=1#L156-L172

https://github.com/godotengine/godot-docs/blob/e4d463e746f0c15e83da79075d42c23e09bc9ca2/tutorials/scripting/gdextension/gdextension_c_example.rst?plain=1#L486-L503

I am also failing to compile the example using scons:

```
/bin/ld: src/gdexample.os:/home/john/Godot/gd-extension-c-example/src/api.h:20: multiple definition of `constructors'; src/api.os:/home/john/Godot/gd-extension-c-example/src/api.h:20: first defined here
```

Which is either on my end (and I could use some help :pray: ) or indicative of other problems within this page.

PR #11389, opened 2025-10-19
Add `extern` to global struct variables in GDExtension C example

dsnopek:
Fixes https://github.com/godotengine/godot-docs/issues/11387

### DIFF
--- a/tutorials/scripting/gdextension/gdextension_c_example.rst
+++ b/tutorials/scripting/gdextension/gdextension_c_example.rst
@@ -306,17 +306,17 @@ We'll start by creating an ``api.h`` file in the ``src`` folder:
 
     // API methods.
 
-    struct Constructors
+    extern struct Constructors
     {
         GDExtensionInterfaceStringNameNewWithLatin1Chars string_name_new_with_latin1_chars;
     } constructors;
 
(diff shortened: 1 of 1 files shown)


charjr:
As well as adding `extern` in `api.h` they also have to then be defined in `api.c` like in vnen's example https://github.com/vnen/gdextension-c-example/blob/b560e60ab5f7f6bfd4b85e71cd68ad736562ebdf/src/api.c#L5-L9

Lastly, this change https://github.com/godotengine/godot-docs/pull/11385.

Vnen's examples use include guards, whereas the docs are using `#pragma once`. `defs.h` is documented with `#pragma once` at the top and then `#endif //DEFS_H` at the bottom. So one or the other should be changed.

dsnopek:
> As well as adding `extern` in `api.h` they also have to then be defined in `api.c` like in vnen's example https://github.com/vnen/gdextension-c-example/blob/b560e60ab5f7f6bfd4b85e71cd68ad736562ebdf/src/api.c#L5-L9

Yep, that's already in this PR

mhilbrunner:
Thanks! Merged.

akien-mga:
Cherry-picked for `4.5`.

PR #11390, opened 2025-10-19
Always cast function pointers in GDExtension C example

dsnopek:
This example is inconsistent about whether or not it casts the function pointers when fetching them

Skipping the cast will cause a compiler warning, so having the cast is probably better!

This PR attempts to add all the missing casts


mhilbrunner:
Thank you!

PR #11397, opened 2025-10-21
Remove obsolete paragraph about Linux power-saving inhibition

wjt:
Since https://github.com/godotengine/godot/pull/108704, the sentence about Flatpak is fully obsolete:

1. All standard Flatpak runtimes contain libdbus;
2. With the addition of support for the XDG Inhibit portal, Godot is able to inhibit idle without any sandboxing adjustments.

More generally, I don't think we need to mention that D-Bus is required to inhibit idle, even outside Flatpak: practically speaking, all desktop Linux systems have had libdbus available for 20+ years.

Remove the whole paragraph.


skyace65 (approved):

mhilbrunner:
Thank you!

PR #11365, opened 2025-10-09
Update release policy with info now that 4.5 is released

clayjohn:
I wasn't able to test this locally as the build seemed to stall at 0%

PR #11406, opened 2025-10-23
Release policy: Mark 3.5 and 4.3 as EOL, partial support for 4.4

akien-mga:
I cherry-picked a bunch of stuff for 4.3.1 many months ago, but never got to making a build and now there would be a ton more work needed to make a useful build (changing platform requirements, etc.).

Since 4.5.1 is out, it's time to EOL the old-old-stable branch (4.3). And since the migration has been pretty successful for most users, there doesn't seem to be a strong need to support 4.4.x for a long time, so we downgrade it to partial support (until 4.6 is released).

As for 3.x, we just released 3.6.2 to support changing platform requirements with some significant delay, so aiming to also release a 3.5.4 is unrealistic with our current resources. So it's also time to end support for 3.5. For 3.x in general I think we should likely pledge to support only the latest stable and not also the previous stable like we do with 4.x.

This needs to be cherry-picked pretty much everywhere - until we finally move this table out of the docs so we don't need to update all versions whenever a new release happens or old releases are EOL.

### DIFF
--- a/about/release_policy.rst
+++ b/about/release_policy.rst
@@ -90,10 +90,9 @@ on GitHub.
 | Godot 4.5    | September 2025       | |supported| Receives fixes for bugs and security issues, as well as      |
 |              |                      | patches that enable platform support.                                    |
 +--------------+----------------------+--------------------------------------------------------------------------+
-| Godot 4.4    | March 2025           | |supported| Receives fixes for bugs and security issues, as well as      |
-|              |                      | patches that enable platform support.                                    |
+| Godot 4.4    | March 2025           | |partial| Receives fixes for security and platform support issues only.  |
 +--------------+----------------------+--------------------------------------------------------------------------+
-| Godot 4.3    | August 2024          | |partial| Receives fixes for security and platform support issues only.  |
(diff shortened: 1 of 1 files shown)


clayjohn:
For context, before 4.5, we were listing the last two minor releases as supported, and the third minor release as "partial". I did the update for 4.5 already here: https://github.com/godotengine/godot-docs/pull/11365. Therefore, this PR has the effect of tightening up our supported releases. 

I am not opposed to doing so, as I think in practice, we have no intention of doing another 4.3 point release.

Calinou (approved):
Makes sense to me.

lawnjelly:
Sounds good for 3.x too, I don't think there was anything significantly breaking between 3.5 and 3.6, so there isn't a whole lot of reason to keep support for 3.5 imo.

akien-mga:
> For context, before 4.5, we were listing the last two minor releases as supported, and the third minor release as "partial". I did the update for 4.5 already here: #11365. Therefore, this PR has the effect of tightening up our supported releases.

Yeah this is a clarification of what we're actually able to do currently. We haven't done a good job at fully supporting the second-to-last minor releases for a while now.

PR #11392, opened 2025-10-21
Update Ragdoll system documentation for Godot 4.5

Calinou:
- Update all images.
- Replace GIFs with videos.
- Document stopping a ragdoll simulation and collision exceptions.
- Add @TranquilMarmot's advice from the user notes on improving visibility, multiple collision shapes per bone, joint descriptions, and recommended joint types.
- Update recommended collision layer/mask to ensure the ragdoll collides with the GridMap, as per @N-o-a-h-t-u-r-n-e-r's comment in the user notes.

This should be cherry-pickable to 4.4, as all features mentioned are also available there.

Modified 3D platformer demo with complete setup: [platformer_ragdoll.zip](https://github.com/user-attachments/files/23011246/platformer_ragdoll.zip)
*Press <kbd>B</kbd> to activate partial ragdoll on arms.*

Image source: [ragdoll_system_collision_layers_masks.xcf.zip](https://github.com/user-attachments/files/24782379/ragdoll_system_collision_layers_masks.xcf.zip)

___

- See https://github.com/godotengine/godot-docs-user-notes/discussions/186#discussioncomment-14639150.

PR #11409, opened 2025-10-23
Link new Ragdoll Physics demo in Ragdoll system

Calinou:
- Follow-up to https://github.com/godotengine/godot-docs/pull/11392.

___

- Add tip to select all PhysicalBone3D nodes in the section about changing collision layers/masks.

I missed this while updating the page earlier this week.

The demo has only been tested on 4.5 onwards, so I suggest not cherry-picking this to 4.4.


mhilbrunner:
Thank you!

PR #11426, opened 2025-10-28
Explicitly destroy Vector2

### LINKED ISSUE #11414: GDExtension C Example - Vector2 misleadingly not cleaned up
## Problem

[Within the section on registering a signal](https://docs.godotengine.org/en/stable/tutorials/scripting/gdextension/gdextension_c_example.html#registering-and-emitting-a-signal) a helper function is defined which states afterwards:

> At the end we need to destruct the Variants we created. While technically the
Vector2 one does not require destructing, it is clearer to cleanup everything.

There are two issues:
1. This message is misleading, as it makes me think _it has_ been cleaned up.
2. Nothing explains why Vector2 doesn't need to be destroyed but other things do.

https://github.com/godotengine/godot-docs/blob/e11d59a2eee6210575734f705387e3a768e5da9d/tutorials/scripting/gdextension/gdextension_c_example.rst?plain=1#L1998-L2030

## Solution

Make it clear that _it hasn't_ been cleaned up _but could be_ (presumably at the expense of more writing more boilerplate).

Alternatively, actually destroy it in the code (if it doesn't expand the docs much).

Ideally, it would be nice to know why Vector2 doesn't need to be destroyed, I noticed in an earlier step it wasn't destroyed either and this was a point of confusion for me.

charjr:
Close #11414 

Tried this locally, my example still works.

@AThousandShips Interesting to note, I also tested the extension with no destruction whatsoever with Godot running `verbose` stdout. I've left it running for over 10 minutes, No warnings of any kind, Godot's built-in monitors show static memory has not changed at all. :shrug:

### DIFF
--- a/tutorials/scripting/gdextension/gdextension_c_example.rst
+++ b/tutorials/scripting/gdextension/gdextension_c_example.rst
@@ -2010,8 +2010,9 @@ implement the helper function.
         // Call the function.
         api.object_method_bind_call(p_method_bind, p_instance, args, 2, &ret, NULL);
 
-        // Destroy the arguments that need it.
+        // Destroy the arguments.
         destructors.variant_destroy(&arg1);
+        destructors.variant_destroy(&arg2);
         destructors.variant_destroy(&ret);
     }
 


dsnopek:
The added `variant_destroy()` call won't actually do anything on the Godot side, because [`Variant::needs_deinit`](https://github.com/godotengine/godot/blob/2cb2c1fb5bbeb2afbac7255841f9c66d9611bf55/core/variant/variant.h#L292) is `false` for `VECTOR2`

However, it also won't cause any issues, and so maybe it's better to demonstrate always calling `variant_destroy()` rather than using internal knowledge to skip it?

Looking at godot-cpp's implementation of `Variant`, it's always calling `variant_destroy()` as well, and not trying to be clever about only calling it for certain types

charjr:
@dsnopek 
> The added `variant_destroy()` call won't actually do anything on the Godot side, because [`Variant::needs_deinit`](https://github.com/godotengine/godot/blob/2cb2c1fb5bbeb2afbac7255841f9c66d9611bf55/core/variant/variant.h#L292) is `false` for `VECTOR2`

@AThousandShips showed me this as well and I just want to say I truly appreciate both of you choosing to point me at the right bit of Godot's internal code. It's not the first time the Godot community have done this for me and each time I come out more confident in my understanding. So thank you. :pray: 

> However, it also won't cause any issues, and so maybe it's better to demonstrate always calling `variant_destroy()` rather than using internal knowledge to skip it?

I think in this instance we should demonstrate it to match the text below that code snippet:

https://github.com/godotengine/godot-docs/blob/ae0c5ee944a1c8f3ba7eae2012785346224a1f18/tutorials/scripting/gdextension/gdextension_c_example.rst?plain=1#L2029-L2030

charjr:
@dsnopek Rebased and seems fine now. :+1:

mhilbrunner:

PR #11437, opened 2025-11-03
Standard_material_3d.rst - Material Shadow Mode names Updated

PHATASS-Studios:
Material Shadow Mode names were wrong. Documentation read  "Do Not Receive Shadows" And "Use Shadow To Opacity"

As of Godot 4.5, Editor's actual fields read:
"Disable Receive Shadows" and "Shadow To Opacity".

Updated to match.


mhilbrunner:

PR #11438, opened 2025-11-04
Update the Object class engine details page.

Ivorforce:
This refreshes some outdated information, and adds a few missing bits and pieces:

- Group related headings together
- Use ``GDREGISTER_CLASS`` instead of ``ClassDB::register_class`` (which is no longer recommended)
- Explain other models of class privateness
    - So far, this has not been explained (afaik), and leads to regular confusion among godot-cpp users.
- Add `Object ownership and casting` heading, where the different ownership models are explained.
    - So far, this has not been explained (afaik), and leads to regular confusion among godot-cpp users.

### DIFF
--- a/engine_details/architecture/object_class.rst
+++ b/engine_details/architecture/object_class.rst
@@ -12,59 +12,73 @@ General definition
 ------------------
 
 :ref:`Object <class_object>` is the base class for almost everything. Most classes in Godot
-inherit directly or indirectly from it. Objects provide reflection and
-editable properties, and declaring them is a matter of using a single
+inherit directly or indirectly from it. Declaring them is a matter of using a single
 macro like this:
 
 .. code-block:: cpp
 
     class CustomObject : public Object {
-
-        GDCLASS(CustomObject, Object); // this is required to inherit
+        GDCLASS(CustomObject, Object); // This is required to inherit from Object.
     };
 
-This adds a lot of functionality to Objects. For example:
+Objects come with a lot of built-in functionality, like reflection and editable properties:
 
 .. code-block:: cpp
 
-    obj = memnew(CustomObject);
+    CustomObject *obj = memnew(CustomObject);
     print_line("Object class: ", obj->get_class()); // print object class
 
-    obj2 = Object::cast_to<OtherClass>(obj); // converting between classes, this also works without RTTI enabled.
+    OtherClass *obj2 = Object::cast_to<OtherClass>(obj); // Converting between classes, similar to dynamic_cast
 
 References:
 ~~~~~~~~~~~
 
 -  `core/object/object.h <https://github.com/godotengine/godot/blob/master/core/object/object.h>`__
 
-Registering an Object
----------------------
+Registering Object classes
+--------------------------
 
-ClassDB is a static class that holds the entire list of registered
-classes that inherit from Object, as well as dynamic bindings to all
-their methods properties and integer constants.
-
-Classes are registered by calling:
+Most ``Object`` subclasses are registered by calling ``GDREGISTER_CLASS``.
 
 .. code-block:: cpp
 
-    ClassDB::register_class<MyCustomClass>()
+    GDREGISTER_CLASS(MyCustomClass)
 
-Registering it will allow the class to be instanced by scripts, code, or
-creating them again when deserializing.
+This will register it as a named, public class in the ``ClassDB``, which will allow the class to be instantiated by
+scripts, code, or by deserialization. Note that classes registered as ``GDREGISTER_CLASS`` should expect to be
+instantiated or freed automatically, for example by the editor or the documentation system.
 
-Registering as virtual is the same but it can't be instanced.
+Besides ``GDREGISTER_CLASS``, there are a few other modes of privateness:
 
 .. code-block:: cpp
 
-    ClassDB::register_virtual_class<MyCustomClass>()
+    // Registers the class publicly, but prevents automatic instantiation through ClassDB.
+    GDREGISTER_VIRTUAL_CLASS(MyCustomClass);
+
+    // Registers the class publicly, but prevents all instantiation through ClassDB.
+    GDREGISTER_ABSTRACT_CLASS(MyCustomClass);
+
+    // Registers the class in ClassDB, but marks it as private,
+    // such that it is not visible to scripts or extensions.
+    // This is the same as not registering the class explicitly at all
+    // - in this case, the class is registered as internal automatically
+    // when it is first constructed.
+    GDREGISTER_INTERNAL_CLASS(MyCustomClass);
+
+    // Registers the class such that it is only available at runtime (but not in the editor).
+    GDREGISTER_RUNTIME_CLASS(MyCustomClass);
+
+It is also possible to use ``GDSOFTCLASS(MyCustomClass, SuperClass)`` instead of ``GDCLASS(MyCustomClass, SuperClass)``.
+Classes defined this way are not registered in the ``ClassDB`` at all. This is sometimes used for platform-specific
+subclasses.
+
+Registering bindings
+~~~~~~~~~~~~~~~~~~~~
 
 Object-derived classes can override the static function
-``static void _bind_methods()``. When one class is registered, this
+``static void _bind_methods()``. When the class is registered, this
 static function is called to register all the object methods,
-properties, constants, etc. It's only called once. If an Object derived
-class is instanced but has not been registered, it will be registered as
-virtual automatically.
+properties, constants, etc. It's only called once.
 
 Inside ``_bind_methods``, there are a couple of things that can be done.
 Registering functions is one:
@@ -95,12 +109,12 @@ documented as thoroughly, the ``D_METHOD()`` macro can safely be ignored and a
 string passing the name can be passed for brevity.
 
 References:
-~~~~~~~~~~~
+^^^^^^^^^^^
 
 -  `core/object/class_db.h <https://github.com/godotengine/godot/blob/master/core/object/class_db.h>`__
 
 Constants
----------
+~~~~~~~~~
 
 Classes often have enums such as:
 
@@ -126,7 +140,7 @@ The constants can also be bound inside ``_bind_methods``, by using:
     BIND_CONSTANT(MODE_SECOND);
 
 Properties (set/get)
---------------------
+~~~~~~~~~~~~~~~~~~~~
 
 Objects export properties, properties are useful for the following:
 
@@ -181,7 +195,7 @@ This creates the property using the setter and the getter.
 .. _doc_binding_properties_using_set_get_property_list:
 
 Binding properties using ``_set``/``_get``/``_get_property_list``
------------------------------------------------------------------
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 An additional method of creating properties exists when more flexibility
 is desired (i.e. adding or removing properties on context).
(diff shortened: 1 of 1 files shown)


engine_details/architecture/object_class.rst
```diff
@@ -244,35 +240,108 @@ Adding signals to a class is done in ``_bind_methods``, using the
...
+
+Godot provides dynamic casting between Object-derived classes, for example:
+
+.. code-block:: cpp
+
+    void somefunc(Object *some_obj) {
+         Button *button = Object::cast_to<Button>(some_obj);
+    }
+
+If cast fails, ``nullptr`` is returned. This works the same as ``dynamic_cast``, but does not use
```
AThousandShips:
```suggestion
If the cast fails, ``nullptr`` is returned. This works the same as ``dynamic_cast``, but does not use
```

engine_details/architecture/object_class.rst
```diff
@@ -244,35 +240,108 @@ Adding signals to a class is done in ``_bind_methods``, using the
...
-method that allows it to respond to engine level callbacks that may relate to it.
-More information can be found on the :ref:`doc_godot_notifications` page.
+Objects are allocated on the heap. There are two different ownership models:
 
-References
-----------
+- Objects derived from ``RefCounted`` are reference counted.
+- All other objects are manually memory managed.
+
+The ownership models are fundamentally different. Refer to the section for each respectively to learn about how to
```
AThousandShips:
```suggestion
The ownership models are fundamentally different. Refer to the section for each respectively to learn how to
```

engine_details/architecture/object_class.rst
```diff
@@ -12,59 +12,72 @@ General definition
...
 
-    ClassDB::register_virtual_class<MyCustomClass>()
+    // Registers the class publicly, but prevents automatic instantiation through ClassDB.
+    GDREGISTER_VIRTUAL_CLASS(MyCustomClass);
+
+    // Registers the class publicly, but prevents all instantiation through ClassDB.
+    GDREGISTER_ABSTRACT_CLASS(MyCustomClass);
+
+    // Registers the class in ClassDB, but marks it as private,
+    // such that it is not visible by scripts or extensions.
```
AThousandShips:
```suggestion
    // such that it is not visible to scripts or extensions.
```

engine_details/architecture/object_class.rst
```diff
@@ -12,59 +12,72 @@ General definition
...
 .. code-block:: cpp
 
-    ClassDB::register_class<MyCustomClass>()
+    GDREGISTER_CLASS(MyCustomClass)
 
-Registering it will allow the class to be instanced by scripts, code, or
-creating them again when deserializing.
+This will register it as a named, public class in ``ClassDB``, which will allow the class to be instantiated by scripts,
+code, or by deserialization. Note that classes registered as ``GDREGISTER_CLASS`` should expect to be instantiated or
+freed automatically, like by the editor or the documentation systems.
```
AThousandShips:
```suggestion
freed automatically, like by the editor or the documentation system.
```

engine_details/architecture/object_class.rst
```diff
@@ -12,59 +12,72 @@ General definition
...
+Most ``Object`` subclasses are registered by calling ``GDREGISTER_CLASS``.
 
 .. code-block:: cpp
 
-    ClassDB::register_class<MyCustomClass>()
+    GDREGISTER_CLASS(MyCustomClass)
 
-Registering it will allow the class to be instanced by scripts, code, or
-creating them again when deserializing.
+This will register it as a named, public class in ``ClassDB``, which will allow the class to be instantiated by scripts,
```
AThousandShips:
```suggestion
This will register it as a named, public class in the ``ClassDB``, which will allow the class to be instantiated by scripts,
```
I'd say, it's unique and we refer to other servers and singletons in definite as well generally AFAIK

engine_details/architecture/object_class.rst
```diff
@@ -244,35 +240,108 @@ Adding signals to a class is done in ``_bind_methods``, using the
...
+
+When storing objects you are not the only owner of, you should store its ``ObjectID`` rather than a pointer:
+
+.. code-block:: cpp
+
+    Node *node = memnew(Node);
+    ObjectID node_id = node.get_instance_id();
+    // ...
+    Object *maybe_node = ObjectDB::get_instance(node_id);
+    ERR_FAIL_NULL(maybe_node); // Node may have been freed in between calls.
```
AThousandShips:
```suggestion
    ERR_FAIL_NULL(maybe_node); // The node may have been freed between calls.
```

engine_details/architecture/object_class.rst
```diff
@@ -244,35 +240,108 @@ Adding signals to a class is done in ``_bind_methods``, using the
...
+    // ...
+    my_ref->get_class_name(); // Calls to Ref are always safe because of shared ownership.
 
-``myref`` is reference counted. It will be freed when no more Ref<>
-templates point to it.
+You should never call ``memdelete`` for ``RefCounted`` subclasses, because there may be other owners of it.
+
+You should also never store ``RefCounted`` subclasses using raw pointers, for example
+``RefCounted *object = memnew(RefCounted)``. This is unsafe because other owners may destruct the object, leaving you
+with a dangling pointer, which will eventually result in a crash.
```
AThousandShips, line 321: A bit of repetition here with "will eventually result in a crash"

engine_details/architecture/object_class.rst
```diff
@@ -244,35 +240,108 @@ Adding signals to a class is done in ``_bind_methods``, using the
...
+
+Dynamic casting
+~~~~~~~~~~~~~~~
+
+Godot provides dynamic casting between Object-derived classes, for example:
+
+.. code-block:: cpp
+
+    void somefunc(Object *some_obj) {
+         Button *button = Object::cast_to<Button>(some_obj);
```
AThousandShips:
```suggestion
    void somefunc(Object *p_some_obj) {
         Button *button = Object::cast_to<Button>(p_some_obj);
```
Let's follow our own style guide in the examples

Ivorforce: I wash my hands of the matter, this code already existed in the old version :P 
(but yes, we should)

engine_details/architecture/object_class.rst
Ivorforce: It's in both sections, yes. But people may not read both.
Do you think we should rephrase one?

AThousandShips: I think a bit of a rephrase would be the best option

engine_details/architecture/object_class.rst
```diff
@@ -244,35 +241,108 @@ Adding signals to a class is done in ``_bind_methods``, using the
...
+Godot provides dynamic casting between Object-derived classes, for example:
+
+.. code-block:: cpp
+
+    void some_func(Object *p_object) {
+         Button *button = Object::cast_to<Button>(p_object);
+    }
+
+If the cast fails, ``nullptr`` is returned. This works the same as ``dynamic_cast``, but does not use
+`C++ .RTTI <https://en.wikipedia.org/wiki/Run-time_type_information>`__.
```
Calinou: @copilot 
```suggestion
`C++ RTTI <https://en.wikipedia.org/wiki/Run-time_type_information>`__.
```

engine_details/architecture/object_class.rst
```diff
@@ -244,35 +241,108 @@ Adding signals to a class is done in ``_bind_methods``, using the
...
+
+Notifications
+-------------
+
+All objects in Godot have a :ref:`_notification <class_Object_private_method__notification>`
+method that allows it to respond to engine level callbacks that may relate to it.
+More information can be found on the :ref:`doc_godot_notifications` page.
```
Calinou, line 347:
```suggestion
All objects in Godot have a :ref:`_notification <class_Object_private_method__notification>`
method that allows it to respond to engine-level callbacks that may relate to it.
More information can be found on the :ref:`doc_godot_notifications` page.
```

engine_details/architecture/object_class.rst
Ivorforce: You got any good ideas? I don't mind a little repetition, if it's addressing the same problem in different contexts.

AThousandShips: I'm not sure, we can leave it as it is for now

engine_details/architecture/object_class.rst
```diff
@@ -244,35 +241,108 @@ Adding signals to a class is done in ``_bind_methods``, using the
...
+    }
+
+If the cast fails, ``nullptr`` is returned. This works the same as ``dynamic_cast``, but does not use
+`C++ RTTI <https://en.wikipedia.org/wiki/Run-time_type_information>`__.
+
+Notifications
+-------------
+
+All objects in Godot have a :ref:`_notification <class_Object_private_method__notification>`
+method that allows it to respond to engine-level callbacks that may relate to it.
```
AThousandShips:
```suggestion
method that allows them to respond to engine-level callbacks that may relate to it.
```
Or use "each object"

engine_details/architecture/object_class.rst
```diff
@@ -244,35 +241,108 @@ Adding signals to a class is done in ``_bind_methods``, using the
...
+.. code-block:: cpp
+
+    class MyRefCounted: public RefCounted {
         GDCLASS(MyReference, RefCounted);
     };
 
-    Ref<MyReference> myref(memnew(MyReference));
+    Ref<MyRefCounted> my_ref = memnew(MyRefCounted);
+    // ...
+    my_ref->get_class_name(); // Calls to Ref are always safe because of shared ownership.
```
AThousandShips: This might be misleading if interpreted as "any time you have a `Ref`" as opposed to "as long as you have a valid `Ref`", might want to clarify that somehow, for example when the `Ref` is a method argument

engine_details/architecture/object_class.rst
```diff
@@ -244,35 +241,108 @@ Adding signals to a class is done in ``_bind_methods``, using the
...
+
+.. code-block:: cpp
+
+    Node *node = memnew(Node);
+    // ...
+    memdelete(node);
+    node = nullptr;
+
+When you are not the sole owner of an object, storing a pointer to it is dangerous: Any other references to the object
+may call ``memdelete`` at any time, causing your pointer to become a dangling pointer, which will eventually result in a
```
AThousandShips: Nitpick but the reference would never call `memdelete` so might be good to reword

engine_details/architecture/object_class.rst
```diff
@@ -12,59 +12,73 @@ General definition
...
 .. code-block:: cpp
 
-    ClassDB::register_class<MyCustomClass>()
+    GDREGISTER_CLASS(MyCustomClass)
 
-Registering it will allow the class to be instanced by scripts, code, or
-creating them again when deserializing.
+This will register it as a named, public class in the ``ClassDB``, which will allow the class to be instantiated by
+scripts, code, or by deserialization. Note that classes registered as ``GDREGISTER_CLASS`` should expect to be
+instantiated or freed automatically, like by the editor or the documentation system.
```
AThousandShips:
```suggestion
instantiated or freed automatically, for example by the editor or the documentation system.
```
I'd say here as in this case "like" would be pretty casual so I'd avoid it

engine_details/architecture/object_class.rst
Ivorforce: I added some more detail about this property.

engine_details/architecture/object_class.rst
Ivorforce: By 'reference' I mean "some other part of the code that has a reference to the object."
I don't know a good word for it otherwise. Got any ideas?

AThousandShips: It's not about the word it's that it says that the object itself might delete itself, I'd say instead something like "any other reference may be deleted", or "`memdelete` may be called on any other reference"

engine_details/architecture/object_class.rst
```diff
@@ -244,35 +241,111 @@ Adding signals to a class is done in ``_bind_methods``, using the
...
+    class MyRefCounted: public RefCounted {
         GDCLASS(MyReference, RefCounted);
     };
 
-    Ref<MyReference> myref(memnew(MyReference));
+    Ref<MyRefCounted> my_ref = memnew(MyRefCounted);
+    // ...
+    // Ref holds shared ownership over the object, so the object
+    // will not destruct . As long as you have a valid, non-null
+    // Ref, it can be safely  assumed the object still valid.
```
AThousandShips:
```suggestion
    // will not be freed. As long as you have a valid, non-null
    // Ref, it can be safely assumed the object is still valid.
```
"Destruct" isn't really used elsewhere and introduces a bit of confusing separate terminology, so I'd say to keep it standard

engine_details/architecture/object_class.rst
Ivorforce: I think i found a nice way of putting it.

Calinou (approved):
Looks good to me.

mhilbrunner (approved):

mhilbrunner:
Thank you!

PR #11439, opened 2025-11-04
Refresh core_types.rst to remove outdated and redundant information.

Ivorforce:
The page currently contains quite a lot of outdated, redundant, or unnecessary information.

In the interest of streamlining what is needed, I'm proposing to remove these parts:

**Definitions**
- The use of C99 types is now so standard that it is not necessary to mention it. I've never seen a codebase that "reinvents the wheel" and provides its own aliases. Anyone even remotely used to C or C++ will recognize `uint32_t`.
- The page makes recommendations about whether to use platform dependent (e.g. `int`) or platform independent (e.g. `uint32_t`) types. Our stance towards this has changed, and we now prefer platform independent types. Anyway, recommendations are better saved for the contributing docs, so this should be removed.
- The claim that `size_t` is always 64 bit is plain wrong, it's 32-bit in 32-bit binaries.
- `CharType` no longer exists, `wchar_t` is used.

**Memory Model**
While a nice read, I don't think this text is needed. There are no practical take-aways from this section. Some of the claims are now wrong or misleading. For example, Godot's types are not always minimal — `Object` types are unexpectedly large, and they are some of Godot's most common types.
The part about allocation and segmentation details has unbacked claims. Since the text was written (earlier than 2015), allocators and memory management have changed. Besides, I've seen reports of modern C++ projects that changed significantly in RAM use when using different allocators, due to fragmentation. It is not relevant to document this, and new users of the engine do not need to read this. 
The final part with memory pools is no longer relevant (or at least, no longer nearly as relevant). Since 4.0, memory pool containers have been removed. Only some memory pools are now used in the engine, most private to specific classes.

**Allocating memory**
- Tidied up the examples
- Removed the recommendations texts. While they are somewhat new, the texts have been superseded by the below table, which contains more and more relevant information.
- The threading safety part has been transferred to its own little section.

### DIFF
--- a/engine_details/architecture/core_types.rst
+++ b/engine_details/architecture/core_types.rst
@@ -9,87 +9,30 @@ and everything is built upon them.
 This reference will try to list them in order for their better
 understanding.
 
-Definitions
------------
-
-Godot uses the standard C99 datatypes, such as ``uint8_t``,
-``uint32_t``, ``int64_t``, etc. which are nowadays supported by every
-compiler. Reinventing the wheel for those is not fun, as it makes code
-more difficult to read.
-
-In general, care is not taken to use the most efficient datatype for a
-given task unless using large structures or arrays. ``int`` is used
-through most of the code unless necessary. This is done because nowadays
-every device has at least a 32-bit bus and can do such operations in
-one cycle. It makes code more readable too.
-
-For files or memory sizes, ``size_t`` is used, which is guaranteed to be
-64-bit.
-
-For Unicode characters, CharType instead of wchar_t is used, because
-many architectures have 4 bytes long wchar_t, where 2 bytes might be
-desired. However, by default, this has not been forced and CharType maps
-directly to wchar_t.
-
--  `core/typedefs.h <https://github.com/godotengine/godot/blob/master/core/typedefs.h>`__
-
-Memory model
-------------
-
-PC is a wonderful architecture. Computers often have gigabytes of RAM,
-terabytes of storage and gigahertz of CPU, and when an application needs
-more resources the OS will swap out the inactive ones. Other
-architectures (like mobile or consoles) are in general more limited.
-
-The most common memory model is the heap, where an application will
-request a region of memory, and the underlying OS will try to fit it
-somewhere and return it. This often works best and is flexible,
-but over time and with abuse, this can lead to segmentation.
-
-Segmentation slowly creates holes that are too small for most common
-allocations, so that memory is wasted. There is a lot of literature
-about heap and segmentation, so this topic will not be developed
-further here. Modern operating systems use paged memory, which helps
-mitigate the problem of segmentation but doesn't solve it.
-
-However, in many studies and tests, it is shown that given enough
-memory, if the maximum allocation size is below a given threshold in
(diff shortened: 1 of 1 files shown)


Calinou:
> The use of C99 types is now so standard that it is not necessary to mention it. I've never seen a codebase that "reinvents the wheel" and provides its own aliases.

[I've seen a few examples out there](https://github.com/ioquake/ioq3/blob/8c48475ca3081c84e168bb0ceaf7b65dcb3cab6a/code/qcommon/q_shared.h#L194) but I agree it's very uncommon now, so it's safe to remove.

engine_details/architecture/core_types.rst
```diff
@@ -98,33 +41,9 @@ For C++-style allocation, special macros are provided:
...
-unrelated types with similar interfaces, both using a buffer as their
-storage strategy.
-
-``List<>`` is another Godot sequence type, using a doubly-linked list as
-its storage strategy. Prefer ``Vector<>`` (or ``LocalVector<>``) over
-``List<>`` unless you're sure you need it, as cache locality and memory
-fragmentation tend to be more important with small collections.
+``memnew``/``memdelete`` also use a little C++ magic to automatically call post-init
+and pre-release functions, for example to notify Objects right after they are created,
+and right before they are deleted.
```
Calinou:
```suggestion
``memnew``/``memdelete`` also use a little C++ magic to automatically call post-init
and pre-release functions. For example, this is used to notify Objects right after they are created,
and right before they are deleted.
```

engine_details/architecture/core_types.rst
```diff
@@ -231,6 +150,12 @@ scripting API.
 .. |typed_dictionary| replace:: `TypedDictionary <https://github.com/godotengine/godot/blob/master/core/variant/typed_dictionary.h>`__
 .. |pair| replace:: `Pair <https://github.com/godotengine/godot/blob/master/core/templates/pair.h>`__
 
+Thread safety
+~~~~~~~~~~~~~
+
+None of Godot's containers are thread-safe. When you expect multiple threads to access them, you must use multithread
+protections, like ``Mutex`` or ``Semaphore``.
```
Calinou, line 157:
```suggestion
None of Godot's containers are thread-safe. When you expect multiple threads to access them, you must use multithread
protections, like :ref:`class_mutex` or :ref:`class_semaphore`.
```

Ivorforce: This is not entirely correct. The `Mutex` and `Semaphore` we use internally are different from the exposed versions. Generally, the exposed versions should not be used internally, and they do not share the same interface.

mhilbrunner (approved):

Calinou (approved):
Looks good to me. I agree the removed text wasn't really relevant in 2025, especially this part:

> Computers often have gigabytes of RAM, terabytes of storage and gigahertz of CPU, and when an application needs more resources the OS will swap out the inactive ones. Other architectures (like mobile or consoles) are in general more limited.

If anything, modern Android devices are known to be [too eager to swap out inactive apps](https://dontkillmyapp.com/) :slightly_smiling_face:

mhilbrunner:
Thank you!

PR #11441, opened 2025-11-04
Document `textureSize()` should be avoided in Built-in functions

Calinou:
For performance reasons, it's best to pass the texture size as a uniform instead.

- See https://github.com/godotengine/godot/pull/100280
and https://github.com/godotengine/godot/issues/105721.


mhilbrunner:
Thanks!

PR #11450, opened 2025-11-06
Fetch redirects based on the current URL.

Ivorforce:
This rewrites the URL fetch to use the current branch of the docs for redirects (e.g. `/en/4.5`) instead of `latest` for everything.

This will be needed in the future when redirects diverge. For now, we only use the js redirects on 4.5 and `master`, which haven't diverged much yet.

> [!WARNING]  
> I can't fully test this locally, because serving with python doesn't seem to incorporate the 404 page correctly.
> However, testing with the REPL in the browser, this should work correctly (even for local testing, if someone gets it to work fully).

_Bugsquad edit: Followup to https://github.com/godotengine/godot-docs/pull/11304_

### DIFF
--- a/404.rst
+++ b/404.rst
@@ -37,7 +37,7 @@ Page not found
     // This is done in JavaScript, as we exceed Read the Docs' limit for the amount of redirects configurable.
 
     const currentPathSegments = window.location.pathname.split('/').filter(segment => segment !== '');
-    // Use the base path (e.g. "/en/latest") when available
+    // Use the base path (e.g. "/en/latest") when available.
     const currentBasePath = (currentPathSegments.length >= 2) ? ("/" + currentPathSegments.slice(0, 2).join("/")) : "/";
 
     fetch(currentBasePath + "/_static/redirects.csv")


Calinou:
> I can't fully test this locally, because serving with python doesn't seem to incorporate the 404 page correctly.

I use https://github.com/TheWaWaR/simple-http-server with the following options while in `_build/html`:

```
--try-file 404.html
```

Python's SimpleHTTPServer also supports serving a custom 404 page, but it's more involved as you can't just use a CLI argument to enable it.

mhilbrunner (approved):

404.rst
```diff
@@ -35,8 +35,12 @@ Page not found
 
     // Check for redirects if on a currently invalid page.
     // This is done in JavaScript, as we exceed Read the Docs' limit for the amount of redirects configurable.
-    // When testing this feature on a local web server, replace the URL below with just `/_static/redirects.csv`.
-    fetch("/en/latest/_static/redirects.csv")
+
+    const currentPathSegments = window.location.pathname.split('/').filter(segment => segment !== '');
+    // Use the base path (e.g. "/en/latest") when available
```
Calinou:
```suggestion
    // Use the base path (e.g. "/en/latest") when available.
```

Calinou (approved):

mhilbrunner:
Well, lets try this on RTD then :) Thank you, merged!

Ivorforce:
Cherry-picked to 4.5 in https://github.com/godotengine/godot-docs/pull/11639

PR #11457, opened 2025-11-10
Tweak suggested energy values to simulate volumetric lighting in Volumetric fog

Calinou:
4.6 has fixed the previously incorrect volumetric fog blending behavior. This has the side-effect of making bright volumetric fog look significantly brighter than before. To get a result similar to the previous look, the Volumetric Fog Energy property on Light3D nodes needs to be reduced significantly (often by a factor of 10, if not more).

- See https://github.com/godotengine/godot/pull/112494.

Not cherry-pickable to `4.5`, as this change will only be in 4.6.


mhilbrunner (approved):

skyace65:
Thanks!

PR #11463, opened 2025-11-13
Change 'subclasses' to 'inner classes' in style guide

wk-y:
Subclasses seems to be a typo here, if the intent is for all `class ClassName: ...` declarations to be put at the end.


mhilbrunner:
Thank you for contributing! 🎉 Merged.

PR #11470, opened 2025-11-15
Minor spelling correction from 'known' to 'know'.

wavyshore:


Calinou:

Contributions to the class reference should be sent on the main Godot repository, not here. This is because the class reference in this repository is generated automatically based on the XML files in the main repository. See [Contribute to the class reference](https://contributing.godotengine.org/en/latest/documentation/class_reference.html) in the documentation :slightly_smiling_face:

PR #11473, opened 2025-11-17
Add Java/Kotlin to the list of other language bindings 

stan220:
Add Java/Kotlin language link

### DIFF
--- a/tutorials/scripting/other_languages.rst
+++ b/tutorials/scripting/other_languages.rst
@@ -23,6 +23,7 @@ The bindings below are developed and maintained by the community:
 
 - `D <https://github.com/godot-dlang/godot-dlang>`__
 - `Go <https://github.com/grow-graphics/gd>`__
+- `Java/Kotlin <https://github.com/utopia-rise/godot-kotlin-jvm>`__
 - `Nim <https://github.com/godot-nim/gdext-nim>`__
 - `Rust <https://github.com/godot-rust/gdext>`__
 - `Swift <https://github.com/migueldeicaza/SwiftGodot>`__


stan220:
Hi, guys @tetrapod00, @paddy-exe, @dsnopek, @skyace65, @aaronfranke, @bruvzg, @AThousandShips, @mhilbrunner, @Mickeon Could you plz approve JVM project? This project is not mine, but I use it and I want more people to know about it.

AThousandShips:
Please do not ping random people to bump your PR, people will take a look at this when they are able, but trying to rush people is not going to make it go faster, quite the opposite

stan220:
> Please do not ping random people to bump your PR, people will take a look at this when they are able, but trying to rush people is not going to make it go faster, quite the opposite

Oh, sorry. I just added the people who approved the last commit to this file. I understand you, I won't do that anymore.

dsnopek:
One thing that may be worth pointing out is that this project appears to be a Godot module; whereas I think the rest of the bindings on the list are GDExtensions?

That said, that's probably fine, and I think it's still worth helping users find it. But should we include a note of some kind about this?

stan220:
Do you have any ideas on how to make it better? As for me, could leave it that way for now

skyace65:
@mhilbrunner are we still requiring Foundation review for linking stuff like this?

stan220:
> One thing that may be worth pointing out is that this project appears to be a Godot module; whereas I think the rest of the bindings on the list are GDExtensions?
> 
> That said, that's probably fine, and I think it's still worth helping users find it. But should we include a note of some kind about this?

What about to change from `Java/Kotlin` to `JVM Module`?

dsnopek:
I think `Java/Kotlin` is better than `JVM` (since it's a list of languages)

But to include the fact that it's a module, it could maybe be like:

```
- `Java/Kotlin <https://github.com/utopia-rise/godot-kotlin-jvm>`__ (Godot module)
```

I'm not sure, though. If we did that, maybe we should also add some other note explicitly stating that the others are GDExtensions? I'm very curious what other folks think, especially the docs maintainers

paddy-exe:
> ```
> - `Java/Kotlin <https://github.com/utopia-rise/godot-kotlin-jvm>`__ (Godot module)
> ```
> 
> I'm not sure, though. If we did that, maybe we should also add some other note explicitly stating that the others are GDExtensions? I'm very curious what other folks think, especially the docs maintainers

I would be in favor of stating a module as `(engine module)` and for GDExtensions as `(GDExtension)` "suffix". That they are specifically Godot modules should be clear from context. Now that we have more of a Godot ecosystem though I think we should be more concise with the language.

mhilbrunner:
I agree with @paddy-exe's suggestion above, could also do that in a followup PR however.

mhilbrunner:
In this goes, thank you for contributing!

PR #11483, opened 2025-11-21
Update recording_with_microphone.rst missing 'var'

toivocat:
Changed "effect = AudioServer.get_bus_effect(idx, 0)" to "var effect = AudioServer.get_bus_effect(idx, 0)" so it will work.


Calinou:
For future reference, this change isn't needed as `effect` is declared as a member variable above. If anything, it will cause an error where there was none previously (as GDScript doesn't allow redeclaring variables).

PR #11484, opened 2025-11-21
Add documentation on how to use Tracy and Perfetto

dsnopek:
This takes the existing `using_cpp_profilers.rst` and moves it to be about "Sampling profilers" (`sampling_profilers.rst`), so that a new page can be added about "Tracing profilers" (like Perfetto and Tracy!) and makes the old URL into an index that explains the difference and links to both pages.

~~This is marked as DRAFT for the moment, because it only covers Perfetto - I actually haven't had a chance try Tracy yet :-)~~


engine_details/development/debugging/profiling/tracing_profilers.rst
```diff
@@ -0,0 +1,115 @@
...
+your game that is exhibiting performance issues), you can use this script that
+comes with the Perfetto source code:
+
+.. code-block:: shell
+
+    cd perfetto
+    ./tools/record_android_trace -c godot.config
+
+This will record for 10 seconds (per the configuration), or until you press
+CTRL-C.
```
AThousandShips:
```suggestion
:kbd:`Ctrl + C`.
```

engine_details/development/debugging/profiling/tracing_profilers.rst
```diff
@@ -0,0 +1,115 @@
...
+To see the Godot events, expand the row for your application by clicking on its
+Android "Unique Name" (Perfetto will also include some events from system
+services in the trace).
+
+.. image:: img/cpp_profiler_perfetto.webp
+
+Then you can use the ``WASD`` keys to navigate the graph:
+
+- Press ``A`` or ``D`` to navigate forward or backward along the timeline
+- Press ``W`` or ``S`` to zoom in or out
```
AThousandShips:
```suggestion
- Press :kbd:`A` or :kbd:`D` to navigate forward or backward along the timeline
- Press :kbd:`W` or :kbd:`S` to zoom in or out
```

engine_details/development/debugging/profiling/tracing_profilers.rst
```diff
@@ -0,0 +1,115 @@
...
+    # Write to file every 1 second to prevent overflowing the buffer.
+    write_into_file: true
+    file_write_period_ms: 1000
+
+    # Track events in the "godot" category.
+    data_sources {
+        config {
+            name: "track_event"
+            track_event_config {
+            enabled_categories: "godot"
```
AThousandShips:
```suggestion
                enabled_categories: "godot"
```

engine_details/development/debugging/profiling/tracing_profilers.rst
```diff
@@ -0,0 +1,115 @@
...
+
+    # Trace for 10 seconds.
+    duration_ms: 10000
+
+    buffers {
+        size_kb: 32768
+        fill_policy: RING_BUFFER
+    }
+
+    # Write to file every 1 second to prevent overflowing the buffer.
```
AThousandShips:
```suggestion
    # Write to file every second to prevent overflowing the buffer.
```
Or
```suggestion
    # Write to file once every second to prevent overflowing the buffer.
```
IMO

engine_details/development/debugging/profiling/sampling_profilers.rst
```diff
@@ -0,0 +1,152 @@
...
+    after compiling the binaries. Otherwise, you will no longer get useful
+    profiling information when running a profiler.
+
+Benchmarking startup/shutdown times
+-----------------------------------
+
+If you're looking into optimizing Godot's startup/shutdown performance,
+you can tell the profiler to use the ``--quit`` command line option on the Godot binary.
+This will exit Godot just after it finished starting.
+The ``--quit`` option works with ``--editor``, ``--project-manager`` or
```
AThousandShips:
```suggestion
The ``--quit`` option works with ``--editor``, ``--project-manager``, and
```

engine_details/development/debugging/profiling/tracing_profilers.rst
dsnopek: Thanks! I've applied all your suggestions :-)

engine_details/development/debugging/profiling/sampling_profilers.rst
dsnopek: All the content of this file isn't new in this PR - I'm just moving it from `using_cpp_profilers.rst` to `sampling_profilers.rst`. However, I've applied your suggestion in any case :-)

dsnopek:
I've filled in the docs for Tracy now too, so taking this out of draft!

Given that I've only just tried Tracy for the first time, I'm very curious to hear if I'm using it correctly or not :-)

engine_details/development/debugging/profiling/tracing_profilers.rst
```diff
@@ -0,0 +1,221 @@
...
+
+In order to use either of them, you'll need to build the engine from source.
+If you've never done this before, please read
+:ref:`these docs <doc_compiling_index>` for the platform you want to profile on.
+You'll need to perform the same steps here, but with the addition of a single
+extra argument for ``scons``.
+
+.. _doc_tracy_profiler:
+
+Tracy for Windows, Linux and MacOS
```
AThousandShips:
```suggestion
Tracy for Windows, Linux, and MacOS
```

engine_details/development/debugging/profiling/tracing_profilers.rst
```diff
@@ -0,0 +1,221 @@
...
+----------------------------------
+
+Tracy is an Open Source profiler that runs on a wide variety of platforms,
+including Windows, Linux and MacOS. While it is primarily a tracing profiler,
```
AThousandShips:
```suggestion
including Windows, Linux, and MacOS. While it is primarily a tracing profiler,
```

engine_details/development/debugging/profiling/sampling_profilers.rst
```diff
@@ -0,0 +1,152 @@
...
+Xcode Instruments
+~~~~~~~~~~~~~~~~~
+
+- Open Xcode. Select **Open Developer Tool** - **Instruments** from the **Xcode** app menu:
+- Double-click on **Time Profiler** in the **Instruments** window:
+
+.. image:: img/cpp_profiler_xcode_menu.png
+
+- In the Time Profiler window, click on the **Target** menu, select **Choose target...**
+  and specify the path to the Godot binary, command line arguments and environment variables
```
AThousandShips:
```suggestion
  and specify the path to the Godot binary, command line arguments, and environment variables
```

engine_details/development/debugging/profiling/tracing_profilers.rst
```diff
@@ -0,0 +1,221 @@
...
+--------------------
+
+Perfetto is the default tracing system for Android. In fact, its system tracing
+service has been built into the platform since Android 9.
+
+Build Godot with Perfetto support
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+First, clone the latest version of the Perfetto source code ("53.0" at the
+time of this writing) using Git:
```
AThousandShips:
```suggestion
time of writing this) using Git:
```

engine_details/development/debugging/profiling/tracing_profilers.rst
```diff
@@ -0,0 +1,221 @@
...
+services in the trace).
+
+.. image:: img/cpp_profiler_perfetto.webp
+
+Then you can use the ``WASD`` keys to navigate the graph:
+
+- Press :kbd:`A` or :kbd:`D` to navigate forward or backward along the timeline
+- Press :kbd:`W` or :kbd:`S` to zoom in or out
+
+You'll probably need to zoom a bit before your able to see the individual
```
AThousandShips:
```suggestion
You'll probably need to zoom a bit before you're able to see the individual
```

engine_details/development/debugging/profiling/sampling_profilers.rst
```diff
@@ -0,0 +1,152 @@
...
+
+.. image:: img/cpp_profiler_hotspot_view_results.png
+
+- Use the tabs at the top to navigate between the different views. These views
+  show the same data, but in different ways. The **Flame Graph** tab is a good
+  way to see which functions take up the most time at a glance. These functions
+  are therefore the most important ones to optimize, since optimizing them will
+  improve performance the most.
+- At the bottom of all tabs except **Summary**, you will also see a list of CPU threads
+  started by the engine among with the CPU utilization for each thread.
```
AThousandShips:
```suggestion
  started by the engine along with the CPU utilization for each thread.
```

engine_details/development/debugging/profiling/sampling_profilers.rst
```diff
@@ -0,0 +1,152 @@
...
+  While not essential for profiling Godot, this will ensure all events can be captured.
+  Otherwise, some events may be missing in the capture.
+  Your settings should now look something like this:
+
+.. image:: img/cpp_profiler_hotspot_record.png
+
+- Click **Start Recording** and perform the actions you wish to profile in the editor/project.
+- Quit the editor/project normally or use the **Stop Profiling** button in HotSpot
+  to stop profiling early. Stopping profiling early can result in cleaner profiles
+  if you're not interested in the engine's quit procedure.
```
AThousandShips:
```suggestion
  if you're not interested in the engine's shutdown procedure.
```
I'd say

engine_details/development/debugging/profiling/sampling_profilers.rst
```diff
@@ -0,0 +1,152 @@
...
+.. image:: img/cpp_profiler_hotspot_welcome.png
+
+- In the next window, specify the path to the Godot binary that includes debug symbols.
+- Specify command line arguments to run a specific project, with or without the editor.
+- The path to the working directory can be anything if an absolute path is used
+  for the ``--path`` command line argument. Otherwise, it must be set to that
+  the relative path to the project is valid.
+- Make sure **Elevate Privileges** is checked if you have administrative privileges.
+  While not essential for profiling Godot, this will ensure all events can be captured.
+  Otherwise, some events may be missing in the capture.
```
AThousandShips:
```suggestion
  Otherwise, some events may be missing from the capture.
```

engine_details/development/debugging/profiling/sampling_profilers.rst
```diff
@@ -0,0 +1,152 @@
...
+~~~~~~~
+
+- Open HotSpot. Click **Record Data**:
+
+.. image:: img/cpp_profiler_hotspot_welcome.png
+
+- In the next window, specify the path to the Godot binary that includes debug symbols.
+- Specify command line arguments to run a specific project, with or without the editor.
+- The path to the working directory can be anything if an absolute path is used
+  for the ``--path`` command line argument. Otherwise, it must be set to that
```
AThousandShips:
```suggestion
  for the ``--path`` command line argument. Otherwise, it must be set so that
```

engine_details/development/debugging/profiling/sampling_profilers.rst
```diff
@@ -0,0 +1,152 @@
...
+    Do *not* strip debugging symbols on the binaries using the ``strip`` command
+    after compiling the binaries. Otherwise, you will no longer get useful
+    profiling information when running a profiler.
+
+Benchmarking startup/shutdown times
+-----------------------------------
+
+If you're looking into optimizing Godot's startup/shutdown performance,
+you can tell the profiler to use the ``--quit`` command line option on the Godot binary.
+This will exit Godot just after it finished starting.
```
AThousandShips:
```suggestion
This will exit Godot just after it's done starting.
```

engine_details/development/debugging/using_cpp_profilers.rst
```diff
@@ -12,150 +12,31 @@ To this end, profilers are useful tools.
...
-Recommended profilers
----------------------
+There are two main types of profilers: sampling profilers and tracing profilers.
 
-- `VerySleepy <http://www.codersnotes.com/sleepy/>`__ (Windows only)
-- `HotSpot <https://github.com/KDAB/hotspot>`__ (Linux only)
-- `Xcode Instruments <https://developer.apple.com/xcode/>`__ (macOS only)
+Sampling profilers periodically interrupt the running program and take a "sample",
+which records which functions are running. The profiler can aggregate this data to
+produce a graph which shows which functions the program spent the most time in.
```
Ivorforce: In XCode, it's not really a graph, but a tree (well, mainly anyway). Either way, what's important is that it's just a statistical estimation, I'd say.

```suggestion
which records which functions are running. Using this information, the profiler estimates
which functions the program spent the most time in.
```

engine_details/development/debugging/using_cpp_profilers.rst
```diff
@@ -12,150 +12,31 @@ To this end, profilers are useful tools.
...
 
-These profilers may not be the most powerful or flexible options, but their
-standalone operation and limited feature set tends to make them easier to use.
+Tracing profilers work by recording application-specific events over a short
+period of time (usually only a few seconds), producing a log called a "trace".
+The profiler can use the trace to produce a graph showing a high-level timeline of
+what happened, when and how long it took.
```
Ivorforce: To me, the most important features of tracing profilers are that they're accurate, and that they require manual setup.

I've also recorded for plenty longer than a few seconds with tracy, without performance issues. I wouldn't be surprised if it supported tracing hours at a time 😅

```suggestion
Tracing profilers work by recording application-specific events (such as the start and end of a single frame), producing a log called a "trace".
The profiler can use the trace to produce a graph showing an accurate high-level timeline of
what happened. However, any code that is not explicitly instrumented will not appear in a tracing
profiler's timeline!
```

engine_details/development/debugging/using_cpp_profilers.rst
```diff
@@ -12,150 +12,31 @@ To this end, profilers are useful tools.
...
 
-Setting up Godot
-----------------
+While a sampling profiler doesn't require any code changes (only a build with
+debug symbols), in order to use a tracing profiler, there needs to be additional
+logging code added.
```
Ivorforce: Since Godot supports both, and needs to be compiled from scratch for both, I don't think this information is needed.

```suggestion
```

engine_details/development/debugging/profiling/tracing_profilers.rst
```diff
@@ -0,0 +1,221 @@
+.. _doc_tracing_profilers:
+
+Tracing Profilers
+=================
+
+Godot currently supports two tracing profilers: Tracy and Perfetto.
```
Ivorforce: Alternatively, we could link to the related headings here?

```suggestion
Godot currently supports two tracing profilers: `Tracy <https://github.com/wolfpld/tracy>`__ and `Perfetto <https://perfetto.dev>`__.
```

engine_details/development/debugging/profiling/tracing_profilers.rst
Ivorforce: Careful, or Apple's marketing team will come after us! :P

```suggestion
including Windows, Linux and macOS. While it is primarily a tracing profiler,
```

engine_details/development/debugging/profiling/tracing_profilers.rst
```diff
@@ -0,0 +1,221 @@
...
+Get the Tracy "server"
+~~~~~~~~~~~~~~~~~~~~~~
+
+In Tracy terminology, the application you are profiling is the "client", and
+the one receiving the data is the "server".
+
+If you are on Windows, you can download a pre-built ``tracy-profiler.exe``
+from the Tracy `releases page <https://github.com/wolfpld/tracy/releases>`_.
+
+However, if you're on Linux or MacOS, you'll need to build it yourself.
```
Ivorforce: I actually had trouble building tracy at first, but later found out both `brew` and `nix` supply pre-built tracy binaries. I would not recommend building it yourself by default.

```suggestion
If you're on Linux or macOS, you may be able to get a Tracy binary from a package manager.
For example, on macOS, you can install Tracy using `homebrew <https://brew.sh>`__:

.. code-block:: shell

    brew install tracy

Alternatively, you can build the Tracy server yourself, as explained below.
```

engine_details/development/debugging/profiling/tracing_profilers.rst
```diff
@@ -0,0 +1,221 @@
...
+First, export your game using the release templates you built above, and run it.
+
+Then, launch the Tracy server - you'll see something like this:
+
+.. image:: img/cpp_profiler_tracy_start.webp
+
+When you're ready to record a trace (for example, when you've hit the part of
+your game that is exhibiting performance issues), press the "Connect" button.
+
+You'll immediately see tracing data start coming in:
```
Ivorforce:
```suggestion
Then, launch the Tracy server - you'll see something like this:

.. image:: img/cpp_profiler_tracy_start.webp

If you press "Connect" now, you can profile the launch of Godot. However,
if you want to profile something specific (for example, a part of your game
that is exhibiting performance issues), wait before pressing the button.

Now, export your game using the release templates you built above, and run it.
As soon as the both the game are running, and you have pressed the "Connect"
button in Tracy, you'll see data coming in:
```

engine_details/development/debugging/profiling/tracing_profilers.rst
AThousandShips:
```suggestion
You'll probably need to zoom a bit before you're able to see the individual
```

engine_details/development/debugging/profiling/tracing_profilers.rst
AThousandShips: Actually 
```suggestion
including Windows, Linux, and macOS. While it is primarily a tracing profiler,
```

engine_details/development/debugging/profiling/tracing_profilers.rst
AThousandShips:
```suggestion
Tracy for Windows, Linux, and macOS
```

engine_details/development/debugging/profiling/tracing_profilers.rst
```diff
@@ -0,0 +1,221 @@
...
+including Windows, Linux and MacOS. While it is primarily a tracing profiler,
+it can also periodically sample data like a
+:ref:`sampling profiler <doc_sampling_profilers>`, giving some of the benefits
+of both approaches.
+
+Build Godot with Tracy support
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+First, clone the latest version of the Tracy source code ("0.13.0" at the
+time of this writing) using Git:
```
AThousandShips:
```suggestion
time of writing this) using Git:
```

engine_details/development/debugging/profiling/tracing_profilers.rst
AThousandShips:
```suggestion
time of writing this) using Git:
```

engine_details/development/debugging/profiling/sampling_profilers.rst
AThousandShips:
```suggestion
  and specify the path to the Godot binary, command line arguments, and environment variables
```

engine_details/development/debugging/profiling/sampling_profilers.rst
AThousandShips:
```suggestion
  started by the engine along with the CPU utilization for each thread.
```

engine_details/development/debugging/profiling/sampling_profilers.rst
AThousandShips:
```suggestion
  if you're not interested in the engine's shutdown procedure.
```
I'd say

engine_details/development/debugging/profiling/sampling_profilers.rst
AThousandShips:
```suggestion
  Otherwise, some events may be missing from the capture.
```

engine_details/development/debugging/profiling/sampling_profilers.rst
AThousandShips:
```suggestion
  for the ``--path`` command line argument. Otherwise, it must be set so that
```

engine_details/development/debugging/profiling/sampling_profilers.rst
AThousandShips:
```suggestion
This will exit Godot just after it's done starting.
```
I'd say

engine_details/development/debugging/profiling/tracing_profilers.rst
bruvzg: It won't work if installed `tracy` version is not matching the client (or at least protocol version). Current `brew` version (0.11.1) doesn't work with current GitHub release client (0.13).

engine_details/development/debugging/using_cpp_profilers.rst
dsnopek: I've accepted your suggestion (thanks!), however...

> I've also recorded for plenty longer than a few seconds with tracy, without performance issues. I wouldn't be surprised if it supported tracing hours at a time 😅

We've barely added any events yet, though. Once something like https://github.com/godotengine/godot/pull/112707 is merged, this is going to change. In [their Godotfest presentation](https://www.youtube.com/watch?v=3BIteSVcO28), one of the "Halls of Torment" devs described how they implemented something similar, and could only run it for a few seconds because otherwise it would generate gigabytes of data.

Ivorforce: I suppose we'll see once we merge https://github.com/godotengine/godot/pull/112707, but I think the interning trick might save us from that problem.
(I watched the talk too and had a suspicion they might not have thought of the same kind of trick)

engine_details/development/debugging/profiling/tracing_profilers.rst
dsnopek: "at the time of this writing" is a set phrase, although, I've also seen "at the time of writing" (with "this" removed) which I guess could be a good compromise :-)

engine_details/development/debugging/profiling/tracing_profilers.rst
dsnopek: Thanks!

In my latest update, I've mentioned that you might be able to find a pre-built binary from a package manager, but that if you do, you should be careful to match the version

engine_details/development/debugging/profiling/sampling_profilers.rst
dsnopek: As I've mentioned previously, the `sampling_profiler.rst` wasn't written in this PR, only moved. That said, your suggestions for it look good to me, so I've accepted them all :-)

dsnopek:
@Ivorforce @AThousandShips Thanks for the review! I think I've integrated all your feedback in my latest push :-)

Ivorforce (approved):
Looks great!

engine_details/development/debugging/using_cpp_profilers.rst
m4gr3d, line 3: Should `C++` be dropped from the page title and page path since both sampling and tracing profilers can be used for more than C++ applications and the Godot codebase contains other languages besides C++.

engine_details/development/debugging/profiling/tracing_profilers.rst
```diff
@@ -0,0 +1,227 @@
...
+Here are some of the basic controls:
+
+- Zoom in/out with the mouse wheel
+- Right click and drag to move forward/backward on the timeline
+- In the top bar, click the left and right arrow buttons by "Frames" to move a single frame on the timeline
+
+To learn more, see the
+`Tracy manual <https://github.com/wolfpld/tracy/releases/latest/download/tracy.pdf>`_.
+
+Perfetto for Android
```
m4gr3d, line 128: Worth including that it supports Win/Mac/Linux as well.

engine_details/development/debugging/profiling/tracing_profilers.rst
```diff
@@ -0,0 +1,227 @@
...
+First, clone the latest version of the Perfetto source code ("53.0" at the
+time of writing) using Git:
+
+.. code-block:: shell
+
+    git clone -b v53.0 --single-branch https://github.com/google/perfetto.git
+
+This will create a ``perfetto`` directory - you can place this anywhere.
+
+Next, build the Android debug or release templates for your architecture using
```
m4gr3d, line 146: Are debug symbols not required for Perfetto?

engine_details/development/debugging/using_cpp_profilers.rst
dsnopek: I believe the "C++" is in there to distinguish it from Godot's profiler?

engine_details/development/debugging/profiling/tracing_profilers.rst
dsnopek: I don't think the way Godot has integrated it (at least so far) really supports Windows and Mac. It's setup to use the "system" backend, which requires connecting with Perfetto's system daemon which I think is only Linux and Android. If we added support for the "in process" backend we could add docs for how to use that later? The steps for using it would be somewhat different

engine_details/development/debugging/profiling/tracing_profilers.rst
dsnopek: Nope! Debug symbols are not required for Perfetto, or Tracy either, if you just use its tracing stuff and not its sampling features

engine_details/development/debugging/profiling/sampling_profilers.rst
```diff
@@ -0,0 +1,152 @@
...
+
+.. image:: img/cpp_profiler_time_profiler_record.png
+
+- Perform the actions you wish to profile in the editor or project. When you're done,
+  click the **Stop** button.
+
+- Wait for the results to appear.
+- At the bottom of the window you will see a call tree for all CPU threads started, and
+  the **Heaviest Stack Trace** overview.
+- Select **Hide system libraries** in the **Call Tree** menu (at the bottom of window) to
```
AThousandShips:
```suggestion
- Select **Hide system libraries** in the **Call Tree** menu (at the bottom of the window) to
```

dsnopek: Thanks! Updated :-)

engine_details/development/debugging/using_cpp_profilers.rst
m4gr3d: I think the section is explicit enough that we could drop `C++`. For example, in the same section we have a `using sanitizers` page.

dsnopek: I think sanitizers is a little different, though: there isn't a "Godot sanitizer" and "C++ sanitizers", whereas we do have a "Godot profiler" and "C++ profilers"

However, if we do rename it, I assume we'd want to keep the path the same so we don't have to deal with redirects?

m4gr3d: > However, if we do rename it, I assume we'd want to keep the path the same so we don't have to deal with redirects?

yeah, keep things simpler.

dsnopek: Given that the page title isn't from this PR (it's pre-existing), I don't think we need to resolve this here. The title can be addressed in another PR

Ryan-000:
When I previously used a Tracy integration, GDExtensions could not reload. (At least on Linux, I'm not sure about Windows). 

If this is still true, it may be worth mentioning.

Ivorforce:
Did you use the [Tracy Gdextension](https://github.com/AndreaCatania/godot_tracy)?

Ryan-000:
> Did you use the [Tracy Gdextension](https://github.com/AndreaCatania/godot_tracy)?

Yes https://github.com/Ryan-000/godot_tracy

Ivorforce:
Aha, that should explain it.
The tracy GDExtension is, well, a GDExtension. It's possible that the hot reload didn't work well with how tracy's / the extension's assumptions.
Godot's integration is in core, and does not use the GDExtension system. As such, I don't think it will affect whether GDExtensions can reload or not. But thanks for reporting, we'll keep an eye on it!

Ryan-000:
Oh my bad, I misinterpreted your question. I actually had it compiled as a custom module, not as GDExtension.

engine_details/development/debugging/profiling/tracing_profilers.rst
```diff
@@ -0,0 +1,227 @@
...
+.. code-block:: shell
+
+    # On Linux, Tracy uses Wayland by default, so if you use X11 add -DLEGACY=1
+    cmake -B profiler/build -S profiler -DCMAKE_BUILD_TYPE=Release
+    cmake --build profiler/build --config Release --parallel
+
+Record a trace
+~~~~~~~~~~~~~~
+
+Launch the Tracy server - you'll see something like this:
```
AdriaandeJongh, line 97: This 'just launch the server!' is a bit too casual for users like me. I went down a rabbit hole just to discover that the previous cmake compile steps place a binary at `tracy/profiler/build/tracy-profiler` which is the thing you'll want to launch. Might be good to add that here somehow.

engine_details/development/debugging/profiling/tracing_profilers.rst
```diff
@@ -0,0 +1,227 @@
...
+
+    If you do use a pre-built binary, be sure to use the same version that
+    you used when building Godot.
+
+Build the Tracy server from source
+++++++++++++++++++++++++++++++++++
+
+The full details can be found in the
+`Tracy manual <https://github.com/wolfpld/tracy/releases/latest/download/tracy.pdf>`_,
+but here is the TL;DR:
```
AdriaandeJongh, line 83: The TL;DR should IMO also contain that you need to install CMake. The easiest way to do that is using a Homebrew command, eg. `brew install cmake`.

engine_details/development/debugging/profiling/tracing_profilers.rst
```diff
@@ -0,0 +1,227 @@
...
+Launch the Tracy server - you'll see something like this:
+
+.. image:: img/cpp_profiler_tracy_start.webp
+
+If you press "Connect" now, you can profile the launch of Godot. However,
+if you want to profile something specific (for example, a part of your game
+that is exhibiting performance issues), wait before pressing the button.
+
+Now, export your game using the release templates you built above, and run it.
+As soon as both are running, and you have pressed the "Connect" button in
```
AdriaandeJongh, line 107: I did all the steps precisely as explained, and yet tracy gets stuck in 'Waiting for connection' after pressing the connect button. Perhaps some basic info on troubleshooting is in order here.

engine_details/development/debugging/profiling/tracing_profilers.rst
```diff
@@ -0,0 +1,227 @@
...
+.. code-block:: shell
+
+    git clone -b v0.13.0 --single-branch https://github.com/wolfpld/tracy.git
+
+This will create a ``tracy`` directory - you can place this anywhere.
+
+Next, build the release templates for your platform using ``scons``, but adding
+the ``profiler_path=path/to/tracy`` argument with the real path to the ``tracy``
+directory, as well as ``debug_symbols=yes`` to allow Tracy's sampling features
+to work.
```
AdriaandeJongh: There is another required argument that is missing from this guide: `profiler=tracy`. Might also have to be set for the other profilers (I don't know) but for this one I confirmed it to be the reason I wasn't able to get tracy to work.

engine_details/development/debugging/profiling/tracing_profilers.rst
```diff
@@ -0,0 +1,227 @@
...
+Record a trace
+~~~~~~~~~~~~~~
+
+Launch the Tracy server - you'll see something like this:
+
+.. image:: img/cpp_profiler_tracy_start.webp
+
+If you press "Connect" now, you can profile the launch of Godot. However,
+if you want to profile something specific (for example, a part of your game
+that is exhibiting performance issues), wait before pressing the button.
```
Ivorforce: I just learned from a comment on the PR (+ the manual) that by default, tracy just stashes the profiler results until a server connection is made. So starting the server later (instead of immediately) just leads to RAM issues.
For now, I'd say we can change the formulation to ask users to profile immediately to avoid issues. In the future, we may want to add support for `TRACY_ON_DEMAND` (which makes profiling slower but supports launching the server later without RAM bloat).

Maybe something like this?

```suggestion
Press "connect". This will ensure tracy makes a connection immediately when the game launches. If you forget to press "connect", tracy will store system events in RAM, which can quickly blow up your memory usage (see ``TRACY_ON_DEMAND`` documentation).
```

engine_details/development/debugging/profiling/tracing_profilers.rst
Ivorforce: Yea, this argument was only added recently. Other profilers also need it (`profiler=perfetto`, as well as `profiler=instruments`, although I think instruments should be documented in a follow-up PR by Stuart).

engine_details/development/debugging/profiling/tracing_profilers.rst
AdriaandeJongh: Perhaps a bit of a checklist:
- make sure you compiled the export templates with the right scons arguments
- make sure you export your game using the custom export template
- ... ?

or maybe this is too convoluted, idk.

AdriaandeJongh:
To make line numbers show up correctly for tracy, **these two checkboxes in the player settings need to be checked**:
Screenshot 2025-12-08 at 12 51 00

This should be in the docs, too.

engine_details/development/debugging/profiling/tracing_profilers.rst
dsnopek: Hrm, I wasn't aware of this change. I'm not sure I like it - having to specify two arguments instead of one seems less friendly :-/

I'll update the docs, though

dsnopek:
> To make line numbers show up correctly for tracy, **these two checkboxes in the player settings need to be checked**:

Let's save this bit for a follow-up. I'd really like to get the simplest "getting started" steps merged in this PR, and then it can be built upon

AdriaandeJongh:
> Let's save this bit for a follow-up. I'd really like to get the simplest "getting started" steps merged in this PR, and then it can be built upon.

I tend to agree with this approach, but it’s very easy to add now, and without it, it will lead to issue reports. it makes the GDScript tracy implementation seem broken.

engine_details/development/debugging/profiling/tracing_profilers.rst
dsnopek: Perhaps we could add a "Troubleshooting" section later? I don't think we should get into too much detail in the middle of the steps

engine_details/development/debugging/profiling/tracing_profilers.rst
dsnopek: Thanks! In my latest push, I've added a note about this in the section about building Tracy from source

engine_details/development/debugging/profiling/tracing_profilers.rst
dsnopek: Thanks! Updated in my latest push

engine_details/development/debugging/profiling/tracing_profilers.rst
dsnopek: Thanks! I've accepted your suggestion in my latest push

engine_details/development/debugging/profiling/tracing_profilers.rst
dsnopek: Thanks! I've added a note about this (similar to the one above about installing Tracy itself via a package manager)

engine_details/development/debugging/profiling/tracing_profilers.rst
AdriaandeJongh: agreed

engine_details/development/debugging/profiling/tracing_profilers.rst
```diff
@@ -0,0 +1,235 @@
...
+
+Build Godot with Tracy support
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+First, clone the latest version of the Tracy source code ("0.13.0" at the
+time of writing) using Git:
+
+.. code-block:: shell
+
+    git clone -b v0.13.0 --single-branch https://github.com/wolfpld/tracy.git
```
akien-mga, line 34: Tracy has relatively frequent releases so this will get out of date quickly, and we don't want to have to update it regularly (which would add churn and also invalidate translations), so I'd suggest looking for an alternative that doesn't hardcode a version number.

You can use https://github.com/wolfpld/tracy/releases/latest to link to the latest tag, but that's not usable for the git clone command.

akien-mga:
Thanks!

engine_details/development/debugging/profiling/tracing_profilers.rst
Ivorforce: > Hrm, I wasn't aware of this change. I'm not sure I like it - having to specify two arguments instead of one seems less friendly :-/

I voiced the same reservation, but was outvoted :D

PR #11489, opened 2025-11-22
Add some information about relocation safety.

Ivorforce:
This is a decision we made in a core meeting some months ago (although trivial relocatability was assumed long before the explicit decision). For more information, see https://github.com/godotengine/godot/issues/100509.

This information is probably rarely useful, since most users won't know what trivial relocatability is, unless they run into issues. Still, documenting it might avoid some obscure bugs down the line.


mhilbrunner:
Thank you!

PR #11494, opened 2025-11-25
Update rotation + movement section of 2d_movement.rst to properly app…

Hjaldar:
…ly the movement in the direction the sprite is facing

The code in the rotation + movement section was applying movement perpendicular to the up direction instead of in the direction the sprite is facing. This can be clearly seen in the animation.


Hjaldar:
Sorry, first attempt at a pull request and made an error. Closing this. Will open an issue so someone who knows what they are doing can fix it properly

PR #11498, opened 2025-11-27
Added missing word to class_astar2d.rst

artieleach:
A word was skipped that confused me. "Called when neighboring enters processing" becomes "Called when neighboring **cell** enters processing".


AThousandShips:

Issue #11497, opened 2025-11-26
Camera3D.set_frustum() description should say the size parameter is the size of the near plane


partybusiness:
**Issue description:**

Current description says "Sets the camera projection to frustum mode (see [PROJECTION_FRUSTUM](https://docs.godotengine.org/en/stable/classes/class_camera3d.html#class-camera3d-constant-projection-frustum)), by **specifying a size,** an offset, and the z_near and z_far clip planes in world space units."

But it doesn't really say what we are specifying the size of. I saw someone very confused on how to use this function because they didn't know. As I understand it, it's the size of the near plane. So maybe we can add a line to clarify:

"The size parameter is the size of the near plane, either the width or height, according to the value of [keep_aspect](https://docs.godotengine.org/en/stable/classes/class_camera3d.html#class-camera3d-property-keep-aspect)."

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/stable/classes/class_camera3d.html#class-camera3d-method-set-frustum

Arjun-singh-rawatt:
Hii, I’d like to work on this.

PR #11499, opened 2025-11-27
Added note about size parameter to Camera3D set_frustum

partybusiness:
The description for the Camera3D.set_frustum() function did not explain what the _size_ parameter. I added a note to explain it's the size of the camera's near plane.

Addressing this issue:

https://github.com/godotengine/godot-docs/issues/11497


AThousandShips:

PR #11501, opened 2025-11-27
GDScript: Clarify the use of `callv()` to call variadic functions

dalexeev:
* Closes https://github.com/godotengine/godot/issues/112832.

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_basics.rst
+++ b/tutorials/scripting/gdscript/gdscript_basics.rst
@@ -1651,12 +1651,19 @@ as a static type of the rest parameter:
 
     ::
 
-        func log_data(...values):
-            # ...
-
-        func other_func(...args):
+        func test_func(...args):
             #log_data(...args) # This won't work.
             log_data.callv(args) # This will work.
+
+        func log_data(...values):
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):
Content as is looks good to me. Should this really be in the comment instead of an explanation outside the code block, thinking about translations and such?

MikeSchulze:
What's about constructors using variadic arguments?
```
func foo(...args: Array) -> Foo:
	return Foo.new(args)
	
class Foo:
   func _init(...args: Array) -> void:
	_values = args	
```

I use this hack to workaround
```
func _init(...args: Array) -> void:
	if args.size() == 1 and args[0] is Array:
		_values = args[0]
		return
	_values = args
```

It would be great instead to write special documentation to support variadic argument delegation over function borders.

dalexeev:
> What's about constructors using variadic arguments?

You can use `callv()` as well:

```gdscript
class Test:
    func _init(...args: Array) -> void:
        print(args)

func test(...args: Array) -> void:
    Test.new.callv(args)

func _run() -> void:
    test()
    test(1)
    test(1, 2)
    test(1, 2, 3)
```

mhilbrunner:
Thanks!

PR #11502, opened 2025-11-27
Update Compiling for Windows following Direct3D 12 being enabled by default

Calinou:
- Follow-up to https://github.com/godotengine/godot/pull/113213.

The recommendation is now to leave `d3d12` to its default `yes` value, which means installing Direct3D 12 dependencies is considered required by default.

You can still skip the dependencies by compiling with `d3d12=no`, which matches the behavior of self-compiled builds prior to Godot 4.6.

This also removes an outdated FAQ question, since a lot of time has passed since that question was written (Godot only had Vulkan and OpenGL support back then).

*Not cherry-pickable to 4.5.*


skyace65 (approved):

mhilbrunner:
Thank you!

PR #11507, opened 2025-12-02
Emphasize script classes over custom types in Making plugins

Calinou:
Custom types were more relevant in the early Godot 3.x days, but nowadays, script classes should be preferred whenever possible (i.e. in any non-C# language).

- See https://github.com/godotengine/godot-docs-user-notes/discussions/137#discussioncomment-15114992.


mhilbrunner:
Thanks!

PR #11508, opened 2025-12-02
Add a reminder to load the CSV as a source in Localization using spreadsheets

Calinou:
A similar reminder is already present in the *Localization using gettext* page.

- See https://github.com/godotengine/godot-docs-user-notes/discussions/579#discussioncomment-15117997.


mhilbrunner:
Thanks!

PR #11510, opened 2025-12-02
Fixed code example in Shaders > Advanced post-processing > An optimization

Aurora-Bee:
If you tried to follow this section of the tutorial now, it wouldn't result in the desired outcome. The post-processing shader wouldn't be applied.

In the code example of that section the vertices are arranged clockwise, which creates a mesh that faces the camera. However, the same article also states, that for QuadMeshes you need to flip the faces by enabling "Flip Faces", so it turns away from the camera. To fix this I rearranged some of code so the vertices are arranged counter-clockwise, which flips the normals around too and resolves the issue.

Note: My formatter also managed to catch and delete some leading whitespaces, which seems to align with the rest of the document's format. If this isn't wished for I can revert those and only leave in the code changes.

Hope this will help some people :)


clayjohn (approved):

mhilbrunner:

akien-mga:
Cherry-picked for `4.5`.

PR #11511, opened 2025-12-03
Fix incorrect claim about visibility range

ttencate:
As mentioned on
https://github.com/godotengine/godot/issues/102799#issuecomment-3603374466, the docs are wrong here: visibility range is not computed to the instance's origin but to its AABB center.


skyace65:
Thanks!

akien-mga:
Cherry-picked for `4.5`.

PR #11516, opened 2025-12-05
godot-cpp: Explain that `ptr()`/`ptrw()` should be used to efficiently access `Packed*Array`s

dsnopek:
It's frequently surprising to folks that accessing individual elements on a `Packed*Array` can be inefficient from godot-cpp.

This attempts to explain, and provide the advice to use `ptr()` or `ptrw()` instead!

### DIFF
--- a/tutorials/scripting/cpp/core_types.rst
+++ b/tutorials/scripting/cpp/core_types.rst
@@ -30,11 +30,48 @@ Packed arrays
 ~~~~~~~~~~~~~
 
 While in Godot, the ``Packed*Array`` types are aliases of ``Vector``, in godot-cpp, they're their own types, using the
-Godot bindings. This is because these types, along with their methods, are exposed through the GDExtension interface,
-which would not be compatible with the templated nature of ``Vector``.
+Godot bindings. This is because ``Packed*Array`` are exposed to Godot and limited to only Godot types, whereas ``Vector``
+can hold any C++ type which Godot might not be able to understand.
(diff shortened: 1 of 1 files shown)


Ivorforce (approved):
Looks good to me, and i agree this is something Godot-cpp users should be aware of. The same kind of access is also slow in Godot, but it's especially slow in godot-cpp.

tutorials/scripting/cpp/core_types.rst
```diff
@@ -30,11 +30,48 @@ Packed arrays
...
+In general, the ``Packed*Array`` types work the same way as their ``Vector`` aliases, however, there are some notable
+differences.
+
+Data access
++++++++++++
+
+``Vector`` keeps its data entirely within the GDExtension, whereas the ``Packed*Array`` types keep their data on the
+Godot side. This means that any time a ``Packed*Array`` is accessed, it needs to call into Godot.
+
+To efficiently read or write a large amount of data into a ``Packed*Array``, you should called ``.ptr()`` (for reading)
```
mhilbrunner:
```suggestion
To efficiently read or write a large amount of data into a ``Packed*Array``, you should call ``.ptr()`` (for reading)
```

dsnopek: Oops! Fixed, thanks :-)

mhilbrunner:
Merged. Thanks a bunch :)

Issue #4945, opened 2021-05-29
Outdated or incorrect binary serialization API


Aurora2500:
**Your Godot version:**
3.3 Stable

**Issue description:**
In the documentation, it is stated that
```
All packets have a 4-byte header representing an integer, specifying the type of data.

...

Following this is the actual packet contents, which varies for each type of packet. 
```

Because of this, I'd expect 4 bytes being taken up as a header, indicating which type it is and some other flag information. And then following that I'd get the data for the type, if any.

So when doing `store_var(null)`, I'd expect there to be 4 bytes all filled with 0:
```
00000000 00000000 00000000 00000000
```

Or when doing `store_var(42)`, I'd expect to get a file with the following binary data:
```
00000010 00000000 00000000 00000000
00101010 00000000 00000000 00000000
```

But instead, there seems to be an extra 4 bytes in front of the header for each data packet.
When I store `null`, I get:
```
00000100 00000000 00000000 00000000
00000000 00000000 00000000 00000000
```
Or when I store `42`:
```
00001000 00000000 00000000 00000000
00000010 00000000 00000000 00000000
00101010 00000000 00000000 00000000
```

The documentation does not indicate what these extra bytes mean, or mention that they should exist at all.

**URL to the documentation page:**
https://docs.godotengine.org/en/stable/tutorials/misc/binary_serialization_api.html

Calinou:
See also https://github.com/godotengine/godot-docs/issues/4887 and https://github.com/godotengine/godot-docs/pull/4893.

BLNJ:
This problem is much bigger, there are actually 38 different packet types which you only discover when directly looking at the source code (for instance Vector2I, Rect2I, Vector3I and so on).
It is also said that all packets have a 4 byte header, but nowhere is mentioned that packets have a size in the header, while packets inside others (like in dictionary) dont have that size component in the header.

KryziK:
Hey all,

Full disclosure, I had Claude help me analyze marshalls.h and marshalls.cpp to produce this. If this isn't allowed, please just inform me rather than berate or attack me. This is a learning experience for me.

I cross-checked it with a save file for a Godot game that I have and it seemed to match. I used ImHex and a custom pattern and the structures it produced looked good to me.

Here's a rough first draft of the new RST documentation. I haven't had a chance to figure out how to "build" it locally so I can preview it in the same format and theme of the Godot website.

This is for `godot-4.4.1-stable`.
[...]

PR #11517, opened 2025-12-06
Update binary serialization documentation

KryziK:
Update binary serialization documentation from Godot version < 4 to 4.5.1.

Disclosure: AI was used as a tool to help me quickly parse the marshalls.h and marshalls.cpp file into the full list of data structures now available in Godot 4.5.

AI was also helpful in answering questions I had about specific data structures while I wrote the new tables and documentation.

One such question was to understand why the existing docs state that a Variant::OBJECT can have 3 different serializations and not just 2, to which the reply was that a NULL object is technically just variant data type 0 (NIL) rather than an object type itself.

Through this back-and-forth Q&A session, I was able to better understand how to write some small nuances and notes into the documentation.

~~

It was definitely a small challenge learning how to use RST, so if there are any formatting or consistency issues I need to fix, please let me know.

I did manually write the final file, mainly leveraging copy/paste of tables from the old documentation to understand the RST format. I did combine this with some manually-reviewed notes based on answers I got from Claude. No part of this document was output directly by Claude without manual review and clean-up to try my best at keeping the changes high-quality.

Thank you for your time.


KryziK:
#4945

AThousandShips:
> No part of this document was output directly by Claude without manual review and clean-up

So parts of this was generated by Claude? 

Edit: reading this I can definitely tell this was written by an LLM, it is not particularly well worded and uses a lot of incorrect terminology and has a lot of grammatical and formatting errors, please go through it again and comment when you've fixed them and I can review after, I am not interested in reviewing it as it is right now

tutorials/io/binary_serialization_api.rst
```diff
@@ -167,19 +222,21 @@ a 64-bit double precision number:
...
-| Offset   | Len   | Type      | Description                |
-+==========+=======+===========+============================+
-| 4        | 4     | Integer   | String length (in bytes)   |
-+----------+-------+-----------+----------------------------+
-| 8        | X     | Bytes     | UTF-8 encoded string       |
-+----------+-------+-----------+----------------------------+
++----------+-------+-----------+------------------------------+
+| Offset   | Len   | Type      | Description                  |
++==========+=======+===========+==============================+
+| 4        | 4     | Integer   | String length (in byte, N)   |
```
AThousandShips:
```suggestion
| 4        | 4     | Integer   | String length (in bytes, N)  |
```

tutorials/io/binary_serialization_api.rst
```diff
@@ -265,17 +457,21 @@ This field is padded to 4 bytes.
 | 16       | 4     | Float   | Real W        |
 +----------+-------+---------+---------------+
 
-11: :ref:`AABB<class_aabb>`
+If flag ``ENCODE_FLAG_64`` is set (flags & 1 == 1), the floats are sent as 64 bit double precision.
```
AThousandShips: At minimum this should as it's the same size as `Vector4`

tutorials/io/binary_serialization_api.rst
```diff
@@ -355,137 +605,207 @@ This field is padded to 4 bytes.
...
+
+27: :ref:`Dictionary<class_dictionary>`
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-+----------+-------+-----------+---------------------------------------------------------------------+
-| Offset   | Len   | Type      | Description                                                         |
-+==========+=======+===========+=====================================================================+
-| 4        | 4     | Integer   | val&0x7FFFFFFF = elements, val&0x80000000 = shared (bool)           |
-+----------+-------+-----------+---------------------------------------------------------------------+
+For typed dictionaries, the header flags indicate the type kind for both keys
```
AThousandShips:
```suggestion
For typed dictionaries, the header flags indicate the type for both keys
```
"Type kind" isn't a terminology used

tutorials/io/binary_serialization_api.rst
```diff
@@ -355,137 +605,207 @@ This field is padded to 4 bytes.
...
+and values (see Header Flags section). Depending on these flags, type
```
AThousandShips:
```suggestion
and values (see the Header Flags section). Depending on these flags, type
```

tutorials/io/binary_serialization_api.rst
```diff
@@ -355,137 +605,207 @@ This field is padded to 4 bytes.
...
+information is encoded before the element count.
 
-Then what follows is, for amount of "elements", pairs of key and value,
-one after the other, using this same format.
+**Container Type Encoding (for each of key type and value type, if not NONE):**
```
AThousandShips:
```suggestion
**Container Type Encoding (for each of key and value type, if not NONE):**
```

KryziK:
> > No part of this document was output directly by Claude without manual review and clean-up
> 
> So parts of this was generated by Claude?
> 
> Edit: reading this I can definitely tell this was written by an LLM, it is not particularly well worded and uses a lot of incorrect terminology and has a lot of grammatical and formatting errors, please go through it again and comment when you've fixed them and I can review after, I am not interested in reviewing it as it is right now

Thank you for your time. 

Could you highlight an example of wrong terminology/grammar? I hand-wrote all of the sentences/statements and used the old documentation as my guide.

And, as we previously discussed, Claude was used to parse all of the variant types and provide me the new list of types, of which now there are many more than v3. However, I output that all into a separate file, and then manually went through that entire file and typed out the new documentation for the PR I submit. I did all of the manual creation of these little tables, their text blurbs, and the numbered types with their references.

Admittedly it was a lot of manual work to keep track of, so I could have made some mistakes or typos, but I'm not seeing anything obvious that is incorrect after a few passes through my own PR. The only one I did notice is my use of "sent" instead of "serialized", which was just semantics (sending values through the serializer to the file) and I can change that if that's your concern.

As far as formatting errors, I'm not sure which pieces or sections are not formatted properly. I took care to make the note sections look similar to the old doc's note sections, and tried to use all the same symbols, separators, etc..

Again, thank you for your time. I am working on this with my own free time and have no intention of disregarding the rules/guidelines; there's nothing I would gain from that anyways. I'm more than happy to redo any of the work necessary to bring it up to your standards, but indeed a pointer on where to start would help a lot.

I apologize if you feel as if your time is being wasted, and I understand any skepticism you may have.

I did take a look at the build log and fixed a typo I made with the spacing at the end of one of the tables. I am not sure how to re-queue the build.

AThousandShips:
Thank you for clarifying, I assumed given the phrasing which seemed to indicate you used it to write it, and the very AI looking errors and word choices, which were very different from the existing documentation and from the rest of the page

There are several unexplained changes, usage of unusual terminology, "type kind" is not used for example 

I will take a look at some point this week, but please go over it and look at existing content to see the difference in terminology

tutorials/io/binary_serialization_api.rst
```diff
@@ -355,137 +605,207 @@ This field is padded to 4 bytes.
...
-19: :ref:`Array<class_array>`
+**Element count (immediately following all type information):**
+
++----------+-------+-----------+----------------------------------------------------------------------+
+| Offset   | Len   | Type      | Description                                                          |
++==========+=======+===========+======================================================================+
+| 0        | 4     | Integer   | val & 0x7FFFFFFF = element count, val & 0x80000000 = shared (bool)   |
++----------+-------+-----------+----------------------------------------------------------------------+
+
+Then what follows is, for each element, a key-value pair encoded one after
```
AThousandShips: This, while from the original is a bit too much like prose, should be more technical

tutorials/io/binary_serialization_api.rst
```diff
@@ -355,137 +605,207 @@ This field is padded to 4 bytes.
...
+**Element count (immediately following type information, if any):**
 
-20: :ref:`PackedByteArray<class_PackedByteArray>`
++----------+-------+-----------+----------------------------------------------------------------------+
+| Offset   | Len   | Type      | Description                                                          |
++==========+=======+===========+======================================================================+
+| 0        | 4     | Integer   | val & 0x7FFFFFFF = element count, val & 0x80000000 = shared (bool)   |
++----------+-------+-----------+----------------------------------------------------------------------+
+
+Then what follows is, for each element, a value encoded using this same format.
```
AThousandShips: Same here

tutorials/io/binary_serialization_api.rst
AThousandShips:
```suggestion
| 4        | 4     | Integer   | String length (in bytes, N)  |
```

tutorials/io/binary_serialization_api.rst
```diff
@@ -235,8 +419,12 @@ This field is padded to 4 bytes.
 | 24       | 4     | Float   | The Y component of the origin vector, accessed via [2][1]     |
 +----------+-------+---------+---------------------------------------------------------------+
 
-9: :ref:`Plane<class_plane>`
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+If flag ``ENCODE_FLAG_64`` is set (flags & 1 == 1), the floats are sent as 64 bit double precision:
```
AThousandShips: I'd say all types should be fully described for their 64 bit version, or none, the longer ones are more useful

tutorials/io/binary_serialization_api.rst
```diff
@@ -340,7 +544,53 @@ This field is padded to 4 bytes.
...
+| 52       | 4     | Float   | w.x (Column 3, Row 0)        |
++----------+-------+---------+------------------------------+
+| 56       | 4     | Float   | w.y (Column 3, Row 1)        |
++----------+-------+---------+------------------------------+
+| 60       | 4     | Float   | w.z (Column 3, Row 2)        |
++----------+-------+---------+------------------------------+
+| 64       | 4     | Float   | w.w (Column 3, Row 3)        |
++----------+-------+---------+------------------------------+
+
+Total of 16 floats (64 bytes for single precision, 128 bytes for double).
```
AThousandShips:
```suggestion
Total of 16 floats (64 bytes for single precision, 128 bytes for double).

```
Not sure why this type alone has a description of the number of bytes it takes up?

tutorials/io/binary_serialization_api.rst
```diff
@@ -551,59 +871,133 @@ The array data is padded to 4 bytes.
...
-| Offset   | Len   | Type      | Description            |
-+==========+=======+===========+========================+
-| X+0      | 4     | Integer   | String length          |
-+----------+-------+-----------+------------------------+
-| X+4      | X     | Bytes     | UTF-8 encoded string   |
-+----------+-------+-----------+------------------------+
++----------+-------+-----------+----------------------------------------+
+| Offset   | Len   | Type      | Description                            |
++==========+=======+===========+========================================+
+| 0        | 4     | Integer   | String length (N, including null term) |
```
AThousandShips:
```suggestion
| 0        | 4     | Integer   | String length (N, including the null terminator) |
```

tutorials/io/binary_serialization_api.rst
```diff
@@ -551,59 +871,133 @@ The array data is padded to 4 bytes.
...
++----------+-------+-----------+----------------------------------------+
+| 4        | N     | Bytes     | UTF-8 encoded string with null         |
```
AThousandShips:
```suggestion
| 4        | N     | Bytes     | UTF-8 encoded string with null terminator        |
```

tutorials/io/binary_serialization_api.rst
```diff
@@ -509,40 +829,40 @@ The array data is padded to 4 bytes.
...
-| 4                | 4     | Integer   | Array length (Floats)                     |
-+------------------+-------+-----------+-------------------------------------------+
-| 8..8+length\*4   | 4     | Integer   | 32-bit IEEE 754 single-precision float    |
-+------------------+-------+-----------+-------------------------------------------+
++------------------+-------+-----------+-----------------------------------+
+| Offset           | Len   | Type      | Description                       |
++==================+=======+===========+===================================+
+| 4                | 4     | Integer   | Array length (Floats)             |
++------------------+-------+-----------+-----------------------------------+
+| 8..8+length\*4   | 4     | Float     | IEEE 754 single-precision float   |
```
AThousandShips:
```suggestion
| 8+i\*4   | 4     | Float     | IEEE 754 single-precision float   |
```
For uniformity, bit confusing otherwise IMO, same for the int and byte versions

tutorials/io/binary_serialization_api.rst
```diff
@@ -509,40 +829,40 @@ The array data is padded to 4 bytes.
...
-| 4                | 4     | Integer   | Array length (Floats)                     |
-+------------------+-------+-----------+-------------------------------------------+
-| 8..8+length\*8   | 8     | Integer   | 64-bit IEEE 754 double-precision float    |
-+------------------+-------+-----------+-------------------------------------------+
++------------------+-------+-----------+-----------------------------------+
+| Offset           | Len   | Type      | Description                       |
++==================+=======+===========+===================================+
+| 4                | 4     | Integer   | Array length (Floats)             |
++------------------+-------+-----------+-----------------------------------+
+| 8..8+length\*8   | 8     | Float     | IEEE 754 double-precision float   |
```
AThousandShips:
```suggestion
| 8+i\*8   | 8     | Float     | IEEE 754 double-precision float   |
```

AThousandShips (commented):
A basic pass, needs some more work and tweaking but a good start

tutorials/io/binary_serialization_api.rst
KryziK: It was one of the lengthiest types, so it was mainly just for clarification. I was actually looking at an example of this type in ImHex and, having done some manual counting, it was just habit to add that information.

Would you prefer I remove it for consistency?

tutorials/io/binary_serialization_api.rst
KryziK: Could you clarify what you mean? Are you suggesting I remove the 32-bit information from each variant? While 64-bit is more useful, we are describing a file structure here, so if it's possible that someone would have a file with a 32-bit value, wouldn't that still be important information as far as file formats go?

tutorials/io/binary_serialization_api.rst
AThousandShips: I'd say to add it to all or to none of the types that aren't a single value

tutorials/io/binary_serialization_api.rst
AThousandShips: No, I'm suggesting either all versions should have a detailed 64 bit version, or none, the 64 bit version is only available in double precision builds so it's pretty unusual

So to be clear, the 32 bit version should always be presented as that's what most people will be working with

tutorials/io/binary_serialization_api.rst
KryziK: I wanted your opinion on this, actually. For some of the types with multiple consecutive tables, I wasn't sure if I should start each table at offset 4 or 0, or instead index them relative to the start of the data (most have offset 4 because of the type ID being specified at offset 0).

I could:
1. Add the type header to every table, at offset 4, that way no single data structure is "composed" of multiple individual tables. Every table would fully define its structure, including the header.
2. For tables with potentially repeating sections, I could specify a small series to showcase the structure, or a subtable.

I also wanted to ask if the godot docs support the [:math: syntax](https://stackoverflow.com/questions/3610551/math-in-restructuredtext-with-latex)? If so, I was going to maybe update the Len column for things like the String variant to be
```
:math:`\lceil N / 4 \rceil \cdot 4`
```
to indicate that the length of the string itself is Ceil(N / 4) * 4 (padded to 4 bytes) (`⌈N/4⌉·4`)- that way we don't have to specify that it's padded to 4 bytes in every single table?

tutorials/io/binary_serialization_api.rst
KryziK: Could you help clarify something for me:

The existing docs state that the "Float" type always uses double precision (line 112 in the prior documentation). Wouldn't this be an 8-byte "double", and not a 4-byte "float" then? This seems to contradict the offsets and length in tables (and most standard IEEE documentation) that state that floats are 4 bytes?

And, having never used Godot, I was unaware that double-precision builds are unusual. Could you elaborate on this a tiny bit? Some things I'm reading say that Godot always uses 64-bit ints and floats. But in that case, they would never be just 4 bytes regardless of the build precision?

Just a little bit confused on this

AThousandShips: This is something that would be critical to understand to write this, so it might be an indication you might need additional help with this, I'd suggest reading things like the documentation for [`Vector2`](https://docs.godotengine.org/en/latest/classes/class_vector2.html#description) to see what this difference is about, also [here](https://docs.godotengine.org/en/latest/tutorials/physics/large_world_coordinates.html#enabling-large-world-coordinates)

But this is a critical point of understanding so please check the documentation and come back when you understand the engine core systems, it is better than me trying to explain to you how to write this page

KryziK: Okay, understood. I think I had a handle on the correct answer, but as I was reading more I was straying from that.

The fact that Godot uses 64-bit floats internally/in the engine is irrelevant to this task, as it will still serialize that into a 32-bit or 64-bit value, depending upon the build flag.

When strictly talking about binary serialization/deserialization, 32-bit values are used (and as you said, much more common) unless the 64-bit build flag is set.

If that's correct, then the only thing I still want to read into (albeit not relevant to this PR) is how Godot ensures that the runtime double will never overflow a 32-bit space.

If I'm incorrect or going down the wrong line of thinking, let me know. But to summarize, it seems like for serialization, the documentation is accurate: the float class, as well as any floats in things like vector2, basis, etc., will use 32 bits unless  otherwise specified by the build flag (and therefore the header flag). And in that case, I can remove all of the 64-bit versions of the tables and instead keep a blurb at the top indicating that the value in the header flag will control this across the board.

AThousandShips: No rather `float` in `Variant` is always 64 bit, but `Vector2/3/4` and other math types like `Basis`, `Projection`, `Transform2D/3D`, etc. use 32 or 64 bits depending on the compile time setting, so it's not strictly serialization related, though it doesn't matter here beyond the serialization part

Now the compiler flag will affect how these values are stored, but they will be parsed as 64 bit and then stored in 32 bit types if loading binary data from a double precision build in a single precision one, which is handled without issue (they just lose precision)

I'd say that the 64 bit entries can be useful, but it might be best to drop them, as they are niche, though they are useful. In that case I'd just keep the note indeed

Adding a note to the top about the encoding flag being controlled is also good

KryziK: So, you said "No", but were you talking about the instantiated/runtime variant types themselves or were you talking about serialization strictly? If you re-read my last message but apply everything I said only to serialization and not to what those types look like while running in the engine, is what I said correct?

If we ignore the runtime types, the compiler, and the engine and focus solely on serialization/deserialization,

Variant Float (type 3) is no different than any other type (Basis, Projection, etc.) in that it may be 32 or 64 bits when serialized, and the header will indicate which one it is.

If this is correct, then the idea that a Float variant is always 64 bits is irrelevant to us because we're not talking about what Godot does internally while running, we're just talking about what it encodes/decodes to a file.

Is this right?

Again, thank you for your time and patience. I think maybe I got a bit derailed by looking too much into the runtime/engine portion of things and not staying strictly focused on the serialization of that information.

AThousandShips: Good clarification, the flag is a bit more complicated in that it signals the 64 bit storage scheme is active, forgot to clarify that the flag is used only if needing to store 64 bit values, this happens when:
* Storing `int` which is outside the 32 signed range
* Storing `float` which requires storing as such, i.e. storing a value as 32 bit would lose precision or range (internally checked with `float(d) != d`)
* Storing any of the precision controlled types on a double prevision build

tutorials/io/binary_serialization_api.rst
KryziK: I've refactored almost the entire document, including these blurbs, the format of tables, and the offset 'calculations' in tables. Please go ahead and take a look when you have the time.

AThousandShips:
I'll try to take another pass of this later today or tomorrow!

tutorials/io/binary_serialization_api.rst
```diff
@@ -33,77 +31,138 @@ The packet is designed to be always padded to 4 bytes. All values are
...
++-------------+----------------------+
+| 9  (0x09)   | Vector3              |
++-------------+----------------------+
+| 10 (0x0A)   | Vector3i             |
++-------------+----------------------+
+| 11 (0x0B)   | Vector4              |
++-------------+----------------------+
+| 12 (0x0C)   | Vector4i             |
++-------------+----------------------+
+| 13 (0x0D)   | Transform2D          |
```
AThousandShips: This is incorrect, please see here:
https://github.com/godotengine/godot/blob/757bba192ef4c3254b44bc0d2c8cf36aa3343e30/core/variant/variant.h#L106-L114

tutorials/io/binary_serialization_api.rst
```diff
@@ -112,498 +171,722 @@ length of "Float" fields within data structures should be 8, and the offset
 should be ``(offset - 4) * 2 + 4``. The "float" type itself always uses double
 precision.
 
-0: null
+.. _nil-section:
+
+0: NIL
 ~~~~~~~
 
+null
```
AThousandShips: Needs a description IMO, will make a suggestion later for wording

tutorials/io/binary_serialization_api.rst
```diff
@@ -112,498 +171,722 @@ length of "Float" fields within data structures should be 8, and the offset
...
-+----------+-------+-----------+--------------------------+
-| Offset   | Len   | Type      | Description              |
-+==========+=======+===========+==========================+
-| 4        | 4     | Integer   | 32-bit signed integer    |
-+----------+-------+-----------+--------------------------+
++----------+----------+-----------+-------------------------+
+| Offset   | Length   | Type      | Description             |
++==========+==========+===========+=========================+
+| 4        | 4        | Integer   | 32-bit signed integer   |
++----------+----------+-----------+-------------------------+
```
AThousandShips: These changes are invalid, my suggestion was to not add the complex different versions between the 32 and 64 bit versions, but this one should be either 64 bits only or the 32 bit as well, but I'd just keep the original text that clarifies how it works

tutorials/io/binary_serialization_api.rst
```diff
@@ -112,498 +171,722 @@ length of "Float" fields within data structures should be 8, and the offset
...
-| 4        | 4     | Float   | IEEE 754 single-precision float   |
-+----------+-------+---------+-----------------------------------+
-
-If flag ``ENCODE_FLAG_64`` is set (``flags & 1 == 1``), the float is sent as
-a 64-bit double precision number:
++----------+----------+---------+-----------------------------------+
+| Offset   | Length   | Type    | Description                       |
++==========+==========+=========+===================================+
+| 4        | 4        | Float   | IEEE 754 single-precision float   |
++----------+----------+---------+-----------------------------------+
```
AThousandShips: Same here

tutorials/io/binary_serialization_api.rst
```diff
@@ -112,498 +171,722 @@ length of "Float" fields within data structures should be 8, and the offset
...
++----------+----------+---------+---------------+
+| 20       | 4        | Float   | y.y           |
++----------+----------+---------+---------------+
+| 24       | 4        | Float   | y.z           |
++----------+----------+---------+---------------+
+| 28       | 4        | Float   | z.x           |
++----------+----------+---------+---------------+
+| 32       | 4        | Float   | z.y           |
++----------+----------+---------+---------------+
+| 36       | 4        | Float   | z.z           |
```
AThousandShips: This removes relevant information IMO, the current wording is not really describing the way to access the information clearly, so I'd keep that, and potentially adding the `x.y` format as well

tutorials/io/binary_serialization_api.rst
```diff
@@ -112,498 +171,722 @@ length of "Float" fields within data structures should be 8, and the offset
...
++----------+----------+---------+---------------+
+| 32       | 4        | Float   | basis.z.y     |
++----------+----------+---------+---------------+
+| 36       | 4        | Float   | basis.z.z     |
++----------+----------+---------+---------------+
+| 40       | 4        | Float   | origin.x      |
++----------+----------+---------+---------------+
+| 44       | 4        | Float   | origin.y      |
++----------+----------+---------+---------------+
+| 48       | 4        | Float   | origin.z      |
```
AThousandShips: Same here, but replace the "accessed via" part with the descriptions you added here

AThousandShips (commented):
Will do a more detailed pass later for the rest but these are immediate details

tutorials/io/binary_serialization_api.rst
KryziK: Maybe I misunderstood - Why are these incorrect? The Header Flags section above, which talks about integers, floats, and math types, clarifies that this type will change based on the header flag, right?

tutorials/io/binary_serialization_api.rst
KryziK: I don't mind changing this to be some hybrid of the descriptions, but again we are focused on the serialization/deserialization of these types and not how to access the information before serialization or after deserialization. The type's respective page will tell you how to access the information and what each value means and is for, right? A basis is 3 axis vectors (Vector3): x, y, and z

Column/row wording and notation seems irrelevant to the concept of reconstructing a Basis object with its properties.

Let me know your thoughts.

tutorials/io/binary_serialization_api.rst
AThousandShips: Yes but the default size is relevant here and as I said above users will face the 64 bit version here regularly even without building with doubles, so this should be kept as the original

tutorials/io/binary_serialization_api.rst
AThousandShips: It remains critical to understand *what* the data that's stored *is*

KryziK: Okay, that I can agree with, yeah. I'll change these

tutorials/io/binary_serialization_api.rst
KryziK: Updated

tutorials/io/binary_serialization_api.rst
KryziK: Okay. Previously you never gave a concrete yes/no correct/incorrect to my questions, so I was not 100%  clear.

Updated

tutorials/io/binary_serialization_api.rst
KryziK: Updated

tutorials/io/binary_serialization_api.rst
KryziK: Updated

tutorials/io/binary_serialization_api.rst
```diff
@@ -112,498 +171,742 @@ length of "Float" fields within data structures should be 8, and the offset
...
-+-------------------+-------+-----------+----------------+
-
-27: :ref:`PackedVector3Array<class_PackedVector3Array>`
++----------+----------+-----------+----------------+
+| Offset   | Length   | Type      | Description    |
++==========+==========+===========+================+
+| 4        | 4        | Integer   | Array length   |
++----------+----------+-----------+----------------+
+
+Following the array length is a contiguous block of Vector2's.
```
AThousandShips:
```suggestion
Following the array length is a contiguous block of Vector2s.
```

tutorials/io/binary_serialization_api.rst
```diff
@@ -112,498 +171,742 @@ length of "Float" fields within data structures should be 8, and the offset
...
-+--------------------+-------+-----------+----------------+
-
-28: :ref:`PackedColorArray<class_PackedColorArray>`
++----------+----------+-----------+----------------+
+| Offset   | Length   | Type      | Description    |
++==========+==========+===========+================+
+| 4        | 4        | Integer   | Array length   |
++----------+----------+-----------+----------------+
+
+Following the array length is a contiguous block of Vector3's.
```
AThousandShips:
```suggestion
Following the array length is a contiguous block of Vector3s.
```

tutorials/io/binary_serialization_api.rst
```diff
@@ -112,498 +171,742 @@ length of "Float" fields within data structures should be 8, and the offset
...
+38: :ref:`PackedVector4Array<class_PackedVector4Array>`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++----------+----------+-----------+----------------+
+| Offset   | Length   | Type      | Description    |
++==========+==========+===========+================+
+| 4        | 4        | Integer   | Array length   |
++----------+----------+-----------+----------------+
+
+Following the array length is a contiguous block of Vector4's.
```
AThousandShips:
```suggestion
Following the array length is a contiguous block of Vector4s.
```

tutorials/io/binary_serialization_api.rst
```diff
@@ -112,498 +171,742 @@ length of "Float" fields within data structures should be 8, and the offset
...
+| Offset   | Length   | Type      | Description    |
++==========+==========+===========+================+
+| 4        | 4        | Integer   | X coordinate   |
++----------+----------+-----------+----------------+
+| 8        | 4        | Integer   | Y coordinate   |
++----------+----------+-----------+----------------+
+| 12       | 4        | Integer   | Z coordinate   |
++----------+----------+-----------+----------------+
+
+11: :ref:`Vector4<class_vector4>`
```
AThousandShips: The order here still needs to be updated

tutorials/io/binary_serialization_api.rst
```diff
@@ -112,498 +171,742 @@ length of "Float" fields within data structures should be 8, and the offset
...
+
+28: :ref:`Array<class_array>`
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-+----------+-------+-----------+---------------------------------------------------------------------+
-| Offset   | Len   | Type      | Description                                                         |
-+==========+=======+===========+=====================================================================+
-| 4        | 4     | Integer   | val&0x7FFFFFFF = elements, val&0x80000000 = shared (bool)           |
-+----------+-------+-----------+---------------------------------------------------------------------+
+For typed arrays, the header flags indicate the element type kind (see Header
```
AThousandShips:
```suggestion
For typed arrays, the header flags indicate the element type (see Header
```

tutorials/io/binary_serialization_api.rst
```diff
@@ -112,498 +171,742 @@ length of "Float" fields within data structures should be 8, and the offset
...
-| 4             | 4     | Integer   | Array length (Bytes)   |
-+---------------+-------+-----------+------------------------+
-| 8..8+length   | 1     | Byte      | Byte (0..255)          |
-+---------------+-------+-----------+------------------------+
++----------+----------+-----------+---------------------------------+
+| Offset   | Length   | Type      | Description                     |
++==========+==========+===========+=================================+
+| 4        | 4        | Integer   | Array length (L)                |
++----------+----------+-----------+---------------------------------+
+| 8 + i    | 1        | Byte      | Array element #i (0 <= i < L)   |
```
AThousandShips:
```suggestion
| 8 + i    | 1        | Byte      | Array element #i ``(0 <= i < L)`` |
```
For all these expressions, for cleaner formatting

tutorials/io/binary_serialization_api.rst
```diff
@@ -112,498 +171,742 @@ length of "Float" fields within data structures should be 8, and the offset
...
+
+Header:
++----------+----------+-----------+--------------------------------------------------------+
+| Offset   | Length   | Type      | Description                                            |
++==========+==========+===========+========================================================+
+| 4        | 4        | Integer   | Class name (String length, in bytes, N)                |
++----------+----------+-----------+--------------------------------------------------------+
+| 8        | N        | Bytes     | Class name (UTF-8 encoded string, padded to 4 bytes)   |
++----------+----------+-----------+--------------------------------------------------------+
+| 8+N      | 4        | Integer   | The number of properties that are serialized (P)       |
```
AThousandShips:
```suggestion
| 8 + N    | 4        | Integer   | The number of properties that are serialized (P)       |
```

tutorials/io/binary_serialization_api.rst
```diff
@@ -112,498 +171,742 @@ length of "Float" fields within data structures should be 8, and the offset
...
+where the offset is relative to the start of each property entry:
+
++----------+--------------+--------------+-----------------------------------------------------------+
+| Offset   | Length       | Type         | Description                                               |
++==========+==============+==============+===========================================================+
+| 0        | 4            | Integer      | Property name (String length, in bytes, N)                |
++----------+--------------+--------------+-----------------------------------------------------------+
+| 4        | N            | Bytes        | Property name (UTF-8 encoded string, padded to 4 bytes)   |
++----------+--------------+--------------+-----------------------------------------------------------+
+| 4+N      | <variable>   | <variable>   | Property value, using this same format                    |
```
AThousandShips:
```suggestion
| 4 + N    | <variable>   | <variable>   | Property value, using this same format                    |
```

tutorials/io/binary_serialization_api.rst
```diff
@@ -112,498 +171,742 @@ length of "Float" fields within data structures should be 8, and the offset
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++----------+----------+-----------+---------------------------------------------------------+
+| Offset   | Length   | Type      | Description                                             |
++==========+==========+===========+=========================================================+
+| 4        | 4        | Integer   | Signal name (String length, in bytes, N)                |
++----------+----------+-----------+---------------------------------------------------------+
+| 8        | N        | Bytes     | Signal name (UTF-8 encoded string, padded to 4 bytes)   |
++----------+----------+-----------+---------------------------------------------------------+
+| 8+N      | 8        | Integer   | Object instance ID (64-bit)                             |
```
AThousandShips:
```suggestion
| 8 + N    | 8        | Integer   | Object instance ID (64-bit)                             |
```

AThousandShips (commented):
Looking a lot better! I'd suggest adding a note to all types depending on the 64 bit flag, just a short note that they'll be stored as such

KryziK:
@AThousandShips Thank you again so much for all your time. I'll get these updated later today.

tutorials/io/binary_serialization_api.rst
KryziK: Updated

tutorials/io/binary_serialization_api.rst
KryziK: Fixed this and for 2 and 3

tutorials/io/binary_serialization_api.rst
KryziK: Removed

tutorials/io/binary_serialization_api.rst
KryziK: Added to all, but need to make sure my formatting doesn't break the PR build

tutorials/io/binary_serialization_api.rst
KryziK: Updated all +N

tutorials/io/binary_serialization_api.rst
```diff
@@ -112,498 +171,744 @@ length of "Float" fields within data structures should be 8, and the offset
...
-+--------------------+-------+-----------+--------------------------------------------------------------+
++----------+----------+-----------+----------------+
+| Offset   | Length   | Type      | Description    |
++==========+==========+===========+================+
+| 4        | 4        | Integer   | Array length   |
++----------+----------+-----------+----------------+
+
+Following the array length is a contiguous block of Colors.
+
+For each Color entry (0 <= i < length):
```
AThousandShips:
```suggestion
For each Color entry ``(0 <= i < length)``:
```

tutorials/io/binary_serialization_api.rst
```diff
@@ -112,498 +171,744 @@ length of "Float" fields within data structures should be 8, and the offset
...
+
++----------+----------+-----------+----------------+
+| Offset   | Length   | Type      | Description    |
++==========+==========+===========+================+
+| 4        | 4        | Integer   | Array length   |
++----------+----------+-----------+----------------+
+
+Following the array length is a contiguous block of Vector4s.
+
+For each Vector4 entry (0 <= i < length):
```
AThousandShips:
```suggestion
For each Vector4 entry ``(0 <= i < length)``:
```

tutorials/io/binary_serialization_api.rst
```diff
@@ -112,498 +171,744 @@ length of "Float" fields within data structures should be 8, and the offset
...
-+----------+-------+-----------+--------------------------+
++----------+----------+-----------+--------------------------+
+| Offset   | Length   | Type      | Description              |
++==========+==========+===========+==========================+
+| 4        | 4        | Integer   | Array length (Strings)   |
++----------+----------+-----------+--------------------------+
+
+Following the array length is a contiguous block of strings.
+
+For each string entry (0 <= i < length):
```
AThousandShips:
```suggestion
For each string entry ``(0 <= i < length)``:
```

AThousandShips:
```suggestion
For each string entry ``(0 <= i < length)``:
```

tutorials/io/binary_serialization_api.rst
```diff
@@ -112,498 +171,744 @@ length of "Float" fields within data structures should be 8, and the offset
...
-27: :ref:`PackedVector3Array<class_PackedVector3Array>`
++----------+----------+-----------+----------------+
+| Offset   | Length   | Type      | Description    |
++==========+==========+===========+================+
+| 4        | 4        | Integer   | Array length   |
++----------+----------+-----------+----------------+
+
+Following the array length is a contiguous block of Vector2s.
+
+For each Vector2 entry (0 <= i < length):
```
AThousandShips:
```suggestion
For each Vector2 entry ``(0 <= i < length)``:
```

tutorials/io/binary_serialization_api.rst
```diff
@@ -112,498 +171,744 @@ length of "Float" fields within data structures should be 8, and the offset
...
-28: :ref:`PackedColorArray<class_PackedColorArray>`
++----------+----------+-----------+----------------+
+| Offset   | Length   | Type      | Description    |
++==========+==========+===========+================+
+| 4        | 4        | Integer   | Array length   |
++----------+----------+-----------+----------------+
+
+Following the array length is a contiguous block of Vector3s.
+
+For each Vector3 entry (0 <= i < length):
```
AThousandShips:
```suggestion
For each Vector3 entry ``(0 <= i < length)``:
```

tutorials/io/binary_serialization_api.rst
AThousandShips:
```suggestion
For each Color entry ``(0 <= i < length)``:
```

tutorials/io/binary_serialization_api.rst
AThousandShips:
```suggestion
For each Vector4 entry ``(0 <= i < length)``:
```

AThousandShips (commented):
Looks much better! Still missing the notes about the 64 bit versions on the types though

When that's been fixed and these details I'll request a review from the core team to go over it too, but it should be good!

AThousandShips:
I'll take a new pass of this this week

tutorials/io/binary_serialization_api.rst
```diff
@@ -112,498 +171,744 @@ length of "Float" fields within data structures should be 8, and the offset
...
++----------+----------+-----------+------------------------------+
+| Offset   | Length   | Type      | Description                  |
++==========+==========+===========+==============================+
+| 4        | 4        | Integer   | 0 (Representing a nullptr)   |
++----------+----------+-----------+------------------------------+
+
+**Non-null object:**
+
+Header:
++----------+----------+-----------+--------------------------------------------------------+
```
Calinou:
```suggestion
Header:

+----------+----------+-----------+--------------------------------------------------------+
```

tutorials/io/binary_serialization_api.rst
```diff
@@ -1,609 +1,915 @@
...
+- Byte 1 (bits 8-15): Unused
+- Bytes 2-3 (bits 16-31): Additional data (type-specific flags)
+
+::
+
+    type = header & 0xFF;
+    flags = header >> 16;
+
++-------------+----------------------+
+| Value       | Type Description     |
```
Calinou:
```suggestion
| Value       | Type description     |
```

AThousandShips (approved):
I think this looks good now

Calinou (commented):
Looks good to me.

PS: Could you share the [ImHex pattern file you mentioned](https://github.com/godotengine/godot-docs/issues/4945#issuecomment-3618314235) as part of working on this PR?

Calinou:
Can you run `dos2unix tutorials/io/binary_serialization_api.rst` or use some editor tool to fix line endings? It seems the first GitHub suggestion I applied changed all line endings to CRLF for some reason (we use LF normally).

I can't push the change myself since this PR was opened from a fork's `master` branch.

Issue #11459, opened 2025-11-11
Provide more specific instructions on "Handling compatibility breakages" page


Meorge:
**Your Godot version:** 4.5, 4.6

**Issue description:**
The last step of adding compatibility bindings – adding the relevant lines to the `extension_api_validation` file – feels rather rushed compared to the line-by-line explanation of the steps before it. For generating the lines of text to put in the file, the page says:
> The instructions for how to add to that file are at the top of the file itself.

The top of the file, in turn, says:
> This file contains the expected output of --validate-extension-api when run against the extension_api.json of the 4.5-stable tag (the basename of this file).
> 
> Only lines that start with "Validate extension JSON:" matter, everything else is considered a comment and ignored. They should instead be used to justify these changes and describe how users should work around these changes.
> 
> Add new entries at the end of the file.

It was unclear to me what exactly this meant I should do _in practice_ to get the text I needed to put in the `extension_api_validation` file. I ended up using a script file in the repository, which seemed to work:
```
> ./misc/scripts/validate_extension_api.sh ./bin/godot.macos.editor.dev.arm64
```
I'm not 100% sure what I did is the intended workflow for contributors, though. Providing more explicit, step-by-step instructions for this process would be very helpful. Given that I doubt the process changes much from version to version, I don't see why this documentation page couldn't just provide the specific commands, like it does for [building the engine itself](https://docs.godotengine.org/en/stable/engine_details/development/compiling/compiling_for_macos.html).

**URL to the documentation page (if already existing):** https://docs.godotengine.org/en/stable/engine_details/development/handling_compatibility_breakages.html

AThousandShips:
We can copy over the instructions from the linked PR directly into the page absolutely, the assumption when writing that was that if needed readers would consult the PR for that information

Meorge:
Perhaps I'm missing something obvious, but looking at https://github.com/godotengine/godot/pull/76446 (linked at the bottom of the documentation page), the closest I see is the following in the initial PR post:

> It is also now possible to ask an engine build whether it breaks compatibility with a previous version by calling like this:
> 
> ```
> $ godot4.1 --validate-extension-api godot4.0-api.json
> ```
> If compatibility was broken, Godot will return with an exit code other than zero.
> 
> Which means it will need to be fixed by providing compatibility versions as described above.
> 
> The idea would be to eventually integrate this to the CI, by storing previous versions of the Godot stable APIs that will need to be tested.

[...]

PR #11519, opened 2025-12-07
Add more details to "Handling compatibility breakages" page

Meorge:
Closes https://github.com/godotengine/godot-docs/issues/11459 .

This PR adds more detailed instructions to the "Handling compatibility breakages" page, explaining how to generate the "Validate extension JSON" text that goes in the `GH-*.txt` file.

It also emphasizes the "Hash changed" error and gives some guidance on things to check when trying to fix it.

This is probably already assumed, but just in case: it'd be very good to get GDExtension peoples' OK on this before merging. The steps I laid out here make logical sense to me and they worked when I tested them with a PR of mine, but I certainly don't want to end up misleading people if there are steps or caveats I'm not aware of.

### DIFF
--- a/engine_details/development/handling_compatibility_breakages.rst
+++ b/engine_details/development/handling_compatibility_breakages.rst
@@ -114,8 +114,53 @@ Next, this should be included in the ``.cpp`` file we're adding compatibility me
 
     #include "core/variant/typed_array.h"
 
-And finally, the changes reported by the API validation step should be added to a validation file. Because this was
-done during the development of 4.3, the validation file would be in the ``misc/extension_api_validation/4.2-stable/`` folder:
+Finally, the GDExtension API changes need to be recorded. To do this, first compile Godot on the ``master`` branch, and
+then run it with the ``--dump-extension-api`` flag:
+
+.. code-block:: shell
+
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):

mhilbrunner:
Thank you! Merged.

Meorge:
Thanks so much! 😄 

One small note: I noticed this received the cherrypick:4.5 label; however, it references the new file-based system for PRs' compatibility breakages (using `4.x-stable/GH-#####.txt` files instead of a single `4.x-stable.expected` file) being introduced with 4.6. As such, would it make sense for this to be cherrypicked back to 4.5's documentation? This is made even trickier by the fact that the example given on the page takes place during development of 4.3, when this new system wasn't there either... 😅

mhilbrunner:
Good point, removed the label for now, thanks for pointing that out :)

Meorge:
No problem! Most of the PR content is also applicable to before the new file-based system; would it make sense for me to submit a pre-4.6-specific version for the sole purpose of cherrypicking? Given that's not where main engine development is now, it may be unnecessary, but otherwise I'd be happy to do that!

PR #11528, opened 2025-12-10
Clarify what is defined by the parent in Scene organization

Calinou:
"Parent-defined" could be understood as the method or signal from the parent being called, but the parent defines just the name of the method or signal, instead of the method body.

- See https://github.com/godotengine/godot-docs-user-notes/discussions/403#discussioncomment-15220997.


mhilbrunner:
Thanks!

PR #11530, opened 2025-12-12
Update documentation around EditorDock

lodetrick:
After EditorDocks was introduced into master and deprecated the `add_control_to_docks` and `add_control_to_bottom_panel` methods, the documentation needs to be updated to remove references to these deprecated methods and to describe the new method of creating docks.


Calinou (approved):
Looks good to me.

Note that this shouldn't be cherry-picked to 4.5, since this relies on new EditorDock functionality in 4.6.

lodetrick:
Sorry, just realized I forgot a `var` before `dock_scene` so it would have given an error when copy-pasted. It should be fixed now

mhilbrunner:
Thank you! Merged.

PR #11533, opened 2025-12-14
Add page on controller-specific features (currently just LED)

Meorge:
Sets the groundwork and first feature for https://github.com/godotengine/godot-docs/issues/11466 .

This PR adds a new page, "Controller features", which describes and provides examples for various controller features that are less essential than ["Controllers, gamepads, and joysticks"](https://docs.godotengine.org/en/stable/tutorials/inputs/controllers_gamepads_joysticks.html#).

Currently, because LED light control is the only merged feature so far, it is the only section on the page; however, as other features like [adaptive triggers](https://github.com/godotengine/godot/pull/111682), [touchpads](https://github.com/godotengine/godot/pull/111714), [accelerometer and gyroscope](https://github.com/godotengine/godot/pull/111679), and [joypad information](https://github.com/godotengine/godot/pull/112680) are added to the engine, corresponding sections on these features can be added to this page.

### DIFF
--- a/tutorials/inputs/controller_features.rst
+++ b/tutorials/inputs/controller_features.rst
@@ -16,7 +16,7 @@ and how you can get started with them in Godot.
 
     Unless you specifically advertise your game as requiring specific controllers,
     remember that there is no guarantee that players will have a controller with
-    given features.
+    any given features.
 
     As a result, we suggest using these features to enhance the gameplay experience
     for players whose controllers support them, without detracting from those who


Meorge:
If anyone has a good camera setup that would allow them to record a DualSense/DualShock controller's LEDs, perhaps that looks a bit like this:



I think that would make a great addition to the code samples, so people could see how it looks in practice without needing to set up their own project 😄

tutorials/inputs/controller_features.rst
```diff
@@ -0,0 +1,90 @@
...
+    given features.
+
+    As a result, we suggest using these features to enhance the gameplay experience
+    for players whose controllers support them, without detracting from those who
+    don't have controllers.
+
+
+LED color
+---------
+.. https://github.com/godotengine/godot/pull/111681
```
skyace65: This comment should be removed. We don't typically include links to PRs that implemented features.

tutorials/inputs/controller_features.rst
```diff
@@ -0,0 +1,90 @@
+.. _doc_controller_features:
+
+Controller features
+===================
+
+Godot supports controller-specific features that can further enhance the gameplay
+experience. This page describes these features, how existing games have used them,
+and how you can get started with them in Godot.
```
skyace65, line 8: Might be worth re-wording this to just be explicit that only LED features are supported for now. I expect that to change in the future, but when 4.6 releases it's going to be the only thing available.

tutorials/inputs/controller_features.rst
```diff
@@ -0,0 +1,90 @@
...
+
+.. warning::
+
+    Unless you specifically advertise your game as requiring specific controllers,
+    remember that there is no guarantee that players will have a controller with
+    given features.
+
+    As a result, we suggest using these features to enhance the gameplay experience
+    for players whose controllers support them, without detracting from those who
+    don't have controllers.
```
skyace65, line 23: This warning is good advice, but I'm 50/50 on if it should be included or not and I'll defer to others. In my mind this just seems so obvious it's not worth having it here.

tutorials/inputs/controller_features.rst
```diff
@@ -0,0 +1,90 @@
...
+
+
+LED color
+---------
+.. https://github.com/godotengine/godot/pull/111681
+
+Games use the LED lights on certain controllers to subtly complement the on-screen gameplay by
```
skyace65:
```suggestion
Games can use the LED lights on certain controllers to subtly complement the on-screen gameplay by
```

minor nitpick.

Nintorch:
> Is there any official list on what controllers support light changing via SDL3?

I think only DualShock 4 and DualSense (and their copies) have LED lights. I haven't checked if copies of those controllers work too but I think they should since to work on their respective consoles they have to implement the original packet "interface" that is also used by SDL in the HIDAPI joystick driver.

Meorge:
Thanks for the review!

I tried digging into the SDL source code to find how it determines whether or not a joypad supports LEDs, but it looks like it eventually defers to platform-specific APIs. As such, I think it should be sufficient to simply state that the recent PlayStation controllers (DualShock and DualSense) reliably have them, but that in general `has_joy_light()` can be used to check for any arbitrary controller.

tutorials/inputs/controller_features.rst
Meorge: I was imagining this page focusing on features that exist mainly to add extra touches of "polish" or "juice" to the game, rather than play an essential part in the experience. It wouldn't surprise me that some (possibly less-experienced) developers might incorporate these features since they're testing with a controller that supports them, forgetting that the features aren't available to everyone.

tutorials/inputs/controller_features.rst
Meorge: I get what you're suggesting here, but I'm not sure at the moment how to rephrase it so that the page overall doesn't sound like it should be entirely dedicated to the LED feature. For example, if we changed

> Godot supports controller-specific features that can further enhance the gameplay experience.

to

> Godot supports changing the LED color on specific controllers to further enhance the gameplay experience.

then as a reader, I would probably wonder why the page as a whole isn't titled something like "LEDs on controllers" instead of the more generic "Controller features". Given that the "LED color" section is the only one on the page, IMO it's self-explanatory that it's the only supported feature for now.

I suppose we could add a sentence saying something like

> For Godot 4.6, only controller LED support is available, but more features will be supported in future versions.

However, that seems like a bit deeper of a dive into Godot's development than the manual tends to do.

tutorials/inputs/controller_features.rst
```diff
@@ -0,0 +1,90 @@
...
+
+Godot supports controller-specific features that can further enhance the gameplay
+experience. This page describes these features, how existing games have used them,
+and how you can get started with them in Godot.
+
+.. warning::
+
+    Unless you specifically advertise your game as requiring specific controllers,
+    remember that there is no guarantee that players will have a controller with
+    given features.
```
AThousandShips:
```suggestion
    any given features.
```

mhilbrunner:
Thank you! Merged.

PR #11535, opened 2025-12-15
CI: Fix `sync_class_ref.yml` workflow

Repiteo:
- Relevant RC thread: https://chat.godotengine.org/channel/documentation?msg=vMPsisoLih9MGr7ez

Makes a minor adjustment to the class sync workflow to call `make_rst.py` through Python directly, sidestepping an issue where the executable flag was erroneously removed


mhilbrunner:
Thanks!

PR #11538, opened 2025-12-16
Update number of built-in BBCode effects

DevPoodle:
The number was not updated when the Pulse effect was added, so the docs currently say there are 5 effects, but then it lists 6.


tomsterBG:
For future maintenance, why not rephrase it as "Several" instead of counting upwards every time the feature set changes? This would make the documentation more accurate over time.

DevPoodle:
Good idea. It now says 'several'.

skyace65:
Thanks!

PR #11545, opened 2025-12-18
[iOS] Change minimum requirement for Metal to A12.

bruvzg:
Depends on https://github.com/godotengine/godot/pull/114098


clayjohn (commented):
I'm fine with making this change, but note we technically support A11, even with the change to export settings because users can easily disable the flag to ship on older devices. 

That being said, in practice, we feel it is a bad idea to ship without the flag that limits running on older devices, so this change makes sense as a base level of communication for users. They don't need to know the nuance of exactly which version is supported and why

skyace65:
Thanks!

PR #11551, opened 2025-12-20
Add documentation page about `gdextension_interface.json`

dsnopek:
This adds documentation on the `gdextension_interface.json` file that was added in PR https://github.com/godotengine/godot/pull/107845

It attempts to address the remaining issues brought up on https://github.com/godotengine/godot/issues/113732

cc @Bromeon 

Fixes https://github.com/godotengine/godot/issues/113732

### DIFF
--- a/tutorials/scripting/gdextension/gdextension_interface_json_file.rst
+++ b/tutorials/scripting/gdextension/gdextension_interface_json_file.rst
@@ -1,0 +1,350 @@
+.. _doc_gdextension_interface_json_file:
+
+The C interface JSON file
+=========================
+
+The ``gdextension_interface.json`` file is the "source of truth" for the C API that
+Godot uses to communicate with GDExtensions.
+
+You can use the Godot executable to dump the file by using the following command:
+
+.. code-block:: shell
+
+    godot --headless --dump-gdextension-interface-json
+
+This file is intended to be used by GDExtension language bindings to generate code for
+using this API in whatever form makes the most sense for that language.
+
+.. note::
+
+    This is not to be confused with the ``extension_api.json``, which is also used by
+    GDExtension language bindings, and contains information about the classes and
+    methods that are exposed by Godot. The ``gdextension_interface.json`` is more
+    low-level, and is used to interact with those higher-level classes and methods.
+
+For languages that can be extended via C, or provide tools for interacting with C code,
+it's also possible to use the Godot executable to dump a generated C header file:
+
+.. code-block:: shell
+
+    godot --headless --dump-gdextension-interface
+
+.. note::
+
+    The header file is compatible with earlier versions of the header file that were included
+    with Godot 4.5 and earlier, which means it preserves some typos in names in order to
+    ensure compatibility.
+
+The goal of this page is to explain the JSON format for the GDExtension language bindings that
+would like to do their own code generation from the JSON.
+
+Overall structure
+-----------------
+
+The JSON file is broken up into 3 sections:
+
+- The header, which includes some miscellaneous information at the top-level of the JSON file.
+- The ``types`` key, which defines all the types used in the GDExtension interface.
+- The ``interface`` key, which defines all the function pointers that can be loaded via the
+  ``GDExtensionInterfaceGetProcAddress`` function pointer, which is passed to all GDExtensions
+  when they are loaded.
+
+There is a complete `JSON schema <https://github.com/godotengine/godot/blob/master/core/extension/gdextension_interface.schema.json>`__
+included in Godot's source code.
+
+Even though we may add new types and interface functions with each minor release of Godot, we
+strive to **never** change them in a backwards incompatible way, or remove them. Every
+interface function is labeled with the version of Godot it was introduced in (the ``since``
+key), so you can always use the latest version of the file, and simply refrain from using
+anything in versions of Godot that are newer than the version you are targeting.
+
+Header
+------
+
+The "header" is made up of 3 miscellaneous keys at the top-level of the file:
+
+- ``_copyright``: The standard copyright and license text that Godot includes in all source
+  code files.
+- ``$schema``: Points to the JSON schema relative to this file. It can be useful to place
+  the schema in the same directory, if you're viewing it with a code editor that understands
+  JSON schema.
+- ``format_version``: An integer for the version of the file format (meaning the schema).
+  Right now, there is only one format version (``1``). If we ever change the file format in
+  an incompatible way, we will increment this number. This *doesn't* reflect the version
+  of the data in the file (so it won't change between Godot versions), only its format.
+  Hopefully, we'll never have to use it, but it allows code generators to error early if they
+  encounter an unexpected value here.
+
+Types
+-----
+
+The ``types`` section is an array of types that will be used by other types, and the interface
+functions that will be in the last section.
+
+The types should be evaluated in order. Later types may refer to earlier types, but earlier
+types will not refer to later types.
+
+There is a small set of built-in types which aren't explicitly listed in the JSON:
+
+- ``void``
+- ``int8_t``
+- ``uint8_t``
+- ``int16_t``
+- ``uint16_t``
+- ``int32_t``
+- ``uint32_t``
+- ``int64_t``
+- ``uint64_t``
+- ``size_t`` (``uint32_t`` on 32-bit architectures, and ``uint64_t`` on 64-bit architectures)
+- ``char``
+- ``char16_t``
+- ``char32_t``
+- ``wchar_t``
+- ``float``
+- ``double``
+
+These correspond to their equivalent C types.
+
+Additionally, types can include modifiers such as:
+
+- ``*`` (e.g. ``int8_t*``) to indicate a pointer to the type
+- ``const`` (e.g. ``const int8_t*``) to indicate a const type
+
+Each type defined in the JSON file falls into one of 5 "kinds":
+
+- ``enum``
+- ``handle``
+- ``alias``
+- ``struct``
+- ``function``
+
+Regardless of the "kind", all types can have the following keys:
+
+- ``kind`` (required): The type's "kind".
+- ``name`` (required): The name of the type, which could be used as a valid C identifier.
+- ``description``: An array of strings documenting the type, where each string is a line of
+  documentation (this format for ``description`` is used throughout the JSON file).
+- ``deprecated``: An object with its own keys for the Godot version the type was deprecated in
+  (``since``), a message explaining the deprecation (``message``), and optionally a replacement
+  to use instead (``replacement``).
+
+Enums
+~~~~~
+
+Enums are 32-bit integers with a fixed set of possible values. In C, they could be represented
+as an ``enum``.
+
+They have the following keys:
+
+- ``is_bitfield``: If true, this enum is a bitfield, where the enum values can be bitwise OR'd together.
+  It is false by default.
+- ``values``: The array of fixed values for this enum, each with a ``name``, ``value``, and ``description``.
+
+An enum should be represented as an ``int32_t``, unless ``is_bitfield`` is true, in which case a ``uint32_t``
+should be used.
+
+Example
++++++++
+
+.. code-block:: json
+
+    {
+        "name": "GDExtensionInitializationLevel",
+        "kind": "enum",
+        "values": [
+            {
+                "name": "GDEXTENSION_INITIALIZATION_CORE",
+                "value": 0
+            },
+            {
+                "name": "GDEXTENSION_INITIALIZATION_SERVERS",
+                "value": 1
+            },
+            {
+                "name": "GDEXTENSION_INITIALIZATION_SCENE",
+                "value": 2
+            },
+            {
+                "name": "GDEXTENSION_INITIALIZATION_EDITOR",
+                "value": 3
+            },
+            {
+                "name": "GDEXTENSION_MAX_INITIALIZATION_LEVEL",
+                "value": 4
+            }
+        ]
+    }
+
+Handles
+~~~~~~~
+
+Handles are pointers to opaque structs. In C, they could be represented as ``void *`` or ``struct{} *``.
+
+They have the following keys:
+
+- ``is_const``: If true, this handle type is to be treated as a "const pointer", meaning its internal
+  data will not be changed. It is false by default.
+- ``is_uninitialized``: If true, this handle type is to be treated as pointing to uninitialized memory
+  (which may be initialized using interface functions). It is false by default.
+- ``parent``: The optional name of another handle type, if this handle type is the const or uninitialized
+  version of the parent type. This only makes sense if either ``is_const`` or ``is_uninitialized`` is true.
+
+Handles are the size of pointers on the given architecture (so, 64-bit on x86_64 and 32-bit on x86_32,
+for example).
+
+Example
++++++++
+
+.. code-block:: json
+
+    {
+        "name": "GDExtensionStringNamePtr",
+        "kind": "handle"
+    }
+
+Aliases
+~~~~~~~
+
+Aliases are alternative names for a type. In C, they could be represented as a ``typedef``.
+
+They have only one additional key:
+
+- ``type``: The type the alias is an alternative name for. It may include modifiers as described above.
+
+These should be represented using the same C type as the type they refer to.
+
+Example
++++++++
+
+.. code-block:: json
+
+    {
+        "name": "GDExtensionInt",
+        "kind": "alias",
+        "type": "int64_t"
+    }
+
+Structs
+~~~~~~~
+
+Structs represent C ``struct``\ s (aka a block of memory made up of the given members in order), and should
+follow all the same layout and alignment rules as C structs.
+
+They have only one additional key:
+
+- ``members``: An array of objects which have a ``name``, ``type`` (which may include modifiers), and
+  ``description``.
+
+Example
++++++++
+
+.. code-block:: json
+
+    {
+        "name": "GDExtensionCallError",
+        "kind": "struct",
+        "members": [
+            {
+                "name": "error",
+                "type": "GDExtensionCallErrorType"
+            },
+            {
+                "name": "argument",
+                "type": "int32_t"
+            },
+            {
+                "name": "expected",
+                "type": "int32_t"
+            }
+        ]
+    }
+
+Functions
+~~~~~~~~~
+
+Functions represent C function pointer types, with a list of arguments and a return type, and should
+follow the same size and alignment requirements as C function pointers.
+
+They have the following members:
+
+- ``return_value``: An object which has a ``type`` (which may include modifiers) and ``description``.
+  It will always be given, and use ``void`` as the ``type`` if the function returns no value.
+- ``arguments``: An array of function arguments which each has a ``type`` (which may include modifiers),
+  ``name``, and ``description``.
+
+
+Example
++++++++
+
+.. code-block:: json
+
+    {
+        "name": "GDExtensionPtrConstructor",
+        "kind": "function",
+        "return_value": {
+            "type": "void"
+        },
+        "arguments": [
+            {
+                "name": "p_base",
+                "type": "GDExtensionUninitializedTypePtr"
+            },
+            {
+                "name": "p_args",
+                "type": "const GDExtensionConstTypePtr*"
+            }
+        ]
+    }
+
+Interface
+---------
+
+The ``interface`` section of the JSON file is the list of interface functions, which can be loaded
+by ``name`` using the ``GDExtensionInterfaceGetProcAddress`` function pointer, which is
+passed to all GDExtensions when they are loaded.
+
+Interface functions have some of the same keys as types, including ``name`` (required),
+``deprecated``, and ``description``.
+
+And they also have ``return_value`` (required) and ``arguments`` (required) that have the same format
+as the equivalent keys on function types (as described in the previous section).
+
+There are only a handful of unique keys:
+
+- ``since`` (required): The Godot version that introduced this interface function.
+- ``see``: An array of strings describing external references with more information, for example,
+  names of classes or functions in the Godot source code, or URLs pointing to documentation.
+- ``legacy_type_name``: The legacy name used for the function pointer type in the header generated
+  by Godot, when the legacy name doesn't match the pattern used for these type names. This field
+  only exists so that we can generate the header in a way that is backwards compatible with the
+  header from Godot 4.5 or earlier, and it shouldn't be used unless you also need to maintain
+  compatibility with the old header.
+
+Example
+~~~~~~~
+
+.. code-block:: json
+
+    {
+        "name": "get_godot_version",
+        "return_value": {
+            "type": "void"
+        },
+        "arguments": [
+            {
+                "name": "r_godot_version",
+                "type": "GDExtensionGodotVersion*",
+                "description": [
+                    "A pointer to the structure to write the version information into."
+                ]
+            }
+        ],
+        "description": [
+            "Gets the Godot version that the GDExtension was loaded into."
+        ],
+        "since": "4.1",
+        "deprecated": {
+            "since": "4.5",
+            "replace_with": "get_godot_version2"
+        }
+    }
--- a/tutorials/scripting/gdextension/index.rst
+++ b/tutorials/scripting/gdextension/index.rst
@@ -23,4 +23,5 @@ articles instead, such as the articles about :ref:`C++ (godot-cpp) <doc_godot_cp
 
    what_is_gdextension
    gdextension_file
+   gdextension_interface_json_file
    gdextension_c_example


dsnopek:
This page needs a shorter title, because it gets weird line breaks currently:

Selection_417

Any suggestions? It could be "GDExtension interface JSON" maybe?

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,344 @@
...
+Godot uses to communicate with GDExtensions.
+
+You can use the Godot executable to dump the file by using the following command:
+
+.. code-block:: shell
+
+    godot --headless --dump-gdextension-interface-json
+
+This file is intended to be used by GDExtension language bindings to generate code for
+using this API in whatever form makes the most sense for that language.
```
Bromeon: Maybe we should briefly mention that there's also `extension_api.json`, and how the two differ? Especially since there are a lot of existing references to "the API JSON", which can now become ambiguous.

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,344 @@
...
+.. code-block:: shell
+
+    godot --headless --dump-gdextension-interface
+
+.. note::
+
+    The header file is compatible with earlier versions of the header file that were included
+    with Godot 4.5 and earlier, which means it preserves some type-o's in names in order to
+    ensure compatibility. These "legacy names" can be found in the ``legacy_type_name`` key
+    in the JSON file.
```
Bromeon: I've never seen this spelling of "type-o", always "typo" 🙂 does it not come from "**typo**graphical error"?

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,344 @@
...
+- The ``interface`` key, which defines all the function pointers that can be loaded via
+  ``GDExtensionInterfaceGetProcAddress`` function pointer, which is passed to all GDExtensions
+  when they are loaded.
+
+There is a complete `JSON schema <https://github.com/godotengine/godot/blob/master/core/extension/gdextension_interface.schema.json>`__
+included in Godot's source code.
+
+Even though we may add a little bit to this file with each minor release of Godot, we strive
+to **never** change anything in a backwards incompatible way, or remove anything. Every type
+and interface function is labeled with the version of Godot it was introduce in (the ``since``
```
Bromeon:
```suggestion
and interface function is labeled with the version of Godot it was introduced in (the ``since``
```

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,344 @@
...
+key), so you can always use the latest version of the file, and simply refrain from using
+anything in versions of Godot that are newer than the version you are targeting.
```
Bromeon: This slightly contradicts the `format_version` just specified below, which foresees breaking changes (even if unlikely). Maybe it should be referenced here, or maybe it's OK since the `format_version` is explained pretty much afterwards?

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,344 @@
...
+  only one format version (``1``), if we ever change the file format in an incompatible way,
+  we will increment this number. This this *doesn't* reflect the version of the data (so it
+  won't change between Godot versions), only the format of the file. Hopefully, we'll never
+  have to use it, but it allows code generators to error early if they encounter an unexpected
+  value here.
+
+Types
+-----
+
+The "types" section is an array of types that will be used by other types, and the interface
```
Bromeon:
```suggestion
The ``types`` section is an array of types that will be used by other types, and the interface
```
Clarifies that the JSON key is verbatim `types`.

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,344 @@
...
+- ``void``
+- ``int8_t``
+- ``uint8_t``
+- ``int16_t``
+- ``uint16_t``
+- ``int32_t``
+- ``uint32_t``
+- ``int64_t``
+- ``uint64_t``
+- ``size_t``
```
Bromeon: Should the size of `size_t` be specified? Does it vary between Godot-supported platforms?

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,344 @@
...
+
+Regardless of the "kind", all types can have the following keys:
+
+- ``name``: The name of the type, which could be used as a valid C identifier.
+- ``description``: An array of strings documenting the type, where each string is a line of
+  documentation (this format for ``description`` is used through out the JSON file).
+- ``since``: The Godot version that introduced this type.
+- ``deprecated``: An object with its own keys for the Godot version the type was deprecated in
+  (``since``), a message explaining the deprecation (``message``) and optionally a replacement
+  to use instead (``replacement``).
```
Bromeon, line 129: Maybe it should be explicitly mentioned that `kind` is one of the JSON keys.

Should we document that `name`/`kind` are required and the others are optional? Or is it enough if that's part of the JSON schema?

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,344 @@
...
+    }
+
+Interface
+---------
+
+The final section of the JSON file is the list of interface functions, which can be loaded by their
+``name`` using the ``GDExtensionInterfaceGetProcAddress`` function pointer, which is passed to all
+GDExtensions when they are loaded.
+
+They have some of the same keys as types, including ``name``, ``since``, ``deprecated``, and
```
Bromeon: "They" is slightly ambiguous, as you mentioned "GDExtensions" last.




```suggestion
Interface functions have some of the same keys as types, including ``name``, ``since``, ``deprecated``, and
```

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,344 @@
...
+``description``.
+
+And they also have ``return_value`` and ``arguments`` that have the same format as the keys of the
+same name on function types, as described in the previous section.
```
Bromeon: Maybe it's my non-native English, but "as .. of .. on .. as" is a bit heavy to digest here


```suggestion
And they also have ``return_value`` and ``arguments`` that have the same format as the equivalent keys
on function types, as described in the previous section.
```
Or maybe "corresponding" or "same-named" or so. Or put the "as described..." part in parentheses -- many ways to break it up a bit.

Bromeon:
> Any suggestions? It could be "GDExtension interface JSON" maybe?

In the future, there might be a section about the `extension_api.json` as well -- or even if not, it would probably be good to align on two official "prose" terms for these two files...

I'm not sure if "interface JSON" and "API JSON" are clear enough because the "I" in "API" is also "interface". Would a different term make sense here? In Rust I will likely keep using "header", which isn't very precise, but at least distinct enough for all practical purposes.

One defines the actual C interface (or FFI), while the other is more a specification/definition of concrete engine symbols (classes, enums, etc.)... Maybe some inspiration here?

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
Bromeon: (On that topic, see also my other comment https://github.com/godotengine/godot-docs/pull/11551#issuecomment-3678070995)

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
dsnopek: Hm, yeah, this'll be tricky to explain clearly. The `format_version` is for changes to the format of the file, and the comment here is about the interface itself - the content rather than format. I'll see how I can clarify

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
dsnopek: Yes, this is `uint32_t` on 32-bit architecures and `uint64_t` on 64-bit architectures, but reading up on it now, apparently that's not guaranteed in C/C++ and there is some compiler discretion. :-/ Because of this, if we were starting from scratch, it would maybe be best not to use this type...

As we can't remove it, I'll just add a note about its size in the docs, which should be true for any of the C++ compilers we currently support

If we want to be paranoid, we could do a follow-up PR for Godot that guarantees these sizes on the Godot side, just in case there's ever a compiler that does something weird for `size_t`?

dsnopek:
@Bromeon Thanks! I attempted to integrate all of your suggestions in my latest push :-)

> One defines the actual C interface (or FFI), while the other is more a specification/definition of concrete engine symbols (classes, enums, etc.)... Maybe some inspiration here?

Hm, yeah. The `extension_api.json` is really like a "`ClassDB` JSON" - it would have been nice if it were named something closer to that

When writing the note I added about `extension_api.json` in my latest push, I ended up using the words "low-level" and "high-level" to describe them. We could talk about the "Low-level GDExtension interface JSON" and "High-level extension API JSON"? The first word makes the big distinction, while still sort of including the file name in there, so people can find them?

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,352 @@
+.. _doc_gdextension_interface_json_file:
+
+The gdextension_interface.json file
+===================================
+
+The ``gdextension_interface.json`` file is the canonical definition of the C API that
+Godot uses to communicate with GDExtensions.
```
paddy-exe:
```suggestion
The ``gdextension_interface.json`` file is the official definition of the C API that
Godot uses to communicate with GDExtensions.
```

I would replace `canonical` with `official`.

Other options: 
`approved`, 
`recognized`, 
`accepted`

paddy-exe (approved):
Small language nitpick. Otherwise this reads well. Great page addition!
As always thanks for including examples :)

Bromeon:
> Hm, yeah. The `extension_api.json` is really like a "`ClassDB` JSON" - it would have been nice if it were named something closer to that
>
> When writing the note I added about `extension_api.json` in my latest push, I ended up using the words "low-level" and "high-level" to describe them. We could talk about the "Low-level GDExtension interface JSON" and "High-level extension API JSON"? The first word makes the big distinction, while still sort of including the file name in there, so people can find them?

I think it's a good start, but "high vs. low" is still very abstract... All of "ClassDB API", "classes API", "class specification", "class definitions" would be more concrete -- do you not think something along that would be possible? Without renaming the file now, of course. But high/low isn't reflected in the filename, either...

The C API could still be "low-level interface", "C interface", "FFI interface" etc. Something that distinguishes the technical _interface_ from the API JSON, which is more a specification or type database.

But I don't know, naming is hard :sweat_smile:

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
Ivorforce: Canonical is rather niche, so I get where you're coming from.
However, I think the alternatives may not be correct enough. In lieu of a better term, maybe we can just remove the adjective altogether?

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,352 @@
...
+.. code-block:: shell
+
+    godot --headless --dump-gdextension-interface
+
+.. note::
+
+    The header file is compatible with earlier versions of the header file that were included
+    with Godot 4.5 and earlier, which means it preserves some typos in names in order to
+    ensure compatibility. These "legacy names" can be found in the ``legacy_type_name`` key
+    in the JSON file.
```
Ivorforce: This may be a bit early to explain this, as the structure of the file hasn't been explained at all by this point.

tutorials/scripting/gdextension/index.rst
```diff
@@ -24,3 +24,4 @@ articles instead, such as the articles about :ref:`C++ (godot-cpp) <doc_godot_cp
    what_is_gdextension
    gdextension_file
    gdextension_c_example
+   gdextension_interface_json_file
```
Ivorforce, line 26: I'd put this above the C example section.

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
dsnopek: > Other options:
> `approved`,
> `recognized`,
> `accepted`

I was trying to make the distinction between the JSON and the header, where the JSON is the canonical "source of truth" and the header isn't. However, unfortunately, all these other options are true for both the JSON _and_ the header (ie the header is also "approved", "recognized" and "official"), so it would no longer make any distinction.

Maybe:

```suggestion
The ``gdextension_interface.json`` file is the "source of truth" for the C API that
Godot uses to communicate with GDExtensions.
```

Although, that feels kind of awkward to me...

What exactly is the problem with `canonical`? Is it just that it's a less common word?

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
dsnopek: Perhaps. In the beginning of the doc, I was attempting to clarify the relationship between the JSON and the header (since folks who have used GDExtension up to now would be familiar with the header), and so I think it makes sense to say something. Maybe just dropping the very specific sentence about `legacy_type_name`?

Ivorforce:
```
The header file was the previous canonical definition of the interface.
It will continue to work, but we recommend using the json file instead.
```

Something like this maybe?

dsnopek: I've removed that last sentence in my latest push

dsnopek:
> The C API could still be "low-level interface", "C interface", "FFI interface" etc. Something that distinguishes the technical _interface_ from the API JSON, which is more a specification or type database.

In my latest push, I've gone for "C interface" such that the sidebar looks like this:

Selection_418

So, now at least it fits! Hopefully, "C interface" is clear enough. Naming _is_ hard and I feel about equally unsure about every option :-)

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
dsnopek: > The header file was the previous canonical definition of the interface.
> It will continue to work, but we recommend using the json file instead.

I don't know if we should say that using the JSON is recommended. The header is a totally valid option for languages that integrate with C well and don't need to do any fancy processing, so I think we should keep it long-term and not discourage using it (although, for Godot 5 I'd like to wipe out the typos)

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
paddy-exe: > Maybe:
> 
> 
> 
> ```suggestion
> 
> The ``gdextension_interface.json`` file is the "source of truth" for the C API that
> 
> Godot uses to communicate with GDExtensions.
> 
> ```

That would also be fine for me 👍🏻 I think this is more descriptive.

> 
> 
> What exactly is the problem with `canonical`? Is it just that it's a less common word?

It's a less common word (I had to look it up myself and I consider myself rather fluent in English) so I am not sure how well the meaning is being transported.

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,351 @@
...
+
+Functions
+~~~~~~~~~
+
+Functions represent C function pointer types, with a list of arguments and a return type, and should
+follow the same size and alignment requirements as C function pointers.
+
+They have the following members:
+
+- ``return_value``: An object which has a ``type`` (which may include modifiers), and ``description``.
```
AThousandShips:
```suggestion
- ``return_value``: An object which has a ``type`` (which may include modifiers) and ``description``.
```

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,351 @@
...
+
+Structs
+~~~~~~~
+
+Structs represent C ``struct``'s (aka a block of memory made up the given members in order), and should
+follow all the same layout and alignment rules as C structs.
+
+They have only one additional key:
+
+- ``members``: An array of objects which have a ``name``, ``type`` (which may include modifiers) and
```
AThousandShips:
```suggestion
- ``members``: An array of objects which have a ``name``, ``type`` (which may include modifiers), and
```

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,351 @@
...
+~~~~~
+
+Enums are 32-bit integers with a fixed set of possible values. In C, they could be represented
+as an ``enum``.
+
+They have the following keys:
+
+- ``is_bitfield``: If true, this enum is a bitfield, where the enum values can be bitwise OR'd together.
+  It is false by default.
+- ``values``: The array of fixed values for this enum, each with a ``name``, ``value`` and ``description``.
```
AThousandShips:
```suggestion
- ``values``: The array of fixed values for this enum, each with a ``name``, ``value``, and ``description``.
```

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,351 @@
...
+- ``alias``
+- ``struct``
+- ``function``
+
+Regardless of the "kind", all types can have the following keys:
+
+- ``kind`` (required): The "kind" of type.
+- ``name`` (required): The name of the type, which could be used as a valid C identifier.
+- ``description``: An array of strings documenting the type, where each string is a line of
+  documentation (this format for ``description`` is used through out the JSON file).
```
AThousandShips:
```suggestion
  documentation (this format for ``description`` is used throughout the JSON file).
```

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,351 @@
...
+- ``char32_t``
+- ``wchar_t``
+- ``float``
+- ``double``
+
+These correspond to their equivalent C types.
+
+Additionally, types can include modifiers such as:
+
+- ``*`` (ex ``int8_t*``) to indicate a pointer to the type
```
AThousandShips:
```suggestion
- ``*`` (e.g. ``int8_t*``) to indicate a pointer to the type
```

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,351 @@
...
+- ``const`` (ex ``const int8_t*``) to indicate a const type
```
AThousandShips:
```suggestion
- ``const`` (e.g. ``const int8_t*``) to indicate a const type
```

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,351 @@
...
+- The header, which includes some miscellaneous information at the top-level of the JSON file.
+- The ``types`` key, which defines all the types used in the GDExtension interface.
+- The ``interface`` key, which defines all the function pointers that can be loaded via the
+  ``GDExtensionInterfaceGetProcAddress`` function pointer, which is passed to all GDExtensions
+  when they are loaded.
+
+There is a complete `JSON schema <https://github.com/godotengine/godot/blob/master/core/extension/gdextension_interface.schema.json>`__
+included in Godot's source code.
+
+Even though we may add new types and interface function with each minor release of Godot, we
```
AThousandShips:
```suggestion
Even though we may add new types and interface functions with each minor release of Godot, we
```

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+Handles
+~~~~~~~
+
+Handles are pointers to opaque structs. In C, they could be represented as ``void *`` or ``struct{} *``.
+
+They have the following keys:
+
+- ``is_const``: If true, this handle type is to be treated as a "const pointer", meaning it's internal
```
Bromeon:
```suggestion
- ``is_const``: If true, this handle type is to be treated as a "const pointer", meaning its internal
```

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,351 @@
...
+  data will not be changed. It is false by default.
+- ``is_uninitialized``: If true, this handle type is to be treated as pointing to uninitialized memory
+  (which may be initialized using interface functions).
+- ``parent``: The optional name of another handle type, if this handle type is the const or uninitialized
+  version of the parent type.
```
Bromeon: Maybe document if we can expect that this is always present when `is_const || is_uninitialized` is true?

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,351 @@
...
+~~~~~~~
+
+Aliases are alternative names for another type. In C, they could be represented with a ``typedef``.
+
+They have only one additional key:
+
+- ``type``: The type the alias is an alternative name for. It may include modifiers as described above.
+
+These should be represented using the same C type as the type they point to, except when a pointer
+modifier is used (in which case, they should be represented as pointers).
```
Bromeon:
```suggestion
These should be represented using the same C type as the type they refer to, except when a pointer
modifier is used (in which case, they should be represented as pointers).
```

I would not use "point" here as it creates slight ambiguity: `IntPtr` can be understood to point to `int` rather than `int*`.

But I also don't fully understand the semantics of the 2nd part here: can a `type` not refer to a pointer type? Or what do you mean by "represented as pointers" -- there is no "pointer" kind? Maybe it's less confusing to omit that? 🤔

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,351 @@
...
+                "name": "expected",
+                "type": "int32_t"
+            }
+        ]
+    }
+
+Functions
+~~~~~~~~~
+
+Functions represent C function pointer types, with a list of arguments and a return type, and should
```
Bromeon, line 265: Nitpick: "parameters" usually refer to the declaration site, where "arguments" are the call site. But this terminology isn't really used in Godot (which uses "arguments" in most places), so might as well leave it 🙂

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,351 @@
...
+follow the same size and alignment requirements as C function pointers.
+
+They have the following members:
+
+- ``return_value``: An object which has a ``type`` (which may include modifiers), and ``description``.
+  It will always be given, and use ``void`` as the ``type`` if the function returns no value.
```
Bromeon: This is more a question about the format:

> It will always be given, and use ``void`` as the ``type`` if the function returns no value.

Is there a good reason to deviate from `extension_api.json` here, which omits the `return_value` key for `void` return types? Notably, the key is called "return _value_", not "return _type_" 🤔

Bromeon (commented):
Looks really good! I only have a few remarks left, and they're bordering nitpicking territory 🙂

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
AThousandShips:
```suggestion
Even though we may add new types and interface functions with each minor release of Godot, we
```

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
AThousandShips:
```suggestion
- ``*`` (e.g. ``int8_t*``) to indicate a pointer to the type
- ``const`` (e.g. ``const int8_t*``) to indicate a const type
```

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
AThousandShips:
```suggestion
  documentation (this format for ``description`` is used throughout the JSON file).
```

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,351 @@
...
+- ``function``
+
+Regardless of the "kind", all types can have the following keys:
+
+- ``kind`` (required): The "kind" of type.
+- ``name`` (required): The name of the type, which could be used as a valid C identifier.
+- ``description``: An array of strings documenting the type, where each string is a line of
+  documentation (this format for ``description`` is used through out the JSON file).
+- ``deprecated``: An object with its own keys for the Godot version the type was deprecated in
+  (``since``), a message explaining the deprecation (``message``) and optionally a replacement
```
AThousandShips:
```suggestion
  (``since``), a message explaining the deprecation (``message``), and optionally a replacement
```

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
AThousandShips:
```suggestion
- ``values``: The array of fixed values for this enum, each with a ``name``, ``value``, and ``description``.
```

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,351 @@
...
+
+    {
+        "name": "GDExtensionStringNamePtr",
+        "kind": "handle"
+    }
+
+Aliases
+~~~~~~~
+
+Aliases are alternative names for another type. In C, they could be represented with a ``typedef``.
```
AThousandShips:
```suggestion
Aliases are alternative names for a type. In C, they could be represented as a ``typedef``.
```

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,351 @@
...
+    {
+        "name": "GDExtensionInt",
+        "kind": "alias",
+        "type": "int64_t"
+    }
+
+Structs
+~~~~~~~
+
+Structs represent C ``struct``'s (aka a block of memory made up the given members in order), and should
```
AThousandShips:
```suggestion
Structs represent C ``struct``\ s (aka a block of memory made up of the given members in order), and should
```
This is plural not possessive

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
AThousandShips:
```suggestion
- ``members``: An array of objects which have a ``name``, ``type`` (which may include modifiers), and
```

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
AThousandShips:
```suggestion
- ``return_value``: An object which has a ``type`` (which may include modifiers) and ``description``.
```

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,351 @@
...
+
+Functions represent C function pointer types, with a list of arguments and a return type, and should
+follow the same size and alignment requirements as C function pointers.
+
+They have the following members:
+
+- ``return_value``: An object which has a ``type`` (which may include modifiers), and ``description``.
+  It will always be given, and use ``void`` as the ``type`` if the function returns no value.
+- ``arguments``: An array of function arguments which each has a ``type`` (which may include modifiers),
+  a ``name`` and ``description``.
```
AThousandShips:
```suggestion
  a ``name``, and a ``description``.
```
Or
```suggestion
  ``name``, and ``description``.
```

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,351 @@
...
+                "type": "const GDExtensionConstTypePtr*"
+            }
+        ]
+    }
+
+Interface
+---------
+
+The ``interface`` section of the JSON file is the list of interface functions, which can be loaded
+by their ``name`` using the ``GDExtensionInterfaceGetProcAddress`` function pointer, which is
```
AThousandShips:
```suggestion
by their ``name``s using the ``GDExtensionInterfaceGetProcAddress`` function pointer, which is
```
Or
```suggestion
by ``name`` using the ``GDExtensionInterfaceGetProcAddress`` function pointer, which is
```

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,351 @@
...
+- ``double``
+
+These correspond to their equivalent C types.
+
+Additionally, types can include modifiers such as:
+
+- ``*`` (ex ``int8_t*``) to indicate a pointer to the type
+- ``const`` (ex ``const int8_t*``) to indicate a const type
+
+Each type defined in the JSON file falls into one of 5 "kind"s:
```
AThousandShips:
```suggestion
Each type defined in the JSON file falls into one of 5 "kinds":
```

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
```diff
@@ -0,0 +1,351 @@
...
+
+- ``enum``
+- ``handle``
+- ``alias``
+- ``struct``
+- ``function``
+
+Regardless of the "kind", all types can have the following keys:
+
+- ``kind`` (required): The "kind" of type.
```
AThousandShips:
```suggestion
- ``kind`` (required): The "kind" of the type.
```
I'd say

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
Bromeon: Both is correct in English grammar -- does Godot have a convention regarding the [Oxford comma](https://en.wikipedia.org/wiki/Serial_comma)? :wink:

(Personally I prefer the flow without the last comma, unless it's a longer part of a sentence)...

AThousandShips: [Yes](https://contributing.godotengine.org/en/latest/documentation/guidelines/docs_writing_guidelines.html#use-the-oxford-comma-to-enumerate-anything)

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
dsnopek: Yeah, I think we should stick with "arguments" given that the key in JSON is also called "arguments"

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
dsnopek: Hm, well, the implementation was easier :-)

We already have "void" as a valid type (so that we can have "void *" and other modifiers) and simply always requiring this is simpler than having to check if it exists

I can look at what it would take to change the parsers. We'd then have to decide if we want to disallow explicitly specifying a return type of "void" (and force it to be omitted in that case), which would need to be reflected in the parsers and the schema

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
dsnopek: "The kind of the type" sounds really wrong to me, but I can't explain why. It logically would make sense because we have "the name of the type", and even "the type of the type" would sound fine, but for some reason with the formulation "the kind of" it just sounds wrong to be followed by "the" :-/

AThousandShips: I'd say `the type "kind"` works better, but to me it reads like "the color of car" or "the height of building"

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
AThousandShips: The pause here is not right I'd say, it's not tying together two complete sentences it's part of an enumeration, and then the comma is not used for two elements, only three or more

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
AThousandShips: I'd say this needs to be changed, it's several functions each with their own names, it'd be like "several people arrived and they parked their car"

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
dsnopek: > but to me it reads like "the color of car" or "the height of building"

Yeah, I see what you mean - in those cases you definitely need some article ("the" or "a")

But for some reason, in my mind, "kind of" cannot be followed by "the". Of course, there are several varieties of English, and maybe this is a peculiarity of the one I speak? It could be useful to get an opinion from another native English speaker

AThousandShips: Alternatively `The "kind" type this type is"

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
dsnopek: I accepted your 2nd suggestion in my most recent push. Or are you saying this needs further changes?

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
dsnopek: Oh, I think I misread the original comment here as adding a comma not removing one - I'll update this in a moment

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
dsnopek: Ack, it looks like that change actually didn't end up in that push - I probably forgot to save before committing. It's in there now :-)

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
Bromeon: It's probably because "kind of" is an idiom on its own, and when followed by a noun, doesn't have an article ("that kind of house", "this kind of food", ...)

Not sure how this would translate to this situation, where "kind" is meant as "category", outside the "kind of" idiom :thinking:

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
dsnopek: Here's a PR that changes the format to omit the `return_value` for `void` functions: https://github.com/godotengine/godot/pull/114684

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
dsnopek: > > What exactly is the problem with `canonical`? Is it just that it's a less common word?
> 
> It's a less common word (I had to look it up myself and I consider myself rather fluent in English) so I am not sure how well the meaning is being transported.

Even if it's less common in the language as a whole, I feel like it's pretty commonly used in technical specifications. For example, a quick search finds it 10 times in the [latest Vulkan spec](https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html).

That said, if everyone is fine with "source of truth", we can go with that - it's slightly more conversational than I'd usually use in docs, but it's definitely something I'd say out loud.

tutorials/scripting/gdextension/gdextension_interface_json_file.rst
akien-mga: How about `The type's "kind".`?

dsnopek: `The type's "kind".` would be fine, so I guess I'll change it to that. However, I still think the original version (`The "kind" of type.`) sounds the best

Bromeon (approved):

akien-mga:
Thanks!

PR #11552, opened 2025-12-20
Correct note on single-line shader documentation comments

ProgrammerOnCoffee:
Corrects the shading language documentation where it incorrectly states that single-line documentation comments aren't supported. From my testing, single-line comments have been supported since 4.3 when shader documentation comments were introduced.

### DIFF
--- a/tutorials/shaders/shader_reference/shading_language.rst
+++ b/tutorials/shaders/shader_reference/shading_language.rst
@@ -124,8 +124,9 @@ using ``//`` for single-line comments and ``/* */`` for multi-line comments:
 Additionally, you can use documentation comments that are displayed in the
 inspector when hovering a shader parameter. Documentation comments are currently
 only supported when placed immediately above a ``uniform`` declaration. These
-documentation comments only support the **multiline** comment syntax and must use
-**two** leading asterisks (``/**``) instead of just one (``/*``):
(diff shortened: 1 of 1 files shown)


ProgrammerOnCoffee:
Ah, my mistake. That makes sense.

In that case, should the comment be clarified somehow?

tutorials/shaders/shader_reference/shading_language.rst
```diff
@@ -124,8 +124,8 @@ using ``//`` for single-line comments and ``/* */`` for multi-line comments:
 Additionally, you can use documentation comments that are displayed in the
 inspector when hovering a shader parameter. Documentation comments are currently
 only supported when placed immediately above a ``uniform`` declaration. These
-documentation comments only support the **multiline** comment syntax and must use
-**two** leading asterisks (``/**``) instead of just one (``/*``):
+documentation comments must use **two** leading asterisks (``/**``) instead of
+just one (``/*``):
```
Calinou: This is how I would clarify it:

```suggestion
documentation comments only support the **multiline** comment syntax
(even if used on a single line) and must use **two** leading asterisks
(``/**``) instead of just one (``/*``):
```

mhilbrunner:
Thanks for contributing, merged!

PR #11558, opened 2025-12-22
Update snipping instructions to include macOS key binding

sherifnosseir:
Update snipping instructions to include Cmd keyboard as the macOS key binding; and updates the phrasing to be platform agnostic after the key bindings are specified.


mhilbrunner:
Thank you for contributing and congrats on your first merged PR for Godot :)

PR #11568, opened 2025-12-28
Fix BBCode Matrix effect example

DevPoodle:
In #5982, when the BBCode tutorial was updated after CharFX API changes, the Matrix effect example was broken. This fixes it by first converting the glyph index to a character, so that when it gets converted back into a glyph index, it's the right one. Here's a video showing the before and after:

https://github.com/user-attachments/assets/1eb96440-08c6-46d6-bd9f-d1e3a64ddb9b


Calinou (approved):
Looks good to me.

@bruvzg Does this look correct to you?

bruvzg (approved):

mhilbrunner (approved):

mhilbrunner:
Thank you!

PR #11569, opened 2025-12-29
Document functions used for changing scenes manually

Cyberhead21:
The functions `CanvasItem.hide()` and `CanvasItem.show()` were documented for hiding the existing scenes and `Node.add_child(Node)` for reattaching the scene that was detached from a tree, as they aren't documented and confuse beginners who might not otherwise have much experience with Godot.


mhilbrunner:
Merged. Thanks and congrats on your first merged contribution! Sorry this took a bit. :)

PR #11570, opened 2025-12-29
Update learning_new_features.rst: grammatical error, missing "of"

lukkasROCKUS:
Very simple grammatical fix. Just adding an "of".


skyace65:
This doesn't fix the grammar. The current text, "each Godot class's available functions" is correct.

PR #11574, opened 2025-12-31
Update class_bitmap.rst

FilipRank:
Fix bad grammar


skyace65:
Hi, we appreciate the effort, however class reference pages need to be updated in the engine repository, the class files there get synced with this one. This is the file you want to edit: https://github.com/godotengine/godot/blob/master/doc/classes/BitMap.xml

PR #11557, opened 2025-12-21
Update documentation for dr_mp3 change

skyace65:
minimp3 was replaced with dr_mp3.

Engine PR: https://github.com/godotengine/godot/pull/96547

PR #11585, opened 2026-01-05
Fix change about MP3 module

DeeJayLSP:
#11557 changed instances of the `minimp3` module into `dr_mp3`.

While the `minimp3` library was in fact replaced with `dr_mp3`, the module was renamed to `mp3`, not `dr_mp3`, to keep parity with other modules' names being about support for something and not their dependency.

Using `module_dr_mp3_enabled=no` will do nothing. The correct is `module_mp3_enabled=no`.


mhilbrunner:
Thanks for noticing and taking the time for the fixup :)

PR #11587, opened 2026-01-06
Revise mold installation steps

wjt:
Previously the documentation stated that mold is not widely packaged as of 2023. It is now 2026 and it is more widely packaged. Update this section.

### DIFF
--- a/engine_details/development/compiling/compiling_for_linuxbsd.rst
+++ b/engine_details/development/compiling/compiling_for_linuxbsd.rst
@@ -547,20 +547,27 @@ Using mold for faster development
 For even faster linking compared to LLD, you can use `mold <https://github.com/rui314/mold>`__.
 mold can be used with either GCC or Clang.
 
-As of January 2023, mold is not readily available in Linux distribution
-repositories, so you will have to install its binaries manually.
+.. tabs::
+    .. tab:: Debian/Ubuntu
 
-- Download mold binaries from its `releases page <https://github.com/rui314/mold/releases/latest>`__.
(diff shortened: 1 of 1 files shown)


Calinou (commented):
I think you can also add Arch Linux here (`pacman -Sy --noconfirm --needed mold`).

engine_details/development/compiling/compiling_for_linuxbsd.rst
```diff
@@ -547,20 +547,27 @@ Using mold for faster development
...
+    .. tab:: Arch Linux
+
+        ::
 
-    PATH="$HOME/.local/share/mold/bin:$PATH"
+            pacman -Sy --noconfirm --needed mold
 
-- Open a new terminal (or run ``source "$HOME/.bash_profile"``),
-  then use the following SCons command when compiling Godot:
+Once mold is installed, use the following SCons command when compiling Godot:
```
wjt, line 570: I have added Arch Linux to this section.

I have to say that I am not sure there is really any value in listing out the install instruction for every distro - if you have reached this point in the document you presumably know how to install the `mold` package with your distro's package manager. Maybe we can just say instead:

```rst
As of January 2026, a ``mold`` package is available in many distributions'
package managers. Once mold is installed, use the following SCons command when
compiling Godot:
```

skyace65:
I'm ok with distro specific instructions, you never know if it's someone's first time trying to compile something, or download software via command line, especially with how common flathub is these days.

Calinou (approved):

mhilbrunner:
Thank you! Merged.

PR #11596, opened 2026-01-10
Clarify meaning of "attached to the screen" for Viewport use_hdr_2d

Arjun-explores-Arjun-learn:
This PR clarifies what "attached to the screen" means in the Viewport
documentation by distinguishing on-screen viewports (like the root
Viewport) from viewports that render to a texture or are used off-screen.

No behavioral change, documentation-only clarification.


AThousandShips:

Arjun-singh-rawatt:
Thanks for the clarification.
I understand the documentation workflow better now and will make sure to update the correct source files in future contributions.

PR #11597, opened 2026-01-10
Change "transparent" to "translucent" to clarify red dot's colour.

asyync1024:
This is my first PR ever! So please ignore any ill-etiquette.

Being transparent means fully see through, and the red dot appearing over the panel is not fully see through, as if it was, it would not even be visible. It's translucent, as it's only partially see through.

### DIFF
--- a/tutorials/scripting/debug/debugger_panel.rst
+++ b/tutorials/scripting/debug/debugger_panel.rst
@@ -26,7 +26,7 @@ debugger broke on.
 
     You can create a breakpoint by clicking the gutter in the left of the script
     editor (on the left of the line numbers). When hovering this gutter, you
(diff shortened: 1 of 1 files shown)


AThousandShips:
I don't think we use this term generally, also "translucent" implies you can't see what's behind without distortion

skyace65:
I looked up the differences and I'm in agreement with AThousandShips. This change doesn't seem correct as translucency involves distortion.

asyync1024:
Correct, I thought that transparent meant that the object's all properties are basically the same, so I concluded that colour is not the same, therefore it must not be transparent, so it should be translucent.

Sorry for wasting maintainer time!

PR #11602, opened 2026-01-12
Added a note with link to an Inkscape tools for Godot

OS-of-S:
I've set up a [convenient pipeline for creating editor graphics](https://github.com/OS-of-S/Inkscape-Tools-for-Godot-Editor-Designing), and I'd like to add a link to it in the documentation. It's not so easy to use the bare hex-codes for colors provided by documentation right now. So I believe my tools would lower the barrier for beginners and would help to improve the appearance of tools they create.

### DIFF
--- a/engine_details/editor/creating_icons.rst
+++ b/engine_details/editor/creating_icons.rst
@@ -48,6 +48,16 @@ Otherwise, your icon may become difficult to read on a light background.
 		1. :ui:`Import > Import As > Texture2D`
 		2. Set ``editor/convert_colors_with_editor_theme`` to ``true``
 
+.. note::
+
+	To simplify designing process, you can use tools from this
+	`unofficial project <https://github.com/OS-of-S/Inkscape-Tools-for-Godot-Editor-Designing>`__,
+	that provides Godot palettes, which can be imported into Inkscape,
+	and an Inkscape extension to check its look for different themes.
+	
+	But priority for relevance remains with official
+	`color mappings file <https://github.com/godotengine/godot/blob/master/editor/themes/editor_color_map.cpp>`__.
+
 Icon optimization
 ~~~~~~~~~~~~~~~~~
 
--- a/engine_details/editor/creating_icons.rst
+++ b/engine_details/editor/creating_icons.rst
@@ -50,7 +50,7 @@ Otherwise, your icon may become difficult to read on a light background.
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


engine_details/editor/creating_icons.rst
```diff
@@ -48,6 +48,16 @@ Otherwise, your icon may become difficult to read on a light background.
 		1. :ui:`Import > Import As > Texture2D`
 		2. Set ``editor/convert_colors_with_editor_theme`` to ``true``
 
+.. note::
+
+	To simplify designing process, you can use tools from this
```
AThousandShips:
```suggestion
	To simplify the design process, you can use tools from this
```

engine_details/editor/creating_icons.rst
```diff
@@ -48,6 +48,16 @@ Otherwise, your icon may become difficult to read on a light background.
...
+	`unofficial project <https://github.com/OS-of-S/Inkscape-Tools-for-Godot-Editor-Designing>`__,
+	that provides Godot palettes, which can be imported into Inkscape,
+	and an Inkscape extension to check its look for different themes.
+	
+	But priority for relevance remains with official
```
AThousandShips, line 58:
```suggestion
	But priority for relevance remains with the official
```
Though I don't understand what this sentence means

OS-of-S: > Though I don't understand what this sentence means

It means, that since my project unofficial,  it may lag behind Godot updates and color scheme palette changes. I can't guarantee the relevance.

Anyway, I don't think developers will change already existing colors, at least - at the _dark theme_ palette. It would break icons in a lot of tools and plugins at the same time if they would. But they could probably add new colors or change colors at the _light theme_ palette.

I'll try to track and follow the changes in the original sources, but it will lag behind anyway. So I think it's honest to left a little remark with a link to the [color mappings file](https://github.com/godotengine/godot/blob/master/editor/themes/editor_color_map.cpp) mentioned in documentation above as reliable source.

OS-of-S: On the other hand, my project have a remark at the README.md file: _Relevant for a Godot 4.5._ So maybe I could erase the whole last sentence from my pull request. What do you think?

skyace65:
Going to close this PR. For right now at least we'd like to avoid adding in any more links to community projects in the docs. But thank you for making this PR.

PR #11615, opened 2026-01-15
Update step by step images for 4.6

Piralein:
- Image updates
- Updated some dock names, because of the node -> signals/groups split.
- Updated the game opens in "new window" to "game view", at least for me it was the default for new projects.


getting_started/step_by_step/img/nodes_and_scenes_11_final_result.webp
skyace65, line 1: Is game view embedded by default in 4.6? It's not for me, but I know that existing defaults remain as is between engine updates

Calinou: IIRC, it's on by default, but in a floating window.

Piralein: so the second option should also be checked?
image
how can I enable it, it doesn't let me change it.

Piralein: Found it, I needed to enable multi window.
I updated the image and removed the text change for `nodes_and_scenes.rst`

mhilbrunner:
Thank you! Merged. 🎉

PR #11616, opened 2026-01-15
Remove "outdated article" banner on index pages

Calinou:
The warnings are relevant on the pages themselves, but not so much on the pages listing them (especially since in most of these index pages, not all pages are outdated).


skyace65:
Thanks!

PR #11618, opened 2026-01-16
Add "Upgrading from Godot 4.5 to Godot 4.6" page

raulsntos:
This documents all the breaking changes in 4.6, like we did for previous versions:

- https://github.com/godotengine/godot-docs/pull/11223

_Updated as of commit godotengine/godot@1549d51865109a27216f2409a35f2c6dab7bb4d1_

_Preview: https://github.com/raulsntos/godot-docs/blob/breaking_changes_4.6/tutorials/migrating/upgrading_to_godot_4.6.rst_

### DIFF
--- a/tutorials/migrating/upgrading_to_godot_4.6.rst
+++ b/tutorials/migrating/upgrading_to_godot_4.6.rst
@@ -1,0 +1,287 @@
+.. _doc_upgrading_to_godot_4.6:
+
+Upgrading from Godot 4.5 to Godot 4.6
+=====================================
+
+For most games and apps made with 4.5 it should be relatively safe to migrate to 4.6.
+This page intends to cover everything you need to pay attention to when migrating
+your project.
+
+Breaking changes
+----------------
+
+If you are migrating from 4.5 to 4.6, the breaking changes listed here might
+affect you. Changes are grouped by areas/systems.
+
+This article indicates whether each breaking change affects GDScript and whether
+the C# breaking change is *binary compatible* or *source compatible*:
+
+- **Binary compatible** - Existing binaries will load and execute successfully without
+  recompilation, and the run-time behavior won't change.
+- **Source compatible** - Source code will compile successfully without changes when
+  upgrading Godot.
+
+Core
+~~~~
+
+========================================================================================================================  ===================  ====================  ====================  ============
+Change                                                                                                                    GDScript Compatible  C# Binary Compatible  C# Source Compatible  Introduced
+========================================================================================================================  ===================  ====================  ====================  ============
+**FileAccess**
+Method ``create_temp`` changes ``mode_flags`` parameter type from ``int`` to ``FileAccess.ModeFlags``                     |✔️|                 |✔️ with compat|      |✔️ with compat|      `GH-114053`_
+Method ``get_as_text`` removes ``skip_cr`` parameter                                                                      |✔️|                 |✔️ with compat|      |✔️ with compat|      `GH-110867`_
+**Performance**
+Method ``add_custom_monitor`` adds a new ``type`` optional parameter                                                      |✔️|                 |✔️ with compat|      |✔️ with compat|      `GH-110433`_
+========================================================================================================================  ===================  ====================  ====================  ============
+
+Animation
+~~~~~~~~~
+
+========================================================================================================================  ===================  ====================  ====================  ============
+Change                                                                                                                    GDScript Compatible  C# Binary Compatible  C# Source Compatible  Introduced
+========================================================================================================================  ===================  ====================  ====================  ============
+**AnimationPlayer**
+Property ``assigned_animation`` changes type from ``String`` to ``StringName``                                            |✔️|                 |❌|                  |❌|                  `GH-110767`_
+Property ``autoplay`` changes type from ``String`` to ``StringName``                                                      |✔️|                 |❌|                  |❌|                  `GH-110767`_
+Property ``current_animation`` changes type from ``String`` to ``StringName``                                             |✔️|                 |❌|                  |❌|                  `GH-110767`_
+Method ``get_queue`` changes return type from ``PackedStringArray`` to ``StringName[]``                                   |✔️|                 |❌|                  |❌|                  `GH-110767`_
+Signal ``current_animation_changed`` changes ``name`` parameter type from ``String`` to ``StringName``                    |✔️|                 |❌|                  |❌|                  `GH-110767`_
+========================================================================================================================  ===================  ====================  ====================  ============
+
+3D
+~~
+
+=================================================================================================================================================================  ===================  ====================  ====================  ============
+Change                                                                                                                                                             GDScript Compatible  C# Binary Compatible  C# Source Compatible  Introduced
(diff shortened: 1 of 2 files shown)


tutorials/migrating/upgrading_to_godot_4.6.rst
```diff
@@ -0,0 +1,208 @@
...
+========================================================================================================================  ===================  ====================  ====================  ============
+**OpenXRExtensionWrapper**
+Method ``_get_requested_extensions`` adds a new ``xr_version`` parameter                                                  |❌|                 |❌|                  |❌|                  `GH-109302`_
+Method ``_set_instance_create_info_and_get_next_pointer`` adds  a new ``xr_version`` parameter                            N/A                  N/A                   N/A                   `GH-109302`_
+========================================================================================================================  ===================  ====================  ====================  ============
+
+.. note::
+
+    The ``OpenXRExtensionWrapper`` type is intended to be subclassed from GDExtensions. The method ``_set_instance_create_info_and_get_next_pointer``
+    has a ``void*`` parameter so it's not exposed to scripting.
```
dsnopek, line 124: `_get_requested_extensions` is also only intended to be used from GDExtension: you return a `Dictionary` which has `bool *` (cast to int) as the keys, so it also can't really be used from scripting

raulsntos: I meant it as the whole class is only intended to be overridden from GDExtension. The `_get_requested_extensions` method still shows up in the C# API because we're only filtering when the method signature contains pointer types and the return type is an untyped `Dictionary` so we don't know.

I wasn't sure how to include these changes since the tables only mention compatibility for scripting languages, and this is only relevant to GDExtension. I thought the note would help, but suggestions to change/improve it are welcomed.

Calinou (commented):
Current changes look good to me.

However, we should add a section on rendering changes related to HDR 2D and glow. I'll let @allenwp handle this one then I'll check back for a review.

We should also mention two default changes in new projects:

- Direct3D 12 is the default on Windows
- Jolt is the default 3D physics engine
 

Existing projects aren't affected by these changes. The procedure to apply the new settings should be documented, in case you want your project to behave as if it was created in 4.6.

allenwp:
`Environment` glow effect has changed in the following ways:

**Changes to default values:**
4.5.1 | 4.6
--|--
glow_levels = { 0.0, 0.0, 1.0, 0.0, 1.0, 0.0, 0.0 } | glow_levels = { 0.0, 0.8, 0.4, 0.1, 0.0, 0.0, 0.0 };
glow_intensity = 1.0 | glow_intensity = 0.3
glow_blend_mode = Soft Light | glow_blend_mode = Screen

**Changes to behaviour:**
Soft Light blend mode now always behaves as it did previously with `use_hdr_2d`, regardless of the Viewport's `use_hdr_2d` setting.

I'm not sure if the last point regarding the change of behaviour should be mentioned because I don't think rendering behaviour changes are normally documented in these upgrade pages. Additionally, there is nothing a user can do to get the old 4.5.1 behaviour, so that makes it seem like it's not valuable to mention.

tutorials/migrating/upgrading_to_godot_4.6.rst
```diff
@@ -0,0 +1,208 @@
...
+    The source sets configuration for Android export templates has been updated to match
+    the default Android Studio project structure (`GH-110829`_). This affects the directory
+    layout of the Android project:
+
+    - Files previously in ``platform/android/java/app/src/`` are now in ``platform/android/java/app/src/main/java/``
+    - Files previously in ``platform/android/java/lib/src/`` are now in ``platform/android/java/lib/src/main/java/``
+    - Resource and manifest files have been moved to ``src/main/`` subdirectories
+
+    For example, ``GodotApp.java`` moved from ``src/com/godot/game/GodotApp.java`` to
+    ``src/main/java/com/godot/game/GodotApp.java``.
```
syntaxerror247:
```suggestion
    - Files previously in ``[Project root]/android/build/src/`` are now in ``[Project root]/android/build/src/main/java/``
    - Android manifest file and assets directory have been moved to ``src/main/`` subdirectories

    For example, ``GodotApp.java`` moved from ``src/com/godot/game/GodotApp.java`` to
    ``src/main/java/com/godot/game/GodotApp.java``.
```

Since this is for the end user, I updated the path to reflect where they would see the changes, rather than the godot source file structure.

tutorials/migrating/upgrading_to_godot_4.6.rst
```diff
@@ -0,0 +1,275 @@
...
+
+Rendering
+~~~~~~~~~
+
+.. note::
+
+    Soft Light blend mode now always behaves as it did previously with ``use_hdr_2d``, regardless of
+    the Viewport's ``use_hdr_2d`` setting (`GH-109971`_).
```
raulsntos, line 192: @allenwp I think it's worth documenting any change in behavior, so at least users are aware of it, even if they can't do anything to revert it. So I added a note with the changes you mentioned to `use_hdr_2d`. Let me know if you want to change the text or if I got the wrong PR.

raulsntos (commented):
Added a note about the `use_hdr_2d` changes and updated the Android note as requested.

Also added a _Changed defaults_ section:
- https://github.com/godotengine/godot-proposals/issues/12763

---

@Calinou

> Direct3D 12 is the default on Windows

Are you sure DirectX 12 is the default driver on Windows now? The `rendering/rendering_device/driver.windows` property still has the default value `vulkan`[^1].

> Jolt is the default 3D physics engine

The `physics/3d/physics_engine` documentation still says `DEFAULT` is equivalent to `GodotPhysics3D` so that should be updated.

[^1]: This was last changed in 4.4: https://github.com/godotengine/godot/pull/103197

tutorials/migrating/upgrading_to_godot_4.6.rst
raulsntos: @dsnopek What do we want to do here? We could just remove the OpenXR table from the list, since the only methods that changed are only relevant to GDExtension, and GDExtension has a virtual method compatibility system so I assume this table is not really useful. Or do you want me to change the note instead?

allenwp:
> Are you sure DirectX 12 is the default driver on Windows now?

D3d12 is now the driver that is used for new projects created with 4.6. The default remains Vulkan. (New projects in 4.6 don't use the default driver on Windows anymore.)

I believe this is the same for physics with Jolt, as Calinou wrote.

Calinou:
> I believe this is the same for physics with Jolt, as Calinou wrote.

This system was implemented in https://github.com/godotengine/godot/pull/105737, and the D3D12 default uses the same mechanism. (This is run on all platforms, so it will apply even if the editor is running on another platform at the point of creation.)

raulsntos:
@Calinou Thanks for linking that PR, I see how these work now. A bit unfortunate that it doesn't change in the XML docs because then I can't easily find these changes with a script/tool.

mhilbrunner (approved):

mhilbrunner:
Thank you! Merged.

PR #11619, opened 2026-01-16
Fix redirects in pages with other base paths than the default (`/en/stable`).

Ivorforce:
Not sure how this slipped us by, but redirects are currently broken.
For example:

https://docs.godotengine.org/en/latest/engine_details/development/debugging/using_cpp_profilers.html

redirects to:

https://docs.godotengine.org/engine_details/development/profiling/index.html

instead of:

https://docs.godotengine.org/en/latest/engine_details/development/profiling/index.html

I don't have a good way to test this locally, since the local version does not have `/en/latest` etc in the pathname. But logically this should be correct.

- Follow-up to https://github.com/godotengine/godot-docs/pull/11304

### DIFF
--- a/404.rst
+++ b/404.rst
@@ -38,7 +38,7 @@ Page not found
 
     const currentPathSegments = window.location.pathname.split('/').filter(segment => segment !== '');
     // Use the base path (e.g. "/en/latest") when available.
-    const currentBasePath = (currentPathSegments.length >= 2) ? ("/" + currentPathSegments.slice(0, 2).join("/")) : "/";
+    const currentBasePath = (currentPathSegments.length >= 2) ? ("/" + currentPathSegments.slice(0, 2).join("/")) : "";
 
     fetch(currentBasePath + "/_static/redirects.csv")
       .then(response => response.text())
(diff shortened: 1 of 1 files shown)


404.rst
```diff
@@ -54,7 +54,7 @@ Page not found
               if (to.startsWith('https://')) {
                 window.location.replace(to);
               } else {
-                const newUrl = window.location.href.replace(window.location.pathname, to);
+                const newUrl = window.location.href.replace(window.location.pathname, currentBasePath + to);
```
Ivorforce, line 57: I guess an alternative fix could be to replace `from` instead of `window.location.pathname`. Maybe that's more stable?

mhilbrunner:
Lets test then, even if we may end up further improving this, redirect loops/broken links are pretty bad.

Another example pointed out by Rémi:
> https://docs.godotengine.org/en/4.5/contributing/development/compiling/compiling_with_script_encryption_key.html
> This redirects to
> https://docs.godotengine.org/engine_details/development/compiling/compiling_with_script_encryption_key.html
> (loses the en/4.5/ domain)

mhilbrunner:
Thanks!

mhilbrunner:
@Ivorforce With this change deployed, your initial case does appear to be fixed :) The other one mentioned above still seems to loop though.

Ivorforce:
Cherry-picked to 4.5 in https://github.com/godotengine/godot-docs/pull/11639

PR #11624, opened 2026-01-17
Edit per-instance texture array workaround

FyiurAmron:
Follow-up on https://github.com/godotengine/godot-docs/pull/11350 due to research from https://github.com/godotengine/godot-docs/issues/11331 - this workaround is only needed for very specific cases and is largely not needed for up-to-date setups. As such, the docs should describe precisely when it's needed, so that users in general don't use it when it's not strictly necessary for them.

I also reduced the amount of branches in the example switch from 4 to 2 - it showcases the idiom in exactly the same way, but doesn't fluff the docs.

CC @Calinou @mhilbrunner @JellyBoonz
 (I wanted to ping everybody directly involved in the original PR, sorry if it's excessive)

### DIFF
--- a/tutorials/shaders/shader_reference/shading_language.rst
+++ b/tutorials/shaders/shader_reference/shading_language.rst
@@ -1112,21 +1112,22 @@ method on a node that inherits from :ref:`class_GeometryInstance3D`:
 
 When using per-instance uniforms, there are some restrictions you should be aware of:
 
-- **Per-instance uniforms do not support textures or arrays**, only regular scalar and vector types.
+- **Per-instance uniforms do not support textures or arrays**, only regular scalar and
+  vector types. As a workaround, you can pass a texture array as a regular uniform,
+  then pass the index of the texture to be drawn using a per-instance uniform.
 
 .. note::
 
-    Due to GLSL limitations, you cannot directly index a texture array
-    using a per-instance uniform. Sampler arrays can only be indexed by
-    compile-time constant expressions.
+    In GLSL versions before 4.0 (i.e. GLSL 3.3 and lower), you cannot directly index
+    a texture array using a per-instance uniform, as sampler arrays can only be indexed by
+    compile-time constant expressions. This affects shaders compiled with the Compatibility
+    renderer.
 
-    As a workaround, pass a texture array as a regular uniform and the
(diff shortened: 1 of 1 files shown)


tutorials/shaders/shader_reference/shading_language.rst
```diff
@@ -1112,21 +1112,22 @@ method on a node that inherits from :ref:`class_GeometryInstance3D`:
...
+  vector types. As a workaround, you can pass a texture array as a regular uniform,
+  then pass the index of the texture to be drawn using a per-instance uniform.
 
 .. note::
 
-    Due to GLSL limitations, you cannot directly index a texture array
-    using a per-instance uniform. Sampler arrays can only be indexed by
-    compile-time constant expressions.
+    In GLSL versions before 4.0 (i.e. GLSL 3.3 and lower), you cannot directly index
+    a texture arrar using a per-instance uniform, as sampler arrays can only be indexed by
```
Calinou:
```suggestion
    a texture array using a per-instance uniform, as sampler arrays can only be indexed by
```

Calinou:
I am not sure we should mention the fact that it *might* work on Compatibility, as this should not be relied upon (e.g. shaders will suddenly break when exporting to mobile/web platforms). I would rather mention it as if it never worked. Most projects that use Compatibility target mobile/web platforms after all, so we want to avoid surprises when exporting.

tutorials/shaders/shader_reference/shading_language.rst
```diff
@@ -1112,21 +1112,22 @@ method on a node that inherits from :ref:`class_GeometryInstance3D`:
...
+    compile-time constant expressions. This can affect shaders compiled with the Compatibility
+    renderer on some particular OS/GPU combinations.
```
FyiurAmron:
```suggestion
    compile-time constant expressions. This affects shaders compiled with the Compatibility
    renderer.
```

@Calinou did I get it right vs https://github.com/godotengine/godot-docs/pull/11624#issuecomment-3775018019 with this change?

Calinou: Yes, that looks fine to me.

mhilbrunner:
Thank you, and thanks Calinou for the review! Merged.

PR #11625, opened 2026-01-17
Update "Editor Introduction" images for 4.6

Piralein:
- update images


tutorials/editor/img/editor_ui_script_editor_replace_all.webp
hsvfan-jan, line 1: Looks like the new theme doesn't apply a darker background here, should an issue be opened in the main repo for this? It looks a bit weird

mhilbrunner (approved):

mhilbrunner:
As discussed on RC, this goes in now, further followup work can be in additional PRs. Thank you so much!

tutorials/editor/img/editor_ui_script_editor_replace_all.webp
Piralein: You can open an issue if you think this should be different, I don't really have an opinion on it, as I haven't used this part with the new theme as much for now.
Maybe its an oversight, maybe not^^"

PR #11630, opened 2026-01-18
Update Audio Buses page images for 4.6

skyace65:
does what the title says. All images have gone through Squoosh compression. This also updates the one image on the effects page since it's the same as what's on the buses page.


mhilbrunner (approved):

mhilbrunner:
Thank you!

PR #11631, opened 2026-01-18
Update data preferences code block to match style guide

JamesSkemp:
Swaps the order of the class_name and extends to match the GDScript style guide.

https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html#code-order


mhilbrunner:
Thank you!

PR #11633, opened 2026-01-18
Split "module API" articles into their own engine details section, away from "architecture"

Ivorforce:
This PR splits the "Engine architecture" section into two separate ones:

| Before | After |
|--------|--------|
| SCR-20260118-qmoa | SCR-20260118-qojb SCR-20260118-qoau |

I did need to write an intro for the new section. This is the only new content in this PR.

## Motivation / Background

The current "Engine architecture" section is pretty big, containing 15 subsections.

I think there are several problems with this:

- This many articles makes it hard to navigate and understand this section.
    - Especially, some of the articles are not very cohesively related.
- Engine contributors usually only need the first half of the articles. Only GDExtension and module authors need both sections.
- It may be hard to find out about "modules" by browsing the docs, because their existence is "hidden" in the architecture.
- Extending engine functionality through its APIs doesn't fit the term "Engine architecture" very well, which may make it hard to find.

There appears to be a "natural split" to the articles already:

- A section which describes how the code _is_ and _works_ (called "architecture" or "Getting started with the source code")
- A section which describes how the code _can be used_, especially without modifying it directly (called "modules" or "Extending Godot by modifying its source code")

All this PR does is "embrace" this existing split, by integrating it into the navigation.


mhilbrunner:
Merging, thanks! 🎉

PR #11634, opened 2026-01-19
Clean up "GDScript" extension use from conf.py.

Ivorforce:
Came across this code from https://github.com/godotengine/godot-contributing-docs/pull/47.

SPHINX_NO_GDSCRIPT was introduced for CI, but appears to be unused now.

The 'include trick' need not be done (and should not), since we can just run the code in the extension implementation.
(i'm not even sure if the `lexers` insertion is needed, but I chose to be agnostic for it)

### DIFF
--- a/_extensions/gdscript.py
+++ b/_extensions/gdscript.py
@@ -1578,7 +1578,10 @@ class GDScriptLexer(RegexLexer):
 
 
 def setup(sphinx):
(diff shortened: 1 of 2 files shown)


mhilbrunner (approved):

mhilbrunner:
Thanks!

PR #11635, opened 2026-01-19
Update images for 2D meshes, 2D lights and shadows, for 4.6

skyace65:
Does what the title says. Lights and shadows is just one image so I figured I might as well  do it as part of this PR. Had to update text for 2D meshes as something got renamed. There's a new option available for converting to a mesh, I have no idea how it works, but the existing text is accurate. I'll need to come back to this at some point and figure out what it does, but for now this is worth updating.

All images have gone through Squoosh compression.


mhilbrunner (approved):

mhilbrunner:
Thank you!

PR #11646, opened 2026-01-21
Improve theme to match Godot 4.6's Modern editor theme

Calinou:
- Follow-up to https://github.com/godotengine/godot/pull/115143.

___

- Use [Inter font with some OpenType features](https://github.com/godotengine/godot/pull/111140) for headings and body.
  - The same font is now used on all platforms for body text, whereas the appearance previously differed on a per-platform basis.
- Remove unused JetBrains Mono Bold weight.
- Update syntax highlighting to match the Godot 4.6 default editor theme.

I'm not sure whether this should be cherry-picked, as we probably want the documentation to match the editor theme for the relevant editor version. In any case, this is safe to cherry-pick to `4.5` if desired. Changes from this PR are also relevant for the contributing docs.

## Preview

Page used for comparison: [Handling quit requests](https://docs.godotengine.org/en/latest/tutorials/inputs/handling_quit_requests.html)

> [!NOTE]
>
> Screenshots were taken on Firefox 146 on Linux, with Inter as the system font. This is why the body font is mostly unchanged. On other platforms, the change between the system font (Segoe UI on Windows, San Francisco on macOS) and Inter will appear to be more significant.

### Light theme

Before | After
-|-
Image | Image

### Dark theme

Before | After
-|-
Image | Image


mhilbrunner:
Thank you!

PR #11647, opened 2026-01-21
Removed an extraneous period.

WraithGlade:
The FAQ rationale for GDScript contained an item with improper grammar that tried to start a sentence in parentheses in lowercase, which is improper. 

This commit is just one possible fix.

### DIFF
--- a/about/faq.rst
+++ b/about/faq.rst
@@ -160,7 +160,7 @@ The main reasons for creating a custom scripting language for Godot were:
 5. Garbage collector results in stalls or unnecessarily large memory
    usage (Lua, Python, JavaScript, ActionScript, etc.).
 6. Difficulty integrating with the code editor for providing code
-   completion, live editing, etc. (all of them).
+   completion, live editing, etc (all of them).
 
 GDScript was designed to curtail the issues above, and more.
 


WraithGlade:
**Note**: Apparently the Godot docs actually seem to use "etc." by convention for _et cetera_ instead of the more common, more easy to use, and more legible plain "etc", which has the consequence of making many spots potentially look like grammatical errors or prematurely terminated sentences due to "etc." (as it did in this example).

I wasn't really aware of this existing convention in Godot's docs when I first made this suggested commit.

So, I personally recommend changing the _et cetera_ convention to be just "etc" throughout all the docs, but I would understand if you choose not to and to instead close this based on whatever your preference is. No worries either way; it isn't very important.

Anyway, have a good day/night and thanks for reading my thoughts on it.

mhilbrunner:
Thanks for taking the time out of your day to improve the docs, even if its indeed a minor thing :)

You could open an issue to discuss changing the convention; personally, I have no opinion either way, but on the other hand this is indeed such a minor change that would also require more work to actually make the changes that I think the same time could be better spent improving other things.

I'm going to close this, still: thanks for taking the time & also being diligent in looking at other examples.

PR #11652, opened 2026-01-22
Move GDExtension section to `engine_details/engine_api`.

Ivorforce:
Moves the GDExtension section to `engine_details/engine_api`.

Needs review / discussion, since I could imagine folks preferring how it is to what I'm proposing.

## Motivation

The GDExtension section is currently located in Manual / Scripting. This section is otherwise concerned with which programming languages you can use with Godot.

While the GDExtension section is sometimes useful for godot-cpp users, it is also plenty cross linked.
However, in general it is more a documentation of what GDExtension _is_ to the engine, and parts of it are fairly low level.

I think the `engine_api` section may be better suited for it. This is next to the module documentation, and indeed the two systems are somewhat similar. It's also close by the documentation for architecture and extension APIs.


AThousandShips:
You need to remove the entry in `tutorials/scripting/index.rst`

dsnopek:
So, the idea is that the godot-cpp stuff stays under scripting and the GDExtension system stuff moves to engine details?

This makes sense to me! We can always link from the godot-cpp docs into the GDExtension system docs where we need to.

There's one place I noticed where we're linking from GDExtension to godot-cpp, that I think would make more sense the other way around: [GDExtension version compatibility](https://docs.godotengine.org/en/latest/tutorials/scripting/gdextension/what_is_gdextension.html#version-compatibility) currently links to [godot-cpp version compatibility](https://docs.godotengine.org/en/latest/tutorials/scripting/cpp/about_godot_cpp.html#doc-what-is-gdextension-version-compatibility). We probably want the more specific case (godot-cpp) referring to the more generic case (GDExtension).

mhilbrunner:
Fine by me! Needs to wait until after 4.6 release though, as we probably don't get to change those URLs in the main repository last minute that make CI unhappy.

Ivorforce:
Rebased. I suppose this needs a companion PR on the main repo to fix the links?
Edit: Done https://github.com/godotengine/godot/pull/118060

dsnopek (approved):
We discussed this at yesterday's GDExtension team meeting, and agree we'd still like to do this move

paddy-exe (approved):

akien-mga:
@Ivorforce Can you update the class ref links here to pass CI?

Ivorforce:
Done

classes/class_gdextension.rst
```diff
@@ -28,7 +28,7 @@ The **GDExtension** resource type represents a `shared library <https://en.wikip
 Tutorials
 ---------
 
-- :doc:`GDExtension overview <../tutorials/scripting/gdextension/what_is_gdextension>`
+- :doc:`GDExtension overview <../engine_details/engine_api/gdextension/what_is_gdextension.html>`
```
akien-mga: The `.html` extension is wrong here (same in the other one).

Ivorforce: Ah, whoops, thanks!

akien-mga:
Thanks!

PR #11659, opened 2026-01-23
Document `IS_MULTIVIEW` built-in constant in Spatial shaders

Calinou:
- Follow-up to https://github.com/godotengine/godot/pull/115147.

### DIFF
--- a/tutorials/shaders/shader_reference/spatial_shader.rst
+++ b/tutorials/shaders/shader_reference/spatial_shader.rst
@@ -190,6 +190,8 @@ Global built-ins are available everywhere, including custom functions.
 |                             | In the Forward+ or Mobile renderers, it's ``0.0``.                                                  |
 |                             | In the Compatibility renderer, it's ``-1.0``.                                                       |
 +-----------------------------+-----------------------------------------------------------------------------------------------------+
+| in bool **IS_MULTIVIEW**    | ``true`` when output is stereoscopic (XR), ``false`` when output is monoscopic.                     |
++-----------------------------+-----------------------------------------------------------------------------------------------------+
 
 Vertex built-ins
 ----------------


tutorials/shaders/shader_reference/spatial_shader.rst
```diff
@@ -190,6 +190,8 @@ Global built-ins are available everywhere, including custom functions.
 |                             | In the Forward+ or Mobile renderers, it's ``0.0``.                                                  |
 |                             | In the Compatibility renderer, it's ``-1.0``.                                                       |
 +-----------------------------+-----------------------------------------------------------------------------------------------------+
+| in bool **IS_MULTIVIEW**    | ``true`` when output is multiview (XR), ``false`` otherwise (flat screen).                          |
```
AThousandShips: Flat in what sense? Isn't this stereo vs mono?

Calinou: Yes, it refers to whether we're rendering in stereo or not. That said, in practice, multiview enabled basically means we're rendering for XR, while multiview disabled means we're rendering for a flat screen.

There are scenarios where this could be not the case (e.g. if 3D glasses ever make a comeback), but Godot doesn't support this natively. You *could* also be viewing a flat screen game through an XR device (theater mode), but it's hardly ideal if the game has native XR support.

I've reworded the description a little, not sure if it's clear enough now.

tutorials/shaders/shader_reference/spatial_shader.rst
```diff
@@ -190,6 +190,9 @@ Global built-ins are available everywhere, including custom functions.
 |                             | In the Forward+ or Mobile renderers, it's ``0.0``.                                                  |
 |                             | In the Compatibility renderer, it's ``-1.0``.                                                       |
 +-----------------------------+-----------------------------------------------------------------------------------------------------+
+| in bool **IS_MULTIVIEW**    | ``true`` when output is stereoscopic (i.e. designed to be viewed on an XR device),                  |
+|                             | ``false`` when output is monoscopic (i.e. designed to be viewed on a flat screen).                  |
```
AThousandShips: I think "flat" might be confusing as we have plenty of curved screens nowadays

tutorials/shaders/shader_reference/spatial_shader.rst
BastiaanOlij: I think for 3DTV, it would still fall under the XR umbrella, especially as the 3D monitors currently being made all are getting OpenXR drivers because they add spatial tracking of the user into the mix.

tutorials/shaders/shader_reference/spatial_shader.rst
BastiaanOlij: Yeah, we really don't have good terms here. I think we can just leave out the i.e. parts, stereoscopic and monoscopic is indicative enough. People can google the meaning if they don't understand it.

skyace65 (approved):

skyace65:
Thanks!

PR #11666, opened 2026-01-23
Add `AStar*` behavior changes to 4.6 upgrade page

mihe:
Relates to godotengine/godot#113988.

We ran into this behavior change with `AStarGrid2D.get_point_path` when migrating a client's project to 4.6, and it took a while to track down, so figured it warranted being mentioned on the upgrade page.

I haven't verified the behavior of the other methods mentioned here. I'm just going by what was said in the PR description.

CC @kleonc @smix8

### DIFF
--- a/tutorials/migrating/upgrading_to_godot_4.6.rst
+++ b/tutorials/migrating/upgrading_to_godot_4.6.rst
@@ -191,6 +191,14 @@ Rendering
     Soft Light blend mode now always behaves as it did previously with ``use_hdr_2d``, regardless of
     the Viewport's ``use_hdr_2d`` setting (`GH-109971`_).
 
+Navigation
+~~~~~~~~~~
+
+.. note::
+
+    ``AStar2D.get_point_path``, ``AStar3D.get_point_path``, ``AStarGrid2D.get_id_path`` and ``AStarGrid2D.get_point_path`` will now return an
+    empty path when ``from_id`` is a disabled/solid point (`GH-113988`_).
+
 Changed defaults
 ----------------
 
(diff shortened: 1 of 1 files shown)


tutorials/migrating/upgrading_to_godot_4.6.rst
```diff
@@ -191,6 +191,17 @@ Rendering
...
+Navigation
+~~~~~~~~~
+
+.. note::
+
+    ``AStar2D.get_point_path``, ``AStar3D.get_point_path``, ``AStarGrid2D.get_id_path`` and ``AStarGrid2D.get_point_path`` will now return an
+    empty path when ``from_id`` is a disabled/solid point (`GH-113988`_).
+
+    ``AStar2D.get_id_path``, ``AStar2D.get_point_path``, ``AStar3D.get_id_path``, ``AStar3D.get_point_path``, ``AStarGrid2D.get_id_path`` and
+    ``AStarGrid2D.get_point_path`` will now return an empty path when ``from_id`` equals ``to_id`` (`GH-113988`_).
```
kleonc: This is not true though (in general case). It's true only if the `from_id == to_id` point is disabled, so the previous note already includes such case. If the point is enabled the mentioned methods still return a 1 element array.

Are the docs I've changed in https://github.com/godotengine/godot/pull/113988 unclear? I've added the `(even if [code]from_id == to_id[/code])` part to be super explicit there, but maybe it only adds potential confusion? 🤔
https://github.com/godotengine/godot/blob/233a5aeb4d23b569bb0c711825bb31978313d4ca/doc/classes/AStar2D.xml#L155

mihe: Ah, I misunderstood then. The documentation you wrote is perfectly fine. I just misread the PR description. I'll remove the second note.

mhilbrunner:
Definitely worth mentioning, thank you!

mhilbrunner:

PR #11672, opened 2026-01-24
Use 4 CPUs on readthedocs by injection.

Ivorforce:
An attempt to improve build times.

I haven't found any easier way to increase the job count, since there is no easy way to alter the build command.

I also tried [using free-threaded python](https://github.com/godotengine/godot-docs/compare/master...Ivorforce:godot-docs:free-threaded), but it turns out while `sphinx` has a unit test for free-threaded python, they aren't making use of it (still creating multiple processes with `-j4` instead of threads).

### DIFF
--- a/_extensions/override_jobs.py
+++ b/_extensions/override_jobs.py
@@ -1,0 +1,12 @@
+# -*- coding: utf-8 -*-
+
+# Hackily injects -j4 into the RTD build process.
+# Can be removed if we find an easier way to inject it.
+
+def setup(app):
+    app.parallel = 4
+
+    return {
+        "parallel_read_safe": True,
+        "parallel_write_safe": True,
+    }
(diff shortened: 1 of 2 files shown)


mhilbrunner:
Will be interesting to see whether we get enough RAM for this on RTD. Still, worth a try.

akien-mga:
Let's try and see what happens I guess :D

mhilbrunner:
Lets gooo (and thank you :))

Ivorforce:
Looks like it [succeeded](https://app.readthedocs.org/projects/godot/builds/31136188/), and with a runtime of ~50m, shaved off ~15m (22%) off the [reference runtime](https://app.readthedocs.org/projects/godot/builds/31130477/).

It won't solve our timeouts for good, but it should give us some wiggle room to enact further optimizations in the meantime :)

mhilbrunner:

PR #11678, opened 2026-01-25
Disambiguated and expanded the unofficial language bindings section.

WraithGlade:
I noticed the wording of the language bindings section was slightly ambiguous in context, and thus added the word "four" for clarity, but then also decided to add a link to what seems to be the most well-maintained of the unofficial community-made lists of language bindings and such, which could be very helpful to new users who want more expedient information.

### DIFF
--- a/getting_started/step_by_step/scripting_languages.rst
+++ b/getting_started/step_by_step/scripting_languages.rst
@@ -25,7 +25,10 @@ Available scripting languages
 
 Godot offers **four gameplay programming languages**: GDScript, C#,
 and, via its GDExtension technology, C and C++. There are more
-community-supported languages, but these are the official ones.
+community-supported languages, often referred to as "language bindings", 
+but these four are the official ones. Unofficial language bindings can be found on
+community-maintained lists such as [this one](https://github.com/Godot-Languages-Support/godot-lang-support) or elsewhere on the internet.
+Not everybody remembers to add their work to shared lists, nor wants to, nor necessarily can be trusted, so be aware of that.
 
 You can use multiple languages in a single project. For instance, in a team, you
 could code gameplay logic in GDScript as it's fast to write, and use C# or C++ to


getting_started/step_by_step/scripting_languages.rst
```diff
@@ -25,7 +25,10 @@ Available scripting languages
 
 Godot offers **four gameplay programming languages**: GDScript, C#,
 and, via its GDExtension technology, C and C++. There are more
-community-supported languages, but these are the official ones.
+community-supported languages, often referred to as "language bindings", 
+but these four are the official ones. Unofficial language bindings can be found on
+community-maintained lists such as [this one](https://github.com/Godot-Languages-Support/godot-lang-support) or elsewhere on the internet.
```
AThousandShips, line 30:
```suggestion
community-maintained lists such as `this one <https://github.com/Godot-Languages-Support/godot-lang-support>`_.
```

getting_started/step_by_step/scripting_languages.rst
```diff
@@ -25,7 +25,10 @@ Available scripting languages
...
+Not everybody remembers to add their work to shared lists, nor wants to, nor necessarily can be trusted, so be aware of that.
```
AThousandShips, line 31: I don't think this note is helpful

getting_started/step_by_step/scripting_languages.rst
```diff
@@ -25,7 +25,10 @@ Available scripting languages
 
 Godot offers **four gameplay programming languages**: GDScript, C#,
 and, via its GDExtension technology, C and C++. There are more
-community-supported languages, but these are the official ones.
+community-supported languages, often referred to as "language bindings", 
+but these four are the official ones. Unofficial language bindings can be found on
```
AThousandShips, line 29:
```suggestion
but these are the official ones. Unofficial language bindings can be found in
```
Let's avoid something we might forget to update

WraithGlade:
Here's a copy of my original suggestion again for context, with my changes in bold font, since the above snippets of commentary are fragmentary and made the context harder to understand:

> Godot offers four gameplay programming languages: GDScript, C#, and, via its GDExtension technology, C and C++. There are more community-supported languages, often referred to as "language bindings", but these **four** are the official ones. **Unofficial language bindings can be found on community-maintained lists such as [this one](https://github.com/Godot-Languages-Support/godot-lang-support) or elsewhere on the internet. Not everybody remembers to add their work to shared lists, nor wants to, nor necessarily can be trusted, so be aware of that.**

I don't understand why the suggested changes could ever be considered unhelpful. In particular, here are my thoughts:

- The expanded text I wrote links to what appears to be the most extensive and well-maintained list of unofficial/community language bindings. Not linking to such a resource only wastes users' time and/or potentially may lead them to worse sites or even conceivably to malware vectors designed to look benign. The open internet is increasingly risky in that regard, especially with so many bots dominating network traffic lately. Providing a link to whatever the current most credible (though still unofficial) source is objectively helpful to users, both in terms of expediency and user safety. Many Godot users will be children, teenagers, and people with imperfect understanding of internet safety and trustworthiness.
- I'm not sure what the suggestion change to the link syntax is supposed to indicate or do, so I don't have any real comment on that except that I thought we are supposed to submit links in Markdown's usual syntax.
- The comment that "Let's avoid something we might forget to update" strikes me as a non-sequitur. Literally the entire documentation of Godot (like any software project) is subject to potential need for revision, but even if one thinks in terms of what may need to be updated more or less often then my suggested set of changes seems clearly among the *least* likely to need updated. It links to an independently updated and maintained community site. It makes only broadly applicable and safe claims designed to help and protect users. It clarifies and disambiguates the discussion of the existing official languages slightly. It's not like adding new official languages is something Godot is likely to do often or to ever forget to update information about something that basic and unilaterally relevant to users. What is it about it that actually bothers you, because I see no objective basis on which it could be "forgetting to update it" and so I can't make much sense of this point and don't even really understand what you mean by that?
- You've suggested turning "There are more community-supported languages, often referred to as "language bindings", but these **four** are the official ones." back to the pre-existing "There are more community-supported languages, but these are the official ones." text, but doing so makes the statement subtly grammatically ambiguous. The original sentence actually grammatically implies that the four official languages (GDScript, C++, C, C#) are merely community supported languages, which is actually self-contradictory and makes the sentence feel subtly wrong. The only reason the original sentence can read "right" to some users is because when grammatical faults exist in text human beings have a strong tendency to gloss over them by subconsciously correcting erroneous grammatical and semantic references in the text based on contextual info and pre-existing knowledge. Existing experts on any subject are more likely to make the mistake of conflating their own internal automatic correction of incorrect information with what is actually there than newbies are. Read by a non-programmer or someone new to Godot though, the original sentence as written will subtly mislead some relatively small but still significant proportion of users into conflating the official and unofficial language bindings in some way. The addition of the word "four" a second time is just one possible way of eliminating that ambiguity, because it clarifies that it is referring to those four and _not_ to the unknown number of other language bindings.
- Mentioning the term "language bindings" is essential for enabling users to actually know what terminology they will need to search for in order to most effectively find more info about what bindings are available for Godot and what the concept even means in general, which many users won't know. As it stands, the original text's lack of inclusion of that terminology will leave users will less vocabulary by which to search for what they need to benefit from such bindings or to be properly aware of what they are or which ones exist, etc.
- The wording of the sentence "Not everybody remembers to add their work to shared lists, nor wants to, nor necessarily can be trusted, so be aware of that." was something I debated at length and rewrote several times from scratch. Its purpose is solely to do the following and I would be fine with any equivalent sentence(s) that convey the same essential points:
  - Just because any list of unofficial language bindings doesn't include a specific language binding doesn't mean it doesn't still exist somewhere. Not everyone who creates such things remembers to add their work to a shared list nor does any list maintainer necessarily have any awareness of such cases.
  - Some bindings may be specific to a project and specialized to it, or may currently be in a state where the creator doesn't think it wise for others to use it yet, and thus the user should be aware that there may be a reason for the absence of such, even if it isn't documented on that binding's own page(s).
  - Large software packages that are widely used (which Godot has become) are highly vulnerable to attack by malware authors who deliberately try to find vulnerable or poorly maintained software packages that they can insert malicious code into. Third party unofficial bindings and other libraries are very natural targets for such malicious parties. Thus, I thought it essential from a user safety standpoint (especially given the wide audience of Godot, which even includes children) that users be at least mildly warned of the existence of potentially untrustworthy unofficial Godot code. Large software packages such as JaveScript's NPM have been hit by such attacks repeatedly and in some cases even jeopardized much of the entire internet. Godot is increasingly heavily used due to its cross-platform ease for many things and is a target. Even though I've never seen untrustworthy Godot code yet it inevitably exists or will soon. Not warning users about that at least in some intentionally mildly worded way (as seen in the wording of my suggestion as "nor necessarily can be trusted") will endanger any users who seek such packages. Not linking to credible 3rd party sources will only force users to search on their own and thus can only increase their risk of blundering into a malware package unwittingly.

I don't actually care how any of the information in this suggested change is worded, nor do I care if someone else writes it or I do, but I do think that it is objectively true that all the provided points above add value and not including them could even harm users in extreme/unlucky cases. Burying one's head in the sand and pretending that the issue of where to find unofficial language bindings and what they are and that such 3rd party packages carry risks doesn't strike me as productive. It is objectively a risk that has to be accounted for and not addressing it explicitly in the docs can only increase the risk to users by forcing them to be entirely on their own in navigating that space.

I was honestly surprised to find myself even needing to defend these points, considering how objectively true they are and how they are essentially universally applicable concerns of _any_ software system supporting 3rd party packages.

In such cases I often find that there are unspoken _real_ reasons why such a benign suggestion like this is being resisted and that the nominal reasons are perhaps not the real ones, and so if you there's some other real reason why you don't like the notion of making users more aware of what 3rd party language bindings exist (maybe you don't want people using them and so are finding any possible nominal reason to not link to them?) and that 3rd parties can't necessarily be trusted nor necessarily will add their work to shared lists on their own volition then feel free to share it. I honestly thought I worded the suggested text as wholesomely and politely as I could, but perhaps it is the notion that 3rd party Godot packages could ever be malicious that is bothering you. That is my best guess so far. So, here's a suggested alternative for the last sentence:

Here's the original again:

> Not everybody remembers to add their work to shared lists, nor wants to, nor necessarily can be trusted, so be aware of that.

... and here's another way of making the same points:

> Lists of community-made language bindings should not be assumed to be complete and more may exist elsewhere. However, be aware that although most Godot community code is good-natured, some third party code could be unsafe to use, whether from a practicality or a security standpoint.

There are countless other ways one could make the same points though and I don't care how it is written or who writes it as long as this hole in the documentation is closed instead of leaving users facing the inconvenience of seeking more info about what language bindings are and what can be trusted on their own, which in contrast can only ever increase the risk and wasted time for the community if omitted.

Accurately debating the exact nuances of the wording of even a small piece of text can take a lot of time and energy to fully explain, as can be seen here in how much verbiage I had to expend in explaining what at the time seemed self-evident to me, but hopefully I have clarified the underlying reasons for my suggestions here adequately in a way that will lead to an optimal compromise or end result. Subtlety is hard to express concisely in words sometimes.

Anyway though, thank you for your time and considerations and for sharing your thoughts. 

In any case, to anyone reading this: Have an awesome day/night/etc!

getting_started/step_by_step/scripting_languages.rst
Ivorforce, line 30: We actually have a list inside the docs as well, at `doc_scripting_languages`.

dsnopek: Yeah, personally, I think we should be promoting the list we have in the docs rather than an external one. Our list has fewer languages listed, but it can certainly be expanded

WraithGlade:
Thanks for the information @Ivorforce and @dsnopek! 🙂

I wasn't aware that there was an official page for the bindings already. 

I will modify my commit soon or submit a new one by the end of the day.

WraithGlade:
I have redone a new commit to reflect the commentary and have just submitted it. I also made a few other additions I think add value. Thanks for sharing all your thoughts everyone.

PR #11679, opened 2026-01-25
Slightly improved the grammar and clarity of a sentence.

WraithGlade:
I also considered changing "an underscore" to "underscores" to reflect the fact that there are multiple underscores doing the separating (not just one, which in contrast is semantically wrong) but then thought it'd be even clearer if it used "each word" and "an underscore".

I also removed a superfluous comma that added nothing but subtle punctuation friction to the sentence.

### DIFF
--- a/getting_started/step_by_step/scripting_first_script.rst
+++ b/getting_started/step_by_step/scripting_first_script.rst
@@ -126,7 +126,7 @@ our node's ``texture``.
 .. note::
 
     By default, the :ui:`Inspector` displays a node's properties in "Title Case", with
-    capitalized words separated by a space. In GDScript code these properties
+    capitalized words separated by a space. In GDScript code, these properties
     are in "snake_case", which is lowercase with each word separated by an underscore.
 
     You can hover over any property's name in the :ui:`Inspector` to see a description and


getting_started/step_by_step/scripting_first_script.rst
```diff
@@ -126,8 +126,8 @@ our node's ``texture``.
 .. note::
 
     By default, the :ui:`Inspector` displays a node's properties in "Title Case", with
-    capitalized words separated by a space. In GDScript code, these properties
-    are in "snake_case", which is lowercase with words separated by an underscore.
+    capitalized words separated by a space. In GDScript code these properties
+    are in "snake_case", which is lowercase with each word separated by an underscore.
```
Calinou: I would keep the first comma, it clearly separates the two propositions ("In GDScript code" and "these properties are in snake_case"):

```suggestion
    capitalized words separated by a space. In GDScript code, these properties
    are in "snake_case", which is lowercase with each word separated by an underscore.
```

WraithGlade:
Ok, fair enough since it is subjective. I've applied your suggested change of adding the comma back in.

skyace65 (approved):

skyace65:
Thanks!

PR #11687, opened 2026-01-27
Change 'Make Unique' to 'New Physics Material' instruction

doremon3210:
The tutorial currently instructs users to click the "Make Unique" button for the PhysicsMaterial, but the button is disabled (grayed out) and cannot be selected.

This update revises the instructions to reflect the correct method for making a PhysicsMaterial unique in the Inspector, so users can modify the ball's PhysicsMaterial properties as intended.


skyace65:
This is actually a bug, not intended behavior. Engine issue can be found here https://github.com/godotengine/godot/issues/115416

PR #11702, opened 2026-01-29
Fix visited profile buttons on homepage breaking in light theme

suhankins:
Right now, visited "Profile buttons" ("I'm new to Godot", "I'm experienced with Godot", etc.) in light theme will use dark text, which doesn't contrast well with Profile button color.

This is because setting color to `unset` for their `visited` state made them inherit their color from `body`. This doesn't cause any issues on dark theme, but breaks on light theme.

This PR sets visited link color to the same color as normal one, fixing this issue.

---

The way page appears right now

image

Note dark text on 2 profile buttons

---

The way this page appears with this change applied

image

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -55,7 +55,7 @@ To begin reading the relevant documentation, select the tile that matches your p
         }
 
         .grid-item:visited {
-            color: unset;
+            color: hsla(0, 0%, 100%, 0.9);
         }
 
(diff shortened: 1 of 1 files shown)


Calinou:
- Superseded by https://github.com/godotengine/godot-docs/pull/11709.

Thanks for the contribution nonetheless!

PR #11705, opened 2026-01-29
Grammar: change singular possessive to plural possessive

Orphaeus:
On the page ["Logic preferences"](https://docs.godotengine.org/en/stable/tutorials/best_practices/logic_preferences.html), the second sentence of the third paragraph currently reads:

> Some property's setters have code to update other corresponding values, and that code can be slow!

In this case, because it is talking about multiple properties, not a single property, it should use the plural possessive ("properties' ") instead of the singular possessive ("property's"). I have made this change.


skyace65 (approved):

skyace65:
Thanks!

PR #11706, opened 2026-01-29
Document Patch PCK files

skyace65:
Documents Patch PCK export functionality as well as Delta Encoding for Patch PCK files.

Patch PCKs were added in https://github.com/godotengine/godot/pull/97118
Delta Encoding was added in https://github.com/godotengine/godot/pull/112011

Security recommendations came from dalexeev in a rocket chat discussion.

### DIFF
--- a/tutorials/export/exporting_pcks.rst
+++ b/tutorials/export/exporting_pcks.rst
@@ -64,7 +64,7 @@ See :ref:`doc_exporting_projects_pck_versus_zip` for a comparison of the two for
     The downside of this approach is that it's less transparent to the game logic,
     as it will not benefit from the same resource management as PCK/ZIP files.
 
-Security Concerns
+Security concerns
 -----------------
 
 It is important to note that loading PCK files for patches, mods, or extra content
@@ -81,8 +81,8 @@ asymmetric cryptography. You could store the public key in the main PCK, and sig
 patch or expansion PCK files with the private key. See the :ref:`class_Crypto` class
 for more information.
 
-Piracy Concerns
----------------
+Copyright concerns
+------------------
 
 If you want to use PCK files to distribute extra paid content, such as expansions,
 keep in mind that Godot provides no out-of-the-box way to prevent someone from
@@ -127,13 +127,15 @@ You can also add any patches you export to your base packs for future use. For
 example, adding ``patch.pck`` will ensure that ``patch2.pck`` will not include any
 resources from that first patch.
 
-Delta Encoding
+Delta encoding
 ~~~~~~~~~~~~~~
 
 Patch PCK files can be made smaller through the use of delta encoding. This makes it
 so that only the parts of a file that have been changed are updated. This does have a
-drawback of longer load times for the patch PCK files. There are two settings for
-delta encoding in addition to the filters:
+drawback of longer load times for the resources that are updated, and each patch for
+a resource cumulatively increases its load time.
+
+There are two settings for delta encoding in addition to the filters:
 
 - **Delta Encoding Compression Level:** Controls how much compression is applied to
   the files. We do not recommend any more than the default of 19. Beyond that more
@@ -148,6 +150,31 @@ delta encoding in addition to the filters:
   delta encoding.
 
 The default compression level, 19, is the highest recommended level.
+
+For the smallest patch size possible we recommend turning off compression for any
+resources you want to patch. Even if your base PCK files were compressed that
+shouldn't cause an issue.
+
+There are several places compression can be disabled for different resources:
+
+- Import settings related to compression on individually imported resources, such as
+  translation or 3D model files
+- :ui:`Compress Binary Resources` in the editor settings
+- :ui:`GDScript Export Mode` in the :button:`Scripts` tab of an export preset
+
+.. Note::
+
+    After disabling :ui:`Compress Binary Resources` you must delete the contents of
+    your projects ``.godot/imported/`` folder, then closing and open the project
+    again to re-generate it.
+
+It's important to note that when you export a delta encoded patch on top of previous
+patches, the packs you list under :ui:`Base Packs` in the :Button:`Patching` tab
+must be the exact same files that are loaded by the game at runtime, in the exact
+order they are loaded in. Re-exporting previous versions may end up being slightly
+different, due to non-determinism in Godot's export process, which can cause the
+patching to fail. This only applies to delta encoded patches, regular ones don't
+have this issue.
 
 Opening PCK or ZIP files at runtime
 -----------------------------------
--- a/tutorials/export/exporting_pcks.rst
+++ b/tutorials/export/exporting_pcks.rst
@@ -64,12 +64,37 @@ See :ref:`doc_exporting_projects_pck_versus_zip` for a comparison of the two for
     The downside of this approach is that it's less transparent to the game logic,
     as it will not benefit from the same resource management as PCK/ZIP files.
 
+Security Concerns
+-----------------
+
+It is important to note that loading PCK files for patches, mods, or extra content
+like expansions, will require you to code a system to automatically load files based
+on their location, and possibly name. This is a security vulnerability in three
+scenarios. One, a user downloads a mod with malicious code. Two, a malicious program
+already exists on an end user's PC and has replaced the PCK file with a malicious
+copy. Three, you are distributing patch PCK files through a game launcher and the
+system has become compromised.
+
+Take this into consideration when determining how to use PCK files in a project.
+For situations where you have a patching system via a launcher, consider using
+asymmetric cryptography. You could store the public key in the main PCK, and sign
+patch or expansion PCK files with the private key. See the :ref:`class_Crypto` class
+for more information.
+
+Piracy Concerns
+---------------
+
+If you want to use PCK files to distribute extra paid content, such as expansions,
+keep in mind that Godot provides no out-of-the-box way to prevent someone from
+copying the PCK file, and putting it on another person's computer. Any kind of DRM
+system is your responsibility to implement if that's what you want.
+
 Generating PCK files
 --------------------
 
 In order to pack all resources of a project into a PCK file, open the project
-and go to **Project > Export** and click on **Export PCK/ZIP**. Also, make sure
-to have an export preset selected while doing so.
+and go to :menu:`Project > Export`, select an export preset, and click on
+:button:`Export PCK/ZIP`.
 
 .. image:: img/export_pck.webp
 
@@ -78,27 +103,51 @@ with ``--export-pack``. The output file must with a ``.pck`` or ``.zip``
 file extension. The export process will build that type of file for the
 chosen platform.
 
-.. note::
+Patch PCK files
+---------------
 
-    If one wishes to support mods for their game, they will need their users to
-    create similarly exported files. Assuming the original game expects a
-    certain structure for the PCK's resources and/or a certain interface for
-    its scripts, then either...
+Generating Patch PCK files
+~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-    1. The developer must publicize documentation of these expected structures/
-       interfaces, expect modders to install Godot Engine, and then also expect
-       those modders to conform to the documentation's defined API when building
-       mod content for the game (so that it will work). Users would then use
-       Godot's built in exporting tools to create a PCK file, as detailed
-       above.
-    2. The developer uses Godot to build a GUI tool for adding their exact API
-       content to a project. This Godot tool must either run on a tools-enabled
-       build of the engine or have access to one (distributed alongside or
-       perhaps in the original game's files). The tool can then use the Godot
-       executable to export a PCK file from the command line with
-       :ref:`OS.execute() <class_OS_method_execute>`. The game itself shouldn't
-       use a tool-build of the engine (for security), so it's best to keep
-       the modding tool and game separate.
+To create a PCK file that only contains resources not present in the original
+release of a project, you would create a patch PCK file. This could be used for
+patches, mods, or expansions. For this to work you'll first need to have a PCK file
+for your project at the point of its initial release.
+
+To generate a patch PCK file, within the export menu, and with your desired preset
+selected, click on the :Button:`Patching` tab. At the bottom is the :ui:`Base Packs`
+section. Click on the :Button:`Add Pack` button, then navigate to the PCK file you
+exported that contains everything in your project for its initial release.
+
+Now, when you go to export a PCK file of your project again, if you have the
+:Button:`Export as Patch` button selected, only resources that have changed
+will be exported in the PCK file.
+
+You can also add any patches you export to your base packs for future use. For
+example, adding ``patch.pck`` will ensure that ``patch2.pck`` will not include any
+resources from that first patch.
+
+Delta Encoding
+~~~~~~~~~~~~~~
+
+Patch PCK files can be made smaller through the use of delta encoding. This makes it
+so that only the parts of a file that have been changed are updated. This does have a
+drawback of longer load times for the patch PCK files. There are two settings for
+delta encoding in addition to the filters:
+
+- **Delta Encoding Compression Level:** Controls how much compression is applied to
+  the files. We do not recommend any more than the default of 19. Beyond that more
+  memory is needed for export and import for significantly fewer gains. Any positive
+  values will have the same decompression speed, however export will take
+  longer the higher the number is. Negative values enable fast mode, which
+  means larger files, but the decompression speed is higher.
+
+- **Delta Encoding Minimum Size Reduction:** Controls how much size has to be saved
+  at minimum for compression to be used on an individual file. For example, at a
+  level of 10%, if the file size can only be reduced by 5%, then the file won't use
+  delta encoding.
+
+The default compression level, 19, is the highest recommended level.
 
 Opening PCK or ZIP files at runtime
 -----------------------------------
@@ -153,7 +202,7 @@ The PCK or ZIP file contains a ``mod_scene.tscn`` test scene in its root.
     ``Assembly.LoadFile("mod.dll")``
 
 Troubleshooting
-^^^^^^^^^^^^^^^
+~~~~~~~~~~~~~~~
 
 If you are loading a resource pack and are not noticing any changes, it may be
 due to the pack being loaded too late. This is particularly the case with menu
@@ -167,11 +216,25 @@ call :ref:`ProjectSettings.load_resource_pack() <class_ProjectSettings_method_lo
 in the autoload script's ``_init()`` function, rather than ``_enter_tree()``
 or ``_ready()``.
 
-Summary
--------
+Modding considerations
+----------------------
 
-This tutorial explains how to add mods, patches, or DLC to a game.
-The most important thing is to identify how one plans to distribute future
-content for their game and develop a workflow that is customized for that
-purpose. Godot should make that process smooth regardless of which route a
-developer pursues.
+If one wishes to support mods for their game, they will need their users to
+create similarly exported files. Assuming the original game expects a
+certain structure for the PCK's resources, and/or a certain interface for
+its scripts, then one of two things has to be done.
+
+1. The developer must document these expected structures/
+    interfaces, expect modders to install Godot Engine, and then also expect
+    those modders to conform to the documentation's defined API when building
+    mod content for the game (so that it will work). Users would then use
+    Godot's built in exporting tools to create a PCK file, as detailed
+    above.
+2. The developer uses Godot to build a GUI tool for adding their exact API
+    content to a project. This Godot tool must either run on a tools-enabled
+    build of the engine or have access to one (distributed alongside or
+    perhaps in the original game's files). The tool can then use the Godot
+    executable to export a PCK file from the command line with
+    :ref:`OS.execute() <class_OS_method_execute>`. The game itself shouldn't
+    use a tool-build of the engine (for security), so it's best to keep
+    the modding tool and game separate.


tutorials/export/exporting_pcks.rst
```diff
@@ -78,27 +78,24 @@ with ``--export-pack``. The output file must with a ``.pck`` or ``.zip``
 file extension. The export process will build that type of file for the
 chosen platform.
 
-.. note::
+This exported pck file, with everything included, is a necessary base for creating
+patch PCK files, which only include new or updated resources.
```
Calinou:
```suggestion
This exported PCK file, with everything included, is a necessary base for creating
patch PCK files, which only include new or updated resources.
```

tutorials/export/exporting_pcks.rst
```diff
@@ -78,27 +78,24 @@ with ``--export-pack``. The output file must with a ``.pck`` or ``.zip``
...
-       :ref:`OS.execute() <class_OS_method_execute>`. The game itself shouldn't
-       use a tool-build of the engine (for security), so it's best to keep
-       the modding tool and game separate.
+Now, when you go to export a PCK file of your project again, if you have the
+:Button:`Export as Patch` button selected. Only files resources that have changed
+will be exported in the PCK file.
+
+You can also add any patches you export to your base packs for future use. For
+example, adding patch.pck will ensure that patch2.pck will not include any resources
+from that first patch.
```
Calinou:
```suggestion
You can also add any patches you export to your base packs for future use. For
example, adding ``patch.pck`` will ensure that ``patch2.pck`` will not include any resources
from that first patch.
```

tutorials/export/exporting_pcks.rst
```diff
@@ -68,8 +68,8 @@ Generating PCK files
 --------------------
 
 In order to pack all resources of a project into a PCK file, open the project
-and go to **Project > Export** and click on **Export PCK/ZIP**. Also, make sure
-to have an export preset selected while doing so.
+and go to :menu:`Project > Export`, select and export preset, and click on
```
AThousandShips:
```suggestion
and go to :menu:`Project > Export`, select an export preset, and click on
```

tutorials/export/exporting_pcks.rst
```diff
@@ -78,27 +78,24 @@ with ``--export-pack``. The output file must with a ``.pck`` or ``.zip``
...
+This exported pck file, with everything included, is a necessary base for creating
+patch PCK files, which only include new or updated resources.
+
+Generating patch PCK files
+--------------------------
+
+To generate a patch PCK file, within the export menu, and with your desired preset
+selected, click on the :Button:`Patching` tab. At the bottom is the :ui:`Base Packs`
+section. Click on the :Button:`Add Pack` button, then navigate to the PCK file you
+exported that contained everything in your project.
```
AThousandShips:
```suggestion
exported that contains everything in your project.
```

tutorials/export/exporting_pcks.rst
```diff
@@ -78,27 +78,24 @@ with ``--export-pack``. The output file must with a ``.pck`` or ``.zip``
...
-    2. The developer uses Godot to build a GUI tool for adding their exact API
-       content to a project. This Godot tool must either run on a tools-enabled
-       build of the engine or have access to one (distributed alongside or
-       perhaps in the original game's files). The tool can then use the Godot
-       executable to export a PCK file from the command line with
-       :ref:`OS.execute() <class_OS_method_execute>`. The game itself shouldn't
-       use a tool-build of the engine (for security), so it's best to keep
-       the modding tool and game separate.
+Now, when you go to export a PCK file of your project again, if you have the
+:Button:`Export as Patch` button selected. Only files resources that have changed
```
AThousandShips:
```suggestion
:Button:`Export as Patch` button selected, only resources that have changed
```

skyace65:
guys, you're killing me here

skyace65:
PR is ready for review now, issues from earlier reviews have been fixed.

tutorials/export/exporting_pcks.rst
```diff
@@ -64,12 +64,37 @@ See :ref:`doc_exporting_projects_pck_versus_zip` for a comparison of the two for
...
     as it will not benefit from the same resource management as PCK/ZIP files.
 
+Security Concerns
+-----------------
+
+It is important to note that loading PCK files for patches, mods, or extra content
+like expansions, will require you to code a system to automatically load files based
+on their location, and possibly name. This is a security vulnerability in three
+scenarios. One, a user downloads a mod with malicious code. Two, a malicious program
+already exists on an end users PC and has replaced the PCK file with a malicious
```
AThousandShips:
```suggestion
already exists on an end user's PC and has replaced the PCK file with a malicious
```

tutorials/export/exporting_pcks.rst
```diff
@@ -64,12 +64,37 @@ See :ref:`doc_exporting_projects_pck_versus_zip` for a comparison of the two for
...
+asymmetric cryptography. In the main PCK you could store the public key, and sign
+patch or expansion PCK files with the private key. See the :ref:`class_Crypto` class
+for more information.
+
+Piracy Concerns
+---------------
+
+If you want to use PCK files to distribute extra paid content, such as expansions,
+keep in mind that Godot provides no way out of the box to prevent someone from
+copying the PCK file, and putting it on another persons computer. Any kind of DRM
```
AThousandShips:
```suggestion
copying the PCK file, and putting it on another person's computer. Any kind of DRM
```

tutorials/export/exporting_pcks.rst
```diff
@@ -78,27 +103,51 @@ with ``--export-pack``. The output file must with a ``.pck`` or ``.zip``
...
+
+Patch PCK files can be made smaller through the use of delta encoding. This makes it
+so that only parts of a file that have been changed are updated. This does have a
+drawback of longer load times for the patch PCK files. There are two settings for
+Delta encoding in addition to the filters:
+
+- **Delta Encoding Compression Level:** Controls how much compression is applied to
+  the files. We do not recommend any more than the default of 19. Beyond that more
+  memory is needed for export and import for significantly fewer gains. any positive
+  number values will have the same decompression speed, however export will take
```
AThousandShips:
```suggestion
  values will have the same decompression speed, however export will take
```

tutorials/export/exporting_pcks.rst
```diff
@@ -78,27 +103,51 @@ with ``--export-pack``. The output file must with a ``.pck`` or ``.zip``
...
+  longer the higher the number is. Negative number values enable fast mode, which
+  means larger files, but the decompression speed is faster.
+
+- **Delta Encoding Minimum Size Reduction:** Controls how much size has to be saved
+  at minimum for compression to be used on an individual file. For example, at a
+  level of 10%, if the file size can only be reduced by 5%, then the file won't use
+  delta encoding.
+
+The default compression level, 19, is the highest recommended level
```
AThousandShips:
```suggestion
The default compression level, 19, is the highest recommended level.
```

tutorials/export/exporting_pcks.rst
```diff
@@ -167,11 +216,25 @@ call :ref:`ProjectSettings.load_resource_pack() <class_ProjectSettings_method_lo
...
-developer pursues.
+Modding considerations
+----------------------
+
+If one wishes to support mods for their game, they will need their users to
+create similarly exported files. Assuming the original game expects a
+certain structure for the PCK's resources, and/or a certain interface for
+its scripts, then one of two things has to be done.
+
+1. The developer must publicize documentation of these expected structures/
```
AThousandShips:
```suggestion
1. The developer must document these expected structures/
```

tutorials/export/exporting_pcks.rst
```diff
@@ -78,27 +103,51 @@ with ``--export-pack``. The output file must with a ``.pck`` or ``.zip``
...
-       the modding tool and game separate.
+You can also add any patches you export to your base packs for future use. For
+example, adding ``patch.pck`` will ensure that ``patch2.pck`` will not include any
+resources from that first patch.
+
+Delta Encoding
+~~~~~~~~~~~~~~
+
+Patch PCK files can be made smaller through the use of delta encoding. This makes it
+so that only parts of a file that have been changed are updated. This does have a
```
AThousandShips:
```suggestion
so that only the parts of a file that have been changed are updated. This does have a
```

tutorials/export/exporting_pcks.rst
```diff
@@ -64,12 +64,37 @@ See :ref:`doc_exporting_projects_pck_versus_zip` for a comparison of the two for
...
+For situations where you have a patching system via a launcher, consider using
+asymmetric cryptography. In the main PCK you could store the public key, and sign
+patch or expansion PCK files with the private key. See the :ref:`class_Crypto` class
+for more information.
+
+Piracy Concerns
+---------------
+
+If you want to use PCK files to distribute extra paid content, such as expansions,
+keep in mind that Godot provides no way out of the box to prevent someone from
```
AThousandShips:
```suggestion
keep in mind that Godot provides no out-of-the-box way to prevent someone from
```

tutorials/export/exporting_pcks.rst
```diff
@@ -64,12 +64,37 @@ See :ref:`doc_exporting_projects_pck_versus_zip` for a comparison of the two for
...
+like expansions, will require you to code a system to automatically load files based
+on their location, and possibly name. This is a security vulnerability in three
+scenarios. One, a user downloads a mod with malicious code. Two, a malicious program
+already exists on an end users PC and has replaced the PCK file with a malicious
+copy. Three, you are distributing patch PCK files through a game launcher and the
+system has become compromised.
+
+Take this into consideration when determining how to use PCK files in a project.
+For situations where you have a patching system via a launcher, consider using
+asymmetric cryptography. In the main PCK you could store the public key, and sign
```
AThousandShips:
```suggestion
asymmetric cryptography. You could store the public key in the main PCK, and sign
```

tutorials/export/exporting_pcks.rst
AThousandShips:
```suggestion
already exists on an end user's PC and has replaced the PCK file with a malicious
```

tutorials/export/exporting_pcks.rst
AThousandShips:
```suggestion
copying the PCK file, and putting it on another person's computer. Any kind of DRM
```

tutorials/export/exporting_pcks.rst
```diff
@@ -78,27 +103,51 @@ with ``--export-pack``. The output file must with a ``.pck`` or ``.zip``
...
-.. note::
+Patch PCK files
+---------------
+
+Generating Patch PCK files
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To create a PCK file that only contains resources not present in the original
+release of a project, you would create a patch PCK file. This could be used for
+patches, mods or expansions. For this to work you'll first need to have a PCK file
```
AThousandShips:
```suggestion
patches, mods, or expansions. For this to work you'll first need to have a PCK file
```

tutorials/export/exporting_pcks.rst
```diff
@@ -78,27 +103,51 @@ with ``--export-pack``. The output file must with a ``.pck`` or ``.zip``
...
+~~~~~~~~~~~~~~
+
+Patch PCK files can be made smaller through the use of delta encoding. This makes it
+so that only parts of a file that have been changed are updated. This does have a
+drawback of longer load times for the patch PCK files. There are two settings for
+Delta encoding in addition to the filters:
+
+- **Delta Encoding Compression Level:** Controls how much compression is applied to
+  the files. We do not recommend any more than the default of 19. Beyond that more
+  memory is needed for export and import for significantly fewer gains. any positive
```
AThousandShips:
```suggestion
  memory is needed for export and import for significantly fewer gains. Any positive
```

tutorials/export/exporting_pcks.rst
AThousandShips:
```suggestion
  values will have the same decompression speed, however export will take
```

tutorials/export/exporting_pcks.rst
```diff
@@ -78,27 +103,51 @@ with ``--export-pack``. The output file must with a ``.pck`` or ``.zip``
...
+  number values will have the same decompression speed, however export will take
+  longer the higher the number is. Negative number values enable fast mode, which
```
AThousandShips:
```suggestion
  longer the higher the number is. Negative values enable fast mode, which
```

tutorials/export/exporting_pcks.rst
AThousandShips:
```suggestion
The default compression level, 19, is the highest recommended level.
```

tutorials/export/exporting_pcks.rst
AThousandShips:
```suggestion
1. The developer must document these expected structures/
```

tutorials/export/exporting_pcks.rst
```diff
@@ -78,27 +103,51 @@ with ``--export-pack``. The output file must with a ``.pck`` or ``.zip``
...
+  means larger files, but the decompression speed is faster.
```
AThousandShips:
```suggestion
  means larger files, but the decompression speed is higher.
```

tutorials/export/exporting_pcks.rst
AThousandShips:
```suggestion
so that only the parts of a file that have been changed are updated. This does have a
```

tutorials/export/exporting_pcks.rst
```diff
@@ -78,27 +103,51 @@ with ``--export-pack``. The output file must with a ``.pck`` or ``.zip``
...
+example, adding ``patch.pck`` will ensure that ``patch2.pck`` will not include any
+resources from that first patch.
+
+Delta Encoding
+~~~~~~~~~~~~~~
+
+Patch PCK files can be made smaller through the use of delta encoding. This makes it
+so that only parts of a file that have been changed are updated. This does have a
+drawback of longer load times for the patch PCK files. There are two settings for
+Delta encoding in addition to the filters:
```
AThousandShips:
```suggestion
delta encoding in addition to the filters:
```

tutorials/export/exporting_pcks.rst
```diff
@@ -78,27 +103,51 @@ with ``--export-pack``. The output file must with a ``.pck`` or ``.zip``
...
+Patch PCK files
+---------------
+
+Generating Patch PCK files
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To create a PCK file that only contains resources not present in the original
+release of a project, you would create a patch PCK file. This could be used for
+patches, mods or expansions. For this to work you'll first need to have a PCK file
+created of your project at the point of its initial release.
```
AThousandShips:
```suggestion
for your project at the point of its initial release.
```

tutorials/export/exporting_pcks.rst
AThousandShips:
```suggestion
keep in mind that Godot provides no out-of-the-box way to prevent someone from
```

tutorials/export/exporting_pcks.rst
AThousandShips:
```suggestion
asymmetric cryptography. You could store the public key in the main PCK, and sign
```

tutorials/export/exporting_pcks.rst
AThousandShips: Still unaddressed

tutorials/export/exporting_pcks.rst
AThousandShips: Ditto

AThousandShips (changes requested):

skyace65:
Fixed. Apologies for the ones I missed before.

tutorials/export/exporting_pcks.rst
```diff
@@ -64,12 +64,37 @@ See :ref:`doc_exporting_projects_pck_versus_zip` for a comparison of the two for
     The downside of this approach is that it's less transparent to the game logic,
     as it will not benefit from the same resource management as PCK/ZIP files.
 
+Security Concerns
+-----------------
```
Calinou: Use sentence case for headings:

```suggestion
Security concerns
-----------------
```

tutorials/export/exporting_pcks.rst
```diff
@@ -64,12 +64,37 @@ See :ref:`doc_exporting_projects_pck_versus_zip` for a comparison of the two for
...
+system has become compromised.
+
+Take this into consideration when determining how to use PCK files in a project.
+For situations where you have a patching system via a launcher, consider using
+asymmetric cryptography. You could store the public key in the main PCK, and sign
+patch or expansion PCK files with the private key. See the :ref:`class_Crypto` class
+for more information.
+
+Piracy Concerns
+---------------
```
Calinou: I am not huge on the "piracy" term personally, I would go for a softer term:

```suggestion
Copyright concerns
------------------
```

tutorials/export/exporting_pcks.rst
```diff
@@ -78,27 +103,51 @@ with ``--export-pack``. The output file must with a ``.pck`` or ``.zip``
...
-       executable to export a PCK file from the command line with
-       :ref:`OS.execute() <class_OS_method_execute>`. The game itself shouldn't
-       use a tool-build of the engine (for security), so it's best to keep
-       the modding tool and game separate.
+You can also add any patches you export to your base packs for future use. For
+example, adding ``patch.pck`` will ensure that ``patch2.pck`` will not include any
+resources from that first patch.
+
+Delta Encoding
+~~~~~~~~~~~~~~
```
Calinou:
```suggestion
Delta encoding
~~~~~~~~~~~~~~
```

AThousandShips (approved):
LGTM

skyace65:
@mhilbrunner I got the build issue fixed.

tutorials/export/exporting_pcks.rst
```diff
@@ -78,27 +103,51 @@ with ``--export-pack``. The output file must with a ``.pck`` or ``.zip``
...
+You can also add any patches you export to your base packs for future use. For
+example, adding ``patch.pck`` will ensure that ``patch2.pck`` will not include any
+resources from that first patch.
+
+Delta encoding
+~~~~~~~~~~~~~~
+
+Patch PCK files can be made smaller through the use of delta encoding. This makes it
+so that only the parts of a file that have been changed are updated. This does have a
+drawback of longer load times for the patch PCK files. There are two settings for
```
mihe: I think it's important to stress that it's not the PCK loading that takes longer, it's the loading of the resources themselves, as the patching happens lazily, every time the resource is loaded (i.e. when not already loaded/referenced by something else).

So you're paying for this overhead at every `ResourceLoader.load` call, not the `ProjectSettings.load_resource_pack` call.

This also means that this overhead increases with every new delta-encoded patch you load.

tutorials/export/exporting_pcks.rst
```diff
@@ -78,27 +103,51 @@ with ``--export-pack``. The output file must with a ``.pck`` or ``.zip``
...
-       executable to export a PCK file from the command line with
-       :ref:`OS.execute() <class_OS_method_execute>`. The game itself shouldn't
-       use a tool-build of the engine (for security), so it's best to keep
-       the modding tool and game separate.
+You can also add any patches you export to your base packs for future use. For
+example, adding ``patch.pck`` will ensure that ``patch2.pck`` will not include any
+resources from that first patch.
+
+Delta encoding
+~~~~~~~~~~~~~~
```
mihe, line 131: One thing not in this section that I think would be worthwhile to mention is the importance of what base packs you add to the "Base Packs" list in the export settings.

Because delta encoding is a patch derived from a new and old state, it's crucial that it's then applied on top of the exact same old state, bit for bit. This means that if you created a delta patch from `v1.pck` and `v2.pck`, and then (for whatever reason) decide to just re-export `v2.pck` locally (because you couldn't find the shipped file or whatever), you're subjecting yourself to any non-determinism in Godot's export code, which has turned out to be a surprisingly big problem.

If you don't apply the patch on top of the same binary data, the resource loading will fail with a checksum error, so it's important that "Base Packs" is the exact same packs that were shipped to your end-users, and listed in the exact same order that they would be loaded at runtime.

For Android, you need to add the published APK/AAB as the first base pack instead, since the PCK is bundled into it in a special format. This only recently became possible, as a result of godotengine/godot#116553.

tutorials/export/exporting_pcks.rst
```diff
@@ -78,27 +103,51 @@ with ``--export-pack``. The output file must with a ``.pck`` or ``.zip``
...
+
```
mihe, line 132: Another thing that might be worth stressing is how delta-encoding more or less requires that you disable compression if you want the smallest possible PCK size.

So, for example, GDScript files should ideally not be left at their default export setting, which is "Compressed binary tokens":

Image

The same goes for translation files, which are also compressed by default, if I remember correctly.

(And yes, we should probably emit a warning about this during export.)

mihe (commented):
Sorry for the late review. This ended up falling through the cracks.

I think this looks good overall, but I have a few suggestions.

tutorials/export/exporting_pcks.rst
Calinou: > The same goes for translation files, which are also compressed by default, if I remember correctly.

It's probably a good idea to disable **Compress Binary Resources** in the Editor Settings, then reimport all resources by closing the editor, removing `.godot/imported/` and reopening the editor.

This setting isn't available on a per-project basis, unfortunately.

See also https://github.com/godotengine/godot/pull/103287 and https://github.com/godotengine/godot/pull/103284. We'll probably want to move the compression settings to the project level, so the setting can behave identically across team members when working on the same project.

tutorials/export/exporting_pcks.rst
skyace65: When you say each patch increases overhead, Is that each delta-encoded patch that includes the specific resource you're loading? Or each delta-encoded patch in general even if it doesn't touch it?

tutorials/export/exporting_pcks.rst
skyace65: Am I correct in assuming if the original project's resources are compressed, and the delta encoded PCK resources are uncompressed that will cause issues?

tutorials/export/exporting_pcks.rst
mihe: Each delta-encoded patch that includes that specific resource is what increases the overhead. It's the actual applying of the patch delta that makes up the overhead.

tutorials/export/exporting_pcks.rst
mihe: Not in the sense of causing errors or anything like that, no. All that the delta encoding does is take a blob of binary data from one shape to another. It doesn't really matter what the actual data is. If the resource was compressed but is now uncompressed you just end up with a potentially big delta.

Disabling compression is just a way to ensure a small delta, since compression often scrambles data in a way where even the smallest change can produce a file that looks nothing like the previous one.

skyace65:
I've made changes based on the reviews.

tutorials/export/exporting_pcks.rst
```diff
@@ -78,27 +103,63 @@ with ``--export-pack``. The output file must with a ``.pck`` or ``.zip``
...
-       executable to export a PCK file from the command line with
-       :ref:`OS.execute() <class_OS_method_execute>`. The game itself shouldn't
-       use a tool-build of the engine (for security), so it's best to keep
-       the modding tool and game separate.
+Delta encoding
+~~~~~~~~~~~~~~
+
+Patch PCK files can be made smaller through the use of delta encoding. This makes it
+so that only the parts of a file that have been changed are updated. This does have a
+drawback of longer load times for the resources that are updated. And each patch for
```
mihe:
```suggestion
drawback of longer load times for the resources that are updated, as each patch for
```

tutorials/export/exporting_pcks.rst
```diff
@@ -78,27 +103,63 @@ with ``--export-pack``. The output file must with a ``.pck`` or ``.zip``
...
+a resource cumulatively increases its load time. There are two settings for
+delta encoding in addition to the filters:
```
mihe: Nitpicky, but some line breaks seem appropriate here.

```suggestion
a resource cumulatively increases its load time.

There are two settings for delta encoding in addition to the filters:
```

tutorials/export/exporting_pcks.rst
```diff
@@ -78,27 +103,63 @@ with ``--export-pack``. The output file must with a ``.pck`` or ``.zip``
...
+  at minimum for compression to be used on an individual file. For example, at a
+  level of 10%, if the file size can only be reduced by 5%, then the file won't use
+  delta encoding.
+
+The default compression level, 19, is the highest recommended level.
+
+For the smallest patch size possible we recommend turning off :ui:`Compress Binary Resources`
+in the editor settings, deleting the contents of your projects ``.godot/imported/``
+folder, then closing and opening the project again to re-generate it. Even if your
+base PCK files were compressed that shouldn't cause an issue.
```
mihe: Note that this Compress Binary Resources editor setting that @Calinou mentioned has no affect the GDScript compression I mentioned earlier.

I don't know if we necessarily want to list every single resource that should have its compression disabled though. It might be better to just phrase this more generally ("we recommend disabling compression for any resource you want to patch") and give one example or two, like this editor setting.

tutorials/export/exporting_pcks.rst
```diff
@@ -78,27 +103,63 @@ with ``--export-pack``. The output file must with a ``.pck`` or ``.zip``
...
+
+It's important to note that when you export a delta encoded patch on top of previous
+patches everything must be exactly the same. If you loose a previous patch you can't
+re-export it to recreate it for a new patch. due to non-determinism in Godot's
+export process it wouldn't be exactly the same. This only applies to delta encoded
+patches, regular ones don't have this issue.
```
mihe: I would say maybe don't mention the case of losing a previous patch. You would need to be shipping the patches to your users in some way, so almost by definition you can't "lose them". My earlier example of not being able to find the shipped file was more along the lines of someone being too lazy to do so.

Also, I do think it's important to refer to the "Base Packs" part of the patching settings specifically, since that's where care has to be taken.

Something like this maybe (forgive the lack of line wrapping):

```suggestion
It's important to note that when you export a delta encoded patch on top of previous patches, the packs you list under :ui:`Base Packs` in the :Button:`Patching` tab must be the exact same files that are loaded by the game at runtime, in the exact order they are loaded in. Re-exporting previous versions may end up being slightly different, due to non-determinism in Godot's export process, which can cause the patching to fail. This only applies to delta encoded patches, regular ones don't have this issue.
```

skyace65:
@mihe I've made changes based on your review. I didn't go with the ``, as`` suggestion as that made the way the sentence is worded not flow right, I did replace the period with a comma though and change to lowercase "and". Other changes were done.

I re-did the compression section so it is a list now. I know you said we might not want to do that, but after thinking it over I think that's the best way to handle this. We don't have any centralized page on compression for resources and where everything is, since that info isn't discoverable I think there's a real risk people read it and just assume "Oh, it's talking about texture compression for image files, and sound compression." and never realize there's other stuff they can also change to save space.

tutorials/export/exporting_pcks.rst
```diff
@@ -78,27 +103,78 @@ with ``--export-pack``. The output file must with a ``.pck`` or ``.zip``
...
+The default compression level, 19, is the highest recommended level.
+
+For the smallest patch size possible we recommend turning off compression for any
+resources you want to patch. Even if your base PCK files were compressed that
+shouldn't cause an issue.
+
+There are several places compression can be disabled for different resources:
+
+- Import settings related to compression on individually imported resources, such as
+  image or 3D model files
```
mihe: I don't think you can disable texture compression even if you wanted to. Translation files are probably worth bringing up if we're going to list a bunch of things here though.

```suggestion
- Import settings related to compression on individually imported resources, such as
  translation or 3D model files
```

mihe (approved):
Apart from the minor comment, LGTM.

tutorials/export/exporting_pcks.rst
Calinou: Setting a texture's compression mode to VRAM Uncompressed makes it fully uncompressed. It should be delta-friendly, but keep in mind VRAM usage will be much higher than when using VRAM Compressed (same as Lossless or Lossy).

skyace65: I'll make changes to this tomorrow. Also need to remove the 3D model files listing since I checked and we don't actually have anything for that, I think I was thinking of textures at the time.

mihe: > since I checked and we don't actually have anything for that

We do for both OBJ and GLTF files (see [here](https://docs.godotengine.org/en/latest/classes/class_resourceimporterobj.html#class-resourceimporterobj-property-force-disable-mesh-compression) and [here](https://docs.godotengine.org/en/latest/classes/class_resourceimporterscene.html#class-resourceimporterscene-property-meshes-force-disable-compression)).

EDIT: Also, based on what @Calinou is saying it sounds like textures should be listed there as well.

Calinou: For 2D games, it's probably fine to use VRAM Uncompressed, but I wouldn't recommend disabling VRAM compression for 3D games with lots of high-resolution textures (1024x1024 or more). Pixel art 3D is fine though.

skyace65: I'll leave off images for now and just update it to translation and 3D model files

skyace65:
Updated. Replaced images with translation files for the example. There's no further changes I want to make to this PR myself

Luxecustomshop24 (approved):

mhilbrunner (approved):

PR #11725, opened 2026-02-02
Change URL of Autoloads versus regular nodes page to match new title

Calinou:
The terminology "internal nodes" did not refer to any specific Godot functionality when it was first written, but now that Godot has a notion of internal nodes, it was bound to create confusion.

This adds a redirect for the old URL too.

- See https://github.com/godotengine/godot-docs-user-notes/discussions/540#discussioncomment-15657035.

PS: Do we usually cherry-pick PRs that change URLs? I forgot.

### DIFF
--- a/about/docs_changelog.rst
+++ b/about/docs_changelog.rst
@@ -470,7 +470,7 @@ Best Practices:
 - :ref:`doc_what_are_godot_classes`
 - :ref:`doc_scene_organization`
 - :ref:`doc_scenes_versus_scripts`
-- :ref:`doc_autoloads_versus_internal_nodes`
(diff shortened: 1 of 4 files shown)


skyace65:
It looks good to me but aren't we still at our redirect limit?

Calinou:
> It looks good to me but aren't we still at our redirect limit?

Redirects past the limit can work thanks to JavaScript now, so the redirect limit is less of a problem.

skyace65 (approved):

skyace65:
Thanks!

mhilbrunner:

Issue #11720, opened 2026-02-01
Add enum of DirAccess.copy method's chmod_flags value


CsloudX:
**Your Godot version:** v4.6.stable.mono.official [89cea1439]

**Issue description:**

<img width="876" height="287" alt="Image" src="https://github.com/user-attachments/assets/34ae5f4d-072b-40ed-a189-1b6b860b9da3" />

I want copy file use DirAccess.copy method. but i don't known how set the chmod_flags.

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/stable/classes/class_diraccess.html#class-diraccess-method-copy

ScatteredComet:
https://en.wikipedia.org/wiki/Chmod provides:

<img width="370" height="301" alt="Image" src="https://github.com/user-attachments/assets/e8f92382-760e-446a-a900-89e6d9d110aa" />

I have had a look through the docs and can't find an analogous case of an externally defined enum, so I'm not sure what the precedent is - maybe someone knows another function like DirAccess.copy() that has an int parameter that corresponds to a flag set by e.g. the OS?

I think there are a few options:

- Make it clearer in the DirAccess.copy() description that these flags correspond to those provided by Unix and/or
- Add a table in the manual (e.g. on this page: [Runtime file loading and saving](https://docs.godotengine.org/en/stable/tutorials/io/runtime_file_loading_and_saving.html)) with the 8 possible values and reference it in the DirAccess.copy() description

I think adding a section titled "Managing directories" after "Plain text and binary files" on the runtime file loading and saving page would be a good option. But perhaps there's some convention for how this should be documented that I haven't managed to find; of course let me know if thats the case.

ScatteredComet:
I think my PR is probably overkill, so I've closed it (I didn't see Calinou's PR until after creating mine). There isn't a section in the manual for DirAccess though, so maybe it could serve as a start for creating one.

PR #11729, opened 2026-02-03
Add "moving directories and files" section to file loading manual

ScatteredComet:
This section's current purpose is to list the values that chmod_flags can take, so that the class reference for DirAccess.copy() can link to here.

This is for https://github.com/godotengine/godot-docs/issues/11720 (but does not resolve it).


ScatteredComet:
Oops! Calinou has already provided a PR for this, so I will close this now.

https://github.com/godotengine/godot/pull/115849

Calinou:
Note that your PR technically still has merit on its own, but using chmod flags is pretty niche and something most applications never need to bother doing manually. I think it's fine to have it documented in the class reference only.

Issue #11665, opened 2026-01-23
Download Button redirects to 404


KaenguruuDev:
**Your Godot version:** 4.5.1, "Stable" on the docs site

**Issue description:** The "Download" button in the info box in the bottom right redirects to a 404, specifically `https://docs.godotengine.org/en/stable/null`

**URL to the documentation page:** https://docs.godotengine.org/en/stable

Screenshots below for illustration:

<img width="512" height="447" alt="Image" src="https://github.com/user-attachments/assets/418ccef6-eae0-498c-9cf9-b72c1630bdaa" />

<img width="2560" height="1432" alt="Image" src="https://github.com/user-attachments/assets/e582ed96-961f-4dfb-b48a-9f8418430647" />

I noticed that there still is the direct link at the bottom of the homepage so fixing this is (hopefully) relatively easy, although I don't have the time to get into how to do this myself.

ScatteredComet:
On the [readthedocs page](https://app.readthedocs.org/projects/godot/) there isn't a downloads tab at the top:

<img width="1163" height="600" alt="Image" src="https://github.com/user-attachments/assets/edc2548e-dba0-412b-9565-57c6697c3b76" />

so I assume that the downloads were moved away from readthedocs at some point?

The button defined [here](https://github.com/godotengine/godot-docs/blob/033e8b8a7802f83bbe4d8d8896441cf386695620/_templates/versions.html#L47) (line 47) references https://app.readthedocs.org/projects/godot/downloads/ which isn't a valid url.

[This](https://docs.godotengine.org/en/stable/#offline-documentation) page references downloadable docs hosted using nightly.link - perhaps we should change this button to go to that instead?

I am unfamiliar with nightly.link, but it looks like it only exposes the zip files themselves, and not a home page that lists all the downloads.

Calinou:
> so I assume that the downloads were moved away from readthedocs at some point?

Indeed, we stopped using Read the Docs' downloads a few years ago as their infrastructure couldn't keep up (the Godot docs are large downloads).

We can probably use some JavaScript code to override the download URLs on the RTD download button to the README of this repository at the correct section: https://github.com/godotengine/godot-docs#download-for-offline-use

ScatteredComet:
I've created a PR that changes the href to https://github.com/godotengine/godot-docs#download-for-offline-use, would that work? (I assume not, since you mentioned JS, but I thought I'd suggest it anyway).

AgentSam:
[...]

PR #11730, opened 2026-02-04
fix: update downloads button to point to a valid URL

ScatteredComet:
This is for #11665.


Calinou (commented):
As per the comment on lines 20-23 of the file you modified, this won't work in production (even though it will appear to work when building locally).

Issue #11735, opened 2026-02-05
NavigationAgent Avoidance Documentation - Warn Reader to Set target_position


rhbourke:
## Your Godot version:
4.6
## Issue description:
NavigationAgent Avoidance documentation does not warn reader that it is necessary to set the target_position attribute of the NavigationAgent to calculate safe velocities.

It may seem obvious that the target_position of the NavigationAgent needs to be set for it to work, but there are many use cases for the NavigationAgent's avoidance feature where users are not interested in calculating a navigation path and might not expect it to be necessary.

Here is another example of someone who was stuck due to this confusion:
https://github.com/godotengine/godot/issues/88648#issuecomment-1979642822

While the documentation does have a very helpful scripting example that does indeed contain a call to NavigationAgent.set_target_position(), there is nothing on the page to assist the reader in understanding that all safe_velocity values received through the velocity_computed signal will be set to (0, 0) if this is not provided. In the absence of a note clarifying this, it just looks like something that is being done to facilitate the other navigation features that are being used in that example. 

## URL to the documentation page:
https://docs.godotengine.org/en/stable/tutorials/navigation/navigation_using_navigationagents.html#navigationagent-avoidance

mhilbrunner:
Fixed by https://github.com/godotengine/godot-docs/pull/11736.

PR #11736, opened 2026-02-05
Add note about target_position requirement for NavigationAgent Avoidance

rhbourke:
Addressing https://github.com/godotengine/godot-docs/issues/11735

### DIFF
--- a/tutorials/navigation/navigation_using_navigationagents.rst
+++ b/tutorials/navigation/navigation_using_navigationagents.rst
@@ -114,7 +114,9 @@ This velocity vector should be used to move the NavigationAgent's parent node in
 
 .. note::
 
-    The NavigationAgent must be supplied with a ``target_position`` attribute even if you are only using the agent for avoidance, otherwise the ``safe_velocity`` received from the ``velocity_computed`` signal will always be (0,0).
+    The NavigationAgent **must** be supplied with a ``target_position`` attribute,
+    even if you are only using the agent for avoidance. Otherwise, the ``safe_velocity``
+    received from the ``velocity_computed`` signal will always be ``(0, 0)``.
 
 The following NavigationAgent properties are relevant for avoidance:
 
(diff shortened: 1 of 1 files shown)


tutorials/navigation/navigation_using_navigationagents.rst
```diff
@@ -112,6 +112,10 @@ This velocity vector should be used to move the NavigationAgent's parent node in
 
     Only other agents on the same map that are registered for avoidance themself will be considered in the avoidance calculation.
 
+.. note::
+
+    The NavigationAgent must be supplied with a ``target_position`` attribute even if you are only using the agent for avoidance, otherwise the ``safe_velocity`` received from the ``velocity_computed`` signal will always be (0,0).
```
Calinou:
```suggestion
    The NavigationAgent **must** be supplied with a ``target_position`` attribute,
    even if you are only using the agent for avoidance. Otherwise, the ``safe_velocity``
    received from the ``velocity_computed`` signal will always be ``(0, 0)``.
```

Calinou (approved):
Looks good to me.

mhilbrunner (approved):

tutorials/navigation/navigation_using_navigationagents.rst
```diff
@@ -112,6 +112,12 @@ This velocity vector should be used to move the NavigationAgent's parent node in
 
     Only other agents on the same map that are registered for avoidance themself will be considered in the avoidance calculation.
 
+.. note::
+
+    The NavigationAgent **must** be supplied with a ``target_position`` attribute,
+    even if you are only using the agent for avoidance. Otherwise, the ``safe_velocity``
+    received from the ``velocity_computed`` signal will always be ``(0, 0)``.
```
AThousandShips: Using `(0, 0)` here is potentially misleading as this is relevant for both 2D and 3D, I'd say to make it more generic like "zero velocity" or something

rhbourke: That's a great catch and good suggestion. 

The technically correct mathematical way to say this might be "...will always be the null vector", but I am not sure if that is in line with how zero-valued vectors are described elsewhere in the Godot docs or if that is the clearest way to describe it since the value, of course, is not 'null' in a programming sense. I suppose another option is "...will always be zero-valued". 

I would prefer to defer, ultimately, to those like yourself or others that have more experience contributing to these docs.

AThousandShips: I think "zero vector" would be most appropriate, it's used elsewhere, or reference the `ZERO` constant for each type

mhilbrunner:

mhilbrunner:

PR #11737, opened 2026-02-05
Add summary of blender nodes that work with  -col etc

ScatteredComet:
- This is aimed to close [#115869](https://github.com/godotengine/godot/issues/115869)

---

However this is a draft as I have a few questions:

1. I'm unfamiliar with Collada but wanted to check whether this sentence is definitely correct. Is it saying that if you import a blender object as a COLLADA / DAE file, the -colonly works with blenders empty objects?

> With Collada files, the option ``-colonly`` can also be used with Blender's empty objects.

Maybe it can be made a little clearer by saying "When exporting a Collada [or DAE] file from blender, the option ``-colonly`` can also be used with Blender's empty objects"

2. Do I need to / is it best to clarify what a mesh object / mesh data is in blender?

Also just a general check on my style would be good.


skyace65:
Closing due to lack of activity.

PR #11738, opened 2026-02-05
LightmapGI note light leak issue double-sided materials

SmartySmart702:
Using double-sided materials for e.g. a simple room with overlapping walls will create light leaks.
Also walls that dont overlap but touch.
Reason for this is the bounce light inside the mesh and the over-bright inside of said mesh.
If a beginner (like me) is using Blender the material cull mode is disabled by default (for imported materials) resulting in a frustrating issue that is not obvious at first.


Calinou:

Contributions to the class reference should be sent on the main Godot repository, not here. This is because the class reference in this repository is generated automatically based on the XML files in the main repository. See [Contribute to the class reference](https://contributing.godotengine.org/en/latest/documentation/class_reference.html) in the documentation :slightly_smiling_face:

PS: This is relevant to mention in [Using LightmapGI](https://docs.godotengine.org/en/stable/tutorials/3d/global_illumination/using_lightmap_gi.html) too. Consider opening a PR for that on godot-docs.

PR #11739, opened 2026-02-06
Update wording in nodes_and_scenes.rst

wallbloggerbeing:
Amount of scenes limited by c++ integer limit depending on host machine. Reference:
https://github.com/godotengine/godot/blob/56f3e2611daadb0b1894b46737ee3000e82e33f9/scene/resources/2d/tile_set.h#L794-L810

Further change would be choosing intx_t in shown code snippet for specific number in docs

### DIFF
--- a/getting_started/step_by_step/nodes_and_scenes.rst
+++ b/getting_started/step_by_step/nodes_and_scenes.rst
@@ -52,7 +52,7 @@ that runs and jumps, a life bar, a chest with which you can interact, and more.
 
 The Godot editor essentially is a **scene editor**. It has plenty of tools for
 editing 2D and 3D scenes, as well as user interfaces. A Godot project can
-contain as many of these scenes as you need. The engine only requires one as
+contain as many of these scenes as your computer can handle. The engine only requires one as
 your application's **main scene**. This is the scene Godot will first load when
 you or a player runs the game.
 


AThousandShips:
> Amount of scenes limited by c++ integer limit depending on host machine

It's always 2^31-1 `int` is 32 bits on all modern CPUs

I don't think this change is relevant, nor is the integer limit likely to be the limit, memory is far more likely to be a limit, 2^31-1 `Object`s is at least hundreds of gigabytes, scene data even more

wallbloggerbeing:
> It's always 2^31-1 `int` is 32 bits on all modern CPUs

Not necessarily, I would assume theres an argument to be made for machines that have 16-bit ints, and assuming you have 2^{15} - 1 scenes, each 4KiB (To play bad apple on them or sth idk), then youd run out one after that. 
I get your point about how this is unnecessary, but the change is minimally invasive and just contributes to some clarity.

AThousandShips:
> Not necessarily, I would assume theres an argument to be made for machines that have 16-bit ints

Godot is not built on any such machine, C++ `int` is only 16 bits on Win16 which is only in the windows kernel

But regardless Godot does not compile on any such platform, due to build system checks 

But I disagree that this adds any clarity

mhilbrunner:
Thanks for contributing! I agree that this does not add clarity; it also IMO changes the intended meaning, which to me is "scenes are not meaningfully limited, use as many as you want". Adding these specifics instead makes the reader think about "okay, how many *does* my computer afford? What about my player's specs?"

So we can then go on and add that definition, but that introduces even more mental overhead and details for a sentence that originally only wanted to say "hey, you can use more than a few scenes".

wallbloggerbeing:
Fair enough, I see the point.

PR #11750, opened 2026-02-12
Add missing RID cleanup in server examples

Calfur:
- Adds cleanup of manually created `RID`s in GDScript.
- Adds cleanup of manually created `Rid`s in C#.

## Why
As memory management does not occur automatically when using RenderingServer directly, it is recommended to implement manual cleanup.

I think it makes sense to do so in the example in the docs as well, as this is not shown on the page yet.

### DIFF
--- a/tutorials/performance/using_servers.rst
+++ b/tutorials/performance/using_servers.rst
@@ -285,11 +285,12 @@ and moves a :ref:`class_CanvasItem` when the body moves.
         # ...
 
 
-    func _exit_tree():
-        # Free manually created RIDs when the node is removed.
-        PhysicsServer2D.free_rid(body)
-        PhysicsServer2D.free_rid(shape)
-        RenderingServer.free_rid(ci_rid)
+    func _notification(what):
+        if what == NOTIFICATION_PREDELETE:
+            # Free manually created RIDs when the node is destroyed.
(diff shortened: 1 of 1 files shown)


tutorials/performance/using_servers.rst
```diff
@@ -283,13 +284,23 @@ and moves a :ref:`class_CanvasItem` when the body moves.
         # See the section above on creating a sprite.
         # ...
 
+
+    func _exit_tree():
+        # Free manually created RIDs when the node is removed.
```
AThousandShips: This is not correct, this will mean that removing and re-adding this node to the tree will error as the RIDs are invalid, `_ready` is called only once, but `_exit_tree` every time

Calfur: Ah I didn't know that difference, thanks. So far I have always destroyed nodes which I removed from the tree, so I didn't run into this issue.

What's the recommended approach for the cleanup? NOTIFICATION_PREDELETE?

AThousandShips: That or recreate the data on `_enter_tree` instead

Calfur: I changed it to `NOTIFICATION_PREDELETE `, let me know if you would prefer `_enter_tree ` instead

PR #11767, opened 2026-02-17
Documents link behaviour on scene reload and add examples for link removal.

### LINKED ISSUE #11766: NavigationLinks created via NavigationServer2D.link_create persist after scene reload
**Your Godot version:**
4.6.stable.arch_linux

**Issue description:**
Creating NavigationLinks described in the documentation (with the link owner being set like: `NavigationServer2D.link_set_owner_id(link_rid, get_instance_id())`) and then reloading the scene results in the created links persisting, but losing the reference to the owner. It seems that duplicate links will also be created on each scene reload.
This can cause issues if you want to change behaviour based on the owner type (eg: Teleport links vs Jump links), as the "owner" in the metadata will be a null instance.

** To reproduce:**

1. Have a script  like:
```
class_name Teleporter extends Node2D


@export var link_start_position: Vector2
@export var link_end_position: Vector2
var link_rid: RID

func _ready() -> void:
	link_rid = NavigationServer2D.link_create()

	var link_owner_id: int = get_instance_id()
	var link_enter_cost: float = 1.0
	var link_travel_cost: float = 1.0
	var link_navigation_layers: int = 1
	var link_bidirectional: bool = true

	NavigationServer2D.link_set_owner_id(link_rid, link_owner_id)
	NavigationServer2D.link_set_enter_cost(link_rid, link_enter_cost)
	NavigationServer2D.link_set_travel_cost(link_rid, link_travel_cost)
	NavigationServer2D.link_set_navigation_layers(link_rid, link_navigation_layers)
	NavigationServer2D.link_set_bidirectional(link_rid, link_bidirectional)

	# Enable the link and set it to the default navigation map.
	NavigationServer2D.link_set_enabled(link_rid, true)
	NavigationServer2D.link_set_map(link_rid, get_world_2d().get_navigation_map())

	# Move the 2 link positions to their intended global positions.
	NavigationServer2D.link_set_start_position(link_rid, link_start_position)
	NavigationServer2D.link_set_end_position(link_rid, link_end_position)
```
3. Have a Node that uses a NavigationAgent to move
4. Connect the NavigationAgent `link_reached` signal to a function like:
```
func _agent_link_reached(details: Dictionary):
	var link_owner = details.get("owner", null)
	if link_owner == null:
		# Handle missing owner
		print("reached a link with no owner")
	elif link_owner is Teleporter:
		# Handle Teleport
		print("reached a teleporter")
		global_position = details.get("link_exit_position")
	else:
		# Handle default link type
		print("reached a link with no special type")

```
4. Add a Teleporter to the scene and navigate the agent through it. The link will work as expected on initial scene load. 
5. Reload the scene (eg: via `get_tree().reload_current_scene()`)
6. Attempt to navigate through the Teleporter again, the link_owner will be null. If you check `NavigationServer2d.map_get_links` you will see that there are now two links in the map.

** Suggested Fix **
If this is the intended behavior I think a note should be added to the documentation and perhaps add a call to `NavigationServer2D.free_rid` when the owner exits the tree in the example like:
```
func _exit_tree() -> void:
	if link_rid.is_valid():
		NavigationServer2D.free_rid(link_rid)

```
This will clear the previously created link when the scene is unloaded.

https://docs.godotengine.org/en/stable/tutorials/navigation/navigation_using_navigationlinks.html

If you know how to fix the issue you are reporting please
consider opening a pull request. We provide a tutorial on
using git here: https://contributing.godotengine.org/en/latest/organization/pull_requests/creating_pull_requests.html,
writing documentation at https://contributing.godotengine.org/en/latest/documentation/guidelines/index.html
and contributing to the class reference here: https://contributing.godotengine.org/en/latest/documentation/class_reference.html

delVhariant:
Fixes #11766

### DIFF
--- a/tutorials/navigation/navigation_using_navigationlinks.rst
+++ b/tutorials/navigation/navigation_using_navigationlinks.rst
@@ -199,7 +199,8 @@ The following script uses the NavigationServer to create a new navigation link.
     Additionally, the links created prior to the scene reload will have a null value for their owner
     as the object instance set via the ``link_set_owner_id`` function no longer exists.
 
-If you are creating navigation links as described during runtime, you may want to remove if the owner Node leaves the tree. This can be done with a call to ``free_rid`` function of your navigation server:
+If you are creating navigation links as described during runtime, you may want to remove it if the owner Node leaves the tree.
+This can be done with a call to the ``free_rid`` function of your navigation server:
 
 .. tabs::
  .. code-tab:: gdscript 2D GDScript


tutorials/navigation/navigation_using_navigationlinks.rst
```diff
@@ -191,3 +191,36 @@ The following script uses the NavigationServer to create a new navigation link.
...
         }
     }
+
+.. warning:: If you reload the scene, navigation links created in this way will still exist. This may result in duplicate navigation links. Additionally, the links created prior to the scene reload will have a null value for their owner as the object instance set via the ``link_set_owner_id`` function no longer exists.
+
+If you are creating navigation links as described during runtime, you may want to remove if the owner Node leaves the tree. This can be done with a call to ``free_rid`` function of your navigation server:
+
+.. tabs::
+ .. code-tab:: gdscript 2D GDScript
+    func _exit_tree() -> void:
```
AThousandShips, line 208:
```suggestion

    func _exit_tree() -> void:
```

tutorials/navigation/navigation_using_navigationlinks.rst
```diff
@@ -191,3 +191,36 @@ The following script uses the NavigationServer to create a new navigation link.
...
+        if link_rid.is_valid():
+            NavigationServer2D.free_rid(link_rid)
+
+    .. code-tab:: csharp 2D C#
+        public override void _ExitTree()
```
AThousandShips, line 214:
```suggestion

        public override void _ExitTree()
```

tutorials/navigation/navigation_using_navigationlinks.rst
```diff
@@ -191,3 +191,36 @@ The following script uses the NavigationServer to create a new navigation link.
...
+        {
+            if (_linkRid.IsValid)
```
AThousandShips, line 216:
```suggestion
            if (_linkRid.IsValid())
```

tutorials/navigation/navigation_using_navigationlinks.rst
```diff
@@ -191,3 +191,36 @@ The following script uses the NavigationServer to create a new navigation link.
...
+            {
+                NavigationServer2D.FreeRid(_linkRid);
+            }
+        }
+    .. code-tab:: gdscript 3D GDScript
```
AThousandShips, line 222:
```suggestion

    .. code-tab:: gdscript 3D GDScript

```

tutorials/navigation/navigation_using_navigationlinks.rst
```diff
@@ -191,3 +191,36 @@ The following script uses the NavigationServer to create a new navigation link.
...
+        func _exit_tree() -> void:
+            if link_rid.is_valid():
+                NavigationServer3D.free_rid(link_rid)
+
+    .. code-tab:: csharp 3D C#
+        public override void _ExitTree()
```
AThousandShips, line 230:
```suggestion

        public override void _ExitTree()
```

tutorials/navigation/navigation_using_navigationlinks.rst
```diff
@@ -191,3 +191,36 @@ The following script uses the NavigationServer to create a new navigation link.
...
+        {
+            if (_linkRid.IsValid)
```
AThousandShips, line 232:
```suggestion
            if (_linkRid.IsValid())
```

tutorials/navigation/navigation_using_navigationlinks.rst
```diff
@@ -191,3 +191,36 @@ The following script uses the NavigationServer to create a new navigation link.
             NavigationServer3D.LinkSetEndPosition(_linkRid, _linkEndPosition);
         }
     }
+
+.. warning:: If you reload the scene, navigation links created in this way will still exist. This may result in duplicate navigation links. Additionally, the links created prior to the scene reload will have a null value for their owner as the object instance set via the ``link_set_owner_id`` function no longer exists.
```
AThousandShips:
```suggestion
.. warning::

    If you reload the scene, navigation links created in this way will still exist.
    This may result in duplicate navigation links.
    Additionally, the links created prior to the scene reload will have a null value for their owner
    as the object instance set via the ``link_set_owner_id`` function no longer exists.
```

tutorials/navigation/navigation_using_navigationlinks.rst
```diff
@@ -191,3 +191,36 @@ The following script uses the NavigationServer to create a new navigation link.
...
+
+If you are creating navigation links as described during runtime, you may want to remove if the owner Node leaves the tree. This can be done with a call to ``free_rid`` function of your navigation server:
```
AThousandShips:
```suggestion
If you are creating navigation links as described during runtime, you may want to remove it if the owner Node leaves the tree.
This can be done with a call to the ``free_rid`` function of your navigation server:
```

AThousandShips (changes requested):

tutorials/navigation/navigation_using_navigationlinks.rst
delVhariant: RID IsValid is a property in C# according to: https://docs.godotengine.org/en/latest/tutorials/scripting/c_sharp/c_sharp_differences.html#rid, is this incorrect?

tutorials/navigation/navigation_using_navigationlinks.rst
delVhariant: Same as https://github.com/godotengine/godot-docs/pull/11767#discussion_r2819810477, the docs indicate IsValid is a property rather than a method.

tutorials/navigation/navigation_using_navigationlinks.rst
AThousandShips: My bad!

PR #11791, opened 2026-02-26
Remove bent normals notice about the compatibility renderer

LunaCapra:
With https://github.com/godotengine/godot/pull/114336 merged, bent normal map support is now available on all renderers.


skyace65:
Thanks!

Issue #11807, opened 2026-03-03
Meta: make `pyproject.toml` more useful when using dependency managers


ScatteredComet:
**Your Godot version:** 4.6

**Issue description:**

Currently, the tutorial for building the manual suggests using the `requirements.txt`. This misses a few things (such as dependency resolution, although we don't really need that) but mainly it means that no python version is actually being specified by the docs. I believe the versions specified in the `requirements.txt` need python 3.10 or above (at least that's what poetry told me).

This can easily be fixed by adding `[project]` metadata.

This has the added benefit of allowing tools such as poetry and uv to be used when creating venv's to build the docs.

**URL to the documentation page (if already existing):**

https://docs.godotengine.org/en/4.4/contributing/documentation/building_the_manual.html

ScatteredComet:
I suppose 1 downside to doing this is now all the dependencies are listed in two places, which kindof defeats the point of using a dependency manager...

However you can generate a `requirements.txt` from the `pyproject.toml` using

```
poetry self add poetry-plugin-export
poetry export -f requirements.txt --output requirements.txt --without-hashes
```

Note that this generates a few other dependencies other than the ones listed in the `requirements.txt`, so there are probably technically some missing from the current requirements. For example, the `requests` module is not included in the current `requirements.txt`, but it is used in `_tools/redirects/create_redirects.py`.

---

Maybe a different solution is just specifying the python_version in the `requirements.txt` like this:

```
...
pygments==2.19.2 ; python_version >= "3.10"
...
```

PR #11808, opened 2026-03-03
Meta: update `pyproject.toml`

ScatteredComet:
- This closes https://github.com/godotengine/godot-docs/issues/11807

This is a very poetry-centric toml; might be worth checking if this works with uv. 

The `.lock` file is committed as per [poetry's recommendations](https://python-poetry.org/docs/basic-usage/#committing-your-poetrylock-file-to-version-control).


ScatteredComet:
This would also mean that local venvs could be being created using a different method to the CI

https://github.com/godotengine/godot-docs/blob/f62bd3c35d8d4e99b4502141ecf4d79edef1106d/.github/workflows/ci.yml

which is probably a bad thing

Issue #11016, opened 2025-06-11
Incorrect link fragments


miruji:
**Issue description:**

I'm not sure if these links are actually incorrect, but while browsing the documentation, I noticed something seemed off.

For example:
https://docs.godotengine.org/en/stable/tutorials/2d/introduction_to_2d.html#d-in-2d
https://docs.godotengine.org/en/stable/tutorials/2d/introduction_to_2d.html#d-workspace

The fragment identifiers here are missing the numbers:
https://docs.godotengine.org/en/stable/tutorials/2d/introduction_to_2d.html#3d-in-2d
https://docs.godotengine.org/en/stable/tutorials/2d/introduction_to_2d.html#2d-workspace

If it's expected behavior that the fragment name does not match the heading text, or if there's another reason, then this is not an error.

ScatteredComet:
I've tried to have a look as to why this might be, but can't find any docs about URL fragments in Sphinx / rST / etc. I would have a guess that this is because the section name starts with a number, so its being removed by something?

I think the 3D in 2D section title isn't that descriptive anyway and could be changed to "Displaying 3D Nodes in 2D", which would fix that heading (although of course not fix the bug).

Calinou:
I guess Sphinx tries to make the heading a valid identifier, which means it can't start with a digit.

This can be worked around by defining a custom anchor in rST. We typically use this form as a convention:

`_doc_<page_slug>_<title_slug>`

For example:

```rst
Previous section text here.

.. _doc_introduction_to_2d_3d_in_2d:

3D in 2D
--------

Text here.
```

PR #11811, opened 2026-03-03
Fix section heading link fragments which start with a digit

ScatteredComet:
- Closes https://github.com/godotengine/godot-docs/issues/11016

---

- Does using `-`, `.`, `/` within a custom anchor break them? If so then some of my changes will need amending [Edit: they do, and have been removed]
- When the convention `_doc_<page_slug>_<title_slug>:` led to non-unique anchors (e.g. in `upgrading_to_godot_4.6.rst` & `docs_changelog.rst` where headings are used twice) I used `_doc_<page_slug>_<section_slug>_<subsection_slug>:`
- I did not check if all pre-existing section anchors fit these 2 conventions
- `binary_serialization_api.rst` headings contain lots of colons: what should we do? [Edit: they are fine as is]

I also think it would be a good idea to add something to the contributing docs and/or unit tests saying that you need to provide a custom anchor if you define a title which starts with a number, and define a convention.

# Notes on this PR

I used `ripgrep` with the commands `rg -U '^[0-9].*\n-'` and `rg -U '^[0-9].*\n~'` to search for these, which corresponds to searching for [line with a number at the start] + [line break] + [~ or -].

I also saw a lot of page anchors which dont align to

> The reference _doc_insert_your_title_here and the title should match.

from the ["writing the manual" page](https://contributing.godotengine.org/en/latest/documentation/manual/contributing_to_the_manual.html#titles), but I'll make another PR for that as its a large diff.

### DIFF
--- a/tutorials/platform/android/android_library.rst
+++ b/tutorials/platform/android/android_library.rst
@@ -167,6 +167,7 @@ Below we break-down the steps used to create the GLTF Viewer app.
       ...
   </activity>
 
+.. _doc_android_library_2_create_the_godot_project:
 
 2. Create the Godot project
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -241,6 +242,7 @@ Below we break-down the steps used to create the GLTF Viewer app.
 
       add_child(current_gltf_node)
 
+.. _doc_android_library_3_build_and_run_the_app:
 
 3. Build and run the app
 ~~~~~~~~~~~~~~~~~~~~~~~~
--- a/engine_details/architecture/internal_rendering_architecture.rst
+++ b/engine_details/architecture/internal_rendering_architecture.rst
@@ -422,6 +422,8 @@ release.
 - `servers/rendering/renderer_rd/effects/fsr.cpp <https://github.com/godotengine/godot/blob/4.2/servers/rendering/renderer_rd/effects/fsr.cpp>`__
 - `thirdparty/amd-fsr/ <https://github.com/godotengine/godot/tree/master/thirdparty/amd-fsr>`__
 
+.. _doc_internal_rendering_architecture_2d_rendering_techniques:
+
 2D rendering techniques
 -----------------------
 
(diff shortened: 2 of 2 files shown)


ScatteredComet:
From the CI, it looks like including those special characters break the links; I'll remove them. Would be nice to know exactly what characters are allowed though if we're going to define a convention.

ScatteredComet:
I tried to look for some specification for valid anchors but didn't manage to find anything, so I just changed all special characters to underscores.

tutorials/platform/android/android_library.rst
```diff
@@ -71,6 +71,8 @@ Below we break-down the steps used to create the GLTF Viewer app.
   - By locking to a specific orientation using the `android:screenOrientation attribute <https://developer.android.com/guide/topics/manifest/activity-element#screen>`_.
   - By declaring that the Activity will handle these configuration events using the `android:configChanges attribute <https://developer.android.com/guide/topics/manifest/activity-element#config>`_.
 
+.. _docandroid_library_1_create_the_android_app:
```
Calinou:
```suggestion
.. _doc_android_library_1_create_the_android_app:
```

Calinou:
> I also saw a lot of page anchors which dont align to
>
> > The reference _doc_insert_your_title_here and the title should match.
>
> from the ["writing the manual" page](https://contributing.godotengine.org/en/latest/documentation/manual/contributing_to_the_manual.html#titles), but I'll make another PR for that as its a large diff.

Changing page anchors will break translations, so I would recommend against this.

ScatteredComet:
`binary_serialization_api.rst`'s heading links are actually fine as is e.g.

https://docs.godotengine.org/en/stable/tutorials/io/binary_serialization_api.html#float

so I don't think there's anything else that needs to be added to this PR

ScatteredComet:
This is a different solution than the contributing docs used e.g. [here](https://github.com/godotengine/godot-contributing-docs/blob/e3de8a307ed0505f8be6ba023032d99d5bbfe62b/organization/areas.rst?plain=1#L4)

https://github.com/godotengine/godot-contributing-docs/pull/75#discussion_r2882697867 suggests that this PR might not work

PR #11826, opened 2026-03-06
Add some info on how to react to changes to an `Array` or `Dictionary` in a `@tool` script

mgjv:
Because normal runtime behaviour for an Array and Dictionary is to not have its setter called on insertions, deletions, or changes to elements, it is probably useful to explicitly call out that Arrays and Dictionaries changed in the Inspector in the Editor, as part of a `@tool` script, _do_ have the setter called. 

I searched quite a bit, before I stumbled onto the comment of dalexeev on https://github.com/godotengine/godot-proposals/issues/14065, explaining that this is (almost certainly) intended but undocumented behaviour. To help others like me, I submit this change to the documentation, which is the first place I looked for this info.

Update to running_code_in_the_editor.rst

### DIFF
--- a/tutorials/plugins/running_code_in_the_editor.rst
+++ b/tutorials/plugins/running_code_in_the_editor.rst
@@ -273,6 +273,66 @@ angle add a setter ``set(new_speed)`` which is executed with the input from the 
     but you can't access user variables. If you want to do so, other nodes have
     to run in the editor too.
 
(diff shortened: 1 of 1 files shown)


mhilbrunner:
Looks good to me too. Congrats on your first merged PR for Godot's documentation, and thanks for contributing! For future PRs, please use more specific commit messages. :)

PR #11837, opened 2026-03-10
Add note for fixing invisible player in Finishing up page of first 2D game tutorial

LSpeakmanOU:
Final page in the first 2D game tutorial proposes a couple ways to make a background. Due to z ordering issues, the player may be invisible. A small note is added to the background section on how to fix this.

### DIFF
--- a/getting_started/first_2d_game/07.finishing-up.rst
+++ b/getting_started/first_2d_game/07.finishing-up.rst
@@ -21,6 +21,7 @@ You could also add a background image, if you have one, by using a
 :ref:`TextureRect <class_TextureRect>` node instead.
 
 .. note::
+
     The ``Player`` may be invisible if the background is drawn on top of it. 
     Select the ``Player`` instance, then set **Ordering > Z Index** to ``1``.
 


getting_started/first_2d_game/07.finishing-up.rst
```diff
@@ -20,6 +20,10 @@ select "Layout" -> "Anchors Preset" -> "Full Rect" either in the toolbar at the
 You could also add a background image, if you have one, by using a
 :ref:`TextureRect <class_TextureRect>` node instead.
 
+.. note::
+    The ``Player`` may be invisible if the background is drawn on top of it.
```
AThousandShips, line 25:
```suggestion

    The ``Player`` may be invisible if the background is drawn on top of it. 
```

LSpeakmanOU:
> How consistently are you seeing this issue if you don't set the Z Index? If this is a consistent thing that happens the tutorial needs to be updated and it shouldn't be a note.

I noticed it happening on Godot 4.6.1(haven't tested any other versions) just when it comes to adding the background in the finishing up section. This issue didn't occur anywhere else in the tutorial.

Do you think it is sufficient to remove the note section header and integrate it into the text? Or would you propose something different? Thanks!

skyace65:
@LSpeakmanOU I'd like to confirm what exactly is happening before you make any changes. Could you please upload a zip file here of your finished project without the Z index change?

LSpeakmanOU:
[nozfix.zip](https://github.com/user-attachments/files/25918338/nozfix.zip)
@skyace65 Sure that makes sense, here you go(see above)

LSpeakmanOU:
@skyace65 Hey I just wanted to make this edit. The tutorial explicitly says to make it the first node under main. That is my bad. I believe no change is needed to the tutorial in this case. Thanks for looking this over

PR #11848, opened 2026-03-14
docs: fixed a typo

Zacky3181V:
deleted comma and put "as" word instead of it. Look better now. 

master branch is targeted


AThousandShips:
Thank you for your contribution, however this change is not correct. The original text is correct and this change breaks the meaning and is not grammatically correct

To clarify: the description says "it takes a start position (the mob's spawn position) and the player position as the arguments for the function"

PR #11849, opened 2026-03-14
Remove unnecessary viewport navigation in Your first 3D game

Zacky3181V:
targeting master branch.

Removed 2 sentences and an image because moving to a 3D viewport is not needed since `Signals` doc is always available. Considering the fact that after moving to 3D viewport, user will proceed with writing a code to the mob's script right after connecting the signal, opening 3D viewport makes no sense.


Calinou (changes requested):
The change makes sense to me, but the `image8` substitution should also be removed at the end of the page, along with its associated image file.

Zacky3181V:
For some reason I didn't get notification. I reflected your feedback. Should I update the image numbering after deleting image8?

Calinou:
> Should I update the image numbering after deleting image8?

I'd keep things as-is to reduce the size of the diff. @mhilbrunner What do you think?

Calinou (approved):
Looks good on my end.

I force-pushed a commit to squash changes, and also remove the image file that is no longer needed.

mhilbrunner (approved):

mhilbrunner:

PR #11861, opened 2026-03-19
[4.5] Release policy: Remove outdated table, link to stable

akien-mga:
This should be cherry-picked to all older branches.

### DIFF
--- a/about/release_policy.rst
+++ b/about/release_policy.rst
@@ -81,69 +81,9 @@ If you experience an issue using an older patch release, please upgrade to the
 latest patch release of that series and test again before reporting an issue
 on GitHub.
 
-+--------------+----------------------+--------------------------------------------------------------------------+
-| **Version**  | **Release date**     | **Support level**                                                        |
-+--------------+----------------------+--------------------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


about/release_policy.rst
```diff
@@ -81,69 +81,9 @@ If you experience an issue using an older patch release, please upgrade to the
...
-Pre-release Godot versions aren't intended to be used in production and are
-provided for testing purposes only.
-
-.. seealso::
-
-    See :ref:`doc_upgrading_to_godot_4` for instructions on migrating a project
-    from Godot 3.x to 4.x.
+For an overview of current, past, and future releases, refer to the table on
+the stable branch of the documentation:
+`Release support table <https://docs.godotengine.org/en/stable/about/release_policy.html#release-support-timeline>`__.
```
akien-mga, line 86: This hardcodes the `en` prefix, because I don't know of a way to make an URL that would be "this page on this branch in the user's detected language".

But this change won't make its way to localized versions of the docs anyway as I don't intend to update them further, so /shrug. If it did, translators would be able to use a localized URL too in their translation.

mhilbrunner:
Makes sense. We should really finally move this to the website...

akien-mga:
Cherry-picked to all branches down to `3.2`. The page doesn't exist before `3.2`.

Issue #11859, opened 2026-03-19
Add warning to use 32-bit version of FFmpeg for Ogg Theora encoding


PiewieP:
**4.6.1.stable**

**Issue description:**
Just spent a few hours trying to figure out why my Ogg Theora file conversions were all artifacted. I tried Shutter Encoder and FFmpeg, but both failed to create an output without artifacts. As a last resort I tried a 32-bit version of FFmpeg, which ended up working. Related FFmpeg issues can be found [here](https://github.com/BtbN/FFmpeg-Builds/pull/513), [here](https://github.com/GyanD/codexffmpeg/issues/196) and [here](https://github.com/GyanD/codexffmpeg/issues/200).

[link to docs to be updated](https://docs.godotengine.org/en/4.6/tutorials/animation/playing_videos.html#doc-playing-videos-recommended-theora-encoding-settings)

berarma:
I don't understand it. You've linked to the issues that have been already closed with the bugs fixed, and to the docs where there's already a warning to use the latest binaries from the official site for the latest bug fixes.

Which FFmpeg version were you using and where did you get it from? Did you build it by yourself?

SnowSquire:
> I don't understand it. You've linked to the issues that have been already closed with the bugs fixed, and to the docs where there's already a warning to use the latest binaries from the official site for the latest bug fixes.
> 
> Which FFmpeg version were you using and where did you get it from? Did you build it by yourself?

it broke again in at least 8.1.2 but now it's fixed in 9.0 64 bit from gyan. seems it gets broken and fixed so recommending a specific build would be beneficial

berarma:
I think the recommendation should be: use always the latest version and build.

If there's an issue, report it to the FFmpeg builder instead of changing Godot documentation or filing a report here.

PR #11865, opened 2026-03-21
Recommend 32-bit FFmpeg on Windows in Playing videos

Calinou:
- This closes https://github.com/godotengine/godot-docs/issues/11859.

### DIFF
--- a/tutorials/animation/playing_videos.rst
+++ b/tutorials/animation/playing_videos.rst
@@ -199,6 +199,10 @@ below with almost any input video format (AVI, MOV, WebM, …).
    multiplexer. It's highly recommended to use one of the latest static daily
    builds, or build from their master branch to get the latest fixes.
 
+   On Windows, make sure to use 32-bit builds of FFmpeg. Windows 64-bit builds
+   have known issues with Theora encoding, which result in artifacts in the
+   output file.
+
 .. UPDATE: When the FFmpeg bugfixes for https://trac.ffmpeg.org/ticket/11451 and
 .. https://trac.ffmpeg.org/ticket/11454 are included in a stable FFmpeg release,
 .. this warning can be removed. That will likely be FFmpeg 7.2 or 8.0, and will


mhilbrunner:
Oof. Thank you, merged.

fire:
Has an upstream report been attached?

berarma:
These problems and others were already reported by me and fixed upstream, that's why I added warning to the docs to use the latest daily builds as their main site already recommends for end users.

This warning shouldn't be required for the latest FFmpeg builds since several months ago, and if the problem had reappeared, I think we could report it and get it fixed again.

Links to upstream issues in the related bug report.

PiewieP:
I had missed the warning about using the latest daily builds while working on this, sorry for the confusion.
Where on the main site does FFmpeg recommend using daily builds?

I used version N-123549-g70537ec8e6-20260318 (32-bit) and I just tested daily build version n7.1.3-43 (64-bit) which I can also confirm works.

Maybe the current warning in the docs can be extended, eg. "On Windows, make sure to use either 64-bit daily builds or 32-bit builds of FFmpeg. Current Windows 64-bit builds have known issues with Theora encoding, which result in artifacts in the output file."?

berarma:
> I had missed the warning about using the latest daily builds while working on this, sorry for the confusion.

Adding more warnings won't help then. I think we could update the warning though and simply recommend the latest builds more prominently.

> Where on the main site does FFmpeg recommend using daily builds?

https://ffmpeg.org/download.html#releases

Not exactly what I said but along those lines. Always test latest builds for fixes.

> 
> I used version N-123549-g70537ec8e6-20260318 (32-bit) and I just tested daily build version n7.1.3-43 (64-bit) which I can also confirm works.
> 
> Maybe the current warning in the docs can be extended, eg. "On Windows, make sure to use either 64-bit daily builds or 32-bit builds of FFmpeg. Current Windows 64-bit builds have known issues with Theora encoding, which result in artifacts in the output file."?

Using the latest version and builds is recommended anyway over any older build. Older builds can have other not so obvious issues. There were many issues fixed for ogg/theora encoding in the last year.

berarma:
There's a point to be made on user builds. Users compiling FFmpeg by themselves must make sure libtheora is compiled with the right flags.

Maybe warn about it.

PR #11866, opened 2026-03-21
Remove outdated info about differences in 2D and 3D physics interpolation

ChudeRonczki:
The info on the page seems to describe how the engine worked before the SceneTreeFTI was introduced in Godot 4.5.

The "Global versus local interpolation" gives information that looks like the exact opposite of the current logic, where the SceneTreeFTI walks the whole hierarchy of both interpolated nodes and non-interpolated ones, but parented to interpolated, and sums the local transforms accordingly. Just like how the 2D inner workings are described.

The "Resetting physics interpolation" describes a difference that also looks to be no longer there. The code for NOTIFICATION_ENTER_TREE in Node3D::_notification seems to handle the reset automatically. The rest of the information in this section looks correct, but it just reiterates info from the "Using physics interpolation" article about resetting, and since there doesn't seem to be any difference anymore between 2D and 3D with this, it doesn't seem to fit this article anymore.

### DIFF
--- a/tutorials/physics/interpolation/advanced_physics_interpolation.rst
+++ b/tutorials/physics/interpolation/advanced_physics_interpolation.rst
@@ -21,6 +21,18 @@ for a Node, the children will recursively also be affected (as they default to
 inheriting the parent setting). This means you can easily disable interpolation for
 an entire subscene.
 
+.. figure:: img/physics_interpolation_mode.webp
+
+It is worth noting that, both in 2D and 3D, physics interpolation is performed
+on the **local transform** of each instance. During rendering, interpolated local 
+transforms are passed down to children.
+
+This means that if a parent has ``physics_interpolation_mode`` set to ``On``, 
+but the child is set to ``Off``, the child will still be interpolated if the parent
+is moving. *Only the child's local transform is uninterpolated.*
+Controlling the on / off behavior of nodes therefore requires some 
+thought and planning.
+
 The most common situation where you may want to perform your own interpolation is
 Cameras.
 
--- a/tutorials/physics/interpolation/2d_and_3d_physics_interpolation.rst
+++ b/tutorials/physics/interpolation/2d_and_3d_physics_interpolation.rst
@@ -6,50 +6,6 @@
 Generally 2D and 3D physics interpolation work in very similar ways. However, there
 are a few differences, which will be described here.
 
-Global versus local interpolation
----------------------------------
-
-- In 3D, physics interpolation is performed *independently* on the **global
-  transform** of each 3D instance.
-- In 2D by contrast, physics interpolation is performed on the **local transform**
-  of each 2D instance.
-
-This has some implications:
-
-- In 3D, it is easy to turn interpolation on and off at the level of each ``Node``,
-  via the :ref:`physics_interpolation_mode<class_Node_property_physics_interpolation_mode>`
-  property in the Inspector, which can be set to ``On``, ``Off``, or ``Inherited``.
-
-.. figure:: img/physics_interpolation_mode.webp
-    :align: center
-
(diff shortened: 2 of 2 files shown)


lawnjelly:
Ah yes, good spot, the existing info on the differences is out of date. I'll check this is okay to remove.

We'll probably need to reintroduce the text about uninterpolated nodes only affecting local interpolation, and still inheriting interpolation from the parent, as this is now shared behaviour in 2D and 3D.

ChudeRonczki:
I can take a stab at rewriting this section into the "Advanced physics interpolation" article, because it's now probably a better fit for it. Is it fine for me to do this under this pull request?

Calinou:
> I can take a stab at rewriting this section into the "Advanced physics interpolation" article, because it's now probably a better fit for it. Is it fine for me to do this under this pull request?

Yes, I think it makes sense to do this in the same PR.

lawnjelly:
Something to note that I just realised:

The old method of scene traversal is still available via `physics/3d/physics_interpolation/scene_traversal`, so technically that text _still makes sense in Legacy mode_.

However, legacy mode was included really for the changeover to prevent any rewrites needed for people who had written under the old system, long term it could probably be removed a few point releases down the line.

So I'm not convinced we need a specific section in the help to describe it, just a note on the `scene_traversal` project setting is enough. If technically possible we could include a link to the previous documentation section on the web, but it's not vital imo.

ChudeRonczki:
> The old method of scene traversal is still available via `physics/3d/physics_interpolation/scene_traversal`, so technically that text _still makes sense in Legacy mode_.

I might be misunderstanding the code, but I think this is not the case. There is indeed the legacy code for calculating transforms the way this doc describes, in Node3D::get_global_transform_interpolated, but it's under #if 1 ... #else, so practically unreachable. And the setting you pointed out seems to just influence the fashion in which the SceneTreeFTI walks the tree, always going from the root instead of starting with relevant branches, but shouldn't influence the results.

lawnjelly:
> I might be misunderstanding the code, but I think this is not the case. 

Ah yes, you are likely correct, I can't even remember myself as the author (can't remember last week, let alone months / years back :grin: ).

ChudeRonczki:
Ok, I reworded the parts from previously deleted info that still might be useful and moved it to the beginning of "Advanced physics interpolation", because that part was already talking about setting the interpolation mode per node, so it seemed fitting.

tutorials/physics/interpolation/advanced_physics_interpolation.rst
```diff
@@ -21,6 +21,20 @@ for a Node, the children will recursively also be affected (as they default to
...
+
+- Interpolated local transforms are passed down to children during rendering. 
+  This means that if a parent has ``physics_interpolation_mode`` set to
+  ``On``, but the child is set to ``Off``, the child will still be interpolated if
+  the parent is moving. *Only the child's local transform is uninterpolated.*
+  Controlling the on / off behavior of nodes therefore requires some
+  thought and planning.
+- On the positive side, pivot behavior in the scene tree is perfectly preserved
+  during interpolation, which gives super smooth behavior.
+
```
lawnjelly, line 35: I think we can omit the lines 35-36 now, as it was only relevant as a comparison between the old 2D and 3D methods. Both now respect pivots correctly, so doesn't really need a mention imo.

lawnjelly:
Aside from removing the line mentioned above this looks fine to me in terms of content. :+1:

tutorials/physics/interpolation/advanced_physics_interpolation.rst
ChudeRonczki: Done. Restructured it a bit more to not leave a list with a single item.

mhilbrunner:

PR #11879, opened 2026-03-26
Update particle builtins description in Spatial shader

Calinou:
- Depends on https://github.com/godotengine/godot/pull/117861.

If we want to merge this now, I can remove the `USERDATA1` note and split it to its own PR.

PS: `USERDATA1-6` aren't documented anywhere in the manual. What would be the best place to introduce it? These are only available in particle shaders.

### DIFF
--- a/tutorials/shaders/shader_reference/spatial_shader.rst
+++ b/tutorials/shaders/shader_reference/spatial_shader.rst
@@ -234,9 +234,16 @@ However, the value passed to the fragment shader still comes from ``VERTEX``.
 For instancing, the ``INSTANCE_CUSTOM`` variable contains the instance custom data. When using particles, this information
 is usually:
 
-* **x**: Rotation angle in radians.
-* **y**: Phase during lifetime (``0.0`` to ``1.0``).
-* **z**: Animation frame.
+* **x**: Current rotation angle in radians.
+* **y**: Time the particle has been active in seconds.
+* **z**: Current animation frame.
+* **w**: Particle lifetime in seconds.
+
+To get the lifetime phase (ranging from ``0.0`` to ``1.0``), use ``CUSTOM.y / CUSTOM.w``.
+
+Additionally, ``USERDATA1`` is defined with the following value in the built-in :ref:`class_ParticleProcessMaterial`:
+
+* **w**: Accumulated angle from angular velocity in radians, as applied by the angular velocity texture.
 
 This allows you to easily adjust the shader to a particle system using default particle material. When writing a custom particle
 shader, this value can be used as desired.
--- a/tutorials/shaders/shader_reference/spatial_shader.rst
+++ b/tutorials/shaders/shader_reference/spatial_shader.rst
@@ -234,9 +234,9 @@ However, the value passed to the fragment shader still comes from ``VERTEX``.
 For instancing, the ``INSTANCE_CUSTOM`` variable contains the instance custom data. When using particles, this information
 is usually:
 
-* **x**: Rotation angle in radians.
-* **y**: Phase during lifetime (``0.0`` to ``1.0``).
-* **z**: Animation frame.
+* **x**: Current rotation angle in radians.
+* **z**: Current animation frame.
+* **y / w** (division): Lifetime fraction from ``0.0`` to ``1.0``.
 
 This allows you to easily adjust the shader to a particle system using default particle material. When writing a custom particle
 shader, this value can be used as desired.
--- a/tutorials/shaders/shader_reference/particle_shader.rst
+++ b/tutorials/shaders/shader_reference/particle_shader.rst
@@ -80,6 +80,11 @@ Start and Process built-ins
 
 These properties can be accessed from both the ``start()`` and ``process()`` functions.
 
+When using the built-in :ref:`class_ParticleProcessMaterial`, ``USERDATA1`` is defined
+with the following value in the built-in:
+
+* **w**: Accumulated angle from angular velocity in radians, as applied by the angular velocity texture.
+
 +------------------------------------+-----------------------------------------------------------------------------------------------------------------------------------------+
 | Function                           | Description                                                                                                                             |
 +====================================+=========================================================================================================================================+


tutorials/shaders/shader_reference/spatial_shader.rst
```diff
@@ -234,9 +234,15 @@ However, the value passed to the fragment shader still comes from ``VERTEX``.
...
 is usually:
 
-* **x**: Rotation angle in radians.
-* **y**: Phase during lifetime (``0.0`` to ``1.0``).
-* **z**: Animation frame.
+* **x**: Current rotation angle in radians.
+* **y**: Time the particle has been active in seconds.
+  To get the lifetime phase (ranging from ``0.0`` to ``1.0``), use ``CUSTOM.y / CUSTOM.w``.
+* **z**: Current animation frame.
+* **w**: Particle lifetime in seconds.
```
paddy-exe:
```suggestion
* **x**: Current rotation angle in radians.
* **y**: Time the particle has been active in seconds.
* **z**: Current animation frame.
* **w**: Particle lifetime in seconds.

To get the lifetime phase (ranging from ``0.0`` to ``1.0``), use ``CUSTOM.y / CUSTOM.w``.
```

Just to make the reading flow cleaner since both CUSTOM.y and CUSTOM.w are connected to lifetime.

paddy-exe (changes requested):

paddy-exe:
> PS: `USERDATA1-6` aren't documented anywhere in the manual. What would be the best place to introduce it? These are only available in particle shaders.

I would argue [in the particle shader page](https://docs.godotengine.org/en/stable/tutorials/shaders/shader_reference/particle_shader.html) as a separate section with a barebone example on how to read/write the values.

paddy-exe (approved):

tutorials/shaders/shader_reference/spatial_shader.rst
```diff
@@ -234,9 +234,16 @@ However, the value passed to the fragment shader still comes from ``VERTEX``.
 For instancing, the ``INSTANCE_CUSTOM`` variable contains the instance custom data. When using particles, this information
 is usually:
 
-* **x**: Rotation angle in radians.
-* **y**: Phase during lifetime (``0.0`` to ``1.0``).
-* **z**: Animation frame.
+* **x**: Current rotation angle in radians.
+* **y**: Time the particle has been active in seconds.
+* **z**: Current animation frame.
+* **w**: Particle lifetime in seconds.
```
QbieShay: This is incorrect. I don't know how to easily phrase it, but it contains the randomness of the lifetime.

So for example a certain particle may have 0.9 in CUSTOM.w, meaning the particle lives for 90% of the specified lifetime. So when this particle dies, CUSTOM.y will contain 0.9. Dividing Y by W gives the correct result.

I count on you all to figure out a phrasing for this, i have no idea how to make it intuitive.

QbieShay (changes requested):

skyace65:
@Calinou Poke. Please see QbieShay's review

mhilbrunner:
Another gentle @Calinou bump :)

tutorials/shaders/shader_reference/spatial_shader.rst
Calinou: I've amended the description based on this, but I haven't tested it in-engine.

Is there a way to access the total particle lifetime in seconds? As I understand it, having access to the lifetime randomness *ratio* alone doesn't make it possible to figure out the total lifetime in seconds, and therefore calculate a phase from `0.0` to `1.0`.

For example, with a lifetime randomness of `0.6`, `CUSTOM.W` would be between `0.4` and `1.0` (randomly determined for each particle).

tutorials/shaders/shader_reference/spatial_shader.rst
```diff
@@ -234,9 +234,17 @@ However, the value passed to the fragment shader still comes from ``VERTEX``.
...
-* **z**: Animation frame.
+* **x**: Current rotation angle in radians.
+* **y**: Time the particle has been active in seconds.
+* **z**: Current animation frame.
+* **w**: Particle lifetime multiplier from lifetime randomness (lower is shorter).
+  For example, if the **Lifetime Randomness** property is set to ``0.6``,
+  this value will be between ``0.4`` and ``1.0``. If no lifetime randomness is used,
+  this is always ``1.0``.
+
+Additionally, ``USERDATA1`` is defined with the following value in the built-in :ref:`class_ParticleProcessMaterial`:
```
QbieShay: You can't read userdata in spatial shaders

tutorials/shaders/shader_reference/spatial_shader.rst
```diff
@@ -234,9 +234,17 @@ However, the value passed to the fragment shader still comes from ``VERTEX``.
 For instancing, the ``INSTANCE_CUSTOM`` variable contains the instance custom data. When using particles, this information
 is usually:
 
-* **x**: Rotation angle in radians.
-* **y**: Phase during lifetime (``0.0`` to ``1.0``).
-* **z**: Animation frame.
+* **x**: Current rotation angle in radians.
+* **y**: Time the particle has been active in seconds.
```
QbieShay: This is not correct. This is a ratio, but with no randomness in it. If a particle's lifetime is 2, CUSTOM.y will go from 0 to 1.

If custom.w is 0.5, custom.y will be from 0 to 0.5, so to get actual lifetime ratio it needs to be CUSTOM.y/CUSTOM.w.

I just wouldn't document what Y and W are, just document Y/W because it's way too convoluted otherwise

tutorials/shaders/shader_reference/spatial_shader.rst
Calinou: I moved the note to the Particle shaders page.

QbieShay (approved):

tutorials/shaders/shader_reference/spatial_shader.rst
```diff
@@ -234,9 +234,9 @@ However, the value passed to the fragment shader still comes from ``VERTEX``.
 For instancing, the ``INSTANCE_CUSTOM`` variable contains the instance custom data. When using particles, this information
 is usually:
 
-* **x**: Rotation angle in radians.
-* **y**: Phase during lifetime (``0.0`` to ``1.0``).
-* **z**: Animation frame.
+* **x**: Current rotation angle in radians.
+* **z**: Current animation frame.
+* **y / w** (division): Lifetime fraction from 0.0 to 1.0.
```
Mickeon, line 239: Mmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm.
```suggestion
* **y / w** (division): Lifetime fraction from ``0.0`` to ``1.0``.
```

tutorials/shaders/shader_reference/spatial_shader.rst
Mickeon: I think this is a question worth answering. Marking as unsolved to make it noticeable.

Mickeon (commented):
It's not _that_ big of a deal, but I would personally document **w** by itself, along with the division, because otherwise we're giving a rather vague, albeit useful instruction based on assumptions. Who knows, they could be needing that value for something else.

skyace65 (approved):

skyace65:
Thanks!

PR #11880, opened 2026-03-27
Explain the return value of undo() and redo() in UndoRedo

guillaumeriousat:
Some may feel that the return values of these methods are self-explanatory but I still felt the need to go look at the C++ source to make sure of what they meant.


AThousandShips:

PR #11885, opened 2026-03-29
Update AR to MR in passthrough documentation

insanepure:
I think it is a bit misleading.

Augmented Reality is often depicting the usage of a smartphone or tablet to see virtual objects in the real world.
While the same is happening here, for VR Headsets this is more reflected as Mixed Reality.
To avoid this confusion, I suggest renaming it to Mixed Reality.

### DIFF
--- a/tutorials/xr/mr_passthrough.rst
+++ b/tutorials/xr/mr_passthrough.rst
@@ -1,9 +1,9 @@
 .. _doc_openxr_passthrough:
 
-AR / Passthrough
+MR / Passthrough
 ================
 
-Augmented Reality is supported through various methods depending on the capabilities of the hardware.
+Mixed Reality is supported through various methods depending on the capabilities of the hardware.
 
 Headsets such as the Magic Leap and glasses such as TiltFive show the rendered result on
 `see-through displays <https://en.wikipedia.org/wiki/See-through_display>`__ allowing the user


skyace65 (approved):
From what I'm seeing online this would be a correct terminology change. However I'll wait for the XR Team to review before merging.

BastiaanOlij:
I very much dislike the MR word choice.

It's currently a Meta only term that has been repurposed.
Apple is abandoning the VR and AR nomenclature all together and calling it spatial computing.
Google is focusing on the XR terminology but reinterpreting it as eXtended Reality, which is problematic as well.
To many people Mixed Reality also still very much means Metas original purpose, e.g. mixing real life camera footage with game footage to better community how people play immersive games (See tools like [LIV](https://liv.tv/), [Meta Mixed Reality Capture](https://developers.meta.com/horizon/resources/mixed-reality-capture-and-casting/) and [MixCast](https://mixcast.me/tilt-five)).
Oh and lets not forget that Microsoft for the longest time hoped that we would classify MR as Microsoft Reality when they introduced the WMR platform.
MR is a hella confusing term.

Also our whole approach change, and what we've been advocating within the OpenXR working group, goes against this change.

Passthrough is a technique to bring Augmented Reality capabilities to a VR headset, so we can start building the AR applications of the future. Calling it anything else has just caused a lot of confusion in the space and puts developers on the wrong path.
We mirror this in our approach in Godot, if you stick to the core features, there is no difference between AR on glasses and AR through passthrough. It enables developers to develop AR applications today, and gain experience today, for the hardware that releases tomorrow.

Yes through the vendor plugin you can access a bunch of funky passthrough only features that are not possible with glasses, but developers need to take real care whether they want to use these features as they will likely find they've followed the wrong path if their end goals are the glasses form factor.

With feasible glasses likely hitting the market this year, and likely additional products launching in the very near future, we're just adding confusion by changing to a terminology which we'll need to change back soon after.

insanepure:
I will close this, as I opened another PR to add an XR Terminology Page that would fix the confusion better.

PR #11886, opened 2026-03-30
Add XR Terminology Page

insanepure:
After discussions around renaming AR to MR, I think another way to remove confusion and declare our status on it is by having a dedicated page explaining all of those terms and how we use them within the engine.

I made sure to use also some visual examples (thanks to Bastiaan for his videos!), I would love to have something for the handheld AR but I could not find a good image.
Overall also this is my first page so tell me if there is anything that should be changed or improved.

### DIFF
--- a/tutorials/xr/xr_terminology.rst
+++ b/tutorials/xr/xr_terminology.rst
@@ -16,7 +16,7 @@ XR (Extended Reality)
 
 In practice, XR refers to the complete system exposed through the
 :ref:`XRServer <class_xrserver>` and related APIs. This system abstracts away platform
-differences and provides a unified way to build extended reality applications.
+differences and provides a unified way to build XR applications.
 
 XR includes:
 
--- a/tutorials/xr/xr_terminology.rst
+++ b/tutorials/xr/xr_terminology.rst
@@ -79,7 +79,7 @@ The user continues to see their physical surroundings, while virtual objects are
 in a way that makes them appear part of that environment.
 
 In Godot, AR is treated as a single concept, regardless of the type of device used.
-This includes XR headsets using camera passthrough, see through glasses with displays, and handheld devices
+This includes XR headsets using camera passthrough, see-through glasses with displays, and handheld devices
 such as phones and tablets.
 
 For XR devices and AR glasses that support standards such as OpenXR or WebXR, AR
--- a/tutorials/xr/xr_terminology.rst
+++ b/tutorials/xr/xr_terminology.rst
@@ -98,7 +98,7 @@ the device.
 
 Implementation is done through platform-specific plugins:
 
-- `ARKit plugin (iOS) <https://docs.godotengine.org/de/4.x/tutorials/platform/ios/plugins_for_ios.html>`_
+- :ref:`ARKit plugin (iOS) <doc_plugins_for_ios>`
 - `ARCore plugin (Android) <https://github.com/godotvr/godot_arcore>`_
 
 These integrations are separate from the XR interface system and must be set up
--- a/tutorials/xr/xr_terminology.rst
+++ b/tutorials/xr/xr_terminology.rst
@@ -79,7 +79,7 @@ The user continues to see their physical surroundings, while virtual objects are
 in a way that makes them appear part of that environment.
 
 In Godot, AR is treated as a single concept, regardless of the type of device used.
-This includes XR headsets using camera passthrough, AR glasses, and handheld devices
+This includes XR headsets using camera passthrough, see through glasses with displays, and handheld devices
 such as phones and tablets.
 
 For XR devices and AR glasses that support standards such as OpenXR or WebXR, AR
@@ -87,7 +87,7 @@ functionality is available through the XR system. In these cases, applications c
 across different devices with minimal changes. Passthrough on VR headsets is one example
 of this, and is simply a technical method used to enable AR capabilities on such devices.
 
-Outside of Godot, this type of experience is often referred to as "Mixed Reality (MR)".
+Outside of Godot, this type of experience is sometimes referred to as "Mixed Reality (MR)".
 In Godot documentation, this is treated as Augmented Reality to avoid ambiguity.
 Passthrough is considered an implementation detail, not a separate category.
 
--- a/tutorials/xr/xr_terminology.rst
+++ b/tutorials/xr/xr_terminology.rst
@@ -3,7 +3,7 @@
 XR Terminology
 ==============
 
-This page defines how terms such as *XR*, *VR*, *AR*, and *MR* are used within Godot.
+This page defines how terms such as *XR*, *VR*, and *AR* are used within Godot.
 
 These terms are not always used consistently across the industry. In Godot, we use clear
 and practical definitions to avoid ambiguity and to better reflect how these technologies
@@ -12,11 +12,11 @@ are actually implemented.
 XR (Extended Reality)
 ---------------------
 
-**XR** is an umbrella term that covers all immersive technologies supported by Godot.
+**XR** is an umbrella term that covers all extended reality technologies supported by Godot.
 
 In practice, XR refers to the complete system exposed through the
 :ref:`XRServer <class_xrserver>` and related APIs. This system abstracts away platform
-differences and provides a unified way to build immersive applications.
+differences and provides a unified way to build extended reality applications.
 
 XR includes:
 
@@ -35,7 +35,7 @@ VR (Virtual Reality)
         <iframe src="https://www.youtube-nocookie.com/embed/xJKQ2ca5zVw" frameborder="0" allowfullscreen style="position: absolute; top: 0; left: 0; width: 100%; height: 100%;"></iframe>
     </div>
 
-**Virtual Reality (VR)** refers to fully immersive experiences where the user is placed
+**Virtual Reality (VR)** refers to fully **immersive** experiences where the user is placed
 inside a completely virtual environment.
 
 When using VR, the user does not see the real world. Instead, everything they see is
@@ -51,6 +51,18 @@ In Godot, VR typically involves:
 This is the most common use of XR in Godot. See
 :ref:`Setting up XR <doc_setting_up_xr>` for how to get started.
 
+.. note::
+
+    Even in VR applications, passthrough can be used if supported by the headset.
+
+    In this case, passthrough is typically used to reveal specific real-world
+    elements, such as a keyboard, mouse, or other peripherals, while the rest
+    of the experience remains fully virtual.
+
+    This is a hybrid use case and differs from Augmented Reality, as passthrough
+    is not used to place virtual content into the real world, but to selectively
+    expose parts of the real world within a VR experience.
+
 AR (Augmented Reality)
 ----------------------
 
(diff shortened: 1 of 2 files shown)


tutorials/xr/xr_terminology.rst
```diff
@@ -0,0 +1,125 @@
...
+.. raw:: html
+
+    <div style="position: relative; padding-bottom: 56.25%; height: 0; overflow: hidden; max-width: 100%; height: auto;">
+        <iframe src="https://www.youtube-nocookie.com/embed/8B8RnFokAFc" frameborder="0" allowfullscreen style="position: absolute; top: 0; left: 0; width: 100%; height: 100%;"></iframe>
+    </div>
+
+**Augmented Reality (AR)** refers to experiences where virtual content is overlaid onto
+the real world.
+
+In these applications, the user continues to see their physical surroundings, while
```
BastiaanOlij: Hmm, I think this section needs to change. The approach we took is that we don't see a difference whether AR works through passthrough, through phones or through glasses. These all fall under Augmented Reality.

In an ideal world you would be able to write your AR application once, and deploy it to all these three variants and that is the architecture we're going for. 

In practice for phone based AR this doesn't really work because no phone developer supports OpenXR, and our implementations of ARKit and ARCore are only mildly portable due to lacking standard approaches between these eco systems. OpenXR is however fully capable to support phone based AR and if someone would write such a runtime, you would be able to run a Godot AR application on such a device. Monado is rumoured to be working on such an implementation.

For passthrough and glasses based AR, provided they are OpenXR based, this portability is there today. We do not see these as two separate ecosystems and the code you write is no different. 
Godot works with MagicLeap glasses for instance, would work with Hololens if it wasn't for the UWP platform requirement, and is expected to work with other OpenXR based AR glasses that are being showcased.

Passthrough is merely a technical solution used to bring AR capabilities to VR headsets.

The bottom line is, as far as Godots architecture is concerned, we support Augmented Reality in all its forms ad see it as one thing.
It's purely what SDKs vendors provide that determine whether you have support out of the box through OpenXR (and/or WebXR, which also support AR), or whether you need a plugin implementing a proprietary API such as ARCore or ARKit.

tutorials/xr/xr_terminology.rst
```diff
@@ -0,0 +1,125 @@
...
+VR (Virtual Reality)
+--------------------
+
+.. raw:: html
+
+    <div style="position: relative; padding-bottom: 56.25%; height: 0; overflow: hidden; max-width: 100%; height: auto;">
+        <iframe src="https://www.youtube-nocookie.com/embed/xJKQ2ca5zVw" frameborder="0" allowfullscreen style="position: absolute; top: 0; left: 0; width: 100%; height: 100%;"></iframe>
+    </div>
+
+**Virtual Reality (VR)** refers to fully immersive experiences where the user is placed
```
BastiaanOlij: I would emphasise or underline immersive. That is really the key word.

tutorials/xr/xr_terminology.rst
```diff
@@ -0,0 +1,125 @@
...
+inside a completely virtual environment.
+
+When using VR, the user does not see the real world. Instead, everything they see is
```
BastiaanOlij, line 41: It might be good to add a note block here to indicate that with VR applications, passthrough can still be applicable. Passthrough, if supported  by the headset, is often used to cut out peripherals, be it a keyboard, mouse or joystick. This is a bit of a grey area where passthrough is not used in the traditional AR sense.

tutorials/xr/xr_terminology.rst
```diff
@@ -0,0 +1,125 @@
...
+
+- :ref:`ARKit plugin (iOS) <doc_plugins_for_ios>`
+- `ARCore plugin (Android) <https://github.com/godotvr/godot_arcore>`_
+
+These integrations are separate from the XR interface system and must be set up
+individually.
+
+MR (Mixed Reality)
+------------------
+
```
BastiaanOlij: I would leave out the MR section. We do not use MR when talking about the core features and XR foundations in Godot. We choose to stick with XR inline with our OpenXR adoption.

Including it only leads to confusion and arguments. Possibly if we ever get LIV, MRC, MixCast, or similar  support (which I really hope we do), it would make sense to have a section on MR indicating that for us Mixed Reality indicates embedding camera footage of the player into a 3rd person rendered view of the play environment.
However, until we actually support this, I think we should leave it out.

tutorials/xr/xr_terminology.rst
```diff
@@ -0,0 +1,125 @@
...
+This page defines how terms such as *XR*, *VR*, *AR*, and *MR* are used within Godot.
+
+These terms are not always used consistently across the industry. In Godot, we use clear
+and practical definitions to avoid ambiguity and to better reflect how these technologies
+are actually implemented.
+
+XR (Extended Reality)
+---------------------
+
+**XR** is an umbrella term that covers all immersive technologies supported by Godot.
```
BastiaanOlij: Immersive might be the wrong term here, Immersive is applicable for VR, it indicates that you are immersing yourself into a virtual world.
AR use cases stand in opposition to this, the point here is that it's **not** immersive but that instead, you're bringing the virtual into the real world and you are very much still in the real world.

I don't have a good suggestion for what term we should use.

tutorials/xr/xr_terminology.rst
insanepure: Maybe adding a small note to the part talking about AR Passthrough in XR Headsets saying that outside Godot, Mixed Reality is referencing this, but we do not do so to avoid confusion?
I just want to cover users asking about Mixed Reality and saying that Godot can deliver this, its just not name like that and should not be.

BastiaanOlij:
@dsnopek @m4gr3d the above is my take on this conversation, would love your input on this as well.

tutorials/xr/xr_terminology.rst
insanepure: I changed the section, emphasizing that we support AR through OpenXR and WebXR and adding a note that for handheld devices, they do not support the standard and have ARKit and ARCore, yet if they would support it, it would just work.

tutorials/xr/xr_terminology.rst
insanepure: highlighted it a bit

tutorials/xr/xr_terminology.rst
insanepure: Added a note

tutorials/xr/xr_terminology.rst
insanepure: You are right, I changed it to extended reality technologies to get back to the XR term that we are talking about.

tutorials/xr/xr_terminology.rst
insanepure: Added small part in AR about Mixed Reality instead of having it as a whole category.

tutorials/xr/xr_terminology.rst
dsnopek: > In practice for phone based AR this doesn't really work because no phone developer supports OpenXR

I don't really know the details, but there is apparently a way to use Monado on Android to implement smartphone AR using OpenXR. Would be a cool approach to document at some point!

tutorials/xr/xr_terminology.rst
```diff
@@ -0,0 +1,109 @@
...
+are actually implemented.
+
+XR (Extended Reality)
+---------------------
+
+**XR** is an umbrella term that covers all extended reality technologies supported by Godot.
+
+In practice, XR refers to the complete system exposed through the
+:ref:`XRServer <class_xrserver>` and related APIs. This system abstracts away platform
+differences and provides a unified way to build extended reality applications.
```
dsnopek: While I'm very happy to see "extended reality" written lower case, I think we should default to using "XR" almost everywhere. It's fine in the earlier paragraph defining XR, but here I think it should be:


```suggestion
differences and provides a unified way to build XR applications.
```

dsnopek (commented):
Thanks! Overall, this looks pretty good to me

tutorials/xr/xr_terminology.rst
insanepure: Alright, renamed that part.

tutorials/xr/xr_terminology.rst
```diff
@@ -0,0 +1,109 @@
...
+    </div>
+
+**Augmented Reality (AR)** refers to experiences where virtual content is overlaid onto
+the real world.
+
+The user continues to see their physical surroundings, while virtual objects are rendered
+in a way that makes them appear part of that environment.
+
+In Godot, AR is treated as a single concept, regardless of the type of device used.
+This includes XR headsets using camera passthrough, AR glasses, and handheld devices
```
m4gr3d: AR glasses are also referred to as see-through, and the difference in technology between see-through and passthrough devices may affect use-cases, and user experiences. 

Though I'm not sure it's worth mentioning the latter here, asides from qualifying AR glasses as see-through devices, since that by itself could fill a full page.

m4gr3d (approved):
Looks good overall.

tutorials/xr/xr_terminology.rst
BastiaanOlij: Agree, maybe something like `This includes XR headsets using camera passthrough, see through glasses with displays, and handheld devices such as phones and tablets`?

I would leave out the 2nd part. That's kind of the point I'm trying to make here.

Yes there are edge cases where passthrough is used in a way that can't be purely categorized as AR or VR, or situations like with Apple where you're really using the headset for immersive purposes primarily but where passthrough is mostly used to stay aware of your surroundings, even if you're not interacting with it.

But for core AR use cases, the technology shouldn't dictate the functionality, to be exact it's a trap. I fear that a lot of people who are too focused on passthrough are going to have a real big problem when their market actually turns out to be glasses which can't do the things they now rely on.
It's better to steer people to ignore HOW AR is done, but focus on WHAT they do with AR.

Long term I fully believe passthrough is going to loose its front and centre status. It's going to be relegated to being a supporting technology for immersive headsets, just so you can see your environment before entering an immersive world, or to still see peripherals, or simply to show you obstacles when you move out of your play space. There will be exceptions, but in the norm people aren't even going to think about pass through when they talk about AR, and MR will be a long forgotten thing.

tutorials/xr/xr_terminology.rst
BastiaanOlij: There was someone from Collabora on the XR channel in the Godot Discord asking about some settings, so I do believe they are working on it. I need to find out who it was and reach out to see how things are going. I agree that if this comes to fruition, this is something we should highlight on our manual pages. I think a lot of people will be interested in this as an alternative to ARKit/ARCore.

tutorials/xr/xr_terminology.rst
```diff
@@ -0,0 +1,109 @@
...
+
+Outside of Godot, this type of experience is often referred to as "Mixed Reality (MR)".
+In Godot documentation, this is treated as Augmented Reality to avoid ambiguity.
+Passthrough is considered an implementation detail, not a separate category.
+
+See :ref:`AR passthrough <doc_openxr_passthrough>` for an example of AR using the XR system.
+
+.. note::
+
+    Handheld platforms such as phones and tablets currently do not provide OpenXR support.
```
BastiaanOlij, line 98: As discussed above, but wanted to add it here as context. Before we publish this we should check on the state of the Monado implementation.
While it doesn't change this directly as this means installing a secondary component on a phone, it's definitely worth adding this as an option here.

tutorials/xr/xr_terminology.rst
```diff
@@ -0,0 +1,109 @@
...
+In Godot, AR is treated as a single concept, regardless of the type of device used.
+This includes XR headsets using camera passthrough, AR glasses, and handheld devices
+such as phones and tablets.
+
+For XR devices and AR glasses that support standards such as OpenXR or WebXR, AR
+functionality is available through the XR system. In these cases, applications can run
+across different devices with minimal changes. Passthrough on VR headsets is one example
+of this, and is simply a technical method used to enable AR capabilities on such devices.
+
+Outside of Godot, this type of experience is often referred to as "Mixed Reality (MR)".
```
BastiaanOlij: Nitpicking now but I would go for "is sometimes referred to". Obviously this usage of MR is big in Meta circles as they have adopted it fully in their marketing (even if they still promote the old meaning too), but as I mentioned before, in other circles this term is shunned.
Apple doesn't go anywhere near this terminology distancing itself totally from AR/VR/MR and XR.
Google doesn't either I think, preferring XR (re-invented as eXtended Reality) in their marketing
Not 100% sure what PICO does.

tutorials/xr/xr_terminology.rst
BastiaanOlij: Close enough I think. We're avoiding the assumption that they are all the same in this context.

tutorials/xr/xr_terminology.rst
insanepure: So I have checked on the Monado Discord and any documentation of Monado.
They said that even through OpenXR supports a mode for hand-held single view AR Applications, Monado does not support it on android yet.
Their focus right now is also highly HMD related so android is more experimental yet.

I would not feel confident in adding this as a current use case. I could add a note stating that Monado could bridge the gap here in the future, but that would require them to spent time on it which as said, is not a prio for them right now.
Same way it is not a prio to support OpenXR directly on android or ios.

insanepure: Some Comments from them about this:

"The android Monado drivers are just 3-dof orientation-only using motion sensor data. There is no passthrough on Android. There is no ARCore integration. You also need hit-testing against real world geometry for convincing AR object placement."

"You would need to write a driver that wraps around ARKit / ARCore as well as a backend for the compositor."

tutorials/xr/xr_terminology.rst
insanepure: Pico is calling their devices VR-MR Headsets btw, and they even have a blog for terminology (https://www.picoxr.com/my/blog/what-is-extended-reality).

But overall they use XR more often.

insanepure: I changed the wording to is sometimes referred to as suggested.

tutorials/xr/xr_terminology.rst
insanepure: I changed AR glasses to see-through glasses with displays. That also makes it more open, while official communications call those AR glasses or AI display glasses even.

tutorials/xr/xr_terminology.rst
dsnopek: > They said that even through OpenXR supports a mode for hand-held single view AR Applications, Monado does not support it on android yet.

Hm, my understanding was that [SOMAR used Monado](https://godotengine.org/article/godot-showcase-somar/) to do this on Android already. Perhaps they're still using a fork or unmerged PR?

insanepure: Yes, but that's just for the VR aspect of it, which they support through cardboard and with 3-dof orientation only. There is no AR yet within the android driver of Monado as seen in the comments I got from their discord.

dsnopek: Ah, ok, that makes sense!

BastiaanOlij: Ok, lets leave it as is then. Shame, I was hoping they were further along with this seeing how little love ARCore seems to get.

tutorials/xr/xr_terminology.rst
BastiaanOlij: The AI terminology is troublesome. It makes perfect sense for the no-display glasses where the emphasis is in AI interpreting what the glasses see and hear, and informing the user, but for display glasses it's just the AI hype that attaches the name, AI becomes just one of the many features.

Also seeing AI is sadly a rather toxic term atm, even though in our use case its mostly a positive technology, I'd leave that focus out for now.

tutorials/xr/xr_terminology.rst
BastiaanOlij: Seeing the discussion up above, I retract this suggestion :)

tutorials/xr/xr_terminology.rst
BastiaanOlij: Just adds to the confusion but I can't say I'm surprised with PICOs approach here.

In the end its all highly subjective what people feel is the better term.

Like I said on Discord, this is something I'm strongly opiniated by, but mine is just one opinion here. Though I do think that since XR is the terminology we use throughout, thats the term I'd like to keep promoting :)

BastiaanOlij (approved):
All looking good now, this is a nice addition to our documentation.

insanepure:
Awesome, my first approved PR :D anything missing?

skyace65 (approved):
Did a spelling and grammar check, everything looks good.

skyace65:

PR #11893, opened 2026-03-31
Improve the discoverability of print_debug()

peterbozso:
As the title says, the goal of this PR is to improve the discoverability of print_debug(). It tries to achieve that by putting a reference to it on the main Output panel doc page. I personally found this method very useful, and I think this would help other people to discover it quicker.


mhilbrunner (approved):

mhilbrunner:

peterbozso:
Thank you both!

@mhilbrunner Hopefully not the last. ;)

PR #11901, opened 2026-04-02
Improve custom iterator tutorial to use built in `iter` state instead of storing state in fields

aurpine:
The current custom iterator tutorial uses an example where the state is stored in a member variable instead of using the `iter` parameter provided by the `_iter_*` functions. Though it works, I believe it is bad practice as it is does not work as expected in the nested for loop case. It is ironic that the nested case is mentioned later despite the example not functioning properly there itself. We should steer users in the right direction by using the proper practice here.

For example, see the confusion documented here: https://forum.godotengine.org/t/deep-er-dive-on-godot-custom-iterators-and-the-mysterious-arg/92474

Associated `Object._iter_init` changes here: https://github.com/godotengine/godot/pull/118102

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_advanced.rst
+++ b/tutorials/scripting/gdscript/gdscript_advanced.rst
@@ -418,36 +418,37 @@ while() loops are the same everywhere:
 Custom iterators
 ----------------
 You can create custom iterators in case the default ones don't quite meet your
-needs by overriding the Variant class's ``_iter_init``, ``_iter_next``, and ``_iter_get``
+needs by overriding ``_iter_init()``, ``_iter_next()``, and ``_iter_get()``
 functions in your script. An example implementation of a forward iterator follows:
 
 ::
 
     class ForwardIterator:
-        var start
-        var current
-        var end
-        var increment
+        var _start
+        var _end
+        var _increment
 
-        func _init(start, stop, increment):
-            self.start = start
-            self.current = start
-            self.end = stop
-            self.increment = increment
+        func _init(start, end, increment):
+            _start = start
+            _end = end
+            _increment = increment
 
-        func should_continue():
-            return (current < end)
+        func _should_continue(current):
+            return current < _end
 
-        func _iter_init(arg):
-            current = start
-            return should_continue()
+        func _iter_init(iter):
+            # Initialize the state to store the current value.
+            iter[0] = _start
+            return _should_continue(iter[0])
 
-        func _iter_next(arg):
-            current += increment
-            return should_continue()
+        func _iter_next(iter):
+            iter[0] += _increment
+            return _should_continue(iter[0])
 
-        func _iter_get(arg):
-            return current
+        func _iter_get(iter):
+            # The state is not wrapped in an array for `_iter_get()`.
+            # The iteration value is the same as the state.
+            return iter
 
 And it can be used like any other iterator:
 
@@ -457,8 +458,19 @@ And it can be used like any other iterator:
     for i in itr:
         print(i) # Will print 0, 2, and 4.
 
-Make sure to reset the state of the iterator in ``_iter_init``, otherwise nested
-for-loops that use custom iterators will not work as expected.
+It is possible but discouraged to store the state in a member variable. 
+Multiple states are necessary in cases such as nested loops where the same
+iterator instance is used simultaneously. The ``iter`` parameter in 
+``_iter_init()`` and ``_iter_next()`` is a single-element array so that updates
+can persist. Whereas in ``_iter_get()``, the state is is not wrapped because it
+is supposed to be read-only.
+
+Returning ``true`` from ``_iter_init()`` and ``_iter_next()`` indicates that the
+iterator is valid. Returning ``false`` will terminate the loop.
+
+For more details see :ref:`_iter_init() <class_Object_private_method__iter_init>`,
+:ref:`_iter_next() <class_Object_private_method__iter_next>`, and
+:ref:`_iter_get() <class_Object_private_method__iter_get>`.
 
 Duck typing
 -----------


dalexeev:
Note that Class Reference links to the online documentation, [mentioning](https://github.com/godotengine/godot/blob/137fa73ed4d71d2e7b3a53e2ab6b2690e1f46149/doc/classes/Object.xml#L238) this as an alternative. Some people may find it easier to ignore the `iter` parameter and use object state instead. Yes, this makes it impossible to use the same iterator instance in nested loops, but you can use multiple instances (by creating them directly in the `for` loop header). This was the reason I didn't update this example after fixing the bug with the `iter` parameter and documenting the iterator API in Class Reference.

If you still think this example in the online documentation is too confusing and encourages bad practice, then we should also update the Class Reference so that it doesn't mention a no-longer-existent example. Also, both documentations should mention that storing iterator state in member variables instead of the `iter` parameter means you can't reuse the same iterator instance in nested loops.

aurpine:
Thanks for the quick reply! Opened https://github.com/godotengine/godot/pull/118102 to update the _iter_init docs.

I think storing the state in the instance is not much simpler than using the parameter method to warrant the dangers of the implementation. As an aside, I think explaining that storing the state fails the nested case takes more effort than showing the proper way to implement it. 

Thanks for your work on the `iter` parameter documentation! I found the tutorial page first and was confused at the unused arg before finding the class reference which clarified things. To be honest, this tutorial is a bit redundant now that it looks very similar to `_iter_init`.

tutorials/scripting/gdscript/gdscript_advanced.rst
```diff
@@ -425,29 +425,29 @@ functions in your script. An example implementation of a forward iterator follow
 
     class ForwardIterator:
         var start
-        var current
         var end
         var increment
 
         func _init(start, stop, increment):
```
dalexeev: It might also make sense to rename the `stop` parameter to `end`, given that's the name of the corresponding member variable. It might also be worth adding the `_` prefix to member variables to emphasize their private role and remove warnings about name shadowing (and remove the `self.` prefix as unnecessary).

tutorials/scripting/gdscript/gdscript_advanced.rst
```diff
@@ -425,29 +425,29 @@ functions in your script. An example implementation of a forward iterator follow
...
-        func should_continue():
-            return (current < end)
+        func should_continue(current):
+            return current < end
 
-        func _iter_init(arg):
-            current = start
-            return should_continue()
+        func _iter_init(iter):
+            # Initialize the state to store the current value
```
dalexeev:
```suggestion
            # Initialize the state to store the current value.
```

tutorials/scripting/gdscript/gdscript_advanced.rst
```diff
@@ -425,29 +425,29 @@ functions in your script. An example implementation of a forward iterator follow
...
-            current += increment
-            return should_continue()
+        func _iter_next(iter):
+            iter[0] += increment
+            return should_continue(iter[0])
 
-        func _iter_get(arg):
-            return current
+        func _iter_get(iter):
+            # The state is not wrapped in an array for _iter_get
```
dalexeev:
```suggestion
            # The state is not wrapped in an array for `_iter_get()`.
```

tutorials/scripting/gdscript/gdscript_advanced.rst
dalexeev, line 422:
```diff
 You can create custom iterators in case the default ones don't quite meet your
-needs by overriding the Variant class's ``_iter_init``, ``_iter_next``, 
+needs by overriding ``_iter_init()``, ``_iter_next()``, and ``_iter_get()``
 functions in your script. An example implementation of a forward iterator follows:
```

tutorials/scripting/gdscript/gdscript_advanced.rst
```diff
@@ -457,8 +457,9 @@ And it can be used like any other iterator:
     for i in itr:
         print(i) # Will print 0, 2, and 4.
 
-Make sure to reset the state of the iterator in ``_iter_init``, otherwise nested
-for-loops that use custom iterators will not work as expected.
+For more details see :ref:`_iter_init<class_Object_private_method__iter_init>`,
+:ref:`_iter_next<class_Object_private_method__iter_next>`, and
+:ref:`_iter_get<class_Object_private_method__iter_get>`.
```
dalexeev:
```suggestion
For more details see :ref:`_iter_init() <class_Object_private_method__iter_init>`,
:ref:`_iter_next() <class_Object_private_method__iter_next>`, and
:ref:`_iter_get() <class_Object_private_method__iter_get>`.
```

tutorials/scripting/gdscript/gdscript_advanced.rst
dalexeev, line 1: It's probably also worth mentioning here why you shouldn't store the iterator state in a member variable, but rather use the `iter` parameter.

aurpine:
Squashed and made more improvements to the code snippet. Made `_should_continue` private and added type hints.

dalexeev:
> and added type hints

See [contributing docs](https://contributing.godotengine.org/en/latest/documentation/guidelines/docs_writing_guidelines.html#dynamic-vs-static-typing):

> The code examples in the documentation should follow a consistent style not to confuse users. As static type hints are an optional feature of GDScript, we chose to stick to writing dynamic code. This leads to writing GDScript that is concise and accessible.
>
> The exception is topics that explain static typing concepts to users.

tutorials/scripting/gdscript/gdscript_advanced.rst
```diff
@@ -457,8 +458,18 @@ And it can be used like any other iterator:
     for i in itr:
         print(i) # Will print 0, 2, and 4.
 
-Make sure to reset the state of the iterator in ``_iter_init``, otherwise nested
-for-loops that use custom iterators will not work as expected.
+It is possible but discouraged to store the state in a member variable. 
+Multiple states are necessary in cases such as nested for-loops where multiple
```
dalexeev:
```suggestion
Multiple states are necessary in cases such as nested ``for`` loops where multiple
```

or

```suggestion
Multiple states are necessary in cases such as nested ``for``-loops where multiple
```

or (if the previous one is invalid, I'm not sure about the RST syntax):


```suggestion
Multiple states are necessary in cases such as nested ``for``\ -loops where multiple
```

or just


```suggestion
Multiple states are necessary in cases such as nested loops where multiple
```

tutorials/scripting/gdscript/gdscript_advanced.rst
```diff
@@ -457,8 +458,18 @@ And it can be used like any other iterator:
...
+loops use the same iterator instance simultaneously. The ``iter`` parameter in 
+``_iter_init()`` and ``_iter_next()`` is a single-element array that acts as a
+pointer to the state. The ``_iter_get`` state is read-only so it is unwrapped.
```
dalexeev: 1. It might be better to avoid using the word "pointer".
2. Technically, this doesn't guarantee a read-only state, you can still use one more nested array.

```suggestion
``_iter_init()`` and ``_iter_next()`` is a single-element array that acts as a
pointer to the state. The ``_iter_get`` state is supposed to be read-only so it is unwrapped.
```

tutorials/scripting/gdscript/gdscript_advanced.rst
```diff
@@ -457,8 +458,18 @@ And it can be used like any other iterator:
...
+
+Return ``true`` from ``_iter_init()`` and ``_iter_next()`` to indicate that the
+iterator is valid. Returning false will terminate the loop.
```
dalexeev:
```suggestion
iterator is valid. Returning ``false`` will terminate the loop.
```

tutorials/scripting/gdscript/gdscript_advanced.rst
```diff
@@ -457,8 +458,18 @@ And it can be used like any other iterator:
...
+
+For more details see :ref:`_iter_init()<class_Object_private_method__iter_init>`,
+:ref:`_iter_next()<class_Object_private_method__iter_next>`, and
+:ref:`_iter_get()<class_Object_private_method__iter_get>`.
```
dalexeev:
```suggestion
For more details see :ref:`_iter_init() <class_Object_private_method__iter_init>`,
:ref:`_iter_next() <class_Object_private_method__iter_next>`, and
:ref:`_iter_get() <class_Object_private_method__iter_get>`.
```

The space is optional, but I find it improves readability and is used frequently in this documentation (except for Class Reference, which is generated from XML). The space does not affect the appearance of the document.

tutorials/scripting/gdscript/gdscript_advanced.rst
```diff
@@ -457,8 +458,19 @@ And it can be used like any other iterator:
     for i in itr:
         print(i) # Will print 0, 2, and 4.
 
-Make sure to reset the state of the iterator in ``_iter_init``, otherwise nested
-for-loops that use custom iterators will not work as expected.
+It is possible but discouraged to store the state in a member variable. 
+Multiple states are necessary in cases such as nested loops where the same
+iterator instance is used simultaneously. The ``iter`` parameter in 
+``_iter_init()`` and ``_iter_next()`` is a single-element array so that updates
+can persist. Whereas in ``_iter_get``, the state is is not wrapped because it
```
dalexeev:
```suggestion
can persist. Whereas in ``_iter_get()``, the state is is not wrapped because it
```

tutorials/scripting/gdscript/gdscript_advanced.rst
```diff
@@ -457,8 +458,19 @@ And it can be used like any other iterator:
...
+is supposed to be read-only.
+
+Return ``true`` from ``_iter_init()`` and ``_iter_next()`` to indicate that the
+iterator is valid. Returning ``false`` will terminate the loop.
```
dalexeev:
```suggestion
Returning ``true`` from ``_iter_init()`` and ``_iter_next()`` will indicate that the
iterator is valid. Returning ``false`` will terminate the loop.
```

dalexeev (approved):
Looks good to me. :+1:

mhilbrunner (approved):

mhilbrunner:

aurpine:
Thanks @mhilbrunner! Can you also look at https://github.com/godotengine/godot/pull/118102 since it's related?

PR #11903, opened 2026-04-03
Document Linux/*BSD support for HDR output

Calinou:
- Clarify that Forward+/Mobile is required for HDR output.
- Remove trailing whitespace throughout the page, so that diffs can be cleaner later on.

### DIFF
--- a/tutorials/rendering/hdr_output.rst
+++ b/tutorials/rendering/hdr_output.rst
@@ -3,8 +3,8 @@
 HDR output
 ==========
 
-HDR output is a feature that enables presentation of High Dynamic Range (HDR) visuals on 
-HDR-capable screens. HDR **output** is not to be confused with the internal HDR rendering that is 
+HDR output is a feature that enables presentation of High Dynamic Range (HDR) visuals on
+HDR-capable screens. HDR **output** is not to be confused with the internal HDR rendering that is
 used by Godot for both Standard Dynamic Range (SDR) output and HDR output modes.
 
 Enabling HDR output in your project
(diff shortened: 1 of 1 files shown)


tutorials/rendering/hdr_output.rst
```diff
@@ -3,111 +3,116 @@
...
-2. Configure the :ref:`Rendering Device Driver<class_ProjectSettings_property_rendering/rendering_device/driver>` project setting to the following:
+3. Configure the :ref:`Rendering Device Driver<class_ProjectSettings_property_rendering/rendering_device/driver>`
+   project setting to the following:
 
-- macOS: metal
-- iOS: metal
-- Windows: d3d12
+- macOS: ``metal`` *(default)*
+- iOS: ``metal`` *(default)*
+- Windows: ``d3d12``
```
paddy-exe, line 29: Windows is default since 4.6 so this should be changed

EDIT: Not sure if that should be done in this PR or as a separate one.


```suggestion
- Windows: ``d3d12`` *(default)*
```

paddy-exe (changes requested):

tutorials/rendering/hdr_output.rst
Calinou: See the note at the end of the section. It's the default for projects created since 4.5 only; projects created on older versions still default to Vulkan.

skyace65:
Superseded by #11920. Though we should note the difference on the X11/Wayland page at some point. I may open a PR for that later.

PR #11904, opened 2026-04-03
Fix: Replace fake constructor with _Ready() initialization

bonskki:
fix: replace void constructor with _Ready() override

MySprite2D() was declared as a void method, making it a no-op that Godot never calls. Replaced with _Ready() so initialization logic (GD.Print) actually executes on node load.

### DIFF
--- a/getting_started/step_by_step/scripting_first_script.rst
+++ b/getting_started/step_by_step/scripting_first_script.rst
@@ -148,7 +148,7 @@ Add the following code to your script:
 
  .. code-tab:: csharp C#
 
-    public MySprite2D()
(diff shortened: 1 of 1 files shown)


raulsntos:
This section is explaining the `_init` GDScript function, which is equivalent to the class constructor in C#, so using `_Ready` is not correct. Perhaps you misunderstood `MyScript()` to be a void method instead of the class constructor because you named your class differently, the constructor matches the name of the class and earlier in this tutorial this class is called `MyScript` so it matches.

skyace65:
Closing per raulsntos' comment

PR #11906, opened 2026-04-04
Mention `RWLock` is similar to `shared_mutex` instead of timed variant

DeeJayLSP:
Technically a follow-up of https://github.com/godotengine/godot/pull/117801/

`RWLock` doesn't use any of the timed features, so it's better to mention the closest STL data type is `shared_mutex` instead of its timed variant.

Also, `RWLock` did in fact use `shared_timed_mutex` internally before https://github.com/godotengine/godot/pull/117801/ changed it to a plain `shared_mutex`.


Ivorforce (approved):

mhilbrunner (approved):

mhilbrunner:
Thank you!

PR #11917, opened 2026-04-09
Use `is` instead of `typeof()` in OpenXR spatial entities

Calinou:
Type checks using `is` are more idiomatic in GDScript.

The method spelling was incorrect anyway, which caused the code not to work.

- See https://github.com/godotengine/godot-docs-user-notes/discussions/630#discussioncomment-16470005.


dsnopek (approved):
Thanks! Looks good to me :-)

m4gr3d (approved):

skyace65:
Thanks!

PR #11918, opened 2026-04-10
Explain the ctrl-scroll can be used to change Grid Floor

oninoshiko:
Using gridmaps I was finding it rather painful to move up and down gridfloor levels, and thought "there has to be a shortcut for this." It's not documented, but it does exist and I had to figure out on my own. So here's a minor change so it's references where I would expect it to be. Hopefully it will help someone else find it.


Calinou:
- Superseded by https://github.com/godotengine/godot-docs/pull/11919.

PR #11920, opened 2026-04-10
Update HDR output manual page based on the 4.7 dev cycle

### LINKED ISSUE #11913: HDR output: No warning against using emissive with max linear value
**Your Godot version:**
4.7

**Issue description:**

The HDR output docs suggest that it is fine to use emissive materials with the `Window.get_output_max_linear_value()` so long as they are not in a scene with a `WorldEnvironment`. But when using emissive materials without a `WorldEnvironment` and with VoxelGI, a scene will have different indirect lighting behaviour depending on the dynamic range of the Window.

In the following screenshots, no colours are clipping in either SDR or HDR, but the glow that is seen on the torus is significantly larger and brighter in the HDR version than the SDR version:

SDR | HDR
-|-
Image|Image

(Note: although the HDR screenshot appears to be clipping, this is just because these are SDR screenshots)

MRP: [hdr_output.zip](https://github.com/user-attachments/files/26547316/hdr_output.zip)

https://docs.godotengine.org/en/latest/tutorials/rendering/hdr_output.html

**Notes to team:**

I will handle updating these docs before the 4.7 release. I have some ideas on how to better structure the page such that readers will not fall into these sorts of traps.

I also have a good idea of how to give a good 3D demo using unshaded materials instead of emissive materials.

allenwp:
### What problem(s) does this PR solve?

- Fixes #11913
- Supersedes #11903

### What is the rationale for the approach used in this PR?

This PR addresses the some of the same points as #11903 (removes trailing whitespace, adds requirement for mobile or forward+ renderer), but intentionally does not add the "On Linux/*BSD, HDR output is not supported on X11 due to display server limitations" comment to these instructions. My opinion is that these initial setup instructions have a lot of text, so adding more explanation to them makes for a more difficult user experience when configuring a project to use HDR output. In short, I take the stance that "less is more" for this list.

This PR also removes the comment about macOS needing the metal driver, as this is [no longer the case](https://github.com/godotengine/godot/pull/118083).

This PR adds instruction relating to new changes to the editor introduced in https://github.com/godotengine/godot/pull/118079 and updates the example GDScript to use the signals added by https://github.com/godotengine/godot/pull/115666 and https://github.com/godotengine/godot/pull/118076

Most importantly, I have reworked the guidance for when to use `Window.get_output_max_linear_value()` to address #11913.

I have not yet included any comments about how to adjust system reference/max luminance. Of specific importance is how Windows laptops behave, which is bonkers. I'll add this information in a followup PR sometime before 4.7 stable, but it is not necessary for beta 1.

### Was AI used to create a portion of this PR?

No.

### Are there any parts of this PR that you are uncertain of or require special attention from reviewers?

Much of these changes are required before https://github.com/godotengine/godot-website/pull/1302 goes live.

Depends on:
- https://github.com/godotengine/godot-demo-projects/pull/1287
- https://github.com/godotengine/godot/pull/118079


allenwp:
Whoops, I forgot to add the note that HDR output is not currently supported on Android. I’ll add that to this PR sometime next week (I’m back in the office on Tuesday). Adding this note won’t affect the rest of this PR, so it’s still ready for review.

Calinou (approved):
Looks good to me.

mhilbrunner (approved):

mhilbrunner:
Thank you! Merged.

PR #11937, opened 2026-04-20
Use 2nd person in godot_notifications.rst

BestAmericanGirl:
Some minor editing to use the 2nd person (you) instead of 3rd person (one). This is more consistent with the rest of the godot documentation and is better suited to educational writing. If this PR is acceptable I can expand it to cover other articles that also use the 3rd person to refer to the reader.

### DIFF
--- a/tutorials/best_practices/godot_notifications.rst
+++ b/tutorials/best_practices/godot_notifications.rst
@@ -39,14 +39,14 @@ And many of the callbacks that *do* exist in Nodes don't have any dedicated
 methods, but are still quite useful.
 
 - :ref:`Node::NOTIFICATION_PARENTED <class_Node_constant_NOTIFICATION_PARENTED>`:
-  a callback that triggers anytime one adds a child node to another node.
+  a callback that triggers anytime you add a child node to another node.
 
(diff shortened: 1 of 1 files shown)


Calinou (approved):
Looks good to me.

mhilbrunner (approved):

mhilbrunner:

tutorials/best_practices/godot_notifications.rst
```diff
@@ -39,14 +39,14 @@ And many of the callbacks that *do* exist in Nodes don't have any dedicated
 methods, but are still quite useful.
 
 - :ref:`Node::NOTIFICATION_PARENTED <class_Node_constant_NOTIFICATION_PARENTED>`:
-  a callback that triggers anytime one adds a child node to another node.
+  a callback that triggers anytime you add a child node to another node.
```
Mickeon, line 42: While the deed is done now and it's a good improvement, in these few cases it *may* have been genuinely beneficial to use passive verbs, instead, because these notifications can also be sent by the engine "doing its thing", not just as a result of user code.
```suggestion
  a callback that triggers anytime a child node is added to another node.
```
(also a "callback" that "triggers" are both *technically* the wrong terms for this but the rest of the tutorial sticks to it so, whatever)

PR #11956, opened 2026-04-29
Fix controller mapping docs

Nintorch:
Currently (more specifically, since Godot 4.5 when we integrated SDL) Godot's `Input.add_joy_mapping()` is broken, it will add a new mapping on top of an already existing mapping from SDL (if it exists), so I'm not sure it should be used as an example at the moment. See https://github.com/godotengine/godot/pull/118606 for more information.
If the mapping doesn't exist, then it (hopefully) should work fine since that way the mapping process is basically the same as before Godot 4.5: raw data from the OS passes to SDL, SDL passes it (hopefully?) without modification to Godot, Godot uses its own mapping system. I haven't tested this possibility yet though.

This PR also adds 2 new methods to create a new mapping that were originally described in the official SDL gamepad database file. The link for the testcontroller application was taken from [here](https://github.com/libsdl-org/SDL/issues/9265#issuecomment-2002565224).

This PR also suggests the users to contribute their mappings (that were created with the methods described by the SDL developers) to the SDL repository.

After https://github.com/godotengine/godot/pull/118606 is merged, we should be able to restore the references to `Input.add_joy_mapping()` on this page.

### DIFF
--- a/tutorials/inputs/controllers_gamepads_joysticks.rst
+++ b/tutorials/inputs/controllers_gamepads_joysticks.rst
@@ -333,11 +333,11 @@ or creating an issue in the `SDL repository <https://github.com/libsdl-org/SDL/i
 
 .. note::
 
-   Note that there are "generic" controllers on the market (usually their
-   ``Input.get_joy_info(device)["raw_name"]`` property contains ``"USB Gamepad"`` string),
-   and different generic controllers may use the same chipset, but they would have a different button placement,
-   so creating a mapping for one of those controllers will most likely conflict with other ones,
-   because the engine has no way of differentiating between controllers with the same chipset.
+    Note that there are "generic" controllers on the market (usually their
+    ``Input.get_joy_info(device)["raw_name"]`` property contains ``"USB Gamepad"`` string),
+    and different generic controllers may use the same chipset, but they would have a different button placement,
+    so creating a mapping for one of those controllers will most likely conflict with other ones,
+    because the engine has no way of differentiating between controllers with the same chipset.
 
 My controller works on a given platform, but not on another platform.
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/tutorials/inputs/controllers_gamepads_joysticks.rst
+++ b/tutorials/inputs/controllers_gamepads_joysticks.rst
@@ -283,8 +283,19 @@ the SDL game controller database used by Godot or the
 `Godot game controller database <https://github.com/godotengine/godot/blob/master/core/input/godotcontrollerdb.txt>`__.
 In this case, you will need to create a custom mapping for your controller.
 
-There are many ways to create mappings. One option is to use the mapping wizard
-in the `official Joypads demo <https://godotengine.org/asset-library/asset/2785>`__.
+.. Nintorch: Currently Godot's Input.add_joy_mapping() is broken, it will add a new mapping
+   on top of an already existing mapping from SDL (if it exists), so I'm not sure it
+   should be used as an example at the moment. See GH-118606 in Godot's main repository
+   for more information.
+
+   One option is to use the mapping wizard
+   in the `official Joypads demo <https://godotengine.org/asset-library/asset/2785>`__.
+
+There are many ways to create mappings.
+One option is to start Steam in Big Picture mode, configure the controller and then look in ``config/config.vdf``
+in the Steam installation directory for the ``SDL_GamepadBind`` entry.
+Another option is to use `SDL's testcontroller application <https://www.libsdl.org/tmp/testcontroller.zip>`__
+(the link only provides a Windows executable).
 Once you have a working mapping for your controller, you can test it by defining
 the ``SDL_GAMECONTROLLERCONFIG`` environment variable before running Godot:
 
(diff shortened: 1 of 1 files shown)


tutorials/inputs/controllers_gamepads_joysticks.rst
```diff
@@ -283,8 +283,18 @@ the SDL game controller database used by Godot or the
...
+   on top of an already existing mapping from SDL (if it exists), so I'm not sure it
+   should be used as an example at the moment. See GH-118606 in Godot's main repository
+   for more information.
+
+   One option is to use the mapping wizard
+   in the `official Joypads demo <https://godotengine.org/asset-library/asset/2785>`__.
+One option is to start Steam in Big Picture mode, configure the controller and then look in "config/config.vdf"
+in the Steam installation directory for the "SDL_GamepadBind" entry.
+Another option is to use `SDL's testcontroller application <https://www.libsdl.org/tmp/testcontroller.zip>`__
+(the link only provides a Windows executable).
```
Calinou: Move the comment above to fix the HTML output. Also use double backticks for file paths and code references:

```suggestion
.. Nintorch: Currently Godot's Input.add_joy_mapping() is broken, it will add a new mapping
   on top of an already existing mapping from SDL (if it exists), so I'm not sure it
   should be used as an example at the moment. See GH-118606 in Godot's main repository
   for more information.

   One option is to use the mapping wizard
   in the `official Joypads demo <https://godotengine.org/asset-library/asset/2785>`__.

There are many ways to create mappings.
One option is to start Steam in Big Picture mode, configure the controller and then look in ``config/config.vdf``
in the Steam installation directory for the ``SDL_GamepadBind`` entry.
Another option is to use `SDL's testcontroller application <https://www.libsdl.org/tmp/testcontroller.zip>`__
(the link only provides a Windows executable).
```

Otherwise, it looks like this currently:

Image

Calinou (changes requested):

Nintorch:
Thank you for your review, Calinou!

I also got [this](https://github.com/libsdl-org/SDL/pull/15480#issuecomment-4346056111) response from slouken after I made mappings for 2 of my controllers that were too generic (they report themselves as "USB Gamepad" and their USB vendor/product IDs conflict with already existing mappings from SDL_GameControllerDB, and these mappings are incorrect for my controllers, so the chipset can be used by several different types of physical controllers and we have no way of differentiating them, and no way to create proper mappings that wouldn't conflict with other controllers).
Should we add a warning for people who might want to contribute the mappings for their controllers in case they're too generic as well?

mhilbrunner (approved):

mhilbrunner:
> Should we add a warning for people who might want to contribute the mappings for their controllers in case they're too generic as well?

May fit some kind of SDL contributor's docs better (not sure I'd look at this docs page for guidance on that), but I can also see how one might end up there from this page, so as long as its not more than a few sentences, sure, why not.

Also getting people to think about "hey, this is all open source, I *can* contribute back stuff I need" isn't the worst. :)

Nintorch:
How about a note like this?
"Note that there are "generic" controllers on the market (usually their `Input.get_joy_info(device)["raw_name"]` property contains `"USB Gamepad"` string), and different generic controllers may use the same chipset, but they would have a different button placement, so creating a mapping for one of those controllers will most likely conflict with other ones, because the engine has no way of differentiating between controllers with the same chipset."

Calinou:
> How about a note like this? "Note that there are "generic" controllers on the market (usually their `Input.get_joy_info(device)["raw_name"]` property contains `"USB Gamepad"` string), and different generic controllers may use the same chipset, but they would have a different button placement, so creating a mapping for one of those controllers will most likely conflict with other ones, because the engine has no way of differentiating between controllers with the same chipset."

Yes, I agree we should add this note.

tutorials/inputs/controllers_gamepads_joysticks.rst
```diff
@@ -304,14 +315,29 @@ the ``SDL_GAMECONTROLLERCONFIG`` environment variable before running Godot:
...
+`official SDL gamepad database <https://github.com/libsdl-org/SDL/blob/main/src/joystick/SDL_gamepad_db.h>`__,
+or creating an issue in the `SDL repository <https://github.com/libsdl-org/SDL/issues>`__.
+
+.. note::
+
+   Note that there are "generic" controllers on the market (usually their
+   ``Input.get_joy_info(device)["raw_name"]`` property contains ``"USB Gamepad"`` string),
+   and different generic controllers may use the same chipset, but they would have a different button placement,
+   so creating a mapping for one of those controllers will most likely conflict with other ones,
+   because the engine has no way of differentiating between controllers with the same chipset.
```
Calinou:
```suggestion
    Note that there are "generic" controllers on the market (usually their
    ``Input.get_joy_info(device)["raw_name"]`` property contains ``"USB Gamepad"`` string),
    and different generic controllers may use the same chipset, but they would have a different button placement,
    so creating a mapping for one of those controllers will most likely conflict with other ones,
    because the engine has no way of differentiating between controllers with the same chipset.
```

Calinou (approved):
Looks good to me.

Nintorch:
Thank you for your reviews, Calinou and mhilbrunner! :)

mhilbrunner (approved):

mhilbrunner:
Thank you! Merged.

PR #11957, opened 2026-04-30
Add physics interpolation warning to class ref method reparent

3ter:
I noticed that the `reparent()` method can cause visible artifacts when physics interpolation is enabled and the node is moved to a very different transform.

This adds a warning to the method reference so users are reminded to call `reset_physics_interpolation()` after reparenting when appropriate.


AThousandShips:

3ter:
Thank you, I didn't realize that the class reference lives somewhere else 🙈 .

AThousandShips:
Easy mistake! We're trying to make it easier to notice and direct to the right place but we haven't found a solution yet

Calinou:
- Superseded by https://github.com/godotengine/godot/pull/119106.

PR #11958, opened 2026-04-30
Fix formatting and indentation in ios_plugin.rst

VojtaStruhar:
I believe the codeblock is not supposed to span all the way to the end of the page. It swallowed up some more documentation that follows it.

I'm not experienced with RST, I view it through GitHub. The docs following the codeblocks relate to the step **6**, but seem indented to me, still. I de-indented  the entire thing by two tabs, but maybe one would be enough.

### DIFF
--- a/tutorials/platform/ios/ios_plugin.rst
+++ b/tutorials/platform/ios/ios_plugin.rst
@@ -116,37 +116,37 @@ To build an iOS plugin:
             "
             StringPlistKeyToInput:string_input="Type something"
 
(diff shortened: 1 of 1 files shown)


Calinou (approved):
Tested locally, it works as expected.

mhilbrunner (approved):

mhilbrunner:
Thank you, and thanks for testing Calinou! Merged.

VojtaStruhar:
Thanks!

PR #11959, opened 2026-04-30
Add "Upgrading from Godot 4.6 to Godot 4.7" page

raulsntos:
This documents all the breaking changes in 4.7, like we did for previous versions:

- https://github.com/godotengine/godot-docs/pull/11618
- https://github.com/godotengine/godot-docs/pull/11223

_Updated as of commit godotengine/godot@3df26a02c446710c979daa541b74f87edeca81b0_

_Preview: https://github.com/raulsntos/godot-docs/blob/breaking_changes_4.7/tutorials/migrating/upgrading_to_godot_4.7.rst_

### DIFF
--- a/tutorials/migrating/upgrading_to_godot_4.7.rst
+++ b/tutorials/migrating/upgrading_to_godot_4.7.rst
@@ -1,0 +1,343 @@
+.. _doc_upgrading_to_godot_4.7:
+
+Upgrading from Godot 4.6 to Godot 4.7
+=====================================
+
+For most games and apps made with 4.6 it should be relatively safe to migrate to 4.7.
+This page intends to cover everything you need to pay attention to when migrating
+your project.
+
+Breaking changes
+----------------
+
+If you are migrating from 4.6 to 4.7, the breaking changes listed here might
+affect you. Changes are grouped by areas/systems.
+
+This article indicates whether each breaking change affects GDScript and whether
+the C# breaking change is *binary compatible* or *source compatible*:
+
+- **Binary compatible** - Existing binaries will load and execute successfully without
+  recompilation, and the runtime behavior won't change.
+- **Source compatible** - Source code will compile successfully without changes when
+  upgrading Godot.
+
+Core
+~~~~
+
+========================================================================================================================  ===================  ====================  ====================  ============
+Change                                                                                                                    GDScript Compatible  C# Binary Compatible  C# Source Compatible  Introduced
+========================================================================================================================  ===================  ====================  ====================  ============
+**Object**
+Method ``is_class`` changes ``class`` parameter type from ``String`` to ``StringName``                                    |✔️|                 |✔️ with compat|      |✔️|                  `GH-118582`_
+**ZIPPacker**
+Method ``start_file`` adds new ``permissions`` and ``modified_time`` optional parameters                                  |✔️|                 |✔️ with compat|      |✔️|                  `GH-115946`_
+**OptimizedTranslation**
+Method ``generate`` changes return type from ``void`` to ``bool``                                                         |✔️|                 |❌|                  |✔️|                  `GH-119563`_
+========================================================================================================================  ===================  ====================  ====================  ============
+
+2D
+~~
+
+========================================================================================================================  ===================  ====================  ====================  ============
+Change                                                                                                                    GDScript Compatible  C# Binary Compatible  C# Source Compatible  Introduced
+========================================================================================================================  ===================  ====================  ====================  ============
+**CPUParticles2D**
+Method ``request_particles_process`` adds new ``process_time_residual`` optional parameter                                |✔️|                 |✔️ with compat|      |✔️|                  `GH-109142`_
+**GPUParticles2D**
+Method ``request_particles_process`` adds new ``process_time_residual`` optional parameter                                |✔️|                 |✔️ with compat|      |✔️|                  `GH-109142`_
+========================================================================================================================  ===================  ====================  ====================  ============
+
+3D
+~~
+
+========================================================================================================================  ===================  ====================  ====================  ============
+Change                                                                                                                    GDScript Compatible  C# Binary Compatible  C# Source Compatible  Introduced
+========================================================================================================================  ===================  ====================  ====================  ============
+**CPUParticles3D**
+Method ``request_particles_process`` adds new ``process_time_residual`` optional parameter                                |✔️|                 |✔️ with compat|      |✔️|                  `GH-109142`_
+**GPUParticles3D**
+Method ``request_particles_process`` adds new ``process_time_residual`` optional parameter                                |✔️|                 |✔️ with compat|      |✔️|                  `GH-109142`_
+========================================================================================================================  ===================  ====================  ====================  ============
+
+GUI nodes
+~~~~~~~~~
+
+=================================================================================================================================================  ===================  ====================  ====================  ============
+Change                                                                                                                                             GDScript Compatible  C# Binary Compatible  C# Source Compatible  Introduced
+=================================================================================================================================================  ===================  ====================  ====================  ============
+**Control**
+Property ``accessibility_live`` changes type from ``DisplayServer.AccessibilityLiveMode`` to ``AccessibilityServer.AccessibilityLiveMode``         |✔️|                 |❌|                  |❌|                  `GH-116839`_
+**RichTextLabel**
+Enum field ``ImageUpdateMask.UPDATE_WIDTH_IN_PERCENT`` renamed to ``ImageUpdateMask.UPDATE_WIDTH_UNIT``                                            |❌|                 |✔️|                  |❌|                  `GH-112617`_
+Method ``add_image`` changes ``width`` parameter type from ``int`` to ``float``                                                                    |✔️|                 |✔️ with compat|      |✔️|                  `GH-112617`_
+Method ``add_image`` changes ``height`` parameter type from ``int`` to ``float``                                                                   |✔️|                 |✔️ with compat|      |✔️|                  `GH-112617`_
+Method ``add_image`` renames ``width_in_percent`` parameter to ``width_unit`` and changes type from ``bool`` to ``RichTextLabel.ImageUnit``        |✔️|                 |✔️ with compat|      |❌|                  `GH-112617`_
+Method ``add_image`` renames ``height_in_percent`` parameter to ``height_unit`` and changes type from ``bool`` to ``RichTextLabel.ImageUnit``      |✔️|                 |✔️ with compat|      |❌|                  `GH-112617`_
+Method ``update_image`` changes ``width`` parameter type from ``int`` to ``float``                                                                 |✔️|                 |✔️ with compat|      |✔️|                  `GH-112617`_
+Method ``update_image`` changes ``height`` parameter type from ``int`` to ``float``                                                                |✔️|                 |✔️ with compat|      |✔️|                  `GH-112617`_
+Method ``update_image`` renames ``width_in_percent`` parameter to ``width_unit`` and changes type from ``bool`` to ``RichTextLabel.ImageUnit``     |✔️|                 |✔️ with compat|      |❌|                  `GH-112617`_
+Method ``update_image`` renames ``height_in_percent`` parameter to ``height_unit`` and changes type from ``bool`` to ``RichTextLabel.ImageUnit``   |✔️|                 |✔️ with compat|      |❌|                  `GH-112617`_
+=================================================================================================================================================  ===================  ====================  ====================  ============
+
+Text
+~~~~
+
+========================================================================================================================  ===================  ====================  ====================  ============
+Change                                                                                                                    GDScript Compatible  C# Binary Compatible  C# Source Compatible  Introduced
+========================================================================================================================  ===================  ====================  ====================  ============
+**Font**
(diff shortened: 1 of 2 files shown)


Zacky3181V:
Should the #11804 issue be added to this PR as well?

raulsntos:
Yes. Thanks for letting me know. Included in 519923acdd9df18cf3c535eac50ad858f5de3774.

Ivorforce:
Heya, could you add a note on https://github.com/godotengine/godot/pull/107679?

I'm not sure how the document is written, but the basic idea is this:

> If you use the `audio_bus_override` feature on `Area2D` / `3D`, AND you use the `AudioStreamPlayer` default `area_mask` (just layer `1` ticked), after the update you will need to reset the mask to layer `1` — otherwise, the bus overrides will stop working. If the mask is set to anything except layer `1`, it will continue to work as expected.

raulsntos:
Thanks, included in 142d191c3b4e5da99e3f50c8dec855c95861b3f9.

kleonc:
These should be included:
- https://github.com/godotengine/godot/pull/113228 (cc @dalexeev)
- https://github.com/godotengine/godot/pull/118948 (cc @mihe)

mihe:
This needs to be included as well, even if the actionable is (unfortunately) just "re-tweak all your soft bodies":

- godotengine/godot#116041

I'll try to draft some section for this and godotengine/godot#118948.

EDIT: Oh, and this as well I guess:

- godotengine/godot#114198

Nintorch:
Should this PR be included?
- https://github.com/godotengine/godot/pull/116274

raulsntos:
Updated to include all mentioned PRs. I'm not sure I fully understood all those changes, so please review that the notes make sense, or let me know if you prefer a different message altogether.

tutorials/migrating/upgrading_to_godot_4.7.rst
```diff
@@ -0,0 +1,321 @@
...
+    The default ``area_mask`` for ``AudioStreamPlayer`` was changed to ``0`` (disabled) (`GH-107679`_).
+    If you use the ``audio_bus_override`` feature on ``Area2D`` or ``Area3D``, **and** you
+    use the ``AudioStreamPlayer`` default ``area_mask`` (just layer ``1`` ticked), you will need
+    to reset the mask to layer ``1`` — otherwise, the bus overrides will stop working.
+    If the mask was set to anything except layer ``1``, it will continue to work as expected.
+
+.. note::
+
+    The distance ``D`` in planes when using the Jolt physics engine has been corrected to use
+    the right sign convention (`GH-118948`_).
```
mihe:
```suggestion
    When using Jolt Physics as the 3D physics engine, ``WorldBoundaryShape3D`` will now use the same
    convention as Godot when applying ``WorldBoundaryShape3D.plane.d``, resulting in the sign of the plane
    distance being interpreted in the opposite way compared to Godot 4.6 (`GH-118948`_). You will need to flip
    the sign yourself to get the same behavior as in Godot 4.6.
```

tutorials/migrating/upgrading_to_godot_4.7.rst
```diff
@@ -0,0 +1,321 @@
...
+
+.. note::
+
+    The default ``SoftBody3D`` mass was changed from ``0`` to ``1`` in the Jolt physics engine (`GH-116041`_).
```
mihe: There's technically two breakages rolled into godotengine/godot#116041, so I'm splitting them up here.

```suggestion
.. note::

    When using Jolt Physics as the 3D physics engine, ``SoftBody3D`` will no longer default its mass to ``0``,
    which resulted in an automatically calculated weight of 1 kg per point, resulting in a very high total
    mass for the body. Now instead it will default to 1 kg for the entire ``SoftBody3D``, same as Godot
    Physics (`GH-116041`_).

.. note::

    When using Jolt Physics as the 3D physics engine, ``SoftBody3D`` will now apply
    ``SoftBody3D.linear_stiffness`` in a way that better matches Godot Physics, and in a way that's more
    appropriate in general (`GH-116041`_). This will affect every ``SoftBody3D`` instance in one way or
    another, meaning you will need to re-tweak properties like ``SoftBody3D.linear_stiffness`` and
    ``SoftBody3D.damping_coefficient`` to achieve your desired behavior.
```

tutorials/migrating/upgrading_to_godot_4.7.rst
```diff
@@ -0,0 +1,321 @@
...
+
+.. note::
+
+    The area's overlap signals are now emitted whenever a soft body enters/exits its shapes in the
+    Jolt physics engine (`GH-114198`_). To work around this breaking change, configure your collision
+    layers/masks so that the relevant interactions between ``Area3D`` and ``SoftBody3D`` are ignored.
```
mihe:
```suggestion
    When using Jolt Physics as the 3D physics engine, ``Area3D`` will now report overlaps with ``SoftBody3D``
    from its various signals and methods (`GH-114198`_). To work around this breaking change, configure your
    collision layers/masks such that any undesirable interactions between ``Area3D`` and ``SoftBody3D`` are
    ignored.
```

mihe:
@jrouwe You might want to take a look at the Jolt-related items listed under the physics heading here, in case there's a more precise way to phrase them, or if I've left out something important.

jrouwe:
> @jrouwe You might want to take a look at the Jolt-related items listed under the physics heading here, in case there's a more precise way to phrase them, or if I've left out something important.

Looks good to me!

tutorials/migrating/upgrading_to_godot_4.7.rst
```diff
@@ -0,0 +1,335 @@
...
+    from its various signals and methods (`GH-114198`_). To work around this breaking change, configure your
+    collision layers/masks such that any undesirable interactions between ``Area3D`` and ``SoftBody3D`` are
+    ignored.
+
+Input
+~~~~~
+
+.. note::
+
+    The device ID for mouse and keyboard has changed from ``0`` because some joypads may use it (`GH-116274`_).
```
AdriaandeJongh: From `0` to what?

tutorials/migrating/upgrading_to_godot_4.7.rst
```diff
@@ -0,0 +1,335 @@
...
+    ``CanvasItem`` now avoids adding the antialiasing feather when drawing lines (`GH-105122`_).
+    The feather made lines appear thicker than intended, projects that relied on this behavior
+    will have to be updated to draw a thicker line width.
+
+Physics
+~~~~~~~
+
+.. note::
+
+    The default ``area_mask`` for ``AudioStreamPlayer`` was changed to ``0`` (disabled) (`GH-107679`_).
```
AdriaandeJongh: To `0` from what?

tutorials/migrating/upgrading_to_godot_4.7.rst
```diff
@@ -0,0 +1,335 @@
...
+    ``InputEvent.DEVICE_ID_MOUSE`` and ``InputEvent.DEVICE_ID_KEYBOARD`` instead.
+
+
+GDScript
+~~~~~~~~
+
+.. note::
+
+    Setting the element of packed arrays no longer calls the setter for the entire packed array property
+    (`GH-113228`_).
```
dalexeev, line 252: Please add a mention of [115763](https://github.com/godotengine/godot/pull/115763). Users must now return a value explicitly in methods that implicitly inherit a return type from a parent class.

tutorials/migrating/upgrading_to_godot_4.7.rst
```diff
@@ -0,0 +1,335 @@
...
+**RichTextLabel**
+Enum field ``ImageUpdateMask.UPDATE_WIDTH_IN_PERCENT`` renamed to ``ImageUpdateMask.UPDATE_WIDTH_UNIT``                                            |❌|                 |✔️|                  |❌|                  `GH-112617`_
+Method ``add_image`` changes ``width`` parameter type from ``int`` to ``float``                                                                    |✔️|                 |✔️ with compat|      |✔️|                  `GH-112617`_
+Method ``add_image`` changes ``height`` parameter type from ``int`` to ``float``                                                                   |✔️|                 |✔️ with compat|      |✔️|                  `GH-112617`_
+Method ``add_image`` renames ``width_in_percent`` parameter to ``width_unit`` and changes type from ``bool`` to ``RichTextLabel.ImageUnit``        |✔️|                 |✔️ with compat|      |✔️ with compat|      `GH-112617`_
+Method ``add_image`` renames ``height_in_percent`` parameter to ``height_unit`` and changes type from ``bool`` to ``RichTextLabel.ImageUnit``      |✔️|                 |✔️ with compat|      |✔️ with compat|      `GH-112617`_
```
bruvzg: C# source compatibility has caveats, if a full list of args is used it will work as is, but if argument is used by name (e.g., `widthInPercent: true`), it won't. So not sure if it should be marked as `❌`, or have a footnote, or renaming argument automatically implies this.

raulsntos: Because the compat method uses the name `widthInPercent`, it won't break users using named arguments.

tutorials/migrating/upgrading_to_godot_4.7.rst
raulsntos: The note mentions later the value used to be `1`. I added `from 1` to this line, is this better?

tutorials/migrating/upgrading_to_godot_4.7.rst
raulsntos: They were changed to the value of the new constants `DEVICE_ID_MOUSE` and `DEVICE_ID_KEYBOARD` which are mentioned later in the note. How would you rewrite the note to make it less confusing?

tutorials/migrating/upgrading_to_godot_4.7.rst
raulsntos: Added in f3cfaeb20. Feel free to suggest different content for the note, my explanation may not be accurate.

tutorials/migrating/upgrading_to_godot_4.7.rst
bruvzg: At least for some cases it does not work (see https://github.com/godotengine/godot/issues/120148):

`AddImage(tex, 110, widthInPercent : true);` works in 4.6.3, but fails to build in 4.7.

bruvzg: Seems like it is caused by compat method not having default arguments (in the C# binding, arguments are set in GDScript binding).

raulsntos: Ah, you are right, in the generated overload all parameters are required so this will break. I have updated the table to say `❌` in this case. Sorry for missing that.

tutorials/migrating/upgrading_to_godot_4.7.rst
```diff
@@ -0,0 +1,342 @@
...
+
+.. note::
+
+    Setting the element of packed arrays no longer calls the setter for the entire packed array property
+    (`GH-113228`_).
+
+.. note::
+
+    Methods that inherit from a method with a typed return now inherit the return type as well,
+    requiring an explicit return statement in the override (`GH-115763`_`).
```
AThousandShips:
```suggestion
    requiring an explicit return statement in the override (`GH-115763`_).
```

tutorials/migrating/upgrading_to_godot_4.7.rst
AdriaandeJongh:
```suggestion
    The device IDs for mouse and keyboard were changed from ``0`` to ``InputEvent.DEVICE_ID_MOUSE`` and ``InputEvent.DEVICE_ID_KEYBOARD`` because some joypads may use ``0`` as their ID (`GH-116274`_).
```

tutorials/migrating/upgrading_to_godot_4.7.rst
```diff
@@ -0,0 +1,343 @@
...
+.. note::
+
+    Setting the element of packed arrays no longer calls the setter for the entire packed array property
+    (`GH-113228`_).
+
+.. note::
+
+    Methods that inherit from a method with a typed return now inherit the return type as well,
+    requiring an explicit return statement in the override (`GH-115763`_).
+    Add ``return null`` to the end of the method to fix the error.
```
Mickeon, line 258: I don't have time as of writing this, but I genuinely urge someone to come up with a better description based on the [RocketChat discussion](https://chat.godotengine.org/channel/gdscript?msg=eYxqZKskQdpHwQwC4) before this is accepted. It's bound to cause some serious confusion.

To make it somewhat brief, and I know it's not-so-difficult to think about for the more "technically-inclined", but from this description:
- It's difficult to infer that both exposed and user-written methods are affected by this change.
- It is especially relevant for **virtual methods** provided by the engine
- It's not clear what happens with overridden methods that return `void` (that is, nothing needs to be done)
	- Speaking of which, this mentions methods that "inherit" which not the engine's terminology. "Override" would likely be more appropriate and less confusing.
- How to fix a certain error is mentioned, but it's not directly explained what the error is.

akien-mga:
Lacking approvals, but merging as we need this live for the imminent 4.7 release. It can be updated further in follow-up PRs.

akien-mga:
Thanks!

PR #11964, opened 2026-05-02
fix @icon example in "Making plugins"

clankill3r:
preload returns a resource which is not possible for @icon. The change shows the exact line that has to be used to make it work.


skyace65 (approved):

skyace65:

PR #11981, opened 2026-05-11
Adding a sample mesh so demo code works 'out of the box'.

brotheramai:
If the gdscript example is put into a scene as is and executed it throws the error "_multimesh_re_create_aabb: Condition "multimesh->mesh.is_null()" is true." 

This patch adds a boxmesh as a placeholder. Tested and working in 4.6.2stable.

Not a hard bug to track down as-is, but it caused me to scratch my head for a minute when I started messing with multimesh so I thought it might be useful.

### DIFF
--- a/tutorials/performance/using_multimesh.rst
+++ b/tutorials/performance/using_multimesh.rst
@@ -62,6 +62,8 @@ efficient for millions of objects, but for a few thousands, GDScript should be f
         multimesh = MultiMesh.new()
         # Set the format first.
         multimesh.transform_format = MultiMesh.TRANSFORM_3D
+        # Set the mesh that will be duplicated
+        multimesh.mesh = BoxMesh.new()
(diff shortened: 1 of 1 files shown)


tutorials/performance/using_multimesh.rst
```diff
@@ -62,6 +62,8 @@ efficient for millions of objects, but for a few thousands, GDScript should be f
         multimesh = MultiMesh.new()
         # Set the format first.
         multimesh.transform_format = MultiMesh.TRANSFORM_3D
+        # Set the mesh that will be duplicated
```
AThousandShips:
```suggestion
        # Set the mesh that will be duplicated.
```

Calinou (approved):
Looks good to me.

mhilbrunner (approved):

mhilbrunner:
Merged, thank you! And thanks everyone for the reviews.

PR #11987, opened 2026-05-14
Update signals.rst to change main scene from Sprite2D to Node_2D

myrual:
I followed the step by step from chapter 1 to chapter signal. But the button is not visible before title "Connecting a signal in the editor". Finally I found the main scene should be updated to show node_2d

### DIFF
--- a/getting_started/step_by_step/signals.rst
+++ b/getting_started/step_by_step/signals.rst
@@ -105,6 +105,7 @@ Save your newly created scene as ``node_2d.tscn``, if you haven't already.
 You can then run it with :kbd:`F6` (:kbd:`Cmd + R` on macOS).
 At the moment, the button will be visible, but nothing will happen if you
 press it.
+If you can not find the button, you may need to open Project → Project Settings → General → Application → Run → Main Scene, replace the Sprite2D.tscn with node_2d.tscn.
 
 Connecting a signal in the editor
 ---------------------------------


getting_started/step_by_step/signals.rst
```diff
@@ -105,6 +105,7 @@ Save your newly created scene as ``node_2d.tscn``, if you haven't already.
 You can then run it with :kbd:`F6` (:kbd:`Cmd + R` on macOS).
 At the moment, the button will be visible, but nothing will happen if you
 press it.
+If you can not find the button, you may need to open Project → Project Settings → General → Application → Run → Main Scene, replace the Sprite2D.tscn with node_2d.tscn.
```
AThousandShips, line 108: Beyond that this is spelled wrong and not formatted correctly, I don't think this is an appropriate suggestion, it is not good advice in cases that's not the reason, and it assumes the user hasn't followed the instructions, which I don't think is good in a basic tutorial

myrual: I follow the tutorial step by step again to reproduce the issue.

AThousandShips: That's because you pressed the wrong button, press `F6` or (`Cmd + R` on macOS) as instructed

myrual: Thank you very much.

PR #11993, opened 2026-05-16
Fix Lychee URL checking on CI

Calinou:
This also adds the cache file to `.gitignore` in case you run Lychee locally with the same options as CI.

There is still one error I'm trying to fix:

```
Issues found in 1 input. Find details below.

[engine_details/architecture/object_class.rst]:
 [ERROR] error: (at 397:3) | Cannot parse 'res://someresource.res::1' into a URL: invalid port number
```

Adding `res://someresource.res::1` to `.lycheeignore` or passing `--exclude res://someresource.res::1` didn't help.

### DIFF
--- a/.github/workflows/check_urls.yml
+++ b/.github/workflows/check_urls.yml
@@ -24,13 +24,14 @@ jobs:
         uses: lycheeverse/lychee-action@v2
         with:
           args: >
-            --base .
+            --root-dir .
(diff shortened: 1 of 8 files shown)


mhilbrunner (approved):

mhilbrunner:
LGTM, thank you for taking the time to fix these!

PR #11997, opened 2026-05-17
Changed "focussed" to "focused" to match OpenXR documentation.  

dgraham-io:
The "session_focussed" signal will have to be changed in a future Godot release.


m4gr3d:
This is the UK/Australian version as @BastiaanOlij the XR lead lives in Australia.

PR #12004, opened 2026-05-19
Clarify that some built-ins are effective when written on any branch in Spatial shader

### LINKED ISSUE #119552: `ALPHA_SCISSOR_THRESHOLD` in `ShaderMaterial` is equal to `1.0` when conditionally skipped.
### Tested versions

4.6.2.stable

### System information

MacOS

### Issue description

The [description for ALPHA_SCISSOR_THRESHOLD](https://docs.godotengine.org/en/stable/tutorials/shaders/shader_reference/spatial_shader.html) says:
```
If written to, values below a certain amount of alpha are discarded.
```
However `ALPHA_SCISSOR_THRESHOLD = 1.0` is applied when a shader conditionally does not write to it, instead of blending alpha.

For example:
`Enable Alpha Scissor` is false, but the texture is incorrectly transparent when `Alpha` is `0.99`
Image

and is opaque when `Alpha` is `1.0`
Image

Notice that this behavior is consistent to when `ALPHA_SCISSOR_THRESHOLD = 1.0` is set in the gdshader
Image
Image

### Steps to reproduce

1. Unzip the MRP and open the project with the Godot editor.
2. Double click on `new_shader_material.tres`
3. In the inspector under `Shader Parameters`, adjust `Alpha` and notice that the texture is correctly transparent from `0.0` through `0.19` and correctly opaque from `0.2` through `1.0`.
4. In the inspector under `Shader Parameters`, uncheck `Enable Alpha Scissor`.  Adjust `Alpha` and notice that the texture is incorrectly transparent from `0.0` through `0.99`, and opaque when `1.0`.
5. Open `new_shader.gdshader` and turn this
```
  if (enable_alpha_scissor) {
    ALPHA_SCISSOR_THRESHOLD = 0.2;
  }
```
into
```
  //if (enable_alpha_scissor) {
    ALPHA_SCISSOR_THRESHOLD = 1.0;
  //}
```
6. Double click on `new_shader_material.tres`
7. Repeat step 4 and notice identical behavior as seen earlier in step 3, except this time it is correct since we set `ALPHA_SCISSOR_THRESHOLD = 1.0`
8. This was seen in with Forward+, Mobile, and Compatibility renderers.

### Minimal reproduction project (MRP)

[alpha-scissor-shader.zip](https://github.com/user-attachments/files/27969866/alpha-scissor-shader.zip)

Calinou:
In other words, the built-ins will have an effect even if the branch never evaluates to `true` (e.g. when placed within an `if (false)` block).

- This closes https://github.com/godotengine/godot/issues/119552.

### DIFF
--- a/tutorials/shaders/shader_reference/spatial_shader.rst
+++ b/tutorials/shaders/shader_reference/spatial_shader.rst
@@ -296,8 +296,8 @@ shader, this value can be used as desired.
 | inout vec3 **BINORMAL**                | Binormal in model space.                               |
 |                                        | In world space if ``world_vertex_coords`` is used.     |
 +----------------------------------------+--------------------------------------------------------+
-| out vec4 **POSITION**                  | If written to, overrides final vertex position in clip |
-|                                        | space.                                                 |
+| out vec4 **POSITION**                  | If written to on any branch, overrides final vertex    |
+|                                        | position in clip space.                                |
 +----------------------------------------+--------------------------------------------------------+
 | inout vec2 **UV**                      | UV main channel.                                       |
 +----------------------------------------+--------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


mhilbrunner (approved):

mhilbrunner:
Thank you! Merged.

QbieShay:
I see this is alreadymerged, but i feel ilke "written to in any branch" isn't really explicit. I would keep a separate list of built-ins that cause pipeline changes, end explain in that list what they do.

PR #12013, opened 2026-05-23
Added Documentation on Camera3D Preview

ic-coder123:
Added documentation for Camera3D preview in the inspector and editor


skyace65:
Hi, while we appreciate the effort, class reference changes can't be made in this repository. You need to change the XML files in the main Godot repository which then get synced with this one. In this case, you would want to edit this file: https://github.com/godotengine/godot/blob/master/doc/classes/Camera3D.xml

However it's important to note that you wouldn't make a brand new section like this, information like that would need to be inside the description section. I did a random check on a few class reference pages and I haven't found anything that breaks typical convention like this.

Putting that aside I'm not sure how beneficial this change is. You're giving users instructions on how to do a very, very specific thing that I personally don't see the utility in. I'm not aware of complaints from users on seeing the preview button, and it takes up very little screen space. We also typically try to avoid specific instructions like this with documentation. As an example, we don't have manual pages going over how you would make an RTS in Godot, but we do go over how to use the navigation system in a non-specific way, which could be used for an RTS game.

PR #11976, opened 2026-05-06
Update build tools requirements.

bruvzg:
- Updates required versions to match changes in https://github.com/godotengine/godot/pull/116550 https://github.com/godotengine/godot/pull/119278 https://github.com/godotengine/godot/pull/119295
- Adds note about Windows SDK version.

PR #11401, opened 2025-10-22
Add CODEOWNERS

AThousandShips:
Several areas missing but as a starting point to figure out the specifics

Todo:
- [x] Update after rework of teams

PR #12016, opened 2026-05-25
Set up `labeler.yml` file for github labeler actions

skyace65:
Github has an action available called labeler (details [here](https://github.com/actions/labeler)), which can be used to automatically label pull requests based on what files are modified.

This file has been setup for automatically assigning area and topic labels (for example area:manual, topic:3d).

This is not a replacement for Triage, labels such as enhancement and bugfix still need to be manually determined, and there's pages where multiple labels could apply so things can't be setup automatically (for example compiling with Xcode)

Once this PR is merged an action will need to be setup to run when a PR is opened.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,209 @@
+# This file is for the github actions labeler: https://github.com/actions/labeler
+# This file is used for stating which files get what label.
+# A github action needs to be run for assignment to happen.
+
+area:about:
+- changed-files:
+  - any-glob-to-any-files: area/**
+
+area:community:
+- changed-files:
+  - any-glob-to-any-files: community/**
+
+area:engine details:
+- changed-files:
+  - any-glob-to-any-files: engine_details/**
+
+area:getting started:
+- changed-files:
+  - any-glob-to-any-files: getting_started/**
+
+area:manual:
+- changed-files:
+  - any-glob-to-any-files: tutorials/**
+
+area:class reference:
+- changed-files:
+  - any-glob-to-any-files: classes/**
+
+content:website:
+- changed-files:
+  - any-glob-to-any-files:
+    - _extensions/**
+    - _static/**
+    - _styleguides/**
+    - _templates/**
+    - _tools/**
+
+platform:android:
+- changed-files:
+  - any-glob-to-any-files:
+    - engine_details/development/compiling/compiling_for_android.rst
+    - engine_details/development/configuring_an_ide/android_studio.rst
+    - tutorials/editor/using_the_android_editor.rst
+    - tutorials/export/android_gradle_build.rst
+    - tutorials/export/exporting_for_android.rst
+    - tutorials/platofrm/android/**
+
+platform:ios:
+- changed-files:
+  - any-glob-to-any-files:
+    - engine_details/development/compiling/compiling_for_ios.rst
+    - engine_details/development/compiling/compiling_for_ios_on_linux.rst
+    - tutorials/export/exporting_for_ios.rst
+    - tutorials/platform/ios/**
+
+platform:macos:
+- changed-files:
+  - any-glob-to-any-files:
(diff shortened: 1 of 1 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,206 @@
...
+    - tutorials/physics/soft_body.rst
+
+topic:animation:
+- changed-files:
+  - any-glob-to-any-files: tutorials/animation/**
+
+# Commented out due to ongoing changes with the change to hte asset store
+#topic:assetstore:
+#- changed-files:
+#  - any-glob-to-any-files: community/asset_library/**
```
Mickeon: Perhaps you should still have `topic:assetlib` for the time being.
```suggestion
topic:assetlib:
- changed-files:
  - any-glob-to-any-files: community/asset_library/**
```
Perhaps you could write a TODO to change it later once new pages are ready for the Asset Store, instead.

.github/labeler.yml
```diff
@@ -0,0 +1,206 @@
...
+
+topic:audio:
+- changed-files:
+  - any-glob-to-any-files: tutorials/audio/**
+
+#Unsure if this would apply to all compiling pages or not
+#topic:buildsystem:
+#- changed-files:
+#  - any-glob-to-any-files:
```
Mickeon: I'm not sure why not? Which pages in `engine_details/development/compiling` are debatably not part of `topic:buildsystem`?

We could use a few past PRs as a point of reference, like:
- https://github.com/godotengine/godot-docs/pull/11976
- https://github.com/godotengine/godot-docs/pull/11169
- Repiteo then labelled https://github.com/godotengine/godot-docs/pull/11535 , but in this case `topic:buildsystem` is used to refer to the documentation's buildsystem itself, rather than Godot Engine's.
- A similar case in https://github.com/godotengine/godot-docs/pull/11401, as well.

Mickeon (commented):
With the way labels are configured in this repository, this is pretty much a no-brainer as it saves a lot of maintenance time. It's worth a deeper look from others.

skyace65:
Fixed

mhilbrunner (approved):

mhilbrunner:
Similar to Mickeon (and as discussed on RC), I think this is a great idea and could save us quite a bit of time and manual work. I gave it a review and it looks fine as is, but its also fine to iteratively improve further if something crops up in practice.

I'm fine with merging as is, but we can also leave this open a few days more to give others a chance to chime in. Especially @AThousandShips and @Calinou :)

.github/labeler.yml
```diff
@@ -0,0 +1,205 @@
...
+- changed-files:
+  - any-glob-to-any-files: engine_details/architecture/unit_testing.rst
+
+topic:xr:
+- changed-files:
+  - any-glob-to-any-files:
+    - engine_details/development/compiling/compiling_for_visionos.rst
+    - tutorials/editor/using_the_xr_editor.rst
+    - tutorials/export/exporting_for_visionos.rst
+    - tutorials/xr/**
```
AThousandShips:
```suggestion
    - tutorials/xr/**

```

AThousandShips (commented):
I'd also suggest adding the `classes` folder under `area:class reference` to more easily catch errors posting such PRs

Also would be good to confirm this works with labels with spaces in them or if we need to handle that specially

Calinou (commented):
Should be good to merge after applying suggestions.

I agree autolabeling makes sense to save some time. However, the downside is that organization members can no longer easily find unlabeled issues that should be labeled by searching for `no:label`. I guess we should test drive it and revert it if this proves to be problematic.

skyace65:
added a newline to the end of the file, and added the class reference label

AThousandShips (approved):
LGTM

mhilbrunner:
Lets give this a try then, merged. Thank you! PR for the GH action welcome.

PR #12017, opened 2026-05-26
Add support for classref admonitions

Cykyrios:
godotengine/godot#111375 adds support for admonitions in the editor, but there were concerns about the size of Sphinx admonitions for the online documentation, especially since some class properties/methods have multiple consecutive notes. That PR now generates a special rST directive, which this PR adds support for, via a custom extension and some CSS.

The extension itself was made using some AI help, then cleaned up by me.

<details>
<summary>Original screenshots</summary>

The following screenshot is from the [Resource](https://docs.godotengine.org/en/latest/classes/class_resource.html) page, with all **Note:**, **Warning:**, **Tip:**, and **Important:** admonitions converted using godotengine/godot#111375:

image
Here are all the available types of admonitions:
image

If someone knows how to get the correct icons from admonitions, I'd like to have those instead of a generic admonition icon.
</details>

Updated screenshots (improved colors, per-type icons):
classref_admonitions_dark
classref_admonitions_light

### DIFF
--- a/_extensions/classref_admonitions.py
+++ b/_extensions/classref_admonitions.py
@@ -1,0 +1,124 @@
+# -*- coding: utf-8 -*-
+"""
+    class_ref_admonitions
+    ~~~~~~~~~~~~~~~~~~~~~
+
+    Sphinx extension to format Godot's classref admonitions.
+
+    :copyright: Copyright 2026 by The Godot Engine Community
+    :license: MIT.
+"""
+
+import re
+from docutils import nodes
+from docutils.parsers.rst import Directive
+from sphinx.writers.html5 import HTML5Translator
+
+
+class classref_admonition(nodes.General, nodes.Element):
+    pass
+
+
+class classref_important(classref_admonition):
+    pass
+
+
+class classref_note(classref_admonition):
+    pass
+
+
+class classref_tip(classref_admonition):
+    pass
+
+
+class classref_warning(classref_admonition):
+    pass
+
+
+class BaseClassrefDirective(Directive):
+    has_content = True
+
+    title = ""
+    css_class = ""
+    node_class = classref_admonition
+
+    def run(self):
+        self.assert_has_content()
+
+        node = self.node_class()
+        node["title"] = self.title
+        node["classes"].append(self.css_class)
+
+        self.state.nested_parse(
+            self.content,
+            self.content_offset,
+            node,
+        )
+
+        return [node]
+
+
+class ClassrefNoteDirective(BaseClassrefDirective):
+    node_class = classref_note
+    title = "Note"
+    css_class = "note classref-admonition"
+
+
+class ClassrefWarningDirective(BaseClassrefDirective):
+    node_class = classref_warning
+    title = "Warning"
+    css_class = "warning classref-admonition"
+
+
+class ClassrefTipDirective(BaseClassrefDirective):
+    node_class = classref_tip
+    title = "Tip"
+    css_class = "tip classref-admonition"
+
+
+class ClassrefImportantDirective(BaseClassrefDirective):
+    node_class = classref_important
+    title = "Important"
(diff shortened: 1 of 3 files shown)


akien-mga:
FYI, it seems you used your no-reply email from GitLab instead of GitHub for your godot-docs clone, so the commit doesn't appear linked to your GitHub account.

Cykyrios:
Fixed my git config and increased color saturation, also using okhcl color space for mixing colors instead of srgb. This makes the admonitions pop a bit more:
classref_admonitions_dark
classref_admonitions_light

Cykyrios:
Fixed saturation being applied twice to the title, and made saturation theme-dependent. @Mickeon mentioned all existing admonitions using the same icon, so this PR will not try to get the "right" icon for each type.

classref_admonitions_dark
classref_admonitions_light

_extensions/classref_admonitions.py
```diff
@@ -0,0 +1,139 @@
...
+    title = "Note"
+    icon = "i"
+    css_class = "note classref-admonition"
+
+
+class ClassrefWarningDirective(BaseClassrefDirective):
+    node_class = classref_warning
+    title = "Warning"
+    icon = "⚠"
+    css_class = "warning classref-admonition"
```
Mickeon, line 70: > so this PR will not try to get the "right" icon for each type.

Was this a typo? It seems like you are doing just that now, which is neat.

Mickeon (approved):
I do approve this in principle as I have https://github.com/godotengine/godot/pull/111375 already. I can't speak much about the code required for it, but as [we discussed the styling](https://chat.godotengine.org/channel/documentation?msg=zgbCiNAT2dRxJvHzs) it's currently pretty good.

In due time we should also update the contributors documentation to mention these admonitions.

_extensions/classref_admonitions.py
Cykyrios: Not a typo, but rather me forgetting to remove this from the first iteration... Character-based icons aren't that reliable as they depend on the font, and the "important" icon was rendered as a colored emoji while the others weren't; they are now just leftovers that I should remove, I think it's better to fetch the icons from the CSS, even if this means that all of them get the same icon for now.

Mickeon: That is completely okay, and I agree with the sentiment.

Even then, the idea of changing the icons would have to be applied to the normal admonitions, as well. When I pointed it out yesterday, it was purely because I found odd that they had been that way from the beginning.

Cykyrios:
I cleaned up the unused icon and paragraph functions in the python file, and added icons using the SVG files from the editor (the same ones as godotengine/godot#111375). They can be added to standard admonitions in a separate PR if needed.

classref_admonitions_dark
classref_admonitions_light

Cykyrios:
Just a quick note on the https://github.com/godotengine/godot-docs/labels/waiting%20on%20pr%20merge label, this is actually the other way around: this PR should be merged first, so that godotengine/godot#111375 won't generate unsupported rST (for methods returning PackedArrays only if I'm not mistaken).

Calinou (approved):
Tested locally on Firefox 150 and Chromium 148, it works as expected.

[`color-mix` is well-supported in CSS](https://caniuse.com/wf-color-mix) overall (it's considered Baseline now).

mhilbrunner (approved):

mhilbrunner:
Left a note on https://github.com/godotengine/godot/pull/111375 that this PR is good to go if a decision is made to merge the feature.

Meorge:
I don't see any pending reviews on this PR, and it doesn't sound like it is dependent on anything else in particular (#111375 sounds like it is waiting on _this_ PR); would it be appropriate to add to the 4.8 milestone given it has several approvals and passes CI?

mhilbrunner:
Thank you! Merged.

PR #12019, opened 2026-05-28
Update XR Images and Asset Library reference for 4.7

skyace65:
Updates the images on the following pages

- Setting up XR
- Deploying to Android
- Introducing XR Tools

All images have gone through Squoosh compression. There's one image on the deploying to android page at the bottom I can't update because I don't have a headset, which is needed for that option to appear.

I've also updated a reference to the Asset Library to the Asset Store


m4gr3d (approved):

mhilbrunner (approved):

mhilbrunner:
Thank you!

dagarsar:
`xr_left_hand.webp` ended up with the wrong path and the old one still appears in the docs page

Calinou:
> `xr_left_hand.webp` ended up with the wrong path and the old one still appears in the docs page

Fixed in https://github.com/godotengine/godot-docs/pull/12060, thanks for the report :slightly_smiling_face:

PR #12020, opened 2026-05-28
An Update to the player input scripting instructions

jarr6:
Clarified instructions for modifying velocity calculation when pressing the 'up' key. It is my first pull, but I saw a misuse of commenting. hope that helps :-)

Instead of uncommenting the code again, why not leave it as a wrong example. commenting then uncommenting it's like holding a cup without its handle.

unchanged code:
var velocity = Vector2.Zero;
if (Input.IsActionPressed("ui_up"))
{
velocity = Vector2.Up.Rotated(Rotation) * _speed;
}
After this code sniplet this should be added = "Position += velocity * (float)delta;"

The full code should look like:
var velocity = Vector2.Zero;
if (Input.IsActionPressed("ui_up"))
{
velocity = Vector2.Up.Rotated(Rotation) * _speed;
}

Position += velocity * (float)delta;
And for GDScript it should look like this:
var velocity = Vector2.ZERO
if Input.is_action_pressed("ui_up"):
velocity = Vector2.UP.rotated(rotation) * speed

position += velocity * delta

### DIFF
--- a/getting_started/step_by_step/scripting_player_input.rst
+++ b/getting_started/step_by_step/scripting_player_input.rst
@@ -105,7 +105,7 @@ Moving when pressing "up"
 -------------------------
 
 To only move when pressing a key, we need to modify the code that calculates the
-velocity. Uncomment the code and replace the line starting with ``var velocity`` with the code below.
+velocity. To modify, add the code starting with ``var velocity`` seen in the code below.
 
 .. tabs::
  .. code-tab:: gdscript GDScript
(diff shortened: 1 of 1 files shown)


AThousandShips:
This change doesn't really make much sense, it's also wrong, and it now has inconsistencies as there's now commented out code that's gone with the current description

I don't think this change makes the description any better

There's no "misuse" of commenting here, and the new changes makes the documentation more confusing and impossible to follow

skyace65:
I'm in agreement with AThousandShips. This change doesn't fix anything, and as is doesn't make sense.

PR #12029, opened 2026-06-01
Added information about an easy pitfall with rand_weighted()

PodunkGames:
This part of the `rand_weighted()` method under the hood causes negative number weights to produce likely-unexpected results.

```cpp
for (int64_t i = 0; i < weights_size; ++i) {
	weights_sum += weights[i];
}
```

Added a brief mention of this fact under the `rand_weighted()` method.

To reproduce what I'm talking about, this code will only ever print the number 1 because the `-6.0` weight is added to the weights_sum canceling out all other values. 

```gdscript
func _ready() -> void:
	var my_array: Array[int] = [1,2,3,4]
	var my_weights: Array[float] = [1.0, 2.0, 3.0, -6.0]
	var rng: RandomNumberGenerator = RandomNumberGenerator.new()
	for i: int in 1000:
		print(my_array[rng.rand_weighted(my_weights)])

```

I'm doubtful this is a bug, but it's definitely an easy pitfall.

### DIFF
--- a/classes/class_randomnumbergenerator.rst
+++ b/classes/class_randomnumbergenerator.rst
@@ -155,7 +155,7 @@ Method Descriptions
 
 :ref:`int<class_int>` **rand_weighted**\ (\ weights\: :ref:`PackedFloat32Array<class_PackedFloat32Array>`\ ) :ref:`🔗<class_RandomNumberGenerator_method_rand_weighted>`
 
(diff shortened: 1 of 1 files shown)


AThousandShips:

Calinou:
- Superseded by https://github.com/godotengine/godot/pull/119948 (itself superseded by https://github.com/godotengine/godot/pull/120004).

PR #12032, opened 2026-06-02
Clarify performance impact of area lights in 3D lights and shadows

Calinou:
- Follow-up to https://github.com/godotengine/godot/pull/119983.

There is a significant cost as soon as at least one AreaLight3D node is visible in the view frustum, due to certain shader features only being enabled in this case.

- See https://github.com/godotengine/godot/pull/119970#pullrequestreview-4411989608.

### DIFF
--- a/tutorials/3d/lights_and_shadows.rst
+++ b/tutorials/3d/lights_and_shadows.rst
@@ -378,14 +378,27 @@ Area light
 
 Sometimes, you want lighting to come from a large area instead of a single
 point. Area lights are useful for simulating soft, diffuse lighting, such as
-light coming from a window or a lit billboard. This type of light is expensive
-to render in real-time, so it should be used sparingly, especially when shadows
-are enabled.
+light coming from a window or a lit billboard.
 
 Godot provides the :ref:`class_AreaLight3D` node for this purpose, which emits
 light from a rectangular area. The node only emits light and has no other visual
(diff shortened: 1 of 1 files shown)


tutorials/3d/lights_and_shadows.rst
```diff
@@ -378,15 +378,21 @@ Area light
...
 
 Godot provides the :ref:`class_AreaLight3D` node for this purpose, which emits
 light from a rectangular area. The node only emits light and has no other visual
 representation in the scene. The screenshots below use a :ref:`class_Sprite3D`
 node as a child of the area light for visualization purposes.
 
+.. warning::
+
+    This type of light is expensive to render in real-time, so it should be used
+    sparingly, especially when shadows are enabled. As soon as one area light is
```
clayjohn: I would change this a bit. The being in the view frustum thing is only applicable to the Forward+ renderer. For the others, only meshes that intersect with the light pay the cost. So it is a lot cheaper. 

So I would say:
1. It is very expensive, use sparingly
2. It's especially bad with the Forward+ renderer since all objects pay the cost of one is in the view frustum.

clayjohn (approved):
Looks great!

mhilbrunner (approved):

mhilbrunner:
Thanks a bunch! Merged.

PR #12041, opened 2026-06-07
Remove default editor shortcuts page

skyace65:
Does what the title says, definitely want discussion before a merge.

I'm making this PR because I do not see the functional benefit of keeping it around. Everything that is listed on this page is viewable in the editor. The documentation has tons of pages that list options available in the editor, the difference is those ones provide extra information you can't get in the editor, and sometimes screenshots to make things clearer. The only thing I can see it providing that's extra is shortcuts on other platforms like MacOS, which is helpful for documentation writing, but that's not useful to end users.

It's also been perpetually out of date as long as I can remember. I don't even consult it for MacOS shortcuts for that reason, I just go digging through the engine's C++ to find what something is supposed to be.

Maybe there's something I'm missing and I'm open to arguments, but with the way I see it right now, I think any further time spent trying to update or maintain it is a waste that would be better spent on anything else documentation related.


tutorials/editor/script_editor.rst
```diff
@@ -93,7 +93,6 @@ column, and whether the indentation is done using tabs, or spaces.
 
 Many of the actions performed in the text editor can also be executed using
 shortcuts. The actions show their corresponding shortcuts next to them.
-For a complete shortcut list, see the :ref:`text editor shortcuts <doc_default_key_mapping_shortcuts_text_editor>`.
```
bruvzg, line 96: Might be worth adding a note where to find and rebind shortcuts in the editor (`Editor Settings` → `Shortcuts` → `Script Text Editor`).

bruvzg (approved):
I agree, it feels redundant, and unless it is auto generated it will always be out of date.

Mickeon (approved):
Let's not be hasty and merge this too soon until more people chime in, but I do agree with the above sentiments.
Ideally this page (& others too!) should be generated automatically, but there's no system in place for this.

##### ~~Just when I was done localizing it, huh?~~

skyace65:
I've added text to the script editor page explaining where shortcuts are listed as requested

tutorials/editor/script_editor.rst
```diff
@@ -93,7 +93,8 @@ column, and whether the indentation is done using tabs, or spaces.
...
+Within Godot itself you can find, and rebind, all shortcuts by going to
```
Calinou:
```suggestion
Within Godot itself, you can find and rebind all shortcuts by going to
```

mhilbrunner (approved):
I've considered this a few times in the past. I agree it adds a lot of churn.

It does give a place for third party authors, content creators and others to link people to, but as noted this is often outdated, so a double edged sword.

Overall, I lean slightly towards agreeing with the removal.

skyace65:
Applied Calinou's fix

Calinou (approved):
Looks good to me.

In the future, I might look into writing a script to generate the page automatically.

akien-mga:
Thanks!

PR #12042, opened 2026-06-08
Add specific icons to admonitions

Cykyrios:
Since #12017 adds icons to the proposed "class reference admonitions", this PR reuses the same icons and applies them to Note, Warning, Tip, and Important admonitions. All other admonitions (Danger, See also, Here be dragons, and anything else I missed) are left untouched.

This also changes the Important color from green (same as Tip) to yellow (same as Warning/Attention), for consistency in the icon color.

| Master | This PR |
| ------ | ------- |
| image | image |

<details>
<summary>More images</summary>
In the following image, only the Important admonition is affected, the other two are identical to the master branch:
image

More examples:
image
image
</details>

Depends on #12017.

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -1208,16 +1208,51 @@ code,
     margin-right: 9px;
 }
 
+.rst-content .admonition.note .admonition-title::before,
+.rst-content .admonition.warning .admonition-title::before,
+.rst-content .admonition.tip .admonition-title::before,
+.rst-content .admonition.important .admonition-title::before {
+    content: "";
+    top: 0.1rem;
+    display: inline;
+    height: 1rem;
+    width: 1rem;
+    padding-right: 1rem;
+    mask-repeat: no-repeat;
+}
+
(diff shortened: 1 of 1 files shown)


Calinou (changes requested):
`:before` should be `::before` instead ([both work, but the former is deprecated](https://stackoverflow.com/questions/17684797/should-i-use-single-colons-or-double-colons-for-before-after-first-le)).

Calinou (approved):
Code and design look good to me.

skyace65 (changes requested):
Just built this branch on Windows 10 and the custom icons do not work. I'm seeing warnings and notes in the manual that are just square boxes in Firefox and Microsoft Edge.

Image

Image

Cykyrios:
Rebased so that this is now properly using the icons from #12017.

skyace65 (approved):
Built the branch and everything looks good.

skyace65:
Thanks!

PR #12048, opened 2026-06-09
Simplify taskbar progress reporting example in Creating applications

Calinou:
Range (which ProgressBar inherits from) provides a `ratio` property, which means `remap()` isn't needed here.


bruvzg (approved):

mhilbrunner (approved):

mhilbrunner:
LGTM, thanks.

PR #12055, opened 2026-06-12
Add labeler github workflow

skyace65:
Follow up to #12016 

I think this is all correct, I just copied what was given as an example that sounded like what we need from the github page [here](https://github.com/actions/labeler). This is my first time trying to make a workflow so please don't assume I read everything correctly.


mhilbrunner (approved):

mhilbrunner:
Alright, lets test this. LGTM :)

PR #12059, opened 2026-06-14
Add warning for back-buffer reading performance

armedturret:
Added a warning about performance impact when reading from the back-buffer in shaders.

This edit's primary purpose is to warn novice users of accidentally causing shaders to have a higher performance impact then intended. This is NOT a bug in Godot, it's just how reading from back buffers work. See [this issue](https://github.com/godotengine/godot/issues/108935#issuecomment-3115851485) for reference.

### DIFF
--- a/tutorials/shaders/screen-reading_shaders.rst
+++ b/tutorials/shaders/screen-reading_shaders.rst
@@ -17,6 +17,12 @@ The workaround is to make a copy of the screen, or a part of the screen,
 to a back-buffer and then read from it while drawing. Godot provides a
 few tools that make this process easy.
 
+.. warning::
+
+    Reading from the back-buffer disrupts the render pipeline and
+    can be performance intensive. For shaders that don't require access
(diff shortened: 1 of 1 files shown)


mhilbrunner:
Needs a CI fix / rebase :)

armedturret:
Not sure what's going on with the CI there. I followed the contribution guide and it didn't mention making any changes file.

Calinou:
> Not sure what's going on with the CI there. I followed the contribution guide and it didn't mention making any changes file.

You need to rebase your PR on top of current `master` to fix CI, as we removed a `labeler` action that kept causing errors. It's still present on your fork's branch, hence the error.

mhilbrunner:
Hey! According to our rendering peeps, Viewports are even more expensive, so the back buffer can't really be avoided if you need it, rendering this notice moot.

armedturret:
Huh, I was certain drawing to a framebuffer through Viewports would be cheaper than copying from the back buffer. Maybe its just all the overhead a Viewport has? Either way, if that's the case I'm happy to close this.

PR #12063, opened 2026-06-15
Mention shader type in Introduction to shaders

Calinou:
The first shader example is a valid `canvas_item` shader, but not a valid `spatial` shader as `COLOR` is read-only in `spatial`'s `fragment()` (it reads the vertex color).


mhilbrunner:
Needs a CI fix / rebase :)

mhilbrunner (approved):

mhilbrunner:
Thanks! Merged.

PR #12066, opened 2026-06-16
Fix incorrect advice on VRR framerate capping in Fixing jitter and stutter

Calinou:
The common "cap 3 frames below the refresh rate" originally added in https://github.com/godotengine/godot-docs/pull/6186 is wrong, as it will frequently trigger V-Sync at higher refresh rates (negating the input lag reduction of being within VRR range).

Instead, the formula `refresh - (refresh * refresh / 3600.0)` should be followed instead, as done by NVIDIA Reflex for its own framerate cap when V-Sync is enabled and a VRR display is used.

### DIFF
--- a/tutorials/rendering/jitter_stutter.rst
+++ b/tutorials/rendering/jitter_stutter.rst
@@ -253,15 +253,42 @@ If your monitor supports it, consider enabling variable refresh rate
 (G-Sync/FreeSync) while leaving V-Sync enabled, then cap the framerate in the
 project settings to a slightly lower value than your monitor's maximum refresh
 rate as per `this page <https://blurbusters.com/howto-low-lag-vsync-on/>`__.
-For example, on a 144 Hz monitor, you can set the project's framerate cap to
-``141``. This may be counterintuitive at first, but capping the FPS below the
+
+This may be counterintuitive at first, but capping the FPS below the
 maximum refresh rate range ensures that the OS never has to wait for vertical
 blanking to finish. This leads to *similar* input lag as V-Sync disabled with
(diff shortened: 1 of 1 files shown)


tutorials/rendering/jitter_stutter.rst
```diff
@@ -253,15 +253,42 @@ If your monitor supports it, consider enabling variable refresh rate
...
+| 144 Hz       | 138 FPS       |
++--------------+---------------+
+| 165 Hz       | 157 FPS       |
++--------------+---------------+
+| 240 Hz       | 224 FPS       |
++--------------+---------------+
+| 360 Hz       | 324 FPS       |
++--------------+---------------+
+| 480 Hz       | 416 FPS       |
++--------------+---------------+
```
Mickeon, line 286: The table _sounds_ useful, but in practice it discourages using the formula described above, in favour of hardcoding the framerate cap to one of these values. I'm not sure it's necessary, unless I'm missing the goal here

Calinou: The table is here to give common examples for people not willing to whip out a calculator :slightly_smiling_face: 

To clarify, it's (unfortunately) rare for games to propose setting VRR-friendly framerate caps out of the box. Users are generally expected to do this on their own, with an exposed FPS cap slider or an external application.

That said, https://github.com/godotengine/godot/pull/66367 adds automatic FPS limiting options (with one of them targeting VRR).

mhilbrunner:
Needs a CI fix / rebase :)

mhilbrunner (approved):

mhilbrunner:
Thank you! 🙏

PR #12071, opened 2026-06-17
Improve Wayland/X11 page, document GNOME < 50 HDR bug in Wayland

### LINKED ISSUE #12043: Document minimum OS/Linux DE requirements for HDR output
**Your Godot version:**
4.7.rc1

**Issue description:**
- See https://github.com/godotengine/godot/issues/118480 for context.

HDR output is relatively "new" feature, at least in the Linux world, with limited or no support on some DEs (e.g. GNOME < 50). It would be good to document known first suitable versions if it's something similar for KDE and Windows or macOS.

If GNOME is the only odd one out, then maybe it's a bit overkill to have a note just for one of many DEs on a niche(r) OS (and I'm the Linux guy of 20 years saying that ;)).

https://docs.godotengine.org/en/latest/tutorials/rendering/hdr_output.html#hdr-output

Calinou:
- Mention Windows 11 being preferred for HDR output over Windows 10.

___

- This closes https://github.com/godotengine/godot-docs/issues/12043.

### DIFF
--- a/tutorials/rendering/hdr_output.rst
+++ b/tutorials/rendering/hdr_output.rst
@@ -10,6 +10,15 @@ used by Godot for both Standard Dynamic Range (SDR) output and HDR output modes.
 HDR output is supported on iOS, Linux (Wayland), macOS, visionOS, and Windows. It is not supported
 on Android, Linux (X11), or web.
 
+.. note::
+
+    Both Windows 10 and Windows 11 support HDR output, but Windows 11 is
+    preferred as it features better HDR calibration tools.
+
+    On Linux, GNOME versions prior to 50 have a bug that prevents HDR output from working
+    on Wayland. If you are using an older version of GNOME, you will need to
+    upgrade to version 50 or later to use HDR output on Wayland.
+
 Enabling HDR output in your project
 -----------------------------------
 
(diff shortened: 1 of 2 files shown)


allenwp:
Thanks! Maybe the comment about Windows could be more specific as to why Windows 11 is preferred:

> Both Windows 10 and 11 support HDR output, but only Windows 11 provides an app that can be used to configure the screen’s maximum luminance that is used by the Godot editor.

This note could be removed when we add an editor setting to control the maximum luminance used by the editor, which is required for macOS and is a nice-to-have for Windows.

I’m also happy to not mention this at all to reduce the overall amount of information on this already dense page, especially given that this limitation will be addressed in the future.

mhilbrunner (approved):

tutorials/rendering/hdr_output.rst
```diff
@@ -10,6 +10,15 @@ used by Godot for both Standard Dynamic Range (SDR) output and HDR output modes.
 HDR output is supported on iOS, Linux (Wayland), macOS, visionOS, and Windows. It is not supported
 on Android, Linux (X11), or web.
 
+.. note::
+
+    Both Windows 10 and Windows 11 support HDR output, but Windows 11 is
+    preferred as it features better HDR calibration tools.
```
akien-mga:
```suggestion
    Both Windows 10 and 11 support HDR output, but only Windows 11
    provides an app that can be used to configure the screen’s maximum
    luminance that is used by the Godot editor.
```

akien-mga:
Thanks!

PR #12074, opened 2026-06-18
Add suppression of RTD popup about outdated stable version

adamscott:
This PR removes the nagging "outdated version" notification for the version specified at the top of the `custom.js` file.

Capture d’écran, le 2026-06-18 à 13 06 31

### DIFF
--- a/_static/js/custom.js
+++ b/_static/js/custom.js
@@ -1,3 +1,5 @@
+const CURRENT_STABLE_VERSION = "4.7";
+
 // Handle page scroll and adjust sidebar accordingly.
 
 // Each page has two scrolls: the main scroll, which is moving the content of the page;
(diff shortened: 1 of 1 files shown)


Calinou (commented):
Tested locally on Chromium 149 using [local overrides in DevTools](https://stackoverflow.com/questions/16494237/chrome-dev-tools-modify-javascript-and-reload#48843321), it sometimes works, sometimes not. It seems to be a 50% chance of it successfully hiding, and it'll show as usual otherwise (with nothing printed to the console).

This might be related to the DevTools override system though.

adamscott:
> Tested locally on Chromium 149 using [local overrides in DevTools](https://stackoverflow.com/questions/16494237/chrome-dev-tools-modify-javascript-and-reload#48843321), it sometimes works, sometimes not. It seems to be a 50% chance of it successfully hiding, and it'll show as usual otherwise (with nothing printed to the console).
> 
> This might be related to the DevTools override system though.

Weird, I developped using Chrome DevTools override. :\

skyace65 (changes requested):
I tested it out with the dev tools in Edge (I imagine they're the same ones as Chrome) and it only kinda works.

The popup does get suppressed for me on the 4.7 pages. The problem is I tested the same code on the 4.6 version of the page and the popup got suppressed there too.

PR #12082, opened 2026-06-20
Replace references to the asset library

skyace65:
Does what the title says. Keep in mind this isn't everything in the docs. Some pages are pointing to demo projects that aren't on the asset store yet. And some pages will need more in depth changes (image updates and or text changes)


mhilbrunner (approved):

mhilbrunner:
Thanks a lot!

PR #12084, opened 2026-06-20
Add the returning vector of `get_gravity()` directly to velocity

leandro-benedet-garcia:
`get_gravity` returns a `Vector2`, so it should be summed directly to gravity instead of Y axis.

That was actually an oversight of mine.


rburing (approved):

skyace65 (approved):

skyace65:
Thanks!

PR #12092, opened 2026-06-23
Update supported VR Headsets for 4.7

skyace65:
Updates the list of supported headsets.

- I've added the Steam Frame, Android XR Headsets, and Vive Focus Vision (I'm assuming we had support for that already but never listed it).
- Also reformatted the android based headsets we support to be bullet point based now.
- Also removed the mention of a plugin being required, that changed with 4.6 (Though it is still recommended).

### DIFF
--- a/about/list_of_features.rst
+++ b/about/list_of_features.rst
@@ -749,15 +749,25 @@ Mobile
 XR support (AR and VR)
 ----------------------
 
-- Out of the box :ref:`support for OpenXR <doc_setting_up_xr>`.
+- Support for desktop headsets using :ref:`OpenXR <doc_setting_up_xr>`. If a headset
+  works with SteamVR, it should work with Godot.
 
-   - Including support for popular desktop headsets like the Valve Index, WMR headsets, and Quest over Link.
+   - Godot also supports Quest over Link, AndroidXR Direct Preview, and Pico
+     Connect.
 
-- Support for :ref:`Android-based headsets <doc_deploying_to_android>` using OpenXR through a plugin.
+- Support for :ref:`Android-based headsets <doc_deploying_to_android>` using OpenXR.
+  Including support for the following standalone headsets:
+  
+   - Meta Quest 1/2/3 and Pro
+   - Pico 4/4 Ultra
+   - Magic Leap 2
+   - Lynx R1
+   - HTC Vive Focus Vision
+   - Android XR headsets
 
-  - Including support for popular standalone headsets like the Meta Quest 1/2/3 and Pro, Pico 4, Magic Leap 2, and Lynx R1.
+- Support for the Linux-based standalone Steam Frame using OpenXR.
 
-- Out of the box limited support for visionOS Apple headsets.
+- Limited support for visionOS Apple headsets.
 
   - Currently only exporting an application for use on a flat plane within the
     headset is supported. Immersive experiences are not supported.


about/list_of_features.rst
```diff
@@ -753,9 +753,17 @@ XR support (AR and VR)
...
    - Including support for popular desktop headsets like the Valve Index, WMR headsets, and Quest over Link.
 
-- Support for :ref:`Android-based headsets <doc_deploying_to_android>` using OpenXR through a plugin.
-
-  - Including support for popular standalone headsets like the Meta Quest 1/2/3 and Pro, Pico 4, Magic Leap 2, and Lynx R1.
+- Support for :ref:`Android-based headsets <doc_deploying_to_android>` using OpenXR.
+  Including support for the following standalone headsets:
+  
+   - Meta Quest 1/2/3 and Pro
+   - Pico 4
```
BastiaanOlij: Should change this to:
```suggestion
   - Pico 4/4 ultra
```

about/list_of_features.rst
```diff
@@ -753,9 +753,17 @@ XR support (AR and VR)
...
+   - Magic Leap 2
+   - Lynx R1.
+   - Vive Focus Vision
```
BastiaanOlij: NItpicking but:
```suggestion
   - HTC Vive Focus Vision
```

about/list_of_features.rst
```diff
@@ -753,9 +753,17 @@ XR support (AR and VR)
 
    - Including support for popular desktop headsets like the Valve Index, WMR headsets, and Quest over Link.
```
BastiaanOlij: I wonder if we should expand on this, we basically support anything that runs on SteamVR. It would be too extensive to list them all, and we don't have access to some heavy hitters that would be good to mention, but...

I would change:
```suggestion
   - Including support for popular desktop headsets like the Valve Index, WMR headsets, Quest over Meta Link, and any headset supporting Steam Link.
```

Though there is also a part of me that wonders if we shouldn't simplify because technically we should also name AndroidXR Direct Preview and Pico's connect as streaming PCVR solutions if we're including Meta Link and Steam Link by name.

about/list_of_features.rst
Calinou: Note that "Ultra" has an uppercase "U", so it should be there too.

about/list_of_features.rst
skyace65: For Meta Link, Direct Preview and Pico's connect, are those working with SteamVR or something else? Like could we include those in an umbrella of "If it can stream SteamVR games" or something

BastiaanOlij: No, they are separate XR Runtimes, they just perform the same function as Steam Link.

mhilbrunner (approved):

skyace65:
@BastiaanOlij I did a more substantial change to the wording of the whole section, let me know if you think this is accurate and phrased in a way that makes sense. I've also removed mention of "out of the box" since I removed the reference to requiring the vendors plugin.

about/list_of_features.rst
```diff
@@ -749,15 +749,25 @@ Mobile
 XR support (AR and VR)
 ----------------------
 
-- Out of the box :ref:`support for OpenXR <doc_setting_up_xr>`.
+- Support for desktop headsets using :ref:`OpenXR <doc_setting_up_xr>`. If a headset
+  works with SteamVR it should work with Godot.
```
Calinou:
```suggestion
- Support for desktop headsets using :ref:`OpenXR <doc_setting_up_xr>`. If a headset
  works with SteamVR, it should work with Godot.
```

about/list_of_features.rst
```diff
@@ -749,15 +749,25 @@ Mobile
...
+     Connect.
 
-- Support for :ref:`Android-based headsets <doc_deploying_to_android>` using OpenXR through a plugin.
+- Support for :ref:`Android-based headsets <doc_deploying_to_android>` using OpenXR.
+  Including support for the following standalone headsets:
+  
+   - Meta Quest 1/2/3 and Pro
+   - Pico 4/4 Ultra
+   - Magic Leap 2
+   - Lynx R1.
```
Calinou:
```suggestion
   - Lynx R1
```

BastiaanOlij:
> @BastiaanOlij I did a more substantial change to the wording of the whole section, let me know if you think this is accurate and phrased in a way that makes sense. I've also removed mention of "out of the box" since I removed the reference to requiring the vendors plugin.

Looks good, small changes Calinou already pointed out not withstanding, thanks!

skyace65:
Fixed.

mhilbrunner (approved):

mhilbrunner:
Thank you! Merged.

PR #12112, opened 2026-06-28
Update AnimationTree guide to explain BlendSpace point names

vaner-org:
Updates "Using AnimationTree" to explain how blend point names work and how renaming works, added to 4.7 by [godot/PR#110369](https://github.com/godotengine/godot/pull/110369), and improves some nearby language.

This will hopefully will lead to fewer issues like [godot/#120680](https://github.com/godotengine/godot/issues/120680) and [godot/#119164](https://github.com/godotengine/godot/issues/119164) cropping up.


TokageItLab (changes requested):
I think it's good, but gif image should be [WebP](https://contributing.godotengine.org/en/latest/documentation/guidelines/docs_image_guidelines.html#format-conversion), I think the text is fine as is.

TokageItLab (approved):

vaner-org:
Found some more opportunities for improvement. If it's better to do the bare minimum per PR, I can split these into another.
- Updated the animated media to fully overlap with and replace `animtree_blendspacepoints.gif`.
- Added a wikipedia link for Delaunay triangulation as is done elsewhere in the docs, rather than just saying "using Delaunay."
- Included a mention of how to manipulate the blend target in-editor, with the shortcut introduced in [godot/PR#109792](https://github.com/godotengine/godot/pull/109792).
- Removed the mention of relabeling X and Y, or changing ranges, as it is purely superficial to the editor and the dimension label does not affect the reference to the value, nor does the visual range of the editor clamp blending range. It also being trivially simple to discover, it doesn't need mention in this overview.
- Moved "Blend mode" to its own subsection, after the "Sync Mode" section, so that they are in the order found in the toolbar, and BlendSpace1D can be mentioned sooner. Text unchanged, updated image.

TokageItLab (changes requested):
PNG files also need to be converted to WebP.

tutorials/animation/animation_tree.rst
```diff
@@ -369,6 +365,15 @@ There are four modes available:
    In that case, use :ref:`AnimationNodeAnimation.use_custom_timeline <class_AnimationNodeAnimation_property_use_custom_timeline>`
    to normalize animation lengths before syncing.
 
+Blend mode
+~~~~~~~~~
+
+Finally, you may also change the blend mode. By default, blending happens by interpolating points inside the closest triangle. When dealing with 2D
+animations (frame by frame), you may want to switch to *Discrete* mode. Alternatively, if you want to keep the current play position when switching
+between discrete animations, there is a *Carry* mode. This mode can be changed in the *Blend* menu:
```
TokageItLab: Since the discrete blending option doesn't apply only to 2D framed animations, it would be better to provide a specific example, such as "when you don't want the intermediate states of the blend to appear in the result".

vaner-org: > By default, blending happens in the *Continuous* mode, by interpolating points inside the closest triangle in ``BlendSpace2D``, or in the line between points in ``BlendSpace1D``. However, this may not always be the best choice. For instance, when dealing with frame-by-frame 2D animations, you may want to switch to the *Discrete* mode, in which the intermediate states of the blend do not appear in the result. Alternatively, if you want to retain the current play position when switching between discrete animations, the *Carry* mode allows you to do so. These modes can be set with the *Blend* menu.

Makes sense?

TokageItLab (approved):

mhilbrunner (approved):

mhilbrunner:
Merged. Thanks and congrats on your first merged contribution! Great work :)

PR #12114, opened 2026-06-28
Update instructions for creating Resource objects

hvdyinv4688hbv:
Clarify instructions for creating a new Resource object in the Inspector.

### DIFF
--- a/tutorials/scripting/resources.rst
+++ b/tutorials/scripting/resources.rst
@@ -189,7 +189,7 @@ and :ref:`Resource <class_Resource>` features:
 
 Godot makes it easy to create custom Resources in the Inspector.
 
-1. Create a new Resource object in the Inspector. This can even be a type that derives Resource, so long as your script is extending that type.
+1. Create a new Resource object in the Inspector by clicking the file icon with the green plus sign. This can even be a type that derives Resource, so long as your script is extending that type.
 2. Set the ``script`` property in the Inspector to be your script.
 
 The Inspector will now display your Resource script's custom properties. If one edits


AThousandShips:
I'm not sure this is needed or if we should really make this so verbose and explicit here, this is already covered in the introduction so I doubt anyone will be confused by this if they've read things that anyone should have read when looking at this part of the documentation

hvdyinv4688hbv:
i guess my question is: is the editor introduction a standard prerequisite for other sections of the documentation? if not, i think it is a good addition. You are right about it being verbose. There is probably a better way to word it. maybe: Create a new Resource object in the Inspector by clicking the top left icon.
or
Create a new Resource object in the Inspector by clicking [image of icon goes here].

AThousandShips:
Yes people are expected to have read the introduction, I do not believe we should write the documentation for the case where users haven't read the basics, otherwise we'll bloat the text with obvious things

Have anyone actually reported issues with this?

Mickeon:
> the file icon with the green plus sign

This is the kind of overly-specific details that are prone to become invalidated in the future, turning what would look helpful in the moment into confusing instructions that don't quite match the current editor. Because they're a big maintenance burden, there's an on-going effort to minimize these cases, so this PR doesn't quite help as much as one may think.

skyace65:
I'm in agreement with Mickeon and AThousandShips. This is far more specific than what we want in this section of the documentation. I appreciate the effort you put into this PR, but unfortunately I'm going to close it. I hope you consider contributing in the future if you find something else you think could be improved.

PR #12116, opened 2026-06-30
Adjust license text according to main repository

notroid5:
It should be identical to https://github.com/godotengine/godot/blob/master/LICENSE.txt again now

### DIFF
--- a/about/complying_with_licenses.rst
+++ b/about/complying_with_licenses.rst
@@ -48,7 +48,7 @@ This text reads as follows:
 
     This game uses Godot Engine, available under the following license:
 
-    Copyright (c) 2014-present Godot Engine contributors (see AUTHORS.md).
+    Copyright (c) 2014-present Godot Engine contributors (see <https://github.com/godotengine/godot/blob/master/AUTHORS.md>).
     Copyright (c) 2007-2014 Juan Linietsky, Ariel Manzur.
 
     Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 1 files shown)


notroid5:
After reading again, should the warning at the top not say: "The recommendations **on** this page"?

Edit: Maybe also change "under this same license" at the start to either "under the same license" or "under this license"? Both "this" and "same" refer to the previously mentioned MIT License, so it feels redundant.

about/complying_with_licenses.rst
```diff
@@ -48,7 +48,7 @@ This text reads as follows:
 
     This game uses Godot Engine, available under the following license:
 
-    Copyright (c) 2014-present Godot Engine contributors.
+    Copyright (c) 2014-present Godot Engine contributors (see AUTHORS.md).
```
Calinou:
```suggestion
    Copyright (c) 2014-present Godot Engine contributors (see <https://github.com/godotengine/godot/blob/master/AUTHORS.md>).
```

Calinou (commented):
This makes sense, but since `AUTHORS.md` isn't distributed alongside with the game, how can people know which `AUTHORS.md` file it's referring to?

In https://github.com/godotengine/godot/pull/51028, I addressed this by replacing `(see AUTHORS.md)` with `(see --about authors command line argument)`.

I think it should be replaced with a full URL to the file in this particular instance.

Calinou (approved):
Looks good to me.

akien-mga:
Thanks for the contribution, but I don't think we should make this change.

I would actually consider removing the "(see AUTHORS.md)" note from the upstream license next time we decide to update it.

PR #12122, opened 2026-07-02
C# [Tool] usage guidelines

Luna-MN:
Added guidelines for using the [Tool] attribute in C#, including initialization of arrays, keeping constructors empty, and avoiding destructive logic in property setters. as per the current tool limitations within Godot C# https://github.com/godotengine/godot-proposals/issues/9001

### DIFF
--- a/tutorials/plugins/running_code_in_the_editor.rst
+++ b/tutorials/plugins/running_code_in_the_editor.rst
@@ -841,22 +841,24 @@ Avoid destructive logic in property setters
         [Tool]
         public partial class TileInfo : Resource
         {
-            private Node3D[] tiles;
+            private Node3D[] _tiles;
+
+            private int _numberOfTiles;
             public int NumberOfTiles 
             {
                 get => _numberOfTiles;
-                set {
+                set 
+                {
                     ClearTiles(); // <--- Avoid this
                     _numberOfTiles = value;
-                            
                 }
             }
-        
-            private int _numberOfTiles;
+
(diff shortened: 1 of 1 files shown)


paulloz (commented):
My first two cents.
- I'm wondering if [Current gotchas and known issues](https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_basics.html#current-gotchas-and-known-issues), or [Common pitfalls](https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_basics.html#current-gotchas-and-known-issues) wouldn't be a better place for this? It's always possible to add a link wherever.
- Are all of the points specific to C#? For instance, the page already contains a danger section about scene tree manipulation at the top.
- This can wait for after more important points have been cleared up, but the code blocks do not follow the [style guide](https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_style_guide.html).

Luna-MN:
as its to do with the tool attribute I added it within that page, if it'd be better in that page i can add it there, and add a warning link somewhere here to redirect there, or the other way around?

these are all problems with C#, due to how it builds the script and applies it to the node, as it was C# related and i know the common user will be using GDScript i added it to the bottom, this can be moved if needed

apologies i will rectify the code style

paulloz:
Documentation maintainers might have a stronger opinion than me about where this should live.

PR #12123, opened 2026-07-03
Clarify `@export` setter not called at startup if the variable has a default in GDScript basics

### LINKED ISSUE #12105: Clarify that the @export setter is not called at startup if the variable has a default value
**Your Godot version:**
4.7.stable

**Issue description:**
When writing some setup logic to the `@export` setter, I expect it always to work, but if the variable has a default value, the setter is not called, and the setup will be incomplete

**improvement**
Clarify that the `@export` setter is not called at startup if the variable has a default value, so the user needs to call the setter in the ready function as well

Calinou:
- This closes https://github.com/godotengine/godot-docs/issues/12105.


mhilbrunner (approved):

mhilbrunner:
Thanks!

Issue #11575, opened 2026-01-01
Add a more complete and simple example of a syntaxically SyntaxHighlighter


1641585051:
**Godot version: 4.5.1**

**Issue description: The Dictionary _GetLineSyntaxHighlighting(int line) function of SyntaxHighlighter si completely icomprehensible that causes me to not be able to accurately locate the highlight,and _GetLineSyntaxHighlighting function lacks more practical instructions and examples**

**URL to the documentation page (if already existing): https://docs.godotengine.org/en/stable/classes/class_syntaxhighlighter.html#class-syntaxhighlighter-method-get-line-syntax-highlighting**

Ayush41:
Yes, The current documentation explains the return format but lacks a complete, practical example of implementing a custom SyntaxHighlighter.
I’ll have to  add a minimal working example that shows how to read a line from TextEdit and return highlighting regions. This should make the method much easier to understand.

1641585051:
@Ayush41 The index of the highlighted data seems to have nothing to do with the wrapped text and the wrapper index,which I noticed after several days of fumbling in the forest of TextEdit's API. I hope to explain the relationship between them.

ScatteredComet:
Here is an example project that implements a syntax highlighter:

https://github.com/BenBlueeeee/godot-syntax-highlighter-tutorial

I think its relatively simple, but let me know if I've overcomplicated things. It has both gdscript and c# code. It modifies a text edit into a "virtual" debugging window, where the user can write lines of distinct types to the textedit. Different line types (error, warning, generic) are highlighted different colors. The user can press the 1,2,3 keys on their keyboard to simulate fake logs.

If anyone has any code style suggestions or comments of course let me know.

I'm also happy to help write a tutorial page for this at some point.

PR #12124, opened 2026-07-04
Add a simple example for _get_line_syntax_highlighting

1641585051:
The original API of function __get_line_syntax_highlighting  was really confusing about the return value and didn't clearly tell users how to write function  _get_line_syntax_highlighting. So, we've added an extra new simple example while keeping the original description. Here’s a zip file of the sample project and some screenshots of the results.There’s a discussion about this issue in #11575 .
project is: [higjtlight-text.zip](https://github.com/user-attachments/files/29652757/higjtlight-text.zip)

result


AThousandShips:

PR #12137, opened 2026-07-09
Fix leftover bracket typo

VojtaStruhar:
I think that the bracket `]` should not be there, seems like a small typo :)

Screenshot-Godot-001479


AThousandShips:

PR #12143, opened 2026-07-12
Ungroup shortcut needs shift to be held

Lonami:
I am trying to press <kbd>Ctrl+G</kbd> after a group is made and nothing happens. However <kbd>Ctrl+Shift+G</kbd> correctly ungroups, and that's what the UI shows as well (v4.7.stable.official \[5b4e0cb0f\]):
image
So I think the documentation is wrong / outdated.


skyace65 (approved):

skyace65:
Thanks!

PR #12146, opened 2026-07-14
Remove pulse from custom effect example

seanhagen:
Seems to have disappeared at some point between now and the v3.2 docs, though it's still referenced in the example below the GDScripts. This just adds back in the script ( and changes `tool` to `@tool` ).

### DIFF
--- a/tutorials/ui/bbcode_in_richtextlabel.rst
+++ b/tutorials/ui/bbcode_in_richtextlabel.rst
@@ -1251,6 +1251,32 @@ the user fixes whatever error cropped up in their custom effect logic.
 
 Here are some examples of custom effects:
 
+Pulse
+~~~~~
+
+::
+   @tool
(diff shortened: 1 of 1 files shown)


Calinou:
That specific example was intentionally removed as it's superseded by the built-in `[pulse]` tag added in 4.2:

- https://github.com/godotengine/godot/pull/77117

I don't think we should readd it, especially since we already have other examples listed below. I would only remove the `[pulse]` usage from the example usage block at the end of the page instead.

seanhagen:
Alrighty, I can re-write the documentation so that's a bit more clear then.

Calinou (approved):
Looks good to me.

AThousandShips:
Title needs an update

skyace65 (approved):

skyace65:

PR #12155, opened 2026-07-17
Add match pattern guard comparison example

Leonardo-Ma:
It's slightly confusing already by using an array match with wildcard to present a new concept. Not being obvious that it can be used for simple direct comparison.


HolonProduction:
> using an array match with wildcard to present a new concept

Array patterns were introduced prior to this though.

IMHO introducing concepts with contrived examples is bad practice. What you added here is contrived. `point` is an array in this codeblock, so checking it with `>` makes no sense. Also all other branches do an actual real world check on the point while this one just makes no sense.

Leonardo-Ma:
I meant specifically the pattern guard and when keyword being introduced using an array and wildcard. Had to see examples elsewhere to actually understand what was happening.
Though I do agree the new examples are wrong. Also couldn't use it in my specific case where I wanted a simple match comparison instead of several ifs. It seems that's just not possible.

PR #1400, opened 2018-04-25
init commit custom godot servers, new file

hungrymonkey:
this doc is made to elaborate on how to create a background thread with godot facilities,
this server would all the engine programmer extend functionality without hindering the main thread
I hope the code is correct but the emit signal binding is kinda hacky

please review

index.rst

updated the side bar

PR #11889, opened 2026-03-30
Remove outdated GDScript grammar page

Calinou:
This page has been outdated for years, and updating it is proving difficult [given how GDScript is currently designed](https://github.com/godotengine/godot-proposals/discussions/14562#discussioncomment-16359640).

- This closes https://github.com/godotengine/godot-docs/issues/6265.

PR #12163, opened 2026-07-20
Remove `Custom Godot servers` page, as it was outdated and serves no real purpose

### LINKED ISSUE #3922: Custom Godot server page is confusing and probably wrong
**Your Godot version:**
3.2
**Issue description:**
The tutorial for advanced topic which are the custom servers seems to be way compliacted than it should be. It focuses more on the Hilbert Hotel issue than the server functionality itself.
Some of code on the page doesn't even compile, there are issue like typos (for example: casting in Variant HilbertHotel::get_bus_info(RID id) function is missing the opening bracket).
Some information seems to also be just wrong, for example there's an information that programmer needs to create a dummy class because, quoting the page "Since a Godot server class creates an instance and binds it to a static singleton, binding the class might not reference the correct instance. Therefore, a dummy class must be created to reference the proper Godot server." Maybe I'm missing something but that doesn't look like a case, programmer is in full control when the class is instanciated and provide get_instance() function to use himself. Programmer is in full control what that function is returning.
https://docs.godotengine.org/en/stable/development/cpp/custom_godot_servers.html

### LINKED ISSUE #8803: Custom Godot server example missing synchronization
In the example provided for implementing a custom Godot server, the `HilbertHotel` class uses a boolean flag `exit_thread` to signal the worker thread to shut down.
This flag is read and written in different threads concurrently without synchronization.
This is technically undefined behavior and can be easily resolved by making the flag atomic -- if that's the desired fix for this.

Side question: is it normal for Godot C++ code to not use RAII for classes like `Thread` or `Mutex`? I was a bit surprised to see manual resource management in `HilbertHotel::finish`.

**URL to the documentation page:** https://docs.godotengine.org/en/stable/contributing/development/core_and_modules/custom_godot_servers.html

### LINKED ISSUE #9525: Custom Godot Server contains out of date information on how to create threads
**Your Godot version:**
4.2
**Issue description:**
Example code in the `HilbertHotel::init()` function on the "Custom Godot Server" page uses static methods `Thread::create` and `Mutex::create` which don't exist in 4.x version of the engine.
https://docs.godotengine.org/en/stable/contributing/development/core_and_modules/custom_godot_servers.html

Ivorforce:
Reverts https://github.com/godotengine/godot-docs/pull/1400

The page is 8 years old now, severely outdated (never maintained much), and partly wrong.
It also doesn't serve any purpose today, as for most of the given examples, people should use scripting, GDExtension, or modules.

Custom servers are rarely needed. If we find people commonly want to add one, we need an up to date guide for it instead.

- Closes https://github.com/godotengine/godot-docs/issues/3922
- Closes https://github.com/godotengine/godot-docs/issues/9525
- Closes https://github.com/godotengine/godot-docs/issues/8803


Calinou (approved):
Makes sense to me.

Like the [GDScript grammar page](https://github.com/godotengine/godot-docs/pull/11889), I trust someone will eventually contribute a new page that better suits modern use cases once there's a strong need for it.

AThousandShips (approved):

Mickeon:
Note that this would _technically_ close the following as the page is burned to the ground:
- https://github.com/godotengine/godot-docs/issues/3922
- https://github.com/godotengine/godot-docs/issues/9525
- https://github.com/godotengine/godot-docs/issues/8803

I'm not sure how these completely went past anyone's mind until now

skyace65 (approved):

skyace65:
Thanks!

PR #12164, opened 2026-07-20
Fix redirect destination validation, partial key overrides, and multiple broken `redirects.csv` entries

Ivorforce:
1. Due to the missing `break;` from the loop, later entries could override the redirect destination.
2. The `contributing.godotengine.org` redirects were currently not accepted in `create_redirects.py` and would have been removed the next time anybody ran the script. They are supported fine though.
3. There were several redirects broken or missing, especially from the contributing move.

> [!NOTE] 
> AI Disclaimer: I used AI to review the `redirects.csv` entries, to find and fix the broken ones. Others were identified using `--validate` — apparently never run since my changes 🙃 Personally I didn't even know the flag existed.


Calinou (approved):
Looks good to me. Thanks for updating the redirects list 🙂

Mickeon (commented):
Perhaps some of these could use more specific sections like `/about/faq.html#which-platforms-are-supported-by-godot` but this is excellent.

Mickeon (approved):

mhilbrunner (approved):

mhilbrunner:
Thank you!

PR #12167, opened 2026-07-21
Document environment variable for export in Compiling with PCK encryption key

Calinou:
This environment variable is also documented on individual platforms' export pages, but it's different from the one used at compile-time. Therefore, it's important to make this more explicit on the page that describes compiling export templates with a PCK encryption key.

- See https://github.com/godotengine/godot-docs-user-notes/discussions/770#discussioncomment-17681073.

### DIFF
--- a/engine_details/development/compiling/compiling_with_script_encryption_key.rst
+++ b/engine_details/development/compiling/compiling_with_script_encryption_key.rst
@@ -77,6 +77,11 @@ Step by step
 
    .. image:: img/encryption_key.png
 
+   If performing an export from the :ref:`command line <doc_command_line_tutorial>`,
+   before exporting, set the ``GODOT_SCRIPT_ENCRYPTION_KEY`` environment variable
+   to the same value as the one used to compile the export templates
+   (``SCRIPT_AES256_ENCRYPTION_KEY``).
+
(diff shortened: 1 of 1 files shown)


bruvzg (approved):

syntaxerror247:
Shouldn't we also document that `GODOT_SCRIPT_ENCRYPTION_KEY` can be used at compile-time too, if `SCRIPT_AES256_ENCRYPTION_KEY` is not set?

Edit: This would be valid after https://github.com/godotengine/godot/pull/121606 is merged.

Calinou:
> Shouldn't we also document that GODOT_SCRIPT_ENCRYPTION_KEY can be used at compile-time too, if SCRIPT_AES256_ENCRYPTION_KEY is not set?

Yes, I'll modify this PR if https://github.com/godotengine/godot/pull/121606 is merged before. Otherwise, I'll open a second PR once https://github.com/godotengine/godot/pull/121606 is merged.

mhilbrunner (approved):

mhilbrunner:
Considering that that PR will only apply to 4.8+ for now, lets merge this and do another one for the 4.8+ change :) Thanks y'all!

PR #12170, opened 2026-07-22
Correct default shortcut for 3D select mode

jspohr:
In the change for the new 3D toolbar, the select mode shortcut wasn't updated


Mickeon:
It would be wise to check if this is the only mention of this specific shortcut in the docs.

skyace65 (approved):

skyace65:
> It would be wise to check if this is the only mention of this specific shortcut in the docs.

It's on the intro to 2D page as well. I'll be working on a PR to update that page and can update the shortcut as part of it. Aside from that it's not mentioned elsewhere.

skyace65:

PR #12171, opened 2026-07-22
Update scripting_first_script.rst - fixed missing "void" for c#

JordanKlooster:
My editor let me know I needed a return type and I found it unclear what should and shouldn't be there so I put comments to help people see what things should look like

### DIFF
--- a/getting_started/step_by_step/scripting_first_script.rst
+++ b/getting_started/step_by_step/scripting_first_script.rst
@@ -148,7 +148,9 @@ Add the following code to your script:
 
  .. code-tab:: csharp C#
 
-    public MySprite2D()
+//public partial class MySprite2d : Sprite2D  
+//{
+    public void MySprite2D()     // class constructor, runs when this class is created as an object
     {
         GD.Print("Hello, world!");
     }


getting_started/step_by_step/scripting_first_script.rst
```diff
@@ -148,7 +148,9 @@ Add the following code to your script:
 
  .. code-tab:: csharp C#
 
-    public MySprite2D()
+//public partial class MySprite2d : Sprite2D
```
AThousandShips, line 151: Why this commented out code?

getting_started/step_by_step/scripting_first_script.rst
```diff
@@ -148,7 +148,9 @@ Add the following code to your script:
 
  .. code-tab:: csharp C#
 
-    public MySprite2D()
+//public partial class MySprite2d : Sprite2D  
+//{
+    public void MySprite2D()     // class constructor, runs when this class is created as an object
```
AThousandShips, line 153: This is incorrect, it is a constructor, they don't have a return type

AThousandShips (changes requested):

getting_started/step_by_step/scripting_first_script.rst
JordanKlooster: because the tutorial wasn't the most clear for the c# side of things, and with the constructor having the same name as the class in c# it's just a little reminder for people that you don't just paste in the GD.Print("Hello, world!");

like:

public partial class MySprite2d : Sprite2D
{
	GD.Print("Hello, world!");
}

which doesn't work.

I thought it would be helpful for people. If you think it shouldn't be there then I don't care if it gets removed.

AThousandShips: It should be formatted properly if nothing else, and doesn't make sense to have it commented out

getting_started/step_by_step/scripting_first_script.rst
JordanKlooster: yeah, now I'm confused why vscode is saying this: 
image

AThousandShips: That's because you spelled it wrong, it should be `public MySprite2d()` not `public MySprite2D()`

JordanKlooster: It's because the Editor made the d on the end of Sprite2d.cs lower case and the tutorial expects it to be upper case, and I just put "My" on the front instead of copying and pasting the file name. That's annoying.

image

AThousandShips: If there's some issue with the UI that's something to report as a bug over [here](https://github.com/godotengine/godot), but this fix isn't correct

AThousandShips:
Closing as this change is not a valid C# change, C# constructors do not have any return type (and the change to `MySprite2d` is also incorrect as it is declared as `MySprite2D` above, that could be changed but it would be a separate discussion)

PR #12182, opened 2026-07-26
Change mentions of "member" to "property" where appropriate

Mickeon:
See https://github.com/godotengine/godot/pull/121793 and [RocketChat discussion](https://chat.godotengine.org/channel/documentation/thread/tPE9nBS6iKFaX2WJm)

~~I will explore and think about this more.
Unlike the class reference, the manual seems particularly keen on distinguishing between "member variable" and "property".~~

### DIFF
--- a/tutorials/3d/introduction_to_3d.rst
+++ b/tutorials/3d/introduction_to_3d.rst
@@ -297,7 +297,7 @@ Node3Ds have a local transform, which is relative to the parent
 node (as long as the parent node is also of **or inherits from** the type
 Node3D). This transform can be accessed as a 3×4
 :ref:`Transform3D <class_Transform3D>`, or as 3 :ref:`Vector3 <class_Vector3>`
-members representing location, Euler rotation (X, Y and Z angles) and
+properties representing location, Euler rotation (X, Y, and Z angles) and
 scale.
 
 .. image:: img/tuto_3d2.webp
--- a/tutorials/plugins/running_code_in_the_editor.rst
+++ b/tutorials/plugins/running_code_in_the_editor.rst
@@ -112,8 +112,8 @@ Important information
 
 The general rule is that **any other GDScript that your tool script uses must
 *also* be a tool**. The editor is not able to construct instances from GDScript
-files without ``@tool``, which means you cannot call methods or reference member
-variables from them otherwise. However, since static methods, constants and
+files without ``@tool``, which means you cannot call methods or reference properties
+from them otherwise. However, since static methods, constants, and
 enums can be used without creating an instance, it is possible to call them or
 reference them from a ``@tool`` script onto other non-tool scripts. One exception to
 this are :ref:`static variables <doc_gdscript_basics_static_variables>`.
--- a/tutorials/scripting/c_sharp/c_sharp_exports.rst
+++ b/tutorials/scripting/c_sharp/c_sharp_exports.rst
@@ -3,9 +3,9 @@
 C# exported properties
 ======================
 
-In Godot, class members can be exported. This means their value gets saved along
+In Godot, member variables can be exported. This means their value gets saved along
 with the resource (such as the :ref:`scene <class_PackedScene>`) they're
-attached to. They will also be available for editing in the property editor.
+attached to. They will also be available for editing in the Inspector.
 Exporting is done by using the ``[Export]`` attribute.
 
 .. code-block:: csharp
@@ -19,7 +19,7 @@ Exporting is done by using the ``[Export]`` attribute.
     }
 
 In that example the value ``5`` will be saved, and after building the current project
-it will be visible in the property editor.
+it will be visible in the Inspector.
 
 One of the fundamental benefits of exporting member variables is to have
 them visible and editable in the editor. This way, artists and game designers
@@ -46,7 +46,7 @@ Exporting works with fields and properties. They can have any access modifier.
     [Export]
     public int Number { get; set; }
 
-Exported members can specify a default value; otherwise, the `default value of the type <https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/default-values>`_ is used instead.
+Exported member variables can specify a default value; otherwise, the `default value of the type <https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/default-values>`_ is used instead.
 
 An ``int`` like ``Number`` defaults to ``0``. ``Text`` defaults to null because
 ``string`` is a reference type.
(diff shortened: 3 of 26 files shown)


paulloz:
I'd be cautious with these on C# pages and code blocks, since the term property holds a different meaning there. Readers might end up with the idea that Godot bindings only work on (C#) properties, and not also on field.

Mickeon:
I acknowledge it. I was waiting for your kind of review to know where to scale back.

tutorials/3d/introduction_to_3d.rst
```diff
@@ -297,7 +297,7 @@ Node3Ds have a local transform, which is relative to the parent
 node (as long as the parent node is also of **or inherits from** the type
 Node3D). This transform can be accessed as a 3×4
 :ref:`Transform3D <class_Transform3D>`, or as 3 :ref:`Vector3 <class_Vector3>`
-members representing location, Euler rotation (X, Y and Z angles) and
+properties representing location, Euler rotation (X, Y and Z angles) and
```
AThousandShips:
```suggestion
properties representing location, Euler rotation (X, Y, and Z angles) and
```
While we're here

tutorials/plugins/running_code_in_the_editor.rst
```diff
@@ -112,8 +112,8 @@ Important information
 
 The general rule is that **any other GDScript that your tool script uses must
 *also* be a tool**. The editor is not able to construct instances from GDScript
-files without ``@tool``, which means you cannot call methods or reference member
-variables from them otherwise. However, since static methods, constants and
+files without ``@tool``, which means you cannot call methods or reference properties
+from them otherwise. However, since static methods, constants and
```
AThousandShips:
```suggestion
from them otherwise. However, since static methods, constants, and
```

tutorials/shaders/shader_reference/shading_language.rst
```diff
@@ -169,11 +169,11 @@ unsigned:
...
     uint a = uint(2); // valid
 
-Members
-~~~~~~~
+Components
+~~~~~~~~~~
 
-Individual scalar members of vector types are accessed via the "x", "y", "z" and
-"w" members. Alternatively, using "r", "g", "b" and "a" also works and is
+Individual scalar components of vector types are accessed via the "x", "y", "z" and
```
AThousandShips:
```suggestion
Individual scalar components of vector types are accessed via the "x", "y", "z", and
```

tutorials/shaders/shader_reference/shading_language.rst
```diff
@@ -169,11 +169,11 @@ unsigned:
...
+"w" components. Alternatively, using "r", "g", "b" and "a" also works and is
```
AThousandShips:
```suggestion
"w" components. Alternatively, using "r", "g", "b", and "a" also works and is
```

Calinou (approved):
Looks good to me.

AThousandShips (approved):
LGTM

Mickeon:
Rebased. Ideally I would still love a proper DotNET review because as @paulloz pointed out, a "property" in C# specifically is not the same concept as a property in Godot.

tutorials/scripting/gdscript/gdscript_advanced.rst
```diff
@@ -458,7 +458,7 @@ And it can be used like any other iterator:
     for i in itr:
         print(i) # Will print 0, 2, and 4.
 
-It is possible but discouraged to store the state in a member variable.
+It is possible but discouraged to store the state in a property.
```
Mickeon, line 461: "Member variable" was correct, too for [the same reason brought up before](https://github.com/godotengine/godot/pull/121793#discussion_r3656769875).
> Looking at other languages, this should probably stay `member variable`. Whether something is exposed to Godot does not matter for this advice.

However, this suggestion is present in the GDScript Advanced page, and all member variables in GDScript are properties by definition. So I'm not sure. I'm asking for careful examination around the GDScript docs.

tutorials/scripting/gdscript/gdscript_styleguide.rst
```diff
@@ -874,14 +874,14 @@ For inner classes, use single-line declarations:
 Signals and properties
 ~~~~~~~~~~~~~~~~~~~~~~
 
-Write signal declarations, followed by properties, that is to say, member
-variables, after the docstring.
+Write signal declarations, followed by properties (also known as member
+variables), after the docstring.
```
Mickeon, line 878: Given what had been discussed [prior](https://github.com/godotengine/godot/pull/121793), "member variables" and "properties" are still distinct concepts, so saying they're exactly equal _may_ be misleading. But also not in this case, since all member variables are properties in GDScript.

Mickeon (commented):
I would like to bring attention to a few edge-cases. I don't want to make the confusion worse by accident.

tutorials/scripting/c_sharp/c_sharp_exports.rst
```diff
@@ -3,9 +3,9 @@
 C# exported properties
 ======================
 
-In Godot, class members can be exported. This means their value gets saved along
+In Godot, properties can be exported. This means their value gets saved along
```
paulloz:
```suggestion
In Godot, fields and properties can be exported. This means their value gets saved along
```

This would be the most precise, but it might get unreadable quite fast to write "fields and properties" everywhere 🤔 

Also, the example below only shows the use of `[Export]` on a property.

Maybe a better solution would be to have a note explaining how everything explained for properties on this page is also true for fields? I'm not sure.

Mickeon: > This would be the most precise, but it might get unreadable quite fast to write "fields and properties" everywhere 🤔

I do not mind generalising these into "member variables" once more. Just so long as it's not "class members" because that's not right.

> Maybe a better solution would be to have a note explaining how everything explained for properties on this page is also true for fields? I'm not sure.

Not a bad idea, actually. Not sure I want to do it in this PR, nor what to write exactly. But the ambiguity does exist, and perhaps a note would be nice.

paulloz: > > Maybe a better solution would be to have a note explaining how everything explained for properties on this page is also true for fields? I'm not sure.
> 
> Not a bad idea, actually. Not sure I want to do it in this PR, nor what to write exactly. But the ambiguity does exist, and perhaps a note would be nice.

If you think this is the best solution, I can write the note in a future PR (if that helps).

Mickeon: I think the "best solution" would be to say "member variables" as far as this page is concerned since, as you say, both C# fields and properties can be exported.

However, we could further discuss this, perhaps note it down in the C# Differences page, because that page currently may use "property" to refer to two different concepts, _technically_.

> In GDScript, the setters/getters of a property can be called directly
> In C#, only the property is defined

I admit that I am overthinking this a lot, but I do find the distinction important. The confusion of "why isn't my property present in `Object.get_property_list()`" stems from this kind of thing.

paulloz: Honestly both options sound good to me. This specific vocabulary thing has also been bothering me for years, so thanks for taking time to address it!

Mickeon:
Updated this PR to reel back on the replacements in the C# Exports page, but still use "member variable" instead of "member" (which, reminder, is misleading)

PR #12186, opened 2026-07-27
Fix 'subemitters' to 'sub-emitters' for consistency with the particle…

is-snndev:
… sub-emitters tutorial page

The docs use "subemitters" (no hyphen) in list_of_features.rst:
https://github.com/godotengine/godot-docs/blob/master/about/list_of_features.rst

But the dedicated tutorial page consistently uses "sub-emitters" (hyphenated) 
throughout its body text:
https://docs.godotengine.org/en/stable/tutorials/3d/particles/subemitters.html

This PR makes the spelling consistent with the tutorial page.

### DIFF
--- a/about/list_of_features.rst
+++ b/about/list_of_features.rst
@@ -384,7 +384,7 @@ Godot 4 includes three renderers:
 
 **Particles:**
 
-- GPU-based particles with support for subemitters (2D + 3D), trails (2D + 3D),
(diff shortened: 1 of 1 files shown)


Mickeon:
The spelling is inconsistent everywhere, not just this one case. The class reference documentation most often refers to them as "subemitters", for example.
I would leave "spellchecking" of hyphened words to someone else until a consensus is reached.

PR #12193, opened 2026-07-28
Change primitive type from TRIANGLES to TRIANGLE_STRIP in example of Using ArrayMesh

Wiechciu:
The original example in the [Using ArrayMesh](https://docs.godotengine.org/en/4.7/tutorials/3d/procedural_geometry/arraymesh.html) documentation uses primitive type TRIANGLES, which requires all triangles to be defined separately (number of vertices should be a multiple of 3), whereas the rectangle in the example is constructed with 4 vertices, which is possible only with TRIANGLE_STRIP primitive type.

The example code without this change results in an error during runtime:
```
E 0:00:01:195   draw_list_draw: Vertex amount (4) must be a multiple of the amount of vertices required by the render primitive (3).
  <C++ Error>   Condition "(to_draw % draw_list.validation.pipeline_primitive_divisor) != 0" is true.
  <C++ Source>  servers/rendering/rendering_device.cpp:6154 @ draw_list_draw()
```

This PR fixes this error by changing the primitive type to TRIANGLE_STRIP.

### DIFF
--- a/tutorials/3d/procedural_geometry/arraymesh.rst
+++ b/tutorials/3d/procedural_geometry/arraymesh.rst
@@ -140,7 +140,7 @@ by adding each array to ``surface_array`` and then committing to the mesh.
     surface_array[Mesh.ARRAY_INDEX] = indices
 
     # No blendshapes, lods, or compression used.
-    mesh.add_surface_from_arrays(Mesh.PRIMITIVE_TRIANGLES, surface_array)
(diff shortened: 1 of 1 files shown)


clayjohn (changes requested):
It should be using indices and TRIANGLES, not a triangle strip

Wiechciu:
Hi @clayjohn, that's true, my bad. It works with indices and TRIANGLES.
Closing this PR.

PR #12200, opened 2026-07-29
Add a subsection for microbenchmarks to the profiling page.

Ivorforce:
- Addresses https://github.com/godotengine/godot-proposals/issues/12319#issuecomment-5118864932

We currently have no documented way of performing microbenchmarks.

There is no perfectly suitable page to add this to, but I think the profiling page is a well suited candidate even though it's not technically profiling, because people are somewhat likely to find it here.
But I'm open to moving the information elsewhere!

### DIFF
--- a/engine_details/development/profiling/index.rst
+++ b/engine_details/development/profiling/index.rst
@@ -113,3 +113,99 @@ All recommended profilers
    perfetto
    tracy
    very_sleepy
+
+Microbenchmarks
+---------------
+
+While not technically profiling, microbenchmarks are a related concept: after you've identified
+your hotspot, you'll want a simple and isolatable way to test whether what you're doing has an impact.
+While re-profiling can be an option, sometimes a microbenchmark can be simpler.
+
+You can see example benchmark setups in the subsections below. Between C++ and GDScript, GDScript
+benchmarks are usually the more appropriate choice. Because they involve the GDScript language and
+its overhead, GDScript benchmarks are truthful to how most people will experience your changes.
+C++ benchmarks are more versatile and can measure smaller performance differences. However, they
+are also more difficult to get right.
+In practice, it can often be good to benchmark both.
+
+.. note::
+
+    To benchmark effectively can be difficult. Benchmarks can easily lead you astray, for example
+    by not representing the situation faithfully, by failing to account for compiler optimizations
+    and other nuances, or by measuring in an unreliable way such that you record noise instead of
+    an actual performance change. Before you start benchmarking, please read up on benchmarking
+    guides and best practices. A good place to start is Gregg's `"Active Benchmarking" <https://www.brendangregg.com/activebenchmarking.html>`__,
+    which provides a high level overview and quick checklist for benchmarking.
+
+GDScript benchmarks
+~~~~~~~~~~~~~~~~~~~
+
+To run a GDScript benchmark, start by creating a ``benchmark.gd`` file with the following
+contents:
+
+.. code-block:: gdscript
+
+    extends SceneTree
+
+    func _init():
+        const N = 1_000_000
+
+        var t0 := Time.get_ticks_usec()
+        for i in N:
+            pass  # Do the thing you want to benchmark here.
+        var t1 := Time.get_ticks_usec()
+
+        var ns_per_op := (t1 - t0) * 1000.0 / N
+        print("Benchmark result: %.1f ns/op" % ns_per_op)
+        quit()
+
+Edit the file to add your benchmark.
+You can run the benchmark using ``godot --headless -s benchmark.gd``.
+
+.. note::
+
+    At the time of writing, GDScript performs few code optimization. Dead code elimination,
+    for example, is generally not performed. Therefore, it is easier to write a decently
+    representative benchmark in GDScript than in C++. This might change in the future.
+
+C++ benchmarks
+~~~~~~~~~~~~~~
+
+To run a C++ benchmark, create the file ``tests/core/test_user_bench.cpp`` with the following contents:
+
+.. code-block:: cpp
+
+    #include "tests/test_macros.h"
+
+    TEST_FORCE_LINK(test_user_bench)
+
+    #include <chrono>
+    #include <cstdio>
+
+    static void user_bench() {
+        const int N = 1'000'000;
+        static uintptr_t sink = 0; // Defeats dead-code elimination.
+
+        auto t0 = std::chrono::steady_clock::now();
+        for (int i = 0; i < N; i++) {
+            // Do the thing you want to benchmark here.
+            // sink += (uintptr_t)something; // Update the sink from the result somehow.
+        }
+        auto t1 = std::chrono::steady_clock::now();
+
+        double ns_per_op = std::chrono::duration<double, std::nano>(t1 - t0).count() / N;
+        printf("Benchmark result: %.1f ns/op (sink %zu)\n", ns_per_op, (size_t)sink);
+    }
+
+    REGISTER_TEST_COMMAND("user-bench", &user_bench)
+
+Edit the file to add your benchmark.
+Compile Godot with ``tests=yes``, and run the benchmark using ``godot --test user-bench``.
+
+.. note::
+
+    C++ benchmarks can be fickle and can easily lead you astray unless you have a strong
+    foundation of knowledge about C++ and compilers. Before you start benchmarking, read
+    up on guides and best practices about how to benchmark C++.  You
+    can find free in-depth guides in Bakhvalov's `"Performance Analysis and Tuning on Modern CPUs" <https://github.com/dendibakh/perf-book>`__,
+    and Agner Fog's `"Software Optimization Resources" <https://www.agner.org/optimize/>`__.


lawnjelly:
More things to mention (specially for c++):

* do the first test runs without timing
* first run to commit pages that have been allocated
* first run to warm the cache
* run the main loop multiple times, the smallest measurement is often most useful
* don't run anything else on your PC while benchmarking (especially e.g. browsers)

These are all very sensitive to other things using your OS.

As an alternative for microbenching you can use RDTSC / RDTSCP (forces processor to finish previous instructions) to measure CPU clock ticks. This used to be easier, but on modern multicore PCs which scale speeds there's more considerations, but most have invariant TSC on the CPU regardless of stepping (check this though for your CPU).

(in fact I must revisit using RDTSC for a recent couple of benchmarks in PRs)

More possibles:
* if you run the same data each loop, it won't stress the cache, but it may do e.g. perfect branch prediction
* good practice often to create a random data set before benchmarking
* the size of the data set will determine how much cache thrashing will go on
* might be worth mentioning websites dedicated to benchmarking (sometimes you can do the benchmark online, with pre-built helpers, especially for code which is localised and doesn't need to pull in the rest of Godot)
* one I discovered today - don't use `rand()` within a benchmark, it's not O(1), and might do locking for threads

Obviously we shouldn't need to be a full reference, but we should probably include some of the major pitfalls, otherwise we'll get new benchmarkers and have to explain the same artifacts each time. And emphasize how difficult it can be to get correct results, I regularly get artifacts even after having done benchmarks here and there for many years :grin: .

engine_details/development/profiling/index.rst
```diff
@@ -113,3 +113,96 @@ All recommended profilers
    perfetto
    tracy
    very_sleepy
+
+Microbenchmarks
+---------------
+
+While not technically profiling, microbenchmarks are a related concept: after you identified
```
AThousandShips:
```suggestion
While not technically profiling, microbenchmarks are a related concept: after you've identified
```

engine_details/development/profiling/index.rst
```diff
@@ -113,3 +113,96 @@ All recommended profilers
...
+    #include "tests/test_macros.h"
+
+    TEST_FORCE_LINK(test_user_bench)
+
+    #include <chrono>
+    #include <cstdio>
+
+    static void user_bench() {
+        const int N = 1'000'000;
+        static uintptr_t sink = 0; // Defeats dead-code elimination
```
AThousandShips:
```suggestion
        static uintptr_t sink = 0; // Defeats dead-code elimination.
```

engine_details/development/profiling/index.rst
```diff
@@ -113,3 +113,96 @@ All recommended profilers
...
+
+        auto t0 = std::chrono::steady_clock::now();
+        for (int i = 0; i < N; i++) {
+            // Do the thing you want to benchmark here
```
AThousandShips:
```suggestion
            // Do the thing you want to benchmark here.
```

engine_details/development/profiling/index.rst
```diff
@@ -113,3 +113,96 @@ All recommended profilers
...
+.. code-block:: gdscript
+
+    extends SceneTree
+
+    func _init():
+        const N = 1_000_000
+
+        var t0 := Time.get_ticks_usec()
+        for i in N:
+            pass  # Do the thing you want to benchmark here
```
AThousandShips:
```suggestion
            pass  # Do the thing you want to benchmark here.
```

Ivorforce:
> More things to mention (specially for c++):
> 
> * do the first test runs without timing
> * first run to commit pages that have been allocated
> * first run to warm the cache
> * run the main loop multiple times, the smallest measurement is often most useful
> * don't run anything else on your PC while benchmarking (especially e.g. browsers)

Those are all good tips, but I deliberately didn't include any in my PR.

If we wanted to be a "best benchmarking practices" reference, we'd have to include a lot more info, for example about how to avoid dead code elimination or why cold checks are often more important than hot checks. I don't think this is in scope for the docs, so I opted for saying "hey be careful about it and read a guide somewhere" instead.

> Obviously we shouldn't need to be a full reference, but we should probably include some of the major pitfalls, otherwise we'll get new benchmarkers and have to explain the same artifacts each time. And emphasize how difficult it can be to get correct results, I regularly get artifacts even after having done benchmarks here and there for many years 😁 .

I'd rather find a blessed guide to link to somewhere and not give any tips here at all. I don't have one at hand, unfortunately, but it sounds like you might be interested in writing one 😄

lawnjelly:
> but it sounds like you might be interested in writing one

Absolutely not :rofl: 
Yeah, agreed maybe we can just link to some good benchmarking websites.

https://quick-bench.com
https://github.com/google/benchmark

Unfortunately most tutorials seem to be directing you to use pre-existing libraries / tools for benchmarking instead of doing it from scratch, which is a bit sad, but maybe we'll find one.

Maybe it's because it's so easy to get wrong (even for the pros!), beginners get steered away from it.

Ivorforce:
I had an AI research credible stable resources we can link, found 3:

- https://www.brendangregg.com/activebenchmarking.html: High level overview, good to link at the top. Site has existed for 15+ years so unlikely to go anywhere. Well known performance engineer.
- https://github.com/dendibakh/perf-book: An in-depth guide on performance analysis and testing.
- https://www.agner.org/optimize/: We all know this one, Agner Fog's guide on CPU optimization.

I've only heard of the last one, but the first two also look like incredible fits, so I think they are all worth linking. I'll adjust in a push.
Regarding your links, I'd rather avoid linking anything that's using libraries. Most benchmarks will probably be hand written and a library will be too much mental overhead for a single PR.

lawnjelly:
> Regarding your links, I'd rather avoid linking anything that's using libraries. Most benchmarks will probably be hand written and a library will be too much mental overhead for a single PR.

I mostly feel the same way, although I'm trying to begrudgingly recognise that for most people spending hours diagnosing artifacts in their benchmarks isn't their idea of fun (like it might be for you or I :grin: ), and I know some quite advanced programmers will still just use off the shelf libs rather than write their own, just because of the pitfalls.

I did check brendangregg's website earlier but didn't spot that particular article (most were quite in depth).

The Bakhvalov book I may have looked through before at some point, seems like a good one to link covering optimization and benchmarking. :+1:

Ivorforce:
If you think some people would prefer a library for it we could mention that and perhaps link one or two examples.

lawnjelly:
Looks good.

Something may be worth adding that gdscript benchmarking can be difficult to use / not hugely enlightening when the cost of the function under test (e.g. c++) is very cheap in relation to the overheads of gdscript.

In many situations you get so much variation from the gdscript overheads, these might be 1000x as e.g. a c++ function that you are trying to optimize. But still I've seen people making PRs and trying to use a gdscript benchmark where it is clearly inappropriate.

Not quite sure how to word it.

lawnjelly (approved):
Although there's a few things we could add, this is fine to get started imo.

engine_details/development/profiling/index.rst
```diff
@@ -113,3 +113,98 @@ All recommended profilers
...
+
+Edit the file to add your benchmark.
+You can run the benchmark using ``godot --headless -s benchmark.gd``.
+
+.. note::
+
+    At the time of writing, GDScript performs few code optimization. Dead code elimination,
+    for example, is generally not performed. Therefore, it is easier to write a decently
+    representative benchmark in GDScript than in C++. This might change in the future. You
+    can find free in-depth guides in Bakhvalov's `"Performance Analysis and Tuning on Modern CPUs" <https://github.com/dendibakh/perf-book>`__, and Agner Fog's `"Software Optimization Resources" <https://www.agner.org/optimize/>`__.
```
AThousandShips:
```suggestion
    can find free in-depth guides in Bakhvalov's `"Performance Analysis and Tuning on Modern CPUs" <https://github.com/dendibakh/perf-book>`__,
    and Agner Fog's `"Software Optimization Resources" <https://www.agner.org/optimize/>`__.
```

engine_details/development/profiling/index.rst
```diff
@@ -113,3 +113,98 @@ All recommended profilers
...
+are also more difficult to get right.
+In practice, it can often be good to benchmark both.
+
+.. note::
+
+    To benchmark effectively can be difficult. Benchmarks can easily lead you astray, for example
+    by not representing the situation faithfully, by failing to account for compiler optimizations
+    and other nuances, or by measuring in an unreliable way such that you record noise instead of
+    an actual performance change. Before you start benchmarking, please read up on benchmarking
+    guides and best practices. A good place to start is Gregg's `"Active Benchmarking" <https://www.brendangregg.com/activebenchmarking.html>`__, which provides a high level overview and quick checklist for
```
AThousandShips:
```suggestion
    guides and best practices. A good place to start is Gregg's `"Active Benchmarking" <https://www.brendangregg.com/activebenchmarking.html>`__,
    which provides a high level overview and quick checklist for
```

AThousandShips (approved):
LGTM

mhilbrunner (approved):

engine_details/development/profiling/index.rst
```diff
@@ -113,3 +113,99 @@ All recommended profilers
...
+
+    To benchmark effectively can be difficult. Benchmarks can easily lead you astray, for example
+    by not representing the situation faithfully, by failing to account for compiler optimizations
+    and other nuances, or by measuring in an unreliable way such that you record noise instead of
+    an actual performance change. Before you start benchmarking, please read up on benchmarking
+    guides and best practices. A good place to start is Gregg's `"Active Benchmarking" <https://www.brendangregg.com/activebenchmarking.html>`__,
+    which provides a high level overview and quick checklist for benchmarking.
+
+GDScript Benchmarks
+~~~~~~~~~~~~~~~~~~~
```
Calinou: Use sentence case for headings:

```suggestion
GDScript benchmarks
~~~~~~~~~~~~~~~~~~~
```

engine_details/development/profiling/index.rst
```diff
@@ -113,3 +113,99 @@ All recommended profilers
...
+You can run the benchmark using ``godot --headless -s benchmark.gd``.
+
+.. note::
+
+    At the time of writing, GDScript performs few code optimization. Dead code elimination,
+    for example, is generally not performed. Therefore, it is easier to write a decently
+    representative benchmark in GDScript than in C++. This might change in the future.
+
+C++ Benchmarks
+~~~~~~~~~~~~~~
```
Calinou:
```suggestion
C++ benchmarks
~~~~~~~~~~~~~~
```

Calinou (approved):
Looks good to me.

engine_details/development/profiling/index.rst
```diff
@@ -113,3 +113,99 @@ All recommended profilers
...
+        auto t1 = std::chrono::steady_clock::now();
+
+        double ns_per_op = std::chrono::duration<double, std::nano>(t1 - t0).count() / N;
+        printf("Benchmark result: %.1f ns/op (sink %zu)\n", ns_per_op, (size_t)sink);
+    }
+
+    REGISTER_TEST_COMMAND("user-bench", &user_bench)
+
+Edit the file to add your benchmark.
+Compile Godot with ``tests=yes``, and run the benchmark using ``godot --test user-bench``.
```
Synzorasize, line 203: Is this code better to use for microbenchmarking than using `doctest`? I used the `doctest` format in my module because it was the easiest to copy (due to established precedence). Just would like to clarify.

Ivorforce: Not necessarily. The code above is just the minimal Godot-integrated way to benchmark. 

Libraries like `doctest` do some things for you, which can be good or bad, really depends on what you're doing.

Synzorasize: Yeah, it seems like `doctest` is more oriented towards checking if things didn't break compat rather than benchmarking. It has a `--duration` param, but it mostly just checks the duration of the entire test case than between each assertion for example, so it's not finely-controllable (format-wise, you probably could still use `Time.get_ticks_usec()` but the logs would be spammed, then).

skyace65 (approved):

skyace65:
Thanks!

Issue #12195, opened 2026-07-29
Clarify array- / dictionary-typed variable defaults


CodeSmith32:
**Your Godot version:**
Godot 4.7 documentation

**Issue description:**
The documentation does clearly state that primitives (`int`, `String`, `bool`) are initialized as empty. This is logical, given that they are passed by copy rather than by reference.

But, because arrays and dictionaries are passed by reference instead, the documentation needs to clarify what happens when we do this:

```gdscript
var array: Array
print(array)
array.append("abc")

var dict: Dictionary
print(dict)
dict["abc"] = 4

var typedArray: PackedInt32Array
print(typedArray)
typedArray.append(1234)

# applies to other Packed*Array types
```

Does this crash at runtime? Does it print an empty array / dictionary? Do the assignments work?

If unassigned arrays / dictionaries are given empty instances (which the docs seem to suggest, but do not state clearly), then, since arrays / dictionaries are passed by reference, does this mean that the following wastefully generates an unused array / dictionary that is immediately replaced?

```gdscript
var array: Array
var dict: Dictionary

array = [1, 2, 3]
dict = {"a": 1, "b": 2}
```

It is evident that the default value cannot be `null`, since that isn't a legal value (even though that would be far more preferable):

```gdscript
var array: Array = null  # Cannot assign a value of type "null" as "Array".
array = null  # Cannot assign a value of type "null" as "Array".
```

Finally, the docs also need to clarify garbage collection protocol for arrays and dictionaries. This is clearly documented for objects, which are not freed automatically, unless we use `RefCounted` and the count reaches 0.

Like objects, dictionaries and arrays are also passed by reference. But, because the docs don't clarify GC protocol, it becomes unclear whether or not something like the following creates a memory leak:

```gdscript
var dict: Dictionary
for i in 50:
    dict = {"a": {"b": {}}}
    dict.a.b.c = dict
# if Dictionary extends something like RefCounted, this creates a memory leak
```

Leaving out details like this makes it very hard for professionals to use a good engine like Godot with confidence that their code doesn't contain severe memory leaks or other issues.

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/4.7/tutorials/scripting/gdscript/gdscript_basics.html#initialization-order

PR #12201, opened 2026-07-30
GDScript: Array / Dictionary / packed array reference counting

CodeSmith32:
- Warn about cyclic reference memory leaks with arrays and dictionaries.
- Clarify that uninitialized arrays and dictionaries are also initialized with empty values.

Resolves issue https://github.com/godotengine/godot-docs/issues/12195

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_basics.rst
+++ b/tutorials/scripting/gdscript/gdscript_basics.rst
@@ -693,6 +693,30 @@ The exceptions are ``Object``, ``Array``, ``Dictionary``, and packed arrays
 All arrays, ``Dictionary``, and some objects (``Node``, ``Resource``)
 have a ``duplicate()`` method that allows you to make a copy.
 
+.. warning::
+
+    ``Array``, ``Dictionary`` and packed arrays are reference-counted. That is,
+    instead of having a ``free()`` method, they are automatically freed when
+    there are no more references to them. This means, however, that creating a
+    cyclic reference can cause a memory leak. As an example, an array containing
+    itself as its first element creates a cyclic reference:
+
+    ::
+
+        # The array creates a cyclic reference to itself.
+        # If the array is not removed from itself before the variable
+        # goes out of scope, then it will never be freed.
+        var array: Array = []
+        array[0] = array
+
+    This caution also applies to dictionaries.
+
+.. note::
+
+    Even though ``Array``, ``Dictionary`` and packed array classes are
+    reference-counted, they do not extend the ``RefCounted`` class. Instead,
+    reference counting is directly handled by these classes internally.
+
 Basic built-in types
 ~~~~~~~~~~~~~~~~~~~~
 
@@ -1105,7 +1129,8 @@ Member variables are initialized in the following order:
 
 1. Depending on the variable's static type, the variable is either ``null``
    (untyped variables and objects) or has a default value of the type
-   (``0`` for ``int``, ``false`` for ``bool``, etc.).
+   (``0`` for ``int``, ``false`` for ``bool``, ``[]`` for ``Array``, ``{}``
+   for ``Dictionary``, etc.).
 2. The specified values are assigned in the order of the variables in the script,
    from top to bottom.
 


CodeSmith32:
@Mickeon / @AThousandShips,

Are Arrays and Dictionaries not ref-counted? If not, then please answer the following:
- How are they garbage collected or memory managed?
- **Where is this clearly outlined in the documentation?**

TMK, the second question cannot be answered, because nothing anywhere in the documentation clarifies how Arrays, Dictionaries, and packed arrays are actually memory-managed. This PR addresses that.

If you have questions / comments, please explain. I'm willing to tweak / disregard this PR if there is reason to, but I was frustrated when I had to request AI to scan the actual Godot codebase for answers to questions that should be plainly available in the docs.

CodeSmith32:
As a side note, "passed by reference" *does not mean* the same thing as "reference-counted".

The docs do clearly state that Arrays, Dictionaries, and packed arrays are passed by reference. This is good. What it does *not* say is that they are 'ref-counted', which is one of many ways that values that are 'passed by reference' can be garbage collected. Other methods include manual freeing, [tracing GC](https://en.wikipedia.org/wiki/Tracing_garbage_collection), etc.

If Godot used a tracing GC for these types, creating a cyclic reference would not create a memory leak, since it would eventually be cleaned up. Because Godot uses ref-counting instead (or at least that's strongly [what the c++ code suggests](https://github.com/godotengine/godot/blob/master/core/variant/array.cpp#L79)), creating cyclic references does cause memory leaks.

Ivorforce:
@CodeSmith32 I understand the frustration of missing information, but your tone to our maintainers is inappropriate. We are only a small group of people and there are thousands of open PRs. It [takes a while](https://contributing.godotengine.org/en/latest/pull_requests/pr_workflow.html#when-will-a-pull-request-get-reviewed) so please be patient.

As a core maintainer, i agree this information should be explained somewhere in the docs. I appreciate you taking the time to add it. However, this isn't the right place. I would suggest to add it to the respective `Variant` type pages. Perhaps a dedicated page for ownership in GDScript could be useful too, but that would be a larger piece of work.

CodeSmith32:
Thanks @Ivorforce, I'm sorry.. I wasn't trying to come across rude. I just wasn't sure of the source of confusion. But maybe it still came out poorly.. :-/
I do understand PRs can take awhile to be merged. Thank you for the link, too. I'll keep that timeframe in mind.

-----

I appreciate your note about where it belongs. I can work on moving it to the correct location. To be clear, you're suggesting it belongs on these pages? (I agree, this would make a lot more sense)
https://docs.godotengine.org/en/stable/classes/class_dictionary.html
https://docs.godotengine.org/en/stable/classes/class_array.html
<details>
<summary>And all PackedArray pages...</summary>

https://docs.godotengine.org/en/stable/classes/class_packedbytearray.html
https://docs.godotengine.org/en/stable/classes/class_packedcolorarray.html
https://docs.godotengine.org/en/stable/classes/class_packedfloat32array.html
https://docs.godotengine.org/en/stable/classes/class_packedfloat64array.html
https://docs.godotengine.org/en/stable/classes/class_packedint32array.html
https://docs.godotengine.org/en/stable/classes/class_packedint64array.html
https://docs.godotengine.org/en/stable/classes/class_packedstringarray.html
https://docs.godotengine.org/en/stable/classes/class_packedvector2array.html
https://docs.godotengine.org/en/stable/classes/class_packedvector3array.html
https://docs.godotengine.org/en/stable/classes/class_packedvector4array.html
</details>

-----

Also, what are your thoughts concerning including a few by-reference types in the default initialization note?
https://github.com/godotengine/godot-docs/commit/015c8127a8f9551ba2258dc98b423e687086a109#diff-9caada978020928a0eb92d8e5f4f59de3712e6a7c3317af593dc83b62b5530a0

```
1. Depending on the variable's static type, the variable is either ``null``
   (untyped variables and objects) or has a default value of the type
   (``0`` for ``int``, ``false`` for ``bool``, ``[]`` for ``Array``, ``{}``
   for ``Dictionary``, etc.).
```

One thing I was looking for was if I can declare a Dictionary / Array typed variable, and then immediately use it without assigning it first. e.g.,

```
var array: Array
array.append("123")

var dict: Dictionary
dict.xyz = 5
```

Since this code works, it means arrays and dictionaries are pre-initialized. Since these types are passed by reference, I figured including a few of them as part of this note might help clarify that the rule also applies to passed-by-reference types too, and not just stack-memory types.

Thank you

Ivorforce:
Yea the individual pages like https://docs.godotengine.org/en/stable/classes/class_array.html seem like the right spot to me.

Something about initialisation also makes sense to me, probably on the pages themselves. Array and Dictionary have no "null" state so if they're type declared as such, you'll always be able to use them as usual eben when not assigned explicitly, 

P.S. I'm not a docs maintainer so I likely won't be able to review this pr to merge.

CodeSmith32:
Thanks @Ivorforce,

No worries, that's fine. I still appreciate the feedback.
I'll work on getting this PR updated!

PR #12205, opened 2026-07-30
Fix outdated mention of "future" interactive music

Mickeon:
Godot `4.3` by the way.

Part of me is inclined to remove this mention outright, as it feels very off-handed and the descriptions of each property mention this already. Not as nicely as I'd like, but they do.


fire:
As far as I know interactive music support is an actively supported part of Godot Engine.

image

https://github.com/godotengine/godot/pull/64488

fire (approved):
Approved as per previous comment about already implemented.

Calinou (approved):
Looks good to me.

skyace65 (approved):

skyace65:
Thanks!

PR #12176, opened 2026-07-25
Add `AreaLight3D` references to tutorials

Arecher:
Added `AreaLight3D` as a reference across the tutorial documentation, where ever it was missing.

- Added AreaLight3D as a clustered element across multiple pages
- Added AreaLight3D as one of the four lights that is referenced throughout the [Lights and Shadows tutorial](https://docs.godotengine.org/en/stable/tutorials/3d/lights_and_shadows.html)
- Added AreaLight3D as a light that uses Lumens in the [Physical light tutorial](https://docs.godotengine.org/en/latest/tutorials/3d/physical_light_and_camera_units.html)

### Requires Confirmation:
While I can confirm they use lumens and have basic light settings, **I am not 100% sure that Arealights are part of the same *clustered elements* approach as omnilight and spotlights.** The[ AreaLight3D documentation](https://docs.godotengine.org/en/latest/classes/class_arealight3d.html#description) seems to hint at that with:

> **Performance:** Area lights are more demanding on the GPU compared to omni and spot lights. In Forward+, there is an additional GPU cost on all rendered objects as soon as one area light is present in the view frustum (due to the nature of **clustered lighting**). 

but perhaps that refers to something else. **This should be double-checked.**

Similarly, **whether AreaLight3D's shadow settings can be set on runtime** and **whether they have their own shadow texture**, like Omni- and Spotlights is something I **presumed** due to the AreaLight working similarly to those positional lights, and thusfar the **DirectionalLight3D** seems to be the only exception. Someone with more in-depth knowledge than me should at least confirm whether that's likely true.

### Other notes
Keep in mind the page for [SDFGI](https://docs.godotengine.org/en/stable/tutorials/3d/global_illumination/using_sdfgi.html) already has an AreaLight3D reference added in PR https://github.com/godotengine/godot-docs/pull/12175, which I made earlier today.

I switched between formatting it as  `AreaLight3D` or `area light` depending on the context and how other lights were referenced. But this seems extremely inconsistent throughout the documentation.

PR #12212, opened 2026-07-31
Fix incorrect probe count limit in Forward+ in Reflection probes

Calinou:
The number of reflection probes that can be rendered at once is actually limited to 256 in Forward+.

[Godot Contributors Chat discussion](https://chat.godotengine.org/channel/rendering?msg=o3qBcSMuf84ofZFmy)

### DIFF
--- a/tutorials/3d/global_illumination/reflection_probes.rst
+++ b/tutorials/3d/global_illumination/reflection_probes.rst
@@ -178,11 +178,11 @@ Limitations
 -----------
 
 When using the Forward+ renderer, Godot uses a *clustering* approach for
-reflection probe rendering. As many reflection probes as desired can be added (as long as
-performance allows). However, there's still a default limit of 512 *clustered
-elements* that can be present in the current camera view. A clustered element is
-an omni light, a spot light, a :ref:`decal <doc_using_decals>` or a
-:ref:`reflection probe <doc_reflection_probes>`. This limit can be increased by adjusting
+reflection probe rendering. Up to 256 reflection probes can be visible at a time.
+However, there's still a default limit of 512 *clustered elements* that can be present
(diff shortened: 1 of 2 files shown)


Mickeon:
Note in advance that this will conflict with https://github.com/godotengine/godot-docs/pull/12176 when merged

Image-unavailable:
Reflection probes being supported to an unlimited amount is also mentioned here: https://docs.godotengine.org/en/stable/tutorials/rendering/renderers.html#global-illumination

skyace65:
Rebase needs to be done as the other PR was merged. Also please look over Image-unavailable's comment

skyace65:
@Calinou poke

skyace65 (approved):

Image-unavailable:
Actually, I tested this, and it seems like it's not a matter of how many probes are *visible* in the camera frustum but rather how many probes are *active and rendered to the reflection atlas* at once (that is to say, how many visible reflection probes are in the scene tree).

For example in a simple test scene, having 5 probes with a max of 4 running in 4.7.1 results in this error being printed, then an error spam, and finally a crash, despite only 4 probes intersecting the camera frustum:

```
ERROR: Reflection probe atlas index invalid. Maximum amount of reflection probes in use (4) may have been exceeded, reflections will not display properly. Consider increasing Rendering > Reflections > Reflection Atlas > Reflection Count in the Project Settings.
```

The test scene:

[reflection-probe-test.zip](https://github.com/user-attachments/files/31923300/reflection-probe-test.zip)

I ran into the error spam before, but not a crash so not sure what's going on there.

If the editor crashes on startup, edit the project file and set the maximum reflection probe count to 8 to prevent that.

skyace65 (approved):

skyace65:
Thanks!

PR #12213, opened 2026-07-31
Update the list of modules that can be excluded from build

Synzorasize:
After godotengine/godot#121548 and godotengine/godot#115474's core-shattering changes, the `Disabling unwanted modules` section of the `engine_details/development/compiling/optimizing_for_size.rst` page needed to be updated. `regex` is now core, so that can't be disabled, whereas `tilemap` can now be because it's a module.

Also swapped the `e.g.` "regex support" part with a mention of ``tilemap`` instead, since `TileMapLayer`s are still very common in 2D games just like `MP3`s/`OGG`s are common, so I think it's of particular mention. (Especially since I wouldn't want people blindly copy-pasting the command into their command-line and wondering why their game broke, especially those who haven't learned that it's a module now. It's what I would do by accident :P ) If it's okay to not replace "regex support" with anything or something else is better there, though, please suggest so.

Also please update me if there's another module/core shift in the works so I can update this PR accordingly. :) Also please tell me if any other doc page mentions specific modules. I quickly searched and I don't see any other pages that reference `TileMapLayer` or `RegEx` as being part of a module/core/etc.

### DIFF
--- a/engine_details/development/compiling/optimizing_for_size.rst
+++ b/engine_details/development/compiling/optimizing_for_size.rst
@@ -270,11 +270,11 @@ a lot of them:
 
 ::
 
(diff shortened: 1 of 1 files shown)


skyace65:
Note for other maintainers. As this is a 4.8 only change hold off on merging it for now.

engine_details/development/compiling/optimizing_for_size.rst
```diff
@@ -270,11 +270,11 @@ a lot of them:
...
 
 If this proves not to work for your use case, you should review the list of
 modules and see which ones you actually still need for your game (e.g. you might
-want to keep networking-related modules, regex support,
+want to keep networking-related modules, ``tilemap`` for 2D levels/maps,
```
Mickeon, line 277:
```suggestion
want to keep networking-related modules, tilemaps for 2D levels,
```
Although, I would put something more on the same "league" as RegEx would be. I'm not sure. E.g.:
```suggestion
want to keep networking-related modules, HDR support,
```

Synzorasize: Sure, I guess since the sentence doesn't need to change that much, although I personally think tilemaps need to be mentioned, lol. I'll wait to see if any other module shift occurs before doing anything, though, to see if there's something more appropriate that comes up.

Calinou: "HDR support" might be understood as [HDR output support](https://godotengine.org/article/hdr-output-arrives-in-godot-4-7/), but what `module_hdr_enabled=no` actually does is disabling import of Radiance HDR images (`.hdr` extension), typically used for panorama skies.

Synzorasize: I guess that makes sense, since I thought it was weird that it was totally supported in core and yet HDR was a module. That would make the module more niche, I guess. So is the tilemap explanation alright then?

Mickeon: Yes it is alright.

Calinou (approved):
Looks good to me.

skyace65 (approved):

skyace65:
Thanks!

PR #12216, opened 2026-08-01
[4.8] Add note to unit testing page about enabling static typing warnings for GDScript tests

Meorge:
Since https://github.com/godotengine/godot/pull/121839 , the `UNTYPED_DECLARATION` and `INFERRED_DECLARATION` warnings can be enabled for an individual GDScript test by including a comment in the test file.

This PR documents the behavior, but also notes that most GDScript tests should _not_ be made statically typed, and that the only ones that use these warnings should be ones specifically testing their functionality.

### DIFF
--- a/engine_details/architecture/unit_testing.rst
+++ b/engine_details/architecture/unit_testing.rst
@@ -451,3 +451,12 @@ If no errors are printed and everything goes well, you're done!
     function present inside the script file,
     you can disable the runtime section of the test by naming the script file so that it matches the pattern ``*.notest.gd``.
     For example, "test_empty_file.notest.gd".
+
+.. note::
+
+    By default, GDScript tests suppress the ``UNTYPED_DECLARATION`` and ``INFERRED_DECLARATION``
+    warnings. Add the comments ``# enable UNTYPED_DECLARATION`` or ``# enable INFERRED_DECLARATION``
+    to the GDScript test file to enable the corresponding warning for that test.
+
+    In general, GDScript tests should not use statically typed variables.
+    Only enable static typing warnings if you are specifically testing them.
--- a/engine_details/architecture/unit_testing.rst
+++ b/engine_details/architecture/unit_testing.rst
@@ -458,5 +458,5 @@ If no errors are printed and everything goes well, you're done!
     warnings. Add the comments ``# enable UNTYPED_DECLARATION`` or ``# enable INFERRED_DECLARATION``
     to the GDScript test file to enable the corresponding warning for that test.
 
-    In general, GDScript tests should not use statically typed variables.
+    In general, GDScript tests should not use statically-typed variables.
     Only enable static typing warnings if you are specifically testing them.


skyace65:
Note for other maintainers. As this is a 4.8 only change hold off on merging it for now.

engine_details/architecture/unit_testing.rst
```diff
@@ -451,3 +451,12 @@ If no errors are printed and everything goes well, you're done!
...
     you can disable the runtime section of the test by naming the script file so that it matches the pattern ``*.notest.gd``.
     For example, "test_empty_file.notest.gd".
+
+.. note::
+
+    By default, GDScript tests suppress the ``UNTYPED_DECLARATION`` and ``INFERRED_DECLARATION``
+    warnings. Add the comments ``# enable UNTYPED_DECLARATION`` or ``# enable INFERRED_DECLARATION``
+    to the GDScript test file to enable the corresponding warning for that test.
+
+    In general, GDScript tests should not use statically typed variables.
```
Mickeon:
```suggestion
    In general, GDScript tests should not use statically-typed variables.
```

Calinou (commented):
Looks good to me.

Is it possible to enable both warnings in the same file? I assume this is done by putting the lines one after the other, but this should be documented as well.

Meorge:
Thanks for the review! Indeed, enabling each warning is done individually with its corresponding comment (see implementation [here](https://github.com/godotengine/godot/pull/121839/changes#diff-c3f85bdfa5af8c2a08bc71eb6e81b563175f0749e7b7feed77f8bb6291a051ac)). Looking back at the wording for this PR, I can definitely see how it sounds like only one of the comments will work at a time; I'll see what I can do to rephrase it better.

engine_details/architecture/unit_testing.rst
```diff
@@ -451,3 +451,12 @@ If no errors are printed and everything goes well, you're done!
     function present inside the script file,
     you can disable the runtime section of the test by naming the script file so that it matches the pattern ``*.notest.gd``.
     For example, "test_empty_file.notest.gd".
+
+.. note::
+
+    By default, GDScript tests suppress the ``UNTYPED_DECLARATION`` and ``INFERRED_DECLARATION``
+    warnings. Add the comments ``# enable UNTYPED_DECLARATION`` or ``# enable INFERRED_DECLARATION``
+    to the GDScript test file to enable the corresponding warning for that test.
```
Meorge, line 459: Minimal changes, probably my preferred one:
```suggestion
    By default, GDScript tests suppress the ``UNTYPED_DECLARATION`` and ``INFERRED_DECLARATION``
    warnings. Add the comments ``# enable UNTYPED_DECLARATION`` and/or ``# enable INFERRED_DECLARATION``
    to the GDScript test file to enable the corresponding warning(s) for that test.
```

More changes and possibly a bit more confusing, but suggesting just in case:
```suggestion
    By default, GDScript tests suppress the ``UNTYPED_DECLARATION`` and ``INFERRED_DECLARATION``
    warnings. The comment ``# enable UNTYPED_DECLARATION`` in a GDScript
    test file will enable the ``UNTYPED_DECLARATION`` warning for that test, and 
    similarly the comment ``# enable INFERRED_DECLARATION`` will enable the
    ``INFERRED_DECLARATION`` warning.
```

Shorter, but does not take into account the fact that currently these two specific warnings are hardcoded to be enabled when the specific string is present. (So if we add more warnings in the future that are suppressed by default, they won't work unless we add cases for them too.)
```suggestion
    By default, GDScript tests suppress the ``UNTYPED_DECLARATION`` and ``INFERRED_DECLARATION``
    warnings. To explicitly enable a warning for a test, add the comment ``# enable [WARNING_NAME]``
    to the GDScript test file.
```

Meorge (commented):
A few potential re-phrasings of the note that make it more clear the warning-enabling comments are independent of each other.

PR #12218, opened 2026-08-01
Replace most leftover "instance" verbs with "instantiate"

Mickeon:
In Godot `4.0`, in a moment of lucidity, PackedScene's `instance()` function was renamed to `instantiate()` and the surrounding documentation followed suite. Despite this, there's some leftover text, or even new text, that didn't quite get the memo.

This PR replaces every occurrence of "instance" being used as a verb with "instantiate".
Similarly, "instancing" has been replaced with "instantiation" where appropriate.
File names are not changed in this PR.

Keep **very strongly** in mind the following:
- **Instantiation** is the whole creation process.
- **Instancing** is a real term in graphics rendering. It refers to the grouping of similar objects to render them all in one swoop. It's not the same concept as instantiation.
	- Similarly, an object is said to be **instanced** when it is, or can be, part of this batch, but this meaning very rare in the docs.

### DIFF
--- a/getting_started/first_2d_game/04.creating_the_enemy.rst
+++ b/getting_started/first_2d_game/04.creating_the_enemy.rst
@@ -7,7 +7,7 @@ Now it's time to make the enemies our player will have to dodge. Their behavior
 will not be very complex: mobs will spawn randomly at the edges of the screen,
 choose a random direction, and move in a straight line.
 
-We'll create a ``Mob`` scene, which we can then *instance* to create any number
+We'll create a ``Mob`` scene, which we can then *instantiate* to create any number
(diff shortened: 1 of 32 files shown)


AThousandShips (approved):
LGTM!

Mickeon:
Addendum, if this PR is accepted, it would be ideal to rename the [Instancing Demo](https://godotengine.org/asset-library/asset/2716) as well.

skyace65 (approved):

skyace65:
Needs a rebase

mhilbrunner (approved):

skyace65:
Thanks!

PR #12221, opened 2026-08-01
Document translation template generation

### LINKED ISSUE #12219: Expand the tutorial about translation template generation
**Your Godot version:**

4.8 dev

**Issue description:**

The POT generator was recently expanded and, now called "Template generator", allows creating CSV templates too. However the only mention of the generator is still in the gettext tutorial.

It likely deserves its own page, and can be expanded with extra information, like best practices or tricks how to prepare scenes/script for extracting.

Things that could be added:
- List of some example fields that get extracted from scenes (like any `text` property etc.)
  - I wouldn't make full list though, or at least clearly state that it's not exhaustive, to avoid needing to update it.
- Node's `auto_translate_mode` affects which strings are extracted.
- Control's new `translation_context` property is included in the template.
- This is a trick, but if you want to mark an arbitrary string in script for translation, without translating it, you can use `set_message_translation()` to disable translation of an object and then call `tr()` on that object, to obtain an extractable, but not translated string.
- The EditorTranslationParserPlugin's new `_customize_strings()` method allows to add/remove/modify extracted strings, which can be useful for some advanced operations.
  - One example is when you have a placeholder string in a scene that you don't want to translate, but the node with the string should not have translation disabled. You can prefix such strings with a fixed prefix, to easily filter it out. This is yet another trick I guess.
- Template generator can be invoked from Command Palette, which is useful when you are actively working on updating the template.

https://docs.godotengine.org/en/latest/tutorials/i18n/localization_using_gettext.html#automatic-generation-using-the-editor

ThomasConway01:
Adds a dedicated guide for the translation template generator and links it from the gettext workflow.

It covers configuring PO and CSV output, Command Palette access, the kinds of strings that are extracted, Auto Translate Mode, translation contexts, and `EditorTranslationParserPlugin._customize_strings()` for custom processing.

Closes #12219.

Validation:

- `codespell` passed on the three edited files.
- `git diff --check` passed.
- A complete local Sphinx build was started after installing the documented dependencies, but did not complete within the available command time. CI can run the full documentation build.

### DIFF
--- a/tutorials/i18n/index.rst
+++ b/tutorials/i18n/index.rst
@@ -8,6 +8,7 @@ Internationalization
    :name: toc-learn-features-i18n
 
    internationalizing_games
+   translation_template_generation
    localization_using_spreadsheets
    localization_using_gettext
    locales
--- a/tutorials/i18n/localization_using_gettext.rst
+++ b/tutorials/i18n/localization_using_gettext.rst
@@ -74,6 +74,9 @@ specified scene and GDScript files. This POT generation also supports translatio
 contexts and pluralization if used in a script, with the optional second
 argument of ``tr()`` and the ``tr_n()`` method.
 
+For configuration, extraction behavior, and customizing generated strings, see
+:ref:`generating translation templates <doc_translation_template_generation>`.
+
 Open :menu:`Project > Project Settings > Localization > Template Generation`, then use the
 :button:`Add…` button to specify the path to your project's scenes and scripts that
 contain localizable strings:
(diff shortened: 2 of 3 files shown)


ThomasConway01:
Understood, and I appreciate the heads-up regarding the policy on AI contributions. I'm a real person and submitted this PR to help document the translation template generation workflow, but I apologize if the phrasing or structure came across poorly.

I'm happy to completely rewrite and streamline the content to match Godot's style guide and standards, or close the PR if you'd prefer. Let me know how you'd like to proceed.

PR #12222, opened 2026-08-02
Rewrite "Using navigation layers" page

Mickeon:
TL:DR: This particular page bothers me.

This PR currently exists as a draft just to showcase my intention.

### DIFF
--- a/tutorials/navigation/navigation_using_navigationlayers.rst
+++ b/tutorials/navigation/navigation_using_navigationlayers.rst
@@ -1,197 +1,152 @@
 .. _doc_navigation_advanced_using_navigationlayers:
 
-Using NavigationLayers
-======================
+Using navigation layers
+=======================
 
-NavigationLayers are an optional feature to further control which navigation meshes are considered in a path query.
-They work similar to how physics layers control collision between collision objects or how visual layers control what is rendered to the Viewport.
+Navigation layers are an optional feature to further specify which navigation
+meshes are considered in a path query.
+They work similarly to how physics layers specify collisions between collision
+objects, or how visual layers specify what is rendered to a viewport.
 
-NavigationLayers can be named in the **ProjectSettings** the same as physics layers or visual layers.
+Navigation layers can be named in the **Project Settings** in the same way
+as physics layers or visual layers.
 
 .. image:: img/navigationlayers_naming.png
 
-If a region has not a single compatible navigation layer with the ``navigation_layers`` parameter of a path query this regions navigation mesh will be skipped in pathfinding.
-See :ref:`doc_navigation_using_navigationpaths` for more information on querying the NavigationServer for paths.
+If a region does not have any compatible navigation layer with the ``navigation_layers``
+parameter of a path query, the region's navigation mesh will be skipped in the pathfinding.
+See :ref:`doc_navigation_using_navigationpaths` for more information on querying
+the NavigationServer for paths.
 
-NavigationLayers are a single ``int`` value that is used as a **bitmask**.
-Many navigation related nodes have ``set_navigation_layer_value()`` and
-``get_navigation_layer_value()`` functions to set and get a layer number directly
+Navigation layers are contained in a single ``int`` value that is used as a **bitmask**.
+Many navigation-related nodes have ``set_navigation_layer_value()`` and
+``get_navigation_layer_value()`` functions to access a layer by number directly,
 without the need for more complex bitwise operations.
 
-In scripts the following helper functions can be used to work with the ``navigation_layers`` bitmask.
+.. tabs::
+ .. code-tab:: gdscript GDScript
+
+    @onready var region = get_node("MyNavigationRegion")
+    @onready var agent = get_node("MyNavigationAgent")
+
+    func _ready():
+        # Enables layer 4 for this region.
+        region.set_navigation_layer_value(4, true)
+        # Disables layer 1 for this region.
+        region.set_navigation_layer_value(1, false)
+
+        # Make future path queries of this agent ignore regions with layer 4.
+        agent.set_navigation_layer_value(4, false)
+
+    .. TODO: C# example.
 
 .. tabs::
- .. code-tab:: gdscript 2D GDScript
+ .. code-tab:: gdscript GDScript (2D)
 
-    func change_layers():
-        var region: NavigationRegion2D = get_node("NavigationRegion2D")
-        # enables 4-th layer for this region
-        region.navigation_layers = enable_bitmask_inx(region.navigation_layers, 4)
-        # disables 1-rst layer for this region
-        region.navigation_layers = disable_bitmask_inx(region.navigation_layers, 1)
+    var map: RID
+    var start_position: Vector2
+    var target_position: Vector2
 
-        var agent: NavigationAgent2D = get_node("NavigationAgent2D")
-        # make future path queries of this agent ignore regions with 4-th layer
-        agent.navigation_layers = disable_bitmask_inx(agent.navigation_layers, 4)
-
-        var path_query_navigation_layers: int = 0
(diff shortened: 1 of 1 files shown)


tutorials/navigation/navigation_using_navigationlayers.rst
```diff
@@ -1,197 +1,151 @@
...
-        path_query_navigation_layers = enable_bitmask_inx(path_query_navigation_layers, 2)
-        # get a path that only considers 2-nd layer regions
-        var path: PackedVector3Array = NavigationServer3D.map_get_path(
-            map,
-            start_position,
-            target_position,
-            true,
-            path_query_navigation_layers
-            )
+    static func is_bitmask_idx_enabled(bitmask: int, _index: int) -> bool:
```
AThousandShips:
```suggestion
    static func is_bitmask_idx_enabled(bitmask: int, index: int) -> bool:
```
It's not unused, same below

tutorials/navigation/navigation_using_navigationlayers.rst
```diff
@@ -1,197 +1,151 @@
...
- .. code-tab:: gdscript 3D GDScript
-
-    func change_layers():
-        var region: NavigationRegion3D = get_node("NavigationRegion3D")
-        # enables 4-th layer for this region
-        region.navigation_layers = enable_bitmask_inx(region.navigation_layers, 4)
-        # disables 1-rst layer for this region
-        region.navigation_layers = disable_bitmask_inx(region.navigation_layers, 1)
+Inside a script, the following helper functions can make it easier to work
+with the  ``navigation_layers`` bitmask:
```
AThousandShips:
```suggestion
with the ``navigation_layers`` bitmask:
```

tutorials/navigation/navigation_using_navigationlayers.rst
```diff
@@ -1,197 +1,151 @@
...
-        return _bitmask & ~(1 << _index)
+    func get_path():
+        # Get a path that only considers layer 2 regions.
+        var path: PackedVector3Array = NavigationServer3D.map_get_path(
+                map,
+                start_position,
+                target_position,
+                true,
+                0b10
+        )
```
AThousandShips, line 92:
```suggestion
            )
```

tutorials/navigation/navigation_using_navigationlayers.rst
```diff
@@ -1,197 +1,151 @@
...
         var path: PackedVector2Array = NavigationServer2D.map_get_path(
-            map,
-            start_position,
-            target_position,
+                map,
+                start_position,
+                target_position,
+                true,
+                0b10
+        )
```
AThousandShips, line 58:
```suggestion
            )
```

tutorials/navigation/navigation_using_navigationlayers.rst
```diff
@@ -1,197 +1,151 @@
...
-``get_navigation_layer_value()`` functions to set and get a layer number directly
+Navigation layers are contained in a single ``int`` value that is used as a **bitmask**.
+Many navigation-related nodes have ``set_navigation_layer_value()`` and
+``get_navigation_layer_value()`` functions to access a layer by number directly,
 without the need for more complex bitwise operations.
 
-In scripts the following helper functions can be used to work with the ``navigation_layers`` bitmask.
+.. tabs::
+ .. code-tab:: gdscript GDScript
+    @onready var region = get_node("MyNavigationRegion")
```
AThousandShips, line 29:
```suggestion

    @onready var region = get_node("MyNavigationRegion")
```

AThousandShips (commented):
Just a pass on pure format, will do a proper pass when it is ready otherwise

Mickeon:
Yeah, I would like to first assess, perhaps with the help of @\smix8, if there's anything worth mentioning because this page is in an awkward spot. The whole thing about bitmask helpers is particularly out-of-place, but there's also a lot of minor details that I find odd.

tutorials/navigation/navigation_using_navigationlayers.rst
Mickeon: I don't know if this is right. It wouldn't match `var path`'s indentation

AThousandShips: No it should be one indent for the closing parentheses, see [here](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html#indentation), so either one indent or after the last argument

Mickeon: I don't see anything about "one indent for the closing parentheses", but to my dismay there is something about "after the last argument"

smix8:
The reason that page has all the manual bitmask examples is because the entire Godot documentation has/had none how to really use them in such context. So telling users to use an `int` as a bitmask with zero example or helpers around the entire engine was cursed documentation which is why it was added to that page here.

The most important part to mention about using navigation_layers is that they do not affect the geometry of a navigation map, they only affect what gets included or excluded within a path query.

The reason this is so important to mention and drive home is because users still think they can use those layers to stack all kinds of navmesh geometry all over each other. All that such stacking does is create bugs and performance problems, it never worked like that and never will. The point of navigation_layers is to filter things out on a specific query WITHOUT changing the entire navigation map geometry as that would be very performance heavy.

Mickeon:
> So telling users to use an `int` as a bitmask with zero example or helpers around the entire engine was cursed documentation which is why it was added to that page here.

This is exactly why one should, instead, make a page about bitmasks, even a bit of a shoddy one, that contains this information in a way that can be shared by multiple other pages. Putting it all here is a hacky solution and very hard one to find.
It doesn't help that the bit operations have the corresponding methods, which makes the example a bit moot because most users will be satisfied with this basic purpose.

> The most important part to mention about using navigation_layers is that they do not affect the geometry of a navigation map, they only affect what gets included or excluded within a path query.

Since it's pertinent to path queries specifically, wouldn't all of this be better explained in "Using NavigationPathQuery objects" instead of having a whole separate page on navigation layers?

Billy-The-Squid:
Just my two cents, but this is just one symptom of the entire navigation tutorial needing reorganization. I've got some additional ideas of how to do that in https://github.com/godotengine/godot-docs/issues/12358

tutorials/navigation/navigation_using_navigationlayers.rst
```diff
@@ -1,197 +1,152 @@
...
 
-NavigationLayers can be named in the **ProjectSettings** the same as physics layers or visual layers.
+Navigation layers can be named in the **Project Settings** in the same way
+as physics layers or visual layers.
 
 .. image:: img/navigationlayers_naming.png
 
-If a region has not a single compatible navigation layer with the ``navigation_layers`` parameter of a path query this regions navigation mesh will be skipped in pathfinding.
-See :ref:`doc_navigation_using_navigationpaths` for more information on querying the NavigationServer for paths.
+If a region does not have any compatible navigation layer with the ``navigation_layers``
```
Billy-The-Squid, line 16:
```suggestion
If a region does not have any navigation layers that are compatible with the ``navigation_layers``
```
The wording still feels awkward

tutorials/navigation/navigation_using_navigationlayers.rst
```diff
@@ -1,197 +1,152 @@
...
+enabling/disabling entire navigation regions. Compared to region changes, a
 navigation path query with different navigation layers does not
-trigger large scale updates on the NavigationServer.
+trigger large-scale updates on the NavigationServer.
 
-Changing the navigation layers of NavigationAgent nodes will have an immediate
+Changing the navigation layers of a navigation agent will have an immediate
 effect on the next path query. Changing the navigation layers of
-regions will have an effect after the next NavigationServer sync.
+a region will have an effect after the next NavigationServer sync.
```
Billy-The-Squid, line 152: An example here might be helpful? Something to show the context/scenario in which this could be useful a little more concretely

PR #12226, opened 2026-08-03
Unlisted gdscript_basics setter exceptions

ISamAtlas:
add more details to the exceptions of when a setter function for a variable isn't called as they can work differently than expected for dictionary variables.

see the following as evidence
[issue 85578](https://github.com/godotengine/godot/issues/85578)
[pull request 62462](github.com/godotengine/godot/pull/62462)

(my first time making a pull request, hope that was sufficient/alright)

### DIFF
--- a/tutorials/scripting/gdscript/gdscript_basics.rst
+++ b/tutorials/scripting/gdscript/gdscript_basics.rst
@@ -2580,6 +2580,26 @@ This also applies to the alternative syntax:
         func set_my_prop(value):
             my_prop = value # Infinite recursion, since `set_my_prop()` is not the setter.
 
+Setters are also not called when editing or instantiating a key in a dictionary variable
+
+::
+
+        var my_dict: Dictionary:
+            set(value):
+                my_dict = value
+
+        my_dict["key"] = "value" # this will not call the setter
+
+Only if you supplant the variable with another dictionary will the setter be called
+
+::
+
+        var my_dict: Dictionary:
+            set(value):
+                my_dict = value
+
+        my_dict = {} # this replaces the dictionary and calls the setter
+
 .. _doc_gdscript_tool_mode:
 
 Tool mode
--- a/tutorials/scripting/gdscript/gdscript_basics.rst
+++ b/tutorials/scripting/gdscript/gdscript_basics.rst
@@ -2590,7 +2590,7 @@ Setters are also not called when editing or instantiating a key in a dictionary 
 
         my_dict["key"] = "value" # This will not call the setter.
 
-Only if you supplant the variable with another dictionary will the setter be called
+Only overwriting the variable will call the setter.
 
 ::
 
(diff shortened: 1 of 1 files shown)


tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -2580,6 +2580,26 @@ This also applies to the alternative syntax:
...
 
+Setters are also not called when editing or instantiating a key in a dictionary variable
+
+::
+
+        var my_dict: Dictionary:
+            set(value):
+                my_dict = value
+
+        my_dict["key"] = "value" # this will not call the setter
```
AThousandShips:
```suggestion
        my_dict["key"] = "value" # This will not call the setter.
```
Per the comment style, same below

tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -2580,6 +2580,26 @@ This also applies to the alternative syntax:
...
+
+Only if you supplant the variable with another dictionary will the setter be called
```
AThousandShips: "supplant" is a bit unusual here I'd say

tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -2580,6 +2580,26 @@ This also applies to the alternative syntax:
         func set_my_prop(value):
             my_prop = value # Infinite recursion, since `set_my_prop()` is not the setter.
 
+Setters are also not called when editing or instantiating a key in a dictionary variable
```
AThousandShips:
```suggestion
Setters are also not called when editing or instantiating a key in a dictionary variable:
```
Same below

But not sure this is the correct place for this example

tutorials/scripting/gdscript/gdscript_basics.rst
ISamAtlas: okay, how's overwrite?

"Only overwriting the variable will call the setter."

tutorials/scripting/gdscript/gdscript_basics.rst
```diff
@@ -2580,6 +2580,26 @@ This also applies to the alternative syntax:
...
+
+::
+
+        var my_dict: Dictionary:
+            set(value):
+                my_dict = value
+
+        my_dict["key"] = "value" # This will not call the setter.
+
+Only overwriting the variable will call the setter.
```
AThousandShips, line 2593:
```suggestion
Only overwriting the variable will call the setter:
```
I'd say

Mickeon:
If this is worth mentioning, so would be:
- adding elements to an array;
- removing entries from a Dictionary;
- changing properties of an Object assigned to the variable;
- and similar.

I don't think listing each is a great idea at all.

There's some nuance between "setting"/"overwriting" and "changing" that, [assuming the original issue is fair](https://github.com/godotengine/godot/issues/85578), is worth making clearer. And I don't believe it should be done by adding a new overly-specific example.

ISamAtlas:
okay, that makes sense, i see what you mean

PR #12228, opened 2026-08-03
Properly link to code regions

przemek-git:
Docs mention code regions, but wrongly link to built-in types instead.

### DIFF
--- a/tutorials/editor/script_editor.rst
+++ b/tutorials/editor/script_editor.rst
@@ -270,7 +270,7 @@ The **Edit** menu provides several options for line operations:
   - **Fold All Lines**: Folds all code blocks or code regions in the open document.
   - **Unfold All Lines**: Unfolds all code blocks and code regions in the open document.
   - **Create Code Region***: Wraps the selected text in a foldable code region to improve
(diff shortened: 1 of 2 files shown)


Mickeon (changes requested):
This change is incorrect. There's no matching [label](https://www.sphinx-doc.org/en/master/usage/referencing.html#ref-role) in the corresponding page, so this reference would fail to work entirely:

The original likely points at `_doc_gdscript_builtin_types` because it was deemed the closest, without needing to change the target page (which is still weird in my opinion)

https://github.com/godotengine/godot-docs/blob/4ee2cbcb7bb7b486cb524766ef65c6c7a4fcc890/tutorials/scripting/gdscript/gdscript_basics.rst?plain=1#L684

For this to work there would have to be a matching `.. _doc_gdscript_code_regions:`

przemek-git:
My bad, added matching label

Calinou (approved):
Looks good to me.

Mickeon (approved):

skyace65 (approved):

skyace65:
Thanks!

PR #12233, opened 2026-08-03
Remove non-breaking and zero-width spaces

Mickeon:
This PR removes all occurrences of the Non-Breaking Space (`00A0`) and Zero-Width Space (`\u200B`) characters from the docs.
Whether this would be a bug, I am not sure. When translating these lines on Weblate, sometimes you would get a message saying "_Applied automatic fix: zero-width space_" because of these imperceptible characters, which means translations can't even make use of them.

You know, ideally we would have checks to forever prevent them, but for now this can do.
The commit will also most certainly look funny because it looks like it does absolutely nothing.


skyace65 (approved):

skyace65:
Thanks for nothing!

PR #12242, opened 2026-08-05
Update Linux paths to use XDG variables

egnrse:
This MR updates Linux paths to use XDG variables and adds an explaining section (about the XDG variables).


bruvzg (approved):

Calinou (commented):
This makes sense on the surface, but I don't remember if these are *defined* by default in a way that makes them expand when entered in a file manager's address bar. This may reduce usability if paths are no longer copy-pastable.

`%APPDATA%` is always defined on Windows, so it's not a problem there.

bruvzg:
> but I don't remember if these are defined by default in a way that makes them expand when entered in a file manager's address bar

It should be defined in desktop distros, but not every file manager expand environment variables in the address bar.

egnrse:
That is a good point. I just tested it in Dolphin (default kde file manager) on Debian Trixie on KDE, and it does not expand there.

We could leave the old paths and add a footnote or `*` that links to the new `XDG Base Directory Specification` section?

bruvzg:
Or list both, the actual logic in the engine is:
- if `XDG_DATA_HOME` is set use it
- else use `HOME` + `.local/share`

skyace65:
Closing due to the lack of a response/activity after bruvzg's last comment. egnrse please ping me if you want this opened again.

PR #12244, opened 2026-08-05
Fix node Type contradiction

glsooogls:
On this page it is said: Area2D is used for players AND enemies. But later in the tutorial, the page after the next one, enemies will be constructed with RigidBody2D nodes.


Mickeon:
> On this page it is said: Area2D is used for players AND enemies

This is not what it says. The note explicitly says **CharacterBody2D** would usually be the better choice for players and enemies. It doesn't say the tutorial will make use **CharacterBody2D** for enemies, and implies the opposite.
> However, this tutorial focuses on core concepts that apply to a wider range of nodes and use cases.

glsooogls:
Thanks for checking.

PR #12257, opened 2026-08-08
Update Intro to 3D page for Godot 4.7

skyace65:
Does what the title says

- All screenshots have been updated and gone through Squoosh compression. The first screenshot now actually shows a scene from the third person shooter demo, I felt it makes Godot's 3D capabilities look better than the random 3 node scene we used before.
- I've added UI roles where I think they work
- Removed the reference to the asset library, the third person shooter demo has not been uploaded to the new asset store yet
- Class reference links have been simplified where applicable. I left a link to a specific method as is because I don't know if it will include the `()` unless we manually set the text.


tutorials/3d/introduction_to_3d.rst
```diff
@@ -256,8 +255,8 @@ Use :kbd:`Keypad 5` to toggle between perspective and orthogonal view.
 Using Blender-style transform shortcuts
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Since Godot 4.2, you can enable Blender-style shortcuts for translating,
-rotating and scaling nodes. In Blender, these shortcuts are:
+You can enable Blender-style shortcuts for translating, rotating and scaling nodes.
```
Mickeon:
```suggestion
You can enable Blender-style shortcuts for translating, rotating, and scaling nodes.
```

tutorials/3d/introduction_to_3d.rst
```diff
@@ -380,8 +379,8 @@ Environments can also be overridden in the Camera.
 Preview environment and light
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-By default, any 3D scene that doesn't have a :ref:`WorldEnvironment <class_WorldEnvironment>`
-node, or a :ref:`DirectionalLight3D <class_DirectionalLight3D>`, will have
+By default, any 3D scene that doesn't have a :ref:`class_WorldEnvironment`
+node, or a :ref:`class_DirectionalLight3D`, will have
 a preview turned on for what it's missing to light the scene.
```
Mickeon, line 384: I believe "what" is the subject here. Also maybe this sounds better:
```suggestion
a preview of what is missing to light the scene.
```

Mickeon (commented):
> I left a link to a specific method as is because I don't know if it will include the `()` unless we manually set the text.

Indeed, we took the lazy way out and the `()` is written manually (or when generating the class reference). We really could use a custom role for these, dang it...

tutorials/3d/introduction_to_3d.rst
skyace65: There's something about that phrasing that sounds awkward, I'll try and think of a better way to write this.

Calinou: Maybe this:

> By default, any scene that doesn't have a WorldEnvironment or DirectionaLLight3D node will use an editor-only preview node to act in its place.

PR #12263, opened 2026-08-10
Add :classref: role for automatic class reference formatting

Cykyrios:
This adds a new `:classref:` role which automatically formats the cross-reference link's title depending on the nature of the target (for instance, linking to a method adds trailing parentheses). It also features a toggle character to include or exclude the class name from the title, see the image below.

The syntax for this role is exactly the same as the target part of `:ref:` links to the class reference, with the optional `-` character for class name display. For comparison, when linking to Node's `_ready()` method:
* `` :ref:`_ready() <class_Node_private_method__ready>` ``
* `` :classref:`-class_Node_private_method__ready` `` (including `-` to hide the leading `Node.`)

More examples in the following image:

classref

The regex includes all variations of theme properties (theme_color, theme_constant, theme_font_size, etc.), as multi-word types could otherwise result in wrong parsing of the identifier; single-word types like method or property are kept generic.

### AI disclosure

I used LLM help to get the extension started and test it, then fixed and cleaned up myself.

### DIFF
--- a/_extensions/classref_links.py
+++ b/_extensions/classref_links.py
@@ -1,0 +1,97 @@
+# -*- coding: utf-8 -*-
+"""
+    classref_links
+    ~~~~~~~~~~~~~~
+
+    Sphinx extension to format links to Godot's class reference,
+    including classes, methods, properties, etc.
+
+    :copyright: Copyright 2026 by The Godot Engine Community
+    :license: MIT.
+"""
+
+import re
+from dataclasses import dataclass
+
+from docutils import nodes
+
+from sphinx.addnodes import pending_xref
+from sphinx.application import Sphinx
+from sphinx.roles import XRefRole
+from sphinx.util.typing import ExtensionMetadata
+
+
+_NO_CLASS = "~"
+_THEME_ITEMS = r"(?:theme_(?:color|constant|font_size|font|icon|style))"
+_CLASS_DEFINITION = rf"class_(@?[A-Za-z0-9]+)(?:_(private_)?({_THEME_ITEMS}|[a-z]+)_([_a-zA-Z0-9]+))?"
+_ENUM = r"enum_(@?[A-Za-z0-9]+)_([A-Za-z0-9]+)"
+_CLASSREF_RE = re.compile(rf"^({_NO_CLASS})?(?:{_CLASS_DEFINITION}|{_ENUM})$")
+
+
+def parse_classref_label(label: str) -> str:
+    match = _CLASSREF_RE.fullmatch(label)
+    if match is None:
+        return label
+
+    no_class = match.group(1) == _NO_CLASS
+    is_enum = match.group(6) != None
+    class_name = match.group(6) if is_enum else match.group(2)
+    private_method = False if is_enum else match.group(3) == "private"
+    definition_type = "enum" if is_enum else match.group(4)
+    identifier = match.group(7) if is_enum else match.group(5)
+
+    if identifier == None:
+        return class_name
+    if private_method:
+        identifier = f"_{identifier}"
+    title = identifier
+    if not no_class:
+        title = f"{class_name}.{identifier}"
+    if definition_type == "method":
+        title += "()"
+    return title
+
+
+class ClassRefRole(XRefRole):
+    def __init__(self) -> None:
+        super().__init__(
+            lowercase=True,
+            innernodeclass=nodes.inline,
+            warn_dangling=True,
+        )
+
+    def process_link(
+        self,
+        env,
+        refnode,
+        has_explicit_title: bool,
+        title: str,
+        target: str,
+    ) -> tuple[str, str]:
+        if target[0] == _NO_CLASS:
+            target = target[1:]
+        return parse_classref_label(title), target
+
+    def run(self) -> tuple[list[nodes.Node], list[nodes.system_message]]:
+        self.has_explicit_title = True
+        result_nodes, messages = super().run()
+        final_nodes: list[nodes.Node] = []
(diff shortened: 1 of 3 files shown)


skyace65:
Unfortunately I'm going to have to close this PR due to the AI usage. Our new AI policy we've talked about ([Blog post here](https://godotengine.org/article/contribution-policy-2026/)) is much more restrictive than the previous one, and applies to the documentation repository as well as the engine one. The contributing docs still show the old policy for now, but that will be updated relatively soon in the future, in the meantime we have started enforcing the new policy. And I thank you for disclosing that you used AI.

Putting aside the AI usage this would be an interesting feature, I was talking about the fact that we have to manually set link text for methods to get the ``()``  in another PR the other day. Though this might require consideration from a doc size perspective as we'd be adding in more text for every class reference link.

Cykyrios:
To be honest, I used AI here as an alternative to documentation, as I find it rather lacking on the Sphinx extension side (do feel free to point me to relevant doc, other than [this very basic tutorial](https://www.sphinx-doc.org/en/master/development/tutorials/extending_syntax.html)). The actual difficult part for me in this was creating the equivalent of `:ref:` links from a custom role, and that's where the LLM pointed me to XRefRole. But again, I never found a proper documentation page for XRefRole, and it was only after some back and forth that I found I was missing `lowercase=True`, which, even when I search specifically for it, only returns a few results.

The rest of the code being a regex and some formatting functions, in the end, the ClassRefRole is  the only one that actually benefitted from AI use here (but it doesn't change the fact that it generated the entire extension before my fix and cleanup pass).

About your last point, I'm not sure I follow: this should result in *less* text per link, not more? The role itself uses 5 more characters, but you don't specify the title and don't need to enclose the label in `<>`, so only extraordinarily short names would be shorter using `:ref:` directly.

Cykyrios:
Thanks for reopening! I separated the theme items part of the regex for better readability, and also switched the `-` prefix to `~` as recommended by @Mickeon.

Cykyrios:
Reworked the extension to subclass XRefRole in a cleaner way.

Calinou:
Tested locally, this doesn't appear to work on my end (the link points to the correct location):

image

```rst
Printing the license text using the :ref:`print() <class_@GlobalScope_method_print>`
function may be enough on platforms where a global output log is readable.
This is the case on desktop platforms, Android and HTML5 (but not iOS).

Printing the license text using the :classref:`class_@GlobalScope_method_print`
function may be enough on platforms where a global output log is readable.
This is the case on desktop platforms, Android and HTML5 (but not iOS).
```

With a leading `-` in the above `:classref:` role, I get a warning and no link:

```
WARNING: undefined label: '-class_@globalscope_method_print' [ref.ref]
```

Also, would it be possible to wrap the link within an "inline code" appearance (similar to using double backticks)? Alternatively, it could use a fixed-width font, even without the background and smaller font size.

Cykyrios:
Right, I forgot about `@GlobalScope` and `@GDScript`, so I'll need to add an optional `@` to the regex. As for the `-` prefix, I mentioned above that I replaced it with `~`, but haven't updated the screenshots in the PR description to reflect that.

Styling the link as "inline code" should be doable by changing `innernodeclass=nodes.inline` in the class initialization to `innernodeclass=nodes.literal`, I will push those changes in a moment.

Cykyrios:
`innernodeclass=nodes.literal` didn't work as Sphinx overrides the node type, so I did actually wrap the result in a literal. Below is a comparison (from the GDScript reference page) of master, updated version with literal wrapping (latest push of this PR), and a variation with CSS removing the background:

comparison_before
comparison_after
comparison_nobg

Calinou:
> Below is a comparison (from the GDScript reference page) of master, updated version with literal wrapping (latest push of this PR), and a variation with CSS removing the background:

I like the backgroundless fixed-width one the most. It doesn't look too distracting but still helps distinguish code links from non-code links.

Cykyrios:
I added the CSS to remove the background, the result is now the last screenshot from my previous post, here's a comparison with the screenshot from the PR description:

classref

(pushed twice to fix a mistake I made while refactoring last time...)

Mickeon:
I have to admit, I do prefer the black background albeit it's distracting.
That's mostly because some docs intermittently use \`\`code\`\` (`code`) and refs ([`code`](12263)) in a somewhat inconsistent manner.

You can even see this [in the example above](https://github.com/godotengine/godot-docs/pull/12263#issuecomment-5406386027). `weakref()` is a link, but `free()` and `queue_free()` aren't.
One could argue this would not be an issue if the new `classref` role supported [suppressing the link with !](https://www.sphinx-doc.org/en/master/usage/referencing.html#xref-modifiers) like `ref`, but it'd be quite niche and I imagine the sheer length of the tag would discourage contributors from doing it.

By the way, for disclosure, the reason I have not approved this myself despite insisting upon this is because I wanted to explore alternative options personally. But by all means don't let that prevent the PR from being merged, because at worst it'd just be a find 'n' replace around in the future.

Cykyrios:
I do agree with the mixed ``` ``code`` ``` and `:ref:` making a bit of a mess of things at times, but I was thinking not having the background for `:classref:` makes the links stand out from code a bit more (but they also have the doc icon for that purpose).

One alternative I wanted to explore initially was formatting the resulting link with syntax highlighting, as it would instantly make the classes pop out because of the green color (as long as official classes are up to date in the lexer, which shouldn't be an issue for that vast majority of nodes). Whether that brings too much color is another issue, but I haven't quite been able to do that anyway, Sphinx seems to make things difficult with its `pending_xref` processing.

PR #12264, opened 2026-08-11
Update the effects menu image

Synzorasize:
This is an update from godotengine/godot#122195.

Out with the bland, in with the glam. :) But also, the menu was outdated by `4.8` because the base classes of some effects were removed from the list (like `Filter` and `EQ`) since the image was added (`4.6`). godotengine/godot#120271 and godotengine/godot#120277 specifically removed them. You can see the old version on this page for comparison: https://docs.godotengine.org/en/latest/tutorials/audio/audio_effects.html (Thanks to @skyace65 for adding the original images ofc!)

I used the same menu image posted in godotengine/godot#122195 and godotengine/godot-proposals#15336. Slight quality was lost because I had to convert it to `.webp` (it was `.png` originally) but it's perfectly readable, I believe.


skyace65 (approved):
Thanks for the image update!

Calinou (approved):
Looks good to me.

skyace65:
Thanks!

PR #12265, opened 2026-08-11
Fixed typo in the "changing scenes manually" documentation.

ThomasCoasters:
In the second method of changing scenes manually described in this documentation, there was a type with "will be become", instead of "will be" or "will become", Here I chose for the latter.


skyace65 (approved):

skyace65:

Issue #11248, opened 2025-08-31
Inline code hightlighting


Qubus0:
while working on #11217 i was curious if it was possible to add inline code highlighting like we have it for the mod loader docs, and it turns out that it is possible by adding this to the files

```rst
.. role:: gd(code)
   :language: gdscript

The following is the list of supported operators and their precedence. All binary operators are `left-associative <https://en.wikipedia.org/wiki/Operator_associativity>`_,
including the :gd:`**` operator. This means that :gd:`2 ** 2 ** 3` is equal to :gd:`(2 ** 2) ** 3`. Use parentheses to explicitly specify precedence you need, for
example :gd:`2 ** (2 ** 3)`. The ternary :gd:`if/else` operator is right-associative.
```

which does result in all the proper code highlights. 

|Before | After |
|---|---|
|<img alt="Image" src="https://github.com/user-attachments/assets/a1e965b3-9693-4a23-b55c-304776226ab4" />|<img alt="Image" src="https://github.com/user-attachments/assets/f3032119-b985-4b49-8dfe-940dda64d0a4" />|

note that `:language: gd` results in the old lexer being used, i think, so not everything is properly highlighted, but `:language: gdscript` works just fine.

now the question - do we want this? it's neat, but also quite some effort to pull through everywhere, especially when tables break. also new contributions should probably use this too, which adds a bit more contribution friction

regex replace with
```regex
``(.*?)``
:gd:`$1`
```
does work alright, but of course not everything is gdscript

Calinou:
This looks pretty nice! I've been wanting to have inline code highlighting for a while. It's certainly something I'd like to see on more websites in general.

The hard part will be to comb through the existing pages and ensure we're only using `:gd:` for inline code that is actually GDScript. Maybe we'll need a `:cs:` role for C# code in the long run too.

Qubus0:
yep we probably would need that. i think we could start with this on the main gdscript pages where we could get the biggest benefit with the smallest chance for error.? for /classes the code that converts from bbcode [code] to backticks would need to be adjusted of course.

Calinou:
>  for /classes the code that converts from bbcode [code] to backticks would need to be adjusted of course.

This code is in the main Godot repository: https://github.com/godotengine/godot/blob/master/doc/tools/make_rst.py

However, we don't have explicit tagging of inline GDScript code right now (everything uses `[code]`), so we'd need to do that first by combing through the class reference in the main Godot repository (and adding editor support for the tags).

[...]

PR #12271, opened 2026-08-12
Add BBCode syntax highlighting

Mickeon:
This PR does a few major things:
- Adds a new lexer for **BBCode**, which on its own makes it possible for code blocks to highlight BBCode in a way that suits the Godot documentation.
- Adds a new `bbcode` role, which can be used to highlight BBCode **inline** by marking it down ```:bbcode:`like this` ```.
- Applies both of the above to the **BBCode in RichTextLabel** page.
	- In the future a few more pages can benefit from this, but this is a good starting point.

Here's some previews on how it looks like (these screenshots may become outdated):
image
image
image

This PR does need some clean-up, but that will be done when we agree it's good to go.

### DIFF
--- a/_extensions/bbcode.py
+++ b/_extensions/bbcode.py
@@ -1,0 +1,55 @@
+
+from pygments.lexer import RegexLexer, bygroups
+from pygments.token import (
+    Name,
+    Comment,
+    String,
+    Operator,
+    Text,
+    Whitespace,
+)
+
+__all__ = ["BBCodeFormatter"]
+
+
+class BBCodeFormatter(RegexLexer):
+    name = "BBCode"
+    aliases = ["bbcode", "bb"]
+
+    STYLE_BRACKETS = Comment.Doc
+    STYLE_TAG_TYPE = Name.Builtin.Function
+    STYLE_EQUAL_SIGN = Operator
+    STYLE_OPTION = String.StringName
+    STYLE_VALUE = String
+
+    tokens = {
+        "root": [
+            (r"(\[)(\w*\b)", bygroups(STYLE_BRACKETS, STYLE_TAG_TYPE), "tag"),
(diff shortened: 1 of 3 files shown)


Calinou:
> - Adds a new `bbcode` role, which can be used to highlight BBCode **inline** by marking it down `:bbcode:`like this``.

See also https://github.com/godotengine/godot-docs/issues/11248, which could be done to implement this for GDScript.

Mickeon:
> See also #11248, which could be done to implement this for GDScript.

I was going to. I actually have another branch for this with a few ideas and fixes.

skyace65 (changes requested):
I built and tested this locally. Is there a way to manually change what color certain things use? The color here (white skin color?) is difficult to read in light mode.

Image

Mickeon:
> The color here (white skin color?) is difficult to read in light mode.

No new colors are being used to highlight BBCode. They're all taking from the existing, GDScript highlight colors defined in the CSS file.

https://github.com/godotengine/godot-docs/blob/da8ea6f72e3a2aa5dd5fd3386ee4a758ab831247/_static/css/custom.css#L1163-L1165

That's to say, that the fault is on the original colors. The StringName color `#ffc2a6` is not appropriate in light mode and no one noticed until now. That's worth fixing separately in my opinion.
I wonder what the editor does, instead? ([I checked, it should be `#cc8f73`](https://github.com/godotengine/godot-docs/pull/12271))

> Is there a way to manually change what color certain things use?

I can also separate the colors from the ones used for GDScript, but that would require making up a new palette that I am not sure is worth thinking about just yet, given that BBCode is not prevalent outside of specific pages.

Mickeon:
Updated the PR to remove commented code

skyace65 (approved):
Color issue fixed by #12351

skyace65:
Thanks!

PR #12272, opened 2026-08-12
Fix accidental usages of RST's default role

Mickeon:
Not sure how to label this one.

So, I've explained this [a while back](https://chat.godotengine.org/channel/documentation?msg=mzPLHQ8TfxrfcDQmK). The short of it is that, unlike the Markdown you may be accustomed to, the single backticks do not denote \`code\` or \`literal\`. Rather, they denote [interpreted text](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#interpreted-text), with its role inferred implicitly rather than explicitly[^1]. 

So what ends up happening is that, because no **default role** is defined ([_yet_](https://github.com/godotengine/godot-docs/pull/12273)), it falls back to emphasis (*italics*):

image

Oops. This is a _very_ easy mistake to make, but I won't divulge why here, [that's for later](https://github.com/godotengine/godot-docs/pull/12273).

-------- 

This PR fixes every occurrence where single backticks are mistakenly used in favour of whatever is more appropriate. Oftentimes it's double backticks for code, sometimes it's our own `:ui:` roles, sometimes it's neither.

[^1]: The role is the string surrounded by colons before the backtick (e.g. `:abbr:` in ``:abbr:`HDR (High Dynamic Range)``)

PR #12273, opened 2026-08-12
Make default role highlight GDScript code

### LINKED ISSUE #11248: Inline code hightlighting

Mickeon:
Happens to close #11248
Related to #12272

Surrounding code with single backticks the docs has been a common mistake for a few reasons:
- Single backticks in typical Markdown denote literal text or code (\`this\` turns to `this`). That's the standard basically everywhere else (including this very website). Not in reStructured text, though.
- It's genuinely hard to infer the difference between \`this\` and \`\`this\`\` at a glance, especially without proper syntax highlighting in your IDE.
	- https://github.com/godotengine/godot-docs/pull/12272 proves it.
- Design-wise, one would assume that, because multiple asterisks denote increasingly stronger levels of emphasis, that backticks would behave similarly. But they don't.

In fact, and I've explained this [a while back](https://chat.godotengine.org/channel/documentation?msg=mzPLHQ8TfxrfcDQmK), single backticks denote [interpreted text](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#interpreted-text), with its role falling back to the **default role**. However, we have not defined a **default role**, so what ends up happening is that \`this text\` appears in italics:

image

---------------------------

This PR not only makes the single backticks purposeful, but it adds a new `gdscript` role. This allows the docs to feature syntax highlighting **inline** for GDScript.

In a nutshell, `` `text` `` becomes a shorthand for `` :gdscript:`text` ``.

It would allow for the following, but as of writing the PR does not currently use it intentionally.

image 
image

### DIFF
--- a/_extensions/gdscript.py
+++ b/_extensions/gdscript.py
@@ -1576,12 +1576,26 @@ class GDScriptLexer(RegexLexer):
         ],
     }
 
+from sphinx.application import Sphinx
+from sphinx.util.docutils import SphinxRole
+from sphinx.util.typing import ExtensionMetadata
 
-def setup(sphinx):
+
+def setup(sphinx: Sphinx) -> ExtensionMetadata:
     from sphinx.highlighting import lexers
+    from sphinx.roles import code_role
+    from docutils import nodes
 
     sphinx.add_lexer("gdscript", GDScriptLexer)
     lexers["gdscript"] = GDScriptLexer()
+
+    class GDScriptRole(SphinxRole):
+        def run(self) -> tuple[list[nodes.Node], list[nodes.system_message]]:
+            return code_role(self.name, self.rawtext, self.text,
+                             self.lineno, self.inliner,
+                             {"language": "gdscript"}, self.content)
+
+    sphinx.add_role("gdscript", GDScriptRole())
 
     return {
         "parallel_read_safe": True,
--- a/conf.py
+++ b/conf.py
@@ -243,6 +243,11 @@ rst_prolog = """
 
 """
 
+# This is the Sphinx role used for text formatted `like this`.
+# Many docs mistakenly use single backticks, instead of double backticks.
+# Default GDScript is the least intrusive, and happens to look quite nice.
+default_role = "gdscript"
+
 # -- I18n settings --------------------------------------------------------
 
 # Godot localization is handled via https://github.com/godotengine/godot-docs-l10n


_extensions/gdscript.py
```diff
@@ -1576,13 +1576,24 @@ def inner_string_rules(ttype):
...
+def setup(sphinx: Sphinx) -> ExtensionMetadata:
     from sphinx.highlighting import lexers
 
     sphinx.add_lexer("gdscript", GDScriptLexer)
     lexers["gdscript"] = GDScriptLexer()
 
+    class GDScriptRole(SphinxRole):
+        # HACK: This role doesn't do anything on its own. It's a placeholder,
+        # so that no warning is sent for the `default_role` not existing.
+        # The actual gdscript role is defined in conf.py's `rst_prolog`.
```
Cykyrios: My understanding is that `rst_prolog` simply adds its contents to the beginning of every rST file, so the `default_role` technically isn't defined. So I think this placeholder is fine, as the alternative would be to create the role properly here and remove the entry in `rst_prolog`.

conf.py
```diff
@@ -241,8 +241,16 @@
 .. role:: ui
     :class: role-ui
 
+.. role:: gdscript(code)
+   :language: gdscript
```
Cykyrios:
```suggestion
    :language: gdscript
```
There's one entry above with 3 spaces, but the rest uses 4.

Cykyrios:
Built and tested locally, this works fine, code looks good to me. There may be one small caveat worth mentioning, which is that including `` ` `` in the code itself requires escaping, but there is little chance of that happening anyway.

image

All samples behave the same when removing `:gdscript:`, since that is the new default role.

Calinou (approved):
Tested locally, it works as expected. Code looks good to me.

Some examples of this being used on the Shading language page (even though inline code here isn't GDScript):

Image

Image

Image

Image

Image

`:csharp:` and `:gdshader:` roles could be beneficial to add in future PRs, but I think using GDScript as a fallback for unlabeled roles makes sense. It ensures translations don't get too long as it's the main use case for inline code, including generic "pseudocode" like standalone expressions.

PR #12223, opened 2026-08-02
Update codespell to include many more checks

Mickeon:
See also:
- https://github.com/godotengine/godot-docs/pull/12051
- https://github.com/godotengine/godot/pull/120906

This PR rewrites the configuration side of `codespell`:
- Update version to `v2.4.3`
- It now checks for more mistakes and converts British spelling to US.
- Move list of ignored words from a separate file to `pyproject.toml` itself.
	- This keeps the list more tightly integrated and allows us to add comments.
- Add more ignored words
- It is now possible to ignore whole regions of code with `codespell:ignore-begin` and `codespell:ignore-end`
	- Under normal circumstances, the documented [`codespell:ignore` or `codespell:ignore-next-line`](https://github.com/codespell-project/codespell#inline-ignore) keywords should be used because they make the intention more explicit.
	However, due of the way RST comments are parsed, this is not feasible.
	- The used multiline regex can be changed, if ever necessary.
	- Maybe someday we could even have a custom RST role for this.
	- Barely documented feature of `codespell`, I know.
- ~~Add a "duplicate" list of file paths, to be used when executing `codespell` locally.~~
	- ~~I tried [absolutely everything](https://github.com/godotengine/godot-docs/pull/12051#discussion_r3384599090) to make it work without this hack to no avail.~~
- Add a new `sub-surface->subsurface` rule to our dictionary.
	- This was for testing the waters but it doesn't hurt.

The affected documentation pages have been changed in this PR, of course.

Some of these changes are made for parity with the configuration file in the current `godot` repository, as well as https://github.com/godotengine/godot/pull/120906.
To be honest I made this PR pondering how [`codespell`](https://github.com/codespell-project/codespell) interacts with this repository and its CI.

PR #12277, opened 2026-08-13
Expand codespell coverage

Repiteo:
- Follow-up to #12223

Takes the foundation laid out by Micky in the above PR and further integrates it across additional files/directories. Slightly trims the excluded words, as a few of those cases only cropped up once or twice & could be suppressed in isolation

This expanded ruleset and coverage actually resulted in a proper bugfix, as `labeler.yml` would fail to properly apply a label for android tutorials, as it was looking for changes in the "platofrm" folder

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,26 +9,21 @@ quiet-level = 3
 dictionary = ["_tools/codespell-dict.txt", "-"]
 builtin = ["clear", "rare", "en-GB_to_en-US"]
 skip = [
-    "_build/*",
-    "_static/*",
+    "_static/css/*",
     "_styleguides/*",
     "classes/*",
-    ".github/*",
     "tutorials/i18n/locales.rst",
     "AUTHORS.md",
     "LICENSE.txt",
 ]
 ignore-words-list = [
     "dialogue",     # Common even in US spelling.
-    "doubleclick",
     "findn",        # Our case-insensitive `find` function.
-    "implementors",
     "inout",        # Shader attribute.
     "lod",          # "Level of Detail".
     "marshalled",   # Common spelling in IT.
     "marshalling",  # Common spelling in IT.
     "thirdparty",   # Prefer as one word.
     "uint",         # "Unsigned integer".
-    "writeable",    # Part of a Python error message.
 ]
 ignore-multiline-regex = "codespell:ignore-begin.*?codespell:ignore-end"
--- a/tutorials/editor/external_editor.rst
+++ b/tutorials/editor/external_editor.rst
@@ -99,7 +99,11 @@ We have official plugins for the following code editors:
 LSP/DAP support
 ---------------
 
+.. codespell:ignore-begin implementors
+
 Godot supports the `Language Server Protocol <https://microsoft.github.io/language-server-protocol/>`_ (**LSP**) for code completion and the `Debug Adapter Protocol <https://microsoft.github.io/debug-adapter-protocol/>`_ (**DAP**) for debugging. You can check the `LSP client list <https://microsoft.github.io/language-server-protocol/implementors/tools/>`_ and `DAP client list <https://microsoft.github.io/debug-adapter-protocol/implementors/tools/>`_ to find if your editor supports them. If this is the case, you should be able to take advantage of these features without the need of a custom plugin.
+
+.. codespell:ignore-end
 
 To use these protocols, a Godot instance must be running on your current project. You should then configure your editor to communicate to the running adapter ports in Godot, which by default are ``6005`` for **LSP**, and ``6006`` for **DAP**. You can change these ports and other settings in the **Editor Settings**, under the **Network > Language Server** and **Network > Debug Adapter** sections respectively.
 
--- a/tutorials/migrating/upgrading_to_godot_4.3.rst
+++ b/tutorials/migrating/upgrading_to_godot_4.3.rst
@@ -269,8 +269,8 @@ Animation
 .. note::
 
     ``AnimationNode`` has a reworked process for retrieving the semantic time info. This ensures that time-related
-    behavior works as expected, but changes the blending behavior. Implementors of the ``_process`` virtual method
-    should also note that this method is now deprecated and will be replaced by a new one in the future (`GH-87171`_).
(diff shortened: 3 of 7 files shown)


pyproject.toml
```diff
@@ -9,26 +9,19 @@ quiet-level = 3
 dictionary = ["_tools/codespell-dict.txt", "-"]
 builtin = ["clear", "rare", "en-GB_to_en-US"]
 skip = [
-    "_build/*",
-    "_static/*",
     "_styleguides/*",
-    "classes/*",
```
Mickeon, line 15: I had happen to spot errors as well in `classes/*` by removing it at some point. However 

I think it'd be wise to keep it in the exceptions and fix the errors on the main repository, instead. These pages are entirely generated from a script. Any fixes you make here will be overridden later, causing future commits to always fail.

tutorials/migrating/upgrading_to_godot_4.3.rst
```diff
@@ -269,7 +269,7 @@ Animation
 .. note::
 
     ``AnimationNode`` has a reworked process for retrieving the semantic time info. This ensures that time-related
-    behavior works as expected, but changes the blending behavior. Implementors of the ``_process`` virtual method
+    behavior works as expected, but changes the blending behavior. Implementers of the ``_process`` virtual method
     should also note that this method is now deprecated and will be replaced by a new one in the future (`GH-87171`_).
```
Mickeon: This sentence kinda sucks, actually. I wanted to do it before but I didn't know how.
```suggestion
    behavior works as expected, but changes the blending behavior. It's also worth noting that the ``_process`` virtual method
is now deprecated and will be replaced by a new one in the future (`GH-87171`_).
```

tutorials/editor/external_editor.rst
```diff
@@ -99,8 +99,12 @@ We have official plugins for the following code editors:
 LSP/DAP support
 ---------------
 
+.. codespell:ignore-begin implementors
```
Mickeon, line 102: You can try if you want, but I do believe `codespell` ignores links by detecting them with a regex. So this may not be necessary.

_static/css/custom.css
```diff
@@ -997,7 +997,7 @@ code,
 
 /* Default Pygments styles make all of these bold */
 .highlight .k /* Keyword */,
-.highlight .ges /* Generic.EmphStrong */,
+.highlight .ges /* Generic.EmphStrong */, /* codespell:ignore ges */
```
Mickeon: I don't think we need codespell to be checking CSS files because of how "strict" CSS is with specific wording. That is to say, if you make a typo you will see it and you'll know on the spot.

pyproject.toml
```diff
@@ -9,26 +9,19 @@ quiet-level = 3
 dictionary = ["_tools/codespell-dict.txt", "-"]
 builtin = ["clear", "rare", "en-GB_to_en-US"]
 skip = [
-    "_build/*",
```
Mickeon, line 12: There is no need to check on `_build/*` indeed, but purely because it's a waste of time?

pyproject.toml
Repiteo: Oh my god, it's just now occurring to me how much space is wasted by these files being versioned *at all*

I'll keep the folder excluded in that case and revert their contents

_static/css/custom.css
Repiteo: That's reasonable enough, especially when the current comments are already using appropriate spelling. I'll re-add the `_static` exclusion, but specifically target the `css` subfolder; `.js` files will still warrant checks

tutorials/editor/external_editor.rst
Repiteo: Bit of a weird one, but I figured it out. Codespell allows for words to be excluded *only* in uri-based contexts, but the parser is overzealous and actually tries to correct the excluded word *in the config*. Worked around by wrapping the config option *itself* in `codespell:ignore-begin/end`, which was a little silly but it worked so hey

pyproject.toml
Repiteo: It's already excluded in `prek` runs thanks to `.gitignore`, so it was redundant

Mickeon:
> Oh my god, it's just now occurring to me how much space is wasted by these files being versioned at all

In my opinion if we could find a way to remove `_classes` from the main repo without disastrous consequences it'd be a net positive, so do keep that on the back of your head if you may.

tutorials/editor/external_editor.rst
Mickeon: I noticed that too. To be honest, I would just add all of `conf.py` to the ignore list. It's not a particularly busy file.

Mickeon: To be clear, I would approve of the PR if this were done, unless you're vehemently against it.

tutorials/migrating/upgrading_to_godot_4.3.rst
```diff
@@ -269,7 +269,7 @@ Animation
 .. note::
 
     ``AnimationNode`` has a reworked process for retrieving the semantic time info. This ensures that time-related
-    behavior works as expected, but changes the blending behavior. Implementors of the ``_process`` virtual method
+    behavior works as expected, but changes the blending behavior. It's also worth noting that the ``_process`` virtual method
     should also note that this method is now deprecated and will be replaced by a new one in the future (`GH-87171`_).
```
skyace65:
```suggestion
    is now deprecated and will be replaced by a new one in the future (`GH-87171`_).
```

skyace65 (changes requested):

tutorials/editor/external_editor.rst
Repiteo: Opted to return the suppression to the file itself, as that was the only area in the repo that used it after other fixes/skips were added

We can consider skipping `pyproject.toml` altogether in the future if something else crops up, but I think it's fine for now. I assume you meant that file instead of `conf.py`, because the latter never caused any issues

Mickeon: Alright. Yeah, I did mean `pyproject.toml`, got my hands on too many cookie jars right now.

Mickeon (approved):

skyace65 (approved):

skyace65:
Thanks!

PR #12278, opened 2026-08-14
Document iOS 16 requirement for web exports

### LINKED ISSUE #12231: Add minimum iOS version to HTML export
**Your Godot version:**
4.7.1

**Issue description:**
I have exported a small game for the web using the HTML5 export in project settings. Nothing fancy mostly all standard. On Android and PC no problem to load the game, but on iOS 15 it stuck on the loading screen. To debug this and see what is going on I installed 'eruda' to see the console and there a strange error popped up:
```
CompileError: WebAssembly.Module doesn't parse at byte 88: invalid opcode 253, in function at index 231
```
and after a reload of the page and starting the game:
```
CompileError: WebAssembly.Module doesn't parse at byte 5: can't get Function local's type in group 1, in function at index 0
```
After some more digging and back-and-forth with an LLM is seems iOS 15 does not support SIMD which is used in the export. Only iOS version 16 and higher supports SIMD.

So IMO it would be a good idea to put a remark in the documentation to let others know that only iOS 16+ is supported.

Or is this a bug because Godot wants to support iOS 15?

https://docs.godotengine.org/en/stable/tutorials/export/exporting_for_web.html

GraceyDugar:
Fixes #12231

Added a note to the web export page explaining that iOS 16 or later is
required, since web export templates use WebAssembly SIMD which iOS 15
does not support. Projects fail to load with a WebAssembly CompileError
on older versions.

Placed alongside the existing WebAssembly/WebGL 2.0 requirement in the
introduction.

### DIFF
--- a/tutorials/export/exporting_for_web.rst
+++ b/tutorials/export/exporting_for_web.rst
@@ -14,6 +14,13 @@ This requires support for `WebAssembly
 <https://webassembly.org/>`__ and `WebGL 2.0 <https://www.khronos.org/webgl/>`__
 in the user's browser.
 
+.. note::
+
+    iOS 16 or later is required to run web exports on iOS devices. Web export
+    templates use WebAssembly SIMD, which is not supported on iOS 15 and
+    earlier. On those versions, the project will fail to load with a
+    WebAssembly ``CompileError``.
+
 .. attention::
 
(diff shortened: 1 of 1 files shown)


AThousandShips:
Please note that the issue for this has not reached consensus, do not merge before this has been decided as the appropriate solution

PR #12285, opened 2026-08-17
Add GDShader syntax highlighting

Mickeon:
Similar to https://github.com/godotengine/godot-docs/pull/12273 and https://github.com/godotengine/godot-docs/pull/12271

This PR adds a GDShader lexer, which on its own makes it possible for code blocks to highlight GDShader code in a manner that matches the Godot Editor more closely. 
It is based on the GLSL lexer. In fact, it inherits straight from it, and adds more rules specifically for the Godot Shading Language.

Here's how it looks like (_this may become a bit outdated_):

| Before| after |
| ---- | ---- |
| image | image |

------------------------------

Do note that this PR contains a lot of unused code and huge lists of hardcoded values. I would like to have further discussions on how to handle them (hardcoding vs. relying on heuristics)

### DIFF
--- a/_extensions/gdshader.py
+++ b/_extensions/gdshader.py
@@ -1,0 +1,244 @@
+
+from pygments.lexer import bygroups, words, inherit
+from pygments.token import (
+    Keyword,
+    Name,
+    Comment,
+    String,
+    Operator,
+    Whitespace,
+)
+from pygments.lexers.graphics import GLShaderLexer
+
+__all__ = ["GDShaderLexer"]
+
+KEYWORDS = (
+    "shader_type",
+    "render_mode",
+
+    # Uniform hints.
+    "source_color",
+    "hint_enum",
+    "hint_range",
+    "hint_normal",
+    "hint_default_white",
+    "hint_default_black",
+    "hint_default_transparent",
+    "hint_anisotropy",
+    "hint_roughness_r",
+    "hint_roughness_g",
+    "hint_roughness_b",
+    "hint_roughness_a",
+    "hint_roughness_normal",
+    "hint_roughness_gray",
+    "filter_nearest",
+    "filter_nearest_mipmap",
+    "filter_nearest_mipmap_anisotropic",
+    "filter_linear",
+    "filter_linear_mipmap",
+    "filter_linear_mipmap_anisotropic",
+    "repeat_enable",
+    "repeat_disable",
+    "hint_screen_texture",
+    "hint_depth_texture",
+    "hint_normal_roughness_texture",
+    "no_editor",
+    "no_storage",
+
+    # Data types.
+    "samplerExternalOES",
+
+    # Other.
+    "group_uniforms",
+    "global",
+    "instance",
+)
+KEYWORDS_CONTROL_FLOW = (
+    "break",
+    "case"
+    "continue",
+    "do",
+    "elif",
+    "else",
+    "for",
+    "if",
+    "return",
+    "switch",
+    "while",
+)
+
+BUILT_IN_FUNCTIONS = (
+    "radians",
+    "degrees",
+    "sin",
+    "cos",
+    "tan",
+    "asin",
+    "acos",
+    "atan",
+    "sinh",
+    "cosh",
+    "tanh",
+    "asinh",
+    "acosh",
+    "atanh",
+
+    "pow",
+    "exp",
+    "exp2",
+    "log",
+    "log2",
+    "sqrt",
+    "inversesqrt",
+    "abs",
+    "sign",
+    "floor",
+    "round",
+    "roundEven",
+    "trunc",
+    "ceil",
+    "fract",
+    "mod",
+    "modf",
+    "min",
+    "max",
+    "clamp",
+    "mix",
+    "fma",
+    "step",
+    "smoothstep",
+    "isnan",
+    "isinf",
+    "floatBitsToInt",
+    "floatBitsToUint",
+    "intBitsToFloat",
+    "uintBitsToFloat",
+
+    "length",
+    "distance",
+    "dot",
+    "cross",
+    "normalize",
+    "reflect",
+    "refract",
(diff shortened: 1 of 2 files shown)


_extensions/gdshader.py
```diff
@@ -0,0 +1,428 @@
...
+        "root": [
+            # (words(RENDER_MODES, prefix=r"\b", suffix=r"\b"), Name.Builtin),
+            # (words(STENCIL_MODES, prefix=r"\b", suffix=r"\b"), Name.Builtin),
+            # (words(BUILT_IN_SPATIAL_GLOBAL, prefix=r"\b", suffix=r"\b"), Name.Builtin),
+            # (words(BUILT_IN_SPATIAL_VERTEX, prefix=r"\b", suffix=r"\b"), Name.Builtin),
+            # (words(BUILT_IN_SPATIAL_FRAGMENT, prefix=r"\b", suffix=r"\b"), Name.Builtin),
+            # (words(BUILT_IN_SPATIAL_LIGHT, prefix=r"\b", suffix=r"\b"), Name.Builtin),
+            # Faster alternative, but less reliable bet.
+            # Anything uppercase is a built-in.
+            (r"\b[A-Z]+\b", Name.Builtin),
```
Cykyrios:
```suggestion
            (r"\b[A-Z_]+\b", Name.Builtin),
```

_extensions/gdshader.py
```diff
@@ -0,0 +1,428 @@
...
+            (r"render_mode\s+", Keyword, "builtin_list"),
+            (r"stencil_mode\s+", Keyword, "builtin_list"),
+            (words(KEYWORDS, prefix=r"\b", suffix=r"\b"), Keyword),
+            (words(BUILT_IN_FUNCTIONS, prefix=r"\b", suffix=r"\b"), Keyword),
+            # Anything after a dot is a component.
+            (r"(?<!\.)(\.)([a-zA-Z_]\w*)\b(?!\s*\()", bygroups(Operator, Name.Variable.Instance)),
+            (r"/(\\\n)?\*\*(.|\n)*?\*(\\\n)?/", Comment.Doc),
```
Cykyrios: I wasn't too sure about this one so I played with it on regex101, it seems to support only doc comments (2 asterisks when opening the comment). It doesn't look like the shader editor requires a leading `*` for each new line though, and this regex doesn't either. However single-line doc comments are not supported here.

It might be worth having single and multi-line comments in one regex and a separate one for doc comments, like in regular GDScript.

_extensions/gdshader.py
Cykyrios:
```suggestion
            (r"\b[A-Z_]+\b", Name.Builtin),
```
For things like `SCREEN_UV` or `MODEL_MATRIX`. But it is indeed less reliable than using the lists, as this allows leading or trailing underscores, in addition to typos.

Mickeon: I will test most built-in variables later. It just dawned on me it also won't catch `UV2`.

The question is always on whether to rely purely on guesswork, or manually writing out each constant (which heavily risks being outdated). The PR implements the former, although you can see I got prepared for the latter as huge lists, despite being unused.

Cykyrios: True, didn't think of `UV2`! I can understand the getting outdated part, but at least the lists should not change often enough to be an issue. I would put them on the same level as the GDScript lexer's built-in classes, although those might be easier to update as you can just get the list when generating the documentation.

Calinou:
This looks like a really good start :slightly_smiling_face: 

To make the appearance closer to the Godot shader editor, I would change the built-in color (such as `ALBEDO` in the example) to light blue instead of green.

Cykyrios:
Here's the example shader from the PR description, typed in the shader editor for comparison (please excuse the typo in the comments):

shader

I haven't checked the exact hex colors, but I'm only seeing 2 differences here:
* `return` should use the control flow keyword color (#ff8ccc)
* `render_mode` entries such as `cull_disabled`, and `ALBEDO`, `UV`, etc. should use the class name color (#c7ffed).

Other than that, it's looking very good!

Mickeon:
Updated the PR after a good chunk of manual testing.
- All control flow keywords are highlighted distinctly
- Render and stencil modes are highlighted the same way as the editor
- [Shader directives](https://docs.godotengine.org/en/stable/tutorials/shaders/shader_reference/shader_preprocessor.html) (`#if`, etc.) no longer use GLSL's whole line highlighting.
- Account for https://github.com/godotengine/godot/pull/122931
- Fixed missing `filter` uniform hints.

image



I also am aware of occasions where the highlighting will fail (comments between `render_mode` and the final `;`), but that is because Pygments's original GLSL highlighting is not very "modular". I'd have to override it in its entirely, which I currently find not worth it.
I'd be better to approach edge cases granularly, while replacing the code block highlighting, instead of trying to find them all right now.

I must note, GDShader's documentation comments are not specially highlighted in the Godot Editor, at all. I will keep that here nonetheless.

Lastly, I'm still undecided on the approach to highlight all built-in variables, as there are so many, and a naive approach that doesn't account for all of them leads to erroneous highlight.
For most snippets in the documentation, this only really happens inside preprocessor directives.

Calinou (approved):
Tested locally (by changing all `code-block:: glsl` to `code-block:: gdshader` in `shading_language.rst`), it works as expected. Code looks good to me.

Some feedback:

- `u` (unsigned), `f` (float), `e` (scientific notation) suffixes in numbers should be colored with the number color:

Image

This was also an issue in the current GLSL highlighter, so it can be improved in a future PR.

Mickeon:
Given the approval, I updated the PR to remove unused code, including the huge list of built-in variables.

> This was also an issue in the current GLSL highlighter, so it can be improved in a future PR.

It appears that Pygments is set up to catch the `e` in their regex but for some reason (possibly a possessive quantifier) it doesn't work.

mhilbrunner (approved):

PR #12286, opened 2026-08-18
Update macOS/iOS system requirements.

bruvzg:
- MoltenVK dropped macOS 11 support, so it is updated to list separate min. version for Compatibility and Forward+/Mobile.
- Changed recommended macOS version to 14, since it is the oldest version still getting updates.
- Min. supported iOS was changed to 15.0, but example SOC was still A7 (iPhone 5S, which does not support iOS 15), changed it to A9 (iPhone 6S).


skyace65 (approved):

mhilbrunner (approved):

mhilbrunner:
Thanks!

PR #12304, opened 2026-08-25
Update description about visible handlers in the getting started tutorial.

leandro-benedet-garcia:
Since https://github.com/godotengine/godot/pull/71092 the behavior of handlers changed so they are always visible even behind things. Now the visible orange dots are 6 most of the time.

But I removed the mention of how many dots appear because even before the change, the amount of visible handlers could be 5 if you looked at the cube on it's side. The same still happens.

(issue reported originally in rocket chat https://chat.godotengine.org/channel/documentation?msg=3tkQRgx2ySFGoDtGP)


skyace65 (approved):

skyace65:
Thanks!

PR #12075, opened 2026-06-19
First3dGame - Change randi_range to randf_range to match source, use explicit floats for min and max speed

samw-26:
Issue #9770
The docs used randi_range and integers for random_speed which does not match the source in https://github.com/godotengine/godot-demo-projects/tree/master/3d/squash_the_creeps and would cause an integer division bug when scaling the animation speed.

Additionally, min and max speed have been changed to floats for clarity. https://github.com/godotengine/godot-demo-projects/pull/1329

PR #12310, opened 2026-08-27
Fix integer division causing unintended rounding

brookmiles:
The line:

`$AnimationPlayer.speed_scale = random_speed / min_speed`

results in the warning `GDScript::reload: Integer division. Decimal part will be discarded.` and `speed_scale` will always evaluate to `1.0`.


skyace65 (commented):
I can't comment on if this PR is likely to be merged or not. But please open a corresponding PR in the demo project repository. https://github.com/godotengine/godot-demo-projects

brookmiles:
I see now that this has already been addressed by https://github.com/godotengine/godot-docs/pull/12075 in the "latest" docs, I didn't see the change as I was looking at "stable".

PR #12327, opened 2026-08-31
Update the setup documentation for the Android SDK

m4gr3d:
The setup documentation was not updated when https://github.com/godotengine/godot/pull/113761 was merged. This fixes that, and update the setup process based on the latest Android best practices.

PR #12328, opened 2026-08-31
Update the Android export documentation for Godot 4.8 and newer

m4gr3d:
Follow-up to https://github.com/godotengine/godot/pull/121849


tutorials/export/img/setup_android_build_screenshot.webp
syntaxerror247: Could you replace this screenshot with the one using Godot's menu bar? This one looks a bit odd because every other screenshot is using editor theme.

tutorials/export/exporting_for_android.rst
```diff
@@ -30,8 +30,25 @@ steps described in :ref:`doc_android_gradle_build`.
 
 You can then skip to :ref:`doc_exporting_for_android_providing_launcher_icons`.
 
-Setup on Windows, macOS, and Linux
-----------------------------------
+Godot 4.8 and newer: Setup on Windows, macOS, and Linux
+-------------------------------------------------------
+
+Godot 4.8 adds support for `automatically setting up the Android and Java SDKs <https://github.com/godotengine/godot/pull/121849>`__.
```
Calinou: I'm not sure we need to link the pull request, unless it has important (but secondary) information we can't put in the docs for conciseness' sake.

tutorials/export/exporting_for_android.rst
```diff
@@ -30,8 +30,25 @@ steps described in :ref:`doc_android_gradle_build`.
...
+
+The feature can be enabled from two locations:
+
+- From the Project menu, select **Setup Android Build...**
+
+.. image:: img/setup_android_build_screenshot.webp
+
+- Alternatively, from the **Export Presets** window a warning and clickable link will show when an Android preset is added and the Android and / or Java SDKs are not set up.
```
Calinou:
```suggestion
- Alternatively, from the **Export Presets** window a warning and clickable link will show when an Android preset is added and the Android and/or Java SDKs are not set up.
```

tutorials/export/exporting_for_android.rst
```diff
@@ -30,8 +30,25 @@ steps described in :ref:`doc_android_gradle_build`.
...
+
+.. image:: img/setup_android_build_from_export_presets_screenshot.webp
+
+Clicking on either will launch several dialogs that will prompt for, download, and install the Android and Java SDKs in a location specific to the Godot editor.
```
Calinou, line 49: When I tried the feature, it installed the Android SDK to `$HOME/Android` on Linux rather than a scoped editor location. I thought this was intended, but I would prefer it to be scoped to the editor's data folder (e.g. `android_sdk` and `jdk` in `EditorPaths::get_singleton()->get_data_dir()`).

tutorials/export/exporting_for_android.rst
```diff
@@ -30,8 +30,25 @@ steps described in :ref:`doc_android_gradle_build`.
...
-Setup on Windows, macOS, and Linux
-----------------------------------
+Godot 4.8 and newer: Setup on Windows, macOS, and Linux
+-------------------------------------------------------
+
+Godot 4.8 adds support for `automatically setting up the Android and Java SDKs <https://github.com/godotengine/godot/pull/121849>`__.
+
+The feature can be enabled from two locations:
+
+- From the Project menu, select **Setup Android Build...**
```
Calinou:
```suggestion
- From the Project menu, select :menu:`Setup Android Build...`.
```

tutorials/export/img/setup_android_build_screenshot.webp
syntaxerror247, line 1: Could you replace this screenshot with the one using Godot's menu bar? This one looks a bit odd because every other screenshots are using editor panels.

tutorials/export/exporting_for_android.rst
syntaxerror247, line 49:
```suggestion
Clicking on either will launch several dialogs that will prompt for, download, and install the Android and Java SDKs.
```
I'd also prefer if it uses editor specific folders. But current implementation is to install JDK in cache and Android SDK in `$HOME/Android/Sdk` (on linux it's the default location).

syntaxerror247 (commented):
I see that you have kept the previous setup instructions.
Instead of doing "Godot 4.8 and newer" and "Godot 4.7 and older", it might be better to say "In Editor method (Automatic)" and "Manual method"? You can use different title text but Idea is to avoid mentioning Godot version becuase doc pages are already versioned.

tutorials/export/exporting_for_android.rst
m4gr3d: That's correct based on the current logic in https://github.com/godotengine/godot/blob/20a52a8988b1170a53e0e04a7d71a0cb26f71656/editor/file_system/editor_paths.cpp#L87.

Although both @Calinou and you prefer using editor specific folders so maybe the implementation should be updated to match the documentation..

m4gr3d:
> I see that you have kept the previous setup instructions. Instead of doing "Godot 4.8 and newer" and "Godot 4.7 and older", it might be better to say "In Editor method (Automatic)" and "Manual method"? You can use different title text but Idea is to avoid mentioning Godot version becuase doc pages are already versioned.

People may be using older versions of the editor and accessing the `stable` version of the docs, so being clear about the version will prevent them from looking for functionality they don't have.

tutorials/export/exporting_for_android.rst
m4gr3d: @Calinou @syntaxerror247 I update the logic to keep the location specific to the Godot editor in https://github.com/godotengine/godot/pull/123079.

skyace65 (commented):
I'm going to strongly disagree with having both versions present in the master branch, the 4.7 section should be removed completely. It has always been the case that we only document how to do things for the current version. Keeping both versions makes the page needlessly long, and opens a can of worms where we'd have to start considering that for other features as well.

m4gr3d:
> I'm going to strongly disagree with having both versions present in the master branch, the 4.7 section should be removed completely. It has always been the case that we only document how to do things for the current version. Keeping both versions makes the page needlessly long, and opens a can of worms where we'd have to start considering that for other features as well.

Regardless of how we title the sections, we are still going to need both sections to provide instructions for folks that want to set up their dependencies manually.

The manual section is also used as setup reference for the 'Compiling for Android' documentation, so it has to stay regardless.

> Keeping both versions makes the page needlessly long

The changes add 17 lines and are built upon https://github.com/godotengine/godot-docs/pull/12327 which removes 18 lines.. so overall 1 line was removed from the current stable version of this page.

skyace65:
In that case please rename the section titles so they don't reference specific Godot versions.

m4gr3d:
> In that case please rename the section titles so they don't reference specific Godot versions.

I've updated the section titles in the PR.

m4gr3d:
@skyace65 @syntaxerror247 @Calinou ping on this PR.

syntaxerror247 (approved):

tutorials/export/exporting_for_android.rst
```diff
@@ -30,8 +30,28 @@ steps described in :ref:`doc_android_gradle_build`.
 
 You can then skip to :ref:`doc_exporting_for_android_providing_launcher_icons`.
 
-Setup on Windows, macOS, and Linux
-----------------------------------
+Automatic Build Setup on Windows, macOS, and Linux
+--------------------------------------------------
```
Calinou: Use sentence case for headings:

```suggestion
Automatic build setup on Windows, macOS, and Linux
--------------------------------------------------
```

tutorials/export/exporting_for_android.rst
```diff
@@ -30,8 +30,28 @@ steps described in :ref:`doc_android_gradle_build`.
...
+- Alternatively, from the **Export Presets** window a warning and clickable link will show when an Android preset is added and the Android and/or Java SDKs are not set up.
+
+.. image:: img/setup_android_build_from_export_presets_screenshot.webp
+
+Clicking on either will launch several dialogs that will prompt for, download, and install the Android and Java SDKs in a location specific to the Godot editor.
+
+.. _doc_exporting_for_android_manual_android_build_setup:
+
+Manual Build Setup on Windows, macOS, and Linux
+-----------------------------------------------
```
Calinou:
```suggestion
Manual build setup on Windows, macOS, and Linux
-----------------------------------------------
```

Calinou (approved):
Should be good to merge after applying suggestions.

mhilbrunner (approved):

mhilbrunner:
Thank you! Merged.

PR #12330, opened 2026-09-01
Fix typo in data_preferences.rst

Kasi-Asa:
The reference links to Object. I assume the reference text should only be 'Object' instead of 'Objects'.


AThousandShips:
This is incorrect, I'd say it should be inside the link for formatting reasons

Calinou:
Tested locally, this does not display correctly in Sphinx:

image

Sphinx does not support MediaWiki-style links where the plural form can be put outside the link and the link will still be created correctly, closing.

Mickeon:
> Sphinx does not support MediaWiki-style links where the plural form can be put outside the link and the link will still be created correctly

Just for disclosure, it does support this. The letter right after the backtick must be escaped (in this case `/s`). The problem is that this would still be a bad idea because 
image

PR #12351, opened 2026-09-05
Fix StringName highlight color in light mode

Mickeon:
Required for https://github.com/godotengine/godot-docs/pull/12271

Due to an oversight, the color used to highlight StringName literals in GDScript is the same between dark and light mode. This wasn't caught sooner because you can count the StringName literals in the docs with one hand.

image

This PR changes the color to match what the Godot Editor actually uses.


skyace65 (approved):

skyace65:
I checked the editor and confirmed the color used in light mode. Thanks!

PR #12361, opened 2026-09-07
Style: update 'Object' s to 'Objects' 

Kasi-Asa:
According to the previous discourse,#12330, should we tweak here?


Mickeon:
Yes, it should be changed here for style.
But please change the title and the commit's name to be more self-explanatory, instead of about the "discourse"

skyace65 (approved):

skyace65:
Thanks! I updated the commit message when merging.

PR #12378, opened 2026-09-13
Fix order of node renaming in spawning mobs guide

JordanAdair:


skyace65:
This change is incorrect. The PathFollow3D node IS the spawn location. And the Path3D node is the spawn path.

PR #12381, opened 2026-09-14
Fix solid background blocks behind series titles in the documentation changelog

### LINKED ISSUE #12352: Remove solid background blocks from documentation series titles in docs_changelog.html
**Your Godot version:**
N/A

**Issue description:**
In the Documentation changelog page, the text lines introducing manual document series ("Procedural geometry series:" and "Your First Shader Series:") are displayed in a solid background that mismatches with the rest of the page background (gray in light mode and black in dark mode).

While this is likely *by design*, it breaks the visual harmony and clean layout of the rest of the documentation theme.

https://docs.godotengine.org/en/latest/about/docs_changelog.html

**Attachments:**
Light mode:
Image

Dark mode:
Image

mateuseap:
Closes #12352

#### Changes made

- Rewrote the "Your First Shader Series:" and "Procedural geometry series:" entries in `about/docs_changelog.rst` as regular nested bullet lists. They were written with definition list syntax (a line ending in `:` immediately followed by an indented list), so Sphinx rendered them as `<dl><dt>` elements, which the theme styles with a solid background that mismatches the rest of the page.
- Kept the fix in the page markup rather than in `_static/css/custom.css`, since the `dt` styling is intentional and shared with the class reference and other pages that legitimately use definition lists.

Verified with docutils that the two entries now parse as nested bullet lists (the page goes from 4 `definition_list` nodes to 0) and that the file still parses without warnings.


about/docs_changelog.rst
```diff
@@ -437,10 +437,12 @@ Shading
 ~~~~~~~
 
 - Your First Shader Series:
```
skyace65:
```suggestion
Your First Shader Series:
```

about/docs_changelog.rst
```diff
@@ -463,7 +465,8 @@ Multi-threading
 Creating content
 ~~~~~~~~~~~~~~~~
 
-Procedural geometry series:
+- Procedural geometry series:
```
skyace65:
```suggestion
Procedural geometry series:
```

skyace65 (changes requested):
Bullet points should be removed for consistency with earlier entries on the page

skyace65 (approved):

skyace65:
Thanks!

PR #12387, opened 2026-09-17
Document the working directory requirement for running unit tests

### LINKED ISSUE #12366: Unit testing: document necessity of being in Godot's root folder
**Your Godot version:**

4.7 Stable (Applies to any Godot version).

**Issue description:**

In order for tests to be run correctly, the command launching Godot with the test option `--test` must be executed with the Godot source repository root as the working directory.

The documentation does instruct to execute `./bin/<godot_binary> --test` in the "Running tests" section. I feel like it's not clear that this is a hard requirement and not just a way to keep the docs lean by skipping CD'ing into `/bin` after executing `scons` from the root folder.

Further down, the "Integration tests for GDScript" does instruct explicitly to CD the root folder before doing further operations. Since this is in the context of writing new GDScript tests, I don't feel like it would be automatic to assume this is a requirement for running them as well. Furthermore, if a user has a problem generically running the tests, they're going to look for help in the "Running tests" section rather than the one about writing new tests specifically for GDScript.

Because of this, I think there should be a a small sentence or perhaps a WARNING or NOTE box (like the one already present for information about `dev_mode`) making it explicit that this is a hard requirement.

[Unit testing](https://docs.godotengine.org/en/stable/engine_details/architecture/unit_testing.html)

MeGaurav4:
The Running tests section showed ./bin/<godot_binary> --test without saying the command must run from the repository root. This adds a note making the requirement explicit, including the post-godot#123329 behavior (bin/ works via automatic directory change with a warning, anywhere else fails early).

One note block, nothing else.

Closes #12366

## Verification

- Read the merged engine PR to describe the current behavior accurately
- Matched the file's existing note directive style


AThousandShips:
Thank you for your contribution! Did you use AI tools when creating this PR or the description? Please see [this](https://contributing.godotengine.org/en/latest/development/contribution_rules.html#for-contributors)

MeGaurav4:
Thanks for asking, and for linking the policy.

Yes, I used AI assistance on this one: it helped draft the note wording, which I then reviewed before submitting. Since your policy asks contributors not to use AI for documentation text, I am withdrawing this rather than take up review time.

AThousandShips:
Thank you! And I hope we see you contributing in the future!

